PR #438, opened 2016-07-05
Implemented B86B exchange and B86BPBE exchange-correlation functionals.

aoterodelaroza:
## Description

## Todos
- [ ] Feature1

## Questions
- [ ]  Question1

## Status
- [ ]  Ready to go

The exchange functional B86B [A. D. Becke, J. Chem. Phys. 85:7184,
1986], and its combination with PBE correlation ("B86BPBE") are
implemented in the DFT part of psi4.


loriab:
Thanks much! Can you suggest a geometry/basis/grid and reference value from the literature for a test case?

dgasmith:
Looks good. Can you add a test in `/tests/dft1/` that compares the result against another program with a prior implementation?

loriab:
nice, @dgasmith . sync is 2 seconds off :-)

aoterodelaroza:
Hi all. Thanks for the quick response! :). I implemented b86bpbe in
nwchem back in the day, but psi4's results are a bit off (not too
much, a bit over 1e-3). I'll try to figure out what's going on when I
have some time (attending a conference right now), add b86bpbe to
the dft1 test, and update the pull request.

* Lori A. Burns <notifications@github.com> [2016-07-05 18:14:09 -0700]:

loriab:
By default, psi4 does density fitting, which would yield discrepancies in the right ballpark. Try `set scf_type pk` if you weren't already accounting for this.

dgasmith:
Sure, take your time.

aoterodelaroza:
OK, it works now. I reimplemented it in a different way and now it matches nwchem and the w=0.0 result of the range-separated functional built using the HJS recipe. Not sure what happened there. I have added a test in dft1, and the reference nwchem input and output.

loriab:
Thanks much. There's a bit of a hold-up just now as several large PR have to be merged in a certain order when ready. And the two people most likely to do this are headed to a conference. So this probably won't be addressed until next week.

aoterodelaroza:
No rush at all! :)

* Lori A. Burns <notifications@github.com> [2016-07-11 19:49:14 -0700]:

PR #437, opened 2016-07-05
Hessians!

andysim:
## Description
Adds analytic RHF Hessians.  Still much cleanup/fixing to be done, but I thought I'd open a PR to allow others to chime in.

## Todos
- [x] Add tests.
- [x] Hook into driver (with help from @loriab).
- [x] Add batching to allow fock derivatives to be computed without using too much memory.
- [x] Fix bug in spherical harmonic second derivative integrals.

## Questions
- [x]  Should we just merge this in when UHF is working, or should we also push to get the XC Hessian terms, @dgasmith / @robparrish ?  I don't know how involved the XC terms will be, but I'm happy to help get them cranked out.

## Status
- [x]  Ready to go

PR #286, opened 2016-02-27
Docspass secondary user-facing driver fns

loriab:
## Description
consolidate cube-writing code

## Todos
Notable points that this PR has either accomplished or will accomplish.
- [x] consolidate three cube-centered pieces of code into one, libcubeprop. history of this in discussions below.
- [x] spruce up some docs, particularly when options text has been copied from read_options into the written docs, which then gets out of sync, and who knows which is right. want to replace these with `.. include:: autodir_options_c/globals__relativistic.rst`` that can't get out of sync
- [x] put in parameter, return, examples docstrings, versionadded (for wfn) for the lesser-used user-facing fns: oeprop, molden, cubeprop and get these into the Psithon Functions chapter of docs, rather than hidden away in the Programmers Manual --> Python Driver section. This is to satisfy the soft fail request for oeprop/molden/cubeprop (http://forum.psicode.org/t/ideas-to-simplify-wave-function-passing/161/10?u=loriab for discussion).
- [x] put off until later real standardization of handling oeprop, molden, cubeprop python fns and associated documentation

## Status
- [x]  Ready to go

PR #447, opened 2016-07-22
Oeprop

andysim:
## OEProp bug fix
Some OEProp routines contracted Cartesian density matrices with spherical harmonic integrals, which is bad (and caused by me, I believe).  Thanks to Coleman Howard in @lothian's group for pointing this out.  For Cartesian basis sets, there was never a problem.  For spherical harmonics, the multipoles were bad only if the origin was chosen such that it broke symmetry; spherical harmonic fields and ESPs were all wrong.

I added a test for the multipoles, because they're available in Psi vars (reference values confirmed by Molpro).  I also confirmed the ESP and Fields on a small grid against Molpro values, but they're not available in Psi vars, so are harder to test.

## Status
- [x]  Ready to go


andysim:
This is ready to go, but I wrote the changes w.r.t. my master.  Will reopen when PR #437 has been merged in, because it's easier that way.

loriab:
If the changes in this PR are on top of the Hessians! PR, the history should clean up automatically as soon as Hessians! is merged. Many thanks for fixing this.

andysim:
@loriab thanks for the heads-up.  I didn't realize GH was that smart, but should have probably expected that.  Reopening now...

andysim:
Actually, I think I have a better idea for testing this stuff.  I can implement "ESP" and "FIELD" methods in OEProp that will compute the ESP and fields at the `PROPERTIES_ORIGIN` and set the results in Psi vars.  I don't think the grid iterators should populate those variables, because the grids could be huge.  By allowing `PROPERTIES_ORIGIN` to specify a single grid point, we can easily store the value for testing.  This will also allow us to construct loops over grids in Python, without having to generate a grid.dat file.  Thoughts, @loriab?

dgasmith:
Your next push will clean up this history. We are trying very hard to remove Psivars (and all globals) so id much rather have a setter on the object itself. 

Grid is a pretty small quantity all things considered, 10M grid points isn't all that much data these days. It would actually be nice if we could access the entire grid python side. We have some neat interactive visualization examples where this would be very interesting.

andysim:
@dgasmith, I didn't realize you were trying to deprecate the variables.  I can just populate V, Ex, Ey and Ez std::vectors (in OEProp) with the data, in the order that the grid points are specified, and bind the getters to python.  The problem is then making a clean interface.  The user is hidden from the oeprop objects themselves through the `property` wrappers.  Perhaps something analogous to the `return_wfn` keyword, to return the `oeprop` instance?

dgasmith:
Pretty much anything that is global should be on its way out. 

I haven't quite figured out how to rationalize two objects that are connected, but are separate entities yet. Currently I have just been appending the object to the Wavefunction python-side:

```python
def prop(wf, ...):
    oeprop = psi4.OEProp(wfn, ...)
    wfn.oeprop = oeprop
   return wfn
```

This works, but you get a vague attribute not found python message if the wavefunction does not have an oeprop object. We can go through and add `None` types for all of these attributes, but that doesn't really work long term.

loriab:
On the psivars, it's the global aspect that we'd like to squash by moving psivars into (perhaps module nested) dictionaries in the Wavefunction object. 

I agree that the ESP gets to be a lot of psivars and is rather more suited to a psiarray than multiple individual psivars. Perhaps atom centers are a special case of grid point that still gets psivars/array set. 

But then I also agree that as much as possible, the OEProp objects should be manipulatable py-side. 

There started to be an effort to standardize these lesser-used user-facing functions (http://forum.psicode.org/t/ideas-to-simplify-wave-function-passing/161/11?u=loriab and https://github.com/psi4/psi4/pull/286) and they degenerated into bug-fixing, so there's still no consensus. And there's the antagonism between API-like function arguments and QC-like long list of options to tweak.

This is a meandering way of saying that I don't presently have a plan for grid/prop interface. I think making the python access to the OEProp object full-functioned would be the most flexible approach.

andysim:
Used the builtin 'Update branch' button just now, thinking it'd rebase automagically, but that doesn't appear to be the case.   Anyways, I just added a first pass at what we were discussing above.  The reference values came from this Molpro input:
```
***,h2o properties
noorient
geometry = {
ang
    O ,,    0.250254404867,    0.126248114412,    0.000000000000
    H ,,    0.428893090449,    1.055731838795,    0.000000000000
    H ,,    1.104987458381,   -0.280303532167,    0.000000000000
}

basis=cc-pvdz
df-hf;
property
density,2100.2
pot,, -3.779451977, -3.779451977,  1.889725989
pot,, -3.779451977,  0.000000000,  1.889725989
pot,, -3.779451977,  3.779451977,  1.889725989
pot,,  0.000000000, -3.779451977,  1.889725989
pot,,  0.000000000,  0.000000000,  1.889725989
pot,,  0.000000000,  3.779451977,  1.889725989
pot,,  3.779451977, -3.779451977,  1.889725989
pot,,  3.779451977,  0.000000000,  1.889725989
pot,,  3.779451977,  3.779451977,  1.889725989
ef,, -3.779451977, -3.779451977,  1.889725989
ef,, -3.779451977,  0.000000000,  1.889725989
ef,, -3.779451977,  3.779451977,  1.889725989
ef,,  0.000000000, -3.779451977,  1.889725989
ef,,  0.000000000,  0.000000000,  1.889725989
ef,,  0.000000000,  3.779451977,  1.889725989
ef,,  3.779451977, -3.779451977,  1.889725989
ef,,  3.779451977,  0.000000000,  1.889725989
ef,,  3.779451977,  3.779451977,  1.889725989
```

andysim:
I just added a small section to the manual, advertising the new functionality.  A number of samples got updated during my local docs build.  With that in place, I think this is good to go.  Given that this is a bug fix, I think we should try to cherrypick it into the tagged release version, if that's not too much hassle.

Issue #322, opened 2016-03-19
scf5 fails intermittently


loriab:
I've seen this happen, though not on identical codebases. Added scf5 to quicktests so we could monitor with TravisCI. Then I saw it on Blueridge (https://testboard.org/cdash/testDetails.php?test=8074282&build=66614) to be mysteriously healed the next day (https://testboard.org/cdash/testDetails.php?test=7019761&build=66801). Then @bennybp reported it happens intermittantly on an identical codebase (https://github.com/psi4/psi4/issues/321#issuecomment-198631685). So now I believe it exists. Compilers (in my case) were Intel 2015.0.3.187.
- `Singlet PK CUHF energy: computed value (-149.6304503) does not match (-149.5905972) to 6 decimal places.`
- `Singlet PK UHF energy: computed value (-149.6484410) does not match (-149.5905972) to 6 decimal places`
- In my case, I think the guess occupancy spontaneously changed.

loriab:
@dgasmith, fyi that the scf5 problem has surfaced on your box with 2016 compilers. Sometimes fails on Singlet PK UHF, like above error. Sometimes passes. Below is diff of the key SCF with new (failing) on L, output.ref on R. Further below is diff of the key SCF with new (passing) on L, output.ref on R. Note the changing occupancies.

```
  ==> Iterations <==									  ==> Iterations <==

                        Total Energy        Delta E     RMS |[F,P]|		      |	                        Total Energy        Delta E     Density RMS

										      >	  Forming PK and K matrices.
										      >	   @UHF iter   1:  -130.87953989478825   -1.78522e+01   0.00000e+00 
    Occupation by irrep:									Occupation by irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 					         Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     3,    0,    0,    0,    0,    2,    1,    1 ]			      |		DOCC [     3,    0,    1,    0,    0,    2,    1,    1 ]
    SOCC [     0,    0,    1,    1,    0,    0,    0,    0 ]			      |		SOCC [     0,    0,    0,    0,    0,    0,    0,    0 ]
										      |	   @UHF iter   2:  -140.04253152363862   -9.16299e+00   2.20869e-01 DIIS
[...]

PR #454, opened 2016-08-02
added the molden_novirtual option to scf - discard unoccupied MOs

aoterodelaroza:
## Description
The MOLDEN file written by psi4, which can be used by our external programs, is HUGE when a reasonably-sized molecule is calculated using a large basis set. Since for many post-processing purposes the virtual orbitals are unnecessary, it would be interesting to have an option to drop writing most virtual MOs to the file, dramatically cutting down its size. This PR implements the molden_novirtual option to SCF. If true (it is false by default to preserve the original behavior), only the occupied orbitals are written to the molden file in a closed-shell molecule. In an open-shell system, 2*max(nocca,noccb) orbitals are written.

### DIFF
--- a/doc/sphinxman/source/molden.rst
+++ b/doc/sphinxman/source/molden.rst
@@ -28,7 +28,9 @@ also used to pass information between |PSIfour| and WebMO, if |PSIfour|
 computations are invoked using the WebMO GUI.  The filename of the 
 Molden file ends in ".molden", and the prefix is determined by 
 |globals__writer_file_label| (if set), or else by the name of the output
-file plus the name of the current molecule.
+file plus the name of the current molecule. If |scf__molden_novirtual|
+is set to true, the unoccupied orbitals are not written to the Molden
+file.
 
 .. autofunction:: driver.molden(wfn, filename)
 
@@ -37,4 +39,5 @@ Options
 
 .. include:: autodir_options_c/scf__molden_write.rst
 .. include:: autodir_options_c/globals__writer_file_label.rst
+.. include:: autodir_options_c/scf__molden_novirtual.rst
 
--- a/doc/sphinxman/source/molden.rst
+++ b/doc/sphinxman/source/molden.rst
@@ -28,8 +28,8 @@ also used to pass information between |PSIfour| and WebMO, if |PSIfour|
 computations are invoked using the WebMO GUI.  The filename of the 
 Molden file ends in ".molden", and the prefix is determined by 
 |globals__writer_file_label| (if set), or else by the name of the output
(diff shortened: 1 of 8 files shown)


aoterodelaroza:
Sorry about the failed tests. I'll take a look tomorrow.

doc/sphinxman/source/molden.rst
```diff
@@ -37,4 +39,5 @@ Options
 
 .. include:: autodir_options_c/scf__molden_write.rst
 .. include:: autodir_options_c/globals__writer_file_label.rst
+.. include:: autodir_options_c/scf__molden_novirtual.rst
```
loriab: very nice, thanks :-)

src/bin/psi4/read_options.cc
```diff
@@ -1210,6 +1210,8 @@ int read_options(const std::string &name, Options & options, bool suppress_print
     (if set), or else by the name of the output file plus the name of
     the current molecule. -*/
     options.add_bool("MOLDEN_WRITE", false);
+    /* Write all the MOs to the MOLDEN file (false) or discard the unoccupied MOs (true). */
+    options.add_bool("MOLDEN_NOVIRTUAL", false);
```
loriab: Make sure the read_options comments are surrounded by `/*-` and `-*/`. That extra dash is a signal to the autodocumentation script.

loriab: For booleans, I've found it easier to reason with if they convey positive instructions. Maybe switch the logic to `MOLDEN_WITH_VIRTUAL` and default `true`.

loriab:
No problem on the failing test case. In fact, due to build infrastructure changes in another branch, these PRs will probably be cherry-picked onto the other branch rather than merged directly.

The failing test case can be remedied by adding the extra arg to https://github.com/psi4/psi4/blob/master/share/python/driver.py#L1825 . Even if you set it up right in C++ with a default argument, Boost takes some convincing to allow that different signature automatically py-side. And since "Kill the Boost" is happening w/i a few weeks, no use working much with Boost on the alternate signature. (We're just replacing Boost with something else that will still allow exporting C++ functionality to Python)

aoterodelaroza:
Ah, I spent ten very confused minutes yesterday wondering about why the compiler didn't like my default argument :). I've just pushed the other changes suggested.

aoterodelaroza:
Meh, I'll make the change anyway. molden1 and molden2 tests are ok now. Thanks for the info! :)

loriab:
btw, don't worry about the Distelli fail. it looks like a case of #322, first we've seen on Mac.

loriab:
Replaced by #472 after rebase through CMakeRewrite-KillTheBoost

PR #456, opened 2016-08-05
CBS wrapper fix

loriab:
## Description
In the last `cbs()` pass, I refined the encoding of "an SCF is produced in the course of a CCSD" to "a HF is produced in the course of a CCSD". Unfortunately, I only refined it in a first pass that hit mp2, ccsd, ccsd(t), etc., not the lesser used fno-ccsd(t), mp3, cisd, etc. which can also be extrapolated. The alarming result is that if you used the latter, you'd get that the extrapolation of HF between 0 and –300 is –600.

## Todos
- [x] Now all methods that are usable in `cbs()` use the same psivar.

## Status
- [x]  Ready to go whenever but not so pressing as to disrupt the CMakeRewrite branch.

### DIFF
--- a/share/python/driver_cbs.py
+++ b/share/python/driver_cbs.py
@@ -351,83 +351,83 @@ def return_energy_components():
                              'hf': 'HF TOTAL ENERGY',
                             'mp2': 'MP2 TOTAL ENERGY'}
     VARH['mp2.5'] = {
-                            'scf': 'SCF TOTAL ENERGY',
+                             'hf': 'HF TOTAL ENERGY',
                             'mp2': 'MP2 TOTAL ENERGY',
                           'mp2.5': 'MP2.5 TOTAL ENERGY',
                             'mp3': 'MP3 TOTAL ENERGY'}
     VARH['mp3'] = {
-                            'scf': 'SCF TOTAL ENERGY',
+                             'hf': 'HF TOTAL ENERGY',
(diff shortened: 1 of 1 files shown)


loriab:
@dgasmith , maybe keep an eye out for test "casscf-fzc-sp". That's what failed on distelli (below), though it passed locally and on travis. May be another clue to the #322 intermittent failure.

```
       ==> Setting up CI strings <==
    
       There are 15 alpha and 15 beta strings
       The CI space requires 65 (6.50E+01) determinants and 4 blocks
    
    
       ==> Setting up MCSCF integrals <==
    
    
       ==> Starting MCSCF iterations <==
    
                            Total Energy        Delta E      RMS Grad   NCI
    
       @MCSCF Iter   1:  -76.02993400953443   -1.2637e-02   9.9525e-03   1   Initial CI
       @MCSCF Iter   2:  -76.06895156963408   -3.9018e-02   1.3537e-02   9   APPROX
       @MCSCF Iter   3:  -76.07302175257405   -4.0702e-03   4.4001e-03   7   APPROX
       @MCSCF Iter   4:  -76.07362055903870   -5.9881e-04   1.6406e-03   7   APPROX
       @MCSCF Iter   5:  -76.07373579006990   -1.1523e-04   7.9473e-04   7   APPROX, DIIS
```

PR #469, opened 2016-09-22
LGTM & CBS wrapper fix

loriab:
## Description
Configures LGTM properly. In the last cbs() pass, I refined the encoding of "an SCF is produced in the course of a CCSD" to "a HF is produced in the course of a CCSD". Unfortunately, I only refined it in a first pass that hit mp2, ccsd, ccsd(t), etc., not the lesser used fno-ccsd(t), mp3, cisd, etc. which can also be extrapolated. The alarming result is that if you used the latter, you'd get that the extrapolation of HF between 0 and –300 is –600.

## Todos
- [x] introduces a `.lgtm` file to define the approval regex and number of approvers.
- [x] Now all methods that are usable in cbs() use the same psivar.

## Status
- [x]  Ready to go

### DIFF
--- a/.lgtm
+++ b/.lgtm
@@ -1,0 +1,3 @@
+approvals = 3
+pattern = "(?i)LGTM"
+
--- a/psi4/share/psi4/python/driver_cbs.py
+++ b/psi4/share/psi4/python/driver_cbs.py
@@ -353,83 +353,83 @@ def return_energy_components():
                              'hf': 'HF TOTAL ENERGY',
                             'mp2': 'MP2 TOTAL ENERGY'}
     VARH['mp2.5'] = {
-                            'scf': 'SCF TOTAL ENERGY',
+                             'hf': 'HF TOTAL ENERGY',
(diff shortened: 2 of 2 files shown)


dgasmith:
Dont we need a maintainers file? Currently our LGTM's are not coming through.

andysim:
Do we still need to use the lgtm module, now that github [has reviews](https://help.github.com/articles/about-pull-request-reviews/)?

loriab:
You can have a `MAINTAINERS` file _or_ it defaults to the (secret) maintainers team. The latter is already set up and seems less controversial. We could ditch lgtm for GH reviews. The former can mandate geographic diversity of approvals (I set it to 3– Daniel and I just won't both do counted responses), whereas the latter I think only insists that if an owner puts a hold, then it must be resolved. Up to you all which you'd like to do.

andysim:
Ah, yes, you're right.  Your LGTM implementation LGTM.  As you were, @loriab!

PR #472, opened 2016-09-23
rebased: added option to discard unoccupied MOs in MOLDEN file

loriab:
## Description
The MOLDEN file written by psi4, which can be used by our external programs, is HUGE when a reasonably-sized molecule is calculated using a large basis set. Since for many post-processing purposes the virtual orbitals are unnecessary, it would be interesting to have an option to drop writing most virtual MOs to the file, dramatically cutting down its size. This PR implements the molden_novirtual option to SCF. If true (it is false by default to preserve the original behavior), only the occupied orbitals are written to the molden file in a closed-shell molecule. In an open-shell system, 2*max(nocca,noccb) orbitals are written.

## Questions
- [x]  @dgasmith, input file seems unresponsive to `set MOLDEN_WRITE true`. Instead have to keep Wfn and call `molden(Wfn)`. Does that sound right for the new scheme? Otherwise, can just have `molden()` called at end of `energy()` if kw is on. The PR's molden writing was in `src/bin/scf` which is now gone. Is the driver `molden()` the replacement location or did I miss it? Currently set up to be sensitive to `MOLDEN_WITH_VIRTUAL` but whatever's decided about `MOLDEN_WRTIE`, should probably go the same way.

## Status
- [x]  Ready to go

### DIFF
--- a/doc/sphinxman/source/molden.rst
+++ b/doc/sphinxman/source/molden.rst
@@ -28,7 +28,9 @@ also used to pass information between |PSIfour| and WebMO, if |PSIfour|
 computations are invoked using the WebMO GUI.  The filename of the 
 Molden file ends in ".molden", and the prefix is determined by 
 |globals__writer_file_label| (if set), or else by the name of the output
-file plus the name of the current molecule.
+file plus the name of the current molecule. If |scf__molden_novirtual|
+is set to true, the unoccupied orbitals are not written to the Molden
+file.
 
 .. autofunction:: driver.molden(wfn, filename)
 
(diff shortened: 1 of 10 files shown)


dgasmith:
@loriab It looks like I forget to reimplement in driver. My plan was to move it to energy/gradient/etc so that it would be universal. I need to add MOWriter to `proc/scf_helper` and probably `proc/mcscf` as well. Should I just add it to this PR?

loriab:
Sure, go ahead and add to this PR. Branch is in my acct now so you can push to it. 

Yeah, the distelli thing is weird. So fat, always fails first time, always passes second time.

loriab:
Molden writing sensitive to `MOLDEN_WRITE` again. I've just rebased, so let's do this one next. And we can see if lgtm ever recognizes @jturney .

LGTM

jturney:
LGTM

*crosses finger*

jturney:
Looks like this one worked.

dgasmith:
LGTM

PR #480, opened 2016-10-12
FISAPT dHF PsiVars

jgonthier:
## Description
Makes all components of the dHF computation in FISAPT available as Psi Variables.

## Todos

- [x] Export detailed components of FISAPT dHF as Psi variables

## Status
- [x]  Ready to go


loriab:
Good idea, now you mention it. I think I would go ahead and spell out `ENERGY`, though. Haven't been sparing characters in psivars thus far.

LGTM

jgonthier:
I understand, but my logic in that case was that the Psi vars would appear exactly like the associated quantity in the output file, just to decrease possible confusion arising form all these different quantities.
If you'd prefer to have ``ENERGY`` spelled out for homogeneity reasons in the Psi vars, just let me know and I'll make the change, though.

andysim:
LGTM

loriab:
I appreciate the `E` vs. `ENERGY` reasoning, but I think I still favor the latter. Text in the output file shows that it was written by a dozen coders, while the psivars have some unifying properties, including spelling out energy. We can just patch it later, if you'd rather not bother.

andysim:
How about defining both, to catch typos/confusion, but only documenting those with syntax in line with existing variables?

loriab:
Not so keen on extra psivars. Just detangling how `DFT TOTAL ENERGY`/`DFT FUNCTIONAL TOTAL ENERGY` or `SAPT EXCH-IND22 ENERGY`/`SAPT EXCH-IND20,R ENERGY` are different is enough without considering aliasing. And then we'd have to consider same for options. And they all have to fit through the [math](https://github.com/psi4/psi4/blob/master/psi4/share/psi4/python/qcdb/psivardefs.py). Not adamant, if everyone likes the idea, but not in favor.

jgonthier:
@loriab No problems, I just pushed the change. I'd rather have it done right now than forgetting to patch it later :D

loriab:
Thanks much, Jerome.

dgasmith:
Great, LGTM. Ill merge it in after tests pass.

Issue #468, opened 2016-09-19
Upcoming Psi4 changes


dgasmith:
Greeting friends,
As you may have heard we are in the process of several fundamental Psi4 change. These include: rewriting the build system, removing Boost from Psi4, the "inversion", and a history rewrite. Read below for summaries of each step and where to find more detailed information. If you have any questions please ask them here or email us directly.
## Steps

**Build system:**
The build system has moved to a "super build" strategy which reduces total compilation time and makes it much easier to add additional external libraries in the future. Developers will also see a reduction in the amount of time it takes to compile internal Psi4 changes and Psi4 can now link all external dependancies (libint for examples). Note that much of the github wiki on building psi4 is now plain wrong. Consult the top-level CMakeLists.txt to get building again. See [here](https://github.com/psi4/psi4/pull/442) for more information. 

**Boost:**
Much of the Boost functionality that we have depended on in Psi4 has been implemented in C++11. In order to keep Psi4 up to date with best C++11 practices, reduce complexity, and generally remove several headaches, Boost has been removed from Psi4. The C++ Python bindings have been moved to PyBind11. See [here](https://github.com/psi4/psi4/issues/450) for more information.

**Inversion:**
Currently the hybrid Python/C++ architecture of Psi4 is in an odd spot where Psi4 itself is a C++ program that calls an input file as a Python executable. This circular process will be removed and Psi4 itself will become a Python library capable of being imported just like any other module. This will allow Psi4 to be used in a normal python workflow with other scientific python programs like OpenMM, iPython, MolDesign, etc. In addition, python-based error messages will be more expressive and accurate. For users who enjoy the current Psithon interface the current Psi4 Psithon parsing equipment will be moved to a script named "psi4" in the binary location. This will allow all current Psi4 input files to be run normally.

**History rewrite:**
Over time the Psi4 git repository has accumulated many large files (mostly boost binaries) that are not currently used; however, they still exists within the repository and are downloaded every time the repository is cloned. This has caused Psi4 to balloon to approximately 650MB, while only about 65MB of that is currently used. In order to reduce the repository size, these files will be removed from history. When this occurs **_all_** current Psi4 clones **_and_** forks will become invalid. These will need to be deleted and reforked/recloned. 
## Schedule

[...]

PR #481, opened 2016-10-14
Psi4 Inversion

dgasmith:
## Description
Converts Psi4 from a binary that calls Python to a Python module that calls a shared object as discussed in #468. 

## Features
- "psi4" is now a python module, the entire driver and suite of C++ functions is now behind this interface.
- The "bin/psi4" binary is now a python script that parses Psithon input to support the current Psi4 interface.
- For any conventional python script the following now works: `import psi4`.
- C++ no longer calls python, all python objects must be passed in from the driver.
   - BasisSet objects are the primary objects effected here. The Wavefunction now stores BasisSets constructed by python and can be accessed by `wfn->get_basis(*basis*)`.
- The C++ side of psi4 is now compiled to the a shared library with python access named "core.so".

## Remaining issues
- There are several outstanding issues, a list of which can be found [here](https://github.com/dgasmith/psi4/projects/1). All issues may not be cleaned up by the time this PR goes through, current full suite coverage of 99%.
- HF: Broken Symmetry guess is currently broken, apparently no test case for this.
- BasisSet is not final with regard to construct_combined and handling of user defined basis sets.
- SAD does not work due to basis set creation.

## Status
- [x]  Ready to go


dgasmith:
Rebased through USAPT0 changes.

dgasmith:
Getting pretty close to merging this in. Failing tests case should be limited to the following:

```
	 22 - cc14 (Failed)
	155 - docs-psimod (Failed)
	187 - mints9 (Failed)
```

Docs and mints both require changes to the test case. `cc14` remains the only undiagnosed test failure.

andysim:
682 changed files!  GitHub struggled with this one.  I think the risk we introduce by delaying and making merges harder is far greater than the couple of failing tests, so I'm going to say LGTM.  Fantastic work to make this happen; sorry I couldn't help out.  Hopefully I'll be able to help the hunt for cc14's issue soon.

jturney:
LGTM

loriab:
``sapt7`` also in temporary broken list. LGTM

Issue #504, opened 2016-11-07
How to write a Molden file with natural orbitals and occupation numbers from a post-SCF calculation?


tovrstra:
My best guess (with the latest development version 6833bd5645022b232b82e4ffc84496e5be143506) would be as follows, for a CCSD calculation:

    molecule test {
     N   -0.0074552142    0.0447633077    0.0549133281
     H    0.7912317854    0.0113343077    0.0235803281
     H   -0.3132442140   -0.8795806918    0.2831263280
     H   -0.3744022140    0.2462893076   -1.0696916713
    units angstrom
    }
    set {
        basis cc-pVDZ
        freeze_core True
        guess sad
    }
    e, wfn = energy('CCSD', return_wfn=True)
    molden(wfn, 'ccsd_no.molden', density_a=wfn.Da())

However, the file `ccsd_no.molden` does not give reasonable charges when loaded into Molden:

                            Mulliken Charges
                            ================
     
                   N  -0.0793
                   H  -0.4048
                   H  -0.2581
                   H  -0.5537
     
    Sum of Mulliken Charges:  -1.2959

(There were no such problems with Molden files from plain SCF calculations.)

The sum of the occupation numbers in the Molden file is 6.0465. Shouldn't that sum be 10 in this example? Furthermore, with a too-low sum of occupation numbers, I would expect a positive sum of Mulliken charges, so presumably, there is something else going wrong too.

I'm not sure if `wfn.Da()` is the right way to get the density matrix? That was just a guess.

dgasmith:
`density_a` must be the MO density while all quantities on the Wavefunction itself are in the AO basis. I can see how this could be confusing and have edited several parts of the documentation.

In addition, most post-SCF methods do not build the density by default as this can be much more costly than the energy itself. Perhaps @amjames can give hints on extracting CC densities.

tovrstra:
Good to know. I may already be able to work with a density matrix in the AO basis, albeit not via the Molden file. Are the internal PSI4 conventions for ordering of basis functions (and sign conventions) the same as that in the Molden file?

dgasmith:
Unfortunately we do not match MOLDEN ordering. From a Psi to molden standpoint the following mapping is used:

``` C++
    // The order Molden expects
    //     P: x, y, z
    //    5D: D 0, D+1, D-1, D+2, D-2
    //    6D: xx, yy, zz, xy, xz, yz
    //
[...]

PR #506, opened 2016-11-07
ccdensity should communicate the OPDM back to ref_wfn

amjames:
## Description
Coupled Cluster methods in general do not compute the OPDM unless necessary, however when they do need to compute the OPDM they do not communicate that back to the wfn object. This has led to some confusion on the part of users trying to access CC density matrices. (see #504 ). 

~~The ccdensity module will now set `ref_wfn.arrays["CC_OPDM_A"/"CC_OPDM_B"] ` with the ground state opdm.~~

CCdensity does the MO/AO transformation and sets the ref_wfn Da/Db members internally, no python, and no longer using the `arrays` map. 
## Features
 - CC OPDM can be accessed via wfn object returned from a property calculation that constructs it.
 - eom root psivars for dipole/quadrupole components are now properly assigned.
## Todos
- [x] ~~Python-side code to convert arrays["CC_OPDM_A/B"] to ref_wfn.Da()/Db()~~
- [x] fix assignment of root psivars for cc routines 
- [x] add tests to check OPDM assignment in wfn through OEProps
## Questions
- [x] @dgasmith re the suggestion you made in #504 : would that not give us a ref_wfn with two copies of the same data? One held in `ref_wfn.arrays["CC_OPDM_A"]` and the second in `ref_wfn.Da` ? I think we would want the entry in the arrays table to be freed? 

## Status
- [x]  Ready to go


dgasmith:
@amjames Not overly worried about duplicate density matrices in a CC computation, especially considering SCF stores pretty much everything. If you are concerned you can grab the raw pointer `double** Dap = ref_wfn->Da()->pointer(h);` and set it at the C++ layer. Just remember that Da/Db are AO quantities.

amjames:
@dgasmith thats probably a better route to go, and by AO i take it you mean I should transform the opdm to the SO basis first?

dgasmith:
LGTM. Did you want to update OEProp Python side or wait for another update?

amjames:
@dgasmith To bring OEProp out to python should probably wait for another update. This one was to address problems related to #504 and I think that is satisfied.

The OEProp code is being called for (possibly) multiple states, and in order to move the OEProp stuff out to python would mean storing the OPDM for each state.  I would rather move the OEProp stuff out to python at a point where the loop over states can be dealt with in python as well. 

So yes, I am going to wait for another update to move the OEProp Python side.

jturney:
@amjames Can you add a test case to ensure the coupled cluster OPDM is used in properties?

amjames:
@jturney do you mean if you use OEprop from python side?

dgasmith:
It may be as simple as extending a test like:

```python
props, wfn = psi4.properties('CCSD/cc-pVDZ", return_wfn=True)
prop = psi4.core.OEProp(wfn)
...
prop.compute()

compare_value(…) # TEST
```

Just to ensure that the OPDM remains valid during changes. Good examples could be scf-properties or ci-properties test cases.

Cheers,
Daniel Smith

amjames:
While setting up a test I have noticed a couple of things. 

 - None of the `CCxx` tests are actually checking `DIPOLE [x|y|z]` against a reference value.
 - When doing an EOM calculation `psi4.core.ccdensity` is (internally c++ side) calling  `psi4::ccdensity::dipole()` once per state. Inside `psi4::ccdensity::dipole()` the OPDM is sent into OEProp with the title set to "CC". That means at the end of this calculation if you try to grab one of the dipole components via `psi4.get_variable("CC DIPOLE X")` you are really getting what you would expect should be listed as `psi4.get_variable("EOM-CC ROOT {last_root} DIPOLE X")`.  In addition to that you can't get the ground state dipole components at all since they were overwritten in the psivars dict with subsequent calls to the `psi4.core.ccdensity->psi::ccdensity::dipole->OEprop` chain.
 - see [psi::ccdensity::dipole](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/ccdensity/dipole.cc#L61) code and follow into [OEProp](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/oeprop.cc#L1184) to see what I mean if you didn't follow the explanation above. 

I think this very confusing when you look at how the dipole components for different roots can be grabbed with `psi4.get_variable` in the [ci-property tests](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/oeprop.cc#L1184) for example. 

I propose ( and volunteer to make) the following changes in order to implement the test here, and avoid confusion in the future:
 - For the 0th root (only one for non EOM-CC calculations) ccdensity will behave as is. 
 - For any root after that it will pass  `EOM-CC ROOT {root #}` into `oe->set_title()` [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/ccdensity/dipole.cc#L107) so that the confusing behavior with the variables dict is no longer as strange. 
 - To test the original modification that this PR introduced I will: 
   1. extend `tests/cc54` to return the wfn which should now have Da/Db set to the CCSD OPDM. Get the `CC DIPOLE/QUADRUPOLE ` components from the first pass. Run OEProp again (python side) with the wfn holding the CCSD OPDM, and make sure it gives the same `DIPOLE/QUADRUPOLE` components. 
   2. extend `tests/cc49` to do the same thing as before but just to be sure that the ground-state density is the only one being put back in ref_wfn, and at the same time confirm that the other roots are not over-writing the `CC DIPOLE/QUADRUPOLE` entries in the variables dict anymore. 

Objections?

loriab:
If you're willing to take on properly assigning root psivars in cc*, that'd be splendid, @amjames. It's very true that CC psivars are pretty much ground state only and that not much has been done in the way of CC property psivars since the testing has been through `output.ref` comparison rather than `compare_values()`.

One thing that will have to be resolved is indexing, as CI starts from 1. I can't even remember what position I took at the Last Great Debate on this, but I think I favor 0-indexed now, as it make the GS distinctive from ESs. Will query detci ppl tomorrow.

But, yes, if you see your way clear toward fixing this up, go for it.

amjames:
@loriab that is the only part Im not sure on. Feedback on indexing would be appreciated. I will open an issue for the lack of proper `compare_values`  checking in the cc property codes. We also don't have tests for UHF properties I can ask @lothian about references and reference values.

amjames:
Alright I have extended two tests none of the quicktests run through ccdensity but I ran both of these and the passed. The changes don't seem to mess up comparison to output.ref so I didn't change them.  

Pending feedback about how to index the roots I have followed what is in place for eom-cc ES energies. 

`CC DIPOLE X` is the ground state

`CC ROOT 1 DIPOLE X` is first ES, and so on.

andysim:
This is great; I really like that you've addressed the inconsistency between CI and CC, which I didn't notice before.  The scheme in your last message looks good to me.  I think that numbering the first excited state `1`, and the second `2`, *etc.* makes the most sense, for obvious reasons.  Zero basing the excited states would lead to an off-by-one error in the cardinal and ordinal notation.  LGTM.

CDSherrill:
I have not followed this thread closely, but I have a couple of comments.

1. Someone mentioned that the OPDM is supposed to be stored in wfn in the
AO basis.  AJ then asked if this meant he should backtransform it to the SO
basis.  I'm not sure that question got answered....and AO and SO are not
equivalent in molecules with point group symmetry.  Do I understand
correctly that OPDMs are stored all the way back in the AO (not SO) basis?

2. I looked through the CI and CC output with Lori yesterday.  Both outputs
are totally logical for what they are doing, yet they are inconsistent with
each other in terms of numbering roots.  For the case of CCEOM, because
excited states are all treated together, it is very logical to number the
EOM states 1, 2, 3, etc.  That then means the ground state, if it is in
this same numbering scheme, has to be labeled root 0.  For the CI outputs,
everything is lumped together, so it is logical to number the ground state
as 1.  If we make things consistent, it is easier to change the CI
numbering to match the CC numbering.  This also has the advantage that the
CC numbering will work well for CIS, TDDFT, etc.

So that we can easily loop over things, I favor any lists of energies or
dipoles as also including the ground state (state 0).  So, we can have

CI/CC ROOT 0 ENERGY [or whatever it's called, I forget]
CI/CC ROOT 1 ENERGY
etc.

as well as

CI/CC ROOT 0 DIPOLE x
CI/CC ROOT 1 DIPOLE x

Including ROOT 0 in the lists above will remove unnecessary annoying logic
when looping over roots for processing/printing output.  For CI use cases,
we will always want to loop over all roots including ROOT 0.

For the ground state only, we can also duplicate the ROOT 0 information
into variable names without the "ROOT n" designation, which is what people
will want to grab if they just want g.s. info.

CI/CC DIPOLE x

(Actually, strictly speaking, the root number that should go here should
probably be FOLLOW_ROOT, which defaults to ROOT1 (now should be ->
ROOT0)... FOLLOW_ROOT is supposed to allow you to optimize an excited state
instead of the ground state in an MCSCF or geometry opt).

The CI code has always assumed numbering starts with 1 in the user
interface, so any input options regarding roots will need to have -1 added
(except the total number of roots, obviously).

Best,
David

dgasmith:
@CDSherrill 
1) It should always be in the SO basis. Apologies this was answered elsewhere.

2) If we are set on this on I can patch up the CI code and add patch notes. The "CI DIPOLE X" is the followed root or the dipole from the state averaged density depending on user specifications.

We also need to think about the `psi4/driver/driver.py:molden` function and #504. `properties` is now guaranteed to return an updated SO basis density; however, the current molden function requires a MO density. We could have a simple switch on that function that indices if the incoming density is in the AO or MO basis and transform accordingly.

CDSherrill:
1) Cool.

2) I discussed this with @lothian, who seemed fine with it (it involves no
changes to his code :)  So, I think we're settled unless anyone else has
objections.

I agree there needs to be some thought about returning SO vs MO densities
--- different quantities will be needed in different circumstances.  Your
proposed solution seems reasonable.  Probably we should review all the
cases where densities are wanted.

amjames:
@dgasmith actually directly referring to #504 if you use the option ` write_nos = True` the natural orbitals are written to a molden file.

Is there anything else I should do here in particular ?

dgasmith:
@amjames That may be true if you go through the CC* modules to write NO's, but not for the Python function as #504 does. 

We could start hacking on these things; however, this is all a bit of mess, and hacking away at it without some serious thought to the overall structure is probably not advisable.

amjames:
Right. I meant since it appeared that CCSD NO's were wanted in particular. 

I think there is defiantly more improvements to make here, just not sure they fall under the scope of this PR.  If there is anything else I should add let me know. If not I made sure that the GS info appears in psivars as `CC ROOT 0` as well so we can loop over states, and I rebased the branch.

dgasmith:
You could add a `density_basis` option to `def molden`. If the basis is SO you can convert the matrix to MO by:

```python
Da_mo = Matrix.triplet(wfn.Ca(), Da_so, wfn.Ca(), True, False, False)
```

Or, perhaps we could just add a small SO->MO demonstration to the `molden` docstring. Im kind of leaning towards the latter as it should make the function less complex.

amjames:
I agree keeping the functions simple is favorable. Adding documentation that demonstrates things like this is useful since the user can go and apply it elsewhere. I will add that now. Thanks!

amjames:
@dgasmith The molden function docstring also indicates that wfn will return Da/Db in the AO basis, should I edit that to be SO? 

Possibly confusing is someone reads it and expects to use wfn.Da to get AO basis density.

dgasmith:
@amjames Absolutely. 99% of the time in Psi when we say AO we really mean SO. As large parts are density fit with C1 restrictions these are interchangeable. There are only a few places where we truly want the pure AO density when irreps are present.

dgasmith:
@amjames Can you rebase this PR, I think we can get it in after that.

dgasmith:
LGTM

PR #509, opened 2016-11-09
Enables serialized JSON jobs

dgasmith:
## Description
Provides serialized input and output capabilities through a JSON interface. This is an experimental feature and the JSON interface itself will likely evolve significantly over various iterations.

## Features
- `psi4 --json data.json` is now a valid Psi input.
- All python based exceptions (not just Psi exceptions) in Psithon should now be printed to the Psi output as well as stdout.
- Dimension objects now have an `__iter__` attribute.

## Status
- [x]  Ready to go

### DIFF
--- a/psi4/driver/__init__.py
+++ b/psi4/driver/__init__.py
@@ -39,6 +39,7 @@ from psi4.driver import aliases
 from psi4.driver import diatomic
 from psi4.driver import wrapper_database
 from psi4.driver import wrapper_autofrag
+from psi4.driver import json_wrapper
 
 from psi4.driver.driver import *
 from psi4.driver.driver_cbs import cbs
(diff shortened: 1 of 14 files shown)


loriab:
second time through, still LGTM

amjames:
awesome! LGTM

andysim:
This is a really clean solution.  lgtm

andysim:
Not sure if you'd rather do a squash or just a plain rebase for the merge, so I'm going to punt.

dgasmith:
I dont think we want to do squash commits unless its like `change1`, `redid change1`, `tweaked change1`. Not a huge deal either way.

andysim:
Good to know.  For some reason I thought you preferred `1st part`, `2nd part`, `3rd part` to be rolled together if they're all contributions to the same overall feature.  I agree with your idea though; squash should be used for Trumpian commits.

dgasmith:
Probably would in general, it just seems that our commits seem to hit a lot of topics. For example, this is JSON, exception, and Dimension object handling.

andysim:
Still lgtm after being rebased (I opted not to go nuclear and skip the branch updating earlier)

PR #511, opened 2016-11-10
Python based options parsing

dgasmith:
## Description
Adds a helper functions to aid in options parsing from a Python input. For example currently one must do:

```python
psi4.core.set_global_option("SCF_TYPE", "DF")
psi4.core.set_global_option("BASIS", "6-31G")
...
```

Through the new helper you can set options by:
```python
psi4.set_options({"SCF_TYPE": "DF",
                  "BASIS": "6-31G"})
```

The question comes into play of what about local options:

Scheme 1:
```python
psi4.set_module_options("SCF", {"SCF_TYPE": "DF", ...})
```
or
```python
psi4.set_module_options({"SCF", {"SCF_TYPE": "DF", ...},
                         "DFMP2", {"MP2_TYPE": "DF", ...}})
```

Scheme 2:
```python
psi4.set_options([["SCF_TYPE", "DF"],          # 2 parameters - global
                  ["SCF", "SCF_TYPE", "DF"]])   # 3 parameters - local
```

Scheme 3:
```python
psi4.core.set_local_option("SCF", "SCF_TYPE", "DF")
```

Scheme 3 is the current setup where we leave local options alone as they are not all that common. 

From writing quite a bit of Python-based Psi4 this is quite annoying and *something* needs to be done with globals at the very least. Hopefully we can figure out a logical local scheme as well. 

As a note we are looking into ways to override global options. For example, the following could be a scheme that would accomplish this:

```python
mol = psi4.geometry("He")
opt = {"SCF_TYPE": "DF"}
psi4.energy('SCF', molecule=mol, options=opt)
```

Thoughts on any of the choices?

## Questions
- [x] Decision on global options input
- [x] Decision on local options input

## Status
- [x]  Ready to go


jturney:
Agreed that something needs to happen to make this simpler. I like the syntax for the global option given.  Of the three schemes given for local options, I'd prefer scheme 1.

loriab:
I'm less keen on scheme 2. Prefer scheme 1. At this point reconciliation looks simple because kwargs trump anything previously set. But by the time it gets complicated, I've got code for that.

andysim:
I like scheme 1.  Also, +1 using the term "module_options" instead of "local_options".  The proposed `set_options` syntax is very clean and I don't think many users will ever want to use module options.

amjames:
I also like scheme 1, and `module_options` over `local_options` is a great idea. 

> As a note we are looking into ways to override global options. For example, the following could be a scheme that would accomplish this

Don't module level options already override global options? In a standard input file doing this:

```python
# molecule stuff
set ccenergy print 1
set globals print 3
energy('ccsd')
```

Would have print=2 for the scf and print=1 for ccsd. Is that not the case when you do it with python? Like:

```python
# molecule stuff
psi4.core.set_local_option("ccenergy", "print", 1)
psi4.core.set_global_option("print",3)
```

dgasmith:
@amjames Those are the globals that I would like to override if I pass in options explicitly.

dsirianni:
I really like @dgasmith's proposal for dictionary passing of options, i.e.,
```python
opt = {'SCF_TYPE':'DF', ...}
psi4.energy('scf', molecule=mol, options=opt)
```
I think this really supports the interactive Psi4 that has only been possible post-inversion.

loriab:
Regarding over-riding, as long as the `energy(..., options={'BASIS', 'STO-3G'})` are executed as actual `psi4.core.set_global_option(...)` commands, there shouldn't be any special treatment needed for overriding options. This is still the global options object that's being set, though, yes? So those `options` are going to need to be reset and revoked in an `OptionsState` before return.

Sorry if this is out of place – with all the "overriding", I'm not quite following if it's the user-facing or programming process that's under discussion.

dgasmith:
I was thinking a head a bit where we would be able to kill off globals or at least provide a way to reduce them. I was hopeful that this scheme would be able to fill that roll as well to prevent duplication.

CDSherrill:
I agree with others that scheme 1 is preferable (fewer brackets than scheme
2), and that "module" is clearer than "local".

Also agree we need to plan ahead to simplifying options handling in general.

dgasmith:
Everyone seemed to like scheme 1 so I went ahead and pushed it along with a few other changes.

psi4/driver/p4util/python_helpers.py
```diff
@@ -93,18 +94,87 @@ def pybuild_wavefunction(mol, basis=None):
...
 core.Molecule.run_dftd3 = qcdb.interface_dftd3.run_dftd3
 
+
+def set_options(options_dict):
+    """
+    Sets Psi4 global options from an input dictionary.
+    """
+
+    for k, v, in options_dict.items():
+        core.set_global_option(k, v)
```
loriab: I think should do `k.upper()`. Since options setting has always been within developer scope, it's not (necessarily) case insensitive. The inputparser does this, so should imitate that here, I think.

psi4/driver/p4util/python_helpers.py
```diff
@@ -93,18 +94,87 @@ def pybuild_wavefunction(mol, basis=None):
...
+
+def set_module_options(module, options_dict):
+    """
+    Sets Psi4 module options from a module specification and input dictionary.
+    """
+
+    for k, v, in options_dict.items():
+        core.set_local_option(module, k, v)
```
loriab: `module` and `k` upper according to inputparser

dgasmith: Good point, should we do this to `v` if its a string as well?

andysim: Be careful with strings; we have case insensitive string vals, as well as case sensitive, to allow file locations to be specified.

loriab: exactly what Andy said. `v` shouldn't be upper-ed. it should be `strip()`-ed, though for whitespace. If you want to be thorough, can check it passes through `re.compile(r'(([$]?)([-+()*.,\w\"\'/\\]+))')` for options with `BASIS` in name, or same regex w/o `,` for all others.

dgasmith:
I uppercased the `module` and `k` strings, but left the value unmodified. Most of the stripping and translating was if this came from a pure-text input where this was needed. Personal thoughts for API level handlers is that these things should be correct from the get go. Any thoughts on this?

andysim:
Sorry, I wasn't specific enough before.  I have no objection to strings being sanitized at this level, but we should check with the options parser what the `type()` is first.  If we get `istring`, it shouldn't be uppercased, but `string` should.

loriab:
The inputparser doesn't alter the caps of `v`, so I think it's all right to leave it unsanitized in `set_options`/`set_module_options` and let the options class handle it. Inputparser _does_ caps `k`, so seems prudent to emulate.

andysim:
lgtm so let me know if you consider this finalized

dgasmith:
LGTM too. If we figure out a better way before 1.1 we could potentially change it then.

loriab:
lgtm

PR #513, opened 2016-11-11
Fixes Psi4 Build System

ryanmrichard:
## Description
One of the goals of the recent Psi4 infrastructure changes is to have Psi4 be callable from other packages.  Pulsar is very interested in doing this.  This PR is being created to hold the changes required to have Pulsar (or really any other program) include Psi4 in their build by using Psi4's `Psi4config.cmake` file.

## Todos
- [x] Fixes Findpybind11.cmake (and makes code around it case consistent)
- [x] Makes it so LD_PRELOAD is not required
- [ ] Create a working `Psi4config.cmake`

## Status
- [ ]  Ready to go


CMakeLists.txt
```diff
@@ -124,8 +122,21 @@ set(STAGED_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/stage${CMAKE_INSTALL_PREFIX})
...
+#
+# Intel has introduced libmkl_rt.so for exactly the scenario the new inverted
+# Psi4 presents, i.e. a library that is dlopened at runtime.  The point
+# of libmkl_rt.so is that at dlopen time it automatically loads the appropriate
+# mkl libraries instead of having to have rpaths to all possible libraries. We
+# should thus link against it and not try to play Intel link line advisor
+# ourselves.
+#
+# The final link line should be:
+# -L${MKLROOT}/lib/intel64 -lmkl_rt -lpthread -lm -ldl
```
loriab: Provided I can find a way around it for the purposes of static linking, `mkl_rt.so` os probably a good idea to detect, thanks. Do you happen to know if it can be made to respect sequential/threads via openmp enabling variable?

CMakeLists.txt
```diff
@@ -161,6 +175,7 @@ ExternalProject_Add(psi4_core
               -DPYTHON_EXECUTABLE=${PYTHON_EXECUTABLE}
               -DPYTHON_INCLUDE_DIR=${PYTHON_INCLUDE_DIR}
               -DPYTHON_LIBRARY=${PYTHON_LIBRARY}
+              -DPYBIND11_ROOT=${PYBIND11_ROOT}
```
loriab: That you're having to do all this FindPYBIND11 stuff is my fault for not getting the Config pushed upstream. I'll get on that.

external/libefp/CMakeLists.txt
```diff
@@ -28,7 +28,8 @@ else()
                    -DLIBC_INTERJECT=${LIBC_INTERJECT}
                    -DFRAGLIB_UNDERSCORE_L=OFF
                    -DFRAGLIB_DEEP=OFF
-        CMAKE_CACHE_ARGS -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}
+                   -DLAPACK_LIBRARIES=${LAPACK_LIBRARIES}
```
loriab: Yeah, I was relying on libefp.a blas symbols being satisfied by psi4's at link-time, so probably good to pass in directly. Will look into whether `LAPACK_LIBRARIES` or `LAPACK_INTERJECT` is more appropriate.

CMakeLists.txt
```diff
@@ -112,9 +112,7 @@ include(autocmake_safeguards)
 find_package(PythonInterp REQUIRED)
 find_package(PythonLibs ${PYTHON_VERSION_MAJOR} REQUIRED)
 
-add_subdirectory(external)
-
-    # external projects manage their own OpenMP and c++YY flags, so only add to CXX_FLAGS for psi4_core
+# external projects manage their own OpenMP and c++YY flags, so only add to CXX_FLAGS for psi4_core
```
ryanmrichard: I go back and forth on what's best for including external projects in a super build.  My current thought is it's best to forward all flags to the underlying projects so everything in the build uses the same options.  You only avoid passing certain options to them if those options interfere with their code.

CMakeLists.txt
ryanmrichard: I presume `MKL_N_THREADS` or whatever it is still works.

CMakeLists.txt
loriab: The trouble comes b/c there's not much facility for _un_-setting flags, and openmp and c++11 settings aren't cmake entities, rather they're expressed through flags. If openmp is enabled in psi4, that adds `-fopenmp` or whatever to `CMAKE_C_FLAGS`. Naturally that var gets sent to libefp extproj. But libefp needs to build w/o openmp (don't ask), yet because the flag is already in `CMAKE_C_FLAGS`, passing `ENABLE_OPENMP=OFF` to the libefp extproj has no effect.

CMakeLists.txt
jturney: Here's the MKL documentation for selecting features at runtime when using ```mkl_rt.so```

https://software.intel.com/en-us/node/528522

psi4/src/core.cc
```diff
@@ -829,33 +826,19 @@ bool py_psi_set_global_option_array(std::string const& key, py::list values, Dat
...
+            DataType* newentry = options.set_global_array_array(nonconst_key, entry);
+            py_psi_set_global_option_array(key,oi.cast<py::list>(),newentry);
         }
+        else if(can_cast<std::string>(oi))
+            options.set_global_array_string(nonconst_key, oi.cast<std::string>(), entry);
+        else if(can_cast<int>(oi))
+            options.set_global_array_int(nonconst_key, oi.cast<int>(), entry);
+        else if(can_cast<double>(oi))
+            options.set_global_array_double(nonconst_key, oi.cast<double>(), entry);
+        else throw PSIEXCEPTION("unrecognized type in py_psi_set_global_option_array");
```
loriab: You checked all this on the [options test case](https://github.com/psi4/psi4/blob/master/tests/options1/input.dat)?

psi4/src/core.cc
```diff
@@ -781,33 +792,19 @@ bool py_psi_set_local_option_array(std::string const& module, std::string const&
...
+            DataType* newentry = options.set_local_array_array(module,nonconst_key, entry);
+            py_psi_set_local_option_array(module,key,oi.cast<py::list>(),newentry);
         }
+        else if(can_cast<std::string>(oi))
+            options.set_local_array_string(module, nonconst_key, oi.cast<std::string>(), entry);
+        else if(can_cast<int>(oi))
+            options.set_local_array_int(module,nonconst_key, oi.cast<int>(), entry);
+        else if(can_cast<double>(oi))
+            options.set_local_array_double(module,nonconst_key, oi.cast<double>(), entry);
+        else throw PSIEXCEPTION("unrecognized type in py_psi_set_local_option_array");
```
loriab: All this looks tidier, to be sure, but what motivated the change? Any more resistant to pb11 updates?

psi4/src/psi4/libsapt_solver/usapt0.cc
```diff
@@ -2340,17 +2340,17 @@ void USAPT0::mp2_terms()
...
 
             if ( na_bblock > 0) {
-                fread(Aa_bsp[0],sizeof(double),na_bblock*nasQ,Aa_bsf);
-                fread(Ba_brp[0],sizeof(double),na_bblock*narQ,Ba_brf);
-                fread(Ca_brp[0],sizeof(double),na_bblock*narQ,Ca_brf);
-                fread(Da_bsp[0],sizeof(double),na_bblock*nasQ,Da_bsf);
+                size_t val=fread(Aa_bsp[0],sizeof(double),na_bblock*nasQ,Aa_bsf);
+                val=fread(Ba_brp[0],sizeof(double),na_bblock*narQ,Ba_brf);
+                val=fread(Ca_brp[0],sizeof(double),na_bblock*narQ,Ca_brf);
+                val=fread(Da_bsp[0],sizeof(double),na_bblock*nasQ,Da_bsf);
```
loriab: This looks like a bugfix where you don't actually care about `val`. What's the bug, that I may recognize it :-)

ryanmrichard: It silences the warning about ignoring the return value, which was cluttering my debug log.

psi4/src/core.cc
ryanmrichard: Only tests that fail are:
 - mints9 (uses functions that don't exist)
 - pywrap-checkrun-rhf (didn't compile with high enough A.M.)
 - pywrap-checkrun-rohf (same)
 - pywrap-checkrun-uhf (same)
 - python-energy (module doesn't define an init function)
-  python-curve (same)
-  json-energy (same)
-  json-gradient (same)

At this point I don't care to fix them.

psi4/src/core.cc
ryanmrichard: The previous code was broken by updates to Pybind11 and this solution encapsulates the offending code.  Is it anymore resistant?  Depends on how Pybind11 changes in the future; at the very least you'll only have to change the code in one place now.

ryanmrichard:
This PR is terribly out of date with the new build changes so disregard.

loriab:
There is the mklrt.so from this that I will incorporate sometime for #537 . But I agree with closing.

PR #514, opened 2016-11-11
Makes Psi4 fail gracefully when there are no active virtuals in DF-MP2

bwb314:
## Description
Previously, Psi4 was segfaulting when there were no occupied or virtual orbitals in a DF-MP2 computation. Now, it handles this a bit more gracefully by throwing.

## Status
- [X]  Ready to go

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -189,13 +189,23 @@ double DFMP2::compute_energy()
 {
 
     print_header();
+    if (Ca_subset("AO","ACTIVE_OCC")->colspi()[0] == 0) {
+        if (Cb_subset("AO","ACTIVE_OCC")->colspi()[0] == 0) {
+            throw PSIEXCEPTION("There are no occupied orbitals with alpha or beta spin.");
+        }
+        throw PSIEXCEPTION("There are no occupied orbitals with alpha spin.");
+    }
+    if (Cb_subset("AO","ACTIVE_OCC")->colspi()[0] == 0) {
+        throw PSIEXCEPTION("There are no occupied orbitals with beta spin.");
+    }
     if (Ca_subset("AO","ACTIVE_VIR")->colspi()[0] == 0) {
-        energies_["Opposite-Spin Energy"] = 0.;
-        energies_["Same-Spin Energy"] = 0.;
-        energies_["Singles Energy"] = 0.;
-        outfile->Printf("No virtual orbitals.\n\n");
-        print_energies();
-        return energies_["Total Energy"];
+        if (Cb_subset("AO","ACTIVE_VIR")->colspi()[0] == 0) {
+            throw PSIEXCEPTION("There are no virtual orbitals with alpha or beta spin.");
+        }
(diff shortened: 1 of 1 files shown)


psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -3639,6 +3648,10 @@ void RODFMP2::print_header()
     outfile->Printf( "\t %7s %7d %7d %7d %7d %7d %7d\n", "ALPHA", focc_a, occ_a, aocc_a, avir_a, vir_a, fvir_a);
     outfile->Printf( "\t %7s %7d %7d %7d %7d %7d %7d\n", "BETA", focc_b, occ_b, aocc_b, avir_b, vir_b, fvir_b);
     outfile->Printf( "\t --------------------------------------------------------\n\n");
+    
+    if (avir_a == 0 || avir_b == 0) {
+      throw PSIEXCEPTION("Number of active virtual orbitals is zero. Correlation energy = 0. a.u.");
+    }
```
dgasmith: You might want to edit this slightly to imply that both alpha and beta spins need virtual orbitals. Hard to imagine a case unless someone does Li in a 2s basis set though.

bwb314: I agree. I will add this soon.

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -187,7 +187,16 @@ void DFMP2::common_init()
...
+
     print_header();
+    if (Ca_subset("AO","ACTIVE_VIR")->colspi()[0] == 0) {
+        energies_["Opposite-Spin Energy"] = 0.;
+        energies_["Same-Spin Energy"] = 0.;
+        energies_["Singles Energy"] = 0.;
+        outfile->Printf("No virtual orbitals.\n\n");
+        print_energies();
+        return energies_["Total Energy"];
+    }
```
dgasmith: I still think this should fail as it has serious implications for anything that uses a `dfmp2` Wavefunction down the line. Gradients, natural orbitals, etc. We should catch the odd ball case here instead of having potential seg faults later in the code.

dgasmith:
This is a new for Travis. After 50 minutes: The job exceeded the maximum time limit for jobs, and has been terminated.

andysim:
They have an official 50 minute time limit on the .org site, but don't seem to have enforced it for a very long time.  I hope this was a one-off, because we're usually around that time.

dgasmith:
Everything ran under 50 minutes this time. Need to watch this closely. LGTM.

loriab:
From what I've watched, travis tends to be 30–40 min under new scheme, so we're getting better, except for the anomolies. LGTM.

andysim:
LGTM

dgasmith:
@bwb314 Can you rebase in current changes? Can be done with the following:

```
git remote add upstream https://github.com/psi4/psi4.git
git push --force origin master
```

bwb314:
Sure thing.

andysim:
Sorry about that, Brandon, it's my fault for closing out some of these PRs without paying attention to the order.

dgasmith:
LGTM

PR #515, opened 2016-11-11
Diatomic code change

andysim:
## Description
Completely changes the algorithm used for fitting diatomic potentials from Lagrange interpolations (was tens of thousands of lines of Python) to using [weighted least squares](http://dx.doi.org/10.1063/1.4862157), which can be implemented with a couple of lines of Python.

## Todos
- [x] Remove the old interpolation code.
- [x] Add weighted least squares routine.
- [x] Add exception to handle non-convergent cases.
- [x] Document changes.
- [x] New function to plot the fit to screen/file using matplotlib.

## Status
- [x]  Ready to go


CDSherrill:
This is diatomic subroutine is a neat feature to have, especially for
pedagogical purposes.

andysim:
Inspired by @CDSherrill's comment, I modified the plots to be a little more educational.  The code now plots the harmonic and anharmonic (quartic) fits on top of each other, and shows the first few energy levels as well as a comparison of r_e and r_0.

example plot

The two commits have been squashed and rebased to the current master, so should be ready to go.

dgasmith:
This is awesome! LGTM

loriab:
Very nice plot! We'll have to train Jupyter to pop that up upon fn call.

andysim:
I mentioned that to @dgasmith earlier, and he pointed out that, if the user has the correct inlining statement, it should work in Jupyter by passing the `plot_fit='screen'` argument.

amjames:
Not that I ever doubted it, but @dgasmith has it right.  [Confirmed](https://gist.github.com/anonymous/602a9fab43d864026c0a84b881a40096) works like a charm. Great Stuff! 

LGTM

andysim:
Thanks for checking on that, @amjames.  Good to know these things before merging.

loriab:
Sphinx definitely thanks you for not making it wade through 50k loc :-) LGTM

CDSherrill:
This new plot is really great!

I'll make sure the Psi4Education team sees this.

David

PR #518, opened 2016-11-16
Added save_xyz() to qcdb.Molecule

dsirianni:
## Description
Added `save_xyz()` function to `qcdb.Molecule` with `save_natom` as additional argument.
Will supersede `LibmintsMolecule.save_xyz()`.  

* **User Interest**
  - [x] Able to save .xyz file of QCDB molecule with (default) or without number of atoms as first line of file.

## Status
- [x]  Ready to go

### DIFF
--- a/psi4/driver/qcdb/molecule.py
+++ b/psi4/driver/qcdb/molecule.py
@@ -321,6 +321,16 @@ class Molecule(LibmintsMolecule):
                     x * factor, y * factor, z * factor)
         return text
 
+    def save_xyz(self, filename, save_ghosts=True, save_natom=False):
+        """Save an XYZ file.
+
+        >>> H2OH2O.save_xyz('h2o.xyz')
+
+        """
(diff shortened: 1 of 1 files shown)


andysim:
lgtm

CDSherrill:
I think this is a very useful little addition.  Will be very convenient for
our users.

jturney:
Isn't this functionality (minus the save_natom) already in the parent class? 
https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/libmintsmolecule.py#L1989

loriab:
Right, @jturney, this is just making the `save_natom`-enabled `save_xyz()` available.

loriab:
If future, we should all probably use the context syntax:

```
with open(filename, 'w') as outfile:
    outfile.write(self.save_string_xyz(save_ghosts, save_natom))
```

But perfectly good, as-is. LGTM

PR #524, opened 2016-11-21
python linking no more

loriab:
## Description
Mostly, get psi4 up-to-date with pybind again, remove extraneous python linking, and sanitize chemps2 & pcmsolver test case names and ctest labels.

## Todos
* **Developer Interest**
  - [x] add message-coloring module, useable by all, and used to warn about <4.9 GCC and highlight addon detection
  - [x] make autodocing tools aware of the detci properties psivars that are now set py-side
  - [x] re-work some list casting that (with pybind11 Nov 16 updates) broke our array options
  - [x] dropped the versioned code object library in psi4/src/CM
  - [x] documented python/pybind linking a bit
  - [x] ctest `shorttests` gone and `minitests` in
  - [x] `dmrg` tests and label now accessed through `chemps2`
  - [x] `pcmsolver` tests dropped "pcm" prefix
  - [x] tests involving external code now accessible with label `addon` rather than `ext-exe-dep`
  - [x] no longer linking to libpython

## Status
- [x]  Ready to go


andysim:
This lgtm, but it does remind me that one fine day we should clarify/unify the role of quicktests and shorttests.  There were a few, such as psithon1, that were not being tested by Travis even though they run in seconds.

loriab:
Yeah, I think `shorttests` should die. There's sense in having `quicktests` and `longtests`, but the intention of `shorttests` is probably met by `ctest -LE longtests`.

It's debatable whether the `quicktests` criterion is time or representative sampling. I'm not against time, but we do want to be mindful that we're regularly hitting the 50 min travis limit when we get old hardware. Though build time is probably a better target the cutting. Quicktests is currently about 10 min on travis.

robertodr:
LGTM @loriab Can I steal the [custom_color_messages.cmake](https://github.com/loriab/psi4/blob/work18/cmake/custom_color_messages.cmake) file for use in [Autocmake](http://autocmake.readthedocs.io/en/latest/)?

dgasmith:
I may not be impartial, but I like the CMake colors a lot. LGTM

We have been tossing around the idea of moving tests around to make a bit more sense.

- quicktests - Does psi work? tu1/scf-property/dfmp2-1/sapt1/cc1/casscf-sp only
- normaltests - Best coverage possible in a 20 minute envelope. (10 minutes on 2-thread travis)
- longtests - Anything that runs in 15 minutes
- no modifier - All psi tests

Obviously not in this PR, but something to ponder.

@robertodr I found the colors on Stackoverflow, so I would say yes :)

loriab:
@robertodr, go for it. I put the SO link at the top of the file.

On the test labels front, I'd suggest

#### Labels
* optests - "Does psi work? tu1/scf-property/dfmp2-1/sapt1/cc1/casscf-sp only" plus one test from each add-on and python, json, psi4numpy ("add-in"?). name definitely up for discussion.
* quicktests - tests that contribute to "Best coverage possible in a 20 min envelope. (10 minutes on 2-thread travis)" just because everyone's trained up to type "quicktests"
* longtests - tests that run in 15 min or more
* in particular, if a test doesn't belong to any of the three above, it shouldn't get _any_ length label. that will help with maintenance.

#### Runs
* `ctest -L optests` - runs the operational tests
* `ctest -L quicktests` - runs the decent coverage tests
* `ctest -LE longtests` - runs everything but the indiv tests that require >15 min
* `ctest` - runs "All psi tests (including the 30 minutes - 1.5 hour tests)"

CDSherrill:
Yes, I think the original idea of "quicktests" was that the developer could
quickly tell whether or not Psi4 was plausibly working correctly (so, a few
tests of major functionality, rather than the idea that all tests that are
quick would go here).  Eventually we crammed in quite a few small tests
here because people started getting lazy and only ever ran quicktests, so
coverage started to become a more major issue.  If we are now running most
of the tests nightly, this is less of an issue than it was, so the coverage
could be scaled back.

With regard to Daniel's suggested breakdown, we might try that but with
different names that might be even clearer:

* mintests - Minimal checking of psi; small number of tests, all of which
are short
* shorttests - Best coverage possible in a 20 minute envelope
* normaltests - Tests that would run automatically each night (could also
be nightlytests)
* longtests - Any extra-long tests that were useful in debugging or
performance evaluation, but that are too long to reasonably run every night
(my understanding is that's what we mean now by this label).

CDSherrill:
Yeah, could also use "quicktests" for the "coverage within 20 mins"

loriab:
@psi4/editors, if consensus builds on the ctest label refactoring, I'm willing to do it now and be done with it.

loriab:
This is ready to go again, if you all approve of the `minitests`/`shorttests` changes.

dgasmith:
Looks fine to me.

CDSherrill:
What gets run every night?  All of them?

loriab:
At the moment, just quicktests (~118) because `cc14` and `mints9` are still failing. The latter is my fault, the former is a mystery. Failures make the conda build fail, so the test suite needs to be clean before go back to running all.

CDSherrill:
If the tests can be fixed and we can go back to testing all test cases
soon, that's fine.  If there are persistent problems then the troublesome
tests could be removed from the list temporarily while they get patched up.

Issue #525, opened 2016-11-21
reducing compile time


loriab:
Kicking out Boost and libint has substantially reduced our compile time. However, we're still running up against Travis-CI's 50 min max (~10 min of that are in tests).

So, here's the facts. Any suggestions for reducing compile time welcome. There are thoughts in the air of moving `bin`s to external projects, where they can be tested separately, but less drastic suggestions preferred.

```
psi4            5216.132     100%
---------------------------------
dfocc            490.372     9.4%
ccdensity        489.191     9.4%
mints            327.399     6.3%
dpd              321.392     6.2%
psimrcc          297.722     5.7%
core             285.259     5.5%
ccenergy         242.782     4.7%
occ              227.323     4.4%
cclambda         167.325     3.2%
dcft             156.295     3.0%
cceom            156.036     3.0%
ccresponse       137.6       2.6%
fock             136.653     2.6%
sapt_solver      125.653     2.4%
detci            123.652     2.4%
optking          118.63      2.3%
ccsort           118.255     2.3%
cchbar           107.15      2.1%
mcscf            106.969     2.1%
cctriple          91.443     1.8%
iwl               90.624     1.7%
fnocc             77.272     1.5%
qt                69.421     1.3%
functional        65.807     1.3%
psio              59.924     1.1%
scf_solver        57.751     1.1%
trans             51.991     1.0%
cctransort        49.189     0.9%
ciomr             47.597     0.9%
3index            45.95      0.9%
fisapt            39.127     0.8%
adc               37.382     0.7%
transqt2          34.06      0.7%
thce              33.156     0.6%
findif            31.708     0.6%
scfgrad           30.889     0.6%
moinfo            26.539     0.5%
parallel          23.011     0.4%
dfmp2             19.045     0.4%
options           12.27      0.2%
psi4util          11.808     0.2%
plugin            10.044     0.2%
cubeprop           9.776     0.2%
efp_solver         9.012     0.2%
diis               7.406     0.1%
dmrg               7.377     0.1%
mrcc               6.772     0.1%
gdma_interface     4.447     0.1%
disp               4.221     0.1%
sapt               4.092     0.1%
cholesky           3.863     0.1%
thermo             3.677     0.1%
efp_interface      3.21      0.1%
filesystem         0.613     0.0%

```

Note to future (http://stackoverflow.com/questions/5962285/cmake-compilation-statistics):

```
>>> cat time-icpc
#!/bin/bash
[...]

PR #527, opened 2016-11-22
cease compiling unused lib files as per psi4/psi4#525

loriab:
## Description
Hopefully reduce compile time

## Todos
* **Developer Interest**
  - [x] Removed all the lib files identified as unused except those in libdpd

## Status
- [x]  Ready to go


amjames:
@lothain commented via email and it seems to have been lost in cyberspace:

>I know that Rollin's calculation ran with both codes, but I do not believe we have confirmed that the disk space requirements of cctranssort are reasonable

I was able to run the calculations that his comment refers to, both using cctransort and using ccsort/transqt2. The disk requirements for the transformation/sorting integrals were identical, so I believe we are safe in moving ccsort/transqt2 to the attic.

psi4/src/psi4/libmints/CMakeLists.txt
```diff
@@ -1,70 +1,70 @@
...
+                 intvector.cc
+                 potential.cc
+                 mintshelper.cc
+                 coordentry.cc
+                 kinetic.cc
+                 tracelessquadrupole.cc
+                 pseudospectral.cc
+                 integral.cc
+                 matrix.cc
+                 #svd.cc
```
jturney: It's save to delete ```svd.cc``` completely.

andysim:
LGTM, but we should consult others first.  Perhaps this is a good one to moot on our next monthly call.

dgasmith:
Just a general git point: This one and #534 are tied to the same branch history so neither should be rebased. Its one of the downsides of a rebase team policy. Merges are fine.

dgasmith:
This is tied into #534 now. Closing.

PR #530, opened 2016-11-28
WIP: style checks w/o git hooks

loriab:
## Description
Modelled (esp. the bash) on pybind11, this imposes a script checking for tabs and trailing whitespace on the C++ and flake8 on the python.

## Todos
* **Developer Interest**
  - [x] a dozen or so python files converted, incl. driver/driver and driver/procedures/proc. Many more skipped for now.
  - [x] flake governed by /.flake8 file
  - [x] C++ goverened by /cmake/check-style.sh file . Right now, just has a couple checks, and it's not even run in the travis b/c so many tabs in C++ code. 

## Questions
- [ ]  Do we want to do something like this? [Here's](https://travis-ci.org/loriab/psi4/builds/179363444) what it looks like. Compared to compiling, it's a very short build. Seems an easy way to get style checking that's non-binding (can still merge if fails) and doesn't involve everyone configuring their git hooks.
- [ ] I set max line width at 120, as I can live with that. It's open for discussion.

## Status
- [ ]  Ready to go


robertodr:
@loriab why not use [`clang-format`](http://clang.llvm.org/docs/ClangFormat.html#clangformat)?

loriab:
clang-format's fine by me, and I'd much rather have a flake8-like tool than hand-crafted bash scripts. We just need consensus on reformatting the whole code. Several people have piped up suggesting it these past few weeks, so I bet the 7 Dec conf call can produce a plan.

psi4/run_psi4.py
```diff
@@ -219,7 +219,7 @@
     tb_str += '\n'
     tb_str += type(exception).__name__
     tb_str += ': '
-    tb_str += exception.message
+    tb_str += str(exception)  # was exception.message so perhaps hybrid needed
```
dgasmith: Whats wrong with message? We dont actually want the full string representation here as its overloaded.

dgasmith:
Git-hooks is probably a non-starter, so +1 for avoiding that. Another +1 for clang-format, I use it internally quite a bit with Google style and 4 spaces.

Im kind of on the fence about this. A few notes:

Pros:
 - Completely uniform style

Cons:
 - We are constantly avoided errors, the `# pyflakes` and `# noqa` messages are messier than most of our python errors IMO.
 - Plenty of times when we want something other than pep8 formatting (e.g. `VARH`).
 - Completely uniform style (see previous point).
 - If we override a single error in pyflakes, pyflakes will continue to fail for all further commits. It would become very easy just to ignore this test.

loriab:
On the `exception.message`, change https://github.com/psi4/psi4/blob/master/psi4/driver/procedures/proc.py#L3934 from `== False` to `is False`, then run cepa2. You should get an exception that doesn't have a message, so it throws another exception.

On the Cons, yes, there are rather more kept errors than I would have expected. We can ignore them more broadly in `.flake8` if we choose, including all line lengths and imports not used. The noqa for things like VARH don't bother me too much. On the overriding a single error, yes, would have to insert to ignore code fairly promptly to keep getting good results. I've had the driver to within a few chosen exceptions of pep8 compliant, so maintaining it as such I don't think is too onerous.

dgasmith:
@loriab This works fine for me in Python 2 and 3. Can you provide the error message? This should be fine as the base Exception class always inits a blank `message`. We may be doing something wrong somewhere.

loriab:
Apparently the `str(mess` error I was getting last night was from a combination of errors. Can't recreate at present. So that part will have to be reverted until I hit the problem again and fully note its circumstances.

dgasmith:
Looks like this has not gained sufficient traction and has fairly serious conflicts with the driver at this point. Closing and we can reconsider at another time.

PR #526, opened 2016-11-22
DFOCC Header Cleanup

andysim:
## Description
Remove some includes from DFOCC, to try and speed up compilation

## Todos
Notable points that this PR has either accomplished or will accomplish.
* **Developer Interest**
  - [x] Removed some includes from DFOCC header files.

## Questions
- [x]  Does this look reasonable @bozkaya?

## Status
- [x]  Ready to go (pending approval from @bozkaya)

…clarations.

PR #534, opened 2016-11-29
Allow ccsort/transqt2 legacy modules from psi4pasture to be enabled at build time. 

amjames:
## Description
Legacy modules that were default in the past, but have been replaced will move to the [psi4pasture](http://github.com/psi4/psi4pasture) repo. The ccsort/transqt2 modules now live there.

This PR introduces some machinery for installing them from psi4's build, and allowing them to be used in the driver at run-time if they are installed.

## Todo
 - [x] cleanup history
 - [x] find specific libs that have missing symbols + add `${PRE/POST_LIBRARY_OPTIONS}` to each not all.
 - [x] Docs: Finalize build instructions for `PastureRequiredError`
 - [x] Docs: Add transqt2/ccsort note to read_options.cc
 - [x] Tests: add tests as lori mentioned 

* **Developer Interest**
  - I would say this functionality is experimental since pasture does not have any travis checks.
  - I am not completely confident in the way I have "injected" pasture.addins.ccsort_transqt2 into proc.py, but it seems to work okay. See the code comments
  - Does the pasture module even need to be installed w/in lib/psi4/driver?

  - ccsort/transqt2 are no longer part of psi4_core
  - adds ENABLE_ccsort and ENABLE_transqt2 flags to optionally build/install them from pasture

* **Self imposed CI checks**
 - [x] MacOS/clang/python2 build + use pasture ok
 - [x] MacOS/clang/python3 build + use pasture ok
 - [x] Linux/gcc/python2 build + use pasture ok
 - [x] Linux/gcc/python3 build + use pasture ok

## Status
- [x]  Ready to go


psi4/driver/procedures/proc.py
```diff
@@ -2172,8 +2172,14 @@ def run_ccenergy(name, **kwargs):
     if core.get_global_option('RUN_CCTRANSORT'):
         core.cctransort(ref_wfn)
     else:
-        core.transqt2(ref_wfn)
-        core.ccsort()
+        # try:
+        from psi4.driver.pasture import addins
```
amjames: I added this try/except block to gracefully deal with some kind of botched install. I commented it out because the raised exception was masking any errors I had from problems linking the ccsort/transqt2.so files to core.so.  I think its okay to remove it all together because if pasture is not installed using the run_cctransort option will throw before we get down here. 

Please try and break this. If you find a case where this line is causing an import error let me know.

psi4/driver/procedures/proc.py
```diff
@@ -2172,8 +2172,14 @@ def run_ccenergy(name, **kwargs):
     if core.get_global_option('RUN_CCTRANSORT'):
```
amjames: I probably didn't clean out my build directory correctly but removing the 'RUN_CCTRANSORT" option does cause this to throw.  Probably the source of Travis failures

psi4/driver/dependency_check.py
```diff
@@ -51,3 +51,8 @@
     import v2rdm_casscf
 except ImportError:
     pass
+
```
dgasmith: If were just going to pass, I dont think we need this here since the inversion.

psi4/driver/procedures/proc.py
dgasmith: Yes, the RUN_CCTRANSORT option needs to stay in `read_options.cc`. If an option is not a valid keyword an error is thrown.

psi4/driver/procedures/proc.py
dgasmith: As `RUN_CCTRANSORT` needs to stay valid, we probably need the try/except block to prevent possible seg faults or weird errors. This sort of area should be the only entry point for the addins.

CMakeLists.txt
```diff
@@ -201,6 +211,14 @@ add_dependencies(psi4_core ambit_external
                            libint_external
                            pcmsolver_external
                            pybind11_external)
+if(ENABLE_psi4pasture)
+  ExternalProject_Add(psi4pasture
+    GIT_REPOSITORY https://github.com/amjames2/psi4pasture
```
dgasmith: Dont forget to switch this over.

loriab: I think this if/endif should go in a `external/pasture/CMakeLists.txt`. I'm fine with keeping it in `external` folder unless ppl really want an `internal` folder.

loriab: Of the usual ones from, for example, [libefp](https://github.com/psi4/psi4/blob/master/external/libefp/CMakeLists.txt#L16-L31), you'll want to include at least https://github.com/psi4/psi4/blob/master/external/libefp/CMakeLists.txt#L16-L19, L22, L31 & CXX analogs for compiler and flags.

psi4/driver/procedures/proc.py
loriab: Guess we need to decide if pasture is (1) purely plugin-like where needn't be built/detected at build time and could instead be loaded at runtime and interacts only with driver code through hooks and no C++ code conditionals or (2) more a conventional addon library that does need to be built/detected at build time and can conditionally build C++ code.

If the latter, can attache a `USING_transqt2ccsort` compile definition to conditionally add `RUN_CCTRANSORT` option to read_options.

psi4/src/CMakeLists.txt
```diff
@@ -25,7 +25,11 @@ set(sources_list export_psio.cc
 add_library(core SHARED ${sources_list})
 
 get_property(psi4_binmodules GLOBAL PROPERTY BINLIST)
-target_link_libraries(core PRIVATE ${psi4_binmodules})
+if(ENABLE_psi4pasture)
+  target_link_libraries(core PRIVATE ${PRE_LIBRARY_OPTION} ${psi4_binmodules} ${POST_LIBRARY_OPTION})
```
loriab: I'll have to devise a better way on this. Because as soon as PCMSolver is turned on, there'll be two sets of _LIBRARY_OPTIONs on the final link line, and I promise CMake won't leave them alone but will instead optimize them to one. :-( Do you know exactly which libs symbols are needed from?

psi4/driver/procedures/proc.py
loriab: Will have to build this to understand what's going on to comment.

psi4/driver/procedures/proc.py
amjames: Here are two possible ways to handle this particular case that I think are okay.

1) Define `RUN_CCTRANSORT` in the pasture plugin. And then the block here would look like:

```python
try:
    from psi4.driver.pasture import addins
    addint.cctransort_override(ref_wfn)
except ImportError:
     # pasture is not installed
    core.cctransort(ref_wfn)
```
where `addins.cctransort_override()` is going to check the value of `RUN_CCTRANSORT` and decide if it should use `psi4.core.cctransort(ref_wfn)` or `psi4.core.plugin(transqt2_sofile,ref_wfn)` + `psi4.core.plugin(ccsort_sofile,ref_wfn)`

This would be business as usual since pasture would defines the default for `RUN_CCTRANSORT` as true. If pasture is not installed `RUN_CCTRANSORT` is not a valid argument so there is an error if it is used file. If `RUN_CCTRANSORT` is not used and pasture is not installed we end up in the `except` and we use the only thing we have. 


An alternate would be to put the `run_cctransort` option back in and then:

```python

if core.get_global_option('RUN_CCTRANSORT'):
    core.cctransort(ref_wfn)
else:
       try:
           from psi4.driver.pasture import addins
           addins.ccsort_transqt2(ref_wfn)
        except ImportError:
           raise PastureNotFoundError('RUN_CCTRANSORT')
```

and then a message like:

```
RUN_CCTRANSORT = false requires that [ccsort, transqt2] are installed from psi4pasture

 - How to build psi4 with pasture enabled 
 - How to install pasture into an existing psi4 install
```

The exception class could lookup `[ccsort,transqt2]` in a dict somewhere with the key `RUN_CCTRANSORT` that gets expanded when/if other modules move there. 

I like the latter, since it is self documenting, and can be expanded when/if more modules move out to pasture. The pasture requirement could be documented with the option in the manual, and once this is all working we can put a link in the manual to the pasture readme or wherever the install instructions end up.

CMakeLists.txt
amjames: Moving this to external/pasture/CMakeLists.txt works fine, as long as I leave it off of https://github.com/psi4/psi4/blob/master/external/CMakeLists.txt and instead put `add_subdirectory(external/pasture)` here. 

I put the pasture block here instead since it seems, everything in external is something that could be built without psi4 existing. Pasture is a slightly different beast and the dependency goes the other way.

loriab: because of the superbuild, externals is odd in that for each of them, either a real or dummy library is always built anyway. i think make an internals directory and `add_subdirectory(internals)` here. as you point out, this needs to happen at a different point in the main cmake processing, and i think we do want these hidden away in a subdir.

i think you've convinced me on using them as plugins instead of link in libraries. according to practice so far, they wouldn't even be included in the psi4 cmake build sys. instead you'd build on your own and make avaailble through pythonpath at runtime. but i think it is nice to be able to order them up though the main build. in which case maybe external --> extlibraries and internal --> extplugins. but that's for general discussion.

internals/pasture/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+    GIT_REPOSITORY https://github.com/psi4/psi4pasture
+    CMAKE_ARGS -DENABLE_ccsort=${ENABLE_ccsort}
+               -DENABLE_transqt2=${ENABLE_transqt2}
+               -Dpsi4_DIR=${STAGED_INSTALL_PREFIX}/share/cmake/psi4
+               -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+               -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+               -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+    CMAKE_CACHE_ARGS -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}
+                     -DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS}
+    INSTALL_COMMAND ${CMAKE_MAKE_PROGRAM} install)
```
loriab: If you build the pasture mods internally (as opposed to detecting pre-built), where are they getting installed? Or actually, do we/should we even have an installation mech for plugins? Mostly I'm curious how `CMAKE_INSTALL_PREFIX` and/or `DESTDIR` aren't needed.

CMakeLists.txt
```diff
@@ -211,14 +211,7 @@ add_dependencies(psi4_core ambit_external
...
                            pybind11_external)
-if(ENABLE_psi4pasture)
-  ExternalProject_Add(psi4pasture
-    GIT_REPOSITORY https://github.com/amjames2/psi4pasture
-    CMAKE_ARGS -DENABLE_ccsort=${ENABLE_ccsort}
-               -DENABLE_transqt2=${ENABLE_transqt2}
-               -Dpsi4_DIR=${STAGED_INSTALL_PREFIX}/share/cmake/psi4
-    INSTALL_COMMAND ${CMAKE_MAKE_PROGRAM} install)
-endif()
+add_subdirectory(internals/pasture)
```
loriab: thanks, trying to keep things tidy

psi4/driver/procedures/proc.py
loriab: I think I agree on the second option. Seems more extensible. And options are bad enough without overwriting them. Maybe stick a commented `USING_transqt2` near the option in read_options.cc, as a guide to searching.

internals/pasture/CMakeLists.txt
amjames: The pasture build systems uses psi4's config file to locate the driver directory and everything gets installed within `psi4_install_prefix/lib/psi4/driver/pasture`. 

I thought this would be a safe way to make sure that a pasture install is local to a *particular* psi4 install. If pasture gets installed into`PYTHONPATH` then any psi4 can find it, but there would probably be strange errors if the compiler-version, math libs, ect. didn't match between the psi4 and the pasture builds. 


We don't have an install procedure for plugins, but I guess I have artificially made one for the "pasture" plugins. The pasture `CMakeLists.txt` is not a clone of the one we give with plugins, but It is heavily inspired by it (since it was my starting point). I am trying to use as much info as possible from the psi4 config file, so that the build env for pasture is as close as possible to the "host" psi4. 

The way things are currently set up, there is no mechanism for detecting a pre-built pasture, rather it will always be built if requested. If you are building pasture outside of psi4's build system, then it will have to be built after, and will require a valid `psi4_DIR` be passed as an argument to cmake. The equivalent using `CMAKE_PREFIX_PATH` would work as well.

amjames: Also, when pasture is built at the same time as psi4:
Pasture's install location ends up being in the `stage` directory. However since it is installed before here before the larger project is copied from `stage` -> `final location`, when psi4 is installed the `stage/…../driver/pasture` directory gets copied to the final install location as well.

psi4/src/CMakeLists.txt
```diff
@@ -26,10 +26,9 @@ add_library(core SHARED ${sources_list})
 
 get_property(psi4_binmodules GLOBAL PROPERTY BINLIST)
 if(ENABLE_psi4pasture)
-  target_link_libraries(core PRIVATE ${PRE_LIBRARY_OPTION} ${psi4_binmodules} ${POST_LIBRARY_OPTION})
-else()
-  target_link_libraries(core PRIVATE ${psi4_binmodules})
+  target_link_libraries(core PRIVATE ${PRE_LIBRARY_OPTION} qt ${POST_LIBRARY_OPTION})
```
loriab: this will be fine for now. I wrote up last week the problems with _LIBRARY_OPTION, but that comment seems to have dematerialized. Basically, in doing this you're adding linker directives (`-Wl,--stuff`) as libraries, and cmake is known to rearrange them according to library rules. So `-llib1 -llib2 -llib3 -llib2` is perfectly fine optimized to `-llib1 -llib3 -llib2`. But when `lib2` isn't a library but `-Wl,--whole-archive` or `-Wl,--start-group`, then that "optimization" is a corruption. 

So adding this is just fine for now, but whenever we get more than one set of archiving directives (there's another one lurking for PCMSolver), then we'll have trouble. Targets get around this, which is partly why I'm gung-ho on them.

amjames:
I added tests that check ccsort/transqt2 give the same results as other tests provided they are installed. These tests will not fiddle with the `run_cctransort` option if pasture is not installed.

I also cleaned up some of my fixup commits since the twin PR has been closed. 

The last things I need to do is finalize the pasture install instructions to go in the PastureNotFoundError message, and then this is ready to go.

loriab:
Thanks for the tests. The `tests/pasture/modules/testdirs` layout is just fine on the face of it. But it's going to trouble those scripts and services that expect input.dat's at either `tests/purepsi4testdirs/` _or_ `tests/addon/integrationpsi4testdirs/`. Could you collapse a dir level, perhaps `add_subdirectory(pasture-transqt2ccsort` in `tests/CM`, then the two driver tests beneath?

amjames:
I moved everything up one level so `test/pasture/ccsorttransqt2/*` —> `pasture-ccsortrransqt2/*`

tests/CMakeLists.txt
```diff
@@ -104,7 +104,7 @@ endforeach()
 add_subdirectory(psi4numpy)
 add_subdirectory(python)
 add_subdirectory(json)
-add_subdirectory(pasture)
+add_subdirectory(pasture-ccsorttransqt2)
```
loriab: Any reason not to add a `if(ENABLE_ccsort OR ENABLE_transqt2)` around this and the "Adding Psi4 + transqt2 tests" message?

amjames: Since the test itself checks if pasture is available I didn't see any reason to. The test will run and pass if pasture is not installed. If pasture is installed it sets `run_cctransort false` and expects the same results to be obtained.

loriab: Looking again, I see what you mean. All the same, for ease of comprehension at a glance, I think seeing a "pasture-ccsorttransqt2/test_name .... PASSED" should mean that ccsorttransqt2 from pasture is installed and working correctly. Similar to the erd test that sets the integrals module to erd, even though it would pass just fine with libint. So I'd remove the logic from the pasture tests and only activate the tests if the transqt2/ccsort field from the pasture is active. Though I'm willing to entertain other views.

loriab:
Ok, I've looked this over again, and I'm willing to brave it. I don't think this [suggestion of @andysim's](https://github.com/psi4/psi4/issues/525#issuecomment-262108439) from the original PR discussion ever made it into a PR. Do you want to add that, @amjames, and rebase to upstream, then this'll be ready to go? After that I can run another timed compile.

amjames:
@loriab removed those lines and rebased. Just waiting on travis.

andysim:
They were removed in [this](https://github.com/psi4/psi4/pull/526/commits/94e6335e88f572bbf3bf57576947ef915a7a9b09) commit.

amjames:
@andysim, I assumed there was a miscommunication and I removed the same lines in arrays.cc. The build failed on my mac, so I will roll that last commit back.

loriab:
Ok, I checked the wrong repo. Sorry, miscommunication my fault.

PR #535, opened 2016-11-30
bug fix of prevention of zero target optimization values by the evil atof()

psi-rking:
## Description
Fixes bug reported [here](http://forum.psicode.org/t/fixed-dihedral-optimization-error/323).  I was sloppy and apparently no one had tried a 0 degree fixed angle before.

  - [ ] Fixes bug preventing fixed (target) values of exactly zero.

## Status
Passed HOOH at 0 degrees.  Still need to run other psi standard tests. 
- [ ]

### DIFF
--- a/psi4/src/psi4/optking/molecule_read_coords.cc
+++ b/psi4/src/psi4/optking/molecule_read_coords.cc
@@ -567,12 +567,15 @@ bool stoi(string s, int *a) {
 
 // convert string to float
 bool stof(string s, double *val) {
-  double i = atof(s.c_str());
-  if (i!=0) {
-    *val = i;
+  const char *word = s.c_str();
+  char *check;
+  double f = strtod(word, &check);
+  if (f == 0.0F && word == check)
+    return false;
+  else {
+    *val = f;
     return true;
   }
-  return false;
 }
 
 // convert string to boolean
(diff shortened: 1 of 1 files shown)


loriab:
Could `std::string` operations serve the purpose here? We had some `char *` issues with pybind11 and although there's no direct interface in optking, there was talk of moving from `char *`s to `std::string`s for readability.

psi-rking:
OK, I pushed a stringstream version.

psi4/src/psi4/optking/molecule_read_coords.cc
```diff
@@ -567,12 +567,24 @@ bool stoi(string s, int *a) {
 
 // convert string to float
 bool stof(string s, double *val) {
-  double i = atof(s.c_str());
-  if (i!=0) {
-    *val = i;
+   double d;
+   std::istringstream iss(s);
+   iss >> d;
+   if (iss.bad() || iss.fail())
```
fevangelista: Just out of curiosity: wouldn't std::stod() be more convenient/compact than the istringstream solution?

psi-rking:
I incorporated Francesco's suggestion of std::stod().  It worked fine in my tests.  Only for a hand-edited intco.dat would these checks ever matter.  (I discovered that if the user tries a string like "hello" for the value, it doesn't even pass through our initial psithon syntax code for some reason.)  The important thing is that we allow zero values.  BTW, I wonder why I can't see Francesco's comment on github's website (https://github.com/psi4/psi4/pull/535).  Are there private comments?

loriab:
There aren't private comments, but GH does hide comments that refer to outdated code. If you click the "Show outdated" above, do you see what you wanted?

loriab:
lgtm

andysim:
lgtm

dgasmith:
LGTM

Issue #483, opened 2016-10-19
SCF occupation and orbital origination


dgasmith:
We have the following SCF functionality as options:
- guess_read - Uses a previously computed guess orbitals, can cast up
- guess_basis - Computes a small-basis DF guess, can cast up
- broken_symmetry (no test cases, currently disabled)
- guess_mix - Conventional orbital guess proceeded by a mixing of orbitals

We can obtain orbitals from:
- Build a guess - HF internal function such as CORE/SAD/etc
- `geuss_Ca/Cb` - HF functions, need to be called before `iterations()`

We can perform the following internal conversions:
- Cast between basis sets, however not between symmetries
- Cast occupations between symmetries
- Build orbitals from a Fock guess

We can set occupation numbers in the following manner:
- Read from Nalpha/Nbeta
- User defined docc/socc
- Broken_symmetry forcing
- Read in occupation

As these are not mutually exclusive options a user can get some strange results for edge cases. Currently the following hierarchy is implemented:

1. User `DOCC`/`SOCC` always takes precedence (occasionally after the first iteration in SAD/cast examples)
2. `guess_read` occupation values *unless* the reference has changed
3. Nalpha/Nbeta as determined by the number of electrons and multiplicity

SCF is a bit strange as it is (effectively) the only wavefunction originator within Psi4 and cannot take in a reference_wavefunction. Some operations could be simplified if SCF could take in a wavefunction and always use the incoming values as its defaults. In addition, the current scheme could lead to some confusion if for example `broken_symmetry` and `DOCC` were both set. Its not overwhelmingly clear to me what the exact hierarchy should be and it may not be possible to catch every invalid input. Those of you who require strict SCF input settings should probably comment on what this logic should look like. As some of this logic has been reworked in the latest commits, difficult test cases should be tried.

See issues, #421, #420, and #419.

PR #538, opened 2016-12-04
Fix for cc14 test case

andysim:
## Description

Fixes cc14 by assigning SOCC and DOCC, which previously changed during the optimization.

## Status
- [x]  Ready to go

### DIFF
--- a/tests/cc14/input.dat
+++ b/tests/cc14/input.dat
@@ -16,6 +16,8 @@ set {
   max_disp_g_convergence  1e-6
   max_force_g_convergence 1.0e-6
   max_energy_g_convergence    7
+  docc [2, 0, 0, 1]
+  socc [1, 0, 1, 0]
   e_convergence 10
   r_convergence 10
 }


loriab:
Splendid, LGTM. Now I'm the only stumbling block with mints9.

dgasmith:
Great you found this! The occupation origination code did change when we removed boost, an issue about it is [here](https://github.com/psi4/psi4/issues/483). 

It looks like the behavior changed for this case. So, is this an issue with the input or with the new occupation code?

andysim:
I guess it's an issue with the occupation code, but I think it gets corrected as the optimization proceeds (which changes dimensioning, which causes the psio problems we were observing).  In the long term we should do a check for the occupation within the optimizer, and issue a `psiclean()` if it changes, so that DPD doesn't choke, but that's beyond the scope of what I have time for right now sadly.

andysim:
@dgasmith has a better fix, which actually addresses the underlying ROHF problem, so I'm going to close this in favor of his fix.

PR #539, opened 2016-12-05
migrate build docs from wiki to sphinxman

loriab:
## Description
Empty wiki and revise build docs for CMR-KTB-INV

## Todos
* **Developer Interest**
  - [x] all the user part of the Wiki migrated
  - [x] `.lgtm` discontinued
  - [x] since Summer 2016, the default installation has actually been to `/usr/local/`. Now back to `/usr/local/psi4` as traditional. Naturally, you can choose your own install loc via `-DCMAKE_INSTALL_PREFIX=/path/to/install/dir`

## Status
- [x]  Ready to go

~~This is very much a WIP, but I don't want it an invisible WIP.~~

Unpolished, but better than the Wiki, and better than nothing, so I'm calling it ready to go.


loriab:
This is ready for evaluation.

doc/sphinxman/source/build_planning.rst
```diff
@@ -0,0 +1,1152 @@
...
+
+How to save configuration settings for a future compilation
+-----------------------------------------------------------
+
+Create a file like ``do-configure`` with the ``cmake`` command and options
+*on one line*. ::
+
+ >>> cd {top-level-psi4-dir}
+ >>> cat do-configure
+     cmake -H. -Bobjdir5 \
```
dgasmith: The `objdir` name is sometimes the mentioned or `build, `objdir5`, and so on. Maybe check these over and always do `objdir` or `build`?

doc/sphinxman/source/build_planning.rst
```diff
@@ -0,0 +1,1152 @@
...
+
+After installation (:samp:`cd {objdir} && make && make install`), a directory
+structure like the below will exist at :samp:`/{prefix}`. This is a full
+installation.
+
+.. code-block:: bash
+
+ /
+ bin/                                                   (executables fro psi4 + any external proj)
+ bin/psi4                                                      (py script masquerading as c++ exe)
```
dgasmith: Maybe just the `psi4 executable` or something similar. Could be too much detail at this level.

doc/sphinxman/source/build_planning.rst
```diff
@@ -0,0 +1,1152 @@
...
+    >>> cat sample.py
+    import psi4
+    mol = psi4.geometry("""
+    He
+    """)
+    psi4.energy('hf/cc-pvdz')
+    psi4.compare_values(-2.85518839, psi4.core.get_variable('current energy'), 5, 'SCF E')
+    >>> python sample.py
+    SCF E.............................................................PASSED
+
```
dgasmith: I really like the paired module/executable details!

doc/sphinxman/source/build_planning.rst
```diff
@@ -0,0 +1,1152 @@
...
```
dgasmith: Could this be too much duplication? Might be more straightforward to just point out that `conda` is handeling the paths for you.

doc/sphinxman/source/build_planning.rst
```diff
@@ -0,0 +1,1152 @@
...
+.. _`faq:gdblldb`:
+
+How to use ``gdb`` and ``lldb`` with Psi4
+-----------------------------------------
+
+Debugging |PSIfour| has gotten a little confusing now that it's running through Python. Here's the syntax ::
+
+  >>> cd {objdir}
+  >>> lldb -- python stage/{prefix}/bin/psi4 ../tests/tu1-h2o-energy/input.dat
+  >>> (lldb) run
```
dgasmith: No `--` is required in `lldb -- python`.

doc/sphinxman/source/introduction.rst
```diff
@@ -498,15 +498,18 @@ Implementation within Psi4
...
+`https://github.com/psi4/psi4 <http://www.github.com/psi4/psi4>`_).
 
 |PSIfour| has been successfully compiled using Intel, GCC, and Clang
-compilers. For the Intel compilers, we recommend at least 12.1 (we have
-had trouble with version 12.0 and 13.0.1). GCC version 4.6 or above is
-recommended. For some architectures, a :ref:`precompiled binary
+compilers. For the Intel compilers, we recommend at least 2016 versions.
+GCC version 4.9 or above is
+required. For some architectures, a :ref:`precompiled binary
 <sec:conda>` is available. See :ref:`Compiling and Installing
```
dgasmith: Good catch, might want to note that this is true for Intel compilers here as well.

doc/sphinxman/source/build_planning.rst
```diff
@@ -0,0 +1,1152 @@
...
+.. 
+.. When you naively use Valgrind with Psi4, you're likely to get incomprehensible mess of garbage or it may just crash with a boost overflow error. This happens because the boost python layer looks really really bad as far as Valgrind is concerned, i.e. it looks like a ton of memory leaks.  It really isn't, so we want to ignore all such errors/warnings. Valgrind has a mechanism for this in the way of suppression files.  Calling Valgrind as:
+.. 
+.. ```bash
+.. valgrind --suppressions=<file_name>
+.. ```
+.. 
+.. will run valgrind with the suppression file located on disk at "file_name". Lucky for you, Psi4 comes with such a suppression file at [``$top-level-psi4-dir/lib/valgrind-python.supp``](../blob/master/lib/valgrind-python.supp). This should remove all the boost python errors.
+..  
+.. The other error, boost overflow error arises from ``src/lib/libmints/sieve.cc`` where the inverse
```
amjames: Are these boost related Valgrind notes still necessary?

doc/sphinxman/source/build_planning.rst
loriab: Good catch. Should be `{objdir}` when trying to be general and `objdir` when doing simplest prescribed path. Fixed the `objdir5`.

doc/sphinxman/source/build_planning.rst
loriab: I'd like it to be a little advanced. Switched to `psi4 executable, actually just a py script`

doc/sphinxman/source/build_planning.rst
loriab: Thought the ``--`` was how you got args in on the command line, instead of having to split the normal command into lldb-launch and run-rest-of-args steps. From [here](http://lldb.llvm.org/lldb-gdb.html). Maybe can do same with gdb.

doc/sphinxman/source/introduction.rst
loriab: I think having this here at all is superfluous as (1) violations caught at build time and (2) this isn't comprehensive and is covered elsewhere. So changed this to a link to section and mention of 4.9

doc/sphinxman/source/build_planning.rst
loriab: No, they're not. But the next time I hear of someone running valgrind, I'm going to quiz them on any procedural changes, then revive this section (commented now) with accurate info. The Boost bit is dead, but the suppression file and possibly something about sieving may endure.

CMakeLists.txt
```diff
@@ -118,6 +118,11 @@ include(GNUInstallDirs)
 include(autocmake_safeguards)
 include(custom_color_messages)
 
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    set(CMAKE_INSTALL_PREFIX "/usr/local/psi4" CACHE PATH "Install path" FORCE)
```
jturney: Wait, why change the default install from ```/usr/local``` to ```/usr/local/psi4```?  We install everything into a ```psi4``` subdirectory:  ```$prefix/include/psi4```, ```$prefix/share/psi4```, ```$prefix/lib/psi4``` but the binaries stay in ```$prefix/bin```.  Changing the default from ```/usr/local``` means users have to modify their path to add ```/usr/local/psi4/bin```.

loriab: Because psi3 and psi4 have always defaulted to `/usr/local/psi4`, with `/bin`, `/include`, etc. underneath, and I supposed we wanted to continue that.

loriab: Will make this a public discussion. I prefer encapsulated, but I'm not adamant.

The main business installs to ``{CMAKE_INSTALL_PATH}/lib/psi4/core.so`` which isn't good python installation protocol anyways (should install to detected python's site-packages for import psi4 to do any good automatically), so I think an encapsulated installation makes more sense for non-conda

PR #540, opened 2016-12-07
CCHBAR remove unnecessary I/O in ovvv intermediates 

amjames:
## Description
Rewrite of CCHBAR's Wabei algorithm for UHF reference. The old algorithm did quite a bit of unnecessary I/O. The replacement has been designed to compute as many sub-intermedates in the target order to avoid storage of additional intermediate terms, and to avoid expensive sorting of ov^3 buffers. 

## Todos
* **Developer Interest**
  - [x] TDC has requested I take any integral sorts that were required within the Wabei_XXXX functions and move them to some setup function so that they are all preformed before any of the functions run. 
  - [x] I also need to remove the first commit that I added from #513 that let me build + run on our cluster. 
  This change results in a performance boost, but no end-user changes should be noticed. 

## Status
- [x]  Ready to go


psi4/src/psi4/cchbar/Wabei_BBBB_UHF.cc
```diff
@@ -59,198 +60,191 @@ namespace psi { namespace cchbar {
...
+   *
+   * --AMJ 02/19/2016
+   **/
+  build_Z1_BBBB();//Z(ia,mf)
+  global_dpd_->buf4_init(&Z, PSIF_CC_TMP0, 0, 30, 30, 30, 30, 0, "Z1(ia,mf)");
   global_dpd_->buf4_init(&F, PSIF_CC_FINTS, 0, 15, 30, 15, 30, 0, "F <ai||bc> (ab,ic)");
-  global_dpd_->buf4_init(&T2, PSIF_CC_TAMPS, 0, 30, 30, 30, 30, 0, "tiajb");
-  global_dpd_->contract444(&F, &T2, &Z, 0, 0, -1, 0);
-  global_dpd_->buf4_close(&T2);
+  global_dpd_->buf4_init(&W, PSIF_CC_TMP0, 0, 15, 30, 15, 30,  0, "W1(be,ia)");
```
loriab: Are items like this changing the identity of psi scratch files, such that the comments in psifiles.h need editing? Or just labels internal to DPD?

amjames: No, `PSIF_CC_TMP0` will still be the same file, I am creating a temporary DPD buffer here. The `PSIF_CC_TMPX` are wiped at the end of each CC* module (see [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cchbar/cchbar.cc#L203) ), so I have changed some of the buffer that get written there as needed.  These won't mess up anything else.

loriab: very good then, thanks.

PR #542, opened 2016-12-08
CI: Moves to zero-root indexing

dgasmith:
## Description
As discussed CI moves to a zero-indexed scheme.

## Todos
* **Developer Interest**
  - [x] Zero-indexing scheme.
  - [x] Returned value for CAS was returning lowest root instead of average for SA computations. 
  - [x] CI Roots are now zero index (the ground state is now 0 instead of 1).

## Status
- [x]  Ready to go

### DIFF
--- a/psi4/driver/procedures/proc.py
+++ b/psi4/driver/procedures/proc.py
@@ -2585,7 +2585,7 @@ def run_detci_property(name, **kwargs):
     states = core.get_global_option('avg_states')
     nroots = core.get_global_option('num_roots')
     if len(states) != nroots:
-        states = range(1, nroots + 1)
+        states = range(nroots)
 
     # Run OEProp
     oe = core.OEProp(ciwfn)
@@ -2602,7 +2602,6 @@ def run_detci_property(name, **kwargs):
         core.print_out("\n   ===> %s properties for all CI roots <=== \n\n" % name.upper())
         for root in states:
             oe.set_title("%s ROOT %d" % (name.upper(), root))
-            root = root - 1
             if ciwfn.same_a_b_dens():
                 oe.set_Da_mo(ciwfn.get_opdm(root, root, "A", True))
             else:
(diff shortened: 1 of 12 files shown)


psi4/src/read_options.cc
```diff
@@ -489,7 +489,7 @@ int read_options(const std::string &name, Options & options, bool suppress_print
     (the one-particle density matrices are written for all roots).
     Useful for a state-specific CASSCF or CI optimization on an
     excited state. -*/
-    options.add_int("FOLLOW_ROOT", 1);
+    options.add_int("FOLLOW_ROOT", 0);
```
loriab: Also,
* https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L318
* https://github.com/psi4/psi4/blame/master/doc/sphinxman/source/glossary_psivariables.rst#L340


Can https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L2075 be changed cosmetically easily?

@fevangelista, https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L2283 is typically ground state, yes? Can it's external ordering be switched to 0?

psi4/src/psi4/detci/opdm.cc
```diff
@@ -571,120 +570,6 @@ void CIWavefunction::opdm_block(struct stringwr **alplist, struct stringwr **bet
 
 }// End OPDM Block
 
-/*
-** Here we compute all properties for opdms in opdm_map_ from form_opdm
-*/
```
loriab: This whole section is now what's handled by `oeprop()` in driver?

psi4/driver/procedures/proc.py
```diff
@@ -2594,7 +2594,7 @@ def run_detci_property(name, **kwargs):
         core.print_out("\n   ===> %s properties for all CI roots <=== \n\n" % name.upper())
         for root in states:
             oe.set_title("%s ROOT %d" % (name.upper(), root))
-            root = root - 1
+            # root = root - 1
```
loriab: Is there a comment in detci somewhere about external/internal root numbering? Are they equivalent now?

psi4/src/psi4/detci/opdm.cc
dgasmith: Yup, no need for it C side.

psi4/src/read_options.cc
loriab: Ah, sorry, on third point. I read the description and thought it was plain SCF.

dgasmith:
This should be ready to go and is rebased up to master.

PR #547, opened 2016-12-13
Analytic RHF-CCSD(T) gradients using libdpd in cctriples module

ashutoshvt:
## Description
Implements RHF-CCSD(T) first derivatives making use of the libdpd infrastructure inside
cctriples module.

## Status
- [X] Ready to go


psi4/driver/procedures/proc.py
```diff
@@ -870,7 +870,7 @@ def select_ccsd_t__gradient(name, **kwargs):
     # Considering only ccenergy
 
     func = None
-    if reference == 'UHF':
+    if reference == 'RHF' or reference == 'UHF':
```
loriab: `if reference in ['RHF', 'UHF']:`

tests/cc13b/input.dat
```diff
@@ -0,0 +1,50 @@
...
+   basis cc-pvdz
+   r_convergence 13
+   e_convergence 13
+   d_convergence 13
+}
+
+analytic, wfn = gradient('ccsd(t)', return_wfn=True)
+
+set findif points 5
+fin_diff, wfn = gradient('ccsd(t)', dertype=0, return_wfn=True)
```
amjames: This test doesn't take very long to run (1-2 min on 1 thread). I wonder if we could hard-code in the results from finite differences in for the comparison, and possibly move this to quick-tests?

I could see RHF-CCSD(T) gradients being a popular feature and it would be a good idea to make sure it hasn't broken.

tests/cc13b/input.dat
```diff
@@ -0,0 +1,50 @@
...
+
+compare_matrices(analytic, fin_diff, 10, "RHF-CCSD(T) gradient test") #TEST
```
loriab: Concur with @amjames that it would be good to have a quick rhf-ccsd(t) grad in quicktests. In the cfour tests, you'll find [some gradients](https://github.com/psi4/psi4/blob/master/tests/cfour/opt-rhf-ccsd_t_/output.ref#L1024-L1034) if you'd like to compare with that.

ashutoshvt:
Sure, will do. Thanks for the feedback !!

tests/cc13b/input.dat
dgasmith: 2 minutes on 1 thread would be 10-20% of the entire time allotted to all 130 quicktests.

amjames: Ah, I realize we were working with that little. @ashutoshvt is going to cut some of the extra/unnecessary stuff out, any guess on what would be the longest we could consider acceptable?

dgasmith: Could something like Water/6-31G fit in a 15-20 second envelope? By all means populate longtests as much as you want, but quicktests should be as bare bones as possible.

ashutoshvt: @dgasmith  I have two test cases: water for checking if my code works with symmetry and N2 in C1 symmetry for checking the degeneracies handling. If I use 6-31 G basis the total wall time is around 6 seconds on 1 thread where I am comparing the results with hard-coded finite difference 5 point formula results to 10 digits. So now can I put these tests to the quick tests ??

dgasmith: Sure! Anything thats a few seconds is perfectly fine. Please do populate long tests with more in-depth tests.

ashutoshvt: Sure, I can compare my geometry optimization results against CFOUR in my long tests.

tests/cc13b/input.dat
ashutoshvt: @loriab Can I hard code the cfour gradient results in my test case or do you want me to run a wrapper like: energy('cfour') and then compare my results?

loriab: Hard code, please. It's faster, and we don't want normal tests depending on an external exe.

dgasmith:
If this is ready to go please rebase it up to the master branch and we will get it in next, `git pull --rebase upstream master`.

ashutoshvt:
@dgasmith I have rebased it to the master and its ready to go now.

tests/cc13b/input.dat
```diff
@@ -0,0 +1,56 @@
...
+
+analytic, wfn = gradient('ccsd(t)', return_wfn=True)
+ref_vals = [
+             [ 0.00000000000000,     0.00000000000000,    -0.02788939642641],  # FINITE DIFFERENCE 5 POINT FORMULA
+             [ 0.00000000000000,    -0.00904503210340,     0.01394469821321],
+             [ 0.00000000000000,     0.00904503210340,     0.01394469821321]
+           ]
+fin_diff = psi4.Matrix(3, 3)
+fin_diff.set(ref_vals)
+compare_matrices(analytic, fin_diff, 10, "1. RHF-CCSD(T) gradient comparison with finite differences") #TEST
```
loriab: 10 is way too tight here and below; I'd go for 6.

loriab: 5–8 is what's in the compare_matrices in tests currently

ashutoshvt: I wanted to compare it to 10 decimal places since I cranked up the convergence to 1e-13. But I can lower it down.

loriab: I agree, it _looks_ like it should work. But all the same, I don't think it will over lots of architectures. Go for 8 if you like, and we'll see how it fares in the wild.

ashutoshvt:
We don't have frozen core capabilities in any of our analytic coupled cluster gradients yet, so it should fail gracefully.

Issue #548, opened 2016-12-13
Goal: Make Psi4 work under MPI (again)


ryanmrichard:
This issue is intended more as a discussion, than a real issue.

At the moment it is not possible to call Psi4 under MPI with more than one process per node because of [this](https://github.com/psi4/psi4/blob/fe3ab0d351e05276185ad838f3bcf2c0321be2fe/psi4/driver/procedures/proc.py#L1340)

Basically this creates a file that has a non-unique name on each process and then when it goes to delete the file each process deletes the same file leading to an error for every process after the first one.  Although this could be fixed easily (either only write to some master process or ensure that the filename is unique.  The latter is easily done [in python](http://stackoverflow.com/questions/2961509/python-how-to-create-a-unique-file-name) and [in C++](http://en.cppreference.com/w/cpp/io/c/tmpnam), or by using the process ID) this has been a problem before and is likely to be a problem again, hence I am hoping to reach a more permanent solution.

To me the most obvious solution is to make a file writer (stream writer is even better as it then also works for printing, in turn killing two birds with one stone) object/library that handles all read/writes and does so in a way that will not interfere with MPI.  In theory `libpsio` could be used for this, but it would have to be extended to writing other types of files aside from binary (I'm under the impression that it only writes binary files).

Thoughts?

dgasmith:
I think its this way so that people can use the `READ` keyword for failed jobs without a PID. We do have a PID restart mechanism; however, its not tested or used AFAIK. 

There are quite a few read/writes coming from Python libraries these days so a pure Psi4 C++ module would not work. You could have a unified namespace that includes the PID. This is what `core.get_writer_file_prefix(scf_molecule.name())` tries to do; however, without a PID.

loriab:
We'll also need to consider [interface codes](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/interface_dftd3.py#L152-L157) that do file writing, too. Is it wanted to keep only mpi process == 1 writing the files or to keep the filenames unique so it doesn't matter if N mpi processes are writing the files? Or either, depending on the situation?

ryanmrichard:
@dgasmith 
- I don't understand what difference the PID makes in terms of a read command
     ~~~~.py
    set{
    READ "some_file_with_no_PID.input"
    READ "some_file_generated_from_process.1234"
    }
    #later...
    file2open=options.get_str("READ")
    f=open(file2open,'r')
    ~~~~
[...]

PR #549, opened 2016-12-14
File names now have PID

ryanmrichard:
## Description
Addresses issue #548 by appending the process identification (PID) number to file prefixes generated with
the `get_writer_file_prefix` function.

## Todos
* **Developer Interest**
  - [x] PID is now appended to all file prefixes generated with the `get_writer_file_prefix` function
  - [x] Fixes `const` correctness in this function as well as simplifies internal guts
  - [x] Now ignores all file extensions (not just `.out` and `.dat`) when determining base file name
  - [x] Some (all at the moment) of the generated ASCII file's names now contain the PID

## Status
- [x]  Ready to go


ryanmrichard:
Before this gets merged, do we want to also do the `set guess file` change in it?  It's sort of in the spirit of this PR.  That of course requires feedback from the community.

dgasmith:
Can you rebase this up to master. We can get it in next.

ryanmrichard:
For the record this PR does not change the `set guess file` thing, largely because it appeared to be a bigger change than I initially thought.

ryanmrichard:
@loriab see modified description

PR #550, opened 2016-12-14
Slight clarification in libdpd documentation

matt-chan:
## Description
It took me way too long to realize buf4_init could have non-zero data. Just adding a minor modification to the docstrings to clarify it a bit. 

If accepted, please do a squash commit. I edited the docstring from github directly, so there's 1 commit per change. 

## Todos
* **Developer Interest**
  - [ ] Feature1
  - [ ] Feature2

## Questions
- [ ]  Question1

## Status
- [ ]  Ready to go

### DIFF
--- a/psi4/src/psi4/libdpd/buf4_init.cc
+++ b/psi4/src/psi4/libdpd/buf4_init.cc
@@ -52,7 +52,8 @@ namespace psi {
 **                   data as it will be stored on disk.
 **   int anti: Boolean flag which indicates whether the data needs to
 **             be antisymmetrized as it is read from disk.
-**   char *label: The string labelling the PSIO TOC entry on disk.
+**   char *label: The string labelling the PSIO TOC entry on disk. If it
+**                exists in cache, it will be loaded.
 */
 
 int DPD::buf4_init(dpdbuf4 *Buf, int inputfile, int irrep, int pqnum, int rsnum,
--- a/psi4/src/psi4/libdpd/file2_init.cc
+++ b/psi4/src/psi4/libdpd/file2_init.cc
@@ -48,7 +48,8 @@ namespace psi {
 **             dpd_init()].
 **   int qnum: The orbital subspace number for the right index [see
 **             dpd_init()].
-**   char *label: A string labelling for this buffer.
+**   char *label: A string labelling for this buffer. If it exists 
+**                in cache, it will be loaded.
 **   int print_flag: A boolean for the print routines.
 **   std::string OutFileRMR: The formatted output file stream.
 */
--- a/psi4/src/psi4/libdpd/file4_init.cc
+++ b/psi4/src/psi4/libdpd/file4_init.cc
@@ -49,7 +49,7 @@ namespace psi {
 **   int rsnum: The index combination for the ket indices for the
 **              data as it will be stored on disk.
 **   char *label: A string labelling for this buffer. If it exists
-**              in cache, it will be loaded.
+**                in cache, it will be loaded.
 */
 
 int DPD::file4_init(dpdfile4 *File, int filenum, int irrep, int pqnum,
--- a/psi4/src/psi4/libdpd/file4_init.cc
+++ b/psi4/src/psi4/libdpd/file4_init.cc
@@ -48,7 +48,8 @@ namespace psi {
 **              data as it will be stored on disk.
 **   int rsnum: The index combination for the ket indices for the
 **              data as it will be stored on disk.
-**   char *label: A string labelling for this buffer.
+**   char *label: A string labelling for this buffer. If it exists
+**              in cache, it will be loaded.
 */
 
 int DPD::file4_init(dpdfile4 *File, int filenum, int irrep, int pqnum,


lothian:
Can you clarify what was giving you trouble?  A buf4_init() call shouldn't yield any data, because it doesn't attempt to allocate memory for the tensor data itself, just meta-data like row/column lookups.  The underlying file4_init() call *can* be associated with data that's already in cache (in which case I wouldn't describe it as "loading", just assigning a pointer to data in memory).  Sorry to be picky; I just want to the docs to match what's happening in the code.

amjames:
I believe I know exactly what you have experienced, please correct me if I am wrong. 

It is easy to come to the conclusion that buf4_init is like a constructor. Leading to the conclusion that when you pass a `dpdbuf4` to the `buf4_init` function you get a brand new instance of a class with data initialized to zero. However these are not objects in a proper OOP sense, and for that reason the when/why of data moving into `buffer->matrix` can be confusing.

The comments here should have some warning that the data that *will be* pointed to by`buffer->matrix`, will only be "blank" if the tensor that this (`filenumber`, `label`) combination refers to does not exist. Exist here has a very broad scope since you may be referring to an entry in a file which was written by some other module, and often it is up to the programmer to know "This quantity was written by module A, and it is safe for me to assume that it will be on disk when the code I am writing in module B will run because I made sure that module A runs before module B in the driver"

If you expect/want a tensor to be zero-ed out you should always do that explicitly when manipulating the data. This detail is easy to overlook since common operations using the `contractXXX` methods may be doing that without you realizing it. For example by setting the pre-factor on the target to zero. 

There are plans in the works for a fairly major overhaul of libDPD to address various problems, one being that the interface is very cumbersome. Hopefully those efforts will make it much easier to get comfortable writing code that uses DPD.

amjames:
To expand on my request for changes a little more, the wording is problematic since the problem I believe you are getting at has nothing to do with the file being in cache, it has to do with the file-entry existing on disk. 

Yes if the quantity is held in cache, the data is accessible right away, but if it is not held in cache you may still be modifying a non-zero quantity depending on how you manipulate the buffer later.

matt-chan:
Yep, @amjames described my confusion exactly (and very clearly too!).

So the wording should be modified to indicate that if the (filenum, label) combination was called at an earlier point, then it will return non-zero data?

One other small thing that would be nice to clarify (but I don't know how to do it clearly) is that filenum or label alone is not enough to determine which data is returned by buf4_init and friends. I didn't have a clear concept of how filenum and label related to each other from reading the original docstring. I ended up realizing later on there was a contradiction in the CCENERGY code if I assumed filenum or label alone were enough.

amjames:
@matt-chan  Regarding the note in the documentation, I think it would be best to leave the comments on the arguments alone, and in the general description of file4/file2 init we can add a line to the end. 

`Note: if label is found in the TOC for FILENUM the data in that file-entry read in/overwritten by subsequent functions using this buffer. If label is not found in the TOC for FILENUM 

If @lothian is fine with that, and you think it would help, I think that is enough to help people out. I wouldn't want to confuse the issue further by putting notes about files in `buf4_init`, or suggesting that the cache system is somehow involved.

Regarding your question: First, could you show me where that contradiction is in CCENERGY?
Second, to clarify the file-number, label combination is enough to specify the quantity you want to retrieve, with the important caveat that you (the programmer) also know *everything* that has been done with that file-entry up to the point where your code is using it. Which may include the file it lives in being having been wiped clean (if it was a temporary file, for example).

Also a side note:
> … which data is returned by buf4_init and friends ..

I will reiterate that `buf4_init()` does not return any data.  It may be if the quantity is stored in cache, and after you `buf4_init()`  you can access the data in `dpdbuf4->matrix` but that is not something you can rely on since setting `cachelevel = 0` in the input file would mean code relying on something staying in cache will break. You need to have calls to the appropriate matrix initialization/reading functions as well. This is not obvious at first glance at libdpd since the commonly used `contractXYZ` and other similar functions are doing a lot on the inside inside.

matt-chan:
Thanks for the clarification @amjames. I think the line at the docstring as you suggested might be better, but it would be even clearer if it explicitly mentions the possibility of non-zero data residing in the file/buffer?

The code I was looking at is here at [line 79-90](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/ccenergy/mp2_energy.cc). If I understand you though, this code would break if someone set `cachelevel = 0`? It makes no sense to take the dot product or direct product of zero arrays.

amjames:
The lines you highlighted are completely valid since within [`dirprod`](https://github.com/psi4/psi4/blob/9618f2716c4ff16a6c48276bcd8e4b96adc29fec/psi4/src/psi4/libdpd/buf4_dirprd.cc#L55) for example:
 - **lines 55+56**: memory is allocated for the elements of each tensor (one irrep at a time) , 
 - **lines 57+58**: The elements are read from the file and used to populate the memory we just allocated 
- ** line 60**: the operation is performed for the block (irrep) we are working with.
- **line 63**: The modified elements are written back to disk, replacing the data that was read in on line 56
- **lines 64 + 65**: The memory for this irrep is freed.

This is why I want to move away from talking about data at all in the comment string, `buf4_init` is not responsible for initializing the "data" you are talking about. That is by *design* so that operations can be performed over the tensors in chunks like I have outlined in the part above. The buffer holds the information necessary to get that data, *when it is explicitly told to*. 

I am sorry for not being clear, but this stuff is very easy to get lost in. I meant that trying to access tensor elements directly after calling `buf4_init`.

Getting back to the subject of this PR: 

 1. Lets leave the comments in `file4_init` alone, since the `buf4/file2_init` functions are the ones being used externally, `file4_init`s are usually only found w/in DPD itself as far as I know.. 
2. In `buf4_init()` and `file2_init()` We can put note/warning separate from the description of  `label`, something to the tune of 
    > Make sure that you use the correct `label` and `inputfile` if you *intend* to read from or write to an existing quantity on disk. If you *intend* to populate a new quantity on disk, ensure that the `label` is not already used in `inputfile`. PSIO::tocprint(int filenum) can be used to display the current list of labels in use in the outputfile. 


BTW: That last hint is particularly useful for debugging I/O related errors.

Also, If you intend … better than it will be loaded… since it indicates this may or may not be done, and it does **not** happen now. 

The case where an tensor is held in core via the cache system does not need to be mentioned since it does not change any of this behavior.

To go on to your question about the inputfile, label relationship. The file is one of those defined in [psifiles.h](https://github.com/psi4/psi4/blob/master/psi4/include/psi4/psifiles.h), and these macros provide an easy way to know which number file a quantity belongs in. Thinking of the file as a dictionary is not so far from reality, and will serve for the purpose of explaining. The label is going to be used to lookup the offset into the file (number of  bytes from the beginning) where some numbers (the elements of the tensor) are found. If writing it will replace these, if reading they used in any operations involving the buffer we are using to manage all of this.

amjames:
Since this has not been updated I added the clarification to the comments and opened a new PR  #573 , this will be closed.  

Thanks for bringing this up @matt-chan. If you have any questions about libdpd or the rest of psi4, I'm always watching our [forums](http://forum.psicode.org/) and would be happy to help!

PR #553, opened 2016-12-17
Update .codecov.yml

stevepeak:
## Description
Per discussion with @dgasmith this pull request fixes the following.

1. Do not ignore python/cobertura reports that have expired
2. Will not wait for CI to complete, nor require it when comparing commits.


dgasmith:
Great, thanks for patching this up!

PR #554, opened 2016-12-19
pybind11, plugins, externals, and targets

loriab:
## Description
Honestly, this started as just updating pybind11 to detect via config. It got out of hand, but to good results.

## Todos
* **Developer Interest**
  - [x] `exception.message` --> `str(exception)` since former sometimes raises an error in the course of handling another.
  - [x] fix python/json testing when PYMOD_INSTALL_LIBDIR is non-default.
  - [x] export psi4::core target. use it in plugin. much simplify the psi4Config. tidy up the share installs.
  - [x] flatten install dir structure. no more `external/`
  - [x] shift importing of official plugins. rename psi4 --plugin* options. add cmake -C line. update CM templates.
  - [x] switch over Py detection to PythonLibsNew. Beautify components printing. 
  - [x] Move threads detection to target.
  - [x] start testing the plugin templates again
  - [x] standardize `Psi4Macros` --> `psi4OptionsTools`.
  - [x] Correct CXX comp names in autocmake_static_lib (@robertodr, I copied this from you, so you may have it wrong, too). Check enabled langs properly rather than compilers defined so OptionsTools usable to other projects with diff enabled langs.
  - [x] switch pybind11 to upstream source. better handle c++ std wrt pybind. `PSI4_CXX_STANDARD` --> `psi4_CXX_STANDARD`.
  - [x] compile test of gcc version able to account for -gcc-name/-gxx-name (@bennybp, here's the code for what we talked about once in `custom_cxxstandard.cmake`)
  - [x] convert restrict test to write its own file.
  - [x] kill off LAB and @jturney config files. replace with `cmake -C psi4PluginCache.cmake` suited config file.
  - [x] `psi4pasture` --> `pasture`.
  - [x] always whole-archive dpd and qt
  - [x] `external` --> `external/upstream`, `internals` --> ``external/downstream`
  - [x] remove old Find*cmake files

  - [x] We changed our python detection module to the one used by NumPy. Also changed so that not actually using libpython anywhere.
  - [x] Used to do `psi4 --new-plugin-makefile` to generate `Makefile`. Now do `psi4 --plugin-compile` to get a cmake command to generate a `Makefile` against that psi4.
  - [x] `--new-plugin` --> `--plugin-name`.
  - [x] `ctest -L mini` --> `ctest -L smoke`
 
- [x] @amjames, the pasture is going to need a slight update to mesh with the new plugin infrastructure.

## Status
- [x]  Ready to go


psi4/share/psi4/plugin/aointegrals/CMakeLists.txt.template
```diff
@@ -27,16 +27,12 @@
...
-set(CMAKE_CXX_COMPILER @CXX@)
-set(CMAKE_Fortran_COMPILER @Fortran@)
+cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 
 project(@plugin@ CXX)
 
 find_package(psi4 1.0 REQUIRED)
 
+include(custom_cxxstandard)
+
```
jturney: Can this include be moved into psi4Config.cmake?

cmake/autocmake_static_library.cmake
```diff
@@ -42,13 +43,13 @@ if(ENABLE_GENERIC)
     endif()
 
     if(DEFINED CMAKE_CXX_COMPILER_ID)
-        if(CMAKE_C_COMPILER_ID MATCHES GNU)
+        if(CMAKE_CXX_COMPILER_ID MATCHES GNU)
```
robertodr: @loriab Thanks for the heads up. It's already fixed in PCMSolver. I would suggest to remove the `autocmake_` prefix from the filename, since this file is not part of Autocmake.

psi4/share/psi4/plugin/aointegrals/CMakeLists.txt.template
loriab: Yes, I think it can go into the `add_psi4_plugin` function. I rejected the idea once, but I think that was before the module was properly checking internal gcc version (as opposed to `gcc --version`).

external/downstream/pasture/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
-if(ENABLE_psi4pasture)
-  ExternalProject_Add(psi4pasture
-    DEPENDS psi4_core
+if(ENABLE_pasture)
+  ExternalProject_Add(pasture
+    DEPENDS psi4-core
     GIT_REPOSITORY https://github.com/psi4/psi4pasture
     CMAKE_ARGS -DENABLE_ccsort=${ENABLE_ccsort}
                -DENABLE_transqt2=${ENABLE_transqt2}
```
amjames: It looks like we will also need to add `-Dpybind11_DIR=${STAGED_INSTALL_PREFIX}/share/cmake/pybind11` to the `CMAKE_ARGS` section or else cmake chokes on `find_package(psi4 1.0 REQUIRED)` due to [this](https://github.com/psi4/psi4/pull/554/files#diff-720577ced4653fc45bc393ee4df2dc42R109) line in the psi4config.cmake

loriab: Are you running against a 554-patched psi4? I killed off the `<INSTALL_PREFIX>/external` dir (collapsed to install alongside psi4 itself) so that that wouldn't happen. Or using a pre-built pybind11 by chance? Because the line you suggested isn't necessary valid if pybind11 is detected external.

amjames: Okay, I thought I was, but I was not. I am checking this again (and I am sure I am using this patch now)

amjames: This still seems to be an issue, I take back my first comment though you are 100% right using an external pybind11 would make that argument invalid. 

I think the best way to go would to be passing `-Dpybind11_DIR=${pybind11_DIR}` to the external project. If an external pybind is used that variable is how it would be communicated to the build, and if we are building pybind internally that variable is set after the external project call for [pybind](https://github.com/psi4/psi4/pull/554/files?diff=unified#diff-5e8400e93d83200f7e000603ba1d3cc7R20)

loriab: You're absolutely right, @amjames. I was thinking of "my way" for plugins (`>>> cmake -C {}/psi4PluginCache.cmake -DCMAKE_PREFIX_PATH={} .`) where the prefix path catches everything. I've made the correction. Just patching up the docs a bit before pushing.

loriab:
After this current PR, I think only [this](https://github.com/psi4/psi4/pull/543/files#diff-3ff5778d1fb0846c66c1c74028288218R77) and [this](https://github.com/psi4/psi4/pull/543/files#diff-3ff5778d1fb0846c66c1c74028288218L91) are left from #543. Sorry I didn't think to include those bits.

dgasmith:
Alright, this is next up I think. Lori can you rebase this up to master and pull in whatever you need from #543.

PR #557, opened 2017-01-06
pybind11 2.0.0 update

loriab:
## Description
Get Psi4 to a tagged version on pybind11 that's likely to be deployed

## Todos
* **Developer Interest**
  - [x] Seeks only >=2.0.0 pybind11. This has the cmake export stuff we need. Note that the current release 2.0.1 has material changes only for Windows.
  - [x] @dgasmith note that I had to add the buffer_protocol flag to your CIVector for runtime functioning. This was part of py11's pypy push. [Details here](https://github.com/pybind/pybind11/commit/1d1f81b278d956445287b3c491125caa11b925bd) if you want to review. I'm a little surprised that neither more nor less was required in the way of changes.
  - [x] `cbs-xtpl-func` broken, but not by this PR.
  - [x] Uses stable 2.0 pybind11 release

## FYI
- @ryanmrichard and @bennybp, shall we agree upon 2.0.0 as our next set point? Not much code adaptation required. Note `pybind11::pybind11` --> `pybind11::module`. There's a one-word patch I'll forward to the pb11 people, but I don't think it'll interfere with you (leftover `::pybind11` in pybind11Config.cmake).

## Status
- [x]  Ready to go

### DIFF
--- a/cmake/psi4OptionsTools.cmake
+++ b/cmake/psi4OptionsTools.cmake
@@ -67,7 +67,7 @@ macro(psi4_add_module binlib libname sources)
     foreach(name_i IN LISTS depend_name)
         target_link_libraries(${libname} PRIVATE ${name_i})
     endforeach()
-    target_link_libraries(${libname} PRIVATE pybind11::pybind11)
+    target_link_libraries(${libname} PRIVATE pybind11::module)
     target_link_libraries(${libname} PRIVATE ${LAPACKBLAS_LIBRARIES})
 endmacro()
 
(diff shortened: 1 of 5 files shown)


dgasmith:
Kinda surprised they went with PyPy over C IR support since the latter is much more useful for JITs in general. This could break a few Python-side GC things, but we can patch those up if they occur.

Still, will be great to have a stable PyBind11 version.

ryanmrichard:
@loriab barring any objection from @bennybp I think locking in version 2.0.0 should be fine with me.

bennybp:
No objection here

PR #567, opened 2017-01-11
Fix a number of memory leaks.

susilehtola:
## Description
This commit fixes a number of memory leaks revealed by cppcheck.

## Status
- [x]  Ready to go


psi4/src/psi4/ccdensity/file_build.cc
```diff
@@ -102,18 +102,33 @@ int file_build(dpdfile4 *File, int inputfile, double tolerance,
               core_left = memoryd - row_length;
 
               /* Make room for another bucket */
-              bucket_offset = (int **) realloc((void *) bucket_offset,
-                                               nbuckets * sizeof(int *));
```
dgasmith: We are a bit hit or miss in catching `alloc`, `realloc`, `calloc`, etc calls. However, this one is particularly large so probably a good idea to do so. For each of these can you include explicit bracers in the if/else calls.

psi4/src/psi4/dcft/dcft_sort_mo_tpdm.cc
```diff
@@ -92,18 +92,35 @@ void DCFTSolver::presort_mo_tpdm_AB()
...
                 coreLeft = memoryd - rowLength;
                 /* Make room for another bucket */
-                bucketOffset = (int **) realloc((void *) bucketOffset,
-                                             nBuckets * sizeof(int *));
+
+		int **p;
+
+		p = static_cast<int **>(realloc(static_cast<void *>(bucketOffset),
+						nBuckets * sizeof(int *)));
+		if(p == NULL)
```
amjames: Should probably get explicit braces in here as well?

psi4/src/psi4/dcft/dcft_sort_mo_tpdm.cc
```diff
@@ -230,18 +247,35 @@ void DCFTSolver::presort_mo_tpdm_AA()
...
```
amjames: Braces in here as well, improves readability.

amjames:
@susilehtola Thanks! this branch just needs a rebase then I think it's ready to go!

PR #571, opened 2017-01-14
Refactoring of Dimension class

fevangelista:
## Description
This PR contains a refactored version of the Dimension class in which data is no longer stored in a `int*` array but on `std::vector<int>`.  This eliminates a potential bug of the old code (memory was potentially freed even for a zero-length vector). After working on this PR I realized that the dependency on `int*` runs pretty deep in Psi4. It affects major libraries (dpd, libtrans, libfock,...). So I still had to leave a way to get a `int*` to the data stored in the class member `std::vector<int>` object.  Things are slightly complicated also by the fact that Wavefunction exposes Dimension objects via functions that return constant references. Some work was done to reduce the number of Dimension->int* casts.

## Todos
* **Developer Interest**
  - [X] Modernize the Dimension class.
  - [X] Fixes a potential problem with memory.
  - [X] Adds exceptions to operations that could potentially involve Dimension objects of different dimensions.
 - [X] Avoids exposing non-const Dimension objects.
 - [X] Refactors MatrixFactory to use Dimension objects.

## Status
- [X]  Ready to go


psi4/src/psi4/ccdensity/get_moinfo.cc
```diff
@@ -71,8 +72,8 @@ void get_moinfo(std::shared_ptr<Wavefunction> wfn)
         moinfo.escf = wfn->reference_wavefunction()->reference_energy();
     else
         moinfo.escf = wfn->reference_energy();
-    moinfo.orbspi = wfn->nmopi();
-    moinfo.openpi = wfn->soccpi();
+    moinfo.orbspi = const_cast<psi::Dimension*>(&wfn->nmopi())->pointer();
+    moinfo.openpi = const_cast<psi::Dimension*>(&wfn->soccpi())->pointer();
     moinfo.clsdpi = init_int_array(moinfo.nirreps);
     for(int h = 0; h < moinfo.nirreps; ++h)
         moinfo.clsdpi[h] = wfn->doccpi()[h];
```
dgasmith: I wouldn't allow de-consting of pointers, instead use the loop below to copy the data and let the user do what they wish. Just as a general philosophy, I think its best to stop supporting behavior that should be changed so that it is not propagated in the future.

psi4/src/psi4/libmints/matrix.h
```diff
@@ -593,6 +593,9 @@ class Matrix : public std::enable_shared_from_this<Matrix> {
     const Dimension& rowspi() const {
         return rowspi_;
     }
+    Dimension& rowspi() {
+        return rowspi_;
+    }
```
dgasmith: As above, I wouldn't allow non-const Dimension objects. A user modifying one of these objects in place could create seriously hard to track down bugs.

psi4/src/psi4/libmints/vector.h
```diff
@@ -204,7 +204,7 @@ class Vector
     int dim(int h = 0) const { return dimpi_[h]; }
 
     /// Returns the dimension array
-    int *dimpi() const { return dimpi_; }
+    int *dimpi() { return dimpi_; }
```
dgasmith: It might be worth switching this over to return a `const Dimension` object instead of the implicit casting here so that the functionality is similar to the Matrix class.

fevangelista:
Thanks for the help with git. I am afraid I still made a little mess with merging the commits. I addressed the issues you raised and once test pass this should be ready to go.

dgasmith:
@fevangelista I like the new changes a lot! Lori fixed up the history somewhat in #574. Thanks for improving this functionality.

jturney:
I like the changes, too!

fevangelista:
Thanks @loriab for the duplicating the PR and @dgasmith @jturney thanks for the positive feedback! In general I think that Matrix/Vector and other infrastructure in libmints could be made leaner and cleaner if we focused on passing only Dimension objects, moved to use stl containers, and modernized the code to take advantage of C++11 features. I'll be interested in discussing more about this with you when we have a chance.

BTW: I went ahead and resolved a conflict both in #571 and #574. Should I go ahead and close this PR so we may merge @loriab's version?

dgasmith:
Closing this out as #574 was merged in. I would say more C++11 modernization is a good thing! Please keep in touch about your future plans.

PR #573, opened 2017-01-15
Clarification in libdpd

amjames:
See discussion in #550.

## Description
Commonly, new users of libdpd will think that DPD::buf4_init() and DPD::file2_init() are initializing blank (zeroed out) tensors to be manipulated.  The note added to the comments above these functions should help some from making this mistake. 

I also removed comments from the file2_init() that describe arguments that do not exist. 
## Status
- [x]  Ready to go


dgasmith:
Looks like were still hitting timeouts occasionally.

PR #579, opened 2017-01-19
Update base compiler version detection with Intel compiler and clang.

jturney:
## Description
According to [clang source](http://llvm.org/svn/llvm-project/cfe/trunk/lib/Frontend/InitPreprocessor.cpp) (about half way down the page), clang always reports version 4.2.1 when using ```__GNUC__```, ```__GNUC_MINOR__```, ```__GNUC_PATCHLEVEL__```. This PR updates the version tester to use ```__clang_major__```, ```__clang_minor__```, ```__clang_patchlevel__``` when clang is detected.

This issue appeared when using Intel compilers on a Mac.

## Status
- [x]  Ready to go

### DIFF
--- a/cmake/custom_cxxstandard.cmake
+++ b/cmake/custom_cxxstandard.cmake
@@ -15,7 +15,11 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
     #include <stdio.h>
     
     int main() {
+        #ifdef __clang__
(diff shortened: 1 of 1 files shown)


jturney:
I'll make the changes that @loriab suggested above. It'll probably be tomorrow before I'm able to commit.

jturney:
Okay, should be ready to go.

PR #582, opened 2017-01-23
Plugin fix

ryanmrichard:
## Description
This factors the code in the `CMakeLists.txt` of the various plugin templates into one function to avoid
copy/paste.  Within this function it then ensures that `CMAKE_PREFIX_PATH` is properly passed to each build in turn allowing the user to build plugins and use externally built Psi4 dependencies.

## Todos
* **Developer Interest**
  - [x] Adding additional plugin templates to the build system is dramatically simplified
  - [x] You can now use external versions of libraries that Psi4 normally compiles with plugins.

## Status
- [x]  Ready to go


ryanmrichard:
I went ahead and squashed the commits so that it wasn't showing the (now) empty math library one.

PR #585, opened 2017-01-25
Added test case for dgasmith's orbital rotation code

bzloink:
## Description
A test case for dgasmith's orbital rotation code using an ROHF reference.

### DIFF
--- a/tests/rohf-orb-rot/input.dat
+++ b/tests/rohf-orb-rot/input.dat
@@ -13,7 +13,6 @@ set {
     e_convergence 10
     d_convergence 10
     r_convergence 10
-    memory 60000
     freeze_core True
     scf_type pk
     docc [7, 1]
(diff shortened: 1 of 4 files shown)


tests/rohf-orb-rot/input.dat
```diff
@@ -0,0 +1,122 @@
...
+
+scf_e, scf_wfn = energy('scf', return_wfn=True)
+ccsd_e, ccsd_wfn = energy('ccsd', return_wfn=True)
+
+compare_values(-150.108136116503147, scf_e, 6, 'X SCF energy') #TEST
+compare_values(-150.347533153677489, ccsd_e, 6, 'X CCSD energy') #TEST
+
+# Rotate orbitals to obtain A state in C1
+# Quick orbital rotation code
+def orb_rotate(C, h, i, j, theta):
```
bzloink: The latter. An orbital rotation function like this has not yet been implemented into psi4.

bzloink: The goal of this test was more instructive than anything - if a user wished to learn how to perform an orbital rotation like this, there would be a reference on how to do so. I understand your rationale, though. Where would something like this example go instead? (Or is the thread concerning this in the forum helpful enough on its own?)

amjames:
Okay, this kind of contribution is certainly welcome I just wasn't sure if we had a place in mind for where this kind of stuff should go. 

Talking with @loriab It appears we do not, so we will probably move this to something like a "cookbook" directory in the future, but for now it is fine where it is.

Would you mind adding in some comments to make it a bit more tutorial-like? Maybe a section at the top with a brief overview (you can put a link to the forum post here as well), and then a section above each calculation is trying to show/do? 

Imagine if you were reading it when you first made your forum post and any comments that would help realize you could use this to get the desired results.

bzloink:
Where would you like these descriptions? On github, in `input.dat`, or both?

loriab:
Just comments in the input.dat, please.

bzloink:
Done and done. Let me know if there's anything else I need to change.

dgasmith:
Can this wait for #592, I added an export for `rotate_columns` on the matrix class so we don't need the extra definition `rotate`.

bzloink:
Sounds good.

dgasmith:
@bzloink Sorry for the long delay. The Matrix class (what the orbitals are) should have the `void rotate_columns(int h, int i, int j, double theta);` signature attached to them so there is no need for a custom rotate function. The main difference is that the angle is now assumed to be in radians.

If you could make the appropriate changes, we can merge this in.

loriab:
Please take a look at [this message](https://github.com/psi4/psi4/pull/644#issuecomment-286609790). I think you've got the same problem.

bzloink:
@loriab Shoot, nooooo. This is exactly what I was trying to avoid. After I'm done recompiling, I'll try your fix on #644.

tests/rohf-orb-rot/input.dat
```diff
@@ -0,0 +1,129 @@
...
+
+   xp = np.cos(theta) * x - np.sin(theta) * y
+   yp = np.sin(theta) * x + np.cos(theta) * y
+
+   C.nph[h][:, i] = xp
+   C.nph[h][:, j] = yp
+
+# Rotate columns here.
+orb_rotate(scf_wfn.Ca(), 0, 7, 8, 90.0)
+
```
loriab: I think this can be switched to use the new `void rotate_columns(int h, int i, int j, double theta);` method on `psi4.Matrix` that @dgasmith mentioned.

bzloink:
Done!

PR #588, opened 2017-01-27
Minor CI and SCF fixes

### LINKED ISSUE #556: CASSCF from DFT reference
Hi,

I'm trying to run some CASSCF calculations for systems where the HF reference is terrible. What I'd like to do is use e.g. BP86 orbitals to run CASSCF. However, when I use

energy('bp86')
set guess read
set maxiter 0
set fail_on_maxiter 0
energy('casscf')

it still runs a single HF iteration even though maxiter=0.

Would it be possible to change the behavior of maxiter so that maxiter=0 only recomputes the energy and orbital eigenvalues, but does not update the orbitals? (Pseudocanonization should be OK.)

### LINKED ISSUE #586: scf guess read fails when molecule is modified
I'm trying to run some FCI calculations without symmetry to ensure I land on the ground state. However, the SCF is sometimes very hard to do without symmetry, so I'd like to run SCF with symmetry, switch symmetry off and check if I'm still at a stable solution.

However, I find I'm unable to change my molecule for symmetry constraints (or charge/multiplicity for that matter). The following input
```
molecule sym {
0 1
H  0.459619  0.000000 0.000000
H  0.000000  0.459619 0.000000
}

set basis STO-3G
energy('scf')

molecule nosym {
0 1
H  0.459619  0.000000 0.000000
H  0.000000  0.459619 0.000000
}

set scf guess read
energy('scf')
```
runs the first SCF, but crashes on the second with
```
RuntimeError: 
Fatal Error:   SCF guess was set to SAD, but sad_basissets_ was empty!


Error occurred in file: /home/work/psi4/psi4/psi4/src/psi4/libscf_solver/sad.cc on line: 659

psi::scf::HF::compute_SAD_guess()
psi::scf::HF::guess()
psi::scf::HF::initialize()
psi::scf::HF::compute_energy()
```
Oddly, it appear that while in the second run
```
  ==> Algorithm <==

  SCF Algorithm Type is DF.
  DIIS enabled.
  MOM disabled.
  Fractional occupation disabled.
  Guess Type is READ.
  Energy threshold   = 1.00e-06
  Density threshold  = 1.00e-06
  Integral threshold = 0.00e+00
```
the initialization uses
```
  ==> Integral Setup <==

  ==> DFJK: Density-Fitted J/K Matrices <==

    J tasked:                  Yes
    K tasked:                  Yes
    wK tasked:                  No
    OpenMP threads:              1
    Integrals threads:           1
    Memory (MB):               375
    Algorithm:                Core
    Integral Cache:           NONE
    Schwarz Cutoff:          1E-12
    Fitting Condition:       1E-12

   => Auxiliary Basis Set <=

  Basis Set: file /home/work/psi4/install/share/psi4/basis/def2-svp-jkfit.gbs
    Number of shells: 12
    Number of basis function: 36
    Number of Cartesian functions: 40
    Spherical Harmonics?: true
    Max angular momentum: 2

  Minimum eigenvalue in the overlap matrix is 2.8114045453E-01.

```
followed by the crash.

### LINKED ISSUE #590: CI natural orbitals occupation numbers not printed out with nat_orbs=true
Somewhat related to issue 556 and pull 588, if you set nat_orbs to true and run a CI calculation, you don't get the natural orbital occupations in the output, whereas if you're running CASSCF you get a nice printout. 

I'm guessing the NOONs should get printed out, since the output file does say
```
   Computing CI Natural Orbitals
   !Warning: New orbitals will be sorted by occuption number,
   orbital spaces (occ/act/vir) may change.
```

Example input
```
molecule {
0 1
symmetry c1
C
C 1 1.0
}

set basis cc-pVDZ
set scf_type pk
set nat_orbs true
set wfn detci
set restricted_docc [2]

# Run CISDT
cas_e, cas_wfn = energy('cisdt', return_wfn=True)
```

dgasmith:
This grew a bit to encompass a few more issues, its still a general bug fix patch.

## Description
Fixes and adds test for two issues that were identified:
 - Fixes #586 where the SAD basis was not being set when READ failed.
 - Fixes #556 as the NO option was not tracked to the new MCSCF driver.
 - CI OPDM/TPDM now correctly closes files when complete.
 - MCSCF Rotate matrices are now built continuously to avoid difficult cases to exponentiate.
 - Patched up Dvec reading between successive CI calls (especially for MCSCF).
 - PK Non-symmetric density matrices should now always be detected correction.
 - Makes CI printing to be more canonical.
 - Prints occupation number for CI, fixes #590

## Status
- [x]  Ready to go


susilehtola:
```
molecule {
0 1
symmetry c1
C
C 1 1.0
}

set basis cc-pVDZ
set scf_type pk
set nat_orbs true
set restricted_docc [2]
set qc_module detci

# Initial guess from CISD natural orbitals
ci_e, ci_wfn = energy('cisd', return_wfn=True)

# Run CAS(10,10)
set active [10]
cas_e, cas_wfn = energy('casscf', return_wfn=True, ref_wfn=ci_wfn)
```

crashes with 90cbafb2ad1e7a0a70780b26e5caa15c1ed02a60 applied. Without qc_module = detci, the CASSCF initial energy is the same as with HF, I guess the CISD module doesn't have NOs implemented?

susilehtola:
The error I get is
```
   ==> Starting MCSCF iterations <==

        Iter         Total Energy       Delta E   Orb RMS    CI RMS  NCI NORB

  File "/home/work/psi4/install/bin/psi4", line 223, in <module>
    exec(content)
  File "<string>", line 35, in <module>
  File "/home/work/psi4/install/lib/psi4/driver/driver.py", line 447, in energy
  File "/home/work/psi4/install/lib/psi4/driver/procedures/proc.py", line 4025, in run_detcas
    ciwfn = mcscf.mcscf_solver(ref_wfn)
  File "/home/work/psi4/install/lib/psi4/driver/procedures/mcscf/mcscf_solver.py", line 153, in mcscf_solver
    nci_iter = ciwfn.diag_h(abs(ediff) * 1.e-2, orb_grad_rms * 1.e-3)

RuntimeError: 
Fatal Error: PSIO Error
Error occurred in file: /home/work/psi4/psi4/psi4/src/psi4/libpsio/error.cc on line: 128

psi::psio_error(unsigned int, unsigned int)
psi::PSIO::read(unsigned int, char const*, char*, unsigned long, psi::psio_address, psi::psio_address*)
psi::psio_read(unsigned int, char const*, char*, unsigned long, psi::psio_address, psi::psio_address*)
psi::psio_read_entry(unsigned int, char const*, char*, unsigned long)
```

dgasmith:
@susilehtola Yup, got it. With a test too.

psi4/src/psi4/libfock/PKmanagers.cc
```diff
@@ -370,6 +370,7 @@ void PKManager::form_D_vec(std::vector<SharedMatrix> D, std::vector<SharedMatrix
 
     D_ = D;
     all_sym_ = true;
+    symmetric_.clear();
```
dgasmith: @jgonthier I think this was the source of the symmetric issues. The vector wasnt being cleared whenever PK was being called so it reach many more elements than density matrices. This fixes up local test cases and the issue I was having, but please confirm.

dgasmith:
@susilehtola If you would, please give the current PR a look over and see if it matches your needs.

susilehtola:
The NOONs are now printed out, and the example calculation above runs nicely with cisd, cisdt, and cisdtq guesses. lgtm

susilehtola:
One more tiny bug:
```
molecule {
0 2      
H 0.0 0.0 0.000000
H 0.0 0.0 0.500000
H 0.0 0.0 1.000000
}

set basis STO-3G
set reference rohf
set ci_maxiter 100
set scf_type pk
set fail_on_maxiter false
set nat_orbs true
ci_e, ci_wfn=energy('fci', return_wfn=True)

# Rerun to get true solution
set num_roots 3
set e_convergence 1e-10
set r_convergence 1e-8
energy('fci', ref_wfn=ci_wfn)
```
prints out
```
$ psi4
(H0block_setup): Can't find CI block!
(H0block_setup): Can't find CI block!
```

dgasmith:
The CI program is trying to tell you something is going wrong. Maybe @CDSherrill can elucidate.

susilehtola:
Sure, I'm just wondering whether this should really be in stdout or the log file?

psi4/src/psi4/libfock/PKmanagers.cc
jgonthier: @dgasmith Sorry for my late answer. Yes, it was a problem that this vector was not being cleared. Thanks for the correction !

PR #594, opened 2017-01-30
PCMSolver

loriab:
## Description
Try to get PCMSolver reintegrated.

## Todos
* **Developer Interest**
  - [x] This is atop lapack and ambit PR's, so you'll want to look only at 410be6f
  - [x] Switched to PCMSolverConfig imported target.
  - [x] Based on https://github.com/loriab/pcmsolver.git branch config6
  - [x] target `PCMSolver::PCMSolver` --> `PCMSolver::pcm`
  - [x] no longer using pcm_placeholder, instead add to path in run_psi4.py
  - [x] I'd rather have pcmsolver as a proper module containing __init__, pcmsolver, getkw, etc., but that would mess with other codes use of pcmsolver, I suspect
  - [ ] This runs into trouble in libpsipcm, where the signatures don't match up. What's missing is the dreaded HostWriter at the end. I daresay that's a change upstream that @robertodr understands.

```
/theoryfs2/ds/cdsgroup/psi4-compile/hrw-temp/psi4/src/psi4/libpsipcm/psipcm.cc(133): error: argument of type "void (*)(const char *, int)" is incompatible with parameter of type "HostWriter"
                               symmetry_info, &host_input, &host_writer);
                                                           ^

/theoryfs2/ds/cdsgroup/psi4-compile/hrw-temp/psi4/src/psi4/libpsipcm/psipcm.cc(136): error #165: too few arguments in function call
                               symmetry_info, &host_input);
                                                         ^
```

## Questions
- [ ]  @robertodr, you'll take a look at this? I can certainly answer any "why" questions.

## Status
- [ ]  Ready to go


external/upstream/pcmsolver/CMakeLists.txt
```diff
@@ -29,45 +30,36 @@ if(${ENABLE_PCMSolver})
                        -DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
                        -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
                        -DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}
+                       -DPYMOD_INSTALL_LIBDIR=${PYMOD_INSTALL_LIBDIR}
                        -DCMAKE_INSTALL_DATADIR=${CMAKE_INSTALL_DATADIR}
                        -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                       -DENABLE_psi4=ON
```
robertodr: I am not sure I want to introduce host-dependent flags. See also [here](https://github.com/PCMSolver/pcmsolver/commit/78dc78d20dd7f178bcd916ac450f865c48d08368#diff-a94fc1ec48247bd2a42eed10a9f07894)

external/upstream/pcmsolver/CMakeLists.txt
```diff
@@ -29,45 +30,36 @@ if(${ENABLE_PCMSolver})
...
-#                  -DENABLE_64BIT_INTEGERS=${ENABLE_64BIT_INTEGERS}
+                       -DENABLE_CXX11_SUPPORT=ON
+                       -DENABLE_64BIT_INTEGERS=ON
 #                  -DENABLE_TESTS=OFF
 #                  -DENABLE_LOGGER=OFF
 #                  -DENABLE_TIMER=OFF
 #                  -DBUILD_STANDALONE=OFF
 #                  -DENABLE_FORTRAN_API=OFF
 #                  -DSTATIC_LIBRARY_ONLY=${ENABLE_STATIC}
 #                  -DPYTHON_INTERPRETER=${PYTHON_EXECUTABLE}
```
robertodr: Why isn't the `PYTHON_INTERPRETER` passed down anymore? Am I missing something obvious?

external/upstream/pcmsolver/CMakeLists.txt
```diff
@@ -29,45 +30,36 @@ if(${ENABLE_PCMSolver})
                        -DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
                        -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
                        -DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}
+                       -DPYMOD_INSTALL_LIBDIR=${PYMOD_INSTALL_LIBDIR}
                        -DCMAKE_INSTALL_DATADIR=${CMAKE_INSTALL_DATADIR}
                        -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                       -DENABLE_psi4=ON
                        -DSTATIC_LIBRARY_ONLY=${_a_only}
+                       -DSHARED_LIBRARY_ONLY=${_so_only}
 #                       -DENABLE_OPENMP=???
```
robertodr: I currently have no OpenMP parallelization in PCMSolver, nor have I enabled the one that comes "for free" with Eigen. For that option, you can pass the value set when configuring Psi4, it won't hurt (that I know of)

external/upstream/pcmsolver/CMakeLists.txt
```diff
@@ -29,45 +30,36 @@ if(${ENABLE_PCMSolver})
...
                        #-DENABLE_XHOST=???  # option missing
```
robertodr: I have no opinion on this. If Psi4 sets it, then it can be set to the same value here, I think

psi4/src/psi4/libscf_solver/CMakeLists.txt
```diff
@@ -9,10 +9,11 @@ set(sources_list uhf.cc
                  stability.cc
 )
 psi4_add_module(lib scf_solver sources_list mints diis efp_solver fock)
-if(TARGET PCMSolver::PCMSolver)
+if(TARGET PCMSolver::pcm)
```
robertodr: I had to change the way the output from PCMSolver is handed over to the host. Instead of a free function that the API expects to be defined somewhere in the host, the host needs to pass this function explicitly to the module, i.e. a function pointer. I will fix this problem in your fork, the fix should do away with the compilation error you report **and** the need for the workaround in this file.

external/upstream/pcmsolver/CMakeLists.txt
loriab: Yeah, I've been ignoring OpenMP for Fortran projects. But read [this new info](https://github.com/psi4/erd/pull/5) from Ben Pritchard today and see what you think.

psi4/src/psi4/libscf_solver/CMakeLists.txt
loriab: Forks of forks can get a little crazy. I gave you write access to my fork, so do whichever you choose– push to my fork or use the web interface or whatever's most convenient. And thanks, I'm glad not to deal with HostWriter ever again.

external/upstream/pcmsolver/CMakeLists.txt
loriab: We absolutely can pass through the [python trio](https://github.com/psi4/psi4/blob/master/CMakeLists.txt#L39-L41) (I'm not yet tempted to make a target out of them). Python just fell in the class of haven't-needed-to-deal-with-yet among the other external projects.

By the way, have you considered switching to [FindPythonLibsNew](https://github.com/psi4/psi4/blob/master/cmake/FindPythonLibsNew.cmake) which is reputed to be far more reliable on matching interp/libs versions? It's the python detection used by NumPy. Psi4's is a slightly modified version by pybind11. It's only one find_package() call since FindPythonInterp called internally.

And FYI, Psi4's not using the python library anymore. We only detect the library to get at the python header files that are needed by pybind11.

external/upstream/pcmsolver/CMakeLists.txt
loriab: Yes, this enables one to toggle the a/so of the whole set of external projects at once (and if only one of a/so is built, then that's the one that gets detected). When CheMPS2 was in packaging by one of the Linux distros, it was requested that a shared-only option be added since the distro forbade static libs. So hopefully this is doing PCMSolver a favor in the end.

external/upstream/pcmsolver/CMakeLists.txt
loriab: done on the PCMSolver PR side. I'll let you make whatever HostWriter changes you want to this PR before editing it again and patching up this and the other things you mention below.

loriab:
>>> I apologize for being overly slow on the promised PCMSolver CMake overhaul.

No problem. By the time I got Lapack, CheMPS2, libefp, and Psi4 back on speaking terms, it seemed easiest to just do a decent pass at the psi4 interface with ambit and PCMSolver, then let you and Jet concentrate on the parts you know better than me.

>>> What do you mean by "I'd rather have pcmsolver as a proper module containing init, pcmsolver, getkw, etc., but that would mess with other codes use of pcmsolver, I suspect"?

I meant that it'd be nice to have an installation dir structure like the below where there's an outer pcmsolver module/namespace.
```
here/pcmsolver/
here/pcmsolver/__init__.py
here/pcmsolver/pcmsolver.py
here/pcmsolver/getkw.py
...
```
Then `sys.path('here')` is the operation, and psi4 does `import pcmsolver; pcmsolver.pcmsolver.parse_pcm_input('pcmsolver.inp')` or whatever and all the getkw and codata are within the pcmsolver namespace. At present, one does `sys.path('here/pcmsolver')` and everything's in the outer namespace. The anticipated problem with other codes is that what I want would probably involve some `from . import getkw` and other codes might be depending on the `import getkw` structure.

>>> I can fix the linking problems (see comment on psi4/src/psi4/libscf_solver/CMakeLists.txt) Should I fork you fork? Should I make the changes from the web-interface on the pcm branch of your fork?

As mentioned above, direct commit, web interface, whatever's most convenient.

external/upstream/pcmsolver/CMakeLists.txt
robertodr: If I interpret @bennybp's comment correctly, then passing down the value of `${ENABLE_OPENMP}` should work.

external/upstream/pcmsolver/CMakeLists.txt
robertodr: Better explicit than implicit. I think I was just being lazy when I based the static/shared logic on one single toggle.

external/upstream/pcmsolver/CMakeLists.txt
robertodr: I only need the `PYTHON_INTERPRETER` to configure some of the '*.py` files, nothing more.
I had considered FindPythonLibsNew, but FindPythonInterp already does a good job at just finding the interpreter.

robertodr:
OK, I see _viz_ "pcmsolver as proper module" I'll stack it on my TODO list. And it should be no problem for other projects, which only access `pcmsolver.py` anyway.

psi4/src/psi4/libscf_solver/CMakeLists.txt
robertodr: I made a bunch of commits on the web interface and accidentally clogged both Travis and Distelli...

robertodr:
The Travis builds timed out... Would it be possible for you to restart just the last one?

dgasmith:
Please email me your Distelli account info.

loriab:
I just added @robertodr and @amjames on Distelli.

loriab:
@robertodr , that seeming GCC version symbol error you reported earlier somehow came about because [psipcm was unlinked](https://github.com/psi4/psi4/pull/594/commits/f89d42b517564f4f46918488554ef889d10b9a16#diff-938e23f9a1850388684b9681e0db1a4aL14).  I'll push a correction shortly.

This branch (pcm) seems to have gotten commits from upstream master added not at the base. Abandoning it and pushing a new branch and PR, this one still atop lapack branch but not atop ambit.

PR #599, opened 2017-01-31
PCMSolver, take #2

loriab:
## Description
Try to get PCMSolver reintegrated.

## Todos
* **Developer Interest**

  - [x] Fix some HDF5 detection problems
  - [x] Adds some colored "MKL" and "AM" to CMake output for lapack and libint, respectively, for ready diagnosis.
  - [x] Fixes rpath on Mac by setting it to `@loader_path/${_rel}"`
  - [x] Some host_writer changes to modernize PCMSolver
 - [x] Re-enables PCMSolver via `-DENABLE_PCMSolver`. No longer hint with `PCMSOLVER_ROOT` but with `PCMSolver_DIR`. Pins to `v1.1.9`.
  - [x] Bumps CMake requirement to 3.2
  - [x] upstream PCMSolver needs py3 compatibility, PCMSolver/pcmsolver#39
  - [x] anyone building with Intel and native gcc < 4.9 needs to have the gcc-name and gxx-name in `CMAKE_Fortran_FLAGS`, too, just like `CMAKE_CXX_FLAGS`

## Status
- [x]  Ready to go


robertodr:
Can you grant me write access to your fork of PCMSolver?
1. I already use the OpenMP Autocmake module, just not its latest version, which I will update.
2. Testing for Python 3 support will be easier starting from your current version than the mess I currently have on my development branch... :disappointed:

robertodr:
I have just noticed that you have `-DENABLE_64BIT_INTEGERS=ON` This is only necessary when compiling PCMSolver within Fortran projects that use 64 bit integers. I'll disable it on the `pcm2` branch.

external/upstream/pcmsolver/CMakeLists.txt
```diff
@@ -1,26 +1,27 @@
...
-        #get_property(_loc TARGET PCMSolver::pcm PROPERTY LOCATION)  # todo
-        get_property(_loc TARGET PCMSolver::PCMSolver PROPERTY LOCATION)
+        get_property(_loc TARGET PCMSolver::pcm PROPERTY LOCATION)
         message(STATUS "${Cyan}Found PCMSolver${ColourReset}: ${_loc} (found version ${PCMSolver_VERSION})")
         add_library(pcmsolver_external INTERFACE)  # dummy
     else()
         include(ExternalProject)
         message(STATUS "Suitable PCMSolver could not be located, ${Magenta}Building PCMSolver${ColourReset} instead.")
 
         if(${BUILD_SHARED_LIBS})
```
robertodr: What is the suggested way of building the add-ons? As static or as shared libraries? The former **does not** currently work with PCMSolver due to PCMSolver/pcmsolver#34 (yes, I am a moron...)

loriab: The AddOns currently build so/a according to the value of `BUILD_SHARED_LIBS`, which I usually set as ON, but the default is OFF. Fortunately for you, the one that I _must_ have working is the shared. If static is no good with PCMSolver, I'd comment out [this](https://github.com/loriab/psi4/blob/25e2e4347953ae027b0176059cffdf8637b63705/external/upstream/pcmsolver/CMakeLists.txt#L37-L38) and just add a new line `SHARED_LIBRARY_ONLY=ON` as a stopgap.

loriab:
@robertodr 
Thanks on looking at Py3 and updating the autocmake omp. And go ahead and set `ENABLE_64BIT_INTEGERS` and anything else to sensible values.

You're correct that Distelli and Travis are testing _only_ Psi4, not any addons. In fact, the addons it does test (libint, libefp, gdma) aren't getting built either. Rather (for Travis at least), it's downloaded prebuilt conda packages and just doing integration testing. PCMSolver compiles _really_ slowly (at least Release with Intel compilers), and we can't afford that. But once PCMSolver is up and running and packaged again, it can be added to the integration tests. We can certainly afford a fast test case or two.

external/upstream/pcmsolver/CMakeLists.txt
robertodr: Yes, I'll follow your suggestion. I can build, link and successfully execute tests with `BUILD_SHARED_LIBS=ON`

robertodr:
Sounds like a great plan. I think I am almost done with Py3 compatibility.

loriab:
@amjames, sure you pulled `loriab:pcm2`, not `loriab:pcm`? Check that [these lines](https://github.com/loriab/psi4/blob/pcm2/psi4/CMakeLists.txt#L104-L109) are matching. It looks like it's still using the cmake/FindPCMSolver `MODULE` mode, not the new `CONFIG` mode.

amjames:
@loriab You are right. When I cloned the branch I forgot the `-b `so I just cloned your fork's master into a directory called pcm2 (I'm a dingus). This works fine, sorry.

loriab:
No problem, @amjames. PCM involves two PRs on two repos both being worked on by two people and has been through some non-trivial rebases. So I'm grateful to you for field-testing it.

psi4/CMakeLists.txt
```diff
@@ -26,20 +26,20 @@ get_property(_ill TARGET tgt::lapk PROPERTY INTERFACE_LINK_LIBRARIES)
...
 if(ENABLE_Ambit OR ENABLE_CheMPS2)
     find_package(TargetHDF5 CONFIG REQUIRED)
     get_property(_ill TARGET tgt::hdf5 PROPERTY INTERFACE_LINK_LIBRARIES)
     list(GET _ill 0 _ill0)
     list(APPEND _addons ${_ill})
-    message(STATUS "${Cyan}Found HDF5${ColourReset}: ${_ill0};... (found version ${${PN}_VERSION})")
+    message(STATUS "${Cyan}Using HDF5${ColourReset}: ${_ill0};... (found version ${${PN}_VERSION})")
 else()
-    message(STATUS "Unfound HDF5")
+    message(STATUS "Disabled HDF5")
```
dgasmith: Cool :)

psi4/run_psi4.py
```diff
@@ -133,6 +133,9 @@
     data_dir = os.path.abspath(os.path.expanduser(args["psidatadir"]))
     os.environ["PSIDATADIR"] = data_dir
 
+if "@ENABLE_PCMSolver@" in ["ON", "YES", "TRUE"]:
+    sys.path.insert(1, "@PCMSolver_PYMOD@")
+
```
dgasmith: Any reason for the three?

loriab: Hmm, didn't actually test it except with `ON`. And you're right that it could use more case-testing. Those are the most common cmake true values. Just looked it up and should test  `1, ON, YES, TRUE, Y`, it seems. I'll look into it. But the current works for the most common and conda cases.

```
if "ON" in ["ON", "YES", "TRUE"]:
    sys.path.insert(1, "/theoryfs2/ds/cdsgroup/miniconda/envs/text35/lib/python3.5/site-packages/pcmsolver")
```

loriab: Thorough test for above queued for next PR. And after https://github.com/PCMSolver/pcmsolver/issues/40, I'll see about getting PCMSolver working with API mode.

PR #604, opened 2017-02-02
add intel RTL Fortran threading initialization/finalization

loriab:
## Description
@bennybp pointed out [here](https://github.com/psi4/erd/pull/5) that we need to be paying attention to thread safety in called libraries, even if they don't have OpenMP, because Psi4 does. This is trying to follow the guidance in [the venerable document (search `ifcoremt`)](https://software.intel.com/en-us/articles/threading-fortran-applications-for-parallel-performance-on-multi-core-systems)

## Todos
* **Developer Interest**
  - [x] Adds rtl_init and rtl_finalize calls. According to the sample code below, also supposed to set reentrancy, but I don't know what to do about that.
  - [x] Drops clang-3.9 travis test that's a little slow. I'm told that clang 3.9 is the same as Mac/Distelli is running anyways.

## Questions
- [x]  I don't know much about this, but it appears harmless, at least for Intel compilers. Going to let Travis test other build systems. Any thoughts?

## Status
- [x]  Ready to go

```
/*
 * Multithreaded C-Fortran mixed language test program
 */

#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <pthread.h>
#include <errno.h>
#include "threadpool.h"

#include "forreent.h"

#define NUM_THREADS 12
#define NUM_JOBS 10

/* definitions for Fortran Runtime library init/finalize */
extern void for_rtl_init_ (int *, char **);
extern int for_rtl_finish_ ( );
extern int for_set_reentrancy (int *);

/* declare the external Fortran function 'fpa' */
extern void fpa(int io, char *filename);  

void job_thread(void *arg1, void *arg2);

int main(int argc, char **argv)
{
    threadpool tp;
    int i;
    int *reent ;
    int fstat;

    char filenames[NUM_JOBS][100];
    int io[NUM_JOBS];

    int io_status;
    /* Call the Fortran Runtime library for initialization */
    for_rtl_init_ (&argc, argv);
    *reent = FOR_K_REENTRANCY_THREADED;
    fstat = for_set_reentrancy( reent );
    printf ("fstat %i \n", fstat);

    /* Create a thread pool of 12 threads */
    tp = create_threadpool(NUM_THREADS);

    for (i = 0; i < NUM_JOBS; i++)
    {
        /* Create a unique file name i.e. "./Files/file1.txt" */
        sprintf(&filenames[i][0], "/home/spd/rwgreen/quad/rwgreen/forums/81088/file%d.txt", i+1);

        /* Create a unique file I/O unit */
        io[i] = 10+i;

        /* Add file name and I/O unit to threadpool job queue */
        printf("Main: queueing %s\n", &filenames[i][0]);
        dispatch_job(tp, job_thread, (void *) &filenames[i][0], &io[i]);
    }

    /* Wait a while for all the threads to do their thing */
    printf("Main: sleeping 20 secs...\n");
    sleep(20);
    printf("Main: waking up and destroying threadpool...\n");

    /* Clean up threadpool */
    destroy_threadpool(tp);

    /* close the Fortran Runtime */
    io_status = for_rtl_finish_ () ;

    printf("Main: exiting...\n");
    return 0;
}

/* Thread job function */
void job_thread(void *arg1, void *arg2)
{
    char *filename = (char *) arg1;
    int *io = (int *) arg2;
    unsigned int len = strlen(filename);

    char string[len+1];
    sprintf(string, "%s", filename);

    fpa(*io, string); 

    printf("->Thread %ld: Opened %s (len=%d) on I/O unit %d\n", pthread_self(), filename, len, *io);
}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,27 +36,27 @@ matrix:
       - NAME='clang'
       - VERSION='3.6'
 
-  - os: linux
-    compiler: clang
-    addons: &2
-      apt:
-        sources:
-        - llvm-toolchain-precise-3.9
-        - ubuntu-toolchain-r-test
-        - george-edison55-precise-backports
-        packages:
-        - liblapack-dev
-        - clang-3.9
-        - libhdf5-serial-dev
-        - gfortran
-    env:
-      - CXX_COMPILER='clang++-3.9'
-      - PYTHON_VER='3.5'
-      - C_COMPILER='clang-3.9'
-      - Fortran_COMPILER='gfortran'
-      - BUILD_TYPE='release'
-      - NAME='clang'
-      - VERSION='3.9'
+  #- os: linux
+  #  compiler: clang
+  #  addons: &2
+  #    apt:
+  #      sources:
+  #      - llvm-toolchain-precise-3.9
+  #      - ubuntu-toolchain-r-test
+  #      - george-edison55-precise-backports
+  #      packages:
+  #      - liblapack-dev
+  #      - clang-3.9
+  #      - libhdf5-serial-dev
+  #      - gfortran
+  #  env:
+  #    - CXX_COMPILER='clang++-3.9'
+  #    - PYTHON_VER='3.5'
+  #    - C_COMPILER='clang-3.9'
+  #    - Fortran_COMPILER='gfortran'
+  #    - BUILD_TYPE='release'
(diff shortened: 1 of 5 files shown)


robertodr:
@bast FYI

Issue #606, opened 2017-02-03
segfault in cisd calculation


susilehtola:
```
molecule {
0 1
symmetry c1
Mn -0.0128264866 -0.0002779912 -0.1091894544
O  -1.1037647823  1.0246639195 -1.2821459116
O  -0.4319309790 -1.5910632792 -1.0423314985
N   0.7243918073  1.6205280457  0.7709992473
N   1.3008767173 -0.9287768925  1.0635839133
C  -1.2886186814  2.2834208702 -1.2181338272
C  -0.7050865946  3.1821062059 -0.3604271375
C   0.3534920334  2.8195308082  0.5207059889
C   0.2177419816 -2.6945497951 -0.9662726987
C   1.2181378357 -3.0250490115 -0.0961995911
C   1.7180430400 -2.1229929449  0.8899899444
C   1.9393113446  1.3525171542  1.5367699106
C   1.8333722479 -0.0604962603  2.1055467704
H   0.8908035853  3.6241374839  1.0039698993
H   2.5024254412 -2.4854165868  1.5399007635
H   2.7858100041  1.4098930461  0.8661311665
H   1.1490336627 -0.0652673733  2.9502674019
H   2.0638511115  2.0769335299  2.3353274192
H   2.7993768711 -0.4085469833  2.4534782102
H  -1.9892347032  2.6686747336 -1.9454961675
H  -0.0986239324 -3.4441353556 -1.6764363233
Cl  1.8821942342  0.3098852687 -1.6599082362
O  -1.1395010178 -0.1703680057  0.9263551082
H  -0.9312248148  4.2251864392 -0.4696082573
H   1.6695378989 -3.9953914470 -0.1633815245
}

# Run SCF, initialize in smaller basis
set basis 6-31G*
set basis_guess 3-21G
set scf_type pk
set soscf true
scf_e, scf_wfn = energy('scf', return_wfn=True)

# Get CISD active space guess. 69 occupied orbitals, out of which 32
# are core electrons. Leaves No=37 orbitals, excitable into Nv=204
# virtuals. Let's take only the first 50 virtuals
set restricted_docc [32]
set restricted_uocc [154]
set qc_module detci
set nat_orbs true
ci_e, ci_wfn = energy('cisd', ref_wfn=scf_wfn, return_wfn=True)
```
segfaults after the 20 most important determinants have been written to output.

Backtrace gives
```
#0  __GI___libc_free (mem=mem@entry=0x550020205841) at malloc.c:2963
[...]

PR #607, opened 2017-02-06
Safer memory operations

susilehtola:
## Description
This PR 
(a) adds checks that alloc's have been succesful in optking
(b) rewrites detci's print_config routine to use dynamic memory allocation instead of static arrays, which I believe caused issue #606.

## Status
- [x]  Ready to go

### DIFF
--- a/psi4/src/psi4/detci/ciwave.h
+++ b/psi4/src/psi4/detci/ciwave.h
@@ -395,9 +395,8 @@ private:
     void H0block_filter_setup(void);
     void H0block_fill(void);
     void H0block_coupling_calc(double E);
-    void print_config(int nbf, int num_alp_el, int num_bet_el,
-       struct stringwr *stralp, struct stringwr *strbet, int num_drc_orbs,
-       char *outstring);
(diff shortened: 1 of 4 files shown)


susilehtola:
I opted to use std::string and ostringstream instead of just increasing the size of the static arrays, because this is not bottleneck code. I'm not sure why they were used in the original code, but that seemed to be over 20 years old, i.e. originally C code?

loriab:
Thanks, `char *` --> `std::string` always welcome.

dgasmith:
I looked at the original error and thought that the debugger was throwing on the wrong line, so this is a really great catch!

Yes, most of the CI code was overhauled last year; however, most parts that deal with determinants have not been touched in awhile. I need to pull it down and double check, but I think this all looks great. Thanks for the contribution.

susilehtola:
Yeah I'm still trying to check if #606 is solved by this, but I'm pretty sure the static arrays were overflown and caused memory corruption.

susilehtola:
But this can be merged regardless.

PR #609, opened 2017-02-07
Fixing thread synchronization in AIOHandler.

### LINKED ISSUE #596: Integrals threading error messages
Running larger calculations with scf_type pk gives me lots and lots of error messages
```
$ psi4
thread already exists.
thread already exists.
thread already exists.
```
and so on.

These seem to arise from src/psi4/libpsio/aio_handler.cc. I'm wondering if these really are errors, if they should be printed, and whether stdout instead of the log file is the right place to do it.

jgonthier:
## Description

This PR fixes thread synchronization in AIOHandler after KTB. The std::thread in C++11 must not be joinable when it is destroyed, or it calls terminate(). Before destroying every thread created in AIOHandler, it is thus joined.

In addition, a thread must not be joined twice, thus the synchronization step now checks whether the thread is joinable before joining.

Fixes #596 and should definitely fix errors in tests cubeprop, sapt2, sapt3 and sapt6 that appeared in KTB and for which @loriab had a workaround.

## Todos
* **Developer Interest**
  - [x] Fixed thread management in AIOHandler.

## Status
- [x]  Ready to go


jgonthier:
I'd rather have some cake than beer ;-)

PR #610, opened 2017-02-07
CIWave: Fixes occasional power issue with the MCSCF DIIS object

dgasmith:
## Description
NumPy linked DSYEV was not robust enough under certain conditions. Switches to Psi4 power inversion to fix this issue.

## Status
- [x]  Ready to go


loriab:
This showed up in dfrasscf-sp on Mac  in a compilation with heavy use of conda packages (and thus 2017 `mkl_rt.so` lurking). Though `otool -L` provided no trace back to MKL, a separate cosseted build free of conda (no links into a conda env with MKL) did _not_ have this problem (a blatant problem in convergence).

So this is a good fix for a less common set of conditions.

PR #613, opened 2017-02-09
Fix for building with Intel without fortran enabled

bennybp:
## Description
Fixes an issue where building with Intel but without fortran enabled would lead to undefined symbols `for_rtl_init_()` and `for_rtl_finish_()`

## Todos
* **Developer Interest**
  - [X] Psi4 now builds and runs with intel without fortran

## Status
- [X]  Ready to go

### DIFF
--- a/psi4/src/CMakeLists.txt
+++ b/psi4/src/CMakeLists.txt
@@ -36,8 +36,9 @@ target_link_libraries(core PUBLIC pybind11::module)
 target_link_libraries(core PRIVATE Threads::Threads)
 
 if(Fortran_ENABLED AND CMAKE_Fortran_COMPILER_ID MATCHES Intel)
-  # Enable call to for_rtl_init_() and link to the correct library
-  target_compile_definitions(core PRIVATE -DFortran_ENABLED)
+  # Enable call to for_rtl_init_() which is required if using the
+  # Intel fortran compiler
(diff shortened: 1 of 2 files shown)


psi4/src/CMakeLists.txt
```diff
@@ -34,6 +34,13 @@ target_link_libraries(core PRIVATE ${psi4_binmodules})
 target_link_libraries(core PRIVATE ${LIBC_INTERJECT})
 target_link_libraries(core PUBLIC pybind11::module)
 target_link_libraries(core PRIVATE Threads::Threads)
+
+if(Fortran_ENABLED AND CMAKE_Fortran_COMPILER_ID MATCHES Intel)
```
loriab: Isn't just `if(Fortran_ENABLED)` enough here?

bennybp: I suppose that might be a little better, but yeah it is equivalent in this case

loriab: Ok, has my approval, though I might edit out the Intel later to avoid confusion.

bennybp: Actually, maybe not. It might be possible to compile the C++ source with intel and the fortran sources with something else. Which means that rtl_init, etc, are not required.

It should only be enabled if using the fortran compiler. So maybe a more descriptive flag would be better.

bennybp:
Latest push makes it a little more descriptive. Otherwise it's the same

PR #615, opened 2017-02-10
re-enable MKL threading

loriab:
## Description
So once upon a time, Psi4 ran ConfigMath to get its lapack/blas, then [tested for mkl](https://github.com/psi4/psi4private/blob/71ea2ea8ee142ee42541c589c42977fff2d3c381/CMakeLists.txt#L173-L174) to set `HAVE_MKL`. Then in the code proper, `#include mkl.h` (that defines `__INTEL_MKL__`) and the [very important `mkl_set_num_threads()`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libparallel/process.cc#L91-L93) were guarded by `HAVE_MKL` ifdefs. Then, on the fateful 23 Jun 2016 (https://github.com/psi4/psi4/commit/e07d645ea7acc85ae90d8c60a30f3426c977fc26) in the course of CMR, ConfigMath was blown away and all the `HAVE_MKL`s were replaced by `__INTEL_MKL__`. But that's circular because the mkl.h inclusions, which define `__INTEL_MKL__` were guarded by `__INTEL_MKL__` ifdefs. So Psi4 hasn't been threading MKL (though OMP threading has still been present). Story continues in bullet below. Identification and debugging of this issue by @dgasmith and Matt Schieber.

## Todos
* **Developer Interest**
  - [x] ~~This PR identifies which of MKL, OPENBLAS, etc. has been detected by ConfigMath.~~ Identify MKL-or-not by searching for "mkl" in the full lapack library list, whether detected math or passed-in math. Colors "MKL LAPACK" to screen if "on". If MKL, attaches the `USING_LAPACK_MKL` compiler definition to the `tgt::lapack` target that every module of Psi4 is linked with. Thus we're getting MKL threading again.
  - [x] Adds min version tags for dkh, gdma, erd
  - [x] Adjusts where the required check comes in for hdf5 so we don't get an warning
  - [x] Add docs with examples of math detection
  - [x] Remove unused routes in math detection: explicit, external, mkl_flag
  - [x] ConfigMath now FATAL_ERRORs if math libs _not_ found
  - [x] Newly pays attention to headers for math. Only required for MKL.
  - [x] Add a threading testing script for general maintenance, see comment below. In general tests where MKL found. Has compare_values, but really better suited for running by hand and judging results, as `-jN` defeats the purpose.
  - [x] Added some CCSD(T) table docs requested by @CDSherrill 
  - [x] Change Mac math precedence to `MKL SYSTEM_NATIVE OPENBLAS ESSL ATLAS ACML` so native accelerate only behind MKL
  - [x] New mechanism to pass in math libs explicitly through `LAPACK_LIBRARIES` and `LAPACK_INCLUDE_DIRS`
  - [x] Collects all the addon libraries into a list, then forms RPATH of internal (relative to ORIGIN), then all non-static, non `/usr/` paths. Redundancy removed and math libs paths _always_ first. This takes over from `INSTALL_RPATH_USE_LINK_PATH`.
  - [x] Bumped CMake minimum version to 3.2
  - [x] Added some extra badges to README

## Questions
- [x]  I had already implemented the isMKL? detection in ConfigMath before I looked back and found  [mkl testing] (https://github.com/psi4/psi4private/blob/71ea2ea8ee142ee42541c589c42977fff2d3c381/CMakeLists.txt#L173-L174). Do we want to do that way instead? Who added the "NEEDS" there? _From discussion below, No, bad idea._
- [x] @robertodr and/or @bast, I've commented out all the EXPLICIT_LIB and EXTERNAL_LIB stuff from ConfigMath because the new CMake scheme isn't doing anything with those libs, and the TargetLAPACK uses only `BLAS/LAPACK_LIBRARIES`. But we do need a way to take in a raw library string. By inclination, I would just have that if user specifies `BLAS/LAPACK_LIBRARIES`, then those get processed immediately into the TargetLAPACK. But I thought I'd see if you had other/opposing opinions. _From discussion below, they approve._
- [x] Currently require CMake 3.1. Can we require 3.2 (for `continue()`) or 3.3 (for HDF5 versioning)? 3.3 and 3.6 available on conda.
- [x] RPATH work focused on Linux. Mac I don't think is a problem, plus it's so easy to adjust after link. I'm hoping to continue ignoring rpath on Mac.

## Status
- [x]  Ready to go


dgasmith:
Looks like the MKL header is always being included:

```
/psi4/psi4/src/psi4/libparallel/process.cc:36:17: fatal error: mkl.h: No such file or directory
 #include <mkl.h>
```

loriab:
Makes sense since it's finding the conda mkl_rt. Will look into where the `mkl.h` ever actually came from. The distelli fault is hdf5 – altered reasoning too late at night. Looking at that now.

loriab:
If the latest patches work, ConfigMath should never again find a conda MKL (which probably means I should revert #611) because it can't find a `mkl.h` at the same MKL_ROOT. For all of us with intel compilers and mkl, this doesn't change a thing because ConfigMath was always finding our intel installations, never conda (the final lib may have shown up as linked to conda, but that's a `rpath`, not a ConfigMath, issue). But in cases w/o an intel installation and w/ a conda installation, ConfigMath should no longer settle upon conda mkl_rt.so because there's no associated `mkl.h`. Instead, should give up on MKL and proceed to OpenBLAS, ..., SYSTEM_NATIVE.

robertodr:
Your suggestion on how to handle explicit math libraries makes perfect sense @loriab

jgonthier:
Does anybody have an idea what happens if a threaded MKL call happens inside a threaded OpenMP call ?

loriab:
@dgasmith and [this](http://stackoverflow.com/questions/21560544/number-of-threads-of-intel-mkl-functions-inside-omp-parallel-regions) say MKL is smart and single-threads the nested call.

loriab:
This PR solves one clear lack-of-threading problem with MKL. But we've had hints of a more elusive one where threading-or-not depends on MKL library load order. Be on the looking for this. To check both threads issues, I've checked in a timings script (originally Daniel's) [here](https://raw.githubusercontent.com/loriab/psi4/b381e6c0865f3e47986044a409e4c7e7b1f21757/tests/python/threading/input.py) to compare p4 vs np threading and n1 vs nmax threading.

loriab:
This is ready for review. It actually requires CMake 3.2 (see Q above). It'll need a rebase and some squashing, but I'll do that once all the comments are in.

bast:
@loriab if I understood correctly the EXPLICIT_LIB change is to avoid having explicit libs baked into the math detection and I agree that it is good to separate these. We have also done this in other codes.

loriab:
@rbast, the EXPLICIT_LIB and EXTERNAL_LIB removal is because nothing outside cmake/math knows about those variables, so they're orphaned. LAPACK_LIBRARIES and LAPACK_INCLUDE_DIRS are being used in (I think, never used them myself) the EXPLICIT_LIB role.

doc/sphinxman/source/build_planning.rst
```diff
@@ -895,6 +911,131 @@ old to compile |PSIfour|, but the Fortran compiler will work.
...
+  immediate solution, however, is to form symlinks between the
+  library names that exist and the names expected. Consult file
+  :source:`cmake/math/MathLibs.cmake` for the library patterns being
+  sought.
+
+* The BLAS/LAPACK interface is standardized, so only libraries, not
+  headers, need to be detected. The exception is MKL, where the ``mkl.h``
+  header defines additional functionality; it must be located to use
+  BLAS threading.
+
```
andysim: This documentation is outstanding!

loriab: Thanks. I've been avoiding writing math detection docs for two years now, so my past few days' learning is now dumped to disk.

README.md
```diff
@@ -1,5 +1,17 @@
...
+<a href="https://gitter.im/psi4devs/Lobby"> <img src="https://img.shields.io/gitter/room/nwjs/nw.js.svg" /></a>
+<a href="http://forum.psicode.org/"> <img src="https://img.shields.io/badge/chat-on_forum-5077AB.svg" /></a>
+<br>
+<a href="https://github.com/psi4/psi4/releases"> <img src="https://img.shields.io/github/commits-since/psi4/psi4/1.0.svg" /></a>
+<a href="http://www.psicode.org"> <img src="https://img.shields.io/badge/home-Psi4-5077AB.svg" /></a>
+<a href="http://psicode.org/psi4manual/master/index.html"> <img src="https://img.shields.io/badge/docs-latest-5077AB.svg" /></a>
+<a href="https://conda.anaconda.org/psi4/psi4"> <img src="https://anaconda.org/psi4/psi4/badges/installer/conda.svg" /></a>
+<br><br>
+<img src="https://github.com/psi4/psi4media/blob/master/logos/psi4banner_half_tight.png" alt="Psi4 banner logo" height=200> <br>
+</p>
```
amjames: This looks real sharp with all of the badges at the top. Awesome stuff!

PR #619, opened 2017-02-14
Should solve integer overflows in SAPT code

jgonthier:
## Description
Substitute some int by size_t, and introduces (size_t) casts wherever multiplication of integers seemed problematic.

## Todos
* **Developer Interest**
  - [x] Integer declarations in all .h files in libsapt_solver have been examined and the ones that could potentially generate overflows were converted to size_t. Probably some overkill there.
  - [x] Searched the whole source in libsapt_solver for a regex representing 4 variables being multiplied, to find all the occ*vir*occ*vir that are bound to overflow. Added (size_t) cast for those that were missing it

## Questions
- [ ]  This originated in a problem reported by a user on the forum. I can't run Psi4 on good enough hardware to actually test the patch on a large system. Anyone want to give it a try ? Or do we let the user test it ?

## Status
- [x]  Ready to go


psi4/src/psi4/libsapt_solver/amplitudes.cc
```diff
@@ -620,7 +620,7 @@ void SAPT2::ijkl_to_ikjl(double *tARAR, int ilength, int jlength, int klength,
 
   for(int i=0; i<ilength; i++) {
     for(int l=0; l<llength; l++) {
-      long int ijkl = i*jlength*klength*llength + l;
+      long int ijkl = (long int)i*jlength*klength*llength + l;
```
susilehtola: I might remember wrong, but don't you need to also cast jlength, klength and llength, since the order of evaluation isn't guaranteed...? What this is doing is typecasting i to long int, but jlength*klength*llength is still int and might overflow.

psi4/src/psi4/libsapt_solver/disp2ccd.cc
```diff
@@ -240,7 +240,7 @@ void SAPT2p::r_ccd_prep(const char *TARBS, const char *ARBS, const char *CA_RBS,
 
   double **tARBS = block_matrix(occA*virA,occB*virB); //!
 
-  C_DCOPY(occA*virA*occB*virB,&(vARBS[0][0]),1,&(tARBS[0][0]),1);
+  C_DCOPY((size_t)occA*virA*occB*virB,&(vARBS[0][0]),1,&(tARBS[0][0]),1);
```
susilehtola: Same thing here.

psi4/src/psi4/libsapt_solver/amplitudes.cc
susilehtola: Nevermind, the compiler should promote all of the operands to long int before the multiplication takes place.

dgasmith:
@jgonthier These are good changes regardless, so pulling them in now. If it turns out it didn't fix the problem we can always go back in later.

PR #622, opened 2017-02-17
DFMP2 gradients with external potential

### LINKED ISSUE #618: MP2 Gradient Incorrect with QMMM charges
The analytic and numerical MP2 gradients disagree when a QMMM charge field is used, while the same calculation runs correctly at the HF level.  [psi4 version 1.0.0]

I found a similar issue which has already been resolved with the RKS gradient: https://github.com/psi4/psi4/issues/184 

My test case is one QM water molecule in the presence of two TIP3P-charge waters.

```
memory 400 mb

molecule water {
0 1
O    -0.167787    1.645761    0.108747
H    0.613411    1.10262    0.113724
H    -0.093821    2.20972    -0.643619
symmetry c1
no_reorient
no_com
}

chgfield = QMMM()

chgfield.extern.addCharge(-0.834, -1.350388,    -0.964879, -0.092208)
chgfield.extern.addCharge(0.417, -1.908991,    -1.211298,  0.626207)
chgfield.extern.addCharge(0.417, -1.263787,    -0.018107, -0.055536)
chgfield.extern.addCharge(-0.834, 1.517569,    -0.667424,    -0.080674)
chgfield.extern.addCharge(0.417, 1.989645,    -1.098799,    0.612047)
chgfield.extern.addCharge(0.417, 0.668397,    -1.091798,   -0.139744)
psi4.set_global_option_python('EXTERN', chgfield.extern)

set basis aug-cc-pvtz
set scf_type df
set freeze_core True

grad_fd = gradient('mp2', molecule=water, dertype=0)
grad_an = gradient('mp2', molecule=water)
compare_matrices(grad_an, grad_fd, 5, "TESTING finite-diff vs. analytic gradient")
```
The identical input file replacing "mp2" with "hf" runs successfully.  It seems strange to me, since as far as I understand the external charges should only modify the core contribution to the Fock matrix.  Is there a possibility that gradient('mp2') somehow fails to include the external potential gradients that are correctly computed in gradient('hf')?

andysim:
## Description
Adds missing CPHF terms, so that DF-MP2 gradients are correct when an external potential is present. Fixes #618.

## Todos

* **Developer Interest**
  - [x] Fixed DF-MP2 gradient code.
  - [x] Added test case for DF-MP2 QM/MM gradients (extern2, not currently in quicktests, but runs in 12 seconds).
  - [x] Fixed bug so that QM/MM gradients are correct for DF-MP2.

## Status
- [x]  Ready to go

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -27,6 +27,7 @@
 
 #include "mp2.h"
 #include "corr_grad.h"
+#include "psi4/liboptions/liboptions_python.h"
 #include "psi4/lib3index/3index.h"
 #include "psi4/libmints/basisset.h"
 #include "psi4/libmints/matrix.h"
@@ -40,6 +41,7 @@
 #include "psi4/psi4-dec.h"
 #include "psi4/physconst.h"
 #include "psi4/psifiles.h"
+#include "psi4/libmints/extern.h"
 #include "psi4/libmints/twobody.h"
 #include "psi4/libmints/integral.h"
 #include "psi4/libmints/oeprop.h"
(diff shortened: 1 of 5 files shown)


psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -2800,6 +2802,18 @@ void RDFMP2::form_gradient()
...
+    if (pyExtern) {
+        std::shared_ptr<ExternalPotential> external = pyExtern.cast<std::shared_ptr<ExternalPotential>>();
+        if (external) {
+            gradient_terms.push_back("External Potential");
+            timer_on("Grad: External");
+            gradients_["External Potential"] = external->computePotentialGradients(basisset_, PAO);
+            timer_off("Grad: External");
+        }  // end external
+    }
+
```
dgasmith: Awesome, thanks Andy! As a note I guess this will effect all other gradients (and Hessians) within Psi4 as well? We might need to add an extra list to filter out gradients that have not considered external potentials.

PR #623, opened 2017-02-18
un-require libefp

loriab:
## Description
Requiring libefp was rather a holdover from the early days of code-in-repo add-ons. This PR makes it slightly easier on packagers and enables us to have only three required compile-in dependencies in the figure in the paper (libint, blas/lapack, pybind11).

## Todos
* **Developer Interest**
  - [x] `#ifdef`-ing the EFP code probably could have been done to varying degrees of thoroughness. But since EFP is tied into `P::e` _and_ Molecule _and_ it's own interface, none of which I wanted to rexamine, this is the one that requires a minimum of recompilation if you change `ENABLE_libefp` ON/OFF. When OFF, the EFP object still exists, but only the constructor, destructor, and nfragments method (that returns 0 fragments). This way the driver needs no mods (which is a good thing, since we can't `#ifdef` python).
  - If you build w/o libefp and try an efp-containing molecule, throws a runtime error with text `EFP fragments detected but are not available`.
  - [x] slight fix of passing `pybind11_DIR` to plugin cache. helps with mixed install/staged psi4/pybind11. we'll see how this goes, as I'm not sure it's the best fix.
  - [x] more thoroughly test cmake variable for PCMSolver path (`if "@ENABLE_PCMSolver@".upper() in ["1", "ON", "YES", "TRUE", "Y"]`)
  - Hope this helps a bit on packaging, @susilehtola 
  - [x] Psi4 can be build without libefp add-on

## Status
- [x]  Ready to go


psi4/src/export_efp.cc
```diff
@@ -39,8 +39,10 @@ void export_efp(py::module& m)
...
     py::class_<EFP, std::shared_ptr<EFP> >(m, "EFP", "Class interfacing with libefp").
         def(py::init<Options&>()).
+#ifdef USING_libefp
         def("compute", &EFP::compute, "Computes libefp energies and, if active, torque").
         def("set_qm_atoms", &EFP::set_qm_atoms, "Provides libefp with QM fragment information").
-        def("nfragments", &EFP::get_frag_count, "Returns the number of EFP fragments in the molecule").
-        def("print_out", &EFP::print_out, "Prints options settings and EFP and QM geometries");
+        def("print_out", &EFP::print_out, "Prints options settings and EFP and QM geometries").
+#endif 
+        def("nfragments", &EFP::get_frag_count, "Returns the number of EFP fragments in the molecule");
```
andysim: Haven't given this much thought, but does the mechanism that generates the docstrings obey the preprocessor macros?

loriab: Didn't think of that either. But probably ok b/c these docstrings are API-extracted, not script-autodoc'd and b/c proper docs build from fully featured psi4. Ok, just tested, and though I can't view them, the docs at least build just fine. Or was the question do docs filter their content by `ENABLE_<AddOn>` directives? That's no, except by accident. The read_options autodoc-ing doesn't even respect commented-out options.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -1845,10 +1851,12 @@ void HF::print_energies()
 
     Process::environment.globals["SCF ITERATIONS"] = iteration_;
 
+#ifdef USING_libefp
     // Only print this alert if we are actually doing EFP or PCM
     if(pcm_enabled_ || ( Process::environment.get_efp()->get_frag_count() > 0 ) ) {
         outfile->Printf("    Alert: EFP and PCM quantities not currently incorporated into SCF psivars.");
     }
+#endif
```
dgasmith: This is still an important alert for PCM. Will need to break PCM/EFP out.

loriab: Good point. I released the print message from ifdef, as `get_efp` and `get_frag_count` are safe to use, libefp enabled or not.

susilehtola:
@loriab yeah it's one less mandatory prerequisite to worry about packaging!

psi4/src/export_efp.cc
andysim: My main concern was the auto-doc stuff, but that's apparently a non-issue.  It'd be nice for the pydocs to be consistent with user compilation settings, but certainly not a big deal.

PR #533, opened 2016-11-29
WIP: Simint integration in Psi4

bennybp:
## Description
**Currently a work-in-progress**

**Currently requires intel compiler**

Implements the simint two-electron integral library into Psi4.

## Todos
Notable points that this PR has either accomplished or will accomplish.
* **Developer Interest**
  - [X] Implement ENABLE_SIMINT option
  - [X] Download & build via ExternalProject_Add
  - [ ] Download & use precompiled library
  - [X] Simint is being used rather than libint
  - [ ] Batching interface to interface classes in libmints (partially done)
  - [ ] Use batching in JK objects (partially done)
  - [ ] Fix some GCC issues in simint
  - [ ] Port new batching interface to libint & erd
  - [ ] Higher AM, or split between simint and erd
  - [ ] Derivatives (WIP in simint...)

* **User-Facing for Release Notes**
  - [ ] Hmmm

## Questions
- [ ]  Where to put tarballs (that is more permanent for psi4). Git is probably not a good idea

## Status
- [ ]  Ready to go

PR #625, opened 2017-02-19
Simint Integration, Continued

loriab:
## Description
Some Psi4-side fixes to getting @bennybp's simint AddOn added

## Todos
* **Developer Interest**
  - [x] Includes all commits of #533 
  - [x] Temporarily based off private repo https://github.com/loriab/simint branch config6 until the integration settles down, and we can return to tarballs. Note that if you want to test this branch, you may need to supply your GH credentials during the build to clone simint.
  - [x] Adds all the pre-built detection, target referencing, docs, standardized variable names, tests that we expect from an addon to simint
  - [x] Continued general documentation on addons.
  - [x] Modestly tested- static and dynamic on Linux, dynamic on Mac
  - [x] New integral library available with `-DENABLE_simint`

## Questions for @bennybp 
- [ ]  There's some questions in external/upstream/simint/CMakeLists.txt
- [ ]  I've made the simint on my acct private and just added you and some core developers who may want to test it. If you don't mind it being public, as a temp repo, that'd be a little easier.
- [ ] There are spots where I supply at most separate home, docs, and source links for addons. I've got your website, http://www.bennyp.org/research/simint/, for all three at the moment, so let me know if there are more specifics.
- [ ] I filled in the installing and configuring sections of simint.rst according to the template. Other addons (http://psicode.org/psi4manual/master/interfacing.html) often have middle sections about interacting with the code. Fill in anything you like.
- [ ] Got the following unexpected failing test cases on Linux and partially confirmed on Mac, all segfaulted. Could you take a look? None are in quicktests.
```
The following tests FAILED:
	 16 - cbs-xtpl-wrapper (Failed)
	144 - dft-b2plyp (Failed)
	269 - pywrap-checkrun-rhf (Failed)
	270 - pywrap-checkrun-rohf (Failed)
	271 - pywrap-checkrun-uhf (Failed)
	319 - cubeprop-esp (Failed)
```

## Status
- [ ]  Ready to go


robertodr:
I get this error when running `cmake -H. -Bobjdir_simint -DCMAKE_BUILD_TYPE=Debug -DENABLE_simint=ON` on my laptop with CMake 3.2.2
```
CMake Error at /usr/share/cmake-3.2/Modules/ExternalProject.cmake:2027 (get_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "_EP_IS_EXTERNAL_PROJECT" is not allowed.
Call Stack (most recent call first):
  /usr/share/cmake-3.2/Modules/ExternalProject.cmake:2324 (_ep_add_configure_command)
  CMakeLists.txt:154 (ExternalProject_Add)
```

loriab:
I can confirm @robertodr's error. But it does away with CMake 3.3 (even though I can't see what the change was from the [release notes](https://cmake.org/cmake/help/v3.7/release/3.3.html)). I've sort of been wanting 3.3 because that detects the HDF5 version number, and I, at least, am susceptible to mixed 1.8.16 and 1.8.17 compilations, which the build system can now avoid. So shall we bump again?

robertodr:
If it solves this **and** library detection problems, then let's go for it. I notice that in [3.6](https://cmake.org/cmake/help/v3.7/release/3.6.html) they added support for compiler features also for Intel compilers, but maybe that's not as interesting.

loriab:
I consider that aspect of 3.6 very interesting. We wouldn't have to manually add `-std=c++11` and could avoid a repeated compiler warning about overwriting `-std=c++11` by ~`-std=gnuc++11` when cmake _is_ >=3.6 . But 3.6 is quite a leap just for our convenience.

robertodr:
The following should do the trick already, no?
```
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
set(CMAKE_CXX_EXTENSIONS FALSE)
```
You basically require C++11 _without_ GNU extensions from the compiler. Configuration should abort if the compiler doesn't have the relevant flag. These three lines would possibly eliminate the need for checking for compiler versions in [cmake_cxxstandard.cmake#L3-L11](https://github.com/psi4/psi4/blob/master/cmake/custom_cxxstandard.cmake#L3-L11)

loriab:
The first two we've got the the inner `psi4-core` project (psi4 itself). The third I tried to get some opinions on (copied below for general interest). The compiler checks at least give a warning people understand better than C++ standards and do it before all the addons (which often don't require c++11) get built. _And_ for the crazy Intel/gnu situation, it's a nice place to check the underlying compiler version. So I'm all for solving the EXTENSIONS warnings and for revamping when we decide to require cmake 3.6.

How do we feel about -std=c++11 vs. -std=gnu++11 via [CXX_EXTENSION](https://cmake.org/cmake/help/v3.4/prop_tgt/CXX_EXTENSIONS.html)?
I can fix the `icpc: command line warning #10121: overriding '-std=c++11' with '-std=gnu++11’` by (a) turning that cmake variable off (its default is on); (b) when omp is enabled, have our cmake add the gnu++ instead of the c++ flag; or (c) suppressing that icpc warning.

robertodr:
Just tested and the check on compiler version is still needed, since earlier compilers might not be feature-complete and CMake doesn't check for all features when setting the standard.

robertodr:
Do I understand correctly that icpc + OpenMP wants `-std=gnu++11`? If that's the case, I would have `set(CMAKE_CXX_EXTENSIONS TRUE)` for all cases.

loriab:
No, the deal is that Psi4 adds `-std=c++11` to `CMAKE_CXX_FLAGS`, and everyone's happy with it. I haven't heard any petitions to make it `-std=gnu++11`. And we pass the CXX_STANDARD on to target `psi4-core`. Then with cmake 3.6 (which newly understands intel capabilities) + intel compilers, cmake takes that CXX_STANDARD + icpc and adds its flag so that `CMAKE_CXX_FLAGS` becomes `... -std=c++11 ... -std=gnu++11 ...` and then icpc throws a warning at every file that c++11 is being overwritten. The simplest thing, I think is to `CMAKE_CXX_EXTENSIONS=OFF` in the `psi4-core` ExternalProject list in main CMakeLists.txt.

bennybp:
One thing to note about the c++ standard is that it is showing up on the command line twice. I'm almost certain that one is coming from pybind11 interface definitions (which might not be a good idea). 

    cd /home/ben/programming/psi4/build.reference/psi4-core-prefix/src/psi4-core-build/src/psi4/liboptions && /opt/intel2017/compilers_and_libraries_2017.1.132/linux/bin/intel64/icpc
    -I/home/ben/programming/psi4/psi4.reference/psi4/include
    -I/home/ben/programming/psi4/psi4.reference/psi4/src
    -isystem /home/ben/programming/psi4/build.reference/stage/home/ben/programming/psi4/install.reference/include
    -isystem /usr/include/python2.7  -xHost -qopenmp
    -std=c++11 -O3 -DNDEBUG -fPIC   -std=c++11
    -o CMakeFiles/options.dir/print.cc.o
    -c /home/ben/programming/psi4/psi4.reference/psi4/src/psi4/liboptions/print.cc

loriab:
Ah, that's a good point, @bennybp. The psi4_STANDARD was designed before `pybind11::module` had the standard attached. And pybind11 supports CMake back to 2.8.12, so it's determined to set a flag, rather than surrendering control of CXX standard to CMake itself.

[`cxxstandard.cmake`](https://github.com/psi4/psi4/blob/master/cmake/custom_cxxstandard.cmake#L47) is really only affecting `psi4-core` and possibly pasture. `psi4-core` is covered by flag from pybind11 _and_ by passing the project `CMAKE_CXX_STANDARD`. If add `CMAKE_CXX_EXTENSIONS=OFF`, then the signals from both aforementioned should be in sync (double flag, but the same flag), and shouldn't need [block](https://github.com/psi4/psi4/blob/master/cmake/custom_cxxstandard.cmake#L43-L49) at all.

loriab:
Turns out @robertodr's original `_EP_IS_EXTERNAL_PROJECT` error [was because](http://public.kitware.com/pipermail/cmake-developers/2015-April/025035.html) I moved all the external_* dependencies from `ExternalProject_Add(psi4-core ...) add_dependencies(psi4-core chemps2_external ...)` to `ExternalProject_Add(psi4-core DEPENDS chemps2_external ...)`, and CMake didn't learn to accept interface libs in that role until 3.3.

So, I still rather want 3.3 because the DEPENDS looks tidier and the HDF5 version thing, but I can make 3.2 work again.

loriab:
Updated this a bit after talking with @bennybp. I think he's going to do the following, then it'll be about ready:
* tarball with higher (h) AM
* also, could you throw an error in libmints if try to access out of bounds AM through simint? or switch to libint?
* absorb the config6 branch into upstream.

loriab:
@bennybp, the hours-long step in preparing the simint tarball is the execution of the generator executable to produce the source files packed up in the simint tarball, yes? Since you're now bundling all vectorizations together in that tarball, how difficult would it be to generate to a quite high AM, say 7 or 8, then make the MAX_AM_ERI that actually gets compiled a build-time parameter?

bennybp:
Yes, that may be possible. It doesn't take *that* long to generate (I just did AM = 7 in about 30 mins). There is just some slight trickery with determining which files must be compiled.

loriab:
Writing out a different CMake list of source to be compiled for each AM? If that's do-able, I think it's a good way to get an all-inclusive tarball so that I'm not bugging you for a set of tarballs, AM=4–8.

loriab:
Ok to push my rebase of this branch, @bennybp?

Also, https://github.com/simint-chem/simint-generator/issues/1.

bennybp:
Yes you can rebase. Shouldn't be too bad since I just did one last week

loriab:
Continuing thread from other repo, @bennybp.
- I _haven't_ removed the `restrict` def psi4-side. Go to town on this `simintlab` branch.
- I have run a simint AM5 full test suite on psi4, and it's clean.
- I have built a Mac conda package, and it builds and "ldd"'s ok
- I haven't rebuilt the Linux conda package after you confirmed refreshed tarball, but will do that next.

loriab:
What's your status on this, @bennybp? Are you deleting the `restrict` def and rebasing or shall I? I don't doubt that this PR _works_, it just hasn't been much exercised. Should we get it merged in at least? What's the status of a higher AM tarball?

bennybp:
I'll try to wrap everything up this afternoon. I just need to do the official release (which is easy, just requires me making a changelog and stuff). And I will remove the restrict keyword from cmake

bennybp:
Ok new version 0.7 is up with max AM=7. Trying it out now, but not anticipating any problems.

Also pushed the removal of the restrict keyword.

loriab:
Thanks! On to review.

bennybp:
Oops spoke too soon. Fixing some minor cmake issues now...

bennybp:
Ok compiles now and the first few tests are passing

doc/sphinxman/source/simint.rst
```diff
@@ -0,0 +1,141 @@
...
+* Downstream Dependencies |w---w| |PSIfour| (\ |dr| optional) simint
+
+* Upstream Dependencies |w---w| simint |dr| None
+
+**CMake Variables**
+
+* :makevar:`ENABLE_simint` |w---w| CMake variable toggling whether |PSIfour| builds with simint
+* :makevar:`CMAKE_PREFIX_PATH` |w---w| CMake list variable to specify where pre-built dependencies can be found. For simint, set to an installation directory containing ``include/simint/simint.h``
+* :makevar:`simint_DIR` |w---w| CMake variable to specify where pre-built simint can be found. Set to installation directory containing ``share/cmake/simint/simintConfig.cmake``
+* :makevar:`CMAKE_DISABLE_FIND_PACKAGE_simint` |w---w| CMake variable to force internal build of simint instead of detecting pre-built
```
andysim: I see that SIMINT_VECTOR takes a default value of SSE.  This will be significantly slower than AVX (which most modern machines support) so maybe it's worth instructing users how to find CPU features (I do `cat /proc/cpuinfo` on Linux and `sysctl -a | grep machdep.cpu.features` on OSX) so that they can set this to the fastest supported vectorization?  Also a listing of the possible vectorization schemes (either here or in CMakeLists.txt) could be helpful.

loriab: When discussing this with Ben before, I found https://github.com/tunabrain/tungsten/blob/master/cmake/OptimizeForArchitecture.cmake, which is admittedly already out of date and would be difficult to maintain. If there's something that could be hooked up to `exeucte_process` to gather the info, would be good to do. Certainly would be good to have a list.

dgasmith: Could we default to AVX and always toss out a warning with suggestions that they should assert the validity of the assumption? AVX has been shipping since 2011 which does not seem particularly onerous.

If SSE is the default I am not quite sure how much faster simint would be over libint (Ben correct me here). This could turn users off from trying to use simint when available.

andysim: @loriab , I'm not suggesting that we should auto-detect the vectorization scheme (although it can be done using `#ifdef __SSE__`,`#ifdef __SSE2__`,`#ifdef __SSE3__`,`#ifdef __AVX__` and `#ifdef __MIC__` at compile time and a call to `cpuid` at run time).  One of the potential pitfalls of that would be compiling on one node, then running on different architectures, which would have to be accounted for using the above mechanism.  I think the easiest thing to do is to set the default to AVX, as @dgasmith suggested, and put a warning in the documentation and/or CMakeLists.txt that this may not be appropriate.  I can provide code snippets for testing cpu features if you like, but I advocate the simple approach.

loriab:
Didn't recognize you without the knives, @bennybp. Great, I was going to ask you for a useable photo for the Psi4 lineup.

bennybp:
Ha yeah. I decided I needed a better profile picture. I have other, similar ones if you'd like to look at them or you needed a better resolution.

PR #627, opened 2017-02-22
threading, scratch, misc. fixes, esp. for PsiAPI

loriab:
## Description
Several general fixes, esp. for PsiAPI mode.

## Todos
* **Developer Interest**
  - [x] Turning off gnu C++ extensions. This means that `-std=gnu++11` will never get passed, even if CMake can detect that your compiler supports it. Instead will get a `-std=c++11` from pybind11 and possibly another `-std=c++11` if cmake detects that your compiler supports.
  - [x] Put docs in order wrt threading
  - [x] Make `psi4/__init__.py` pay attention to `PSI_SCRATCH`. Otherwise scratch is non-adjustible in PsiAPI mode.
  - [x] Default to serial execution in `psi4/__init__.py`. Can be overruled in Psithon mode through `-nN` or `set_num_threads(N)`. Can be overruled in PsiAPI mode through `set_num_threads(N)`.
  - [x] Fix `atexit` deregistering in Py3
  - [x] Removed get_/set_num_threads driver wrappers. Still looks the same to users. Prints "Threads set to %d by Python driver" in output file so there's a record of the change.
  - [x] `psi4.compare_values()` and friends adapted to work with pytest upon success. Recommended pattern: `ans = psi4.compare_values(normal test args here); assert ans`.
  - [x] Set up pytest so can test installed psi4 free of cmake/ctest. For the moment, you need to be in `${top-level-psi4-dir}/tests/pytest`, have pytest installed (`conda install pytest`), have psi4 importable (`psi4 --psiapi-path`), and run `python -m pytest` or for more info, `python -m pytest -vs`.
  - [x] Revive a ccsd(t) sp and an mp2 opt in the cfour interface. Add a pysmoke test.
  - [x] Vital tweaks so that psiapi can handle more than vanilla calcs. Includes initializing efp in psi4.geometry, moving pcmsolver path detect to __init__, and the options clearing mentioned above.
  - [x] Fix TargetHDF5 detection for detect pre-built case (i.e., conda case).

  - [x] CMake 3.3 now required
  - [x] Using `OMP_NUM_THREADS` and/or `MKL_NUM_THREADS` to guide parallelism in Psi4 is _now deprecated_. Psi4 won't pay the least attention to these variables.
  - [x] Added `psi4 --psiapi-path` command-line argument. This prints the bash commands you need to execute to set up python interp and pythonpath in your shell to `python -c "import psi4"`
  - [x] Psi4 molecule strings passed to `psi4.geometry()` no longer balk at python-style comments, as these are filtered out.
  - [x] Added a `psi4.clean_options()` fn (alongside `psi4.clean()` and `psi4.clean_variables()`) to reset the `P::e.options` object to its initial state. This is necessary for independent tests in pytest and will probably be useful for psiapi sessions.
  - [x] Add pytest smoke tests for psi4 itself and psi4+addons (addon tests skip if addon not enabled or detectable). From source, `cd tests/pytest && python -m pytest [-v -s]`. From staged or installed, `python -c "import psi4; psi4.test()"` or `psi4 --test`. From objdir, `make pytest`.

## Questions
- [ ]  I can work around the 3.3 cmake. It'll just be less tidy and won't ward off the HDF5 version detection issue. Speak up if there are objections.
- [ ] Everyone should make sure you're ok with the table below. No envvar control over parallelism anymore.
- [ ] Everyone read 12cd carefully, as I'm a bit nervous about this. Searched the code and Options object is cleared _nowhere_ except for a commented out in read_options. So I think it's safe to commandeer this function. And it works to completely reset options.
- [ ] The threading test is now in test_psi4_profiling.py. It works fine as the only pytest, but it's threading is still messed up when in the whole pytest suite. Anyone have a guess why? I haven't read anything about pytest controlling threads.

## Status
- [x]  Ready to go

`OMP_NUM_THREADS` | `MKL_NUM_THREADS` | `-nN` | `psi4.set_num_threads()` | PsiAPI | Psithon
--- | --- | --- | --- | --- | ---
_ | _ | _ | _ | _111 | 1111
1 | _ | _ | _ | _111 | 1111
3 | _ | _ | _ | _111 | 1111
_ | 1 | _ | _ | _111 | 1111
_ | 3 | _ | _ | _111 | 1111
_ | _ | 1 | _ | N/A | 1111
_ | _ | 3 | _ | N/A | 3333
_ | _ | _ | 1 | _111 | 1111
_ | _ | _ | 3 | _333 | 1333
_ | _ | 4 | 2 | _222 | 4222

Where the four numbers are extracted from the output file: threads in psi4 header, threads in MintsHelper, threads in SCF header, threads used by JK. The first of these in inapplicable in PsiAPI mode.


amjames:
I have no issues with requiring cmake version 3.3 for the sake of cleaner CMakeLists.txt's I think it is worth it. 

w.r.t to the changes in handling threading, you have a big :+1: from me on this:
 - Ignoring the envars is a good move. If `__init__.py` was going to look at `os.eviron` to set the default number of threads which should take precedence `MKL` or `OMP`? 
   -  > *In the face of ambiguity, refuse the temptation to guess.*
 - With the command line argument, or `psi4.set_num_theads()` it is simple to honor the envars *explicitly*.
   -  > *Explicit is better than implicit* 


One suggestion /comment (If the general consensus is in favor of this):
 - Should the number of threads be removed from the psi4 header? I can see confusion coming from that line when `psi4.set_num_threads()` is used. Maybe just append something that indicates it is the value from the command line?

psi4/run_psi4.py
```diff
@@ -213,9 +213,12 @@
 if args["messy"]:
     import atexit
 
-    for handler in atexit._exithandlers:
-        if handler[0] == psi4.core.clean:
-            atexit._exithandlers.remove(handler)
+    if sys.version_info >= (3, 0):
+        atexit.unregister(psi4.core.clean)
```
dgasmith: Ah, so glad they made this easy.

dgasmith:
I pretty much echo what AJ said.

@amjames I see you have been reading `import this`.

CMakeLists.txt
```diff
@@ -220,7 +223,8 @@ include(ConfigTesting)
 
 install(DIRECTORY ${STAGED_INSTALL_PREFIX}/
         DESTINATION ${CMAKE_INSTALL_PREFIX}
-        USE_SOURCE_PERMISSIONS)
+        USE_SOURCE_PERMISSIONS
+        PATTERN "*.pyc" EXCLUDE)
```
dgasmith: Good catch.

doc/sphinxman/source/build_planning.rst
```diff
@@ -1300,6 +1309,23 @@ a few other factors. Either way, this file should contain CMake's testing
...
+(staged installation). To test an installed |PSIfour| (full or staged
+installation), a limited number of "smoke" tests are available to be
+run via pytest. ::
+
+    python -c "import psi4; psi4.test()"
+
+For the above to work, |PSIfour| must be detectable as a Python
+module. See the note at :ref:`faq:runordinarymodule` for setup if needed.
+
+
```
dgasmith: Could we add this under ctest? Perhaps something like `ctest -L pytest`?

doc/sphinxman/source/external.rst
```diff
@@ -208,23 +208,27 @@ explained below. Note that each deeper level trumps all previous levels.
...
+..     setenv MKL_NUM_THREADS 4
+..
+.. .. code-block:: bash
+..
+..     # sh, bash: add to shell or ~/.bashrc (Linux) or ~/.bash_profile (Mac) file
+..     export OMP_NUM_THREADS=4
+..     export MKL_NUM_THREADS=4
+..
+.. |PSIfour| then detects these value via the API routines in ``<omp.h>`` and
+.. ``<mkl.h>``, and runs all applicable code with 4 threads.
```
dgasmith: Is this still true? It reads like we are purely using P::e now.

psi4/__init__.py
```diff
@@ -71,11 +77,19 @@
 # Make official plugins accessible in input
 from .driver import endorsed_plugins
 
+# Manage threads. Must be after endorsed plugins, honestly.
+core.set_num_threads(1, quiet=True)
+
```
dgasmith: Nice, I think thats everything set in `__init__` that is required now.

psi4/extras.py
```diff
@@ -58,3 +59,77 @@ def exit_printing():
...
+
+def addons(request=None):
+    """Returns boolean of whether Add-On *request* is available to Psi4,
+    either compiled in or searchable in $PSIPATH:$PATH, as relevant. If
+    *request* not passed, returns list of available Add-Ons.
+
+    """
+    if request is None:
+        return sorted([k for k, v in _addons_.items() if v])
+    return _addons_[request.lower()]
```
dgasmith: I think if `request is None` then this would return a list that cannot index `_addons_` or `lower`.

tests/pytest/test_psi4_proper.py
```diff
@@ -0,0 +1,270 @@
...
+        "reference"       : 'rhf',
+        "scf_type"        : 'pk',
+        "mcscf_algorithm" : 'ah',
+        "qc_module"       : 'detci',
+        "nat_orbs"        : True})
+
+    cisd_energy, cisd_wfn = psi4.energy("CISD", return_wfn=True)
+
+    ans = psi4.compare_values(-76.2198474477531, cisd_energy, 6, 'CISD Energy')
+    assert ans
```
dgasmith: Any reason not to `assert psi4.compare...`?

doc/sphinxman/source/build_planning.rst
loriab: Can't add to `ctest`. But can make a `cmake` target of it so can `make pytest` and that'll run it. List is getting extensive.

* from `{top-level-psi4-dir}/tests/pytest`, set PYTHONPATH, `python -m pytest -v`
* from anywhere, set PATH, `psi4 --test`
* from anywhere, set PYTHONPATH, `python -c "import psi4; psi4.test()"`
* from `{objdir}`, `make pytest`

doc/sphinxman/source/external.rst
loriab: If I mention lines starting with `.. ` are commented out, does that change your question?

psi4/__init__.py
loriab: I'd still like `print_header()` to happen somewhere in psiapi mode. But admittedly, I can't find a place for it that wouldn't be annoying or unpythonic in some circumstance.

tests/pytest/test_psi4_proper.py
loriab: hmm, makes sense, now you mention it

psi4/extras.py
loriab: Not following. It seems to work fine:

```
python -c "import psi4; print(psi4.addons())"
['chemps2', 'dftd3', 'dkh', 'gcp', 'gdma', 'libefp', 'pcmsolver']
```

doc/sphinxman/source/external.rst
dgasmith: Ah, I didnt realize. Very much so in that case.

psi4/extras.py
dgasmith: Apparently I am blind and mixed up `request/return`.

andysim:
I see a 3c smoke test in here, so I think it's best for me to update the branch with the 3c PR contents, before merging.  Thoughts, @loriab ?  Excellent stuff in this one - sorry for the delayed review.

loriab:
Right @andysim, the 3c test was shuffled btwn PRs. The other had 3c but no pytest. This has pytest but no 3c. Either order would have been fine, as nothing's counting on the pytest to pass at this point (and it wouldn't even have run if you didn't have gcp installed), which is good b/c the threading test fails in the multi-pytest mode. So moot, but that's what was going on.

PR #630, opened 2017-02-28
Add HF-3c and PBEh-3c

loriab:
## Description
Add HF- and PBEh-3c capabilities, now that the former have been verified from Orca and the latter through S. Grimme private communications.

## Todos
* **Developer Interest**
  - [x] gcp smoke tests going to pytest PR
  - [x] write a bit of docs
  - [x] So this is where my compare_values improvements ran off to. Now have a toggle to fail or not and can check mismatch by val, not just by order of magnitude value.
  - [x] Add `gcp` interface, but not as advanced as `dftd3` b/c not writing to a file, just takes one of the pre-def values. 
  - [x] Semi-semiempirical methods HF-3c and PBEh-3c of Grimme can now be accessed (with `dftd3` and `gcp` executables detectable in PATH). Access through `energy('hf3c')` and `energy('pbeh3c')`. Can run with or without preferred basis set specified (minix and def2-msvp, respectively).

## Questions
*  @dgasmith, this'll need a little adaption for 1.2

## Status
- [x]  Ready to go. Not averse to a squash.


hokru:
oh this cool! I suspected some work on this, because Grimme forwarded me some email about psi4 and asked me to fix our gcpd3 webserver.

loriab:
Aha, you were the number supplier behind the email chain, @hokru? Thank you very much. Those gave me enough confidence to believe the PBEh-3c implementation correct. And the gcpd3 webserver and an Orca run yesterday convinced me that HF-3c is in good order, too. Note, though, that we're not changing scf algorithm defaults to follow the *-3c definitions. Psi4 default is still `DF`, even though non-DF and DF were recommended for HF-3c and PBEh-3c, respectively.

hokru:
no numbers, i just restarted the webserver some time ago and there was something about psi4 in the forwarded email. It brakes occasionally and I am still responsible for it. I will probably test the methods once it goes to the master.

the choice of the scf algorithm was purely for practical reasons of getting the best performance in ORCA and/or Turbomole, i believe. RI-J combined with exact exchange is pretty good for these methods and their target molecule size.

loriab:
Thanks, @hokru, any testing or comments would be welcome, either now or once merged.

Btw, the doi link nearest your name on [the gCP page](http://www.thch.uni-bonn.de/tc/downloads/gcp/index.html) is broken, if that's a concern.

psi4/driver/qcdb/interface_gcp.py
```diff
@@ -0,0 +1,307 @@
...
+    elif dertype == 1:
+        return psi_dashdderiv
+
+try:
+    # Attach method to libmints psi4.Molecule class
+    core.Molecule.run_gcp = run_gcp
+except (NameError, AttributeError):
+    # But don't worry if that doesn't work b/c
+    #   it'll get attached to qcdb.Molecule class
+    pass
```
dgasmith: Out of curiosity, we didnt want to tack this onto the current dispersion function? It has the flexibility to wrap as many underlying programs that it needs.

loriab: Sorry, didn't see this earlier. I wasn't sure of the best way to append the gcp corrections, so at some point (I guess in December) went with _a_ way. The gcp now is only two functionals (hf3c and pbeh3c), so before it gets generalized, I should see what scheme you have in mind to work best with 1.2 and adapt it there, I think.

dgasmith: I was thinking of adding it [here](https://github.com/psi4/psi4/blob/b1a604bf799a08b05debfe468ac5ee0b61b393a8/psi4/driver/procedures/empirical_dispersion.py#L191) or something similar.

loriab: lines 304--307 are (iirc) just so the run_gcp method can be applied to a pure qcdb.Molecule with psi4 nowhere in sight. There's three modes: standalone on qcdb.Mol, standalone on psi4.core.Mol, and as part of a psi4 DFT. I think 304--307 are just handling standalone mode. Unless you think `energy('gcp')` or `energy('d3m')` should be allowed? Does that explain things or does your rearrangement suggestion still hold?

dgasmith: What I mean is specifically in the future a superfunctional is really a tuple of (Superfunctional, EmpericalDispersion). The EmpericalDispersion object can be duck typed so that it doesnt really matter what it it is along as you follow `compute_energy`, `compute_gradient`, etc. However, adding GCP as an option to EmpericalDispersion would be a favorable route as it already contains psi4.core.Dispersion and dftd3. Its just a convenient place to put more atom pairwise corrections without having to add them anywhere else.

Issue #631, opened 2017-02-28
Build error with gcc 7.0.1


susilehtola:
My pybind11 package was accepted in Fedora, so I started updating the psi4 package (which was at rc1!). It built fine on the released versions, but failed in the development version that has the newest GCC.

The error(s) I get is

```
cd /builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/objdir-x86_64-redhat-linux-gnu/psi4-core-prefix/src/psi4-core-build/src/psi4/libpsio && /usr/bin/g++    -I/builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/psi4/include -I/builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/psi4/src -isystem /usr/include/python3.6m  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -fopenmp -DUSING_CheMPS2 -O2 -DNDEBUG -fPIC   -std=c++11 -std=gnu++11 -o CMakeFiles/psio.dir/aio_handler.cc.o -c /builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/psi4/src/psi4/libpsio/aio_handler.cc
/builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/psi4/src/psi4/libpsio/aio_handler.cc: In member function 'long unsigned int psi::AIOHandler::read(unsigned int, const char*, char*, psi::ULI, psi::psio_address, psi::psio_address*)':
/builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/psi4/src/psi4/libpsio/aio_handler.cc:91:48: error: 'bind' is not a member of 'std'
   thread_ = std::make_shared<std::thread>(std::bind(&AIOHandler::call_aio,this));
                                                ^~~~
/builddir/build/BUILD/psi4-926879e243ce44dcc0ecfe7cda680a01383982fd/psi4/src/psi4/libpsio/aio_handler.cc:91:48: note: suggested alternative: 'find'
[...]

PR #633, opened 2017-03-02
Add missing include.

susilehtola:
## Description
Add missing include to fix build with gcc 7 (see #631).


susilehtola:
It's still building, but it's gone further than before. I'll let you know.

susilehtola:
Confirmed - builds succesfully with gcc7 with this patch.

PR #634, opened 2017-03-02
Updates MRCC interface to allow perturbative methods with ROHF reference.

jturney:
## Description
Recent versions of MRCC are able to compute perturbative energies [e.g. CCSDT(Q)] with an ROHF reference wave function. This updates Psi4's interface to allow these methods to be performed.

Credit goes to @bzloink for updating the interface.

  - [x] Perturbative methods using MRCC with an ROHF reference are now supported.

## Status
- [x]  Ready to go


psi4/src/psi4/mrcc/mrcc.cc
```diff
@@ -640,8 +642,11 @@ PsiReturnType mrcc_load_ccdensities(SharedWavefunction wave, Options &options, c
     if (options.get_str("REFERENCE") == "UHF")
         restricted = false;
 
-    if (pertcc && options.get_str("REFERENCE") == "ROHF")
-        throw PSIEXCEPTION("Perturbative methods not implemented for ROHF references.");
+    if (pertcc && options.get_str("REFERENCE") == "ROHF") {
+        outfile->Printf("\n");
+        outfile->Printf("  WARNING: ROHF references are not implemented for perturbative\n");
+        outfile->Printf("           methods on older versions of MRCC. Proceed with caution.\n\n");
+    }
```
loriab: Do you know what MRCC versions have the new capability?

jturney: MRCC-2014-07-01 and newer versions appear to have this ability.  I can't speak for older versions.

jturney:
I'll have time tomorrow to update the docs and tests.

bzloink:
I'd actually thought about clarifying that warning. MRCC will only run perturbative CC on semicanonical orbitals - this is how CFour passes the integrals, and that's mostly what I changed. Even the newest versions of MRCC are unable to run perturbative CC using standard ROHF orbitals - this may have initially led to the flags in both CFour and Psi4.

psi4/src/psi4/mrcc/mrcc.cc
```diff
@@ -753,13 +758,17 @@ PsiReturnType mrcc_generate_input(SharedWavefunction ref_wfn, Options &options,
...
+    if (options.get_str("REFERENCE") == "ROHF")
+        canonical = false;
 
-    if (pertcc && options.get_str("REFERENCE") == "ROHF")
-        throw PSIEXCEPTION("Perturbative methods not implemented for ROHF references.");
+    if (pertcc && options.get_str("REFERENCE") == "ROHF") {
+        outfile->Printf("\n");
+        outfile->Printf("  WARNING: ROHF references are not implemented for perturbative\n");
+        outfile->Printf("           methods on older versions of MRCC. Proceed with caution.\n\n");
+    }
```
andysim: Oops.  @loriab already asked this.  I forgot to click "start review" earlier when I typed this.  I guess it's better just to leave line comments in future.

loriab: Any chance the MRCC executable can report it's version through `dmrcc --version or the like`, @bzloink? I'm expecting not, but if it can, we can set psi4 up to test.

bzloink: Nothing that I know of, but in response to the new capabilities, as far as I can tell from the [2008 JCP paper,](http://aip.scitation.org/doi/10.1063/1.2988052) perturbative schemes for semicanonical references have been available for almost a decade.

dgasmith:
I think if we can get a test for this in the `tests/mrcc` folder this will be good to go.

tests/mrcc/rohf_ccsdt_q_/input.dat
```diff
@@ -0,0 +1,34 @@
+memory 500 Mb
```
loriab: I don't know that we need to set mem in test cases unless it particularly needs more than the default (500 MiB).

tests/mrcc/rohf_ccsdt_q_/input.dat
```diff
@@ -0,0 +1,34 @@
...
+O  1.7242024334  0.1244981797  0.0000000000
+H -0.3510151994  1.1579201425 -0.8912164899
+H -0.3510151994  1.1579201425  0.8912164899
+C -1.8420084989 -0.1447691015  0.0000000000
+H -1.9545447709 -0.7674520567 -0.8865786907
+H -1.9545447709 -0.7674520567  0.8865786907
+H -2.6301702073  0.6086421759  0.0000000000
+}
+
+energy('mrccsdt(q)')
```
loriab: The last item in the `ene` file is what gets set as `CURRENT ENERGY`. Is that sensible for the ROHF case? I'm not sure what the `/A` and `/B` are. Do they have any analogies in cc* modules that we should standardize?

tests/mrcc/rohf_ccsdt_q_/input.dat
jturney: Ok , I'll remove the memory keyword.

tests/mrcc/rohf_ccsdt_q_/input.dat
jturney: The ```/A``` and ```/B``` are from Kallay's paper: Kallay, M.; Gauss, J. J. Chem. Phys. 2008, 129, 144101.  ```/B``` is found to be more robust: Martin, J. M. L. Mol. Phys. 2014, 112, 785. So it sounds like a sensible default for CCSDT(Q). A note should probably be made in the documentation.

jturney:
It doesn't appear to be honoring the ```MRCC_OMP_NUM_THREADS``` keyword or family. Looks like how the environment variables being passed to mrcc was changed but not all the settings made the trip.

loriab:
It shouldn't be honoring `OMP_NUM_THREADS` or `MKL_NUM_THREADS` since psi4 is ignoring them, but should preferentially honor `MRCC_OMP_NUM_THREADS` keyword. If not that should probably default to `core.get_num_threads()`.

jturney:
Ah, I was just looking for ```core.get_num_threads()```. I'll change it to ignore ```OMP_NUM_THREADS```.

loriab:
And I think the new test case is missing a CMakeLists.txt

jturney:
Now uses ```core.get_num_threads()``` to determine the number of threads to use with MRCC; this can be overridden by ```MRCC_OMP_NUM_THREADS```. Updates documentation with reference to Kallay's CCSDT(Q)/A and CCSDT(Q)/B paper and makes note about CCSDT(Q)/B being used by default and cites Martin's paper.  I couldn't compile the documentation so please let me know if the formatting is off.

doc/sphinxman/source/mrcc.rst
```diff
@@ -78,8 +78,14 @@ input file for MRCC for the methods listed in table below.
...
-Note that perturbative methods (``ccsd(t)``, ``ccsdtqp(h)_l``, etc.)
-are not available with |scf__reference| ROHF.
+Perturbative methods (``ccsd(t)``, ``ccsdtqp(h)_l``, etc.)
+are available with |scf__reference| ROHF in versions of MRCC published
+at least after July 1, 2014.
+
+When using ROHF-CCSDT(Q), MRCC will compute and report two variants:
+CCSDT(Q)/A and CCSDT(Q)/B.[Kallay:2008:144101]_ Psi4 will save both energies but will use
+the CCSDT(Q)/B as the CCSDT(Q) energy. CCSDT(Q)/B has been found to be
+more robust by Martin.[Matrin:2014:785]_
```
loriab: typo in author citation?

jturney: Whoops

PR #638, opened 2017-03-06
Optexit and pytest

loriab:
## Description
Make `psi4.test()` safe to run. And beer safe to grep.

## Todos
* **Developer Interest**
  - [x] The psi4 vs numpy threading test is more likely to work if run first, so moved it there. But since that's still no guarantee, made it an expected failure, so it doesn't count against pytest either way.
  - [x] with `ENABLE_GENERIC`, blas/lapack detection looks for dynamic libm.
  - [x] Geometry optimizations that exceed maxiter no longer just print a warning and exit successfully (beer). Now they exit unsuccesfully (coffee) and raise a `psi4.ConvergenceError`.

## Status
- [x]  Ready to go


psi4/driver/driver.py
```diff
@@ -1139,12 +1141,12 @@ def optimize(name, **kwargs):
 
         n += 1
 
-    core.print_out('\tOptimizer: Did not converge!')
     if core.get_option('OPTKING', 'INTCOS_GENERATE_EXIT') == False:
         if core.get_option('OPTKING', 'KEEP_INTCOS') == False:
             core.opt_clean()
 
     optstash.restore()
+    raise ConvergenceError("""geometry optimization""", n - 1)
```
dgasmith: Cool, as discussed I think this will alleviate the `grep 'beer' output.dat` problem.

PR #641, opened 2017-03-07
Feature add: string argument to API's psi4.set_memory()

dsirianni:
## Description
Adds ability for psi4.set_memory() to take unit-containing string as an argument, so that
`psi4.set_memory('30 GB')` behaves as would be expected.

## Todos
* **User API Changes**
  - `int`, `float`, and `str` argument types supported
  - `int` and `float` type arguments taken literally as number of bytes of memory to be set
  - `str` type arguments must have units attached, e.g., `'20 mb'`, and are case-insensitive
  - Added return of the amount of memory set
  - renamed underlying fn to `psi4.core.set_memory_bytes()
  - Arranged for inputparser to use the same psi4.set_memory() as psiapi, so now kib/mib/gib/tib avail. in input
  - Correct the Psi4 and SCF headers which were computing MB and labeling MiB
  - Add and test some bad examples of mem setting
  - Add a minimum memory (250 MiB), adjust the default mem to 500 MiB

## Questions
- [x] Currently, arguments of type `str` **MUST** have units.  Should we allow users to input unitless numbers as `str`, to be taken literally?  Or, if they wish to do that, should they just input as `int` or `float`? Consensus is, No, string args must have unit.

## Status
- [x] Ready to go


psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,70 @@ def cubeprop(wfn, **kwargs):
...
-def set_memory(bytes):
-    """Function to reset the total memory allocation."""
-    core.set_memory(bytes)
+def set_memory(inputval):
+    """Function to reset the total memory allocation. Takes memory value 
+    *inputval* as type int, float, or str; int and float are taken literally
+    as bytes to be set, string taken as a unit-containing value (e.g., 30 mb)
+    which is case-insensitive.
+    
+    :returns: None
```
dgasmith: Could return the number of bytes that it set.

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,70 @@ def cubeprop(wfn, **kwargs):
...
+     '100.0 KiB': 102400L,
+     '500 MB': 500000000L,
+     '2 eb': 2000000000000000000L}
+
+    """
+    # Handle memory given in bytes directly (int or float)
+    if isinstance(inputval, int) or isinstance(inputval, float):
+        val = inputval
+        units = ''
+    # Handle memory given as a sring
```
dgasmith: Space between the indents here, might also want a `units = units.upper()` to ensure matches.

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,70 @@ def cubeprop(wfn, **kwargs):
...
+            units = str(matchobj.group(2))
+            val = float(str(matchobj.group(1)))
+        except AttributeError as e:
+            raise ValidationError("""Invalid memory specification {}.""".format(inputval))
+        
+    # Units decimal or binary?
+    multiplier = 1000
+    if "i" in units.lower():
+        multiplier = 1024
+        units = ''.join(units.split('i'))
```
dgasmith: Might needs something like `units.lower().replace("i", "").upper()` to handle cases. If there is more than one `i` it should throw.

dsirianni: Good catch.  Also, the current RE won't match more than one `i`, but that can be rectified with `i{,}?B`

dgasmith: Well if someone does something like `KiiB` we should probably throw an error.

dsirianni: That's already baked in then, since the RE not matching will raise the `ValidationError: Invalid memory specification 32 KiiB`

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,72 @@ def cubeprop(wfn, **kwargs):
...
+     '2 eb': 2000000000000000000L}
+
+    """
+    # Handle memory given in bytes directly (int or float)
+    if isinstance(inputval, int) or isinstance(inputval, float):
+        val = inputval
+        units = ''
+    # Handle memory given as a string
+    if isinstance(inputval, basestring):
+        memory_string = re.compile(r'\s*?([+-]?\d*\.?\d+)\s+([KMGTPBE]i?B)', re.IGNORECASE)
```
loriab: Can drop `[+-]?` from regex.

dsirianni: So, throw the aforementioned `ValidationError` if any sign is given?

psi4/driver/p4util/util.py
```diff
@@ -138,13 +138,12 @@ def set_memory(inputval):
         units = ''
     # Handle memory given as a string
     if isinstance(inputval, basestring):
```
amjames: The builtin `basestring` has been removed in python 3. This will throw:

```python
NameError  File "/Users/admin/Workspace/Git/psi4/objdir/stage/Users/admin/psi4-dev-installs/test/lib/psi4/driver/p4util/util.py", line 140, in set_memory
    if isinstance(inputval, basestring):
NameError: name 'basestring' is not defined
```

Which points out another problem; `psi4.set_memory()` Should be checked somewhere in quick-tests, but we have python3 builds on travis and this PR passed.

Also, should probably add a `psi4.set_memory()` call somewhere in quick-tests that uses a string as input.

Possible Solution (I haven't tested this)
- Just check if input is `float` or `int` if not str is assumed and any non-`str` inputs will lead to exceptions raising in `re.search`

dsirianni: Changing this line to a simple `else:` would fix that, and should still work for unicode or raw strings as well

dsirianni: Could also check `basestring` as in 
```
try:
    basestring
except NameError:
    basestring = str
```
which would then work with the current checks

loriab: I was going to suggest [this](https://github.com/psi4/psi4/blob/2118f2f5b5ffbe66932b8a1f20f4553cd3ee9415/psi4/driver/p4util/procutil.py#L40) for py3, but yes, an `else:` works too, and more simply. Good point, AJ, I'm going to add some tests of this to options1.

amjames: That is what I am thinking. 

As far as testing goes @loriab might want to put this check in a specific place, but it should be tested *somewhere* in quicktests.

dgasmith: @amjames Good catch! We could add a new test in `tests/python/*` which tests things like `set_output_file`, `set_memory`, etc.

@dsirianni In python 3, `isinstance(s, str)` is sufficient which is also backwards compatible.

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,71 @@ def cubeprop(wfn, **kwargs):
...
+     '': AttributeError("'NoneType' object has no attribute 'group'"),
+     '21986': AttributeError("'NoneType' object has no attribute 'group'"),
+     '100 kb': 100000L,
+     '100.0 KiB': 102400L,
+     '500 MB': 500000000L,
+     '2 eb': 2000000000000000000L}
+
+    """
+    # Handle memory given in bytes directly (int or float)
+    if isinstance(inputval, int) or isinstance(inputval, float):
```
dgasmith: This can be collapsed to `isinstnace(inputval, (int, float))`.

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,71 @@ def cubeprop(wfn, **kwargs):
...
+        val = inputval
+        units = ''
+    # Handle memory given as a string
+    if isinstance(inputval, basestring):
```
dgasmith: Yes, this does need to be `else if isinstance(inputval, str)`.

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,71 @@ def cubeprop(wfn, **kwargs):
...
+        memory_string = re.compile(r'^\s*?(\d*\.?\d+)\s+([KMGTPBE]i?B)', re.IGNORECASE)
+        matchobj = re.search(memory_string, inputval)
+        if matchobj:
+            val = float(matchobj.group(1))
+            units = matchobj.group(2)
+        else:
+            raise ValidationError("""Invalid memory specification '{}'.""".format(inputval))
```
dgasmith: Need an `else` below here that throws if the above two do not match.

psi4/driver/p4util/util.py
dgasmith: `try/except` `if/else` when dealing with Python 3 is not a good idea these days IMO. It adds extra code that will need to be ripped out sooner rather than later.

psi4/driver/p4util/util.py
dsirianni: Just checked, this is be compatible with either unicode or raw string input types.

psi4/driver/p4util/util.py
dsirianni: Echo the same `ValidationError` as on 147?

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,71 @@ def cubeprop(wfn, **kwargs):
...
+
+    # Build conversion factor, convert units        
+    unit_list = ["", "KB", "MB", "GB", "TB", "PB", "EB"]
+    mult = 1
+    for unit in unit_list:
+        if units.upper() == unit:
+            break
+        mult *= multiplier
+            
+    memory_amount = int(val * mult)
```
amjames: Still Working on my morning coffee, so forgive me if this is obvious.

How does this handle decimal input

For example do
`psi4.set_memory("0.5 GB")` 
and 
`psi4.set_memory("500 MB")` 
both result in the same `memory_amount` getting passed to `core.set_memory()`?

dsirianni: Building the multiplier is based on the unit, not on `val`.  So, regardless of `4000 GB` or `0004 GB` or `0.004 GB` the multiplier is the same, so `val` is scaled appropriately.

psi4/driver/p4util/util.py
amjames: Well the outer `else` would be catching input that is not `str`,`int`, or `float` so maybe a `TypeError` would be more appropriate.

psi4/driver/p4util/util.py
amjames: Great! :+1:

psi4/driver/p4util/util.py
dsirianni: Added outer `TypeError` in latest commit to handle this.

loriab: I'm voting for (and implementing with a similar helpfully TypeError msg) ValidationError since it prints more helpfully

psi4/driver/inputparser.py
```diff
@@ -796,7 +789,9 @@ def future_replace(m):
 
     # Move up the psi4.core namespace
     for func in dir(core):
-        temp = temp.replace("psi4." + func, "psi4.core." + func)
+        # Except fns that have valid wrappers in driver
+        if func not in ['set_memory']:
+            temp = temp.replace("psi4." + func, "psi4.core." + func)
```
dgasmith: This could get rough, would you consider moving the `psi4.core.set_memory` to a different internal name?

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,91 @@ def cubeprop(wfn, **kwargs):
...
+    for unit in unit_list:
+        if units.upper() == unit:
+            break
+        mult *= multiplier
+
+    memory_amount = int(val * mult)
+
+    min_mem_allowed = 524288000
+    if memory_amount < min_mem_allowed:
+        raise ValidationError("""Please, sir, I want some memory. ({} < {} b)""".format(memory_amount, min_mem_allowed))
```
dgasmith: This needs to be a bit more explicit in its error. Perhaps something like (note the formatted quantities):
```
Minimum allowed memory setting is 500 MiB (300 MiB requested). Please increase the amount of memory specified.
```

`psi4/header.py` has an automatic formatter for you, `sizeof_fmt`.

psi4/driver/p4util/util.py
```diff
@@ -98,10 +99,91 @@ def cubeprop(wfn, **kwargs):
...
+
+    800000000         # 800000000
+    2004088624.9      # 2004088624
+    1.0e9             # 1000000000
+    '600 mb'          # 600000000
+    '600.0 MiB'       # 629145600
+    '.6 Gb'           # 600000000
+    ' 100000000kB '   # 100000000000
+    '2 eb'            # 2000000000000000000
+
```
dgasmith: Do we explicitly test these anywhere?

loriab: Sure, [here](https://github.com/psi4/psi4/pull/641/files#diff-cce9d0a4205503ec53f080f5c4210c36)

psi4/driver/inputparser.py
loriab: Yes, it is awkward, and renaming is a better idea than the excepting I did. How about `psi4.core.set_memory_bytes(bytes)` and `psi4.core.get_memory()`. The symmetry of the core fn names is lost, but for the most part, the user still sees `psi4.set_memory()` and `psi4.get_memory()`.

psi4/driver/p4util/util.py
loriab: You'd rather not have the quote? How about:

`set_memory(): requested 300 MiB; minimum 500 MiB. Please, sir, I want some more.`

dsirianni: For what it's worth, I like the quote.  But it does seem like a good idea to have the minimum requirement given explicitly.

amjames: I think @loriab's revised error looks great, explicit short, and includes the quote (I like it by the way). 

Just wondering, is the limit only enforced here? In the past I have tested out-of-core  algorithms by turning the memory way down to force them into binning without running a large job on my laptop. 

I think the 500MiB limit is fine I just want to know if I can temporarily side step it by removing this check.

loriab: Yes, the limit is only enforced here. But contrary to the current pre-PR state, this function now has broader scope, controlling both psiapi and psithon memory setting, since now inputparser uses this fn.

You can still do a bypass like you describe by calling the core fn directly, like [here](https://github.com/loriab/psi4/blob/dom_setmem2/tests/cc44/input.dat#L13-L15). Though that will shortly become `psi4.core.set_memory_bytes` (`set_memory_bytes` in psithon) after Daniel's suggestion is implemented. The core fn has no limit (except positive).

dsirianni: If the user requests too little memory but with KB, MB, etc. units, should we present the minimum amount in the unit they originally requested the memory in? So, `set_memory('300 MiB')` would still produce `ValidationError: set_memory(): requested 300 MiB; minimum 500 MiB. Please sir, I want some more.`, but `set_memory('400 KB')` would return the error message with `requested 400 KB, minimum 525,000 KB`?  Or should we leave the user to fend for themselves in the unit conversion?

loriab: Personally, I think they can fend for themselves (of course I thought that bytes were sufficient), but I defer to consensus.

dgasmith: Bytes is probably not a good way to go `(5000000000 < 500000000 bytes)`.

dsirianni: Since (post PR) the `set_memory()` command here will be used by the inputparser, I am sympathetic to the novice user who may have no idea what they're doing.  I could see that having the error read `set_memory(): requested 300 KB; minimum 524288 KB.  Please, sir, I want some more.` would be more immediately helpful to those users.

loriab:
FYI, Dom is going to push a broken commit that I want to debug in a controlled env. So still blocked (and probably all failing).

CDSherrill:
Would 256MB minimum memory work just as well as 500MB?  I'm a little
concerned about imposing this as a minimum when the real minimum required
to make the program work is actually lower.

loriab:
@CDSherrill, nominally, yes. We've a test case that sets `memory 2 mb` to test an algorithm, and it runs just fine. Of course, we've never tested psi4 on a platform that has only 2–256 MB available.

In general, I'm with you that I'd rather the minimum at which it complains was lower than the default: 100 or 256 mb. But it's not too hard to get around: `psi4.core.set_memory_bytes(2000)` will happily set internal memory to 2kb.

dsirianni:
So, consensus on changing the minimum allowable memory with `set_memory()` to 100 mb?

Issue #347, opened 2016-04-16
reorganize Hessian, vib, thermo analysis


loriab:
### Current

Presently the `psi4.fd_freq_1()` and `psi4.fd_freq_0()` calls from `driver.hessian()`
- take in a Molecule and list of G or E, respectively
- stitch together the mass-weighted (mw) Hessian and compute freqs & normco
- pass vibs to the infamous `// This print function also saves frequencies in wavefunction.\n print_vibrations(mol, modes);`
- form the nmw Hessian and returns it

Then `driver.hessian()`
- initializes a new Wavefunction
- grabs from globals the frequencies that findif has stuffed there and puts them in Wfn
- takes the returned H matrix and puts it in Wfn

Then `driver.frequency()`
- extracts the frequencies from Wfn
- passes Wfn and Wfn.frequencies into thermo

Then `psi4.thermo()`
- extracts Molecule and E0 from Wfn
- uses those and frequencies to do the thermodynamic analysis

Key lines from driver below (note order of execution)

```
1 frequency(name):
2    H, wfn = hessian(name, return_wfn=True, molecule=molecule, **kwargs)  # send away for hessian
7    psi4.thermo(wfn, wfn.frequencies())  # pass frequencies into thermo analysis

3  hessian(name):   
4    H = psi4.fd_freq_1(molecule, gradients, irrep)  # stores freq in global psi4.frequencies(), returns H
5    wfn.set_hessian(H)  # store H in fresh wfn
6    wfn.set_frequencies(psi4.get_frequencies())  # store freq in fresh wfn
```

So there are three stages: "assemble H", "vibrational analysis", & "thermo analysis". Keeping in mind that the nmw Hessian is isotope-independent while freqs are isotope-dependent, you'll notice that the present arrangement of groups (1, 2) and (3) with frequencies as intermediate are such that can't reuse H for vibrational analysis of different isotopes (though it does allow for reuse of freq for thermo analysis at diff T, P).
### Future

assemble H
- this is doing just fine c-side in `psi4.fd_freq_0/1`
- should focus on returning nmw Hessian (as currently) does and trim out the vibrational analysis part
- @psi-rking, I don't remember seeing any projection of rot/trans out of Hessian. Is this just because by using SALCs you're assured of only displacements in vib-active directions? Assuming we someday get an analytic Hessian mtd, projection would be needed, right? I'm inclined to group projection with the vib analysis for comparison of proj and unproj freqs, so psi4.fd module needn't be involved.
[...]

Issue #225, opened 2016-01-26
freq oddity


loriab:
So one does a frequency calc, either by energies or gradients

```
#! Optimization followed by frequencies H2O HF/cc-pVDZ

memory 250 mb

molecule h2o {
  O 
  H 1 0.96
  H 1 0.96 2 104.5
}

set basis cc-pVDZ
optimize('scf')
#frequencies('scf', dertype='energy')  # TOGGLE
frequencies('scf', dertype='gradient')  # TOGGLE

eval = psi4.get_frequencies()
eval.print_out()
evec = psi4.get_normalmodes()
evec.print_out()

ref_freqs = psi4.Vector(3) #TEST
ref_freqs.set(0, 0, 1775.65) #TEST
ref_freqs.set(0, 1, 4113.38) #TEST
ref_freqs.set(0, 2, 4212.18) #TEST

fd_freqs = psi4.wavefunction().frequencies() #TEST
compare_vectors(ref_freqs, fd_freqs, 1, "Reference vs computed frequencies to 0.1 cm^-1") #TEST
```

Both versions pass the frequencies check. But if you diff their printed normal modes, as below, the coordinate system seems unaligned. This job doesn't have the sow/reap complication, though it does have the zmat one. Basically, is this expected or correctable?

```
   Frequency:       1775.58				      |	   Frequency:       1775.65
   Force constant:   0.1193					   Force constant:   0.1193
	     X       Y       Z           mass				     X       Y       Z           mass	
  O 	   0.000   0.000  -0.068      15.994915		      |	  O 	   0.000   0.000  -0.270      15.994915
  H 	   0.000   0.416   0.536       1.007825		      |	  H 	   0.000   0.417   0.538       1.007825
  H 	   0.000  -0.416   0.536       1.007825		      |	  H 	   0.000  -0.417   0.538       1.007825

   Frequency:       4113.41				      |	   Frequency:       4113.38
   Force constant:   0.6403					   Force constant:   0.6403
	     X       Y       Z           mass				     X       Y       Z           mass	
  O 	   0.000   0.000   0.049      15.994915		      |	  O 	   0.000   0.000   0.197      15.994915
  H 	   0.000   0.569  -0.392       1.007825		      |	  H 	   0.000   0.571  -0.393       1.007825
  H 	   0.000  -0.569  -0.392       1.007825		      |	  H 	   0.000  -0.571  -0.393       1.007825

   Frequency:       4212.15				      |	   Frequency:       4212.18
[...]

PR #644, opened 2017-03-09
Add a function to save normal modes and geometry after a vibrational analysis

fevangelista:
## Description
This PR adds code to the findif codes that allows to save the current geometry and normal modes to  a Molden file.

Normal mode info is only written if explicitly requested by the user (via the boolean keyword NORMAL_MODES_WRITE). The normal modes are saved in [Molden format](http://www.cmbi.ru.nl/molden/molden_format.html).

## Todos
* **Developer Interest**
  - [x] Exposes (read only-mode) some variables in the VIBRATIONS class.
  - [x] Adds a function to save normal modes info in Molden format.
  - [x] Adds an option to write normal modes (false by default).
  - [x] Documentation

## Questions
- [x]  Where should the documentation go?
- [x]  Suggestions to improve output format?

## Status
- [x] Ready to go

### DIFF
--- a/doc/sphinxman/source/freq.rst
+++ b/doc/sphinxman/source/freq.rst
@@ -50,6 +50,16 @@ the scenes, :py:func:`~psi4.frequency` is a light wrapper over
 :py:func:`~psi4.hessian` that computes the Hessian then adds a
 thermochemical analysis.
 
+Visualization of Normal Modes
+-----------------------------
+
+|PSIfour| has the ability to export a Molden file that stores information about
+the harmonic frequancies and normal modes computed via :py:func:`~psi4.frequency`.
+This feature can be enabled by setting the option |globals__normal_modes_write| to true.
+The filename of the Molden file ends in ”.molden_normal_modes, and the prefix is
+determined by |globals__writer_file_label| (if set), or else by the name of the
+output file plus the name of the current molecule.
+
 .. autofunction:: psi4.frequency(name [, molecule, return_wfn, func, mode, dertype, irrep])
 
 .. autofunction:: psi4.hessian(name [, molecule, return_wfn, func, dertype, irrep])
--- a/doc/sphinxman/source/freq.rst
+++ b/doc/sphinxman/source/freq.rst
@@ -50,17 +50,24 @@ the scenes, :py:func:`~psi4.frequency` is a light wrapper over
 :py:func:`~psi4.hessian` that computes the Hessian then adds a
 thermochemical analysis.
 
-Visualization of Normal Modes
------------------------------
-
-|PSIfour| has the ability to export a Molden file that stores information about
-the harmonic frequencies and normal modes computed using :py:func:`~psi4.frequency`.
-This feature can be enabled by setting the option |globals__normal_modes_write| to true.
-The filename of the Molden file ends in ”.molden_normal_modes, and the prefix is
-determined by |globals__writer_file_label| (if set), or else by the name of the
-output file plus the name of the current molecule.
-
 .. autofunction:: psi4.frequency(name [, molecule, return_wfn, func, mode, dertype, irrep])
 
 .. autofunction:: psi4.hessian(name [, molecule, return_wfn, func, dertype, irrep])
 
+Visualization of Normal Modes
+-----------------------------
+
+|PSIfour| has the ability to export a Molden file that stores information about
+the harmonic frequancies and normal modes computed via :py:func:`~psi4.frequency`.
+This feature can be enabled by setting the option |globals__normal_modes_write| to true.
+The filename of the Molden file ends in ”.molden_normal_modes, and the prefix is
+determined by |globals__writer_file_label| (if set), or else by the name of the
+output file plus the name of the current molecule.
+The normal coordinates saved in the Molden file are mass-weighted and are not normalized.
+
+Molden Interface Keywords
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. include:: autodir_options_c/globals__normal_modes_write.rst
+
+.. include:: autodir_options_c/globals__writer_file_label.rst
--- a/doc/sphinxman/source/freq.rst
+++ b/doc/sphinxman/source/freq.rst
@@ -36,8 +36,8 @@
 
 .. _`sec:freq()`:
 
-Harmonic Vibrational Analysis |w---w| :py:func:`~psi4.frequency` and :py:func:`~psi4.hessian`
-=============================================================================================
+Harmonic Vibrational Analysis and Visualization of Normal Modes |w---w| :py:func:`~psi4.frequency` and :py:func:`~psi4.hessian`
+===============================================================================================================================
 
 * :ref:`Psi4 Native Hessian Methods <table:freq_gen>`
 
@@ -49,6 +49,16 @@ need to access directly to perform frequency calculations. Behind
 the scenes, :py:func:`~psi4.frequency` is a light wrapper over
 :py:func:`~psi4.hessian` that computes the Hessian then adds a
 thermochemical analysis.
+
+Visualization of Normal Modes
+-----------------------------
+
+|PSIfour| has the ability to export a Molden file that stores information about
+the harmonic frequancies and normal modes computed via :py:func:`~psi4.frequency`.
+This feature can be enabled by setting the option |globals__normal_modes_write| to true.
+The filename of the Molden file ends in ”.molden_normal_modes, and the prefix is
+determined by |globals__writer_file_label| (if set), or else by the name of the
+output file plus the name of the current molecule.
 
 .. autofunction:: psi4.frequency(name [, molecule, return_wfn, func, mode, dertype, irrep])
 
--- a/doc/sphinxman/source/freq.rst
+++ b/doc/sphinxman/source/freq.rst
@@ -54,7 +54,7 @@ Visualization of Normal Modes
 -----------------------------
 
 |PSIfour| has the ability to export a Molden file that stores information about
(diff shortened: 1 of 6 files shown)


loriab:
Haven't looked over this closely, but could VMD format and [Molden Format](http://www.cmbi.ru.nl/molden/molden_format.html) work together?

fevangelista:
Let me take a look at this. I wonder if Molden is ok with just getting the normal modes and the geometry without the orbitals. Otherwise things are a bit more involved and I need to grab data from Wavefunction and the vib. analysis to create the molden file.

CDSherrill:
As a possible alternative, conceivably there could be some advantage in
just using the vibrational frequencies portion of a Molden file even if a
full-blown Molden file is inconvenient (although I'm not familiar enough
with the specifics to know if that really makes sense).

loriab:
I don't know what the Molden program requires, but I'm pretty sure that the Molden format is mix-and-match so no Wfn info req'd.

jturney:
I think @loriab is right that you can mix and match what you want in the Molden file.

fevangelista:
Ok, I tested it and it's a piece of cake. I'll save the info in Molden format and change the script for VMD to use this interface.

loriab:
Oh, good. We'll want the data structures accessible in python, too (plan for the last three years), but for text, Molden format should do for convention.

fevangelista:
The changes I pushed should do it. The option that writes the molden file is documented. Since visualizing vibrations is a very useful feature I was thinking of adding a small manual section "Normal Mode Visualization with Molden and VMD".

loriab:
Either a separate section like your "Orbital Visualization with VMD" or add to `freq.rst`, which will put it [here](http://psicode.org/psi4manual/master/freq.html), I think.

fevangelista:
I added some documentation so I think this PR is close to being ready for a commit.

I am curious: are there any ideas on how to make normal modes available via Python? From what I have seen in findif, it might be necessary to store this information somewhere. Exactly where, I don't know.

doc/sphinxman/source/freq.rst
```diff
@@ -50,6 +50,16 @@ the scenes, :py:func:`~psi4.frequency` is a light wrapper over
...
 
+Visualization of Normal Modes
+-----------------------------
+
+|PSIfour| has the ability to export a Molden file that stores information about
+the harmonic frequencies and normal modes computed using :py:func:`~psi4.frequency`.
+This feature can be enabled by setting the option |globals__normal_modes_write| to true.
+The filename of the Molden file ends in ”.molden_normal_modes, and the prefix is
+determined by |globals__writer_file_label| (if set), or else by the name of the
+output file plus the name of the current molecule.
```
loriab: fyi, to include text from read_options keywords, you can do something like [this](https://github.com/psi4/psi4/blame/master/doc/sphinxman/source/sapt.rst#L409) which will copy it in bodily and keep the manual in sync with read_options.

loriab:
Oh, and I meant to add that docs should say whether un-mass-weighted vs. mass-weighted and normalized vs. unnormalized normal coordinates are printed to the molden file. Pretty sure that it's the former on both choices.

On the Python accessibility issue, there's some discussion [here](https://github.com/psi4/psi4/issues/347) (start in the paragraph above "Future"). Findif's vibrational analysis capabilities have to be moved out of that module because it's organized on the assumption the psi4 has _no_ analytic Hessians available. Some of the shift to py-side is already in the works (@bwb314), and I expect this molden-writing can be moved py-side, too, once the data is accessible.

fevangelista:
@loriab: I think I agree with you but perhaps @psi-rking can help disentangle this. This is the code in `findidf/fd_freq_0.cc`:

```
        if (print_lvl >= 3) {
            outfile->Printf("\n\tForce Constants for irrep %s in mass-weighted, ", irrep_lbls[h]);
            outfile->Printf("symmetry-adapted cartesian coordinates.\n");
            mat_print(H_irr[h], salcs_pi[h].size(), salcs_pi[h].size(), "outfile");
        }

        // diagonalize force constant matrix
        int dim = salcs_pi[h].size();
        double *evals = init_array(dim);
        double **evects = block_matrix(dim, dim);

        sq_rsp(dim, dim, H_irr[h], evals, 3, evects, 1e-14);

        // Build Bu^1/2 matrix for this irrep
        SharedMatrix B_irr_shared = salc_list.matrix_irrep(h);
        double **B_irr = B_irr_shared->pointer();

        for (int i = 0; i < dim; ++i)
            for (int a = 0; a < Natom; ++a)
                for (int xyz = 0; xyz < 3; ++xyz)
                    B_irr[i][3 * a + xyz] /= sqrt(mol->mass(a));

        double **normal_irr = block_matrix(3 * Natom, dim);
        C_DGEMM('t', 'n', 3 * Natom, dim, dim, 1.0, B_irr[0], 3 * Natom, evects[0],
                dim, 0, normal_irr[0], dim);

        if (print_lvl >= 3) {
            outfile->Printf("\n\tNormal coordinates (non-mass-weighted) for irrep %s:\n", irrep_lbls[h]);
            eivout(normal_irr, evals, 3 * Natom, dim, "outfile");
        }
```

It looks to me that contrary to what the print functions say, we first compute the Hessian (non-mass weighted), diagonalize it, and then mass-weight and symmetrize the eigenvectors?

@loriab, the normal coordinates that we print out now are not normalized. It's unclear what the Molden requirement are when it comes to mass-weighting or normalization. The Molden Format page does not specific it.

loriab:
Sorry, written hastily. Units in [molden format](http://www.cmbi.ru.nl/molden/molden_format.html) `The atomic coordinates x,y,z and atomic displacements dx,dy,dz are all in Bohr (Atomic Unit of length)` seem to indicate non-mass-weighted. And I thought we got this standardized in 64dc61a2bb4 to solve psi4/psi4#225. I'm not sure that analytic freq are adhering to the nmw print standard, but the freq by E and by G should be, for what's written to output.

psi-rking:
I think the print statements are correct.  The displacements are all done by displace_cart() in fd_misc.cc.  These are in mass-weighted coordinates.  The 'disp_size' is therefore in mass-weighted coordinates.  The Hessian then computed (using energies and disp_size) is also the mass-weighted Hessian.  So "evects" starting at 239 are the eigenvectors of the mass-weighted Hessian.  I think (without thinking about it too long) that the division of "B_irr" coordinates by the square-root of the masses before multiplying is just an ugly way to undo the mass-weighting.

psi-rking:
That said, I can't recall if I've ever compared element-for-element these modes with those from other programs.  Perhaps Lori has.

fevangelista:
@psi-rking: Thanks, I missed the fact that the displacement were in mass-weighted coordinates. That is consistent with dividing by the sqrt of the atomic mass. Thanks for the clarifying this point. I think I'll go ahead and normalize the normal modes. Any objection?

loriab:
I haven't reloaded all my freq familiarity, but I did various comparisons around the time of the thermo update, but [molpro](https://github.com/psi4/psi4/blob/master/tests/freq-isotope/molpro.out#L469-L481) and [psi4](https://github.com/psi4/psi4/blob/master/tests/freq-isotope/output.ref#L4113-L4136) match, and there's other examples in that folder. Psi4 output vibs through both findif routes I'm confident are nmw (through analytic I haven't looked at).

Normalization sounds good to me.

fevangelista:
Rebased, pushed corrections to the docs, and added code to normalize normal modes.

loriab:
Regarding rebasing, on the Commits tag, there should be only your commits. I think the problem arises when you do:

```
# wrong
git pull origin master
git push origin master
```

instead of
```
# right (except that it's good to use a feature branch rather than master)
git push -f origin master
```

[Andy's diagram](http://psicode.org/psi4manual/master/build_obtaining.html#faq-githubworkflow) from the Nov workshop is now up, in case that helps.

fevangelista:
Sorry for messing this PR as well. I promise, next time I'll use `git push -f origin master`.

loriab:
Do you want me to fix it up, or do you want to give it a try? The below should do the trick.

```
# on master
git checkout -b moldennormco
# last commit should be "Fix typo in doc and specify non-mass-weighted/normalized normal modes"
# last commit should be "Fix typo ..." and 
#    first non-you commit should be "Merge pull request #641 " with hash 0c08 
#    (from https://github.com/psi4/psi4/commits/master)
git push origin moldennormco:master
# fails because you're trying to rewrite published history
git push -f origin moldennormco:master
# success
```

fevangelista:
I went ahead and rewrite the history. I think your instructions worked.

doc/sphinxman/source/freq.rst
```diff
@@ -50,7 +50,21 @@ the scenes, :py:func:`~psi4.frequency` is a light wrapper over
 :py:func:`~psi4.hessian` that computes the Hessian then adds a
 thermochemical analysis.
 
-.. autofunction:: psi4.frequency(name [, molecule, return_wfn, func, mode, dertype, irrep])
+Visualization of Normal Modes
+-----------------------------
 
-.. autofunction:: psi4.hessian(name [, molecule, return_wfn, func, dertype, irrep])
+|PSIfour| has the ability to export a Molden file that stores information about
```
loriab: Did the following lines get deleted by accident? Not a show-stopper, I can add them back in later.

```
.. autofunction:: psi4.frequency(name [, molecule, return_wfn, func, mode, dertype, irrep])

.. autofunction:: psi4.hessian(name [, molecule, return_wfn, func, dertype, irrep])
```

fevangelista: These got lost in the merging process. Thanks for spotting it.

Issue #617, opened 2017-02-13
Bug running X2C with basis not included in basisfamily_list


tyzhang1993:
Input file:
```
molecule h2o {
  O
  H 1 0.96
  H 1 0.96 2 104.5
}
set {
  basis roos-ano-dz
  relativistic x2c
}
energy('scf')
```
Error Info:
```
Traceback (most recent call last):
  File "bin/psi4", line 210, in <module>
    exec(content)
  File "<string>", line 28, in <module>
  File "lib/psi4/driver/driver.py", line 447, in energy
    wfn = procedures['energy'][lowername](lowername, molecule=molecule, **kwargs)
  File "lib/psi4/driver/procedures/proc.py", line 1917, in run_scf
    scf_wfn = scf_helper(name, **kwargs)
  File "lib/psi4/driver/procedures/proc.py", line 1241, in scf_helper
    scf_wfn = scf_wavefunction_factory(core.get_option('SCF', 'REFERENCE'), base_wfn)
  File "lib/psi4/driver/procedures/proc.py", line 1035, in scf_wavefunction_factory
    puream=wfn.basisset().has_puream())
  File "lib/psi4/driver/p4util/python_helpers.py", line 62, in pybuild_basis
    key, target, fitrole, other, return_atomlist=return_atomlist)
  File "lib/psi4/driver/qcdb/libmintsbasisset.py", line 631, in pyconstruct
    return_atomlist=return_atomlist)
  File "lib/psi4/driver/qcdb/libmintsbasisset.py", line 753, in construct
    filename = cls.make_filename(basgbs)
  File "lib/psi4/driver/qcdb/libmintsbasisset.py", line 1292, in make_filename
    basisname = basisname.lower()

AttributeError: 'NoneType' object has no attribute 'lower'
```
This happens not only with roos-ano-dz basis, but also when a user-defined basis is used.
This problem is actually caused by line 726~730 in lib/psi4/driver/qcdb/libmintsbasisset.py. After line 730: seek['basis'] = [item for item in tmp if item != (None, None, None)], seek['basis'] have only one entry (None, None, BasisSet.decontract), which will cause basgbs == None at line 753.
Actually, after line 730, seek['basis'] should have an entry ('roos-ano-dz-decon', 'roos-ano-dz', BasisSet.decontract) for the program to run correctly.

One possible solution is that add a guard:
```
if basgbs == None:
      continue
```
before line 753: filename = cls.make_filename(basgbs)

[...]

PR #649, opened 2017-03-15
the fabled basis set pass

loriab:
## Description
All the basis set stuff that simply had to be done for 1.1 to be at least as capable as 1.0

## Todos
* **Developer Interest**
  - [x] Fixes psi4/psi4#617
  - From psi4/psi4#43, fixes
    - [x] 1. `decontract` now py-side and signaled by `-decon`. This was done some time ago.
    - [x] 2. `basis_rel` also now `basis_relativistic`, etc. Also some time ago
    - [x] 3. user-def basis sets have names/handles and are recallable as shown in mints9 and python/mints9. docs corrected and elaborated on this point
    - [x] 7. got taken out last fall, and its replacement has not been added.
    - [x] 12. fixed up the basisset.rst docs
    - [x] 14. covered by comments at 6
  - [x] fixed mints9 and enabled symmetry detection to show through to C-side mol
  - [x] assigned a rough zeta to most built-in basis sets

  - [x] Extends `basis <opt_name> {...}` blocks to PsiAPI via `psi4.basis_helper(name='mybas', key='DF_BASIS_JKFIT', block="""
assign ...
""")`
  - From psi4/psi4#43, fixes
    - [x] 4. didn't try it on a `db()`, but basis blocks aren't anonymous and can be reused at will. Even `basis mybas {....}; energy('hf/mybas')`
    - [x] 6. vastly improved basis printing by always printing file and line origin on each basis set load (called in the driver so no longer mass rebuilding of basissets). separated name into a name field that is the handle by which it can be recalled (that is, the BASIS* keyword value) and a composition that'll look like `aug-cc-pvdz + cc-pvdz + 6-31g`
    - [x] 8. fixed. now if first-preference fitting basis not avail and orbitals basis is _known_ to be >quad-zeta, _won't_ default to def2-qz* fitting sets

## Questions
- [ ] @WDerricotte, there was that decon example in the manual that we agreed was wrong. Can you fix that?
- [ ] PsiAPI folks: getting a lot of these "helper" functionals for processing input blocks: pcm_helper, basis_helper. Shall we settle on a name? Others are "geometry" and "set_options".

## Status
- [x] Ready to go


psi4/driver/p4util/python_helpers.py
```diff
@@ -185,3 +189,84 @@ def pcm_helper(block):
...
+                raise TestComparisonError(message)
+        else:
+            # case with specs separated by [basname] markers
+            for idx in range(0, len(basblock), 2):
+                basstrings[basname(basblock[idx])] = basblock[idx + 1]
+
+        return basstrings
+    anon.__name__ = 'basisspec_psi4_yo__' + cleanbas
+    qcdb.libmintsbasisset.basishorde[name.upper()] = anon
+    core.set_global_option(key, name)
```
amjames: The docstring should mention that the function does this. Also, what do you think about adding a kwarg to control this? Defaulting to True would be fine, but personally, I would prefer if the default was to skip `core.set_global_option()` that way api mode has less "magic" in the background. 
The call to `core.set_global_option` is handy, but I think it's a good idea to avoid implicitly messing with global state *especially* in apimode. If I was to read the docstring for this new function and try it out with something like this:

```python
mymol = # setup molecule, 
# setting options
psi4.core.set_global_option("BASIS","sto-3g") # set my default
psi4.basis_helper("""
…
""", name="custom-bs1")
psi4.basis_helper("""
…
""", name="custom-bs2")
scf_energies = {"default_basis": 0.0, "custom-bs1":0.0, "custom-bs1": 0.0}
scf_energies["deafult_basis"] = psi4.energy('scf', mol=mymol)
psi4.core.set_global_option("BASIS", "custom-bs1")
scf_energies["custom-bs1"] = psi4.energy('scf', mol=mymol)
psi4.core.set_global_option("BASIS", "custom-bs2")
scf_energies["custom-bs2"] = psi4.energy('scf', mol=mymol)

# look at the energies, make table, whatever
```

I may be surprised to see that the entries for "custom-bs2" and "default_basis" are the same, or worse I might not notice if the "custom-bs2" is some very small modification to sto-3g,  and I was expecting the energies to be close.

loriab: Good point about the docstring. I'm fine with adding a kwarg to suppress setting the option. But at least in Psithon, I regard `df_basis_scf mybas {...}` as just an elaborate `set df_basis_scf mybas` that just happens to need multiline strings. And so the function name and arg order of the PsiAPI `basis_helper()` could use some improvements, but should basically behave the same way, just as `geometry(...)` sets active molecule. So I'm in the default `basis_helper(..., set_option=True`) camp.

psi4/share/psi4/basis/def2-qzvpp-ri.gbs
```diff
@@ -542,7 +542,7 @@ Ne    0
 S  1  1.00
  522.238   1.0
 S  1  1.00
- 110       1.0
+ 110.0     1.0
 S  1  1.00
```
andysim: is this just for cleanliness, or is there a bug in the parser?

loriab: A bug, indeed, in that the parser won't accept integers here.

PR #651, opened 2017-03-16
scf: moves read wavefunction file into scratch directory

### LINKED ISSUE #646: Saving SCF npy files in a different directory
It's convenient that the orbitals are saved in [`scf_helper`](https://github.com/psi4/psi4/blob/4e8c34a1744559f7f7714c52338f3f3f430766a5/psi4/driver/procrouting/proc.py#L1361-L1381), but I think it would be convenient to be able to control the directory.

Obviously I could `cd` the whole process, but that feels a little clumsy. Potentially adding an option to scf_helper to control the directory, or adding a `set_writer_file_directory()` to  [`get_writer_file_prefix.cc`](https://github.com/psi4/psi4/blob/40e4680d55f3f342f0fc50f8146873cb3eed318d/psi4/src/psi4/libmints/get_writer_file_prefix.cc) would be cleaner (that defaults to `.`).

dgasmith:
## Description
The Wavefunction `read` command was using the current working directory to read/write its files. This function now properly uses the scratch directory. Closes #646.

## Status
- [x] Ready to go


psi4/driver/procrouting/proc.py
```diff
@@ -1250,7 +1250,9 @@ def scf_helper(name, **kwargs):
     scf_wfn = scf_wavefunction_factory(core.get_option('SCF', 'REFERENCE'), base_wfn)
     core.set_legacy_wavefunction(scf_wfn)
 
-    read_filename = core.get_writer_file_prefix(scf_molecule.name()) + ".180.npz"
+    fname = core.get_writer_file_prefix(scf_molecule.name()) + ".180.npz"
+    read_filename = os.path.join(core.get_environment("PSI_SCRATCH"), fname)
```
loriab: Any reason to prefer envvar `PSI_SCRATCH` to  `core.IOManager.shared_object().get_default_path()`?

dgasmith: Seems more understandable I think. Also, PSIO is probably going to be deprecated sooner or later.

PR #663, opened 2017-03-23
The nightly conda, docs build one.

loriab:
## Description
The stuff needed to get nightly builds working again, this time driven by Py36.

## Todos
* **Developer Interest**
  - [x] Something about my new nightly build script makes it hypersensitive to unicode. So conducted a purge of docs, source (except for comments and @jgonthier), and basis sets.
  - [x] Make sure to return a `str` version from our fake setup.py, no matter if called by py2 or 3.
  - [x] After two reports of Intel 2016.0.1 not building, officially disallowing it.

## Questions
  - [x] @amjames and/or @jgonthier, could you copy the relevant two lines into your `cmake/custom_cxxstandard.cmake` and see if that properly scuppers your 2016.0.1 build attempts?

## Status
- [x] Ready to go


cmake/custom_cxxstandard.cmake
```diff
@@ -6,8 +6,8 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
     endif()
 
 elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 16)
-        message(FATAL_ERROR "ICPC version must be at least 2016!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "16.0.2")
+        message(FATAL_ERROR "ICPC version must be at least 2016.0.2!")
```
andysim: So just to clarify, this is not a check on the minor version number of intel 16, it's really saying that, *e.g.* Intel 15 won't work, right?

psi4/share/psi4/basis/def2-qzvp-ri.gbs
```diff
@@ -3,7 +3,7 @@ spherical
 !
 ! Basis sets obtained from http://www.cosmologic.de/basis-sets/basissets.php  03/17/14
 !
-! C. Hättig; PCCP 7, 59 (2005).
+! C. Hattig; PCCP 7, 59 (2005).
 !
```
andysim: Not important, but I think "Haettig" is the correct way to avoid an umlaut.

cmake/custom_cxxstandard.cmake
loriab: Supposedly [this works](https://github.com/psi4/psi4/blob/master/cmake/autocmake_omp.cmake#L55). So I'm hoping that the check here will allow 2016.0.2 and above and prevent the buggy 2016.0.1. I'll add a note asking for testers.

psi4/share/psi4/basis/def2-qzvp-ri.gbs
loriab: Non unicode H¨attig was evenly divided into `a` and `ae`. I can do a standardization pass at some point.

cmake/custom_cxxstandard.cmake
amjames: Might be useful to print the version found as well, but yes this prevents building with intel 16.1. Nice!

loriab: I'd agree except just about the first thing the CMake config prints is the compiler ID. So that info's there, if needed. Thanks for the confirmation – I'll check the box.

andysim: @loriab My concern isn't the CMake syntax, but I think your check will prevent *any* version of Intel 15 from being usable, unless I'm missing something, when the intent is to really check `major version == 16` and `minor version > 0.02`.  If that makes any sense.

andysim: I just realized that the string you replaced already demands version 16, so it is what you intended.  My bad.  I thought we supported Intel 15.

loriab: Yeah, pybind11 only supported Intel 2015 with an involved hack, so we just never supported it.

PR #664, opened 2017-03-26
typo fix periodictable.py

hokru:
## Description
one less problem in lawrencium calculations ;-)

## Status
- [x] Ready to go


andysim:
Good catch!  Thanks for the fix

PR #665, opened 2017-03-27
Fix various issues with PCM

robertodr:
## Description
Fixes a [build](http://forum.psicode.org/t/crc32-undefined-symbol-at-runtime-when-built-with-pcmsolver-gcc-4-9-4/449/10) and an [execution](http://forum.psicode.org/t/pcmsolver-inp-file-not-found-bug/438/7) issue with PCMSolver.

## Todos
* **Developer Interest**
  - [x] Fix build issue with Zlib. Thanks @loriab for the pull request on the PCMSolver repo.
  - [x] The scratch directory is now unique-ified using UUID. The PCM input parsing step happens in scratch, preventing overwriting of `pcmsolver.inp` and/or `@pcmsolver.inp` files.

## Questions
- [x] The `cavity.off`, `cavity.npz` and `PEDRA.OUT` files are **not** copied back to the execution directory. Is there functinality already in place to salvage them after successful execution?

## Status
- [x] Ready to go

### DIFF
--- a/external/upstream/pcmsolver/CMakeLists.txt
+++ b/external/upstream/pcmsolver/CMakeLists.txt
@@ -1,5 +1,5 @@
 if(${ENABLE_PCMSolver})
-    find_package(PCMSolver 1.1.9 CONFIG QUIET)
+    find_package(PCMSolver 1.1.10 CONFIG QUIET)
 
     if(${PCMSolver_FOUND})
         get_property(_loc TARGET PCMSolver::pcm PROPERTY LOCATION)
@@ -19,7 +19,7 @@ if(${ENABLE_PCMSolver})
 
         ExternalProject_Add(pcmsolver_external
             GIT_REPOSITORY https://github.com/PCMSolver/pcmsolver
-            GIT_TAG v1.1.9
+            GIT_TAG v1.1.10
             UPDATE_COMMAND ""
             CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
                        -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -185,6 +185,15 @@ def set_module_options(module, options_dict):
 def pcm_helper(block):
     """Passes multiline string *block* to PCMSolver parser."""
 
+    # Setup unique scratch directory and move in, as done for other add-ons
+    psioh = core.IOManager.shared_object()
+    psio = core.IO.shared_object()
+    os.chdir(psioh.get_default_path())
+    pcm_tmpdir = 'psi.' + str(os.getpid()) + '.' + psio.get_default_namespace() + \
+        'pcmsolver.' + str(uuid.uuid4())[:8]
+    if os.path.exists(pcm_tmpdir) is False:
+        os.mkdir(pcm_tmpdir)
+    os.chdir(pcm_tmpdir)
     with open('pcmsolver.inp', 'w') as handle:
         handle.write(block)
     import pcmsolver
--- a/psi4/CMakeLists.txt
+++ b/psi4/CMakeLists.txt
@@ -106,7 +106,7 @@ list(APPEND _addons ${_loc})
 message(STATUS "${Cyan}Using libint ${libint_MAX_AM_ERI}${ColourReset}: ${_loc} (version ${libint_VERSION})")
 
 if(${ENABLE_PCMSolver})
-    find_package(PCMSolver 1.1.9 CONFIG REQUIRED)
+    find_package(PCMSolver 1.1.10 CONFIG REQUIRED)
     get_property(_loc TARGET PCMSolver::pcm PROPERTY LOCATION)
     list(APPEND _addons ${_loc})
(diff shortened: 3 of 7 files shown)


external/upstream/pcmsolver/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ if(${ENABLE_PCMSolver})
 
         ExternalProject_Add(pcmsolver_external
             GIT_REPOSITORY https://github.com/PCMSolver/pcmsolver
-            GIT_TAG v1.1.9
+            GIT_TAG v1.1.10
```
loriab: One more spot for 3 total, https://github.com/psi4/psi4/blob/master/psi4/CMakeLists.txt#L109

psi4/driver/p4util/python_helpers.py
```diff
@@ -185,6 +185,15 @@ def set_module_options(module, options_dict):
 def pcm_helper(block):
     """Passes multiline string *block* to PCMSolver parser."""
 
+    # Setup unique scratch directory and move in, as done for other add-ons
+    psioh = core.IOManager.shared_object()
+    psio = core.IO.shared_object()
+    os.chdir(psioh.get_default_path())
+    pcm_tmpdir = 'psi.' + str(os.getpid()) + '.' + psio.get_default_namespace() + \
+        'pcmsolver.' + str(uuid.uuid4())[:8]
```
loriab: The above sequence of commands is probably a candidate for a util function. Adding to developer development list.

loriab:
On the preserving cavity, pedra, etc. files, I think a keep keyword like https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L2444-L2445 to not delete or to copy back would work.

external/upstream/pcmsolver/CMakeLists.txt
robertodr: Ooops that last spot slipped away. Corrected now.

robertodr:
Since these files should always be copied back from scratch, I think it would be enough to have a check on whether PCM was set to true in the input. Would `driver.py` be a good place for that? Should these PCM specific helpers be located in a separate file?

loriab:
Sure they should always be copied back? Usually there's no additional files written w/o user's specific request. (Even `intco.dat` bowed out.) And they're probably not named to associate with an in/out pair, are they (e.g., mytest.in, mytest.out, mytest.pedra)?

robertodr:
Now that you made me think of it, you're right, they do not need to be always copied back. However, I don't think these files are so important that there needs to be a `KEEP` input keyword. I would rather copy them back "by hand" in the input script and the tests can serve as an example for the pattern. Is there a Python function to copy them back from scratch? I see [this](https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/util.py#L436-L485) but then the namespace is "mangled" using UUID and I don't quite know how to get the full path to the execution directory.

loriab:
I don't think we want to register scratch subdirs, so best thing may be to just glob within subdirs of main scratch.

robertodr:
Sorry for taking such a long time on this. @loriab the `input.dat` in `pcmsolver/scf` now shows how to recover files from scratch:
```python
pid = str(os.getpid())
scratch_dir = os.path.join(core.IOManager.shared_object().get_default_path(), 'psi.' + pid + '.pcmsolver')
pcm_save = reduce(lambda l1, l2 : l1 + l2, (glob.glob(t) for t in [scratch_dir + '*/' + x for x in ['cavity.off__' + pid, 'PEDRA.OUT__' + pid, 'cavity.npz']]))
filter(lambda x : shutil.copy(x, psi4.extras.get_input_directory()), pcm_save)
```
Maybe it's an interesting bit for `p4util/util.py`.
The `input.dat` for the `pcmsolver/dipole` test now uses the `perturb_h`, `perturb_with` and `perturb_dipole` parameters. Surprise, surprise, the test is **failing**. After checking the PCM set up, basis, SCF start guess and all other obvious culprits, I found out that the nuclear energy is **different** from the one in the reference file. Do you and/or @andysim and/or @dgasmith have any idea why this could be?

loriab:
Good idea, will make a note of the copy syntax for a developer development exercise.

On the pcmsolver/dipole problem, there were extra fns added to molecule (https://github.com/psi4/psi4/pull/668/files#diff-aac3b6c47032ba056103b4f355643ca7) that are perhaps relevant, though I don't see that the plain NRE should have changed. Also, the old ~`set perturb_with fieldx` is still valid. Any chance that's working while perturb_with dipole not?

andysim:
@robertodr , #668 corrected the nuclear repulsion energy, which didn't feel the effect of a field before (and it should).  I missed the fact that there's a PCM test that uses the `perturb_h` machinery, so I didn't fix the test.  Sorry about that.  [This](https://github.com/psi4/psi4/blob/master/tests/props1/input.dat) test does essentially the same thing that you do in the dipole test.  Because the NRE now feels the applied field, you can just remove the nuclear_dipole terms entirely (that function still exists, but now lives in molecule).  The `perturb_with` syntax will work, but will be deprecated.  Please change it to `perturb_with dipole` and `perturb_dipole [0, 0, $l]` as in the test case that I linked.  Sorry for the confusion!

robertodr:
No problem @andysim It was just, as you say, confusing! The updated `input.dat` is basically [`tests/props1/input.dat`](https://github.com/psi4/psi4/blob/master/tests/props1/input.dat) with CPCM for HF and KS-B3LYP. I will just update the `output.ref` and this one in ready to go :wink:

PR #668, opened 2017-03-29
Perturbed SCF gradients

andysim:
## Description
There are a few bugs and usability issues with calculations in the presence of a dipole field; this PR will address them.

## Todos
* **Developer Interest**
  - [x] Fix and test dipole derivative integrals.
  - [x] Add dipole derivative terms to SCF gradients.
  - [x] Add dipole derivative terms to MP2 gradients.
  - [x] Ensure findif calculations in the presence of a field don't project rotations/translations.
  - [x] Add tests for SCF and MP2 gradients.
  - [x] Fix nuclear repulsion energies and gradients when a dipole field is present.
  - [x] Document method for specifying dipole perturbation.

  - [x] Improve handling of calculations in the presence of a field.

## Questions
- [x] I don't like the fact that we have to set `perturb_h`, `perturb_with` and `perturb_magnitude`.  The biggest problem is that we can only specify x, y, or z perturbations as-is.  I propose something like keeping `perturb_h` for easy toggling of the field, but then we should deprecate `dipole_{x,y,z}` in favor of `dipole`.  Then an option like `perturb_dipole` or similar that would take a 3-vector to specify the magnitudes would be better.  Please weigh in on this.

## Status
- [x] Ready to go


andysim:
My latest fix raises yet another problem: if the user requests 'set scf perturb_h true`, the variable is invisible to `findif`.  Perhaps we should strongly suggest that people don't scope their options in future?

loriab:
Yeah, scoping is not encouraged by example.

The `perturb_` situation is pretty awful, now you mention it. And diple_x, dipole_y aren't even simultaneously specifyable so as to allow a non-axis perturbation. Cfour uses three indep options: xfield, yfield, zfield, but I think your 3-vector suggestion is sensible. And I favor keeping all these kw grouped by `perturb_`.

andysim:
Ok, I think this one is ready for review now.  Thanks for the feedback @loriab

psi-rking:
This looks great, Andy!  Thanks for your speedy help on this.  Reading your code is good training!

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -2820,6 +2822,222 @@ void RDFMP2::form_gradient()
...
+            ylambda = options_.get_double("PERTURB_MAGNITUDE");
+        else if (perturb_with == "DIPOLE_Z")
+            zlambda = options_.get_double("PERTURB_MAGNITUDE");
+        else if (perturb_with == "DIPOLE") {
+            if(options_["PERTURB_DIPOLE"].size() !=3)
+                throw PSIEXCEPTION("The PERTURB dipole should have exactly three floating point numbers.");
+            xlambda = options_["PERTURB_DIPOLE"][0].to_double();
+            ylambda = options_["PERTURB_DIPOLE"][1].to_double();
+            zlambda = options_["PERTURB_DIPOLE"][2].to_double();
+        } else {
```
andysim: I only kept the former as an option so that the update doesn't break existing inputs.  The manual and the code have deprecation warnings about `dipole_y` and `perturb_magnitude` so they should be removed eventually.

andysim:
Good idea about the asymmetric test, @loriab .  I think mintshelper handles the case where the field breaks symmetry, but it would be nice to have a C1 symmetry test that has a [1,1,1] field applied, to test all components of the dipole derivatives in case they get broken at some point (only z components are tested now, for speed, but I have verified the other components manually).  We can think about these in the next phase, where some code (mintshelper?) will provide the matrices of dipole derivatives for @psi-rking to use in the IR intensities.

PR #670, opened 2017-03-30
Analytic RHF-CC2 gradients

ashutoshvt:
## Description
Implements Analytic RHF-CC2 gradients and builds up the 
cc2-densities of UHF and ROHF cases which will become functional
once the corresponding  lambda equations are coded up.

## Status
- [X] Ready to go


andysim:
This is fantastic, Ashutosh.  I think you may have forgotten to activate the test, which can be done by adding cc13d to [this line](https://github.com/psi4/psi4/blob/master/tests/CMakeLists.txt#L40)  in CMakeLists.txt of the tests.  Apart from that, I think removing the 3pt finite difference test would be a good idea (there's nothing the 3pt test would catch that the 5pt finite difference test below wouldn't, unless I'm missing something).  After that, if you could follow [these directions](http://www.psicode.org/psi4manual/master/build_obtaining.html#what-is-the-suggested-github-workflow) to update your branch, I suggest to @loriab , @jturney and @dgasmith that we pull this one in next (sorry for leapfrogging the Simint PR, Lori).

ashutoshvt:
Thanks for the feedback @andysim I have made the changes that you suggested and I think its ready to be merged now.

tests/cc13d/input.dat
```diff
@@ -0,0 +1,25 @@
...
+H 1 R 2 A
+
+R = 1.0
+A = 104.5
+}
+
+set {
+  reference RHF
+  basis cc-pVDZ
+  dertype first
```
loriab: Why setting `dertype`? It really shouldn't ever be used by user, instead being set by fn in proc called by gradient. Is this superfluous? Also, fyi, I think we can stop setting memory in test cases unless significantly more or less than the default of 500mib is needed.

psi4/driver/procrouting/proc.py
```diff
@@ -786,6 +786,30 @@ def select_ccsd_gradient(name, **kwargs):
     else:
         return func(name, **kwargs)
 
+def select_cc2_gradient(name, **kwargs):
```
loriab: I'm a little confused over why do a select fn for cc2 gradient when they can only be generated by one module (cc*) and there isn't one for cc2 energy. "select" fns are generally for the purpose of routing calculations through the code that should take different paths depending on reference, df-ness, etc. I'm perfectly fine with having a select fn for cc2, especially if there'll be further variations on it, and I can fill them in myself (needs a block [here](https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/notes_c.rst), too), I just want to make sure it's deliberate and consistent.

tests/cc13d/input.dat
ashutoshvt: Sure, I can get rid of 'dertype' as it is indeed superfluous. will remove the set memory option too. thanks for the feedback.

psi4/driver/procrouting/proc.py
ashutoshvt: Yup, we don't need this function as of now and I have removed it. I have updated the docs as you suggested. Thanks again!

loriab:
Sorry, I wasn't clear. Either cc2 is under select control, in which case the notes_c section is needed, the select_cc2_gradient fn should come back and a select_cc2 fn should be added. Or, cc2 _not_ under select control, the notes_c can be removed, and the procedures table should direct cc2 calcs to run_ccenergy and run_ccenergy_gradient (or the appropriate non-select fn). It's not a good idea to have cc2 calcs directed toward select_ccsd. Sorry about the confusion.

ashutoshvt:
Sorry for the wrong commit. As you rightly pointed out that this feature is limited to the cc module, I am going to use run_ccenergy_gradient function for cc2 gradient calcs instead of any select functions for now.  Also, I will have to remove the cc2 section from notes_c.

andysim:
Good catch, @loriab .  I completely missed that, but your interventions vindicates our three review strategy.

Issue #661, opened 2017-03-22
parallel cfour with psi4


hokru:
(I feel a bit bad for opening so many issues..but here we go..)
Hi,

In order to make parallel xcfour work with psi4, I needed to remove the `env` parameter the subprocess: 
`retcode = subprocess.Popen([cfour_executable], bufsize=0,stdout=subprocess.PIPE)`
(compare with [original](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/interface_cfour.py#L213))

Otherwise xinitenv (from cfour) would exist with exit code 265 and then psi4 is upset by the non-existing output.

Maybe is has to do with our cfour installation using "bash" and not "sh" in the various cfour init scripts? But env does not change the shell type, so I am confused.

loriab:
No problem on all the issues. Glad you're testing the code.

The `env=lenv` is basically to fulfill the "addons may be in PSIPATH or PATH" promise and to remove envvar noise from the execution environment. [This](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/interface_cfour.py#L142-L148) is pretty much it. My first thought is that there might be some envvar you have set that is important for `xcfour` to run correctly but is getting cut from `lenv`. Maybe `printenv` and see if anything looks likely? We can certainly pass more through if warranted.

hokru:
ah ok, I remember that cfour needs an environmental variable to figure out the number of mpi threads. At the configure step of the source code one sets something like:
`configure [...] --with-mpirun=mpirun -np ${CFOUR_NUM_CORES}`

Not sure if this is just our configuration or if `CFOUR_NUM_CORES` is the default name described in some installation documentation.

I see that `OMP_NUM_THREADS` is handled correctly with `env=lenv`, though I don't see it is being added to `lenv`.

loriab:
From the cfour email chain today, it sounds like `CFOUR_NUM_CORES` is widespread. I'll add it to the whitelisted envvars.

> export CFOUR_NUM_CORES=20
>
> From my experience I can say that using more than 4 processors will not increase performance and even sometimes the worse. Of course that depends also from the kind of computation (optimization, frequency, …).
> 
> Best regards,
> Alberto

hokru:
thank you.

with SSDs 8 processors seem to do well enough with the tremendous amount of I/O in CCSD(T) gradients.

loriab:
I think this was resolved by psi4/psi4#672

PR #672, opened 2017-04-01
much miscellany

loriab:
## Description
Adds minao Br. Update CheMPS2 upstream. Remove Fort checking. Whiteli……st Cfour parallelism variable.

## Todos
* **Developer Interest**
  - [x] Remove copied-in CMake CheckFortranCompilerFlag module now that min cmake is 3.3 which always provides it. Also echo-target, which was more of a devel utility.
  - [x] Fixed the suggest-near-option-match for py3.
  - [x] Whitelist CFOUR_NUM_CORES envvar for Cfour, closing psi4/psi4#661, fix `psi4 -s` not handling relpaths property with Cfour.
  - [x] Pull CheMPS2 from upstream v1.8.3-12, now that the lapack & hdf5 cmake target have been included upstream.
  - [x] From Rob, add Br to cc-pvtz-minao.gbs
  - [x] Adjusted the molecule name parsing in inputparser. Now allow practically any name through the `molecule ... {...}`, then test if valid Py object name, printing error msg if not.

## Status
- [x] Ready to go


psi4/driver/p4util/text.py
```diff
@@ -200,10 +200,10 @@ def print_stderr(stuff):
 def levenshtein(seq1, seq2):
     """Function to compute the Levenshtein distance between two strings."""
     oneago = None
-    thisrow = range(1, len(seq2) + 1) + [0]
-    for x in xrange(len(seq1)):
+    thisrow = list(range(1, len(seq2) + 1)) + [0]
+    for x in range(len(seq1)):
         twoago, oneago, thisrow = oneago, thisrow, [0] * len(seq2) + [x + 1]
-        for y in xrange(len(seq2)):
+        for y in range(len(seq2)):
```
dgasmith: Hmm good catch. Is there anywhere else that we use `xrange`?

loriab: Safe. Only in qcdb/mpl.py and that's nothing to do with Psi4.

PR #678, opened 2017-04-07
Orientationmgr diag

lazaroid:
## Description
This cleans up a matrix before diagonalisation by removing extremely small values arising from numerical accuracy during matrix generation. Diagonalisation of matrices that have off-diagonal elements of ca <10-14 with DSYEV appears to lead to unexpected eigenvalues, which ultimately lead to different DFT grid generation on different processor types that show subtly different behaviour with numerical accuracy.

Ultimately, this enables the dftd3-version test to pass on core i7-5930X processors (all Haswell?) where before it failed due to generation of an odd DFT grid compared with other processors.

## Todos
* **Developer Interest**
  - I don't think this should affect anything other than within OrientationMgr but the main developers will know better!
  - None

## Status
- [ ] Ready to go

### DIFF
--- a/psi4/src/psi4/libfock/cubature.cc
+++ b/psi4/src/psi4/libfock/cubature.cc
@@ -3070,6 +3070,7 @@ OrientationMgr::LMatrix OrientationMgr::RotMatrixFromOneAxis(LVector axis3)
     return Q;
 }
 
+#define EPSILON 1e-14
 // I hate to write wrapper functions, but it's easier this way...
 void OrientationMgr::diagonalize(LMatrix const& M, LMatrix *Q_out, LVector *D_out)
 {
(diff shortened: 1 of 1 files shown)


andysim:
Thanks a lot for you hard detective work to track this down.  I'm happy to merge the PR, but I'd like to dig around and see if I can figure out why the off-diagonals appear in the first place.

lazaroid:
I added a whole load of printfs and compared the output produced on a couple of different processors. I put a fair bit about it on the [forum](http://forum.psicode.org/t/different-dft-grid-behaviour-on-different-machines-very-different-dft-enrgies/460/9?u=laz). The Iq matrix obtains these "nearly zero" off diagonals at ca. line 3400 of cubature.cc in OrientationMgr::OrientationMgr() where the summation involves products of x, y, and z coordinates. For some processors "-a + a = 0" but for others, not quite by the looks of it.

Some of the cartesian coordinates are things like 10-17, so products become extremely small. I've not looked into whether they've already been transformed form the input coordinates or not but they must have been processed in some way to get that sort of value. Also, the value of center was essentially "0.0, 0.0, 1e-17" on one machine but "1e-17, 1e-17, 1e-16" on an i7-5960x. When this gets subtracted off (line 3414), it either has no effect or adds to the non-zero-ness of the matrix elements (although, subtraction from elements with a significant value, such as the diagonals, has a negligible effect).

Should coordinates have values as small as 1e-17? How close to zero is zero?! That's something I don't really have a feel for.

lazaroid:
Exactly why I started trying to track it down: it only occurs on some processors and it shouldn't! I'm just perusing libmints/molecule.cc. In Molecule::inertia_tensor(), inertia tensor elements less that ZERO (#defined as 1e-14) are set to 0.0 before diagonalisation, presumably for similar reasons.

dgasmith:
Thanks @lazaroid for the fix. Interesting that we had the same problem in Molecule before.

psi4/src/psi4/libfock/cubature.cc
```diff
@@ -3070,19 +3070,22 @@ OrientationMgr::LMatrix OrientationMgr::RotMatrixFromOneAxis(LVector axis3)
...
     double** Ip = I.pointer();
-    Ip[0][0]            = M.xx;
-    Ip[1][1]            = M.yy;
-    Ip[2][2]            = M.zz;
-    Ip[1][0] = Ip[0][1] = M.xy;
-    Ip[2][0] = Ip[0][2] = M.xz;
-    Ip[2][1] = Ip[1][2] = M.yz;
+    // Clean matrix by setting tiny values arising from numerical errors to zero
+    // to prevent generation of weird eigenvectors
+    Ip[0][0]            = (fabs(M.xx) < EPSILON ? 0.0 : M.xx);
```
dgasmith: I dont think `EPSILON` will be used anywhere else. Can you just init epsilon as a double when you start setting `Ip`?

andysim:
Ok, I've had a chance to poke around a bit and I have found that I can reproduce the issue on an E5-2680 v3 chip, so your hunch that this affects all Haswells is likely correct.  I found that putting `#pragma optimize("", off )` immediately before `OrientationMgr::OrientationMgr(std::shared_ptr<Molecule> mol)` on 3389 and `#pragma optimize("", on ) ` after that function on line 3418 fixes the issue.  I have a very slight preference for this approach because it's unambiguous, but I'm also fine to go with zeroing offending elements out.  Thoughts anybody?  Either way, this is a very important fix, and we should either merge the current PR or my proposed solution.

andysim:
I tried my suggested fix using GCC and it didn't work.  Neither did [this](https://gcc.gnu.org/onlinedocs/gcc-4.8.4/gcc/Function-Specific-Option-Pragmas.html) or [this](http://odellconnie.blogspot.com/2012/02/controlling-gcc-optimization-with.html) so I'd rather stick with zeroing out offending elements, as in this PR; portability is far more important than worrying about some arbitrary cutoff.

andysim:
Thanks again @lazaroid; this is extremely useful

PR #679, opened 2017-04-17
Minor modification to FISAPT machinery

bwb314:
This PR 1) adds a slightly more user-friendly message to a possible exception raised in the F-SAPT post-analysis script and 2) makes the FISAPT_FSAPT_FILEPATH, FISAPT_FSSAPT_FILEPATH, and FISAPT_PLOT_FILEPATH options work without adding a '/' to the end of the option. 

On 2), this option only worked properly if the '/' is appended to the end of the directory name (example: "FISAPT_FSAPT_FILEPATH myfsapt/"), and this only works if done in a set block. Now, this way of setting the option will work, as well as "FISAPT_FSAPT_FILEPATH myfsapt" inside and outside of a set block.

- [x] Ready to go


dgasmith:
Restarted the timed out test cases. This should be ready to go after the tests pass.

psi4/share/psi4/fsapt/fsapt.py
```diff
@@ -209,14 +209,20 @@ def partitionFragments(fragkeys,frags,Z,Q,completeness = 0.85):
         inds = sorted(range(len(sums)),key=lambda x:-sums[x])
         sum = sums[inds[0]] + sums[inds[1]]
         if sum <= completeness:
-            raise Exception('Orbital %d is not complete over two fragments.' % (a+1))
+            raise Exception('Orbital %d is not complete over two fragments. ' \
+                            'To avoid this error, please try to avoid cutting ' \
+                            'multiple bonds, aromatic rings, etc., in your ' \
+                            'definitions of fragments.' % (a+1))
```
loriab: Line continuations make me nervous. How about something like the below from [SO](http://stackoverflow.com/a/10660443)?

```
s = ("this is a very"
      "long string too"
      "for sure ..."
     )
```

bwb314: Thanks for the suggestion Lori!

PR #681, opened 2017-04-19
Effective Core Potentials

andysim:
## Description
Adds ECP capabilities by interfacing the integrals written @robashaw.

## Todos
* **Developer Interest**
  - [x] ECP SCF energies.
  - [x] ECP gradients by finite differences in Python.
  - [x] Test case(s) and documentation.
  - [x] Make sure frozen core code doesn't add contributions from ECP bearing atoms.
  - [x] Add / document more basis sets, and add ECP info to existing Def2 series (w/help from @loriab).

  - Effective Core Potentials available for SCF energies and gradients, as well as post-SCF energies.
  - Def2 family of basis sets available for all atoms supported by those basis sets.

Followup projects include:
  - Add fully analytic when derivative integrals are available.
  - Add CPHF terms and consolidate the OE derivatives from scfgrad and mp2grad.
  - Try to clean up Python/C++ parsing and processing of basis sets.

## Status
- [x] Ready to go


dgasmith:
Cool! Ill look into this later, I think I patched up Distelli with help from GT folks. Ping me if it goes down again.

fevangelista:
@andysim: I am super excited about this code! I was wondering if you could answer a few questions so that I can start to think about how to write the interface with forte. How do you handle the reduced number of electrons? Do you redefine the atomic charge (Z) or just subtract the number of core electrons? Is this info available from Wavefunction? Let us know if we can be of any help!

CDSherrill:
Changing the nuclear charge is the easiest thing to do.  But it makes me
nervous because now the nuclear charge is not really the nuclear charge,
and others who are not "in the know" could fail to take account of this
when accessing the nuclear charge.

andysim:
@CDSherrill modifying the nuclear charge is the correct thing to do, to be consistent with the ECP formulation, where some of the nuclear charge is folded into the definition of the core.  I have verified this by running other codes with and without ECPs; this changes the nuclear repulsion energy in all cases.  I'll drop a note in the manual that anybody doing advanced stuff should use [this funciton](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/molecule.h#L286) if they want the "real" charge.

@fevangelista sorry this took me some time to get working.  I'll throw some finite difference gradients in there either tonight or tomorrow.  The energies should be all correct, so feel free to play with this.  Wavefunction has an [`ecpbasisset()`](https://github.com/andysim/psi4/blob/ecp/psi4/src/psi4/libmints/wavefunction.h#L312) that you can query to find the [total number of core electrons excluded](https://github.com/andysim/psi4/blob/ecp/psi4/src/psi4/libmints/basisset.h#L262) or [the number of core electrons excluded on a given atom](https://github.com/andysim/psi4/blob/ecp/psi4/src/psi4/libmints/basisset.h#L265).  Between these functions and that I linked above, you should have all the info you need.  Please let me know if you need any extra information / code to get your work done.

psi4/driver/driver_cbs.py
```diff
@@ -1415,8 +1415,11 @@ def cbs(func, label, **kwargs):
     core.set_variable('CBS NUMBER', Njobs)
 
     # new skeleton wavefunction w/mol, highest-SCF basis (just to choose one), & not energy
-    basis = core.BasisSet.build(molecule, "ORBITAL", 'sto-3g')
-    wfn = core.Wavefunction(molecule, basis)
+    basis,ecpbasis = core.BasisSet.build(molecule, "ORBITAL", 'sto-3g')
+    if ecpbasis:
+        wfn = core.Wavefunction(molecule, basis, ecpbasis)
+    else:
+        wfn = core.Wavefunction(molecule, basis)
```
dgasmith: These lines should probably be replaced by the modified `core.Wavefunction.build` function.

psi4/driver/p4util/python_helpers.py
```diff
@@ -75,7 +75,16 @@ def pybuild_basis(mol, key=None, target=None, fitrole='ORBITAL', other=None, pur
...
 
     psibasis = core.BasisSet.construct_from_pydict(mol, basisdict, puream)
-    return psibasis
+    ecpbasis = None
+    if 'ecp_shell_map' in basisdict:
+        ecpbasis = core.BasisSet.construct_ecp_from_pydict(mol, basisdict, puream)
+
+    if key == 'BASIS':
+        # For orbitals basis sets, we need to return ECP also
+        return psibasis, ecpbasis
```
dgasmith: This will break a great deal of code. Is there any way we can simplify this such as adding the ECP basis object to the canonical Basis or only returning the ECP part if requested?

psi4/driver/p4util/python_helpers.py
```diff
@@ -75,7 +75,16 @@ def pybuild_basis(mol, key=None, target=None, fitrole='ORBITAL', other=None, pur
...
+    else:
+        # There is no ECP basis for auxilliary basis sets
+        return psibasis
```
dgasmith: This transparent heterogeneous return could cause issues as well.

psi4/src/export_mints.cc
```diff
@@ -507,6 +511,9 @@ void export_mints(py::module& m)
             def("so_potential", &MintsHelper::so_potential, "docstring").
 
             // One-electron properties and
+            def("ao_ecp", oneelectron(&MintsHelper::ao_ecp), "AO basis effective core potential integrals.").
+            def("ao_ecp", oneelectron_mixed_basis_ecp(&MintsHelper::ao_ecp), "AO basis effective core potential integrals.").
```
dgasmith: Can you expand in the docs exactly how the three basis input should be called?

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -584,13 +771,16 @@ BasisSet::BasisSet(const std::string &basistype, SharedMolecule mol,
     // The unique primitives
     uexponents_ = new double[n_uprimitive_];
     ucoefficients_ = new double[n_uprimitive_];
+    uns_ = new int[n_uprimitive_];
```
dgasmith: Did I miss the corresponding delete call?

psi4/src/psi4/libmints/bessel.cc
```diff
@@ -0,0 +1,178 @@
...
+	K = new double*[N+1];
+	for (int i = 0; i < N+1; i++) K[i] = new double[lMax + TAYLOR_CUT + 1];
+	C = new double[lMax+TAYLOR_CUT];
+	
+	// Tabulate values
+	tabulate(accuracy);
+}
+
+BesselFunction::~BesselFunction() {
+	free(K);
```
dgasmith: Would you mind postfixing internal quantities with an underscore? Such as `K_`.

psi4/driver/p4util/python_helpers.py
andysim: Much of this stuff is the result of me spending a few days trying to understand the python basis set stuff, and failing miserably.  The full test suite passes as-is, so I think I caught the majority of any damage done by my py-side changes.  I don't disagree about stuffing the ECP stuff inside basis set objects if needed, but I'm reluctant to do that without feedback from @loriab, who wrote the py-side infrastructure.  My intent with this PR is to get something in that works, so that we can focus on a proper derivative implementation, with a basis set machinery cleanup pass to follow.  I'm willing to reconsider that approach if you have a better idea.

psi4/src/psi4/libmints/bessel.cc
andysim: I'll do that; QtCreator allows simple refactoring

psi4/driver/p4util/python_helpers.py
dgasmith: For example, I think this will break most of the Psi4NumPy like scripts as the basis is usually built explicitly.

Since we have a bit of time before this goes in (as we are waiting for 1.1), I wouldn't mind cleaning up this particular aspect. Reading through your code (and as you mention) it looks like the ECP info stuff inside a conventional basis object should be fairly straightforward. This may help with the book keeping as well. Of course @loriab would be great to comment on this.

dgasmith:
Rebased in #698!

andysim:
Want me to close this one then?

dgasmith:
Sure, as long as everyone is finished commenting on the source. Id rather not intertwine the LibXC and ECP discussion.

andysim:
Good point.  I'll leave it open until the merge.  I'll work on the changes suggested (fold ECP into basis set) on the libxc branch.

dgasmith:
This was pulled in via #698. Closing.

PR #682, opened 2017-04-19
Fix misplaced regexp flags arguments 

jwilk:
This fixes multiple invalid uses of flags in `re.compile(...).sub()` and `re.split()` calls.

Found using [pydiatra](https://github.com/jwilk/pydiatra).

### DIFF
--- a/psi4/driver/inputparser.py
+++ b/psi4/driver/inputparser.py
@@ -693,8 +693,8 @@ def process_input(raw_input, print_level=1):
     #    check_parentheses_and_brackets(temp, 1)
 
     # First, remove everything from lines containing only spaces
-    blankline = re.compile(r'^\s*$')
-    temp = re.sub(blankline, '', temp, re.MULTILINE)
+    blankline = re.compile(r'^\s*$', re.MULTILINE)
+    temp = re.sub(blankline, '', temp)
 
     # Look for things like
     # set matrix [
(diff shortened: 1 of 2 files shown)


psi4/driver/qcdb/cfour.py
```diff
@@ -49,7 +49,7 @@ def harvest_output(outtext):
     pass_coord = []
     pass_grad = []
 
-    for outpass in re.split(r'--invoking executable xjoda', outtext, re.MULTILINE):
```
dgasmith: Shouldn't `re.pslit(..., flags=re.MULTILINE)` work?

psi4/driver/inputparser.py
```diff
@@ -693,8 +693,8 @@ def process_input(raw_input, print_level=1):
     #    check_parentheses_and_brackets(temp, 1)
 
     # First, remove everything from lines containing only spaces
-    blankline = re.compile(r'^\s*$')
-    temp = re.sub(blankline, '', temp, re.MULTILINE)
+    blankline = re.compile(r'^\s*$', re.MULTILINE)
+    temp = re.sub(blankline, '', temp)
```
dgasmith: Same as below, the function is `re.sub(pattern, repl, string, count=0, flags=0)`. So this should work, unless I misunderstand something. I do not use advanced regex much so I am likely overlooking something.

loriab:
Thanks for investigating the code. Can you confirm that these lines weren't working and that your changes fixed them? I think one'd have to run some mildly exotic Psi4+Cfour to hit those lines.

dgasmith:
@loriab Any thoughts on pulling this in or rejecting it? Since its a bot we are probably not going to get a response on raised questions.

PR #683, opened 2017-04-20
PCM-CCSD in the PTE approximation

robertodr:
## Description
This PR enables CCSD calculations with the PCM in the [PTE (Perturbation-To-Energy) approximation](http://dx.doi.org/10.1063/1.3245400). The CCSD amplitude equations are solved using the solvated Fock matrix and MOs, _i.e._ running SCF with PCM and then using the solvated SCF determinant as the reference.
It boils down to adding the PCM polarization energy in the reference energy used in the CC modules.

## Todos
  - [x] CCSD with the PCM in the PTE approximation.

## Questions
- [x] Accessing the PCM polarization energy of the reference with `Process::environment.globals["PCM POLARIZATION ENERGY"]` is probably not the cleanest solution. Are there alternatives?
- [x] This PR lacks documentation. Where should I add some notes about this?

## Status
- [x] Ready to go

### DIFF
--- a/tests/pcmsolver/ccsd-pte/CMakeLists.txt
+++ b/tests/pcmsolver/ccsd-pte/CMakeLists.txt
@@ -1,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(pcmsolver-ccsd-pte "psi;quicktests;pcmsolver;addon;ccsd;pte")
--- a/doc/sphinxman/source/bibliography.rst
+++ b/doc/sphinxman/source/bibliography.rst
@@ -33,11 +33,11 @@ Bibliography
 |
 
 .. [Gonthier:2016:134106]
-   J. F. Gonthier and C.D. Sherrill, 
+   J. F. Gonthier and C.D. Sherrill,
    *J. Chem. Phys.* **145**, 134106 (2016).
 
 .. [Parrish:2015:051103]
-   R. M. Parrish, J. F. Gonthier, C. Corminboeuf, and C.D. Sherrill, 
+   R. M. Parrish, J. F. Gonthier, C. Corminboeuf, and C.D. Sherrill,
    *J. Chem. Phys.* **143**, 051103 (2015).
 
 .. [Parrish:2014:17386]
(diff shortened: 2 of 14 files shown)


dgasmith:
For Wavefunction objects we suggest moving over to the `variables` attribute. See the [Wavefunction](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.h#L232) and [MP2](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dfmp2/mp2.cc#L180) classes. Ultimately we want all variables to move here and remove the global object.

loriab:
I echo Daniel on using the Wavefunction.variables map for storage of the pcm energy. Eventually, Wfn.variables will be primary and P::e dropped, but for now, make them mirror.

https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/pcmsolver.rst seems a good place for all Psi4+PCMSolver documentation. And maybe mention on the CC page that CC+PCM is available if PCMSolver enabled.

robertodr:
Forgot to mark it as "Ready to go". I suggest this gets squashed if merged in.

tests/pcmsolver/ccsd-pte/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(pcmsolver-ccsd-pte "psi;quicktests;pcmsolver;addon;ccsd;pte")
```
dgasmith: I dont think CCSD is part of the regression test lineup.

andysim: It's not, but I don't think extra labels hurt at all, do they?

loriab: no, they don't hurt. except for quick and smoke, nobody uses the labels except maybe developers for their own convenience.

dgasmith: I use the labels quite extensively for testing. If we have them it should be consistent IMO. A new CCSD tag here would mean that we would need to tag all cases that are CCSD only.

loriab: Ok, this branch could use rebasing, the `ccsd` to `cc` could go in then. Otherwise, could get this merged in as-is, then I can fix it in a later PR.

andysim: Ah, ok.  Was just asking for my own edification, in case there's some CTest standards that I'm not aware of.  Approvals are in place, but I'll punt to one of y'all to figure out how you'd like to rebase and fix the test label.  +1 for the documentation with citations.

PR #684, opened 2017-04-23
Update LICENSE to permit newer GPL

andysim:
## Description

When changing the LICENSE to make it parsable by GitHub (to correctly display our license), @loriab inadvertently removed the "or (at your option) any later version."  I did this on my master branch so that GitHub would parse it, and it seems to still display the license information correctly with the extra text added.

## Todos
  - [x] LICENSE fixed

## Status
- [x] Ready to go


dgasmith:
Thanks for getting this Andy. Looks like a merge or rebase gone wrong!

andysim:
I think @loriab intended to put the GPLv2 text verbatim into LICENSE, but with the text in CITATION indicating that we also permit newer GPLs.  That way, GitHub could parse LICENSE and provide a simple digest to users, but the CITATION code would clarify our intent.  Adding the extra text to LICENSE removes any ambiguity resulting from the mismatch between these files and, it appears, doesn't break GitHub's license parser.

susilehtola:
No this is just wrong.

What is now LICENSE should be called COPYING, and its contents should match
  https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt
which it currently appears to.

LICENSE should contain the stuff that is now in CITATION.

andysim:
Thanks for the info @susilehtola; I'll close and try again.

andysim:
Ok, after @susilehtola's heads-up and some Googling, I think I now have a working solution.  The COPYING file now contains GPLv2 **verbatim** (no mention of newer ones).  I got rid of both LICENSE and CITATION and essentially stuck them into README.md so they're very accessible; the statement that newer GPL versions are allowed is in this README.md file.  Removing LICENSE also forces the GitHub parser to parse CITATION, so that it correctly accredits us with the GPLv2 license.  As you can see, this took a few failed attempts, so this should be squashed when merged.

PR #685, opened 2017-04-25
Fixes an integer overflow in libdpd

lazaroid:
## Description
I tracked down a seg fault during a largish cc2 calculation to an integer overflow in the value of `size` in DPD::buf4_scmcopy(). I think it was triggered because I allocated a large amount of RAM so a huge chunk of data was to be copied in one go rather than in bits. In my case,
`size = rowtot*coltot = 47252*47252 = -2062215792`
Funnily enough, memcpy wasn't very happy being called with a negative size to copy!

It's a simple case of `int*int` being larger than an int!

Changing `size` to a `long int` and adding some casts matches other sections of the file (although perhaps replacing them all with `size_t` might be better?).

## Questions
This fix got my calculation past the point where it caused the seg fault (although it's still running: it might take a few days!).

This bug probably only got triggered by large jobs with a lot of RAM so the in-core copy was performed: there could be similar overflows to track down (similar to that [recently found in the SAPT code](http://forum.psicode.org/t/sapt2-calculation-segmentation-fault-during-exch12-computation/415/6)).

I guess a safer option would be test the values for overflow (I've not thought how it behaves 32 vs 64 bit!).

## Status
- [ ] Ready to go

### DIFF
--- a/psi4/src/psi4/libdpd/buf4_scmcopy.cc
+++ b/psi4/src/psi4/libdpd/buf4_scmcopy.cc
@@ -54,8 +54,8 @@ namespace psi {
 int DPD::buf4_scmcopy(dpdbuf4 *InBuf, int outfilenum, const char *label, double alpha)
 {
     int h, row, col, rowtot, coltot, all_buf_irrep;
-    int nbuckets, incore, n, size;
-    long int memoryd, rows_per_bucket, rows_left;
+    int nbuckets, incore, n;
+    long int size, memoryd, rows_per_bucket, rows_left;
     dpdbuf4 OutBuf;
 
     all_buf_irrep = InBuf->file.my_irrep;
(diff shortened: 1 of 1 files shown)


andysim:
You're gaining quite the reputation for hunting down difficult and obscure bugs.  Thanks a lot!  At first glance, I think you're right about the `size_t` being a good solution, but I'll hold off saying anything definitive about that until I can look at it in a less sleep deprived state.

amjames:
@andysim I agree that `size_t` is probably a better solution, but I will have to take a closer look to say for sure. 

@lazaroid I agree that there are probably more int overflows to be tracked down in DPD, any that you find you are welcome to patch I will go hunting as well.

loriab:
@amjames, @ashutoshvt, should this be accepted as-is, or is there a comprehensive hunt and/or type renaming underway?

ashutoshvt:
@loriab @amjames I was not able to find any other integer overflows yet in libdpd. As mentioned before, I see both `size_t` and `long int` has been used in the files. I am wondering if we decided on the use of `size_t` consistently throughout?

loriab:
@jturney went on a `size_t` campaign a couple years ago, I think. It may have been confined to libmints.

Shall we agree on pulling this in now, as it fixes a known and seemingly solitary issue, and punting any more radical re-typing-for-consistency changes past 1.1?

andysim:
Yup, we should pull this one in now because it won't get in the way of any future size_tifying

PR #686, opened 2017-04-26
libefp bump, cbs refinements, GH workflow expansion

loriab:
## Description
Mostly xtpl refinements, plus some misc.

## Todos
* **Developer Interest**
  - [x] Fixed the typo @fevangelista noticed on the GitHub workflow slide (it was definitely mine, not Andy's, in my zeal to use https). Added the link to this on the front page that CDS wanted. Added a do/don't git pull/push slide
  - [x] Added ability to forward the extrapolation scheme from a method/basis spec
  - [x] Tested and added docs and test for user-specified extrapolation functions. Added refs for xtpl fns.
  - [x] Changed import of xtpl fns to `from driver_cbs import *` in inputparser after they had previously been named explicitly. This should be fine b/c the targeted ones are the only fns in file not beginning with underscore. This way anyone can add an extrapolation fn to driver w/o having to know to edit inputparser.
  - [x] Bump libefp source to upstream tag 1.4.2
  - [x] Fix more int/float in basis set files that wouldn't get read correctly by parser. In particular, def2 RI fitting for Mg.

## Status
- [x] Ready to go


psi4/share/psi4/basis/def2-qzvpp-ri.gbs
```diff
@@ -718,7 +718,7 @@ S  1  1.00
 S  1  1.00
  0.0665549  1.0
 P  1  1.00
- 217       1.0
+ 217.0     1.0
 P  1  1.00
```
andysim: We should fix the the parser in the the next pass through the py-side basis set machinery.  I also modified this file for the ECP code, so will be on the lookout for similar issues.

loriab: Agreed. It's on the develdevel projects list.

Issue #688, opened 2017-04-27
FYI: Psi4 license change


loriab:
Since Psi4:
  * is primarily a shared library / Python module (to C++/Py folks, respectively) rather than an executable (`bin/psi4` is light script loading Python module);
  * is not necessarily at the top of the call stack, and we've heard from projects who would like to import us;
  * is benefiting from the non-attaching and permissive licenses of numerous external projects to supplement Psi4's capabilities, and we want to return the favor,

**we plan to change our license from GPL-2.0+ to LGPL-3.0+, effective by the the 1.1 release.**

EDIT: now targeting LGPL-3.0, not +, with PR psi4/psi4#690

The current and historical core developers have signed off on this. The historical peripheral developers are discussing this without any initial objections. We now want to get feedback from the GitHub contributor, user, and packaging community (or anyone not in e-mail communication with the core developers). So please reply with any concerns, constraints, preferences, etc.

##### Dependencies licensing situation
  * required library dependencies (current & anticipated): PSFL, LGPL-3.0+, BSD-3-Clause-LBNL, BSD-3-Clause 
  * optional library dependencies: BSD-2-Clause, BSD-3-Clause, GPL-2.0+, LGPL-3.0+
  * optional executable dependencies: GPL-1.0+, Custom
  * Psi4 plugins: GPL-2.0+, GPL-3.0+
  * conda-wise, we'll build an alternate conda package that doesn't link any GPL libraries or plugins

##### Queries
  * Psi4 is a Python module and hence a dynamic library. I don't know right off how to even build a `psi4.a`. With that in mind, does anyone care about license terms for static linking?
  * Any concrete problems that would be solved by a permissive license instead of LGPL?
  * Anyone care about sublicensing?

j3mdamas:
Is there a release date for 1.1? :)

Cc @mj-harvey

loriab:
Thanks, @j3mdamas, I had meant to cc mj-harvey and @susilehtola and forgot.

Target release is probably our regional QC conference ~May 18. We're currently sending back responses to reviewers of the accompanying paper. Todo items are: building Mac plugins from conda distribution, updating some add-on integrations we promised in the paper, paper, relicensing.

susilehtola:
Well, I'm okay with LGPL. But I'm not 100% sure that'd be the right license for psi4, ref.
  https://www.gnu.org/licenses/why-not-lgpl.en.html
What are the other projects that would *need* psi4 to be LGPL instead of GPL?

[...]

PR #690, opened 2017-05-01
license change

loriab:
## Description
Relicense from GPL-2.0+ to LGPL-3.0

## Todos
* **Developer Interest**
```
find psi4/ -type f -exec sed -i '' 's/GNU General/GNU Lesser General/g' {} \;
find psi4/ -type f -exec sed -i '' 's/[Tt]his program/Psi4/g' {} \;
find psi4/ -type f -exec sed -i '' '/(at your option)/d' {} \;
find psi4/ -type f -exec sed -i '' 's/; either version 2 of the License, or/, version 3./g' {} \;
find psi4/ -type f -exec sed -i '' $'s/ * Psi4 is free software/ This file is part of Psi4.\\\n *\\\n * Psi4 is free software/g' {} \;
```
  - [x] Relicense from GPL-2.0+ to LGPL-3.0

## Status
- [x] Ready to go

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,281 +1,622 @@
-                GNU GENERAL PUBLIC LICENSE
-                   Version 2, June 1991
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
- Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
- 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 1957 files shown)


mbanck:
Uhm, was that checked by legal? As I understand it, PSI does not mandate copyright assignments, so you would need to be sure that every developer ever contributing code to PSI4 (and in extension PSI3-retained code, which I assume is also relicensed?) either agreed to the license change or had their copyright handled by their research institution, which in turn agreed to the license change.

Don't get me wrong, I don't have anything against the LGPL, but relicensing larger projects is usually multi-year effort.

mbanck:
Ok, I see this is being discussin in issue #688 and you have a sign-off from all the past core contributors. In that case, I think you're fine, sorry for the noise.

PR #695, opened 2017-05-09
Use upstream libint1

loriab:
## Description
Updates Libint 1.2.0 @ evaleev/libint <-- libint 1.1.6 @ psi4/libint, evaleev/libint#77

## Todos
* **Developer Interest**
  - [x] All your Psi4 build will start building Libint again b/c wrong version detected until your pre-compiled libints get updated.
  - [x] Now using github.com/evaleev/libint (branch v1) rather than our own github.com/psi4/libint after pushing some CMake build details upstream. The evaleev repo didn't have CMake before v1.2.0, so that version is now required. Also, the project name changed from `libint` to `Libint` and CMake cares about this, so Psi4 repo and libint repo will have to be carefully matched for a bit.

## Status
- [x] Ready to go – Go ahead and review, but I want to look at travis/distelli before quite ready to go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -150,7 +150,7 @@ install:
 - conda config --set always_yes yes --set changeps1 no
 - conda update -q conda
 - conda info -a
-- conda create -q -n p4env python=$PYTHON_VER numpy cmake dftd3 libint=1.1.6 libxc gdma libefp -c psi4 -c psi4/label/dev -c psi4/label/test
+- conda create -q -n p4env python=$PYTHON_VER numpy cmake dftd3 libint=1.2.0 libxc gdma libefp -c psi4 -c psi4/label/dev -c psi4/label/test
 - source activate p4env
 - conda list
 before_script:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 #    - PYTHON_INCLUDE_DIR "Path to the python include files (e.g., /path/to/include/python2.7)"
 #    - SPHINX_ROOT "Root directory for Sphinx: 'bin/sphinx-build' (or similar) should be in this dir."
 #
-#    For any ${AddOn} of: ambit, CheMPS2, dkh, libefp, erd, gdma, libint, PCMSolver, pybind11, simint
+#    For any ${AddOn} of: ambit, CheMPS2, dkh, libefp, erd, gdma, Libint, PCMSolver, pybind11, simint
 #    - CMAKE_PREFIX_PATH "Set to list of root directories to look for externally built add-ons and dependencies
 #      (e.g., /path/to/install-libint;/path/to/install-gdma where exists /path/to/install-libint/lib/libderiv.a)"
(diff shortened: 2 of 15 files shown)


loriab:
Drat, didn't think about `MAXNODE` differences btwn repos, and it only shows up `MAX_AM_ERI` >=7. Anyone ever need to reset it higher than the psi4 settings? If so, I can make it configurable.

```
# psi4/libint
/libint/src/libint_compiler/emit_order.c:#define MAXNODE 20000
/libint/src/libderiv_compiler/emit_deriv12_managers.c:#define MAXNODE 20000
/libint/src/libderiv_compiler/emit_deriv1_managers.c:#define MAXNODE 20000
```

```
# evaleev/libint
/libint/src/bin/libint/emit_order.c:#define MAXNODE 20000
/libint/src/bin/libderiv/emit_deriv12_managers.c:#define MAXNODE 10000
/libint/src/bin/libderiv/emit_deriv1_managers.c:#define MAXNODE 5000
/libint/src/bin/libr12/emit_gr_order.c:#define MAXNODE 3000
/libint/src/bin/libr12/emit_grt_order.c:#define MAXNODE 20000
```

susilehtola:
@loriab I guess you should ask Ed about that. But, I believe this affects only some static allocation in the library generator (not the generated library!), so changing the defaults to handle any MAX_AM_ERI shouldn't make that big of a difference..?

loriab:
@susilehtola Right, I was just trying to work out the settings to get an AM=8 build, which is as high as I ever expect to go. This mostly (for me) affects the conda build (which I can patch), but it will also place a hard limit on the AM that Psi4 can build internally. I suppose I should look at your Fedora build to see what's useful. I agree that I don't see why a large constant in the generator should harm the efficiency of the built libraries.

I just wish I would have thought of this issue _before_ declaring everything was ready to go this morning.

susilehtola:
So the patch I use in Fedora is
http://pkgs.fedoraproject.org/cgit/rpms/libint.git/tree/libint-1.1.5-maxnode.patch

which makes building with options
 --with-libint-max-am=10 --with-libint-opt-am=6 \
 --with-libderiv-max-am1=6 --with-libderiv-max-am2=5 \
 --with-libr12-max-am=5 --with-libr12-opt-am=4
work. I guess I haven't tried playing with the libderiv max am's too much...

loriab:
Update:
* There'll be another commit to this to fix some docs. Today, there's 1.2 libint in the conda channels, so we'll be able to look at travis/distelli built on that rather than the internal builds that the above reflects.
* Thanks for the config params, @susilehtola. The CMake ties libderiv AM to libint AM, so you may not have hit some of the problems. What I'm seeing now is that psi4/libint allows an AM=8 build (libderiv AM = libint AM - 1), but evaleev/libint/v1 segfaults in the libderiv generator execution. So I'm investigating the differences.

susilehtola:
@loriab does AM here refer to max am or opt am?

loriab:
@susilehtola , AM is `MAX_AM_ERI` the CMake variable. Its translation to the quantities that appear in libint.h and libderiv.h is [here](https://github.com/evaleev/libint/blob/v1/README_CMake.md#building)

loriab:
Note to future: AM issue solved via https://github.com/evaleev/libint/pull/78 for Libint 1.2.1

PR #635, opened 2017-03-03
Threads MintsHelper

dgasmith:
## Description
Adds threading to the MintsHelper object. Notably we no longer build SO integrals directly in the SO basis, but build AO and transform AO->MO. This appears to be about as efficient as the former when threading, we can revisit this if the potential integrals can be improved.

Benchmark is a carbon chain in the cc-pvdz basis set, tuned to compute about a million basis functions per row.

I am separating this out from #592 which will wait for further tests.

```
# Original - 1 thread
   nbf  ntrial  ao_overlap  ao_kinetic  ao_potential  so_overlap  so_kinetic  so_potential
0   28    1275    0.002474    0.000240      0.000936    0.000253    0.000311      0.001072
1   98     104    0.005268    0.002996      0.031417    0.002182    0.002873      0.031835
2  168      35    0.007805    0.007087      0.149304    0.005561    0.007733      0.150582
3  238      17    0.019714    0.014981      0.415791    0.010558    0.014752      0.418562
4  308      10    0.029242    0.023728      0.890593    0.017159    0.023903      0.938164
5  378       6    0.038581    0.034306      1.624236    0.023958    0.034886      1.637466
6  448       4    0.058684    0.050234      2.687780    0.033669    0.048659      2.867169
7  518       3    0.073995    0.067988      4.212363    0.043464    0.063137      4.170965

# New algorithm - 1 thread
   nbf  ntrial  ao_overlap  ao_kinetic  ao_potential  so_overlap  so_kinetic  so_potential
0   28    1275    0.002221    0.000152      0.000583    0.002632    0.000274      0.000731
1   98     104    0.004644    0.001872      0.016447    0.004653    0.002240      0.016920
2  168      35    0.005474    0.004001      0.076230    0.006461    0.004780      0.078292
3  238      17    0.007382    0.006899      0.210317    0.010013    0.008758      0.214130
4  308      10    0.011478    0.011354      0.449083    0.015937    0.015050      0.452936
5  378       6    0.017539    0.016836      0.817148    0.025508    0.022129      0.825095
6  448       4    0.022457    0.023191      1.350382    0.028970    0.031135      1.363716
7  518       3    0.029087    0.030661      2.079129    0.040195    0.042085      2.086033

# New algorithm -  6 threads
   nbf  ntrial  ao_overlap  ao_kinetic  ao_potential  so_overlap  so_kinetic  so_potential
0   28    1275    0.001724    0.000046      0.001282    0.001815    0.000171      0.001468
[...]

PR #699, opened 2017-05-10
Fixed issue where dovirt is undefined if dovirtual is passed in as a …

mzott:
…keyword argument.

## Description
When `dovirtual` is passed in as a keyword argument, `dovirt` is never assigned. Assigning `dovirt = dovirtual` allows the methods that use `dovirt` as an argument later to function correctly.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1857,8 +1857,8 @@ def molden(wfn, filename=None, density_a=None, density_b=None, dovirtual=None):
 
         if density_b:
             NO_Rb = core.Matrix("NO Beta Rotation Matrix", nmopi, nmopi)
-            NO_occa = core.Vector(nmopi)
-            density_b.diagonalize(NO_Ra, NO_occa, core.DiagonalizeOrder.Descending)
+            NO_occb = core.Vector(nmopi)
+            density_b.diagonalize(NO_Rb, NO_occb, core.DiagonalizeOrder.Descending)
             NO_Cb = core.Matrix("Cb Natural Orbitals", nsopi, nmopi)
             NO_Cb.gemm(False, False, 1.0, wfn.Cb(), NO_Rb, 0)
 
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1842,6 +1842,9 @@ def molden(wfn, filename=None, density_a=None, density_b=None, dovirtual=None):
     if dovirtual is None:
         dovirt = bool(core.get_option("SCF", "MOLDEN_WITH_VIRTUAL"))
 
+    else:
+        dovirt = dovirtual
+
     if density_a:
         nmopi = wfn.nmopi()
         nsopi = wfn.nsopi()


loriab:
Good catch and fix, @mzott.

The real question, though, is how did we end up with both a kwarg and an option doing the same thing. The option wins, esp. as its [used elsewhere](https://github.com/psi4/psi4/blob/da39266fcb0b84203cdd9a3935e6e98600a688a4/psi4/driver/procrouting/proc.py#L1353). Is there a good use case for using the kwarg instead of the option?

loriab:
@amjames, were you looking into some aspect of `molden()` recently?

amjames:
@loriab, I was working in here recently, but related to other stuff. I don't know what the option does for other modules, but I think the kwarg is a good way to go for this kind of function. 

As far as a use case goes I am working to get all EOMCC root densities into wfn.arrays, and if you wanted to generate molden files for different states some with `dovirtual = True` some with `dovirtual = False` going back to options everytime would be a pain.

They can co-exisit, but I would say it makes more sense for the kwarg to take precedence if passed for this case at least.

mzott:
I'm looking around in the same area, and I think it might be a nice option to allow natural orbital occupancies to be gathered from the SCF wavefunction when the densities are requested as keyword arguments. In the current code, if you specify `density_a`, the occupancies are taken from the `core.Wavefunction.nmopi` which is the number of molecular orbitals per irrep. In the `write.cc` file in `libmints` which is called when `set print mos` and `set molden_write True` are set, the occupancy is reported correctly. If you use the molden interface directly, I think you would simply need to compute the reference wavefunction first, return this wavefunction, and pass it into molden with a new kwarg `scf_wfn` which would then use `core.HF.occupation_a()` and `core.HF.occupation_b()` to get the correct occupations. Then you could pass the SCF wavefunction into the post-SCF methods.

amjames:
@mzott When you use the python side molden driver passing `density_a` or `density_b`it will generate molden files for the natural orbitals, the occupancies of the NOs will be determined at the same time as the natural orbitals are computed (via diagonalization of the provided density matrices). When densitites aren't provided the molden driver is going to generate a file with the canonical molecular orbitals and take the occupations from the `occupation_a/b` functions.

In both cases the same method you mention in write.cc is used to write the data to file. The method and the class it belongs to are exposed python side.

mzott:
@amjames Sorry for not being more clear - the output generated by the different calls to `MoldenWriter` is different if you directly interact with the interface via `molden()` versus the `set print mos` route. I have attached two examples. `test.molden` is output generated by using the `molden()` call, and `test.default.22773.molden` is generated using the `set print mos` route (I added to .txt extension to the attachments so that Github would upload them). Per the molden example in the manual (example 4 [](http://www.psicode.org/psi4manual/master/molden.html)), I passed in the densities from the returned wavefunction. I'm not sure if the DF-MP2 generates densities (I would think it needs to in order to do the density fitting), but if not the SCF call certainly generates a density which should be available in the wavefunction if no MP2 density is generated. Below is my input.

> E, wfn = energy('mp2/aug-cc-pvdz', properties=['dipole'], return_wfn=True)
> 
> Da_so = wfn.Da()
> Db_so = wfn.Db()
> 
> Da_mo = Matrix.triplet(wfn.Ca(), Da_so, wfn.Ca(), True, False, False)
> Db_mo = Matrix.triplet(wfn.Cb(), Db_so, wfn.Cb(), True, False, False)
> 
> molden(wfn, 'test.molden',  density_a=Da_mo, density_b=Db_mo, dovirtual=True)
> 

[test.molden.txt](https://github.com/psi4/psi4/files/992976/test.molden.txt)
[test.default.22773.molden.txt](https://github.com/psi4/psi4/files/992977/test.default.22773.molden.txt)

The energies end up being set equal to the occupancies (which are incorrect - noninteger and in some instances in the thousands). I think this is due to the call to `MoleculeWriter.writeNO`. I did not find this call in `writer.cc`; all I could find was a header file. It does not appear to take energies in as arguments, so I'm not sure why/where from it is passing energies to the molden output.

I don't understand all of the procedure in going from the density to the occupancies; why is the number of MOs per irrep used as the rotation matrix in `NO_Ra = core.Matrix("NO Alpha Rotation Matrix", nmopi, nmopi)`?

amjames:
@mzott I think I see where you are getting tripped up. 

When you pass density matrices to the (python side) molden driver function it will put Natural Orbitals in the molden file that is generated. The line you reference is not generating the rotation matrix it is only initializing the matrix object where the rotation matrix is going to be stored. 
The line: 

`density_b.diagonalize(NO_Rb, NO_occb, core.DiagonalizeOrder.Descending)`

I where the rotation matrix is actually computed. This method computes the rotation matrix and the occupations as eigenvectors/eigenvalues of the MO density matrix. 

When you use the option `set print mo` the molecular orbitals are put into the molden file. The natural orbitals are never computed. If you want to write the molecular orbitals into a molden file via the driver function you shouldn't pass the `density_a` or `density_b` kwargs.

If there is a reference to the `MoldenWriter::writeNO` function anywhere I would check your version is up to date, since that function was completely removed, this change was tacked on to a pretty larger PR #635.

dgasmith:
@loriab Its the war between our options and kwargs class. The option may not be visible while the kwarg is when calling that function. Its a local override to a global option which I think we support in general?

loriab:
@dgasmith Yes, I was alarmed at first to see option/kwarg to do the same thing. But I can see the kwarg use case now, and I'm ok with it. Esp. now that Mike has seen to it that it's read.

PR #700, opened 2017-05-11
Be able to dump CI vectors into file.

susilehtola:
## Description
This pull request makes it possible to dump CI vectors into files for subsequent analysis with other programs.

## Status
- [ ] Ready to go


susilehtola:
There is one slight problem, though. The orbitals appear to be blocked by symmetry group, so they don't come out in energy ordering. I would need to modify the bitstring part of the dump_vec routine to map the bitstrings into Aufbau order.

I couldn't find with a quick look if there's already something that would give me a mapping between detci's orbital numbering vs. the energy ordering. I note that I could build the mapping from calcinfo's scfeigval members, but is there a cleaner way to do this?

dgasmith:
Doesnt look like there is a function that lays around. Prototype code below:

```
        std::vector <std::tuple<double, int, int>> order;
        for (int h = 0; h < nirrep_; h++) {
            for (int i = 0; i < (int) positions[h].size(); i++) {
                order.push_back(std::tuple<double, int, int>(epsilon->get(h, positions[h][i]), i, h));
            }
        }

        std::sort(order.begin(), order.end(), std::less < std::tuple < double, int, int > > ());
```

Can you also add a test for this?

susilehtola:
I'm not sure how this could be tested, since the dumps are pretty lengthy.

loriab:
I don't know how detailed the FCIDUMPs are to know if this'd be easy or hard, but would a `compare_fcidump` like [compare_cubes](https://github.com/psi4/psi4/blob/da39266fcb0b84203cdd9a3935e6e98600a688a4/psi4/driver/p4util/util.py#L364) as used in https://github.com/psi4/psi4/blob/master/tests/cubeprop/input.dat be appropriate?

susilehtola:
So the dump file just contains the CI vector coefficients and bitstrings. The problem is that due to finite numerical precision the ordering of the vectors may change, as several configurations may have the same coefficient. I can probably write a check for that but it'll easily be O(N^2) so I'll have to restrict to rather small calculations.

susilehtola:
Turns out the ordering is still not correct. It appears the orbitals are in some jumbled "CI ordering", which is not the Pitzer i.e. symmetry blocked ordering.

What is the global orbital index of ```stralp->occs[k]```? ```CalcInfo_->order``` appears to contain a translation from CI order to Pitzer order, but I'm having trouble making sense of its contents...

susilehtola:
OK, this should be working 100% now. The output looks correct on all the systems I've run so far, and I've  added tests.

I initially tried to use H2O, but it seems there's something weird going on with the CI wave functions there. Namely, when the CI wave function appears to flip its overall sign between the computed and the reference wave function, there are some configurations that *do not* flip their sign. I guess this must be some weird thing to do with symmetry, in that the wave function can individually flip signs within symmetry subblocks...?

susilehtola:
Actually, it seems the ordering is *still* wrong. Comparing coefficients with and without symmetry, there are huge differences. I'm at my wit's end...

So, for a simple test case
```
molecule {
H 0.0 0.0 0.0
H 0.0 1.0 0.0
H 0.0 0.0 1000.0
H 0.0 1.0 1000.0
}

set basis 6-31G
scf_e, scf_wfn = energy('s_x', return_wfn=True)

set reference rhf
set r_convergence 1e-10
set dump_vectors true
energy('fci', ref_wfn=scf_wfn)
```
when run without symmetry, the 00000202 configuration turns up with zero weight. When symmetry is used, however, the configuration gets an amplitude of 5.949876123577e-04...

dgasmith:
@CDSherrill Any thoughts on this?

loriab:
Hi @susilehtola, I'm sorry it doesn't look like we'll get to this in 1.1. We're busy preparing for a conference later this week, but I'll alert @CDSherrill in person that this could do with a look after the furor dies down. Do you have any known barrier-to-1.1 issues? The plan is to tag later today.

susilehtola:
@loriab not a problem. Just a moment, I'll try updating the Fedora package.

fevangelista:
Just out of curiosity, do the orbital break symmetry (are localized) in the C1 computation?

susilehtola:
@fevangelista they don't break symmetry, but they're also localized. With symmetry on, the orbitals are Ag and B1u. That is, one MO is positive on both molecules, while the other is positive on one and negative on the other. That's just the bonding and antibonding superpositions of the bonding orbital in the H2 molecule.

susilehtola:
@loriab I was able to port the patches to the newest snapshot, good to go.

fevangelista:
Then I suspect that the degeneracy of the orbitals might be the cause of the problem. If the orbitals are degenerate then any arbitrary rotation would give you the same SCF energy but different CI wave functions. Am I missing something?

loriab:
very good, @susilehtola, thanks for checking! There's nothing substantive yet to go in.

susilehtola:
That's certainly true, but the expansion I get with the current code doesn't make sense for other reasons into which I won't go here.

susilehtola:
Ew, looks like I managed to destroy the commit history. Closing this ticket and opening up a new, cleaner one.

PR #702, opened 2017-05-11
Fix DKH

jturney:
## Description
Fixes the DKH interface in Psi4.

## Todos
  - [X] Test case added comparing to a Molpro calculation.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/libmints/mintshelper.cc
+++ b/psi4/src/psi4/libmints/mintshelper.cc
@@ -506,7 +506,7 @@ SharedMatrix MintsHelper::ao_overlap()
     one_body_ao_computer(ints_vec, overlap_mat, true);
 
     // Is this needed?
-    overlap_mat->save(psio_, PSIF_OEI);
+//    overlap_mat->save(psio_, PSIF_OEI);
     return overlap_mat;
 }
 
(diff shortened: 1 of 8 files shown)


psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -619,9 +618,25 @@ SharedMatrix MintsHelper::ao_dkh(int dkh_order)
     H_dk->subtract(Torig);
 
     H_dk->set_name("AO-basis DKH Ints");
-    //H_dk->print();
+//    H_dk->print();
+
+    // DKH is solved in the decontracted basis ... transform to the contracted.
+
+    SharedMatrix S_inv = S->clone();
+    S_inv->general_invert();
```
dgasmith: Any common way around this? Seems like it could become unstable quickly.

tests/dkh/molpro-2order/input.dat
```diff
@@ -0,0 +1,23 @@
...
+relativistic dkh
+dkh_order 2
+print 2
+scf_type pk
+}
+
+e = energy('scf')
+
+print_variables()
+compare_values(-128.66891610, e, 6, '2nd order vs Molpro')  #TEST
```
dgasmith: Woo tests!

psi4/src/psi4/libmints/mintshelper.cc
andysim: there's an SVD-based pseudoinverse in Matrix that could be used for singular cases

PR #706, opened 2017-05-16
Fixing integral overflow problem in (T) energy code

bozkaya:
## Description
The integral overflow problem in the (T) energy code was solved. 

## Todos
* **Developer Interest**
  - [ ] Feature1
  - [ ] Feature2

## Questions
- [ ] Question1

## Status
- [x] Ready to go


bozkaya:
I know that I do not need to use "long int" for "naoccA" and "navirA" variables. However, there are many variables (eg. V^3 type) that requires the usage of long int. Instead of changing specific variables one by one, I have just applied a "find/replace".

You are right that size_t would be a better solution. But long int is also okay as it is used in the ccenergy and fnocc modules.

Further, in the near future I will replace ccsd_triples.cc file with a newer version which includes (T) gradients. At that stage I will consider to use the "size_t" instead of "long int". Hence, I think that the present change is  okay.

PR #711, opened 2017-05-18
Devdevel bugs

dgasmith:
## Description
Adds more CC variables and fixes docstring.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/export_wavefunction.cc
+++ b/psi4/src/export_wavefunction.cc
@@ -67,7 +67,7 @@ void export_wavefunction(py::module& m) {
         .def(py::init<std::shared_ptr<Molecule>, std::shared_ptr<BasisSet>>())
         .def(py::init<std::shared_ptr<Molecule>, std::shared_ptr<BasisSet>,
                       std::shared_ptr<BasisSet>>())
-        .def("reference_wavefunction", &Wavefunction::reference_wavefunction, "docstring")
+        .def("reference_wavefunction", &Wavefunction::reference_wavefunction, "Returns the reference wavefunction.")
         .def("set_reference_wavefunction", &Wavefunction::set_reference_wavefunction, "docstring")
         .def("shallow_copy", take_sharedwfn(&Wavefunction::shallow_copy),
              "Copies the pointers to the internal data.")
--- a/psi4/src/psi4/ccenergy/ccenergy.cc
+++ b/psi4/src/psi4/ccenergy/ccenergy.cc
@@ -341,6 +341,13 @@ double CCEnergyWavefunction::compute_energy()
         update();
         checkpoint();
     }  // end loop over iterations
+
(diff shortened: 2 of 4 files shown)


jturney:
Woot!

psi4/src/psi4/ccenergy/ccenergy.cc
```diff
@@ -341,6 +341,13 @@ double CCEnergyWavefunction::compute_energy()
         update();
         checkpoint();
     }  // end loop over iterations
+
+    // DGAS Edit
```
andysim: You probably didn't mean to leave this in here

psi4/src/psi4/ccenergy/ccenergy.cc
```diff
@@ -341,6 +341,13 @@ double CCEnergyWavefunction::compute_energy()
...
+    Process::environment.globals["CC T1 DIAGNOSTIC"] = moinfo_.t1diag;
+    Process::environment.globals["CC D1 DIAGNOSTIC"] = moinfo_.d1diag;
+    Process::environment.globals["CC NEW D1 DIAGNOSTIC"] = moinfo_.new_d1diag;
+    Process::environment.globals["CC D2 DIAGNOSTIC"] = moinfo_.d2diag;
+
```
amjames: These should be duplicated into `ref_wfn.variables` since `P::e.globals` is going to be deprecated in the future.

psi4/src/export_wavefunction.cc
```diff
@@ -67,7 +67,7 @@ void export_wavefunction(py::module& m) {
         .def(py::init<std::shared_ptr<Molecule>, std::shared_ptr<BasisSet>>())
         .def(py::init<std::shared_ptr<Molecule>, std::shared_ptr<BasisSet>,
                       std::shared_ptr<BasisSet>>())
-        .def("reference_wavefunction", &Wavefunction::reference_wavefunction, "docstring")
+        .def("reference_wavefunction", &Wavefunction::reference_wavefunction, "Returns the reference wavefunction.")
         .def("set_reference_wavefunction", &Wavefunction::set_reference_wavefunction, "docstring")
```
susilehtola: Is this guy also supposed to have a docstring other than "docstring"?

jturney: Yes, we've got someone working on this.

psi4/src/psi4/ccenergy/ccenergy.cc
dgasmith: Id actually push back on this and say lets do it at a large scale when we get to it. Id rather not have a half baked transition and would like to do entire modules at the same time.

amjames: That's actually a very good point.

PR #712, opened 2017-05-18
adds t1/d1 diagnostics computed in fnocc to psi4 global variables

edeprince3:
## Description
T1 and D1 diagnostics computed by ccsd/df-ccsd routed through fnocc are added to psi4 global variables

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/fnocc/df_ccsd.cc
+++ b/psi4/src/psi4/fnocc/df_ccsd.cc
@@ -405,6 +405,9 @@ PsiReturnType DFCoupledCluster::CCSDIterations() {
   // add T1 diagnostic to globals
   Process::environment.globals["CC T1 DIAGNOSTIC"] = t1diag;
 
+  // add T1 diagnostic to wavefunction variables
+  reference_wavefunction_->set_variable("CC T1 DIAGNOSTIC",t1diag);
+
   std::shared_ptr<Matrix>T (new Matrix(o,o));
   std::shared_ptr<Matrix>eigvec (new Matrix(o,o));
   std::shared_ptr<Vector>eigval (new Vector(o));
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/fnocc/df_ccsd.cc
```diff
@@ -418,6 +422,9 @@ PsiReturnType DFCoupledCluster::CCSDIterations() {
   outfile->Printf("        D1 diagnostic:                  %20.12lf\n",sqrt(eigval->pointer()[0]));
   outfile->Printf("\n");
 
+  // add D1 diagnostic to globals
+  Process::environment.globals["CC D1 DIAGNOSTIC"] = sqrt(eigval->pointer()[0]);
+
```
amjames: We should be duplicating these into `wfn.variables` since `P::e.globals` is going to be deprecated at some point in the future

andysim: good catch

edeprince3:
since I accidentally coupled this PR to PR #715, should I close both and start over, or is there a way to decouple them?

loriab:
Ok, so you've got two branches, "devdevel_bugs" and "cepa_test" and they both have the same commits. Make note of which commit hashes ought to be in each PR. Then go back to your master branch and update it wrt upstream master (`git checkout master`, `git pull --rebase upstream master`). Make a _new_ branch and apply only the appropriate commits (`git checkout -b cepa_new`, `git cherry-pick aaaa`, `git cherry-pick dddd`). Then push the new branch onto github using the name of the old branch so that the new set of fewer commits show up under the existing PR. It'll have to be a force push because you're rewriting the history of the github branch (`git push -f cepa_new:cepa_test`). Then go back to master (`git checkout master`) and repeat with the t1/d1 commits.

edeprince3:
thanks @loriab I think I have the commits separated into different PRs now.

psi4/src/psi4/fnocc/df_ccsd.cc
amjames: @dgasmith brought up a good point in discussion of #711 that we might want to hold of on duplicating `P::e.globals` into `wfn.variables` until we just do an entire module at the same time so we don't have a half done transition.

To keep things consistent, maybe remove the `set_variable` lines or if it isn't too complicated go through and duplicate all of them up to you @edeprince3.

edeprince3: OK, I'll remove those lines for now.  I should have time to duplicate them in the whole module later this week.

dgasmith:
@jturney Could you re-review this?

Issue #714, opened 2017-05-18
Consistency check on energies by different Psi4 modules


CDSherrill:
Carlos is going to look into writing us a few test cases that make sure that all MP2 methods and all CC methods in Psi4 give consistent energies.

- [ ] DF-MP2: Rob's code vs dfnocc is already in test case test/fnocc4, at the end.   Split this part out to a new test case that compares these two as well as from the occ module.  set qc_module occ will run Ugur's code (give it method 'mp2').
- [ ] Conventional MP2: Crawford's code vs fnocc vs occ
- [ ] DF-CCSD and DF-CCSD(T): Eugene's code with no FNO (df-fnocc), Ugur's code
- [ ] CCSD and CCSD(T): fnocc vs Crawford's code (ccenergy)
- [ ] CEPA comparison: energy('cepa(0)') vs LCCD if available for non-OO, and vs DCFT [FYI some CEPA definitions have singles, others don't]

edeprince3:
energy('lccd') calls fnocc's CEPA(0) (without singles).  You can get an equivalent energy (for closed shells) for

set reference uhf
set ao_basis    none
set dcft_functional cepa0
set algorithm   twostep
print energy('dcft')

edeprince3:
test cepa4 tests detci's cisd vs fnocc's cisd already.

loriab:
cisd, cepa, mp2 are covered. only cc to consider really.

loriab:
closed by #966. thanks, @edeprince3 and @amjames

PR #715, opened 2017-05-19
Cepa test

edeprince3:
## Description
In response to #714 adds a test checking LCCD against DCFT/CEPA0

## Status
- [x] Ready to go


dgasmith:
Looks like you need to add a CMakeLists.txt to the cepa4 folder. You can basically copy one from say `cepa3` and change over the a few pieces of text within that file (this should be obvious).

PR #717, opened 2017-05-22
Devdevel bugs

dgasmith:
## Description
Finishes fixing up the full test suite. Full tests should now pass.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/driver/procrouting/proc_util.py
+++ b/psi4/driver/procrouting/proc_util.py
@@ -221,4 +221,4 @@ def print_ci_results(ciwfn, rname, scf_e, ci_e, print_opdm_no=False):
         ciwfn.print_vector(dvec, root)
 
     # True to keep the file
-    #dvec.close_io_files(True)
+    dvec.close_io_files(True)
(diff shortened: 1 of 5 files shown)


psi4/driver/procrouting/proc_util.py
```diff
@@ -221,4 +221,4 @@ def print_ci_results(ciwfn, rname, scf_e, ci_e, print_opdm_no=False):
         ciwfn.print_vector(dvec, root)
 
     # True to keep the file
-    #dvec.close_io_files(True)
+    dvec.close_io_files(True)
```
andysim: Can't believe it was something so simple.  I was searching for something screwy in libtrans

dgasmith:
@loriab YAPF is set to 120 chars.

psi4/driver/procrouting/proc_util.py
dgasmith: Well it *is* something screwy in PSIO. For some reason that line wipes deletes that file in certain conditions causing SAPT(CASSCF) to fail. Looks like I accidentally committed that at one point, causing others to fail. This *should* be the right way to do it however.

PR #718, opened 2017-05-22
Reading one electron hamiltonian from wavefunction

ashutoshvt:
## Description
The one electron hamiltonian is read from PSIF_OEI inside  libtransform library. This 
leads to a loss of flexibility in adding external potentials to the wavefunction directly 
for post-scf methods through the python interface.  This PR fixes this problem and makes sure the hamiltonian is read consistently from the wavefunction in all the modules.

## Status
- [x] Ready to go


ashutoshvt:
I think this PR is ready to go now. Thanks in advance for any feedback.

andysim:
Thanks for handling this, Ashutosh.  Libtrans is showing its age, so it's great to see it being modernized.  There are many CI and CC tests that don't run on Travis, so as a sanity check it'd be great if you could run the full test suite from the build directory: `ctest -j 12` if you have 12 cores available.

ashutoshvt:
Thanks for the suggestion, Andy. Some test cases do fail and I am going to let you know once I resolve the problem.

loriab:
Good idea, Andy. @ashutoshvt, just in case you haven't rebased recently, there were 5 tests in full tests that weren't passing for a while but are fixed now in upstream. So rebase before investigating.

andysim:
Have you pulled in master in the last day or so?  @dgasmith committed some fixes for the full test suite, which corrected about 5 or 6 failures.  As far as I know, the full suite should pass with that change.

andysim:
Haha, once again we comment at the same time, @loriab.  We don't normally pull things in if we know of small regressions, but needed to get things hurried through for the workshop last week and made an exception.  Thankfully Daniel figured out the problems very quickly.

ashutoshvt:
I have rebased to the upstream master. The problem seems to on my my end right now. I will let you guys know once I correct my code.

loriab:
Indeed. Happily with the new `.devN` versioning scheme we're not guaranteeing that `master` is always passing full tests. (conda pkgs _are_ so guaranteed.) Glad all healed now.

ashutoshvt:
All test cases pass now (using ctest). There wasn't any problem in the code. It was an issue in my perl environment settings which caused failure of some cc test cases.

PR #722, opened 2017-05-27
MCSCF: Add orbital semicanonicalization

fevangelista:
## Description
Adds Python interfaces and code to semicanonicalize MCSCF orbitals.

## Todos
* **Developer**
- [x] Tested to make sure resulting orbitals diagonalize docc, actv, and virt blocks of average Fock matrix.

## Questions
  - [x] By default MCSCF orbitals are semicanonicalized, unless user requests natural orbitals. This seems the most sensible choice, OK with other devs?

## Status
- [x] Ready to go


loriab:
There's some `xrange` that py3 wants converted to `range`. They don't look to be over big lists, so shouldn't hurt py2 performance.

psi4/driver/procrouting/mcscf/mcscf_solver.py
```diff
@@ -44,6 +44,53 @@ def print_iteration(mtype, niter, energy, de, orb_rms, ci_rms, nci, norb, stype)
...
+    U = core.Matrix("U to semi", nmopi, nmopi)
+
+    # Diagonalize each block of Favg
+    for block in [noccpi,nactpi,nvirpi]:
+        F = core.Matrix("Fock",block,block)
+        for h in xrange(nirrep):
+            offset = offsets[h]
+            for i in xrange(block[h]):
+                for j in xrange(block[h]):
+                    F.set(h, i, j, Favg.get(h, i + offset, j + offset))
```
dgasmith: You could consider the numpy slicing notation here:

```
for h in range(nirrep):
     offset = offsets[h]
     sl = slice(offset, offset + block[h])
     F.nph[h][:] = Favg.nph[h, sl, sl]
```

The same thing can be done with U below. This will also be much faster. Last comment: `xrange` is no longer valid with Python3, probably best to drop that.

psi4/driver/procrouting/mcscf/mcscf_solver.py
```diff
@@ -44,6 +44,53 @@ def print_iteration(mtype, niter, energy, de, orb_rms, ci_rms, nci, norb, stype)
...
+                for j in xrange(block[h]):
+                    U.set(h, i + offset, j + offset, evecs.get(h, i, j))
+
+        # update offset for next block
+        for h in xrange(nirrep):
+            offsets[h] += block[h]
+
+    # rotate MOs and push them to the ciwfn
+    Cnew = core.Matrix.doublet(ciwfn.Ca(), U, False, False)
+    ciwfn.Ca().copy(Cnew)
```
dgasmith: Need to be careful here as `Ca()` spans the whole space, we want just the rotable pieces:
```
Cnew = core.Matrix.doublet(ciwfn.get_orbitals("ROT"), U, False, False)
ciwfn.set_orbitals("ROT", Cnew)
```

The `get/set` orbitals are specific to the CIWavefunction as they know much more about the available spaces. It would be nice to bake this back into Wavefunction.

psi4/driver/procrouting/mcscf/mcscf_solver.py
```diff
@@ -355,7 +402,11 @@ def mcscf_solver(ref_wfn):
 
     # Do we diagonalize the opdm?
     if core.get_option("DETCI", "NAT_ORBS"):
+        core.print_out("  \n   Transforming MOs to the natural basis\n")
```
dgasmith: As a note this already has a message from the C side, did you want to add additional text here?

psi4/src/export_fock.cc
```diff
@@ -127,7 +127,10 @@ void export_fock(py::module &m) {
         .def("current_IFock", &SOMCSCF::current_IFock)
         .def("zero_redundant", &SOMCSCF::zero_redundant)
         .def("gradient", &SOMCSCF::gradient)
-        .def("gradient_rms", &SOMCSCF::gradient_rms);
+        .def("gradient_rms", &SOMCSCF::gradient_rms)
+        .def("noccpi", &SOMCSCF::noccpi)
+        .def("nactpi", &SOMCSCF::nactpi)
+        .def("nvirpi", &SOMCSCF::nvirpi);
```
dgasmith: These can be acquired from the `CIWavefunction::get_dimension` function as well.

psi4/driver/procrouting/mcscf/mcscf_solver.py
```diff
@@ -44,6 +44,53 @@ def print_iteration(mtype, niter, energy, de, orb_rms, ci_rms, nci, norb, stype)
     core.print_out("%s %2d:  % 18.12f   % 1.4e  %1.2e  %1.2e  %3d  %3d  %s\n" %
                     (mtype, niter, energy, de, orb_rms, ci_rms, nci, norb, stype))
 
+def build_semicanonical_mos(ciwfn):
```
dgasmith: To be `build` sounds like it should return those when its acting on the incoming Wavefunction. We may be consider calling this `semicanonical_orbs` and binding it to the CIWavefunction itself which should be very analogous to the `ci_nat_orb` function. What do you think?

dgasmith:
Looks good! Defaulting to semicanonical MO's is probably a good idea. Can you think of a reason that we wouldnt want to do this?

We may want to add some sort of test to this just to ensure that we are not actually changing the energy. One idea would to be in a test to do:

```
# Solve and leave the Wavefunction active
set MCSCF_CI_CLEANUP False
set MCSCF_DPD_CLEANUP False
cas_e, cas_wfn = energy("CASSCF", return_wfn=True)

# Update energy
cas_wfn.transform_mcscf_integrals(approx_integrals_only)
nci_iter = cas_wfn.diag_h(1.e-10, 1.e-10)
current_energy = core.get_variable("MCSCF TOTAL ENERGY")

# Compare the energy
compare_values(cas_e, cas_wfn, 8, "Rotated Orbital Energy Invariance")
```

fevangelista:
I just fixed the issue with xrange and the frozen MOs. Will go over some of @dgasmith's comments soon. I just realized I could access the spaces from ciwfn instead of mcscf_obj. I'll remove the python bindings that I added since they seem redundant now.

psi4/driver/procrouting/mcscf/mcscf_solver.py
fevangelista: Sounds like a good solution. How do I bind semicanonical_orbs() to CIWavefunction from Python?

psi4/driver/procrouting/mcscf/mcscf_solver.py
fevangelista: I am not familiar with the syntax. Tried but got this:
```
    F.nph[h][:] = Favg.nph[h, sl, sl]

TypeError: tuple indices must be integers, not tuple
```

dgasmith: Ah, whoops. It should be: ` F.nph[h][:] = Favg.nph[h][sl, sl]`. The `.nph` accessor returns a list of NumPy arrays who's data is shared with the underlying matrix. So the `[h]` accesses the irrep and then the next `[]` uses the normal NumPy slicing syntax.

fevangelista:
@dgasmith: I think this version is a good improvement. I need some direction on how to bind semicanonical_orbs().

psi4/driver/procrouting/mcscf/mcscf_solver.py
fevangelista: Thanks, it worked an is now pushed!

dgasmith:
Python functions are currently bound to C++ classes in the `driver/p4util/python_helpers.py` file. This should be something like:
```
core.CIWavefunction.semicanonical_orbs = semi_func

@classmethod
def semi_func(self):
     ...
```

Where `self` should be equivalent to `ciwfn`. Hmm,, hope this helps.

dgasmith:
Looks like the `molden1` writer is failing which makes sense. However, `ci-properties` is still failing as the `dipole` changed quite considerably. I would expect a small change, but ~30% seems to be a bit large. Need to think on why thats the case a bit.

dgasmith:
This will close #723.

fevangelista:
I pushed a fix to the molden1 test. The issue with the dipole moment is more complex. I think the OPDM and the MOs are certainly not consistent. 

One solution would be to converge the orbitals and then do the following:
- transform orbs. to either canonical or natural basis
- transform integrals
- solve the CI problem
- compute 1/2-RDMs

@dgasmith: since you are more familiar with the mcscf code, can you think of a place where we can put these orbital transformations (semi/nat) before the computation of the OPDM? Better, if you have a few minutes could you try to implement the above code?

Also, is there a way I can access the full Fock matrix with frozen docc and virt blocks? This is of course irrelevant now, but I am thinking that in the future if we try to do gradients of MRPTs it might be better if we have all blocks diagonalize the Fock matrix.

dgasmith:
To do the above you just need to copy [these lines.](https://github.com/fevangelista/psi4/blob/aac087f1c604787c1dd85d42bf2fcdc366ae858e/psi4/driver/procrouting/mcscf/mcscf_solver.py#L201-L210). Looking at the loops im not entirely sure that we can avoid another CI iteration/density build. 

For better or worse the `SOSCF` module by and large does not handle frozen spaces. When you call `update` you can probably just set the `Cocc` and `Cvir` spaces to span the full space. The only catch is that you would need to turn off the tech that tries to save the frozen Fock and inactive Fock matrix when it is formed in other places. This would be the functions `set_AO_IFock` and `set_frozen_orbitals` and the booleans that they set. Maybe you want an option like `clean_focks` or some-such that will reset these booleans. The code will compute these quantities itself and the `Cocc` expansion to also include frozen orbitals should just work.

fevangelista:
One more try... The two RASSCF tests fail because the energy is not invariant with respect to arbitrary rotations of active orbitals. There should be as simple as turning off all orbital transformations if calculations are RASSCF.

The DFCASSCF test was affected by the new algorithm to compute CI natural orbitals. The new one does not scramble different orbital classes, which is what I suspect is what happened with the old code. All the entries in the `nov0` array are the same but are sorted differently. (I take it that you check the first element of each row).

dgasmith:
@fevangelista If its Travis thats failing you can go in and restart individual ones on their website instead of retriggering the test.

fevangelista:
Don't merge yet please. I have got some new neat Matrix/Vector slicing functions that I'd like to add to this repo.

fevangelista:
Just cleaned from obsolete code/comments. Ready to go when passes tests.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -747,6 +747,53 @@ void Matrix::set_column(int h, int m, SharedVector vec)
...
+
+/**
+ * Set a matrix block
+ *
+ * @param rows Rows slice
+ * @param cols Columns slice
+ * @return SharedMatrix object
+ */
+void set_block(Slice rows,Slice cols,SharedMatrix block);
+
```
dgasmith: This is a bit different than `View`'s, but I think they are fairly close. We probably dont want to duplicate the tech... any comments on that?

fevangelista: For what I can see `View` was not integrated extensively and it is used only in exactly three lines of psi4. View is really a shared pointer plus a set of start/end Dimension objects. It uses pointers to store what are dimension-like objects and should be modernized. Slice here is just a set of start/end Dimension objects. Indeed I think it might be a good idea to write `View` using `Slice` (perhaps my choice of class name is not idea, we could also call it `DimensionRange`, `Range`, etc.). So I think this is not a case of repetition.

dgasmith:
I do like the ability to do slices (and it's the correct name IMO). Perhaps if View is not used much we can just cut it out and promote the Slice class which is probably ultimately more extensible. 

Sorry to grab you for more changes, but you keep bringing up good ones :)

-Daniel Smith

fevangelista:
No problem. I think it would be good to get rid of `View` for now and base everything on `Slice`. I'll work on this tomorrow and do some improvements to get/set_block (e.g., forgot passing `Slice` as a const ref). Any desired features in addition to what I added so far?

dgasmith:
I think this is good for now. The next thing that comes to mind is sliced AXPYs and GEMMs which may take a bit more doing.

psi4/src/psi4/libmints/matrix.cc
fevangelista: I must have done a lousy search. `View` is used in about 15-20 lines. Still, I think that using slices might be a good way to go about this.

dgasmith:
This is looking pretty cool! Some of the functionality really cleans up some of the code IMO.

Do you have any other planned changes? This looks ready to go to me.

fevangelista:
@dgasmith: Yes this is ready to go. I just pushed two minor commits that clean existing code so once tests past all can be merged. I think there is potentially more we can do with Slices (merging, getting Slices from MO spaces, which would simplify even further some of the get/set_ commands) but I think that needs some experimentation. Thanks for all the useful feedback!
This also closes issue #723.

PR #720, opened 2017-05-24
Expose P::e.options in Python.

fevangelista:
## Description
Expose P::e.options in Python and enable pure pybind11 plugins.

## Status
- [x] Ready to go

PR #725, opened 2017-05-31
Apifix

fevangelista:
## Description
Exposes options class functions necessary for pybind11 plugins.

## Status
- [x] Ready to go


psi4/src/export_plugins.cc
```diff
@@ -61,14 +61,16 @@ int py_psi_plugin_load(std::string fullpathname)
...
-        ret = 2;
-
+//    // Make sure the plugin isn't already loaded.
+//    if (plugins.count(uc) == 0) {
+//        plugins[uc] = plugin_load(fullpathname);
+//        outfile->Printf("%s loaded.\n", fullpathname.c_str());
+//        ret = 1;
+//    } else
+//        ret = 2;
+    plugins[uc] = plugin_load(fullpathname);
```
rmcgibbo: Out of curiosity, what is the benefit of removing the check on whether the plugin is already loaded?

loriab: We're doing some work to make plugins more flexible. To allow plugins to work in PsiAPI mode and still go through the plugin machinery, they needed some options commands _every_ time initialized that are in `libplugin/plugin_load.cc`. As soon as you go a step further with plugins and drop the plugin machinery for a plugin itself exported to python via pybind11, these changes aren't necessary, and we actually plan to revert them.

fevangelista:
@loriab: can you please restart the gcc:python 3.5 build on travis? I don't have write permission on psi4 so I cannot see the ↻ button.

fevangelista:
I think the file `export_options.cc` was created with PR #720.

loriab:
Ah, of course. I forgot that was already in upstream.

Btw, in general if anyone wants to avoid "undefined" version, just `git fetch upstream` to get the tags. Undefined version doesn't hurt a thing.

fevangelista:
@loriab: I reverted changes made in `export_plugin.cc`.

loriab:
thanks!

dgasmith:
Can wait on merge until #726 is ready to go as well.

Issue #727, opened 2017-06-01
Add cc-pwCVXZ for Li, Be, Na, Mg


rmcgibbo:
Would there be any objection to me filing a PR to add some of the cc-pwCVXZ basis sets for alkali and alkali earth metals that are currently missing from Psi4 (i.e. https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/cc-pwcvtz.gbs)

 As far as I can tell, they're not on the basis set exchange, but thet are available at http://tyr0.chem.wsu.edu/~kipeters/basissets/alkal-nr.html

andysim:
Given that they're from a non-standard location, it'd be nice if you could drop a note to that effect in [this](https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/NOTES) informal collection of developer notes.  Apart from that, I think @loriab has the user-facing documentation nicely automated.  The only other gotcha is that they'll need to have appropriate defaults set for the corresponding RI basis sets (I'm assuming that custom RI bases don't exist, so the cc-pCVXZ versions will have to be used).

susilehtola:
It wouldn't be hard to upload them on the basis set exchange either :D

loriab:
Thanks! Could you instead just update the files tight-cc-pwcvXz.gbs and tight-cc-pwcvXz-ri.gbs that live [here](https://github.com/psi4/psi4/tree/master/psi4/share/psi4/basis/primitives)? Those are the "w-core" and "w-core-ri" that get composed (e.g., [here](https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/NOTES#L336-L337) and [here](https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/NOTES#L456-L457)) into the actual basis set files to keep everything consistent. I'll run the scripts afterward to do the composing, so don't feel you need to follow the README in primitives/ . Let me know if you've any questions – I wasn't an experienced scripter when I put this together.

On Andy's comments, any comment that you add about the origin to the top of the primitives/*.gbs files will get carried through to the final basis set file. And you're not (I think) creating new basis sets, just adding elements to existing basis sets, so there shouldn't be any need to set fitting defaults for orbital bases.

@susilehtola, I am exceedingly grateful for EMSL. But if I trusted their internal consistency on basis sets that should contain the same info, I wouldn't have the crazy basis file summing scheme in Psi4. Admittedly, I haven't reexamined EMSL in a few years.

[...]

PR #331, opened 2016-03-29
Adds non-relativistic core-valence basis sets for Sc-Zn.

fevangelista:
## Description
Adds cc-pwCVXZ-NR basis sets (X = T,Q,5) of Balabanov and K.A. Peterson (J. Chem. Phys. 2005) for elements Sc-Zn.

## Todos
Notable points that this PR has either accomplished or will accomplish.
- [X] Adds missing basis sets for Sc-Zn.

## Questions
- [ ]  Should this be merged with the cc-pwCVXZ basis set file?  In EMSL this basis set has the suffix "-NR" to distinguish it from the "-DK" version.

## Status
- [ ]  Ready to go

PR #728, opened 2017-06-02
Update cc-pVXZ and add cc-pwCVXZ for Li, Be, Na, Mg

rmcgibbo:
## Description
Update cc-pVXZ (D through 5) for Li, Be, Na, and Mg, and add cc-pwCVXZ (D through 5) for Li, Be, Na and Mg with values from Prascher, D.E. Woon, K.A. Peterson, T.H. Dunning, Jr., A.K. Wilson, Theor. Chem. Acc. 128, 69 (2011), accessed from http://tyr0.chem.wsu.edu/~kipeters/basissets/alkal-nr.html. Closes gh-727.

Add cc-pwCVXZ-RI (D through 5) for Li, Be, Na and Mg with values from J. G. Hill and K. A Peterson, "Correlation consistent basis sets for explicitly correlated wavefunctions: Valence and  core-valence basis sets for Li, Be, Na, and Mg" Phys. Chem. Chem. Phys. (2010). The values were accessed from http://www.rsc.org/suppdata/cp/c0/c0cp00020e/c0cp00020e.pdf.

Update cc-pV5Z-RI for Na from Hill and Peterson (2010).

## Todos

- [x] Add RI
- [x] Check locally against molpro.
- [x] Fix cc-pwCV5Z for Na and Mg

## Status
- [x] Ready to go


loriab:
Uggh. Just put the full basis for Na & Mg 5-zeta in the tight file, please, and I'll add except them by hand.

rmcgibbo:
This is ready for review now. I've compared the energies vs. molpro and double checked the equivalence of the basis sets.

loriab:
Yeah, I'll run the scripts so it percolates up, then PR them to here.

loriab:
Sorry for the delay – scripts needed some serious updating. And I found that psi4/psi4#331 had the same non-additivity (pwCV != pV + tight) problems as found here for Na/Mg 5-zeta but just got added to the outer basis sets rather than the basis composing machinery. So machinery got reworked to allow for the case – see NOTES for details. This is all in [PR](https://github.com/rmcgibbo/psi4public/pull/3).

In the end, 100+ basis sets change, as represented in the reports below (yellow means element has changed from before; blue is new element). I think they look sane, but @rmcgibbo and @fevangelista might look them over to confirm.

screen shot 2017-06-11 at 3 03 49 am
screen shot 2017-06-11 at 3 04 07 am

@robertodr, I had to edit pcmsolver-ccsd-pte to adjust for the new Mg basis set (makes sense). But I only had to edit the CC ref values, not the SCF. Does it make sense to you why?

loriab:
This is probably ready to go after @rmcgibbo accept the upstream PR. Needs a commit anyway to re-trigger Distelli.

loriab:
We have a clear policy (at least for what was originally added) of using optimized contractions from EMSL. From [Kirk Peterson's website](http://tyr0.chem.wsu.edu/~kipeters/basis.html), we don't have that option, so we get what he supplies. Not sure what to do about it, unless implement a fn to re-optimize contractions.

CDSherrill:
Perhaps Kirk would be willing to add this feature for us.

andysim:
Looking at the EMSL outputs with and without optimization, it looks like a simple case of ripping out any functions from contractions that are already present as standalone functions.  This should be really easy to put together in Python if we need it.

loriab:
Admittedly I can't support this (I see what Andy sees), but I think there are many cases where the coefficients change between optimized and unoptimized general contractions.

Also, Distelli just doesn't acknowledge this PR. It's built others recently, so I don't know what its problem is. Can override again.

robertodr:
Regarding the PTE test case, I get the following `Integral setup` running the code on `master`:
```
Using in-core PK algorithm.
 Calculation information:
    Number of atoms:                   1
    Number of AO shells:               8
    Number of primitives:             50
    Number of atomic orbitals:        19
    Number of basis functions:        18
                                             
    Integral cutoff                 1.00e-12
    Number of threads:                 1
```
 while this one with the code on this PR:
```
 Using in-core PK algorithm.
  Calculation information:
     Number of atoms:                   1
     Number of AO shells:               8
     Number of primitives:             55
     Number of atomic orbitals:        19
     Number of basis functions:        18
     Integral cutoff                 1.00e-12
     Number of threads:                 1
```
The SCF results are the same within 10^-11 The CC results are different already at the first iteration of the T amplitude solver, _i.e._ the MP2 correlation energy. This despite the fact that the SCF energies reported by the two versions at the end of cctransort are the same.

susilehtola:
I'm confused: in the printout, what is the difference between AOs and basis functions?
I assume this is referring to linearly independent basis functions - a key word is missing!

fevangelista:
@robertodr: I think because the only real difference in the cc-pVDZ basis is a change in the exponent of a d function. No d orbitals are occupied in Mg, so an energy difference would show up only in the correlation energy.

andysim:
@susilehtola the AOs [are the number of Cartesian functions](https://github.com/psi4/psi4/wiki/OrbitalDimensions), while basis functions are the number of functions actually present (fewer than AOs if AM>=2 and pure functions are used).

susilehtola:
@andysim that's confusing, since the AOs should be spherically symmetric! It should say cartesian functions instead of atomic orbitals.

robertodr:
@fevangelista indeed, you're right! Should've seen it myself!

Il mer 14 giu 2017, 09:31 Susi Lehtola <notifications@github.com> ha
scritto:

PR #729, opened 2017-06-03
Minor changes in ccdensity and ccresponse

ashutoshvt:
## Description
1. Default gauge has been set to "velocity" for ccresponse calculations. Un-necessary
printing removed from pertbar.cc
2. cc-energies are reproduced from densities only when debug's value is non-zero.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/ccdensity/fold_UHF.cc
+++ b/psi4/src/psi4/ccdensity/fold_UHF.cc
@@ -95,81 +95,81 @@ void fold_UHF(struct RHO_Params rho_params)
 
 
   if(params.debug_) {
-  outfile->Printf( "\n\tEnergies re-computed from Fock-adjusted CC density:\n");
-  outfile->Printf(   "\t---------------------------------------------------\n");
+    outfile->Printf( "\n\tEnergies re-computed from Fock-adjusted CC density:\n");
+    outfile->Printf(   "\t---------------------------------------------------\n");
 
-  global_dpd_->file2_init(&D, PSIF_CC_OEI, 0, 0, 0, rho_params.DIJ_lbl);
-  global_dpd_->file2_init(&F, PSIF_CC_OEI, 0, 0, 0, "h(I,J)");
-  this_energy = global_dpd_->file2_dot(&D, &F);
-  global_dpd_->file2_close(&F);
-  global_dpd_->file2_close(&D);
-  /*  outfile->Printf( "\tDIJ = %20.15f\n", this_energy); */
-  one_energy += this_energy;
+    global_dpd_->file2_init(&D, PSIF_CC_OEI, 0, 0, 0, rho_params.DIJ_lbl);
+    global_dpd_->file2_init(&F, PSIF_CC_OEI, 0, 0, 0, "h(I,J)");
+    this_energy = global_dpd_->file2_dot(&D, &F);
+    global_dpd_->file2_close(&F);
+    global_dpd_->file2_close(&D);
+    /*  outfile->Printf( "\tDIJ = %20.15f\n", this_energy); */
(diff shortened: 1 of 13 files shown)


psi4/src/read_options.cc
```diff
@@ -1937,11 +1939,11 @@ int read_options(const std::string &name, Options & options, bool suppress_print
     options.add_int("CACHELEVEL",2);
     /*- Specifies the choice of representation of the electric dipole operator.
     Acceptable values are ``LENGTH`` for the usual length-gauge representation,
-    ``VELOCITY`` for the modified velocity-gauge representation in which the
+    ``VELOCITY``(default) for the modified velocity-gauge representation in which the
     static-limit optical rotation tensor is subtracted from the frequency-
     dependent tensor, or ``BOTH``. Note that, for optical rotation calculations,
     only the choices of ``VELOCITY`` or ``BOTH`` will yield origin-independent results. -*/
-    options.add_str("GAUGE","LENGTH", "LENGTH VELOCITY BOTH");
+    options.add_str("GAUGE","VELOCITY", "LENGTH VELOCITY BOTH");
```
robertodr: This makes perfect sense for optical rotation, but it might be a bit of a surprise for any other property. For SCF-level properties, the default is usually always length-gauge and it might then generate confusion to have different defaults between uncorrelated and correlated approaches.

ashutoshvt: Thanks for the feedback. Our ccresponse module calculates three properties right now: optical-rotation, dipole-polarizability and raman optical activity. The intention was to change the gauge for rotation and raman optical activity. For other properties like polarizabilities we only calculate the length gauge version in psi4 irrespective of the gauge provided in input file.

robertodr: OK, thanks. Much more informative.

psi4/src/psi4/ccdensity/fold_UHF.cc
```diff
@@ -481,9 +490,11 @@ void fold_UHF(struct RHO_Params rho_params)
     global_dpd_->buf4_mat_irrep_close(&G, h);
   }
 
+  if(params.debug_){
   global_dpd_->buf4_init(&E, PSIF_CC_EINTS, 0, 23, 27, 23, 27, 0, "E <iJ|kA>");
   two_energy += 2 * global_dpd_->buf4_dot(&E, &G);
   global_dpd_->buf4_close(&E);
+  }
```
dgasmith: To nitpick would it be possible to indent these blocks as well?

psi4/src/read_options.cc
loriab: How about "Specifies the choice of representation of the electric dipole operator. For polarizability, this keyword is ignored and ``LENGTH`` gauge is computed. For optical rotation and raman optical activity, this keyword is active, and acceptable values are ``LENGTH`` for length-gauge representation, ``VELOCITY`` (default) for the modified velocity-gauge representation in which the static-limit tensor is subtracted from the frequency dependent tensor, or ``BOTH``. Note that only velocity gauge results are origin independent."

Is that correct and any clearer?

psi4/src/psi4/ccdensity/fold_UHF.cc
ashutoshvt: Sure, will do that.

ashutoshvt:
Yes, it is correct and clearer. I am going to add that now.

dgasmith:
@ashutoshvt Double checking, this is ready to go right?

ashutoshvt:
@dgasmith yup, its ready to go.  I ran ctest and all tests passed. I will go ahead and rebase.

PR #732, opened 2017-06-07
minor tweak of finite difference atomic_displacements order

psi-rking:
## Description
Minor tweaks to finite differences in the atomic_displacements() function that is used only in prototype ROA calculations at present.  

## Todos
Now supports 3-point and 5-point formulae finite differences in the atomic_displacements() function.
Now returns displacements in canonical most negative to most positive order.

* **Developer Interest**
I have used this function in dipole moment derivative/file17 computations.  Further development along these lines is likely.

None at this juncture.

## Questions
None

## Status
Should be ready to go.

### DIFF
--- a/psi4/src/psi4/findif/fd_misc.cc
+++ b/psi4/src/psi4/findif/fd_misc.cc
@@ -270,36 +270,52 @@ void displace_atom(SharedMatrix geom, const int atom, const int coord, const int
 std::vector< SharedMatrix > atomic_displacements(std::shared_ptr<Molecule> mol, Options &options) {
 
   // This is the size in bohr because geometry is in bohr at this point
-  // This equals 0.1 angstrom displacement
   double disp_size = options.get_double("DISP_SIZE");
+  int pts = options.get_int("POINTS");
 
   int natom = mol->natom();
 
-  // Geometry seems to be in bohr at this point
   Matrix ref_geom_temp = mol->geometry();
   SharedMatrix ref_geom(ref_geom_temp.clone());
 
   std::vector< SharedMatrix > disp_geoms;
 
   // Generate displacements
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/findif/fd_misc.cc
```diff
@@ -270,36 +270,52 @@ void displace_atom(SharedMatrix geom, const int atom, const int coord, const int
...
+        SharedMatrix m1_geom(ref_geom->clone());
+        displace_atom(m1_geom, atom, coord, -1, disp_size);
+        disp_geoms.push_back(m1_geom);
+        // plus displacements
+        SharedMatrix p1_geom(ref_geom->clone());
+        displace_atom(p1_geom, atom, coord, +1, disp_size);
+        disp_geoms.push_back(p1_geom);
+        SharedMatrix p2_geom(ref_geom->clone());
+        displace_atom(p2_geom, atom, coord, +2, disp_size);
+        disp_geoms.push_back(p2_geom);
```
loriab: Could just wrap `if (pts == 5)` around the "p2"/"m2" blocks and condense. But not vital.

dgasmith: This would be cleaner, would be good to put an except in the `else` statement as well.

psi-rking:
I rather prefer the readability of separate blocks, particularly if other options might be supported in the future, but I'll change if you want.  Just added the exception (probably too used to relying on our options class for that).

psi-rking:
I just discovered I had to switch over to the new "properties()" keyword, but it looks great!

dgasmith:
@psi-rking I think is ready to go, all set?

psi-rking:
Yes, it's ready to go.

PR #734, opened 2017-06-07
Basis set parsing cleanup

andysim:
## Description
During the 'inversion' we accumulated some duplicate machinery for handling basis sets in the C++ and Python sections of the code; this PR aims to streamline the process.  My goal is to keep the flexible Python parser, have it create a Python ShellInfo object, and use that to pass fixed-format information into the C++ basis set construction routines.  The superfluous code in the Python and C++ layers should then be nuked.  I also want to clean up the horrible mess I made when implementing ECPs.  Very much a work in progress with feedback welcomed, especially from @loriab, @jturney and @rmcgibbo.

## Todos
* **Developer Interest**
  - [x] Pass pre-formatted data structure to C++ basis set constructor.
  - [x] Get rid of unused `center` info from ShellInfo and GShell.
  - [x] Remove unneeded code:-
     - [x] C++ basis set parser.
     - [x] Python GaussianShell.
     - [x] C++ ShellInfo.
  - [x] Tolerate integers where floats are expected in basis set input.
  - [x] Fold ECPs into BasisSet objects.
  - [x] Fixed bug that prevented integers in basis set exponents/coefficients from being handled correctly, and another that truncated them to 8 digits after the decimal.

## Status
- [x] Ready to go


psi4/driver/p4util/python_helpers.py
```diff
@@ -105,16 +96,13 @@ def _resolve_target(key, target):
...
     elif (sys.version_info[0] > 2) and isinstance(basis, str):
-        basis, ecpbasis = core.BasisSet.build(mol, "ORBITAL", basis)
+        basis = core.BasisSet.build(mol, "ORBITAL", basis)
 
-    if ecpbasis:
-        wfn = core.Wavefunction(mol, basis, ecpbasis)
-    else:
-        wfn = core.Wavefunction(mol, basis)
+    wfn = core.Wavefunction(mol, basis)
     return wfn
```
dgasmith: Cool! This will help prevent quite a bit of confusion I think.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -192,24 +191,19 @@ void BasisSet::print(std::string out) const
...
+    printer->Printf("    Number of Cartesian functions: %d\n", nao());
+    printer->Printf("    Spherical Harmonics?: %s\n", has_puream() ? "true" : "false");
+    printer->Printf("    Max angular momentum: %d\n\n", max_am());
+    if(has_ECP()) {
         printer->Printf("  Core potential: %s\n", name_.c_str());
-        printer->Printf("    Blend: %s\n", target_.c_str());
-        printer->Printf("    Number of shells: %d\n", nshell());
-        printer->Printf("    Number of terms: %d\n", nprimitive());
+        printer->Printf("    Number of shells: %d\n", n_ecp_shell());
+        printer->Printf("    Number of terms: %d\n", n_ecp_primitive());
```
dgasmith: A bit naive on my part, but should this mix the words "terms" and "primitives"?

andysim: Yeah, not really sure what to call these.  We normally think of "primitives" as GTOs combined to approximate an STO.  That's not what's happening in ECPs, so I thought a different word would be better.  Happy to take suggestions on this

andysim:
There are further changes that I'd like to make (getting rid of the extra basis set constructor that extracts a single atom basis, in favor of a member to directly get atom hashes) but extra cleanup will be determined by whether we move to py-side construction.  This is a decent start that I think we can use as a basis for further debate, once #736 is also merged.  Therefore, I think this one's ready to review.

psi4/src/psi4/libmints/basisset.cc
loriab: I've resorted to calling pieces of a gbs file "blocks", if that appeals. Only other info I can find is Cfour output:

```
 ECP: maximum angular momentum =                     2
 ECP: maximum angular projector=                     2
 total number of basis functions                    39
```
Should the number of electrons covered by ECP be printed? (EDIT: guess I just need to scroll)

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -571,133 +574,122 @@ std::shared_ptr <BasisSet> BasisSet::zero_ao_basis_set()
...
 }
 
-std::shared_ptr<BasisSet>
-BasisSet::construct_ecp_from_pydict(std::shared_ptr <Molecule> mol, py::dict pybs, const int forced_puream){
+std::shared_ptr<BasisSet> BasisSet::construct_from_pydict(const std::shared_ptr <Molecule> &mol, py::dict pybs, const int forced_puream){
 
     std::string key = pybs["key"].cast<std::string>();
     std::string name = pybs["name"].cast<std::string>();
     std::string label = pybs["blend"].cast<std::string>();
-    std::string message = pybs["message"].cast<std::string>();
```
loriab: This is covered by the main basis set message, or do I need to write a modified ecp-extracted-from message string? I may well have left that code overly concise.

tests/mints10/input.dat
```diff
@@ -0,0 +1,46 @@
...
+thisSCF = energy('scf')
+compare_values(refSCF, thisSCF, 10, "Reference energy (float input)")         #TEST
+
+basis {
+    [ sto-3g ]
+    cartesian
+    ****
+    H     0
+    S   1   1.00
+          3         1.0
```
loriab: Yay! I have ambitions of making it comment tolerant, too at some point, like `def geometry` in molutil.py

loriab:
One more thing, should [these names](https://github.com/andysim/psi4/blob/basisset/psi4/src/export_mints.cc#L890-L891) be updated now that BasisSet is unified? And is it clear where to get QC-frozen core vs. ECP-frozen core?

andysim:
I think the name `ncore` is ok, since it's still asking for the number of core electrons (maybe `n_ecp_core` is more explicit?).  The docstring certainly needs an update - good catch!  If you have a suggestion on `ncore` nomenclature, fire away.

tests/mints10/input.dat
andysim: A lot of the really heavy element def2 series basis sets have integers, so this is quite an important fix, if we don't want to have to reformat all those new basis sets that came in with ECPs.

loriab:
I slightly prefer `n_ecp_core` in case a user thinks `n_core` valid for cc-pwCVXZ, but I'm fine as-is as long as docstring clear.

tests/mints10/input.dat
loriab: After Robert's goes in, we'll have psi4/share/psi4/basis/primitives/diff_gbs.py that takes a gbs file (or two) as arguments and forms one-elements basis sets for the full periodic table. Can be useful as a check.

psi4/src/psi4/libmints/basisset.cc
andysim: [This](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/libmintsbasisset.py#L824) line results in [this](https://github.com/psi4/psi4/blob/master/tests/scf-ecp/output.ref#L67) output, to tell the user where the ECP was found.  I'm currently assuming that ECPs and bases live together.

andysim:
Oh, I missed your point about the `ncore` meaning the number of frozen core.  Yes, that could very easily confuse users.  I'll switch to `n_ecp_core`.

psi4/src/psi4/libmints/basisset.cc
susilehtola: Wait what? That's not what a primitive means. What you mean is a contraction.

andysim: Oops, I meant to type "that get combined".  Multitasking was never my forte.

PR #735, opened 2017-06-08
Adds ambit helper files to Psi4

jturney:
## Description
Adds a few helper functions for populating ambit tensors from Psi4 included only when ambit add on is enabled.

## Todos
* **Developer Interest**
  - [x] `convert` functions for converting from Matrix and Vector to ambit Tensor.
  - [x] `integrals` functions for computing one- and two-electron integrals storing the results directly in an ambit Tensor.

## Status
- [x] Ready to go


loriab:
Yay! Should the headers be removed from the ambit plugin template?

jturney:
Indeed they should. Forthcoming....

PR #736, opened 2017-06-09
General maintenance

dgasmith:
## Description
I originally had grand plans for a few of the base classes, but this turned out to be harder then expected so this PR kind of took a turn. Best description of this is perhaps "general cleanup".

## Todos
* **Developer Interest**
  - [x] Might compile faster.

As a note two libraries have moved:
```
"psi4/libparallelPsiOutStream.h" -> "psi4/libpsi4util/PsiOutStream.h"
"psi4/libparallel/process.h"     -> "psi4/libpsi4util/process.h"
```
There is no longer a `"psi4/libparallel/libparallel.h"`, you may need to include one of the two above libraries.

- [x] Removed several IWL functions
- [x] Removed some `libciomr` functions
- [x] Cleaned up `libparallel`
- [x] Fixed a few bugs and syntax issues in Wavefunction
- [x] Removed some headers from `Process` and `psi4-dec` which required a larger header shake up. I hope we are finally starting to peel away the points where we accidentally include too many headers and those headers go viral.
- [x] Still working to pull pybind11 out of a few layers
- [x] A GEMM routine with the tag line "a reasonably fast matrix multiply (at least on the DEC3100)" was removed.
- [x] `extern1` was temporarily removed from quickest. Still need to decided on what to do here.

## Questions
- [x] Anything else we can do?

## Status
- [ ] Ready to go


dgasmith:
Looks like everything is on track to finish with a single exception: `EXTERN`. This keyword is used to supply QM/MM charges through the `ExternalPotential` object to a variety of internal programs. Currently the `EXTERN` keyword is a PythonObject which is then cast to a `ExternalPotential` object in the C++ layer. When pushing pybind11 out of the layers I had deleted the ability to add Python objects to the options object (apparently I grepped in too deep of a folder). I think the following are solutions:

 - Add the PythonObjects back to the `options` object. This will unwind most of the work in this PR as pybind11 will go back into basically every Psi file and add ~10 minutes to the compile.
 - Add a new class to the options object that takes `void*`'s and then cast as is currently done. This is about as safe as currently done as there is currently no typechecking on the PyObject cast.
 - Bind Python objects to the Options object Python-side and then insert these objects into the class members in the driver layer.

I object a bit to the first as I really think we need to keep pybind11 out of Psi except in the `export` layer. The second and third are about the same to me. Although in the last one we gain all the C++ type checking at the cost of a few more lines in the driver.

loriab:
Against option 1, for 2 or 3, with suggestion that 3 is what we do for basis sets.

psi4/src/psi4/detci/h0block.cc
```diff
@@ -358,11 +363,13 @@ void CIWavefunction::H0block_xy(double *x, double *y, double E)
       outfile->Printf("H0block_->s0bp[%d] = %lf\n",i,H0block_->s0bp[i]);
   */
 
-   dot_arr(H0block_->c0b, H0block_->c0bp, H0block_->size, &tx);
+   //dot_arr(H0block_->c0b, H0block_->c0bp, H0block_->size, &tx);
+   tx = C_DDOT(H0block_->size,  H0block_->c0bp, 1, H0block_->c0b, 1);
    *x += tx;
-   dot_arr(H0block_->s0b, H0block_->c0bp, H0block_->size, &ty);
+   //dot_arr(H0block_->s0b, H0block_->c0bp, H0block_->size, &ty);
+   ty = C_DDOT(H0block_->size,  H0block_->c0bp, 1, H0block_->s0b, 1);
```
loriab: In this and immediate above, c0bp and s0b are switched wrt dot_arr call. This ok?

dgasmith: `DDOT` should be agnostic to input array order. If you see this in GEMM however...

dgasmith:
Pushed out a new round of updates mostly removing more pybind11 from the core `size_t`ifying the code base. Also merged through @andysim ECP Basis changes without too much issue.

dgasmith:
Ready for review, good luck! 

If anyone has EXTERN suggestions I would be glad to hear them.

andysim:
Couldn't see anything obviously wrong on first pass through.  I wonder if the EXTERN stuff is best handled by having a C++ class that's exposed using pybind11, so that it can be constructed py-side but easily stored in wfn.  That way we wouldn't be passing a py object around.  This is really great stuff - some long overdue cleanup and standardization in here.

loriab:
Indeed, these are great improvements. Only the CheMPS2 print handling needs fixing before merging, I think. The PR has also greatly deranged the CMake module dependency chain, so `make -j8` fails over and over (but restartable). That needs to be handled soon but not necessarily immediately.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -720,23 +724,34 @@ void HF::find_occupation()
...
+            for (int i = 0; i < epsilon_a_->dimpi()[h]; ++i){
                 pairs_a.push_back(std::make_pair(epsilon_a_->get(h, i), h));
+            }
         }
-        for (int h=0; h<epsilon_b_->nirrep(); ++h) {
-            for (int i=0; i<epsilon_b_->dimpi()[h]; ++i)
-                pairs_b.push_back(std::make_pair(epsilon_b_->get(h, i), h));
+        sort(pairs_a.begin(), pairs_a.end());
+
+        // Do we need to sort beta?
```
dgasmith: @fevangelista @CDSherrill @andysim I think this should fix #738. We simply take the sort of `pairs_a` if we are a singlet, as I believe we effectively have a RHF reference for UHF/CUHF/ROHF reference singlets. This will ensure the resulting code is always a singlet regardless of degenerate noise in the sort. I think this handles all cases, but some thought to make sure this doesn't trip up an edge case is required.

fevangelista: The only issue I can see is with broken symmetry computations. However, I am convincing myself this is not really an issue. If ran in C1 symmetry, the lack of beta sorting should not make any difference. On the other hand, I don't think Psi4 can do broken symmetry computations with non-C1 groups and use different alpha and beta occupation vectors. This is necessary to represent determinants like:

`|(core)^2 (4a' alpha) (2a" beta)>` (That's what we do in OCDFT, and there we have tricks to avoid psi4 sorting the MOs).

So in the end this might not be an issue at all.

PR #737, opened 2017-06-09
Desymmetrize SCF for non-symmetry-aware modules

CDSherrill:
## Description
Allow symmetry to be used in SCF, but then convert the symmetrized wavefunction to a desymmetrized (C1) format for use by non-symmetry-aware modules

## Todos
* **Developer Interest**
  - [x] Correct bug in D_subset_helper() and F_subset_helper() for nonsymmetric matrices
  - [x] Refactor D_subset_helper() and F_subset_helper() to a generic matrix_subset_helper()
  - [x] Implement Wavefunction::c1_deep_copy() function
  - [ ] Update bassisset_ to be C1 inside c1_deep_copy() function
  - [ ] Call the new c1 deep copy from the driver in the appropriate places

  - [ ] Allow symmetry to be used in SCF even when post-SCF module is C1-only

## Questions
- [x] The wavefunction object has a bassiset_ member that needs to be updated to C1

## Status
- [x] Ready to go

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1193,20 +1193,6 @@ def scf_helper(name, post_scf=True, **kwargs):
         p4util.banner('  Computing high-spin triplet guess  ')
         core.print_out('\n')
 
-    # C1: OLD WAY
-    # If we force c1 copy the active molecule
-    if use_c1:
-        scf_molecule.update_geometry()
-        if scf_molecule.schoenflies_symbol() != 'c1':
-            core.print_out("""  A requested method does not make use of molecular symmetry: """
-                           """further calculations in C1 point group.\n""")
-            scf_molecule = scf_molecule.clone()
-            scf_molecule.reset_point_group('c1')
-            scf_molecule.fix_orientation(True)
-            scf_molecule.fix_com(True)
-            scf_molecule.update_geometry()
-    # C1: NEW WAY: comment out above
-
     # If GUESS is auto guess what it should be
     if core.get_option('SCF', 'GUESS') == "AUTO":
         if (core.get_option('SCF', 'REFERENCE') in ['RHF', 'RKS']) and \
@@ -1424,37 +1410,20 @@ def scf_helper(name, post_scf=True, **kwargs):
 
     optstash.restore()
 
-    if not use_c1:
+    if (not use_c1) or (scf_molecule.schoenflies_symbol() == 'c1'):
         return scf_wfn
     else:
-        # C1: OLD WAY
-        return scf_wfn
-        # C1: NEW WAY
-        #return scf_wfn.deep_copy(scf_wfn)
-
-#        # If we force c1 copy the active molecule
-#        scf_molecule.update_geometry()
-#        if scf_molecule.schoenflies_symbol() != 'c1':
-#            core.print_out("""  A requested method does not make use of molecular symmetry: """
-#                           """transforming Wavefunction so that further calculations in C1 point group.\n""")
-#            scf_molecule = scf_molecule.clone()
-#            scf_molecule.reset_point_group('c1')
-#            scf_molecule.fix_orientation(True)
-#            scf_molecule.fix_com(True)
-#            scf_molecule.update_geometry()
-
-## Not really needed, but we cannot set the orbitals explicitly-- need Python bindings
-#scf_e, nosym_wfn = psi4.energy("SCF", molecule=mol_nosym, return_wfn=True)
-#
-## Copy SCF quantities using built in functions to cast from SO (c2v symmetry) to AO (C1 symmetry).
-#nosym_wfn.Ca().np[:] = sym_wfn.Ca_subset("AO", "ALL")
-#nosym_wfn.Cb().np[:] = sym_wfn.Cb_subset("AO", "ALL")
-#
-#nosym_wfn.epsilon_a().np[:] = sym_wfn.epsilon_a_subset("AO", "ALL")
-#nosym_wfn.epsilon_b().np[:] = sym_wfn.epsilon_b_subset("AO", "ALL")
-#
-## Zip up alpha/beta/nfzc/ndocc/etc, cant actually do this yet. Need Python setters.
-#nosym_wfn.ndocc()[0] = sym_wfn.ndocc().sum()
+        # If we force c1 copy the active molecule
+        scf_molecule.update_geometry()
+        core.print_out("""  A requested method does not make use of molecular symmetry: """
+                           """further calculations in C1 point group.\n""")
+        c1_molecule = scf_molecule.clone()
+        c1_molecule.reset_point_group('c1')
+        c1_molecule.fix_orientation(True)
+        c1_molecule.fix_com(True)
+        c1_molecule.update_geometry()
+        c1_basis, c1_ecpbasis = core.BasisSet.build(c1_molecule, "ORBITAL", core.get_global_option('BASIS'))
+        return scf_wfn.deep_copy(scf_wfn, c1_basis)
 
 
 def run_dcft(name, **kwargs):
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1193,18 +1193,19 @@ def scf_helper(name, post_scf=True, **kwargs):
         p4util.banner('  Computing high-spin triplet guess  ')
         core.print_out('\n')
 
-
-#    # If we force c1 copy the active molecule
-#    if use_c1:
-#        scf_molecule.update_geometry()
-#        if scf_molecule.schoenflies_symbol() != 'c1':
-#            core.print_out("""  A requested method does not make use of molecular symmetry: """
-#                           """further calculations in C1 point group.\n""")
-#            scf_molecule = scf_molecule.clone()
-#            scf_molecule.reset_point_group('c1')
-#            scf_molecule.fix_orientation(True)
-#            scf_molecule.fix_com(True)
-#            scf_molecule.update_geometry()
+    # C1: OLD WAY
+    # If we force c1 copy the active molecule
+    if use_c1:
+        scf_molecule.update_geometry()
+        if scf_molecule.schoenflies_symbol() != 'c1':
+            core.print_out("""  A requested method does not make use of molecular symmetry: """
+                           """further calculations in C1 point group.\n""")
+            scf_molecule = scf_molecule.clone()
+            scf_molecule.reset_point_group('c1')
+            scf_molecule.fix_orientation(True)
+            scf_molecule.fix_com(True)
+            scf_molecule.update_geometry()
+    # C1: NEW WAY: comment out above
 
     # If GUESS is auto guess what it should be
     if core.get_option('SCF', 'GUESS') == "AUTO":
@@ -1426,7 +1427,11 @@ def scf_helper(name, post_scf=True, **kwargs):
     if not use_c1:
         return scf_wfn
     else:
-        return scf_wfn.c1_deep_copy()
+        # C1: OLD WAY
+        return scf_wfn
+        # C1: NEW WAY
+        #return scf_wfn.deep_copy(scf_wfn)
+
 #        # If we force c1 copy the active molecule
 #        scf_molecule.update_geometry()
 #        if scf_molecule.schoenflies_symbol() != 'c1':
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1422,8 +1422,8 @@ def scf_helper(name, post_scf=True, **kwargs):
         c1_molecule.fix_orientation(True)
         c1_molecule.fix_com(True)
         c1_molecule.update_geometry()
-        c1_basis, c1_ecpbasis = core.BasisSet.build(c1_molecule, "ORBITAL", core.get_global_option('BASIS'))
-        return scf_wfn.deep_copy(scf_wfn, c1_basis)
+        c1_basis = core.BasisSet.build(c1_molecule, "ORBITAL", core.get_global_option('BASIS'))
+        return core.Wavefunction.c1_deep_copy(scf_wfn, c1_basis)
 
 
 def run_dcft(name, **kwargs):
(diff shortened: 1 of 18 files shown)


psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -235,6 +235,110 @@ void Wavefunction::deep_copy(const Wavefunction *other)
 
 }
 
+void Wavefunction::c1_deep_copy(SharedWavefunction other)
+{
+    c1_deep_copy(other.get());
+}
+
+void Wavefunction::c1_deep_copy(const Wavefunction *other)
```
dgasmith: It might be better if this simply returned a new Wavefunction (make this a static method).

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -235,6 +235,110 @@ void Wavefunction::deep_copy(const Wavefunction *other)
...
+        other->epsilon_subset_helper(other->epsilon_b_, other->nsopi_, "AO", "ALL");
+
+
+    // these are simple SharedMatrices of size 3*natom_, etc., so should
+    // not depend on symmetry ... can just copy them
+    if (other->gradient_) gradient_ = other->gradient_->clone();
+    if (other->hessian_) hessian_ = other->hessian_->clone();
+    if (other->tpdm_gradient_contribution_)
+        tpdm_gradient_contribution_ = other->tpdm_gradient_contribution_->clone();
+
```
dgasmith: This is more of a tag for me, the internal `array` and `variable` maps were not copied over until #736. Should probably add that here also. Although, what to do with the `array` objects that might be irreped is interesting.

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -779,6 +883,98 @@ SharedMatrix Wavefunction::D_subset_helper(SharedMatrix D, SharedMatrix C, const
...
+            C_DGEMM('N', 'N', nsor, nmor, nsor, 1.0, Srp[0], nsor, Crp[0], nmor, 0.0, SC, nmor);
+            C_DGEMM('N', 'N', nsol, nmor, nsor, 1.0, Msop[0], nsor, SC, nmor, 0.0, temp, nmor);
+            C_DGEMM('N', 'N', nsol, nmol, nsol, 1.0, Slp[0], nsol, Clp[0], nmol, 0.0, SC, nmol);
+            C_DGEMM('T', 'N', nmol, nmor, nsol, 1.0, SC, nmol, temp, nmor, 0.0, Mmop[0], nmor);
+        }
+        delete[] temp;
+        delete[] SC;
+        return M2;
+    } else {
```
dgasmith: For the above it might be good to use the `Matrix->remove_symmetry` and `Matrix->doublet/triplet` functionality to simplify this code down to 4-5 lines. The only downside is that the internals will allocate a few extra matrices in the `MO` return as the chain GEMM functionality isnt smart enough to reuse buffers. On that note, we have a `chain_dot` function Python side- might be good to move it C-side.

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -235,6 +235,110 @@ void Wavefunction::deep_copy(const Wavefunction *other)
...
+    name_ = other->name_;
+    molecule_ = std::shared_ptr<Molecule>(new Molecule(other->molecule_->clone()));
+    molecule_->reset_point_group("c1");
+    molecule_->set_orientation_fixed(true);
+    molecule_->set_com_fixed(true);
+    molecule_->update_geometry();
+
+    basisset_ = other->basisset_; // CDS: think we need to reset this
+                                  // b/c it contains a molecule
+    basissets_ = other->basissets_; // Still cannot copy basissets
```
dgasmith: As it stand this basisset still has a pointer to the old molecule with symmetry. In this particular case we could just add a `molecule` setter to the BasisSet object as the previous wavefunction will be destroyed. This wont work in the general case as it would screw up anything downstream that used the previous Wavefunction. We could consider making the `to_c1` an attribute so that this call is a *destructive* process (just updates the current Wavefunction). This could be a elegant way around this particular issue until the BasisSet object is cleaned up a bit. I think @andysim is working on some parts of this.

dgasmith:
Hmm, tricky issues from Travis. We downcast to Wavefunction so we loose things like `cphf_solve` and `semicanonicalize` and we need to remove symmetry from not only the orbital, but also the density fitted basis sets as well. We may need to add `c1_deep_copy` to the SCF-level Wavefunctions.

loriab:
We were thinking that the build_scf_wavefunction auxiliary basissets had already been consumed (except maybe decon) and the post-scf aux ones hadn't been added to Wfn. Thought we might be safe for the very particular location that c1_deep_copy is called at present, though, very true, not in general. Hopefully BasisSet will be copyable and/or rebuildable by then. Or our thing may be all wrong.

CDSherrill:
Ran complete test suite, looks ok.  Ready to go.  Did not have time to address all previous comments.

psi4/src/psi4/dfocc/get_moinfo.cc
```diff
@@ -340,34 +340,33 @@ void DFOCC::get_moinfo() {
...
+    //IWL::read_one(psio_.get(), PSIF_OEI, PSIF_SO_V, so_ints, ntri_so, 0, 0, "outfile");
+    //Vso_->set(so_ints);
+    //IWL::read_one(psio_.get(), PSIF_OEI, PSIF_SO_S, so_ints, ntri_so, 0, 0, "outfile");
+    //Sso_->set(so_ints);
+    //free(so_ints);
+    //Hso_->copy(Tso_);
+    //Hso_->add(Vso_);
+    //Tso_.reset();
+    //Vso_.reset();
+    // CDS: Migrate these from disk reads to grabbing off Wavefunction
```
dgasmith: Nice, the more of this the better. We noticed similar issues when in the gradient terms where potentials and perturbations were not being applied correctly.

psi4/src/psi4/libscf_solver/rhf.cc
```diff
@@ -937,5 +937,34 @@ bool RHF::stability_analysis() {
...
+        epsilon_subset_helper(epsilon_a_, nsopi_, "AO", "ALL");
+    hf_wfn->epsilon_b_ = hf_wfn->epsilon_a_;
+    // H_ ans X_ reset in the HF constructor, copy them over here
+    SharedMatrix SO2AO = aotoso()->transpose();
+    if (H_) hf_wfn->H_->remove_symmetry(H_, SO2AO);
+    if (X_) hf_wfn->X_->remove_symmetry(X_, SO2AO);
+
+    // pretty sure the next 2 things are really set already
+    same_a_b_dens_ = true;
+    same_a_b_orbs_ = true;
```
dgasmith: Dont these need to be set on the returned Wavefunction? If people are blindly copying could be good to do `hf_wfn-> same_a_b_orbs_ = same_a_b_orbs_`.

psi4/src/psi4/libscf_solver/rohf.cc
```diff
@@ -1383,4 +1383,32 @@ bool ROHF::stability_analysis()
...
+    if (Fb_) hf_wfn->Fb_ = Fb_subset("AO");
+    if (epsilon_a_) hf_wfn->epsilon_a_ =
+        epsilon_subset_helper(epsilon_a_, nsopi_, "AO", "ALL");
+    hf_wfn->epsilon_b_ = hf_wfn->epsilon_a_;
+    hf_wfn->epsilon_b_->set_name("beta orbital energies");
+    // H_ ans X_ reset in the HF constructor, copy them over here
+    SharedMatrix SO2AO = aotoso()->transpose();
+    if (H_) hf_wfn->H_->remove_symmetry(H_, SO2AO);
+    if (X_) hf_wfn->X_->remove_symmetry(X_, SO2AO);
+
```
dgasmith: Needs same dens/orbs.

psi4/src/psi4/libscf_solver/uhf.cc
```diff
@@ -1069,4 +1069,32 @@ void UHF::compute_nos()
...
+    if (Fb_) hf_wfn->Fb_ = Fb_subset("AO");
+    if (epsilon_a_) hf_wfn->epsilon_a_ =
+        epsilon_subset_helper(epsilon_a_, nsopi_, "AO", "ALL");
+    if (epsilon_b_) hf_wfn->epsilon_b_ =
+        epsilon_subset_helper(epsilon_b_, nsopi_, "AO", "ALL");
+    // H_ ans X_ reset in the HF constructor, copy them over here
+    SharedMatrix SO2AO = aotoso()->transpose();
+    if (H_) hf_wfn->H_->remove_symmetry(H_, SO2AO);
+    if (X_) hf_wfn->X_->remove_symmetry(X_, SO2AO);
+
```
dgasmith: Needs same dens/orbs.

psi4/src/psi4/libscf_solver/uhf.cc
```diff
@@ -1069,4 +1069,32 @@ void UHF::compute_nos()
...
+    if (Fa_) hf_wfn->Fa_ = Fa_subset("AO");
+    if (Fb_) hf_wfn->Fb_ = Fb_subset("AO");
+    if (epsilon_a_) hf_wfn->epsilon_a_ =
+        epsilon_subset_helper(epsilon_a_, nsopi_, "AO", "ALL");
+    if (epsilon_b_) hf_wfn->epsilon_b_ =
+        epsilon_subset_helper(epsilon_b_, nsopi_, "AO", "ALL");
+    // H_ ans X_ reset in the HF constructor, copy them over here
+    SharedMatrix SO2AO = aotoso()->transpose();
+    if (H_) hf_wfn->H_->remove_symmetry(H_, SO2AO);
+    if (X_) hf_wfn->X_->remove_symmetry(X_, SO2AO);
```
dgasmith: If X is non symmetric, will this work here? Needs to be handled very carefully when we snip out small S eigenvalues.

psi4/src/psi4/libscf_solver/rhf.cc
```diff
@@ -937,5 +937,34 @@ bool RHF::stability_analysis() {
...
+{
+    std::shared_ptr<Wavefunction> wfn = Wavefunction::c1_deep_copy(basis);
+    auto hf_wfn = std::make_shared<RHF>(wfn, functional_, wfn->options(), wfn->psio());
+    // now just have to copy the matrices that RHF initializes
+    // include only those that are not temporary (some deleted in finalize())
+    if (Ca_) hf_wfn->Ca_ = Ca_subset("AO", "ALL");
+    hf_wfn->Cb_ = hf_wfn->Ca_;
+    if (Da_) hf_wfn->Da_ = Da_subset("AO");
+    hf_wfn->Db_ = hf_wfn->Da_;
+    hf_wfn->D_  = hf_wfn->Da_;
```
loriab: Won't `hf_wfn->Da_` not be available if `!Da_`. Maybe the logic needs reoorganization, even though the absence of `Da_` is unlikely.

psi4/driver/procrouting/proc.py
```diff
@@ -1422,7 +1409,26 @@ def scf_helper(name, post_scf=True, **kwargs):
         core.tstop()
 
     optstash.restore()
-    return scf_wfn
+
+    if (not use_c1) or (scf_molecule.schoenflies_symbol() == 'c1'):
```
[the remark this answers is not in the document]
loriab: confirmed that whole test suite passes with `if (not use_c1):`

psi4/src/psi4/libscf_solver/rhf.cc
CDSherrill: Good point.  Actually these are already set in the RHF constructor so I'm deleting these lines.

psi4/src/psi4/libscf_solver/rohf.cc
CDSherrill: Would, but already set in ROHF constructor

psi4/src/psi4/libscf_solver/uhf.cc
CDSherrill: I don't see anything in the code that would fail if X is not symmetric... maybe there is something subtle I am missing though?  Perhaps someone else can check.

psi4/src/psi4/libscf_solver/uhf.cc
CDSherrill: Taken care of by UHF constructor

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
CDSherrill: Cool!  Good to know.

psi4/src/psi4/libscf_solver/uhf.cc
loriab: Was DGAS warning about when the space is truncated for linear dependency?

psi4/src/psi4/libscf_solver/rhf.cc
CDSherrill: Good catch, fixing that

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3256,7 +3260,8 @@ bool Matrix::equal(const Matrix *rhs) {
     for (int h = 0; h < nirrep(); ++h) {
         for (int m = 0; m < rowspi()[h]; ++m) {
             for (int n = 0; n < colspi()[h ^ symmetry_]; ++n) {
-                if (get(h, m, n) != rhs->get(h, m, n)) return false;
+                if (fabs(get(h, m, n) - rhs->get(h, m, n)) > TOL)
```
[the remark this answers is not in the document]
CDSherrill: Fixed

psi4/src/export_wavefunction.cc
```diff
@@ -206,19 +208,24 @@ void export_wavefunction(py::module& m) {
              "Semicanonicalizes the orbitals for ROHF.");
 
     py::class_<scf::RHF, std::shared_ptr<scf::RHF>, scf::HF>(m, "RHF", "docstring")
-        .def(py::init<std::shared_ptr<Wavefunction>, std::shared_ptr<SuperFunctional>>());
+        .def(py::init<std::shared_ptr<Wavefunction>, std::shared_ptr<SuperFunctional>>())
+        .def("c1_deep_copy", &scf::RHF::c1_deep_copy,
+             "Returns a new wavefunction with internal data converted to C_1 symmetry, using pre-c1-constructed BasisSet *arg0*");
```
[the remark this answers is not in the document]
CDSherrill: Ok, I updated these

PR #739, opened 2017-06-13
Fix possible unsigned subtraction wrap around.

### LINKED ISSUE #716: DF-HF segfault with very large system
I see a segfault when running an SCF with some of the systems in L7 (in particular [circumcoronene](http://www.begdb.com/index.php?action=oneMolecule&state=show&id=4108)). I've traced the issue, at least in part, to integer sizing bug [in this check](https://github.com/psi4/psi4/blob/6044281f7a94c23852a4fd8291ab67dbcc3280fd/psi4/src/psi4/libfock/DFJK.cc#L699) when the user requests less than Q^2 memory. This is due to a wraparound during subtraction, which results in the error condition [here](https://github.com/psi4/psi4/blob/6044281f7a94c23852a4fd8291ab67dbcc3280fd/psi4/src/psi4/libfock/DFJK.cc#L761) erroneously evaluating to `false`.

rmcgibbo:
## Description
Fixes #716

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/libfock/DFJK.cc
+++ b/psi4/src/psi4/libfock/DFJK.cc
@@ -696,7 +696,7 @@ void DFJK::initialize_JK_disk()
 
     // ==> Memory Sizing <== //
     ULI two_memory = ((ULI)auxiliary_->nbf())*auxiliary_->nbf();
-    ULI buffer_memory = memory_ - 2*two_memory; // Two is for buffer space in fitting
+    ULI buffer_memory =(memory_ > 2 * two_memory) ? memory_ - 2*two_memory : 0; // Two is for buffer space in fitting
 
     //outfile->Printf( "Buffer memory = %ld words\n", buffer_memory);
 


psi4/src/psi4/libfock/DFJK.cc
```diff
@@ -696,7 +696,7 @@ void DFJK::initialize_JK_disk()
 
     // ==> Memory Sizing <== //
     ULI two_memory = ((ULI)auxiliary_->nbf())*auxiliary_->nbf();
-    ULI buffer_memory = memory_ - 2*two_memory; // Two is for buffer space in fitting
+    ULI buffer_memory =(memory_ > 2 * two_memory) ? memory_ - 2*two_memory : 0; // Two is for buffer space in fitting
```
fevangelista: How about using the max function? e.g.
```
ULI buffer_memory = std::max(memory_ - 2 * two_memory, 0);
```

rmcgibbo: but when `memory_ - 2 * two_memory` itself wraps around, it's larger than zero.

fevangelista: Ok, I see, both `memory_` and  `two_memory` are ULI.

PR #742, opened 2017-06-14
DF_Helper Updates

schiebermc:
## Description
This pull request replaces replaces the `DFERI` class with `DF_Helper` in DFEP2, MCSCF, SAPT-DFT, FSAPT, and USAPT. 

Update 08/16/17: This pull request demolishes all traces of libthce.

* **Developer Interest**
  - [x] Sparsity utilization in 3-index integral transforms
  - [x] Workflows optimized to fit procedure contexts
  - [x] Intermediate recycling
  - [x] Excellent parallel scaling
  - [x] Simple UI including pythonic tensor indexing

## Status
- [x] Ready to go


schiebermc:
Also updates the unit tests for DF_Helper.

psi4/src/export_fock.cc
```diff
@@ -164,5 +168,24 @@ void export_fock(py::module &m) {
...
+        .def("set_JK_hint", &df_helper::DF_Helper::set_JK_hint)
+        .def("get_fun_mask", &df_helper::DF_Helper::get_fun_mask)
         .def("build_JK", &df_helper::DF_Helper::build_JK);
+
+// MATT'S HACKING FIXME
+    py::class_<scf::SADGuess, std::shared_ptr<scf::SADGuess>>(m, "SADGuess", "docstring")
+        .def_static("build_SAD",
+        [](std::shared_ptr<BasisSet> basis, std::vector<std::shared_ptr<BasisSet>> atomic_bases, int i, int j) 
+        { 
+           return scf::SADGuess(basis, atomic_bases, i, j, Process::environment.options);
```
dgasmith: +1 for lambda's.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -137,42 +137,51 @@ void DF_Helper::initialize()
...
     outfile->Printf("      nao           = %zu\n", nao_);
     outfile->Printf("      naux          = %zu\n", naux_);
     outfile->Printf("      Scwarz cutoff = %E\n", cutoff_);
     outfile->Printf("      mem (doubles) = %zu\n", memory_);
-    outfile->Printf("      method        = %s\n", method_.c_str());
-    outfile->Printf("      core?         = %d\n", (int)core_);
+    outfile->Printf("      nthreads      = %zu\n", nthreads_);
+    outfile->Printf("      Algorithm     = %s\n", method_.c_str());
+    outfile->Printf("      AO_core?      = %d\n", (int)AO_core_);
+    outfile->Printf("      MO_core?      = %d\n", (int)MO_core_);
```
dgasmith: We might want to cleanup the text a bit and make sure it looks very similar to the current DF-JK and DF Tensor print outs. One thing to note is the `=> Something <=` headers.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -137,42 +137,51 @@ void DF_Helper::initialize()
...
-    if(!(std::fabs(mpower_ - 0.0) < 1e-13)){
-        if(core_)
-            prepare_metric_core();
-        else{
-            if(!hold_met_ || !direct_)
-                prepare_metric();
-        }
-    }
+    timer_off("DF_Helper~ prepare_sparse   ");
+    outfile->Printf(" Mask size vs nao^2 : (%d) vs (%d)\n", small_skips_[nao_], nao_*nao_);
```
dgasmith: Might want to give this in terms of percentages, not raw numbers.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -182,71 +191,89 @@ void DF_Helper::initialize()
...
+                for (int q = 0; q < nQ; q++) {
+                    max_val = std::max(max_val, fabs(buffer[p * (nQ * nP * nQ + nQ) + q * (nP * nQ + 1)]));
+                }
+            }
+            max_ = std::max(max_, max_val);
+            shell_prints[P * pshells_ + Q] = shell_prints[Q * pshells_ + P] = max_val;
+            for (int p = 0; p < nP; p++) {
+                for (int q = 0; q < nQ; q++) {
+                    fun_prints[(p + oP) * nao_ + (q + oQ)] =
+                    fun_prints[(q + oQ) * nao_ + (p + oP)] = max_val;
```
dgasmith: Is there a reason we removed threading from this? Can we add it back?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1341,183 +1575,194 @@ void DF_Helper::transform_disk() {
...
-        std::vector<double> M;
+// scope buffer declarations
+{
+    // declare bufs
+    std::vector<double> T; T.reserve(max_block*nao*wtmp);
+    std::vector<double> F; F.reserve(max_block*wfinal);
+    std::vector<double> N; N.reserve(max_block*wfinal);
+    double *Tp = &T[0];
+    double *Fp = &F[0];
+    double *Np = &N[0];
```
dgasmith: Your right, if we need to ensure that these are aligned and switch back to `malloc`'s it should be fine. It was probably good at the start to use `std::vectors` so we didn't have to worrying about freeing them at the end, but now that we are close to complete the raw pointers shouldn't be as big of a deal.

 Also, any reason you did not do `double *Tp = T.data();`?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1341,183 +1575,194 @@ void DF_Helper::transform_disk() {
...
+                size_t sp_size = small_skips_[k];
+                size_t jump = (AO_core_ ? big_skips_[k] + bcount * sp_size : (big_skips_[k] * block_size) / naux_);
+
+        #ifdef _OPENMP
+                rank = omp_get_thread_num();
+        #endif
+                for (size_t m = 0, sp_count = -1; m < nao_; m++) {
+                    if (schwarz_fun_mask_[k * nao_ + m]) {
+                        sp_count++;
+                        C_DCOPY(bsize, &Bpt[m * bsize], 1, &C_buffers[rank][sp_count * bsize],
```
dgasmith: In VTune, whats the total (relative) time spent on this out of curiosity?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -2062,54 +2313,64 @@ void DF_Helper::JK_disk(std::vector<SharedMatrix> Cleft, std::vector<SharedMatri
     std::shared_ptr<BasisSet> zero = BasisSet::zero_ao_basis_set();
     std::shared_ptr<IntegralFactory> rifactory(new IntegralFactory(aux_, zero, primary_, primary_));
     std::vector<std::shared_ptr<TwoBodyAOInt>> eri(nthreads_);
-    #pragma omp parallel private(rank) num_threads(nthreads_)
-    {
-        #ifdef _OPENMP
+    
+    // manual cache alignment..(poor std vectors..) (assumes cache size is 64bytes)
+    size_t remainder = nao*nao;
```
dgasmith: `remainder` is a C++ [function](http://www.cplusplus.com/reference/cmath/remainder/).

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -2183,51 +2447,58 @@ void DF_Helper::compute_J(std::vector<SharedMatrix> D, std::vector<SharedMatrix>
...
+        
+        // align cache to reduce false sharing
+        align_size = (nao % 8 ? nao + (8 - nao % 8) : nao);
+        
+        // complete pruned J
+        #pragma omp parallel for schedule(guided) num_threads(nthreads_)
+        for(size_t k=0, count=0; k<nao; k++){
+            size_t sp_size = small_skips_[k];
+            size_t jump = (AO_core_ ? big_skips_[k] + bcount * sp_size : (big_skips_[k] * block_size) / naux);
+            C_DGEMV('T', block_size, sp_size, 1.0, &Mp[jump], sp_size, T1p, 1, 0.0, &T2p[k*align_size], 1);
```
dgasmith: I would mark this as a point to revisit, we could still save ~50% on J due to symmetry. It wouldnt impact the total walltime that much, but could help the total efficiency marks.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -2236,16 +2507,21 @@ void DF_Helper::compute_K(std::vector<SharedMatrix> Cleft,
         size_t cright = Cright[i]->colspi()[0];
         double* Kp = K[i]->pointer()[0];
 
+        // cache alignment (is this over-zealous?)
+        size_t remainder = block_size*cleft;
+        size_t align_size = (remainder%8 ? remainder + (8-remainder%8) : remainder);
```
dgasmith: Ditto from above.

tests/df-helper/JK_input.py
```diff
@@ -7,27 +7,26 @@
 C 1 20
 """)
 
+psi4.set_num_threads(1)
+memory = 50000
 primary = psi4.core.BasisSet.build(mol, "ORBITAL", "cc-pVDZ")
```
dgasmith: +3 for many actual unit tests, not just regression tests.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: Since this only scans through a 2-dimensional quantity (nao*nao), threading slows things down for smaller systems and does not boost much for larger ones.  I can revert it back and show you some timings.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: Yes, I have been eager to revert back to get proper alignment.  I often use `data()` for pointer assignment, but I was not consistent with this when I first started using `std::vector` for this code.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: whoops

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: Oh yeah, did we want to revert this code back to the previous sieving version?  The current code collects `max_` on a rolling basis (and therefore requires a critical section) whereas the previous code applies a `max_` per shell pair.  

The old code screens more and does not require a critical section.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: done

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: implemented

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: I believe I use those headers everywhere.  One thing I do that is probably not preferable is print out block info when performing a blocking operation.

psi4/src/psi4/lib3index/df_helper.cc
dgasmith: Why does this code screen more? The sieving size should be the same regardless of how you approach it.

schiebermc: My previous explanation was incorrect, let me try again:

If you look at the `prepare_sparsity()` function, you will see the two screening methods.  The uncommented code (which you highlighted above) is more of an upper bound on screening.  The reason for this is the `max_val` variable which is the maximum value amongst each shell pair.  This `max_val` is written to the masks and later checked for screening.  On the other hand, the commented code uses no such shell maxim and instead applies each individual value of the shell pair into the mask.  Therefore the values of the mask generated by the commented code can be smaller and more screening will result.

For example, the masks for the uracil dimer test I showed you:
1. Using current code:       `sig_fun_pairs = 349102`
2. Using commented code `sig_fun_pairs = 342410`
convergence is the same
 
For the linear alkene (hexadecaoctaene)
1. Using current code:       `sig_fun_pairs = 423660`
2. Using commented code `sig_fun_pairs = 415301`
convergence is the same

dgasmith: If I am reading this correctly the uncommented code is applying the shell maximum to all pairs in the function values. For the function values we want the individual contributions, as we want to screen per function not per shell. For example, look at the line

```
236: if (fun_prints[omu*nao_+onu] <= val)
```
In this case this should always be true since the `fun_prints` should always be zero until this line is reached. 

While for 
```
232: if (shell_prints[MU*pshells_+NU] <= val){
```

We hit this `nummuu x numnu` times as we are searching for the shell maxima.

The result should be identical as when we screen the function pairs we are effectively saying that single contribution is negligible. A bit surprised that the difference between shell and function screening is that small however (~2%). I would have thought orthogonal P functions (for example) would cut out more, especially in the alkene chain.

schiebermc: Yes, I agree with everything said above!  

The difference is small but adds a bit of an oopmh to the speedup, which ought to be preferable for Psi4.  Shall I revert to the function screening?

dgasmith: Yes please!

schiebermc:
Hey Daniel,

Not sure how you responded by email, but I copied and pasted the VTune
comment below.  I went to answer this question for both DF_Helper and DFJK
and I discovered something very strange.  It seems that dcopy requires
about ~6% of OpenMP time for DF_Helper.  On the other hand, it requires
nearly ~28% of OpenMP time for DFJK.  That being said, I noticed the DFJK
times are about 3-4x slower after I rebased.  Would you have any idea why
that is?


Best,

Matt


In psi4/src/psi4/lib3index/df_helper.cc
<https://github.com/psi4/psi4/pull/742#discussion_r122868449>:

> +            timer_on("DF_Helper~transform - first ");
+        // form temp, thread over spM (nao)
+        #pragma omp parallel for firstprivate(nao, naux, bsize,
block_size) private(rank) schedule(guided) \
+                                      num_threads(nthreads_)
+            for (size_t k = 0; k < nao_; k++) {
+                // truncate transformation matrix according to fun_mask
+                size_t sp_size = small_skips_[k];
+                size_t jump = (AO_core_ ? big_skips_[k] + bcount *
sp_size : (big_skips_[k] * block_size) / naux_);
+
+        #ifdef _OPENMP
+                rank = omp_get_thread_num();
+        #endif
+                for (size_t m = 0, sp_count = -1; m < nao_; m++) {
+                    if (schwarz_fun_mask_[k * nao_ + m]) {
+                        sp_count++;
+                        C_DCOPY(bsize, &Bpt[m * bsize], 1,
&C_buffers[rank][sp_count * bsize],

In VTune, whats the total (relative) time spent on this out of curiosity?

dgasmith:
@schiebermc For this PR, what else do you want to add? At some point we need to pop this into the general JK builders as well.

schiebermc:
DF_Helper has totally replaced DFERI in both MCSCF and DFEP2.

psi4/src/psi4/libfock/soscf.cc
```diff
@@ -1088,31 +1085,26 @@ SharedMatrix DFSOMCSCF::compute_Qk(SharedMatrix TPDM, SharedMatrix U, SharedMatr
...
     NNQ.reset();
 
     // Read aaQ
-    std::shared_ptr<Tensor> aaQT = dfints["aaQ"];
     SharedMatrix aaQ(new Matrix("aaQ", nact_ * nact_, nQ));
     double* aaQp = aaQ->pointer()[0];
-    FILE* aaQF = aaQT->file_pointer();
-    fseek(aaQF,0L,SEEK_SET);
-    fread(aaQp, sizeof(double), nact_ * nact_ * nQ, aaQF);
+    dfh_->fill_tensor("aaQ", aaQ);
```
dgasmith: This came together really nicely I think.

psi4/src/psi4/libfock/soscf.cc
```diff
@@ -910,43 +911,49 @@ void DFSOMCSCF::transform(bool approx_only)
...
+    SharedMatrix AO_R(new Matrix("AO_R", nao_, nrot));
+    SharedMatrix AO_a(new Matrix("AO_a", nao_, aoc_rowdim - nrot));
+    
+    double** rp = AO_R ->pointer(); 
+    double** ap = AO_a ->pointer(); 
+    
+    for(size_t i=0; i<nao_; i++){
+        C_DCOPY(nrot, &AO_Cp[i][0], 1, &rp[i][0], 1); 
+        C_DCOPY((aoc_rowdim - nrot), &AO_Cp[i][nrot], 1, &ap[i][0], 1); 
     }
```
dgasmith: Don't forget to remove this.

psi4/src/psi4/lib3index/df_helper.h
```diff
@@ -109,6 +117,14 @@ class DF_Helper {
     void fill_tensor(std::string name, SharedMatrix M, std::pair<size_t, size_t> a1,
       std::pair<size_t, size_t> a2, std::pair<size_t, size_t> a3);
 
+    // only use this one if you now what you're doing!
```
dgasmith: This should be a common operation, can you add a little info on what this does here and with the SharedMatrix. Worth commenting on how you fill the the matrix when matrix size is not equal to the tensor slice size.

psi4/src/psi4/lib3index/df_helper.h
```diff
@@ -56,38 +55,47 @@ class DF_Helper {
...
     bool get_hold_met() { return hold_met_;}
 
-    // tell me what the worst MO index size is!
+    // Tell me the worst MO size to improve blocking. (defaults to 0.5*AO_SIZE, must specify if greater!)
     void set_MO_hint(size_t wMO) {wMO_ = wMO;}
     size_t get_MO_hint() { return wMO_;}
-    // user options, must set before build---------------------------------
+    
+    // Enhanced memory use if (all Cleft = Cright). (defaults to FALSE!)
+    void set_JK_hint(bool hint) {JK_hint_ = hint;}
```
dgasmith: Is this setting for memory considerations?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -2186,51 +2872,58 @@ void DF_Helper::compute_J(std::vector<SharedMatrix> D, std::vector<SharedMatrix>
...
+        }
+        timer_off("DF_Helper~JK        - Jfirst");
+        timer_on("DF_Helper~JK       - Jreduce");
+        // reduce
+        for(size_t k=1; k<nthreads_; k++){
+            for(size_t l=0; l < naux; l++)
+                T1p[l] += T1p[k * align_size + l];
+        }
+        timer_off("DF_Helper~JK       - Jreduce");
+        timer_on("DF_Helper~JK       - Jcomple");
```
dgasmith: General point, but after you rebase you will have a new timer printer. Might be worth running to see how the new scheme looks with your labeling. You may want to rejig your names slightly.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -2239,16 +2932,21 @@ void DF_Helper::compute_K(std::vector<SharedMatrix> Cleft,
         size_t cright = Cright[i]->colspi()[0];
         double* Kp = K[i]->pointer()[0];
 
+        // cache alignment (is this over-zealous?)
```
dgasmith: Probably a bit, did this profile better?

psi4/src/psi4/libsapt_solver/usapt0.cc
```diff
@@ -2305,57 +2252,41 @@ void USAPT0::mp2_terms()
...
-        fseek(Ca_brf,0L,SEEK_SET);
-        fseek(Da_bsf,0L,SEEK_SET);
-        fseek(Ab_bsf,0L,SEEK_SET);
-        fseek(Bb_brf,0L,SEEK_SET);
-        fseek(Cb_brf,0L,SEEK_SET);
-        fseek(Db_bsf,0L,SEEK_SET);
-        for (int bstart = 0; bstart < max(nab, nbb); bstart += maxa_b) {
-            int na_bblock = (bstart + maxa_b >= nab ? nab - bstart : maxa_b);
-            int nb_bblock = (bstart + maxb_b >= nbb ? nbb - bstart : maxb_b);
+        for (int abstart = 0, bbstart = 0 ; abstart < max(nab, nbb); abstart += maxa_b, bbstart += maxb_b) {
```
dgasmith: Can you explain why you altered this loop slightly?

psi4/src/psi4/libfock/soscf.cc
schiebermc: Right?

psi4/src/psi4/lib3index/df_helper.h
schiebermc: Not sure what you mean.  If the procedure knows beforehand that all Cleft will be equal to all Cright for the J/K builds, then I will know at runtime to not declare two temporary buffers. Instead only one buffer is used and all buffers will get beefed up.

psi4/src/psi4/lib3index/df_helper.h
schiebermc: Yes better header file documentation en route!

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: Admittedly for larger systems the benefit is nominal.  The C_buffer alignment isnice as it completely partitions resources for threads when the ends of O(np) matrices share a cache line.  However this is for false sharing and not vectorization and these overlaps will become a tiny fraction for large n.

schiebermc: For smaller systems the scaling improves a bit.

schiebermc: Although I may need a more thorough testing..  But it is free so why not?

psi4/src/psi4/libsapt_solver/usapt0.cc
schiebermc: As you know DF_Helper fills tensors using tuple indexing.  However this loop did not previously keep track of the indexing for the `astart`.  This was intrinsically done for DFERI as it used `fread`, which moved the file pointer for each loop, but DF_Helper needs to keep track of the index.  The exit condition and nature of the loops is the same, i just kept track of one more variable.

schiebermc:
Rebased and pushed, some timings en route.

psi4/src/psi4/detci/ciwave.h
```diff
@@ -41,12 +41,15 @@
...
-class DFERI;
 class IntegralTransform;
 class MOSpace;
 typedef std::shared_ptr<Matrix> SharedMatrix;
 class SOMCSCF;
 
+namespace df_helper {
+class DF_Helper;
+}
+
```
dgasmith: @psi4/maintainers Out of curiosity how do we feel about this in a namespace? We usually leave it flat. Im not sure I really have a preference either way, but not being consistent isn't great.

psi4/src/psi4/detci/ints.cc
```diff
@@ -136,17 +141,27 @@ void CIWavefunction::setup_dfmcscf_ints() {
     } else {
         jk_ = JK::build_JK(basisset_, BasisSet::zero_ao_basis_set(), options_);
     }
+    
     jk_->set_do_J(true);
     jk_->set_do_K(true);
-    jk_->initialize();
     jk_->set_memory(Process::environment.get_memory() * 0.8 / sizeof(double));
+    jk_->initialize();
```
dgasmith: Whoops, great catch.

psi4/src/psi4/detci/ints.cc
```diff
@@ -724,19 +740,15 @@ void CIWavefunction::read_dpd_ci_ints() {
...
 
     // Read RaQ
-    std::shared_ptr<Tensor> RaQT = dferi_->ints()["RaQ"];
     SharedMatrix RaQ(new Matrix("RaQ", nrot, nact * nQ));
-
     double* RaQp = RaQ->pointer()[0];
-    FILE* RaQF = RaQT->file_pointer();
-    fseek(RaQF, 0L, SEEK_SET);
-    fread(RaQp, sizeof(double), nrot * nact * nQ, RaQF);
+    dfh_->fill_tensor("RaQ", RaQ);
```
dgasmith: This looks really clean now I think!

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -2768,56 +2717,40 @@ void FISAPT::felst()
...
-    AaaT.reset();
 
     std::shared_ptr<Matrix> QbC(new Matrix("QbC", nb, nQ));
     double** QbCp = QbC->pointer();
-    FILE* Abbf = AbbT->file_pointer();
-    fseek(Abbf,0L,SEEK_SET);
     for (size_t b = 0; b < nb; b++) {
-        fseek(Abbf,(b * nb + b) * nQ * sizeof(double), SEEK_SET);
-        size_t statusvalue=fread(QbCp[b], sizeof(double), nQ, Abbf);
+        dfh->fill_tensor("Abb", QbCp[b], std::make_pair(b, b + 1), std::make_pair(b, b + 1));
```
dgasmith: For this signature do you really want it as a `std::pair`? In C++17 we can skip the `make_pair`, but for now if we used `std::vectors` we could do: ` dfh->fill_tensor("Abb", QbCp[b], {b, b + 1}, {b, b + 1});`. You would have to add one more check ensuring that you are only giving it two elements. Just a comment on the syntactic sugar.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -58,10 +58,9 @@ DF_Helper::DF_Helper(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet
 {
```
dgasmith: Can you run this whole thing through clang-format when you get a chance?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -132,82 +131,97 @@ void DF_Helper::filename_maker(std::string name, size_t a0, size_t a1, size_t a2
     std::tuple<std::string, std::string> files(pfilename.c_str(), filename.c_str());
     files_[name] = files;
 
-    std::tuple<size_t, size_t, size_t> sizes = std::make_tuple(a0, a1, a2);
+    std::tuple<size_t, size_t, size_t> sizes;
+    if(op)
+        sizes = std::make_tuple(a1, a2, a0);
+    else
+        sizes = std::make_tuple(a0, a1, a2);
```
dgasmith: Please use braces here.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -239,17 +254,45 @@ void DF_Helper::prepare_sparsity()
...
+//                for (int q = 0; q < nQ; q++) {
+//                    fun_prints[(p + oP) * nao_ + (q + oQ)] =
+//                    fun_prints[(q + oQ) * nao_ + (p + oP)] = max_val;
+//                }
+//            }
+//        }
+//    }   
+//    tolerance_ = cutoff_*cutoff_/max_;
+
+    //#pragma omp parallel for simd num_threads(nthreads_) schedule(static)
```
dgasmith: Lets clean this up if you are not going to use it again.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -269,153 +312,212 @@ void DF_Helper::prepare_sparsity()
...
+        std::vector<double> metric; 
+        
+        if(!hold_met_){
+            metric.reserve(naux_ * naux_);
+            metp = metric.data();
+            std::string filename = return_metfile(mpower_);
+            get_tensor_(std::get<0>(files_[filename]), metp, 0, naux_-1, 0, naux_-1);
+        } 
+        else 
+            metp = metric_prep_core(mpower_);
```
dgasmith: More braces.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -868,114 +1122,183 @@ void DF_Helper::contract_metric(std::string file, double* metp, double* Mp, doub
...
+//            size_t skips = big_skips_[begin+j]-big_skips_[begin];
+//            C_DGEMM('N', 'N', naux_, mi, naux_, 1.0,
+//              metp, naux_, &Mp[skips], mi, 0.0, &Fp[skips], mi);
+//        }
+//
+//        // put
+//        put_tensor_AO(putf, Fp, size, count, "ab");
+//        count += size;
+//    }
+//}
```
dgasmith: Can we cleanup this as well?

psi4/src/psi4/lib3index/df_helper.cc
dgasmith: Same comment, time to do it!

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1841,10 +2471,10 @@ void DF_Helper::transpose_disk(std::string name, std::tuple<size_t, size_t, size
     for(size_t i=0; i<M0; i++){
         current += M1*M2;
         count++;
-        if(current*2 > memory_ +1000000|| i==M0-1){
+        if(current*2 > 1000000 + memory_ || i==M0-1){
```
dgasmith: Might want to use a few parentheses just to be safe here.

psi4/src/psi4/detci/ints.cc
schiebermc: Yes, all forms of `fill_tensor`, `write_disk_tensor`, etc. greatly simply I/O.  The recursive signatures work wonderfully IMO.

psi4/src/psi4/fisapt/fisapt.cc
schiebermc: Ahh, good point.  Do you want this now or later?

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: done.  poor pragmas..

psi4/src/psi4/libfock/soscf.cc
schiebermc: done

psi4/share/psi4/plugin/dfmp2/plugin.cc.template
```diff
@@ -108,17 +107,21 @@ SharedWavefunction @plugin@(SharedWavefunction ref_wfn, Options& options)
...
-    df->set_memory(memory / 8L);
-    df->print_header();
-    df->compute();
-
-    std::shared_ptr<Tensor> B = df->ints()["B"];
-
-    FILE* Bf = B->file_pointer();
-    df.reset();
+    std::shared_ptr<df_helper::DF_Helper> dfh = std::shared_ptr<df_helper::DF_Helper>(new
+    df_helper::DF_Helper(primary, auxiliary));
```
loriab: Should this be a `auto dfh (std::make_shared<df_helper::DF_Helper>(primary, auxiliary))`?

dgasmith:
@psi4/maintainers Wouldn't mind one more review on this. I know its a bit heavy, but also fairly fundamental.

psi4/share/psi4/plugin/dfmp2/plugin.cc.template
andysim: I agree with Lori here.  I think that setting an example of how to instantiate this cleanly will mean that we get cleaner implementations when people copy/paste this example.

psi4/src/psi4/detci/ciwave.h
andysim: I would probably have gone with a flat namespace, but I don't feel strongly enough to request changing it now.  As long as there are no symbol clashes (which there aren't, obviously) I don't really have much of an opinion.

psi4/src/psi4/detci/ints.cc
```diff
@@ -131,66 +133,69 @@ void CIWavefunction::setup_dfmcscf_ints() {
...
+#endif
+
     /// Build DF object
-    dferi_ = DFERI::build(get_basisset("ORBITAL"), get_basisset("DF_BASIS_SCF"), options_);
-    dferi_->set_memory(Process::environment.get_memory() * 0.8 / sizeof(double));
-    dferi_->print_header();
+    // ==> Init DF object <== /
+    dfh_ = std::shared_ptr<df_helper::DF_Helper>(
+        new df_helper::DF_Helper(get_basisset("ORBITAL"), get_basisset("DF_BASIS_SCF")));
+    dfh_->set_memory(Process::environment.get_memory() * 0.8 / sizeof(double));
```
andysim: As above, this could be cleaned up by using `make_shared` to instantiate `dfh_`.

psi4/src/psi4/dfep2/dfep2.cc
```diff
@@ -86,53 +82,56 @@ DFEP2Wavefunction::DFEP2Wavefunction(std::shared_ptr<Wavefunction> ref_wfn)
...
-    #endif
-}
+#ifdef _OPENMP
+    num_threads_ = omp_get_max_threads();
+#endif
 
-std::vector<std::vector<std::pair<double, double>>> DFEP2Wavefunction::compute(std::vector<std::vector<size_t>> solve_orbs){
+    // ==> Init DF object <== /
+    dfh_ = std::shared_ptr<df_helper::DF_Helper>(
+        new df_helper::DF_Helper(get_basisset("ORBITAL"), get_basisset("DF_BASIS_SCF")));
```
andysim: Another candidate for `make_shared`ifying

psi4/src/psi4/lib3index/df_helper.h
```diff
@@ -47,50 +46,62 @@ class Matrix;
...
 
-    // user options, must set before calling build()---------------------------------
-    // memory in doubles
-    void set_memory(size_t mem) { memory_ = mem;}
-    size_t get_memory() { return memory_; }
-
-    // workflow method (store or direct)
-    void set_method(std::string met) { method_ = met;}
+    // => user options - set these before calling initialize() <=
+    // workflow method (store or direct) (defaults to STORE)
```
andysim: I know we're horribly inconsistent with Doxygen in the rest of the code, but could you use the `///` comment marker instead of `//`, please?  That will trigger this class to be Doxygened, which could be useful for developers.

schiebermc:
Thank you!  I have addressed the aforementioned comments.

loriab:
@schiebermc was seeing problems with `scf-coverage`, but my fresh build of this branch (icpc) came back clean on full tests. So no barriers to merge on my part. Keep an eye out for `scf-coverage` errors on the 3rd test.

dgasmith:
Thanks for all the work on this Matt, pulling it in now.

loriab:
FYI, a failing `scf-coverage` has been seen again in a 3rd independent build. @carlosborca is investigating triggering compilation circumstances.

PR #743, opened 2017-06-14
Implementation of parallel timer for OpenMP threading

tyzhang1993:
## Description
It implements a new version of `timer` in `libqt`, which enables calling of `timer_on` and `timer_off` in OpenMP parallel blocks.

## Todos
* **Parallel based on original code**
  - [x] Have a working version of parallel timer
* **Introducing new C++ features and optimize performance**
  - [x] Replace the original timer data structure (doubly linked list) by std::list
  - [x] Replace old timing C functions by C++11 std::chrono library
  - [x] Optimization
* **Timer format and output**
  - [x] Nested timer
  - [x] Format output
* **Test and debug**
  - [x] Debug
* **Clean up**
  - [x] Clean up

## Questions
- [x] Is there a way to get correct thread rank in `timer_on` and `timer_off` functions rather than introducing an extra arg.

## Status
- [x] Ready to go

## Note
The signature of timer operations are:
(1) `void timer_on(const std::string& key)`
(2) `void timer_off(const std::string& key)`
(3) `void parallel_timer_on(const std::string& key, int thread_rank)`
(4) `void parallel_timer_off(const std::string& key, int thread_rank)`

In order to run timer correctly in parallel, (1) and (2) should only be called in seriel part of code and (3) and (4) should only be called in OpenMP parallel sections. Calling to (3) and (4) requires to pass an additional arg `thread_rank` unique to every thread. Generally, call
- `parallel_timer_on("your_key_string", omp_get_thread_num());`
- `parallel_timer_off("your_key_string", omp_get_thread_num());`
will work.

## An example code
```
double SCF::compute_energy()
{
    timer_on("energy");

    #pragma omp parallel
    for (int a = 0; a < 100; a++)
    {
        parallel_timer_on("inner",omp_get_thread_num());
        // some running code
        parallel_timer_off("inner",omp_get_thread_num());
    } 

    timer_off("energy");
    return 0.0;
}
```

## An example output
```
Host: DESKTOP-JOFFIR0

Timers On : Fri Jul  7 15:47:10 2017
Timers Off: Fri Jul  7 15:47:13 2017

Wall Time:        3.50 seconds

V: Grid                     :      0.333u      0.000s      0.209w      2 calls
JK: (A|mn)                  :      0.000u      0.000s      0.004w      2 calls
JK: (A|Q)^-1/2              :      0.783u      0.000s      0.256w      2 calls
JK: (Q|mn)                  :      0.000u      0.000s      0.001w      2 calls
HF: Form H                  :      0.150u      0.017s      0.007w      2 calls
HF: Form S/X                :      0.000u      0.000s      0.000w      2 calls
HF: Guess                   :      0.000u      0.000s      0.001w      2 calls
HF: Form G                  :     18.817u      0.017s      1.883w     24 calls
RKS: Form V                 :     16.567u      0.017s      1.265w     24 calls
V: Properties               :      2.047p                          14064 calls
V: Functional               :      6.249p                          14064 calls
V: V_XC                     :      0.532p                          14064 calls
V: LSDA                     :      0.182p                          28128 calls
V: GGA                      :      0.218p                          14064 calls
JK: D                       :      0.000u      0.000s      0.000w     24 calls
JK: USO2AO                  :      0.000u      0.000s      0.000w     24 calls
JK: JK                      :      2.250u      0.000s      0.616w     24 calls
JK: J                       :      0.000u      0.000s      0.000w     16 calls
JK: J1                      :      0.000u      0.000s      0.000w     16 calls
JK: J2                      :      0.000u      0.000s      0.000w     16 calls
JK: K                       :      0.000u      0.000s      0.001w     16 calls
JK: K1                      :      0.000u      0.000s      0.000w     16 calls
JK: K2                      :      0.000u      0.000s      0.000w     16 calls
JK: AO2USO                  :      0.000u      0.000s      0.000w     24 calls
HF: Form F                  :      0.000u      0.000s      0.001w     24 calls
HF: DIIS                    :      0.183u      0.000s      0.013w     24 calls
DIISManager::add_entry      :      0.183u      0.000s      0.009w     24 calls
DIISManager::extrapolate    :      0.000u      0.000s      0.003w     22 calls
DIISManager::extrapolate: bMatrix setup:      0.000u      0.000s      0.001w     22 calls
DIISManager::extrapolate: bMatrix pseudoinverse:      0.000u      0.000s      0.001w     22 calls
DIISManager::extrapolate: form new data:      0.000u      0.000s      0.001w     22 calls
HF: Form C                  :      0.217u      0.000s      0.002w     24 calls
HF: Form D                  :      0.000u      0.000s      0.000w     24 calls

-----------------------------------------------------------
V: Grid                     :      0.333u      0.000s      0.209w      2 calls
JK: (A|mn)                  :      0.000u      0.000s      0.004w      2 calls
JK: (A|Q)^-1/2              :      0.783u      0.000s      0.256w      2 calls
JK: (Q|mn)                  :      0.000u      0.000s      0.001w      2 calls
HF: Form H                  :      0.150u      0.017s      0.007w      2 calls
HF: Form S/X                :      0.000u      0.000s      0.000w      2 calls
HF: Guess                   :      0.000u      0.000s      0.001w      2 calls
HF: Form G                  :     18.817u      0.017s      1.883w     24 calls
| RKS: Form V               :     16.567u      0.017s      1.265w     24 calls
| | V: Properties           :      2.047p                          14064 calls
| | V: Functional           :      6.249p                          14064 calls
| | V: V_XC                 :      0.532p                          14064 calls
| | | V: LSDA               :      0.182p                          28128 calls
| | | V: GGA                :      0.218p                          14064 calls
| JK: D                     :      0.000u      0.000s      0.000w     24 calls
| JK: USO2AO                :      0.000u      0.000s      0.000w     24 calls
| JK: JK                    :      2.250u      0.000s      0.616w     24 calls
| | JK: J                   :      0.000u      0.000s      0.000w     16 calls
| | | JK: J1                :      0.000u      0.000s      0.000w     16 calls
| | | JK: J2                :      0.000u      0.000s      0.000w     16 calls
| | JK: K                   :      0.000u      0.000s      0.001w     16 calls
| | | JK: K1                :      0.000u      0.000s      0.000w     16 calls
| | | JK: K2                :      0.000u      0.000s      0.000w     16 calls
| JK: AO2USO                :      0.000u      0.000s      0.000w     24 calls
HF: Form F                  :      0.000u      0.000s      0.001w     24 calls
HF: DIIS                    :      0.183u      0.000s      0.013w     24 calls
| DIISManager::add_entry    :      0.183u      0.000s      0.009w     24 calls
| DIISManager::extrapolate  :      0.000u      0.000s      0.003w     22 calls
| | DIISManager::extrapolate: bMatrix setup:      0.000u      0.000s      0.001w     22 calls
| | DIISManager::extrapolate: bMatrix pseudoinverse:      0.000u      0.000s      0.001w     22 calls
| | DIISManager::extrapolate: form new data:      0.000u      0.000s      0.001w     22 calls
HF: Form C                  :      0.217u      0.000s      0.002w     24 calls
HF: Form D                  :      0.000u      0.000s      0.000w     24 calls

***********************************************************
```


jturney:
I think you should be able to use `omp_get_thread_num()` to determine the thread rank. Refer to [this](https://gcc.gnu.org/onlinedocs/libgomp/omp_005fget_005fthread_005fnum.html).

jturney:
I'd like to have the ability to nest timers:

```
Timing information:

  157 ms : 1 calls : 157 ms per call :                        1. Wbaef = T1mb * Gmaef
    157 ms : 1 calls : 157 ms per call :                        #: 0.000000 Wbaef[b,a,e,f] = 1.000000 T1[m,b] * Giabc[m,a,e,f]
      157 ms : 1 calls : 157 ms per call :                        BLAS
      0 ms : 1 calls : 0 ms per call :                            pre-BLAS: internal overhead
  1308 ms : 1 calls : 1308 ms per call :                      2. Wabef = T1mb * Gamef
    1308 ms : 1 calls : 1308 ms per call :                      #: 0.000000 Wabef[a,b,e,f] = 1.000000 T1[m,b] * Gaibc[a,m,e,f]
      166 ms : 1 calls : 166 ms per call :                        BLAS
      250 ms : 1 calls : 250 ms per call :                        post-BLAS: internal C permutation
        250 ms : 1 calls : 250 ms per call :                        P: 0.000000 C2[a,b,e,f] = 1.000000 C2[b,a,e,f]
      82 ms : 1 calls : 82 ms per call :                          pre-BLAS: internal B allocation
      66 ms : 1 calls : 66 ms per call :                          pre-BLAS: internal B permutation
        66 ms : 1 calls : 66 ms per call :                          P: 0.000000 Gaibc[a,e,f,m] = 1.000000 Gaibc[a,m,e,f]
      666 ms : 1 calls : 666 ms per call :                        pre-BLAS: internal C allocation
      0 ms : 1 calls : 0 ms per call :                            pre-BLAS: internal overhead
  1043 ms : 1 calls : 1043 ms per call :                      3. Waebf = T1mb * Gaemf
    1043 ms : 1 calls : 1043 ms per call :                      #: 0.000000 Waebf[a,e,b,f] = 1.000000 T1[m,b] * Gabic[a,e,m,f]
      188 ms : 1 calls : 188 ms per call :                        BLAS
      285 ms : 1 calls : 285 ms per call :                        post-BLAS: internal C permutation
        285 ms : 1 calls : 285 ms per call :                        P: 0.000000 C2[a,e,b,f] = 1.000000 C2[b,a,e,f]
      86 ms : 1 calls : 86 ms per call :                          pre-BLAS: internal B allocation
      77 ms : 1 calls : 77 ms per call :                          pre-BLAS: internal B permutation
        77 ms : 1 calls : 77 ms per call :                          P: 0.000000 Gabic[a,e,f,m] = 1.000000 Gabic[a,e,m,f]
      334 ms : 1 calls : 334 ms per call :                        pre-BLAS: internal C allocation
      0 ms : 1 calls : 0 ms per call :                            pre-BLAS: internal overhead
  257 ms : 1 calls : 257 ms per call :                        4. Waefb = T1mb * Gaefm
    257 ms : 1 calls : 257 ms per call :                        #: 0.000000 Waefb[a,e,f,b] = 1.000000 T1[m,b] * Gabci[a,e,f,m]
      257 ms : 1 calls : 257 ms per call :                        BLAS
      0 ms : 1 calls : 0 ms per call :                            pre-BLAS: internal overhead
  1606 ms : 9 calls : 178 ms per call :                       Tensor::build
```

You can refer to [ambit's timer](https://github.com/jturney/ambit/blob/master/src/tensor/timer.cc) for how I do it there.

tyzhang1993:
Thank @jturney !
However, `omp_get_thread_num()` always return 0 in this code when I call `timer_on()` from `forte` plugin. I don't understand why. Maybe because `forte` and psi4 are different libraries so that when psi4 is called by `forte`, it cannot figure out which thread it is on. Just my guess.

fevangelista:
@tyzhang1993: might it be snippets like this in `forte-def.h`?
```
#ifdef _OPENMP
#include <omp.h>
#else
#define omp_get_max_threads() 1
#define omp_get_thread_num() 0
#define omp_get_num_threads() 1
#endif
```

tyzhang1993:
@fevangelista I have similar code in `timer.cc`, and have identified that `_OPENMP` is defined in compilation. Haven't identified the problem.

jturney:
`omp_get_thread_num()` will work even in a plugin:

```
void timer_on()
{
    printf("timer thread rank: %d\n", omp_get_thread_num());
}

extern "C"
SharedWavefunction play(SharedWavefunction ref_wfn, Options& options)
{
    int print = options.get_int("PRINT");

    int nthreads, tid;

    /* Fork a team of threads giving them their own copies of variables */
    #pragma omp parallel for private(nthreads, tid)
    for (int a = 0; a < 8; a++)
    {

        /* Obtain thread number */
        tid = omp_get_thread_num();
        printf("Hello World from thread = %d\n", tid);
        timer_on();

        /* Only master thread does this */
        if (tid == 0)
        {
            nthreads = omp_get_num_threads();
            printf("Number of threads = %d\n", nthreads);
        }
    } /* All threads join master thread and disband */


    // Typically you would build a new wavefunction and populate it with data
    return ref_wfn;
}
```

You must call `psi4 -n 4` or some other number because psi4 tells OpenMP to use 1 thread by default.

```
Attention! This SCF may be density-fitted.
Hello World from thread = 0
Hello World from thread = 3
Hello World from thread = 2
Hello World from thread = 1
timer thread rank: 0
timer thread rank: 3
timer thread rank: 2
timer thread rank: 1
Number of threads = 4
Hello World from thread = 3
Hello World from thread = 2
Hello World from thread = 1
Hello World from thread = 0
timer thread rank: 3
timer thread rank: 2
timer thread rank: 1
timer thread rank: 0
Number of threads = 4
```

tyzhang1993:
@jturney It is true if you have `timer` and `#pragma omp parallel` both in the plugin.
However, if I have
```
void timer_on(const char *key, int thread_rank)
{
  ...
  if (thread_rank != 0 and thread_rank != omp_get_thread_num()) {
    printf("\nTimer_on::Thread rank does not agree! arg=%d, %d=omp_get_thread_num()",thread_rank, omp_get_thread_num());
  }
  ...
}
```
in `/psi4/src/psi4/libqt/timer.cc`, and a plugin with
```
double SCF::compute_energy()
{
    timer_on("energy");

    #pragma omp parallel
    for (int a = 0; a < 100; a++)
    {
        timer_on("inner",omp_get_thread_num());
        printf("Hello World from thread = %d\n", omp_get_thread_num());
        timer_off("inner",omp_get_thread_num());
    } 
    timer_off("energy");
    return 0.0;
}
```
It will print:
```
Timer_on::Thread rank does not agree! arg=2, 0=omp_get_thread_num()
```
That's what I don't understand here.

jturney:
Still works:

```
psi4: timer_on: 0 thread
psi4: timer_on: 1 thread
psi4: timer_on: 2 thread
psi4: timer_on: 3 thread
Hello World from thread = 0
Hello World from thread = 1
Hello World from thread = 2
Hello World from thread = 3
Number of threads = 4
psi4: timer_on: 2 thread
psi4: timer_on: 1 thread
psi4: timer_on: 3 thread
psi4: timer_on: 0 thread
Hello World from thread = 2
Hello World from thread = 1
Hello World from thread = 3
Hello World from thread = 0
Number of threads = 4
```

Plugin:
```
extern "C"
SharedWavefunction play(SharedWavefunction ref_wfn, Options& options)
{
    int print = options.get_int("PRINT");

    int nthreads, tid;

    /* Fork a team of threads giving them their own copies of variables */
    #pragma omp parallel for private(nthreads, tid)
    for (int a = 0; a < 8; a++)
    {
        std::string timer_name = "play " + std::to_string(a);
        timer_on(timer_name.c_str());

        /* Obtain thread number */
        tid = omp_get_thread_num();
        printf("Hello World from thread = %d\n", tid);
        //timer_on();

        /* Only master thread does this */
        if (tid == 0)
        {
            nthreads = omp_get_num_threads();
            printf("Number of threads = %d\n", nthreads);
        }

        timer_off(timer_name.c_str());

    } /* All threads join master thread and disband */


    // Typically you would build a new wavefunction and populate it with data
    return ref_wfn;
}
```

Change to `timer_on` timer.cc in libqt with corresponding `#include <omp.h>`:
```
  printf("psi4: timer_on: %d thread\n", omp_get_thread_num());
```

Then executed with `psi4 -n 4`.

Not sure what's going on with your system.

psi4/src/psi4/libqt/qt.h
```diff
@@ -85,8 +85,8 @@ void mat_print(double **A, int rows, int cols, std::string OutFileRMR);
 
 void timer_init(void);
 void timer_done(void);
-void timer_on(const char *key);
-void timer_off(const char *key);
+void timer_on(const char *key, int thread_rank = 0);
+void timer_off(const char *key, int thread_rank = 0);
```
dgasmith: Can you switch these to `std::string` over `char *`? Trying to kill off `char *` in Psi4.

psi4/src/psi4/libqt/timer.cc
```diff
@@ -66,243 +67,997 @@
...
-  char *host;
-  extern time_t timer_start, timer_end;
-  struct timer *this_timer, *next_timer;
-  extern struct timer *global_timer;
-
-  timer_end = time(NULL);
-
-  host = (char *) malloc(40 * sizeof(char));
-  gethostname(host, 40);
+public:
```
dgasmith: Can you reformat this with 4-space indentions? Simplest way would be to use `clang-format` with google style options (4-spaces).

psi4/src/psi4/libqt/qt.h
tyzhang1993: Sure, it is easy.

psi4/src/psi4/libqt/timer.cc
tyzhang1993: Thanks! It is done.

dgasmith:
I really like the way this turned out, thanks for the example. It looks like we should probably go back and look into the timing print outs a bit. For example:

```
| | DIISManager::extrapolate:       0.00u       0.00s   0.000530w     10 calls
| | | DIISManager::extrapolate: bMatrix setup:       0.00u       0.00s   0.000085w     10 calls
```

Since `bMatrix setup` is only called with `extrapolate` we can safely shorten this text so that we do not skew the columns. This can be done in later passes however.

 Did the threading issue get resolved?

tyzhang1993:
@dgasmith Current solution to the threading issue is to pass an additional arg `thread_rank` unique to every thread. I don't know how to get the correct `thread_rank` in the `timer_on` function. I think this can be modified if we find a better way in the future.

dgasmith:
Sure, I'm happy with that for now. On that regard can you switch on `psi4/libfock/v.cc:483,525` and the like timer functions back on to test out the new functionality? The rank information is unsurprisingly called `rank`.

tyzhang1993:
@dgasmith By running test dft2 I got:
```
HF: Form G                  :      18.68u       0.00s   1.955138w     24 calls
| RKS: Form V               :      16.30u       0.00s   1.255284w     24 calls
| | V: Properties           :                           2.093465p  14064 calls
| | V: LSDA                 :                           0.074432p  14064 calls
| JK: D                     :       0.00u       0.00s   0.000135w     24 calls
```
after the corresponding modifications.

dgasmith:
Cool! Thats really neat and I see whats going on now. Can we actually tuck `p` (for parallel?) under `u` time as this is what this somewhat represents? It looks a little odd now since the parent took `1.2s`(Form V) and the children (V: Properties/LSDA) took `~2.2s`. 

If you wouldn't mind could you also make all of the timers in `v.cc` live in this way?

tyzhang1993:
Oh, it is because the parent is the wall time, but the child one is the parallel time (ending `p`, actually cpu time). I will change the parallel time printing under the same column of cpu time (ending `u`).

tyzhang1993:
I got
```
HF: Form G                  :      18.28u       0.00s   1.871290w     23 calls
| RKS: Form V               :      16.18u       0.00s   1.252431w     23 calls
| | V: Properties           :   2.017924p                          13478 calls
| | | V: Functional         :   1.220899p                           2682 calls
| | V: Functional           :   4.852314p                          10796 calls
| | | V: V_XC               :   0.305338p                           7999 calls
| | | | V: LSDA             :   0.108765p                          15998 calls
| | | | V: GGA              :   0.126083p                           7999 calls
| | V: V_XC                 :   0.205600p                           5479 calls
| | | V: LSDA               :   0.072038p                          10958 calls
| | | V: GGA                :   0.084044p                           5479 calls
| JK: D                     :       0.00u       0.00s   0.000137w     23 calls
```
with all the timers in `v.cc` turned on.
I think I want to make the output numbers the same number of significant figures so that they will look better together.

dgasmith:
Looks like we have some issues with the nesting when we run these in serial. Properties, Functional, and V_XC should all be on the same level with LSDA and GGA nested in V_XC. Hopefully this could be verified with a serial run. Could we give the nesting hints in these cases?

tyzhang1993:
@dgasmith A serial run prints:
```
HF: Form G                  :       8.92u       0.10s   5.404107w     22 calls
| RKS: Form V               :       8.73u       0.10s   5.286138w     22 calls
| | V: Properties           :       1.93u       0.00s   1.173858w  12892 calls
| | V: Functional           :       6.27u       0.00s   3.723876w  12892 calls
| | V: V_XC                 :       0.42u       0.03s   0.316659w  12892 calls
| | | V: LSDA               :       0.15u       0.00s   0.107187w  25784 calls
| | | V: GGA                :       0.18u       0.03s   0.119156w  12892 calls
| JK: D                     :       0.00u       0.00s   0.000112w     22 calls
```
Is the correct? Is the timer calls different in serial and parallel?

dgasmith:
The serial run is correct. I think your automatic nesting scheme is breaking down as some of the above can turn off/on before other pieces as each block is not a static amount of work. I wonder if we can create a region where we freeze nesting like `timer_nesting_freeze(true)`. We would loose the LSDA/GGA nesting, but I think thats ok.

tyzhang1993:
@dgasmith It means I have a bug in my code. It is supposed to solve the correct nesting pattern no matter serial or parallel. I will look into this issue.
Thank you for providing this test.

tyzhang1993:
@dgasmith The problem was not that nesting cannot be solved correctly in parallel. The difficulty was actually there was no way to identify between which `timer_on` calls on thread 0 did the OpenMP parallel section starts. In previous versions, there were assumptions to guess the starting point of the parallel section, but no guess can deal with the situation where a timer on thread 0 start the first and end the last.
The newly pushed solution (as described in the pull request description) distinguishes serial timer calls and parallel timer calls, which gives an explicit hint where the parallel section starts and solves the problem.
The only disadvantage is that in parallel section. Instead of calling to `timer_on` and `timer_off`, `parallel_timer_on` and `parallel_timer_off` should be called. They actually can also be distinguished by whether an additional `thread_rank` arg is passed (`timer_on(key)` vs `parallel_timer_on(key, rank)`), but this solution may be confusing and misleading.

tyzhang1993:
@dgasmith The example output in PR description is the corresponding print out for updated `v.cc` in 8 threads.

dgasmith:
Great, thanks.

General question: What if we use a static print width rather than a responsive one. The print statement would move to something like `%10.3f` or the like. That gives us ms accuracy which is probably about as accurate as the timer really is.

tyzhang1993:
That's a good suggestion.
In the implementation, the `utime` and `stime` are using those defined in `<sys/times.h>`, and `wtime` is using `std::chrono::high_resolution_clock`. It seems that the resolution of `utime` and `stime` are around ms level, but `wtime` can be much more accurate.
It is easy to change the format, but I am not sure which is better.

tyzhang1993:
@dgasmith The number format as already been changed to `%10.3f`.

psi4/src/psi4/libqt/qt.h
```diff
@@ -86,8 +86,10 @@ void mat_print(double **A, int rows, int cols, std::string out_fname);
 
 void timer_init(void);
 void timer_done(void);
-void timer_on(const char *key);
-void timer_off(const char *key);
+void timer_on(std::string key);
+void timer_off(std::string key);
```
andysim: I don't think the timer is intended to be used for ultra fine grained purposes but, out of paranoia, shouldn't the key be passed by reference to avoid the tiny amount of copying overhead?  It's not likely to be an issue, but I see no harm in doing that, at first glance.

tyzhang1993: Thanks! I am changing immediately.

PR #746, opened 2017-06-20
Remove pybind from Libmints

andysim:
## Description
Pybind includes are gone from libmints and a couple of other locations, to speed up compilation.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/libmints/basisset.h
+++ b/psi4/src/psi4/libmints/basisset.h
@@ -35,8 +35,6 @@
 #include "psi4/libmints/typedefs.h"
 #include "psi4/psi4-dec.h"
 
-#include "psi4/pybind11.h"
-
 #include <cstdio>
 #include <string>
 #include <vector>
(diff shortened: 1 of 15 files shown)


psi4/src/psi4/libmints/basisset.h
```diff
@@ -390,7 +396,7 @@ class BasisSet
      * @param py::dict      Python dictionary containing the basis information
      * @param forced_puream Force puream or not
     **/
-    static std::shared_ptr<BasisSet> construct_from_pydict(const std::shared_ptr <Molecule> &mol, py::dict pybs, const int forced_puream);
+    static std::shared_ptr<BasisSet> construct_from_pydict(const std::shared_ptr <Molecule> &mol, py::dict &pybs, const int forced_puream);
```
loriab: Declaring fn in class header but not implementing it until export_mints — I didn't know you could do that.

andysim:
Works fine on my laptop, but it looks like i need to hop onto my linux box for more testing...

psi4/src/psi4/libmints/basisset.h
andysim: In this case, it'll be undefined in libmints.a, but defined in core.so.  If somebody wanted to link in libmints.a to another application, this could be problematic.  However, the static function is specific to out Python interface, so it's fine for it to only be defined in core.so, IMHO.

andysim: Having said that, it'd probably be best to simply remove this function from BasisSet entirely, and just make it a static function that only lives in `export_mints.cc`.  That would avoid the janky forward declaration machinery in `basisset.h`.

andysim:
After indecision on my part, and pickiness from clang, this is ready to go.  It should be squashed when merging.

PR #747, opened 2017-06-26
Proper update of post-scf wave-functions for fchk interface

ashutoshvt:
## Description
This pull request aims to properly update energy, orbital energies, densities etc of the wave function for (conventional and density-fitted Mp2) and density-fitted CCSD calculations, so as to ensure proper checkpointing using FCHK interface. 
Also, there seems to be a problem in using FCHKWriter for conventional mp2 gradient calculations as I get a segmentation fault. This PR is meant to fix these problems which was pointed out by abdale on the psi4 forum. http://forum.psicode.org/t/problems-with-wavefunction-object-and-fchk-interface/532

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -229,6 +229,7 @@ double DFMP2::compute_energy()
     form_energy();
     timer_off("DFMP2 Energy");
     print_energies();
+    energy_ = variables_["MP2 TOTAL ENERGY"];
 
     return variables_["MP2 TOTAL ENERGY"];
 }
@@ -314,6 +315,7 @@ SharedMatrix DFMP2::compute_gradient()
 
     if(options_.get_bool("ONEPDM")){
         print_energies();
+        energy_ = variables_["MP2 TOTAL ENERGY"];
         return SharedMatrix(new Matrix("NULL", 0, 0));
     }
 
@@ -322,6 +324,7 @@ SharedMatrix DFMP2::compute_gradient()
     timer_off("DFMP2 grad");
 
     print_energies();
+    energy_ = variables_["MP2 TOTAL ENERGY"];
 
     print_gradients();
 
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -2310,6 +2310,9 @@ void RDFMP2::form_Z()
     SharedMatrix AP(new Matrix("A_mn^ls P_ls^(2)", nso, nso));
     double** APp = AP->pointer();
     SharedMatrix Dtemp;
+    SharedMatrix Ca__;
+    SharedVector epsilon_a__;
+    
 
     if(options_.get_bool("OPDM_RELAX")){
         psio_->read_entry(PSIF_DFMP2_AIA, "W", (char*) Wpq1p[0], sizeof(double) * nmo * nmo);
@@ -2403,9 +2406,10 @@ void RDFMP2::form_Z()
         // Add in the reference contribution
         for (int i = 0; i < nocc; ++i)
             Dtemp->add(i, i, 1.0);
-        Ca_ = SharedMatrix(new Matrix("DF-MP2 Natural Orbitals", nsopi_, nmopi_));
-        epsilon_a_ = SharedVector(new Vector("DF-MP2 NO Occupations", nmopi_));
+        Ca__ = SharedMatrix(new Matrix("DF-MP2 Natural Orbitals", nsopi_, nmopi_));
+        epsilon_a__ = SharedVector(new Vector("DF-MP2 NO Occupations", nmopi_));
         Da_ = SharedMatrix(new Matrix("DF-MP2 relaxed density", nsopi_, nsopi_));
+        
     }else{
         // Don't relax the OPDM
         Dtemp = Ppq->clone();
@@ -2414,16 +2418,18 @@ void RDFMP2::form_Z()
         Dtemp->scale(0.5);
 
         // Add in the reference contribution
+
         for (int i = 0; i < nocc; ++i)
             Dtemp->add(i, i, 1.0);
-        Ca_ = SharedMatrix(new Matrix("DF-MP2 (unrelaxed) Natural Orbitals", nsopi_, nmopi_));
-        epsilon_a_ = SharedVector(new Vector("DF-MP2 (unrelaxed) NO Occupations", nmopi_));
+        Ca__ = SharedMatrix(new Matrix("DF-MP2 (unrelaxed) Natural Orbitals", nsopi_, nmopi_));
+        epsilon_a__ = SharedVector(new Vector("DF-MP2 (unrelaxed) NO Occupations", nmopi_));
         Da_ = SharedMatrix(new Matrix("DF-MP2 unrelaxed density", nsopi_, nsopi_));
+        
     }
 
-    compute_opdm_and_nos(Dtemp, Da_, Ca_, epsilon_a_);
-    Cb_ = Ca_;
-    epsilon_b_ = epsilon_a_;
+    compute_opdm_and_nos(Dtemp, Da_, Ca__, epsilon_a__);
+    //Cb_ = Ca_;
+    //epsilon_b_ = epsilon_a_;
     Db_ = Da_;
 
     if(options_.get_bool("ONEPDM")){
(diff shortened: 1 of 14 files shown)


psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -2411,16 +2418,18 @@ void RDFMP2::form_Z()
         Dtemp->scale(0.5);
 
         // Add in the reference contribution
+
         for (int i = 0; i < nocc; ++i)
             Dtemp->add(i, i, 1.0);
-        Ca_ = SharedMatrix(new Matrix("DF-MP2 (unrelaxed) Natural Orbitals", nsopi_, nmopi_));
-        epsilon_a_ = SharedVector(new Vector("DF-MP2 (unrelaxed) NO Occupations", nmopi_));
+        Ca__ = SharedMatrix(new Matrix("DF-MP2 (unrelaxed) Natural Orbitals", nsopi_, nmopi_));
+        epsilon_a__ = SharedVector(new Vector("DF-MP2 (unrelaxed) NO Occupations", nmopi_));
```
dgasmith: I would think that we actually want the non relaxed natural orbitals for better guesses downstream. I wouldn't add a Ca__ attribute as it's difficult to tell between Ca_ and Ca__.

ashutoshvt: As you pointed out, many of the post-scf methods don't set their energy_ and just update the process environment variable. This needs a major overhaul. Here, I just wanted to fix the problem faced by a user on the forum. Also, I was uncertain if we needed the unrelaxed natural orbitals coefficients and occupation numbers as I didn't see them getting used anywhere. How do you suggest I retain these as I want the Ca_ and epsilon_a_ to be the conventional MO coefficients and orbital energies respectively ?

dgasmith: I don't think we want to retain `Ca_` and `epsilon_a_` as the original SCF values, its ok if Wavefunctions update themselves. For example, the MCSCF Wavefunction rotates its own orbitals.

These values will not get used within the MP2 code, but are useful for things like:
```
set mp2_opdm true # or whatever
mp2_grad, mp2_wfn = psi4.energy("MP2")
psi4.energy("OCCSD", ref_wfn = mp2_wfn)
```

The relaxed densities are odd because they rotate the `occupied x virtual` parts of the OPDM while the unrelaxed density only rotates the `occupied x occupied` and `virtual x virtual` areas. The former actually changes the Wavefunction while the latter helps compress the information. (Hopefully I didn't screw this up, its been awhile since I have done MP2 densities)

ashutoshvt: I see. However, its confusing when one uses the fchk_interface as `Ca_` and `epsilon_a_` which are natural orbitals and occupation numbers get printed as `Alpha MO coefficients` and `Alpha Orbital Energies` respectively in the checkpoint file. Do you think we should provide the original MO coefficients and orbital energies of the reference wavefunction as well in the df-mp2 gradient case since all the other modules do the same?

dgasmith: I don't think all other modules provide the original MO coefficients, in fact in many cases they alter the coefficients as described above. If someone wants SCF level `fchk`, they should give the `fchk` a SCF Wavefunction? We want the `fchk` written to write the current state of the Wavefunction, regardless of what that state is I would think.

ashutoshvt: You are right, I was thinking of including the SCF Wavefunction data for post-scf methods which  i guess can be done quite easily from the python's side:
```
scf_e, scf_wfn = energy('scf', return_wfn=True)
scf_fchk_writer = core.FCHKWriter(scf_wfn)
scf_fchk_writer.write('scf.fchk')
mp2_e, mp2_wfn = gradient('mp2', ref_wfn = scf_wfn, return_wfn=True)
mp2_fchk_writer = core.FCHKWriter(mp2_wfn)
mp2_fchk_writer.write('mp2.fchk')
```
Also, do you think we should change the name `Alpha Orbital Energies` which actually are the occupation numbers in the case of df-mp2 to something more suitable?

dgasmith: Good question. @andysim @loriab I forget who wrote the `fchk` interface. I unfortunately do not know a lot about the spec.

loriab: Its not unheard of to have occ num in place of energies. That's how I confirmed was dealing with natorbs in cfour. There does seem to be an `Occup` slot in Molden spec, though.

```
 Sym= symmetry_label_1
 Ene= mo_energy_1
 Spin= (Alpha|Beta)
 Occup= mo_occupation_number_1
 ao_number_1 mo_coefficient_1
 ...
 ao_number_n mo_coefficient_n
 ....
```

ashutoshvt: Thanks for the reply @loriab I was wondering if we can print the name of the matrix on the fchk file, something like 
`wfn_->Ca()->name()`
 instead of hardcoding it like "Alpha Orbital Energies"
`write_matrix("Alpha Orbital Energies", wavefunction_->epsilon_a_subset("AO"));` 
This way it will reflect whatever change was made by the corresponding wavefunction ?

ashutoshvt: I will push appropriate changes in this regard which would make it easier to get feedback from you guys.

ashutoshvt:
@dgasmith I think this is ready for feedback.

psi4/src/psi4/libscf_solver/rhf.cc
```diff
@@ -87,9 +87,10 @@ void RHF::common_init()
     // Allocate matrix memory
     Fa_        = SharedMatrix(factory_->create_matrix("F"));
     Fb_        = Fa_;
-    Ca_        = SharedMatrix(factory_->create_matrix("C"));
+    Ca_        = SharedMatrix(factory_->create_matrix("MO coefficients (C)"));
```
dgasmith: I think these new names must be identical to that of the replaced names in `writer.cc` for this for `fichk` to work?

ashutoshvt: `fchk` will work even if the names were different. For example, If one looks at mp2.cc inside dfmp2, ```Ca_ = SharedMatrix(new Matrix("DF-MP2 Natural Orbitals", nsopi_, nmopi_));```
I wanted this to be reflected when we write the wfn's info to the `fchk_writer` If the wfn doesn't do anything special, the names would stay the same as reference wfn's. Does this make sense?

psi4/src/psi4/dfocc/dfocc.cc
```diff
@@ -33,6 +33,8 @@
 #include "defines.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
 #include "psi4/liboptions/liboptions.h"
+#include "psi4/libpsi4util/process.h"
```
loriab: Should this header go in `dfocc/manager.cc` so as to minimize its impact on the build?

ashutoshvt:
@loriab Yup, removed that header now.

dgasmith:
LGTM @andysim Can you check this one out, I think you may know a bit about `fchk`.

psi4/src/psi4/fnocc/df_ccsd.cc
```diff
@@ -102,6 +102,9 @@ double DFCoupledCluster::compute_energy() {
   Process::environment.globals["CCSD SAME-SPIN CORRELATION ENERGY"] = eccsd_ss;
   Process::environment.globals["CCSD TOTAL ENERGY"] = eccsd + escf;
   Process::environment.globals["CURRENT ENERGY"] = eccsd + escf;
+  /* updates the wavefunction for checkpointing */
+  energy_ = Process::environment.globals["CCSD TOTAL ENERGY"];
+  name_ = "DF-CCSD";
 
   if (options_.get_bool("COMPUTE_TRIPLES")){
       long int o = ndoccact;
```
andysim: If we get in here, shouldn't the name be updated to `DF-CCSD(T)`?

ashutoshvt: thanks for pointing that out @andysim. I took this opportunity to update the names and energies of all the wfns inside `manager.cc`

ashutoshvt:
I agree, there are no well-defined standards as of yet. @dgasmith suggested that one can do a scf calculation to get scf level properties as this info is meant for the state of the wfn requested.

dgasmith:
I think in terms of changes for this particular patch this is all ready to go. Any objections?

ashutoshvt:
No objections from my side.

PR #751, opened 2017-06-30
Numpy update

dgasmith:
## Description
Occasionally we had issues cropping up where users were assigning views that were not tied to the Matrix object. Consider the following:

```
mat = Matrix(5, 5)
view = mat.np
del mat
```

Previously, `mat` and `view` lifetimes were not tied together. The NumPy array would point to deleted memory in the above example. With this PR, the object lifetime are tied together and `mat` is not actually deleted in this example (just a handle to it) so that the `view` remains valid. This also allows operations like `view = Matrix(5, 5).np` and `Cocc = wfn.Ca_subset("AO", "OCC").np` to work without issue.

A small downside is that the views cannot be cached as this would lead to circular nurse/patient relationships and prevent the objects from ever being garbage collected. This makes the `.np` accessors take slightly more time (~2-5us depending on number of irreps vs 0.5us). However, unless you are looping over individual indices and setting values this really shouldn't matter and the `mat.get/mat.set` functionality can be used for this with the ~0.4us latency. There is a chance this is fixable and we can revisit if it comes an issue.

I also was able to delete a great deal of rather hideous code which makes me quite happy.

Special thanks to @rmcgibbo for poking me about this and helping with idea and also to the PyBind11 team for "fixing" this a few months ago (doh!).

## Todos
* **Developer Interest**
  - [x] Can now obtain stand-alone views with correct reference counting (`view = Matrix(5, 5).np`).
  - [x] Dimension objects can now be built python-side from list `Dimension([5, 5, 5])`.
  - [x] The NumPy interface is now more robust. 

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/libpsi4util/PsiOutStream.cc
+++ b/psi4/src/psi4/libpsi4util/PsiOutStream.cc
@@ -51,6 +51,8 @@ PsiOutStream::PsiOutStream(std::string fname, std::ios_base::openmode mode) {
         is_cout_ = false;
     }
 
+    // This is sort of big, but vsnprintf does not appear to work correctly on all OS's.
+    // Im looking at YOU CentOS
     buffer_.resize(512000);
 }
 
@@ -61,8 +63,7 @@ PsiOutStream::~PsiOutStream() {
 }
 
 void PsiOutStream::Printf(const char* format, ...) {
-    // We don't know how long the fully expanded string is so lets guess our average print is about
-    // a line
+    // We can check if the buffer is large enough
     va_list args;
     va_start(args, format);
     int left = vsnprintf(buffer_.data(), buffer_.size(), format, args);
@@ -81,10 +82,10 @@ void PsiOutStream::Printf(const char* format, ...) {
     // Everything is cool
 
     va_end(args);
-    (*stream_) << buffer_.data();
+    (*stream_) << buffer_.data() << std::flush;
 }
 void PsiOutStream::Printf(std::string fp) {
-    (*stream_) << fp;
+    (*stream_) << fp << std::flush;
 }
 
 } // End Psi Namespace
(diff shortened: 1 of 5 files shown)


rmcgibbo:
`py::return_value_policy::reference_intern` FTW. This should probably revert the extra copy in #750 too.

dgasmith:
The real magic is in the `py::array` call where you pass it a handle to the matrix/vector object its spawning from. Im not completely sure the `reference_internal` is doing anything here since the spawned `keep_alive<0, 1>` function is trying to connect two weak pointers which might not be possible.

I actually kind of like the copy in `#750` it prevents the rather awkward problem of:
```
mat = Matrix(5, 5)
data = mat.write(None, "my_mat"
...

mat.np[:] = 5
```

rmcgibbo:
Oh yeah, there is still the "snapshotting" feature of it.

psi4/src/psi4/libpsi4util/PsiOutStream.cc
```diff
@@ -51,6 +51,8 @@ PsiOutStream::PsiOutStream(std::string fname, std::ios_base::openmode mode) {
         is_cout_ = false;
     }
 
+    // This is sort of big, but vsnprintf does not appear to work correctly on all OS's.
+    // Im looking at YOU CentOS
     buffer_.resize(512000);
```
andysim: I think this large value was randomly chosen by me when investigating one of your errors in a much older PR.  Does it need to be this large?  Probably not a big deal either way, but figured I'd check

dgasmith: Probably doesn't need to be this large, but at 0.5 MB it is probably not hurting anything incase someone does a massive print statement. The `vsnprintf` could be related to [this](https://stackoverflow.com/questions/29369974/vsnprintf-with-s-segfaults-with-gcc).

dgasmith:
While not formally part of our tests all Psi4NumPy and SAPT(CASSCF) test cases pass with these changes.

PR #753, opened 2017-06-30
Add clang-format/YAPF style files

### LINKED ISSUE #752: Psi4 Style Guide
There was some discussion on coming up with a coding style guide for Psi4. I don't think the idea is that we reformat all of Psi, but if you are going to reformat a section you should *try* to adhere to the following.

My personal `clang-format` settings for C++ are:
```
{
    "BasedOnStyle": "Google",
    "IndentWidth": 4,
    "ColumnLimit": 120,
}
```

Personally, I feel fairly strongly about the 4 indent width, I like the Google style in general, and the column limit is sort of an arbitrary number.

My YAPF setting for Python are:
```
{
   "config": {
        "COLUMN_LIMIT": 119,
    }
}
```
I seem to recall there is a good reason that it's 119 not 120, but I cannot remember what that is for the life of me.

Personally, not a huge fan of enforced style guides through hook mechanisms, but I can go with the flow on it. I mostly feel that being a consistent style is really nice to have, but is not required. And those hook warnings like `# noflakes...` is worse than having a style thats a little off, again, IMO.

Thoughts?

fevangelista:
## Description
Adds a .clang-format file to psi4 based on discussion of issues #752. (closes #752)

To format C++ files (with inplace edit) run:
```
 clang-format -style=file -i [<file> ...]
```

## Questions
- [x] All agree on style?
- [x] Add YAPF file.

## Status
- [x] Ready to go

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,5 @@
+Language: Cpp
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 120
+SortIncludes: false
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,4 @@
+[style]
+COLUMN_LIMIT = 119
+INDENT_WIDTH = 4
+USE_TABS = False


dgasmith:
You can add a `.style.yapf` with the following content as well:
```
[style]
COLUMN_LIMIT = 119
```

loriab:
for `.style.yapf`, maybe also
* `INDENT_WIDTH = 4`
* `USE_TABS = False`

There are others I have opinions on, but I think this is supposed to be minimal.

dgasmith:
@loriab Sure, I those those are default. Always good to add the cases were not willing to budge on however.

loriab:
Ah, thanks, @dgasmith. I thought it started from nothing, but now I see it's all of pep8 by default. All the better, imo.

fevangelista:
You got it.

fevangelista:
We should also remove these files from .gitignore.

fevangelista:
@loriab or @dgasmith : can you provide snippet of yapf usage?

dgasmith:
```
pip install yapf
yapf -i script.py
```

Done! Best way to do this is use some sort of editor like Sublime and grab PyYapf so it will only format your selection.

loriab:
also, `conda install yapf -c conda-forge`

PR #760, opened 2017-07-10
Can now run SOUKS for UKS GGA's

### LINKED ISSUE #59: cartesian PEC without nocom / no_reorient
If a user wants to make a potential energy curve or surface, it is very easy to generate nonsense when using cartesian coordinates.  Using nocom and no_reorient fixes this, so maybe the driver should be setting these options if it detects such a case.

``` python
molecule beh2 {
Be 0.0  0.0    z
H  0.0  1.0  0.0
H  0.0 -1.0  0.0
}
molecule beh2_nocom {
Be 0.0  0.0    z
H  0.0  1.0  0.0
H  0.0 -1.0  0.0
nocom
no_reorient
}
activate(beh2)
set basis sto-3g
set guess sad
for i in range (1,41):
    z = i*0.1
    beh2.z = z
    e = energy('scf')
    print z,e
    set guess read
print ''

activate(beh2_nocom)
set basis sto-3g
set guess sad
for i in range (1,41):
    z = i*0.1
    beh2_nocom.z = z
    e = energy('scf')
    print z,e
    set guess read
```

### LINKED ISSUE #759: Something weird with PES scan: SCF suddenly not converging
So here's another weird issue with PES scans in Psi4. Namely, it looks like the scan crashes in the SCF cycle every 4 or 5 points, even when I'm not feeding in the previous scf solution. If you try a new calculation in the point where it failed to converge, there's no problem.

Test input
```
molecule mol {
0 1
symmetry c1
H 0.0 0.0 0.0
H 0.0 0.0 R1
H 0.0 0.0 R2
H 0.0 0.0 R3
}   
    
set basis STO-3G
set ci_maxiter 100

# This crashes at R=0.68
#for ir in range(60, 80, 2):
# This crashes at R=0.74
for ir in range(68, 80, 2):
    mol.R1 = 1*(ir/100)
    mol.R2 = 2*(ir/100)
    mol.R3 = 3*(ir/100)
    ci_e=energy('fci')
```

### LINKED ISSUE #765: cann't get atomic positions
In [1]: import psi4

In [2]: mole = psi4.geometry('H 0.0 0.0 0.0\nH 1.5 0.0 0.0')

In [3]: atoms = mole.geometry()

In [4]: import numpy as np

In [5]: np.array(atoms)
Out[5]: array([], shape=(0, 3), dtype=float64)

dgasmith:
## Description
This simplifies some CAS machinery and fixes UKS GGA XC derivatives.

## Todos
* **Developer Interest**
  - [x] UKS XC derivates are now (likely) correct through GGA's.
  - [x] Fixes silly bug where SCF-level linear response would fail if no OEProp quantities were specified. @ashutoshvt brought this issue up.
  - [x] Freezes molecular geometry when setting a variable for cartesian coordinates. @loriab please check this out. Closes #759 and closes #59. If we like this we need fix the odd zmat/cart combined cases.
  - [x] Adds SAPT midbond functions for @konpat @loriab 
  - [x] Fixes HF3C gradients.
  - [x] Makes building molecules slightly safer, closes #765.

## Status
- [x] Ready to go


dgasmith:
@andysim I finally added your symmetrize Hessian code to your Hessian module. Looks like for some geometries the something goes slightly awry:

`test/gibbs` for the H20 dimer:

```
  No.    Vib. Freq. [cm^-1]      Vib. Temp. [K]
    1              1799.290            2588.774
    2              3812.368            5485.140
    3              3945.823            5677.152
```

```
  No.    Vib. Freq. [cm^-1]      Vib. Temp. [K]
    1                 0.002               0.003
    2              1799.290            2588.774
    3              3812.368            5485.140
```

which is actually a pretty good trick. I don't immediate see anything wrong in the symmetrize function `libmints/matrix.cc:1079`. The N2 dimer seems work however.

andysim:
Weird.  Is it only non-totally symmetric modes that are affected?

dgasmith:
Maybe, for methane (Td) we get:

```
  No.    Vib. Freq. [cm^-1]      Vib. Temp. [K]
    1                -0.000              -0.000
    2                -0.000              -0.000
    3                -0.000              -0.000
    4                -0.000              -0.000
    5                 0.000               0.000
    6              1520.326            2187.407
    7              1739.780            2503.152
    8              3186.765            4585.039
    9              3280.126            4719.364
```

andysim:
Hmmm.  Just looked at my implementation and saw this gem
```
    if(H->nirrep() > 1 || H->rowdim() != H->coldim() || H->rowdim() != 3*molecule->natom())
        throw PSIEXCEPTION("Molecule::symmetrize_hessian: Matrix cannot be symmetrized.");
```
The routine is just to clean up an already symmetric Hessian, not to apply symmetry (years since I wrote it, so I'm having to trawl through code to put things back in RAM).  It should be easy to use the CDSalc class to apply symmetry to the Hessian.  I remember chatting with @loriab about that, btu can't find the code snippet.  Will ping back when I find it.  There's no reason I can think of for the cleanup code being restricted to C1 symmetry; it should only need small tweaks to work for all point groups.

dgasmith:
This is ready for final review. Completed during SciPy Sprints woo!

psi4/src/psi4/scfgrad/scf_grad.cc
```diff
@@ -712,7 +699,7 @@ SharedMatrix SCFGrad::compute_gradient()
 
     // => XC Gradient <= //
     timer_on("Grad: XC");
-    if (potential_) {
+    if (functional_->needs_xc()) {
```
andysim: much clearer!

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -1287,7 +1297,8 @@ std::shared_ptr <Molecule> Molecule::create_molecule_from_string(const std::stri
         }
     }
 
-    mol->set_has_zmatrix(zmatrix);
+    mol->set_has_zmatrix(has_zmatrix);
+    mol->set_has_cartesian(has_cart);
```
loriab: is this planned to work ok for mixed cart/zmat?

dgasmith: Mixed will have both and we freeze the COM/Orient on detection of any cart. Its why we needed the extra cart detection.

psi4/driver/molutil.py
```diff
@@ -311,6 +311,12 @@ def geometry(geom, name="default"):
     molecule = core.Molecule.create_molecule_from_string(geom)
     molecule.set_name(name)
 
+    # Attempt to go ahead and construct the molecule
+    try:
+        molecule.update_geometry()
+    except:
+        pass
```
dgasmith: Noted.

psi4/driver/procrouting/proc.py
```diff
@@ -3393,43 +3326,20 @@ def run_sapt(name, **kwargs):
...
-        core.print_out('  SAPT does not make use of molecular symmetry, further calculations in C1 point group.\n')
-        sapt_dimer = sapt_dimer.clone()
-        sapt_dimer.reset_point_group('c1')
-        sapt_dimer.fix_orientation(True)
-        sapt_dimer.fix_com(True)
-        sapt_dimer.update_geometry()
+    sapt_basis = 'dimer'
+    if 'sapt_basis' in kwargs:
+        sapt_basis = kwargs.pop('sapt_basis')
+    sapt_basis = sapt_basis.lower()
```
dgasmith: That was actually already in, I just didn't clean it up. Thanks for the catch.

psi4/driver/procrouting/proc.py
```diff
@@ -2504,14 +2433,18 @@ def run_scf_property(name, **kwargs):
...
+    if len(oe_properties):
+        proc_util.oeprop_validator(oe_properties)
 
     if len(linear_response):
         optstash_jk = p4util.OptionsState(["SAVE_JK"])
         core.set_global_option("SAVE_JK", True)
 
     # Compute the Wavefunction
-    scf_wfn = run_scf(name, scf_do_dipole=False, do_timer=False, **kwargs)
+    if name == "scf":
```
dgasmith: Yup.

dgasmith:
@loriab @andysim Can you guys check into the EFP failure when you get a moment, not completely sure what to do about that one.

If we *only* have EFP frags:
```
  File "/home/travis/build/psi4/psi4/build/stage/home/travis/build/psi4/psi4/Install/bin/psi4", line 260, in <module>
    exec(content)
  File "<string>", line 83, in <module>
  File "/home/travis/build/psi4/psi4/build/stage/home/travis/build/psi4/psi4/Install/lib/psi4/driver/driver.py", line 427, in energy
    molecule.update_geometry()
RuntimeError: 
Fatal Error: Molecule::geometry(): molecule does not contain any atoms. Try calling `molecule.update_geometry()
     to ensure the molecule is properly constructed.
Error occurred in file: /home/travis/build/psi4/psi4/psi4/src/psi4/libmints/molecule.cc on line: 541
psi::PsiException::PsiException(std::string, char const*, int)
psi::Molecule::geometry() const
psi::Molecule::set_full_point_group(double)
psi::Molecule::update_geometry()
```

tests/opt-freeze-coords/input.dat
```diff
@@ -6,6 +6,8 @@ O   1.000000   1.000000   1.000000
 H   2.000000   1.000000   1.000000
 H   1.000000   2.000000   1.000000
 units angstrom
+no_com
+no_reorient
```
loriab: Didn't we want this test making sure that the user could offer any old molecule and then the driver would be smart enough to set no_com/reorient when it noticed the frozen coordinates?

PR #767, opened 2017-07-18
fix MKL linking for non-Intel compilers

loriab:
## Description
misc.

## Todos
* **Developer Interest**
  - [ ] MKL _can_ link with libgomp on Linux+Gnu, but it can link with libiomp5 everywhere, incl. Mac (icpc, g++, clang++) and Linux (icpc, g++). Moreover in practice, Mac+Gnu+MKL+gomp leads to n-thread-dependent energies (note that none of our testing can catch this). So switching clang and gnu to intel interface and thread libraries when built with MKL. Note that the `-fno-openmp` is b/c gnu takes that slight provocation as signal to link in libgomp, and then you get _two_ threading libraries!

## Questions
- [x] Notice to @robertodr that math detection getting shaken up again.

## Status
- [x] Ready to go --- Nope, I just want to see what CI thinks of this


robertodr:
Thanks for the heads up! Ping @bast.

bast:
Interesting - thanks for info.

cmake/math/MathLibs.cmake
```diff
@@ -170,10 +170,17 @@ else()
     set(_end_group)
 endif()
 
+# LAB Jul 2017: Because gcc thinks that if OpenMP is on, then implicitly one _must_ want their
+#   libgomp, when we'd actually prefer libiomp5 that doesn't give nthread-dependent energies.
```
andysim: this must have been a nightmare to track down

psi4/CMakeLists.txt
```diff
@@ -236,11 +237,11 @@ if((DEFINED ENV{CONDA_BUILD}) AND (UNIX AND NOT APPLE))
...
     set(cache_pybind11_DIR           "/opt/anaconda1anaconda2anaconda3/share/cmake/pybind11")
-elseif((DEFINED ENV{CONDA_BUILD}) AND (APPLE))
+elseif((DEFINED ENV{CONDA_BUILD}) AND (APPLE) AND (CMAKE_CXX_COMPILER_ID MATCHES AppleClang))
     # for stdlib compatibility, must be consistent with Darwin section of
     #   https://github.com/psi4/psi4meta/blob/master/conda-recipes/psi4-core/build.sh
-    set(cache_CMAKE_C_COMPILER       "clang")
-    set(cache_CMAKE_CXX_COMPILER     "clang++")
+    set(cache_CMAKE_C_COMPILER       "/usr/bin/clang")
+    set(cache_CMAKE_CXX_COMPILER     "/usr/bin/clang++")
     set(cache_CMAKE_CXX_FLAGS        " -stdlib=libc++")
```
andysim: I'm curious, why the full path?  Is there some system where /usr/bin isn't in the user paths by default?

cmake/math/MathLibsFunctions.cmake
```diff
@@ -203,28 +205,28 @@ macro(config_math_service _SERVICE)
...
+        #    endif()
+        #    if(HAVE_MKL_${_SERVICE})
+        #        if(APPLE)
+        #            set(${_SERVICE}_LIBRARIES ${${_SERVICE}_LIBRARIES} ${_omp_flag})
+        #        else()
+        #            set(${_SERVICE}_LIBRARIES -Wl,--start-group ${${_SERVICE}_LIBRARIES} ${_omp_flag} -Wl,--end-group)
+        #        endif()
+        #    endif()
+        #    unset(_omp_flag)
+        #endif()
```
amjames: If this code block is dead should it just be removed?

cmake/math/MathLibs.cmake
loriab: happily, attending a conference means lots of time to contemplate what on earth just happened and dream of solutions. the I-am-linking-to-pristine-intel-mkl_rt-yet-managed-to-completely-suppress-all-threading bug of the week before was much scarier (soln: any statically-linked mkl can suppress mkl_rt threading).

psi4/CMakeLists.txt
loriab: It came up in the devdevel workshop where some ppl had odd clangs installed by homebrew causing problems. we mean the system clang, so making it official.

andysim: good job intel now ships mkl in conda then!

cmake/math/MathLibsFunctions.cmake
loriab: I've no objections in another pass

psi4/CMakeLists.txt
andysim: oops, meant to comment on the conversation above.   sounds like a good way to get system clang over homebrew - was just curious about the motivation

loriab: indeed. only remaining problem is they only distribute 27 and 35

PR #769, opened 2017-07-20
Fixes basis_guess with -D combo

bwb314:
## Description
DFT-D plus the use of a guess basis set produced an error because of a typo in setting the dispersion energy: set_variables -> set_variable. 

## Status
- [X] Ready to go

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1114,7 +1114,7 @@ def scf_helper(name, **kwargs):
 
     if cast:
 
-        # A use can set "BASIS_GUESS" to True and we default to 3-21G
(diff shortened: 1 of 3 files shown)


dgasmith:
Better question: How was this not tested? Should Travis go through smoke tests.

loriab:
Aren't a whole lot of tests of guess_basis. I think "smoke" is a subset of "quick", through I would like the full set of smoke addons.

dgasmith:
Oh, I see its only if its cast and no other scenario. @bwb314 Can you add a *single* cast to `tests/dftd3/energy/input.dat`?

dgasmith:
Looks good to me, will pull in once everything passes.

PR #776, opened 2017-08-07
ENH: Adds conv thresh for supermolec complex opt

dsirianni:
## Description
Adds convergence threshold option for geometry optimization of supermolecular complexes which is slight relaxation of `g_convergence gau_tight` criteria which allows for:
- tight convergence of energy and forces, and
- relaxed convergence of `max_disp` and `rms_disp` displacement criteria. 

These criteria allow for successful optimization on flat potential surfaces, especially those associated with interfragment (supermolecular) geometry optimizations.

## Todos
Adds `interfrag_gau_tight` value to `g_convergence` OPTKING convergence threshold control keyword.  

  - Basic users can use keyword to achieve quality geometry optimizations of supermolecular complexes, without manual manipulation of individual `g_convergence` components.

## Questions
- [x] `interfrag_gau_tight` vs. `gau_tight_interfrag`? One one hand, it's a modification of `gau_tight`, but on the other hand, it has no counterpart within the Gaussian package.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/optking/set_params.cc
+++ b/psi4/src/psi4/optking/set_params.cc
@@ -322,7 +322,7 @@ void set_params(void)
       Opt_params.conv_max_disp  = 6.0e-4;  Opt_params.i_max_disp = true;
       Opt_params.conv_rms_disp  = 4.0e-4;  Opt_params.i_rms_disp = true;
     }
-if (Opt_params.general_conv == "GAU_VERYTIGHT") {
+    else if (Opt_params.general_conv == "GAU_VERYTIGHT") {
       Opt_params.i_untampered = true;
       Opt_params.conv_max_force = 2.0e-6;  Opt_params.i_max_force = true;
       Opt_params.conv_rms_force = 1.0e-6;  Opt_params.i_rms_force = true;
--- a/psi4/src/psi4/optking/set_params.cc
+++ b/psi4/src/psi4/optking/set_params.cc
@@ -314,7 +314,15 @@ void set_params(void)
       Opt_params.conv_max_disp  = 6.0e-5;  Opt_params.i_max_disp = true;
       Opt_params.conv_rms_disp  = 4.0e-5;  Opt_params.i_rms_disp = true;
     }
-    else if (Opt_params.general_conv == "GAU_VERYTIGHT") {
+    else if (Opt_params.general_conv == "INTERFRAG_GAU_TIGHT") {
+      Opt_params.i_untampered = true;
+      Opt_params.conv_max_DE    = 1.0e-6;  Opt_params.i_max_DE = true;
+      Opt_params.conv_max_force = 1.5e-5;  Opt_params.i_max_force = true;
+      Opt_params.conv_rms_force = 1.0e-5;  Opt_params.i_rms_force = true;
+      Opt_params.conv_max_disp  = 6.0e-4;  Opt_params.i_max_disp = true;
+      Opt_params.conv_rms_disp  = 4.0e-4;  Opt_params.i_rms_disp = true;
+    }
+if (Opt_params.general_conv == "GAU_VERYTIGHT") {
       Opt_params.i_untampered = true;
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/optking/set_params.cc
```diff
@@ -314,7 +314,15 @@ void set_params(void)
...
-    else if (Opt_params.general_conv == "GAU_VERYTIGHT") {
+    else if (Opt_params.general_conv == "INTERFRAG_GAU_TIGHT") {
+      Opt_params.i_untampered = true;
+      Opt_params.conv_max_DE    = 1.0e-6;  Opt_params.i_max_DE = true;
+      Opt_params.conv_max_force = 1.5e-5;  Opt_params.i_max_force = true;
+      Opt_params.conv_rms_force = 1.0e-5;  Opt_params.i_rms_force = true;
+      Opt_params.conv_max_disp  = 6.0e-4;  Opt_params.i_max_disp = true;
+      Opt_params.conv_rms_disp  = 4.0e-4;  Opt_params.i_rms_disp = true;
+    }
+if (Opt_params.general_conv == "GAU_VERYTIGHT") {
```
loriab: why `if` not `else if`?

dsirianni: Whoops, that was meant to be an `else if`.

psi4/src/psi4/optking/set_params.cc
```diff
@@ -314,7 +314,15 @@ void set_params(void)
       Opt_params.conv_max_disp  = 6.0e-5;  Opt_params.i_max_disp = true;
       Opt_params.conv_rms_disp  = 4.0e-5;  Opt_params.i_rms_disp = true;
     }
-    else if (Opt_params.general_conv == "GAU_VERYTIGHT") {
+    else if (Opt_params.general_conv == "INTERFRAG_GAU_TIGHT") {
+      Opt_params.i_untampered = true;
+      Opt_params.conv_max_DE    = 1.0e-6;  Opt_params.i_max_DE = true;
```
loriab: since you're changing two convcrit (disp) and actually adding one (dE), I might advocate just `INTERFRAG_TIGHT` or  `INTERMOL_TIGHT`.

psi4/src/psi4/optking/set_params.cc
loriab: can add to docs table (https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/optking.rst) while you're at it.

CDSherrill:
Yeah, I think I would prefer a new name like INTERMOL_TIGHT instead of
INTERFRAG_GAU_TIGHT.  One reason is that this gives us more flexibility to
change it in the future as we get more experience.... I consider the
current thresholds only draft values for now.

psi4/src/read_options.cc
```diff
@@ -2419,7 +2419,7 @@ int read_options(const std::string &name, Options & options, bool suppress_print
       /*- Set of optimization criteria. Specification of any MAX_*_G_CONVERGENCE
       or RMS_*_G_CONVERGENCE options will append to overwrite the criteria set here
       unless |optking__flexible_g_convergence| is also on.      See Table :ref:`Geometry Convergence <table:optkingconv>` for details. -*/
-      options.add_str("G_CONVERGENCE", "QCHEM", "QCHEM MOLPRO GAU GAU_LOOSE GAU_TIGHT GAU_VERYTIGHT TURBOMOLE CFOUR NWCHEM_LOOSE");
+      options.add_str("G_CONVERGENCE", "QCHEM", "QCHEM MOLPRO GAU GAU_LOOSE GAU_TIGHT INTERFRAG_GAU_TIGHT GAU_VERYTIGHT TURBOMOLE CFOUR NWCHEM_LOOSE");
```
loriab: going to have to change kw here, too.

dgasmith: @dsirianni Can you patch this so we can get it in?

PR #780, opened 2017-08-15
Char star replace for irrep_labels()

tsexton:
## Description
Replace char** irrep_labels() with std::vector<std::string> irrep_labels()
Convert with .c_str() when needed in print statements.
Remove calls to free()
Removed calls to delete in dcft_scf_UHF.cc

## Todos
* **Developer Interest**
  - [ ] Part of the change to std::string
  - [ ] None

## Status
- [ ] Test jobs cc1, dfmp2-1, scf-occ  all passed, and Irrep labels printed were the same as the reference.

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1055,17 +1055,6 @@ void export_mints(py::module& m)
              "Print the bond angle geometrical parameters")
         .def("print_out_of_planes", &Molecule::print_out_of_planes,
              "Print the out-of-plane angle geometrical parameters to output file")
-        /*.def("irrep_labels",
-             [](Molecule& mol) {
-                 std::vector<std::string> ret;
-                 std::vector<std::string> labels = mol.irrep_labels();
-                 int nirrep = mol.point_group()->char_table().nirrep();
-                 for (size_t h = 0; h < nirrep; h++) {
-                     std::string lh(labels[h]);
-                     ret.push_back(lh);
-                 }
-                 return ret;
-             })*/
         .def_property("units", py::cpp_function(&Molecule::units),
                       py::cpp_function(&Molecule::set_units),
                       "Units (Angstrom or Bohr) used to define the geometry")
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1055,7 +1055,7 @@ void export_mints(py::module& m)
              "Print the bond angle geometrical parameters")
         .def("print_out_of_planes", &Molecule::print_out_of_planes,
              "Print the out-of-plane angle geometrical parameters to output file")
-        .def("irrep_labels",
+        /*.def("irrep_labels",
              [](Molecule& mol) {
                  std::vector<std::string> ret;
                  std::vector<std::string> labels = mol.irrep_labels();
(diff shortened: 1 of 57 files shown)


psi4/src/export_mints.cc
```diff
@@ -1058,7 +1058,7 @@ void export_mints(py::module& m)
         .def("irrep_labels",
              [](Molecule& mol) {
                  std::vector<std::string> ret;
-                 char** labels = mol.irrep_labels();
+                 std::vector<std::string> labels = mol.irrep_labels();
                  int nirrep = mol.point_group()->char_table().nirrep();
```
amjames: Actually  since `mol.irrep_labels()` returns `std::vector<std::string>` then the whole lambda can go now. The function should be bound directly without the need for the wrapper.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -1264,9 +1264,9 @@ void HF::print_orbitals()
                 orb_order[orb_e[a].second] = a;
 
             for (int a = 0; a < nalphapi_[h]; a++)
-                occ.push_back(std::make_pair(epsilon_a_->get(h,a), std::make_pair(labels[h],orb_order[a] + 1)));
+                occ.push_back(std::make_pair(epsilon_a_->get(h,a), std::make_pair(labels[h].c_str(),orb_order[a] + 1)));
```
amjames: I don't see any reason to keep using `std::pair<const char*, int>` can probably just  switch to std::pair<const std::string, int>.

There would be no need to use `labels[h].c_str()` all over the place in `HF::print_orbitals()`. you would only need to using it once in the overload above (after changing the argument types too)

psi4/src/psi4/cctriples/get_moinfo.cc
```diff
@@ -243,9 +243,9 @@ void cleanup(void)
 //    free(moinfo.uoccpi);
 //    free(moinfo.fruocc);
 //    free(moinfo.frdocc);
-    for(i=0; i < moinfo.nirreps; i++)
-        free(moinfo.labels[i]);
-    free(moinfo.labels);
+    //for(i=0; i < moinfo.nirreps; i++)
+    //    free(moinfo.labels[i]);
+    //free(moinfo.labels);
```
dgasmith: Please go ahead and remove this when you get a chance.

psi4/src/psi4/dcft/dcft_scf_UHF.cc
```diff
@@ -297,9 +297,9 @@ namespace psi{ namespace dcft{
           Ca_->print();
           Cb_->print();
       }
-      for (int h = 0; h < nirrep_; ++h)
-          delete [] irrepLabels[h];
-      delete[] irrepLabels;
+      //for (int h = 0; h < nirrep_; ++h)
+      //    delete [] irrepLabels[h];
+      //delete[] irrepLabels;
```
dgasmith: Same as above and in the remaining spots.

loriab:
I think you have to add `.def("irrep_labels", &Molecule::irrep_labels, "docstring")` back to `export_mints`. But now it can be an ordinary export, not the complicated lambda. Same for any other export lambda you deleted as unnecessary. That should fix the tests. After that, lgtm!

tsexton:
Looks like this may have caused some trouble in eom-ccsd (#786)  that the tests didn't catch.

PR #790, opened 2017-08-25
ECP housekeeping

loriab:
## Description
ECP basissets, counter, AM lookup

## Todos
* **Developer Interest**
  - [x] Fixed some `****` around ECP that weren't situated so as to satisfy our picky parser
  - [x] Fixed unititialized ECP printing counter variable noticed by @fevangelista 
  - [x] Extended dict looking up ECP AM noticed by @fevangelista 
  - [x] Add `ao_efp_multipole_potential` and `electric_field` integrals to mintshelper

## Questions
- [ ] Should we suppress [this printing](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/ecpint.cc#L500)? or redirect to outfile?
- [ ] Should consider origin question with mintshelper. Some ints use the PROPERTIES option to set origin. Just added ints take origin as an arg.

## Status
- [x] Ready to go


psi4/driver/qcdb/libmintsbasissetparser.py
```diff
@@ -195,7 +195,11 @@ def parse(self, symbol, dataset):
                     # or:    H_ECP    O_ECP ...     0
                     if atom_ecp.match(line):
                         ecp_msg = """line %5d""" % (lineno)
-                        symbol_to_am = { 0:0, 1:1, 2:2, 3:3, "s":0, "S":0, "p":1, "P":1, "d":2, "D":2, "f":3, "F":3 }
+                        symbol_to_am = {0: 0,   's': 0,   'S': 0,
+                                        1: 1,   'p': 1,   'P': 1,
+                                        2: 2,   'd': 2,   'D': 2,
+                                        3: 3,   'f': 3,   'F': 3,
+                                        4: 4,   'g': 4,   'G': 4}
                         # This is an ECP spec like "KR-ECP    3     28"
```
andysim: Are we confident that G is high enough?  I had thought that F was sufficient, based on that periodic table thingy, but apparently I misunderstood the exact meaning of the projectors.

loriab: I searched the existing ecp basis sets, and `g` was the highest I found. I don't mind adding higher.

andysim: I figured you would have checked.  I'm fine with up to g in that case.  Want me to merge this now?

loriab:
NP, you did the real work. I'm glad to shift a few lines.

This does only get ecp's running more broadly. I think some ecp convergence problems are another PR.

psi4/driver/qcdb/libmintsbasissetparser.py
susilehtola: AFAIK/IIRC to build an ECP for an an atom with maximum angular momentum L, projectors up to L+1 are necessary. So, for the lanthanides and actinides you have F electrons, meaning G projectors. Should be fine.

andysim: Yup, you're right.  I'd missed / forgotten about the `+1` in, *e.g.* eqn. 1 of [this paper](http://dx.doi.org/10.1063/1.4922844).  I was wondering why G electrons were being accounted for when the G projectors were

Issue #789, opened 2017-08-24
OPTKING PSIO error: optimization data write failure


dsirianni:
Hi all,

The following PSIO error occurs for identical geometry optimizations on two separate machines, when OPTKING attempts to write opt data to binary file:
```
Fatal Error: PSIO Error
Error occurred in file: /theoryfs2/ds/sirianni/Gits/psi4/psi4/src/psi4/libpsio/error.cc on line: 129
The most recent 5 function calls were:

psi::PsiException::PsiException(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, char const*, int)
psi::psio_error(unsigned long, unsigned long)
psi::PSIO::write(unsigned long, char const*, char*, unsigned long, psi::psio_address, psi::psio_address*)
psi::PSIO::write_entry(unsigned long, char const*, char*, unsigned long)
psi::psio_write_entry(unsigned long, char const*, char*, unsigned long)
```
This then results in a second PSIO error upon each subsequent optimization iteration, when attempting to read the binary optimization data from file:
```
Fatal Error: PSIO Error
Error occurred in file: /theoryfs2/ds/sirianni/Gits/psi4/psi4/src/psi4/libpsio/error.cc on line: 129
The most recent 5 function calls were:

psi::psio_error(unsigned long, unsigned long)
psi::PSIO::open(unsigned long, int)
psi::psio_open(unsigned long, int)
opt::opt_io_is_present()
```
This finally results in the optimizer spinning its wheels for the remainder of the allotted iterations. The usual PSIO error-culprit, an un-touchable filesystem, was ruled out for both machines on which this behavior was observed.

Input:
```
# index A24-7-dimer label Dimer from ammmonia_methane_C3v

memory 14000 mb

molecule mol {
       units Angstrom
       no_com
       no_reorient
       0 1
    N            0.001106462041    -0.000130846927     1.849133409958
    H            0.938627974105    -0.000139345380     2.221618285244
    H           -0.467316580071    -0.811849948182     2.222477040520
    H           -0.467307391010     0.811579719379     2.222474714333
    H           -0.000457075852    -0.000011938197    -0.942103507206
    C           -0.001019750816     0.000109523659    -2.027776337389
    H            0.509840850903     0.885353515687    -2.394362733084
    H            0.509834400679    -0.885039296153    -2.394623104348
    H           -1.023308879978     0.000128616114    -2.393761958029

}

set e_convergence 11
set basis cc-pvtz
[...]

PR #798, opened 2017-09-08
Optbug

psi-rking:
## Description
Two bug fixes.  First is wrong projected energy change in P-RFO transition state optimization.
Second, is an i/o error that occured because the file1 was not always deleted when the "dynamic level" changed.  Default behavior should be unaffected.
## Todos
* **Developer Interest**
Should resolve issue #789 
 
## Questions
None
## Status
Should be ready to go.

### DIFF
--- a/psi4/src/psi4/optking/io.h
+++ b/psi4/src/psi4/optking/io.h
@@ -43,7 +43,7 @@ namespace opt {
 enum OPT_IO_FILE_STATUS {OPT_IO_OPEN_NEW, OPT_IO_OPEN_OLD} ;
 
 bool opt_io_is_present(void);
-void opt_io_remove(void);
+void opt_io_remove(bool force=false);
 void opt_io_open(OPT_IO_FILE_STATUS status);
 void opt_io_close(int keep);
 void opt_io_read_entry(const char *key, char *buffer, size_t size);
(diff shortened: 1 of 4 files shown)


loriab:
macpsinet seems out of sorts even though I rebooted it earlier this week. Will attend to it tomorrow.

dsirianni:
Just observed a separate issue, commenting here to raise awareness before merge at the advice of @loriab.  

If the OPTKING bad-step exception occurs in the highest dynamic level (7), e.g.,
```
    Current energy   :      -154.6485078443

    Energy change for the previous step:
        Projected    :        -0.0449999979
        Actual       :         0.0000000000
    The BAD_STEP_EXCEPTion handler:
    Energy has increased in a minimization.

    Dynamic level is 7.
    Consecutive backsteps is 1.
    The INTCO_EXCEPTion handler:
    Too many bad steps.
    Dynamic level is 7.
    exc.g_really_quit() is 0.

  **** Optimization has failed! (in 9 steps) ****
            --------------------------
             OPTKING Finished Execution
            --------------------------
```
the error isn't translated to a Py-side exception, and the program prints the "Buy a developer a beer!" successful exit message even though the optimization has, in fact, failed.  This should raise some error, akin to the new `ConvergenceError` exception, so that the output does not fly under the radar as successful.

andysim:
Great catch by @dsirianni.  I'll pull this one in (with squash) and the exception error can be fixed separately.

PR #799, opened 2017-09-18
Dft logic

dgasmith:
## Description
The DFT_CUSTOM_FUNCTIONAL keyword died at some point so I started to examine how to re-enable it (@loriab) and I was also bugged about HF-3C gradients (@bwb314). Luckily it turns out I had already done most of the work at some point (I have 3 more of these branches apparently that I need to merge). Basically this works finishes the movement of trying to distinguish between DFT and SCF where SCF is now a very basic DFT "functional". This could lead towards "functionals" being just SCF + stuff where the stuff could be XC, PCM, -D, external potentials, etc. 

I recall @adabbott is willing to document this.

## Todos
Notable points that this PR has either accomplished or will accomplish.
* **Developer Interest**
  - [x] SCF is now basically a very vanilla case of DFT
  - [x] Adds finite difference Hessians from gradients for -D corrections
* **User-Facing for Release Notes**
  - [x] Enables HF-3C gradients and Hessians.
  - [x] Tentatively re-enables ip_fitting and other frac procedures
  - [x] Remove keyword `DFT_CUSTOM_FUNCTIONAL` in favor of `energy("SCF", dft_functional=...)`

## Status
- [x] Ready to go

PR #805, opened 2017-09-25
New DFT Logic Documentation

adabbott:
## Description
Documentation update for the new dft logic

## Todos
- [x] Ensure currently existing docs are updated with the new dft logic
- [x] Include new examples on functional customization

Waiting on #799 to be merged to test new examples for the docs.

## Status
- [x] Ready to go


doc/sphinxman/source/dft.rst
```diff
@@ -717,3 +717,100 @@ The "best-practice" input file for KS-DFT is::
     }
 
     energy('b3lyp')
+
+
+Advanced Functional Use and Manipulation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Using the ``dft_functional`` option in the energy procedure call, one is able to use the orbitals generated by DFT for correlated wavefunction methods::
```
loriab: I'd say "keyword argument" rather than "option" so as to reserve the latter for things in read_options.cc

loriab: Also please add a warning that if you tweak calculations (esp. MP2+PBE0) that only generic psivars will be defined so as to avoid contamination.

Issue #782, opened 2017-08-18
orientation, cartesian, and props2


loriab:
https://github.com/psi4/psi4/pull/760/files#diff-aac3b6c47032ba056103b4f355643ca7R3113 makes a lot of sense and eliminates a pit that's easy to fall into. But need to revisit its impact on properties of charged molecules.

Basically, props2 is currently broken, and need to figure out if the recent PR was too aggressive in `no_reorient`ing or if `props2` has been falling into the aforementioned pit all along.

loriab:
w/o reinvestigating, this is my memory of what's going on:

props2 is a Cartesian-with-variable geometry (y-coord gets updated in a loop) with _no explicit_ freeze com/orient. System is a cation so origin of *poles matters. Test ref values are computed at COM. For a long time this was fine b/c in the absence of user forbidding it, molecule was moved to COM & std orientation and computed==expected.

But there's a deep pit that both users and developers can fall into wrt scanning over coordinates of a Cartesian system -- values you're updating aren't in same coord sys as values being specified. Results can be jarring. So DGAS fixed with https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/molecule.cc#L3110-L3117 . This fills the pitfall, but now the props2 case doesn't get moved to COM, so computed != expected. That is, until psi4/psi4#808 where we explictly instruct moving to COM so that test passes again.

Naturally, a Z-Mat molecule would always go to COM and wouldn't have this distinction.

This is all perfectly logical. Trouble is consistency btwn Cart/ZMat and what users will be least confused by.

dgasmith:
Im still not completely sure that there is a way to bridge the two. Ultimately the modification in-place leads to quite a few errors; however, the Psithon API make the in-place a logical operation.

It would be better to do something like:
```
new_mol = molecule.update({"var1": 5, "var2": 6})
```
or something similar. Getting that across in something like Psithon would be difficult I think.

CDSherrill:
Ok thanks for the summary.  I agree the current behavior is logical, and
also that there is an inconsistency between Zmat and Cartesians that could
confuse a user.  I am very happy there is a little warning printed by the
code if the user is updating the Cartesian coordinates.  Perhaps that
warning could be updated to say something like "(multipole moments of
charged molecules may be affected)" or something?  And/or a sentence or two
in whatever section of the manual might cover something like this?

I also recommend expanding the comment in the props2/input.dat from "#
[...]

PR #808, opened 2017-09-26
FIX: Tightens up a few tests

dgasmith:
Based off the forums and a few issues there are a few tests that are not quite robust enough to test the code rather than a operational tweaks.
 - `props2` - The new orientation code requires that this be moved to the COM at every step.
 - `props3` - Overall convergence needs to be tightened to move flutter in 64-pole below test threshold.
 - `scf-cov` - Convergence needs to be tightened in addition to not using a C2v molecule during a orientation check.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -71,8 +71,7 @@ foreach(test_name adc1 adc2 casscf-fzc-sp casscf-semi casscf-sa-sp ao-casscf-sp 
                   omp3-3 omp3-4 omp3-5 omp3-grad1 omp3-grad2 opt-lindep-change 
                   opt1 opt1-fd opt2 opt2-fd opt3 opt4 opt5 opt6 opt7 opt8 opt9 
                   opt11 opt12 opt13 opt14 opt-irc-1 opt-irc-2 opt-freeze-coords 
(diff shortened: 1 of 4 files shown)


loriab:
I presume this actually fixes props2 (in which case should also add back to tests/CM). My concern is the change that people won't automatically get their multipoles at the origin w/o explicitly asking and that behaviour is different btwen Cart/ZMat (psi4/psi4#782). Anyone else concerned?

The scf-coverage problems we've been seeing have been compilation dependent after DF_Helper merge. Do you consider the changes to this test a fix to slightly flawed test?

CDSherrill:
I pulled up issue 782 to remind my self what is going on, but the comments
there are somewhat terse and they don't do enough to jog my memory even
though I had a conversation with Lori about it.  I'm not sure the wider
community will be able to comment much on this point without doing a little
homework to figure out what is being asked about.  Could we have a bit more
commentary describing the issue?

dgasmith:
I view `props3` and `scf-cov` fixing up a slightly shaky/flawed test.

For `props2` this is fixing an issue caused by the Cart/ZMat changes. Let me comment there further to keep that discussion contiguous.

loriab:
[props2 commentary](https://github.com/psi4/psi4/issues/782#issuecomment-332296506), as I don't think you're on that thread, @CDSherrill 

props3, scf-coverage LGTM

PR #809, opened 2017-09-26
Set up integration with Danger.Systems

robertodr:
The integration runs some checks on PRs, notably that C++ code style conforms to `.clang-format`

## Description
[Danger.Systems](http://danger.systems/ruby/) provides a DSL to perform checks on PRs. It's a Ruby gem that runs on Travis, reads the `Dangerfile` and comments on a PR through a bot.
The current version of the `Dangerfile`:
1. **Fails PR** if C++ code style does not comply with `.clang-format`
1. **Fails PR** if Python code style does not comply with `.style.yapf`
2. **Warns** if PR is classed as work in progress, _i.e._ contains `[WIP]` in the title
3. **Warns** if PR is classed as trivial, _i.e._ contains `#trivial` in the title
4. **Fails PR** if there are merge commits in the history
5. **Warns** if PR introduced code changes (in subdirectories `psi4`, `external`, `cmake`) without adding tests and/or docs.

## Todos
* **Developer Interest**
  - [x] Integrate with Danger.Systems
  - [x] Danger checks that C++ code style complies with `.clang-format`
  - [x] Clean up `.travis.yml` from now useless PPA
  - [x] Ignore `build*` directories
  - [x] Danger checks that Python code style complies with `.style.yapf`
  - [x] Enabled caching of some dependencies on Travis

## Questions
1. I think it's better to pull this in once `clang-format` has been run over the whole code base.
2. Where is a good place to document the behavior of the bot?
3. Are the commented sections in `.travis.yml` supposed to be there for future use?

## Status
- [x] Ready to go


.travis.yml
```diff
@@ -1,5 +1,7 @@
+dist: trusty
```
loriab: I heard a rumor from the summer school that psi4 and trusty don't play well together, hence pinning precise. I guess we'll see from this PR.

Commented-out sections in .travis file are just old py/compiler combinations. I think we're pretty satisfied with the time/coverage balance of 3 tests, so the others could be removed. Just leave [this commented section](https://github.com/psi4/psi4/blob/master/.travis.yml#L4-L13) as it took me a lot of searching to figure out allowed fails.

Gemfile.lock
```diff
@@ -0,0 +1,67 @@
+GIT
```
loriab: Don't suppose there's any way the Danger and Gem files can be `.Something`'s so ppl can see they're not vital to runtime operations?

robertodr: Good point! I moved `Dangerfile`, `Gemfile` and `Gemfile.lock` to `.github`, since they are related to repo maintenance :smiley:

.travis.yml
robertodr: It seems to be working on trusty.

Gemfile.lock
loriab: oh, good, I'm glad it's flexible. `.scripts/` is another good place for them.

robertodr: I prefer under `.github` since they are not actual scritps. What do you think about the random LGTM picture? Should I keep it in?

loriab: If you like, could rename `.scripts` to `.ci` since it just contains travis run/test scripts. But not vital.

I'd prefer dropping the random LGTM picture. Too big and obscures the actual result message.

robertodr: :+1:

dgasmith:
Do you have an example of what this would look like in practice?

robertodr:
Something [like this](https://github.com/PCMSolver/danger-cpp-example/pull/3)

robertodr:
By the way, I've managed to modify the checker to also run the code style validation on Python using YAPF. [Here is an example](https://github.com/PCMSolver/danger-python-example/pull/1) I'll integrate that too as soon as the modifications to the checker make it upstream.

loriab:
On the topic of Q2, I think the best place to document would be an FAQ answer to "How and what C++ style is enforced?" https://github.com/loriab/psi4/blob/wiki2docs/doc/sphinxman/source/style_c.rst would be just the thing. Maybe a py one, too. Then just link them in https://github.com/loriab/psi4/blob/wiki2docs/doc/sphinxman/source/prog_faq.rst .

doc/sphinxman/source/code_style.rst
```diff
@@ -0,0 +1,100 @@
...
+repository are formatted according to these style files using ``clang-format``
+and ``yapf``. It is then helpful if these tools be part of your development toolchain.
+Once ``clang-format`` and ``yapf`` are installed, then there are three ways in which
+formatting of the code can be accomplished, in decreasing order of automation:
+
+1. By integrating the formatters into your editor.
+2. By installing Git hooks to run the formatters when committing.
+3. By running the formatters manually on the modified files.
+
+.. _`faq:editor`:
```
loriab: Also go ahead and add the tags `faq:editor`, etc. to the prog-faq.rst file, I think.

PR #814, opened 2017-09-29
Pybind11 2.2.1

loriab:
## Description
Bump some dependencies. Update psi to pybind11 2.2.1. Deal accompanying visibility issues through Jet's `PSI_API` macro.

## Todos
* **Developer Interest**
  - [x] CheMPS2 to 1.8.4
  - [x] ~Libint to latest v1 head~ (RDR already covered this)
  - [x] Pybind11 to 2.2.1
  - [x] had to add some psio unsavoriness (recomputing integrals whether present or not) since pb11-221 won't seem to give me my `None` for `nullptr`. This only affects stability analyses, and I figure they're due for an overhaul anyways.
  - [x] add enough `PSI_API` to get shipped plugins to work.

## Questions
- [ ] Big question here is do we want to make the pybind jump?

  * Not immediately until https://github.com/pybind/pybind11/issues/1121 is cleared up
  * Likely to cause some building headaches and entanglements of 200 and 221 versions at build-time, no matter how tightly I try to instruct CMake where to find it.
  * Pybind folks are [pretty adamant about symbol visibility](http://pybind11.readthedocs.io/en/stable/upgrade.html#stricter-enforcement-of-hidden-symbol-visibility-for-pybind11-modules). We do something they truly disapprove of and use pb as a plain dependency rather than their (very nice) module maker, so I think our plugins (double use of `core.so` as python module and ordinary Linux lib) should be fine, but we'll want to test Forte, @fevangelista.
  * Can take a look at the [Changelog](http://pybind11.readthedocs.io/en/stable/changelog.html) for new goodies. Note that Psi4 is currently pinned at v2.0.0 (1 Jan 2017).
  * When @ryanmrichard was dealing with two pybind projects (Psi4, Pulsar), he came to the conclusion that it was very important to have all pybind projects on exactly the same version. Right now we control all ecosystem projects using pybind (Psi4, Forte, PylibEFP (which needs >=2.2.0, incidentally)) so can just made decisions in consensus. That may not always be the case. I believe some of pybind's asserted control over symbols is actually meant to allow better mixing of pb projects, in which case we should also consider doing as they advise and only exporting the symbols we need.

## Status
- [x] Ready to go. This is now pointing to scfitertopy branch, so ok to merge.

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -366,19 +366,22 @@ def scf_finalize_energy(self):
         # compute it if needed
         if core.get_option('SCF', 'REFERENCE') != "UHF":
             psio = core.IO.shared_object()
-            psio.open(constants.PSIF_SO_TEI, 1)  # PSIO_OPEN_OLD
-            try:
-                psio.tocscan(constants.PSIF_SO_TEI, "IWL Buffers")
-            except TypeError:
-                # "IWL Buffers" actually found but psio_tocentry can't be returned to Py
-                psio.close(constants.PSIF_SO_TEI, 1)
-            else:
-                # tocscan returned None
-                psio.close(constants.PSIF_SO_TEI, 1)
-                core.print_out("    SO Integrals not on disk, computing...")
-                mints = core.MintsHelper(self.basisset())
-                mints.integrals()
-                core.print_out("done.\n")
+            #psio.open(constants.PSIF_SO_TEI, 1)  # PSIO_OPEN_OLD
+            #try:
+            #    psio.tocscan(constants.PSIF_SO_TEI, "IWL Buffers")
+            #except TypeError:
+            #    # "IWL Buffers" actually found but psio_tocentry can't be returned to Py
+            #    psio.close(constants.PSIF_SO_TEI, 1)
+            #else:
+            #    # tocscan returned None
+            #    psio.close(constants.PSIF_SO_TEI, 1)
+
+            # logic above foiled by psio_tocentry not returning None<--nullptr in pb11 2.2.1
+            #   so forcibly recomputing for now until stability revamp
+            core.print_out("    SO Integrals not on disk, computing...")
+            mints = core.MintsHelper(self.basisset())
+            mints.integrals()
+            core.print_out("done.\n")
 
             # Q: Not worth exporting all the layers of psio, right?
 
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -378,7 +378,7 @@ def scf_finalize_energy(self):
 
             # logic above foiled by psio_tocentry not returning None<--nullptr in pb11 2.2.1
             #   so forcibly recomputing for now until stability revamp
-            core.print_out("    SO Integrals not on disk, computing...")
+            core.print_out("    SO Integrals not on disk. Computing...")
             mints = core.MintsHelper(self.basisset())
             mints.integrals()
             core.print_out("done.\n")
--- a/psi4/src/psi4/libqt/qt.h
+++ b/psi4/src/psi4/libqt/qt.h
@@ -140,7 +140,7 @@ void C_DTRMV(char uplo, char trans, char diag, int n, double* a, int lda, double
 void C_DTRSM(char side, char uplo, char transa, char diag, int m, int n, double alpha, double* a, int lda, double* b, int ldb);
 
 // BLAS 3 Double routines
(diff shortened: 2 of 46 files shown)


loriab:
Update: `-fvisibility=hidden` is indeed active in 2.2.1 and quite breaking plugins. Shelving this PR and recommending that symbols/c-layer-/py-layer-plugin-interfacing be a topic at WWDC.

dgasmith:
I skimmed the functionality for new goodies, I dont see too many that effects us (and nothing in a major way). So nothing too promising there.

Do the PyBind11 folks have an official stance on cross version compatibility. I dont see a huge reason for major releases not to work together at the moment. Although I do see possible issues as they tweak things like `py::keep_alive` between 2.1 and 2.2.

loriab:
The pybind stance on cross-version compatibility is slightly mentioned in the "pretty adamant" link above. Basically, they are alert to the problem of multiple py-bound modules and made changes so they _can_ work together at runtime, provided each is modest with its symbols, as is enforced with newer versions of pb11. Unfortunately, Psi relies on free symbol visibility.

Note that this _is_ likely to hit us as pylibefp uses a feature in 2.2. We could backtrack on that, but I don't suppose we want to put much technical effort into remaining pinned to a single version forever.

dgasmith:
@loriab Looks like Pybind11 #1121 is clearing up. Do you have thoughts on getting this or should I tweak up gau2grid to avoid 2.2 features?

loriab:
@dgasmith, I really prefer to upgrade to pb11 2.2. pylibefp is using a nice 2.2 feature (though I do know how to downgrade). But we do have to agree and solve symbols visibility or face breaking plugins (slide from wwdc attached). Since gau2grid doesn't need the python interface in psi (or do you have plans for it?), I think keep the 2.2 features.

screen shot 2017-11-26 at 10 20 24 am

robertodr:
Notice that CMake can generate a header with the export macros and can set the appropriate compiler flags accordingly: https://cmake.org/cmake/help/v3.3/module/GenerateExportHeader.html

psi4/src/psi4/libqt/qt.h
```diff
@@ -140,7 +140,7 @@ void C_DTRMV(char uplo, char trans, char diag, int n, double* a, int lda, double
 void C_DTRSM(char side, char uplo, char transa, char diag, int m, int n, double alpha, double* a, int lda, double* b, int ldb);
 
 // BLAS 3 Double routines
-void C_DGEMM(char transa, char transb, int m, int n, int k, double alpha, double* a, int lda, double* b, int ldb, double beta, double* c, int ldc);
+PSI_API void C_DGEMM(char transa, char transb, int m, int n, int k, double alpha, double* a, int lda, double* b, int ldb, double beta, double* c, int ldc);
```
andysim: is there a reason why only `DGEMM` is being exported?

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -366,19 +366,22 @@ def scf_finalize_energy(self):
...
+            #except TypeError:
+            #    # "IWL Buffers" actually found but psio_tocentry can't be returned to Py
+            #    psio.close(constants.PSIF_SO_TEI, 1)
+            #else:
+            #    # tocscan returned None
+            #    psio.close(constants.PSIF_SO_TEI, 1)
+
+            # logic above foiled by psio_tocentry not returning None<--nullptr in pb11 2.2.1
+            #   so forcibly recomputing for now until stability revamp
+            core.print_out("    SO Integrals not on disk, computing...")
```
andysim: This is a good opportunity to remove one of the comma splices we have in our code methinks.

loriab: In the world of quantum chemistry, someone's probably banking on that comma splice. :-) Will do.

psi4/src/psi4/libqt/qt.h
loriab: The current exports are exactly enough to get the packaged plugin templates to compile. I figure we need an en banc judgement on what exactly is the public c-side psi api.

psi4/driver/procrouting/scf_proc/scf_iterator.py
andysim: That's a really good point.  I guess we should be careful to avoid too many changes in output.  The most notable comma splice we had, which was probably used in greps, was the old "buy a developer a beer" banner.  Thankfully @lothian fixed that a long time ago, before it drove me nuts and affected users' grepping.

loriab: I actually considered it more of a joke than a good point. My attitude has been that we shouldn't be tentative on changing text output — that we strive for consistent psivar access, not text access. I've already changed up a lot of `Bohr` --> `[a0]` in related branches on those grounds.

PR #822, opened 2017-10-13
cubeprop unit labels

carlosborca:
## Todos
* **Developer Interest**
  - [x] Added unit labels to density and ESP.

## Questions
- [ ] @fevangelista, how do the units look to you?
- [ ] Should units be applied to https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libcubeprop/cubeprop.cc#L192-L196

## Status
- [ ] Ready to go

### DIFF
--- a/psi4/src/psi4/libcubeprop/cubeprop.cc
+++ b/psi4/src/psi4/libcubeprop/cubeprop.cc
@@ -131,10 +131,10 @@ void CubeProperties::compute_properties()
             Ds->copy(Da_);
             Dt->add(Db_);
             Ds->subtract(Db_);
-            compute_density(Dt, "Dt");
-            compute_density(Ds, "Ds");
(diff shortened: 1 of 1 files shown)


dgasmith:
Thanks for the PR! Definitely add units where applicable. 

Looks like you will need to tweak that test after your changes. Everything else looks great.

carlosborca:
What I did before failed because I was putting the units into a string that was also used to generate file names. As the string contained a slash, and other illegal characters the execution ran into an error.

Should I add a general comment (third argument) by default the units to compute_density and related functions or hard-wire the units into the cube file header?

https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libcubeprop/cubeprop.h#L98-L108

dgasmith:
I think adding a third argument makes the most sense.

PR #826, opened 2017-10-24
Rid P::e of envvar

loriab:
## Description
I hit a system that was raising [putenv and setenv not avail](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libpsi4util/process.cc#L136). Rather than solve it, decided (with support) to clean environment variables out of `P::e`. Only active uses were PSI_SCRATCH and PSIDATADIR.

## Todos
* **Developer Interest**
  - [x] Replaced all the PSI_SCRATCH with calls directly to psio.get_default_path(). Also cleared that fn out so it truly defaults to `/tmp/`, not TMP, TEMP, TMPDIR, then /tmp/.
  - [x] Once all the parsing's py-side might be able to drop PSIDATADIR entirely. But for now, moved it to its own slot as `P::e.[gs]et_datadir()`
  - [x] Can't tell if you're dealing with Clang or AppleClang by `#defines`, so lowered cxxstandard cutoff to permit Intel + !AppleClang to pass.
  - [x] `FindOpenMP.cmake` attaches a lib to try_run that isn't needed just to extract versions. Shift OMP detection so cxxstandard doesn't complain about unfindable lib

## Status
- [x] Ready to go

### DIFF
--- a/cmake/custom_cxxstandard.cmake
+++ b/cmake/custom_cxxstandard.cmake
@@ -31,7 +31,7 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
     file(REMOVE ${_testfl})
 
     if (APPLE)
-        if (${GCC_VERSION} VERSION_LESS 6.1)
+        if (${GCC_VERSION} VERSION_LESS 3.6)
             message(FATAL_ERROR "${BoldYellow}Intel ICPC makes use of CLANG (detected: ${GCC_VERSION}; required for C++11: 6.1) so this build won't work without CLANG intervention: http://psicode.org/psi4manual/master/build_planning.html\n${ColourReset}")
         endif()
     else ()
--- a/cmake/custom_cxxstandard.cmake
+++ b/cmake/custom_cxxstandard.cmake
@@ -26,17 +26,17 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
     try_run(GCCV_COMPILES
             GCCV_RUNS
             ${CMAKE_BINARY_DIR} ${_testfl}
-            RUN_OUTPUT_VARIABLE GCC_VERSION)
-    message(STATUS "Found base compiler version ${GCC_VERSION}")
+            RUN_OUTPUT_VARIABLE CPLR_VERSION)
+    message(STATUS "Found base compiler version ${CPLR_VERSION}")
     file(REMOVE ${_testfl})
 
     if (APPLE)
-        if (${GCC_VERSION} VERSION_LESS 3.6)
-            message(FATAL_ERROR "${BoldYellow}Intel ICPC makes use of CLANG (detected: ${GCC_VERSION}; required for C++11: 6.1) so this build won't work without CLANG intervention: http://psicode.org/psi4manual/master/build_planning.html\n${ColourReset}")
+        if (${CPLR_VERSION} VERSION_LESS 3.6)
+            message(FATAL_ERROR "${BoldYellow}Intel ICPC makes use of CLANG (detected: ${CPLR_VERSION}; required for C++11: Clang 3.6 or AppleClang 6.1) so this build won't work without CLANG intervention.\n${ColourReset}")
         endif()
     else ()
-        if (${GCC_VERSION} VERSION_LESS 4.9)
-            message(FATAL_ERROR "${BoldYellow}Intel ICPC makes use of GCC (detected: ${GCC_VERSION}; required for C++11: 4.9) so this build won't work without GCC intervention: http://psicode.org/psi4manual/master/build_planning.html#faq-modgcc\n${ColourReset}")
+        if (${CPLR_VERSION} VERSION_LESS 4.9)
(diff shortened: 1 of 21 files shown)


cmake/custom_cxxstandard.cmake
```diff
@@ -31,7 +31,7 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
     file(REMOVE ${_testfl})
 
     if (APPLE)
-        if (${GCC_VERSION} VERSION_LESS 6.1)
+        if (${GCC_VERSION} VERSION_LESS 3.6)
             message(FATAL_ERROR "${BoldYellow}Intel ICPC makes use of CLANG (detected: ${GCC_VERSION}; required for C++11: 6.1) so this build won't work without CLANG intervention: http://psicode.org/psi4manual/master/build_planning.html\n${ColourReset}")
```
dgasmith: You moved this back to 3.6, but didn't update the message. Can we really support GCC 3.6 now?

psi4/driver/plugin.py
```diff
@@ -87,7 +88,7 @@ def sanitize_name(name):
 
 # Determine the available plugins
 available_plugins = []
-psidatadir = os.environ.get('PSIDATADIR', None)
+psidatadir = core.get_datadir()
```
dgasmith: We keep building up functions off `core`. In the future it could be nice to move this off to something else like `core.environment.get_datadir()`.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -715,7 +715,12 @@ def construct(cls, parser, mol, key, deffit=None, basstrings=None, return_atomli
         mol.update_geometry()
 
         # Paths to search for gbs files: here + PSIPATH + library
-        psidatadir = os.environ.get('PSIDATADIR', None)
+        try:
+            from psi4 import core
+            psidatadir = core.get_datadir()
+        except ImportError:
+            pass
```
dgasmith: Doesn't this allow the possibility for `psidatadir` to be undefined?

psi4/src/core.cc
```diff
@@ -1343,6 +1350,8 @@ PYBIND11_PLUGIN(core) {
         "Assigns the global frequencies to the values stored in the 3N-6 Vector argument.");
     core.def("set_memory_bytes", py_psi_set_memory, py::arg("memory"), py::arg("quiet")=false, "Sets the memory available to Psi (in bytes).");
     core.def("get_memory", py_psi_get_memory, "Returns the amount of memory available to Psi (in bytes).");
+	core.def("set_datadir", &py_psi_set_datadir, "Sets the path to shared text resources, PSIDATADIR");
+	core.def("get_datadir", &py_psi_get_datadir, "Returns the path to shared text resources, PSIDATADIR");
```
dgasmith: Did GitHub mess up, or is there a weird indent there? Also, any reason not to do lambda functions again? It seems to increase the readability of these tiny wrapper layers.

psi4/src/psi4/libpsio/filemanager.cc
```diff
@@ -42,17 +42,7 @@ namespace psi{
...
-    // TMP environment variables is set
-    if(std::getenv("TMPDIR"))
-        set_default_path(std::getenv("TMPDIR"));
-    else if(std::getenv("TEMP"))
-        set_default_path(std::getenv("TEMP"));
-    else if(std::getenv("TMP"))
-        set_default_path(std::getenv("TMP"));
-    else
-        set_default_path("/tmp");
+    set_default_path("/tmp");
```
dgasmith: Just to be absolutely sure, you checked that `psidatadir` and `PSIO` still play nicely together? It might be good to engineer a unittest for that at some point.

cmake/custom_cxxstandard.cmake
loriab: Ah, whoops, I'll fix it to "C++11: 3.6". Don't be mislead by the variable names; in this case, `${GCC_VERSION}` contains Clang version. Maybe I should rename to CPLR_VERSION.

psi4/driver/qcdb/libmintsbasisset.py
loriab: Yes, but only in QCDB standalone mode and only for the library part of the search path. Smooth sailing (no envvar config) not guaranteed outside of P4.

psi4/src/psi4/libpsio/filemanager.cc
loriab: I checked with all the cleaning commented out:
* plain `psi4`
* `psi4 -s scr`
* `PSI_SCRATCH=scr2 psi4 -s scr`
* `PSI_SCRATCH=scr2 psi4`
* `python input.py`
* `PSI_SCRATCH=scr2 python input.py`

and they all _seem_ to work as advertised. But yes, I'm nervous, too.

Actually, now that I read again, what do `psidatadir` and `PSIO` have to do with each other?

I did also test `psi4 -l alt_psidatadir` and it at least prints in the right place and runs.

psi4/driver/plugin.py
loriab: Yeah, I'm hoping this particular fn will be short-lived.

PR #836, opened 2017-11-03
Use std::make_shared almost everywhere

robertodr:
## Description
This PR is an effort to move towards canonical C++11 usage: 
 1. it removes use of `new` when creating `std::shared_ptr`s in favor of `std::make_shared`.
 2. it uses `auto` on the left-hand side of the `std::shared_ptr` declarations.
 3. it replaces occurrences of `NULL`, signalling the null pointer,  with `nullptr`.
These are changes suggested by the modern C++ best-practices and are explained in detail in Scott Meyers' [Effective Modern C++]. 

#### Prefer `std::make_shared` to direct use of `new`
Using `std::make_shared`:
 1. Reduces code verbosity, especially when coupled with `auto`:
 ```
 std::shared_ptr<Matrix> F = std::shared_ptr<Matrix>(new Matrix("Fock matrix", nso, nso)); // Type information written down 3 TIMES!!!
 std::shared_ptr<Matrix> F = std::make_shared<Matrix>("Fock matrix", nso, nso); // So much typing...
 auto F = std::make_shared<Matrix>("Fock matrix", nso, nso); // Much better!!!!
 ```
 2. Ensures exception safety and prevents resource leaks.
 3. Improves efficiency:
 ```
 // Performs TWO allocations
 std::shared_ptr<Matrix> F = std::shared_ptr<Matrix>(new Matrix("Fock matrix", nso, nso)); 
 // Performs ONE allocation
 auto F = std::make_shared<Matrix>("Fock matrix", nso, nso); 
 ```

_Reference_: Item 21 in [Effective Modern C++].
#### Prefer `auto` to explicit type declarations
Using `auto` reduces and/or avoids:
 1. Verbosity in variable declarations:
 ```
 std::shared_ptr<Matrix> F = std::make_shared<Matrix>("Fock matrix", nso, nso); // So much typing...
 auto F = std::make_shared<Matrix>("Fock matrix", nso, nso); // Much better!
 ```
 2. Problems with uninitialized variables. `auto` works like template type deduction, hence the right-hand side of the declaration **needs to have** an initializer:
  ```
  int x1; // fine, but initialized :(
  auto x2; // WON'T COMPILE!!!
  auto x3 = 1; // fine and initialized
  ```
 3. Problems with unintended type casts and type mismatches that are hard to impossible to catch:
  ```
  std::vector<int> v;
  // !!! The size of a vector is of type std::vector<int>::size_type and is compiler- AND architecture-DEPENDENT
  unsigned sz = v.size(); // might not be correct on some compiler/machines
  auto size = v.size(); // size is ALWAYS of the correct type
  ```
_Reference_: Items 2 and 5 in [Effective Modern C++].
#### Prefer `nullptr` to `0` or `NULL`
`0` is an `int` not a pointer. Almost the same goes for `NULL` though implementations of the language can differ in the details. If you want to overload on pointer types and/or use pointer types with templates, use `nullptr` to signal the null pointer. The correct overload/template parameter will then be deduced. Using `nullptr` also makes the code more readable, especially if `auto` is used consistently throughout.
_Reference_: Item 8 in [Effective Modern C++].
_Note_: This PR only replaces `NULL` with `nullptr`.

## Todos
* **Developer Interest**
  - [x] Converts `std::shared_ptr<TYPE> VARIABLE = std::shared_ptr<TYPE>(new TYPE(PARAMETERS));` to `std::shared_ptr<TYPE> VARIABLE = std::make_shared<TYPE>(PARAMETERS);`
  - [x] Use `auto` on the left-hand side of some of the now converted declarations: `auto VARIABLE = std::make_shared<TYPE>(PARAMETERS);`
  - [x] Replace occurrences of `NULL` with `nullptr`

## Status
- [x] Ready to go

[Effective Modern C++]:
 https://edisciplinas.usp.br/pluginfile.php/1995323/mod_resource/content/1/Effective%20Modern%20C%2B%2B%202014.pdf


loriab:
Yay! Sort of pointless when the type is already there, but do we instead want the first type to be `auto` so that the type info only gets expressed once?

robertodr:
I'll fix the failing build. I had considered `auto` but it seemed to me its usage had been discouraged in Psi4. If that's not the case, I'll do the `auto`-ing.

loriab:
It was, but the `auto` ban has been lifted, esp. for clarity as here.

robertodr:
In the docs or in the PR description?

loriab:
There's an explicit rule & eqn in that C++ book Andy sent around a couple months ago, if you need a citation.

robertodr:
Yes, I am re-reading Scoot Meyers' "Effective Modern C++" items 2 and 5 (on `auto`), 8 (on `nullptr`) and 21 (on `make_shared`).

CDSherrill:
+1 on explaining why the new syntax is preferred.

loriab:
Nice justification write-up! Need to get that into sphinxman rather than just PR comment.

Did you mean _un_init here? `int x1; // fine, but initialized :(` ?

psi4/src/core.cc
```diff
@@ -968,77 +887,52 @@ py::object py_psi_get_option(std::string const& module, std::string const& key)
...
 }
 
-void py_psi_set_active_molecule(std::shared_ptr<Molecule> molecule)
-{
-    Process::environment.set_molecule(molecule);
-}
-void py_psi_set_legacy_molecule(std::shared_ptr<Molecule> legacy_molecule)
-{
+void py_psi_set_active_molecule(std::shared_ptr<Molecule> molecule) { Process::environment.set_molecule(molecule); }
+void py_psi_set_legacy_molecule(std::shared_ptr<Molecule> legacy_molecule) {
```
loriab: No reason to do it immediately (that is, this PR), but many of these should just become lambdas in the pb11 `.def`.

psi4/src/export_functional.cc
```diff
@@ -213,25 +213,25 @@ void export_functional(py::module &m) {
                       std::shared_ptr<BasisExtents>>())
         .def("x",
              [](BlockOPoints &grid) {
-                 SharedVector ret(new Vector("X Grid points", grid.npoints()));
+                 auto ret = std::make_shared<Vector>("X Grid points", grid.npoints());
```
loriab: Yay, this will save a lot of people time on their first explorations to find where on earth `SharedVector` and `SharedMatrix` are defined.

jturney: Oh nice

fevangelista:
Bravissimo!

PR #840, opened 2017-11-06
Patches up wiggles in SAPT(DFT) FDDS Dispersion

dgasmith:
## Description
Small tweak to SAPT(DFT) to help eliminate small ~30cal/mol wiggles in the PES. Tracked this down to issues in allowing the derivatives of very small densities leading to quite spurious results. This was changed to the same tolerance as the V derivative builds. Does change the result by ~1cal/mol for a test case- this is lost in the noise of DFT.

## Status
- [x] Ready to go


dgasmith:
SAPT_QUIET mostly tells the other SAPT programs to be quiet when called so you can do it more in a API fashion. Its not really quite the same as SAPT_PRINT or PRINT_LVL unfortunately.

loriab:
Ah, `SAPT_QUIET` is an I'm-the-driver-so-you-shut-up signal? Got it.

psi4/src/read_options.cc
```diff
@@ -941,6 +941,8 @@ int read_options(const std::string &name, Options & options, bool suppress_print
     options.add_int("SAPT_FDDS_DISP_NUM_POINTS", 10);
     /*- Lambda shift in the space morphing for the FDDS Dispersion time integration !expert -*/
     options.add_double("SAPT_FDDS_DISP_LEG_LAMBDA", 0.3);
+    /*- Minima rho cutoff for the in the LDA response for FDDS -*/
+    options.add_double("SAPT_FDDS_V2_RHO_CUTOFF", 1.e-6);
```
andysim: Agree with @loriab on the minimum/minima switcharoo, but should this be `!expert`?  I would think so.

dgasmith: Good point, fixed.

PR #844, opened 2017-11-09
External field cleanup

andysim:
## Description
@dgasmith and I had to do some preliminary work to get the Hessian code extended, and here's the result.  Essentially, we've centralized the construction of the core Hamiltonian and its derivatives, which ensures that different codes will correctly obey any new one electron contributions (field perturbation, MM point charges, ECPs, *etc.*) and derivatives thereof.

## Todos
* **Developer Interest**
  - [x] @psi-rking: your perturbed CC gradient request should be addressed by this fix.
  - [x] Test cases for DFOCC and CCENERGY gradients, with dipole field.
  - [x] Improved support for ECPs and external fields, and their gradients, in correlated calculations. 

## Questions
- [x] How should we handle the new signature for mol's `nuclear_repulsion_energy()`?  The current solution is to mandate that the field be specified C++-side, but py-side just provides a default zero field.  Another option would be to overload with an arg-less function, and provide a deprecation warning for that version (to force the programmer to think about whether they really mean to compute the nuc. energy ignoring the field).

## Status
- [x] Ready to go


andysim:
Heads-up, @bozkaya, there are some changes in DFOCC that should make your code play nicely with external potentials and the like.  The full test suite passes right now, but I welcome any input from you.

psi4/src/psi4/libmints/deriv.cc
```diff
@@ -714,7 +680,8 @@ SharedMatrix Deriv::compute()
     }
 
     // Obtain nuclear repulsion contribution from the wavefunction
-    SharedMatrix enuc(new Matrix(molecule_->nuclear_repulsion_energy_deriv1()));
+    std::vector<double> field_strength = wfn_->get_dipole_field_strength();
+    SharedMatrix enuc(new Matrix(molecule_->nuclear_repulsion_energy_deriv1(field_strength)));
```
robertodr: `SharedMatrix enuc = std::make_shared<Matrix>(molecule_->nuclear_repulsion_energy_deriv1(field_strength));` just because I was on a recent `make_shared` rampage :wink:

psi-rking:
Wow.  What a lot of tedious changes.  But looks very good to me.

psi4/src/psi4/libmints/molecule.h
```diff
@@ -364,13 +364,15 @@ class Molecule
     /// Computes center of mass of molecule (does not translate molecule)
     Vector3 center_of_mass() const;
     /// Computes nuclear repulsion energy
-    double nuclear_repulsion_energy() const;
+    /// The dipole field is a vector of length 3, containing the dipole field strength in the {x,y,z} directions.
+    double nuclear_repulsion_energy(std::vector<double> dipole_field) const;
```
jturney: Maybe these should be `const std::array<double, 3>` then you can remove the length check in the function itself.

andysim: Oh, that's a great idea.  I originally wanted `Vector3`, but didn't want to use a pointer, or include the header in `wavefunction.h`.  Using `std::array` gives the right behavior, with none of the drawbacks.

psi4/src/psi4/libmints/deriv.cc
andysim: Good catch!  Will change this...

psi4/src/psi4/dfocc/get_moinfo.cc
```diff
@@ -339,35 +339,10 @@ void DFOCC::get_moinfo() {
...
-    IWL::read_one(psio_.get(), PSIF_OEI, PSIF_SO_V, so_ints, ntri_so, 0, 0, "outfile");
-    Vso_->set(so_ints);
-    IWL::read_one(psio_.get(), PSIF_OEI, PSIF_SO_S, so_ints, ntri_so, 0, 0, "outfile");
-    Sso_->set(so_ints);
-    free(so_ints);
-    Hso_->copy(Tso_);
-    Hso_->add(Vso_);
-    Tso_.reset();
-    Vso_.reset();
     Hso = SharedTensor2d(new Tensor2d("SO-basis One-electron Ints", nso_, nso_));
```
loriab: FYI, @CDSherrill, I think they're out to change your dfocc OEI yet again.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -258,64 +258,7 @@ void HF::common_init()
...
-            else if (perturb_with == "SPHERE") {
-                perturb_ = sphere;
-                perturb_dipoles_[0] = 1.0;
-            }
-            else {
-
-                    outfile->Printf( "Unknown PERTURB_WITH. Applying no perturbation.\n");
-
-            }
-        } else {
```
loriab: yay, between us, we're much shrinking `HF::common_init()`

psi4/src/export_mints.cc
```diff
@@ -1011,7 +1021,7 @@ void export_mints(py::module& m)
              "Prints the molecule in Cartesians in input units adding fragment separators")
         .def("rotational_constants", &Molecule::rotational_constants,
              "Prints the rotational constants of the molecule")
-        .def("nuclear_repulsion_energy", &Molecule::nuclear_repulsion_energy,
+        .def("nuclear_repulsion_energy", &Molecule::nuclear_repulsion_energy, py::arg("dipole_field") = std::vector<double>(3,0.0),
```
loriab: I gather C-side Mol.nre() users are required to explicitly pass in any perturbation, while py-side callers can do so or remain oblivious? Sounds fair.

loriab:
Probably a patch or two needed to the plugin templates to satisfy Travis.

psi4/src/psi4/dfocc/get_moinfo.cc
andysim: Yup, centralizing it will allow various terms like ECPs to work transparently.  IF you know of any other offending lines of code stuck in T+V world, please let me know.

psi4/src/export_mints.cc
andysim: That's the status quo, yes.  If you're fine with that, I am too.  The hidden danger is that somebody coding stuff py-side may just expect external fields to work transparently with their code and it won't.  Kinda like what happened to us C-side before now...

loriab: I wonder, can one overload `Wavefunction.molecule().nuclear_repulsion_energy()` so that it pulled the perturbation off the wfn and passed it into the mol.nre(). Would one even want to? Clearly for an un-owned Mol, a programmer couldn't expect any perturbation considerations w/o passing them explicitly.

psi-rking:
Just a note to whomever that I will use this all-but-done update as soon as we can get it merged.  Let me know if I can help.  Whose problem is the dfocc one to solve?

dgasmith:
Ok fixed up the plugins, modernized the SCF plugin example, and a few other odds and ends. This should be ready to go.

andysim:
Oops, I ran the tests when I submitted this one, but didn't check the plugins.  Thanks a lot for dealing with that, @dgasmith

dgasmith:
Hmm, neither Andy or I can really approve this. Can someone else from @psi4/maintainers take a look (holiday time permitting)?

dgasmith:
@psi4/maintainers This is ready to go, could use one more review and merge.

PR #845, opened 2017-11-09
Re-enables sanitizer flags with the CMake superbuild

jturney:
## Description
During the switch over to the super build the `ENABLE_{ASAN,TSAN,UBSAN}` flags stopped working. This PR fixes them.

To use these sanitizers you must compile a copy of Python with the corresponding flags added.  If you do not then you'll receive runtime errors.

## Status
- [x] Ready to go


CMakeLists.txt
```diff
@@ -110,12 +113,6 @@ option_with_flags(ENABLE_CODE_COVERAGE "Enables details on code coverage" OFF
                   "-ftest-coverage")
 option_with_flags(ENABLE_BOUNDS_CHECK "Enables bounds check in Fortran" OFF
                   "-ftrapuv -check all -fpstkchk" "-fcheck=all" "-fbounds-check -fcheck-array-temporaries")
-option_with_flags(ENABLE_ASAN "Enables address sanitizer" OFF
```
robertodr: I know only recent Clang and GCC support these flags, but why skip the `check_cxx_compiler_flag` in `option_with_flags`?

jturney: On my system the flags are not passed to the linker when using `check_cxx_compiler_flag` which are needed and the check fails.

CMakeLists.txt
```diff
@@ -77,6 +77,9 @@ option(ENABLE_erd "Enables use of ERD instead of Libint (requires Fortran)" OFF)
 option(ENABLE_simint "Enables use of SIMINT two-electron integral library" OFF)
 option(ENABLE_gdma "Enables Stone's GDMA multipole code (requires Fortran)" OFF)
 option(ENABLE_PCMSolver "Enables PCMSolver library (requires Fortran)" OFF)
+option(ENABLE_ASAN "Enables address sanitizer" OFF)
+option(ENABLE_TSAN "Enables thread santizer" OFF)
+option(ENABLE_UBSAN" Enables undefined behavior sanitizer" OFF)
```
robertodr: What about also adding the MemorySanitizer?

jturney: Can do.

loriab: Add a "(requires similarly-compiled Python and Numpy" to the string, maybe.

CMakeLists.txt
loriab: try_compile doesn't like it. But if it could be made to work, that'd def. be better.

robertodr: `try_compile` or `check_cxx_source_compiles` would probably be more flexible/robust in this situation. I'll try it out.

robertodr:
This is very cool! I have an instance of [Drone CI](https://www.drone-ci.science/) that can be used to run these sanitizer jobs on each commit.

CMakeLists.txt
robertodr: The following works:
```
cmake_minimum_required(VERSION 3.1 FATAL_ERROR)

project(recipe-01 CXX)

include(CheckCXXSourceCompiles)
set(_source "int main() { return 0; }")
set(CMAKE_REQUIRED_FLAGS "-fsanitize=address -fno-omit-frame-pointer")
check_cxx_source_compiles("${_source}" _asan_works)
```
 `check_cxx_source_compiles` forwards the flags to the linker too.

loriab: interesting! maybe `check_cxx_source_compiles` should replace the fn in cmake/psi4OptionsTools? (though that's in a lot of codes now and shouldn't be done casually.)

jturney:
I've added memory sanitizer and comments that Python and Numpy need to be compiled similarly.  I'm happy to change to using `check_cxx_source_compiles` if that's the way it needs to go.

robertodr:
What do you say @loriab?

loriab:
Ok, out of a mtg. 

I don't see why https://cmake.org/cmake/help/v3.8/module/CheckCXXCompilerFlag.html (which we have now and supposedly calls https://cmake.org/cmake/help/v3.9/module/CheckCXXSourceCompiles.html under the hood _doesn't work_ (Jet and I saw Thursday) while the latter _does work_ (for RDR above). Could you try it out in situ in psi, @robertodr? Why is the former setting `CMAKE_REQUIRED_DEFINES`, not `CMAKE_REQUIRED_FLAGS`?

If only CheckCXXSourceCompiles works, I'd advocate replacing https://github.com/psi4/psi4/blob/master/cmake/psi4OptionsTools.cmake#L102 (and friends) by `check_cxx_source_compiles` and CMAKE_REQUIRES_..., switching ASAN and Co. back to `option_with_flags` and deleting the new `custom_sanitize` workaround file. Also, the quotes grouping of some of the flags in `option_with_flags` needs revisiting ( for this, should only be one string, since only one set of flags known for clang/gcc compilers).

If we  instead stick with the current working setup, LGTM.

robertodr:
Since `check_<lang>_compiler_flag` calls `check_<lang>_source_compiles` defining `CMAKE_REQUIRED_FLAGS` right before the former is called does the trick. I don't know why this isn't done automatically or why it isn't clearly documented, but it works. This is the full patch:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index ca7e20054..6cc4ee09e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,10 +77,6 @@ option(ENABLE_erd "Enables use of ERD instead of Libint (requires Fortran)" OFF)
 option(ENABLE_simint "Enables use of SIMINT two-electron integral library" OFF)
 option(ENABLE_gdma "Enables Stone's GDMA multipole code (requires Fortran)" OFF)
 option(ENABLE_PCMSolver "Enables PCMSolver library (requires Fortran)" OFF)
-option(ENABLE_ASAN "Enables address sanitizer (requires similarly compiled Python and Numpy)" OFF)
-option(ENABLE_TSAN "Enables thread sanitizer (requires similarly compiled Python and Numpy)" OFF)
-option(ENABLE_UBSAN" Enables undefined behavior sanitizer (requires similarly compiled Python and Numpy)" OFF)
-option(ENABLE_MSAN" Enables memory sanitizer (requires similarly compiled Python and Numpy)" OFF)
 # These options are relevant to pasture
 option(ENABLE_ccsort "Enables ccsort plugin installed from psi4pasture" OFF)
 option(ENABLE_transqt2 "Enables transqt2 plugin installed from psi4pasture" OFF)
@@ -114,6 +110,14 @@ option_with_flags(ENABLE_CODE_COVERAGE "Enables details on code coverage" OFF
                   "-ftest-coverage")
 option_with_flags(ENABLE_BOUNDS_CHECK "Enables bounds check in Fortran" OFF
                   "-ftrapuv -check all -fpstkchk" "-fcheck=all" "-fbounds-check -fcheck-array-temporaries")
+option_with_flags(ENABLE_ASAN "Enables address sanitizer (requires similarly compiled Python and Numpy)" OFF
+                  "-fsanitize=address -fno-omit-frame-pointer")
+option_with_flags(ENABLE_TSAN "Enables thread sanitizer (requires similarly compiled Python and Numpy)" OFF
+                  "-fsanitize=thread -fPIE -pie -fno-omit-frame-pointer")
+option_with_flags(ENABLE_UBSAN "Enables undefined behavior sanitizer (requires similarly compiled Python and Numpy)" OFF
+                  "-fsanitize=undefined -fno-omit-frame-pointer")
+option_with_flags(ENABLE_MSAN "Enables memory sanitizer (requires similarly compiled Python and Numpy)" OFF
+                  "-fsanitize=memory -fPIE -pie -fno-omit-frame-pointer")
 option_with_default(MAX_AM_ERI "Maximum angular momentum for integrals" 5)
 option_with_default(CMAKE_BUILD_TYPE "Build type (Release or Debug)" Release)
 option_with_default(FC_SYMBOL "The type of Fortran name mangling" 2)
diff --git a/cmake/custom_sanitize.cmake b/cmake/custom_sanitize.cmake
deleted file mode 100644
index bdcda4962..000000000
--- a/cmake/custom_sanitize.cmake
+++ /dev/null
@@ -1,44 +0,0 @@
-if(ENABLE_ASAN)
-    if((CMAKE_CXX_COMPILER_ID MATCHES Clang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES AppleClang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES GNU))
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -fno-omit-frame-pointer")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address -fno-omit-frame-pointer")
-    else()
-        message(WARNING "ASAN flags are not known for your compiler ${CMAKE_CXX_COMPILER_ID}")
-    endif()
-endif()
-
-if(ENABLE_TSAN)
-    if((CMAKE_CXX_COMPILER_ID MATCHES Clang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES AppleClang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES GNU))
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=thread -fno-omit-frame-pointer -pie")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=thread -fno-omit-frame-pointer -pie")
-    else()
-        message(WARNING "TSAN flags are not known for your compiler ${CMAKE_CXX_COMPILER_ID}")
-    endif()
-endif()
-
-if(ENABLE_UBSAN)
-    if((CMAKE_CXX_COMPILER_ID MATCHES Clang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES AppleClang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES GNU))
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=undefined -fno-omit-frame-pointer")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=undefined -fno-omit-frame-pointer")
-    else()
-        message(WARNING "UBSAN flags are not known for your compiler ${CMAKE_CXX_COMPILER_ID}")
-    endif()
-endif()
-
-if(ENABLE_MSAN)
-    if((CMAKE_CXX_COMPILER_ID MATCHES Clang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES AppleClang) OR
-        (CMAKE_CXX_COMPILER_ID MATCHES GNU))
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=memory -fno-omit-frame-pointer")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=memory -fno-omit-frame-pointer")
-    else()
-        message(WARNING "MSAN flags are not known for your compiler ${CMAKE_CXX_COMPILER_ID}")
-    endif()
-endif()
-
diff --git a/cmake/psi4OptionsTools.cmake b/cmake/psi4OptionsTools.cmake
index 024e66285..f95c5081b 100644
--- a/cmake/psi4OptionsTools.cmake
+++ b/cmake/psi4OptionsTools.cmake
@@ -74,7 +74,7 @@ endmacro()
 include(CheckCCompilerFlag)
 include(CheckCXXCompilerFlag)
 if(CMAKE_Fortran_COMPILER)
-    include(CheckFortranCompilerFlag)  # CMake >= 3.3, so local copy in cmake/
+    include(CheckFortranCompilerFlag)
 endif()
 
 #The guts of the next two functions, use the wrappers please
@@ -95,6 +95,7 @@ set(CMAKE_REQUIRED_QUIET_SAVE ${CMAKE_REQUIRED_QUIET})
          break()
       endif()
       unset(test_option CACHE)
+      set(CMAKE_REQUIRED_FLAGS "${flag_i}")
       if(${is_C} EQUAL 0)
           CHECK_C_COMPILER_FLAG("${flag_i}" test_option)
           set(description_to_print CMAKE_C_FLAGS)
@@ -117,6 +118,7 @@ set(CMAKE_REQUIRED_QUIET_SAVE ${CMAKE_REQUIRED_QUIET})
            message(STATUS "${msg_base} Failed")
         endif()
       endif()
+      unset(CMAKE_REQUIRED_FLAGS)
    endforeach()
    set(CMAKE_REQUIRED_QUIET ${CMAKE_REQUIRED_QUIET_SAVE})
 endmacro()
```

loriab:
If it's been tested to work correctly on 
```
option_with_flags(ENABLE_ASAN "Enables address sanitizer (requires similarly compiled Python and Numpy)" OFF
                  "-fsanitize=address -fno-omit-frame-pointer" "-nonsense")
```
and
```
"-nonsense" "-fsanitize=address -fno-omit-frame-pointer")
```
then I really like @robertodr 's patch.

Do you think the `unset(CMAKE_REQUIRED_FLAGS)` needs to be `CACHE`? I don't understand these things; just know that they can go very wrong.

robertodr:
It still works for the `-xHost` `-march=native` case, so I would say it passes your test (haven't tried though!) I was asking myself the same about `CACHE`, but I think it's never set in cache, so we should be safe.

loriab:
Ok, great, @robertodr, so long as xHost/march=native (or the nonsense one) works in both directions.

robertodr:
I was a bit careless making modifications on your fork @jturney and a merge commit made it into the PR. Just squash it away.

jturney:
Great, thanks for making the changes!

jturney:
I can confirm this works on my system. We should be good to go!

robertodr:
I'll set up CI builds for `AddressSanitizer` on my [Drone CI](https://www.drone-ci.science/) instance as soon as this makes it into `master` :smile_cat:

PR #810, opened 2017-09-27
[WIP] PCM-CCSD in the PTED approximation

robertodr:
## Description
This PR enables CCSD calculations with the PCM in the [PTED (Perturbation-To-Energy-and-Density) approximation](http://dx.doi.org/10.1063/1.3245400). The PCM potential is calculated from the CCSD _density_. Thus macroiterations are introduced to solve the CCSD amplitude **and** multiplier equations, obtain the CCSD density and update the PCM potential. @jH0ward the macroiterations are handled in the Python layer. 

There are two test cases, both currently failing, and the code is quite ugly.

## Todos
* **User-Facing for Release Notes**
  - [ ] CCSD with the PCM in the PTED approximation.

## Status
- [ ] Ready to go

PR #847, opened 2017-11-10
SCF iteration py-side

loriab:
## Description
In order to bring back SCF hooks, kill the C-side EFP octopus, and provide a foundation for a smart SCF driver, want control over SCF iterations to be at the driver level.

Features will have to be filled out more fully from the commit msgs once this stabilizes, but current status is (1) rebased, (2) no EFP or PCM yet, (3) otherwise, all tests pass, (4) haven't done a final output vs. output.ref comparison.

Please to assess mercilessly the general structure and stable parts of the C/Py handshake. There's very little that I'll defend energetically. There's a few questions and notes marked with `Q:` in the code.

Hack away in the "pre-review" time. That is, this can't be merged b/c it breaks addons. But it'd be nice if this was tentatively pre-approved before continuing on into efp, pcm, and json stages.

## Status
- [ ] Ready to go


robertodr:
Nice @loriab! As we, very briefly, discussed at WWDC: 
1. I had to rework the way the PCM potential is added to the Fock matrix in `hf.cc` in #810. I ended up adding a `std::vector<SharedMatrix> external_potentials_` to the base HF objects to store the PCM potential computer at each iteration. This data member is then used in `form_F()` in the derived classes. This is in `rohf.cc`, for example:
```
// ... some more code ...

if (!external_potentials_.empty()) {
       std::for_each(external_potentials_.begin(), external_potentials_.end(),
                     [this](SharedMatrix Vext) { Fa_->add(Vext); });
}

// ... some more code ...

if (!external_potentials_.empty()) {
       std::for_each(external_potentials_.begin(), external_potentials_.end(),
                     [this](SharedMatrix Vext) { Fb_->add(Vext); });
}
```
For the PCM (and I'm pretty sure also EFP) potential, this is enough: these do not need to distinguish between alpha and beta.
The caveat of this approach is that `compute_E()` cannot rely on tracing the Fock matrix with the density matrix to calculate the energy anymore, since there might be more (or less) polarization energy included in that way. In the mentioned PR, I modified the `compute_E()` functions to only use traces of one- and two-electron integrals to calculate the energy. I think this approach could be ported Python side and would fit very nicely with the infrastructure changes you are proposing.
2. Since PCMSolver itself is quite far from being pybind-able, I suggest I pybind the wrapper within Psi4 (`libpsipcm/psipcm.h` and `libpsipcm/psipcm.cc`)

psi4/driver/procrouting/scf/scf_iterator.py
```diff
@@ -0,0 +1,583 @@
...
+    # #endif
+
+
+def scf_iterate(self, e_conv=None, d_conv=None):
+
+    df = core.get_option('SCF', "SCF_TYPE") == "DF"
+    verbose = core.get_option('SCF', "PRINT")
+    reference = core.get_option('SCF', "REFERENCE")
+
+    def converged(e_delta, d_rms, e_conv=e_conv, d_conv=d_conv):
```
dgasmith: Would add this as a `_converged` function outside of `scf_iterate`.

psi4/driver/procrouting/scf/scf_iterator.py
```diff
@@ -0,0 +1,583 @@
...
+    if enabled:
+        start = core.get_option('SCF', "MOM_START")
+        if enabled < 0:
+            raise ValidationError('SCF MOM_START ({}) must be at least 1'.format(start))
+
+    # note that this is excited-MOM-enabled, which is the more interesting flag
+    return enabled
+
+
+def _validate_soscf():
```
dgasmith: Hate to ask, but can you doc some of the logic trees on these?

loriab: I didn't change any of the logic on damp/diis/soscf/mom/frac (I hope). Doc for review, doc for general interest in sphinxman/hf.rst or read_options.cc, doc for programmers meeting the SCF code?

The checks in the `def _validate_*` fns are pretty much just sanity checks that may/may not have been elsewhere.

psi4/driver/procrouting/__init__.py
```diff
@@ -31,3 +31,5 @@
 from .empirical_dispersion import EmpericalDispersion
 from . import dft_funcs
 from . import response
+
+from . import scf # FIXME! DONT DO THIS.
```
loriab: btw, @dgasmith, what did you have in mind instead here?

psi4/driver/procrouting/scf/scf_iterator.py
dgasmith: Mostly just a doc for someone reading the code, probably something as simple as "Validates SOSCF start, convergences, and iteration options" would work just fine.

psi4/driver/procrouting/__init__.py
dgasmith: I think I was worried about fudging up the namespace. The name `scf` is a bit too generic I think. Perhaps something like `scf_procrouting` or some such.

loriab: Got it. This, and your other points, should be cleaned up now.

dgasmith:
We have a bit of a chicken and egg problem here with regard to getting this. This is likely waiting on EFP/PCM Python side stuff which is waiting on getting this in as some of the issues overlap. We can either merge the EFP/PCM/Python SCF iterations into this single PR or merge in the SCF iterations in *now* and patch up EFP/PCM later. Waiting on this however is likely a bad idea however as SCF is so fundamental to Psi and frequently is changed.

While we have stated that psi4/psi4:master is only required to pass quick tests and only full releases require the full test suite to pass we have often stayed in the realm of having the full test suite pass all of the time. This has led users to pick up the current HEAD and using that as their main to get up to date features. This goes a bit into a larger discussion that we likely should mint more frequent minor updates so that we can feel free to break psi4/psi4:master as needed. That being said as long as we properly fence PCM/EFP being broken with large flashing warning lights a temporary break should be fine as long as its resolved in a timely manner.

@robertodr @loriab @psi4/admins

loriab:
Quite right on the chicken/egg problem. One approach I considered was to get normal 3-round approval of this PR with the current scf-only changes. The re-rebase it, `c-f` it, and post it to psi4/psi4 as a branch. _Then_ ppl can make additional EFP/PCM/etc. PRs to that branch rather than master, so all the discussion will stay on psi4/psi4 but we won't have an enormous final PR that no one wants to review.

For longer term, though, yes, if we had a stable, vetted, low-key release every couple months, there wouldn't be this considerable pressure to maintain master in top-notch working order. And if we wanted to deprecate something, wouldn't take 2 years to get through warning release, then error release.

robertodr:
I agree with @loriab's proposed roadmap.

dgasmith:
Ok let’s get that going then. @loriab any changes to make before last review and pulling this in as a branch?

-Daniel Smith

loriab:
It hasn't had a final comprehensive diff vs `output.ref`s to make sure every change is legit (that is, I know there's differences in the 0 iter guess, but I haven't traced if they're anything to do with this PR), but it's been extensively spot-checked to make sure all the little SCF pieces are still working. So if you're all ok with that level, yes, this is ready for pre-branch review (guess that doesn't actually close this PR, since I'll just push to another branch).

dgasmith:
Moved this base to `psi4:scfitertopy`. I say merge this in now IMO and then start rebasing the other PR's off that branch.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -1756,7 +1756,7 @@ void HF::iterations()
                 base_name = "SOSCF, nmicro = ";
             }
 
-            if (!test_convergency()) {
+            if (!false) {
```
jturney: Is this right, or did it get changed in a later commit?

jturney: Never mind, this part of the C-side SCF removal.

Issue #846, opened 2017-11-09
UnboundLocalError in libmintsgshell


rmcgibbo:
[This block of code](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/libmintsgshell.py#L222-L228) in libmintsgshell.py throws a UnboundLocalError when trying to recover from the ZeroDivisionError

```
        try:
            norm = math.sqrt(1.0 / (tmp * e_sum))
        except ZeroDivisionError:
            self.PYcoef[i] = [1.0 for i in range(self.nprimitive())]
        # Set the normalization
        for i in range(self.nprimitive()):
            self.PYcoef[i] *= norm
```

Traceback
```
Traceback (most recent call last):
  File "/prefix/bin/psi4", line 260, in <module>
    exec(content)
  File "<string>", line 27, in <module>
  File "/prefix/lib//psi4/driver/driver.py", line 460, in energy
    wfn = procedures['energy'][lowername](lowername, molecule=molecule, **kwargs)
  File "/prefix/lib//psi4/driver/procrouting/proc.py", line 1990, in run_scf
    scf_wfn = scf_helper(name, post_scf=False, **kwargs)
  File "/prefix/lib//psi4/driver/procrouting/proc.py", line 1279, in scf_helper
    base_wfn = core.Wavefunction.build(scf_molecule, core.get_global_option('BASIS'))
  File "/prefix/lib//psi4/driver/p4util/python_helpers.py", line 101, in pybuild_wavefunction
    basis = core.BasisSet.build(mol, "ORBITAL", basis)
  File "/prefix/lib//psi4/driver/p4util/python_helpers.py", line 68, in pybuild_basis
    key, resolved_target, fitrole, other, return_atomlist=return_atomlist)
  File "/prefix/lib//psi4/driver/qcdb/libmintsbasisset.py", line 661, in pyconstruct
    return_atomlist=return_atomlist)
  File "/prefix/lib//psi4/driver/qcdb/libmintsbasisset.py", line 876, in construct
    ecpbasisset = BasisSet(key, mol, ecp_atom_basis_shell)
  File "/prefix/lib//psi4/driver/qcdb/libmintsbasisset.py", line 148, in __init__
    self.constructor_role_mol_shellmap(*args)
  File "/prefix/lib//psi4/driver/qcdb/libmintsbasisset.py", line 349, in constructor_role_mol_shellmap
[...]

PR #849, opened 2017-11-11
Issue 846 Fix; Changed 0 Coefficient Handling

JonathonMisiewicz:
## Description
Issue #846 has been fixed, so a contraction set where all primitives have a coefficient of 0 will remain untouched. While I was poking around the file, I decided to clean up some other things.

## Todos
* **Developer Interest**
  - [X] 0 coefficient primitives will no longer be normalized. This should prove useful to planned ECP changes.
  - [X] Miscellaneous code cleanup.

## Questions
- [ ] ECP team, I assume you're happy with this? 
  - [ ] In the zero coefficient case, do we want to explicitly check this is an ECP by checking radial powers? If so, what would we do otherwise, raise some error?
- [ ] I also assume that I'm within the Psi4 coding style here?

## Status
- [X] Ready to go

### DIFF
--- a/psi4/driver/qcdb/libmintsgshell.py
+++ b/psi4/driver/qcdb/libmintsgshell.py
@@ -26,9 +26,6 @@
 # @END LICENSE
 #
 
-"""Class to
-
-"""
 from __future__ import absolute_import
 from __future__ import print_function
 import math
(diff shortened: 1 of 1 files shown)


psi4/driver/qcdb/libmintsgshell.py
```diff
@@ -193,7 +188,7 @@ def __init__(self, am, c, e, pure, nc, center, start, pt='Normalized', rpowers=N
             self.normalize_shell()
             self.erd_normalize_shell()
         else:
-            self.PYerd_coef = [0.0 for i in c]
+            self.PYerd_coef = [0.0] * self.nprimitive()
```
andysim: The `[..]*integer` version here makes a list of references to the same float object, which has different consequences for mutable vs. immutable types.  The list comprehension `[ 0.0 for i in c]` creates distinct objects, removing the need to know whether the type is immutable or not, which is why I tend to prefer that version.  In this case both are safe, as the float is immutable, so it's fine to use either.  I'm commenting here to point out this difference for future reference, and to solicit for opinions from other devs who know Python better than I, in case what I said is wrong.

PR #850, opened 2017-11-11
SNS-MP2 Plugin Addition

shannonhouck:
## Description
This PR adds the [SNS-MP2 plugin](https://github.com/DEShawResearch/sns-mp2) to the list of endorsed Psi4 plugins such that the SNS-MP2 method can be called by energy() without the use of any import statements.

## Todos
  - [x] Added SNS-MP2 (via endorsed_plugins.py)
  - [x] Added documentation for SNS-MP2
  - [ ] Make sure a full Psi4 install includes SNS-MP2
  - [ ] Add tests

## Questions
- [ ] Is there anything I should add to or change about the documentation? (Tagging @rmcgibbo because he wrote the code, as well as @dgasmith and @loriab because of their general Psi4 and documentation knowledge.)
- [ ] The plugin itself comes with several tests already. Should I incorporate those tests into the standard Psi4 test suite?
- [ ] I've borrowed several lines of code for the installation and example [from the plugin's README](https://github.com/DEShawResearch/sns-mp2/blob/master/README.md). Is this ok, or should I change it?

## Status
- [ ] Ready to go

### DIFF
--- a/doc/sphinxman/source/plugin_snsmp2.rst
+++ b/doc/sphinxman/source/plugin_snsmp2.rst
@@ -49,7 +49,7 @@ Spin-Network-Scaled MP2 (SNS-MP2) by D. E. Shaw
 This plugin is an implementation of the SNS-MP2 algorithm developed by McGibbon 
 et. al. [McGibbon:2017:161725]_. The SNS-MP2 method uses neural networking to 
 improve the accuracy of MP2 (:ref:`dfmp2`) interaction energies for dimer molecules. 
-The plugin is distributed under the BSD-2 license.
+The plugin is distributed under the 2-clause BSD license.
 
 Installation
 ~~~~~~~~~~~~
--- a/doc/sphinxman/source/plugin_snsmp2.rst
+++ b/doc/sphinxman/source/plugin_snsmp2.rst
@@ -33,7 +33,7 @@
 Spin-Network-Scaled MP2 (SNS-MP2) by D. E. Shaw
 ===============================================
 
-.. codeauthor:: D. E. Shaw
+.. codeauthor:: D. E. Shaw Research
 .. sectionauthor:: Shannon E. Houck
 
 .. image:: https://img.shields.io/badge/home-sns--mp2-5077AB.svg
@@ -46,11 +46,9 @@ Spin-Network-Scaled MP2 (SNS-MP2) by D. E. Shaw
 .. image:: https://img.shields.io/badge/docs-latest-5077AB.svg
    :target: https://github.com/DEShawResearch/sns-mp2/blob/master/README.md
 
-This plugin is an implementation of the SNS-MP2 algorithm, originally 
-developed by McGibbon et. al. [McGibbon:2017:161725]_, which uses neural networking to 
-improve the accuracy of MP2 (:ref:`dfmp2`) interaction energies for dimers. 
-This |PSIfour| plugin allows the user to compute both energies and 
-confidence intervals.
+This plugin is an implementation of the SNS-MP2 algorithm developed by McGibbon 
+et. al. [McGibbon:2017:161725]_. The SNS-MP2 method uses neural networking to 
+improve the accuracy of MP2 (:ref:`dfmp2`) interaction energies for dimer molecules.
 
 Installation
 ~~~~~~~~~~~~
@@ -77,7 +75,7 @@ Installation
 Sample Input
 ~~~~~~~~~~~~
 
-A sample input file, borrowed from D. E. Shaw's documentation, is shown below::
+A sample input file, adapted from the documentation, is shown below::
 
    # Sample SNS-MP2 calculation for two helium atoms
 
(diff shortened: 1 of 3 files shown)


doc/sphinxman/source/plugin_snsmp2.rst
```diff
@@ -0,0 +1,99 @@
...
+.. raw:: html
+
+   <br>
+
+.. image:: https://img.shields.io/badge/docs-latest-5077AB.svg
+   :target: https://github.com/DEShawResearch/sns-mp2/blob/master/README.md
+
+This plugin is an implementation of the SNS-MP2 algorithm developed by McGibbon 
+et. al. [McGibbon:2017:161725]_. The SNS-MP2 method uses neural networking to 
+improve the accuracy of MP2 (:ref:`dfmp2`) interaction energies for dimer molecules.
```
andysim: It's probably worth mentioning that their code is distributed under the [BSD 2-clause](https://opensource.org/licenses/BSD-2-Clause) license, because it's different from our LGPL license.  It's not a big deal, as BSD-2 is more permissive than LGPL, but it could help out some legal folks.

doc/sphinxman/source/plugin_snsmp2.rst
```diff
@@ -0,0 +1,99 @@
...
+   He 0 0 0
+   --
+   He 2 0 0
+   }
+
+   energy('sns-mp2')
+    
+Note that the two monomers are separated by double dashes, indicating that
+they should be treated as separate molecules. (See 
+:ref:`sec:analysis-of-intermolecular-interactions` for more details on
```
andysim: +1 for the explanation of the dash syntax - that'll really help casual readers who just want a quick look at the docs

rmcgibbo:
Would it make sense to add the SNS-MP2 plugin to any of the psi4 conda packages concurrent with this, to make it easier to access for users? (I'm not sure if that's standard practice for "endorsed plugins" or not.)

loriab:
Thanks, @rmcgibbo, it actually is standard practice to make a conda package of "endorsed plugins". I haven't gotten to that, but if you pre-approve, that's very helpful. :-)

I also have a plan that even optional run-time dependencies (e.g., plugins) can be built at CMake superbuild-time with `-DENABLE_rt_dep=ON`. This is purely user-convenience since they're downstream of psi.

Are you good with either/both of these options for SNS-MP2?

doc/sphinxman/source/plugin_snsmp2.rst
shannonhouck: I've added this info to the docs. Thank you!

rmcgibbo:
Absolutely.

doc/sphinxman/source/plugin_snsmp2.rst
```diff
@@ -49,7 +49,7 @@ Spin-Network-Scaled MP2 (SNS-MP2) by D. E. Shaw
 This plugin is an implementation of the SNS-MP2 algorithm developed by McGibbon 
 et. al. [McGibbon:2017:161725]_. The SNS-MP2 method uses neural networking to 
 improve the accuracy of MP2 (:ref:`dfmp2`) interaction energies for dimer molecules. 
-The plugin is distributed under the BSD-2 license.
+The plugin is distributed under the 2-clause BSD license.
```
andysim: Oops - I didn't see this sentence when I left my previous comment, which is why I requested the license info be added.  Must caffeinate better in future; sorry for the confusion.

shannonhouck: Nope, you were right! Both commits were added after your commentary-- the second commit was just to fix the phrasing of it. Thank you!

dgasmith:
Thanks for making this PR, everything looks great so far to me. I think the last two items are:
 - Ensure a full install of Psi4 includes this plugin.
 - Manage the tests issue. It looks like SNS-MP2 does not contain tests within the project directory so they cannot be run after installed. I think we can either add a few tests to psi4 with `@requires_sns_mp2` decorators and pytest or have the tests moved within the project folder so they can be run after installation. Any thoughts @rmcgibbo @loriab?

loriab:
Yes, I was confused by the testing in `sns-mp2`. When I tried to run `pytest` in a repo, it complained about `import garden`. And `python setup.py [py]test` wasn't happy, either.

On the other hand, the simple input worked just fine, and there are reference values in the two pytest input files, so I can just copy them over to become conditional psi test cases.

By the way, @rmcgibbo, I can get the version from `python setup.py --version` (`1.0.1.dev1`, I think). Is there a way to get that from the imported module (e.g., `python c "import snsmp2; print(snsmp2.__version__)"` or similar)?

loriab:
Also, @rmcgibbo, did you want to settle on `energy('sns-mp2')` or `energy('nnmp2')` for the call? I'm guessing the former, so changing w/i the p4 test suite. Latter appears [here](https://github.com/DEShawResearch/sns-mp2/blob/master/tests/2x%7BCC%7D.in#L30-L35). (Note that unless you're doing something I didn't notice, setting memory w/i `set ...` isn't valid; memory is a plain `memory 7 gb` statement.)

dgasmith:
Thanks, @shannonhouck. Great first commit, I knew going into it there would be a bit that we needed to clean up as this is a lot of firsts for us (see Lori's PR), but this certainly catalyzed quite a bit of action and thought behind the scenes. Let me know if you are up for more, we certainly have a list of things to do. #WWDC (?)

PR #856, opened 2017-11-13
address build issues for FreeBSD packaging

loriab:
## Description
address build issues for FreeBSD packaging @yurivict 

## Todos
* **Developer Interest**
  - [x] Killed off the Apple `environ` setting that wasn't being used anymore
  - [x] AddOns now have a `CMAKE_INSIST_FIND_PACKAGE_${AddON}` variable whereby you can fail configuration when the right package not detected. That is, no falling back to internal build

## Status
- [x] Ready to go

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,11 +42,12 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 #    - PYTHON_INCLUDE_DIR "Path to the python include files (e.g., /path/to/include/python2.7)"
 #    - SPHINX_ROOT "Root directory for Sphinx: 'bin/sphinx-build' (or similar) should be in this dir."
 #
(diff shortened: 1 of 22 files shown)


loriab:
On RDR's point 2,

* Looks like CMake has the following hash algorithms available `file(<MD5|SHA1|SHA224|SHA256|SHA384|SHA512> filename variable)`
* And ExternalProject has `URL_HASH ALGO=value` to specify them
* `conda install openssl -c conda-forge` followed by `openssl sha256 your_sdist.tar.gz` to compute them

Only problem is that I suspect that's a tarball (single file) procedure, not the GH hash we usually do.

robertodr:
Found this [Running FreeBSD in Travis-CI
](http://erouault.blogspot.com/2016/09/running-freebsd-in-travis-ci.html) by googling a bit around.

yurivict:
Thanks!

loriab:
In case we do ever switch to tarball+sha256 downloading in ExternalProject, here's a GH blogpost on accessing non-tagged versions. https://github.com/blog/651-annotated-downloads

PR #857, opened 2017-11-13
Fix to dump all rotation and polarizability values calculated to get_variables() dictionary

bgpeyton:
## Description
Originally, only the last rotation or polarizability values computed were actually dumped to the dictionary accessed by `core.get_variables`. This fix pushes each one into the dictionary with its corresponding wavelength (in NM). I've added a few lines to `tests/cc39/input.dat` and `tests/cc29/input.dat` to test the polarization and rotation variables, respectively. I've run the tests separately since the quick tests on Travis may not cover them, the results can be found [here](https://gist.github.com/bgpeyton/6a7842adb77a03a10bf696c8364a9c54) and [here](https://gist.github.com/bgpeyton/a6654acd0f7fcaa89b5023bebd40b025). 

## Todos
- [x] Make polarizabilities push each calculated value to the dictionary just like I did with rotation
- [x] Add to polarizability / rotation test cases to ensure new variables don't get overwritten by "accident"

## Questions
None

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/ccresponse/optrot.cc
+++ b/psi4/src/psi4/ccresponse/optrot.cc
@@ -70,8 +70,8 @@
 #include <cstdio>
 #include <cstring>
 #include <cstdlib>
-#include <cmath> /* needed for lround*/
-#include <sstream> /* needed for stringstream */
+#include <cmath>
+#include <sstream>
 
 #include "psi4/libpsi4util/process.h"
 #include "psi4/libciomr/libciomr.h"
--- a/psi4/src/psi4/ccresponse/optrot.cc
+++ b/psi4/src/psi4/ccresponse/optrot.cc
@@ -70,6 +70,8 @@
 #include <cstdio>
 #include <cstring>
 #include <cstdlib>
+#include <cmath> /* needed for lround*/
+#include <sstream> /* needed for stringstream */
 
 #include "psi4/libpsi4util/process.h"
 #include "psi4/libciomr/libciomr.h"
@@ -447,10 +449,21 @@ void optrot(std::shared_ptr<Molecule> molecule)
       outfile->Printf( "\n   Specific rotation using length-gauge electric-dipole Rosenfeld tensor.\n");
       outfile->Printf( "\t[alpha]_(%5.3f) = %10.5f deg/[dm (g/cm^3)]\n", params.omega[i], rotation_rl[i]);
 
-      if(params.wfn == "CC2")
-        Process::environment.globals["CC2 SPECIFIC ROTATION (LEN)"] = rotation_rl[i];
-      else if(params.wfn == "CCSD")
-        Process::environment.globals["CCSD SPECIFIC ROTATION (LEN)"] = rotation_rl[i];
+      /* grab omega in nm, rounded to nearest int */
+      long om_nm = std::lround((pc_c*pc_h*1e9)/(pc_hartree2J*params.omega[i]));
+
+      if(params.wfn == "CC2"){
+        /* build up a string */
+        std::stringstream tag;
+        tag << "CC2 SPECIFIC ROTATION (LEN) @ " << om_nm << "NM";
+        Process::environment.globals[tag.str()] = rotation_rl[i];
+      }
+      else if(params.wfn == "CCSD"){
+        /* build up a string */
+        std::stringstream tag;
+        tag << "CCSD SPECIFIC ROTATION (LEN) @ " << om_nm << "NM";
+        Process::environment.globals[tag.str()] = rotation_rl[i];
+      }
     }
 
     if(compute_pl) {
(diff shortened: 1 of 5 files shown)


psi4/src/psi4/ccresponse/optrot.cc
```diff
@@ -70,6 +70,8 @@
 #include <cstdio>
 #include <cstring>
 #include <cstdlib>
+#include <cmath> /* needed for lround*/
```
amjames: I think these comments should go. If someone is doing a header cleanup and they remove these two the file won't compile, so I don't think they are really necessary IMO.

bgpeyton: Removed the comments, thanks!

psi4/src/psi4/ccresponse/optrot.cc
```diff
@@ -447,10 +449,21 @@ void optrot(std::shared_ptr<Molecule> molecule)
...
 
-      if(params.wfn == "CC2")
-        Process::environment.globals["CC2 SPECIFIC ROTATION (LEN)"] = rotation_rl[i];
-      else if(params.wfn == "CCSD")
-        Process::environment.globals["CCSD SPECIFIC ROTATION (LEN)"] = rotation_rl[i];
+      /* grab omega in nm, rounded to nearest int */
+      long om_nm = std::lround((pc_c*pc_h*1e9)/(pc_hartree2J*params.omega[i]));
+
+      if(params.wfn == "CC2"){
+        /* build up a string */
```
loriab: keep `/* grab omega in nm, rounded to nearest int */`, drop the two `/* build up a string */`

same below for LEN & MVG

psi4/src/psi4/ccresponse/polar.cc
```diff
@@ -137,11 +139,21 @@ void polar(void)
...
+      tag << "CC2 DIPOLE POLARIZABILITY @ " << omega_nm_rd << "NM";
+      Process::environment.globals[tag.str()] = trace[i]/3.0;
+    }
+    else if(params.wfn == "CCSD"){
+      /* build up a string */
+      std::stringstream tag;
+      tag << "CCSD DIPOLE POLARIZABILITY @ " << omega_nm_rd << "NM";
+      Process::environment.globals[tag.str()] = trace[i]/3.0;
+    }
   }
```
loriab: For all of these, please make a comment section like https://github.com/psi4/psi4/blob/master/psi4/src/psi4/detci/opdm.cc#L644-L666 so the autodocs with catch them. Lowercase for variable (non-fixed) parts of the string. And note the nonstandard comment markers.

psi4/src/psi4/ccresponse/optrot.cc
```diff
@@ -474,10 +487,21 @@ void optrot(std::shared_ptr<Molecule> molecule)
...
+        std::stringstream tag;
+        tag << "CC2 SPECIFIC ROTATION (VEL) @ " << om_nm << "NM";
+        Process::environment.globals[tag.str()] = rotation_pl[i];
+      }
+      else if(params.wfn == "CCSD"){
+        /* build up a string */
+        std::stringstream tag;
+        tag << "CCSD SPECIFIC ROTATION (VEL) @ " << om_nm << "NM";
+        Process::environment.globals[tag.str()] = rotation_pl[i];
+      }
```
loriab: Please add all of these entries to [the glossary](https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/glossary_psivariables.rst), particularly including units and the fact that rounded to the nanometer.

psi4/src/psi4/ccresponse/optrot.cc
bgpeyton: Done

psi4/src/psi4/ccresponse/optrot.cc
bgpeyton: Done, let me know if I missed anything!

psi4/src/psi4/ccresponse/polar.cc
bgpeyton: Done, let me know if I missed anything!

bgpeyton:
I thought about the explicit checking of psivars. In fact, I figured that if anything would fail the test suite, it would be if there were tests which checked these (evidently there was not). I could possibly add one at a later date- it would help keep plugins which parse a dictionary created by `core.get_variables()` from randomly breaking.

loriab:
There aren't tests that check these because the whole `tests/cc*` got only sparsely ported to the new `compare_values()` testing scheme. Mostly cc tests are compared via a perl script against output.ref. I hope optical rotation is checked there. I'd strongly encourage throwing in a couple lines in one of the existing optrot and polarizability test cases to `compare_values()` against your new psivars.

bgpeyton:
Understood. I will add new test cases to this PR either tonight or tomorrow morning.

bgpeyton:
Added a few lines to tests `cc39` and `cc29` for checking the variables for polarization and rotation. I tested them myself, in case Travis doesn't use those tests, and put the outputs in the PR description. I hope everything is in order!

dgasmith:
@bgpeyton Can you `c-f` those files so we can get this in?

bgpeyton:
@dgasmith Yes, we're working on that today. There is an issue with getting `clang-format` on the VT-ARC systems that we mean to overcome this afternoon, sorry for the delay.

loriab:
We had a trial to get `c-f` at gatech, too. maybe this helps.

https://psi4.slack.com/archives/C5BLBG6VB/p1503008447000049

any poor souls trying to get `clang-format` on RHEL7 w/ broken `yum`, the magic llvm download is 3.9.0 (being early enough to have req. library abis (unlike 4.0.1) and late enough to recognize psi4's clang-format keywords(unlike 3.6.2)) for pkg Fedora23_x64_64. System `libstdc++` is still too old, but you can feed it any of psi4's gcc 5.2 via `LD_LIBRARY_PATH=${CONDA_PREFIX}/lib`

dgasmith:
Dont forget you can just `conda create -n cformat -c conda-forge clangdev`. Really recommend a venv for this as it installs a lot of things you may not want floating around.

bgpeyton:
Ran `clang-format` on both files. Thanks for the tip @dgasmith !

PR #861, opened 2017-11-15
[WIP] Pybind-ing the PCM object

robertodr:
## Description
This PR will expose the `PCM` object in Psi4 out to Python. The `PCM` object wraps the PCMSolver library and offers three methods:
1. `compute_E` to calculate the polarization energy due to the continuum;
2. `compute_E_electronic` to calculate the polarization energy due to the continuum, but only using the electron density;
3. `compute_V` to calculate the PCM potential, which is added on top of the Fock matrix.

## Todos
* **Developer Interest**
  - [ ] Clean-up the `PCM` object and corresponding sources:
    * Forward-declare as much stuff as possible instaed on `include`-ing
    * Remove raw pointers in favor of `std::vector`, `Vector` or `std::shared_ptr`
    * Use a [scoped `enum`](http://en.cppreference.com/w/cpp/language/enum) instead of an [unscoped `enum`](http://en.cppreference.com/w/cpp/language/enum)
  - [ ] Exposing the `PCM` object to Python.

## Questions
- The initialization of the `PCM` object requires a `BasisSet`. I am unsure what is the best way to grab that Python-side.
- Still regarding initialization, I think it can be done better. Meaning that after reading the user input, the `PCM` object can be initialized by passing the printlevel and the `BasisSet` _under the hood_. @loriab @dgasmith can you point me to existing examples in the codebase? I've looked around in `core.cc` but it's a bit daunting.
- I think this will be useful for the ongoing work on #847 
- @andysim and/or @lothian, the file [psi4/libmints/integral.h](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/integral.h#L487) has a comment `/// Want to change the name of this after the PCM dust settles`. Do you remember what was the intention there?

## Status
- [ ] Ready to go

### DIFF
--- a/psi4/src/export_pcm.cc
+++ b/psi4/src/export_pcm.cc
@@ -26,17 +26,28 @@
  * @END LICENSE
  */
 
-#include "psi4/libmints/potentialint.h"
+#include "psi4/pybind11.h"
 
-namespace psi {
+#include "psi4/libmints/basisset.h"
+#include "psi4/libmints/matrix.h"
+#include "psi4/liboptions/liboptions.h"
+#include "psi4/libpsipcm/psipcm.h"
 
-PCMPotentialInt::PCMPotentialInt(std::vector<SphericalTransform>& trans, std::shared_ptr<BasisSet> bs1,
-                                 std::shared_ptr<BasisSet> /* bs2 */, int /* deriv */)
-    : PotentialInt(trans, bs1, bs1) {
-    // We don't want to transform the integrals from Cartesian (6d, 10f, ...) to Pure (5d, 7f, ...)
-    // for each external charge.  It'll be better to backtransform the density / Fock matrices to
-    // the Cartesian basis, if necessary.
-    force_cartesian_ = true;
+using namespace psi;
+#ifdef USING_PCMSolver
+
+void export_pcm(py::module& m) {
+    py::class_<PCM, std::shared_ptr<PCM>> pcm(m, "PCM", "Class interfacing with PCMSolver");
+
+    py::enum_<PCM::CalcType>(pcm, "CalcType")
+        .value("Total", PCM::CalcType::Total)
+        .value("NucAndEle", PCM::CalcType::NucAndEle)
+        .value("EleOnly", PCM::CalcType::EleOnly);
+
+    pcm.def(py::init<int, std::shared_ptr<BasisSet>>())
+        .def("compute_E", &PCM::compute_E, "Compute PCM polarization energy", py::arg("D"), py::arg("type"))
+        .def("compute_V", &PCM::compute_V, "Compute PCM potential")
+        .def("compute_V_electronic", &PCM::compute_V_electronic,
+             "Compute PCM potential due to electronic charge density only");
 }
-
-}  // Namespace
+#endif
--- a/psi4/src/psi4/libpsipcm/psipcm.cc
+++ b/psi4/src/psi4/libpsipcm/psipcm.cc
@@ -27,417 +27,414 @@
  */
 #ifdef USING_PCMSolver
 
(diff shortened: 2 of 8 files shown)


psi4/src/psi4/libpsipcm/psipcm.cc
```diff
@@ -157,31 +155,21 @@ PCM::PCM(int print_level, std::shared_ptr<BasisSet> basisset) {
...
     // The charge and {x,y,z} coordinates (in bohr) for each tessera
     tess_Zxyz_ = std::make_shared<Matrix>("Tess Zxyz", ntess_, 4);
     double **ptess_Zxyz = tess_Zxyz_->pointer();
     // Set the tesserae's coordinates (note the loop bounds; this function is 1-based)
     for (int tess = 1; tess <= ntess_; ++tess) pcmsolver_get_center(context_.get(), tess, &(ptess_Zxyz[tess - 1][1]));
 
     // Compute the nuclear potentials at the tesserae
-    double **patom_Zxyz = atom_Zxyz_->pointer();
-    ::memset(tess_pot_n_, 0, ntess_ * sizeof(double));
+    std::fill(tess_pot_n_, tess_pot_n_ + ntess_, 0.0);
```
dgasmith: Or better yet, use `SharedVector` or `std::vector` over raw pointers.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -74,25 +74,19 @@
 #include <omp.h>
```
dgasmith: Is `c-f`ing this a good idea at the moment due to the large changes caused by #847? Same with the `libmints` folder, there are several PR's making deep changes there.

robertodr: No, it's not. I'll de-format before un-WIP-ping.

psi4/src/psi4/libpsipcm/psipcm.cc
robertodr: It's not really a problem (I think) since we're adhering to RAII here and the lifetime of these resources is tied to constructor and destructor. I would like to transition to `Vector`, however, because it can keep track of the symmetry blocking. I need to figure out how to get the symmetry information in and out of PCMSolver first, though. I have it working for DALTON, but the symmetry handling is slightly different between the two codes.

dgasmith: It is not really a problem, but your `memset` was perfectly valid as well. I just noted that you modernized one part of the code, but left the underlying data structure unchanged when it would probably be about the same amount of work to just update the underlying data model.

psi4/src/export_pcm.cc
```diff
@@ -0,0 +1,53 @@
...
+using namespace psi;
+#ifdef USING_PCMSolver
+
+void export_pcm(py::module& m) {
+    py::class_<PCM, std::shared_ptr<PCM>> pcm(m, "PCM", "Class interfacing with PCMSolver");
+
+    py::enum_<PCM::CalcType>(pcm, "CalcType")
+        .value("Total", PCM::CalcType::Total)
+        .value("NucAndEle", PCM::CalcType::NucAndEle)
+        .value("EleOnly", PCM::CalcType::EleOnly);
```
loriab: pb11 just accepted a PR that lets you add docstrings to enums that get autodoc-ed. May want to throw a comment out to the side that can someday become a docstring when we catch up.

jturney: That's a nice feature.

robertodr:
As per Slack discussions, closing this one as SCF iterations are moving to Python. I'll open a new PR targeting the `scfitertopy` branch.

PR #864, opened 2017-11-19
CUBEPROP: Compute adaptive isocontour range for cube files

fevangelista:
## Description
This is an enhancement to Psi4 cubeprop library. Cube files are usually plotted using isocontour surfaces with a common iso value. This PR implements the computation of adaptive isocontour values that capture a certain fraction of a MO density (orbital squared) using the least amount of points. This is similar to the approach of Lehtola and Jónsson [J. Chem. Theory Comput. 10, 642–649 (2014)], but appears to be different in the fact that the implementation in this PR uses two isocontour values.

The added code just sorts a cube file's grid points and finds a pair of positive and negative isocontour values the satisfy the above definition. The default is to capture 85% of the density, but the user can specify a different value with the new option `CUBEPROP_ISOCONTOUR_THRESHOLD`.  The adaptive isocontour range is included in the second comment line of the MO cube files and looks like this
```
Psi4 Gaussian Cube File.
Property: Psi_a_1_1-Ag. Isocontour range for 85% of the density: (  0.053504,  0.000000)
```
The user can then inspect the cube file to find the value of the range. However, the recommended way to use this feature is to use an updated version of `vmd_cube` that will be posted shortly and that will automatically plot all cube files using the adaptive ranges saved in the cube files.

## Todos
* **Developer Interest**
  - [x] Add ability to compute range for densities
  - [x] Add documentation of feature
  - [x] Add documentation of new options

## Questions
- [x] Merge the changes done in #822 with this PR?
## Status
- [x] Ready to go
- [x] Closes #822

### DIFF
--- a/psi4/driver/p4util/util.py
+++ b/psi4/driver/p4util/util.py
@@ -367,9 +367,14 @@ def compare_cubes(expected, computed, label):
     Performs a system exit on failure. Used in input files in the test suite.
 
     """
-    # Skip the first six elemets which are just labels
-    evec = [float(k) for k in expected.split()[6:]]
-    cvec = [float(k) for k in computed.split()[6:]]
+    # Grab grid points and skip the first two lines which are just labels
+    evec = []
+    for line in expected.split("\n")[2:]:
+        evec.extend([float(k) for k in line.split()])
+    cvec = []
+    for line in computed.split("\n")[2:]:
+        cvec.extend([float(k) for k in line.split()])
+
     if len(evec) == len(cvec):
         for n in range(len(evec)):
             if (math.fabs(evec[n]-cvec[n]) > 1.0e-4):
--- a/psi4/driver/p4util/util.py
+++ b/psi4/driver/p4util/util.py
@@ -367,21 +367,15 @@ def compare_cubes(expected, computed, label):
     Performs a system exit on failure. Used in input files in the test suite.
 
     """
-    # Grab grid points and skip the first two lines which are just labels
-    evec = []
-    for line in expected.split("\n")[2:]:
-        evec.extend([float(k) for k in line.split()])
-    cvec = []
-    for line in computed.split("\n")[2:]:
-        cvec.extend([float(k) for k in line.split()])
-
-    if len(evec) == len(cvec):
-        for n in range(len(evec)):
-            if (math.fabs(evec[n]-cvec[n]) > 1.0e-4):
-                message = ("\t%s: computed cube file does not match expected cube file." % label)
-                raise TestComparisonError(message)
+    # Grab grid points. Skip the first nine lines and the last one
+    evec = np.genfromtxt(expected,skip_header=9,skip_footer=1)
+    cvec = np.genfromtxt(computed,skip_header=9,skip_footer=1)
+    if evec.size == cvec.size:
+        if not np.allclose(cvec, evec, rtol=5e-02, atol=1e-10):
+            message = ("\t%s: computed cube file does not match expected cube file." % label)
+            raise TestComparisonError(message)
     else:
-        message = ("\t%s: computed cube file does not match expected cube file." % (label, computed, expected))
+        message = ("\t%s: computed cube file does not match size of expected cube file." % label)
         raise TestComparisonError(message)
     success(label)
     return True
--- a/psi4/src/psi4/libcubeprop/csg.cc
+++ b/psi4/src/psi4/libcubeprop/csg.cc
@@ -228,14 +228,15 @@ void CubicScalarGrid::print_header() {
 
     primary_->print();
 }
-void CubicScalarGrid::write_gen_file(double* v, const std::string& name, const std::string& type) {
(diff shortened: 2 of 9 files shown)


psi4/driver/p4util/util.py
```diff
@@ -367,9 +367,14 @@ def compare_cubes(expected, computed, label):
...
-    evec = [float(k) for k in expected.split()[6:]]
-    cvec = [float(k) for k in computed.split()[6:]]
+    # Grab grid points and skip the first two lines which are just labels
+    evec = []
+    for line in expected.split("\n")[2:]:
+        evec.extend([float(k) for k in line.split()])
+    cvec = []
+    for line in computed.split("\n")[2:]:
+        cvec.extend([float(k) for k in line.split()])
+
```
dgasmith: You might want to use [np.genfromtxt](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.genfromtxt.html) and `np.allclose` instead of using pure python. This would be more canonical NumPy/Psi4 usage here.

psi4/src/psi4/libcubeprop/csg.cc
```diff
@@ -228,14 +228,15 @@ void CubicScalarGrid::print_header() {
 
     primary_->print();
 }
-void CubicScalarGrid::write_gen_file(double* v, const std::string& name, const std::string& type) {
+void CubicScalarGrid::write_gen_file(double* v, const std::string& name, const std::string& type,
+                                     const std::string& comment) {
```
dgasmith: Does this interact with #822, @carlosborca?

psi4/src/psi4/libcubeprop/csg.cc
```diff
@@ -649,4 +664,36 @@ void CubicScalarGrid::compute_ELF(std::shared_ptr<Matrix> D, const std::string&
...
+        double value = p.second;
+        if (value >= 0.0) {
+            positive_isocontour = value;
+        } else {
+            negative_isocontour = value;
+        }
+        sum += p.first / sum_weight;
+        if (sum > cumulative_threshold) break;
+    }
+    return std::make_pair(positive_isocontour, negative_isocontour);
```
dgasmith: Oh C++17!

psi4/src/psi4/libcubeprop/csg.cc
```diff
@@ -649,4 +664,36 @@ void CubicScalarGrid::compute_ELF(std::shared_ptr<Matrix> D, const std::string&
     write_gen_file(v, name, type);
     delete[] v;
 }
+std::pair<double, double> CubicScalarGrid::compute_isocontour_range(double* v2, double exponent) {
```
dgasmith: Not sure where this is called from, but would it make more sense to pass in a managed object so that it can be called Python side?

fevangelista: Let me think about this one.

psi4/driver/p4util/util.py
fevangelista: Thanks, it's a very good idea.

psi4/src/psi4/libcubeprop/csg.cc
fevangelista: This function is used internally and I think it would be difficult to interface it to Python. What we really need is a Cubefile object with py-side interface (for another PR). That would allow us to find the isocontour values directly from Python and play with different values without having to rerun Psi4. However, I think the current default density threshold (85%) is pretty good. BTW, the reason I am not doing all of this in Python is because by the time we write all the data to a cube file it takes about 2 seconds/cube to re-read the data even with  np.genfromtxt.

fevangelista:
The last changes address two comments from @dgasmith. I was able to port the changes in #822 so I think that that PR is now obsolete. It should be ready to go once it passes.

dgasmith:
Cool, LGTM!

psi4/src/read_options.cc
```diff
@@ -226,6 +226,8 @@ int read_options(const std::string &name, Options & options, bool suppress_print
   /*- List of basis function indices for which cube files are generated
   (1-based). All basis functions computed if empty.-*/
   options.add("CUBEPROP_BASIS_FUNCTIONS", new ArrayType());
+  /* Fraction of density captured by adaptive isocontour values */
+  options.add_double("CUBEPROP_ISOCONTOUR_THRESHOLD",0.85);
```
loriab: I feel like we should be standardizing these names. There's precedent for `_PERCENTAGE` (though no consensus on 0-->1, 0-->100). I actually like `_FRACTION` better, but there's no precedent for that. There are no `_THRESHOLD`. I'm fine with it as-is, but if either of the former appeal, go for it (sticking with 0--1, imo).

fevangelista: I like `_FRACTION`, will fix when free this week.

dgasmith:
@psi4/maintainers This looks good to go, could use one more review.

PR #865, opened 2017-11-22
sns-mp2, the build

loriab:
## Description
Continues from psi4/psi4#850 in getting sns-mp2 integrated into psi as an endorsed plugin. Also various infrastructure additions to accommodate downstream and python projects.

## Todos
* **Developer Interest**
  - [x] First downstream project buildable during superbuild
  - [x] First pure python project buildable during superbuild. That is, `setup.py`, not `addonConfig.cmake`, so opt-in/detect-from machinery a lot shakier. Have to use FindPythonModule adapted from internet, rather than sturdy `find_package()`. For where-module?, relying totally on :envvar:PYTHONPATH. Seems to be behaving sensibly, but report oddities.
  - [x] Add some dev docs for how to do workflows with addons
  - [x] Adapt some snsmp2 test to work with psi ctest and pytest
  - [x] Fix up tests/pytest/test_numpy_views so that it runs with pytest and doesn't interfere when `memory_profiler` not installed
  - [x] ~Until https://github.com/DEShawResearch/sns-mp2/pull/2 goes in, pulling from my fork~
  - [x] Adds some (mildly tested) versioning capabilities to FindPythonModule
  - [x] Enable [sns-mp2](https://github.com/DEShawResearch/sns-mp2) to be built at build-time via `cmake -DENABLE_snsmp2=ON`. Can also detect it if you put an external snsmp2 in PYTHONPATH (but as it's an optional runtime dependency, latter's only really useful for testing)
  - [x] Fix nasty bug that made the number of points computed for cube file dependent on rounding error. Could show up as differences in cube files btwn compilers. Tightened down cubefile checking.

## Questions
- [x] @rmcgibbo, I'd still like to know about getting the version from the imported plugin. Also, it'd be nice to get pytest in the sns-mp2 repo working in its own right.
- [x] @shannonhouck, there'll probably be a few more docs to add, if you don't mind.

## Status
- [x] Ready for review
- [x] Ready for merge


cmake/FindPythonModule.cmake
```diff
@@ -0,0 +1,35 @@
...
+    if(ARGC GREATER 1 AND ARGV1 STREQUAL "REQUIRED")
+        set(${module}_FIND_REQUIRED TRUE)
+    endif()
+    # A module's location is usually a directory, but for binary modules
+    # it's a .so file.
+    execute_process(COMMAND "${PYTHON_EXECUTABLE}" "-c" 
+                            "import re, sys; \
+                             sys.path.insert(0, '${STAGED_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}${PYMOD_INSTALL_LIBDIR}'); \
+                             import ${module}; \
+                             print(re.compile('/__init__.py.*').sub('', ${module}.__file__))"
```
dgasmith: Cool! Shouldn't this be `sys.path.insert(1,...)`. Granted this could be an off case due to how this is being run, but worth pondering + a small note why it is `0`.

doc/sphinxman/source/build_planning.rst
```diff
@@ -1447,3 +1447,53 @@ Ways to refer to |PSIfour| in text, in order of decreasing goodness:
...
+
+If you're changing the |PSIfour| repo codebase between compiles, there's
+nothing more to do as CMake will handle the code rebuild deps for you.
+
+If you're changing the local Add-On repo codebase between compiles,
+CMake *does not* know when ``libaddon.a/so/dylib`` needs rebuilding. It
+is recommended that the |PSIfour| build be initially configured with
+``-DBUILD_SHARED_LIBS=ON`` (easier to notice changes). And to trigger
+Add-On library rebuild, ``rm -rf {objdir}/external/upstream/addon/``
+and ``rm -rf {objdir}/stage/{prefix}/share/cmake/AddOn``. This should
```
dgasmith: Would it be possible to add this to our CMake/make `make clean add-on`?

doc/sphinxman/source/build_planning.rst
```diff
@@ -1447,3 +1447,53 @@ Ways to refer to |PSIfour| in text, in order of decreasing goodness:
...
+re-clone the Add-On, rebuild and install it, rebuild any parts of
+|PSIfour| that interface to it, and relink the main ``core.so``.
+If you're modifying the Add-On's file or directory structure, be
+smart and ``rm`` all traces of it within ``{objdir}/stage/{prefix}``,
+especially any ``*.pyc`` files.
```
dgasmith: Ditto here `make clean python`.

external/downstream/snsmp2/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+            GIT_REPOSITORY https://github.com/loriab/sns-mp2
+            GIT_TAG psifixes
+            CONFIGURE_COMMAND ""
+            UPDATE_COMMAND ""
+            BUILD_COMMAND ${PYTHON_EXECUTABLE} setup.py build
+            BUILD_IN_SOURCE 1
+            INSTALL_COMMAND ${PYTHON_EXECUTABLE} setup.py install
+                                                          --record=record.txt
+                                                          --single-version-externally-managed
+                                                          --install-lib=${STAGED_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}${PYMOD_INSTALL_LIBDIR})
```
dgasmith: Cool!

tests/pytest/test_np_views.py
```diff
@@ -25,48 +28,105 @@ def check_leak(func, tol=1.e6):
...
+
+    # Double check totals
+    diff = abs(start - snapshot_memory())
+    if diff > 1.e6:
+        raise MemoryError("\nA function leaked %d bytes of memory!" % diff)
+    else:
+        print("\nNo leaks detected!")
+
+
 if __name__ == "__main__":
```
dgasmith: You can just nuke the main now thats its properly wrapped.

tests/snsmp2/cc-cc/input.dat
```diff
@@ -0,0 +1,41 @@
...
+H 1.70129500 4.27523200 -0.74662600
+H 0.79282300 3.82975100 0.73173200
+
+symmetry c1
+no_com
+no_reorient
+}
+
+
+energy('sns-mp2')
```
dgasmith: We are getting more of these methods where the basis set is fixed (`HF-3C`). Do we want to throw some kind of warning/error if a basis is specified?

doc/sphinxman/source/build_planning.rst
loriab: Do we have _any_ `make clean` capabilities? I've never trusted them enough to experiment.

Those would be nice, but, in any case, these particular commands are hitting the install folders or the written-upon-cmake-configure folders, so they're not a usual use of `make clean`. Even devs should be hitting this FAQ rarely.

doc/sphinxman/source/build_planning.rst
loriab: For this and previous, yes, looks like we can define external commands that will forcibly clean up whatever. They will be whatever names we choose, though, and not have any relation to actual `make clean` (in fact, cmake isn't big on "clean" b/c it's only really defined for GnuMake, not Ninja, etc. https://cmake.org/pipermail/cmake/2016-December/064695.html)

I tend to prefer the direct `rm` approach, since at this detail of dev one should know where files are going. But I'll certainly entertain seconds to the proposal.

dgasmith: I like the rm approach as well. I worry a bit that while nice, this could get lost in the noise of features if we do implement it. However, it is handy on those off cases in which you do need it.

tests/pytest/test_np_views.py
loriab: Nuke the `if __name__ == "__main__":`? Then you can't run it as a script.

Or nuke the `def test_totals()`? That seemed to have an extra "end minus beginning" check.

tests/snsmp2/cc-cc/input.dat
loriab: Probably a good idea. This is where I wish we had logging, so could emit a warning event.

tests/pytest/test_np_views.py
dgasmith: Oh, I would say nuke the entire main. But if you want it as a script, just have it call the function instead of duplicating?

cmake/FindPythonModule.cmake
loriab: Normally, yes, `insert(1`. I was having such problems with it detecting the clone on my system that I threw the hammer at it with `0`. 

The whole detecting python mods setup is flimsy and w/o many of the tweaking tools we have for `projectConfig.cmake` projects. We may need to substantially develop the machinery. For instance, there's legitimate reason a user would want his own location in `sys.path[0]`. But then, even if we delicately tuned detection, we can't do the same to runtime (don't want to set the user's `PYTHONPATH`), just as `core.so`'s `RPATH` is crude compared to all the `_DIR` cmake vars.

So, long story short, I don't know what's best here in the long term. But for now, I'll clobber with `0` and add a note. :-)

tests/snsmp2/cc-cc/input.dat
loriab: I want to reason through this a little more, first. Technically, `energy(cbs, ...)` should hit this, too. But it could just be left over from a previous calc. Should one raise like below? Just putting this on hold.

```
if mtd in ['sns-mp2', 'hf-3c', 'pbeh-3c'] and not IKnowWhatImDoing:
    raise YoureProbablyDoingItWrong("Method ({}) specified with basis ({}) does not use a basis")
```

dgasmith: I like that personally. Could help alleviate some confusion.

rmcgibbo:
@loriab I've merged your PR into DEShawResearch/sns-mp2, and updates the testing code to work better outside of desres internal infrastructure (no `import garden)`.

loriab:
Thanks, @rmcgibbo. I'll be away for a while, but should be able to get this finished off next week. Any prospects for version-from-import? (It could be there already – I'm not as proficient in setup.py as I ought.)

rmcgibbo:
@loriab: done

loriab:
Provided the CI passes, this is reasonably ready for final review. I think further docs can be in a follow-up PR. This one should actually get the docs building.

Warnings for basis specified for non-basis-changing mtds still not implemented.

tests/pytest/test_np_views.py
```diff
@@ -25,48 +28,86 @@ def check_leak(func, tol=1.e6):
...
+
+
+@using_memory_profiler
+def test_build_copy_mat():
+    assert(check_leak(build_copy_mat))
+
+
+@using_memory_profiler
+def test_totals():
     start = snapshot_memory()
```
andysim: Didn't know about `memory_profiler` before.  Neat!

loriab: Me, neither, nor the `mpmath` for multiprecision math that DGAS used for gau2grid.

dgasmith: The wonder of Python, I have yet to be left wanting for a library that did not involve quantum mechanics in some way.

PR #866, opened 2017-11-23
Enable v2rdm_casscf build from psi4

loriab:
## Description
Enable v2rdm_casscf build from psi4. Add tests. This gets Eugene's renovated 0.5 plugin that now has gradients working again with psi4.

Suppresses the plugin "module loaded" text since that command isn't called directly anymore and the text was interfering with CMake by corrupting the cache.

## Status
- [x] Ready to go


loriab:
Not quite so great as it looks on the testing front, as I'm copying tests from one repo to another. Properly, I wanted to shift psi's testing so that each native ctest dir has a one-line mirror run through pytest. Then pytest can automatically pick up other projects' (also pytest) tests from their install locations. Thus no test duplication, tests stay synched, and all testing available for _installed_ software, too.

But that seemed like a deep rabbit hole, so just copied the tests :-) Maybe at least I can keep v2rdm in working order now.

PR #876, opened 2017-12-06
Adding force_soccpi and force_doccpi functions

shannonhouck:
## Description
This PR adds implementation of `force_soccpi` and `force_doccpi`, allowing the user to change the singly and doubly occupied orbitals per irrep in the Wavefunction object. The alpha and beta electron counts are also updated accordingly. This functionality should only be used by expert users in special circumstances, since it causes inconsistency in the Wavefunction object for SCF purposes. (Let me know if I should add some sort of "expert" flag in the documentation to reflect this, or if I need to add any other documentation besides what I've put in already!)

## Todos
  - [x] Add implementation of `force_soccpi` and `force_doccpi` C++-side
  - [x] Add Python-side accessibility to `force_soccpi` and `force_doccpi`

## Status
- [ ] Ready to go

### DIFF
--- a/psi4/src/psi4/libmints/wavefunction.cc
+++ b/psi4/src/psi4/libmints/wavefunction.cc
@@ -609,8 +609,8 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
 
 void Wavefunction::force_doccpi(const Dimension &doccpi) {
     for (int h = 0; h < nirrep_; h++) {
-        if(doccpi[h] > nmopi_[h]) {
-            throw PSIEXCEPTION("Wavefunction::force_doccpi: Number of doubly occupied orbitals in an irrep cannot exceed the number of molecular orbitals.");
+        if((soccpi_[h] + doccpi[h]) > nmopi_[h]) {
+            throw PSIEXCEPTION("Wavefunction::force_doccpi: Number of doubly and singly occupied orbitals in an irrep cannot exceed the total number of molecular orbitals.");
         }
         doccpi_[h] = doccpi[h];
         nalphapi_[h] = doccpi_[h] + soccpi_[h];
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -360,6 +360,16 @@ class Wavefunction : public std::enable_shared_from_this<Wavefunction>
 
     void set_doccpi(const Dimension& doccpi);
     void set_soccpi(const Dimension& soccpi);
```
loriab: Erase `set_doccpi` & `set_soccpi`, I think, as they're never used as-is (and have no implementation?) and are replaced by your new fns.

shannonhouck: Done! (You're right-- it seems like they didn't have any implementation.) Thank you!

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -607,6 +607,26 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
...
+    }
+    nalpha_ = doccpi_.sum() + soccpi_.sum();
+    nbeta_ = doccpi_.sum();
+}
+
+void Wavefunction::force_soccpi(const Dimension &soccpi) {
+    for (int h = 0; h < nirrep_; h++) {
+        soccpi_[h] = soccpi[h];
+        nalphapi_[h] = doccpi_[h] + soccpi_[h];
+        nbetapi_[h] = doccpi_[h];
```
dgasmith: Slight pedantic, but no need to change `beta` here.

shannonhouck: I've removed it. Thank you!

dgasmith:
Looks good, thanks for the changes.

PR #877, opened 2017-12-06
additional XC superfunctionals

hokru:
## Description
Defines couple new XC functionals in the driver and adds few D3BJ parameters

## Todos
  - [x] Double-Hybrids: B2GPPLYP, PWPB95, PTPSS, DSD-BLYP
  - [x] hybrid-GGAs: PW6B95
  - [x] mega-GGAs: TPSS, revTPSS
  - [x] GGAs: mPWPW
  - [x] tests
  - [x] adding of some reference outputs
  - [x] clean-up 
  - [x] praying libxc 4.0 switch does not break the tweaking options.

* **Developer notes/issues**
  - [x] tests put into dft-custom-gga/mgga/hybrid/dhdf
   supersedes a number of older tests (removed only from CMakeLists.txt):
  -`dft-dsd + dft-pbe0-2 `moved into `dft-custom-dhdf`
  -`dft-dldf` moved into `dft-custom-mgga`
  - [x] new tests for mgga_ms0/1/2 based on libxc test data.
  - incomplete dft-custom-hybrid test: missing wpbe, wpbe0, b5050lyp

## Status
- [x] Ready to go


loriab:
Thanks! Yes, a single test file will be fine (unless multiple is more convenient for you). Just make a dir in `/tests/`, copy in a `CMakeLists.txt` from another test and modify at least the name. Then add your new dir name to the long list in `/tests/CMakeLists.txt`. Model on any of the dft tests and feel free to check in output files from other programs into the test directory for future consultation. If you're trying to do anything that isn't obvious, just post. Are you on the slack channel?

dgasmith:
I think what we need to do is organize the non-XC custom functionals into a series of tests like `tests/test-custom-gga` where all of the custom functionals that we build in `gga_superfuncs.py` should be there. Similar folders for hybrid/double hybrid. Right now its kind of scattered all over the place. If you could start your tests in that manner it would be very helpful. This could be a good project for a new-dev to clean up.

@susilehtola One more example why we need access to parameters in LibXC.

dgasmith:
Also @hokru you might want to check out the `tests/libxc` folder for a list of tests and see if your new functionals are present. We need to hook this up to long tests at some point as well.

@bwb314 At some point you added many DFT LibXC tests against Q-Chem that missed the rebasing and was finally closed [here](https://github.com/dgasmith/psi4/pull/18). Do you still have these tests that we can add to Psi4?

susilehtola:
@dgasmith bad example, since these should go in libxc proper ;)

hokru:
@susilehtola @dgasmith They should maybe be included, yes, but exposing functional parameters to the python layer is a _dream_ for functional development in general!

susilehtola:
@hokru as if the world needs more functionals :P

psi4/driver/procrouting/dft_funcs/gga_superfuncs.py
```diff
@@ -217,6 +217,81 @@ def build_pw91_superfunctional(name, npoints, deriv, restricted):
...
+    expo_pw6=3.72 # paper values
+
+    alpha_pw6=c_pw/X2S/X2S 
+    a_pw6=6.0*bt/X2S
+    b_pw6=1.0/X2S
+    c_pw6=bt/(X_FACTOR_C*X2S*X2S)
+    d_pw6=-(bt-beta)/(X_FACTOR_C*X2S*X2S)
+    f_pw6=1.0e-6/(X_FACTOR_C*X2S**expo_pw6)
+    pw6.set_tweak([a_pw6,b_pw6, c_pw6, d_pw6, f_pw6, alpha_pw6, expo_pw6])
+    sup.add_x_functional(pw6)
```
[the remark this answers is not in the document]
hokru: yes, but this custom mPWPW was basically just a test to understand what i need to do to modified pw91 for pwpb95. It will be removed. Further below is the mPWPW build using exactly what you suggest.

hokru:
> I think what we need to do is organize the non-XC custom functionals into a series of tests like tests/test-custom-gga where all of the custom functionals that we build in gga_superfuncs.py should be there.  

I added tests this way. I still want to add some reference outputs.
 
> Are you on the slack channel?

I am not, I also don't see it mentioned anywhere? I can join if that helps.

psi4/driver/procrouting/dft_funcs/mgga_superfuncs.py
```diff
@@ -161,6 +161,52 @@ def build_mgga_ms2_superfunctional(name, npoints, deriv, restricted):
...
+    sup = core.SuperFunctional.blank()
+    sup.set_max_points(npoints)
+    sup.set_deriv(deriv)
+
+    # => User-Customization <= #
+
+    # No spaces, keep it short and according to convention
+    sup.set_name('tpss')
+    sup.set_description('    TPSS Meta-GGA XC Functional\n')
+    sup.set_citation('   J. Tao, J. P. Perdew, V. N. Staroverov, and G. E. Scuseria Phys. Rev. Lett. 91 146401, 2003   \n')
```
dgasmith: Comma missing between vol/page and extra lines at end.

psi4/driver/procrouting/dft_funcs/mgga_superfuncs.py
```diff
@@ -161,6 +161,52 @@ def build_mgga_ms2_superfunctional(name, npoints, deriv, restricted):
...
+
+    # Call this first
+    sup = core.SuperFunctional.blank()
+    sup.set_max_points(npoints)
+    sup.set_deriv(deriv)
+
+    # => User-Customization <= #
+
+    # No spaces, keep it short and according to convention
+    sup.set_name('tpss')
```
dgasmith: Names should be upper case to match other (applies to all new funcs). Very surprised this doesn't come from LibXC, looks like they only give TPSSh.

tests/dft-custom-gga/input.dat
```diff
@@ -0,0 +1,97 @@
...
+set basis cc-pvdz
+set reference rks
+set dft_radial_points 99
+set dft_spherical_points 770 
+set scf_type direct
+
+for func in ggalist[:]:
+  edft = energy(func)
+  label=func.upper()
+  val=float(ggaval[func])
```
dgasmith: Can be replaced with:
```python
for func, val in ggaval.items():
    edft = energy(func)
   ...
```

The `ggalist` can be removed.

psi4/driver/procrouting/dft_funcs/mgga_superfuncs.py
susilehtola: Well, that's one of the open questions about libxc standards that hasn't really been addressed. I think opinions vary, but so do functionals. TPSSh could follow other ab initio (non-fitted) functionals like most LDAs and GGAs and some meta-GGAs and have a separate exchange part and correlation part.

psi4/driver/procrouting/dft_funcs/double_hyb_superfuncs.py
```diff
@@ -84,24 +83,25 @@ def build_dsd_blyp_superfunctional(name, npoints, deriv, restricted):
...
 
     # Set GKS up after adding functionals
     sup.set_x_omega(0.0)
     sup.set_c_omega(0.0)
-    sup.set_x_alpha(0.71)
+    sup.set_x_alpha(0.70)
     sup.set_c_alpha(1.0)
     sup.set_c_os_alpha(0.46)
-    sup.set_c_ss_alpha(0.43)
+    sup.set_c_ss_alpha(0.40)
```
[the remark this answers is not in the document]
hokru: dsd-blyp was commented out. So i guess something was up.

hokru:
I added your requested changes @dgasmith, including running yapf. I did all my planned changes. Pending further review this is ready to go, i think.

psi4/driver/procrouting/dft_funcs/double_hyb_superfuncs.py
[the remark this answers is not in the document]
dgasmith: Ah ok, good to know it was never live :)

dgasmith:
@psi4/maintainers Lets get a few more eyes on this.

psi4/driver/procrouting/dft_funcs/double_hyb_superfuncs.py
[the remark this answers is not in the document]
loriab: IIRC, the SCS DH plus variants were alive c. beta3. Chris Cook had charge of them, as he was trying out the different -D parameterizations. Then SCS DH were disabled for years after DH were reformed. If there was any uncertainty, esp. in -D parameters, I? didn't uncomment them after the libxc reform. Glad the dust has settled on some of these DSD variations, and thanks for sorting them out, @hokru.

tests/dft-custom-hybrid/input.dat
```diff
@@ -0,0 +1,61 @@
...
+# PBE0
+##Gaussian 09: 5D SP Int(Grid=99770)
+# wb97x
+
+# MISSING: 
+#          "wpbe"     : build_wpbe_superfunctional,
+#          "wpbe0"    : build_wpbe0_superfunctional,
+#          "b5050lyp" : build_b5050lyp_superfunctional,
+
+psi4.core.set_output_file("output.dat", False)
```
loriab: Are these output file resets leftovers or still doing something?

tests/dft-custom-mgga/input.dat
```diff
@@ -0,0 +1,111 @@
...
+set dft_spherical_points 302
+
+edft = energy('dldf+d')
+val = -153.80439592484987    #TEST
+compare_values(val, edft, 8, "dlDF+D (2010 +D correction)")  #TEST
+clean()
+
+edft = energy('dldf+d09')
+val = -153.8046388583318276   #TEST
+compare_values(val, edft, 8, "dlDF+D09 (2009 +D correction)")  #TEST
```
loriab: Are ref `val` coming from psi runs of this input file? If so, the ref value only good to ~6 decimal places (default e_conv for energy), so checking to 8 won't be reliable. If this is the case, need to generate the `val` with e_/d_conv 10 or so, save that to `val`, then set e_/d_conv to 8 in this file, and continue checking to 8.

hokru: All dldf related lines are copy & paste from the individual test (dft-dldf). Unknown to me when the ref values were generated.

tests/dft-custom-hybrid/input.dat
hokru: I took it initially from the libxc test. Wasn't sure if i needed to use it. Just delete them?

tests/dft-custom-mgga/input.dat
loriab: Ok, then we'll leave the ref as-is. May have to turn down the number of decimal places in test case in future.

tests/dft-custom-hybrid/input.dat
loriab: Yes, delete them. That libxc is a psiapi test masquerading as psithon. Yours are normal psithon.

loriab:
@dgasmith and @bwb314, the libxc test case _looks_ like the one mentioned https://github.com/psi4/psi4/pull/877#issuecomment-349732113 . Is it, or does that still need to be added?

tests/dft-custom-hybrid/input.dat
hokru: ah ok! Will do so.

dgasmith:
@loriab They are a different kind of test, there may be some overlap however.

PR #880, opened 2017-12-12
Added new tests for libXC

bwb314:
## Description
This PR is designed to add more tests for libXC in Psi4.

## Todos
- [x] output.ref files need to be generated
- [x] CMakeLists.txt needs to added
- [x] Test names need to be added to psi4/tests/CMakeLists.txt

## Status
- [x] Ready to go


tests/libxc/dft-water-dimer/input.dat
```diff
@@ -0,0 +1,132 @@
...
+monA = h2odimer.extract_subsets(1)                                      #TEST
+monB = h2odimer.extract_subsets(2)                                      #TEST 
+for func in data.keys():                                                #TEST
+    try: dimer_energy = energy(func, molecule = dimer)                  #TEST
+    except: continue                                                    #TEST
+    monoA_energy = energy(func, molecule = monA)                        #TEST
+    monoB_energy = energy(func, molecule = monB)                        #TEST
+    func_ie = 627.5095 * (dimer_energy - monoA_energy - monoB_energy)   #TEST
+    try: compare_values(data[func], func_ie, 3, func)                   #TEST
+    except: continue                                                    #TEST
```
loriab: `except: continue` looks like a Get Out of Jail Free line. is that really wanted?

tests/libxc/dft-water-dimer/input.dat
```diff
@@ -0,0 +1,132 @@
...
+
+dimer = h2odimer.extract_subsets([1,2])                                 #TEST
+monA = h2odimer.extract_subsets(1)                                      #TEST
+monB = h2odimer.extract_subsets(2)                                      #TEST 
+for func in data.keys():                                                #TEST
+    try: dimer_energy = energy(func, molecule = dimer)                  #TEST
+    except: continue                                                    #TEST
+    monoA_energy = energy(func, molecule = monA)                        #TEST
+    monoB_energy = energy(func, molecule = monB)                        #TEST
+    func_ie = 627.5095 * (dimer_energy - monoA_energy - monoB_energy)   #TEST
```
loriab: psi_bohr2angstroms

tests/libxc/dft-anion/input.dat
```diff
@@ -0,0 +1,121 @@
+memory 1 Gb
```
loriab: nix all the `memory` unless it's really needed.

tests/libxc/dft-anion/input.dat
```diff
@@ -0,0 +1,121 @@
...
+                                           
+molecule h2o {                             
+   -1 2                                    
+   O                                       
+   H 1 1.0                                 
+   H 1 1.0 2 104.5                         
+   units angstrom                          
+   no_reorient
```
loriab: ang is default and `no_reorient` shouldn't actually be needed, right?

tests/libxc/dft-anion/input.dat
```diff
@@ -0,0 +1,121 @@
...
+   basis 6-31G                             
+   dft_radial_points 100                   
+   dft_spherical_points 302                
+   guess sad                               
+   d_convergence 11                        
+}                                          
+
+data = {                                   #TEST
+      'BMK':  -76.1817717171999931,        #TEST
+ 'MGGA_MS1':  -76.2292773247000071,        #TEST
```
loriab: origin of all these nice references? at least program, preferably in/out files and options, too.

tests/libxc/dft-water-dimer/input.dat
andysim: [Try again!](https://github.com/psi4/psi4/blob/master/psi4/include/psi4/physconst.h#L77) (can tell you've been working with molecule class lately)

loriab: ack, good catch. yeah, it's `psi_` in qcdb which took its patterns from molecule and didn't change with the times. (indeed, too much of molecule in all languages – py, c++, json.)

andysim: Oh yeah, the prefix is also wrong.  I was referring to the fact that you typed out the length conversion factor, not the energy conversion.  I surmise you've had to type that one out quite a bit lately, so it's become a reflex.  Much like how I can't type "multiple" without first typing "multipole."

loriab: wow, I totally missed that aspect. they both fall under "the important one with a 5 and a 9", so must be the same, surely.

andysim: Good point.  As long as it's not `NaN`, what's a few orders of magnitude among friends?

tests/libxc/dft-anion/input.dat
dgasmith: DFT is not completely invariant to its orientation, wouldn't mind having explicit coordinates in case things shift.

tests/libxc/dft-anion/input.dat
```diff
@@ -0,0 +1,121 @@
...
+  'MN12-SX':  -76.1521811698000022,        #TEST
+  'wB97X-D':  -76.1994640198000042,        #TEST
+  'MPW1LYP':  -76.1938387614000021,        #TEST
+    'PBE50':  -76.1459052103000005,        #TEST
+ 'MGGA_MS0':  -76.2427928287000043,        #TEST
+      'N12':  -76.1636902505999984         #TEST
+}                                          #TEST
+                                           
+for func in data.keys():                                   #TEST
+    try: func_energy = energy(func)                        #TEST
```
dgasmith: The original try/except here was functionals that we did not implement. For actual tests I would very much prefer just commenting out the functionals that we do not have as it serve as a list of functional would should look at implementing and ensure that we cannot regress in functionality without noticing.

tests/libxc/dft-water-dimer/input.dat
dgasmith: Any reason not to use `energy(func, molecule=dimer, bsse_type='nocp')` here?

loriab:
aha, here's where the psi vs libxc vs qchem tests went – knew I had seen them somewhere. this is a simple PR and shouldn't need much work to get it ready to merge when you have the time, @bwb314.

tests/libxc/dft-anion/input.dat
```diff
@@ -0,0 +1,121 @@
+memory 1 Gb
+                                           
+molecule h2o {                             
+   -1 2                                    
+   O                                       
+   H 1 1.0                                 
+   H 1 1.0 2 104.5
```
dgasmith: This should XYZ with no_com and no_reorient NOT z-matrix due to the variance in DFT with respect to orientation.

tests/libxc/dft-anion/input.dat
```diff
@@ -0,0 +1,121 @@
...
+   no_reorient                             
+}                                          
+                                           
+set {                                      
+   reference uhf                           
+   basis 6-31G                             
+   dft_radial_points 100                   
+   dft_spherical_points 302                
+   guess sad                               
+   d_convergence 11
```
dgasmith: `e_convergence` and `d_convergence` should both be set to 9 (a reasonable cutoff) radial and spherical should be bumped to 200/590 for testing.

tests/libxc/dft-anion/input.dat
dgasmith: Please see `tests/libxc` for a reasonable way of checking these cutoffs. The test must actually fail if a certain threshold is crossed.

dgasmith:
This is quite important to have before the next release. If you will not have a chance to work on this, please let us know and we can open a general issue.

bwb314:
The only real problem with this test (that I'm aware of) is compare_values stops the test if psi4 is not within the given threshold (10^-4 for qchem and 10^-6  for psi4 without libXC). Are these acceptable thresholds?

tests/libxc/dft-water/input.dat
```diff
@@ -0,0 +1,152 @@
...
+for func in qchem_data.keys():                                #TEST
+    psi4_energy = energy(func, molecule=h2o)                  #TEST
+                                                              #TEST 
+    qchem_energy = qchem_data[func]                           #TEST
+                                                              #TEST 
+    if func in old_psi4_data:                                 #TEST
+        old_psi4_energy = old_psi4_data[func]                 #TEST
+        compare_values(old_psi4_energy, psi4_energy, 6, func) #TEST
+                                                              #TEST
+    compare_values(qchem_energy, psi4_energy, 4, func)        #TEST
```
loriab: could be a bit tighter. perhaps ...

```
for func in qchem_data:
    psi4_energy = energy(func, molecule=h2o)

    compare_values(qchem_data[func], psi4_energy, 4, func + ': Psi4 vs. Q-Chem')

    if func in old_psi4_data:
        compare_values(old_psi4_data[func], psi4_energy, 6, func + ': Psi4 vs. Psi4 1.0 (pre-Libxc)')
```

tests/libxc/dft-water/input.dat
```diff
@@ -0,0 +1,152 @@
...
+    'B97-3':    -76.365757473,        #TEST
+ 'MGGA_MS1':   -76.3863611957,        #TEST
+      'M11':   -76.3548246737,        #TEST
+   'PW6B95':   -76.4636801397,        #TEST
+   'B3LYP5':   -76.3479565264,        #TEST
+    'X3LYP':   -76.3550695373,        #TEST
+     'BLYP':   -76.3649513839,        #TEST
+   'HCTH93':   -76.3690959506,        #TEST
+   'M06-HF':   -76.3438580561         #TEST
+#   'PBEOP':   -76.3113060897,        #TEST
```
loriab: throw in a note about why these commented out – libxc doesn't have them? or psi can't match them?

tests/libxc/dft-water/input.dat
```diff
@@ -0,0 +1,152 @@
...
+#  'PBEsol':   -76.0376457772,        #TEST
+#   'SCAN0':   -76.3393392953,        #TEST
+#'MGGA_MVS':   -76.3432912834,        #TEST
+#'MGGA_MVSh':  -76.3375136868,        #TEST
+#  'wM05-D':    -76.355280694,        #TEST
+#  't-HCTH':   -76.3695334542,        #TEST
+# 'BP86VWN':   -76.3897099046,        #TEST
+}                                     #TEST
+
+old_psi4_data = {                     #TEST
```
loriab: throw in a comment about the version. If it's the one I sent you, its ~April 2017 and you can get a sha from the output file.

dgasmith:
Can you add the water dimer and excitation energy tests back in? Meta-GGA UKS should be fixed in latest master.

tests/libxc/dft-water/input.dat
```diff
@@ -0,0 +1,156 @@
...
+
+for func in qchem_data.keys():                                                                        #TEST
+    psi4_energy = energy(func, molecule=h2o)                                                          #TEST
+                                                                                                      #TEST 
+    qchem_energy = qchem_data[func]                                                                   #TEST
+                                                                                                      #TEST 
+    compare_values(qchem_energy, psi4_energy, 4, func + ': Psi4 vs. Q-Chem')                          #TEST
+                                                                                                      #TEST
+    if func in old_psi4_data:                                                                         #TEST
+            old_psi4_energy = old_psi4_data[func]                                                     #TEST
```
dgasmith: Watch your formatting, this should only be indented 4 spaces.

tests/libxc/dft-water/input.dat
```diff
@@ -0,0 +1,156 @@
...
+  'B97-D': -76.34490642131334,        #TEST
+    'M05':  -76.3495817363072,        #TEST
+  'B3LYP': -76.38509296980465,        #TEST
+'HCTH407': -76.37515280900291,        #TEST
+'HCTH147': -76.37958046443578,        #TEST
+  'SOGGA': -75.97737386477169,        #TEST
+  'B97-2': -76.35924563701468         #TEST
+}                                     #TEST
+
+for func in qchem_data.keys():                                                                        #TEST
```
dgasmith: Is `qchem_data` a formal superset of psi4 data?

bwb314: Yes, all functionals `old_psi4_data` are in `qchem_data`. However, note that `SOGGA`, which is in `old_psi4_data`, has been commented out in `qchem_data` since it is not currently implemented in Psi4.

tests/libxc/dft-water/input.dat
bwb314: Good catch!

tests/libxc/dft-water/input.dat
PeterKraus: I've tried implementing `SOGGA`, however there's a problem:

> Water test:
> SOGGA Psi4 dev:       -75.9775599
> SOGGA Psi4 1.0:        -75.9773739
> SOGGA Q-Chem:       -75.9775523

My implementation comfortably passes against Q-Chem, but doesn't pass the tight criterium compared to old Psi4.

dgasmith:
Hmm, more inclined to trust qchem and libxc here. The old Psi4 probably handled singularities in a different way. I think I’m ok with these differences. 

-Daniel Smith

bwb314:
It appears that the wB97X-D3 is currently being built incorrectly. According to Psi4's print out when running wB97X-D3, wB97X is used and -D3 is appended: 
```   => Exchange-Correlation Functionals <=

   1.0000   wB97X range-separated functional

  => Exact (HF) Exchange <=

   0.8423            HF,LR [omega = 0.3000]
   0.1577               HF
```

However, there should be 19.57% SR HF and 100% LR HF. Also, omega should be 0.25.

hokru:
> It appears that the wB97X-D3 is currently being built incorrectly

Yes. Part of the problem will be solved by https://github.com/psi4/psi4/issues/920.

The other problem is that the current dashparam-dftd3 interface does not support setting a custom rs8 parameter for D3ZERO (rs8 is 1.0 except in Chai's wB97X-D3 re-fit where s8 is 1.0 and rs8 is fitted).

Until above is solved the d3 dict entry should probably be commented out.

loriab:
Oh! That we can fix.

My psi is a wreck at present. @bwb314, you want to do as @hokru advises and change [this](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/dashparam.py#L265) to `dashcoeff['rs8']`, add `rs8: 1.0` to everything [here](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/dashparam.py#L84-L129) and add `rs8: ??` to wB97X [here](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/dashparam.py#L130)? Then run tests.

PeterKraus:
PR #922 should address the `DSD-xxx`-class of functionals. I have also modified the `dashparams.py` in that PR, however I plan to expand the dashparams even further (to include references as well and to automatically generate the functionals listed). 

It would be really good to get these tests merged. My wish would be to have reference values from Psi4 1.1 (or even 1.2a1.dev781 - which is the latest conda release, if we trust them) in addition to Q-Chem, to make the comparison with PR #922 easier.

dgasmith:
@bwb314 Any reason not to go ahead and merge this in?

loriab:
Looking good. Still need the TODOs in your original post.

bwb314:
Alright, pending any suggested changes, I think this PR is ready to merge.

loriab:
ok, @bwb314, concurrence on killing off the `libxc/` subfolder.

loriab:
ok to lose `devl/cpks.py` and the like, @dgasmith?

tests/dft-cation/input.dat
```diff
@@ -0,0 +1,193 @@
...
+# Current version of Psi4 does not match Q-Chem for these tests
+expected_fail_qchem = ['wB97X-D3', 'O3LYP'] #TEST
+
+# These tests fail to converge with the current version of Psi4
+excluded_qchem_convergence_probs = {
+'CAM-B3LYP':     0.4568003603999955,        #TEST
+}
+
+# These tests are expected to fail with the current version of Psi4
+# due to lack of UKS support
```
susilehtola: Wait, hasn't this been solved already? VV10 works the same way with unrestricted densities - it just operates on the total density!

psi4/driver/qcdb/dashparam.py
```diff
@@ -81,53 +81,53 @@
...
-        'pwpb95'      : {'s6': 0.82, 's8': 0.705, 'sr6': 1.557, 'alpha6': 14.0},
-        'revpbe0'     : {'s6': 1.0,  's8': 0.792, 'sr6': 0.949, 'alpha6': 14.0},
-        'revpbe38'    : {'s6': 1.0,  's8': 0.862, 'sr6': 1.021, 'alpha6': 14.0},
-        'rpw86pbe'    : {'s6': 1.0,  's8': 0.901, 'sr6': 1.224, 'alpha6': 14.0},
-        'wb97x'       : {'s6': 1.0,  's8': 1.094, 'sr6': 1.281, 'alpha6': 14.0},
+        'hf'          : {'s6': 1.0,  's8': 1.746, 'sr6': 1.158, 'rs8': 1.000, 'alpha6': 14.0},  # in psi4  #IN
+        'b1b95'       : {'s6': 1.0,  's8': 1.868, 'sr6': 1.613, 'rs8': 1.000, 'alpha6': 14.0},
+        'b2gpplyp'    : {'s6': 0.56, 's8': 0.760, 'sr6': 1.586, 'rs8': 1.000, 'alpha6': 14.0},
+        'b3lyp'       : {'s6': 1.0,  's8': 1.703, 'sr6': 1.261, 'rs8': 1.000, 'alpha6': 14.0},  # in psi4  #IN
+        'b97-d'       : {'s6': 1.0,  's8': 0.909, 'sr6': 0.892, 'rs8': 1.000, 'alpha6': 14.0},  # in psi4  #IN
```
andysim: Just out of interest, what does `#IN` signify?

loriab: The big dict once upon a time was every param set on Grimme's website or in dftd3 itself and `# in psi4` was a tag that psi had actually implemented the parent fctl. I suspect `#IN` was yet another pass at that. All those tags can go, as @PeterKraus and @hokru have examined these rather more closely now.

tests/dft-cation/input.dat
dgasmith: There is some sort of missing factor of 2 running around. The energies are correct, but the Fock contributions are wrong. These tests should help narrow it down. Feel free to play with it.

PR #881, opened 2017-12-13
Get useful info back after optimizer errors

loriab:
## Description
Allow access to wfn (incl. geom) after optking errors

## Todos
* **Developer Interest**
  - [x] If optking fails internally or through exceeding max cycles and throws a `ConvergenceError`, you can now catch it with:

```
try:
    optimize('hf')
except psi4.ConvergenceError as err:
    gotcha = err.wfn
    print(gotcha.molecule().nuclear_repulsion_energy())
```

## Status
- [x] Ready to go

### DIFF
--- a/psi4/driver/p4util/exceptions.py
+++ b/psi4/driver/p4util/exceptions.py
@@ -81,10 +81,11 @@ class ConvergenceError(PsiException):
     error message *msg* to standard output stream and output file.
 
     """
-    def __init__(self, eqn_description, maxit):
+    def __init__(self, eqn_description, maxit, wfn=None):
         msg = "Could not converge %s in %d iterations." % (eqn_description, maxit)
         PsiException.__init__(self, msg)
         self.message = msg
+        self.wfn = wfn
         core.print_out('\nPsiException: %s\n\n' % (msg))
 
 
--- a/psi4/driver/p4util/exceptions.py
+++ b/psi4/driver/p4util/exceptions.py
@@ -81,12 +81,19 @@ class ConvergenceError(PsiException):
     error message *msg* to standard output stream and output file.
 
     """
-    def __init__(self, eqn_description, maxit, wfn=None):
+    def __init__(self, eqn_description, maxit):
         msg = "Could not converge %s in %d iterations." % (eqn_description, maxit)
         PsiException.__init__(self, msg)
         self.message = msg
-        self.wfn = wfn
         core.print_out('\nPsiException: %s\n\n' % (msg))
+
+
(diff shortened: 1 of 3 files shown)


psi4/driver/p4util/exceptions.py
```diff
@@ -81,10 +81,11 @@ class ConvergenceError(PsiException):
     error message *msg* to standard output stream and output file.
 
     """
-    def __init__(self, eqn_description, maxit):
+    def __init__(self, eqn_description, maxit, wfn=None):
```
dgasmith: Hmm, is convergence error used elsewhere? Having this as an optional arg could be dangerous when folks are always expecting for this exception to contain a Wfn.

loriab: I thought it was only used in py-side SCF iterations, so that's a problem for another branch.

Turns out it is used by `diatomic`; that one's off the hook b/c there's no `core.driver` calls involved, so the user has no reasonable expectation of wfn-returnedness.

Also used in `frac`. There, there's simply too many wfns around to single out one to return.

So I think it's ok, subject to fixing in scfitertopy. Other thoughts?

dgasmith: Hmm, could we separate out iterative and geometry convergence errors? Could help folks catch the right errors, but subclassing Exceptions can go too far.

loriab: Sure, though I don't think iterative vs geometry convergence errors will be the separator in the end
```
ConvergenceError (no wfn)
-- OptimizationConvergenceError (wfn) (though keep in mind this could be a dinky wfn if it's from cbs() for example)
-- SCFConvergenceError(wfn) (presently in another branch)
-- -- OscillatingSCFConvergenceError(wfn) (some of these may be internal-only for the smart SCF driver)
-- CCConvergenceError(wfn) (should cc* shift its language boundary)
-- FracConvergenceError(no wfn) (still don't know what to return)
etc.
```

I'm all for a more complete error set now that we can do interesting things with them. Presently still in minimal-working-state exception set.

Would you rather I subclassed out `GeometryConvergenceError` now, so `ConvergenceError` could stay wfn-less? That's fine by me.

dgasmith: What about something simpler:
```
ConvergenceError (no wfn)
-- OptimizationConvergenceError (wfn)
-- MethodConvergenceError(wfn) (name?)
```

I dont think we want to go too deep into the individual convergence issues until there is a real need?

dgasmith:
Cool, thanks for the update. Big fan of this.

Issue #860, opened 2017-11-14
Non-Deterministic OPTKING Steps with Dihedral 0


JonathonMisiewicz:
Under certain conditions, OPTKING's steps are non-deterministic. A dihedral of 0 seems necessary, but it is not clear to me if this is sufficient.

I've only found this in the IRC code, though I haven't checked for this in the non-IRC code. The below three files use the exact same input file but all have different steps. (This is on my own ircfix branch. I had conv_check return false on the first iteration as described [here](https://github.com/psi4/psi4/issues/859) and hardcoded the filename for my external hessian, to speed up my tests. The hessian itself was generated at SCF with this geometry.)

[output.txt](https://github.com/psi4/psi4/files/1472907/output.txt)
[output1.txt](https://github.com/psi4/psi4/files/1472908/output1.txt)
[output2.txt](https://github.com/psi4/psi4/files/1472909/output2.txt)

The below cases do not show the non-deterministic behavior. The only change in input file from the previous runs to this run was that the dihedral of 0 has been changed to a dihedral of 2.

[output.txt](https://github.com/psi4/psi4/files/1472911/output.txt)
[output1.txt](https://github.com/psi4/psi4/files/1472912/output1.txt)

psi-rking:
I don't know how you are reading in the Hessian here, since I don't know how to automatically read files like "hooh.21221.hess" now that we have pid's in the name.   (BTW, it appears the current psi4 test cases actually use an empirical Hessian.)

These jobs are failing because the "Point 1" on the IRC is being judged 'converged' immediately before a gradient is even run.  (The logic in the code is messy, since the optimizer doesn't actually call the gradient.)  So optking needs to do displacements from the TS (where the gradient is zero) but return a not-converged signal to the python optimize.  This usually works.  I tried the following:

```
set {
  g_convergence gau_verytight
  basis dzp
  opt_type                   irc
  cart_hess_read true
  hessian_write true
}

hessian('scf', dertype=1)
h2o2.reset_point_group('c2')
energy = optimize('scf')

```
and everything worked fine.  The problem in my testing is not indeterminism.  There is something about the calculation setup that is causing optking not to compute the gradient after the first irc_step().  Perhaps @avcopan could figure out what is going on?

JonathonMisiewicz:
[...]

Issue #859, opened 2017-11-14
IRC Converges to Initial TS


JonathonMisiewicz:
Suppose we are performing an IRC starting from a transition state that is very tightly converged. An IRC will converge to that transition state, rather than the minima.

## Psi4 Internals Explanation

optking.cc::349 will call the irc_step function beginning opt_data.cc::98. Now proceed to line opt_data.cc::221. The lines below that will return true if the molecule is converged. Because the transition state was tightly converged, the function will return true. **Comparison with a non-bugged case shows this is unexpected behavior.**

converged is true and this is an IRC type optimization, so we enter optking.cc::392, the point_converged function. molecule_irc_step.cc::83 shows this causes an IRC step to be taken immediately - before the energy or the gradient have been updated from the last step. OPTKING then sees we are not at the first step but the gradient is zero and assumes we have truly converged. It terminates the IRC.

## Previous Fix Attempts

I attempted to fix this just by having the IRC return false for Step 0 of an IRC, but in some cases I found that it would converge on Step 1 instead. We'll need a smarter way to make sure we don't converge to a TS, possibly checking for negative Hessian eigenvalues. I would do this myself, but after some of the other Optking issues I've run into this, this is much less of a personal priority.

Hopefully, PyOptKing will automatically take care of this.

## Minimal Demonstration Files

Demonstrated for four atoms at the SCF level. Even including the Hessian calculation, this took less than two minutes on our cluster.

[input.txt](https://github.com/psi4/psi4/files/1471977/input.txt)
[output.txt](https://github.com/psi4/psi4/files/1471978/output.txt)

PR #882, opened 2017-12-15
#859, #860 Fix: Tightly Converged IRCs Now Run

JonathonMisiewicz:
## Description
As documented in #859, an IRC performed on a tightly converged transition state would converge to the initial transition state. The obvious workaround showed non-deterministic optking steps, as documented in #860. Both bugs originate from neglecting the possibility of near-zero gradients and have been fixed. Also as documented in #860, there is still some non-deterministic behavior after the minimum has been found. As optking is going to be moved Python-side soon and this behavior is irrelevant for most users, I'll hold off from investigating that until we see if the Python-version has the problem.

I have added a test case for tightly converged transition states, thanks to the oddly prescient #881. For the sake of other developers, I added a page on how to add test cases.

## Todos

  - [x] Fixes a bug disabling IRCs for tightly converged transition states.

## Status
- [x] Ready to go

### DIFF
--- a/doc/sphinxman/source/add_tests.rst
+++ b/doc/sphinxman/source/add_tests.rst
@@ -28,12 +28,12 @@
 
 .. include:: autodoc_abbr_options_c.rst
 
-.. _`sec:add_tests`:
+.. _`faq:add_tests`:
 
 Adding Test Cases
-=========
+=================
 
-To create a new test case, first make a folder in :source:`tests`. This directory will need two files. The first is ``CMakeLists.txt``, which is necessary to add the test case to the suite. This file should have the following lines::
+To create a new test case, first make a folder in :source:`tests`. The directory name may not contain an underscore; to indicate spaces, use a hyphen instead. This directory will need two files. The first is ``CMakeLists.txt``, which is necessary to add the test case to the suite. This file should have the following lines::
 
     include(TestingMacros)
     
--- a/doc/sphinxman/source/add_tests.rst
+++ b/doc/sphinxman/source/add_tests.rst
@@ -1,0 +1,79 @@
+.. #
+.. # @BEGIN LICENSE
+.. #
+.. # Psi4: an open-source quantum chemistry software package
+.. #
+.. # Copyright (c) 2007-2017 The Psi4 Developers.
+.. #
+.. # The copyrights for code used from other parties are included in
+.. # the corresponding files.
+.. #
+.. # This file is part of Psi4.
+.. #
+.. # Psi4 is free software; you can redistribute it and/or modify
+.. # it under the terms of the GNU Lesser General Public License as published by
+.. # the Free Software Foundation, version 3.
+.. #
+.. # Psi4 is distributed in the hope that it will be useful,
+.. # but WITHOUT ANY WARRANTY; without even the implied warranty of
+.. # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+.. # GNU Lesser General Public License for more details.
+.. #
+.. # You should have received a copy of the GNU Lesser General Public License along
+.. # with Psi4; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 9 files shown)


doc/sphinxman/source/add_tests.rst
```diff
@@ -0,0 +1,79 @@
...
+.. #
+.. # @END LICENSE
+.. #
+
+.. include:: autodoc_abbr_options_c.rst
+
+.. _`sec:add_tests`:
+
+Adding Test Cases
+=========
```
loriab: sphinx, not me, demands that the number of `=` underlining the section title equal or exceed the length of the section title.

doc/sphinxman/source/add_tests.rst
```diff
@@ -0,0 +1,79 @@
...
+
+To create a new test case, first make a folder in :source:`tests`. This directory will need two files. The first is ``CMakeLists.txt``, which is necessary to add the test case to the suite. This file should have the following lines::
+
```
loriab: maybe also mention that directory/testcase names must have `-`, not `_`

doc/sphinxman/source/add_tests.rst
```diff
@@ -0,0 +1,79 @@
...
+.. # You should have received a copy of the GNU Lesser General Public License along
+.. # with Psi4; if not, write to the Free Software Foundation, Inc.,
+.. # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+.. #
+.. # @END LICENSE
+.. #
+
+.. include:: autodoc_abbr_options_c.rst
+
+.. _`sec:add_tests`:
```
loriab: add a ref to this [here](https://raw.githubusercontent.com/psi4/psi4/master/doc/sphinxman/source/prog_faq.rst) so ppl have another chance to find it.

tests/opt-irc-3/input.dat
```diff
@@ -0,0 +1,44 @@
...
+}
+
+# g_convergence needs to be tight for IRCs:
+set g_convergence gau_verytight
+
+set {
+  basis                      sto-3g
+  opt_type                   irc
+  geom_maxiter               10
+  g_convergence              gau_verytight
```
loriab: redundant

tests/opt-irc-3/input.dat
```diff
@@ -0,0 +1,44 @@
...
+
+# An ideal test would check the energy at the minima, but there's
+# an unpredictable last step; see #860. Instead, check after 10 steps.
+# We just need to make sure this makes it past the second point.
+try:
+    energy = optimize('scf')
+except psi4.OptimizationConvergenceError as ex:
+    wfn = ex.wfn
+
+compare_values(-148.752629, wfn.energy(), 5, "Energy of IRC point after 10 iterations")  #TEST
```
loriab: just FYI, `wfn.energy()` isn't rolled out to all the modules. that is, if you run it on a ccsd wfn, you're likely to get the hf energy. `psi4.get_variable('CURRENT ENERGY')` is still the most thorough at present.

doc/sphinxman/source/add_tests.rst
JonathonMisiewicz: Weird, neither sphinxmini nor sphinxman gave me an error when I made the documentation. I shouldn't be surprised that sphinx is so cryptic.

loriab: it may be just a warning. once upon a time I had a clean sphinx build and so could notice these things. if you misalign a table cell delimiter character on their fancy tables (e.g., the energy docstring one in driver.py), it will error.

dgasmith:
This looks good, but lets get a review from @avcopan or @psi-rking before pulling this in.

tests/opt-irc-3/input.dat
dgasmith: I agree here, this is unfortunate and could be a good new-dev project.

avcopan:
Looks good to me -- and props to Jonathon for working with my undergrad C++ code.

PR #891, opened 2018-01-06
Update memory docs

TiborGY:
## Description
Default memory had been raised to 500 MiB, update documentation to match it.  Add some extra information about the handling of IEC/SI units. 

## Todos
  - [ ] Documentation about setting memory has been updated

## Status
- [x] Ready to go

### DIFF
--- a/doc/sphinxman/source/psithoninput.rst
+++ b/doc/sphinxman/source/psithoninput.rst
@@ -98,7 +98,7 @@ to |PSIfour|::
     
     memory 2000000 kB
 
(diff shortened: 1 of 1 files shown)


TiborGY:
As far as I can tell the memory minimum is currently set to 250 MiB, [in here](https://github.com/TiborGY/psi4/blob/master/psi4/driver/p4util/util.py).

loriab:
Yes, `set_memory` will reject anything under 250 MiB. But it calls `set_memory_bytes` under the hood, and that latter function will take any paltry amount.

PR #895, opened 2018-01-18
CdSalcList Pybinding, Cleanup

JonathonMisiewicz:
## Description
The CdSalcList code is made fully accessible to the Python layer, per #884. This code does that and makes a few other changes to clean up code I found while trying to write the documentation for Pybind.

Pinging @loriab and @dgasmith by request.

## Todos
* **Developer Interest**
  - [x] Moves the factory argument from the CdSalcList constructor to an argument of the method that uses it, getting rid of several null pointers and also meaning that a list of displacements no longer needs to care about orbital irreps.
  - [x] Adds missing `const` declarations
  - [x] The documentation no longer tells developers to use the non-existent SimpleMatrix class
  - [x] Other misc. cleanup, from removing an overloaded term to simplification
  - [x] Exposed `CdSalcList` to Python
  - [x] Made the needed irreps for Cartesian displacements more readable

## Questions
- [x] ~~I am playing with reworking the print function. needed_irreps displays as an integer, so that part of the output is unintelligible to people who do not know that it is internally a bitstring. Would this be worth changing?~~ Changed!

## Status
- [x] Ready to go!

### DIFF
--- a/psi4/src/psi4/libmints/pointgrp.cc
+++ b/psi4/src/psi4/libmints/pointgrp.cc
@@ -83,6 +83,7 @@
 #include "psi4/libpsi4util/PsiOutStream.h"
 #include "psi4/libpsi4util/exception.h"
 
+#include <bitset>
 #include <cstdlib>
 #include <cstring>
 #include <ctype.h>
@@ -370,6 +371,19 @@ PointGroup::print(std::string out) const
     printer->Printf("PointGroup: %s\n", symb.c_str());
 }
 
+std::string PointGroup::irrep_bits_to_string(int irrep_bits) const {
+    std::string irrep_str;
+    const CharacterTable c_table = char_table();
+    for (int irrep = 0; irrep < c_table.nirrep(); ++irrep) {
+        if ((1 << irrep) & irrep_bits) {
+            if (!irrep_str.empty()) {
+                irrep_str += ", ";
+            }
+            irrep_str += c_table.gamma(irrep).symbol();
+        }
+    }
+    return irrep_str;
+}
 }
 
 /////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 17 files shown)


andysim:
In terms of the print function, `"{0:b}".format(needed_irreps)` would give the binary representation.  Perhaps that's the cleanest.  If you go that route, you'd have to ensure that only the trailing `nirreps` characters are printing, which is easy to do using Python's slicing syntax.

JonathonMisiewicz:
@andysim I understand if the print code was in Python, but the plan was to keep the CdSalcList code in C++. Are you proposing I use Python code in the method of a C++ class? I have no idea how I would do that.

andysim:
Oops, thought the printing was to be done in Python.  To print the bits in C++ you can make a `bitset` from the integer and call the [to string](http://en.cppreference.com/w/cpp/utility/bitset/to_string) method.

loriab:
pybind11 in its wonder does [allow python printing in c++](http://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html), but I've never used it. In any case, we want to keep pb11 away from core classes and isolated in the export_* and core.cc files for header/compilation reasons.

JonathonMisiewicz:
And, ready to go. Many thanks to Jet for help with this last part.

psi4/src/psi4/libmints/pointgrp.cc
```diff
@@ -83,6 +83,7 @@
 #include "psi4/libpsi4util/PsiOutStream.h"
 #include "psi4/libpsi4util/exception.h"
 
+#include <bitset>
```
jturney: Did you remove the `bitset` from the new function? If so, please remove the header.

JonathonMisiewicz: Good catch. Fixed.

PR #897, opened 2018-01-21
UKS Meta-GGA fixes and UKS VV10

dgasmith:
## Description
Apparently UKS Meta-GGA's have been wrong since the LibXC patch. I have fixed this error and get quite a nice agreement across our test set. I have changed the `dft1` test to cover these cases and be a bit more specific in what we are testing.

I have also started on UKS VV10; however, I am not quite confident it is correct. The VV10 energy is certainly correct, but the gradient might not be scaled correctly. This is the last chance I have to look at it for several weeks so it will need to wait.

## Status
- [x] Ready to go


PeterKraus:
Dear @dgasmith: currently, optimisations with VV10, LC-VV10 and B97M-V fail due to the lack of the analytical gradients for VV10. This can be, of course, circumvented by passing `dertype="energy"` to the optimize command. However, it's not necessary to do this for wB97M-V and wB97M-X - they do numerical gradients automatically. I was wondering if it's possible to set the default for VV10 etc. in the meantime?

PR #902, opened 2018-01-27
getting the value of CC static polarizability properly 

ashutoshvt:
## Description
The CC2/CCSD static polarizability values i.e when omega = 0 ev or Inf nm can be properly accessed now. Ex for CC2: ``psi4.get_variable("CC2 DIPOLE POLARIZABILITY @ INF NM")``. Right now, the tag is erroneously set to ``"...POLARIZABILITY @ 0NM"`` for such cases. So, this is a very minor bug.
## Status
- [X] Ready to go


psi4/src/psi4/ccresponse/polar.cc
```diff
@@ -132,8 +132,12 @@ void polar(void) {
...
         outfile->Printf("\n\talpha_(%5.3f) = %20.12f a.u.\n", params.omega[i], trace[i] / 3.0);
 
-        /* make sure omega in nm is rounded */
-        long omega_nm_rd = std::lround(omega_nm);
+         std::string omega_nm_rd;
+         if (params.omega[i] == 0.0)
+            omega_nm_rd = "INF ";
+         else
+            /* make sure omega in nm is rounded */
+            omega_nm_rd = std::to_string(std::lround(omega_nm));
```
dgasmith: Can we get parenthesis around this if statement?

ashutoshvt: you mean braces?

dgasmith: Sorry yes, getting a little late for my poor brain.

ashutoshvt: No problem, formatting has never been my strength :(

Issue #863, opened 2017-11-17
SCAN/libXC segmentation fault


hokru:
This was already shortly discussed on the forum http://forum.psicode.org/t/libxc-density-functionals/589, possibly caused by libxc. No idea if this can be hot-fixed without upgrade to libxc4, but SCAN-D3(BJ) gave excellent results for the GMTKN55 so it would be really nice to have it.

Here is a test than runs scan and scan0: [scan.txt](https://github.com/psi4/psi4/files/1481932/scan.txt)

also tagging @dgasmith because he did already some checks on this.

dgasmith:
After chatting with the LibXC dev's I think we can do a few workarounds so that LibXC 4 works for us. If I get a second I might work on getting that upgrade through.

susilehtola:
Hm, any reason why Psi4 isn't asking libxc for the default exact exchange parameters?

susilehtola:
And yeah, scan correlation crashes because it's not implemented in libxc 3.0.0...

dgasmith:
@susilehtola Then why do the SCAN primitives exist in 3.0? See [here](https://github.com/loriab/libxc/blob/4c765cc3b9186e5b690d8da41f9a03d6293fba3d/src/mgga_c_scan.c) for a 3.0 clone.

susilehtola:
@dgasmith .. because they went ahead and published 3.0.0 before scan correlation had been filled in. See
https://gitlab.com/libxc/libxc/blob/3.0.0/src/mgga_c_scan.c#L314

dgasmith:
@susilehtola Ah, I see thank you. Is there a way to detect this in LibXC 4.0?

susilehtola:
Miguel implemented it after 3.0.0, but after that we switched over to machine generated code. Libxc 4 is probably better for you anyway.

susilehtola:
Detect what? Libxc 4 has SCAN.

dgasmith:
Detect non-operational functionals ahead of time.

susilehtola:
Not really. Libxc4 doesn't have anything that crashes.

I've backported the SCAN implementation to libxc 3 at
 https://gitlab.com/libxc/libxc/commit/96568b1dd36130df57d19f7037fab7afecde48a4
as it appears there's going to be a new release in that series soon.

hokru:
@susilehtola good news, thank you!

[...]

PR #904, opened 2018-01-29
Added some functionals

PeterKraus:
## Description
Added and tested the following functionals. For the last two, unfortunately no bench value is provided.
```
           Functional:            libxc :      rob    qchem
Comparison    SOGGA11:   -76.3720792686 :      nan 3.84e-05 
Comparison  SOGGA11-X:   -76.3569745367 :      nan 1.10e-05 
Comparison      M06-L:   -76.3766432198 :      nan 1.95e-06 
Comparison     MN12-L:   -76.3109178757 :      nan 3.84e-05 
Comparison     MN15-L:   -76.3001901669 :      nan 5.49e-06 
Comparison       MN15:   -76.3017521834 :      nan      nan 
Comparison    PBE0-DH:   -76.1613318950 :      nan      nan 
```

Added worker functions for SCAN family of functionals, however without XC_MGGA_C_SCAN (which is available in LibXC 3.0.1) they won't work - commented out.

See http://forum.psicode.org/t/libxc-density-functionals/589/


hokru:
Unfortunately, SCAN doesn't work with LibXC 3.0.1, either. https://github.com/psi4/psi4/issues/863

psi4/driver/procrouting/dft_funcs/hyb_superfuncs.py
```diff
@@ -239,7 +232,6 @@ def build_hf3c_superfunctional(name, npoints, deriv, restricted):
     sup.allocate()
     return (sup, ('HF3C', '-d3bj'))
 
-
```
dgasmith: There should be two spaces between these functions. Can you format your files with YAPF (`pip install yapf`)?

PeterKraus:
@dgasmith, is just the Ne test enough? I don't have ORCA, but I have access to G09.D01 (SOGGA, SOGGA11-X, M06-L, MN12-L) and a colleague of mine has G16 (MN15, MN15-L and PBE0-DH), that should be good enough.

tests/dft-custom-dhdf/input.dat
```diff
@@ -87,6 +93,13 @@ compare_values(-0.062529936, e_mp2, 5,  'Ne: PWPB95_PT2')   #TEST
 compare_values(-128.920706595202, edhdf, 5,  'Ne: PWPB95')   #TEST
 clean()
 
+edhdf = energy('pbe0-dh')
```
hokru: Does it fail already for the SCF part, or at the MP2 part?
Your G16 test has frozen-core is on for the MP2 part, which is actually recommended for DHDFs. But my test has it globally off. Maybe that is the reason the test fails? The RI-MP2 error should be small enough.

PeterKraus: Already the DFT reference energy fails. I am a bit at a loss as to why this is. I've checked the original paper and the handy reference table from Q-Chem [1] and I think I set it up correctly, but apparently not...

[1] http://www.q-chem.com/qchem-website/manual/qchem44_manual/sect-DFT.html

PeterKraus: Attached out file from the current version (frozen core set to true).
[dhdf.txt](https://github.com/psi4/psi4/files/1689537/dhdf.txt)

hokru: I used the cc-pVTZ basis set for the DHDF tests to reproduce values from a publication. The other tests used cc-pVDZ. My bad for the confusion, sorry.
PSI4 should be much closer with the correct basis ;-). If you check my PBE values in dft-custom-gga, you can see that only TM reproduces the PSI4 values to the 5th decimal. ORCA and Gaussian deviate a bit. Very sensitive to the constants used, i think.

This comparison with ORCA passes:
https://gist.github.com/hokru/fd97a2cc67404be4a8835925c06616a4

PeterKraus: OK I see. I need to learn how to read. The DSD-BLYP is also with FC=TRUE already, so I'll just lump them together.

PeterKraus:
This should be all.

psi4/driver/qcdb/dashparam.py
```diff
@@ -115,8 +115,10 @@
         'lcwpbe'      : {'s6': 1.0,  's8': 1.279, 'sr6': 1.355, 'alpha6': 14.0},
         'm05-2x'      : {'s6': 1.0,  's8': 0.00 , 'sr6': 1.417, 'alpha6': 14.0},  # in psi4  #IN
         'm05'         : {'s6': 1.0,  's8': 0.595, 'sr6': 1.373, 'alpha6': 14.0},  # in psi4  #IN
+        'm06-2x'      : {'s6': 1.0,  's8': 0.00 , 'sr6': 1.619, 'alpha6': 14.0},
         'm062x'       : {'s6': 1.0,  's8': 0.00 , 'sr6': 1.619, 'alpha6': 14.0},
         'm06hf'       : {'s6': 1.0,  's8': 0.00 , 'sr6': 1.446, 'alpha6': 14.0},
+        'm06-l'       : {'s6': 1.0,  's8': 0.00 , 'sr6': 1.581, 'alpha6': 14.0},
         'm06l'        : {'s6': 1.0,  's8': 0.00 , 'sr6': 1.581, 'alpha6': 14.0},
```
loriab: Looks like some duplicates `m06l` and `m062x` probably from when I copied over Grimme's parameter tables. Since these are now actually hooked up to fctls with `-` keys, go ahead and remove the duplicates.

tests/dft-custom-mgga/ref_outputs/g09_m06l.ref
```diff
@@ -0,0 +1,240 @@
...
+ L/cc-pvdz 5D SP int=(grid=99770)\\Ne\\0,1\Ne,0,0.,0.,0.\\Version=AS64L
+ -G09RevD.01\State=1-A1G\HF=-128.9031949\RMSD=4.624e-10\Dipole=0.,0.,0.
+ \Quadrupole=0.,0.,0.,0.,0.,0.\PG=OH [O(Ne1)]\\@
+
+
+ THE THINKER DIES, BUT HIS THOUGHTS ARE BEYOND THE 
+ REACH OF DESTRUCTION.  MEN ARE MORTAL; BUT IDEAS 
+ ARE IMMORTAL.
+                              -- WALTER LIPPMANN
+ Job cpu time:       0 days  0 hours  0 minutes 14.0 seconds.
```
loriab: This program can be sensitive. It'd be good to remove the "Job cpu time" lines from the output files. And you may want to zip these files. And if you're not the license holder, you may want to make sure the license holder is ok with posting the files.

tests/dft-custom-dhdf/input.dat
```diff
@@ -87,6 +78,23 @@ compare_values(-0.062529936, e_mp2, 5,  'Ne: PWPB95_PT2')   #TEST
...
+set freeze_core true 
+edhdf = energy('dsd-blyp')
+e_dft=get_variable('SCF TOTAL ENERGY')
+e_mp2=get_variable('DOUBLE-HYBRID CORRECTION ENERGY')
+compare_values(-128.75560404, e_dft, 6, 'Ne: DSD-BLYP_DFT')   #TEST
+compare_values(-0.126886573, e_mp2, 7,  'Ne: DSD-BLYP_PT2')   #TEST 
+compare_values(-128.882490613961, edhdf, 6,  'Ne: DSD-BLYP')   #TEST
+clean()
+
+edhdf = energy('pbe0-dh')
```
loriab: In definition above says `pbe0-dh` is disabled, and it is disabled in the superfuncs dict. Yet testing it here. Is there an inconsistency or how does this work?

Issue #907, opened 2018-02-01
SAPT0 printing


jgonthier:
The manual states that when NO_RESPONSE option is set, SAPT0 should print a warning that uncoupled induction is computed. There is no such warning and the final printing contains Ind20,r which seems to imply response.

I can probably do the necessary changes but I create the issue so it does not get forgotten.

Example input:

```
import psi4
psi4.set_output_file("output_ref.dat")
water_dimer = psi4.geometry("""
O  -1.551007  -0.114520   0.000000
H  -1.934259   0.762503   0.000000
H  -0.599677   0.040712   0.000000
--
0 1
O   1.350625   0.111469   0.000000
H   1.680398  -0.373741  -0.758561
H   1.680398  -0.373741   0.758561
units angstrom
no_com
no_reorient
symmetry c1
""")

psi4.set_options({"BASIS"  : "cc-pVDZ",
                 "SCF_TYPE": "DF",
                  "SAVE_JK" : True,
                  "NO_RESPONSE" : True,
})

psi4.energy('sapt0')
```

dgasmith:
@jgonthier I think #908 closes this right?

PR #908, opened 2018-02-02
Corrected printing for uncoupled induction

jgonthier:
## Description
When requesting uncoupled induction in RHF-based SAPT0, the final summary still printed Ind20,r instead of Ind20,u as the line title, and no warning was issued about the uncoupled induction stored in the Psi4 variable for coupled induction, contrary to what is happening for open-shell SAPT0.
Solves #907.

## Todos
  - [x] Correct printing of titles and warning for uncoupled induction in RHF-based SAPT0

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/libsapt_solver/sapt0.cc
+++ b/psi4/src/psi4/libsapt_solver/sapt0.cc
@@ -249,6 +249,7 @@ void SAPT0::print_results() {
             outfile->Printf("    with Alpha = %12.6f \n", exch_scale_alpha_);
         }
         std::string scaled = (scal_it != Xscal.begin() ? "sc." : "   ");
+        std::string coupled = (no_response_ ? "u" : "r")
         outfile->Printf(
             "  "
             "--------------------------------------------------------------------------------------------------------"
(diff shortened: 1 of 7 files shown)


psi4/src/psi4/libsapt_solver/sapt0.cc
```diff
@@ -265,13 +265,23 @@ void SAPT0::print_results() {
...
+        if(no_response_) {
+            outfile->Printf("      Ind20,u                 %16.8lf [mEh] %16.8lf [kcal/mol] %16.8lf [kJ/mol]\n",
+                            e_ind20_ * 1000.0, e_ind20_ * pc_hartree2kcalmol, e_ind20_ * pc_hartree2kJmol);
+            outfile->Printf("      Exch-Ind20,u %3s        %16.8lf [mEh] %16.8lf [kcal/mol] %16.8lf [kJ/mol]\n",
+                            scaled.c_str(), *scal_it * e_exch_ind20_ * 1000.0,
+                            *scal_it * e_exch_ind20_ * pc_hartree2kcalmol, *scal_it * e_exch_ind20_ * pc_hartree2kJmol);
+            outfile->Printf("      delta HF,u (2) %3s      %16.8lf [mEh] %16.8lf [kcal/mol] %16.8lf [kJ/mol]\n",
+                            scaled.c_str(), dHF2 * 1000.0, dHF2 * pc_hartree2kcalmol, dHF2 * pc_hartree2kJmol);
+        } else {
+            outfile->Printf("      Ind20,r                 %16.8lf [mEh] %16.8lf [kcal/mol] %16.8lf [kJ/mol]\n",
```
JonathonMisiewicz: The only change I see in the`else` block close compared to the code in the `if` block code is r vs. u. Why not turn r vs. u into an argument of `Printf`, just like `scaled.c_str()`?

loriab:
Thanks very much, Jerome.  I was worried about psivar consequences. I think I'd rather make a different set for when induction is uncoupled, just so that the definition of SAPT0 doesn't get broadened. `NO_RESPONSE` is only for SAPT0, right, not higher SAPT levels? I'll make a PR to your PR.

jgonthier:
`NO_RESPONSE` is only for SAPT0, that's right. The open-shell SAPT0 uses a different implementation and a different variable (I think it's `COUPLED_INDUCTION`) that already issues the proper warning, so I thought it would be more consistent to have closed-shell SAPT0 do the same.

Currently, when induction is uncoupled, the induction energy gets put in both an UNCOUPLED and a COUPLED induction Psivar, with a warning in the output file. Would you rather have us not set the COUPLED variables ? We could then delete the warning.

dgasmith:
I'd rather just not set the coupled variables if we do not compute them. I guess the argument the other way is folks might have a script expecting these values either in the printing or psivars. I still think not setting the coupled variables is a bit cleaner overall.

psi4/src/psi4/libsapt_solver/sapt0.cc
jgonthier: Sure, I added this in there.

jgonthier:
Ok, since everyone seems to agree (and I also think it makes sense), I corrected the code to now only store uncoupled induction values in uncoupled induction Psivars.

jgonthier:
@dgasmith : Do you mean set up empty `Ind20,u` variable in case it's not computed ? As it is, the closed-shell SAPT0 code either computes coupled or uncoupled induction.

dgasmith:
Ah, thats right only FISAPT and SAPT(DFT) build tensors that can be convolved with the amplitudes easily. It is fairly straightforward to do, but probably more work than the scope of this. Please ignore my comments in this case.

jgonthier:
Well, I needed to add a commit for making the code compile, so I also made sure the `Ind20,u` Psivars are filled up whenever Ind20,u is available (i.e. always except in the closed-shell SAPT0 code). Also, I added all the energy components Psivars in the FISAPT code. Once the tests are done I'll commit.

jgonthier:
I noticed that sapt8 test case is broken (no idea if it's because of me or not). It has 0.0 stored in `SAPT IND ENERGY` Psivar instead of the actual uncoupled value. I tried to modify sapt_util.py to make it compute the uncoupled induction energy for `SAPT IND ENERGY`but that did not fix the test case. I don't see where else `SAPT IND ENERGY` is set. Are the uncoupled energy values even available to sapt_util.py ?

dgasmith:
In `sapt_jk_terms.py` we build the following variables:

```python
   ret["Ind20,u (A<-B)"] = unc_ind_ab
    ret["Ind20,u (A->B)"] = unc_ind_ba
    ret["Ind20,u"] = unc_ind_ab + unc_ind_ba
    ret["Exch-Ind20,u (A<-B)"] = unc_indexch_ab
    ret["Exch-Ind20,u (A->B)"] = unc_indexch_ba
    ret["Exch-Ind20,u"] = unc_indexch_ba + unc_indexch_ab
```

There are also the associated `r` terms as well.

Hmm, SAPT8 passes on my laptop with current master.

jgonthier:
I confirmed that sapt8 got broken because we stopped storing `Ind20,u` in the `Ind20,r` variable in usapt0.cc. I am not sure where the `SAPT IND ENERGY` variable that is used for the test is defined. Does open-shell SAPT0 even go through `sapt_jk_terms.py` ? It seems like the code flow changed a bit since I implemented it.

jgonthier:
Ok, the problem arises because in `psivardefs.py` the `SAPT0 IND ENERGY` variable is always defined in terms of coupled induction. I remember we had discussions about that at the time. Since usapt0.cc does not put uncoupled induction in the coupled induction variables, the `SAPT0 IND ENERGY` is not defined properly any more.

Possible solution: define uncoupled variants of the variables in `psivardefs.py`, and when we create `SAPT IND ENERGY` (in `proc.py`) we copy `SAPT0 IND,U ENERGY` value if `SAPT0 IND ENERGY` is not defined. 
- Would that work ?
- Is there a way to check whether a Psivar is defined other than checking if it is exactly 0 ?

dgasmith:
Currently on SAPT(DFT) goes through `sapt_jk_terms.py`. I kind of hope that we can move more of the SAPT0 code over to something similar with Disp staying fully C++. We are going to gain a lot of SAPT0 capabilities soon like spin-flip SAPT, Sinf Exch-Ind and Exch-Disp, regularized SAPT, ISA Exch etc and the Python level interface should make it easier to handle without performance degradation.

I think the proposed would work fine. You can use either of the following:
```python
>>> psi4.core.get_variables() # Returns a dict of all variables
>>> psi4.core.has_variable("CURRENT DIPOLE X") # Returns a bool checking if the variable is set
```

jgonthier:
Exciting news for the SAPT code then! I just pushed a new commit that should fix sapt8. At least it's working on my machine. Let me know if my fix is convenient.

dgasmith:
Looks good to me, but I think is mostly a @loriab question. I believe this is ready to go now?

loriab:
Yeah, strangely for C++, psivars behaves like defaultdict – as soon as you ref a psivar, it exists. I agree some sort of has_key is needed.

Do `NO_RESPONSE` and `COUPLED_INDUCTION` need a reconciliation or are they differentiated?

jgonthier:
`NO_RESPONSE` is the option for the closed-shell sapt0 code while `COUPLED_INDUCTION` is the one for the open-shell SAPT0 code. I don't remember why I felt the need to name the options differently at the time. As far as I can see they are doing the same thing in both codes. Shall we replace both of them with `NO_RESPONSE` ?

jgonthier:
Also, @loriab, should we keep both options working in the input file for backward compatibility? Or is that not an issue ?

loriab:
I'm in breaking mode, so I'd say choose one. I like "coupled induction" rather better, but "no_response" may be more widely known. Up to you. :-)

jgonthier:
Sorry, that took me some time to actually get to. ``NO_RESPONSE`` has been replaced by ``COUPLED_INDUCTION`` and the documentation should be up to date as well. Auto-doc should build the proper keyword documentation and ``COUPLED_INDUCTION`` is now included in the advanced SAPT keywords and in the open-shell SAPT keywords. This is because the default for ROHF is ``False`` instead of ``True`` because the related equations are not implemented yet.

jgonthier:
@loriab, @dgasmith : I see the Travis CI failed because one of the builds timed out. Is there anything I need to do to correct that?

dgasmith:
Restarted. The land of timeouts is annoying. Is the build time going back up?

dgasmith:
Can we get one more review here? Time to get this in I think.

PR #912, opened 2018-02-07
turning off iwl integrals for df-dcft

hokru:
## Description
The IWL integrals are calculated but apparently not needed for DF-DCFT. Added IF/ELSE clause to avoid this.

## Status
- [x] Ready to go


jturney:
Oops, that slipped through our fingers when we implemented DF-DFCT. Good catch.

PR #913, opened 2018-02-08
New DF_Helper Technology

schiebermc:
## Description
Furthers the generality of integral transformations and enables a special workflow, `DIRECT_iaQ`, to alleviate disk IO when transforming to `pqQ` forms.  Timings comparing the `DIRECT_iaQ` and `DIRECT` methods are inbound.

Edit: Here are some tests that (almost comically) display the efficacy of `DIRECT_iaQ` over `DIRECT` + `pqQ`.

1. Test1: 
[input.txt](https://github.com/psi4/psi4/files/1740970/input.txt)
[timer.txt](https://github.com/psi4/psi4/files/1740977/timer.txt)

2. Test2:  
[input.txt](https://github.com/psi4/psi4/files/1740966/input.txt)
[timer.txt](https://github.com/psi4/psi4/files/1740974/timer.txt)

Notice the difference between the `DFH: MO to disk` timers.  

Edit2: I added a reduction on `max_val` in `prepare_sparsity` to ensure thread safety.  This was incorrect before.

## Todos
* **Developer Interest**
  - DF_Helper now has three methods for integral transformations:
    `STORE`:  Contracts metric with AO integrals, stores, then transforms.
    `DIRECT`: Transforms integrals, contracts with metric.
    `DIRECT_iaQ`: Optimized `DIRECT` workflow when using `pqQ` transformed integrals.

  - DF_Helper outputs any transformed integral form you want: `Qpq`, `pQq`, or `pqQ`.  The following can be used for each of the respective methods listed above:
    `STORE`:  `Qpq`, `pQq`, or (ill-advised) `pqQ`
    `DIRECT`:  `Qpq`, `pQq`, or (ill-advised) `pqQ`
    `DIRECT_iaQ`:   `pqQ`

## Status
- [x] Ready to go

### DIFF
--- a/psi4/src/psi4/dfep2/dfep2.cc
+++ b/psi4/src/psi4/dfep2/dfep2.cc
@@ -91,7 +91,7 @@ DFEP2Wavefunction::DFEP2Wavefunction(std::shared_ptr<Wavefunction> ref_wfn)
 
     // ==> Init DF object <== /
     dfh_ = std::make_shared<DF_Helper>(get_basisset("ORBITAL"), get_basisset("DF_BASIS_SCF"));
-    dfh_->set_method("DIRECT");
+    dfh_->set_method("DIRECT_iaQ");
     dfh_->set_memory(memory_doubles_);
     dfh_->set_nthreads(num_threads_);
     dfh_->initialize();
@@ -198,6 +198,9 @@ std::vector<std::vector<std::pair<double, double>>> DFEP2Wavefunction::compute(
 
     // ==> Transform DF integrals <== /
 
+    // safety check
+    dfh_->clear_spaces();
+    
     // add spaces
     dfh_->add_space("i", AO_Cocc_);
     dfh_->add_space("a", AO_Cvir_);
--- a/psi4/src/psi4/dfep2/dfep2.cc
+++ b/psi4/src/psi4/dfep2/dfep2.cc
@@ -199,7 +199,6 @@ std::vector<std::vector<std::pair<double, double>>> DFEP2Wavefunction::compute(
     // ==> Transform DF integrals <== /
 
     // add spaces
-    dfh_->clear_all();
     dfh_->add_space("i", AO_Cocc_);
     dfh_->add_space("a", AO_Cvir_);
     dfh_->add_space("E", AO_CE);
--- a/psi4/src/psi4/fisapt/fisapt.cc
+++ b/psi4/src/psi4/fisapt/fisapt.cc
@@ -2049,7 +2049,7 @@ void FISAPT::disp(std::map<std::string, SharedMatrix> matrix_cache, std::map<std
     // => Get integrals from DF_Helper <= //
     auto dfh(std::make_shared<DF_Helper>(primary_, auxiliary));
     dfh->set_memory(doubles_ - Cs[0]->nrow() * ncol);
-    dfh->set_method("DIRECT");
+    dfh->set_method("DIRECT_iaQ");
     dfh->set_nthreads(nT);
     dfh->initialize();
     dfh->print_header();
@@ -2067,16 +2067,16 @@ void FISAPT::disp(std::map<std::string, SharedMatrix> matrix_cache, std::map<std
     dfh->add_space("a4", Cs[10]);
     dfh->add_space("b4", Cs[11]);
 
-    dfh->add_transformation("Aar", "a", "r", "pqQ");
-    dfh->add_transformation("Abs", "b", "s", "pqQ");
-    dfh->add_transformation("Bas", "a", "s1", "pqQ");
-    dfh->add_transformation("Bbr", "b", "r1", "pqQ");
-    dfh->add_transformation("Cas", "a2", "s", "pqQ");
-    dfh->add_transformation("Cbr", "b2", "r", "pqQ");
-    dfh->add_transformation("Dar", "a", "r3", "pqQ");
-    dfh->add_transformation("Dbs", "b", "s3", "pqQ");
-    dfh->add_transformation("Ear", "a4", "r", "pqQ");
-    dfh->add_transformation("Ebs", "b4", "s", "pqQ");
+    dfh->add_transformation("Aar", "a" , "r" );
+    dfh->add_transformation("Abs", "b" , "s" );
+    dfh->add_transformation("Bas", "a" , "s1");
+    dfh->add_transformation("Bbr", "b" , "r1");
+    dfh->add_transformation("Cas", "a2", "s" );
+    dfh->add_transformation("Cbr", "b2", "r" );
+    dfh->add_transformation("Dar", "a" , "r3");
+    dfh->add_transformation("Dbs", "b" , "s3");
+    dfh->add_transformation("Ear", "a4", "r" );
+    dfh->add_transformation("Ebs", "b4", "s" );
 
     dfh->transform();
 
@@ -2637,7 +2637,7 @@ void FISAPT::felst() {
     // => Get integrals from DF_Helper <= //
     auto dfh(std::make_shared<DF_Helper>(primary_, reference_->get_basisset("DF_BASIS_SCF")));
     dfh->set_memory(doubles_);
-    dfh->set_method("DIRECT");
+    dfh->set_method("DIRECT_iaQ");
     dfh->set_nthreads(nT);
     dfh->initialize();
     dfh->print_header();
@@ -2645,8 +2645,8 @@ void FISAPT::felst() {
     dfh->add_space("a", matrices_["Locc0A"]);
     dfh->add_space("b", matrices_["Locc0B"]);
 
-    dfh->add_transformation("Aaa", "a", "a", "pqQ");
-    dfh->add_transformation("Abb", "b", "b", "pqQ");
+    dfh->add_transformation("Aaa", "a", "a");
+    dfh->add_transformation("Abb", "b", "b");
 
     dfh->transform();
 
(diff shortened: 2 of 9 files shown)


psi4/src/psi4/dfep2/dfep2.cc
```diff
@@ -199,7 +199,6 @@ std::vector<std::vector<std::pair<double, double>>> DFEP2Wavefunction::compute(
     // ==> Transform DF integrals <== /
 
     // add spaces
-    dfh_->clear_all();
```
dgasmith: Why pop this? Seems like a good general safety check.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -3807,16 +3808,16 @@ void FISAPT::fdisp() {
...
+    dfh->add_transformation("Aar", "r" , "a" );
+    dfh->add_transformation("Abs", "s" , "b" );
+    dfh->add_transformation("Bas", "s1", "a" );
+    dfh->add_transformation("Bbr", "r1", "b" );
+    dfh->add_transformation("Cas", "s" , "a2");
+    dfh->add_transformation("Cbr", "r" , "b2");
+    dfh->add_transformation("Dar", "r3", "a" );
+    dfh->add_transformation("Dbs", "s3", "b" );
+    dfh->add_transformation("Ear", "r" , "a4");
+    dfh->add_transformation("Ebs", "s" , "b4");
```
dgasmith: I really like how simple this is to change underlying algorithms.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -776,7 +864,71 @@ void DF_Helper::get_tensor_(std::string file, double* b, const size_t start1, co
...
+    // Here, we compute dense AO integrals in the Qpq memory layout.
+    // Sparsity and permutational symmetry are used in the computation,
+    // but not in the resulting tensor.
+ 
+    size_t begin = Qshell_aggs_[start];
+    size_t end = Qshell_aggs_[stop + 1] - 1;
+    size_t block_size = end - begin + 1;
+
+    // stripe the buffer
+    ::memset(Mp, 0, sizeof(double) * block_size * nao_ * nao_);
```
dgasmith: Any reason not to use `std::fill` here? Just to keep things consistent.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1047,6 +1200,55 @@ std::string DF_Helper::compute_metric(double pow) {
...
+    std::string op = "wb";
+
+    // determine blocking, block through l
+    std::vector<std::pair<size_t, size_t>> steps;
+    for (size_t i = 0; i < l; i++) {
+        count++;
+        if (tots < count * Q * r || i == l - 1) {
+            if (count == 1 && i != l - 1) {
+                std::stringstream error;
+                error << "DF_Helper:contract_metric: not enough memory.";
```
dgasmith: Can we do anything here so that we can tell the user how much memory they need?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1047,6 +1200,55 @@ std::string DF_Helper::compute_metric(double pow) {
...
+                maxim = (maxim < count ? count - 1 : maxim);
+                steps.push_back(std::make_pair(i - count + 1, i - 1));
+                i--;
+            } else {
+                maxim = (maxim < count ? count : maxim);
+                steps.push_back(std::make_pair(i - count + 1, i));
+            }
+            count = 0;
+        }
+    }
```
dgasmith: I trust that this works, but I think it would be a bit difficult to understand the logic. Can you add a comment on how this works or see if there is a simpler way to figure out the blocking?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1551,282 +1747,303 @@ void DF_Helper::transform_core() {
...
-                    // setup putf
-                    std::string putf =
-                        (!direct_ ? std::get<1>(files_[order_[count + k]]) : std::get<0>(files_[order_[count + k]]));
+        #ifdef _OPENMP
+        rank = omp_get_thread_num();
+        #endif
+        for (size_t m = 0, sp_count = -1; m < nao_; m++) {
+            if (schwarz_fun_mask_[k * nao_ + m]) {
+                sp_count++;
+                C_DCOPY(bsize, &Bp[m * bsize], 1, &C_buffers[rank][sp_count * bsize], 1);
```
dgasmith: If we run this through VTune, how much time do we spending copying these orbitals to the new format? This is an aligned copy, so much better than previous implementations with strides copies. However, any rank 1 operation these days is pretty painful.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1551,282 +1747,303 @@ void DF_Helper::transform_core() {
...
+            // (w|Qb)->(bQw)
+            #pragma omp parallel for num_threads(nthreads_)
+            for (size_t x = 0; x < rblock_size; x++) {
+                for (size_t y = 0; y < bsize; y++) {
+                    for (size_t z = 0; z < wsize; z++) {
+                        Np[y * lblock_size * wsize + (bcount + x) * wsize + z] 
+                            = Fp[z * bsize *rblock_size + x * bsize + y];
+                    }
+                }
+            }
```
dgasmith: Ouch, I was worried about the aligned COPY and then I find this. Be curious to see how much time we spend on these operations as well.

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -1551,282 +1747,303 @@ void DF_Helper::transform_core() {
...
+
+            // (w|Qb)->(Q|wb)
+            #pragma omp parallel for num_threads(nthreads_)
+            for (size_t x = 0; x < rblock_size; x++) {
+                for (size_t z = 0; z < wsize; z++) {
+                    #pragma omp simd
+                    for (size_t y = 0; y < bsize; y++) {
+                        Np[(bcount + x) * wsize * bsize + z * bsize + y] 
+                            = Fp[z * rblock_size * bsize + x * bsize + y];
+                    }
```
dgasmith: Any reason to use SIMD here and C_DCOPY below?

psi4/src/psi4/lib3index/df_helper.h
```diff
@@ -335,19 +348,34 @@ class DF_Helper {
...
     std::vector<std::string> bspace_;
     std::vector<size_t> strides_;
 
     // => FILE IO maintenence <=
-    struct stream {
-        FILE* fp;
-        std::string op;
-    };
-    std::map<std::string, stream> file_status_;
+    typedef struct StreamStruct {
```
dgasmith: One thing id like to keep in mind while we are doing new IO tech is to always ask the question "Can I port this to HDF5 easily"? Curious on your thoughts here.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: No reason other than inconsistency.  I will revert to C_DCOPY.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: These operations are timed and included in `timer.dat` as `DFH: (w|Qb)->(bw|Q)`.  Although it might be preferable to switch to `DFH: Transpose (w|Qb)` after these latest edits.  I agree these are ugly operations, but they are necessary and fairly inconsequential in the wake of some of the O(N^4) operations occurring.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: Actually, this copy also occurs in previous implementations:
https://github.com/psi4/psi4/blob/4c3ce1aa1fb8fe6914448071af33d1ed7cf7c5f2/psi4/src/psi4/libfock/DFJK.cc#L1823
`DF_Helper` avoids the strided copy altogether.  
Still, profiling this on VTune is necessary, will get back to you about that.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: Good point, this is getting a bit blurred as I try to make it general with all possibilities.  I will see what I can do.

psi4/src/psi4/lib3index/df_helper.cc
dgasmith: Keep in mind that our N isn't all that large, in the DFJK example we noticed that we spent ~50-70% of our time in a strided N^3 copy while the N^4 GEMM that *should* have been the bottleneck.

psi4/src/psi4/lib3index/df_helper.h
```diff
@@ -280,15 +283,24 @@ class DF_Helper {
     // => AO building machinery <=
     void prepare_AO();
     void prepare_AO_core();
-    void compute_AO_Q(const size_t start, const size_t stop, double* Mp,
+    void compute_dense_Qpq_blocking_Q(const size_t start, const size_t stop, double* Mp,
+                      std::vector<std::shared_ptr<TwoBodyAOInt>> eri);
+    void compute_sparse_pQq_blocking_Q(const size_t start, const size_t stop, double* Mp,
```
andysim: +1 for the verbose function names here.  Makes it so much easier to understand at first glance

psi4/src/psi4/lib3index/df_helper.h
schiebermc: The DF_Helper IO tech is mostly the same, I just organized it differently.  I am not sure how this impacts portability to HDF5.  I would be happy to help move DF_Helper IO to a new regime once we get there.

dgasmith:
@schiebermc Is this ready to go? Please click the check mark and we can pull it in.

schiebermc:
@dgasmith Yes, just checked. Sorry about that.

Issue #870, opened 2017-11-29
vv10 correction usage


hokru:
I would like to use the vv10 kernel with custom parameters to use it in the DFT-NL fashion (http://pubs.acs.org/doi/full/10.1021/ct200644w).
One needs to modify the b and c parameters. I figured out that those are obtained from libxc in `LibXCfunctional.cc`. I suppose that within `LibXCFunctional::build_worker` one has to add a user-tweak somehow to modify the vv10 parameters.

So to the actual questions:

- Can I already request the VV10 kernel to be included for a given functional through the python interface? (parameters aside)
- Can I already request the VV10 kernel in a post-SCF fashion for a given density (a non-SCF VV10 is perfectly fine for energies, it is the default in ORCA). I think I saw this mentioned somewhere, but cannot find it.
- Any further helpful pointers to within code where I would need to make changes or need to take a look? (I am more a Fortran than C++ guy, my head hurts already)

dgasmith:
I dont have access to that paper at the moment, but a few notes for you:
- If you need to tweak functionals (@susilehtola this is a good example) please checkout [this](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/dft_funcs/hyb_superfuncs.py#L230) example.

- For VV10 you can specify it via the python interface:
```python
def build_myvv10_superfunctional(name, npoints, deriv, restricted):
    sup = core.SuperFunctional.blank()
    sup.set_name('HF+VV10')
    sup.set_x_alpha(1.0)

    # Set VV10 parameters. If these are changed we assume you want VV10.
    sup.vv10_b(...)
    sup.vv10_c(...)

    return sup

func_call = energy('SCF', dft_functional= build_myvv10_superfunctional)
```

The DFT code is likely the most class/inheritance/convoluted C++ code we have besides perhaps integrals. So probably not a great place to start in :). 

- Post-SCF only VV10 is a good idea and just a matter of code logic. We need a flag somewhere that computes [this](https://github.com/psi4/psi4/blob/f03a896a2d70eac47137cf471dda4805e483586d/psi4/src/psi4/libfock/v.cc#L213) at the end of a SCF computation. The SCF iterations are in a bit of a state of flux due to the python porting; however, this should be pretty easy once merged as we have access to that Py-side.

hokru:
Ah wonderful! Yes, `sup.set_vv10_b()` is what I was looking for.

[...]

PR #914, opened 2018-02-08
adding DFT-NL scheme

hokru:
## Description
DFT-NL (=non-local) adds the VV10 correlation energy to arbitrary DFT functionals using functional specific  vv10_b parameter. Conceptually similar to DFT-D. DOI: 10.1021/ct200644w

Fixes https://github.com/psi4/psi4/issues/870

  - [x] SCF option `DFT_VV10_B` enables and adds VV10 correlation according to DFT-NL scheme (fixed `vv10_c`, user-supplied `vv10_b`). 
  - [x] `DFT_VV10_C` also gives access to the C parameter.
  - [x] `NL_DISPERSION_PARAMETERS` similar to `DFT_DISPERSION_PARAMETERS`
  - [x] `energy('functional-nl')` sets pre-defined vv10_b parameters for over 15 functionals. Recommended usage for most users.
  - [X] `DFT_VV10_POSTSCF` enables a post-scf VV10 calculation. Large gain in speed with minimal loss of VV10 correlation energy.
  - [x] added documentation. 

* **Developer notes**
  - [x] rebase+adjustment to PR https://github.com/psi4/psi4/pull/922
  - [x] figure out DSD variant and param citations
  - [x] re-integrate tests

* ** Notes**
   * forcing pure HF with VV10  will seg. fault. (no integration grid). Only `energy('hf')` with `set DFT_VV10_B` is caught, not `energy('scf')` to allow custom access to the scf.
   * NL parameters were entered manually, someone please check over them for strange values or typos
   * tests are kept short, but we also don't test every dispersion dash parameter out there.
   * for double-hybrids with multiple versions, the frozen core and/or D3BJ parent variant was chosen for a sensible consistency. 

## Status
- [x] Ready to go

### DIFF
--- a/doc/sphinxman/source/dftnl.rst
+++ b/doc/sphinxman/source/dftnl.rst
@@ -1,0 +1,73 @@
+.. #
+.. # @BEGIN LICENSE
+.. #
+.. # Psi4: an open-source quantum chemistry software package
+.. #
+.. # Copyright (c) 2007-2018 The Psi4 Developers.
+.. #
+.. # The copyrights for code used from other parties are included in
+.. # the corresponding files.
+.. #
+.. # This file is part of Psi4.
+.. #
+.. # Psi4 is free software; you can redistribute it and/or modify
+.. # it under the terms of the GNU Lesser General Public License as published by
+.. # the Free Software Foundation, version 3.
+.. #
+.. # Psi4 is distributed in the hope that it will be useful,
+.. # but WITHOUT ANY WARRANTY; without even the implied warranty of
+.. # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+.. # GNU Lesser General Public License for more details.
+.. #
+.. # You should have received a copy of the GNU Lesser General Public License along
+.. # with Psi4; if not, write to the Free Software Foundation, Inc.,
+.. # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+.. #
+.. # @END LICENSE
+.. #
+
+.. include:: autodoc_abbr_options_c.rst
+
+.. index:: DFTNL
+.. _`sec:dftnl`:
+
+
+DFT-NL
+======
+
+.. sectionauthor:: Holger Kruse
+
+Non-local (NL), density based correlation energy from the VV10 kernel can be added
+to arbitrary functionals.
+
+.. math:: E_{DFT-NL}=E_{DFT}+E_{NL}
+
+For pre-defined functionals (see Functional overview in :ref:`this Table <table:dft_all>` ) it is sufficient to add `-NL` to 
+the functional name::
+
+    energy('b3lyp-nl')
+
+Modification of the parameters `b` and `C` is done setting |scf__DFT_VV10_B| and |scf__DFT_VV10_C|. The `C` is usually left unchanged and the originally proposed
+value of `C=0.0093` is used. 
+
+Adding |scf__DFT_VV10_B| to any functional activates the calculation of the VV10 kernel. A BLYP-NL calculation can be set as follows::
+   
+    set DFT_VV10_B 4.0
+    energy('blyp')
+
+The default `C` parameter will be used.
+
+Similar to |scf_DFT_DISPERSION_PARAMETERS| the tuple |scf_NL_DISPERSION_PARAMETERS| can used::
+    
+    set NL_DISPERSION_PARAMTERS [4.0]
+    energy('blyp')
+
+which is equivalent to the example above.
+
+Further examples can be found in the respective :source:`regression test <tests/dft-vv10/input.dat>`
+
+post-SCF time savings
+~~~~~~~~~~~~~~~~~~~~~
+
+Substantial time-savings for energy calculations are available by evaluating the VV10 kernel only at the converged electron density, i.e. in a post-SCF fashion.
+The deviations from the fully self-consistent treatment are usually minimal. To activate this set |scf_DFT_VV10_POSTSCF| to `true`.
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -348,6 +348,32 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
            sup.set_vv10_c(d_params["params"]["c"])
         dispersion = d_params
 
+    # custom options for NL dispersion
+    # wants to be where the functional is build
+    if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS")):
+       nl_tuple = core.get_option("SCF", "DFT_DISPERSION_PARAMETERS")
+       sup.set_vv10_b(nl_tuple[0])
+       if len(nl_tuple) > 1:
+          sup.set_vv10_c(nl_tuple[1])
+       if len(nl_tuple) > 2:
+          raise ValidationError("too many entries in DFT_DISPERSION_PARAMETERS for DFT-NL")
+    if (core.has_option_changed("SCF", "DFT_VV10_B") and core.has_option_changed("SCF", "DFT_VV10_C")):
+       if(name.lower()=='scf'):
+          raise ValidationError("SCF: HF with -NL not implemented")
+       vv10_b = core.get_option("SCF", "DFT_VV10_B")
+       vv10_c = core.get_option("SCF", "DFT_VV10_C")
+       sup.set_vv10_b(vv10_b)
+       sup.set_vv10_c(vv10_c)
+    elif core.has_option_changed("SCF", "DFT_VV10_B"):
+       if(name.lower()=='scf'):
+          raise ValidationError("SCF: HF with -NL not implemented")
+       vv10_b = core.get_option("SCF", "DFT_VV10_B")
+       sup.set_vv10_b(vv10_b)
+       if (abs(sup.vv10_c() - 0.0) <= 1e-8):
+           core.print_out("SCF: VV10_C not specified. Using default (C=0.0093)!")
+           sup.set_vv10_c(0.0093)
+
+
     sup.set_max_points(npoints)
     sup.set_deriv(deriv)
     sup.set_name(func_dictionary["name"].lower())
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -324,7 +324,6 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
                 sup.set_c_os_alpha(c_params["os"])
                 sup.set_c_alpha(1.0)
 
-
         # Merge descriptions and citations from above components obtained from LibXC as a fallback.
         descr = "\n".join(descr)
         citation = "\n".join(citation)
@@ -342,10 +341,10 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
     if "dispersion" in func_dictionary:
         d_params = func_dictionary["dispersion"]
         if "citation" not in d_params:
-           d_params["citation"] = False
-        if d_params["type"]=='nl':
-           sup.set_vv10_b(d_params["params"]["b"])
-           sup.set_vv10_c(d_params["params"]["c"])
+            d_params["citation"] = False
+        if d_params["type"] == 'nl':
+            sup.set_vv10_b(d_params["params"]["b"])
+            sup.set_vv10_c(d_params["params"]["c"])
         dispersion = d_params
 
     sup.set_max_points(npoints)
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -348,31 +348,6 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
            sup.set_vv10_c(d_params["params"]["c"])
         dispersion = d_params
 
-    # custom options for NL dispersion
-#    if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS")):
-#       nl_tuple = core.get_option("SCF", "DFT_DISPERSION_PARAMETERS")
-#       sup.set_vv10_b(nl_tuple[0])
-#       if len(nl_tuple) > 1:
-#          sup.set_vv10_c(nl_tuple[1])
-#       if len(nl_tuple) > 2:
-#          raise ValidationError("too many entries in DFT_DISPERSION_PARAMETERS for DFT-NL")
-#    if (core.has_option_changed("SCF", "DFT_VV10_B") and core.has_option_changed("SCF", "DFT_VV10_C")):
-#       if(name.lower()=='hf'):
-#          raise ValidationError("SCF: HF with -NL not implemented")
-#       vv10_b = core.get_option("SCF", "DFT_VV10_B")
-#       vv10_c = core.get_option("SCF", "DFT_VV10_C")
-#       sup.set_vv10_b(vv10_b)
-#       sup.set_vv10_c(vv10_c)
-#    elif core.has_option_changed("SCF", "DFT_VV10_B"):
-#       if(name.lower()=='scf'):
-#          raise ValidationError("SCF: HF with -NL not implemented")
-#       vv10_b = core.get_option("SCF", "DFT_VV10_B")
-#       sup.set_vv10_b(vv10_b)
-#       if (abs(sup.vv10_c() - 0.0) <= 1e-8):
-#           core.print_out("SCF: VV10_C not specified. Using default (C=0.0093)!")
-#           sup.set_vv10_c(0.0093)
-
-
     sup.set_max_points(npoints)
     sup.set_deriv(deriv)
     sup.set_name(func_dictionary["name"].lower())
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -349,29 +349,28 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
         dispersion = d_params
 
     # custom options for NL dispersion
-    # wants to be where the functional is build
-    if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS")):
-       nl_tuple = core.get_option("SCF", "DFT_DISPERSION_PARAMETERS")
-       sup.set_vv10_b(nl_tuple[0])
-       if len(nl_tuple) > 1:
-          sup.set_vv10_c(nl_tuple[1])
-       if len(nl_tuple) > 2:
-          raise ValidationError("too many entries in DFT_DISPERSION_PARAMETERS for DFT-NL")
-    if (core.has_option_changed("SCF", "DFT_VV10_B") and core.has_option_changed("SCF", "DFT_VV10_C")):
-       if(name.lower()=='scf'):
-          raise ValidationError("SCF: HF with -NL not implemented")
-       vv10_b = core.get_option("SCF", "DFT_VV10_B")
-       vv10_c = core.get_option("SCF", "DFT_VV10_C")
-       sup.set_vv10_b(vv10_b)
-       sup.set_vv10_c(vv10_c)
-    elif core.has_option_changed("SCF", "DFT_VV10_B"):
-       if(name.lower()=='scf'):
-          raise ValidationError("SCF: HF with -NL not implemented")
-       vv10_b = core.get_option("SCF", "DFT_VV10_B")
-       sup.set_vv10_b(vv10_b)
-       if (abs(sup.vv10_c() - 0.0) <= 1e-8):
-           core.print_out("SCF: VV10_C not specified. Using default (C=0.0093)!")
-           sup.set_vv10_c(0.0093)
+#    if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS")):
+#       nl_tuple = core.get_option("SCF", "DFT_DISPERSION_PARAMETERS")
+#       sup.set_vv10_b(nl_tuple[0])
+#       if len(nl_tuple) > 1:
+#          sup.set_vv10_c(nl_tuple[1])
+#       if len(nl_tuple) > 2:
+#          raise ValidationError("too many entries in DFT_DISPERSION_PARAMETERS for DFT-NL")
+#    if (core.has_option_changed("SCF", "DFT_VV10_B") and core.has_option_changed("SCF", "DFT_VV10_C")):
+#       if(name.lower()=='hf'):
+#          raise ValidationError("SCF: HF with -NL not implemented")
+#       vv10_b = core.get_option("SCF", "DFT_VV10_B")
+#       vv10_c = core.get_option("SCF", "DFT_VV10_C")
+#       sup.set_vv10_b(vv10_b)
+#       sup.set_vv10_c(vv10_c)
+#    elif core.has_option_changed("SCF", "DFT_VV10_B"):
+#       if(name.lower()=='scf'):
+#          raise ValidationError("SCF: HF with -NL not implemented")
+#       vv10_b = core.get_option("SCF", "DFT_VV10_B")
+#       sup.set_vv10_b(vv10_b)
+#       if (abs(sup.vv10_c() - 0.0) <= 1e-8):
+#           core.print_out("SCF: VV10_C not specified. Using default (C=0.0093)!")
+#           sup.set_vv10_c(0.0093)
 
 
     sup.set_max_points(npoints)
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -324,6 +324,7 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
                 sup.set_c_os_alpha(c_params["os"])
                 sup.set_c_alpha(1.0)
 
+
         # Merge descriptions and citations from above components obtained from LibXC as a fallback.
         descr = "\n".join(descr)
         citation = "\n".join(citation)
@@ -341,7 +342,10 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
     if "dispersion" in func_dictionary:
         d_params = func_dictionary["dispersion"]
         if "citation" not in d_params:
-            d_params["citation"] = False
+           d_params["citation"] = False
+        if d_params["type"]=='nl':
+           sup.set_vv10_b(d_params["params"]["b"])
+           sup.set_vv10_c(d_params["params"]["c"])
         dispersion = d_params
 
     sup.set_max_points(npoints)
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -349,7 +349,6 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
 
     sup.set_max_points(npoints)
     sup.set_deriv(deriv)
-#    sup.set_name(func_dictionary["name"].lower())
     sup.set_name(func_dictionary["name"].upper())
     sup.allocate()
     return (sup, dispersion)
--- a/psi4/driver/procrouting/dft_funcs/dict_builder.py
+++ b/psi4/driver/procrouting/dft_funcs/dict_builder.py
@@ -349,6 +349,7 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
 
     sup.set_max_points(npoints)
     sup.set_deriv(deriv)
-    sup.set_name(func_dictionary["name"].lower())
+#    sup.set_name(func_dictionary["name"].lower())
+    sup.set_name(func_dictionary["name"].upper())
     sup.allocate()
     return (sup, dispersion)
--- a/psi4/driver/procrouting/dft_funcs/superfuncs.py
+++ b/psi4/driver/procrouting/dft_funcs/superfuncs.py
@@ -99,14 +99,15 @@ def build_superfunctional(name, restricted):
     if core.has_option_changed("SCF", "DFT_ALPHA_C"):
         sup[0].set_c_alpha(core.get_option("SCF", "DFT_ALPHA_C"))
 
-    # customization of NL dispersion:
-    if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS")):
+    # customization of existing VV10 dispersion:
+    if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS") and sup[0].vv10_b() > 0.0):
        nl_tuple = core.get_option("SCF", "DFT_DISPERSION_PARAMETERS")
        sup[0].set_vv10_b(nl_tuple[0])
        if len(nl_tuple) > 1:
           sup[0].set_vv10_c(nl_tuple[1])
        if len(nl_tuple) > 2:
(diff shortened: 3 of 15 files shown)


susilehtola:
If the user has to supply the b parameter anyhow, wouldn't it make sense just to make using VV10 (specify b and C) easier on top of any functional....?

hokru:
Not sure what you exactly mean. Setting vv10_b in the superfunctional is straightforward. I followed what was done for the LRC omega. 

Ideally the user has not do anything for the pre-defined functionals once I get the -nl dashparam to work. The manual setting is more for developers who want to fit the vv10_b for a new functional. Using DFT-NL should be like using DFT-D.

Having the vv10 integrator exposed to python and supplying a user-defined density would be nice. And for a post-scf VV10 correction that is probably the way to go for the scfitertopy implementation. Though this pybind is currently a level too much for me, i dont even understand the errors i get. Small steps for me ;-)

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -92,6 +92,7 @@
             sup.set_name(sup.name() + '-D3M')
             superfunctional_list.append(sup)
```
dgasmith: Right here add the following to get `energy('func-NL')` to work:

```
        if dashlvl == 'nl':
            # -D2 overide
            sup = superfunctionals[func](func, 1, 1, True)[0]
            sup.set_name(sup.name() + '-NL')
            superfunctional_list.append(sup)
```

dgasmith:
As a general note, if you are working on Psi4 please drop us a line if you are having issues. Contributing should be an easy process and we are here to help! I believe you are on our Slack channel, feel free to comment or ping me there. If not we can certainly shoot out an invite (thats general to anyone reading this).

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -194,6 +200,14 @@ def build_superfunctional(name, restricted):
         else:
             for x_func in sup[0].x_functionals():
                 x_func.set_omega(omega)
+
+    # add VV10 correlation to DFT aka DFT-NL (Hujo/Grimme) 
+    if core.has_option_changed("SCF", "DFT_VV10_B"):
+        vv10_b = core.get_option("SCF", "DFT_VV10_B")
+        sup[0].set_vv10_b(vv10_b)
+        sup[0].set_vv10_c(0.0093) # unchanged in DFT-NL scheme
```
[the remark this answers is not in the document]
loriab: `dashparams` should be already set up to allow on-the-fly parameter changes, e.g., https://github.com/psi4/psi4/blob/master/tests/dftd3/energy/input.dat#L93-L103. Will have to be hooked up to your code rather than the dftd3-runner.

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -175,45 +178,68 @@ def print_out(self, level=1):
...
 
         # Psi auto sets this with no change of deviation
         if (self.disp_type == 'p4') and (self.dtype == "-D2"):
             core.print_out("    %6s = %14.6E\n" % ("A6", 20.0))
 
-        core.print_out("\n");
+        core.print_out("\n")
 
     def compute_energy(self, molecule):
+        if self.disp_type == 'nl':
```
hokru: this looks even to me poorly solved. Just needed it to stop giving errors to get a working version.

hokru:
The planned functionality works and my code does no longer cause obvious trouble in the various tests.

However, especially some of the modifications in`empirical_dispersion.py` seem not elegant.  I force `return 0 `in the compute_energy/gradient/hessian parts. Otherwise I get errors like:
`AttributeError: 'NoneType' object has no attribute 'compute_energy`
Here I would need advice if the solution is not acceptable.

I will come with questions in slacks soon enough ;)

dgasmith:
Yea, this one is on us. We need to talk a bit about how to make operations like you are showing easier. We were only thinking of -D when originally building this interface, but allowing a bit more flexibility when mixing these functionals together would be useful.

The other option is to use a function generator in `custom_hyb_funcs.py` or `superfuncs.py` that loops over all nl `dash_params`, generates a new function, and appends it the dictionary.

PeterKraus:
Just out of interest, what's the status of rVV10? [This](http://dx.doi.org/10.1021/acs.jpclett.6b02527) recent Mardirossian paper seems to suggest it's mostly a drop-in replacement for VV10.

hokru:
Looks like one needs to only needs to change phi? That is here:
https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfunctional/superfunctional.cc#L869

dgasmith:
Adding rVV10 could be a good new developer project as it should be fairly straightforward to do.

This PR has gotten a bit complex as we didn't quite anticipate this use case. We might want to roll with this for now as long as we are considering redoing each Superfunctional as a dictionary.

dgasmith:
I think #922 makes this quite a bit easier?

PeterKraus:
@hokru the idea is that the functional would simply specify:
```
    ...
    "dispersion": {"type": "nl", "params": {"b": X, "c": Y}, "citation": "My Tetrahedron Letters Paper"
    ...
```
The type is obviously up for discussion - `nl  == VV10` or `rVV10` etc.

The builder would then set all the required parameters. If this gets merged before PR #922 , and there's at least one superfunctional that uses this path, I can fix the other PR to work with this new code.

hokru:
Yes, but that needs to be done automatically when the user specifies -nl in the energy call.
I just cant picture all the work flow in my head right now. Maybe next week I have more time for this.

As a side note: the current changes in compute_gradient+hessian in empirical_dispersion.py  are not working correctly. (they should never be called since we need numerical gradients, but still...)

dgasmith:
@hokru Can you and @PeterKraus look at updating this to the new DFT scheme? Happy to help here as well.

hokru:
Yes, I will look into it in the next few days.

psi4/driver/procrouting/dft_funcs/superfuncs.py
PeterKraus: This bit is now gone - the handling of dispersion is done in `dict_builder.py`. As long as `qcdb.dashparam.get_dispersion_aliases("nl")` returns `"nl"`, the `"dispersion"` key in the dictionary of the functional will be sourced from `qcdb.dashparam.dashcoeff["nl"][func]` automatically.

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -182,6 +187,23 @@ def build_superfunctional(name, restricted):
...
     sup[0].set_lock(False)
 
+    # pre-defined -NL dashparam first, then special dft_disp_parm option.
+    # not applicable for XC generating function
+    if not hasattr(name, '__call__'):
+        if (any(name.lower().endswith(al) for al in dftd3.full_dash_keys)):
+            dashparam = [x for x in dftd3.full_dash_keys if name.endswith(x)]
+            if (dashparam[0] == 'nl'):
+                base_name = name.replace('-' + dashparam[0], '')
+                sup[0].set_vv10_b(dftd3.dashcoeff['nl'][base_name.lower()]['b'])
```
PeterKraus: In my opinion, this should be handled in the EmpiricalDispersion class - we pass the `"dispersion" : {"type": "nl", "params": {"b": B, "c": C}, "citation": "STR"}` in, analogously to all other functionals, and the class figures it out. The `if`-`else` on lines 75-106 of empirical_dispersion.py will need a third fork for `nl`.

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -182,6 +187,23 @@ def build_superfunctional(name, restricted):
...
+                sup[0].set_vv10_c(dftd3.dashcoeff['nl'][base_name.lower()]['c'])
+                if (core.has_option_changed("SCF", "DFT_DISPERSION_PARAMETERS")):
```
PeterKraus: As above; they could be passed into the `tuple_params` argument to empirical dispersion class, and then processed inside.

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -87,22 +85,23 @@ def __init__(self, alias, dtype, **kwargs):
...
                 self.dash_params = {}
                 for k, v in zip(['s6', 'p1', 'p2', 'p3'], tuple_params):
                     self.dash_params[k] = v
 
-
         # Build the C++ dispersion class
         if self.disp_type == 'p4':
             self.disp = core.Dispersion.build(self.dtype, **self.dash_params)
         else:
             self.disp = None
```
PeterKraus: This probably needs expanded so that the `compute_things()` functions below don't have to return 0.

psi4/driver/qcdb/dashparam.py
```diff
@@ -210,6 +210,21 @@
...
+        'revpbe'      : {'b': 3.700, 'c': 0.0093},
+        'b3lyp'       : {'b': 4.800, 'c': 0.0093},
+        'b3pw91'      : {'b': 4.500, 'c': 0.0093},
+        'revpbe0'     : {'b': 4.300, 'c': 0.0093},
+        'bp86'        : {'b': 4.400, 'c': 0.0093},
+        'pbe0'        : {'b': 6.900, 'c': 0.0093},
+        'pbe'         : {'b': 6.400, 'c': 0.0093},
+        'tpss0'       : {'b': 5.500, 'c': 0.0093},
+        'dsd-pbepbe'  : {'b': 9.600, 'c': 0.0093},
+        'dsd-pbeb95'  : {'b': 12.500, 'c': 0.0093},
```
PeterKraus: These three will have to be defined in `dict_dh_funcs` manually, unless they happen to use the dispersionless variant anyway!

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -187,19 +203,36 @@ def __init__(self, alias, dtype, **kwargs):
...
+
     def print_out(self, level=1):
 
         core.print_out("   => %s: Empirical Dispersion <=\n\n" % self.dtype)
         core.print_out(self.description + "\n")
 
         core.print_out(self.citation + "\n\n")
+       
+        if self.disp_type=='nl':
+           return
```
PeterKraus: Don't we want to print out the B, C params? Or do they get printed out somewhere else anyway?

psi4/driver/qcdb/dashparam.py
```diff
@@ -309,6 +310,26 @@ def get_dispersion_aliases():
...
+        'pbe0'        : {'b': 6.900, 'c': 0.0093},
+        'pbe'         : {'b': 6.400, 'c': 0.0093},
+        'tpss0'       : {'b': 5.500, 'c': 0.0093},
+        'tpss'        : {'b': 5.000, 'c': 0.0093},
+        'dsd-pbepbe'  : {'b': 9.600, 'c': 0.0093},
+        'dsd-pbeb95'  : {'b': 12.500, 'c': 0.0093},
+        'dsd-pbep86'  : {'b': 12.800, 'c': 0.0093},
+        'b2gpplyp'    : {'b': 9.900, 'c': 0.0093},
+        'b2plyp'      : {'b': 7.800, 'c': 0.0093}, # 10.1021/acs.jctc.5b00002
+        'dsd-blyp'    : {'b': 12.000, 'c': 0.0093}, #https://pubs.acs.org/doi/10.1021/ct500642x
```
PeterKraus: Any of these parametrised in "Hujo, W.; Grimme, S; (2011), J. Chem. Theory Comput.; 7:3866"? If not, I'd add: 
`... "citation": '     citation in std format, as above' },` 
to each of the dicts, so that the `6b) custom citation` section in EmpiricalDispersion picks them up.

psi4/driver/procrouting/proc.py
```diff
@@ -1227,7 +1227,7 @@ def scf_helper(name, post_scf=True, **kwargs):
         core.set_legacy_wavefunction(ref_wfn)
 
         # Compute dftd3
-        if "_disp_functor" in dir(ref_wfn):
+        if "_disp_functor" in dir(scf_wfn) and not ref_wfn._disp_functor.excludelist:
             disp_energy = ref_wfn._disp_functor.compute_energy(ref_wfn.molecule())
             ref_wfn.set_variable("-D Energy", disp_energy)
         ref_wfn.compute_energy()
```
PeterKraus: I still think this is the wrong place to do these checks. We're printing the "Empirical Dispersion Energy =" for all DFT's anyway, as it seems to be always printed; otherwise it's always ignored:

https://github.com/psi4/psi4/blob/a7fc0506050facb3b544cc4200f708a5dd63c56a/psi4/src/psi4/libscf_solver/hf.cc#L1825-L1829

The EmpiricalDispersion class can, IMO, return 0 for `-NL` safely - unless we want to be ready for when somebody writes a method using both -D3 and -NL at the same time!

The more exceptions we add here, the more often we'll have to touch `proc.py`. I'm sure '-NL' is not the last - I believe `-D4` will need its own handling, as it calculates partial charges on the fly.

psi4/driver/qcdb/dashparam.py
```diff
@@ -309,6 +310,26 @@ def get_dispersion_aliases():
         'pbe0'        : {'s6': 1.000, 's8': 0.528823, 'a1': 0.007912, 'a2': 6.162326},
         'lcwpbe'      : {'s6': 1.000, 's8': 0.906564, 'a1': 0.563761, 'a2': 3.593680},
     },
+    'nl': { # TODO: citations
```
PeterKraus: I am wondering about `-rVV10`: you mentioned there is a third parameter which is the only difference between `-V` and `-rV` in eg. B97M-[r]V?

psi4/driver/procrouting/empirical_dispersion.py
hokru: They are printed out in the functional part. Once is enough.

psi4/driver/procrouting/proc.py
hokru: Short-range NL with long-range D3, why not? I am indifferent to the solution. I thought this is smoother and the exclude list can easily be expanded.

I assume -D4 will all be handled by an external program, since gfn-xtb data is recommended as default ingredient iirc.

psi4/driver/qcdb/dashparam.py
hokru: The vv10 kernel is simplified, the external parameters remain the same i think. Would need to check.

psi4/driver/qcdb/dashparam.py
hokru: only partially, I have to dig out the citations.

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -108,6 +108,32 @@ def build_superfunctional(name, restricted):
...
+        sup[0].set_vv10_b(vv10_b)
+        sup[0].set_vv10_c(vv10_c)
+    elif core.has_option_changed("SCF", "DFT_VV10_B"):
+        if(name.lower()=='hf'):
+           raise ValidationError("SCF: HF with -NL not implemented")
+        vv10_b = core.get_option("SCF", "DFT_VV10_B")
+        sup[0].set_vv10_b(vv10_b)
+        if (abs(sup[0].vv10_c() - 0.0) <= 1e-8):
+            core.print_out("SCF: VV10_C not specified. Using default (C=0.0093)!")
+            sup[0].set_vv10_c(0.0093)
```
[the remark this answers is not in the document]
hokru: It is for just provided vv10_b for an unknown functional and using the default C, which is the typical use case.

Different mechanisms are about user expectation. One can set omega manually, so also vv10_b/c should be possible. And since it behaves like -D in many ways, one would expect a similar functionality.

psi4/driver/procrouting/proc.py
hokru: Seems I am breaking the psi4-side D2 detection with it as well.

PeterKraus: Part of your tests fail because you're using `scf_wfn` instead of `ref_wfn`.

hokru: i know, but my excludelist function does not work for "p4". Getting too late today, though.

psi4/driver/qcdb/dashparam.py
```diff
@@ -309,6 +310,26 @@ def get_dispersion_aliases():
...
+        'pbe0'        : {'b':  6.900, 'c': 0.0093, 'citation' : '    M. K. Kesharwani, A. Karton, J.M. L. Martin, J. Chem. Theory Comput. 12, 444-454, 2016 \n'}, # unclear if this is the real origin
+        'pbe'         : {'b':  6.400, 'c': 0.0093, 'citation' : '    M. K. Kesharwani, A. Karton, J.M. L. Martin, J. Chem. Theory Comput. 12, 444-454, 2016 \n'}, # unclear if this is the real origin
+        'tpss0'       : {'b':  5.500, 'c': 0.0093, 'citation' : ' W. Hujo, S. Grimme, J. Chem. Theory Comput. 9, 308-315, 2013 \n'},
+        'tpss'        : {'b':  5.000, 'c': 0.0093, 'citation' : ' W. Hujo, S. Grimme, J. Chem. Theory Comput. 9, 308-315, 2013 \n'},
+        #'dsd-pbeb95'  : {'b': 12.500, 'c': 0.0093},
+        'b2gpplyp'    : {'b':  9.900, 'c': 0.0093},
+        'b2plyp'      : {'b':  7.800, 'c': 0.0093, 'citation': '    J. Calbo, E. Orti, J. C. Sancho-Garcia, J. Arago, J. Chem. Theory Comput. 11, 932-939, 1015 \n'}, 
+        'dsd-blyp'    : {'b': 12.000, 'c': 0.0093, 'citation': '    F. Yu J. Chem. Theory Comput. 10, 4400-4407, 2014 \n'},
```
PeterKraus: Shouldn't this one be defined as a dict_func too?

hokru: you are right! Changed in the next commit.

hokru:
Oh sweet madness that is everything DSD double-hybrid...

All DSD-DHDF-NL versions I put in will use the frozen-core parameters and will also use the D3BJ variants. No matter what others have done. This makes the most sense and brings some needed consistency. The description does mention what is used.

The impact of these small variations are very minimal from my experience anyway. No thorough testing was done of course. I think it is less severe than different LDA versions for GGAs, where are often ignored.

I hope everyone agrees.

For the sake of future users all DHDF should use frozen-core by default, imo. This is done already for standard DF-MP2, but I cannot find how.

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -108,6 +108,33 @@ def build_superfunctional(name, restricted):
...
```
PeterKraus: The empirical dispersion counterpart of this has a rather extensive testing suite in the `dftd3/energy` test. Is this multiple parameter-setting functionality for VV10 also covered in a test?

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -102,6 +103,11 @@ def __init__(self, alias, dtype, **kwargs):
                 if len(tuple_params) > 4:
                     raise Exception("Too many parameter in input tuple param.")
 
+        # DFT-NL dispersion. 
+        elif self.dtype in ["-NL"]:
+             self.disp_type = 'nl'
```
PeterKraus: If #986 gets merged before this, I've changed the suffixes to lowercase (-d2p4, -d3 etc.) to be consistent with the rest of the code. I guess the "-NL"'s scattered throughout the file should be lowercased too - I suppose you can fix that during rebase from master.

psi4/driver/qcdb/dashparam.py
```diff
@@ -309,6 +310,24 @@ def get_dispersion_aliases():
         'pbe0'        : {'s6': 1.000, 's8': 0.528823, 'a1': 0.007912, 'a2': 6.162326},
         'lcwpbe'      : {'s6': 1.000, 's8': 0.906564, 'a1': 0.563761, 'a2': 3.593680},
     },
+    'nl': { 
+        'blyp'        : {'b':  4.000, 'c': 0.0093, 'citation' : '    W. Hujo, S. Grimme J. Chem. Theory Comput. 7, 3866-3871, 2011 \n'}, 
+        'hf'          : {'b':  3.900, 'c': 0.0093, 'citation' : '    W. Hujo, S. Grimme J. Chem. Theory Comput. 7, 3866-3871, 2011 \n'}, # not implemented
```
PeterKraus: Not implemented?

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -330,7 +331,10 @@ def build_superfunctional_from_dictionary(name, npoints, deriv, restricted):
     if "dispersion" in functionals[name]:
         d_params = functionals[name]["dispersion"]
         if "citation" not in d_params:
-            d_params["citation"] = False
```
dgasmith: Small note, it looks like these are 3-space indented, not 4. Can you run this through YAPF?

psi4/driver/procrouting/dft_funcs/superfuncs.py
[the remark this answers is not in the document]
dgasmith: Maybe call `DFT_DISPERSION_PARAMETERS` `VV10_DISPERSION_PARAMETERS` or `NL_`?

If someone sets `DFT_DISPERSION_PARAMETERS` and `DFT_VV10_B` it could be a bit confusing about order of operations, best to throw I think?

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -264,6 +284,12 @@ def compute_hessian(self, molecule):
         #magic (if magic was easy)
         """
 
+        if self.disp_type in 'nl':
+             nat3=3*molecule.natom()
+             ZeroMat = np.zeros((nat3,nat3))
+             ZeroHess= core.Matrix.from_array(ZeroMat)
+             return ZeroHess
+
```
dgasmith: Does this still have to be routed through Empirical Dispersion? I think with the dictionary functional changes this can now be skipped.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -328,6 +330,13 @@ void HF::common_init()
         potential_ = nullptr;
     }
 
+    // post-scf vv10 correlation
+    if(options_.get_bool("DFT_VV10_POSTSCF")) {
```
dgasmith: Can you format these blocks? We normally use clang-format, but basically add spaces around operations, `if(` should be `if (`, and the code block should be indented by 4 spaces.

tests/dft-vv10/input.dat
```diff
@@ -17,7 +17,28 @@ set DFT_VV10_RADIAL_POINTS 20
...
+0.0238582087067217
+compare_values(0.0238582087067217,psi4.get_variable('DFT VV10 ENERGY') , 6, 'BLYP-NL custom b ') # TEST
+
+set DFT_VV10_B  4.0
+scf_e, scf_wfn = energy("BLYP", return_wfn=True)
+compare_values(0.032852665973,psi4.get_variable('DFT VV10 ENERGY') , 6, 'BLYP + custom VV10 ENERGY') # TEST
+
+set DFT_VV10_POSTSCF true
+scf_e, scf_wfn = energy("BLYP-NL", return_wfn=True)
+compare_values(0.032852665973,psi4.get_variable('DFT VV10 ENERGY') , 6, 'BLYP-NL postscf') # TEST
```
dgasmith: This value looks identical to the self consistent one? Can you also add spaces after commas here.

psi4/driver/procrouting/empirical_dispersion.py
PeterKraus: I thought the agreement was to get NL in, and then tackle the whole Dispersion - QCDB split / rework later (perhaps for 1.3)?

dgasmith: I agree for sure, just curious if this was still needed with the `dict` changes that you made. I think we might have separated these issues out.

psi4/driver/procrouting/dft_funcs/superfuncs.py
hokru: I admit only in a minimal fashion.

tests/dft-vv10/input.dat
hokru: yes, I used the self consistent value to check if the deviation to the post-scf one is small, which is perhaps not in the spirit these tests are to be used.

psi4/driver/procrouting/empirical_dispersion.py
hokru: It can be partially avoided, i.e. still using its citation formatting powers, but avoiding the energy/gradient/hessian calls.


I do this by deleting the dispersion object `del wfn._disp_functor`, if this is acceptable.

hokru:
upcoming changes:

- [x] formatting request
- [x] modifying tests
- [x] NL_DISPERSION_PARAMETERS name
- [x] partially avoiding EmpiricalDispersion

psi4/driver/procrouting/dft_funcs/superfuncs.py
```diff
@@ -108,6 +108,36 @@ def build_superfunctional(name, restricted):
...
+    # add VV10 correlation to any functional or modify existing
+    # custom procedures using name 'scf' without any quadrature grid like HF will fail and are not detected
+    if (core.has_option_changed("SCF", "DFT_VV10_B") and core.has_option_changed("SCF", "DFT_VV10_C")):
+        if(name.lower()=='hf'):
+           raise ValidationError("SCF: HF with -NL not implemented")
+        vv10_b = core.get_option("SCF", "DFT_VV10_B")
+        vv10_c = core.get_option("SCF", "DFT_VV10_C")
+        sup[0].set_vv10_b(vv10_b)
+        sup[0].set_vv10_c(vv10_c)
+    elif core.has_option_changed("SCF", "DFT_VV10_B"):
```
PeterKraus: Wouldn't it be more readable if these ifs were merged:
```
+    if (core.has_option_changed("SCF", "NL_DISPERSION_PARAMETERS") and core.has_option_changed("SCF", "DFT_VV10_B")):
+        raise ValidationError("SCF: Decide between NL_DISPERSION_PARAMETERS and DFT_VV10_B !!")
+    elif (core.has_option_changed("SCF", "NL_DISPERSION_PARAMETERS") and sup[0].vv10_b() > 0.0):
+        ...
+    elif (core.has_option_changed("SCF", "DFT_VV10_B"):
+        if(name.lower()=='hf'):
+           raise ValidationError("SCF: HF with -NL not implemented")
+        vv10_b = core.get_option("SCF", "DFT_VV10_B")
+        if (core.has_option_changed("SCF", "DFT_VV10_C"):
+            vv10_c = core.get_option("SCF", "DFT_VV10_C")
+        elif sup[0].vv10_c() != 0.0:
+            vv10_c = sup[0].vv10_c()
+        else:
+            core.print_out("SCF: VV10_C not specified. Using default (C=0.0093)!")
+            vv10_c = 0.0093
+        sup[0].set_vv10_b(vv10_b)
+        sup[0].set_vv10_c(vv10_c)
```

The `if (abs(sup[0].vv10_c() - 0.0) <= 1e-8)` check below only for overriding functional-specific "B" but keeping functional-specific non-default "C"? The way I read the code, unless set by `dict_func`, the `vv10_c` is initialised as 0.0:

https://github.com/psi4/psi4/blob/d5b1e133eb4fd2bebd695e8520938f14743fdd16/psi4/src/psi4/libfunctional/superfunctional.cc#L71

Why compare against a threshold?

tests/dft-vv10/input.dat
```diff
@@ -27,3 +30,52 @@ scf_e, scf_wfn = energy("VV10", return_wfn=True)
 
 for k, v in bench.items():                        # TEST
     compare_values(v, psi4.get_variable(k), 9, k) # TEST
+scf_nl=psi4.get_variable('TOTAL SCF ENERGY')
+
+# dft-nl tests:
+
+set reference rks
+scf_e, scf_wfn = energy("BLYP-NL", return_wfn=True)
+compare_values(0.032852665973, psi4.get_variable('DFT VV10 ENERGY') , 6, 'BLYP-NL dashtest') # TEST
```
PeterKraus: Could you perhaps give the reference values sensible names and move them all to the top? Such as:
```
ref_blyp_nl_b_4 = 0.032852665973
ref_blyp_nl_b_5 = 0.0238582087067217
ref_blyp_nl_b_5_c_0_1 = 0.0246750
```

PeterKraus:
@dgasmith: I don't know the "citation", "description" and also ["dispersion"]["citation"] maximum lengths. However, it shouldn't be too difficult to "pad" all three of these strings with the right number of leading spaces, and throw a warning (or even raise an error) if they are too long. However, perhaps not in this PR? If you make an issue (and tell me the limits), I can tackle this next week.

psi4/driver/procrouting/dft_funcs/superfuncs.py
hokru: This is essentially testing if `c = 0`. The threshold avoids the "equal test" on two floats.

I noticed that the comments around the if block are outdated.
I can merge them, yes.

tests/dft-vv10/input.dat
hokru: ok

hokru:
after rebasing to #986 the functional names in the print out are lower case. Do we want to keep it, or make them upper case again?

`=> Composite Functional: bp86-d2gr <=`

loriab:
I think uppercase is preferred for the output file.

PeterKraus:
Yeah, I think in print outs (ie. sup.name) we want "uppercase" (in general it makes sense to set the "name" in the dictionary to the preferred spelling, ie. "wB97X-D" as opposed to all-uppercase). Internally, I've modified all lookup calls (which use "name" and "alias" entries) to lowercase, but I might've been overzealous. The rest of the code uses lowercase names of methods internally.

Then there are the "PBEh-3c" and "HF-3c" cases, where dftd3/gcp require the dashes to be stripped, but the functionals as defined in the papers have a dash.

In all changes to these upper/lower mechanics, the `gcp` and `dftd3` tests should be checked; `dft-smoke` is not enough.

hokru:
All test but these pass:
```
        368 - dft-custom-gga (Failed) (see https://github.com/psi4/psi4/issues/994)
        370 - dft-bench-interaction (Timeout)
        372 - pywrap-align (Failed)
        373 - pywrap-align-chiral (Failed)

```

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -342,10 +342,14 @@ def build_superfunctional_from_dictionary(func_dictionary, npoints, deriv, restr
...
             d_params["citation"] = False
+        if d_params["type"] == 'nl':
+            sup.set_vv10_b(d_params["params"]["b"])
+            sup.set_vv10_c(d_params["params"]["c"])
         dispersion = d_params
 
     sup.set_max_points(npoints)
     sup.set_deriv(deriv)
-    sup.set_name(func_dictionary["name"].lower())
+#    sup.set_name(func_dictionary["name"].lower())
```
dgasmith: Can you clean this?

doc/sphinxman/source/dftnl.rst
```diff
@@ -0,0 +1,73 @@
...
+   
+    set DFT_VV10_B 4.0
+    energy('blyp')
+
+The default `C` parameter will be used.
+
+Similar to |scf_DFT_DISPERSION_PARAMETERS| the tuple |scf_NL_DISPERSION_PARAMETERS| can used::
+    
+    set NL_DISPERSION_PARAMTERS [4.0]
+    energy('blyp')
```
loriab: Someday I want to make sure that an `energy('b3lyp')` doesn't produce a `B3LYP TOTAL ENERGY` that is anything but the certified B3LYP energy. But that involved checking alpha, -D, -NL, etc. so way out of the frame of this PR.

PR #921, opened 2018-02-14
Add Sinf to induction in sapt_jk_terms

jwaldrop107:
Add the option for SAPT to compute the second order exchange-induction term, coupled or uncoupled, without the Single-Exchange Approximation.

## Todos
* __User-Facing for Release Notes__
    - [ ] Add Sinf option to second order exchange-induction


dgasmith:
Thanks for submitting this. Can you edit the description of this PR to include a bit about what you are doing and possible issues that need to be taken care of? See [here](https://github.com/psi4/psi4/pull/908) as an example, similar to the template you were originally given.

psi4/driver/procrouting/sapt/sapt_jk_terms.py
```diff
@@ -417,6 +551,34 @@ def induction(cache, jk, do_print=True, maxiter=12, conv=1.e-8, do_response=True
                 core.print_out(print_sapt_var(name, ret[name], short=True))
                 core.print_out("\n")
 
+        # Exch-Ind without S^2
+        if Sinf:
+            cT_A = np.dot(cache["Cvir_A"], x_B_MOA.transpose()).dot(Tmo_AA).dot(cache["Cocc_A"].np.T)
```
dgasmith: Can you use the `chain_dot` function here as well?

psi4/driver/procrouting/sapt/sapt_jk_terms.py
```diff
@@ -389,6 +389,140 @@ def induction(cache, jk, do_print=True, maxiter=12, conv=1.e-8, do_response=True
...
+        jk.C_left_add(core.Matrix.chain_dot(cache["Cocc_A"], Tmo_AB))
+        jk.C_right_add(cache["Cocc_B"])
+    
+        jk.compute()
+    
+        J_AA, J_BB, J_AB = jk.J()
+        K_AA, K_BB, K_AB = jk.K()
+
+        # A <- B
+        EX_AA = V_B.clone()
```
dgasmith: Can you add an `_inf` to the names of the tensors so it is very clear that they differ from the S^2 tensors.

tests/sapt-exch-ind-inf/input.dat
```diff
@@ -0,0 +1,30 @@
+#! SAPT(DFT) aug-cc-pVDZ interaction energy between Ne and Ar atoms.
+
+Eref = {'Exch-Ind20,u (A<-B) (Sinf)':   0.00749701, # TEST
+        'Exch-Ind20,u (A->B) (Sinf)':   0.07677809, # TEST
+        'Exch-Ind20,u (Sinf)':          0.08427511, # TEST
+        'Exch-Ind20,r (A<-B) (Sinf)':   0.00839011, # TEST
+        'Exch-Ind20,r (A->B) (Sinf)':   0.08666563, # TEST
+        'Exch-Ind20,r (Sinf)':          0.09505574} # TEST
```
dgasmith: Please test the `S^2` values in this test as well.

psi4/driver/procrouting/sapt/sapt_jk_terms.py
```diff
@@ -438,74 +433,74 @@ def induction(cache, jk, do_print=True, maxiter=12, conv=1.e-8, do_response=True
...
-        J_AA, J_BB, J_AB = jk.J()
-        K_AA, K_BB, K_AB = jk.K()
+        J_AA_inf, J_BB_inf, J_AB_inf = jk.J()
+        K_AA_inf, K_BB_inf, K_AB_inf = jk.K()
 
         # A <- B
         EX_AA = V_B.clone()
         EX_AA.axpy(-1.00, core.Matrix.chain_dot(S, T_AB, V_B, trans=[False, True, False]))
         EX_AA.axpy(-1.00, core.Matrix.chain_dot(S, T_B, V_B))
-        EX_AA.axpy(2.00, J_AB)
```
dgasmith: I was thinking of renaming these `EX_AA` quantities as well.

psi4/driver/procrouting/sapt/sapt_jk_terms.py
```diff
@@ -417,6 +546,33 @@ def induction(cache, jk, do_print=True, maxiter=12, conv=1.e-8, do_response=True
...
+                trans=[False, True, True, True])
+
+            ind_ab_total = 2.0 * (cT_A.vector_dot(EX_AA) + cT_AB.vector_dot(EX_AB))
+            ind_ba_total = 2.0 * (cT_B.vector_dot(EX_BB) + cT_BA.vector_dot(EX_BA))
+            indexch_ab_inf = ind_ab_total - ind_ab
+            indexch_ba_inf = ind_ba_total - ind_ba
+        
+            ret["Exch-Ind20,r (A<-B) (Sinf)"] = indexch_ab_inf
+            ret["Exch-Ind20,r (A->B) (Sinf)"] = indexch_ba_inf
+            ret["Exch-Ind20,r (Sinf)"] = indexch_ba_inf + indexch_ab_inf
```
dgasmith: Can you call these quantities `(S^inf)` to match `(S^2)` as well.

dgasmith:
Right now we name `E^10_Exch (S^inf)` "Exchange" and `E^10_Exch (S^2)` "Exchange (S^2)" while `E^20_Exch-Ind (S^2)` is "Exchange-Induction" in the output which is not consistent. Historically this makes perfect sense; however, this PR makes everything a bit rougher because we are now adding the possibility of `E^20_Exch-Ind (S^inf)`. I believe there are plans to add `E^20_Exch-Disp (S^inf)` as well which can be computed in roughly the same time as the S^2 version.

We have three issues:
 - Do we start calling the current Exch-Ind (S^2) "Exch-Ind (S^2)"?
 - Do we always compute Exch-Ind (S^inf) as it is quite cheap?
 - What do we do with our "SAPT0" nomenclature? We currently define SAPT0 as: S^inf Exchange, S^2 Exchange-Induction, S^2 Exchange-Dispersion. What is the all S^inf version called?

@CDSherrill @loriab @jgonthier @konpat

loriab:
@dgasmith Uggh. And to think that DFT-SAPT is worse ...

For reference, alpha is presently incorporated into the psivar name (https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/psivardefs.py#L42).

The way I'd frame the question is:
   * Is `S^2`/`S^inf` an _option_ that modifies the infinite-precision value of the same physical quantity (like `PK`/`DF` `HF TOTAL ENERGY`)?
   * or Are `S^2`/`S^inf` terms distinctly different physical quantities computed (e.g., Ind20/Ind30 or singlet/triplet MP2 correlation energy).

* Another point, is there ever safe mixing of s2/sinf?
* Can we compute Sinf terms and incorporate them into post-0 SAPT?

dgasmith:
- No S^2 vs S^Inf is, in some ways, a different level theory.
- S^2/S^Inf are separate quantities
- Mixing is as safe as any other mixing in SAPT. Our current version is fine, but S^Inf should help more for charged complexes. Exch/Ind-Exch at S^Inf and Disp-Exch at S^2 is probably quite reasonable.
- Yes and no. S^Inf is limited to SAPT0 for the time being, higher levels could get quite complex.

@alston One more person to ping and can correct me: @ajmisquitta

dgasmith:
In `psi4/src/read_options.cc:847` can you add the following lines:

```
    /*- For SAPT(DFT) computes the S^inf Exchange-Induction terms !expert -*/
    options.add_bool("DO_IND_EXCH_SINF",false);
```

and in `psi4/driver/procrouting/sapt/sapt_proc.py:165` change the induction call there to:

```
       ind = sapt_jk_terms.induction(
            hf_cache,
            sapt_jk,
            True,
            maxiter=core.get_option("SAPT", "MAXITER"),
            conv=core.get_option("SAPT", "D_CONVERGENCE"),
            Sinf=core.get_option("SAPT", "DO_IND_EXCH_SINF"))
```

Recompile the code. You should now be able to run the test by setting `DO_IND_EXCH_SINF True` in the options block.

tests/sapt-exch-ind-inf/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(sapt-exch-ind-inf "psi;quicktests;sapt")
```
dgasmith: Can you remove this from quick tests.

tests/sapt-exch-ind-inf/input.dat
```diff
@@ -0,0 +1,36 @@
...
+  Ar 1 6.5
+  units bohr
+}
+
+set {
+    basis         aug-cc-pvdz
+    scf_type      df
+    sapt_dft_grac_shift_a 0.203293
+    sapt_dft_grac_shift_b 0.138264
+    Sinf          True
```
dgasmith: This shouldn't work as shown, can you remove the `Sinf` option here and change it for the one presented in the comments? Also can delete the GRAC shifts, they are not valid for SAPT(HF).

dgasmith:
Great, so that test case should work now?

jwaldrop107:
It should yes. It did when I tried it.

dgasmith:
As the code is in place and we have a long test for it I say lets go ahead and get this in. This is a good unit of work currently and I think the discussion on how to integrate this properly into Psi4 is a long one.

Issue #920, opened 2018-02-14
different parameter versions for DFT functionals with "dash correction"


hokru:
Title is kept general, but this is especially relevant for the newer DSD double hybrids. Issue made from short discussion in slacks.

For example: http://onlinelibrary.wiley.com/doi/10.1002/jcc.23391/full 
Different functional compositions for the exchange, correlation, mp2 parts for (among others) -D2 and -D3BJ.

The dashparam.py functionality cannot handle these multiple versions, but
`energy('dsd-pbepbe')`,  `energy('dsd-pbepbe-d2') `or `energy('dsd-pbepbe-d3bj')` should provide the proposed parameters automatically. 

A possible workaround is to define everything, including the e.g. the -D parameters, in the superfunctional object and either avoid the "dash" in the method name or avoid invoking the dashparam functionality for specific names.

PeterKraus:
The references for the dashparams are currently handled in ``empirical_dispersion.py``. It might be convenient to keep those as defaults, and pass overrides or supplements in the dashparam.py file.

There's already the case of "B97" and "B97-D" functionals, where B97 refers to the hybrid of Becke, and B97-D is just the GGA part + D2 dispersion correction of Grimme. Another weird case is "wPBE" and "LCwPBE"; it would be cool if we could support both variants (along with "M062X" and "M06-2X" etc). Perhaps the functional translator part needs expanding.

Is there a reason why the "functional parts" (HF/GGA X, MP2/SS/OS/GGA C) couldn't be overridden by contents of dashparam.py? We could check whether the dashparam contains the keys, and if yes, override the defaults using lines 186-203 in superfuncs.py...

dgasmith:
I was thinking it might be nice to have a dictionary that can define functionals rather than a function. For example `PBE0` could be defined like:

```python
func_dict["PBE0"] = {
    "name": "PBE0",
    "citation": ...,
    "x_functionals": [("XC_GGA_X_PBE", 0.75), ]
     "c_functionals": [("XC_GGA_C_PBE", 1.0), ]
    "set_x_alpha": 0.25,
    }
```

Here it would be much easier to programmatically build functionals. We could have it so that any valid SuperFunctional setter (`x_omega`, `vv10_c`, etc) could be used. This would also be very useful for the `-NL` functionals.

PeterKraus:
I had a bit more of a look and here's what I'd propose:

[...]

PR #922, opened 2018-02-18
Dict funcs

PeterKraus:
## Description
Added a dictionary-based method for implementing density functionals. All non-dispersion versions of original functionals are working with some exceptions, the new definition of functionals has "TEST-" prefix for now.

Not working functionals are: HF+D and HJS-B88 which do not run, HF3C for unknown reason, B97-D which should be -D2 not -D3 corrected in master.

Some dispersion corrected methods already implemented - all BLYP functionals, for example.

Script used for testing attached:
[test.dat.txt](https://github.com/psi4/psi4/files/1735039/test.dat.txt)

## Todos
* **Developer Interest**
  - [x] Fix for Issue #920 
  - [x] Easy implementation of new functionals
  - [x] LibXC parameters (ALPHA, BETA, OMEGA) now exposed to python
  - [x] Check functionals against Q-Chem
  - [ ] oPBE-D, oTPSS-D, oPWLYP-D ,oBLYP-D, TPSSh-D3BJ - not included in PR #880 
  - [x] B97-D
  - [x] wB97X-D3 - needs tweaks - was disabled.
  - [x] DSD-***
  - [x] [PBE]HF3C
  - [ ] Eventually: let user pass a dict to psi4 with custom functionals? 

## Questions
- [ ] What do you think about the current interface?

## Status
- [x] Dashparams for functionals need to be migrated too.
- [x] Ready to go - from my PoV at least.


psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
```diff
@@ -0,0 +1,142 @@
...
+#
+"""
+List of XC functionals
+"""
+
+
+b2plyp = {
+    "name": "B2PLYP",
+    "x_functionals": {"GGA_X_B88": {"alpha": 0.47}, "X_HF": {"alpha": 0.53}},
+    "c_functionals": {"GGA_C_LYP": {"alpha": 0.73}, "C_MP2": {"alpha": 0.27}},
```
dgasmith: Not entirely sure I like `X_HF` and `C_MP2` as part of the "functionals" dictionary. Mostly this is because these need to be handled completely separately internally and usually I think of these as completely different components. I can easily be convinced, but its just something that struck me.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
```diff
@@ -0,0 +1,142 @@
...
+
+b2gpplyp = {
+    "name": "B2GPPLYP",
+    "x_functionals": {"GGA_X_B88": {"alpha": 0.35}, "X_HF": {"alpha": 0.65}},
+    "c_functionals": {"GGA_C_LYP": {"alpha": 0.64}, "C_MP2": {"alpha": 0.36}},
+    "citation": '    A. Karton, A. Tarnopolsky, J.-F. Lamere, G. C. Schatz, J.M. L. Martin, J.Phys. Chem. A, 112, 12868-12886,2008   \n',
+    "description": '    B2GPPLYP Double Hybrid Exchange-Correlation Functional\n',
+}
+
+beta = 5.0 * (36.0 * 3.141592653589793)**(-5.0 / 3.0)
```
dgasmith: We need to do something to namespace this bit. As a first thought, can you build a function that looks like:

```python
def pwpb95_pw91_tweaks():
    ....
   return [a_pw6, b_pw6, c_pw6, d_pw6, f_pw6, alpha_pw6, expo_pw6]

 {"GGA_X_PW91": {"tweak": pwpb95_pw91_tweaks() ,"alpha": 0.50}
```

psi4/driver/procrouting/dft_funcs/dict_gga_funcs.py
```diff
@@ -0,0 +1,194 @@
...
+    "name": "BLYP",
+    "x_functionals": {"GGA_X_B88": {}},
+    "c_functionals": {"GGA_C_LYP": {}},
+    "citation": '    P.J. Stephens et. al., J. Phys. Chem., 98, 11623-11627, 1994\n    B. Miehlich et. al., Chem. Phys. Lett., 157(3), 200-206 1989\n',
+    "description": '    BLYP GGA Exchange-Correlation Functional\n',
+}
+
+blyp_d2 = copy.deepcopy(blyp)
+blyp_d2["name"] = "BLYP-D2"
+blyp_d2["dispersion"] = {"type": "d2", "params": {"s6": 1.20}}
```
dgasmith: You could consider building these in a loop like the following:

```python
blyp_dashparams = {{"d2": {"s6": 1.20}, ...}

for dtype, params in blyp_dashparams.items():
    data = copy.deepcopy(blyp)
    data["dispersion"] = {"type": dtype , "params":params}
    functional_list["BLYP-" + params.upper()] =  data
```

psi4/driver/procrouting/dft_funcs/dict_gga_funcs.py
```diff
@@ -0,0 +1,194 @@
...
+    "TEST-PW91": pw91,
+    "TEST-FT97": ft97,
+    "TEST-BOP": bop,
+    "TEST-MPWPW": mpwpw,
+    "TEST-SOGGA11": sogga11,
+    "TEST-SOGGA": sogga,
+    "TEST-OPBE-D"    : {"name": "OPBE-D"   , "xc_functionals": {"GGA_XC_OPBE_D"   : {}}, "dispersion": {"type": "d3zero", "params": {'s6': 1.0,  's8': 1.494, 'sr6': 1.128, 'alpha6': 14.0}}},
+    #"TEST-OPWLYP_D"  : {"name": "OPWLYP-D" , "xc_functionals": {"GGA_XC_OPWLYP_D" : {}}}, # no dispersion
+    #"TEST-OBLYP_D"   : {"name": "OBLYP-D"  , "xc_functionals": {"GGA_XC_OBLYP_D"  : {}}}, # no dispersion
+    "TEST-HCTH407P" : {"name": "HCTH/407P", "xc_functionals": {"GGA_XC_HCTH_407P": {}}},
```
dgasmith: Might be nice to automatically build the `xc_functionals` list as well like we do in `xc_funcs.py`.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
```diff
@@ -0,0 +1,244 @@
...
+
+mn15 = {
+    "name": "MN15",
+    "description": '   MN15 Hybrid Meta-GGA Exchange-Correlation Functional\n',
+    "citation": '    H. S. Yu, X. He, S. L. Li, and D. G. Truhlar, Chem. Sci. 7, 5032-5051, 2016\n',
+    "x_functionals": {"HYB_MGGA_X_MN15": {"use_libxc": True}},
+    "c_functionals": {"MGGA_C_MN15": {}},
+}
+
+beta = 0.0018903811666999256  # 5.0*(36.0*math.pi)**(-5.0/3.0)
```
dgasmith: Ditto here on the namespace like above.

psi4/driver/procrouting/proc.py
```diff
@@ -1017,8 +1017,12 @@ def scf_wavefunction_factory(name, ref_wfn, reference):
         raise ValidationError("SCF: Unknown reference (%s) when building the Wavefunction." % reference)
 
     if disp_type:
-        wfn._disp_functor = empirical_dispersion.EmpericalDispersion(
-            disp_type[0], disp_type[1], tuple_params=modified_disp_params)
+        if len(disp_type) > 2:
```
dgasmith: If possible, it might be nice to be able to check this off keys rather than length.

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,186 @@
...
+    # require explicit X_HF or C_MP2 with defined alpha for complex definitions
+    #if len(components) > 1 and "X_HF" not in components.keys() and "C_MP2" not in components.keys():
+    #    return(False)
+    if "X_HF" in components.keys() and "alpha" not in components["X_HF"].keys():
+        return(False)
+    if "C_MP2" in components.keys() and "alpha" not in components["C_MP2"].keys():
+        return(False)
+    return(True)
+
+def build_superfunctional_from_dictionary(name, npoints, deriv, restricted):
```
dgasmith: Once we settle on this, can we add a long doctoring to this to explain what the new format looks like.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
PeterKraus: We could change it to:
```
    ...
    "x_functionals": {"GGA_X_B88": {"alpha": 0.47}}, 
    "x_hf": {"alpha": 0.53},
    "c_functionals": {"GGA_C_LYP": {"alpha": 0.73}}, 
    "c_mp2": {"alpha": 0.27},
    ...
```
or even:
```
    "x_functionals": {"HYB_GGA_X_SOGGA11_X": {"use_libxc": True}}, # sets omega for GGA from libxc, alpha=1.0
    "x_hf": {"use_libxc": "HYB_GGA_X_SOGGGA11_X"}, # sets alpha, beta, omega for HF from libxc
    "c_functionals": {"GGA_C_SOGGA11_X": {}},
```

The question then remains what to do if one asks for a hybrid with `"use_libxc": True` and doesn't provide the `x_hf` key - do we add HF exchange as specified by libxc (even though the ALPHA actually applies for `x_hf` not for the GGAs in `x_functionals`), or do we just run it without HF exchange?

psi4/driver/procrouting/dft_funcs/dict_gga_funcs.py
PeterKraus: Is there a reason why you want to keep `xc_functionals` separate from the others? I basically just went through the `xc_funcs_list` and split them across the other files, so that when one looks for a hybrid there's only one place where it's defined, not two.

dgasmith: Mostly because the XC list is either copied as a single block from LibXC itself or in the future we will query the library to see what XC funcs it has. Keeping it all in separate files will make it hard to figure out where new functionals go as LibXC adds new functionals at quite a ferocious rate to keep up with literature.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
dgasmith: I do like the change in format, but you are right we introduce some ambiguity.

I really wish that the SOGGA functionals (for example) just came as full XC functionals rather than the pieces. I dont quite understand why they deprecated the M05-2X macros and you now have to rebuild them on your own in LibXC 4.0+. This puts a lot of onus on the downstream users (us) to make sure this is correctly specified.

To make the parsing simpler I would do `{"HYB_GGA_X_SOGGA11_X": {"use_libxc": True}}` and then throw an error if `x_hf` has any data in it or if `x_functionals` has more than one functional in it. AFAIK there is no current functional that will violate this rule and we can handle it when the time arrises. 

I would say that the deprecation of XC functional macros (M05-2X and similar) puts us in a position of continuing to need expansive custom functions lists like this. Unless I am missing something?

PeterKraus:
OK, 2nd iteration. I removed most dispersion-corrected functionals for now - the plan is to have the "standard" dispersion funcs generated from the content of `dashparams.py`, the non-standard ones which need changes in the underlying XC terms (DSD- and wB97X-D3) are/will be defined explicitly (along with `XC_[...]_D` functionals).

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,207 @@
...
+                x_params["alpha"] = 0.0
+            if "beta" in x_params.keys():
+                sup.set_x_beta(x_params["beta"])
+            if "omega" in x_params.keys():
+                sup.set_x_omega(x_params["omega"])
+        else:
+            sup.set_x_alpha(x_HF["ALPHA"] + x_HF["BETA"])
+            sup.set_x_beta(x_HF["ALPHA"] - x_HF["BETA"])
+            sup.set_x_omega(x_HF["OMEGA"])
+        if "c_functionals" in dict_functionals[name].keys():
```
dgasmith: Can you add spaces between this `if` block and the previous along with a comment to describe what each ones does?

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
```diff
@@ -0,0 +1,442 @@
...
+        "citation": "    S. Kozuch, J.M.L. Martin, J. Comp. Chem., 34, 2327-2344, 2013\n"
+    },
+})
+
+funcs.append({
+    "name": "CORE-DSD-BLYP",
+    "x_functionals": {"GGA_X_B88": {"alpha": 0.31}},
+    "x_hf": {"alpha": 0.69},
+    "c_functionals": {"GGA_C_LYP": {"alpha": 0.54}},
+    "c_mp2": {"alpha": 1.0, "os": 0.46, "ss": 0.37},
```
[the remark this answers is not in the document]
PeterKraus: We do, for normal DH functionals (like B2PLYP and PBE0-2). It is true that the presence of `"ss"` and `"os"` currently implies `"alpha" = 1`, but who knows what turns up later.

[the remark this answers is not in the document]
dgasmith: Hmm, ok this is a bit awkward in how this works. If we are a `scs_hybrid` we use the `c_os/c_ss alpha`, otherwise we just add in the `MP2 correlation energy * alpha`. Currently we determine scs_hybrid by the following:

```
    bool is_c_scs_hybrid() const { return c_os_alpha_ != 0.0 || c_ss_alpha_ != 0.0; }
```

So the following will give the same result for any value of alpha.
```
"c_mp2": {"alpha": 0.5, "os": 0.46, "ss": 0.37}
```

What is your personal view of how this should work regardless of how the code currently is (we can fix that later)?

PeterKraus:
Dear all, I had another bash at this. It's rather a large amount of changes, but here's the current workflow:

- functionals get defined in `dict_*.py`, including aliases for the base names (M06-2X = M062X, wPBE = LC-WPBE = LCWPBE etc.)
- special cases of dispersion-corrected functionals are defined there (B97-D and friends, OPBE-D, HF3C etc.)
- all this stuff gets combined into a single dict `dict_functionals` in `dict_builder.py`.
- aliases for dispersion names (D2 = D = D2P4) are defined as `dispersion_names` (VV10 should go here)
- we loop through the proto-functionals in `dict_functionals`, figure out all base functional aliases, add them into the final dict `functionals`
- if the base functional is *not* dispersion corrected, we look through `psi4.driver.qcdb.dashparam.dashcoeff` for a matching dispersion correction; we match all base aliases (so M062X-D3 and M06-2X-D3ZERO are equivalent)
- the newly generated dispersion corrected functional gets added (incl. aliases)
- the `build_superfunctional` function in `superfuncs.py` now has to find the requested functional in `dict_builder.functionals`, if there is a match, we build the `(sup, dispersion)` tuple
- the `disp_type` assigned in `proc.py` is a `dict()`, so we don't need to match against length
- we pass this onto an `EmpiricalDispersion()` constructor

The handling of the latter two points is not quite figured out yet. I'm sure it could be simplified.

As for tests, compared to `master`, the following fail:
```
OPBE-D: wrong dispersion
B97-D: wrong dispersion (B97-D2, -D3 and other variants work)
OTPSS-D: wrong dispersion
HF3C, HF3C-D3BJ: wrong dispersion
DSD-* and co: expected to fail
WB97X-D3 and -D3ZERO: expected to fail
```
I will try to figure those out later, but I'm assuming it is just the dispersion component that is misassigned.

psi4/driver/procrouting/proc.py
```diff
@@ -1017,8 +1017,12 @@ def scf_wavefunction_factory(name, ref_wfn, reference):
         raise ValidationError("SCF: Unknown reference (%s) when building the Wavefunction." % reference)
 
     if disp_type:
-        wfn._disp_functor = empirical_dispersion.EmpericalDispersion(
-            disp_type[0], disp_type[1], tuple_params=modified_disp_params)
+        if type(disp_type) == type(dict()):
```
dgasmith: Can you use `isinstance(disp_type, dict)` here?

dgasmith:
This continues to look better and better IMO. Please do poke me if you need help on specific parts.

For HF3C do you know the error here? Is it just a an issue with the correction dispersion.

PeterKraus:
@dgasmith: the HF3C fail is weird - the dispersion is set correctly, but something about the functional is wrong. Hence the failing tests. I think the problem is related to dispersion, or GCP - lines 195 - 208 in `EmpiricalDispersion` would be my guess.

The OPBE-D and OTPSS-D: I need to check which dispersion they are actually meant to have. In master, they currently have none, as far as I see.

dgasmith:
Hmm, might want to just print the alias when you call energy to see what is printed out.

For the other two functionals we popped them out of the functional list as we do not have dispersion terms from them I think.

PeterKraus:
I think this is on the final stretch. Once the new tests are in, I'll go through the functionals again and make sure they match Q-Chem.

I don't particularly like the way YAPF formats the dictionaries - sometimes the name is on the next line, which is a bit odd. 

Also, the B97-D case is weird. It seems that in `master`, calling B97-D doesn't apply dispersion at all!

psi4/driver/procrouting/proc.py
PeterKraus: Done.

psi4/driver/procrouting/dft_funcs/dict_builder.py
PeterKraus: Did you have anything in particular in mind?

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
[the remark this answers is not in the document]
PeterKraus: I'm not knowledgeable enough about XC functionals to judge that. In my opinion, it doesn't hurt to have `alpha` explicit, but then it shouldn't get overridden in the code... Perhaps @susilehtola has an opinion?

loriab:
I won't have time to look at this is detail until the end of the month, but from skimming, these look like very nice changes. I'm modifying how `dftd3` gets called, so that should work in well later on. If any particular `yapf` lines are troubling, just add `  # yapf: disable`.

PeterKraus:
Just a quick note on the failing tests so I don't forget: `dft-smoke` fails because it includes DSD-PBEP86 (which was parametrised for -D2, now is parametrised without it); `dftd3-energy` fails because of `dashparams['sr8']` not being passed in. It would make sense to generalise that interface somewhat - set defaults and then update from the passed-in dict.

PeterKraus:
@loriab @dgasmith Thanks for merging PR #880, the tests are very helpful. All of the original funcs pass with the dict_func patch - only wB97X doesn't match the old pre-libxc psi4 value. Unfortunately, wB97X-D3 requires tweaks to the underlying functional and I'm not sure whats wrong with O3LYP, both of which are expected fails against Q-Chem. 

From the funcs I implemented in the meantime (SOGGA, SOGGA11, revPBE, MN12-L, M06-L, MN15-L, SOGGA11-X, MN12-SX), all pass against Q-Chem and only SOGGA fails against old psi4, but we knew that already.

A lot of the other functionals, especially DSD-hybrids, do not have a reference value for the water dimer. This also crashes the dft-smoke test, as DSD-PBEP86 is part of it. Gaussian 2016 implements the "Kozuch 2011" version of DSD-PBEP86, I am not sure it includes any dispersion in their form: it would be prudent to have our implementation of that one match G16. Alternatively, we can stick to Kozuch's 2013 paper (where he supplies non-, D2-, D3- and D3BJ- forms of at least 10 DSD functionals); but then we're on our own. 

In any case I suggest replacing DSD-PBEP86 in `dft-smoke` by a more widely used functional - B2PLYP or PBE0-2 or PBE0-DH come to mind; even though they are not spin-component-scaled.

loriab:
I'm recalling something about pre-libxc psi4 having a wrong coeff for wB97X omega (0.3 vs 0.4); it affected wB97X but _not_ wB97X-D. I'd have to look up when (and if, considering Libxc may have lead to the discovery) it was corrected, but that's a possibility for that discrepancy.

I think the idea for dft-smoke was to have one of each category, so a different DSD-DH would be good, if there any that aren't contentious. That DSD-PBEP86 helped me fix my bug just this weekend when I did something only affecting SS/OS.

Not sure what to say on the Kozuch 2011/2013. Both matching Gaussian and using dispersion have merit. Maybe DSD-PBEP86-2011 for the former and normal names for the latter? I take it Kozuch 2013 isn't obliging in the way of suppmat?

hokru:
i remember the omega issue from a forum post: http://forum.psicode.org/t/wb97x-definition-bug/480

I think ORCA was used for all the DSD developments. There you can defined DHDF compositions in the input file (see e.g. my reference outputs). If manual defining parameters for another program really counts as a good test is perhaps debatable because typos never come alone ;)

Are the non-dispersion forms of DSD-PBEP86 from both papers really different? I hope not...
Anyway, I'd stick with the 2013 paper in doubt because it is the latest "update".

PeterKraus:
Yes, they are different. To add insult to injury, the G16 manual references DSD-PBEP86 as [Kozuch2011](http://dx.doi.org/10.1039/C1CP22592H), however the implementation is the DSD-PBEP86-D3BJ version (including dispersion!) from the [SI](http://onlinelibrary.wiley.com/wol1/doi/10.1002/jcc.23391/suppinfo) in [Kozuch2013](http://dx.doi.org/10.1002/jcc.23391):

```
code:      G16    K. 2011  K. 2013
gga_x:     0.31    0.30     0.31
gga_c:     0.44    0.43     0.44
mp2_ss:    0.22    0.25     0.22  
mp2_os:    0.52    0.53     0.52
d3bj_s6:   -??-    0.418    0.48
d3bj_a2:   -??-    5.65     5.6
d3bj_s8:   -??-    0.0      0.0
d3bj_a1:   -??-    0.0      0.0
------------------------------------
code:      G16                  K. 2013
E_GGA:    -76.1911169141   -76.1910816827
E_D3(BJ):  -0.0002162453    -0.0002162500
E_MP2:     -0.1114294518    -0.1124581459
E_DSD:    -76.3025463659   -76.3035398286
```
G16 output is cryptic enough that it doesn't show the dispersion coefficients, but the D3(BJ) component is almost the same, so I'm assuming it's the 2013 values. The disagreement with G16 is already at the GGA component, which is somewhat annoying - hopefully it's just numerics.

The water cation test passes to 4 D.P. (12 meH difference):

````
         G16            psi4 Kozuch2013
neutral -76.2031337     -76.2035504
cation  -75.7571549     -75.7575598
IE        0.4459788       0.4459906   
````

I'd suggest:
1) use Kozuch's 2013 functionals throughout and forget the 2011 paper exists
2) use DSD-PBEP86-D3BJ as a test in `dft-smoke`
3) set DSD-PBEP86 and DSDPBEP86 to return the formulation without dispersion correction (unlike G16)

**Comment by CDSherrill** [comment]
I have vague memories about some of the DSD methods changing parameters
from one paper to another.  I discussed this with Jan Martin way back when
via email, and thought the situation was somewhat confusing.  Without
remembering this any more clearly than I do, I suppose Holger's idea of
using the more recent paper is probably a good one.

**Comment by CDSherrill** [comment]
Weird, the GGA agreement is pretty good (all things considered), but then
the MP2 is off by a millihartree, possibly (presumably?) because the
orbitals are different from the GGA.  Could all just be numerics.

It would be nice to be able to get the non-D-corrected energies... however,
would a user asking for  DSD-PBEP86 be expecting a -D correction to be
automatically included?  (I don't remember enough about the method).

**Comment by hokru** [comment]
For the dft-custom-gga test I used Turbomole's PBE value because G09's one was too different. It just really sometimes matters a lot to which decimal points all the constant and conversion factors are defined. Don't know about q-chem.

Just to have it asked: same frozen core treatment in MP2?

> would a user asking for  DSD-PBEP86 be expecting a -D correction to be
automatically included? 

I dont think so.

**Comment by PeterKraus** [comment]
So, the difference in the MP2 part was due to frozen core - I did forget that in psi4 it's not on by default. With `set freeze_core true` I get:
````
E_GGA:   -76.1910816826976429 (G16: -76.1911169141)
E_D3(BJ): -0.0002162500000000 (G16:  -0.0002162453)
E_MP2:    -0.1114064924058353 (G16:  -0.1114294518)
E_TOT:   -76.3024881751034769 (G16: -76.3025463659)
````
Which seems good enough agreement. 

From my point of view, this is good to go.

**Comment by dgasmith** [comment]
Cool! I really like these new changes. Ill do a deep review this weekend. Thank you for all of the work!

**Comment by CDSherrill** [comment]
Agreement looking good

File: psi4/driver/procrouting/dft_funcs/dict_builder.py [hunk 11]
```diff
@@ -0,0 +1,242 @@
...
+    of the functional, including detecting duplicate requests for LibXC params, inconsistent
+    requests for HF exchange and missing correlation.
+    """
+    # 1a) sanity checks definition of xc_functionals
+    if "xc_functionals" in func_dictionary.keys():
+        if "x_functionals" in func_dictionary.keys() or "x_hf" in func_dictionary.keys():
+            raise ValidationError("SCF: Duplicate specification of exchange (XC + X) in functional %s." % (name))
+        elif "c_functionals" in func_dictionary.keys() or "c_mp2" in func_dictionary.keys():
+            raise ValidationError("SCF: Duplicate specification of correlation (XC + C) in functional %s." % (name))
+    # 1b) require at least an empty exchange functional entry or X_HF
```
dgasmith: Can you add a space above a comment/code section to break up this code a bit for readability?

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
```diff
@@ -0,0 +1,539 @@
...
+    "x_hf": {
+        "alpha": 0.75
+    },
+    "c_functionals": {
+        "GGA_C_LYP": {
+            "alpha": 0.53
+        }
+    },
+    "c_mp2": {
+        "alpha": 1.0,
```
dgasmith: You probably want to remove `alpha` from all `c_mp2` blocks. The two do not interact.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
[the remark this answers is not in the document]
dgasmith: From above, lets just remove alpha for now from these dicts.

psi4/driver/procrouting/dft_funcs/dict_builder.py
PeterKraus: Fixed. I've added another docstring further below.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
PeterKraus: Done.

psi4/driver/procrouting/proc.py
```diff
@@ -1446,20 +1462,22 @@ def run_dcft(name, **kwargs):
...
-        aux_basis = core.BasisSet.build(ref_wfn.molecule(), "DF_BASIS_DCFT",
-                                        core.get_global_option("DF_BASIS_DCFT"),
+        aux_basis = core.BasisSet.build(ref_wfn.molecule(), "DF_BASIS_DCFT", core.get_global_option("DF_BASIS_DCFT"),
                                         "RIFIT", core.get_global_option("BASIS"))
         ref_wfn.set_basisset("DF_BASIS_DCFT", aux_basis)
 
-        scf_aux_basis = core.BasisSet.build(ref_wfn.molecule(), "DF_BASIS_SCF",
-                                            core.get_option("SCF", "DF_BASIS_SCF"),
-                                            "JKFIT", core.get_global_option('BASIS'),
-                                            puream=ref_wfn.basisset().has_puream())
```
PeterKraus: I'm not sure if I'm doing something wrong here, but I didn't touch this part of the code - after running `yapf --style .style.yapf psi4/driver/procrouting/proc.py -i` it splits these function calls differently!

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
```diff
@@ -0,0 +1,577 @@
...
+        }
+    },
+    "c_mp2": {
+        "alpha": 0.27
+    },
+    "citation":
+    '    S. Grimme, J. Chem. Phys., 124, 034108, 2006\n',
+    "description":
+    '    B2PLYP Double Hybrid Exchange-Correlation Functional\n',
+})# yapf: disable
```
PeterKraus: I've added `yapf: disable`'s to the dicts so they remain formatted somewhat nicely.

psi4/driver/procrouting/dft_funcs/dict_xc_funcs.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+List of XC functionals
+"""
+
+import copy
+
+funcs = []
+
+# yapf: disable
+funcs.append({"name": "TETER93"        , "xc_functionals": {"LDA_XC_TETER93"            : {}}})
```
PeterKraus: I'm not 100% happy with this how this is done, but I think for now it will do - once we migrate to LibXC 4.0, perhaps more changes will be necessary?

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -35,24 +34,50 @@
 from psi4.driver import p4util
 
 
+def get_default_dashparams(dtype):
```
PeterKraus: I've added this so that each dispersion type will return sensible defaults. The downside is that it might hide poorly defined dashparams in other parts of the code.

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,254 @@
...
+            "SCF: Libxc parameters requested for an exchange functional not defined as a component of %s." % (name))
+
+
+def build_superfunctional_from_dictionary(name, npoints, deriv, restricted):
+    """
+    This returns a (core.SuperFunctional, dispersion) tuple based on the requested name.
+    The npoints, deriv and restricted parameters are also respected.
+    """
+    check_consistency(functionals[name], name)
+    if "xc_functionals" in functionals[name].keys():
```
dgasmith: Can you add comments and spread this function out a bit with blank lines?

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,254 @@
...
+functionals = {}
+for functional_name in dict_functionals.keys():
+    functional_aliases = get_functional_aliases(dict_functionals[functional_name])
+    # first process aliases of parent functional
+    for alias in functional_aliases:
+        functionals[alias] = dict_functionals[functional_name]
+    # if the functional is already dispersion corrected, skip to next
+    if "dispersion" in dict_functionals[functional_name].keys():
+        continue
+    # loop through dispersion types in dashparams incl aliases
```
dgasmith: Ditto here to add a few more comments and blank lines.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
```diff
@@ -0,0 +1,494 @@
...
+#
+"""
+List of hybrid functionals
+"""
+
+import copy
+
+funcs = []
+
+
```
dgasmith: Can we keep it consistent with YAPF enable/disable blocks rather than for every function?

psi4/driver/procrouting/dft_funcs/dict_xc_funcs.py
dgasmith: Yea, LibXC 4.0 has more changes and the next iteration will be even different.

psi4/driver/procrouting/proc.py
```diff
@@ -25,7 +25,6 @@
 #
 # @END LICENSE
 #
-
```
dgasmith: Hmm, @loriab Do we have outstanding changes to `proc.py`? We might need to revert these changes so that this file is still mergeable later.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
PeterKraus: What happens is that sometimes yapf prefers the `"name": "method_name"` on the same line, sometimes it splits them onto newline. The same for `"citation"` and `"description"`. If you're not fussed about this, I can just remove the yapf disables...

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,347 @@
...
+# with Psi4; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# @END LICENSE
+#
+"""
+Superfunctional builder function & handlers.
+The new definition of functionals is based on a dictionary with the following structure
+dict = {
+           "name":  "",       name of the functional - matched against name.upper() in method lookup
```
PeterKraus: Hopefully this is the right place to do it. Let me know if it should be moved.

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,347 @@
...
+    if "alias" in functional_dict.keys():
+        aliases = [each.upper() for each in functional_dict["alias"]]
+        aliases.append(functional_dict["name"].upper())
+    else:
+        aliases = [functional_dict["name"].upper()]
+    return aliases
+
+
+# alias table for dispersion
+dispersion_names = {}
```
PeterKraus: We should probably move this to EmpiricalDispersion (or perhaps dashparams), so that when new corrections are added (VV10), it doesn't need to be edited in 20 places.

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,347 @@
...
+            c_params = functionals[name]["c_mp2"]
+            if "alpha" in c_params.keys():
+                sup.set_c_alpha(c_params["alpha"])
+            else:
+                sup.set_c_alpha(0.0)
+            
+            # The value of alpha is locked to 1.0 C++-side when SCS is detected
+            if "ss" in c_params:
+                sup.set_c_ss_alpha(c_params["ss"])
+                sup.set_c_alpha(1.0)
```
PeterKraus: I think this is the current behaviour enforced C++-side. However, without setting alpha to 1 here, the SCS are not activated!

tests/dft-cation/input.dat
```diff
@@ -81,47 +81,47 @@ qchem_data = {                              #TEST
      'EDF2':     0.4571615420000086,        #TEST
     'B97-2':    0.45620779479999385,        #TEST
     'B97-1':     0.4551136999999983,        #TEST
- 'wB97X-D3':    0.45707443810000825,        #TEST
```
PeterKraus: Disabled.

psi4/driver/procrouting/dft_funcs/dict_builder.py
dgasmith: +1 for that.

psi4/driver/procrouting/dft_funcs/dict_builder.py
dgasmith: Hmm, I see that in the C++ code. Is there a better way to handle the double hybrid business? This seems a bit awkward at the moment... happy to make some C++ changes based on your recommendations.

psi4/driver/procrouting/dft_funcs/dict_builder.py
dgasmith: Probably as good of place as any. At some point we probably want to convert this to an actual [schema](http://json-schema.org) and go through and test these.

psi4/driver/procrouting/dft_funcs/dict_builder.py
PeterKraus: I've moved it to `EmpiricalDispersion`. I don't like how "messy" the current interface is, scattered over a few files. Ideally, we'd load the `dashparams` from a `.json` file, and the `EmpiricalDispersion` module would be self-contained. I might make a pull request with some ideas once this is one is through... It would be good to get input from @hokru 's work on VV10 so that we're all on the same page.

psi4/driver/procrouting/dft_funcs/dict_builder.py
PeterKraus: I implemented it this way cause you suggested removing all of the `"c_mp2": {"alpha":x}` entries - the alpha has to be set python-side for MP2 correlation to be triggered at all. The current version works for me...

psi4/driver/procrouting/dft_funcs/dict_builder.py
PeterKraus: Let me know your ideas. I think another candidate for `json`-ising is the dashparams dictionary, see below.

PeterKraus:
I am not 100% convinced the dispersion handling - the way it's currently implemented - is very good. I'd propose the following:

- move all method and functional specific parameters from elsewhere into `qcdb/dashparam.py`. This includes citation handling that's currently done in `procrouting/empirical_dispersion.py`

- the rest of psi4 python-side should be "parameter-agnostic", ie there are obvious differences whether the dispersion _type_ is das2009, d2, or d3 (or later vv10) - this should get figured out (as it currently is) in `procrouting/empirical_dispersion.py` accordingly. 

- the parameters and citations should be passed into the `EmpiricalDispersion` class from outside, whether it's the functional's dictionary or user overrides or dashparams database. It might be helpful to settle on _one_ interface as opposed to the currently-used two (tuple_params tuple or dashparams dict in kwargs). In fact, EmpiricalDispersion class shouldn't be doing any parameter lookup on its own - the `dashparam.dashcoeff` dict is already trawled through in `procrouding/dft_funcs/dict_builder`, in addition to the modified name, the dispersion dashparam dict can (should?) be passed in accordingly

However, I have a feeling that this is a little bit of a feature creep. I'm quite happy with the functional part as it is now and wouldn't mind having this merged as is (if the tests pass), and leave the empirical / non-local dispersion re-work for eg. the VV10 implementation...

andysim:
This is fantastic!  I really like the design; kudos for coming up with such a clean scheme.  I know that these functions will not even be close to becoming a bottleneck but it would be nice to have the `for X in A.keys()` replaced with `for X in A:`.  Unless I'm mistaken (likely) the latter is faster, and more Pythonic (this could well have changed with Py3's change to the `keys()` method).  I think the shorter version is also easier to read.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
```diff
@@ -0,0 +1,494 @@
...
+#     },
+#     "description":
+#     '    This functional is currently broken - wB97X needs enabled tweaks.\n',
+#     "citation":
+#     '    Y.-S. Lin, G.-D. Li, S.-P. Mao, J.-D. Chai, J. Chem. Theory Comput., 9, 263-272, 2013\n',
+# })# yapf: disable
+
+funcs.append({
+    "name": "HF",
+    "alias": ["SCF"],
```
loriab: @psi4/editors do we want SCF as a HF alias – I've been trying to push the more specific HF, with SCF being potentially HF or DFT. I don't mind `energy('scf')` in an input, but if this goes here, then `SCF-D3BJ` shows up in the DFT functionals docs table.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
```diff
@@ -0,0 +1,577 @@
...
+    },
+    "c_mp2": {
+        "os": 0.46,
+        "ss": 0.37
+    },
+    "citation":
+    '    S. Kozuch, Phys. Chem. Chem. Phys., 13, 20104, 2011\n',
+    "description":
+    '    DSD-BLYP Dispersion-corrected SCS Double Hybrid XC Functional\n' + \
+    '    (full-core parameterization) \n'
```
loriab: Make this edit to appease the Sphinx. :-)

```
-    '    DSD-BLYP Dispersion-corrected SCS Double Hybrid XC Functional\n' + \
-    '    (full-core parameterization) \n'
+    '    DSD-BLYP Dispersion-corrected SCS Double Hybrid XC Functional (full-core param.)\n'
```

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -0,0 +1,338 @@
...
+    else:
+        aliases = [functional_dict["name"].upper()]
+    return aliases
+
+
+# alias table for dispersion
+dispersion_names = get_dispersion_aliases()
+
+functionals = {}
+for functional_name in dict_functionals.keys():
```
loriab: I'm with Andy on removing `.keys()` where you notice them. I don't think it matters for efficiency in py3, but imo it helps readability. Any contrary opinions chime in.

PeterKraus: I think I did it this way because I saw you use `.keys()` in other parts of the code. I'll change it, I too prefer the shorthand.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
PeterKraus: FWIW, it was a superfunc before this rework.

psi4/driver/procrouting/dft_funcs/dict_builder.py
loriab: Thanks for trying to match style. A good number of people (myself included) worked on the driver early in their python careers, so some of its style is more worth matching than others. :-)

psi4/driver/qcdb/dashparam.py
```diff
@@ -35,162 +35,261 @@
...
-        'dsd-blyp'    : {'s6': 0.50, 's8': 0.705, 'sr6': 1.569, 'rs8': 1.000, 'alpha6': 14.0},  
-        'ptpss'       : {'s6': 0.75, 's8': 0.879, 'sr6': 1.541, 'rs8': 1.000, 'alpha6': 14.0},
-        'pwpb95'      : {'s6': 0.82, 's8': 0.705, 'sr6': 1.557, 'rs8': 1.000, 'alpha6': 14.0},
-        'revpbe0'     : {'s6': 1.0,  's8': 0.792, 'sr6': 0.949, 'rs8': 1.000, 'alpha6': 14.0},
-        'revpbe38'    : {'s6': 1.0,  's8': 0.862, 'sr6': 1.021, 'rs8': 1.000, 'alpha6': 14.0},
-        'rpw86pbe'    : {'s6': 1.0,  's8': 0.901, 'sr6': 1.224, 'rs8': 1.000, 'alpha6': 14.0},
-        'wb97x'       : {'s6': 1.0,  's8': 1.000, 'sr6': 1.281, 'rs8': 1.094, 'alpha6': 14.0},
+        # S. Grimme, J. Antony, S. Ehrlich, H. Krieg, J. Chem. Phys 132, 154104, 2010
+        # 'b2plyp'      : {'s6': 0.5,  's8': 1.000, 'sr6': 1.332, 'alpha6': 14.0}, # superseded by a value below.
+        'pw6b95'      : {'s6': 1.0,  's8': 0.862,  'sr6': 1.532 , 'sr8': 1.000, 'alpha6': 14.0},
```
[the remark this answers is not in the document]
PeterKraus: To my understanding, a "proper" dispersion citation should reference both the method (which is currently done in the `EmpiricalDispersion` class) and the source of the parameters (which is done here or in the functional dictionaries). I've somewhat arbitrarily chosen to omit the latter if it's the same as the former, but it wouldn't be much work to be consistent throughout. Let me know if that's what you want - or we can leave it until we tackle the dispersion stuff properly!

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
```diff
@@ -0,0 +1,494 @@
...
+        "type": "d3bj",
+        "params": {
+            's6': 1.000,
+            's8': 0.8777,
+            'a1': 0.4171,
+            'a2': 2.9149
+        }
+    },
+    "description": 
+    '    Hartree Fock based 3C composite method with minimal basis set, gCP and D3(BJ)\n',
```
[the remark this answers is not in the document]
PeterKraus: This should all be working fine - the tests pass, and I am assuming that means it works.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
```diff
@@ -0,0 +1,494 @@
...
+    },
+    "c_functionals": {},
+    "dispersion": {
+        "type": "d3bj",
+        "params": {
+            's6': 1.000,
+            's8': 0.8777,
+            'a1': 0.4171,
+            'a2': 2.9149
+        }
```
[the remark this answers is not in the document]
PeterKraus: For a non-dispersion functional, like B3LYP, the `dashparam.dashcoeff[dtype]` is trawled through in `psi4/driver/procrouting/dft_funcs/dict_builder.py` lines 126-144 and the dispersion-corrected versions are generated from the `dashcoeff` data.

For a dispersion corrected functional, like B97-D, DSD-PBEPBE-D3BJ, or wB97X-D, I've defined the dispersion contributions directly in the dictionary of the functional. B97-D is a special case as B97 is a hybrid and B97-D is a dispersion corrected GGA; DSD functionals have different SCS parameters for each dispersion version, and wB97X-D has been parametrised with -CHG dispersion.

psi4/driver/procrouting/dft_funcs/dict_mgga_funcs.py
```diff
@@ -0,0 +1,214 @@
...
+        "MGGA_X_M06_L": {}
+    },
+    "c_functionals": {
+        "MGGA_C_M06_L": {}
+    },
+    "description": 
+    '    M06-L Meta-GGA XC Functional\n',
+    "citation": 
+    '    Y. Zhao and D. G. Truhlar, J. Chem. Phys. 125, 194101, 2006\n',
+})# yapf: disable
```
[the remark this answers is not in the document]
PeterKraus: I'll remove the `# yapf: disable`'s and see what happens. The only place where they should be strictly necessary is in the `dict_xc_funcs.py` file.

psi4/driver/procrouting/dft_funcs/dict_mgga_funcs.py
```diff
@@ -0,0 +1,214 @@
...
+    },
+    "description": 
+    '    MN15-L Meta-GGA XC Functional\n',
+    "citation": 
+    '    H. S. Yu, X. He, and D. G. Truhlar, J. Chem. Theory Comput. 12, 1280-1293, 2016\n',
+})# yapf: disable
+
+funcs.append({
+    "name": "mGGA_MS0",
+    "alias": ["MGGA-MS0"],
```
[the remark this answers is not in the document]
PeterKraus: I thought of this as a feature (especially for the `M06-2X-D3 = M062X-D3ZERO` case), and the current aliases are more of a "proof of concept". If you've already decided _not_ to have aliases, I can remove them from the offending cases. However, at least for things like `HCTH = HCTH120` or `B97-D = B97-D2`, I think it makes sense to keep the option there.

dgasmith:
Thanks again for so much work, and keeping with us through the review. Its a pretty big change so taking a bit longer than our typical PR's.

psi4/driver/qcdb/dashparam.py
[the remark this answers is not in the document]
loriab: Ah, got it, thanks. Let's defer until the dispersion revamp.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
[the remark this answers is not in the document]
loriab: Got it, thanks for the clarification.

psi4/driver/procrouting/dft_funcs/dict_mgga_funcs.py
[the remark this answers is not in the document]
loriab: Aliases _are_ a feature, and I thank you for it. Just alarmed to see the further expansion of the functional zoo. I'll have to be careful to always use `SuperFunctional.name()` and never `name` from `energy(name, ...)` in labeling things.

psi4/driver/procrouting/dft_funcs/dict_hyb_funcs.py
loriab: I believe it. More trying to gauge if others feel a crack-down is in order. I'll need to filter `SCF` out of the functionals table.

psi4/driver/procrouting/dft_funcs/dict_dh_funcs.py
loriab: This is the only really necessary change, otherwise the docs build will fail. I'm afraid the Sphinx is more particular than I am. :-)

psi4/driver/procrouting/dft_funcs/dict_xc_funcs.py
```diff
@@ -0,0 +1,153 @@
...
+funcs.append({"name": "HCTH/P76"       , "xc_functionals": {"GGA_XC_HCTH_P76"           : {}}, "alias": ["HCTHP76"]})
+funcs.append({"name": "HCTH/P14"       , "xc_functionals": {"GGA_XC_HCTH_P14"           : {}}, "alias": ["HCTHP14"]})
+funcs.append({"name": "B97-GGA1"       , "xc_functionals": {"GGA_XC_B97_GGA1"           : {}}})
+funcs.append({"name": "KT2"            , "xc_functionals": {"GGA_XC_KT2"                : {}}})
+funcs.append({"name": "TH1"            , "xc_functionals": {"GGA_XC_TH1"                : {}}})
+funcs.append({"name": "TH2"            , "xc_functionals": {"GGA_XC_TH2"                : {}}})
+funcs.append({"name": "TH3"            , "xc_functionals": {"GGA_XC_TH3"                : {}}})
+funcs.append({"name": "TH4"            , "xc_functionals": {"GGA_XC_TH4"                : {}}})
+funcs.append({"name": "HCTH/93"        , "xc_functionals": {"GGA_XC_HCTH_93"            : {}}, "alias": ["HCTH", "HCTH93"]})
+funcs.append({"name": "HCTH/120"       , "xc_functionals": {"GGA_XC_HCTH_120"           : {}}, "alias": ["HCTH120"], })
```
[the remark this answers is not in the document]
dgasmith: Your right, the "/" won't work. I think the most canonical thing would be to just remove the "/" here?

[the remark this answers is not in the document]
loriab: I was trying to trace through the CBS parsing, and I don't see anything fancy enough to allow the `HCTH/` through unscathed. Maybe `HCTH-P14` since there's adjacent letters for some variants.

Excellent catch. I think we already have `sapt2+(3)` confusing CBS — don't need more.

dgasmith:
Ok, lets do this once Travis passes. @PeterKraus any last thoughts?

PeterKraus:
One feature I'd like to add (later) is the ability to do something like:

```
my_new_method = { "name": "my_func", ... functional_dictionary ... }
my_superfunctional = add_functional_from_dictionary(my_new_method)
energy("my_func")
```
That way one can fit dispersion parameters or benchmark functionals in a simple loop... However, I think it is a bit of work, so I'd prefer to leave that for later.

dgasmith:
Whoo! Thanks again for this great change.

Issue #923, opened 2018-02-22
Compilation error with icc 18.0.1


rmcgibbo:
Compiling psi4 on -O3 with icc 18.0.1 seems to trigger a crash in the compiler.

The crash is not observed with the icc 17.0.2 with -O3, nor is it observed with icc 18.0.1 if I turn the optimization flag down to -O1.

```
/u/nyc/mcgibbon/gerrit/psi4-gerrit/psi4/src/psi4/dcft/dcft_gradient_UHF.cc(3903) (col. 72): internal error: 20000_6805

[ 85%] Building CXX object src/psi4/optking/CMakeFiles/optking.dir/opt_data_io.cc.o
compilation aborted for <prefix>/psi4/src/psi4/dcft/dcft_gradient_UHF.cc (code 4)
make[5]: *** [src/psi4/dcft/CMakeFiles/dcft.dir/dcft_gradient_UHF.cc.o] Error 4
make[4]: *** [src/psi4/dcft/CMakeFiles/dcft.dir/all] Error 2
make[4]: *** Waiting for unfinished jobs....
```

With 18.0.1, I can fix the error by turning down the optimization flag for that file in particular by adding the following snippet to psi4/src/psi4/dcft/CMakeLists.txt:
```
if (CMAKE_CXX_COMPILER_ID STREQUAL "Intel" AND
    CMAKE_CXX_COMPILER_VERSION VERSION_LESS 18.1 AND
    CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 18.0)

    message(WARNING "Intel C++ (ICC) 18.0 fails to compile dcft_gradient_UHF.cc with aggressive optimization flags")
    set_source_files_properties(dcft_gradient_UHF.cc PROPERTIES
        COMPILE_FLAGS -O1)
endif ()
```

Has anyone else tried using icc 18.0.1? Is there precedent for adding compiler- and file-specific flags in the CMakeLists? I'm not sure if this is suitable as a PR.

loriab:
I think @jturney has built just fine with 18.0.0 – have you tried 18.0.1, Jet?

No problem with compiler- and file-specific flags. We've done this [before](https://github.com/psi4/psi4/blob/f5560a0fe5a88a64dd6c040339b9f682e277da61/psi4/src/psi4/detci/CMakeLists.txt#L3) though in the other direction. Might see if Jet sees a code solution first.

jturney:
[...]

PR #931, opened 2018-03-06
allows for old cmake and new intel compilers

schiebermc:
## Description
Allows for #925 to work for #923 for those with old cmake and new intel compilers.

## Status
- [x] Ready to go


schiebermc:
By the way, why not just check for equality with 18.0?

susilehtola:
If Intel were to release 18.0.1 for instance?

schiebermc:
Makes sense, thanks!

Issue #926, opened 2018-03-03
Issues with ECP's from EMSL


PeterKraus:
Further to our discussion at Slack and e-mail, I made some tests on molecules with ECP's. Please see the attached G09 and psi4 input and log files.

To obtain the psi4 results, the g- and h- shell coefficient of 0.000 in the Xe/Hg ECP has been replaced with 1D-10. A simple trapping of `tsum==0` in `erd_normalize_shell()` in `libmintsgshell.py` doesn't work properly - the term has to be sanity-checked in the loop before on lines 241-252. 

For Hg, the `libmintsbasissetparser.py` had to be extended at line 202 to include the H-shell (thanks, Jesse).

![image](https://user-images.githubusercontent.com/5336995/36932575-1957f002-1ecb-11e8-84dc-b365f16b84b3.png)

As can be seen from the table above, the absolute energies of the dimers are fine when compared to G09. The `nocp` interaction energy is also OK, however when CP-correction is involved, the results are not good - I imagine the ECP doesn't get properly ghosted.

Finally, the psi4 calculations crash as `libmintsbasisset.py` is not able to find the basis sets for some reason - perhaps the basis set file is not passed into the routine properly? This also occurs when the basis set is defined in the job file using `basis { ... }`.

[ECP-test.zip](https://github.com/psi4/psi4/files/1777311/ECP-test.zip)

PeterKraus:
Possible fixes for the angular momentum and contraction coefficients are here:

https://github.com/psi4/psi4/compare/master...PeterKraus:ecp_fix

dgasmith:
For the failure at the end, check out `driver/driver_nbody.py:471`. Which looks like `wfn = core.Wavefunction.build(molecule, 'sto-3g')`. This should be replaced with `wfn = core.Wavefunction.build(molecule)` so that the global basis set is picked up.

You may want to try a fix more like `[x if x > 1.e-10 else x + 1.e10 for x in values]` to be a bit more canonical.

PR #937, opened 2018-03-15
Some ECP fixes.

PeterKraus:
## Description
Minor ECP fixes, so that more ECP's from EMSL Basis set exchange work out of the box. See issue #926 .

## Todos
* **Developer Interest**
  - [x] Extended naming tables to include H-shell (Stuttgart-Cologne ECP's eg. for Hg need it)
  - [x] Trapping "0.0" in contraction coefficients

## Status
- [x] Ready to go

### DIFF
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -468,7 +468,7 @@ def nbody_gufunc(func, method_string, **kwargs):
         ret_ptype = ret_energy
 
     # Build and set a wavefunction
-    wfn = core.Wavefunction.build(molecule, 'sto-3g')
+    wfn = core.Wavefunction.build(molecule, 'def2-svp')
     wfn.nbody_energy = energies_dict
     wfn.nbody_ptype = ptype_dict
     wfn.nbody_body_energy = energy_body_dict
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -468,7 +468,7 @@ def nbody_gufunc(func, method_string, **kwargs):
         ret_ptype = ret_energy
 
     # Build and set a wavefunction
-    wfn = core.Wavefunction.build(molecule)
+    wfn = core.Wavefunction.build(molecule, 'sto-3g')
     wfn.nbody_energy = energies_dict
     wfn.nbody_ptype = ptype_dict
     wfn.nbody_body_energy = energy_body_dict
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -468,7 +468,7 @@ def nbody_gufunc(func, method_string, **kwargs):
         ret_ptype = ret_energy
 
     # Build and set a wavefunction
-    wfn = core.Wavefunction.build(molecule, 'sto-3g')
+    wfn = core.Wavefunction.build(molecule)
     wfn.nbody_energy = energies_dict
     wfn.nbody_ptype = ptype_dict
     wfn.nbody_body_energy = energy_body_dict
--- a/psi4/driver/qcdb/libmintsbasisset.py
+++ b/psi4/driver/qcdb/libmintsbasisset.py
@@ -150,7 +150,7 @@ class BasisSet(object):
             isinstance(args[0], basestring) and \
             isinstance(args[1], Molecule) and \
             isinstance(args[2], OrderedDict) and \
-            args[3] == True:
+            isinstance(args[3], bool):
             self.constructor_role_mol_shellmap(*args)
         else:
             raise ValidationError('BasisSet::constructor: Inappropriate configuration of constructor arguments')
--- a/psi4/driver/qcdb/libmintsbasisset.py
+++ b/psi4/driver/qcdb/libmintsbasisset.py
@@ -352,7 +352,7 @@ class BasisSet(object):
                     self.uoriginal_coefficients[tst:tsp],
                     self.uexponents[tst:tsp],
                     'Pure' if self.puream else 'Cartesian',
-                    n, xyz_ptr, bf_count, pt='Normalised' if is_ecp else 'Unnormalized',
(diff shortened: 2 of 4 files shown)


JonathonMisiewicz:
I'm a little worried about the 0 coefficient handling. The 0 coefficient means "no local potential." (See the discussion in #newdev on Nov. 9.) Replacing that with a "really small but existing local potential" just _sounds_ like it will create nasty surprises down the road.

Good luck getting the tests passing!

psi4/driver/driver_nbody.py
```diff
@@ -468,7 +468,7 @@ def nbody_gufunc(func, method_string, **kwargs):
         ret_ptype = ret_energy
 
     # Build and set a wavefunction
-    wfn = core.Wavefunction.build(molecule)
+    wfn = core.Wavefunction.build(molecule, 'sto-3g')
```
dgasmith: Instead of `sto-3g`, could we use `def2-svp` here? Slightly large, but a much large atom coverage.

PeterKraus:
@JonathonMisiewicz Thanks, that was a good point. I've modified the function somewhat - as opposed to messing with the read-in original contraction coefficients, I've simply disabled the `erd_normalize_shell` call for ECP's (presumably when `self.rpowers` is not just full of `None`'s).

PeterKraus:
OK, so instead of messing around with the ShellInfo, I've decided to rather switch off the normalisation using pt="Normalized" for ECP shells. This was already done in the parser in `qcdb/libmintsbasissetparser.py` on lines 235-236. I have simply added a `True` flag as a 4th parameter to the `BasisSet` constructor, which indicates that the shellmap consists of ECP shells, which shouldn't be normalised.

This seems more correct than the guesswork in the previous commit.

andysim:
I agree with you; there shouldn't be any normalisation applied to ECPs.  When I put the ECP code in I wanted re-use as much of the existing basis set handling code as possible, in the spirit of the DRY principle.  Unfortunately there are a few subtle differences between how Gaussian shells and ECPs should be handled; I thought I had caught them all, but I didn't get the normalisation right.  Thank you very much for pointing it out and fixing it.  The parsing machinery used to take several trips between Python and C++ layers, so I attempted to consolidate it all Py-side.  In the process I may have left some stuff in there that's not needed, so cleanup like this is very welcome.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -346,7 +352,8 @@ def constructor_role_mol_shellmap(self, role, mol, shell_map):
                     self.uoriginal_coefficients[tst:tsp],
                     self.uexponents[tst:tsp],
                     'Pure' if self.puream else 'Cartesian',
-                    n, xyz_ptr, bf_count, pt='Unnormalized', rpowers=rpowers[tst:tsp])
+                    n, xyz_ptr, bf_count, pt='Normalised' if is_ecp else 'Unnormalized',
```
andysim: For consistency, this should be spelled `Normalized`

andysim:
I also agree that switching the 3 parameter function call for the 4 parameter version will be cleaner.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -352,7 +352,7 @@ def constructor_role_mol_shellmap(self, role, mol, shell_map, is_ecp = False):
                     self.uoriginal_coefficients[tst:tsp],
                     self.uexponents[tst:tsp],
                     'Pure' if self.puream else 'Cartesian',
-                    n, xyz_ptr, bf_count, pt='Normalised' if is_ecp else 'Unnormalized',
+                    n, xyz_ptr, bf_count, pt='Normalized' if is_ecp else 'Unnormalized',
```
andysim: As a Englishman living in the States, I often interchange these spellings.  Anywhere else in the code it wouldn't matter but these keywords are used in comparison operations, so very strange bugs could result.  Thanks for fixing :)

PeterKraus:
Hmm, I'm struggling to get my tests to give me a good agreement. From the [original publication](http://dx.doi.org/10.1007/s00214-005-0681-9) for the cc-pvdz-PP pseudopotentials for the Cu (and other) metals, the ionisation energy of Cu should be 165.26 kcal/mol with CCSD(T). In Gaussian 09, with the basis set downloaded from EMSL, I get an energy of 165.75 kcal/mol, which I suppose is reasonable. In Psi4, I get an energy of 166.02 kcal/mol.

In case of Au ionisation energy, the difference (200.23 kcal/mol in the paper, 202.92 kcal/mol in Psi4) is more than 1%.

I'm quite stumped. If anyone has any idea what's going on, please let me know.

PeterKraus:
Attached gaussian inputs:
[cu_plus.txt](https://github.com/psi4/psi4/files/1852217/cu_plus.txt)
[cu_0.txt](https://github.com/psi4/psi4/files/1852218/cu_0.txt)

And Psi4 inputs for the above test:
[cc-pvdz-pp-gbs.txt](https://github.com/psi4/psi4/files/1852221/cc-pvdz-pp-gbs.txt)
[input.txt](https://github.com/psi4/psi4/files/1852222/input.txt)

andysim:
I don't have access to G*, so can't easily verify.  Is the difference due to Psi4's use of density fitting by default?

PeterKraus:
As mentioned on Slack, RHF tests with Au+ and Cu+ yield the same total energies in G09 and Psi4 (8 decimals) with both CCSD(T) and HF using the cc-pvdz-PP basis sets. These basis tests both have `0.0` coefficient in the `N+1`th shell. The orbital eigenvalues match as well.

Unless we need more testing, I'm satisfied with this - the discrepancy in the ionisation energy is clearly coming from the open-shell part (and likely not from the basis set).

andysim:
That works for me - thanks for carefully checking the results

PR #942, opened 2018-03-24
SAPT(DFT) flexibility changes

### LINKED ISSUE #939: SAPT(DFT) : Allow calculations without the asymptotic-correction (AC)
Hi Daniel, 
It would be a good idea to allow SAPT(DFT) calculations without the AC, but this does not seem to be a possibility at the moment as the GRAC shifts are needed and they need to be non-zero. 

GRAC should be an optional part of the SAPT(DFT) calculation. And all parameters in GRAC should be user accessible (this is normally the case in other codes) so the way the splicing is done can be altered if needed. At present only the shifts are user-defined.

This would allow range-separated functionals to be used in SAPT(DFT). Right now it does not seem like they can. 
Thanks!

dgasmith:
## Description
Makes SAPT(DFT) quite a bit more flexible by allowing a potential user to build SCF wavefunctions (any will work, HF or DFT). An example can be found in `tests/sapt-dft-api`.

I have also added LRC references with SAPT(DFT). @ajmisquitta please checkout `tests/sapt-dft-lrc` and verify these are working as expected.

Closes #939 .

## Todos
* **Developer Interest**
  - [x] Users can now call SAPT(DFT) directly with reference wave functions.

## Status
- [x] Ready to go

### DIFF
--- a/tests/sapt-dft-api/input.dat
+++ b/tests/sapt-dft-api/input.dat
@@ -1,0 +1,41 @@
+#! SAPT(DFT) aug-cc-pVDZ interaction energy between Ne and Ar atoms.
+
+Eref = {"SAPT ELST ENERGY":  -0.10190449, # TEST
+        "SAPT EXCH ENERGY":   0.36545706, # TEST
+        "SAPT IND ENERGY":    0.00349281, # TEST
+        "SAPT DISP ENERGY":  -0.24398704, # TEST
+        "CURRENT ENERGY":     0.02305835} # TEST
+
+
+molecule dimer {
+  Ne
+  --
+  Ar 1 6.5
+  units bohr
+}
+
+# Set options
+set {
+    basis         aug-cc-pvdz
+    scf_type      df
+}
+
+# Prepare the fragments
+sapt_dimer, monomerA, monomerB = proc_util.prepare_sapt_molecule(dimer, "dimer")
+
+# Run the first monomer
+set DFT_GRAC_SHIFT 0.203293
+energyA, wfnA = energy("PBE0", molecule=monomerA, return_wfn=True)
+
+# Run the second monomer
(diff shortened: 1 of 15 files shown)


tests/sapt-dft-api/input.dat
```diff
@@ -0,0 +1,41 @@
+#! SAPT(DFT) aug-cc-pVDZ interaction energy between Ne and Ar atoms.
+
+Eref = {"SAPT ELST ENERGY":  -0.10190449, # TEST
+        "SAPT EXCH ENERGY":   0.36545706, # TEST
+        "SAPT IND ENERGY":    0.00349281, # TEST
+        "SAPT DISP ENERGY":  -0.24398704, # TEST
+        "CURRENT ENERGY":     0.02305835} # TEST
```
loriab: actually has to be `#TEST` for the regex to heed the signal, in case you end up making other changes.

dgasmith: Fixed, thanks.

dgasmith:
@ajmisquitta I may have pinged you incorrectly the first time. Let me know what you think of these changes and apologies for the possible double ping.

dgasmith:
I say review and pull this in. We can patch it up based on @ajmisquitta's comments at a later date.

tests/sapt-dft2/input.dat
```diff
@@ -29,6 +29,6 @@ set {
 
 energy('sapt(dft)', molecule=dimer)
 
-for k, v in Eref.items():                                  # TEST
-    compare_values(v / 1000.0, psi4.get_variable(k), 6, k) # TEST
+for k, v in Eref.items():                                  #TEST
```
andysim: A few people have made the `# TEST` mistake.  We may want to modify the regex in future to allow it.  Not related to this PR in any way, of course.

PR #945, opened 2018-04-04
testing whether DSD trouble is just in my build

loriab:
## Description
DSD-BLYP and DSD-PBE-PBE86 subtests in dft-custom-dhdf are failing. This is the easiest way to get it tested externally.

### DIFF
--- a/tests/dft-custom-dhdf/CMakeLists.txt
+++ b/tests/dft-custom-dhdf/CMakeLists.txt
@@ -1,3 +1,3 @@
 include(TestingMacros)
 
-add_regression_test(dft-custom-dhdf "psi;dft;scf")
+add_regression_test(dft-custom-dhdf "psi;dft;quicktests;scf")


loriab:
@bwb314, @PeterKraus, were errors for DSD-BLYP and DSD-PBE-PBE86 to be expected? I didn't see any comments to that respect but there's been a lot of edits to DFT and its testing of late. Thanks for any clues.

PeterKraus:
All DSD functionals, with the exception of the CORE-DSD-BLYP functional, were changed to the version from the 2013 paper. The versions without dispersion correction are now parameterized differently from the D2 and D3BJ versions. They will differ from previous version of Psi4, however our DSD-PBEPBE-D3BJ was checked against G16 and with frozen core on it passed.

loriab:
Ok, so there's a likely reason DSDs would have changed. Did the full test suite pass for dict_funcs on your local branch?

PR #947, opened 2018-04-05
Adds hostname to header, stop time info to footer

dsirianni:
## Description
Adds hostname printing to output file header, and stop time to output file "footer" before success/failure beer/coffee message.

## Status
- [x] Ready to go

### DIFF
--- a/psi4/extras.py
+++ b/psi4/extras.py
@@ -28,6 +28,7 @@
 
 import os
 import atexit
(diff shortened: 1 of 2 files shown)


loriab:
@robertodr, yeah, this was a group programming exercise, so minimal changes. But in general, I like the `.format`, too.

Issue #934, opened 2018-03-08
`properties` call fails with the method/basis syntax.


dgasmith:
The following input:

```
import psi4

mol = psi4.geometry("""
He 0 0 0
""")

wfn = psi4.properties('scf/sto-3g')
```

fails with the following exception:
```
Traceback (most recent call last):
  File "/Users/daniel/Gits/psi4/psi4/run_psi4.py", line 259, in <module>
    exec(content)
  File "<string>", line 23, in <module>
  File "/Users/daniel/Gits/psi4/psi4/driver/driver.py", line 826, in properties
    optstash = driver_util._set_convergence_criterion('properties', lowername, 6, 10, 6, 10, 8)
  File "/Users/daniel/Gits/psi4/psi4/driver/driver_util.py", line 78, in _set_convergence_criterion
    _method_exists(ptype, method_name)
  File "/Users/daniel/Gits/psi4/psi4/driver/driver_util.py", line 50, in _method_exists
    raise ValidationError('%s method "%s" is not available.%s' % (Cptype, method_name, alternatives))

ValidationError: Properties method "scf/sto-3g" is not available.
```

How `frequency` uses the code [here](https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L1674) to overcome the issue.

The `properties` call in Psi4 is located [here](https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L754) and could use a similar fix (or a nice little function to prevent duplication with `frequency`).

loriab:
Btw, is there concurrence on killing of `property()` in favor of `properties()`? I like the former better, but using a python syntax word makes me nervous.

dgasmith:
I incorrectly wrote that down, `property` was killed off in favor of `properties`. I do like the former better as well, but it messes with Python.

loriab:
`energy('B3lyp/6-311++G**')` is another one the regex needs to be smarter for.

loriab:
`+\s*D:` probably

dgasmith:
FYI for the new developers, the relevant line is here:
https://github.com/psi4/psi4/blob/master/psi4/driver/driver_cbs.py#L1516

PR #949, opened 2018-04-07
Allow for "/" in the names of methods.

PeterKraus:
Standardized handling of "/" in method name between frequency, hessian, gradient, properties and energy.

## Description
- Use of `_cbs_gufunc` helper function consistently across `energy`, `hessian`, `gradient`, `frequency` and `properties`
- `_cbs_gufunc` throws an exception for unsupported "true" CBS calls to `frequency` and `properties`, "method/basis" syntax works
- all relevant tests pass, see below
- modified regex in `_parse_cbs_gufunc_string()` to correctly handle `sapt2+(3)/basis` calls or `method/6-311++G` calls; splitting now only happens on variants of `+ D:`.

## Todos
* **Developer Interest**
  - [x] Fix for #934 

## Status
- [x] Ready to go

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -103,7 +103,7 @@ def _find_derivative_type(ptype, method_name, user_dertype):
             % (method_name, str(dertype), alternatives))
 
     return dertype
-    
+
 
 def energy(name, **kwargs):
     r"""Function to compute the single-point electronic energy.
@@ -471,7 +471,6 @@ def energy(name, **kwargs):
         postcallback(lowername, wfn=wfn, **kwargs)
 
     optstash.restore()
-    
     if return_wfn:  # TODO current energy safer than wfn.energy() for now, but should be revisited
 
         # TODO place this with the associated call, very awkward to call this in other areas at the moment
@@ -510,7 +509,7 @@ def gradient(name, **kwargs):
     # Bounce to CP if bsse kwarg (someday)
     if kwargs.get('bsse_type', None) is not None:
         raise ValidationError("Gradient: Cannot specify bsse_type for gradient yet.")
-    
+
     # Figure out what kind of gradient this is
     if hasattr(name, '__call__'):
         if name.__name__ in ['cbs', 'complete_basis_set']:
@@ -615,7 +614,6 @@ def gradient(name, **kwargs):
         wfn = procedures['gradient'][lowername](lowername, molecule=molecule, **kwargs)
 
         optstash.restore()
-        
         if return_wfn:
             return (wfn.gradient(), wfn)
         else:
@@ -824,10 +822,10 @@ def properties(*args, **kwargs):
     lowername, level = driver_util.parse_arbitrary_order(lowername)
     if level:
         kwargs['level'] = level
-    
+
     if "/" in lowername:
         return driver_cbs._cbs_gufunc(properties, lowername, ptype='properties', **kwargs)
-        
+
     return_wfn = kwargs.pop('return_wfn', False)
     props = kwargs.get('properties', ['dipole', 'quadrupole'])
 
@@ -1435,7 +1433,7 @@ def hessian(name, **kwargs):
         core.set_parent_symmetry('')
         optstash.restore()
         optstash_conv.restore()
-     
+
         if return_wfn:
             return (wfn.hessian(), wfn)
         else:
@@ -1676,7 +1674,6 @@ def frequency(name, **kwargs):
     lowername = name.lower()
 
     if "/" in lowername:
-        print(frequency.__name__, name)
         return driver_cbs._cbs_gufunc(frequency, name, ptype='frequency', **kwargs)
 
     if kwargs.get('bsse_type', None) is not None:
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -103,13 +103,6 @@ def _find_derivative_type(ptype, method_name, user_dertype):
             % (method_name, str(dertype), alternatives))
 
     return dertype
-
-
-def _is_cbs_method(methodname):
-    if ("+" in methodname) or ("[" in methodname) or (methodname.count("/") > 1):
-        return True
-    else:
-        return False
     
 
 def energy(name, **kwargs):
@@ -422,15 +415,9 @@ def energy(name, **kwargs):
     if kwargs.get('bsse_type', None) is not None:
         return driver_nbody.nbody_gufunc(energy, name, ptype='energy', **kwargs)
 
-    old_global_basis = None
     # Bounce to CBS if "method/basis" name
     if "/" in lowername:
-        if _is_cbs_method(lowername):
-            return driver_cbs._cbs_gufunc(energy, name, ptype='energy', **kwargs)
-        else:
-            old_global_basis = core.get_global_option("BASIS")
-            lowername, new_basis = lowername.split('/')
-            core.set_global_option('BASIS', new_basis)
+        return driver_cbs._cbs_gufunc(energy, name, ptype='energy', **kwargs)
 
     # Commit to procedures['energy'] call hereafter
     return_wfn = kwargs.pop('return_wfn', False)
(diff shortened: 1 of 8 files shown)


PeterKraus:
Hmm, there seems to be an issue with temporary files being left behind in the `energy` and `gradient` calls if they're not routed through the `cbs_gufunc` routines. If I call `clean()` in between the failing tests [like so for `cbs-xtpl-energy`](https://github.com/psi4/psi4/files/1886155/cbs-xtpl-energy-mod.txt), the test passes fine.

PeterKraus:
There are some issues with temporary files sticking around. In the current version, if I do (excerpt from `props4` test):
```
E, wfn = prop('scf/aug-cc-pvdz', properties=["GRID_ESP", "GRID_FIELD"], return_wfn=True)
E, wfn = prop('scf/cc-pvdz', properties=["GRID_ESP", "GRID_FIELD"], return_wfn=True)
```
the code works fine, but when the bases are very different:
```
E, wfn = prop('scf/sto-3g', properties=["GRID_ESP", "GRID_FIELD"], return_wfn=True)
E, wfn = prop('scf/cc-pvdz', properties=["GRID_ESP", "GRID_FIELD"], return_wfn=True)
```
I have to call clean() in between them.

dgasmith:
The basis sets are different in both examples? I wonder why this is exclusively for prop, can you post the full error, it would be good to track down the extra temps.

loriab:
IIRC, big-basis-followed-by-smaller-basis is ok but not the reverse.

PeterKraus:
OK, then I suggest that for all cases where we're switching basis sets, we should do a `clean()` - it's done in the CBS routine which is used for `energy()` and `gradient()` (as the CBS routines can handle "method/basis" syntax), we might as well do it for `properties()`.

loriab:
Wondering if it mightn't be better to just continue routing everything with a `/` to `driver_cbs._cbs_gufunc` only now with an extra kwarg `cbs_allowed=True` that is false for `def properties()`. I worry about not breaking (or complicating) recursive driver cases (where cbs and nbody function re-enter energy, hessian, properties etc. after being broken into smaller pieces (like `mp2/cc-pv[dt]z` --> `mp2/cc-pvdz`)).

PeterKraus:
Well, that's an obvious and simple solution I should've thought about...

PeterKraus:
The failing test is weird: it used to do FD, now it does analytical frequencies, and that's why there's a difference. Should I enforce `dertype="energy"`?

PeterKraus:
@dgasmith, @loriab : provided the `cbs-xtpl-freq` test passes (it does on my PC), this should be good to go from my PoV. Should I run a full test suite on this?

loriab:
So https://github.com/psi4/psi4/blob/master/tests/cbs-xtpl-freq/input.dat#L22 is properly catching the `dertype=1` again and doing Hess by findif of Grad?

PeterKraus:
Yeah, it was getting popped from the kwargs dict and not passed into the `func()` call. I moved the pop after single-basis methods are dealt with.

psi4/driver/driver.py
```diff
@@ -1706,10 +1703,6 @@ def frequency(name, **kwargs):
     for postcallback in hooks['frequency']['post']:
         postcallback(lowername, wfn=wfn, **kwargs)
 
-    # Reset old global basis if needed
-    if not old_global_basis is None:
-        core.set_global_option("BASIS", old_global_basis)
```
loriab: Nice to get this old inconsistency of `old_global_basis` and `lowername, new_basis = lowername.split('/')` out of here and put freq on par with gradient.

psi4/driver/driver_cbs.py
```diff
@@ -33,6 +33,7 @@
 import math
 import sys
 import numpy as np
+import copy
```
loriab: Did you end up using `copy`?

dgasmith:
As a tangent I would like to figure out why we need clean at the SCF level. Cannot wait to shove everything into UUID-based HDF groups.

PeterKraus:
@loriab I'm confused about the test results. The ctest returns:
```
The following tests FAILED:
	184 - fcidump (Failed)
	359 - dft-custom-dhdf (Failed)
	363 - dft-cation (Timeout)
	364 - dft-water-dimer (Timeout)
	384 - chemps2-caspt2-n2 (Failed)
```
where the `dft-` and `fcidump` tests failing are expected as your PR that fixes them is not yet pulled in. The `chemps2/caspt-n2` test passes when I run it manually and passed in the test log file (see attached)
[testresults.log](https://github.com/psi4/psi4/files/1889030/testresults.log)

psi4/driver/driver_cbs.py
```diff
@@ -1586,6 +1583,16 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
         else:
             return ptype_value
 
+    # Drop out for props and freqs
+    elif ptype == "properties":
+        raise ValidationError("Properties: Cannot extrapolate or delta correct properties yet.")
+    elif ptype == "frequency":
+        raise ValidationError("Frequency: Cannot extrapolate or delta correct frequencies yet.")
```
dgasmith: No reason we cannot do this for `freqs`, should get around to testing that someday. Can we also add an `else` here at the very end of `ptype` is not matched?

PeterKraus: should i do:
```
else:
    pass
```
or shift everything in by one indent?

PeterKraus:
So I have tried to set-up a good test case, and I run into the `clean()` issue again, see attachments.
[cbs-parser.txt](https://github.com/psi4/psi4/files/1889912/cbs-parser.txt)
[cbs-parser.log](https://github.com/psi4/psi4/files/1889914/cbs-parser.log)

loriab:
On the `clean` issue, I don't see a change in behavior from your PR. The below (on master) also requires the intervening `clean()` to avoid a PSIO error. Maybe the rules of how much the user should need to call clean need rethinking, but at least they are steady (one used to have to call it after practically everything).

```
molecule ne2 {
    0 1
    Ne 0 0 0
    --
    0 1
    Ne 2 0 0
}

set  basis cc-pvdz
sapt_global = energy("sapt2+(3)", molecule=ne2)

clean()  # PSIO error without this line

# these two calls are equivalent
#cbs_string = energy("mp2/6-311++G** + D:ccsd/6-311+G")
cbs_string = energy(cbs, corl_wfn='mp2', corl_basis='6-311++G**', delta_wfn='ccsd', delta_basis='6-311+G')
compare_values(-257.50097969, cbs_string, 5, 'cbs')
```

On the caspt2, runs fine for me, too. If you want to see the error try [looking in the testing log file](http://psicode.org/psi4manual/master/build_planning.html#faq-testsoutput). But it's certainly outside the scope of this PR.

PeterKraus:
OK, I've re-ordered the two tests so that they don't crash without `clean()`. I had hoped that this PR might have sorted out this `clean()` issue, but for true CBS method, it's unfortunately still there when switching basis sets.

The test is meant to test the "new" functionality: allowing "method/basis" syntax with `sapt2+(3)` and + D: extrapolation with augmented Pople basis sets (both previously crashed due to "+").

@dgasmith do you want to include some `property()` or `frequencies()` calls in the test too?

dgasmith:
@PeterKraus Yea, to round things out we should probably add those single calls tests as well to ensure that we do not regress at some point in the future (although dont make this a quickest). As a note its fine to call clean in the tests.

PeterKraus:
The `frequency()` call is already checked in `cbs-xtpl-freq` with `dertype=1`; the handling was just moved from driver to cbs_driver in this PR.

For the `properties()`, I think it'd be easier to extend one of the `props[1-4]` tests; I've extended the `props4` call to compare the already-computed V-values from SCF with global cc-pVDZ basis with ones from an explicit SCF/cc-pVDZ calculation.

PeterKraus:
@loriab Done, I added it to the other cbs-tests at the beginning of the list.

PR #950, opened 2018-04-10
pybind11 2.2.2

loriab:
## Description

## Todos
* **Developer Interest**
  - [x] Bump CheMPS2 to v1.8.4
  - [x] Bump Libint to Ninja-able version, v1.2.1+2
  - [x] Bump Pybind11 to v2.2.2. Note that this invokes symbol hiding, so ...
  - [x] Include Jet's `PSI_API` macro to re-expose classes for plugins.
  - [x] Note that anyone wanting to re-use an `objdir` will need to **thoroughly** remove the old pybind11 v2.0.0 from detectability. This means:
     - `<objdir> rm -rf stage/<TAB-TAB-...-TAB>/includes/pybind11`
     - `<objdir> rm -rf stage/<TAB-TAB-...-TAB>/share/cmake/pybind11`
     - `<objdir> rm -rf external/upstream/pybind11`

## Status
- [x] Ready for review
- [x] Ready for merge

@psi4/editors, these should be quick and uncontroversial. Getting this in would help the PR assembly line.

### DIFF
--- a/external/upstream/chemps2/CMakeLists.txt
+++ b/external/upstream/chemps2/CMakeLists.txt
@@ -28,7 +28,7 @@ if(${ENABLE_CheMPS2})
             DEPENDS lapack_external
                     hdf5_external
             GIT_REPOSITORY https://github.com/SebWouters/CheMPS2
-            GIT_TAG 55d6b38  # v1.8.3-12
+            GIT_TAG v1.8.4
             UPDATE_COMMAND ""
             CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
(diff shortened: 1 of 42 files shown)


loriab:
Very true on exposing more symbols later, @dgasmith. I haven't even attempted v2rdm.

@edeprince3, FYI, this affects you. There's a discussion in https://github.com/psi4/psi4/pull/814 if you want the details, but the tl;dr is that Psi4's `core.so` built with this PR will not have all the symbols exposed like a normal Linux `.so` for a plugin's use. If there's extra classes or functions that v2rdm needs, they'll need a `PSI_API` added to the (psi) source code. If you get to the v2rdm/psi reconciliation process before me, please ping with any trouble. I'm updating and rebuilding all the conda packages now, but v2rdm is always last, having psi as a dep. :-)

PR #951, opened 2018-04-10
Bug fix in building orbital cube files

jeffschriber:
## Description
When running a simple RHF calculation and computing cube files, I was experiencing seg faults or orbitals that look like:
image
after updating my Psi for the first time in a while. After a bit of digging, I think I've tracked the problem--no more seg faults and my orbitals look like:
image

Not sure why the cubeprop test (which was passing) didn't catch this.

## Todos
- [X] Fix copy call in points.cc

## Questions
- [ ] Should I add another cubeprop test?

## Status
- [X] Ready for review
- [ ] Ready for merge


loriab:
was there some symmetry in the other cubeprop tests that didn't see the bad index transpose? or is the problem present in the test cubes and just no one looked at them? By all means add another cubeprop test if it spans more testing space, but probably not otherwise. Do we need new reference cubes for the existing tests?

psi4/src/psi4/libfock/points.cc
```diff
@@ -250,7 +250,7 @@ void RKSFunctions::compute_orbitals(std::shared_ptr<BlockOPoints> block) {
     double** Ca2p = C_local_->pointer();
     for (int ml = 0; ml < nlocal; ml++) {
         int mg = function_map[ml];
-        C_DCOPY(na, Cap[ml], 1, Ca2p[mg], 1);
+        C_DCOPY(na, Cap[mg], 1, Ca2p[ml], 1);
```
dgasmith: We might be able to track this back to the `memcpy` -> `DCOPY` move some time ago. Can you also fix UKS orbitals if needed?

loriab:
That "last 10 lines or so different" sounds eerily familiar from the last cubeprop kerfluffle. But that, too, was a definite bug located. Regenerate all the ref cubes, I think. You use gcc, right? And I'll be sure to test them pronto with icpc.

jeffschriber:
@fevangelista said the same thing about the last 10 line issue. The densities are all very similar, with Ds being identical

tests/cubeprop/Da.cube.ref
```diff
@@ -77,7 +77,7 @@ Property: Da [e/a0^3]. Isocontour range for 85% of the density: (0.0697875,0)
  1.18433E-02  1.04189E-02  8.63150E-03  6.77687E-03  5.08580E-03  3.68099E-03 
  2.15389E-03  2.70243E-03  3.33736E-03  4.05961E-03  4.86617E-03  5.74235E-03 
  6.64919E-03  7.51158E-03  8.21632E-03  8.62834E-03  8.62836E-03  8.16425E-03 
- 7.28822E-03  6.14651E-03  4.92188E-03  3.76965E-03  2.78419E-03  1.41093E-03 
+ 7.28821E-03  6.14651E-03  4.92188E-03  3.76965E-03  2.78419E-03  1.41093E-03 
  1.76081E-03  2.16825E-03  2.63544E-03  3.16152E-03  3.73740E-03  4.33769E-03
```
loriab: Last time the numbers were way different (or maybe it was extra lines), so the close numbers here are reassuring. As is Ds unchanged.

loriab:
cubeprop tests pass with `icpc`, so home free.

PR #952, opened 2018-04-10
Exporting N-Body intermediates to psivars

dsirianni:
## Description
Exports intermediate energies computed within N-Body driver to psivars named `N-BODY {cluster-identity} TOTAL ENERGY` for workflow incorporation and sanity-checking.

Convention for `cluster-identity`: `(monomer_tuple)@(basis_tuple)` enumerates the particular monomers involved in the cluster computation in `monomer_tuple`, and enumerates the basis set used within the cluster computation in `basis_tuple`.  In this way, ghost monomers are easily identified as any fragment with basis functions (included in `basis_tuple` not present in `monomer_tuple`.   For example,
```
"N-BODY (1, 2)@(1, 2, 3) TOTAL ENERGY"                  =>    -529.655058639489
```
indicates that the total energy for the fragment (1, 2) in basis set (1, 2, 3) is -529.66 [Eh].  

## Todos
- [x] Variable documentation (pending notation consensus, see questions below)

  - [x] Psivar export of intermediate energies in N-Body for workflow integration

## Questions
- [ ] Is the cluster identification syntax described above suitably intuitive?  Should the `@` symbol be replaced with another, perhaps `:`, to remove potential confusion with ghost atoms notation in the molecule block?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/glossary_psivariables.rst
+++ b/doc/sphinxman/source/glossary_psivariables.rst
@@ -602,6 +602,19 @@ PSI Variables by Alpha
    .. math:: E_{NN} = \sum_{i, j<i}^{N_{atom}}\frac{Z_i Z_j}{|\mathbf{R}_i - \mathbf{R}_j|}
       :label: ENN
 
+.. psivar:: NBODY (i, j, ..., k)@(a, b, ..., c) TOTAL ENERGY
+
+   The total energy [Eh] of a component of the requested N-Body energy.
+   The first parenthetical list over *i*, *j*, ..., *k* enumerates 
+   molecular fragments included in the computation in 1-indexed, 
+   input-file order, while the second enumerates list over *a*, *b*, 
+   ..., *c* enumerates which fragments contribute basis functions to the
+   computation.  For example, ``(1, 2)@(1, 2, 3, 4)`` indicates that the
+   fragments 1 and 2 are explicitly included in the energy computation,
+   with basis functions from each of fragments 1, 2, 3, & 4 included in 
+   the basis set.  Therefore, the basis functions from fragments 3 and 4
+   are included as ghost functions within the energy computation.
+
 .. psivar:: OCEPA(0) TOTAL ENERGY
    OCEPA(0) CORRELATION ENERGY
 
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -310,10 +310,14 @@ def nbody_gufunc(func, method_string, **kwargs):
 
             core.clean()
 
-    # Set vairables for N-Body intermediates
+    # Set variables for N-Body intermediates, add to dict for wavefunction attachment
+    intermediates_dict = {}
     for pair in energies_dict.keys():
-        core.set_variable("N-BODY (%s)@(%s) TOTAL ENERGY" %
-                          (', '.join([str(i) for i in pair[0]]), ', '.join([str(i) for i in pair[1]])), energies_dict[pair])
+        var_key = "N-BODY (%s)@(%s) TOTAL ENERGY" % (', '.join([str(i) for i in pair[0]]), 
+                                                      ', '.join([str(i) for i in pair[1]]))
+        core.set_variable(var_key, energies_dict[pair])
+        # Redundantly add var to energies_dict for wavefunction attachment
+        intermediates_dict[var_key] = energies_dict[pair]
 
     # Final dictionaries
     cp_energy_by_level   = {n: 0.0 for n in nbody_range}
@@ -346,6 +350,7 @@ def nbody_gufunc(func, method_string, **kwargs):
 
 
     # Sum up all of the levels
+    nbody_vars = {}
     for n in nbody_range:
 
         # Energy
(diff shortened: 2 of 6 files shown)


tests/nbody-intermediates/input.dat
```diff
@@ -0,0 +1,42 @@
+#! HF/cc-pVDZ many body energies of an arbitrary noble gas trimer complex
+#! Size vs cost tradeoff is rough here
+
+comparison_dict = {}                                         # TEST
```
loriab: regex wants `#TEST`

psi4/driver/driver_nbody.py
```diff
@@ -304,13 +304,17 @@ def nbody_gufunc(func, method_string, **kwargs):
...
             core.print_out("\n       N-Body: Complex Energy (fragments = %s, basis = %s: %20.14f)\n" %
                                                                 (str(pair[0]), str(pair[1]), energies_dict[pair]))
-
             # Flip this off for now, needs more testing
             #if 'cp' in bsse_type_list and (len(bsse_type_list) == 1):
             #    core.set_global_option('DF_INTS_IO', 'LOAD')
 
             core.clean()
 
+    # Set vairables for N-Body intermediates
```
loriab: sp.

psi4/driver/driver_nbody.py
```diff
@@ -304,13 +304,17 @@ def nbody_gufunc(func, method_string, **kwargs):
...
+    for pair in energies_dict.keys():
+        core.set_variable("N-BODY (%s)@(%s) TOTAL ENERGY" %
```
loriab: Did you find a handy `wfn` to attach these to, too?

tests/nbody-intermediates/input.dat
loriab: nope, alignment doesn't matter. just needs the five chars together.

psi4/driver/driver_nbody.py
loriab: But you're not adding your psivars to `energy_body_dict` to be subsequently attached to `wfn`, right? If say, the `psi4.get_variable('your psivar')` in the tests got its psivars from `wfn`, not `P::e`, it wouldn't find it currently, I think. Future-proof by making that test pass no matter which loc psivars are coming from.

dsirianni: That's correct, yes.  I will add them to `energies_dict` so that when they get attached to the wavefunction with `wfn.nbody_energy = energies_dict` they are recognized.

dgasmith:
While you are here can you bind the `core.set_variable` psivars to the `wfn.set_variable` ones instead. This will involve you needing to create the return wavefunction earlier, but an important change.

dsirianni:
@dgasmith Would you prefer I sub out all `core.set_variable(...)` with `wfn.set_variable(...)` or just ensure that all variables set are also bound to the wavefunction? I've was about to commit the latter as I've just got that working properly, but I'm not sure how important it is to keep access to the psivars w/ `P::e`

psi4/driver/driver_nbody.py
```diff
@@ -467,19 +471,22 @@ def nbody_gufunc(func, method_string, **kwargs):
...
-    wfn.nbody_energy = energies_dict
-    wfn.nbody_ptype = ptype_dict
-    wfn.nbody_body_energy = energy_body_dict
-    wfn.nbody_body_ptype = ptype_body_dict
+    dicts = [intermediates_dict, energies_dict, ptype_dict, energy_body_dict, ptype_body_dict, nbody_dict]
+    for d in dicts:
+        if d is not None:
+            for var in d.keys():
+                wfn.set_variable(str(var), d[var])
+                core.set_variable(str(var), d[var])
```
amjames: Using the dictionary like this is much cleaner! Awesome

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -602,6 +602,19 @@ PSI Variables by Alpha
...
+   The first parenthetical list over *i*, *j*, ..., *k* enumerates 
+   molecular fragments included in the computation in 1-indexed, 
+   input-file order, while the second enumerates list over *a*, *b*, 
+   ..., *c* enumerates which fragments contribute basis functions to the
+   computation.  For example, ``(1, 2)@(1, 2, 3, 4)`` indicates that the
+   fragments 1 and 2 are explicitly included in the energy computation,
+   with basis functions from each of fragments 1, 2, 3, & 4 included in 
+   the basis set.  Therefore, the basis functions from fragments 3 and 4
+   are included as ghost functions within the energy computation.
+
```
amjames: Docs, SWEET! I am I big fan of the syntax (i,j...) @ (a,b...) syntax by the way.

dsirianni: I like it too.  Originally I was using a parenthetical notation, but @loriab suggested the `@` notation which I think is much more clean and clear.

psi4/driver/driver_nbody.py
```diff
@@ -467,19 +471,22 @@ def nbody_gufunc(func, method_string, **kwargs):
...
-    # Build and set a wavefunction
+    # Build wfn and bind variables
     wfn = core.Wavefunction.build(molecule, 'def2-svp')
-    wfn.nbody_energy = energies_dict
-    wfn.nbody_ptype = ptype_dict
-    wfn.nbody_body_energy = energy_body_dict
-    wfn.nbody_body_ptype = ptype_body_dict
+    dicts = [intermediates_dict, energies_dict, ptype_dict, energy_body_dict, ptype_body_dict, nbody_dict]
+    for d in dicts:
+        if d is not None:
```
loriab: Sure this is `if d is not None:` rather than `if d:`? Those empty objects are `{}`, right, so they evaluate to `False`, but they're not the singleton `None`, I don't think.

dsirianni: Other than explicitly casting `str(var)` for variable setting, since `energies_dict` does not use string keys, this seems the proper way to do it.  Good catch, thanks!

psi4/driver/driver_nbody.py
```diff
@@ -336,12 +340,13 @@ def nbody_gufunc(func, method_string, **kwargs):
         nocp_ptype_body_dict = {n: np.zeros(arr_shape) for n in nbody_range}
         vmfc_ptype_body_dict = {n: np.zeros(arr_shape) for n in nbody_range}
     else:
-        cp_ptype_by_level, cp_ptype_body_dict = None, None
-        nocp_ptype_by_level, nocp_ptype_body_dict = None, None
-        vmfc_ptype_body_dict = None
+        cp_ptype_by_level, cp_ptype_body_dict = {}, {}
```
dsirianni: For posterity's sake, the reason I had used the `NoneType` check in the final var setting loop [here](https://github.com/psi4/psi4/pull/952/files#diff-8348f203a0eebe7b8ff993f7e6b17c43R474) is that these dicts were originally initialized to `None`, then set to the variable-containing dicts of interest [here](https://github.com/psi4/psi4/pull/952/files#diff-8348f203a0eebe7b8ff993f7e6b17c43R444).

PR #919, opened 2018-02-14
Py-side EFP and Mol parsing

loriab:
## Description
This gets the scfitertopy (scf iterations py-side, pb221, pcm py-side) fully operational again with efp py-side. Also extends pyvib2's Molecule dict (often labeled "molrec") init and validation to input parsing. This requires two extensions to molrec – one for efp and one for zmat/variables in geometry. Replaces all the c-side constructors (i.e., create_molecule_from_string, init_with_xyz), replacing them with a unified dict, str, arrays constructor.

This branch is presently master (reviewed) + pyvib2 (unreviewed) + scfitertopy (semi-reviewed) + newstuff, all rebased so it doesn't align with anything. After pyvib2 makes it into master, I'll rebase and separate into scfitertopy and newstuff, so it can be readily reviewed. This is just getting the commits posted so I can break psi locally yet again.

## Todos
Notable points that this PR has either accomplished or will accomplish.
* **Developer Interest**
  - [ ] Feature1
* **User-Facing for Release Notes**
  - [ ] Feature2

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

Issue #886, opened 2017-12-17
Generalized Davidson solver


dgasmith:
We have started a folder that contains arbitrary solvers in `psi4/driver/p4util/solvers.py`. A generalized Davidson solver which can find the first `n` roots would be greatly useful for both TDDFT, and SCF (DFT) root following. 

Requirements:
 - First n roots
 - Allow preconditions
 - Specific root following (say only positive roots)
 - Support initial guesses
 - Subspace collapse
 - Any requirements or suggestions @CDSherrill might have as well.

I believe Ruhee is on this project, but opening a stub.

mdav2:
Hi, is anyone on this/should I get started on it? Looking for a first project and this is something I found myself wanting this afternoon - seems like a good fit!

rdcunha:
@mdav2 @dgasmith I've been working on this for some time, and I have a decent bit of the code written. I'll open a pull request from my fork soon.

dgasmith:
@rdcunha Cool! It would be very useful to make a PR this week so folks can understand what parts are being worked on even if it is not complete.

dgasmith:
Closed by #1349. Thanks for everyones help here!

PR #953, opened 2018-04-10
Py-side SCF, PCM, EFP, Mol parsing 2

loriab:
## Description
~This replaces #919, which I don't want to overwrite just yet b/c it has history and some review comments. This is up-to-date wrt master (10 Apr) and has had a recent pyvib2 rebased through (9 Apr). All tests pass. This is **not** ready to look at b/c it still includes #834 . When that makes it into master, I'll rebase it through again and then~ this PR should only include fully-working py-side scf and mol parsing (`create_molecule_from_string` is gone in this code). This is my last sequence PR before splitting away qcdb. I suggest it form the start of `v1.3` .

~Since originally posted (+27k/–10k), this PR has had #834 subtracted from it (now +13k/–7k). It is built atop #965, so I wouldn't review until that gets pulled in and subtracted from this. What does need answering soonish is whether for 1.2/1.3 purposes, the remaining commits of this PR should be separated further.~

~Now this has had #965 rebased through, it should be stable for a good while. It now has size (+9k/–6k). There doesn't seem much interest in promoting the below to 1.2, so no need for further subdivision.~

This has now had **v1.2 rebased through** so I sincerely hope it's been kneaded for the last time.

## Todos: Py-side SCF/PCM/EFP (1.3)
* **Developer Interest**
  - [x] SCF iterations moved py-side
  - [x] includes control of MOM, DIIS, damping, soscf, frac, efp
  - [x] SCF finalize, stability, post-iterations, printing moved py-side
  - [x] Rework a good bit of Wfn::HF::common_init() to minimize the convergence helper controllers that are c-side
  - [x] A couple export/def bugs fixed in Wfn::HF and more fns exported and moved to public in class
  - [x] PCMSolver interface reworked for py-side
  - [x] libefp EFP object moved from c-side `P::e.EFP` to a py-side attached EFP attribute on the `psi4.core.Molecule` object.
  - [x] Psi4 forgot about the c-side EFP object in bin, lib, globals, exports (still in options)
  - [x] EFP/EFP calcs moved purely py-side
  - [x] SCF/EFP calcs carry out operations on the EFP object py-side
  - [x] efp_torque no longer in `P::e` but `EFP TORQUE` in `P::e:arrays`

## Todos: Py-side Mol Parsing (could be split out for 1.2)
* **User Interest**
  - [x] `molecule {...}` will take xyz (element symbol _or_ atomic number) or psi4 formats.
  - [x] overall chgmult now specifiable through `1 3\n--` before any fragments.
  - [x] can specify atomic number, mass number, mass, ghosting, extra-label in mol spec.
* **Developer Interest**
  - [x] `*.Molecule` forgot `charge_specified` and `multiplicity_specified` attributes. They were only used once in deciding mult defaulting in the SCF and that logic is now handled at mol parsing time.
  - [x] `*.Molecule` forgot `create_molecule_from_string()` and good riddance. Also `*.Molecule` `init_with_xyz` (all the format variations can be handled by plain `from_string`. There's an optional `dtype` if you want to be specific.
  - [x] Remove singlet/doublet defaulting logic in `Wavefunction::common_init()`, as this is taken care of by mol init
  - [x] In keywords (`read_options`), `EFP_POL`, `QMEFP_POL`, `EFP_POL_DAMPING`, all `POL` --> `IND`.
  - [x] Main `molutil.set_geometry()` that all psithon and psiapi calls use to process user molecule switch from c-side parsing to `qcdb.molparse.from_string()` parsing. This fn handles QM mol, EFP frag, and PubChem contents.
  - [x] `libefp` accessed through `PylibEFP` and `-DENABLE_libefp=ON` activates both
  - [x] `psi4.core.Molecule.molecule_from_string(string)` constructor calls py-side parsing. Can override string contents with `fix_com`, `fix_orientation`, `fix_symmetry` args. `string` can have QM mol, EFP frag, and pubchem contents.

## Questions
- [ ] @robertodr, feel free to elaborate PCMSolver changes
- [x] ~**Everyone** the previous PR #965 gets both Mol classes to a point where they _can_ be instantiated by py-side parsing (or arrays, or dict). Do we want the second set of changes above where inputparser is calling py-side parsing into 1.2? It will require PylibEFP. SCF/EFP will remain c-side.~
- [x] @robertodr https://github.com/psi4/psi4/pull/953#discussion_r203921398 full test suite (incl. pcm) passes with form_F commented out

## Checklist
- [ ] symmetrize pubchem inputs, see https://github.com/psi4/psi4/pull/965#issuecomment-382854005
- [ ] max_attempts for instability following isn't through

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Say, @psi-rking, does the `6` signify anything special [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/findif/fd_freq_1.cc#L433)? Looks like another was confused, too (https://github.com/psi4/psi4/blob/master/psi4/src/psi4/optking/geom_gradients_io.cc#L376). If it was `3 * nat`, it'd look a lot like Cfour's FCMFINAL file.

JonathonMisiewicz:
I know that 6*natom is also found in the header of the second derivative output file from WDA's `INTDER`, but I'm still not sure why. In addition to my `findif` work, I'm using `INTDER` for my work at UGA, so I'm doubly interested in learning what it means.

Our `anharm` program calls this number `NSTORE` and `NNSTOR`, but it never does anything with them.

psi-rking:
I haven't the slightest idea!  But I was mimicking the old file15 output:
http://vergil.chemistry.gatech.edu/psi/devel/progman/node47.html

The description here is plain as day "sixtimesn" but there is no explanation - a memory issue on an ancient architecture?  An equally interesting question would be "_who_ knows the answer to this question?"

loriab:
That's really great – it acknowledges our curiosity but does not satisfy it. Sounds like an office-by-office survey at CCQC would be required.

"The cartesian Hessian matrix is found in file15. The first line of this file gives the number of atoms (`n`) and, in case you are curious, six times the number of atoms (`sixtimesn`)."

JonathonMisiewicz:
I'll look into this. There aren't many people who might know here, which (sadly) cuts the detective work short.

psi-rking:
The tone of the documentation to which Lori alludes sounds like Justin
Fermann to me, but the convention is probably older than the arrival of the
Quantum Trio to the CCQC.

loriab:
Origin aside, do I gather correctly that optking (or its successors; optking being the only visible consumer of the `.hess` file) isn't dependent on the `6` and if I could absorb the format into `FCMFINAL`, it'd be ok with your code, @psi-rking?

lothian:
The convention is definitely older and may connect to other quantum chemistry codes.  I believe that ACES and ACESII used the same format for the Hessian file, at least many years ago.  Perhaps the source is in one of the really old codes, like the HONDO code or Pulay’s TEXAS code?

-Daniel

psi-rking:
Not a big deal.  My preference would be to produce an equivalent file to
CFOUR.  For example, the Crawford group has some chiroptical scripts
related to ROA that use second derivatives (file15-like) or dipole moment
derivatives (file17-like) but the headers vary slightly between PSI and
CFOUR, and this causes headaches, or worse, incorrect results if errors
aren't thrown.

loriab:
@psi-rking, I should explicitly generate both and do a diff, but the `3*nat` vs. `6*nat` is the sole difference I see between Psi and Cfour hessian formats. I was hoping to completely collapse the format names, but I guess I'll keep the `6` one generate-able (`hondo` or `intder` name? I'll have to confirm hondo involvement) in case it turns up to matter. Only one line of non-shared code. So the current plan is for both to be generate-able but Psi printing to switch to `3` for Cfour consistency.

JonathonMisiewicz:
Dr. Allen said the convention predates _him_, and he doesn't know why it exists either. It looks like this bit of quantum chemistry lore has been lost.

CDSherrill:
To my knowledge, nothing in the Psi ecosystem needs the 6*natom, so making
us Cfour compatible sounds good.  (Of course, in the future, passing this
some other way, like JSON, is really better).

loriab:
fwiw, I searched in the HONDO manual and couldn't find the 6. But if anyone runs across a `HSSFIL` spec, take a look. I put a bounty comment on it in the code. Thanks for your investigations, @JonathonMisiewicz

loriab:
scfitertopy is ready for review. again.

psi4/driver/procrouting/scf_proc/efp.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # get multipoles count, pos'n, values from libefp
+    #   charge + dipoles + quadrupoles + octupoles = 20
+    nmp = efpobj.get_multipole_count()
+    xyz_mp = np.asarray(efpobj.get_multipole_coordinates(verbose=verbose)).reshape(nmp, 3)
+    val_mp = np.asarray(efpobj.get_multipole_values(verbose=verbose)).reshape(nmp, 20)
+
+    #                    0  X  Y  Z  XX   YY   ZZ   XY   XZ   YZ
+    prefacs = np.array([ 1, 1, 1, 1, 1/3, 1/3, 1/3, 2/3, 2/3, 2/3,
+        1/15, 1/15, 1/15, 3/15, 3/15, 3/15, 3/15, 3/15, 3/15, 6/15])
```
andysim: I know we're dropping py2 support, but I don't think this division works correctly there.  If we're dropping it now, that's totally find with me, else we should probably have 1.0/3, etc.

loriab: it's got `from __future__ import division` so I think it's ok.

andysim: Oh, good call.  As you were!

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -771,210 +501,189 @@ void HF::find_occupation()
...
-    outfile->Printf( "\n");
-    outfile->Printf( "  ==> Geometry <==\n\n");
+#ifdef _OPENMP
+    nthread = Process::environment.get_n_threads();
+#endif
 
+    outfile->Printf("\n");
+    outfile->Printf("         ---------------------------------------------------------\n");
+    outfile->Printf("                                   SCF\n");
+    outfile->Printf("            by Justin Turney, Rob Parrish, Andy Simmonett\n");
```
andysim: This overhaul is more than enough justification to add your name to the list of authors of this chunk of code, Lori.  Especially with the number of times you've had to rebase this branch!  If we're going to continue the Psi3 convention of tagging the code segments, it may as well be accurate

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+    # post-scf vv10 correlation
+    if core.get_option('SCF', "DFT_VV10_POSTSCF"):
+        self.functional().set_lock(False)
+        self.functional().set_do_vv10(True)
+        self.functional().set_lock(True)
+        core.print_out("  ==> calculating VV10 correction on post-scf density <==\n\n")
+        SCFE = 0.0
+        self.form_V()
+        SCFE += self.compute_E()
+        self.set_energies("Total Energy", SCFE)
```
dgasmith: Pretty simple in a python iter world :). Can we change the string to "Computing non-self-consistent VV10 Energy Correction", would match literature better.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+
+    # Perform wavefunction stability analysis before doing
+    # anything on a wavefunction that may not be truly converged.
+    if core.get_option('SCF', 'STABILITY_ANALYSIS') != "NONE":
+        # We need the integral file, make sure it is written and
+        # compute it if needed
+        if core.get_option('SCF', 'REFERENCE') != "UHF":
```
dgasmith: Going to be super easy to insert #886 here!

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+        if core.get_option("PCM", "PCM_SCF_TYPE") == "SEPARATE":
+            calc_type = core.PCM.CalcType.NucAndEle
+        Dt = self.Da().clone()
+        Dt.add(self.Db())
+        _, Vpcm = self.get_PCM().compute_PCM_terms(Dt, calc_type)
+        self.push_back_external_potential(Vpcm)
+
+    # recompute the Fock matrices as they are modified during the SCF
+    #   iteration and might need to be dumped to checkpoint later
+    self.form_F()
```
dgasmith: Is this normal? Seems like a needed JK build. The Fock matrix should be fine I think?

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+    if core.get_option('SCF', 'PCM'):
+        core.print_out("    PCM Polarization Energy =         {:24.16f}\n".format(epcm))
+    if hasattr(self.molecule(), 'EFP'):
+        core.print_out("    EFP Energy =                      {:24.16f}\n".format(eefp))
+    core.print_out("    Total Energy =                    {:24.16f}\n".format(total_energy))
+
+    self.set_variable('NUCLEAR REPULSION ENERGY', enuc)
+    self.set_variable('ONE-ELECTRON ENERGY', e1)
+    self.set_variable('TWO-ELECTRON ENERGY', e2)
+    if abs(exc) > 1.0e-14:
```
dgasmith: Can we change this to `self.functional().needs_xc()`. Will make this logic cleaner.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+        self.set_variable('DFT XC ENERGY', exc)
+        self.set_variable('DFT VV10 ENERGY', evv10)
+        self.set_variable('DFT FUNCTIONAL TOTAL ENERGY', hf_energy + exc + evv10)
+        self.set_variable('DFT TOTAL ENERGY', dft_energy)
+    else:
+        self.set_variable('HF TOTAL ENERGY', hf_energy)
+    if abs(ed) > 1.0e-14:
```
dgasmith: Would like to switch the logic here to `hasattr(self, "_disp_functor")`, but might be overkill.

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -738,100 +711,6 @@ void Molecule::rotate_full(const Matrix &R) {
     set_full_geometry(new_geom);
 }
 
-void Molecule::init_with_xyz(const std::string &xyzfilename) {
```
dgasmith: Woohoo!

psi4/src/psi4/libscf_solver/hf.cc
dgasmith: +1 here

psi4/src/psi4/libscf_solver/rhf.cc
```diff
@@ -792,7 +783,7 @@ std::vector<SharedMatrix> RHF::cphf_solve(std::vector<SharedMatrix> x_vec, doubl
     return ret_vec;
 }
 
-int RHF::soscf_update() {
+int RHF::soscf_update(float soscf_conv, int soscf_min_iter, int soscf_max_iter, int soscf_print) {
```
dgasmith: `float -> double`?

psi4/src/psi4/libscf_solver/rohf.cc
```diff
@@ -512,36 +481,28 @@ void ROHF::form_D()
...
-        if (na == 0)
-            memset(static_cast<void*>(Da[0]), '\0', sizeof(double)*nso*nso);
-        if (nb == 0)
-            memset(static_cast<void*>(Db[0]), '\0', sizeof(double)*nso*nso);
-
-
-        C_DGEMM('N','T',nso,nso,na,1.0,Ca[0],nmo,Ca[0],nmo,0.0,Da[0],nso);
-        C_DGEMM('N','T',nso,nso,nb,1.0,Ca[0],nmo,Ca[0],nmo,0.0,Db[0],nso);
+        if (na == 0) memset(static_cast<void*>(Da[0]), '\0', sizeof(double) * nso * nso);
+        if (nb == 0) memset(static_cast<void*>(Db[0]), '\0', sizeof(double) * nso * nso);
```
dgasmith: Nitpick, but while I see it. Can we do `Da_->zero(); Db_->zero();` above the `nirrep_` loop and remove the memset?

psi4/src/psi4/libscf_solver/rohf.cc
```diff
@@ -844,22 +774,22 @@ void ROHF::Hx(SharedMatrix x, SharedMatrix ret)
...
+    Da_->axpy(damping_percentage, Da_old_);
+    Db_->scale(1.0 - damping_percentage);
+    Db_->axpy(damping_percentage, Db_old_);
+    Dt_->copy(Da_);
+    Dt_->add(Db_);
 }
 
-int ROHF::soscf_update()
-{
+int ROHF::soscf_update(float soscf_conv, int soscf_min_iter, int soscf_max_iter, int soscf_print) {
```
dgasmith: Ditto from above.

psi4/src/psi4/libscf_solver/uhf.cc
```diff
@@ -312,45 +287,39 @@ void UHF::form_D()
...
 
-        if (na == 0)
-            ::memset(static_cast<void*>(Da[0]), '\0', sizeof(double)*nso*nso);
-        if (nb == 0)
-            ::memset(static_cast<void*>(Db[0]), '\0', sizeof(double)*nso*nso);
-
-        C_DGEMM('N','T',nso,nso,na,1.0,Ca[0],nmo,Ca[0],nmo,0.0,Da[0],nso);
-        C_DGEMM('N','T',nso,nso,nb,1.0,Cb[0],nmo,Cb[0],nmo,0.0,Db[0],nso);
+        if (na == 0) ::memset(static_cast<void*>(Da[0]), '\0', sizeof(double) * nso * nso);
+        if (nb == 0) ::memset(static_cast<void*>(Db[0]), '\0', sizeof(double) * nso * nso);
```
dgasmith: Ditto from above.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+        self.form_F()
+        core.timer_off("HF: Form F")
+
+        if verbose > 3:
+            self.Fa().print_out()
+            self.Fb().print_out()
+
+        SCFE += self.compute_E()
+        if efp_enabled:
+            global efp_Dt
```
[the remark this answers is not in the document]
loriab: honestly, no. I'd like to, but I don't think it's possible (open to suggestions). There's a callback fn with a very specific signature that the C library is expecting. No way to get the density into it except by global (`self` is out b/c that's part of the signature; no trouble using member data if this is in a C++ class).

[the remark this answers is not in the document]
dgasmith: Hmm, do you have the function to point me to?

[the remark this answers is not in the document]
loriab: https://github.com/ilyak/libefp/blob/master/src/efp.h#L196-L219 and https://github.com/loriab/pylibefp/blob/master/pylibefp/core.cc#L461-L489

psi4/driver/molutil.py
```diff
@@ -299,12 +298,22 @@ def geometry(geom, name="default"):
     the string are filtered.
 
     """
-    core.efp_init()
-    geom = pubchemre.sub(process_pubchem_command, geom)
-    geom = filter_comments(geom)
-    molecule = core.Molecule.create_molecule_from_string(geom)
+    molrec = qcdb.molparse.from_string(geom,
```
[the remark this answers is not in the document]
loriab: Certainly do-able for the Mol constructor. Just have to check if dict arg is a valid schmea repr [here](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/molecule.py#L100-L101) and call `molparse.from_schema`, which will have to be created by moving [`Molecule.from_schmea`](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/molecule.py#L1419-L1420) deeper into molparse.

When that's in place, can decide if `psi4.geometry` should expand its takings from {string} to {string, json}. I've no objection.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+    if self.iteration_ < 2:
+        core.print_out("  ==> Iterations <==\n\n")
+        core.print_out("%s                        Total Energy        Delta E     RMS |[F,P]|\n\n" % ("   "
+                                                                                                      if is_dfjk else ""))
+
+    # SCF iterations!
+    SCFE_old = 0.0
+    SCFE = 0.0
+    Drms = 0.0
+    while True:
```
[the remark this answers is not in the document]
loriab: In general I entirely agree with for/else over "while True"/break. But in this case where `self.iterations_` is the true counter and we may not enter the loop at `=0` (thus throwing off where to give up upon exceeding MAXITER), I wonder if having the shadow `iter` isn't more confusing.

psi4/driver/procrouting/scf_proc/scf_iterator.py
loriab: Good call. My first associations to earlier string were with correlated methods.

psi4/driver/procrouting/scf_proc/scf_iterator.py
loriab: @robertodr, this `form_F` used to [_precede_ the pcmsolver part] (https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libscf_solver/hf.cc#L592-L615). now it follows pcmsolver. All the pcm tests run fine with it commented out. Do you recall if it's related to you and if you still need it?

psi4/src/psi4/libscf_solver/rohf.cc
loriab: There aren't two memset pairs in rohf, but I went ahead and applied pattern to cuhf, rhf, rohf, uhf

psi4/src/psi4/libscf_solver/rhf.cc
loriab: I don't know what I was thinking

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -0,0 +1,766 @@
...
+"""
+
+import numpy as np
+
+from psi4.driver import p4util
+from psi4.driver import constants
+from psi4.driver.p4util.exceptions import ConvergenceError, ValidationError
+from psi4 import core
+
+from .efp import get_qm_atoms_opts, modify_Fock_permanent, modify_Fock_induced
```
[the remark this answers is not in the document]
loriab: `__all__` only controls `from psi4 import *`, I believe. `psi4.os` will still be accessible and in `dir(psi4)`. that's why I started deleting some temps https://github.com/psi4/psi4/blob/master/psi4/__init__.py#L73 . We can delete more. But since `__all__` doesn't actually restrict access, for us it only serves as guide-to-eye, and I think `public`/`_private` can do for that.

Another tidbit I found is that program chokes if you `from psi4 import *` and `__all__ = ['DO_NOT_WILD_IMPORT']` is present b/c that member is not in module. We could disable wild imports!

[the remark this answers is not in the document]
dgasmith: Yea `__all__` will only effect `from psi4.submodule import *`, at the same time we do not say `from psi4.driver import os`. `os` comes from many places, I was mostly thinking that this particular would be a good spot to start the `__all__` revolution.

psi4/src/psi4/libscf_solver/hf.cc
robertodr: :+1:

psi4/driver/procrouting/scf_proc/scf_iterator.py
robertodr: If tests work then I guess it was not needed :) I can't recall if I did that or not, but the way I modified `form_F` should always included the PCM part correctly.

PR #961, opened 2018-04-13
Enable faster density fitting

schiebermc:
## DESCRIPTION
Using `DF_Helper`'s density fitting technology, both integral and exchange matrix computations can be sped up for in-core investigations.  I have created a `JK` derived class, `symm_JK`, which wraps the technology of `DF_Helper` and enables its use for J/K building in Psi4.

## RESULTS
Here are some timing results for entire SCF procedures across various systems, using two different architectures.  Notice that relative speedup between `symm_JK` and `DFJK` improves for non-uniform memory access (NUMA) architectures.

april research update

april research update-1

## TODO
The new JK building object, `symm_JK`, will be faster than DFJK for in-core implementations.  However, it will be slower for disk-based implementations.  
- [x] Add check to figure out which object to use for J/K building. (in-core vs disk)


andysim:
This is the first time I've seen the timings.  Really impressive stuff; performance boosts on NUMA arch is a really big deal.

schiebermc:
Thanks!!  Forgot to mention that symm_JK = algorithms 10 and 13, DFJK = algorithms 11 and 14.

jturney:
Nice, looking forward to getting this in.

psi4/src/psi4/libfock/jk.h
```diff
@@ -938,5 +939,85 @@ class CDJK : public DFJK {
...
+
+/**
+ * Class symm_JK
+ *
+ * JK implementation using sieved, threaded
+ * density-fitted technology
+ * under slightly different paradigm than DFJK
+ * wraps lib3index/DF_Helper class
+ */
+class symm_JK : public JK {
```
dgasmith: Could we call this guy `symm_DFJK` so that we are clear that this is a flavor of DFJK.

tests/python/symm_JK/input.dat
```diff
@@ -0,0 +1,65 @@
...
+
+# get integrals
+DFJK_ints = [DFJK.J(), DFJK.K()]
+SYMM_JK_ints = [SYMM_JK.J(), SYMM_JK.K()]
+
+# compare
+for j, t in enumerate(['J', 'K']):
+    for i in range(len(DFJK_ints[0])):
+        print(t+str(i), ': %e' % 
+            (np.linalg.norm(np.asarray(DFJK_ints[j][i]) -
```
dgasmith: Could you use `compare_array` here similar to other Psi4 tests so that an exception can be raised.

tests/python/symm_JK/input.dat
```diff
@@ -0,0 +1,65 @@
+import psi4
```
dgasmith: I think by default the tests are all lower case, it would be good to stick with that. (I may be wrong here, worth checking).

tests/python/JK_builds/input.dat
```diff
@@ -1,259 +0,0 @@
-import psi4
```
dgasmith: Out of curiosity, why kill off this guy?

psi4/src/psi4/lib3index/df_helper.cc
```diff
@@ -165,26 +176,54 @@ void DF_Helper::initialize() {
...
     timer_off("DFH: initialize()");
+    
+    if(debug_) {
+        outfile->Printf("Exiting DF_Helper::initialize\n");
+    }
 }
+void DF_Helper::AO_core() {
 
+    // total size of sparse AOs
+    size_t required = ( do_wK_ ? 3 * big_skips_[nao_] : big_skips_[nao_]);
```
dgasmith: On this note, do we raise anywhere if `do_wK_` is currently called?

psi4/src/psi4/libfock/jk.cc
```diff
@@ -75,20 +75,38 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
...
-        if (options["BENCH"].has_changed()) jk->set_bench(options.get_int("BENCH"));
-        if (options["DF_INTS_IO"].has_changed()) jk->set_df_ints_io(options.get_str("DF_INTS_IO"));
-        if (options["DF_FITTING_CONDITION"].has_changed())
-            jk->set_condition(options.get_double("DF_FITTING_CONDITION"));
-        if (options["DF_INTS_NUM_THREADS"].has_changed())
-            jk->set_df_ints_num_threads(options.get_int("DF_INTS_NUM_THREADS"));
-
-        return std::shared_ptr<JK>(jk);
+        
+        if(options.get_str("DF_SCF_TYPE") == "DFJK") {
```
dgasmith: We might want to consider calling the current `DFJK` to something like `DiskDFJK` or something so that `DFJK` defines a type with multiple algorithms under it. We haven't handled something like this before.

schiebermc: Are you thinking another level of inheritance? Namely something like `JK` -> `DFJK` -> (`symm_DFJK` and `disk_DFJK`)?
What I mean is that we could create a generic `DFJK` derived from `JK` that would handle all the reused density fitting stuff.  Then the current `symm_JK` and `DFJK` classes would inherit from that class.  Would require some code movement.

dgasmith: This comes from the fact that Im not really sure the `DF_SCF_TYPE` option is correct here. I think we should have `SymmDF` and `DiskDF` options here which call the different JK objects. We should also have a `DF` input option which chooses between `SymmDFJK` and `DiskDFJK` via something like:

```
elif (jk_type == "DF") {
    if (something){
        return build_JK(primary, auxiliary, options, "SymmJK")
   } else {
        return build_JK(primary, auxiliary, options, "DiskJK")
   }
}
```

This give users the ability to call `build_JK` with either: `SymmDF`, `DiskDF` or `DF` so they can be explicit or implicit.

Happy to take other ways of doing this, but something simple would be good here.

As a note please change the class name of `symm_JK` to `SymmJK`, classes should be in Pascal case.

dgasmith: I dont know if it's at the programmer level, but certainly `SymmDF|DiskDF` is not fully standard user territory.

psi4/src/psi4/lib3index/df_helper.cc
schiebermc: No, I set up the template for the wK code to be added, but there are currently no throws.  Shall I go ahead and do that? or are we planning to handle that logic elsewhere?

dgasmith: This will be caught at other layers in typical workflows. But if its not implemented you should still raise an error as you are introducing an API that anyone can use.

psi4/src/psi4/libfock/jk.cc
schiebermc: So did we decide on what `something` should be?

dgasmith: Lets go with `SYMM_DF` and `PACKED_DF` as `jk_type` options keys where `SymmDFJK` and `PackedDFJK` are their class names. Please note that `symm_DF` as a class name conflicts with out style choices.

schiebermc: Also, I can change `SCF_TYPE` like so:
Before:
```
options.add_str("SCF_TYPE", "PK", "DIRECT DF PK OUT_OF_CORE CD GTFOCK");
```
After:
```
options.add_str("SCF_TYPE", "PK", "DIRECT PACKED_DFJK SYMM_DFJK PK OUT_OF_CORE CD GTFOCK");
```
But then how do we want to assign the `SCF_TYPE` in the pertinent locations?  The reason we went with `DF_SCF_TYPE` originally was that it allowed the reuse of most DF logic with just one difference in `jk.cc`.

loriab: Is there ever going to be logic switching btwn `SYMM_DF` and `PACKED_DF`? If so, then `DF` (default; apply best of symm/packed), `SYMM_DF`, and `PACKED_DF` (latter two for power user control) are sensible.

If not, then `DF` (which name we mustn't change) should map internally to `SYMM_DF` and `PACKED_DF` can let power users access the orig. implementation. And the `read_options` docstring should clarify all.

schiebermc: The logic is something like this:
1.  `PackedDFJK` if `wK` else `SymmDFJK`
2.  `PackedDFJK` if procedure relies intrinsically on old DFJK else `SymmDFJK`
3.  `SymmDFJK` if `SymmDFJK->AO_core()` else `PackedDFJK` 

I am not sure where this logic will be applied.  From our discussions, (3.) will probably be a conservative fully dense estimation, but we will need all of `nbf`, `naux`, and `doubles` where it is applied.  Unfortunately, `doubles` (memory allocated to JK object) is procedure dependent.

loriab: @psi4/maintainers welcome to chime in

I think this comes down to do we collapse SCF algorithm and DFJK algorithm into one option, `SCF_TYPE` or two `SCF_TYPE` and `DF_SCF_TYPE`/`DFJK_TYPE`/? I keep being persuaded by the last person who posted, so casting a wider net for opinions.

schiebermc: My take is that having the two options might be easiest.  A power user can use `DF_SCF_TYPE / DFJK_TYPE` if they want, otherwise we can map to the correct object (SymmDFJK or PackedDFJK) under the hood if `SCF_TYPE == DF`.  I guess it comes down to whoever has enough experience with Psi4 to imagine how to apply the above logic in the cleanest way.

amjames: I agree with @schiebermc two options to allow the user who want to control it have control. possibly with a generic default like `DFJK_TYPE = DEFAULT` which would be documented to explain logic used to select the implementation by default. 

The selection logic would check the non-defualt option values, and if none of them catch the "else" implements the logic described above? 
 
I am also in favor of the `DFJK_TYPE` name for the proposed option over the `DF_SCF_TYPE` name. It is shorter and more descriptive IMO. In all cases the `SCF_TYPE` is `DF` the `DFJK` engine is what is getting selected.

dgasmith: I really like this logic encoded inside the JK constructor rather than another option that I have to track down so that when I request DF it makes the correct decision for me, but I can overload it if needed. `DF_SCF_TYPE` is going to be difficult as JK is used in many places than just SCF.

dgasmith:
Matt can you run your test cases for a pure GGA (BLYP is fine). Be curious to see the timing difference there. I have a hunch that you may be a bit faster than you would expect not due to pure J construction cost, but data movement for the entire build.

schiebermc:
Okay, here are some other considerations that must be accounted for when constructing the correct DFJK object:

1.  PackedDFJK if do_wK_ else SymmDFJK
2.  SymmDFJK if SymmDFJK->AO_core() else PackedDFJK

Note that the above two will require knowledge of `do_wK_` and `doubles_` **in the JK constructor**.

3.  PackedDFJK if procedure relies intrinsically on old DFJK else SymmDFJK

(3.) might be tricky.  `DFJK` used `df_ints_io_` to prepare, save, and load the 3-center integrals.  Some procedures in Psi4 intrinsically rely on being able to load files produced by DFJK.  So not only will we need avoid using SymmDFJK in those procedures, but we might also need to do so in preemptive procedures (such as SCF) so that the files can be properly prepared for use later on.

loriab:
After a discussion, the plan is to:

* Implement SymmDFJK vs PackedDFJK logic in the DFJK constructor, thus requiring memory and wK info at initialization time that we don't now.
* Overload the constructor so that there's a default and an elaborate signature
* For methods that require PackedDFJK because of downstream data structure assumptions, that will be handled in proc.py with a force_*dfjk flag passed to scf_helper
* SymmDFJK aka Matt's will be named MemDFJK. PackedDFJK aka Rob's will be named DiskDFJK. since the actual implementations aren't as orthogonal as those names suggest, references/more explicit explanations will be in HF and/or SCF_TYPE docstrings
* `SCF_TYPE` option will have three relevant choices
  * `DF` formerly went to Rob's code. now will choose the best DFJK from available info. will be conservative on memory estimation
  * `MEM_DF` will always go to Matt's code so that a user can test both or be smarter than the logic depending on known hardware/mem considerations
  * `DISK_DF` will always go to Rob's code
* The read_options docstring for SCF_TYPE will label the MEM_DF and DISK_DF as expert choices
* In future, other algorithms will follow similar DF|MEM_DF|DISK_DF options in their MTD_TYPE algorithm controllers

Any further ideas on labeling these things welcome. @CDSherrill

schiebermc:
Note that `DFJKGrad` in `psi4/src/psi4/scfgrad/` is unaffected by name changes.

psi4/src/psi4/libfock/MemDFJK.cc
```diff
@@ -0,0 +1,113 @@
...
+    dfh_->set_fitting_condition(condition_);
+    dfh_->set_memory(memory_ - memory_overhead());
+    dfh_->set_do_wK(do_wK_);
+    dfh_->set_omega(omega_);  
+ 
+    // we need to prepare the AOs here, and that's it.
+    // DFHelper takes care of all the housekeeping
+
+    // TODO add wK integrals.
+    if (!do_wK_) {
```
dgasmith: Can you add a mention that `set_do_wK` on duh throws an error?

psi4/src/psi4/libfock/jk.cc
```diff
@@ -74,9 +74,9 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
 
         return std::shared_ptr<JK>(jk);
```
dgasmith: What happens if someone hits this constructor with the "DF" type? Should we add a DiskDFJK fallback for that here.

psi4/src/psi4/scfgrad/response.cc
```diff
@@ -1052,16 +1052,18 @@ std::shared_ptr<Matrix> SCFGrad::rhf_hessian_response()
...
 
+    size_t mem = 0.9 * memory_ / 8L;
+    size_t per_A = 3L * nso * nso + 1L * nocc * nso;
+    size_t max_A = (mem / 2L) / per_A;
+    max_A = (max_A > 3 * natom ? 3 * natom : max_A);
+    
     std::shared_ptr<JK> jk;
     if (options_.get_str("SCF_TYPE") == "DF"){
-        jk = JK::build_JK(basisset_, get_basisset("DF_BASIS_SCF"), options_);
+        jk = JK::build_JK(basisset_, get_basisset("DF_BASIS_SCF"), options_, false, mem);
```
dgasmith: Why go into an `if` statement here? This signature should be valid for all SCF_TYPE's?

psi4/src/psi4/libfock/jk.cc
schiebermc: There will be a throw for unknown `SCF_TYPE`.  The only way to avoid this is to use the correct constructor that maps `DF` to the correct type.  I think the throw might be better than a fallback, since it would be the programmers fault and not the user.

psi4/src/psi4/scfgrad/response.cc
schiebermc: Yes, the signature is valid for all `SCF_TYPE`'s, I was merely using the `if` statements that were already in place.  This check occurs in almost every location of `build_JK()`

psi4/src/psi4/libfock/MemDFJK.cc
```diff
@@ -0,0 +1,139 @@
...
+        message << "  If you are not a developer or using Psi4NumPy please report this issue at github.com/psi4/psi4." << std::endl;
+        throw PSIEXCEPTION(message.str());
+    }
+}
+void MemDFJK::postiterations() {
+}
+void MemDFJK::print_header() const {
+    // dfh_->print_header();
+    if (print_) {
+        outfile->Printf("  ==> MeMDFJK: Density-Fitted J/K Matrices <==\n\n");
```
loriab: lowercase "M"

dgasmith:
` psi4numpy-rhf-hessian` for whatever reason is simply not completing on Travis, works fine on my laptop. Unsure how to debug this.

loriab:
gcc-only for the `psi4numpy-rhf-hessian` trouble does seem strange. In any case, can you rebase when you have a chance, @schiebermc, and retry a full tests (libxc is going to build from source now)?

loriab:
I've run full tests with both py27 & py36 and both report the following. The last two are b/c hungarian not installed – not a problem. The first two are b/c MemDFJK isn't handling Cartesian basis sets right. I'm not sure how those other errors you were seeing came about, @schiebermc. Was there a point in this branch where full tests was clean? There's been upheaval to `master`, so we could reason from there.

```
The following tests FAILED:
	187 - fd-freq-energy-large (Failed)
	212 - mints8 (Failed)
	375 - pywrap-align (Failed)
	376 - pywrap-align-chiral (Failed)
```

dgasmith:
Well, still confused why the p4n-rhf-hessian is simply timing out. Anyone have a GCC that they can try to replicate with?

@schiebermc Thats why code review exists!

loriab:
The clean intel builds have been atop gcc, but I'll try an outright gcc 7.2.0.

loriab:
I'm shocked. `psi4numpy-rhf-hessian` hanging is completely reproducible with gcc locally. Investigating.

loriab:
Trouble is nothing to do with Hessian (it's hanging at the preliminary energy calc) and everything to do with setting more than one thread.

```
(p4dev36) psilocaluser@bash:psinet:/home/psilocaluser/gits/hrw-qcdb/objdir-gnu36: ((detached from matt/SYMM_JK)) cat ../tests/psi4numpy/rhf-hessian/input.py 
# -*- coding: utf-8 -*- 

import time
import numpy as np
np.set_printoptions(precision=15, linewidth=200, suppress=True)
import psi4

#psi4.set_memory(int(1e9), False)
#psi4.core.set_output_file('output.dat', False)
psi4.core.set_num_threads(2)

mol = psi4.geometry("""
O
H 1 1.1
H 1 1.1 2 104
symmetry c1
""")

psi4.core.set_active_molecule(mol)

options = {'BASIS':'STO-3G', 'SCF_TYPE':'PK',
           'E_CONVERGENCE':1e-10,
           'D_CONVERGENCE':1e-10
           }

psi4.set_options(options)

rhf_e, wfn = psi4.energy('SCF', return_wfn=True)
```

Debug traceback shows hang is in SAD guess at `DFHelper::contract_metric_AO_core_symm`.

```
(p4dev36) psilocaluser@bash:psinet:/home/psilocaluser/gits/hrw-qcdb/objdir-gnu36: ((detached from matt/SYMM_JK)) gdb --args python ../tests/psi4numpy/rhf-hessian/input.py 
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-94.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /home/psilocaluser/toolchainconda/envs/p4dev36/bin/python3.6...done.
(gdb) run
Starting program: /home/psilocaluser/toolchainconda/envs/p4dev36/bin/python ../tests/psi4numpy/rhf-hessian/input.py
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Missing separate debuginfo for /home/psilocaluser/toolchainconda/envs/p4dev36/lib/python3.6/site-packages/numpy/../../../libiomp5.so
Detaching after fork from child process 16215.
Detaching after fork from child process 16216.
Detaching after fork from child process 16217.
Detaching after fork from child process 16218.
Detaching after fork from child process 16219.
  Threads set to 2 by Python driver.

*** tstart() called on psinet
*** at Wed Apr 25 23:55:59 2018

   => Loading Basis Set <=

    Name: STO-3G
    Role: ORBITAL
    Keyword: BASIS
    atoms 1   entry O          line    81 file /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/share/psi4/basis/sto-3g.gbs 
    atoms 2-3 entry H          line    19 file /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/share/psi4/basis/sto-3g.gbs 

    There are an even number of electrons - assuming singlet.
    Specify the multiplicity in the molecule input block.


         ---------------------------------------------------------
                                   SCF
            by Justin Turney, Rob Parrish, Andy Simmonett
                             and Daniel Smith
                              RHF Reference
                        2 Threads,    500 MiB Core
         ---------------------------------------------------------

...

  ==> Pre-Iterations <==

   -------------------------------------------------------
    Irrep   Nso     Nmo     Nalpha   Nbeta   Ndocc  Nsocc
   -------------------------------------------------------
     A          7       7       0       0       0       0
   -------------------------------------------------------
    Total       7       7       5       5       5       0
   -------------------------------------------------------

[New Thread 0x7fffd6116780 (LWP 16220)]
  ==> Integral Setup <==

  Using in-core PK algorithm.
   Calculation information:
      Number of atoms:                   3
      Number of AO shells:               5
      Number of primitives:             15
      Number of atomic orbitals:         7
      Number of basis functions:         7

      Integral cutoff                 1.00e-12
      Number of threads:                 2

  Performing in-core PK
  Using 812 doubles for integral storage.
  We computed 133 shell quartets total.
  Whereas there are 120 unique shell quartets.
    10.83 percent of shell quartets recomputed by reordering.
  ==> DiskJK: Disk-Based J/K Matrices <==

    J tasked:                  Yes
    K tasked:                  Yes
    wK tasked:                  No
    Memory (MB):               375
    Schwarz Cutoff:          1E-12

    OpenMP threads:              2
  Minimum eigenvalue in the overlap matrix is 4.3422219878E-01.

^C
Program received signal SIGINT, Interrupt.
0x00007fffe3ca696f in psi::DFHelper::contract_metric_AO_core_symm(double*, double*, unsigned long, unsigned long) [clone ._omp_fn.16] ()
   from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
Missing separate debuginfos, use: debuginfo-install glibc-2.17-157.el7_3.1.x86_64 libuuid-2.23.2-33.el7.x86_64
(gdb) bt
#0  0x00007fffe3ca696f in psi::DFHelper::contract_metric_AO_core_symm(double*, double*, unsigned long, unsigned long) [clone ._omp_fn.16] ()
   from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#1  0x00007fffec18dc21 in __kmp_api_GOMP_parallel_40_alias () from /home/psilocaluser/toolchainconda/envs/p4dev36/lib/python3.6/site-packages/numpy/../../../libiomp5.so
#2  0x00007fffe3cab007 in psi::DFHelper::contract_metric_AO_core_symm(double*, double*, unsigned long, unsigned long) ()
   from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#3  0x00007fffe3cbb4ae in psi::DFHelper::prepare_AO_core() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#4  0x00007fffe3cbbb99 in psi::DFHelper::initialize() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#5  0x00007fffe3cd9c2c in psi::MemDFJK::preiterations() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#6  0x00007fffe3c6d3bc in psi::scf::SADGuess::get_uhf_atomic_density(std::shared_ptr<psi::BasisSet>, std::shared_ptr<psi::BasisSet>, int, int, std::shared_ptr<psi::Matrix>) ()
   from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#7  0x00007fffe3c70ed3 in psi::scf::SADGuess::form_D_AO() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#8  0x00007fffe3c7205f in psi::scf::SADGuess::form_D() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#9  0x00007fffe3c7291f in psi::scf::SADGuess::compute_guess() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#10 0x00007fffe3c72e7a in psi::scf::HF::compute_SAD_guess() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#11 0x00007fffe3c40435 in psi::scf::HF::guess() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#12 0x00007fffe3c3b566 in psi::scf::HF::initialize() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#13 0x00007fffe3c346ed in psi::scf::HF::compute_energy() () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#14 0x00007fffe2fa46ce in _ZZN8pybind1112cpp_function10initializeIZNS0_C4IdN3psi12WavefunctionEJEJNS_4nameENS_9is_methodENS_7siblingEA41_cEEEMT0_FT_DpT1_EDpRKT2_EUlPS4_E_dJSJ_EJS5_S6_S7_S8_EEEvOSA_PFS9_SC_ESI_ENUlRNS_6detail13function_callEE1_4_FUNESQ_ () from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#15 0x00007fffe2dba8b2 in pybind11::cpp_function::dispatcher(_object*, _object*, _object*) ()
   from /home/psilocaluser/gits/hrw-qcdb/objdir-gnu36/stage/usr/local/psi4/lib/psi4/core.so
#16 0x0000555555663754 in _PyCFunction_FastCallDict ()
#17 0x00005555556f2c1e in call_function ()
#18 0x00005555557154ba in _PyEval_EvalFrameDefault ()
#19 0x00005555556eda39 in PyEval_EvalCodeEx ()
#20 0x00005555556ee956 in function_call ()
#21 0x000055555566355e in PyObject_Call ()
#22 0x0000555555716b39 in _PyEval_EvalFrameDefault ()
#23 0x00005555556eda39 in PyEval_EvalCodeEx ()
#24 0x00005555556ee956 in function_call ()
#25 0x000055555566355e in PyObject_Call ()
#26 0x0000555555716b39 in _PyEval_EvalFrameDefault ()
#27 0x00005555556ebf06 in _PyEval_EvalCodeWithName ()
#28 0x00005555556ecf4f in fast_function ()
#29 0x00005555556f2ba5 in call_function ()
#30 0x0000555555716279 in _PyEval_EvalFrameDefault ()
#31 0x00005555556eda39 in PyEval_EvalCodeEx ()
#32 0x00005555556ee7cc in PyEval_EvalCode ()
#33 0x0000555555768ee4 in run_mod ()
#34 0x00005555557692e1 in PyRun_FileExFlags ()
#35 0x00005555557694e4 in PyRun_SimpleFileExFlags ()
#36 0x000055555576cfdf in Py_Main ()
#37 0x0000555555634bfe in main ()
```

This is linked to mkl_rt & iomp5. I also tried mkl_rt & gomp with the link line advisor syntax. No change.

dgasmith:
@loriab Could potentially be [this](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/lib3index/df_helper.cc#L1335) line. Could you change the schedule to simply `schedule(static)`? Either way we now have a reproducible error that can be fixed, thanks so much!

loriab:
Yep, suggested change heals it. Matt or you can include it among your commits.

```
-#pragma omp parallel for num_threads(nthreads_) schedule(static, nao_ / nthreads_)
+#pragma omp parallel for num_threads(nthreads_) schedule(static) //, nao_ / nthreads_)
```

dgasmith:
Yup, bet were hitting a `num_threads(nthreads_) schedule(static, 0)` there. We can add a bit of logic to ensure this doesn't go below 1.

dgasmith:
Two remaining issues:
1) Non-symmetric JK builds use too much memory.
2) Results are incorrect when there is a cartesian auxiliary basis (primary is correct with both spherical and cartesian).

loriab:
##### Watch Out For

Matt's seeing consistent failures among certain functionals in `dft-bench-ionization`. Particularly the first and third (`PWB6K` & `wB97X-D`) while the second is fine. I don't see the problem with Intel `2018.0.2` compilers, and travis doesn't see it in a truncated test case (clang and gcc).

So watch out for this, and even if you cut off full tests prematurely b/c the benches are so long, please go ahead and run at least the first few functionals of the test outright for a few weeks so we can get more info.

CDSherrill:
I'm ok with adding a throw for Cartesian auxiliary basis sets, although
post-PR I'd like work to continue on trying to track this down.

loriab:
As I recall, switching the to disk_jk does heal them. Forcing that does seem the best route, as we want to get this in tomorrow.

schiebermc:
Forcing to disk_jk fixes the spherical-cartesian issue, but reproduces the same incorrect energy for the `dft-bench-ionization`.

dgasmith:
Ok, this sounds like a XC issue in this case. Can someone open an issue on it showing the different SCF iterations under different compilers?

@schiebermc For the cartesian auxiliary the syntax should look something like `aux_->has_puream()`. This is `True` for spherical, `False` otherwise. I think this logic needs to be handled in both the JK constructor to switch between different DF setups and in the `MemDFJK` to throw if someone passes a cartesian auxiliary. Can you handle these changes?

schiebermc:
Thanks for the tip @dgasmith, should be fixed now.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -168,8 +168,9 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
     if (jk_type == "DF") {
 
         // logic for MemDFJK vs DiskDFJK
-        if (do_wK || options["DF_INTS_IO"].has_changed()) {
+        if (do_wK || !auxiliary->has_puream() || options["DF_INTS_IO"].has_changed()) {
```
dgasmith: Please add a throw in DFHelper as well, this will likely effect your 3-index transforms and especially if someone builds `MemDFJK` directly.

psi4/src/read_options.cc
```diff
@@ -1199,7 +1199,7 @@ int read_options(const std::string &name, Options & options, bool suppress_print
     /*- What algorithm to use for the SCF computation. See Table :ref:`SCF
     Convergence & Algorithm <table:conv_scf>` for default algorithm for
     different calculation types. -*/
```
dgasmith: @loriab Where is a good place to add discussion between `DF/MEM_DF/DISK_DF`?

psi4/src/psi4/libfock/jk.cc
schiebermc: done

psi4/src/read_options.cc
loriab: somewhere below [here](https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/scf.rst#eri-algorithms) or another section if need be.

hokru:
The MINI basis is originally cartesian? Grimme's MINIX is used in a spherical context and the "3c" part is fitted as such. The additional p-functions and the heavier elements are all taken from spherical basis sets.

I recommend to label MINIX spherical.

dgasmith:
Pretty easy fix for SAD. Im ok doing it there as we are pretty loose on the fitting basis set anyways. Is there a good reason the the auxiliary basis is always in the basis sets puream? Spherical auxiliary would probably save a bit of time overall as the angular momenta there is always pushed quite a bit higher.

loriab:
It was thought more physically sound for orbital puream to govern all. After we noticed inconsistency, there was a crackdown and [long test](https://github.com/psi4/psi4/blob/master/tests/castup3/input.dat). Though my faint recollection was that the main trouble came from cast-up basis sets influencing the puream of the main calc. So a pure-sph might be ok.

loriab:
@hokru, happily the `hf3c` test case tests interaction energies, so a switch of MINIX to spherical passes w/o a hitch. Any objections from @psi4/maintainers on switch?

andysim:
Super happy to see this one merged - great work!

PR #963, opened 2018-04-13
Allow more basis sets into CBS methods

PeterKraus:
## Description
Allow Alhrichs (`def2-`) and custom Dunning (*cc-[...]z*) families of basis sets into CBS expansion.

## Todos
* **Developer Interest**
  - [ ] Feature1
  - [x] CBS extrapolation possible with `def2-` bases
  - [x] CBS extrapolation possible with eg. cc-pv[dt]z-PP bases
  - [ ] handle fallbacks for light atoms in `[aug]-cc-pv[dtq]z-PP` bases using non-PP bases
  - [ ] tests for `def2-` extrapolation

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/driver_cbs.py
+++ b/psi4/driver/driver_cbs.py
@@ -59,26 +59,29 @@ def _expand_bracketed_basis(basisstring, molecule=None):
     the qcdb.BasisSet object for *molecule* or for H2 if None. Allows
     out-of-order zeta specification (e.g., [qtd]) and numeral for number
     (e.g., [23]) but not skipped zetas (e.g., [dq]) or zetas outside [2,
-    8] or non-Dunning sets or non-findable .gbs sets.
+    8] or non-Dunning or non-Ahlrichs sets or non-findable .gbs sets.
 
     """
     BSET = []
     ZSET = []
-    legit_compound_basis = re.compile(r'^(?P<pre>.*cc-.*)\[(?P<zeta>[dtq2345678,]*)\](?P<post>.*z)$', re.IGNORECASE)
+    legit_compound_basis = re.compile(r'^(?P<pre>.*cc-.*|def2-)\[(?P<zeta>[dtq2345678,s]*)\](?P<post>.*z.*)$', re.IGNORECASE)
 
     if legit_compound_basis.match(basisstring):
         basisname = legit_compound_basis.match(basisstring)
         # filter out commas and be forgiving of e.g., t5q or 3q
-        bn_gz = basisname.group('zeta')
+        bn_gz = basisname.group('zeta').replace("s","d")
         zetas = [z for z in zeta_values if (z in bn_gz or str(zeta_values.index(z) + 2) in bn_gz)]
         for b in zetas:
             if ZSET and (int(ZSET[len(ZSET) - 1]) - zeta_values.index(b)) != 1:
-                    raise ValidationError("""Basis set '%s' has skipped zeta level '%s'.""" % (basisstring, b))
-            BSET.append(basisname.group('pre') + b + basisname.group('post'))
+                raise ValidationError("""Basis set '%s' has skipped zeta level '%s'.""" % (basisstring, zeta_val2sym[zeta_sym2val[b] - 1]))
+            if basisname.group('pre') == "def2-" and b == "d":
+                BSET.append(basisname.group('pre') + "s" + basisname.group('post')[1:])
+            else:
+                BSET.append(basisname.group('pre') + b + basisname.group('post'))
             ZSET.append(zeta_values.index(b) + 2)
     elif re.match(r'.*\[.*\].*$', basisstring, flags=re.IGNORECASE):
         raise ValidationError("""Basis series '%s' invalid. Specify a basis series matching"""
-                              """ '*cc-*[dtq2345678,]*z'.""" % (basisstring))
+                              """ '*cc-*[dtq2345678,]*z*'. or 'def2-[sdtq]zvp*'""" % (basisstring))
     else:
         BSET.append(basisstring)
         ZSET.append(0)
@@ -109,8 +112,8 @@ def _contract_bracketed_basis(basisarray):
         zetaindx = [i for i in range(len(basisarray[0])) if basisarray[0][i] != basisarray[1][i]][0]
         ZSET = [bas[zetaindx] for bas in basisarray]
 
-        pre = basisarray[0][:zetaindx]
-        post = basisarray[0][zetaindx + 1:]
+        pre = basisarray[1][:zetaindx]
+        post = basisarray[1][zetaindx + 1:]
         basisstring = pre + '[' + ''.join(ZSET) + ']' + post
         return basisstring
 


loriab:
Do you happen to know if there's a lit citation for extrapolating the def2 basis sets or if they need a different extrapolation formula? I know Orca does def2 extrapolations; I've just never looked up how or what procedure they use.

I'd defer the -PP fallbacks from this PR. That means changing the basis set compositing which is a huge pain. I'm hoping the MolSSI basisset project can be API-accessed to cover this (very sensible) case.

Hand-pieced, other program, or `energy(cbs, ...)` call with calculator spot-checks will do fine for tests. If def2's need their own extrapolation formula, tests should probably check that those default correctly.

hokru:
The fitted exponents using in ORCA for standard Halkier 2-point extrapolation are from F. Neese, E.F. Valeev, JCTC 7, 33-43 (2011), where also the def2s are reported, which is perhaps the most appropriate citation.

No special formula is needed, but only the def2-XZPP variant have correlation consistent polarization functions. The others will work less well, especially for heavier atoms.

Older, mostly application works by Grimme already used def2 in extrapolations, but the Neese/Valeev is a good citation.

PeterKraus:
@hokru, thanks for the reference. It would seem based on that paper that we should:
- allow the `roos-ano-[dt]z` basis sets into CBS as well
- are the `pcseg-[0-4]` basis sets (double- to six-zeta, I believe) the segmented versions of `pc-X` bases? There might be an argument in letting those in as well, based on the Neese paper.

I don't suppose we can handle the generally contracted ANO's in Psi4? They're not included in our basis set library anyway.

In terms of coefficients, I probably don't understand exactly what's going on, but:
- for `corl_xtpl_helgaker_2` we use a hard-coded `beta = 3`, which is the physically correct exponent, but for extrapolation between from `[DT]` basis sets the value of `2.46` seems more appropriate. There is currently no way to supply a custom value
- for `scf_xtpl_helgaker_2` (both of these should really be called `scf_xtpl_halkier_[2,3]`): we use the exponential form which seems to perform a lot better than the power form according to Halkier et al. The `alpha=1.63` seems to be a general parameter here, however a zeta-based `alpha` seems to perform better (as it is fitted). The interface to supply `alpha` is in the function, but doesn't seem to be plugged into the CBS machinery
- the "power form" for SCF extrapolation is not implemented in Psi4, but the coefficent `alpha` seems to be very system and basis and zeta dependent, as shown by both Halkier et al. and Neese and Valeev.

@loriab My suggestions for myself for this PR, in addition to allowing the two basis set classes above, would be:
- [ ] add power form for SCF extrapolation (eg. `scf_xtpl_power_2`) with user-supplied `alpha`, a sane default is `alpha=3.54` from Halkier et al.
- [ ] add ability to supply custom `beta` for `corl_xtpl_helgaker_2` for better performance with [DT]'s
- [ ] wire up `scf_xtpl_helgaker_2` parameter `alpha` and the other two from above with the rest of the CBS machinery, perhaps using `scf_alpha` and `corl_beta` parameters to `cbs()`

The `cc-pVXZ-PP` issue can certainly wait. The user can combine their own basis sets into user-supplied `.gbs`'s (which is what I was doing for my systems anyway) and the CBS function will happily use them, as long as they match the regex.

hokru:
Several 2-point extrapolation formula are mathematically equivalent as shown by Kanton and Martin
10.1007/s00214-005-0028-6

Several fitted exponents are found here: 10.1039/c6cp00688d
Note that in the latter paper the extrapolation formula is wrong, E_inf and E_L needs to be swapped. 

IMO, the fitted exponents nicely show the suitability of the basis sets for extrapolation and how the double-zeta basis sets form a poor foundation.

loriab:
* Hmm, if you click on "More information" on pc-seg4 on EMSL, you get "Polarization Consistent Valence Pentuble Zeta, Segmented Contracted". So that's my source iirc, but EMSL has low authority these days. Yes, please do switch to `N+2` in NOTES in this PR. Thanks for the better information.
* `delta N > 2` should probably be rewritten to be extensible w/o copy-paste-sed, so don't bother future-proofing in that respect.
* General plan is that parts of the driver that are generic quantum chemistry should be useable w/o a psi4 dependence, hence the separate qcdb repo. Glad to explain in more detail later. No need to rebase off anything but psi4/psi4 master.
* Sorry, I should have worded that better. Yes, the xtpl fns want `psi4.core.Matrix` for gradient/Hessian extrapolations as input and output. But for all the manipulations of those arrays, you have the choice of keeping them as `Matrix` and using psi internals or to grabbing them as numpy arrays and using all the numpy manipulation syntax then returning to `Matrix` at the end. I just meant to say that if you're facing this choice, choose NumPy because the next version is going to take and return `ndarray`, not `psi4.core.Matrix`.

Thanks!

PeterKraus:
@loriab I think this is now ready for review. The values in the new test are hand-calculated from the single point values using a spreadsheet. Let me know if I should just boilerplate the docstrings from the `qcdb.cbs_helpers` file - and whether I should change just the missing functions or the old ones as well.

One thing that is perhaps missing is handling of different `corl_alpha`'s for delta-corrections; this might actually be a useful feature, as one is more likely to drop down to [DT] extrapolation (for which `alpha = 2.4` gives better results) with things like CCSD(T) while using MP2/[TQ] as the `corl_wfn/basis`. It should be an easy fix.

I left the `pcseg-N` basis sets out for now; I'll read the literature and see whether it makes sense. I still plan to allow the `roos-ano-[dt]z` basis sets into the CBS functions.

psi4/driver/driver_cbs.py
```diff
@@ -278,21 +434,26 @@ def scf_xtpl_helgaker_3(functionname, zLO, valueLO, zMD, valueMD, zHI, valueHI,
...
 
 #def corl_xtpl_helgaker_2(functionname, valueSCF, zLO, valueLO, zHI, valueHI, verbose=True):
-def corl_xtpl_helgaker_2(functionname, zLO, valueLO, zHI, valueHI, verbose=True):
+def corl_xtpl_helgaker_2(functionname, zLO, valueLO, zHI, valueHI, verbose=True, **kwargs):
     r"""Extrapolation scheme for correlation energies with two adjacent zeta-level bases.
     Used by :py:func:`~psi4.cbs`.
     Halkier, Helgaker, Jorgensen, Klopper, Koch, Olsen, & Wilson, Chem. Phys. Lett. 286 (1998) 243-252.
 
-    .. math:: E_{corl}^X = E_{corl}^{\infty} + \beta X^{-3}
+    .. math:: E_{corl}^X = E_{corl}^{\infty} + \beta X^{-alpha}
```
loriab: `, \alpha = 3.0`

psi4/driver/p4util/p4regex.py
```diff
@@ -29,8 +29,8 @@
 from __future__ import absolute_import
 import re
 
-yes = re.compile(r'^(yes|true|on|1)', re.IGNORECASE)
-no = re.compile(r'^(no|false|off|0)', re.IGNORECASE)
+yes = re.compile(r'^(yes|true|on|1\.?0*$)', re.IGNORECASE)
+no = re.compile(r'^(no|false|off|0\.?0*$)', re.IGNORECASE)
```
loriab: I'm uncomfortable with `1.0` turning `True`. Undo these, I think, and add `if isinstance(lvalue, float):\n\tpass` [here](https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/procutil.py#L77).

psi4/driver/driver_cbs.py
```diff
@@ -212,7 +220,155 @@ def scf_xtpl_helgaker_2(functionname, zLO, valueLO, zHI, valueHI, verbose=True,
...
+        raise ValidationError("scf_xtpl_truhlar_2: datatype is not recognized '%s'." % type(valueLO))
+
+
+def scf_xtpl_karton_2(functionname, zLO, valueLO, zHI, valueHI, verbose=True, **kwargs):
+    r"""Extrapolation scheme using root-power form for reference energies with two adjacent zeta-level bases.
+    Used by :py:func:`~psi4.cbs`.
+    Karton, Martin, Theor. Chem. Acc. 115 (2006) 330-333.
+
+    .. math:: E_{total}^X = E_{total}^{\infty} + \beta e^{-\alpha\sqrt{X}}, \alpha = 6.3
+
```
loriab: These docstrings will do fine for now. The eqn and ref are the crucial points.

psi4/driver/driver_cbs.py
```diff
@@ -855,7 +1017,9 @@ def cbs(func, label, **kwargs):
     return_wfn = kwargs.pop('return_wfn', False)
     verbose = kwargs.pop('verbose', 0)
     ptype = kwargs.pop('ptype')
-
+    scf_alpha = kwargs.pop('scf_alpha', False)
+    corl_alpha = kwargs.pop('corl_alpha', False)
```
loriab: Throw an item about the alphas into the docstring above, I think. scf_scheme, scf_alpha, corl_scheme, corl_alpha, delta_scheme, delta_alpha

psi4/driver/driver_cbs.py
```diff
@@ -1332,20 +1496,26 @@ def cbs(func, label, **kwargs):
     finalenergy = 0.0
     finalgradient = core.Matrix(natom, 3)
     finalhessian = core.Matrix(3 * natom, 3 * natom)
+    hiloargs = dict()
+    if scf_alpha:
+        hiloargs["scf_alpha"] = scf_alpha
+    if corl_alpha:
+        hiloargs["corl_alpha"] = corl_alpha
```
loriab: I think I have a plan on how to get independent `delta_alpha`. Let me look a bit further.

psi4/driver/driver_cbs.py
PeterKraus: Fixed.

psi4/driver/p4util/p4regex.py
PeterKraus: Well, it was returning true before for all numerals starting with 1 and 0... Is the current version (matching only "1" and "0" - no floats) acceptable?

psi4/driver/driver_cbs.py
PeterKraus: Fixed - I'm not super confident with sphinx syntax, I only ever used it for small things, so let me know if there are some bugs.

psi4/driver/driver_cbs.py
PeterKraus: I've implemented it for the first delta. If you plan to re-work the delta handling in the future anyway, perhaps it's not that important to extend it to further deltas - what would a two-delta method look like, anyway?

PeterKraus:
Passes all tests but:
```
The following tests FAILED:
	370 - pywrap-align (Failed)
	371 - pywrap-align-chiral (Failed)
```
The two crash with import error; I guess that's just an environment issue.

psi4/driver/driver_cbs.py
```diff
@@ -278,21 +444,25 @@ def scf_xtpl_helgaker_3(functionname, zLO, valueLO, zMD, valueMD, zHI, valueHI,
 
 
 #def corl_xtpl_helgaker_2(functionname, valueSCF, zLO, valueLO, zHI, valueHI, verbose=True):
-def corl_xtpl_helgaker_2(functionname, zLO, valueLO, zHI, valueHI, verbose=True):
+def corl_xtpl_helgaker_2(functionname, zLO, valueLO, zHI, valueHI, verbose=True, alpha=None):
```
dgasmith: Can we do a `alpha=3` and indicate this default in the description. Using the workaround below is only really useful for non-atomics like lists.

PeterKraus: Just to be clear, you mean this for all functions, not just the one above?

dgasmith: Ah, sorry I didn't know you were around so I fixed them directly. Please review the following: https://github.com/psi4/psi4/pull/963/commits/948cc7d4d7eddb5434715a12b34deb9840ccd215

loriab: Your comment makes sense, @dgasmith, but I think it doesn't work b/c `alpha` is always passed, so the `None` is needed to indicate no user involvement.

`hiloargs = {'alpha': cbs_alpha[stage['d_stage']]}`

```
cbs_alpha = {
+        'scf': kwargs.get('cbs_scf_alpha', kwargs.get('scf_alpha', None)),
+        'corl': kwargs.get('cbs_corl_alpha', kwargs.get('corl_alpha', None)),
+        'delta': kwargs.get('cbs_delta_alpha', kwargs.get('delta_alpha', None)),
+        'delta2': kwargs.get('cbs_delta2_alpha', kwargs.get('delta2_alpha', None)),
+    }
```

PR #965, opened 2018-04-16
Mol & Vib polishing

loriab:
## Description
This separates all the additional Molecule work, including py-side parsing, out from the py-side SCF/PCM/EFP PR. I recommend this go into 1.2 so that 1.2 Mol isn't half-baked. This **does not switch Mol parsing py-side**. I can, but that's an extra step to pull out of py-side SCF/PCM/EFP and a more drastic change to Psi.

## Todos
* **Developer Interest**
  - [x] `qcdb.Molecule` constructor learned to take, str (psi4, xyz, xyz+, psi4+), dict, or arrays and pass things on to `from_string`, `from_dict`, or `from_arrays` accordingly. In fact, these fns are simply different entry points. `from_arrays` validates arguments then calls `from_dict`. `from_string` parses into arrays then calls `from_arrays`. The three `from_*` methods are also available for `psi4.core.Molecule` (completely bypassing `create_molecule_from_string`), but the primary c-side Mol constructor in molutil.py is still using `c_m_f_s`.
  - [x] py-side & c-side Molecule & CoordEntry classes learned `A` or `mass_number`. Pretty much this is a pass-through/holder slot for info populated in PQ Mol JSON. `set_mass` voids `A`.
  - [x] `*.Molecule` learned `add_unsettled_atom` for adding `ZMatrixEntry`s analogous to `add_atom` `CartesianEntry`s for the zmat extension.
  - [x] `qcdb.BasisSet.pyconstruct` learned to act on anything it can make a `qcdb.Molecule` of: Mol, str, or dict. It also learned to return BasisSet object or dict based on `return_dict` value, not on type of molecule-ish entity passed in.
  - [x] changed some output unit labels to be more specific or use known abbr
  - [x] In keywords (`read_options`), remove global `UNITS` since unused and covered by Mol parser
  - [x] Write `.hess`, `.normal_modes_molden`, and a proto-JSON for vib/thermo analysis py-side.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


cmake/custom_cxxstandard.cmake
```diff
@@ -6,8 +6,9 @@ if (CMAKE_CXX_COMPILER_ID MATCHES GNU)
     endif()
 
 elseif (CMAKE_CXX_COMPILER_ID MATCHES Intel)
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "16.0.2")
-        message(FATAL_ERROR "ICPC version must be at least 2016.0.2!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS "17.0.0")
+        message(FATAL_ERROR "ICPC version must be at least 2017.0.0 to work with pybind11 2.2.2!")  # v1.2
```
andysim: I *think* we also determined that 2015 is ok.  Not suggesting we change the minimum, but just making a note in case we encounter user pushback.

psi4/driver/procrouting/proc.py
```diff
@@ -2140,52 +2140,8 @@ def run_scf_hessian(name, **kwargs):
...
-    # This is a drop-in replacement for the code that lives in findif
-    if core.get_option('FINDIF', 'HESSIAN_WRITE'):
-        molname = ref_wfn.molecule().name()
-        prefix = core.get_writer_file_prefix(molname)
-        with open(prefix+".hess", 'w') as fp:
-            fp.write("%5d%5d\n" % (natoms, 6*natoms))
-            for row in np.reshape(H, (-1, 3)):
-                fp.write("%20.10f%20.10f%20.10f\n" % tuple(row))
-
     optstash.restore()
```
andysim: YAY!  Thanks for the massive upgrade of this ugly hack

psi4/driver/qcdb/molparse/from_string.py
```diff
@@ -0,0 +1,721 @@
...
+            ----------
+            Specifiable: units, fix_com/orientation/symmetry, fragment_files, hint_types, geom_hints
+            Inaccessible: anything atomic or fragment details -- geom, elem/elez (element identity),
+                          mass, real (vs. ghost), elbl (user label), fragment_separators, fragment_charges,
+                          fragment_multiplicities, molecular_charge, molecular_multiplicity
+
+        psi4+ - Psi4 non-Cartesian molecule {...} format
+        ------------------------------------------------
+        Like `dtype=psi4` (although combination with EFP not tested) except
+        that instead of pure-Cartesian geometry, allow variables, zmatrix,
```
andysim: I forget if we have documentation of the fact that zmatrices are not really used for optimization purposes.  Might be worth noting that here.

cmake/custom_cxxstandard.cmake
loriab: So there was an earlier episode where we (or pb11) found that 2015 would work, with a hack, and 2016 would work out-of-the-box. That was c. pb11 2.0.0.

And later there was [this](https://github.com/pybind/pybind11/issues/1121) where we found that with pb11 2.1.0 and above that not even Ben could tweak the pb11 code to please Intel 2016. Hence the minimum 2017 in the snippet above.

Which episode are you remembering the 2015-is-ok from?

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1347,15 +1347,15 @@ void OEProp::compute_quadrupole(bool transition)
 
     // Print multipole components
     double dfac = pc_dipmom_au2debye * pc_bohr2angstroms;
-    outfile->Printf( "  %sQuadrupole Moment: (Debye Ang)\n", (transition ? "Transition " : ""));
+    outfile->Printf( "  %sQuadrupole Moment: [D Å]\n", (transition ? "Transition " : ""));
```
andysim: Doesn't the funny A cause weird output on some terminals?

loriab: Drat, I thought I had run my non-ascii checker over the codebase again. I don't know about users, but sphinx and conda really don't like non-ascii.

Thanks, I'll fix it. Perils of reordering and splitting all your commits.

cmake/custom_cxxstandard.cmake
andysim: The most unreliable source - my memory.  I think I did some experimentation and got it working but, as you noted, that was likely an earlier PB.  If we get requests for older compiler support, I can mess around with icpc 15 for Linux but not OSX.  Either way, I fully support the icpc17 minimum.

psi4/driver/procrouting/proc.py
loriab: A bit of a hack, yes, but these 40 lines accomplished for analytic Hessians what half a c++ module did for findif Hessians.

psi4/driver/qcdb/molparse/from_string.py
loriab: Huh, apparently we did (".. warning:: The selection of a Z-matrix input, and in particular the inclusion
   of dummy atoms, has no effect on the behavior of the optimizer, which begins
   from a Cartesian representation of the system."), not that I remembered it.

I added it here, too, for good measure. Good idea — it needs drumming in.

loriab:
@andysim , reading https://github.com/psi4/psi4/pull/965#pullrequestreview-112407645 again, I think I may have misinterpreted it. I read it as: since the original pubchem interface, pubchem has started to send back symmetrized mols, so the symmetrization currently applied by libmints Mol upon "I'm from pubchem" trigger is no longer necessary.

Reading again, I'm betting you mean: no one's doing the symmetrize step that pubchem inputs need.

Still a piece missing, right?

andysim:
@loriab Correct: I believe there's a symmetrization step missing.  I forget exactly where it was done before, but things like benzene are commonly retrieved with coords that are effectively C1.  There's (molecule?) member function somewhere that allows symmetrization to a given tolerance (something really sleazy like 0.01A is needed, IIRC) so it's just a case of calling that.  A simple STO-3G optimization could be a good followup step, but that's not essential.

andysim:
Absolutely.  Not intended to be a block, but something to keep in mind if we have issues with C1 symmetry structures coming from pubchem.

doc/sphinxman/document_tests.pl
```diff
@@ -126,7 +126,7 @@
     my $TestOnlyNoSample = 0;
     while(<INPUT>){
         # If this line isn't associated with testing, put it in the sample
-        print SAMPLE unless /\#TEST/;
+        print SAMPLE unless /\#\s*TEST/;
```
dgasmith: This may, unfortunately, be the most critical change of the PR.

psi4/driver/qcdb/molecule.py
```diff
@@ -51,14 +56,101 @@ class Molecule(LibmintsMolecule):
...
+                    fragment_multiplicities=fragment_multiplicities,
+                    molecular_charge=molecular_charge,
+                    molecular_multiplicity=molecular_multiplicity,
+                    domain='qm',
+                    missing_enabled_return=missing_enabled_return,
+                    tooclose=tooclose,
+                    zero_ghost_fragments=zero_ghost_fragments,
+                    nonphysical=nonphysical,
+                    mtol=mtol,
+                    verbose=verbose)
```
dgasmith: When we started out serialize and deserialize this kind of thing, never thought we would end up with structures like this. Makes sense however.

psi4/src/export_mints.cc
```diff
@@ -225,32 +225,57 @@ std::shared_ptr<Molecule> from_dict(py::dict molrec) {
...
+        std::vector <double> geom = molrec["geom"].cast<std::vector <double>>();
+        nat = geom.size() / 3;
+        unsettled = false;
+
+        for (size_t iat=0; iat<nat; ++iat) {
+            std::string symbol = elem[iat];
+            std::string label = elbl[iat];
+            std::transform(symbol.begin(), symbol.end(), symbol.begin(), ::toupper);
+            std::transform(label.begin(), label.end(), label.begin(), ::toupper);
+            mol->add_atom(elez[iat] * real[iat],
```
dgasmith: Should we have more checks here to make sure say the lengths are the same? You can also use `eleze.at(iat)` for bounds check access.

tests/python/mints13/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
 include(TestingMacros)
 
-add_regression_test(python-mints13 "psi;mints;quicktests;python")
+add_regression_test(python-mints13 "psi;mints;quicktests;python;cart")
```
dgasmith: What is `cart`?

loriab: Those are the test cases whose Mols use no zmat spec or variables or efp or pubchem, just plain Cartesian. It was tedious enough to find and add them that I left them in for future.

psi4/src/export_mints.cc
loriab: Thanks. `from_dict`s aren't revalidated (so hand-editing a molrec is not a safe operation), but it makes sense to add checks like this where easy.

PR #975, opened 2018-04-23
N-Body driver refactor (fixes #885)

### LINKED ISSUE #885: Split up n-body calls
Currently the `n-body` code is a single call that figures out required computations, computes those requirements, and then stitches the whole thing back together again. It would be great if these were in three separate functions so that future sow/reap tech could parallelize these across many nodes.

In `psi4/drive/nbody_driver.py` the primary function should be split at lines 291 and 312. into these three separate parts. We will then need one new function which runs through all three steps so that current functionality can be reproduced.

dsirianni:
## Description
Addresses #885 by refactoring N-Body driver to three newly independent functions, driven by the existing `nbody_gufunc()`:
- `build_nbody_compute_list()`: Builds list of all N-body computations to be performed
- `compute_nbody_components()`: Performs given N-body computation(s), could be exploited as pleasantly parallel
- `assemble_nbody_components()`: Stitches components back together into _bona fide_ N-body energies (currently standard behavior of N-body driver)

## Todos
* **Developer Interest**
  - Opportunity for "pleasantly parallel" N-body computations
  - UI to N-Body driver unchanged
  - Standard N-body calls, e.g., `energy(..., bsse_type='cp')` retain current functionality via `nbody_gufunc()`

## Questions
- [x] Docstring styleguide: For the new functions, should the docstrings be in NumPy format or should they look more like the docstring in `nbody_gufunc()` so that sphinxman can scrape them for the manual?

## Checklist
- [x] Tests added for any new features (existing tests sufficient)

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -110,129 +110,12 @@ def _print_nbody_energy(energy_body_dict, header):
             previous_e = energy_body_dict[n]
         core.print_out("\n")
 
-def nbody_gufunc(func, method_string, **kwargs):
+def build_nbody_compute_list(bsse_type_list, max_nbody, max_frag):
     """
-    Computes the nbody interaction energy, gradient, or Hessian depending on input.
-    This is a generalized univeral function for computing interaction quantities.
-
-    :returns: *return type of func* |w--w| The interaction data.
-
-    :returns: (*float*, :py:class:`~psi4.core.Wavefunction`) |w--w| interaction data and wavefunction with energy/gradient/hessian set appropriately when **return_wfn** specified.
-
-    :type func: function
-    :param func: ``energy`` || etc.
-
-        Python function that accepts method_string and a molecule. Returns a
-        energy, gradient, or Hessian as requested.
-
-    :type method_string: string
-    :param method_string: ``'scf'`` || ``'mp2'`` || ``'ci5'`` || etc.
-
-        First argument, lowercase and usually unlabeled. Indicates the computational
-        method to be passed to func.
-
-    :type molecule: :ref:`molecule <op_py_molecule>`
-    :param molecule: ``h2o`` || etc.
-
-        The target molecule, if not the last molecule defined.
-
-    :type return_wfn: :ref:`boolean <op_py_boolean>`
-    :param return_wfn: ``'on'`` || |dl| ``'off'`` |dr|
-
-        Indicate to additionally return the :py:class:`~psi4.core.Wavefunction`
-        calculation result as the second element of a tuple.
-
-    :type bsse_type: string or list
-    :param bsse_type: ``'cp'`` || ``['nocp', 'vmfc']`` || |dl| ``None`` |dr| || etc.
-
-        Type of BSSE correction to compute: CP, NoCP, or VMFC. The first in this
-        list is returned by this function. By default, this function is not called.
-
-    :type max_nbody: int
-    :param max_nbody: ``3`` || etc.
-
-        Maximum n-body to compute, cannot exceed the number of fragments in the moleucle.
-
-    :type ptype: string
-    :param ptype: ``'energy'`` || ``'gradient'`` || ``'hessian'``
-
-        Type of the procedure passed in.
-
-    :type return_total_data: :ref:`boolean <op_py_boolean>`
-    :param return_total_data: ``'on'`` || |dl| ``'off'`` |dr|
-
-        If True returns the total data (energy/gradient/etc) of the system,
-        otherwise returns interaction data.
     """
-
-    ### ==> Parse some kwargs <==
-    kwargs = p4util.kwargs_lower(kwargs)
-    return_wfn = kwargs.pop('return_wfn', False)
-    ptype = kwargs.pop('ptype', None)
-    return_total_data = kwargs.pop('return_total_data', False)
-    molecule = kwargs.pop('molecule', core.get_active_molecule())
-    molecule.update_geometry()
-    core.clean_variables()
-
-    if ptype not in ['energy', 'gradient', 'hessian']:
-        raise ValidationError("""N-Body driver: The ptype '%s' is not regonized.""" % ptype)
-
-    # Figure out BSSE types
-    do_cp = False
-    do_nocp = False
-    do_vmfc = False
-    return_method = False
-
-    # Must be passed bsse_type
-    bsse_type_list = kwargs.pop('bsse_type')
-    if bsse_type_list is None:
-        raise ValidationError("N-Body GUFunc: Must pass a bsse_type")
-    if not isinstance(bsse_type_list, list):
-        bsse_type_list = [bsse_type_list]
-
-    for num, btype in enumerate(bsse_type_list):
-        if btype.lower() == 'cp':
-            do_cp = True
-            if (num == 0): return_method = 'cp'
-        elif btype.lower() == 'nocp':
-            do_nocp = True
-            if (num == 0): return_method = 'nocp'
-        elif btype.lower() == 'vmfc':
-            do_vmfc = True
-            if (num == 0): return_method = 'vmfc'
-        else:
-            raise ValidationError("N-Body GUFunc: bsse_type '%s' is not recognized" % btype.lower())
-
-    max_nbody = kwargs.get('max_nbody', -1)
-    max_frag = molecule.nfragments()
-    if max_nbody == -1:
-        max_nbody = molecule.nfragments()
-    else:
-        max_nbody = min(max_nbody, max_frag)
-
     # What levels do we need?
     nbody_range = range(1, max_nbody + 1)
     fragment_range = range(1, max_frag + 1)
-
-    # Flip this off for now, needs more testing
-    # If we are doing CP lets save them integrals
-    #if 'cp' in bsse_type_list and (len(bsse_type_list) == 1):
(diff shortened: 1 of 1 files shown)


PeterKraus:
Just out of interest, does this fix the `energy(cbs, ... , bsse_type=[...])` calls? The "method/basis" syntax for CBS works fine, but "explicit" returns the total energy. I'm not sure what's the expected behaviour, but I think it should be consistent:

```
molecule ne2 {
    0 1
    Ne 0 0 0
    --
    0 1
    Ne 0 0 2
}

set {
    scf_type      df
    mp2_type      df
    cc_type       df
    e_convergence 7
    reference     rhf
}

E_method_basis = energy("MP2/cc-pv[tq]z + D:CCSD(T)/cc-pv[dt]z", bsse_type="nocp")
E_cbs          = energy(cbs, bsse_type="nocp", corl_wfn="mp2", corl_basis="cc-pv[tq]z",
                                               delta_wfn="ccsd(t)", delta_basis="cc-pv[dt]z")

compare_values(E_method_basis, E_cbs, 6, "Interaction energy")   #TEST                                              
```
returns:
```
TestComparisonError: 	Interaction energy: computed value (445.7515388) does not match (0.0131014) to 6 digits.
```

dsirianni:
@PeterKraus Not currently -- I arrive at exactly the same `TestComparisonError` that you have provided.  It appears that `bsse_type` isn't a recognized kwarg of the CBS driver, so `energy(cbs, bsse_type='cp', ...)` still just returns a total energy.  In the spirit of smaller PR's, I'd suggest that you make it into a new issue, which I'm happy to work on :)

psi4/driver/driver_nbody.py
```diff
@@ -225,35 +206,116 @@ def nbody_gufunc(func, method_string, **kwargs):
...
 
+    # Get compute list
+    compute_dict = build_nbody_compute_list(bsse_type_list, max_nbody, max_frag)
+
+    # Build size and slices dictionaries
+    fragment_size_dict = {frag: molecule.extract_subsets(frag).natom() for
+                                           frag in range(1, max_frag+1)}
+
+    # Compute N-Body components
+    e_dict, ptype_dict, int_dict = compute_nbody_components(func, method_string, molecule, compute_dict['all'], **kwargs)
```
dgasmith: These function calls are quite complex. They either need to be simplified by duplicating small amounts of computation (broken out into functions) or a more complex data structure should be returned. Perhaps something like:

```
>>> build_nbody_compute_list(bsse_type_list, max_nbody, max_frag)
{
    "max_nbody": ...,
    "bsse_type_list": ...,
    ...
}

```

dsirianni: To an extent, I believe that I had done that already for `build_nbody_compute_list()`; are you requesting similar data-structuring for the returned quantities of `compute_nbody_components()` and `assemble_nbody_components()`?  Should the arguments be `dict`-ified as well, for clarity?

dgasmith: Effectively yes, but it must start with `build_nbody_compute_list` as this is the root that all methods will derive from. The `assemble_nbody_components` call is unwieldy enough that it likely is not very usable. I would suggest something like:

```
compute_metadata = build_nbody_compute_list(...)
component_results = compute_nbody_components(func, method_string, molecule, compute_metadata)
result = assemble_nbody_components(compute_metadata, component_results)
```

You may also want to consider making the last call simpler if `component_results` owns `compute_metadata`.

dsirianni: Structure passing added.  For ease of driving w/in the `nbody_gufunc()` (so some things don't need recomputing) I've passed & returned the `metadata` dict itself between functions.  To call these functions independently, however, this isn't actually necessary -- so, I've documented in each function's docstring exactly _which_ `key: value` pairs in the `metadata` dictionary are required and/or added.  

I have no idea if Sphinx will scrape this properly -- but at @loriab's behest I've added an [item](https://github.com/orgs/psi4/projects/1#card-9213776) to the Documentation day topic list regarding standardization of the doc style for this structure passing.

psi4/driver/driver_nbody.py
```diff
@@ -225,65 +211,188 @@ def nbody_gufunc(func, method_string, **kwargs):
...
-        bsse_str =  str(bsse_type_list)
+    bsse_str = metadata['bsse_type_list'][0]
+    if len(metadata['bsse_type_list']) > 1:
+        bsse_str =  str(metadata['bsse_type_list'])
     core.print_out("\n\n")
     core.print_out("   ===> N-Body Interaction Abacus <===\n")
     core.print_out("        BSSE Treatment:                     %s\n" % bsse_str)
 
+    # Build size and slices dictionaries
+    metadata['fragment_size_dict'] = {frag: metadata['molecule'].extract_subsets(frag).natom() for
```
dgasmith: It doesn't look like this is used in `build_nbody_compute_list` so its a good candidate to be constructed inside `component_results` so that it has a signature of:
```
component_results = compute_nbody_components(func, method_string, mol, metadata)
```

psi4/driver/driver_nbody.py
```diff
@@ -225,65 +211,188 @@ def nbody_gufunc(func, method_string, **kwargs):
...
+            ``'nocp'``: dict of int: set
+                Set containing list of computations required for non-CP procedure
+            ``'vmfc_compute'``: dict of int: set
+                Set containing list of computations required for VMFC procedure
+            ``'vmfc_levels'``: dict of int: set
+                Set containing list of levels required for VMFC procedure
+    """
+    # What levels do we need?
+    metadata['nbody_range'] = range(1, metadata['max_nbody'] + 1)
+    metadata['fragment_range'] = range(1, metadata['max_frag'] + 1)
```
dgasmith: As these are one-liners that can be easily regenerated these are good cases to simply build in the functions and not put in the metadata to simplify the interface somewhat.

psi4/driver/driver_nbody.py
```diff
@@ -225,65 +211,188 @@ def nbody_gufunc(func, method_string, **kwargs):
...
+
+    cp_compute_list = {x:set() for x in metadata['nbody_range']}
+    nocp_compute_list = {x:set() for x in metadata['nbody_range']}
+    vmfc_compute_list = {x:set() for x in metadata['nbody_range']}
+    vmfc_level_list = {x:set() for x in metadata['nbody_range']} # Need to sum something slightly different
 
     # Build up compute sets
-    if do_cp:
+    if 'cp' in metadata['bsse_type_list']:
```
dgasmith: It would be good to verify the `bsse_type_list` here again. Fun with sets: `if len({'cp', 'nocp', 'vmfc'} - set(metadata['bsse_type_list'])): raise Exception()`.

psi4/driver/driver_nbody.py
dgasmith: It could potentially be. I mostly just signaled this out as you are specifically inserting extra requirements into `metadata` not covered by a function which makes it more difficult to use. I think when you compute QM results a molecule should be associated with the data no matter what you do so when you sum up the results you know exactly what molecule you are dealing with. Happy to discuss this more, but I have no real thoughts on which path to take to solve this- up to you.

dsirianni: Makes sense.  I'm all in favor of removing complication in the dict contents expected to be passed -- but I agree, passing the molecule around makes it so much more clear exactly what is going on.  This might actually be worth thinking about in the context of passing around a QC JSON object instead of custom `dict`'s, too.

psi4/driver/driver_nbody.py
```diff
@@ -225,35 +211,126 @@ def nbody_gufunc(func, method_string, **kwargs):
...
+    # What levels do we need?
+    nbody_range = range(1, metadata['max_nbody'] + 1)
+    fragment_range = range(1, metadata['max_frag'] + 1)
 
     cp_compute_list = {x:set() for x in nbody_range}
     nocp_compute_list = {x:set() for x in nbody_range}
     vmfc_compute_list = {x:set() for x in nbody_range}
     vmfc_level_list = {x:set() for x in nbody_range} # Need to sum something slightly different
 
+    # Verify proper passing of bsse_type_list
```
dsirianni: Is this more what you had in mind for the `bsse_type_list` verification, @dgasmith? Also, is this `ValidationError` sufficiently verbose?

dsirianni: The set difference returns a new set without the items contained in both sets.  

So, if `set(metadata['bsse_type_list'])` contains none of the accepted BSSE types, `{'cp', 'nocp', 'vmfc'} - set(metadata['bsse_type_list'])` will not remove any items -- leaving it with length 3.  Thus, any length less than three should be acceptable, as one of the accepted BSSE types is contained in the list.  

Commuting the subtraction (what you've just commented) can provide the same check -- except, it also allows for any non-recognized BSSE types to be captured for raising, which does seem like a preferable solution.

dsirianni: This should do the trick:
```
bsse_type_remainder = set(metadata['bsse_type_list']) - {'cp', 'nocp', 'vmfc'}
if bsse_type_remainder:
    raise ValidationError("""Unrecognized BSSE type(s): %s
Possible values are 'cp', 'nocp', and 'vmfc'.""" % ', '.join(str(i) for i in bsse_type_remainder))
```
Is this error message sufficient?

PR #978, opened 2018-04-24
Handle ECP's in n_frozen_core(), fixes #976

### LINKED ISSUE #976: Correlated frozen core calculations with ECP's are wrong
The interaction between FC MP2 and ECP's leads to weird results. Sample input + basis sets attached:

[files.zip](https://github.com/psi4/psi4/files/1939111/files.zip)

The calculation does some no-CP interaction energies, as it was the easiest thing I could replicate this with.

For Kr-Ne "complex", everything seems OK. Perhaps when the shells are full...
```
Method, basis,         no-FC,     FC,          diff:
HF      aug-cc-pvdz    0.09214454 0.09214454 +0.00000000
HF      aug-cc-pvdz-pp 0.09069275 0.09069275 +0.00000000
HF      aug-cc-pvtz    0.09242370 0.09242370 +0.00000000
HF      aug-cc-pvtz-pp 0.09059479 0.09059479 +0.00000000
HF      aug-cc-pvqz    0.09207624 0.09207624 +0.00000000
HF      aug-cc-pvqz-pp 0.09035155 0.09035155 +0.00000000
MP2     aug-cc-pvdz    0.08587835 0.08625893 -0.00038058
MP2     aug-cc-pvdz-pp 0.08432522 0.08437474 -0.00004951
MP2     aug-cc-pvtz    0.08048658 0.08222003 -0.00173345
MP2     aug-cc-pvtz-pp 0.07943063 0.07988880 -0.00045817
MP2     aug-cc-pvqz    0.07956698 0.08046580 -0.00089882
MP2     aug-cc-pvqz-pp 0.07743521 0.07758772 -0.00015251
```
For Kr-OCS complex, HF results are fine, but [FC,no-FC]-MP2 results with triple- and quadruple-zeta start deviating a lot when -pp basis set is used. Notice the sign shift. The no-FC calculations are comparable to the non-ECP basis sets, the FC deviate significantly.
```
Method, basis,         no-FC,       FC,          diff:
HF      aug-cc-pvdz     0.00081347  0.00081347 -0.00000000
HF      aug-cc-pvdz-pp  0.00086360  0.00086360 +0.00000000
HF      aug-cc-pvtz     0.00091359  0.00091359 +0.00000000
HF      aug-cc-pvtz-pp  0.00091948  0.00091948 +0.00000000
HF      aug-cc-pvqz     0.00094070  0.00094070 +0.00000000
HF      aug-cc-pvqz-pp  0.00094875  0.00094875 +0.00000000
MP2     aug-cc-pvdz    -0.00197150 -0.00172722 -0.00024428
MP2     aug-cc-pvdz-pp -0.00204477 -0.00184466 -0.00020012
MP2     aug-cc-pvtz    -0.00293888 -0.00211075 -0.00082813
MP2     aug-cc-pvtz-pp -0.00264185  0.00219334 -0.00483518
MP2     aug-cc-pvqz    -0.00247647 -0.00211852 -0.00035795
MP2     aug-cc-pvqz-pp -0.00262450  0.01283294 -0.01545744
```

PeterKraus:
## Description
Freeze additional electron pairs on atoms with ECPs. This stops erratic behaviour in MP2 correlation energies as described in #976 .

## Todos
* **Developer Interest**
  - [x] `BasisSet::n_frozen_core()` now doesn't skip over ECP-containing atoms.

## Checklist
- [x] dfmp2-ecp modified to reflect new default behaviour

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -188,8 +188,6 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
         // will still have 3d electrons active.  Alkali earth atoms will
         // have one valence electron in this scheme.
         for (int A = 0; A < mymol->natom(); A++) {
-            // If this center as an ECP present, move along.
-            if (n_ecp_core(mymol->label(A))) continue;
             double Z = mymol->Z(A);
             if (Z > 2) nfzc += 1;
             if (Z > 10) nfzc += 4;
@@ -200,6 +198,9 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
             if (Z > 108) {
                 throw PSIEXCEPTION("Invalid atomic number");
             }
+            // If this center has an ECP, some pairs are already frozen
+            double ECP = n_ecp_core(mymol->label(A));
+            if (ECP > 0) nfzc -= ECP/2;
         }
         return nfzc;
     } else {
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -189,7 +189,7 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
         // have one valence electron in this scheme.
         for (int A = 0; A < mymol->natom(); A++) {
             double Z   = mymol->Z(A);
-            // Add ECPs to the number of electrons
+            // Add ECPs to Z, the number of electrons less ECP-treated electrons
             double ECP = n_ecp_core(mymol->label(A));
             if (Z + ECP > 2) nfzc += 1;
             if (Z + ECP > 10) nfzc += 4;
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -188,18 +188,19 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
         // will still have 3d electrons active.  Alkali earth atoms will
         // have one valence electron in this scheme.
         for (int A = 0; A < mymol->natom(); A++) {
-            double Z = mymol->Z(A);
-            if (Z > 2) nfzc += 1;
-            if (Z > 10) nfzc += 4;
-            if (Z > 18) nfzc += 4;
-            if (Z > 36) nfzc += 9;
-            if (Z > 54) nfzc += 9;
-            if (Z > 86) nfzc += 16;
-            if (Z > 108) {
+            double Z   = mymol->Z(A);
+            // Add ECPs to the number of electrons
+            double ECP = n_ecp_core(mymol->label(A));
+            if (Z + ECP > 2) nfzc += 1;
+            if (Z + ECP > 10) nfzc += 4;
+            if (Z + ECP > 18) nfzc += 4;
+            if (Z + ECP > 36) nfzc += 9;
+            if (Z + ECP > 54) nfzc += 9;
+            if (Z + ECP > 86) nfzc += 16;
+            if (Z + ECP > 108) {
                 throw PSIEXCEPTION("Invalid atomic number");
             }
             // If this center has an ECP, some pairs are already frozen
-            double ECP = n_ecp_core(mymol->label(A));
             if (ECP > 0) nfzc -= ECP/2;
         }
         return nfzc;
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1336,7 +1336,8 @@ void export_mints(py::module& m) {
         .def("shell", center_version(&BasisSet::shell), py::return_value_policy::copy,
              "Return the si'th Gaussian shell on center", py::arg("center"), py::arg("si"))
         .def("n_frozen_core", &BasisSet::n_frozen_core,
-             "Returns the number of frozen core electrons, accounting for the presence of any ECPs.")
+             "Returns the number of orbital (non-ECP) frozen core electrons. For a given molecule and "
+             "|globals__freeze_core|, `(n_ecp_core()/2 + n_frozen_core()) = constant`.")
         .def("n_ecp_core", ncore_no_args(&BasisSet::n_ecp_core),
              "Returns the total number of core electrons associated with all ECPs in this basis.")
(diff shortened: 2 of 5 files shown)


PeterKraus:
```
The following tests FAILED:
	132 - dfmp2-ecp (Failed)
	368 - dft-bench-interaction (Timeout)
	370 - pywrap-align (Failed)
	371 - pywrap-align-chiral (Failed)
```
The 132 is not surprising, as with this patch up to `[Kr]` is frozen for Xe, without it only up to `[Ar] + 3d` from the ECP. I guess this is a question of what we want as default behaviour for `freeze_core` - always freeze up to the previous Rg (as done here and in other codes), or ignore further freezing for ECP's (as is done in `master`).

loriab:
Closest we have to a C++ style is http://psicode.org/psi4manual/master/style_c.html plus the `.clang-format`, which is slowly being run over the various modules if they substantial upheaval.

PeterKraus:
@loriab yeah the test needs updating: question is with what? Do we want to keep checking against just the ECP's frozen, or the full [Kr] frozen, or perhaps both?

My suggestion would be both: do a full-core Ne-Xe, then `frozen_core = True` Ne-Xe (with [He] and [Kr] frozen respectively), and then `num_frozen_docc = 1` which tests Ne ([He] frozen) - Xe (only ECP frozen) which is the current test.

loriab:
The "both" test sounds good to me.

PeterKraus:
More ECP woes here: http://forum.psicode.org/t/frozen-core-for-cations/891

Could be tacked onto this PR or made into a new one...

loriab:
How timely. Unless he's really quick with a test case, I think sep PR. It definitely qualifies as bugfix.

loriab:
Any reservations to final review and merge, @PeterKraus?

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -188,18 +188,20 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
...
-            if (Z > 2) nfzc += 1;
-            if (Z > 10) nfzc += 4;
-            if (Z > 18) nfzc += 4;
-            if (Z > 36) nfzc += 9;
-            if (Z > 54) nfzc += 9;
-            if (Z > 86) nfzc += 16;
-            if (Z > 108) {
+            double Z   = mymol->Z(A);
+            // Add ECPs to Z, the number of electrons less ECP-treated electrons
+            double ECP = n_ecp_core(mymol->label(A));
```
amjames: So when using ECPs the molecule's function `Z(atom_index)` returns a differnt value? 

That is not something that concerns this PR, but I just find that concerning. Should the molecule care about ECPs? Is that not something for the basis set to care about exclusively?

loriab: Yup, `Molecule->Z(A)` returns something that is neither the atomic number, nor `0.0` for ghost. That's why my initial pencil-and-paper check of the PR went wrong. I agree disconcerting and worth revisiting. Wish-list maybe?

loriab: FYI, @bgpeyton, a further variant on `Molecule.Z()`'s behavior.

dgasmith:
As a note we are going to wait on this until @andysim gets off a mountain just to triple check.

psi4/src/psi4/libmints/basisset.cc
PeterKraus: It's probably also the reason why sometimes when using eg. Xe with an ECP, psi4 talks about Fe and sometimes about Xe. I believe the offending part is in Optking:
```
        Writing optimization data to binary file.
        Structure for next step:
        Cartesian Geometry (in Angstrom)
           FE    -1.2297359984  -0.0385946949   0.0000000000
            C     2.5758026217   0.5881468581   0.0000000000
            O     2.2878670730   1.7087928078   0.0000000000
            S     2.9767793599  -0.9094270890   0.0000000000
                        --------------------------
                         OPTKING Finished Execution 
                        --------------------------

    Structure for next step:
    Molecular point group: cs
    Full point group: Cs

    Geometry (in Angstrom), charge = 0, multiplicity = 1:

    XE          -1.232187566437    -0.039755314177     0.000000000000
    C            2.573351053672     0.586986238783     0.000000000000
    O            2.285415504952     1.707632188472     0.000000000000
    S            2.974327791842    -0.910587708272     0.000000000000
```

PeterKraus:
I believe this PR fixes the bug in #976 correctly, without adverse side effects (famous last words). Does the overall charge (from the charge/multiplicity entries) affect `Z` values of the component atoms? What happens when one would - perhaps foolishly - try to do a "frozen core" `Li^2+` calculation?

However, I am wondering whether our approach to core freezing is perhaps a bit simplistic: a point was raised on the forums that one might want to be able to automatically detect when the valence shell is empty (eg. alkali metal cations), and then perhaps unfreeze the previous shell. Similarly, some other QM packages allow freezing up to `N-th` previous rare gas shell (the current behaviour is `N=1`).

psi4/src/psi4/libmints/basisset.cc
loriab: That's funny — we have the only transmuting geometry optimizer. Fortunately, optking's getting replaced this summer.

loriab:
Neither overall chg/mult nor fragment chg/mult will effect Mol.Z(at); all that info is stored separately. BasisSets are associated with each atom in Mol, so it does make some, though not a lot of, sense that ECPs should affect Z(at).

Once upon a time there were true/false/large/small choices for `freeze_core`. They got dropped. I don't remember why, but I'd venture b/c they caused more confusion than convenience. But a lot more was handled c-side then. I expect a more flexible interface (that also takes frozen_docc into account) would have support.

psi4/src/psi4/libmints/basisset.cc
andysim: Sorry for my tardiness - I've been traveling nonstop for a couple of weeks.  At the time I threw the ECP stuff together my thinking was that any center with an ECP should have no valence electrons frozen at all.  I think that's what this line is supposed to achieve.  Your fix is much easier to read and allows for some valence electrons to be frozen in addition to the ECP (something that I didn't account for in my haste).  LGTM - thanks for catching this.

PR #979, opened 2018-04-24
QC JSON Schema

dgasmith:
## Description
Adds the ability to consume and produce MolSSI QC JSON Schema.

## Todos
* **Developer Interest**
  - [x] A prototype MolSSI QC Schema has been implemented (http://molssi-qc-schema.readthedocs.io/en/latest/index.html).

## Questions
- [x] How much validation do we want?
- [x] A full test suite should like come from the schema itself?
- [x] How we will we handle extra psi vars? Is the current setup sufficient.

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,132 @@
...
+                 ("DISPERSION CORRECTION ENERGY", "scf_dispersion_correction_energy")]
+    for pkey, skey in other_scf:
+        if (pkey in psi_props) and (psi_props[pkey] != 0):
+            props[skey] = psi_props[pkey]
+
+    # Write out MP2 keywords
+    if "MP2 CORRELATION ENERGY" in psi_props:
+        props["mp2_same_spin_correlation_energy"] = psi_props["MP2 SAME-SPIN CORRELATION ENERGY"]
+        props["mp2_opposite_spin_correlation_energy"] = psi_props["MP2 OPPOSITE-SPIN CORRELATION ENERGY"]
+        props["mp2_singles_energy"] = 0.0
```
loriab: If you care about singles energy, I have some reform (https://github.com/loriab/psi4/commit/263c102469c1ed065badfa3069ea3d2f60c1a8c8). If not, it can wait.

psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,132 @@
...
+
+    if return_output:
+        with open(outfile, 'r') as f:
+            json_data["raw_output"] = f.read()
+        os.unlink(outfile)
+
+    return json_data
+
+
+def run_json_qc_schema(json_data):
```
loriab: Am I overlooking keywords/options setting in qc_schema as opposed to original_v1_1?

psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,132 @@
...
+
+    """
+
+    # This is currently a forced override
+    json_data["schema_version"] = 0
+
+    json_data["provenance"] = {"creator": "QM Program", "version": psi4.__version__, "routine": "psi4.json.run_json"}
+
+    # QCDB, help Lori!
+    mol = core.Molecule.from_arrays(geom=json_data["molecule"]["geometry"], elem=json_data["molecule"]["symbols"], units="Bohr")
```
loriab: on it.

psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,133 @@
...
+    !Warning! This function is experimental and likely to change in the future.
+    Please report any suggestions or uses of this function on github.com/MolSSI/QC_JSON_Schema.
+
+    Examples
+    --------
+
+    """
+
+    # This is currently a forced override
+    json_data["schema_version"] = "v0.1"
```
loriab: Note that I had to change this from `0` to correctly ID the schema.

psi4/driver/json_wrapper.py
loriab: Code's in, but it's untested except insofar as your json tests still work.

psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,132 @@
...
+    method = json_data["model"]["method"] + "/" + json_data["model"]["basis"]
+    val, wfn = methods_dict[json_data["driver"]](method, return_wfn=True, molecule=mol)
+
+
+    if json_data["driver"] == "energy":
+        json_data["return_result"] = val
+    elif json_data["driver"] in ["gradient", "hessian"]:
+        json_data["return_result"] = val.np.ravel().tolist()
+    else:
+        raise KeyError("Did not understand key %s")
```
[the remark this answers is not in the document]
dgasmith: Depends in how much you believe in the magic.

psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,132 @@
...
+
+    # Pull out a standard set of SCF properties
+    psi_props = psi4.core.get_variables()
+    json_data["psi4:properties"] = psi_props
```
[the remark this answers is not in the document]
dgasmith: Sure, it was an open question above.

psi4/driver/json_wrapper.py
```diff
@@ -50,9 +49,132 @@
...
+        "scf_dipole_moment": [psi_props[x] for x in ["SCF DIPOLE X", "SCF DIPOLE Y", "SCF DIPOLE Z"]],
+        "scf_iterations": int(psi_props["SCF ITERATIONS"]),
+        "scf_total_energy": psi_props["SCF TOTAL ENERGY"],
+    }
+
+    # Pull out optional SCF keywords
+    other_scf = [("DFT VV10 ENERGY", "scf_vv10_energy"), ("DFT XC ENERGY", "scf_xc_energy"),
+                 ("DISPERSION CORRECTION ENERGY", "scf_dispersion_correction_energy")]
+    for pkey, skey in other_scf:
+        if (pkey in psi_props) and (psi_props[pkey] != 0):
```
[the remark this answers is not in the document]
dgasmith: Sure, whats the syntax?

psi4/driver/json_wrapper.py
dgasmith: Eh, I think it can wait for now.

psi4/driver/json_wrapper.py
dgasmith: The `keywords` field is currently a free for all according to the spec?

psi4/driver/json_wrapper.py
dgasmith: Hmm, I thought we had discussed going to an integer based scheme so that parsers were not required to parse versioning semantics?

psi4/driver/json_wrapper.py
loriab: But if present, we need to set them, yes? I'll add something, then see if we're at cross-purposes.

psi4/driver/json_wrapper.py
loriab: Ah yes, so we did. In that case, I should have altered the test cases. Will switch out.

psi4/driver/json_wrapper.py
```diff
@@ -32,27 +32,202 @@
...
+        props["mp2_total_energy"] = psi_props["MP2 TOTAL ENERGY"]
+
+    json_data["properties"] = props
+    json_data["success"] = True
+    print(json.dumps(json_data, indent=2))
+
+    # Reset state
+    optstash.restore()
+    if clean:
+        core.clean()
```
loriab: do you want either of `psi4.core.clean_options()` or `psi4.core.clean_variables()` here, too?

psi4/driver/json_wrapper.py
```diff
@@ -169,70 +345,47 @@ def run_json(json_data):
...
-        # Deep copy kwargs
-        if "kwargs" in list(json_data):
-            kwargs = copy.deepcopy(json_data["kwargs"])
-        else:
-            kwargs = {}
+    if json_data["driver"] not in methods_dict_:
+        raise KeyError("Driver type '%s' not recognized." % str(json_data["driver"]))
 
-        kwargs["molecule"] = molutil.geometry(json_data["molecule"])
+    val, wfn = methods_dict_[json_data["driver"]](*args, **kwargs)
```
loriab: I'll need to define an extension for add'l kwargs, yes?

loriab: Yeah, add'l kwargs and the ability to record whole output record goes down dramatically. But for the getting-things-running aspect of the separated driver, I'll need it. record aspect comes later.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1611,6 +1611,13 @@ void OEProp::compute_mulliken_charges()
     outfile->Printf( "\n   Total alpha = %8.5f, Total beta = %8.5f, Total charge = %8.5f\n", \
         suma, sumb, nuc - suma - sumb);
     wfn_->set_atomic_point_charges(apcs);
+
+    auto vec_apcs = std::make_shared<Matrix>("Mulliken Charges: (a.u.)", 1, mol->natom());
+    for (size_t i = 0; i < mol->natom(); i++){
+        vec_apcs->set(0, i, (*apcs)[i]);
+    }
+    wfn_->set_array("MULLIKEN_CHARGES", vec_apcs);
+
```
amjames: Since we are starting to use wfn arrays more heavily. It may be a good idea to provide overloads for `set_array` that accept `std::vector<double>`, `psi::Vector` and `*double` + `size`, and do the conversion to a 1 row matrix automatically. Just to that is does not need to be done over and over again.

dgasmith: I would argue that we really need a N-dimensional tensor class before allowing folks to arbitrary overload `set_arrays` like that. Happy to take PR's on this however.

psi4/driver/json_wrapper.py
```diff
@@ -32,27 +32,206 @@
...
+        if "quadrupole" in kwargs["properties"]:
+            ret["quadrupole"] = [psi_props[mtd + " QUADRUPOLE " + x] for x in ["XX", "XY", "XZ", "YY", "YZ", "ZZ"]]
+        if "mulliken_charges" in kwargs["properties"]:
+            ret["mulliken_charges"] = wfn.get_array("MULLIKEN_CHARGES").np.ravel().tolist()
+        if "lowdin_charges" in kwargs["properties"]:
+            ret["lowdin_charges"] = wfn.get_array("LOWDIN_CHARGES").np.ravel().tolist()
+        if "wiberg_lowdin_indices" in kwargs["properties"]:
+            ret["wiberg_lowdin_indices"] = wfn.get_array("WIBERG_LOWDIN_INDICES").np.ravel().tolist()
+        if "mayer_indices" in kwargs["properties"]:
+            ret["mayer_indices"] = wfn.get_array("MAYER_INDICES").np.ravel().tolist()
```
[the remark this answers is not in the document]
dgasmith: Small misconception: The JSON parser is here to match the MolSSI QC JSON Schema which may not match Psi4's expected output. For `properties` we are in a bit uncharted waters here and are out to provide something that solves industry requests now rather than wait for the Schema to full catch up.

The two issues you bought up seems like a module consistency issue rather than a schema issue. If I recall correctly I patch this up to work everywhere as expected except for the cc* star modules which seemed to be much more involved.

NO Occupations stores a `std::vector<std::vector<std::tuple<double, int, int> >>` which I am not yet sure how to handle. I mostly changed the charges as the data is overridden if you request both types of charges which leads to poor behavior in the output. We can add a change for the NO Occupations if you wish, but it is much more likely they will change in the future.

psi4/driver/json_wrapper.py
dgasmith: Main difference here is energies are energies and will always be produced. For properties, what happens if someone asks for `energy('MP2')` and then several properties? You will get SCF properties which can lead to significant confusion. The `properties` fields are explicitly labeled for each level of theory while the `return_result` is simple what was requested.

psi4/driver/json_wrapper.py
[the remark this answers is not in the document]
loriab: Even if downstream QC_JSON wants properties divorced from method (that is, store DIPOLE_MOMENT unadorned and assumed to be attached to the method MP2), think it'd help psi-side in compensating for wave-like routes through the code to label properties as precisely as possible at the earliest point of calculation, usually c-side?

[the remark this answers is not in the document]
dgasmith: Only the explicit return field is unadorned without further annotation as this is the result that was requested. If we ever want to store all properties computed at multiple levels of theory, this will likely have to be fit in somewhere in the properties or a new field. It would be great to label everything, I just worry about the onus it puts on us and downstream users.

amjames:
My thoughts on the questions:
> How much validation do we want?

I would think that input would need to be validated always as the source is unknown, and output should be validated as part of the test suite, but then does not need to be validated at run-time before they are returned since (ideally) the test-suite ensures that our output jsons always adhere to the schema. Is there a best practice followed by applications using JSON APIs? For example does the JSON "consumer" always hold responsibility for checking that the object received is valid, or is it up to the JSON "producer" to ensure that objects are valid before sending them? 

>A full test suite should like come from the schema itself?

Not sure what you mean here. Should the schema project provide functionality to validate objects that can be used by applications adhering to the schema to test that their JSON generating functions are generating valid objects? That would be great, and would reduce the amount of work that needs to be done by devs of individual QC packages to adopt the interface.

> How we will we handle extra psi vars? Is the current setup sufficient.
If the schema allows for extra/arbitrary fields somewhere,  we could provide all psivars/arrays there. If not then I guess we can just add handling of additional psivars in the future as they apply to new revisions of the schema?

psi4/driver/json_wrapper.py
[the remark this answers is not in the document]
dgasmith: If you could patch up the cc* that would be fantastic and get everyone on a consistent page.

dgasmith:
Ultimately we want to run the JSON through a `jsonschema` validator based on the QC schema spec. However, we are not quite there yet as we would need quite a few extensions. Mostly the question at the moment is how much validation do we want at this moment.

For the QC schema test suite: We have about a dozen examples currently of input/ouput. I would imagine that this will expand quiet considerably over time and ultimately represent a test suite in of itself. The QC schema project will produce a `jsonschema` specification of input and output which can validate structure in most languages, see [here](http://json-schema.org/implementations.html). The sample could be turned into a test suite to validate the information contained.

A general comment about schema's. They unfortunately have no hope of validating content, just structure.

psi4/driver/json_wrapper.py
```diff
@@ -32,27 +32,206 @@
...
+    Examples
+    --------
+
+    """
+
+    # Clean a few things
+    _clean_psi_environ(clean)
+
+    # This is currently a forced override
+    json_data["schema_version"] = "0"
```
loriab: @dgasmith, did you have `v0` or `0` in mind for schema_version?

dgasmith: `0` as an int I think. If we are going to make it easy to compare versions we should make it really easy, some ideas from [here](https://en.wikipedia.org/wiki/Magic_number_(programming)).

loriab: ok, just fixed it up here. RTG now, imo.

dgasmith:
@andysim We may have run out of people able to review this. Going to pull it in as this is this is a versioned WIP. We should be minting the "version 1" Schema soon so can update this in a week or so under the guise of a bug fix.

PR #981, opened 2018-04-25
Cghf psiapi

edeprince3:
## Description
Adds a few classes to the PSI_API list so one of my plugins will run.

## Todos
* **Developer Interest**
  - [x] plugins can access DFJK, PKJK, VBase, and BasisFunctions classes.

## Checklist
- [x] passes smoke, quicktests subsets of tests

## Status
- [x] Ready for review
- [x] Ready for merge


edeprince3:
The only test that failed in quicktests is x2c1, and I don't believe this PR is the culprit:

TestComparisonError:    Non-relativistic CCSD energy: computed value (-76.2807318207) does not match (-76.2807318195) to 9 digits.

Issue #909, opened 2018-02-03
Ghost atom masses and molecular weight


bgpeyton:
This is an issue I came across when needing to calculate the molecular weight of a molecule (which I was surprised is not already part of the Molecule class). 

Currently, ghost atoms still return a mass when calling `mol.mass(i)`. The best way I have found to avoid this on the Python side is to check the Z value `if mol.Z():` because ghost atoms (correctly) return false here. `optrot.cc` does not do this correctly and so uses an incorrect molecular weight for molecules with ghost atoms, and there may be more codes that have this issue. I thought it may be better to fix `mass()` rather than try to find every single code which calculates a molecular weight and add in the Z value check.

I feel ghost atoms should not have any mass in the first place, as they are usually just coordinates on which to place basis functions corresponding to the atom. I can't think of any places where you would want the mass of a ghosted atom, so please correct me if I'm mistaken. 

Within `masses.h`, a note is found:
```
** 10/08/99 EFV - Changed an2masses[0] to 0.0000
**                (ghost atom's weight should be 0)
```
But in `molecule.cc` the an2masses list is only used if `atoms_[atom]->mass() != 0` returns false, which will not happen for ghost atoms as they are currently done. Instead, it should probably be checked first `if atoms_[atom]->Z() == 0` and return a zero mass. (I thought this was better than pushing the Z == 0 case into the else statement, because doing that throws a _lot_ of warnings at the user). This is an easy fix, but as this effects the molecule class I wanted to check some other opinions before opening my PR. There also may be a more fundamental place to fix this problem, such as when the `atoms_` list is generated, but I wasn't sure.

loriab:
Your analysis of inconsistency and proposed fix are totally reasonable. But I disagree and rather have the opposite complaint about the Molecule class – that Z returns 0 instead of atomic number for ghost atoms so that function is actually conflating two independent quantities. mass is doing the right thing.

The reason nonzero quantities are stored for ghost atoms in a CoordEntry is because the atoms can change their ghostedness upon `extract_subsets`, `activate_all_fragments`, etc. This leads to two tiers of ghostedness. The first are those that are read in as ghosts `@C` and get 0 Z and charge from the start. The others are those that are read in as real, then have their ghostedness changed on the fly. Note that Molecule internal fns sometimes have to check both `Molecule.fragment_types_` and `CoordEntry.ghost_` to get the full state of the Molecule.

[...]

PR #990, opened 2018-04-28
Molecule Mass and Z Update for Ghosts

bgpeyton:
## Description
Provides a remedy for [#909](https://github.com/psi4/psi4/issues/909). `Molecule::mass` and `Molecule::Z` show inconsistent behavior for ghosts. This PR requires a new argument `bool zero_ghost` on the C++ side and optionally in the Python layer which gives a choice for behavior.

## Todos
* **Developer Interest**
  - [x] Required C++ argument `bool zero_ghost` for `Molecule::mass`
  - [x] Optional python argument `bool zero_ghost` for `Molecule::mass`
  - [x] Required argument `bool zero_ghost` for `Molecule::Z`
  - [x] Optional python argument `bool zero_ghost` for `Molecule::Z`
  - [x] Optional C++ argument `bool zero_ghost` for `Molecule::center_of_mass`
  - [x] Optional C++ argument `bool zero_ghost` for `Molecule::inertia_tensor`
  - [x] Check all C++ calls for "intended behavior"
  - [x] New optional python arguments for `Molecule.mass()` and `Molecule.Z()`

## Questions
- Each file changed on the C++ side needs to be passed either `True` or `False`. I'm defaulting to previous behavior in all cases except `optrot.cc` where I know functionality should change. 
- "Previous behavior" is that `mass` returns non-zero for ghost atoms (equivalent to `zero_ghost = false`), and `Z` returns zero for ghost atoms (`zero_ghost = true`). This behavior may want to be changed in some places, so this is a good time for a logic pass-through. 

## Checklist
- [x] Add argument for `mass`
- [x] Add argument for `Z`
- [x] Tests
- [x] Docs

## Status
- [x] Ready for review
- [x] Ready for merge


bgpeyton:
Now is a good time for the logic pass-through for `mass` and `Z`. I know there's a few places where molecular weight is being calculated, and those probably don't want to have ghost atom masses counted, so I will change those to `zero_ghost = true`. Many of the other cases I'm unfamiliar with their "intended behavior" so if anyone else sees places where the behavior should change let me know. Otherwise I'm defaulting to previous behavior. @loriab

psi4/src/psi4/ccresponse/optrot.cc
```diff
@@ -406,7 +406,7 @@ void optrot(std::shared_ptr<Molecule> molecule) {
             }
 
         /* compute the specific rotation */
-        for (j = 0, M = 0.0; j < moinfo.natom; j++) M += molecule->mass(j); /* amu */
+        for (j = 0, M = 0.0; j < moinfo.natom; j++) M += molecule->mass(j,true); /* amu */
```
bgpeyton: Definitely need ghost mass to return zero here

psi4/src/psi4/ccresponse/scatter.cc
```diff
@@ -301,7 +301,7 @@ void scatter(std::shared_ptr<Molecule> molecule, Options &options, double step,
     // Mass-weighting the co-ordinates //
     //double massi[natom];
     outfile->Printf("\tAtomic Masses for Raman Computation:\n");
-    for(i=0; i < natom; i++) outfile->Printf("\t%d %12.8f\n", i, molecule->mass(i));
+    for(i=0; i < natom; i++) outfile->Printf("\t%d %12.8f\n", i, molecule->mass(i,true));
```
bgpeyton: Mass calls in `scatter.cc` are for molecular weights or mass weighting, which probably means they do not intend to include the masses of ghost atoms.

psi4/src/psi4/libmints/cdsalclist.cc
```diff
@@ -121,7 +121,7 @@ CdSalcList::CdSalcList(std::shared_ptr<Molecule> mol, int needed_irreps, bool pr
         double atomx = molecule_->x(i);
         double atomy = molecule_->y(i);
         double atomz = molecule_->z(i);
-        double smass = sqrt(molecule_->mass(i));
+        double smass = sqrt(molecule_->mass(i, false));
```
bgpeyton: I'm _pretty_ sure this should return zero mass so that COM isn't effected in a nonphysical way, but can I get a second opinion on this one?

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -359,9 +359,11 @@ void Molecule::add_unsettled_atom(double Z, std::vector<std::string> anchor, std
 
 
 
-double Molecule::mass(int atom) const {
+double Molecule::mass(int atom, bool zero_ghost) const {
     double ret = 0.0;
-    if (atoms_[atom]->mass() != 0.0)
+    if (zero_ghost && (atoms_[atom]->Z() == 0.0))
```
bgpeyton: I'm not totally happy with checking `CoordEntry::Z` to check for Type A (Gh(x)) ghosting, but `is_ghosted` only works for type B (`extract_subsets`) ghosting. So, this is the only way that gives consistent behavior for both types to my knowledge.

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -1689,9 +1691,9 @@ void Molecule::print() const {
 
         for (int i = 0; i < natom(); ++i) {
             Vector3 geom = atoms_[i]->compute();
-            outfile->Printf("    %8s%4s ", symbol(i).c_str(), Z(i) ? "" : "(Gh)");
+            outfile->Printf("    %8s%4s ", symbol(i).c_str(), Z(i, true) ? "" : "(Gh)");
             for (int j = 0; j < 3; j++) outfile->Printf("  %17.12f", geom[j]);
-            outfile->Printf("  %17.12f", mass(i));
+            outfile->Printf("  %17.12f", mass(i, true));
```
bgpeyton: Printed masses of ghost atoms should definitely be zero.

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -1940,14 +1942,14 @@ Matrix *Molecule::inertia_tensor() const {
 
     for (i = 0; i < natom(); i++) {
         // I(alpha, alpha)
-        temp(0, 0) += mass(i) * (y(i) * y(i) + z(i) * z(i));
-        temp(1, 1) += mass(i) * (x(i) * x(i) + z(i) * z(i));
-        temp(2, 2) += mass(i) * (x(i) * x(i) + y(i) * y(i));
+        temp(0, 0) += mass(i, true) * (y(i) * y(i) + z(i) * z(i));
```
bgpeyton: Ghost atoms shouldn't effect moments of inertia, so assuming all of these should return zero mass.

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -424,7 +426,7 @@ Vector3 Molecule::center_of_mass() const {
     total_m = 0.0;
 
     for (int i = 0; i < natom(); ++i) {
-        double m = mass(i);
+        double m = mass(i, false);
```
amjames: This one is probably going to be `true` most of the time, but you may want to have this function take an optional for explicit inclusion of ghost atoms.

psi4/src/psi4/libmints/molecule.cc
amjames: If Checking `CoordEntry::Z` is the only way to catch both type A and type B ghosts then its fine. Changing the behavior of `CoordEntry` is far more complicated and outside the scope of this PR

psi4/src/psi4/libmints/molecule.cc
amjames: Including the mass of ghosted atoms here would lead to incorrect rotational constants, but there may be some use case (ie when orienting if you want A-Gh(B) you want it to be consistent with A-B/Gh(A-)B counterparts).

Same thing would go for the COM calculation should there be optionals for how ghosts are treated in these functions as well?

bgpeyton: I was a little worried about these sorts of use cases. Presets could harm n-body / fragmentation schemes in general (or manual counterpoise corrections), or anyone trying to do "apples to apples" comparisons on systems with potentially different COM's, no matter which way you do them. An optional would be my first thought Py-side, not sure about here. Any other opinions?

loriab: This is changing current behavior, yes? Sounds dangerous. Except for when you're correcting existing ghost misunderstandings, I'd encode existing behavior rather than higher notions of what it should be. The optional arg you could add to inertia_tensor then allows that higher definition to be got, while the existing code stays the same.

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -2835,7 +2837,19 @@ Vector3 Molecule::fxyz(int atom) const { return input_units_to_au_ * full_atoms_
...
+        return atoms_[atom]->Z();
+    }
+    else if (atoms_[atom]->Z() == 0.0) {
+        static const double val = z_list[atoms_[atom]->symbol()];
+        return val;
+    }
+    else {
+        return atoms_[atom]->Z();
+    }
+}
```
loriab: Can't this all be
```
if (!zero_ghost && (atoms_[atom]->Z() == 0.0)) {
    Element_to_Z z_list = Element_to_Z();
    return z_list[atoms_[atom]->symbol()];
}
else {
    return atoms_[atom]->Z();
}
```

tests/python/ghost_types/input.py
```diff
@@ -0,0 +1,33 @@
...
+    Gh(He) 1 1
+""")
+
+B_frag = A_frag.extract_subsets(1,2)
+
+# mass defaults to zero_ghost = false
+assert psi4.compare_values(m_He, A_frag.mass(1), 9, 'A_frag mass default')
+assert psi4.compare_values(m_Gh, A_frag.mass(1, True), 9, 'A_frag mass True')
+assert psi4.compare_values(m_He, B_frag.mass(1), 9, 'B_frag mass default')
+assert psi4.compare_values(m_Gh, B_frag.mass(1, True), 9, 'B_frag mass True')
```
loriab: thanks for the test! may as well check `False` also and `True` for Z.

tests/python/ghost_types/input.py
```diff
@@ -0,0 +1,33 @@
...
+    He 1 1
+""")
+
+A_frag = psi4.geometry(""" 
+    He
+    --
+    Gh(He) 1 1
+""")
+
+B_frag = A_frag.extract_subsets(1,2)
```
loriab: B_frag is rather just echoing the explicit ghosting set up by A_frag. Did you mean `B_frag = frag.extract_subsets(1, 2)` for testing the other type of ghosting? And maybe throw in a NRE and total mass check for mixed extract_subsets/`@` ghosting with `zero_ghosts` T/F?

psi4/src/psi4/libmints/molecule.cc
bgpeyton: I'll go ahead and add an optional for `inertia_tensor` and `center_of_mass` defaulting to `false` (current behavior), then I'll probably do some testing to see whether current tests even care what the default is (my guess is these functions are probably not tested with ghost atoms). Either way, I like this solution.

tests/python/ghost_types/input.py
bgpeyton: Yes, absolutely meant `frag.extract_subsets`. I even had that in my test jupyter notebook, not sure how I missed it here, thanks. I'll increase the tests to look at a mixed-ghost case, and total masses. What do you mean by `NRE`?

loriab: nuclear_repulsion_energy

psi4/src/psi4/libmints/molecule.cc
bgpeyton: Much nicer! I agree on all accounts except `return z_list[atoms_[atom]->symbol()];`. This throws a warning at compile-time: 
```
/Users/mash/Gits/psi4/psi4/src/psi4/libmints/molecule.cc:2843:16: warning: returning reference to local temporary object
      [-Wreturn-stack-address]
        return z_list[atoms_[atom]->symbol()];
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
And I end up getting garbage for this return (Z = 6.95321009529437e-310 for Gh(He), type A and B).

loriab: ok, I don't really know my C++ catechism. :-)

bgpeyton: I'm not 100% sure on the reason why, but declaring the type definitely fixes the problem. Someone with far more C++ knowledge than I have may be able to explain it! ( @amjames  )

tests/python/ghost_types/input.py
bgpeyton: Done, looking much better!

bgpeyton:
As for `ghosts_join` versus `zero_ghost`, when you first used the phrase `zero_ghost` @loriab it was immediately clear what it would mean in both cases- you know exactly what you're getting when it's `true` or `false`. I feel like `ghosts_join` takes just a little bit more thought. Maybe I'm just a bit simpler-minded.

psi4/src/psi4/libmints/molecule.cc
bgpeyton: So, thanks to @amjames I've added a bit more to the test case: the molecule now includes an H and an He atom. This is because he predicted a failure when returning Z values from one atom type to another due to the `static const double val`. Basically, whichever atom got through to the z_list part first would determine all other `val`'s returned for that molecule. The new test cases catch this when checking all Z values for a mixed-element molecule when `zero_ghost = False` and both atoms are ghosted. 

This can really only be fixed by changing the return type of `Molecule::Z` itself to just `const double Molecule::Z` instead of a reference. By doing this, I can now directly `return z_list[atoms_[atom]->symbol()]`, meaning I won't have to set it as `static const` (so it can be overwritten properly).

bgpeyton:
Moved to [PR#1355](https://github.com/psi4/psi4/pull/1355); I left this for too long and the merge conflicts became insurmountable. Sorry!

PR #993, opened 2018-04-30
~hungarian~, EFP, `Mol.to_string()`, bumps

loriab:
## Description
Lots o' stuff. **I recommend reviewing commit-by-commit.**

## Todos
* **Developer Interest**
 - [x] Fix "write() argument must be str, not bytes" error that occurs in new NumPy
 - [x] QCDB: migrate Hungarian alg from hungarian module to hacked scipy. now "align" tests should pass w/o incident or add'l software
 - [x] Mol: migrate Molecule.to_string() back from QCDB. not used, but may as well be handy.
 - [x] QCDB: squash some printing. migrate some changes back from QCDB.
 - [x] AddOns: Don't let scipy missing (for snsmp2) pass silently.
   - Discovered faulty logic where snsmp2 will rebuild more than necessary b/c detection involves "import psi4" which probably doesn't exist at the start of a psi4 build. not fixed.
 - [x] AddOns: Bump pybind11 to 2.2.3
 - [x] AddOns: renegotiate the psi vs. libefp options compromise. We generally try to follow addon's naming conventions unless they directly contradict our own. For SAPT we do have our own, but when libefp was implemented, we used our sapt for two (elst & exch) and let libefp win on (pol). I've since put a translation of input and printing system into pylibefp. that isn't active here, but thought may as well change keywords now.
```
    'psi': {
        'elec': 'elst',
        'pol': 'ind',
        'xr': 'exch',
        'elec_damp': 'elst_damping',
        'pol_damp': 'ind_damping',
        'disp_damp': 'disp_damping',
        'pol_driver': 'ind_driver',
        'ai_elec': 'ai_elst',
        'ai_pol': 'ai_ind',
        'ai_xr': 'ai_exch',
```
 - [x] Add warnings accumulated so far -- ECP, DFT-SAPT, DFT
 - [x] set up plugins via native conda gxx
 - [x] AddOns: bump libefp to 1.5.0, chemps2 to 1.8.7
 - [x] returns align tests to quicktest
 - [x] adds `psi4 --test` to travis to test psiapi and addons coverage
 - [x] switches MINIX to spherical as advocated [here](https://github.com/psi4/psi4/pull/961#issuecomment-385644036)

## Questions
- [ ] I took the warn DFT functionals from tests and notes, not one unified place, so see if they're right.
- [x] Check the EFP bit is what we want.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
@CDSherrill, you might want the opportunity to weigh in on [this commit](https://github.com/psi4/psi4/pull/993/commits/609a84f64a50e5dee8728fcaa9a19cfaebffb4c2), as it involves keyword names for EFP and our compromise between SAPT-like and libefp-like. Please see top post of PR for discussion.

CDSherrill:
Cool, glad to see these updates.  Thanks for pinging me about the naming.

I don't have a strong preference for POL vs IND.  I suppose being
consistent with IND is a good thing.  Does it make sense to alias these
variables so either version works?  Might be less confusing for people
consulting the EFP docs.

I don't really like EFP_AI_XXX.  I do like pre-pending with EFP so that all
EFP keywords start with that.  But AI connotes Artificial Intelligence more
easily than ab initio, which I assume it's supposed to mean.  I think I
like QM better.  (EFP_QM_XXX).  This makes sense in a QM/MM context, which
this basically is (EFP being a very advanced MM method).

Additionally, I'm not sure the docs are not totally clear... I'm not
totally sure what this means:

/*- Do include polarization energy term in EFP computation? (QMEFP_POL c.
v1.1)  -*/

Is this the polarization of the EFP by the QM region, or the polarization
of the QM region by the EFP?

loriab:
Thanks for the comments, @CDSherrill.

* I'm definitely in favor of IND over POL to be consistent with SAPT. And we haven't been in the habit of aliasing options names within Psi. People going directly to the libefp docs are already going to be confused b/c their ELEC is our ELST and their XR is our EXCH. So at least this way POL/IND are translated into psi self-consistency.

* I wasn't sure where the AI/QM compromise should fall. Glad to switch it all the way back to `QMEFP_ELST` or compromise at `EFP_QM_ELST`.

* For `QMEFP_POL`, I think it's both directions. But for other QMEFP terms it should be more specifically EFP influencing the SCF. Will edit.

psi4/driver/endorsed_plugins.py
```diff
@@ -40,6 +40,8 @@
 
 try:
     import snsmp2
-except ImportError:
-    pass
-
+except ImportError as e:
+    if 'scipy' in e.msg:
+        raise ImportError("""Psi4 plugin 'snsmp2' available, but scipy missing. Try `conda install scipy` or `pip install scipy`.""")
```
dgasmith: Cool.

psi4/driver/procrouting/sapt/sapt_proc.py
```diff
@@ -79,6 +79,8 @@ def run_sapt_dft(name, **kwargs):
     core.print_out("         ---------------------------------------------------------\n")
     core.print_out("\n")
 
+    core.print_out("  !!!  WARNING:  SAPT(DFT) capability is in beta. Please use with caution. !!!\n\n")
```
dgasmith: Thanks!

psi4/driver/qcdb/molecule.py
```diff
@@ -1354,8 +1366,10 @@ def _raw_to_dict(self, force_c1=False, force_au=False, np_out=True):
         if self.name() not in ['', 'default']:
             molrec['name'] = self.name()
 
-        if force_au:
+        if force_units == 'Bohr':
```
dgasmith: I noticed throughout you are not case insensitive for these quantities. Just checking if this is intentional as we are usually not case sensitive for input values.

loriab: We're case insensitive for user-facing QC functions but not usually for ordinary python fn args, I thought. These are essentially the enums from libmints.Mol so I didn't really consider it. I'll consider bohr/angSTROM over in qcdb, but I'd really rather not get into yes/1/on- and Bohr/au/a.u./bohr-style flexibility.

PR #995, opened 2018-05-01
CC one electron properties logic

amjames:
## Description
Modifies how one electron properties are computed to allow the `properties` kwarg to behave consistently with other modules.

## Todos
* **Developer Interest**
  - [x] CCDENSITY no longer calls into OEPROPs this is now done in the python layer. 
  - [x] After calling CCDENSITY the `wfn.arrays['CC ROOT {n} D{a|b}']` can be used to access the OPDM for excited states in the SO basis as in the convention for matrix quantities set on the wfn. 
  - [x] Excited state OPDMs are now set in the wfn.arrays, so that they can be used to compute 1e properties for all roots requested in EOM-CC* calls.
 - [x] Now additional valid one electron properties can be passed as part of the `properties=[]` list. These would previously cause a validation error. Valid one electron properties are now:
    - dipole *
    - quadrupole *
    - mulliken_charges **
    - lowdin_charges
    - mayer_indices
    - wiberg_lowdin_indices
    - no_occupations **
  - [x] The previous behavior was such that requesting the properties marked with * above would lead to all * -ed properties and additional (**) properties being computed automatically. The new logic implemented in the driver still has this behavior so existing workflows do no need to change, at least for now. 

## Questions
- [x] @loriab The block of code that I removed from CCDENSITY had comments that were used to trigger documentation of some psivars. Where should these go now? 

## Checklist
  -  I ran all of the CC tests and there were no errors. The new logic changes where in the output file the 1 electron properties appear, but the perl script still finds them since they are labeled/formatted the same way. 

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/ccdensity/ccdensity.cc
+++ b/psi4/src/psi4/ccdensity/ccdensity.cc
@@ -53,16 +53,16 @@
 #include "psi4/libqt/qt.h"
 #include "psi4/libmints/mintshelper.h"
 #include "psi4/libmints/matrix.h"
-#include "psi4/libmints/oeprop.h"
-#include "psi4/libmints/writer.h"
-#include "psi4/libmints/molecule.h"
-namespace psi { namespace ccdensity {
+namespace psi {
+
+class Molecule;
+namespace ccdensity {
 
 void init_io(void);
 void title(void);
 void get_moinfo(std::shared_ptr<Wavefunction> wfn);
 void get_frozen(void);
-void get_params(Options& options);
+void get_params(Options &options);
 void exit_io(void);
 void onepdm(struct RHO_Params);
 void sortone(struct RHO_Params);
@@ -112,8 +112,8 @@ void x_xi_oe_intermediates(void);
 // void G_norm(void);
 void zero_onepdm(struct RHO_Params rho_params);
 void zero_twopdm(void);
-void get_rho_params(Options& options);
-void get_td_params(Options& options);
+void get_rho_params(Options &options);
+void get_td_params(Options &options);
 void td_setup(struct TD_Params S);
 void tdensity(struct TD_Params S);
 void td_print(void);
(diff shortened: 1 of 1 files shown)


psi4/driver/procrouting/proc.py
```diff
@@ -2551,6 +2548,12 @@ def run_cc_property(name, **kwargs):
     else:
         raise ValidationError("""The "properties" keyword is required with the property() function.""")
 
+    # People are used to requesting dipole/quadrupole and getting dipole,quadrupole,mulliken_charges and NO_occupations
+    if ('dipole' in one) or ('quadrupole' in one):
+        for extra_1e in ['dipole', 'quadrupole', 'mulliken_charges', 'no_occupations']:
+            if extra_1e not in one:
+                one.append(extra_1e)
```
loriab: Is this the same logic?
```
if ('dipole' in one) or ('quadrupole' in one):
    one = list(set(one.extend(['dipole', 'quadrupole', 'mulliken_charges', 'no_occupations'])))
```

psi4/src/psi4/ccdensity/ccdensity.cc
```diff
@@ -383,72 +383,29 @@ PsiReturnType ccdensity(std::shared_ptr<Wavefunction> ref_wfn, Options& options)
...
-      if(params.nstates > 1){
-        Process::environment.globals["CC ROOT 0 DIPOLE X"] = Process::environment.globals["CC DIPOLE X"];
-        Process::environment.globals["CC ROOT 0 DIPOLE Y"] = Process::environment.globals["CC DIPOLE Y"];
-        Process::environment.globals["CC ROOT 0 DIPOLE Z"] = Process::environment.globals["CC DIPOLE Z"];
-        Process::environment.globals["CC ROOT 0 QUADRUPOLE XX"] = Process::environment.globals["CC QUADRUPOLE XX"];
-        Process::environment.globals["CC ROOT 0 QUADRUPOLE XY"] = Process::environment.globals["CC QUADRUPOLE XY"];
-        Process::environment.globals["CC ROOT 0 QUADRUPOLE XZ"] = Process::environment.globals["CC QUADRUPOLE XZ"];
-        Process::environment.globals["CC ROOT 0 QUADRUPOLE YY"] = Process::environment.globals["CC QUADRUPOLE YY"];
-        Process::environment.globals["CC ROOT 0 QUADRUPOLE YZ"] = Process::environment.globals["CC QUADRUPOLE YZ"];
-        Process::environment.globals["CC ROOT 0 QUADRUPOLE ZZ"] = Process::environment.globals["CC QUADRUPOLE ZZ"];
```
loriab: Just leave the P::e block here, only commented. psivar scraper for docs is so dumb it doesn't notice.

psi4/src/psi4/ccdensity/ccdensity.cc
```diff
@@ -383,72 +383,29 @@ PsiReturnType ccdensity(std::shared_ptr<Wavefunction> ref_wfn, Options& options)
...
-      oe->compute();
-      /*- Process::environment.globals["CC ROOT n DIPOLE X"] -*/
-      /*- Process::environment.globals["CC ROOT n DIPOLE Y"] -*/
-      /*- Process::environment.globals["CC ROOT n DIPOLE Z"] -*/
-      /*- Process::environment.globals["CC ROOT n QUADRUPOLE XX"] -*/
-      /*- Process::environment.globals["CC ROOT n QUADRUPOLE XY"] -*/
-      /*- Process::environment.globals["CC ROOT n QUADRUPOLE XZ"] -*/
-      /*- Process::environment.globals["CC ROOT n QUADRUPOLE YY"] -*/
-      /*- Process::environment.globals["CC ROOT n QUADRUPOLE YZ"] -*/
-      /*- Process::environment.globals["CC ROOT n QUADRUPOLE ZZ"] -*/
```
loriab: Same here wrt P::e psivars.

psi4/driver/procrouting/proc.py
```diff
@@ -2514,16 +2514,13 @@ def run_cc_property(name, **kwargs):
...
     """
-    optstash = p4util.OptionsState(
-        ['WFN'],
-        ['DERTYPE'],
-        ['ONEPDM'],
-        ['PROPERTY'],
-        ['CCLAMBDA', 'R_CONVERGENCE'],
-        ['CCEOM', 'R_CONVERGENCE'],
-        ['CCEOM', 'E_CONVERGENCE'])
+    optstash = p4util.OptionsState(['WFN'], ['DERTYPE'], ['ONEPDM'], ['PROPERTY'], ['CCLAMBDA', 'R_CONVERGENCE'],
```
dgasmith: For whatever reason, we usually don't format this as I think the other way seems cleaner.

psi4/driver/procrouting/proc.py
```diff
@@ -2514,16 +2514,13 @@ def run_cc_property(name, **kwargs):
...
+                                   ['CCEOM', 'R_CONVERGENCE'], ['CCEOM', 'E_CONVERGENCE'])
 
-    oneel_properties = ['dipole', 'quadrupole']
+    oneel_properties = [
+        'dipole', 'quadrupole', 'mulliken_charges', 'lowdin_charges', 'mayer_indices', 'wiberg_lowdin_indices',
+        'no_occupations'
+    ]
```
dgasmith: You might want to check the other functions but OEProp should carry around its valid list. It would be good to pull from there.

psi4/driver/procrouting/proc.py
amjames: Should I just block it off with `yapf: enable/disable`?

loriab: add `# yapf: disable` at the end of the line so it's self-contained.

psi4/driver/procrouting/proc.py
amjames: Good point. I thought that list would include transition dipole/quadrupole but I should have checked.

amjames:
I fixed the `oneel_properties` issues, so that the list is not hard-coded in there any more. I also ran `ccdensity.cc` through `c-f` and I removed some includes that are not needed anymore.

amjames:
Yep, merge this at your leisure!

PR #997, opened 2018-05-02
enables plugin gpu_dfcc as external plugin compiled with psi4

edeprince3:
## Description
Enables plugin gpu_dfcc as external plugin compiled with psi4

## Todos
  - [x] enables GPU-accelerated DF-CCSD

## Checklist
- [x] tests comparing energies from DF-CCSD and GPU DF-CCSD -- present in psithon and psiapi
- [x] check for NVIDIA compilers and skip compilation if they are missing -- no need since user's `ENABLE_gpu_dfcc=ON` means build-or-die-trying
- [x] For `ENABLE_gpu_dfcc=ON` and built, on a sys w/o GPUs, cmake tests will fail, pytest test will skip

## Status
- [x] Ready for review
- [x] Ready for merge


edeprince3:
There are still some issues with option handling in the plugin itself that we're working on.  I just wanted to get this PR in today with the hopes it could be accepted for 1.2.

external/downstream/gpu_dfcc/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+        if(${CMAKE_INSIST_FIND_PACKAGE_gpu_dfcc})
+            message(FATAL_ERROR "Suitable gpu_dfcc could not be externally located as user insists")
+        endif()
+
+        include(ExternalProject)
+        message(STATUS "Suitable gpu_dfcc could not be located, ${Magenta}Building gpu_dfcc${ColourReset} instead.")
+
+        ExternalProject_Add(gpu_dfcc_external
+            DEPENDS psi4-core
+            GIT_REPOSITORY https://github.com/edeprince3/gpu_dfcc
```
loriab: add a `GIT TAG <address>` here, where <address> is a tag or a commit

psi4/driver/endorsed_plugins.py
```diff
@@ -33,6 +33,11 @@
 except ImportError:
     pass
 
+try:
+    import gpu_dfcc
+except ImportError:
+    pass
```
loriab: does gpu_dfcc have any python module dependencies that aren't already dependencies of psi, like `import scipy`? If so, add something like [this](https://github.com/loriab/psi4/blob/b8f18eceaa8b09af4b0d3527de5d47421fc14704/psi4/driver/endorsed_plugins.py#L42-L53).

external/downstream/gpu_dfcc/CMakeLists.txt
loriab: or a branch name, during dev

psi4/driver/endorsed_plugins.py
edeprince3: no, there aren't any additional python dependencies

dgasmith:
@edeprince3 Can you add in the tests? I dont see any blockers to getting this in beyond that. See `psi4/tests/chemps2` for an external project example.

loriab:
https://github.com/psi4/psi4/search?utf8=%E2%9C%93&q=v2rdm&type= should be a good guide of places to fill in gpu_dfcc links.

My main reservation is how to supply a conda package for this; but that's not on your plate.

tests/gpu_dfcc/gpu_dfcc1/input.dat
```diff
@@ -0,0 +1,34 @@
...
+
+import gpu_dfcc
+
+molecule {
+           O          0.000000000000     0.000000000000    -0.068516219310   
+           H          0.000000000000    -0.790689573744     0.543701060724   
+           H          0.000000000000     0.790689573744     0.543701060724   
+}
+
+memory 32000 mb
```
loriab: What happens on a computer with only CPUs? Do we need to disable if hardware not detected? Only import the plugin if `num_gpus>1`? Or is it just harmless or non-obliging if no GPUs to work with? Clearly I've just no experience with hardware-dependent tests.

In any case, I hope memory req. can be toned down?

edeprince3:
The code checks the number of GPUs and throws an exception if there are none.

Regarding the memory, the code in its current form requires at least as much CPU memory as the GPU has.  For GP100, that is 16 GB.  I’ll reduce this test case to 16 GB, but I need to think carefully how to reduce it safely to a “normal” amount.

edeprince3:
I think we need to add a check to skip compilation if there is no cuda.

edeprince3:
OK, I've reverted that change and updated gpu_dfcc to have find_package(CUDA REQUIRED).  Aside from the large memory requirement @loriab mentioned, I think everything is good on the psi4 and gpu_dfcc sides.

CMakeLists.txt
```diff
@@ -90,6 +90,7 @@ option(ENABLE_gdma "Enables Stone's GDMA multipole code (requires Fortran)" OFF)
 option(ENABLE_PCMSolver "Enables PCMSolver library (requires Fortran)" OFF)
 option(ENABLE_snsmp2 "Enables SNSMP2 plugin (can also be added at runtime)" OFF)
 option(ENABLE_v2rdm_casscf "Enables V2RDM_CASSCF plugin (can also be added at runtime)" OFF)
+option(ENABLE_gpu_dfcc "Enables gpu accelerated dfcc" OFF)
```
loriab: `Enables gpu accelerated df-cc plugin (requires CUDA; can also be added at runtime)`

May as well add the `(requires Fortran;` to v2, too.

external/downstream/gpu_dfcc/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+                       -DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
+                       -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
+                       -DPYMOD_INSTALL_LIBDIR=${PYMOD_INSTALL_LIBDIR}
+                       -DENABLE_XHOST=${ENABLE_XHOST}
+                       -DENABLE_GENERIC=${ENABLE_GENERIC}
+                       -Dpsi4_DIR=${STAGED_INSTALL_PREFIX}/share/cmake/psi4
+                       -Dpybind11_DIR=${pybind11_DIR}
+            CMAKE_CACHE_ARGS -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}
+                             -DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS}
+                             -DCMAKE_Fortran_FLAGS:STRING=${CMAKE_Fortran_FLAGS}
```
loriab: No need for Fortran COMPILER or FLAGS, I think. On the other hand, you will want to pass through `CMAKE_CUDA_COMPILER` and `CUDA_TOOLKIT_ROOT_DIR`.

tests/gpu_dfcc/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(gpu_dfcc1)
+add_subdirectory(gpu_dfcc2)
```
loriab: Would you humor me and just name them `gpudfcc1` or similar. No one can remember `-`/`_`, so I've banned the latter from test case names (plugin names can't be helped).

tests/gpu_dfcc/gpu_dfcc1/input.dat
```diff
@@ -0,0 +1,29 @@
+#! aug-cc-pvdz H2O Test DF-CCSD vs GPU-DF-CCSD
+
+sys.path.insert(0,'../../../.')
+
+import gpu_dfcc
```
loriab: shouldn't need the `sys.path.insert` or the `import gpu_dfcc` lines.

edeprince3: oh duh, I just copied that from the plugin.  removed!

loriab:
@edeprince3, be advised that I've pushed to your PR (pull before continued dev)

@psi4/editors, this is ready for review. a working conda package exists, so I'm comfortable merging.

tests/pytest/addons.py
```diff
@@ -31,6 +31,30 @@ def is_numpy_new_enough(version_feature_introduced):
...
+        try:
+            ngpu = len(GPUtil.getGPUs())
+        except OSError:  # py3 FileNotFoundError
+            # no `nvidia-smi`
+            return False
+        else:
+            return ngpu > 0
+
+
+hardware_nvidia_gpu = pytest.mark.skipif(is_nvidia_gpu_present() is False,
```
dgasmith: Rought.

loriab: agreed. exporting the nGPUs from `gpu_dfcc` would be better, but that needs plugin overhaul, pybind11. Or I could write my own subprocess/nvidia-smi/count_gpus fn, but I don't dare do that w/o an `nvidia-wmi` cmd to test on. hence, a hack.

dgasmith: Agree that this is probably the best solution for now. I wouldn't recommend writing out our own logic here.

PR #1001, opened 2018-05-03
Force use of finite difference gradient with PCM

robertodr:
## Description
The user needs not set `dertype='energy'` explicitly anymore when calling `optimize`. The test case has been extended accordingly.

## Todos
  - [x] FInite difference gradient is always forced when using PCM 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


andysim:
Is it worth trying something like what we do for ECPs, where the SCF analytic gradients are computed, with finite difference PCM terms added on?

Issue #1005, opened 2018-05-03
General HF Stability Analysis w/ root following.


amjames:
RHF/ROHF stability analysis is already implemented but following instabilities (`STABILITY_ANALYSIS=FOLLOW`) is only available with UHF.

Once we have the generalized Davidson solver (#886) it will be possible to:
 - Add RHF/ROHF root following
 - Move the UHF stability analysis code to use the solver and maintain a single interface
 
See #973 for some more information.

dgasmith:
Importantly the RKS/UKS Hessian vector products are available as well, so root following will be enabled for all SCF Wavefunctions.

PR #1006, opened 2018-05-03
Stop RHF/ROHF stability_analysis from silently skipping a "FOLLOW"

### LINKED ISSUE #973: SCF stability analysis with FOLLOW needs some love
SCF stability analysis with FOLLOW is currently only implemented for UHF. RHF and ROHF have stability analysis, but do not have the capability to follow the instability. This is not documented at all, as far as I can tell, and what is even worse is that if the user requests a FOLLOW with RHF, it "fails" silently.
That is, the SCF and the stability analysis is run, and even if there is an instability Psi4 just carries on.

So I propose two short term solutions:
1. Update the docs to make it clear that FOLLOW is currently an UHF only feature.
2. Implement checks against the user requesting FOLLOW with RHF or ROHF. I think exiting with error is the reasonable action when the user requests an unavailable feature.

Of course the long term solution would be to implement follow for R(O)HF, but that is probably more of a wishlist item, than anything else.

I found these issues a while ago and planned to fix them myself, but never got around to actually do so, and will not have the time to work on them before the 1.2 release.

amjames:
## Description
closes #973 

## Todos
  - [x] setting `stability_analysis = "FOLLOW"` will cause a `ValidationError` when used with `REFERENCE` that is not `UHF`
  - [x] SCF docs have been updated to include a note that follow is only implemented for UHF, and the stability analysis methods table has been expanded to include a column showing valid `STABILITY_ANALYSIS` values for each reference.

#973 Contained the report that RHF stability analysis = follow would silently check for instabilities and then continue on doing nothing about them, and a request to implement FOLLOW for RHF/ROHF. This PR fixes the first part, The feature request has been added under a separate issue (#1005)

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -667,6 +667,9 @@ if the instability still exists. For more attempts, set |scf__max_attempts|;
 the default value of 1 is recommended. In case the SCF ends up in the same minimum, modification
 of |scf__follow_step_scale| is recommended over increasing |scf__max_attempts|.
 
+.. note:: Setting the option |scf__stability_analysis| to ``FOLLOW`` is only avalible for UHF. When using 
+   RHF and ROHF instabilities can be checked, but not followed. If you want to attempt to find a lower energy solution
+   you should re-run the calculation with |scf__reference| set to ``UHF``.
 
 The main algorithm available in |PSIfour| is the Direct Inversion algorithm. It can *only*
 work with |scf__scf_type| ``PK``, and it explicitly builds the full electronic Hessian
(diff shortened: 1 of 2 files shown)


doc/sphinxman/source/scf.rst
```diff
@@ -667,6 +667,9 @@ if the instability still exists. For more attempts, set |scf__max_attempts|;
 the default value of 1 is recommended. In case the SCF ends up in the same minimum, modification
 of |scf__follow_step_scale| is recommended over increasing |scf__max_attempts|.
 
+.. note:: Setting the option |scf__stability_analysis| to ``FOLLOW`` is only avalible for UHF. When using 
+   RHF and ROHF instabilities can be checked, but not followed. If you want to attempt to find a lower energy solution
+   you should re-run the calculation with |scf__reference| set to ``UHF``.
```
andysim: +1 for this note.  Seems to be point of confusion among many people

PR #1010, opened 2018-05-07
FINDIF Overhaul

JonathonMisiewicz:
## Description
FINDIF's geometry generator is brought Py-Side, per [#884](https://github.com/psi4/psi4/issues/884).

## Todos
* **Developer Interest**
  - [x] Geometry generators are now unified, so we don't need to edit all three generators when one is bugged.
  - [x] Various PyBinding has been done. Among other things, the new Py-side compute_atom_map function from pyvib2 now works with psi4 molecules.
  - [x] Fixes a bug in the `CharacterTable` copy constructor member initialization list, which used to require a string to be an integer.
  - [x] Getting geometries for finite difference calculations is now Py-side.
  - [x] Computing gradients from energies is now Py-side.
  - [x] Computing hessians from gradients is now Py-side.
  - [ ] Computing hessians from energies _will be_ Py-side.
  - [ ] Add the banner.
  - [ ] Finish renaming the finite difference functions.
  - [ ] Review the findif test suite and add any tests as needed. In particular, make sure the functions are agnostic as to psi4 vs. qcdb molecules.
  - [ ] Run the entire test suite, to make sure my PyBind'ing hasn't caused any deep problems.
  - [x] A warning now displays if you attempt to access-by-index or iterate over an irreped object, instead of using a NumPy accessor.

## Questions
  - [ ] I had to change the internals of SymmetryOperation to make PyBind happy. I trust there are no objections?

## Status
- [ ] Ready to go


JonathonMisiewicz:
The git history is obviously pretty screwy. I'll... get this fixed before this is actually accepted.

JonathonMisiewicz:
Due to severe history contamination, it's easier to just PR with a different branch, and thankfully, no comment history will be lost.

In Lori's words: _Yes, that one suggestion of git’s that you merge instead of force push is catastrophic. I did it twice he first time I did a rebase pr. Once burned, you never do it again._

PR #1022, opened 2018-05-14
Samples update

loriab:
## Description
updates the `samples/` directory. spare yourself the diff. first commit is the only non-copy change.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
The `documents_tests` script prints out a list of missing tests. So when I actually look at the output annually (since docs are built by CI), it's not hard to patch up the strays.

PR #1020, opened 2018-05-14
SCF_TYPE and SAPT mem

loriab:
## Todos
* **Developer Interest**
  - [x] Moves SCF_TYPE to globals, psi4/psi4#1019
  - [x] Change SAPT mem to long int, psi4/psi4#1018

## Questions
- [ ] someone who knows SAPT & C++ should check over the `long int` business

## Checklist
- [x] full tests on SCF_TYPE, sapt tests on long int

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1023, opened 2018-05-15
Additional JK logic

### LINKED ISSUE #1021: New SCF types unrecognized by several downstream modules
Several downstream codes do not recognize the new SCF types `mem_DF` and `disk_DF` (introduced in #961) when set directly within the input file.  I've attached [a tarball](https://github.com/psi4/psi4/files/2002517/eff-scftype-df.tar.gz) containing several such tests, adapted directly from the existing Psi4 test suite, which attempt to highlight these problems.  I've also attached the summary file (from the tarball) separately for easy web viewing [here](https://github.com/psi4/psi4/files/2002535/summary.txt).

Additionally worth note is that a few of these downstream codes fail gracefully (e.g., FNOCC for `scf_type mem_df`, as it assumes a disk algorithm) with helpful error messages, but others do not (e.g., CISD energies fail with PSIO errors and DLDF+D09 gradients segfault).  However, if failure is unavoidable, we should probably at least be verbose in the error message.

dgasmith:
## Description
Corrects additional JK logic edge cases found and increases the verbosity of the error messages. At the same time we should continuously discourage users to set a `SCF_TYPE` to anything besides `DF`. `MEM_DF` and `DISK_DF` are truly expert user options and I believe have been marked as such in a variety of places.

Also added a throw if someone requests dlDF+D gradients as these have never been implemented. 

Fixes #1021.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
From this, I realize that a consequence of #1020 is that most driver calls to `('SCF', 'SCF_TYPE')` will need to drop the first SCF.

dgasmith:
The failure is that somewhere in the cc-module test `SCF_TYPE` is not correctly restored to defaults. This is convolved with #1020 and Lori's comment. Lets get that in and come back to this.

psi4/driver/procrouting/proc.py
```diff
@@ -3862,8 +3833,7 @@ def run_fnodfcc(name, **kwargs):
         ['FNOCC', 'RUN_CEPA'],
         ['FNOCC', 'DF_BASIS_CC'],
         ['SCF', 'DF_BASIS_SCF'],
-        ['SCF', 'DF_INTS_IO'],
-        ['SCF', 'SCF_TYPE'])
```
loriab: I'm not quite seeing why to separately add SCF_TYPE to the optstash later, just because it's in another fn. As long as it doesn't have to be restored from other fn, harmless to add it here, even if it never gets used.

But this way works, too.

dgasmith: I thought it reduced the cognitive load. Anyone writing this code doesn't have to *know* that this random function requires this opt stash state or you leak internal options.

loriab: Ok. Here’s to a better future of decorators and context managers.

PR #1024, opened 2018-05-18
FINDIF Overhaul

JonathonMisiewicz:
## Description
FINDIF's geometry generator is brought Py-Side, per [#884](https://github.com/psi4/psi4/issues/884).

## Todos
* **Developer Interest**
  - [x] Geometry generators are now unified, so we don't need to edit all three generators when one is bugged.
  - [x] Various PyBinding has been done. Among other things, the new Py-side compute_atom_map function from pyvib2 now works with psi4 molecules.
  - [x] Fixes a bug in the `CharacterTable` copy constructor member initialization list, which used to require a string to be an integer.
  - [x] Fixes a bug in vibrational analysis where the zero K enthalpy variable was misspelled.
  - [x] Getting geometries for finite difference calculations is now Py-side.
  - [x] Computing gradients from energies is now Py-side.
  - [x] Computing hessians from gradients is now Py-side.
  - [x] Computing hessians from energies is now Py-side.
  - [x] Finish renaming the finite difference functions.
  - [x] Overly picky changes for consistent style and commenting.
  - [x] Review the findif test suite and add any tests as needed.
  - [x] Add the banner.
  - [x] Remove old FINDIF code (after migrating to attic).
  - [x] Removed the C-side gradient writer.
  - [x] Added several new gradient writing formats.
  - [x] A warning now displays if you attempt to access-by-index or iterate over an irreped object, instead of using a NumPy accessor.
  - [x] Gradients can now be written, even for analytic hessians.
  - [x] The .grad file will now only write the latest hessian.

## Questions
  - [x] I had to change the internals of SymmetryOperation to make PyBind happy. I trust there are no objections?
  - [x] Since pyvib2 made hessian writing no longer FINDIF's responsibility, shall I make gradient writing no longer FINDIF's responsibility as well? Then `comp_grad_from_hess` could return a Numpy array, like the `comp_hess_from_X` functions.

## Checklist
- [x] Tests added for any new features (or rather, edited in)
- [x] [Relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
The commit history is duplicating some others' old commits again... I'll work on this.

JonathonMisiewicz:
Unless something horribly unexpected happens in Travis, I say this is ready for review.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,812 @@
...
+
+    Returns
+    -------
+    gradient : psi4.core.Matrix
+        The gradient in Cartesians, as a matrix with dimensions
+        number-of-atoms by 3. """
+
+    # NOTE: Yes, this returns psi4.core.Matrix, not an ndarray.
+    # Due to GradientWriter, the function internally needs the gradient as a psi4.core.Matrix.
+    # TODO: Move GradientWriter py-side, as well? This function would simplify nicely.
```
jturney: How difficult would it be to move GradientWriter py-side, too? I'm not too familiar with where it is C-side.

JonathonMisiewicz: It doesn't look hard at all. GradientWriter is defined in `libmints/writer.*` and only ever used in `findif/fd_1_0.cc`. I don't know how we write analytic gradients (would have to check), but if the grep on `GradientWriter` didn't catch anything else, it's probably Python tech that can be reused for findif.

jturney: Then I say let's move it py-side.

JonathonMisiewicz: Shall I write it as a qcdb module, or a standalone in the driver? We'd be fine with a standalone for now, but I can very easily envision somebody wanting a load_gradient option. Maybe they optimized in one job and want to skip the initial gradient computation when doing a hessian in another. If we have writing and reading gradients, it would make more sense to have a module for the both of them, just like we do with hessians.

loriab: I think it makes sense to parallel hessian as much as possible, so add to qcdb and give the format a name so the fun can handle variations in future.

JonathonMisiewicz: Next question for the core devs: The gradients are currently printed in file11 format, where the first line is the molecule name, the word "(wfn)" and the word "(dertype)". This is followed by the number of atoms and current energy, the masses and coordinates, and _then_ the gradient.

Keep all the extra stuff, or prune the gradient file to be an analogue of the hessian file? @jturney couldn't quite decide when I asked him.

psi-rking: Oh, I just realized you were debating whether to print the geometry too.  I will just say that in practice I find myself working with, e.g., CFOUR hessian file (with which I am less familiar) and wondering if I am completely sure what Cartesian geometry it belongs to.  Perhaps with modern object packaging of molecular data this problem will go away, but it's nice to be able to verify.

JonathonMisiewicz: Another picky question: We currently write multiple gradients to the same file. See fd-gradient test for an example. Keep this behavior or scrap it? My vote would be to have one gradient for one file, but there may be some use case I'm not seeing.

psi-rking:
We used to use files like file11 for subsequent parsing for
finite-difference computations and for analysis of optimization
trajectories.  Now we can so easily return the trajectory through python,
this doesn't seem necessary.  I presume someone has an eye on what WebMO
parses, or else we do what we want and let WebMO accommodate it.

jturney:
@JonathonMisiewicz Can you resolve the conflict in chartab.cc?  Then I think we're good to go.

JonathonMisiewicz:
Latest batch of commits were to resolve a merge conflict caused by [this line](https://github.com/psi4/psi4/pull/953/commits/30fab96f0902493685f79e48c9182d1f590072bf#diff-fdb076736e7005bb09b8e65240248818R10) of #953.

psi4/driver/driver.py
```diff
@@ -1493,8 +1495,8 @@ def hessian(name, **kwargs):
 
         # Assemble Hessian from gradients
         #   Final disp is undisp, so wfn has mol, G, H general to freq calc
-        H = core.fd_freq_1(molecule, gradients, irrep)  # TODO or moleculeclone?
-        wfn.set_hessian(H)
+        H = driver_findif.comp_hess_from_grad(molecule, gradients, irrep)  # TODO or moleculeclone?
+        wfn.set_hessian(core.Matrix.from_array(H))
         wfn.set_gradient(G0)
         wfn.set_frequencies(core.get_frequencies())
```
loriab: I think the global get_/set_frequencies in [`P::e`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libpsi4util/process.h#L60) can be killed off at this happy juncture. The wfn setting is already happening in `def freq`

psi4/driver/driver.py
```diff
@@ -1798,6 +1800,32 @@ def frequency(name, **kwargs):
...
+    project_trans : boolean
+        Should translations be projected in the harmonic analysis?
+    project_rot : boolean
+        Should rotations be projected in the harmonic analysis?
+
+    Returns
+    -------
+    vibinfo : dict
+        A dictionary of vibrational information. See psi4/driver/qcdb/vib.py:harmonic_analysis
+    """
```
loriab: Ah, thank you! I was just thinking today that this needed a docstring. A few changes:

```
wfn : :py:class:`~psi4.core.Wavefunction`
```

```
hess : ndarray of float, optional
    Hessian to analyze, if not the hessian in wfn.
    (3*nat, 3*nat) non-mass-weighted Hessian in atomic units, [Eh/a0/a0].
```
```
molecule : :py:class:`~psi4.core.Molecule` or qcdb.Molecule, optional
```
```
See :py:func:`~psi4.driver.qcdb.vib.harmonic_analysis`
```

psi4/driver/qcdb/gradparse.py
```diff
@@ -0,0 +1,94 @@
...
+    None
+
+    """
+    # Validate input.
+    try:
+        assert grad.ndim == 2 and grad.shape[1] == 3
+    except AssertionError:
+        raise ValidationError("gradient shape much be number of atoms * 3. Received shape: {}".format(grad.shape))
+
+    if dtype in {'file11', 'GRD'}:
```
loriab: square, not curly, brackets, I think.

also, default dtype is 'psi4', which then isn't an option.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,799 @@
+from __future__ import division
+from psi4.driver.p4util.exceptions import ValidationError
+from psi4.driver.qcdb import libmintsmolecule
```
loriab: Help me keep libmintsmolecule hidden, please. Perhaps try it import it from molecule instead (https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/__init__.py#L40) or failing that just move it to molecule.py or add it as a staticmethod to qcdb.Molecule.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,799 @@
...
+        },
+        5: {
+            "sym": ((-2, ), (-1, ), (1, ), (2, )),
+            "asym": ((-2, ), (-1, )),
+            "off": ((-1, -2), (-2, -1), (-1, -1), (1, -1), (-1, 1), (1, 1), (2, 1), (1, 2))
+        }
+    }
+
+    # We won't need disps for a while, but let's validate num_pts early.
+    try:
```
dgasmith: Can we write this as `if num_pts not in pts_dict: raise...`? Slightly more canonical.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,799 @@
...
+            core.print_out("\tNumber of displacements per irrep:\n")
+            for i, ndisp in enumerate(Ndisp_pi, start=1):
+                core.print_out("\t  Irrep {:d}: {:d}\n".format(i, ndisp))
+
+    if print_lvl > 1 and verbose:
+        for salc in salc_list:
+            salc.print_out()
+
+    data.update({
+        "Ndisp_pi": Ndisp_pi,
```
dgasmith: Mixing cases here a bit, for pure snake case this should be like `ndisp_pi` or `n_disp_pi`.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,799 @@
...
+            Ndisp += len(indices) * (len(indices) - 1) // 2 * len(disps["off"])
+        Ndisp_pi.append(Ndisp)
+
+    # Let's print out the number of geometries, the displacement multiplicity, and the CdSALCs!
+    if print_lvl and verbose:
+        core.print_out("\tNumber of geometries (including reference) is {:d}.\n".format(sum(Ndisp_pi) + 1))
+        if method_allowed_irreps != 0x1:
+            core.print_out("\tNumber of displacements per irrep:\n")
+            for i, ndisp in enumerate(Ndisp_pi, start=1):
+                core.print_out("\t  Irrep {:d}: {:d}\n".format(i, ndisp))
```
dgasmith: I think we should probably avoid `\t` as its definition can vary and can mess up other alignment.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,799 @@
...
+        core.print_out("\n\t Coord      " + energy_string + "    Force\n")
+        for salc in range(data["Nsalc"]):
+            print_str = "\t{:5d}" + " {:17.10f}" * (e_per_salc) + " {force:17.10f}" + "\n"
+            energies = E[e_per_salc * salc:e_per_salc * (salc + 1)]
+            print_str = print_str.format(salc, force=g_q[salc], *energies)
+            core.print_out(print_str)
+        core.print_out("\n")
+
+    B = data["salc_list"].matrix()
+    g_cart = np.matmul(g_q, B)
```
dgasmith: `matmul` was implemented for the `@` python operator. This is still a bit of an experimental function in NumPy despite it being in many releases. Would use `np.dot` in general.

psi4/driver/qcdb/gradparse.py
JonathonMisiewicz: I'll change 'psi4' to 'file11' in the next commit, after tests pass. Curse my indecisiveness on naming the various gradient formats!

Curly braces is another way to initialize a set. Didn't this come up earlier in the review process, too?

psi4/driver/driver.py
JonathonMisiewicz: `wfn.set_frequencies` is called in `def freq`, yes. `wfn.set_hessian` and `wfn.set_gradient` are not. I'll strip the set_frequencies call in `def hessian`, but keep the others.

And yes, delighted to formally kill off some globals while I'm at it.

psi4/driver/qcdb/gradparse.py
loriab: You're right -- I need to remember sets. But does this have an advantage over plain `if dtype in ['file11', 'GRD']:`?

JonathonMisiewicz: Practically, no. From how I learned Python, I was under the impression that while lists were usually taught first, lists were no more plain than sets were. If you think lists are the plainer/more-Pythonic type, I can change to a list.

psi4/driver/driver_findif.py
JonathonMisiewicz: I can, but I'll still need to define `disps` somewhere. Probably closer to when I actually need it.

psi4/driver/qcdb/gradparse.py
dgasmith: Mosty ordered vs unordered difference (doesn't matter here). For a medium number of items sets will be faster and to `log(N)` lookup. Here, sets may be slower due to overhead. Not entirely sure why, but this kind of operation is, seemingly, exclusively implemented as a list.

JonathonMisiewicz:
All requested changes made. I have also run the _entire_ test suite, modulo plugins and dft benchmarks. I stumbled across a rather critical bug that way: empirical_dispersion needed to be updated with the new FINDIF.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,795 @@
...
+    # TODO: Replace with a generator from a stencil to a set of points.
+    pts_dict = {
+        3: {
+            "sym": ((-1, ), (1, )),
+            "asym": ((-1, ), ),
+            "off": ((1, 1), (-1, -1))
+        },
+        5: {
+            "sym": ((-2, ), (-1, ), (1, ), (2, )),
+            "asym": ((-2, ), (-1, )),
```
loriab: Any particular reason to choose neg rather than pos for the single-sided step?

Are "sym" and "asym" good labels, considering that symmetric modes have asymmetric potentials and asymmetric modes have symmetric potentials?

JonathonMisiewicz: 90% sure that choosing the negative step was the C!FINDIF convention. I don't care either way.
`sym_irr` and `asym_irr` would be better, now that you mention it. I'll hold off on making any changes until you say you're done with the review. No need to clog the commit history.

psi-rking:
If I say the "asymmetric stretch for water has a higher frequency than the
symmetric stretch?", you know what I mean.

There is a century of convention here that we use the symmetry of the
coordinate to describe the motion (not the potential).  Right?


On Mon, Aug 6, 2018 at 2:29 PM, Jonathon Misiewicz <notifications@github.com

psi4/driver/driver_findif.py
loriab: Ok on the negative step. Just curious.

I'll throw "oneside" and "twoside" into the ring as possible labels.

I confess to reviewing bits at a time. Push at convenience.

loriab: @psi-rking oh, I definitely agree once you throw "stretch" in there, it's clear that the subject of the "(a)sym" is mode. It just struck me that in describing a stencil, there's room for confusion.

psi4/driver/p4util/numpy_helper.py
```diff
@@ -496,6 +501,8 @@ def _chain_dot(*args, **kwargs):
 core.Matrix.to_serial = _to_serial
 core.Matrix.from_serial = classmethod(_from_serial)
 core.Matrix.chain_dot = _chain_dot
+core.Matrix.__iter__ = _irrep_access
+core.Matrix.__getitem__ = _irrep_access
```
dgasmith: Thanks, great fix!

psi4/driver/p4util/numpy_helper.py
```diff
@@ -578,3 +587,32 @@ def _dimension_iter(dim):
 core.Dimension.from_list = _dimension_from_list
 core.Dimension.to_tuple = _dimension_to_tuple
 core.Dimension.__iter__ = _dimension_iter
+
+# General functions for NumPy array manipulation
+def block_diag(*args):
```
dgasmith: Can we expand the name to `block_diagonal_array` to be very explicit here?

psi4/driver/qcdb/gradparse.py
```diff
@@ -0,0 +1,94 @@
...
+# @END LICENSE
+#
+
+import numpy as np
+
+from .util import filter_comments
+
+# TODO: Include gradient loading feature, analagously to hessian loading.
+
+def to_string(grad, handle, dtype='file11', mol=None, energy=None):
```
dgasmith: Can we change this to `gradeint_to_string`, again to be very explicit?

psi4/src/psi4/libmints/pointgrp.h
```diff
@@ -157,13 +160,17 @@ class SymmetryOperation
     double trace() const
     { return d[0][0] + d[1][1] + d[2][2]; }
 
+    // For the sake of Python export, expose d to reading
+    const std::array<std::array<double, 3>, 3>& matrix()
+    { return d; }
```
dgasmith: I see these are historical, but this should be on a single line. `const std::array<std::array<double, 3>, 3>& matrix() { return d; }`. Can either fix now or when c-f gets it next pass.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,795 @@
...
+
+    Notes
+    -----
+    Only symmetric displacements are necessary, so user specification of
+    symmetry is disabled.
+    """
+    return _geom_generator(molecule, -1, "1_0")
+
+
+def geoms_hess_from_grad(molecule, irrep):
```
dgasmith: Up to you but `fd_2_1` is about as clear as the this line to me. Can we either go back to a shorthand or expand this out to `hessian_from_gradient_geometries` or something else very explicit.

psi4/driver/qcdb/gradparse.py
JonathonMisiewicz: That would break mirroring of `hessparse.py`, which also has a `to_string` method. I can rename that one to `hessian_to_string`, while I'm at it.

loriab: can we just always use it namespaced `gradparse.to_string()` unless in-class? there's a few of these -- `hessparse.to_string()`, `molparse.to_string()`, etc.

JonathonMisiewicz: Good point. The namespace would make the renaming redundant, and the namespacing is standard anyway.

psi4/src/psi4/libmints/pointgrp.h
JonathonMisiewicz: I'll leave this as part of next week's great clang-formatting. Breaking consistency would make the section less readable IMO, and formatting pointgrp would clutter this PR.

JonathonMisiewicz:
The last batch of changes should address @dgasmith's comments.

psi4/driver/qcdb/gradparse.py
```diff
@@ -0,0 +1,94 @@
...
+    Returns
+    -------
+    None
+
+    """
+    # Validate input.
+    try:
+        assert grad.ndim == 2 and grad.shape[1] == 3
+    except AssertionError:
+        raise ValidationError("gradient shape much be number of atoms * 3. Received shape: {}".format(grad.shape))
```
loriab: If you always apply `grad.reshape(-1, 3)`, should suffice to check `shape[0] == mol.natom()` and input can be free-er.

psi4/src/core.cc
```diff
@@ -306,46 +306,6 @@ PsiReturnType py_psi_mrcc_load_densities(SharedWavefunction ref_wfn, const py::d
...
-
-SharedMatrix py_psi_fd_freq_1(std::shared_ptr<Molecule> mol, const py::list& grads, int irrep) {
-    py_psi_prepare_options_for_module("FINDIF");
-    return findif::fd_freq_1(mol, Process::environment.options, grads, irrep);
-}
-
-void py_psi_displace_atom(SharedMatrix geom, const int atom, const int coord, const int sign,
-                                  const double disp_size) {
-    findif::displace_atom(geom, atom, coord, sign, disp_size);
-}
```
loriab: deletions, hooray

psi4/src/export_mints.cc
```diff
@@ -691,6 +693,23 @@ void export_mints(py::module& m) {
              "Returns a new Matrix object named name with default dimensions");
 //              py::arg("name"), py::arg("symmetry"));
 
+    py::class_<CdSalc::Component, std::shared_ptr<CdSalc::Component>>(
+        m, "SalcComponent", "Component of a Cartesian displacement SALC in the basis of atomic displacements.")
+        .def_readwrite("coef", &CdSalc::Component::coef, "The coefficient of the displacement")
+        .def_readwrite("atom", &CdSalc::Component::atom, "The index of the atom being displaced")
+        .def_readwrite("xyz", &CdSalc::Component::xyz, "The direction of tthe displacement, given by x as 1, y as 2, z as 3.");
```
loriab: For `atom`, whenever there's atom indexing mentioned in user-facing docs, please add "0-indexed" or "1-indexed". For `xyz`, typo in "tthe" and is it really 1-indexed?

Thanks for the docstrings.

psi4/src/export_mints.cc
```diff
@@ -1191,6 +1235,7 @@ void export_mints(py::module& m) {
         .def("x", &Molecule::x, "x position [Bohr] of atom arg0 (0-indexed without dummies)")
         .def("y", &Molecule::y, "y position [Bohr] of atom arg0 (0-indexed without dummies)")
         .def("z", &Molecule::z, "z position [Bohr] of atom arg0 (0-indexed without dummies)")
+        .def("xyz", vector_by_index(&Molecule::xyz), "Return the Vector3 for atom i", py::arg("i"))
```
loriab: Please to add "0-indexed"

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,798 @@
+from __future__ import division
```
loriab: Add license header and delete the future import (can be the first py2-free file).

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,798 @@
...
+from psi4.driver.p4util.exceptions import ValidationError
+from psi4.driver.qcdb import molecule
+from psi4 import core
+import numpy as np
+#import scipy.linalg as la
+from psi4.driver.p4util import block_diagonal_array
```
loriab: Go ahead and delete the unused import. And maybe move numpy to the top. Informally, I try to separate into three import blocks:-- std lib, external, internal.

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,798 @@
...
+    mol : qcdb.molecule or psi4.core.Molecule
+        The molecule to compute the gradient of.
+    E : list of floats
+        A list of energies of the molecule at displaced geometries.
+
+    Returns
+    -------
+    gradient : np.array
+        The gradient in Cartesians, as a matrix with dimensions
+        number-of-atoms by 3. """
```
loriab:
```
gradient : ndarray
    (nat, 3) Cartesian gradient [Eh/a0].

"""
```

psi4/driver/driver_findif.py
```diff
@@ -0,0 +1,798 @@
...
+    irrep : int
+        The Cotton ordered irrep to get frequencies for. Choose -1 for all
+        irreps.
+
+    Returns
+    -------
+    disp_geoms : list of psi4.core.Matrix
+        A list of displaced geometries to compute energies at.
+    """
+    return _geom_generator(molecule, irrep, "2_0")
```
loriab: I didn't notice any definite violations, but go ahead and run `yapf` over this file (and add `# yapf: disable` (I think)) if it messes up any formatting you like) so `git blame` doesn't get messed up by a future formatting pass.

JonathonMisiewicz: There were two or three lines worth of newline and whitespace corrections, but that's all.

And with that, all of your points should be taken care of. Once the tests pass again, this should be done!

PR #1032, opened 2018-06-06
fix local options for plugins

loriab:
## Description
So `psi4.clean_options()` is a mainstay of psiapi mode because it allows you to do two calculations after another both starting from a clean slate of options. Likewise, it is important for testing b/c we do one behind the scenes (in pytest) between each test so that your MP2 doesn't get PCM solvation, etc. For plugins, this has been broken since the `clean_options()` command was invented, in that as soon as you tried to set an option unique to the plugin (as opposed to one that another module in read_options.cc defined) _after_ a options cleanse, it would complain about an invalid option. This PR reloads the plugin options after a cleanse.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -547,13 +547,11 @@ void py_psi_clean_options() {
     Process::environment.options.clear();
     Process::environment.options.set_read_globals(true);
     read_options("", Process::environment.options, true);
-    Process::environment.options.set_read_globals(false);
     for (std::map<std::string, plugin_info>::iterator it=plugins.begin(); it!=plugins.end(); ++it) {
         // Get the plugin options back into the global space
-        Process::environment.options.set_read_globals(true);
         it->second.read_options(it->second.name, Process::environment.options);
-        Process::environment.options.set_read_globals(false);
     }
+    Process::environment.options.set_read_globals(false);
 }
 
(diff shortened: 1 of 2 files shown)


psi4/src/core.cc
```diff
@@ -548,6 +548,12 @@ void py_psi_clean_options() {
     Process::environment.options.set_read_globals(true);
     read_options("", Process::environment.options, true);
     Process::environment.options.set_read_globals(false);
+    for (std::map<std::string, plugin_info>::iterator it=plugins.begin(); it!=plugins.end(); ++it) {
+        // Get the plugin options back into the global space
+        Process::environment.options.set_read_globals(true);
```
dgasmith: Should move `Process::env` out of the loop? You may be bale to move this inside the `set_read_globals` two lines above, but I may be misunderstanding exactly what you are doing here.

loriab: I treat both options and plugins as fragile snowflakes. What you say makes a lot of sense, but I'd want Jet/Andy's approval and testing on multiple failing plugins to believe it. :-)

jturney: Should be safe to move the `Process:env...` to outside the loop.  Do you want to make the change before I approve?

loriab:
thanks, @jturney, that should do it. feel free to squash.

PR #1034, opened 2018-06-06
fix PSI_API in libciomr

loriab:
- [x] PSI_API doesn't work for gnu compiler and pointer returning fns whe not at start of decl. Affects `v2rdm_casscf/tests/v2rdm6/input.dat`
- [x] Adds to python module, `psi4.executable` returning the psithon script/binary/executable. Also cleaned up main psi4 namespace by deleting some intermediate strings.
- [x] To the `psi4Config.cmake` consumed by downstream, defines `psi4_SYS_PATH` which is the full path needed to be added to `sys.path` in order to `import psi4`.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Sorry, I added a couple more commits. I don't think they're controversial, but you may want to take another look (4 lines; summary in main PR box). @dgasmith @robertodr

dgasmith:
Looks fine.

robertodr:
Yep, LGTM

PR #1035, opened 2018-06-07
gau2grid: ensure we get the right bf ordering

loriab:
## Description
gau2grid now provides spherical basis sets in `gaussian` _or_ `cca` ordering. let's make sure we get the right one!

~**This won't work until a PR comes into g2g and versions/git tags get bumped here.**~

## Todos
* **Developer Interest**
  - [x] detect/build the gaussian g2g
  - [x] bump min g2g version so the components are there

## Checklist
- [x] ran quicktests against g2g dgasmith/gau2grid#11
- [x] full tests pass locally

## Status
- [x] Ready for review
- [x] Ready for merge


dgasmith:
Gau2grid version minted at v1.1.

loriab:
Ok, Linux conda packages built, and this should be RTG after CI checks.

PR #1039, opened 2018-06-11
misc., mostly build

loriab:
## Description
Accumulating changes here

## Todos
* **Developer Interest**
  - [x] AddOns: make resp an endorsed plugin
  - [x] CMake: fix OMP option name. add `otool -L` parsing
  - [x] CMake: refactoring this to psi4-dev, which it approximately depends on. Current is one layer of templating too many.

## Questions
- [x] Name & home of RESP plugin - shall be GH:cdsgroup/resp and `import resp`

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/endorsed_plugins.py
```diff
@@ -51,3 +51,8 @@
         raise ImportError("""Psi4 plugin 'snsmp2' available, but scipy missing. Try `conda install scipy` or `pip install scipy`.""")
     else:
         pass
+
+try:
+    import resp
+except ImportError:
+    pass
```
andysim: We may want a mechanism like the `XX_FOUND` CMake vars eventually, so we know which plugins are available in the script.  Still, that's another conversation for another day

psi4/share/psi4/scripts/test_threading.py
```diff
@@ -187,22 +187,27 @@ def print_math_ldd(args):
...
+        lddish = 'otool -L'
     else:
-        print('Not available w/o `ldd`')
+        print('Not available w/o `ldd` or `otool`')
+        return True
+
+    cmd = """{} {} | grep -e ':' -e 'mkl' -e 'openblas' -e 'iomp5' -e 'gomp'""".format(lddish, modcore)
+    print('Running {} ...'.format(cmd))
+    subprocess.call(cmd, shell=True)
+    lddout = subprocess.getoutput(cmd)
```
andysim: isn't the `call` command redundant here?

andysim: Not that performance matters - just curious

andysim: Did a little googling on this just because I'm not familiar with `getoutput` (I use `communicate` normally).  Looks like it's new to py3, so just a heads-up that this could break our soon-to-be deprecated py27 support, if I'm right

dgasmith: Hmm, shouldn't this just `lddout = subprocess.check_output(cmd, shell=True)`?

andysim: I think so, which is why I commented initially

psi4/driver/endorsed_plugins.py
loriab: There's [this at runtime](https://github.com/psi4/psi4/blob/master/psi4/extras.py#L126-L152). I would agree with you above the whole `external/downstream/resp/CMakeLists.txt` treatment except that I realized it has a fatal flaw for pure-py plugins. Namely, it tries to import them to see if they're available, but that involves `import psi4`, but that's not built yet ....

psi4/share/psi4/scripts/test_threading.py
loriab: The pipe complicates matters a great deal. It didn't like the substitution.

Guess I'll know why py27 fails, then. Thanks for the warning.

PR #1049, opened 2018-06-22
PSI_API for eigsort function in libciomr

edeprince3:
## Description
PSI_API for eigsort function in libciomr.

## Todos
Nothing notable... just fixes missing symbol in one of my plugins.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libciomr/eigsort.cc
+++ b/psi4/src/psi4/libciomr/eigsort.cc
@@ -35,6 +35,8 @@
 
 #include <cstdlib>
 
+#include "psi4/pragma.h"
+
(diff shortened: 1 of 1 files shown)


loriab:
How many of these plugins do you have? :-)

Are we racing btwn you working through your plugin library and adding `PSI_API`s and me finishing off 1.2?

edeprince3:
@loriab - many, but I think I might be nearing the end of these updates.   I apparently hadn't used this plugin since the days of boost ... it probably would have been easier to rewrite the stupid thing from scratch than to repair it.

PR #1057, opened 2018-06-27
Implementation of interface to i-pi to enable molecular dynamics simulations with psi

tomspur:
## Description
The purpose of this PR is to implement the first bits and pieces to connect psi to [i-pi](https://github.com/i-pi/i-pi) to enable molecular dynamics simulations with psi. For instance, it would be possible to carry out path integral molecular dynamics simulations such as [in this reference](https://doi.org/10.1039/C4CP05192K) with psi.

It would be great to get a first review of the basic idea in the code and I can add some documentation of the above infrastructure after that as well.

## Todos
* **Developer Interest**
  - [x] Infrastructure to connect to [i-pi](https://github.com/i-pi/i-pi) is included in `broker.py`
  - [ ] Infrastructure to connect to [i-pi](https://github.com/i-pi/i-pi) is included (but not yet documented yet)

## Questions
- [ ] Is the dynamic dependency to [i-pi](https://github.com/i-pi/i-pi) to setup the interface on runtime fine? Or would you prefer to have the Client side implementation to have fully here in psi as well? If the latter is the case, this has the disadvantage, that the client side might need adjustments with a new version of i-pi, which is why I left the client implementation in i-pi for now.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


dgasmith:
This looks really cool, thanks for starting this PR. It looks like the CC* changes and the I-PI changes are orthogonal to each other. Could you separate these out into two different PRs/branches so that we can work through them separately. You can use `git cherry-pick` to separate out the commits.

tomspur:
Thank you @dgasmith for the quick feedback.
I just put the I-PI changes now in this PR/branch and tried to add also a small smoke test of the broker, when i-pi is not available on runtime in the unittests to show the concept here.

The other CC changes and minor bug fixes will be put into a different PR then as those changes are indeed not required at this point, but only when also using the amplitudes as well later on in the broker.

psi4/CMakeLists.txt
```diff
@@ -192,6 +192,9 @@ install(DIRECTORY driver
         DESTINATION ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}${PYMOD_INSTALL_LIBDIR}/psi4
         FILES_MATCHING PATTERN "*.py")
 
+install(FILES broker.py
```
dgasmith: We can do this dynamically in Python a bit simpler than this. See below.

psi4/__init__.py
```diff
@@ -98,3 +98,4 @@
 if "@ENABLE_PCMSolver@".upper() in ["1", "ON", "YES", "TRUE", "Y"]:
     sys.path.insert(1, "@PCMSolver_PYMOD@")
 
+from .broker import broker
```
dgasmith:
```
try:
    import ipi
    from .broker import broker
except ImportError:
    # Set some variable noting that this is not available
```

Not CMake required.

psi4/broker.py
```diff
@@ -0,0 +1,132 @@
...
+        print("PSI4 options:")
+        for item, value in self.options.items():
+            print(item, value)
+            if item not in ["LOT", "multiplicity", "charge"]:
+                psi4.core.set_global_option(item, value)
+        psi4.core.IO.set_default_namespace("xwrapper")
+
+        self.timing = {}
+
+    def run_frc(self, pos):
```
dgasmith: Please expand names for clarity (`compute_force` or similar).

psi4/broker.py
```diff
@@ -0,0 +1,132 @@
...
+        """Fetch force, energy of PSI.
+        """
+        if len(pos.shape) == 1:
+            pos = pos.reshape((len(pos) / 3, 3))
```
dgasmith: Probably should use floor division here `len(pos) // 3` to prevent float oddness. Or even better `pos.reshape(-1, 3)`.

psi4/broker.py
```diff
@@ -0,0 +1,132 @@
...
+        self.frc, self.pot = self.callback(pos)
+        return self.frc, self.pot
+
+    def get_molecule(self, pos):
+        ret = "".join(
+                    ["%s %f %f %f\n" % (self.atoms_list[i],
```
dgasmith: Numpy arrays should be indexed as `pos[i, 0]`.

dgasmith: You might want to also join as `"\n".join()` to avoid a lot of these newlines.

psi4/broker.py
```diff
@@ -0,0 +1,132 @@
...
+                                        pos[i][0], pos[i][1],
+                                        pos[i][2])
+                                        for i in range(len(pos))]
+                    + ["units bohr\n"]
```
dgasmith: Be best to write this as a single list.

psi4/broker.py
```diff
@@ -0,0 +1,132 @@
...
+    names = [mol.symbol(i) for i in range(len(atoms))]
+
+    print("Initial atoms", names)
+    b.atoms_list = names
+
+    try:
+        if b.serverdata:
+            b.run()
+        else:
+            print("ATOMS", atoms)
```
dgasmith: Please replace all print statements with `psi4.core.print_out`, keep in mind you will need to manually add newlines.

psi4/broker.py
```diff
@@ -0,0 +1,132 @@
+# Copyright (c) 2013-2018 Thomas Spura <thomas.spura@gmail.com>
```
dgasmith: For files within Psi4 you will need to use the Psi4 copyright header found in all other Python files.

dgasmith:
Shifting this to 1.4 release. @tomspur any thoughts on the current status here?

psi4/__init__.py
tomspur: I wanted to be able to import the broker below anyway so that it is possible to test parts of the infrastructure in the testsuite. Would it be possible to keep the `from .broker import broker` line due to that?

loriab: Would it be possible to move broker into driver/ ? That way it’ll be installed automatically (as all .py files are) and you can add the from .broker import broker to psi4/driver/__init__.py

tomspur: Yes that works fine. Just pushed as well

tomspur:
@dgasmith I think this is ready for review - there is also one test that tests the minimal functionality of the Broker.
I have more things to implement once the amplitudes are available from python as well and could add another PR for that once that is in as well

psi4/driver/broker.py
```diff
@@ -0,0 +1,159 @@
...
+
+        When bypass_scf=True a hf energy calculation has been done before.
+        """
+        start = time.time()
+        self.grd = psi4.gradient(LOT, bypass_scf=bypass_scf)
+        time_needed = time.time() - start
+        self.timing[LOT] = self.timing.get(LOT, []) + [time_needed]
+
+
+def broker(serverdata=False, options=None):
```
dgasmith: Perhaps `ipi_broker` here. There should be a `molecule=None` area as well which will default back to the active molecule. I would argue instantiation as `ipi_broker("CCSD", molecule=mol)` would be more canonical Psi4, but I am not sure to the extent that I-PI specifies these interfaces.

psi4/driver/broker.py
```diff
@@ -0,0 +1,159 @@
...
+    from ipi.interfaces.sockets import Client
+    ipi_available = True
+except ImportError:
+    ipi_available = False
+    # Define Client to enable testing of the Broker in the unittests
+    class Client(object): pass
+
+
+class Broker(Client):
+    def __init__(self, options, genbas=None, serverdata=False):
```
dgasmith: Perhaps this should be `IPIBroker` or similar and a molecule should be passed in here.

psi4/driver/broker.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def calculate_force(self, pos):
+        """Fetch force, energy of PSI.
+        """
+        if len(pos.shape) == 1:
+            pos = pos.reshape((-1, 3))
+        self.frc, self.pot = self.callback(pos)
+        return self.frc, self.pot
+
+    def get_molecule(self, pos):
```
dgasmith: If you pass in a Psi4 molecule perhaps a better way to instantiate this will be `molclone = self.mol.clone(); mol clone.set_geometry(psi4.core.Matrix.from_array(pos))`. In this case all of the multiplicity, symmetry, etc will be carried over automatically.

psi4/driver/broker.py
```diff
@@ -0,0 +1,159 @@
...
+from numpy.testing import assert_array_almost_equal as assert_equal
+
+import psi4
+try:
+    from ipi.interfaces.sockets import Client
+    ipi_available = True
+except ImportError:
+    ipi_available = False
+    # Define Client to enable testing of the Broker in the unittests
+    class Client(object): pass
```
dgasmith: `class Client: pass` for Py 3+

dgasmith:
I think this is pretty close pending the review comments, any chance we can get this patched up and in?

psi4/driver/broker.py
tomspur: This function translates the molecule to the array specification, so directly passing it should work as well. I added it as you suggested here.

psi4/driver/broker.py
tomspur: Hmm, maybe `get_molecule` is a bit misleading as a function name:
I would like to have the final string representation in cartesian coordinates with c1 symmetry that can be used to calculate the force. The reason is, that the molecule will mostly have no symmetry anyway due to the dynamics and this will simplify other algorithms to speed up the force calculation by reusing previous orbitals without changing symmetry/charge/multiplicity.

Would you prefer to name this function like `get_geometry` instead?

psi4/driver/broker.py
```diff
@@ -0,0 +1,159 @@
...
+            print("FORCE:")
+            frc, pot = b.calculate_force(atoms)
+            print(frc, pot)
+
+            atoms *= -1.0
+            frc2, pot2 = b.calculate_force(atoms)
+            print("FORCE MIRROR:")
+            print(frc2, pot2)
+            assert_equal(pot, pot2)
+            assert_equal(frc, -1.0*frc2)
```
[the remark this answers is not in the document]
tomspur: I moved the testing to the testsuite

tomspur:
Sorry for the delay .. Most of the review comments are fixed now

psi4/driver/broker.py
loriab: do you need the string repr for any other reason than to pass into `psi4.geometry(` to create a new active molecule?

if you need the string, an easier way is `molinstance.to_string(dtype='xyz', units='bohr')`. if you're after an actual `Molecule` with the new geometry, @dgasmith's suggestion is the way to go. Has the parent molecule already had c1/nocom/noreorient set?

susilehtola:
I assume this is also valid for DFT? If yes, it might be a good thing to also put in a DFT test case.

tomspur:
> I assume this is also valid for DFT? If yes, it might be a good thing to also put in a DFT test case.

I added a further test that uses `bp86-d2` from another test.

tomspur:
It is unclear to me why the new ipi2 test fails as I get a slightly different result for the final gradient:
Gradient on travis:
```
  -Total Gradient:
     Atom            X                  Y                   Z
    ------   -----------------  -----------------  -----------------
       1        0.000471409888    -0.006768777549    -0.000000000000
       2        0.000447921170    -0.006988127109     0.000000000000
       3       -0.000919134197     0.013753786479     0.000000000000
```
Local gradient:
```
  -Total Gradient:
     Atom            X                  Y                   Z
    ------   -----------------  -----------------  -----------------
       1        0.000471373615    -0.006768232833    -0.000000000005
       2        0.000447935618    -0.006988074312     0.000000000003
       3       -0.000919106219     0.013753539258     0.000000000001
```

If this variation is expected, I could loose the `compare_array` criterion from 7 digits to e.g. 6?

dgasmith:
7 is fairly tight unless your energy convergence thresholds are commensurately tight. I think we usually test in the 5-6 range... something we have talked about standardizing, but have yet to do so.

loriab:
Also, the Bragg–Slater radii changed recently (tag `1.4a1`) and many DFT refs had to change slightly.

tomspur:
> 7 is fairly tight unless your energy convergence thresholds are commensurately tight. I think we usually test in the 5-6 range... something we have talked about standardizing, but have yet to do so.

I used 6 now and seems to work fine with the old ref values

psi4/driver/broker.py
```diff
@@ -0,0 +1,145 @@
...
+
+    try:
+        if b.serverdata:
+            b.run()
+        else:
+            return b
+
+    except KeyboardInterrupt:
+        psi4.core.print_out("Killing IPIBroker\n")
+        b.__del__()
```
dgasmith: Usually write this as `del b` and no need to call `sys.exit()` here.

tests/ipi2/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(ipi2 "psi;quicktests;smoketests;cc")
```
loriab: external deps tests (e.g., dftd3) usually go in a separate subdir of tests. if you don't want to, I'll reposition the tests after merge. in that case, this is just a heads-up for when you notice the tests aren't where you left them.

psi4/driver/broker.py
```diff
@@ -0,0 +1,145 @@
...
+    ipi_available = True
+except ImportError:
+    ipi_available = False
+    # Define Client to enable testing of the Broker in the unittests
+    class Client(): pass
+
+
+class IPIBroker(Client):
+    def __init__(self, options, genbas=None, serverdata=False,
+                 molecule=None):
```
loriab: (a) does cfour's genbas make an appearance?

(b) please run `yapf` over this file

psi4/driver/broker.py
```diff
@@ -0,0 +1,145 @@
...
+            super(IPIBroker, self).__init__(address=address, port=port,
+                                      mode=mode)
+        else:
+            super(IPIBroker, self).__init__(_socket=False)
+        self.options = options
+
+        if molecule is None:
+            molecule = psi4.core.get_active_molecule()
+        self.initial_molecule = molecule
+        assert self.initial_molecule.orientation_fixed() == True, "Orientation must be fixed!"
```
loriab: note that you can do things like https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1442-L1446 to keep the c1 and orientation that the code needs without cooperation from the user.

psi4/driver/broker.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Arguments:
+        - pos: positions of the atoms as array.
+        """
+
+        self.initial_molecule.set_geometry(psi4.core.Matrix.from_array(pos))
+
+        self.calculate_gradient(self.options["LOT"], **kwargs)
+
+        self.pot = psi4.core.scalar_variable('CURRENT ENERGY')
```
loriab: `psi4.variable` will also do

dgasmith:
@loriab Can we get a final review here?

tests/ipi2/CMakeLists.txt
tomspur: I put both tests into a subfolder `ipi`. If you prefer another location, feel free to reposition them as you wish. Thank you for the heads-up

psi4/driver/broker.py
tomspur: (a) I dropped genbas, which was a leftover from a previous version with cfour indeed.

(b) done, with not that many changes though

psi4/driver/broker.py
tomspur: I had to call `__del__` explicitly to close the port immediately. with `del b` it seems to be done on the next garbage collection cycle, which does not always seem to happen on termination and the port was still in use after termination. Not sure if this is still required though ..

tests/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ foreach(test_name adc1 adc2 casscf-fzc-sp casscf-semi casscf-sa-sp ao-casscf-sp
                   fd-freq-energy fd-freq-energy-large fd-freq-gradient
                   fd-freq-gradient-large fd-gradient freq-isotope1 freq-isotope2 fnocc1 fnocc2
                   fnocc3 fnocc4 frac frac-ip-fitting frac-traverse ghosts gibbs matrix1
+                  ipi
```
loriab: This is the wrong place -- instead needs to look like https://github.com/psi4/psi4/pull/1057/files#diff-3dea97423f83302e647a3bab62bf1157L267-L270 for the tests to be added. This can be the sole appearance for an `ENABLE_ipi` cmake var. What I'll do is move your two tests from ctest to tests/pytests/ . Pytest can easily check if ipi is available in the current env, whereas this ctest mechanism is looking at the env at configuration time.

tomspur: This took me longer than expected .. Now it should be at the correct place in the pytests location. Is this the correct location? The tests should work no matter if ipi is available or not so that I did not use the new variable `using_ipi` in the pytests.

The current tests mainly check the interaction with the rest of psi4 to see if the variables are still correctly passed on and all calculations are as expected. Those will be passed onto ipi, if it is installed as well and executed with correct serverdata.

dgasmith:
@loriab I think this can be merged. Please give a once over based on your comments.

tests/pytests/test_addons.py
```diff
@@ -99,6 +101,110 @@ def test_gdma():
...
+    })
+
+    options = {
+        "LOT"                   :   "ccsd",
+        "charge"                :   0,
+        "multiplicity"          :   1,
+    }
+
+    #ipi_broker(serverdata="inet:localhost:21340", options=options)
+    b = ipi_broker(serverdata=False, options=options)
```
loriab, line 129: Would it be violating any of the conventions i-pi users are familiar with to convert to something like https://github.com/psi4/psi4/pull/1676/commits/403e742cab2fb8cd358d09df8e5e44c31aa7f97d#diff-c4d6e39b37f0195e3587302260658d00R504-R506 so that `energy(LOT, use_ipi=True, molecule=molecule)` still with an initialization step like https://github.com/psi4/psi4/pull/1676/files#diff-0fc31e8d2012cf40cdddc336b0b899ecR25 ?

tests/pytests/test_addons.py
```diff
@@ -99,6 +101,110 @@ def test_gdma():
...
+        'dft_radial_points':    99,
+        'dft_spherical_points': 302,
+        'e_convergence':        8,
+        'd_convergence':        8
+    })
+
+    options = {
+        "LOT"                   :   "bp86-d2",
+        "charge"                :   0,
+        "multiplicity"          :   1,
```
loriab, line 199: I couldn't find mention of LOT, charge, multiplicity in i-pi docs. Are these options i-pi users are familiar with? If not, can these be extracted from `molecule.molecular_charge()`, `molecule.multiplicity()`, and `energy(LOT)` so that there aren't conflicting ways to set the same information?

loriab:
@tomspur @andysim @dgasmith 

I added another commit to this that largely does away with the options dict passed to `ipi_broker`. chg/mult weren't being used and should be read off the molecule anyway to avoid input clash. LOT is now a req'd option so the fn looks a lot more like `energy`. (It's returning a broker object, so I didn't work it into `energy()` itself.) Any concerns, particularly if I'm impairing its intended usage?

Also, at present any calls to `ipi_broker` with `import ipi` unavailable run just fine as ordinary psi4 calls. I was a little surprised by this compensating for user's expressed intent. Is this the wanted behavior?

If there aren't actually any wanted changes, I have another commit to make for lgtm and to remove the `@using_ipi` that aren't strictly needed, then it's rtg.

tests/pytests/test_addons.py
tomspur: Yes, getting them from the molecule directly is perfectly fine as well. Thank you for dropping it from this PR

tests/pytests/test_addons.py
tomspur: That would be fine too, I guess. I'm just not familiar enough with psi to directly implement it like this. Calling `energy(LOT, use_ipi=True, molecule=molecule)` along an initialization to provide the serverdata (see an example above commented out `serverdata="inet:localhost:21340"` to listen on the specific network interface/port) can be also used.
Where would the documentation belong into psi to ease the usage of the above convention from psi?

tests/pytests/test_addons.py
loriab: gtk, thanks. I didn’t find that chgmult were actually being used or sent to ipi, so lmk if a hookup is needed and where.

tomspur:
Is it possible to use `options=None` in the class definition? Then the options are mostly not required at all in the current commits above.

Yes, I implemented the Client so that you always can calculate the current energy/force of the initialised molecule, which enables testing of the passed on configuration. When ipi is available, it will also be used for communicating back to the driver.

tests/pytests/test_addons.py
loriab: I think I'll skip the `energy(..., ipi=True)` interface until there's an accustomed ipi user to plan with. Not following the "documentation belong into psi to ease" comment.

I'm content with the last 9ee06b1 commit. That'll fail the 2nd Azure lane until a fix for #1943 is in. After that, I'll rebase, then **force push** (so don't pull your branch again). CI will pass, then ready for merge.

PR #1061, opened 2018-06-28
Expose wavefunction and amplitudes to python

tomspur:
## Description
The purpose of this PR is to expose the wavefunction as well as access to all amplitudes from CC to the python side. This has the advantage to sort out the relationship of Wavefunction, CCEnergyWavefunction and CCLambdaWavefunction in a more natural C++ way through inheritance.

## Todos
* **Developer Interest**
  - [x] CCLambdaWavefunction inherits from CCEnergyWavefunction to facilitate better runtime access from the python side.
  - [x] Setter and getter for all T & L amplitudes are included in the python interface to enable more sophisticated methods in the future, e.g. in the broker from #1057.
  - [x] Setter and getter for all T & L amplitudes are included in the python interface to enable more sophisticated methods in the future, e.g. in the broker from #1057.

## Questions
- [ ] Does the more technical details from above (inheritance change and exposing of the amplitudes to python) require more text for the release notes?
- [ ] The python documentation should be automatically built, is there more documentation for this part required?

## Checklist
- [x] Tests added for any new features (not required as no new functionality added)

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/ccenergy/ccenergy.cc
+++ b/psi4/src/psi4/ccenergy/ccenergy.cc
@@ -330,7 +330,7 @@ double CCEnergyWavefunction::compute_energy()
             outfile->Printf( "\n");
             amp_write();
 
-			// Get T amplitudes
+            // Get T amplitudes
             global_dpd_->file2_init(&t1, PSIF_CC_OEI, 0, 0, 1, "tIA");
             SharedMatrix new_t1 = SharedMatrix(new Matrix(&t1));
             set_T1(new_t1);
--- a/psi4/src/psi4/ccenergy/ccenergy.cc
+++ b/psi4/src/psi4/ccenergy/ccenergy.cc
@@ -96,8 +96,6 @@ double CCEnergyWavefunction::compute_energy()
     double **geom, *zvals, value;
     FILE *efile;
     int **cachelist, *cachefiles;
-    dpdfile2 t1;
-    dpdbuf4 t2;
     double *emp2_aa, *emp2_ab, *ecc_aa, *ecc_ab, tval;
 
     moinfo_.iter=0;
@@ -329,16 +327,6 @@ double CCEnergyWavefunction::compute_energy()
 
             outfile->Printf( "\n");
             amp_write();
-
-            // Get T amplitudes
-            global_dpd_->file2_init(&t1, PSIF_CC_OEI, 0, 0, 1, "tIA");
-            SharedMatrix new_t1 = SharedMatrix(new Matrix(&t1));
-            set_T1(new_t1);
-            global_dpd_->file2_close(&t1);
-            global_dpd_->buf4_init(&t2, PSIF_CC_TAMPS, 0, 0, 5, 0, 5, 0, "tIjAb");
-            SharedMatrix new_t2 = SharedMatrix(new Matrix(&t2));
-            set_T2(new_t2);
-            global_dpd_->buf4_close(&t2);
 
             if (params_.analyze != 0) analyze();
             break;
@@ -663,11 +651,21 @@ void CCEnergyWavefunction::one_step(void) {
     return;
 }
 
-SharedMatrix CCEnergyWavefunction::T1() const {
+SharedMatrix CCEnergyWavefunction::T1() {
+    dpdfile2 t1;
+    global_dpd_->file2_init(&t1, PSIF_CC_OEI, 0, 0, 1, "tIA");
+    SharedMatrix new_t1 = SharedMatrix(new Matrix(&t1));
+    T1_ = new_t1;
+    global_dpd_->file2_close(&t1);
     return T1_;
 }
 
-SharedMatrix CCEnergyWavefunction::T2() const {
+SharedMatrix CCEnergyWavefunction::T2() {
+    dpdbuf4 t2;
+    global_dpd_->buf4_init(&t2, PSIF_CC_TAMPS, 0, 0, 5, 0, 5, 0, "tIjAb");
+    SharedMatrix new_t2 = SharedMatrix(new Matrix(&t2));
+    T2_ = new_t2;
+    global_dpd_->buf4_close(&t2);
     return T2_;
 }
 
--- a/psi4/src/psi4/ccenergy/ccenergy.cc
+++ b/psi4/src/psi4/ccenergy/ccenergy.cc
@@ -43,6 +43,7 @@
 #include "psi4/libdpd/dpd.h"
 #include "psi4/libqt/qt.h"
 #include "psi4/libmints/wavefunction.h"
+#include "psi4/libmints/matrix.h"
 #include "psi4/libpsio/psio.h"
 #include "psi4/libpsio/psio.hpp"
 #include "psi4/libpsi4util/PsiOutStream.h"
@@ -328,6 +329,17 @@ double CCEnergyWavefunction::compute_energy()
 
             outfile->Printf( "\n");
             amp_write();
+
+			// Get T amplitudes
+            global_dpd_->file2_init(&t1, PSIF_CC_OEI, 0, 0, 1, "tIA");
+            SharedMatrix new_t1 = SharedMatrix(new Matrix(&t1));
+            set_T1(new_t1);
+            global_dpd_->file2_close(&t1);
+            global_dpd_->buf4_init(&t2, PSIF_CC_TAMPS, 0, 0, 5, 0, 5, 0, "tIjAb");
+            SharedMatrix new_t2 = SharedMatrix(new Matrix(&t2));
+            set_T2(new_t2);
+            global_dpd_->buf4_close(&t2);
+
             if (params_.analyze != 0) analyze();
             break;
         }
@@ -649,6 +661,22 @@ void CCEnergyWavefunction::one_step(void) {
         }
     }
     return;
+}
+
+SharedMatrix CCEnergyWavefunction::T1() const {
+    return T1_;
+}
+
+SharedMatrix CCEnergyWavefunction::T2() const {
+    return T2_;
+}
+
+void CCEnergyWavefunction::set_T1(SharedMatrix& T1_new) {
+    T1_ = T1_new;
+}
+
+void CCEnergyWavefunction::set_T2(SharedMatrix& T2_new) {
+    T2_ = T2_new;
 }
 
 }} // namespace psi::ccenergy
--- a/psi4/src/psi4/cclambda/cclambda.h
+++ b/psi4/src/psi4/cclambda/cclambda.h
@@ -48,9 +48,9 @@ public:
     double compute_energy();
 
     /// Returns the L1 amplitudes
-    virtual SharedMatrix L1() const;
+    virtual SharedMatrix L1();
     /// Returns the L2 amplitudes
-    virtual SharedMatrix L2() const;
+    virtual SharedMatrix L2();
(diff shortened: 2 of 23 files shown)


psi4/src/psi4/ccenergy/ccenergy.cc
```diff
@@ -331,6 +332,17 @@ double CCEnergyWavefunction::compute_energy()
...
+
+			// Get T amplitudes
+            global_dpd_->file2_init(&t1, PSIF_CC_OEI, 0, 0, 1, "tIA");
+            SharedMatrix new_t1 = SharedMatrix(new Matrix(&t1));
+            set_T1(new_t1);
+            global_dpd_->file2_close(&t1);
+            global_dpd_->buf4_init(&t2, PSIF_CC_TAMPS, 0, 0, 5, 0, 5, 0, "tIjAb");
+            SharedMatrix new_t2 = SharedMatrix(new Matrix(&t2));
+            set_T2(new_t2);
+            global_dpd_->buf4_close(&t2);
```
dgasmith: If I understand this correctly this will set T1/T2 in memory for every run which isn't feasible for large computations. This will need to be an option where a CCWavefunction optionally reads this into memory on demand.

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -403,6 +412,20 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction>
     /// Returns the core Hamiltonian matrix
     SharedMatrix H() const { return H_; }
 
+    /// Sets the alpha electrons MO coefficients
+    void set_Ca(SharedMatrix& Ca_new);
```
dgasmith: I think we have explicitly avoided `set_Ca` to prevent confusion and some possible errors. Is there a reason that `wfn->Ca()->clone(new_Ca)` will not work for you as a way of override?

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -403,6 +412,20 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction>
...
+    /// Sets the beta electrons MO coefficients
+    void set_Cb(SharedMatrix& Cb_new);
+    /// Sets the T1 amplitudes
+    virtual void set_T1(SharedMatrix& T1_new) { throw PSIEXCEPTION("There are no T1 amplitudes for this wavefunction."); };
```
dgasmith: I dont think should go on the base Wavefunction. Please either 1) add them to the CC* Wavefunctions or 2) (recommend) use the `get/set_array` values on the wavefunction itself which can set/get arbitrary arrays.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -2828,6 +2851,41 @@ void Matrix::write_to_dpdfile2(dpdfile2 *outFile) {
     global_dpd_->file2_mat_close(outFile);
 }
 
+void Matrix::write_to_dpdbuf4(dpdbuf4 *outBuf)
```
dgasmith: Cool, these are very useful.

psi4/src/psi4/ccenergy/ccenergy.cc
tomspur: Hmm, how would you call such an option? Maybe it would also work to set this dynamically in memory, when it is accessed via the getter?

psi4/src/psi4/libmints/wavefunction.h
tomspur: How would that work from the python side? This does not seem to work for me:
    wfn = psi4.core.legacy_wavefunction()
    new_ca = wfn.Ca()
Neither `new_ca` nor `wfn.Ca()` has a clone function which accepts another matrix to clone from. This was the reason for adding the setter functions, which are then exposed to python as well, so that it can be set from there. Is there another way to do this?

psi4/src/psi4/libmints/wavefunction.h
tomspur: Ah, right, they can also be set according to 1) this way in `export_wavefunction.cc`. I thought that it is only possible to expose there on the base Wavefunction. :+1: for adding them to the CC*Wavefunctions where needed.

I'm not sure, I understand the `get/set_array` machinery fully: Would 2) mean, that T1/T2 setting in memory would set it in the `array_`  map instead of a direct property of the wavefunctions?
That should work as well, yet it seems to be not heavily used yet in other parts in the code (e.g. how can you access Ca/Cb with this approach), which is why I would prefer option 1) above.

psi4/src/psi4/libmints/wavefunction.h
dgasmith: Sorry `->copy(new_Ca)`.

psi4/src/psi4/ccenergy/ccenergy.cc
dgasmith: You might want to consider an option does not delete T1/T2 and then a function that pulls this from memory `SharedMatrix get_T2()`. The CC guys could comment more.

amjames: I agree with @dgasmith T1 would probably be fine, but T2 will be too large to store in core for many use cases.

Personally I think a good solution would be to leave the `T1_` and `T2_` `SharedMatrix` objects uninitialized until some function is explicitly called by the user to populate them. If the functions returning T1/T2 are called before T1/T2 are initialized they could throw, or populate at that time. 

To address @dgasmith 's statement above the only situation where T1/T2 would be deleted on disk is if they were overwritten by performing a new calculation, or if all of the scratch files are removed such as by calling `clean()`.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -237,6 +237,29 @@ Matrix::Matrix(dpdfile2 *inFile)
     global_dpd_->file2_mat_close(inFile);
 }
 
+Matrix::Matrix(dpdbuf4 *inBuf)
+    : name_("None"), rowspi_(inBuf->params->nirreps), colspi_(inBuf->params->nirreps)
+{
+    matrix_ = NULL;
+    symmetry_ = 0; //inBuf->my_irrep;
```
amjames: Is there a reason you set `symmetry_` to 0? The commented `inbuf->my_irrep` should be the correct value.

psi4/src/psi4/cclambda/cclambda.h
```diff
@@ -38,16 +39,46 @@ class Options;
 
 namespace psi { namespace cclambda {
 
-class CCLambdaWavefunction : public Wavefunction
+class CCLambdaWavefunction : public psi::ccenergy::CCEnergyWavefunction
```
robertodr: Could you mark the `CCLambdaWavefunction` as `final`?

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -141,6 +141,10 @@ void Wavefunction::shallow_copy(const Wavefunction *other) {
     Db_ = other->Db_;
     Fa_ = other->Fa_;
     Fb_ = other->Fb_;
+    T1_ = other->T1_;
```
robertodr: Is it strictly necessary to have Ts and Ls as data members of the base class?

tomspur:
Yes, I do. I am trying to move the initialization of the amplitudes inside T1/T2, but currently getting a segmentation fault. Those functions are now not const anymore, might that be the source of this problem?

dgasmith:
Can you explain a bit more of the issue? From discussions it seems that we will need to have a `get/set T1/T2` that reads/writes the tensors from disk rather than holding them in memory.

tomspur:
I moved the code to get the amplitudes to their getters that should read it from disk in 3fdf1741
It seems reading the file does not work in this location, and I'm unclear why this is the case. I'd work on the other requested changes above after fixing this issue...

dgasmith:
@robertodr @amjames Any thoughts?

amjames:
@tomspur I had a feeling you would run into something like this eventually. I would suspect that you are getting a segfault when calling these functions because the `global_dpd_` object has been broken down.

When `CCEnergyWavefunction::compute_energy` is called the `DPD` object that `global_dpd_` is pointing to is constructed  [(example for UHF)](https://github.com/psi4/psi4/blob/3fdf17415282884ec86e78ba34087ddd6c3645e8/psi4/src/psi4/ccenergy/ccenergy.cc#L122-L131) before iterations begin, and then [deleted](https://github.com/psi4/psi4/blob/3fdf17415282884ec86e78ba34087ddd6c3645e8/psi4/src/psi4/ccenergy/ccenergy.cc#L504) before the function returns. 

So if when you do something like this:

```python
e, wfn = psi4.energy('ccsd', return_wfn=True)
t2 = wfn.T2()
```
At the time when the python bound function is called and the code for `CCEenergyWavefunction::T2()`  is executed the `global_dpd_` the dpd object that the `global_dpd_` points to has been deleted out from under it (you can see [here](https://github.com/psi4/psi4/blob/3fdf17415282884ec86e78ba34087ddd6c3645e8/psi4/src/psi4/libdpd/init.cc#L59) that `dpd_set_default` will assign the `global_dpd_` pointer to `dpd_list[0]`, and below it is the code for `dpd_close` that will delete the object.

I see two possible solutions.  

 - The easiest to implement would be to have the getters/setters check that `global_dpd_` is not `nullptr`, and if it is construct the dpd. Before doing the geting/setting. The function would also have to free the dpd object as well as other modules will assume that `global_dpd_` is null until they initialize it and `dpd_init` will throw when they try to create it.

 - The other option is to simply start phasing out the `global_dpd_` entirely. A data member can be added to the `CCEnergyWavefunction` and the member functions of the class can be modified to use that data member instead of the `global_dpd_`. The initialization of the data member is moved into the construtor and the dpd will be valid until the wavefunction is destroyed. The same would need to be done for the cclambda wavefunction as well.

amjames:
@tomspur Have you gotten a chance to look at my last comments or begun work on any of those changes? @robertodr and myself are prepared to complete the work needed to get this feature working as it would be a great step forward for improving the api of the CC codes. 

We could submit PRs to your working branch if you are able to merge, you could give us access to push directly to this branch, or a third option if you don't plan on working on this any further we could open a new PR (retaining the commits you have already authored of course) to complete the feature. 

Please let us know how you would prefer we proceed, thanks!

tomspur:
> @tomspur Have you gotten a chance to look at my last comments or begun work on any of those changes? @robertodr and myself are prepared to complete the work needed to get this feature working as it would be a great step forward for improving the api of the CC codes.

I had ineed a look into your comments and tried to understand the infrastructure around the global dpd... Unfortunately, It would be quite difficult for me to properly fix the above comments fully...

> We could submit PRs to your working branch if you are able to merge, you could give us access to push directly to this branch, or a third option if you don't plan on working on this any further we could open a new PR (retaining the commits you have already authored of course) to complete the feature.
> 
> Please let us know how you would prefer we proceed, thanks!

I added you two directly access to push to my repository at tomspur/psi4release, so feel free to add anything you like there. Please let me know, how else I could help to move this feature forward.
I just rebased my branch against the current master and hope this helps to get this included

dgasmith:
Shifting this to the 1.4 release at earliest. @tomspur @amjames @robertodr Should this remain open or is it superseded by other PR's?

dgasmith:
Closed by #1344.

PR #1068, opened 2018-07-03
Oeprop printing

### LINKED ISSUE #1055: OEProp direct call printing
I have seen folks try the following several times:

```
e, wfn = energy('MP2', return_wfn=True)
oeprop(wfn)
```

which will return the HF properties as MP2 densities are costly to compute, the same is true with CCSD and others. To help alleviate confusion I would add a few lines [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/oeprop.cc#L866) in the OEProp compute call that would highlight the name of the density matrix. A line like the following could be appropriate:

```
if (title_ == ""){
    outfile->Printf("OEProp: No title given, name of density matrix used for the following properties is %s\n", Da_so_->name().c_str());
}
```

This should give folks some hint that things may not be as they appear.

mlestep:
## Description
Makes print statement more clearly communicate to user which density matrix is being employed. c closes #1055 
## Todos
* **Developer Interest**
  - [ ] Feature1
  - [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/oeprop.cc
+++ b/psi4/src/psi4/libmints/oeprop.cc
@@ -865,7 +865,7 @@ void OEProp::compute()
 
     
     if (title_ == "") {
-        outfile->Printf("OEProp: No title given, name of density matrix used for the following properties is \'%s\'\n", Da_so_->name().c_str());
+        outfile->Printf("OEProp: No title given, name of density matrix used for the following properties is '%s'\n", Da_so_->name().c_str());
     } else {
         outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
     }
--- a/psi4/src/psi4/libmints/oeprop.cc
+++ b/psi4/src/psi4/libmints/oeprop.cc
@@ -863,11 +863,12 @@ bool from_string(T& t,
 void OEProp::compute()
 {
 
-    outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
     
-    if (title_ == ""){
+    if (title_ == "")
         outfile->Printf("OEProp: No title given, name of density matrix used for the following properties is %s\n", Da_so_->name().c_str());
-    }
+    
+    else
+        outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
 
     // Search for multipole strings, which are handled separately
     std::set<std::string>::const_iterator iter = tasks_.begin();
--- a/psi4/src/psi4/libmints/oeprop.cc
+++ b/psi4/src/psi4/libmints/oeprop.cc
@@ -864,6 +864,10 @@ void OEProp::compute()
 {
 
     outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
+    
+    if (title_ == ""){
+        outfile->Printf("OEProp: No title given, name of density matrix used for the following properties is %s\n", Da_so_->name().c_str());
+    }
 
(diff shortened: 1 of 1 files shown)


mlestep:
I have now, and unfortunately it doesn't. I'll look into it now.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -863,7 +863,12 @@ bool from_string(T& t,
 void OEProp::compute()
 {
 
-    outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
+    
+    if (title_ == "")
```
dgasmith: Please use brackets for if/else statements here.

mlestep: Thanks, will do. Is there a particular reason the if statements on lines 882-907 don't use brackets?

bgpeyton: Could use `strcmp()` and just compare to `title_.c_str()` directly

dgasmith: In general please avoid C functions, the current code is fine. There is something else missing here.

dgasmith: Legacy semantics, trying to adhere a bit more to style guides these days.

bgpeyton: Will keep that in mind, thanks!

mlestep: Makes sense, thanks!

loriab:
Good news: it works fine for me. The patch below (n/c from @mlestep, I don't think) gives the output below (it just needed to be told to compute something).
```
-
-    outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
+    if (title_ == "") {
+        outfile->Printf("OEProp: No title given, name of density matrix used for the following properties is %s\n", Da_so_->name().c_str());
+    } else {
+        outfile->Printf( "\nProperties computed using the %s density matrix\n\n", title_.c_str());
+    }
```
```
e, wfn = energy('MP2', return_wfn=True)
oeprop(wfn, "MULTIPOLES(1)")
```
```
Properties will be evaluated at   0.000000,   0.000000,   0.000000 [a0]
OEProp: No title given, name of density matrix used for the following properties is SCF density

 Multipole Moments:

 ------------------------------------------------------------------------------------
     Multipole             Electric (a.u.)       Nuclear  (a.u.)        Total (a.u.)
 ------------------------------------------------------------------------------------

 L = 1.  Multiply by 2.5417462300 to convert to Debye
 Dipole X            :          0.0000000            0.0000000            0.0000000
 Dipole Y            :          0.0000000            0.0000000            0.0000000
 Dipole Z            :         -0.2704549            1.1770271            0.9065722

 --------------------------------------------------------------------------------
```

Bad news: we [lost a test](https://github.com/psi4/psi4/commit/71bf174e60bc54b6a7424ce13dd28cef264fa850) (one of the two cc properties ones) and it fails. @amjames or TDC group, could you take a look?

dgasmith:
Small tidbit, but can we use quotes here: `following properties is 'SCF density'.`

loriab:
@mlestep, I think just add in the brackets `{}` and the single quotes around `'%s'`, and this is good to go. The cc46 is a separate issue.

PR #1080, opened 2018-07-06
Generalized Davidson Solver

rdcunha:
## Description
Adds a generalized Davidson solver to solvers.py. Issue #886. @dgasmith @amjames 

## Todos
* **Developer Interest**
  - [x] General Davidson solver in python, takes an arbitrary number of roots, the convergence tolerance and maximum number of guess vectors

## Questions
- [ ] Should the solver do all operations with psi4.core.Matrix functions? Currently, I’m using a QR decomposition in numpy to orthogonalize the added guess vectors with respect to the previous guess vectors.

- [ ] I also need to change the code to handle a list of psi4.core.Matrix objects. Can I assume that this would be a list of guess vectors, with their second dimension being 1?

- [ ] Since this is a general solver, should I be worrying about irreps?

## Checklist
- [ ] Currently takes in and returns numpy arrays, needs to take in and return a list of Psi4.core.Matrix objects
- [ ] May implement RHF stability analysis as a test case

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/solvers.py
+++ b/psi4/driver/p4util/solvers.py
@@ -310,3 +310,159 @@ class DIIS(object):
             out.axpy(c, self.state[num])
 
         return out
+
+
+# Generalized Davidson solver
+
+def davidson_solver(ax_function, preconditioner, guess, e_conv=1.0E-8,
+        r_conv=None, no_eigs=1, max_vecs_per_root=20, maxiter=100):
+    """
+    Solves for the lowest few eigenvalues and eigenvectors of the given real symmetric matrix
+
+    Parameters
+    -----------
+    ax_function : function
+        Takes in a guess vector and returns the Matrix-vector product.
+    preconditioner : function
+        Takes in a list of :py:class:`~psi4.core.Matrix` objects and a mask of active indices. Returns the preconditioned value.
+    guess : list of :py:class:`~psi4.core.Matrix`
+        Starting vectors, required
+    e_conv : float
+        Convergence tolerance for eigenvalues
+    r_conv : float
+        Convergence tolerance for residual vectors
+    no_eigs : int
+        Number of eigenvalues needed
+    maxiter : int
+        The maximum number of iterations this function will take.
+
+    Returns
+    -----------
+
+    Notes
+    -----------
+
+    Examples
+    -----------
+
+    """
+
+    print("Starting Davidson algo:")
+    print("options:")
+    for k,v in dict(e_conv=e_conv, r_conv=r_conv, no_eigs=no_eigs,
+            max_vecs_per_root = max_vecs_per_root, maxiter=maxiter).items():
+        if v == None:
+            v = 'None'
+        print("{:<25}{:.>100}".format(k,v))
+    if r_conv == None:
+        r_conv = e_conv * 100
+    d_tol = 1.0E-8
+
+    # using the shape of the guess vectors to set the dimension of the matrix
+    N = guess.shape[0]
+
+    #sanity check, guess subspace must be at least equal to number of eigenvalues
+    nli = guess.shape[1]
+    if nli < no_eigs:
+        raise ValueError("Not enough guess vectors provided!")
+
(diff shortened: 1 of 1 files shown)


psi4/driver/p4util/solvers.py
```diff
@@ -310,3 +310,159 @@ def extrapolate(self, out=None):
...
+    """
+
+    print("Starting Davidson algo:")
+    print("options:")
+    for k,v in dict(e_conv=e_conv, r_conv=r_conv, no_eigs=no_eigs,
+            max_vecs_per_root = max_vecs_per_root, maxiter=maxiter).items():
+        if v == None:
+            v = 'None'
+        print("{:<25}{:.>100}".format(k,v))
+    if r_conv == None:
```
loriab: `None` is singleton, so this and `v` above should be `is None:`

psi4/driver/p4util/solvers.py
```diff
@@ -310,3 +310,159 @@ def extrapolate(self, out=None):
...
+# Generalized Davidson solver
+
+def davidson_solver(ax_function, preconditioner, guess, e_conv=1.0E-8,
+        r_conv=None, no_eigs=1, max_vecs_per_root=20, maxiter=100):
+    """
+    Solves for the lowest few eigenvalues and eigenvectors of the given real symmetric matrix
+
+    Parameters
+    -----------
+    ax_function : function
```
JonathonMisiewicz: What's the advantage of having the user pass in a function to perform matrix multiplication, as opposed to having the user pass in the matrix and letting the program do the multiplication?

susilehtola: A CI matrix can be a billion x billion. You don't want to form that. Instead, you contract directly.

susilehtola: Same thing for SCF stability analysis and TD-DFT. Say you have 100 occupied orbitals, and 1000 virtual ones. Then your matrix dimension will be 100*1000 = 100 000 x 100 000. This will take about 80 GB of memory, whereas even 20 trial vectors only take 16 MB.

psi4/driver/p4util/solvers.py
```diff
@@ -310,3 +310,159 @@ def extrapolate(self, out=None):
...
+
+    """
+
+    print("Starting Davidson algo:")
+    print("options:")
+    for k,v in dict(e_conv=e_conv, r_conv=r_conv, no_eigs=no_eigs,
+            max_vecs_per_root = max_vecs_per_root, maxiter=maxiter).items():
+        if v == None:
+            v = 'None'
+        print("{:<25}{:.>100}".format(k,v))
```
dgasmith: Can you have this function print out a header similar to the CG algorithm above? Also please use `core.print_out`.

psi4/driver/p4util/solvers.py
```diff
@@ -310,3 +310,159 @@ def extrapolate(self, out=None):
...
+        # Apply QR decomposition on B to orthogonalize the new vectors wrto all other subspace vectors
+        ## orthogonalize preconditioned residuals against all other vectors in the search subspace
+        B, r = np.linalg.qr(B)
+        nl = B.shape[1]
+
+        print("Davidson: Iter={:<3} nl = {:<4}".format(count, nl))
+        # compute sigma vectors corresponding to the new vectors sigma_i = A B_i
+        sigma = np.zeros_like(B)
+        for i in range(nl):
+            sigma[:,i] = ax_function(B[:,i])
```
dgasmith: It would be good if you allowed the ax_function to compute all of these vectors at once. Quite a bit of savings can usually be had this way. See the above CG algorithm for similar logic.

rdcunha:
@dgasmith just need a little time, I've worked on making the changes, but I still need to add a test. I'll work on it this week.

dgasmith:
This is folded into #1349. Thanks for the hard work on this and very happy it has been used in the upcoming TDDFT module.

PR #1081, opened 2018-07-07
Gauge invariant atomic orbitals (GIAOs) for TD-HF/DFT optical rotation calculations

ashutoshvt:
## Description
This PR attempts to enable calculations of TD-HF/DFT optical rotations (ORs) with GIAOs, also called as London orbitals. This ensures that the length gauge OR results are independent of the choice of origin used in the geometry. 
## Todos
* **Developer Interest**
  - [] adds the gradients of one and two electron integrals of GIAOs
  - [] constructs the modified RHS corresponding to magnetic response vectors. 
  - [] adds a driver for calculating SCF dynamic properties using DIIS
## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


dgasmith:
This would be really fantastic to have in! If I understand correctly we will need both the one and two-electron GIAO integrals. For the two-electron is it possible to convert current libmints tech or will something else need to be implemented?

ashutoshvt:
I am not sure if we need the two electron GIAO integrals for TD-HF/DFT specific rotations. Even if we do need them at some point, I am fairly sure that they could be handled by the current libmints machinery.

jturney:
Let me know if you need help with this!

ashutoshvt:
@dgasmith You were right about the need for two-electron GIAO integrals and the modification of libmints. The one electron magnetic dipole becomes a two electron operator now and I do need to build Fock like pieces for them. I am talking to @jturney about this and it seems like libmints need to be modified for this.

loriab:
Is the what-needs-to-change-and-by-whom clear for this PR?

ashutoshvt:
@loriab I talked to @jturney about the required changes. I hope to start working on this soon.

loriab:
How fares this? Can tests be added (even integral-level ones like https://github.com/psi4/psi4/blob/master/tests/mints3/input.dat if that's where the development is at)?

ashutoshvt:
Hi Lori, I am hoping that this would be ready for review early next week. I hope to calculate Hartree-Fock optical rotations on the psi4numpy side making use of the exported integrals from this PR. I have already exported the 1-electron integrals (and they are correct!) and in the process of calculating the 2-electron GIAO integrals which should be finished soon!

loriab:
New commits! Is this working?

ashutoshvt:
@loriab almost! I am comparing my results to psi3 and only some eri elements are not matching. Will push once I get them to match.

loriab:
Sounds promising! We'd be glad of a status update by Monday morning.

loriab:
near-term prospects?

PR #1097, opened 2018-07-18
enh: adds mem printout to DFHelper

obrien951:
## Todos
* **Developer Interest**
  - [x] Feature1
Added printout for memory availability in DFHelper
## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Main memory for Psi4 is reported in MiB (1024-based; MB is 1000-based). Many modules report 1024-based memory quantities but label them MB. Would be good to global fix this. This PR does something different and reports MB based on 1000 quantities. What should be our standard?

TiborGY:
@loriab I prefer 1024-based quantities, with the correct MiB/GiB label, especially for RAM. Labeling base-2 quantities as MB is just plain wrong. (Windows has the nasty tradition of using the MB/GB label when it actually means MiB/GiB)

PR #1098, opened 2018-07-20
clear filesystem from FISAPT

loriab:
## Description
avoid Win/Linux filesystem conventions by moving selected operations to python

## Todos
* **Developer Interest**
  - [x] Moves fisapt driver py-side
  - [x] Moves writing arrays and filesystem parts of fisapt py-side
  - [x] There were some options with names `sSAPT`. There's machinery that thinks all option names are all-caps, so that wasn't good.
  - [x] Moves filesys in libcubeprop/cubeprop.cc py-side
  - [x] ~Currently can't run cmake multiple times w/o hitting `fsapt/` dir exists~ Now just overwrites existing dirs
  - [x] Adds Molecule docstrings and fixes spelling mistakes

## Questions
- [x] @jturney  what’s this supposed to do when directory already exists? https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfilesystem/path.cc#L53 According to `mkdir` docs, it should throw `EEXIST` error? that’s the behavior i’ve imitated in python. but if that’s the case, how has ctest be working on re-run all this time.
- [ ] Rob was exporting `FISAPT.scalars()` as direct map. I changed to map& to match FISAPTSCF. Rob was never using this in code, but since it was exported to python, possibly scirpts are affected?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/procrouting/sapt/fisapt_proc.py
```diff
@@ -0,0 +1,213 @@
...
+
+    geomfile = filepath + os.sep + 'geom.xyz'
+    xyz = self.molecule().to_string(dtype='xyz', units='Angstrom')
+    with open(geomfile, 'w') as fh:
+        fh.write(xyz)
+
+    self.raw_plot(filepath)
+
+
+def drop(array, filepath):
```
dgasmith: Can these be `_drop` looks like an internal function.

psi4/src/export_misc.cc
```diff
@@ -29,10 +29,13 @@
 #include "psi4/pybind11.h"
 
 #include "psi4/libciomr/libciomr.h"
+#include "psi4/libqt/qt.h"
 
 using namespace psi;
 
 void export_misc(py::module &m) {
+    m.def("timer_on", timer_on, "Start timer with argument as label");
+    m.def("timer_off", timer_off, "Stop timer of label argument");
```
dgasmith: I think this was added in the SAPT(DFT) timer PR? Will cause a conflict.

psi4/src/psi4/fisapt/wrapper.cc
```diff
@@ -1,52 +0,0 @@
-/*
- * @BEGIN LICENSE
```
dgasmith: Cool!

psi4/src/export_misc.cc
loriab: True. It's actually in the scfitertopy PR, three. But all need it. This one has docstrings!

jturney:
@loriab For 
https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfilesystem/path.cc#L53 the code returns `true` if it created the directory and `false` if it didn't. I just looked at the usages of the function in fist and the return value was never checked.

loriab:
Thanks, @jturney, I switched it to `os.makedirs(..., exist_ok=True)`, so it overwrites and `mkdir -p`s.

All, this has had a couple more commits added, been rebased, and is ok for review/merge.

loriab:
This is rebased and fulltests pass locally. There's more on this topic that could be added, but this is a good size for a PR.

PR #1100, opened 2018-07-20
dfh: fix direct_iaq memory estimate in AO_core module

obrien951:
## Todos
* **Developer Interest**
  - [x] fix direct_iaq memory estimate in AO_core module

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/lib3index/dfhelper.cc
+++ b/psi4/src/psi4/lib3index/dfhelper.cc
@@ -214,8 +214,16 @@ void DFHelper::initialize() {
 }
 void DFHelper::AO_core() {
 
-    // total size of sparse AOs
-    size_t required = ( do_wK_ ? 3 * big_skips_[nao_] : big_skips_[nao_]);
+    size_t required;
+
+    if(direct_iaQ_) {
+        // the direct_iaQ method does not use sparse storage
+        // if do_wK added to code, the following will need to be changed to match
+        required = naux_ * nao_ * nao_ ;
+    } else {
+        // total size of sparse AOs
+        required = ( do_wK_ ? 3 * big_skips_[nao_] : big_skips_[nao_]);
+    }
 
     // C_buffers (conservative estimate since I do not have max_nocc TODO)
     required += nthreads_ * nao_ * nao_;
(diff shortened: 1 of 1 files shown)


psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -214,8 +214,16 @@ void DFHelper::initialize() {
...
 void DFHelper::AO_core() {
 
-    // total size of sparse AOs
-    size_t required = ( do_wK_ ? 3 * big_skips_[nao_] : big_skips_[nao_]);
+    size_t required;
+
+    if(direct_iaQ_) {
+        // the direct_iaQ method does not use sparse storage
+        // if do_wK added to code, the following will need to be changed to match
+        required = naux_ * nao_ * nao_ ;
```
dgasmith: @schiebermc Can you review this?

schiebermc: I don't think this changes anything.
Having dense integrals implies that `big_skips_[nao_] == naux_ * nao_ * nao_`.
Is there a reason this was added in? did something break?

loriab: @CDSherrill can explain more fully, but something did break. In a direct_iaq (fisapt) test with a certain amount of memory, it was passing through this memory estimator cleanly, then at AO_core transform, it was failing that memory estimator and throwing an error. With this fix in place, it noticed memory was insufficient for AO_core so switched to disk and completed correctly. Perhaps you can post the triggering input file, @obrien951.

schiebermc: Aha, upon further inspection this is probably the case. I thought that I had switched cutoff_ = 0.0 when direct_iaQ_, but this appears to not be the case. Easily confirmable if the mask sparsity printed to the output is not 0.0%. Therefore big_skips_[nao_] represents the total number of doubles for the sparse form, even when direct_iaQ_.

schiebermc: In fact line 484 uses the correct memory estimate (direct from `naux_`, `nao_`).  I think the issue was that `AO_core()` was added during the JK building push, so I did not consider the dense integrals from `direct_iaQ`.  Probably the discrepancy between this and lines 605-607 causes the throw.  I approve of the changes.

CDSherrill: Thanks for checking this, @schiebermc

Issue #1093, opened 2018-07-14
Density Fitting Incompatible with Frequencies


JonathonMisiewicz:
Any computation that attempts to compute frequencies of a density fitted method will crash unless the symmetry is explicitly set to c1. Gradients do not have this problem. The below input file demonstrates the problem and has been taken from the test cases. The only modifications are changing gradients to frequencies and CCSD(T) to MP2. The bug is there with both methods, and also CCSD. I haven't checked other DF methods.

```
#! DF-MP2 cc-pVDZ energy for the H2O molecule.


molecule h2o {
0 1
o
h 1 0.958
h 1 0.958 2 104.4776 
}

set {
  basis cc-pvdz
  df_basis_scf cc-pvdz-jkfit
  df_basis_cc cc-pvdz-ri
  scf_type df
  guess sad 
  freeze_core true
  cc_type df
  qc_module occ 
}

frequencies('mp2', dertype=0)
```
The crash gives the following error message:
```
Traceback (most recent call last):
  File "/Users/jonathonmisiewicz/psi4/objdir2/stage/usr/local/psi4/bin/psi4", line 264, in <module>
    exec(content)
  File "<string>", line 36, in <module>
  File "/Users/jonathonmisiewicz/psi4/objdir2/stage/usr/local/psi4/lib/psi4/driver/driver.py", line 1774, in frequency
    H, wfn = hessian(lowername, return_wfn=True, molecule=molecule, **kwargs)
  File "/Users/jonathonmisiewicz/psi4/objdir2/stage/usr/local/psi4/lib/psi4/driver/driver.py", line 1595, in hessian
    E, wfn = energy(lowername, return_wfn=True, molecule=moleculeclone, **kwargs)
  File "/Users/jonathonmisiewicz/psi4/objdir2/stage/usr/local/psi4/lib/psi4/driver/driver.py", line 492, in energy
    wfn = procedures['energy'][lowername](lowername, molecule=molecule, **kwargs)
  File "/Users/jonathonmisiewicz/psi4/objdir2/stage/usr/local/psi4/lib/psi4/driver/procrouting/proc.py", line 858, in select_ccsd_t_
    return func(name, **kwargs)
  File "/Users/jonathonmisiewicz/psi4/objdir2/stage/usr/local/psi4/lib/psi4/driver/procrouting/proc.py", line 1608, in run_dfocc
    "RIFIT", core.get_global_option("BASIS"))
[...]

PR #1108, opened 2018-07-26
Make Density Fitting Compatible with Frequencies (#1093 Fix)

JonathonMisiewicz:
## Description
Fixes issue #1093, where the use of density fitting caused a crash when computing frequencies.

As for why these fixes are necessary:

Previously, the clone function would have separate memory addresses for elements of the two lists, which would be fixed by update_geometry assuming that coordinate reinterpretation was allowed. This is precisely what the driver does not allow when getting a findif hessian. The basis set setting code assumed that the memory addresses were the same. This assumption failing led to the originally reported error with setting basis sets. As this seems a rather sane assumption to me, it is Now enforced.

This uncovered a second error when ZMAT coordinates were used: energies at displacements would be off by ~4 hartrees. When the molecule is cloned in the process of getting a c1 wavefunction, I found that the coordinates of the cloned molecules differed from those of the original molecule, right before the basis sets were set. Telling the basis sets that the atoms are not what they actually are seems like the kind of thing that would lead to garbage energies, though I didn't get a detailed mechanism for this. To get the geometries consistent, I had to change the atom cloning procedure.

## Todos
* **Developer Interest**
  - [x] Molecule clone now guarantees that each elements of the atom_ list is identical to some element of the full_atom_ list.
  - [x] Atom clone now guarantees that ZMATs with computed coordinates carry the computed coordinates over to the clone.
  - [x] Fixed a bug where using density fitting for a frequency computation would cause an error unless the symmetry was explicitly set to c1.

## Questions
- [x] Symmetry is used for the purposes of generating displacements, but not for labeling normal modes. Even if we can't take advantage of symmetry for the energetics, we should be able to use symmetry for displacements and normal modes. My hunch is that this is a problem with changing the symmetry of the molecule between the two findif calls, but I'd like Py-side findif to be accepted before looking into this, just in case. Sound good?

## Checklist
- [x] Tests added for any newly functional features
- [x] [All of full tests run, modulo DFT benchmarks and plugins](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge (as soon as the last review is in)

### DIFF
--- a/psi4/src/psi4/libmints/coordentry.h
+++ b/psi4/src/psi4/libmints/coordentry.h
@@ -369,6 +369,8 @@ class ZMatrixEntry : public CoordEntry {
                                                   atoms[ato_->entry_number()], aval_->clone(map),
                                                   atoms[dto_->entry_number()], dval_->clone(map));
         }
+        if (computed_)
+            temp->set_coordinates(coordinates_[0], coordinates_[1], coordinates_[2]);
         return temp;
     }
 };
--- a/psi4/src/psi4/libmints/molecule.cc
+++ b/psi4/src/psi4/libmints/molecule.cc
@@ -219,13 +219,15 @@ Molecule &Molecule::operator=(const Molecule &other) {
 
     // Deep copy the map of variables
     full_atoms_.clear();
+    atoms_.clear();
+    std::shared_ptr<CoordEntry> new_atom;
     std::vector<std::shared_ptr<CoordEntry> >::const_iterator iter = other.full_atoms_.begin();
     for (; iter != other.full_atoms_.end(); ++iter) {
-        full_atoms_.push_back((*iter)->clone(full_atoms_, geometry_variables_));
-    }
-    atoms_.clear();
-    for (iter = other.atoms_.begin(); iter != other.atoms_.end(); ++iter) {
-        atoms_.push_back((*iter)->clone(atoms_, geometry_variables_));
+        new_atom = (*iter)->clone(full_atoms_, geometry_variables_);
+        full_atoms_.push_back(new_atom);
+        // To guarantee every atom is identically a full_atom, add the needed full_atom into atoms
+        // If we lose any atoms this way, we have a bigger problem.
+        if (new_atom->symbol() != "X") atoms_.push_back(new_atom);
     }
 
     // This is called here, so that the atoms list is populated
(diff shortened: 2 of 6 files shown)


loriab:
Good detective work, @JonathonMisiewicz.

@andysim, @jturney, sharing `CoordEntry`s between `atoms_` and `full_atoms_` sounds like it could cause trouble and could undermine `update_geometry` as the one-stop-shop for Molecule self-consistency. Do you anticipate problems with this sol'n?

@JonathonMisiewicz, which `reinterpret_coordentry(False)` in the driver for findif did you have in mind as necessitating this work-around?

JonathonMisiewicz:
@loriab [This line](https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L1382) for hess-by-energies and [this line](https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L1529) for hess-by-grad. The mechanism is the same in both cases. The reinterpretation-forbidden-molecule eventually goes to the density-fitted energy/gradient step. To force the computation to be done in c1, the scf helper gets called in the [energy case](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1591) and in the [gradient case](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1669). Now the two mechanisms unify. The atom_/full_atom_ mirroring gets broken at the [clone step](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1436), which would be re-synced during [update geometry](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1440) if not for that forbidding of reinterpretation.

We might be able to forgive reorienting in the case of hess-from-energies, but I don't see a way for difference of gradients. We need all our gradient vectors to be in the same coordinate system.

jturney:
The sharing of `CoordEntry`s between `atoms_` and `full_atoms_` is what actually happens behind the scenes in `update_geometry` via [reinterpret_coordentries](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/molecule.cc#L1444). This just makes the `clone` function mimic that behavior; I foresee no issues with this change.

I agree, let's get py-findif in first.

Issue #1102, opened 2018-07-23
ECPs, freeze_core and ghosted atoms do not get along


PeterKraus:
When trying to compute counterpoise-corrected energies of complexes with atoms including ECPs, and frozen core applied, the results are incorrect. I believe it is due to incorrect assignment of frozen core orbitals (yet again). See attached input and output files. The relevant part is quoted below

```
$: grep frozen bug.out 
        Use frozen natural orbitals?           no
        Number of frozen core orbitals:         4 (Kr + OCS ghost)
        Number of frozen virtual orbitals:      0
        Use frozen natural orbitals?           no
        Number of frozen core orbitals:         4
        Number of frozen virtual orbitals:      0
        Use frozen natural orbitals?           no
        Number of frozen core orbitals:         3 (Kr ghost + OCS)
        Number of frozen virtual orbitals:      0
        Use frozen natural orbitals?           no
        Number of frozen core orbitals:         3
        Number of frozen virtual orbitals:      0
        Use frozen natural orbitals?           no
        Number of frozen core orbitals:        11 (Kr + OCS)
        Number of frozen virtual orbitals:      0
        Use frozen natural orbitals?           no
        Number of frozen core orbitals:        11
        Number of frozen virtual orbitals:      0
```

For the Kr + OCS(Gh) combination, we want to freeze the [Ar] shell, ie 18 electrons. Part of them is frozen using the ECP (10e-), and the remaining 4 pairs are frozen correctly.

For the Kr(Gh) + OCS combination, we want to freeze the [He] electrons on O and C, and the [Ne] electrons on S, ie. 14 electrons. But we freeze only 3 orbitals, we are 8 electrons short.

For Kr + OCS, we want to freeze 2x[He], 1x[Ne] and 1x[Ar] shells, ie. 32 electrons. As above, 10 are in Kr's ECP, and the additional 22 are correctly ghosted by the 11 frozen orbitals.

The problem seems to be happening when an ECP-containing atom is ghosted. I think it happened both with explicit ghosting using `(Gh)` and with `extract_subsets`; I didn't keep the other logfile.

[bug_inp.txt](https://github.com/psi4/psi4/files/2219016/bug_inp.txt)
[bug_out.txt](https://github.com/psi4/psi4/files/2219017/bug_out.txt)

loriab:
fixed by #1109

PR #1109, opened 2018-07-27
Do not freeze electrons on ghost atoms.

### LINKED ISSUE #843: Incorrect ghosting behavior for Iodine interaction energy calculations
Hi, I noticed some odd ghosting behavior when doing interaction energy calculations on molecules containing Iodine. 

When manually doing the counterpoise correction for a non-I-containing molecule, Psi4 reads the ghosted atoms properly and computes a sensible interaction energy  (see `correctly-ghosted.txt`).

However, when manually doing the counterpoise correction for an I-containing molecule, Psi4 does not ghost the atoms at all, and does the exact same calculation three times, resulting in an incorrect interaction energy (see `incorrectly-ghosted.txt`).

The (much easier) workaround here is to counterpoise correct using the bsse_type="cp" interface, but I wanted to bring this to your attention anyways.

[correctly-ghosted.txt](https://github.com/psi4/psi4/files/1458283/correctly-ghosted.txt)
[incorrectly-ghosted.txt](https://github.com/psi4/psi4/files/1458284/incorrectly-ghosted.txt)

Thanks!

PeterKraus:
## Description
Fix for #1102. We were subtracting ECP's, as (Z + ECP) was > 0 (usually around 10), even when the atom was ghosted (Z == 0).

## Questions
- [x] Should this be applied against master or against #953?

## Checklist
- [x] [All or relevant fraction of full tests run.](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests) Test `dfmp2-ecp` modified to check for correct behaviour.
- [x] Fix #843.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -175,12 +175,14 @@ std::shared_ptr<BasisSet> construct_basisset_from_pydict(const std::shared_ptr<M
     // Modify the nuclear charges, to account for the ECP.
     if (totalncore) {
         for (int atom = 0; atom < mol->natom(); ++atom) {
-            const std::string& basis = mol->basis_on_atom(atom);
-            const std::string& label = mol->label(atom);
-            int ncore = basis_atom_ncore[basis][label];
-            int Z = mol->true_atomic_number(atom) - ncore;
-            mol->set_nuclear_charge(atom, Z);
-            basisset->set_n_ecp_core(label, ncore);
+            if (mol->Z(atom) > 0) {
+                const std::string& basis = mol->basis_on_atom(atom);
+                const std::string& label = mol->label(atom);
+                int ncore = basis_atom_ncore[basis][label];
+                int Z = mol->true_atomic_number(atom) - ncore;
+                mol->set_nuclear_charge(atom, Z);
+                basisset->set_n_ecp_core(label, ncore);
+            }
         }
     }
 
--- a/psi4/src/psi4/libcubeprop/csg.cc
+++ b/psi4/src/psi4/libcubeprop/csg.cc
@@ -286,7 +286,7 @@ void CubicScalarGrid::write_cube_file(double* v, const std::string& name, const 
 
     // Atoms of molecule (Z, Q?, x, y, z)
     for (int A = 0; A < mol_->natom(); A++) {
-        fprintf(fh, "%3d %10.6f %10.6f %10.6f %10.6f\n", (int)mol_->Z(A), 0.0, mol_->x(A), mol_->y(A), mol_->z(A));
+        fprintf(fh, "%3d %10.6f %10.6f %10.6f %10.6f\n", (int)mol_->true_atomic_number(A), 0.0, mol_->x(A), mol_->y(A), mol_->z(A));
     }
 
     // Data, striped (x, y, z)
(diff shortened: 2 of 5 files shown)


andysim:
In response to the question, I vote that this goes into master.  That said, I'm not the one that has to rebase #953.  Thoughts, @loriab?

loriab:
I've no objection to rebasing 953 again. I'm just trying to figure out how this and #843 interact to make sure things are getting fixed in the broadest place.

andysim:
Agreed.  I'm certain that #1113 is also a symptom of the same problem.

PeterKraus:
@loriab, #843 is related but not fixed: see attached calculations with (a relatively old) master and patched versions:
[master.txt](https://github.com/psi4/psi4/files/2246477/master.txt)
[patched.txt](https://github.com/psi4/psi4/files/2246478/patched.txt)
```
pk@dream:~/psi4/tests/ecp_fix_tests/issue-843::grep "Final E" master.txt 
  @DF-RHF Final Energy:  -594.38702051330188
  @DF-RHF Final Energy:  -297.22774750033523
pk@dream:~/psi4/tests/ecp_fix_tests/issue-843::grep "Final E" patched.txt 
  @DF-RHF Final Energy:  -594.38702051366886
  @DF-RHF Final Energy:  -297.22774750056516
```
While the current patch fixes _a_ bug, it doesn't fix _the_ bug. I'll have a look if I can figure out why ghosts still posess ECP's.

loriab:
I've PR'd to this PR a fix for all the recent ghost/ECP issues (https://github.com/PeterKraus/psi4/pull/2). This is all sidestepping the true problem, which is that we're overloading `mol->Z()` to be atomic number, ghostedness, and non-ECP charge. Errors downstream of libmints Molecule are going to keep popping up, methinks. In fact, I should probably check the mol->json procedure. Thoughts on a forward strategy?

andysim:
Yeah, the use of `Z` is going beyond what it was intended to do and I'm to blame for the ECP instances of that.  I think we need to track both the effective charge and the atomic number separately.  And perhaps a different entry to handle ghostedness.  Happy to help to fix this, but will need to consult with SAPT/CP folks first to make sure the design is right in terms of ghosted behavior.  Sounds like a good conference call topic.  Finding all usages and changing them will be easy with QtCreator.

PeterKraus:
Thanks @loriab, I don't think I'd have found where this is handled. What puzzles me is: how does `extract_subsets` work? I see we tag the whole molecule as `Absent` first and then the appropriate fragments as `Ghost` second (in molecule.cc - `Molecule::extract_subsets`). We then loop over the atoms in molecule.cc `Molecule::reinterpret_coordentries` and apply `set_ghosted` on them, but how does this translate into modifying `Z`?

Incidentally, why don't we just apply `set_ghosted` to each coordentry that's actually ghosted, instead of messing with Z and charge in `Molecule::create_molecule_from_string`?

loriab:
> Incidentally, why don't we just apply set_ghosted to each coordentry that's actually ghosted, instead of messing with Z and charge in Molecule::create_molecule_from_string?

Good question. Partly expedience wrt ECP. Partly my uneven approach to libmints Molecule through the years (sacrosanct, change!, sacrosanct, change!). In any case, in the conf call yesterday, we pledged to reform and have ghosted mean ghosted, charge mean effective nuclear charge and Z mean atomic number (or something similarly sane). I think that'll happen in a separate pass, though. I'm good with this PR fixing the immediate known issues.

PeterKraus:
> In any case, in the conf call yesterday, we pledged to reform and have ghosted mean ghosted, charge mean effective nuclear charge and Z mean atomic number (or something similarly sane).

Super, I'm keen to help when that comes around.

PR #1121, opened 2018-08-04
shields.io update

loriab:
## Description
Correct (py versions), expand (commit activity), and organize the README badges. Also fix the slack invite issue.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Ready for another look. I know there are a lot of badges, but they really are useful, I think. Note that this will never pass appveyor w/o a rebase, so it'll have to be forced.

PR #1130, opened 2018-08-08
SCF: Add SCFConvergence Error 

obrien951:
## Description
SCF: Add SCFConvergence Error such that the error will have the wavefunction as a member.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/exceptions.py
+++ b/psi4/driver/p4util/exceptions.py
@@ -85,12 +85,21 @@ class ConvergenceError(PsiException):
     def __init__(self, eqn_description, maxit):
         msg = "Could not converge %s in %d iterations." % (eqn_description, maxit)
         PsiException.__init__(self, msg)
+        self.maxit = maxit
         self.message = msg
         core.print_out('\nPsiException: %s\n\n' % (msg))
 
 
 class OptimizationConvergenceError(ConvergenceError):
     """Error called for problems with geometry optimizer."""
+
+    def __init__(self, eqn_description, maxit, wfn):
+        ConvergenceError.__init__(self, eqn_description, maxit)
+        self.wfn = wfn
+
+
+class SCFConvergenceError(ConvergenceError):
+    """Error called for problems with SCF iterations."""
 
     def __init__(self, eqn_description, maxit, wfn):
         ConvergenceError.__init__(self, eqn_description, maxit)
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -34,7 +34,7 @@ import numpy as np
 
 from psi4.driver import p4util
 from psi4.driver import constants
-from psi4.driver.p4util.exceptions import ConvergenceError, ValidationError
+from psi4.driver.p4util.exceptions import SCFConvergenceError, ValidationError
 from psi4 import core
 
 from .efp import get_qm_atoms_opts, modify_Fock_permanent, modify_Fock_induced
@@ -71,8 +71,8 @@ def scf_compute_energy(self):
             self.initialize()
             try:
                 self.iterations()
-            except ConvergenceError:
-                raise ConvergenceError("""SCF DF preiterations""", self.iteration_)
+            except SCFConvergenceError:
+                raise SCFConvergenceError("""SCF DF preiterations""", self.iteration_, self)
         core.print_out("\n  DF guess converged.\n\n")
 
         # reset the DIIS & JK objects in prep for DIRECT
@@ -84,7 +84,7 @@ def scf_compute_energy(self):
 
     try:
         self.iterations()
-    except ConvergenceError as e:
+    except SCFConvergenceError as e:
         if core.get_option("SCF", "FAIL_ON_MAXITER"):
             core.print_out("  Failed to converge.\n")
             # energy = 0.0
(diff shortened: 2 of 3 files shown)


loriab:
@CDSherrill 

[This](https://github.com/psi4/psi4/blob/1.2.x/psi4/src/psi4/libscf_solver/hf.cc#L631-L639) was the v1.2 behavior so print error msg or die, depending on settings.

Previous scfitertopy behavior was always to throw generic ConvergenceError with msg and niter.

Now, throw SCFConvergenceError with msg, niter, and wfn attached.

psi4/driver/p4util/exceptions.py
```diff
@@ -85,6 +85,7 @@ class ConvergenceError(PsiException):
     def __init__(self, eqn_description, maxit):
         msg = "Could not converge %s in %d iterations." % (eqn_description, maxit)
         PsiException.__init__(self, msg)
+        self.maxit = maxit
```
loriab: Maybe `self.iteration` or `self.iter`? This could be thrown under < maxit circumstances. In fact, someday there might be `MaxiterExceededSCFConvergenceError`, `OscillationDetectedSCFConvergenceError`, etc.

loriab:
@CDSherrill 

Oh, and if an error's raised but not caught, it just shows the traceback and triggers coffee:

```
PsiException: Could not converge SCF iterations in 4 iterations.

  Failed to converge.

  File "stage/usr/local/psi4/bin/psi4", line 269, in <module>
    exec(content)
  File "<string>", line 30, in <module>
  File "/home/psilocaluser/gits/hrw-patch/objdir/stage/usr/local/psi4/lib/psi4/driver/driver.py", line 492, in energy
  File "/home/psilocaluser/gits/hrw-patch/objdir/stage/usr/local/psi4/lib/psi4/driver/procrouting/proc.py", line 2008, in run_scf
  File "/home/psilocaluser/gits/hrw-patch/objdir/stage/usr/local/psi4/lib/psi4/driver/procrouting/proc.py", line 1362, in scf_helper
    e_scf = scf_wfn.compute_energy()
  File "/home/psilocaluser/gits/hrw-patch/objdir/stage/usr/local/psi4/lib/psi4/driver/procrouting/scf_proc/scf_iterator.py", line 93, in scf_compute_energy
    raise e
  File "/home/psilocaluser/gits/hrw-patch/objdir/stage/usr/local/psi4/lib/psi4/driver/procrouting/scf_proc/scf_iterator.py", line 86, in scf_compute_energy
    self.iterations()
  File "/home/psilocaluser/gits/hrw-patch/objdir/stage/usr/local/psi4/lib/psi4/driver/procrouting/scf_proc/scf_iterator.py", line 358, in scf_iterate
    raise ConvergenceError("""SCF iterations""", self.iteration_)

ConvergenceError: Could not converge SCF iterations in 4 iterations.

    Psi4 stopped on: Wednesday, 08 August 2018 05:42PM
    Psi4 wall time for execution: 0:00:01.01

*** Psi4 encountered an error. Buy a developer more coffee!
*** Resources and help at github.com/psi4/psi4.
```

It looks like [`die_if_not_converged`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libpsi4util/process.cc#L105-L113) toggled print-error-msg/die+traceback previously for SCF and DETCI, with defaulting toward print-error-msg. I think it's better to default toward die (which this PR now does) and ppl can catch the error if they want print-error-msg.

For a long time we had geometry optimizations exit smoothly even if failed for iterations exceeded. That caused some problems (I think @dsirianni hit them), so we switched toward the die+traceback. Now SCF will behave like optimizations.

CDSherrill:
@loriab: This sounds good --- I think die+traceback is probably the right
default if the exception isn't caught.

psi4/driver/p4util/exceptions.py
dgasmith: `self.iteration` would be a huge help. I would also suggest `self.e_convergence` and `self.d_convergence` would be a huge future boon so that it allows users to inspect some of the circumstances of the failure and determine if it passes a "good enough" mark.

psi4/driver/p4util/exceptions.py
```diff
@@ -97,6 +98,14 @@ def __init__(self, eqn_description, maxit, wfn):
         self.wfn = wfn
 
 
+class SCFConvergenceError(ConvergenceError):
+    """Error called for problems with SCF iterations."""
+
+    def __init__(self, eqn_description, maxit, wfn):
```
dgasmith: I would try to put more important args first. So something more like:
`def __init__(self, wfn, wfn_name, maxiter, e_conv, d_conv):`

psi4/driver/p4util/exceptions.py
loriab: `self.maxit` _is_ the iteration number. You're agreeing to rename the exception member data?

I think e_ and d_convergence can be extracted from the wfn. You're suggesting appending current energy and current rms density to error?

dgasmith: Yes to renaming the member data.

Unless I am missing something it does not appear that you can extract these quantities from the Wavefunction. At the same time I would argue iter/e_conv/d_conv data makes up the current "convergence state" of the wavefunction and should be carried around together so either get all of these variables from the wavefunction or get none.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -71,8 +71,8 @@ def scf_compute_energy(self):
             self.initialize()
             try:
                 self.iterations()
-            except ConvergenceError:
-                raise ConvergenceError("""SCF DF preiterations""", self.iteration_)
+            except SCFConvergenceError:
+                raise SCFConvergenceError("""SCF DF preiterations""", self.iteration_, self)
```
dgasmith: Looking through the code we probably need to call `scf_finalize_energy` here. There are many variables set in the `scf_finalize_energy/scf_print_energies` functions which we should probably change at some point (or just rename the functions).

loriab:
ok, various issues have been addressed during the Sherrill Group programming exercise. Ready for re-review.

Issue #1116, opened 2018-08-02
Implementation of I/O less OEProps in-memory


timostrunk:
I would like to implement a faster way to do ESP fits than what is found here:
https://github.com/psi4/psi4numpy/tree/master/One-Electron-Property/Restrained-Electrostatic-Potential

The solution above is writing multiple grids on disk, yet the strength of the python and C++ APIs of Psi4 should be to have these in memory and allow for fast multithreaded evaluations.

My development proposal would be a little more involved, but also give the Psi4 API a lot more power. I would split the whole OEProp class into:

OEPropCalc
This class will get all currently existing logic to calculate the requested properties. Most properties are already calculated as SharedMatrix. This class will not be allowed to have any kind of output, unless a bulk property is requested (such as a grid).
OEProp
This class gets all output handling, it will only call the respective OEPropCalc functions and then use the existing code for output. Data will be passed between OEPropCalc and OEProp as SharedPointer and therefore not cost overhead. Exception to this rule would be the functions writing grids, which will get an in-memory and a direct write solution.
Reason:
Currently OEProp sometimes exposes data in memory (such as Mulliken charges using wfn.set_array) and sometimes not (such as Multipoles only via stdout or outfile). The approach above would allow internal code use of SharedMatrices directly (OEPropCalc would get a public python API) and keep existing functionality untouched.

Once the split is done, I would also split the GridEvaluation routines to either write a grid-XXX.dat (the existing functionality) or to keep the grids in memory (new functionality) and generate them with a grid taken from memory using the existing GridIterator interface.

Are there any objections or is somebody else currently working on this?

dgasmith:
Thanks for the suggestion. As a first pass this sounds great and I don’t think anyone else is working on it currently. I agree that OEProp should be a bit more straightforward about the output. Let me think on this a bit and perhaps source some other feedback. 

Thanks for delving into this!

loriab:
libcubeprop will also be getting some I/O refactoring soon, so feel free to chime in if there's anything relevant you see going wrong or right with your planned uses.

timostrunk:
This issue is adressed in https://github.com/psi4/psi4/pull/1133
I did not use the GridIterator interface as the memory overhead of passing the complete grid from python is currently neglibible. 
Also only one of the Grid functions is converted. Converting the others after the split will be trivial.
Therefore closing the issue.

PR #1133, opened 2018-08-13
Split of Prop and OEProp to allow I/O less computation of Properties without globals.

timostrunk:
## Description
This pull request adresses https://github.com/psi4/psi4/issues/1116 :

1.) Split Prop into 
- - Prop: All One Electron convenience accessors and setters and getters, but no access to globals at all. Prop is now only a convenience class, which sets up a few nice accessors.
- - TaskListComputer: This class handles all the compute baseclass elements of Prop, adding, removing tasks from a task queue and computing them. It's very low complexity.
This was done to allow Prop to be easily inheritable without having to implement all the Taskqueue things, which honestly did not have a lot do with Prop itself. This commit makes using Prop inside the code a ton easier.
This was done in commit 
528c0d7 : https://github.com/psi4/psi4/commit/528c0d719e8bb49b12652417b00db506241a4bee

2.) Split OEProp in three distinct classes:
- - MultipolePropCalc: This class calculates all multipole properties and mo extents in commit: 
59005ee https://github.com/psi4/psi4/commit/59005eed1c0bc8b893927284c31386eb4b715f1e
- - PopulationAnalysisCalc: This class calculates Mulliken / Loewdin charges and Bond Orders in commit: a6caee7 https://github.com/psi4/psi4/commit/a6caee7091afc6b0e8071ef0d077f93fcb6c9370
- - ESPPropCalc: This class allows calculation of fields on Grids in commit: 
325149b https://github.com/psi4/psi4/commit/325149b31751b629486f0478cc5cfe1ee1ec9fb1

The rest of the commits are mostly fixups and linking up the three new classes to the old completely unchanged OEProp class, which now does not calculation itself. Also I implemented an in-memory routine calculating grid properties based on a grid, which was passed on.
The three new classes do not touch any global and no environment. They mostly exist without data storage for the grids, all functions have explicit (and mostly shared pointer return values).
Their complete setup is explicit. All environment variables are only set by OEProp.

The reason is again: This allows to the new helpers to be used directly without any interference from Python and C++. The reason they were split this way: ESPPropCalc and PopulationAnalysisCalc and MultipolePropCalc do not share any members or variables.  Example: only MPC requires an explicit origin to be set.

## Questions
- [x]  I changed a return type to const & to avoid a large copy: https://github.com/psi4/psi4/commit/325149b31751b629486f0478cc5cfe1ee1ec9fb1#r30055651
- [x] The return type in compute_mo_extents is an actual vector instead of a Shared pointer. It contains maximum three elements, so I left it at that: https://github.com/psi4/psi4/commit/59005eed1c0bc8b893927284c31386eb4b715f1e#r30055734
- [x] I could remove the Prop baseclass completely from OEProp in the future. Only one function uses something from this.

## Checklist
  The tests are all still called from OEProp. The only new thing is the in-memory grid routine.
- [x] [All or relevant fraction of full tests run]
ctest -L quick was run. No errors

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/oeprop.cc
+++ b/psi4/src/psi4/libmints/oeprop.cc
@@ -2152,12 +2152,13 @@ PopulationAnalysisCalc::compute_wiberg_lowdin_indices(bool print_output) {
 }
 
 void OEProp::compute_no_occupations() {
-    std::vector<std::vector<std::tuple<double, int, int>>> metrics;
-    pac_.compute_no_occupations(metrics, max_noon_, true);
-    wfn_->set_no_occupations(metrics);
+    auto metrics_ptr = pac_.compute_no_occupations(max_noon_, true);
+    wfn_->set_no_occupations(*metrics_ptr);
 }
-void PopulationAnalysisCalc::compute_no_occupations(
-    std::vector<std::vector<std::tuple<double, int, int>>>& output_metrics, int max_noon, bool print_output) {
+std::shared_ptr<std::vector<std::vector<std::tuple<double, int, int>>>> PopulationAnalysisCalc::compute_no_occupations(
+    int max_noon, bool print_output) {
+    std::shared_ptr<std::vector<std::vector<std::tuple<double, int, int>>>> output_metrics_ptr =
+        std::make_shared<std::vector<std::vector<std::tuple<double, int, int>>>>();
     std::vector<std::string> labels = basisset_->molecule()->irrep_labels();
 
     if (print_output) {
@@ -2165,7 +2166,7 @@ void PopulationAnalysisCalc::compute_no_occupations(
     }
 
     // Terminally, it will be [metric_a , metric_b, metric] or [metric] depending on same_dens
-    std::vector<std::vector<std::tuple<double, int, int>>>& metrics = output_metrics;
+    std::vector<std::vector<std::tuple<double, int, int>>>& metrics = *output_metrics_ptr;
 
     if (!same_dens_) {
         SharedVector Oa;
@@ -2279,7 +2280,7 @@ void PopulationAnalysisCalc::compute_no_occupations(
         }
         outfile->Printf("\n");
     }
-
+    return output_metrics_ptr;
     //for(int h = 0; h < epsilon_a_->nirrep(); h++) free(labels[h]); free(labels);
 
 }
--- a/psi4/src/psi4/libmints/oeprop.cc
+++ b/psi4/src/psi4/libmints/oeprop.cc
@@ -760,7 +760,7 @@ Vector3 Prop::compute_center(const double *property) const
 }
 
 
-OEProp::OEProp(std::shared_ptr<Wavefunction> wfn) : Prop(wfn), mpc(wfn)
+OEProp::OEProp(std::shared_ptr<Wavefunction> wfn) : Prop(wfn), mpc(wfn), pac(wfn)
 {
     common_init();
 }
@@ -1655,22 +1655,40 @@ std::vector<SharedVector> MultipolePropCalc::compute_mo_extents(bool print_outpu
     return mo_es;
 }
 
+typedef PopulationAnalysisCalc PAC;
+
 void OEProp::compute_mulliken_charges()
 {
-    outfile->Printf( "  Mulliken Charges: (a.u.)\n");
+    PAC::SharedStdVector Qa_ptr,Qb_ptr,apcs ;
+    std::tie(Qa_ptr,Qb_ptr,apcs) = pac.compute_mulliken_charges(true);
+    wfn_->set_atomic_point_charges(apcs);
 
+    auto vec_apcs = std::make_shared<Matrix>("Mulliken Charges: (a.u.)", 1, apcs->size());
+    for (size_t i = 0; i < apcs->size(); i++){
+        vec_apcs->set(0, i, (*apcs)[i]);
+    }
+    wfn_->set_array("MULLIKEN_CHARGES", vec_apcs);
+}
+
+std::tuple<PAC::SharedStdVector,PAC::SharedStdVector,PAC::SharedStdVector>  PopulationAnalysisCalc::compute_mulliken_charges(bool print_output)
+{
+    if (print_output)
+    {
+        outfile->Printf( "  Mulliken Charges: (a.u.)\n");
+    }
     std::shared_ptr<Molecule> mol = basisset_->molecule();
+
+    auto Qa_ptr = std::make_shared<std::vector<double>>(mol->natom());
+    auto Qb_ptr = std::make_shared<std::vector<double>>(mol->natom());
+    std::vector<double> & Qa = *Qa_ptr;
+    std::vector<double> & Qb = *Qb_ptr;
+
     auto apcs = std::make_shared<std::vector<double>>(mol->natom());
-    double* Qa = new double[mol->natom()];
     double* PSa = new double[basisset_->nbf()];
     double suma = 0.0;
 
-    double* Qb = new double[mol->natom()];
     double* PSb = new double[basisset_->nbf()];
     double sumb = 0.0;
-
-    ::memset(Qa, '\0', mol->natom()*sizeof(double));
-    ::memset(Qb, '\0', mol->natom()*sizeof(double));
 
     SharedMatrix Da;
     SharedMatrix Db;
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1931,33 +2221,44 @@ void OEProp::compute_wiberg_lowdin_indices()
...
+    return std::make_tuple(WBI_total,WBI_alpha,WBI_beta,WBI_valence);
 }
 
 void OEProp::compute_no_occupations()
+{
+    std::vector<std::vector<std::tuple<double, int, int> >> metrics;
+    pac.compute_no_occupations(metrics, max_noon_,true);
+    wfn_->set_no_occupations(metrics);
+}
+void PopulationAnalysisCalc::compute_no_occupations(std::vector<std::vector<std::tuple<double, int, int> > > & output_metrics, int max_noon, bool print_output)
```
dgasmith: Looking at other parts of the code, would it make more sense to return the output_metrics rather than have to build them and pass them in?

psi4/src/psi4/libmints/oeprop.h
```diff
@@ -266,26 +430,18 @@ class OEProp : public Prop {
     /// Compute field at specified grid points
     void compute_field_over_grid();
 
+    MultipolePropCalc mpc;
```
dgasmith: Protected variables must have a appended `_`, so `mpc_` etc.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -764,13 +759,72 @@ Vector3 OEProp::compute_center(const double *property) const
...
 
+    // Determine number of NOONs to print; default is 3
+    if(options.get_str("PRINT_NOONS") == "ALL") max_noon_ = wfn_->nmo();
+        else max_noon_ = to_integer(options.get_str("PRINT_NOONS"));
+}
+
+
+MultipolePropCalc::MultipolePropCalc(std::shared_ptr<Wavefunction> wfn, Vector3 const & origin) : Prop(wfn), origin_(origin)
+{
+    typedef MultipolePropCalc MPC;
```
dgasmith: This does not appear to be used.

jturney:
I say try to remove the `Prop` base class. Like you said, only one function uses something from it.

psi4/src/psi4/libmints/oeprop.h
timostrunk: Thank you for noticing this, I will amend the pull request!

psi4/src/psi4/libmints/oeprop.cc
timostrunk: True, it is a leftover, which I will remove, while amending the pull request. It was used similar to the PAC typedef to simplify outputs and make everything a bit shorter.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1519,47 +1733,79 @@ void OEProp::compute_mo_extents()
...
+    wfn_->set_atomic_point_charges(apcs);
+
+    auto vec_apcs = std::make_shared<Matrix>("Mulliken Charges: (a.u.)", 1, apcs->size());
+    for (size_t i = 0; i < apcs->size(); i++){
+        vec_apcs->set(0, i, (*apcs)[i]);
+    }
+    wfn_->set_array("MULLIKEN_CHARGES", vec_apcs);
+}
 
+std::tuple<PAC::SharedStdVector,PAC::SharedStdVector,PAC::SharedStdVector>  PopulationAnalysisCalc::compute_mulliken_charges(bool print_output)
```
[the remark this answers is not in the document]
timostrunk: Done, will be in the commits coming soon. I ran git-clang-format HEAD~12 on my rebased branch and fixed a few indentations of mismatching format with old code.

One thing though:
clang-format indented my
    public:
    protected: 
etc.

I haven't seen this yet in the psi4 codebase.

psi4/src/psi4/libmints/oeprop.cc
timostrunk: It would be very nice to have this as an actual return value.
The reason i did not do it yet is simply:
The return type would be 
std::shared_ptr<std::vector<std::vector<std::tuple<double, int, int> > > >
which I found quite excessive. Everything else would lead to extra copies.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1602,52 +1848,73 @@ void OEProp::compute_mulliken_charges()
...
+std::tuple<PAC::SharedStdVector,PAC::SharedStdVector,PAC::SharedStdVector> PopulationAnalysisCalc::compute_lowdin_charges(bool print_output)
+{
+    if (print_output)
+    {
+        outfile->Printf( "  Lowdin Charges: (a.u.)\n");
+    }
     std::shared_ptr<Molecule> mol = basisset_->molecule();
+
+    auto Qa_ptr = std::make_shared<std::vector<double>>(mol->natom());
+    auto Qb_ptr = std::make_shared<std::vector<double>>(mol->natom());
```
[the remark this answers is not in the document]
timostrunk: The valid question is: Why did I return Qa_ptr and Qb_ptr in the first place. They are used nowhere in the codebase except for in this function.

I thought them to be useful extra arrays in some usecases, which are calculated anyways. Most people will only be interested in the charges, but the overhead is so small (2 extra pointers, which will be reference counted to 0 immediately, if not used) that i left it.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1353,44 +1549,60 @@ void OEProp::compute_quadrupole(bool transition)
...
+    double xx = qe[0]*dfac;
+    double yy = qe[3]*dfac;
+    double zz = qe[5]*dfac;
+    double xy = qe[1]*dfac;
+    double xz = qe[2]*dfac;
+    double yz = qe[4]*dfac;
 
+    auto output = std::make_shared<Matrix>(3,3);
+    Matrix & outmat = *output;
+    outmat.set(0,0,xx);
```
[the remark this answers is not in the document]
timostrunk: In some cases there is no benefit at all. I do it quite often, because it is a null operation and makes my life easier sometimes, when working with operator[].
In case of Matrix and Vector, there are explicit set and get functions, so I can remove it.

On occasion, where it did have some benefit is for example here:
    auto Qa_ptr = std::make_shared<std::vector<double>>(mol->natom());
    auto Qb_ptr = std::make_shared<std::vector<double>>(mol->natom());
    std::vector<double>& Qa = *Qa_ptr;
    std::vector<double>& Qb = *Qb_ptr;
In this case i can just write
Qa[6] instead of (*Qa_ptr)[6].

Should i remove the superfluos occasions?

timostrunk:
I removed the Prop baseclass in the upcoming commits (pending tests). i had to add a Da_name accessor to prop though, because OEProp uses the Da matrix as a fallback if no title is given.

psi4/src/psi4/libmints/oeprop.cc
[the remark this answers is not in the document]
dgasmith: Canonically we would use `Qa_ptr->set(6, value)` or `Qa_ptr->get(6)` and generally rename `Qa_ptr` to `Qa`. In general, we pull out the raw pointer quite a bit. The requested changes are not necessary, just observing current patterns.

psi4/src/psi4/libmints/oeprop.cc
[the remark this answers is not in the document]
dgasmith: It should indent them by two if I remember correctly.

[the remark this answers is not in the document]
loriab: I recall being surprised by clang-format's treatment of class/public/protected, too. But I can live with it.

The other answer is that while we're trying to format new code or code edits, clang-format hasn't been run over the whole codebase.

timostrunk:
I commited all requested changes except for the reference aliases. Commits should be styled using clang-format.

psi4/src/psi4/libmints/oeprop.cc
dgasmith: Hmm, this way a user can write `auto my_pac-> compute_no_occupations(5, true);` which would be more natural, I think. Either way.

timostrunk: I did not consider the auto keyword. You are absolutely right and I will change this immediately, because it is so much cleaner.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1519,47 +1672,67 @@ void OEProp::compute_mo_extents()
...
+std::tuple<PAC::SharedStdVector, PAC::SharedStdVector, PAC::SharedStdVector>
+PopulationAnalysisCalc::compute_mulliken_charges(bool print_output) {
+    if (print_output) {
+        outfile->Printf("  Mulliken Charges: (a.u.)\n");
+    }
     std::shared_ptr<Molecule> mol = basisset_->molecule();
+
+    auto Qa_ptr = std::make_shared<std::vector<double>>(mol->natom());
+    auto Qb_ptr = std::make_shared<std::vector<double>>(mol->natom());
+    std::vector<double>& Qa = *Qa_ptr;
```
dgasmith: On this we would typically name `Qa_ptr` as Qa and then `double* Qa_ptr = Qa.data()`.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1519,47 +1672,67 @@ void OEProp::compute_mo_extents()
...
+    std::vector<double>& Qb = *Qb_ptr;
+
     auto apcs = std::make_shared<std::vector<double>>(mol->natom());
-    double* Qa = new double[mol->natom()];
     double* PSa = new double[basisset_->nbf()];
```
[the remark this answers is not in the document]
dgasmith: Thanks, we are slowly moving towards a more modern C++ layout, so there are some growing pains.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -916,6 +925,23 @@ void OEProp::compute()
...
+        /*- Process::environment.globals["DIPOLE X"] -*/
+        /*- Process::environment.globals["DIPOLE Y"] -*/
+        /*- Process::environment.globals["32-POLE XXXXX"] -*/
+        /*- Process::environment.globals["32-POLE XXXXY"] -*/
+        Process::environment.globals[name] = total_mpole;
+    }
+}
+
+MultipolePropCalc::MultipoleOutputType MultipolePropCalc::compute_multipoles(int order, bool transition,
+                                                                             bool print_output, bool verbose) {
```
loriab: For clarification, `print_output` signals the difference between a silent function and a normal modest-print-level function, while `verbose` signals additional debug level printing?

This is a bikeshedding comment, so fine for now. I just want to understand the intent in case we ever reform our print levels.

timostrunk: print_output = True outputs the unchanged output from the OEProp library.

The Tasklist part of the Prop class had two print levels: int print_ == 2 and print_ == 4, so there was the attempt to implement some private print_level system here.
They were only used for debugging purposes printing extra information about
- the multipole integrals
- the ESP at Nuclei integrals

If there was a debuglevel enum, or a general different logger class with .INFO, .DEBUG etc. functions, the verbose function argument could be dropped.

dgasmith:
@timostrunk BTW can you update your issue showing that your points are complete?

PR #1139, opened 2018-08-17
Fix Casting Incompatibility with Multiple Geometries

JonathonMisiewicz:
## Description
There was a bug (see conversation between myself and Andy Simmonett in Slack channel earlier today) where if you attempted to use orbital casting on a frequency computation with symmetry, the first symmetry-breaking displacement would succeed at the small basis computation but crash at the large basis computation, saying that it couldn't cast orbitals between symmetries.

The code was previously set so that GUESS would be set to READ. This would cause Psi to attempt to read the last saved orbitals. As [the casting mechanism has nothing to do with reading](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1336-L1341), the last saved orbitals weren't the orbitals from the small basis computation, but the orbitals from the last geometry, which may well be of the wrong point group.

As I can find no reason to have GUESS set to READ, the offending code has been removed.

## Todos
- [x] Fix a bug causing `basis_guess` to crash if the user has geometries of different symmetries, e.g., a finite difference hessian for a system with symmetry

## Checklist
- [x] [Quicktests and cast tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)
- [x] Confirmed this fixes the problem I saw

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1255,9 +1255,7 @@ def scf_helper(name, post_scf=True, **kwargs):
         core.IO.change_file_namespace(180, guesspace, namespace)
         core.IO.set_default_namespace(namespace)
 
-        # Set to read and project, and reset bases to final ones
         optstash2.restore()
-        core.set_local_option('SCF', 'GUESS', 'READ')
 
         # Print the banner for the standard operation
         core.print_out('\n')


JonathonMisiewicz:
Note to Self: Optimizations do not play nicely with this. It sets the orbital guess type to read, so it displays the following error message...
```Warning! Guess was READ without Ca set, switching to CORE! 
           This option should have been configured at the driver level.```

...even though it should not have been set at the driver level! Within the assumption that we have no way to cast orbitals between symmetries, the best option seems to be to have the big basis read from the projected orbitals while the small basis uses a core guess during any findif computations and reads during optimizations. Thoughts? This will probably end with a lot of poking at the driver code.

JonathonMisiewicz:
Additional testing confirms that the error message I reported in my last message displays regardless of whether this fix is in place. This should be investigated in more detail later, but it's outside the scope of this PR.

PR #1140, opened 2018-08-17
Remove initialization of DFHelper object in fisapt fexch and find.

obrien951:
## Description
Removes initialization of DFHelper object in fexch and find by having a dfh object as member data.

## Todos
- [x] Circumvents recalculation of molecular integrals in the fisapt package.
- [x] Provides a corresponding speed-up 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/fisapt/fisapt.h
+++ b/psi4/src/psi4/fisapt/fisapt.h
@@ -33,8 +33,6 @@
 #include "psi4/libmints/basisset.h"
 #include "psi4/libmints/wavefunction.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
-//#include "psi4/lib3index/dfhelper.h"
-
 
 #include <map>
 #include <tuple>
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/fisapt/fisapt.h
```diff
@@ -33,6 +33,8 @@
 #include "psi4/libmints/basisset.h"
 #include "psi4/libmints/wavefunction.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
+#include "psi4/lib3index/dfhelper.h"
```
dgasmith: Can you forward declare the class by adding `class DFHelper::DFHelper` around line 44 instead of including this header?

loriab: fyi, [this](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/basisset.h#L51) is what a forward declaration looks like.

PR #1141, opened 2018-08-17
doc: Reorganize programmer's manual

CDSherrill:
## Description
Begin reorganizing Programmers' Manual.  Put in some stub pages.

## Status
- [x] Ready for review
- [x] Ready for merge


doc/sphinxman/source/documentation.rst
```diff
@@ -30,8 +30,8 @@
 
 .. _`sec:documentation`:
 
-Documentation
-=============
+Updating the |PSIfour| User and Programmer Documentation
```
loriab: Note that the page title is what appears in the front-page TOC. If you want detailed for the page title but short-and-snappy for the TOC, I suggest something like the below.

```
.. toctree::
   :maxdepth: 2

   FAQ <build_faq>
   build_obtaining
   build_planning
   Installing from Binary <conda>
```

CDSherrill: Good tip @loriab , thank you.  In this case, I want to be more explicit because I want to distinguish between the user manual and programmer manual, which is what this page is about, and other kinds of documentation that a programmer's manual might discuss, such as how to document your code (this will need to be another section of our outline, which otherwise might also be called "documentation").

CDSherrill: Hm, actually I mean to say something more like "User's Manual and Programmer's Manual" rather than "Documentation", just to be totally clear.  I'll fix that.

loriab: Just a pattern to consider.

"Updating the Users' and Programmers' Manual" very reasonably falls under the user question of "how to build the docs" and the dev question of 'how to add a page and build the docs' and the build system question of 'what targets are available and what dependencies do they entail'. Whereas "How to document your code" itself has multiple sections for c++ doxygen, c++ export, py docstring, and written docs, which are themselves module, interface/addon, or plugin docs. Some of these also fall under "coding style". And both intended headings fall under the topic of "Documentation".

When the outline goes circular, I find the FAQ model of lots of small-scope questions and answers to be helpful. That way, one can have the overall organizational [TOC outline](http://psicode.org/psi4manual/master/index.html), the all-available-details [index](http://psicode.org/psi4manual/master/build_faq.html), and the [grand topic](http://psicode.org/psi4manual/master/build_planning.html#faq-builddetailed) with overview and condensed details, all linking to the same docs chunks.

CDSherrill: I agree that having multiple links to this section from other parts of the manual (or even from outside the manual) would be a good idea.  It could possibly be broken up into smaller chunks some of which could be collected in a FAQ.  We could consider that at a next pass.

PR #1150, opened 2018-08-24
Fix #1149 - VV10 gradients now findif by default

### LINKED ISSUE #1149: Have energy derivatives intercept missing analytic methods on auto
If the user doesn't specify how a gradient or hessian is to be computed (difference of energies, difference of gradients, or analytically), Psi will choose to compute it by the most analytic method available to the general method. This leads to annoying crashes when that method isn't available for the more particular variant of the method requested. For instance, the following input file will crash:
```
molecule {
 O 
 H 1 1.0 
 H 1 1.0 2 105 
}

set {
    reference     rks 
    basis sto-3g
}

gradient('wb97x-v')
```

The error is that:
```
Fatal Error: V: RKS cannot compute VV10 gradient contribution.
```

But in that case, Psi shouldn't have tried to do an analytic gradient computation at all but just done finite difference of energies.

I'm not familiar with how the method selection works, but I have a bad feeling that writing this would be rather painful.

JonathonMisiewicz:
## Description
Fixes #1149.

## Todos
- [x] Gradient computations that need VV10 are now done by finite difference of energies by default.
- [x] Ran YAPF on `proc_table.py`

## Checklist
- [x] Confirmed it fixes the original issue

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/procrouting/proc_table.py
```diff
@@ -43,196 +42,197 @@
...
-      # Gradients
-      if ((not ssuper.is_c_hybrid()) and (not ssuper.is_c_lrc())):
-            procedures['gradient'][key] = proc.run_scf_gradient
+    # Gradients
+    if ((not ssuper.is_c_hybrid()) and (not ssuper.is_c_lrc()) and (not ssuper.needs_vv10())):
+        procedures['gradient'][key] = proc.run_scf_gradient
 
-      # Hessians
-      if not ssuper.needs_xc():
-            procedures['hessian'][key] = proc.run_scf_hessian
```
loriab: thanks. who knows when these got out of whack.

psi4/driver/procrouting/proc_table.py
```diff
@@ -43,196 +42,197 @@
...
+        'cisd': proc.run_detci_property,
+        'cisdt': proc.run_detci_property,
+        'cisdtq': proc.run_detci_property,
+        'ci': proc.run_detci_property,  # arbitrary order ci(n)
+        'fci': proc.run_detci_property,
+        'rasscf': proc.run_detci_property,
+        'casscf': proc.run_detci_property,
+        # Upon adding a method to this list, add it to the docstring in property() below
+    }
+}
```
loriab: I think this is a bit less readable. Add ` # yapf: disable` to the closing dict curly brace will make yapf leave it alone. Revert, I think (or check in unformatted copy and we'll squash the PR).

JonathonMisiewicz: I don't think I'm following. You seem to be to saying that you want the two closing braces on the same line, so I should revert the entire commit, add the yapf disable command to the brace line, repeat the yapf/commit/push, so you can squash. Is that right? If so, why not just have me change the brace and add the yapf disable without reverting the commit?

loriab: We want all the changes to the procedures dictionary to go away. But while you’re at it, may as well make it invulnerable to future “yapf -i proc_table.py”. Adding the yapf disable on the final dict line will prevent formatting of the whole dict.

JonathonMisiewicz: Does "all the changes" include the fact that yapf wants energy/gradient/properties indented 4 spaces instead of 8? I can't see how keeping all 8 spaces improves readability.

loriab: All I really want is for the `: proc.` to remain vertically aligned. All the other reformatting yapf has done to the dictionary (incl. 8->4) is good by me. That will require some hand formatting -- I was just trying to offer an easy way out by "revert all".

PR #1151, opened 2018-08-24
refactor empirical dispersion

loriab:
## Description
So I wrote `run_dftd3` back in the day. Then @dgasmith unified dftd3 and libdisp handling through the `EmpiricalDispersion` class. Then @PeterKraus added the DFT dict_builder that allowed dispersion parts of a functional. Then @hokru added NL dispersion into the mix. And we were all very courteous and worked around each other's code, yet all used selected pieces from dashparam.py. The result was a tangle. This is a redesign and a backport of dftd3-running tech from qcdb.

## Todos
- [x] libdisp's -D2 made more flexible so it can use same s6/alpha6/sr6 params as Grimme's -D2. existing D2 dict defs needed filling in further
- [x] no more d2p4/d2gr. that choice now made through `energy(..., engine='dftd3|libdisp')`
- [x] _all_ dispersions served out of dashparam for dftd3/libdisp/nl. accordingly much of the metadata from EmpiricalDispersion moved to dashcoeff.
- [x] EmpiricalDispersion much reduced, sorry. On the other hand, I suspect it's now the slim unification class DGAS actually wanted, if it hadn't needed to contort around run_dftd3.
- [x] All the different combos of fctl/dash/fctl-dash/params/default-params/partial-user-params sorted out in `dashparam.from_arrays()`. This is used as the backend to everything. `qcdb.Molecule.run_dftd3()`, `psi4.core.Molecule.run_dftd3()`, `qcdb.energy('b3lyp-d3')`, `psi4.energy('b3lyp-d3')`, `EmpiricalDispersion.compute_energy()`
- [x] Use QCJson_Schema model to drive and harvest from dftd3 rather than one big function.
- [x] This passes all existing tests. However, I've devised one it doesn't pass (user-def fctl incl. disp). Working on that.
- [x] Not yet begun to trace how citations are passed around.
- [x] Yes, I need to rebase to get Jonathan's findif changes incorporated.
- [x] Bump pylibefp to 0.4 and finish its c->py transition in build paraphernalia.
- [x] current commits are rich in docstrings but actual docs need a pass. some stuff, notably DFT_DISPERSION_PARAMETERS is out of date.

## Questions
- [ ] This touches a lot of people's code. I'm fairly happy that this is a simplification and a workable design. But it's just a better starting point, and others may have new ideas and improvements. Or see holes in my logic and planning.

## Checklist
- [x] All full tests run

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -25,259 +25,243 @@
...
+                ptype='gradient',
+                verbose=1)
+
+            dashd_part = core.Matrix.from_array(jobrec['qcvars']['DISPERSION CORRECTION GRADIENT'].data)
+            for k, qca in jobrec['qcvars'].items():
+                if not isinstance(qca.data, np.ndarray):
+                    core.set_variable(k, qca.data)
+
+            if self.fctldash in ['hf3c', 'pbeh3c']:
+                gcp_part = gcp.run_gcp(molecule, self.fctldash, verbose=False, dertype=1)
```
PeterKraus: I suppose while there are only 2 functionals using `gcp`, it's not really a big deal, but it wouldn't hurt to standardize this by adding a `"gcp"` key to the `dict_func` parsers...

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,598 @@
...
+        'formal'            : "D2",
+        'alias'             : ['d'],
+        'description'       : "    Grimme's -D2 Dispersion Correction",
+        'citation'          : "    Grimme, S. (2006), J. Comp. Chem., 27: 1787-1799",
+        'bibtex'            : "Grimme:2006:1787",
+        'default'           : collections.OrderedDict([('s6', 1.0), ('alpha6', 20.0), ('sr6', 1.1)]),
+        'definitions': {
+            'blyp'          : {'s6': 1.2,  'alpha6': 20.0, 'sr6': 1.1},
+            'bp86'          : {'s6': 1.05, 'alpha6': 20.0, 'sr6': 1.1},
+            'b97'           : {'s6': 1.25, 'alpha6': 20.0, 'sr6': 1.1},  # formerly b97-d
```
PeterKraus: I don't understand why this was necessary; the situation with B97, B97-1 and B97-D is rather confusing as is...

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,598 @@
...
+    elif dashlevel_candidate_1 is not None and dashlevel_candidate_2 is not None:
+        if dashlevel_candidate_1 != dashlevel_candidate_2:
+            raise ValidationError(
+                """Inconsistent -D correction level from name_hint ({}) and level_hint ({})""".format(
+                    dashlevel_candidate_2, dashlevel_candidate_1))
+    dashleveleff = dashlevel_candidate_1 or dashlevel_candidate_2
+
+    try:
+        allowed_params = dashcoeff[dashleveleff]['default'].keys()
+    except KeyError:
```
PeterKraus: I wonder when would this ever be the case. Aren't both `dashlevel_candidate_{1,2}` populated from `get_dispersion_aliases()`?

hokru:
On my wishlist is the addition of the D3(ABC) term (`dftd3 -abc`). I haven't really gone through all the changes, but it should now be easier to add now, right?

Envisioning `energy('pbe-d3bjabc')`

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -107,10 +107,11 @@ def get_functional_aliases(functional_dict):
     return aliases
 
 
-# alias table for dispersion
-dispersion_names = get_dispersion_aliases()
+_dispersion_aliases = intf_dftd3.get_dispersion_aliases()
 
 functionals = {}
+dashcoeff_supplement = collections.defaultdict(lambda: collections.defaultdict(dict))
```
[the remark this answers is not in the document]
loriab: Yes, defaultdict's are a little unlovely to print b/c there's lambda's around. But great for autovivification.

psi4/driver/procrouting/empirical_dispersion.py
loriab: Yeah, I need to rewrite the gcp interface. Unlike in dftd3, where everything is controlled through `.dftd3_params` and internal-to-dftd3-exe definitions of, say, PBE, are ignored, gcp just [passes the keyword through](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/interface_gcp.py#L215). So parameters added to the `dict_func` parsers would have no place to go.

You're completely right, though, about explicit parameter handling (assuming gcp params are as well-defined as dftd3) and a place in the dft dict definition being the desired end.

loriab:
@hokru, I think @dsirianni has dreams of `dftd3 -abc`, too. I think that'll be easier to add after this.

psi4/driver/procrouting/empirical_dispersion.py
hokru: the standard/short version of the gcp parameter file includes the name of basis set, and otherwise 4 numbers. The basis set related values can be supplied in an 'extended' version, which is automatically detected by gcp.

However, hf3c and pbeh3c require additional ingredients which are currently only detected by the `-level` keyword.

loriab: thanks, gtk. As the 3c's were our target cases, explains why (even in my haste) I didn't reuse the `dftd3_param` machinery in gcp. Glad to rework it should need arise.

psi4/driver/qcdb/intf_dftd3/dashparam.py
loriab: I think you're right. May have been a remanent of earlier logic (been a few passes). I'll remove it, and at worst, it'll `KeyError`, not proceed on a wrong path.

psi4/driver/procrouting/dft_funcs/dict_builder.py
```diff
@@ -120,27 +121,31 @@ def get_functional_aliases(functional_dict):
 
     # if the parent functional is already dispersion corrected, skip to next
     if "dispersion" in dict_functionals[functional_name]:
+        disp = dict_functionals[functional_name]['dispersion']
+        dashcoeff_supplement[disp['type']]['definitions'][functional_name] = disp['params']
+        # TODO still needs citation
+        # this is to "bless" dft_funcs dispersion definitions
```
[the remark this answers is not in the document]
loriab: So the `func['dispersion']` only gets passed once, via `dashcoeff_supplement` (which is of the same stature, if present, as `intf_dftd3.dashcoeff` itself). This saves the param_tweaks role in `intf_dftd3.from_arrays(..., param_tweaks)` for partial user modifications to defined things, rather than having three sources of truth to shuffle through: literature (in dashcoeff), dict_builder, and user.

You've hit upon my unsolved case. :-) Yes, I think a disp-bearing fctl defined in input should be "custom". I have a test case, still making it work.

loriab:
Thanks for all the comments, @PeterKraus and @hokru. I feel like I understand dispersion-handling again, and I'm glad it seems like an improvement or more easily extensible to you, too.

FYI, this needed a rebase after Jonathon's findif went it. I'm planning to wait until after #1168 (which will also trigger a rebase) goes in Wednesday to rebase and make this thing pass.

psi4/driver/qcdb/intf_dftd3/dashparam.py
PeterKraus: Thanks for clarifying. I suppose now that your thought process is documented on github, we'll be able to remember what's going on once we forget.

dsirianni:
@hokru We've started added DFTD3-ABC in https://github.com/loriab/psi4/pull/7.  While testing on ethene--ethyne, we got the following values for the E6(ABC) dispersion correction (in Eh):
```
BP86-D3 = -0.000000175296
BP86-D3m = -0.000000175296
BP86-D3BJ = -0.000869639424
BP86-D3MBJ = -0.000869639424
```
Do you happen to know if they look reasonable, especially that `-D3` is different than `-D3BJ`, since the DFTD3 manual mentions using the same zero damping for the ABC terms?

hokru:
You should always get the same value, with both zero and bj damping.
Does dftd3 or psi4 give you these results?

loriab:
It's `dftd3` run through the psi4 interface. All we've done is tack on `-abc` to the command and then parse the `E6(abc)` line. @dsirianni, maybe page up in your terminal and check (or re-run) d3zero vs d3bj that the output line is exactly the same (for our regex to handle). Another thing to test is whether `dftd3 -func bp -abc <molfile>` outright gives exactly the same thing (`bp` is the turbomole for bp86, iirc). you can grab the text of the molfile from that big dictionary that gets printed with verbose=4. That will make sure that the ATM value when run from `.dftd3param` file (which is what psi always does) matches when `dftd3` uses its internal parameters for bp86.

Thanks for the help, @hokru. I was really hoping those four numbers were supposed to be the same.

hokru:
For BP86 it is: `dftd3 <molfile> -func b-p -bj -abc`

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,594 @@
...
+
+## ==> Dispersion Aliases and Parameters <== ##
+
+# The dashcoeff dict below defines the -D parameters for most of the DFT methods. Some 'd2' are
+#   taken from already defined functionals in psi4. Other parameters taken from the
+#   references indicated in the "citation" parameter. The remainder of the parameters are
+#   from http://toc.uni-muenster.de/DFTD3/ on September 25, 2012, with the dict keys
+#   translated from Turbomole to Psi4 functional names.
+#   Note: most DSD-functionals have their dispersion correction defined in dict_dh_funcs.py
+dashcoeff = {
```
dgasmith: I really like metadata approaches for requirements like this. Big +1

loriab:
Ok, I've cleared up the last two missing items.
- [x] user dft dicts `dft_functional={..., 'dispersion': {...}}` now work. they go through the same route, but the pieces are treated a little differently. "params" are the final say -- the user option `DFT_DISPERSION_PARAMETERS` has no effect. "name" and "citation" aren't "blessed" like the official ones, so they're not going to get printed unless they match official ones. I think that's reasonable -- don't want to be printing citations unquestioned.
- [x] now tracing citations again. see below for examples. always printing the disp level citation. printing the dash param citation if available (not listed for many common functionals. No opportunity for user-defined citation. It can be put it, but I didn't see the point of trusting unverified values.
- [x] put in a few more checks that'll catch some problems when adding new official fctl dictionaries.
- [x] renamed some files based on slack discussion.

```
================================================================================
   => DLDF+D09: Empirical Dispersion <=

    Podeszwa and Szalewicz Dispersion Correction
    Pernal, K.; Podeszwa, R.; Patkowski, K.; Szalewicz, K. (2009), Phys. Rev. Lett., 103: 263201

        s6 =       1.000000


================================================================================
   => OTPSS-D: Empirical Dispersion <=

    Grimme's -D2 Dispersion Correction
    Grimme, S. (2006), J. Comp. Chem., 27: 1787-1799
    Parametrisation from:    L. Goerigk, S. Grimme, J. Chem. Theory. Comput. 6, 107-126, 2010

        s6 =       1.000000
    alpha6 =      20.000000
       sr6 =       1.150000


================================================================================
   => Custom: Empirical Dispersion <=

    Grimme's -D2 Dispersion Correction
    Grimme, S. (2006), J. Comp. Chem., 27: 1787-1799

        s6 =       2.750000
    alpha6 =      20.000000
       sr6 =       1.100000


================================================================================
```

Also been rebased. Windows is failing b/c it's downloading py27. I'll have to investigate why. In particular, 7921cf1 and 2a273ca ready for review.

psi4/driver/procrouting/empirical_dispersion.py
```diff
@@ -25,260 +25,243 @@
...
+            return dashd_part
         else:
             return self.disp.compute_gradient(molecule)
 
     def compute_hessian(self, molecule):
-        """
-        #magic (if magic was easy)
-        """
+        """Compute dispersion Hessian based on engine, dispersion level, and parameters in `self`.
+        Uses finite difference, as no dispersion engine has analytic second derivatives.
```
PeterKraus: This presumably means that we won't (ever) have analytic 2nd derivatives for `-D3` functionals, as dftd3 doesn't support it?

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,608 @@
...
+            disp_params = dashcoeff[dashlevel_candidate_1]['definitions'][name_hint]['params']
+        elif name_hint == '':
+            dashlevel_candidate_2 = None
+            name_key = None
+            disp_params = {}
+        else:
+            dashlevel_candidate_2 = None
+            raise ValidationError("""Can't guess -D correction level from ({})""".format(name_hint))
+
+    disp_params = copy.deepcopy(disp_params)
```
PeterKraus: I don't follow?

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,608 @@
...
+
+    if disp_params.keys() != allowed_params:
+        raise ValidationError('Requested keys ({}) insufficient ({}) for dispersion level ({})'.format(
+            list(param_tweaks.keys()), list(allowed_params), dashleveleff))
+
+    # << 3 >> use final dashlevel and disp_params to determine if a defined "fctl-disp" label exists
+    # * if/elif chooses right label when some fctls have identical param sets
+    if ((name_hint is not None) and (dashcoeff_supplement is not None)
+            and (name_key in dashcoeff_supplement[dashleveleff]['definitions'])
+            and (disp_params == dashcoeff_supplement[dashleveleff]['definitions'][name_key]['params'])):
```
PeterKraus: Is this line here why you needed the `deepcopy` above - so that you don't overwrite `supplement` when you modify `disp_params`?

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,608 @@
...
+            list(param_tweaks.keys()), list(allowed_params), dashleveleff))
+
+    disp_params.update(param_tweaks)
+
+    if disp_params.keys() != allowed_params:
+        raise ValidationError('Requested keys ({}) insufficient ({}) for dispersion level ({})'.format(
+            list(param_tweaks.keys()), list(allowed_params), dashleveleff))
+
+    # << 3 >> use final dashlevel and disp_params to determine if a defined "fctl-disp" label exists
+    # * if/elif chooses right label when some fctls have identical param sets
```
PeterKraus: Perhaps a note on citation handling done here would be cool too.

psi4/driver/qcdb/intf_dftd3/dashparam.py
```diff
@@ -0,0 +1,608 @@
...
+            if disp_params == params['params']:
+                fctldasheff = '-'.join([func, dashcoeff[dashleveleff]['formal'].lower()])
+                citeff = params.get('citation', '')
+                break
+        else:
+            fctldasheff = ''
+            citeff = ''
+
+    # TODO right now fctldasheff is empty if undefined. use '-dash' or 'custom dash' instead?
+    # TODO right now citation is empty if undefined. remove key or use None or False instead?
```
PeterKraus: I think leaving both empty is fine. The first one should be always checked against `get_dispersion_aliases`, and the second one is only used for print, so we get a newline at worst.

psi4/driver/molutil.py
```diff
@@ -312,7 +313,7 @@ def geometry(geom, name="default"):
             import pylibefp
         except ImportError as e:  # py36 ModuleNotFoundError
             raise ImportError("""Install pylibefp to use EFP functionality. `conda install pylibefp -c psi4` Or build with `-DENABLE_libefp`""") from e
-        #print('pylibefp (found version {})'.format(pylibefp.__version__))
+        #print('Using pylibefp: {} (version {})'.format(pylibefp.__file__, pylibefp.__version__))
```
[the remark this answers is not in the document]
loriab: Yes. I uncomment it pretty often when I want to see what's going on, but for the most part, don't want the pylibefp version every time a Mol has an EFP part.

psi4/driver/procrouting/dft/dft_builder.py
```diff
@@ -71,31 +71,31 @@
      "dispersion":  {          definition of dispersion corrections
                "type": "",       dispersion type - "d2", "d3zero", "d3bj" etc., see empirical_dispersion.py
              "params": {},       parameters for the dispersion correction
-           "citation": "",       special reference for the dispersion correction, appended to output
+           "citation": "",       special reference for the dispersion correction parameters, appended to output
```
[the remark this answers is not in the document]
loriab: good catch

psi4/driver/procrouting/empirical_dispersion.py
loriab: I haven't heard that analytic Hessians are on anyone's agenda. My impression is that findif H by G is pretty stable. But if any engine wants to support them, glad to change the EmpiricalDispersion class to accommodate. Is that what you were asking?

psi4/driver/qcdb/intf_dftd3/dashparam.py
loriab: Yes. The `disp_params` on rhs of 535 are still pointing to `dashcoeff` or `dashcoeff_supplement` objects. Because want to manipulate here and in 564 `disp_params.update(param_tweaks)`, best to grab an independent copy.

psi4/driver/qcdb/intf_dftd3/dashparam.py
loriab: Yes, it's fine for here, I think, too, since this is the as-much-as-can-be-validated form of the data. When it gets to the QCSchema level, the maximal-known-fields form may be preferred.

doc/sphinxman/source/dft.rst
```diff
@@ -724,28 +724,83 @@ The "best-practice" input file for KS-DFT is::
...
+Supported keywords include:
+
+ - `name`: string, name of the functional. for custom defined functionals used for printing only.
+ - `xc_functionals`: dict, definition of a complete (X + C) functional based in LibXC name
+ - `x_functionals`: dict, definition of exchange functionals using LibXC names
+ - `c_functionals`: dict, definition of correlation functionals using LibXC names
+ - `x_hf`: dict, parameters dealing with exact (HF) exchange settings for hybrid DFT
+ - `c_mp2`: dict, parameters dealing with MP2 correlation for double hybrid DFT
+ - `dispersion`: dict, definition of dispersion corrections
+ - `citation`: string, citation for the method, for printing purposes
```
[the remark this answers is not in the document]
loriab: Perhaps inconsistently, the main (as opposed to dispersion) citation is handled elsewhere. So I think even user-spec'd citation will show up in output.

psi4/driver/molutil.py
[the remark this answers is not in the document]
PeterKraus: I was wondering if this isn't something one would want to see in higher `verbose` levels.

psi4/driver/procrouting/empirical_dispersion.py
PeterKraus: No, but it answers my question. :+1:

Issue #1144, opened 2018-08-20
format c-side codebase


loriab:
Due to a lull in impending c-side changes, we’d like to take the opportunity to auto-format the c-side codebase according to [the existing clang-format file](https://github.com/psi4/psi4/blob/master/.clang-format). You (and we, reviewing and merging) have until Friday, 24 Aug to get code submitted for review/merged. After that, we will be making per-module PRs of formatted changes. None of these will be merged until Wed, 29 Aug. You have until then to indicate (on the PR) that this will interrupt your development. If that happens, the PR can be closed and postponed to accommodate you. Unchallenged PRs will be merged over the week. Modules to be skipped include occ/dfocc/findif/optking/interface_dmrg because of outstanding PRs, imminent demise, or need to compare easily with upstream plugin

- [x] adc #1152
- [x] ambit_interface #1172 
- [x] ccdensity -- delay to 29 Aug #1202
- [x] ccenergy -- delay to 29 Aug #1203
- [x] cceom #1204
- [x] cchbar #1205
- [x] cclambda -- delay to 29 Aug #1206
- [x] ccresponse #1207
- [x] cctransort #1208
- [x] cctriples -- ~delay to 29 Aug (wait for #1199) #1209~ #1252 
- [x] dcft #1154
- [x] detci #1155
- [x] dfep2 #1157 
- [x] dfmp2 #1159
- [x] dfocc -- mostly done in #797, touch-ups in #1392
- [x] dmrg -- turning off formatting so can diff with plugin in CheMPS2 repo #1392
- ~findif~
- [x] fisapt #1175
- [x] fnocc #1158
- [x] gdma_interface #1166
- [x] lib3index ~#1169 (wait for #1200)~ #1389
- [x] libciomr #1173
- [x] libcubeprop -- delay to 29 Aug (wait for #1138) #1210 
- [x] libdiis #1167
- [x] libdisp #1168
- [x] libdpd -- ~delay to 29 Aug (wait for #1199) #1211~ #1251 
- [x] libfilesystem #1174
- [x] libfock ~#1164 (wait for #1224 and for @dgasmith )~ #1397
- [x] libfunctional -- ~delay to 29 Aug #1195 (wait for @dgasmith)~ #1396
- [x] libiwl #1163 
- [x] libmints -- delay to 29 Aug #1156
- [x] libmoinfo #1165
- [x] liboptions #1187
- [x] libplugin #1190
- [x] libpsi4util #1176
- [x] libpsio #1170
[...]

PR #1170, opened 2018-08-24
fmt: clang-format libpsio

andysim:
## Description
clang-format libpsio

## Questions
- [ ] If merging this PR will hinder your research b/c you have local edits, protest here or in #1144

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libpsio/aio_handler.cc
+++ b/psi4/src/psi4/libpsio/aio_handler.cc
@@ -43,475 +43,439 @@
 
 namespace psi {
 
-AIOHandler::AIOHandler(std::shared_ptr<PSIO> psio)
-    : psio_(psio)
-{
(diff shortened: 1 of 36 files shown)


JonathonMisiewicz:
At risk of encouraging procrastination, the alignment of error codes in `config.h` has been lost.

loriab:
reformat and disable/enable if you like for `config.h`, @JonathonMisiewicz, but that particular loss of alignment doesn't bother me. the nice thing (for me, b/c I enjoy formatting too much) is that yapf/c-f put a slight barrier to opt-ing in to hand formatting.

loriab:
good catch, though. I tried to scan the changes with `git add -p` so that I could catch and undo and disable autoformatting of data structures, for instance. but it helps to have more eyes on the `diff`s.

PR #1200, opened 2018-08-31
Fix OpenMP for Windows in lib3index

raimis:
## Description
This is part of *Psi4* porting to Windows (#933).

~~MSVC supports only OpenMP 2.0, but *Psi4* needs higher at some parts.~~

*clang-cl* supports OpenMP 4.5!

## Todos
- [x] Clean up OpenMP directives
- ~~Make threaded loop indices signed~~
- ~~Conditional compilation for `simd`~~
- [x] Add missing `for` clauses
-  ~~Replace `max` reduction~~

## Questions
- ~~In some cases `size_t` (`unsigned long int`) are replaced with `int`. Do I need to use `long` to prevent potential overflows in any case? --> Keep `size_t` or change to `long`~~
- [x] In some cases `omp_get_thread_num()` is called in a loop, which is inefficient. --> Insignificant
- [x] Some OpenMP blocks seems to be missing `for` clause. --> Add the clauses

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


jgonthier:
I think it is safer to replace `size_t` by `long int` because in some cases `int` may overflow. I haven't looked at the specific cases in this PR though.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -278,35 +278,34 @@ void DFHelper::prepare_sparsity() {
...
-                for (nu = 0; nu < numnu; ++nu) {
-                    onu = primary_->shell(NU).function_index() + nu;
+            for (int mu = 0; mu < nummu; ++mu) {
+                int omu = primary_->shell(MU).function_index() + mu;
+                for (int nu = 0; nu < numnu; ++nu) {
+                    int onu = primary_->shell(NU).function_index() + nu;
                     if (omu >= onu) {
-                        index = mu * (numnu * nummu * numnu + numnu) + nu * (nummu * numnu + 1);
-                        val = fabs(buffer[rank][index]);
+                        int index = mu * (numnu * nummu * numnu + numnu) + nu * (nummu * numnu + 1);
```
dgasmith: @jgonthier makes a great point, this line will begin to fail for medium to large systems. Keeping this as `size_t` anywhere that does not interact with OMP would be preferable. Cannot believe I missed this.

raimis:
@dgasmith

What is the point of this OpenMP block? `nthreads_` threads are started, where each one is doing the same operations, i.e. no parallelization.

https://github.com/psi4/psi4/blob/a91fd6c8edf809fc9cd1afd070a6c0040ab4603a/psi4/src/psi4/lib3index/dfhelper.cc#L2524-L2531

Shouldn't it contain `for` clause?
```c++
 #pragma omp parallel for num_threads(nthreads_) 
```

schiebermc:
@raimis You are correct! That is a typo.  To my knowledge these transpose functions are never used and the complexity O(N^3) was low enough to allow the mistake to go unnoticed.  Although I only see 10 blocks, where are the other two?

schiebermc:
Ah I see them now around 2197 and 2287.  Those two are definitely used and all of the incorrect lines need the said `for`.

raimis:
@schiebermc I have added the missing `for` clauses.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -1886,16 +1881,15 @@ void DFHelper::transform() {
 void DFHelper::first_transform_pQq(size_t nao, size_t naux, size_t bsize, size_t bcount, size_t block_size,
     double* Mp, double* Tp, double* Bp, std::vector<std::vector<double>>& C_buffers){
 
-    size_t rank = 0;
-
     // perform first contraction on pQq, thread over p.
-    #pragma omp parallel for firstprivate(nao, naux, bsize, block_size, rank) schedule(guided) num_threads(nthreads_)
-    for (size_t k = 0; k < nao_; k++) {
```
schiebermc: The `firstprivate` here with `nao`, `naux`, `bsize` and `block_size` guaranteed all threads got their own copy. Not sure how much (if at all) this affects latency of memory fetches, but that's why it was used.   The cleanup of most other pragmas is definitely warranted.  Of course defining the variables within the loops makes them private. Just a note.

schiebermc: Also worth mentioning this is one of the main computational kernels.

raimis: Which tests could used to measure the impact to the performance?

raimis: Also `nao` and `naux` are not even used in the OpenMP block!

schiebermc: touché!  The tests would require timings of the different versions.. but it might be easiest just to keep it in.  Maybe someone (possibly at GT) can play with some of these ideas and get the best configuration.  This is the only `private` or `firstprivate` that caught my eye as being useful.

schiebermc:
@raimis I think you are correct about this, although I suspect the performance tradeoff is likely trivial compared to other things.  I believe this switch would be better practice in Psi4, but let @dgasmith decide on this one.  Good catch!

raimis:
@dgasmith @schiebermc

As discussed in https://github.com/psi4/psi4/issues/933#issuecomment-423183802, *clang-cl* supports *OpenMP* 4.0, so the adaptation to *OpenMP* 2.0 is not necessary.

So this PR just fixes some missing `for` clauses and clean up *OpenMP* pragmas.

PR #1233, opened 2018-09-07
DFT Grid Cache and VV10 Gradients

dgasmith:
## Description
This PR is intended to implement VV10 gradients and cache collocation matrices to prevent reconstruction every iteration.

The VV10 gradient is on hold until I build a new Grid that uses only atom-centered blocks rather than blocks that possibly contain functions from multiple centers. This is due to the fact that the VV10 nuclear gradient has a component which requires a gradient with respect to centers and not just with respect to the density as there are explicit `1/r` dependance in the VV10 energy.

## Todos
- [x] Cache collocation matrices to remove DFT bottleneck for small molecules.
- [x] Figure out how many blocks to store based on available memory.
- [ ] Implement VV10 gradients
  - [x] Orbital VV10 gradient piece implemented
  - [ ] Nuclear VV10 gradient piece implemented
- [ ] A DFT Grid that uses only Atom centered blocks.

## Questions
- [ ] Anyone have thoughts on how to figure out how much memory JK and grid caching should receive?

Timing for a small molecule without a cache using TPSS (a MGGA):
```
| UV: Form V                        :     22.583u      1.117s     13.702w     15 calls
| | Properties                      :     12.333u      0.883s      7.756w  10530 calls
| | Functional                      :      5.617u      0.033s      3.217w  10530 calls
| | V_xc                            :      4.400u      0.117s      2.567w  10530 calls
```

With a cache:
```
| UV: Form V                        :     13.750u      0.967s      8.316w     15 calls
| | Properties                      :      4.083u      0.783s      2.422w  10530 calls
| | Functional                      :      4.950u      0.000s      3.186w  10530 calls
| | V_xc                            :      4.383u      0.100s      2.564w  10530 calls
```

Note DFT V matrix takes a good portion of JK time up to ~2k basis functions with our DF code, though this depends on the density of the grid. Cache sizes:

Grid (99, 302)
```
LDA - 6MB/atom
GGA - 24MB/atom
MGGA - 60MB/atom
```

Grid (120, 590)
```
LDA - 16MB/atom
GGA - 48MB/atom
MGGA - 160MB/atom
```

A cache of every block point seems to make the computation 40-80% faster.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


dgasmith:
This is ready for review. I doubt I will get VV10 gradients in this PR as it will need a deep overhaul to the MolecularGrid class.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -104,13 +104,58 @@ def scf_compute_energy(self):
...
+        jk_size = total_memory * 0.1
+    elif scf_type.endswith('DF'):
+        jk_size = naux * nbf * nbf
+    else:
+        jk_size = nbf ** 4
+
+    # Give remaining to collocation
+    if total_memory > jk_size:
+        collocation_memory = total_memory - jk_size
+    # Give up to 10% to collocation
```
andysim: This looks like a good heuristic to me - it safely accounts for JK, which is good.

psi4/src/psi4/libfock/v.cc
```diff
@@ -400,6 +438,124 @@ double VBase::vv10_nlc(SharedMatrix D, SharedMatrix ret) {
...
+        //     zcontrib += z_grid[i];
+        // }
+        for (int P = 0; P < npoints; P++) {
+            std::fill(Tp[P], Tp[P] + nlocal, 0.0);
+            C_DAXPY(nlocal, z_grid[P], phi[P], 1, Tp[P], 1);
+        }
+        for (int ml = 0; ml < nlocal; ml++) {
+            int A = primary_->function_to_center(function_map[ml]);
+            // Gp[A][0] += C_DDOT(npoints, &Tp[0][ml], max_functions, &phi_x[0][ml], max_functions);
+            // Gp[A][1] += C_DDOT(npoints, &Tp[0][ml], max_functions, &phi_y[0][ml], max_functions);
```
andysim: i'm guessing the x and y terms are to be uncommented eventually?

psi4/src/psi4/libfock/v.cc
```diff
@@ -1156,6 +1085,10 @@ SharedMatrix RV::compute_gradient() {
     for (size_t i = 0; i < num_threads_; i++) {
         point_workers_[i]->set_deriv(old_deriv);
     }
+    if (functional_->needs_vv10()) {
+        // throw PSIEXCEPTION("V: RKS cannot compute VV10 gradient contribution.");
```
andysim: i think this throw should be reactivated, until the gradients are implemented in the followup PR

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -801,84 +807,124 @@ std::map<std::string, SharedVector> SuperFunctional::compute_vv10_cache(
...
-            double* r_rho = r_block["RHO"]->pointer();
-            double* r_W0 = r_block["W0"]->pointer();
-            double* r_kappa = r_block["KAPPA"]->pointer();
+            const double* r_x = r_block["X"]->pointer();
+            const double* r_y = r_block["Y"]->pointer();
+            const double* r_z = r_block["Z"]->pointer();
+            const double* r_w = r_block["W"]->pointer();
+            const double* r_rho = r_block["RHO"]->pointer();
+            const double* r_W0 = r_block["W0"]->pointer();
+            const double* r_kappa = r_block["KAPPA"]->pointer();
```
andysim: These pointers may have to be marked `RESTRICT` to get the SIMD benefits below

andysim: oops, they're const so it should be fine

psi4/src/psi4/libfock/v.cc
dgasmith: Yea, it is all wildly incorrect until we can isolate blocks to single basis functions. Noted this in the code.

psi4/src/psi4/libfock/v.cc
dgasmith: Thanks, now throws again.

psi4/driver/procrouting/scf_proc/scf_iterator.py
dgasmith: Good deal, I figure we can go back and tune the percentages later if someone feels up for trying a few scenarios.

psi4/src/psi4/libfunctional/superfunctional.cc
dgasmith: The opt report shows that this is properly vectorized. Still getting this in VTune would be worthwhile checking- it looks like an easy vectorization, but isn't always the case.

psi4/src/psi4/libfock/cubature.h
```diff
@@ -156,7 +156,7 @@ class MolecularGrid {
     /// Maximum number of funtions in a block
     int max_functions() const { return max_functions_; }
     /// Total collocation size of all blocks
-    size_t collocation_size() { return collocation_size_; }
+    const size_t collocation_size() { return collocation_size_; }
```
andysim: sorry, I mean the function should be const (doesn't change any members), not the return value (which is returned by value anyway, so the const doesn't apply)

dgasmith: Oh, sure I thought that was a bit weird. We need to const a very large number of functions in Psi4. Is it worth doing it for this one off case without changing all of cubature or libfock?

andysim: Yeah, there are many const qualifiers missing for sure.  Was just pointing it out so that we can try to add correct code going forwards, but it's not a big deal if we skip this one.  Just reviewing what I see, that's all

psi4/src/psi4/libfock/dft_integrators.h
```diff
@@ -0,0 +1,331 @@
...
+            C_DGEMM('T', 'N', nlocal, nlocal, npoints, 1.0, phiw[0], coll_funcs, Tp[0], max_functions, 1.0, V2p[0],
+                    max_functions);
+        }
+        // parallel_timer_off("Meta", rank);
+    }
+}
+
+inline void rks_gradient_integrator(std::shared_ptr<BasisSet> primary, std::shared_ptr<BlockOPoints> block,
+                                    std::shared_ptr<SuperFunctional> fworker, std::shared_ptr<PointFunctions> pworker,
+                                    SharedMatrix G, SharedMatrix U, int ansatz = -1) {
```
loriab: this been `c-f`ed? may as well do the file from the start.

psi4/src/psi4/dfmp2/corr_grad.cc
```diff
@@ -113,7 +113,7 @@ void DFCorrGrad::print_header() const {
 
         outfile->Printf("    OpenMP threads:    %11d\n", nthreads_);
         outfile->Printf("    Integrals threads: %11d\n", df_ints_num_threads_);
-        outfile->Printf("    Memory (MB):       %11ld\n", (memory_ * 8L) / (1024L * 1024L));
+        outfile->Printf("    Memory [MiB]:      %11ld\n", (memory_ * 8L) / (1024L * 1024L));
```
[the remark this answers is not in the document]
dgasmith: Sure.

psi4/src/psi4/libfock/dft_integrators.h
dgasmith: Yes, this was `c-f`ed.

loriab:
Bad luck -- opt13 hits memory problems. It passes the "in-core AOs" check with 200 kb to spare (14042044 and 14074472 below are need and supplied in ints), then runs out of memory for AO blocking. The result of the print statement in second snippet is `AAAA    1:   14176380 > 14074472 ||    3 ==   65`. Should we put padding back or go another route?

```
  ==> Integral Setup <==

  DFHelper Memory: AOs need 0.104621 GiB; user supplied 0.104863 GiB. 14042044 14074472
Using in-core AOs.


  File "stage/bin/psi4", line 271, in <module>
    exec(content)
  File "<string>", line 43, in <module>
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/driver.py", line 1113, in optimize
    G, wfn = gradient(lowername, return_wfn=True, molecule=moleculeclone, **kwargs)
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/driver.py", line 639, in gradient
    wfn = procedures['gradient'][lowername](lowername, molecule=molecule, **kwargs)
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/procrouting/proc.py", line 2068, in run_scf_gradient
    ref_wfn = run_scf(name, **kwargs)
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/procrouting/proc.py", line 2006, in run_scf
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/procrouting/proc.py", line 1360, in scf_helper
    e_scf = scf_wfn.compute_energy()
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/procrouting/scf_proc/scf_iterator.py", line 84, in scf_compute_energy
    self.initialize()
  File "/home/psilocaluser/gits/hrw-release/objdir37/stage/lib/psi4/driver/procrouting/scf_proc/scf_iterator.py", line 178, in scf_initialize
    self.initialize_jk(self.memory_jk_)

RuntimeError:
Fatal Error: DFHelper: not enough memory for (p shell) AO blocking! required memory: 0.105622[GiB].
Error occurred in file: /home/psilocaluser/gits/hrw-release/psi4/src/psi4/lib3index/dfhelper.cc on line: 568
```

```
@@ -556,8 +557,11 @@ std::pair<size_t, size_t> DFHelper::pshell_blocks_for_AO_build(const size_t mem,
         constraint = total;
         constraint += full_3index;
         constraint += (hold_met_ ? naux_ * naux_ : total);
+        printf("AAAA %4d: %10d > %d || %4d == %4d\n", count, constraint, mem, i, pshells_-1);
         if (constraint > mem || i == pshells_ - 1) {
+            printf(" AAA %d || %d\n", constraint > mem, i == pshells_ - 1);
             if (count == 1 && i != pshells_ - 1) {
+                printf("  AA %d && %d\n", count == 1, i != pshells_ - 1);
                 std::stringstream error;
                 error << "DFHelper: not enough memory for (p shell) AO blocking!"
                       << " required memory: " << constraint * 8 / (1024 * 1024 * 1024.0) << "[GiB].";
```

loriab:
Nevermind, possibly. I had tried one increase of user memory and it still failed but when I double it, it works fine. So this test can be easily fixed.

But the real question is are we willing to have more jobs pass through the in-core/disk choice stage with a false negative and instead hit the error further on. I'm thinking that's not good.

PR #1235, opened 2018-09-09
Rework of CBS to use dictionaries

PeterKraus:
## Description
The general idea is to allow more flexible definition of CBS functions, and clean up the `cbs()` function a little. The goal is to allow further corrections to the energy into CBS, eg differences of basis sets (eg. `\Delta E_{diff} = E(aug-cc-pvtz) - E(cc-pvtz)`) or frozen core corrections (`\Delta E_{ae} = E(freeze_core True) - E(freeze_core False)` or DKH ...

Added `frequency` to cbs-compatible calls.

## Todos
- [x] Simple, dict-based interface to CBS.
- [x] Handling of an arbitrary number of correction functions.
- [x] Support for passing of arbitrary options (such as `freeze_core`). 
- [x] Fix `pywrap-cbs1` which fails on `cbs_number` checks.
- [x] Documentation.
- [x] Frequency calculations via cbs are now possible.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/cbs.rst
+++ b/doc/sphinxman/source/cbs.rst
@@ -191,24 +191,6 @@ existing examples are below.
 
 .. autofunction:: psi4.driver.aliases.allen_focal_point
 
-Other helper functions
-^^^^^^^^^^^^^^^^^^^^^^
-
-.. autofunction:: psi4.driver.driver_cbs._expand_bracketed_basis
-
-.. autofunction:: psi4.driver.driver_cbs._contract_bracketed_basis
-
-.. autofunction:: psi4.driver.driver_cbs._process_cbs_kwargs
-
-.. autofunction:: psi4.driver.driver_cbs._expand_scheme_orders
-
-.. autofunction:: psi4.driver.driver_cbs._contract_scheme_orders
-
-.. autofunction:: psi4.driver.driver_cbs._cbs_wrapper_methods
-
-.. autofunction:: psi4.driver.driver_cbs._parse_cbs_gufunc_string
-
-.. autofunction:: psi4.driver.driver_cbs._cbs_gufunc
 
 
 
--- a/doc/sphinxman/source/cbs.rst
+++ b/doc/sphinxman/source/cbs.rst
@@ -43,37 +43,41 @@ Complete Basis Set
 
    cbs_eqn
 
-.. codeauthor:: Lori A. Burns and Daniel G. A. Smith
-.. sectionauthor:: Lori A. Burns
+.. codeauthor:: Lori A. Burns, Daniel G. A. Smith and Peter Kraus
+.. sectionauthor:: Lori A. Burns and Peter Kraus
 
 The :py:func:`psi4.cbs` function described below is
 powerful but complicated, requiring many options. For most common
 calculations, a shorthand can be accessed directly though
 :py:func:`psi4.energy`, :py:func:`psi4.gradient`, *etc.* For example,
 a MP2 single-point DT extrapolation can be accessed through the first item
-below more conveniently than the equivalent second item.
+below more conveniently than the equivalent second or third items.
 
 * ``energy('mp2/cc-pv[dt]z')``
 
 * ``energy(cbs, corl_wfn='mp2', corl_basis='cc-pv[dt]z')``
 
+* ``energy(cbs, cbs_metadata=[{"wfn": "hf", "basis": "cc-pvtz"}, {"wfn": "mp2", "basis": "cc-pv[dt]z"}])``
+
 A CCSD(T) DT coupled-cluster correction atop a TQ MP2 extrapolation
-geometry optimization can be accessed through the first item below more
-conveniently than the equivalent second item.
+geometry optimization can also be accessed through the first item below more
+conveniently than the equivalent second and third items.
 
 * ``optimize('mp2/cc-pv[tq]z + D:ccsd(t)/cc-pvdz')``
 
 * ``optimize(cbs, corl_wfn='mp2', corl_basis='cc-pv[tq]z', delta_wfn='ccsd(t)', delta_basis='cc-pvdz')``
 
+* ``optimize(cbs, cbs_metadata=[{"wfn": "hf", "basis": "cc-pvqz"}, {"wfn": "mp2", "basis": "cc-pv[tq]z"}, {"wfn": "ccsd(t)", "basis": "cc-pvdz"}])``
+
 Many examples can be found at :srcsample:`cbs-xtpl-energy`,
 :srcsample:`cbs-xtpl-gradient`, :srcsample:`cbs-xtpl-opt`,
 :srcsample:`cbs-xtpl-freq`, :srcsample:`cbs-xtpl-func`,
-:srcsample:`cbs-xtpl-wrapper`.
+:srcsample:`cbs-xtpl-wrapper`, :srcsample:`cbs-xtpl-dict`.
 
 
-.. autofunction:: psi4.cbs(name [, scf_basis, scf_scheme, corl_wfn, corl_basis, corl_scheme, delta_wfn, delta_wfn_lesser, delta_basis, delta_scheme, delta2_wfn, delta2_wfn_lesser, delta2_basis, delta2_scheme, delta3_wfn, delta3_wfn_lesser, delta3_basis, delta3_scheme, delta4_wfn, delta4_wfn_lesser, delta4_basis, delta4_scheme, delta5_wfn, delta5_wfn_lesser, delta5_basis, delta5_scheme])
+.. autofunction:: psi4.cbs(name [, scf_basis, scf_scheme, corl_wfn, corl_basis, corl_scheme, delta_wfn, delta_wfn_lesser, delta_basis, delta_scheme, delta2_wfn, delta2_wfn_lesser, delta2_basis, delta2_scheme, cbs_metadata])
 
-.. note:: Presently (May 2016), only two of the five delta possibilities are active. Also, temporarily extrapolations are performed on differences of target and scf total energies, rather than on correlation energies directly. This doesn't affect the extrapolated values of the particular formulas defined here (though it does affect the betas, which are commented out), but it is sloppy and temporary and could affect any user-defined corl extrapolations.
+.. note:: As of October 2018, only two explicit ```deltaN_[wfn,basis,scheme]``` sets of options are active; if more delta functions are required, use the ```cbs_metadata``` interface. Also, temporarily extrapolations are performed on differences of target and scf total energies, rather than on correlation energies directly. This doesn't affect the extrapolated values of the particular formulas defined here (though it does affect the betas, which are commented out), but it is sloppy and temporary and could affect any user-defined corl extrapolations.
 
 .. index::
    pair: cbs(); output
--- a/doc/sphinxman/source/cbs.rst
+++ b/doc/sphinxman/source/cbs.rst
@@ -167,9 +167,15 @@ Extrapolation Schemes
 
 .. autofunction:: psi4.driver.driver_cbs.scf_xtpl_helgaker_2
 
+.. autofunction:: psi4.driver.driver_cbs.scf_xtpl_truhlar_2
+
+.. autofunction:: psi4.driver.driver_cbs.scf_xtpl_karton_2
+
 .. autofunction:: psi4.driver.driver_cbs.scf_xtpl_helgaker_3
 
 .. autofunction:: psi4.driver.driver_cbs.corl_xtpl_helgaker_2
+
+.. autofunction:: psi4.driver.driver_cbs._get_default_xtpl
 
 Aliases
 ^^^^^^^
@@ -184,4 +190,25 @@ existing examples are below.
 .. autofunction:: psi4.driver.aliases.sherrill_gold_standard
 
 .. autofunction:: psi4.driver.aliases.allen_focal_point
+
+Other helper functions
+^^^^^^^^^^^^^^^^^^^^^^
+
+.. autofunction:: psi4.driver.driver_cbs._expand_bracketed_basis
+
+.. autofunction:: psi4.driver.driver_cbs._contract_bracketed_basis
+
+.. autofunction:: psi4.driver.driver_cbs._process_cbs_kwargs
+
+.. autofunction:: psi4.driver.driver_cbs._expand_scheme_orders
+
+.. autofunction:: psi4.driver.driver_cbs._contract_scheme_orders
+
+.. autofunction:: psi4.driver.driver_cbs._cbs_wrapper_methods
+
+.. autofunction:: psi4.driver.driver_cbs._parse_cbs_gufunc_string
+
+.. autofunction:: psi4.driver.driver_cbs._cbs_gufunc
+
+
 
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1767,19 +1767,6 @@ def frequency(name, **kwargs):
     """
     kwargs = p4util.kwargs_lower(kwargs)
 
-    # Let hessian() handle this!
-    # # Bounce (someday) if name is function
-    # if hasattr(name, '__call__'):
-    #     raise ValidationError("Frequency: Cannot use custom function")
-    #
-    # lowername = name.lower()
-    #
-    # if "/" in lowername:
-    #     return driver_cbs._cbs_gufunc(frequency, name, ptype='frequency', **kwargs)
-    #
-    # if kwargs.get('bsse_type', None) is not None:
-    #     raise ValdiationError("Frequency: Does not currently support 'bsse_type' arguements")
-
     return_wfn = kwargs.pop('return_wfn', False)
 
     # are we in sow/reap mode?
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1280,14 +1280,17 @@ def hessian(name, **kwargs):
     else:
         gradient_type = 'conventional'
 
-    if gradient_type != 'conventional':
-        raise ValidationError("Hessian: Does not yet support more advanced input or custom functions.")
+    #if gradient_type != 'conventional':
+    #    raise ValidationError("Hessian: Does not yet support more advanced input or custom functions.")
 
     lowername = name.lower()
 
     # Check if this is a CBS extrapolation
+    print("hessian()")
+    print(lowername)
+    print(kwargs)
     if "/" in lowername:
-        return driver_cbs._cbs_gufunc('hessian', lowername, **kwargs)
+        return driver_cbs._cbs_gufunc(hessian, lowername, **kwargs, ptype="hessian")
 
     return_wfn = kwargs.pop('return_wfn', False)
     core.clean_variables()
@@ -1746,18 +1749,19 @@ def frequency(name, **kwargs):
 
     """
     kwargs = p4util.kwargs_lower(kwargs)
-
-    # Bounce (someday) if name is function
-    if hasattr(name, '__call__'):
-        raise ValidationError("Frequency: Cannot use custom function")
-
-    lowername = name.lower()
-
-    if "/" in lowername:
-        return driver_cbs._cbs_gufunc(frequency, name, ptype='frequency', **kwargs)
-
-    if kwargs.get('bsse_type', None) is not None:
-        raise ValdiationError("Frequency: Does not currently support 'bsse_type' arguements")
+ 
+    # Let hessian() handle this!
+    # # Bounce (someday) if name is function
+    # if hasattr(name, '__call__'):
+    #     raise ValidationError("Frequency: Cannot use custom function")
+    # 
+    # lowername = name.lower()
+    # 
+    # if "/" in lowername:
+    #     return driver_cbs._cbs_gufunc(frequency, name, ptype='frequency', **kwargs)
+    # 
+    # if kwargs.get('bsse_type', None) is not None:
+    #     raise ValdiationError("Frequency: Does not currently support 'bsse_type' arguements")
 
     return_wfn = kwargs.pop('return_wfn', False)
 
@@ -1771,7 +1775,8 @@ def frequency(name, **kwargs):
     molecule.update_geometry()
 
     # Compute the hessian
-    H, wfn = hessian(lowername, return_wfn=True, molecule=molecule, **kwargs)
+    print(name)
+    H, wfn = hessian(name, return_wfn=True, molecule=molecule, **kwargs)
 
     # S/R: Quit after getting new displacements
     if freq_mode == 'sow':
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -25,7 +25,6 @@
 #
 # @END LICENSE
 #
-
 """Module with a *procedures* dictionary specifying available quantum
 chemical methods and functions driving the main quantum chemical
 functionality, namely single-point energies, geometry optimizations,
@@ -50,7 +49,9 @@ from psi4.driver import p4util
 from psi4.driver import qcdb
 from psi4.driver.procrouting import *
 from psi4.driver.p4util.exceptions import *
+
 # never import wrappers or aliases into this file
+
 
 def _find_derivative_type(ptype, method_name, user_dertype):
     r"""
@@ -105,7 +106,7 @@ def _find_derivative_type(ptype, method_name, user_dertype):
             alternatives = """ Did you mean? %s""" % (' '.join(alt_method_name))
 
         raise ValidationError("""Derivative method 'name' %s and derivative level 'dertype' %s are not available.%s"""
-            % (method_name, str(dertype), alternatives))
+                              % (method_name, str(dertype), alternatives))
 
     return dertype
 
@@ -121,11 +122,9 @@ def _energy_is_invariant(gradient, stationary_criterion=1.e-2):
     mol = core.get_active_molecule()
     efp_present = hasattr(mol, 'EFP')
 
-    translations_projection_sound = (not core.get_option('SCF', 'EXTERN') and
-                                     not core.get_option('SCF', 'PERTURB_H') and
-                                     not efp_present)
-    rotations_projection_sound = (translations_projection_sound and
-                                  stationary_point)
+    translations_projection_sound = (not core.get_option('SCF', 'EXTERN') and not core.get_option('SCF', 'PERTURB_H')
+                                     and not efp_present)
+    rotations_projection_sound = (translations_projection_sound and stationary_point)
 
     return translations_projection_sound, rotations_projection_sound
 
@@ -425,11 +424,11 @@ def energy(name, **kwargs):
 
     """
     kwargs = p4util.kwargs_lower(kwargs)
-    
+
     # Bounce to CP if bsse kwarg
     if kwargs.get('bsse_type', None) is not None:
         return driver_nbody.nbody_gufunc(energy, name, ptype='energy', **kwargs)
-    
+
     # Bounce if name is function
     if hasattr(name, '__call__'):
         return name(energy, kwargs.pop('label', 'custom function'), ptype='energy', **kwargs)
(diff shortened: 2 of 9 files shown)


PeterKraus:
I've tried to re-write some of the cbs docs more thoroughly, however my sphinx-fu is not up to scratch. Especially the bits in `driver_cbs.py` under Combined interface, I am not sure what the best way of markup is - the current one is quite ugly.

PeterKraus:
Also, as a to-do for a different PR, it might be a good idea to unify the whole metadata structure with what's currently in `NEED`, `GRAND_NEED` and `MODELCHEM`, so that it's all in one place and not copied across 3 dictionaries.

Another wished-for feature is allowing `freq()` calls into cbs, superficially it doesn't look that hard, at least for methods where we can compute a cbs-`hessian()` already... But as I said above, it's on my radar, but not for this PR.

dgasmith:
Can you rebase this with master so that appveyor will pass.

PeterKraus:
So I've added the option to be able to run frequency calculations via the cbs interface, both using the `"method/basis"` and `frequency(cbs, ...)` syntax. I've added the tests for this too - I wanted to get a reference value for the `HF/cc-pV[DT]Z` extrapolation from other codes (thanks @hokru), but we failed to get it run correctly. When compared to large basis sets, one gets these results:

Freq 1 | Freq 2 | Freq 3 | zeta | RMS
-- | -- | -- | -- | -- 
1809.246 | 3923.151 | 4020.177 | 2 |    174.1217 
1753.040 | 4126.975 | 4226.872 | 3 |     4.1162
1750.381 | 4130.709 | 4230.122 | 4 |     1.4733
1748.147 | 4130.445 | 4230.985 | 5 |     0.2300
1748.003 | 4130.108 | 4230.827 | 6 |     
1747.741 | 4130.226 | 4230.575 | [dt] |    0.2207
1749.876 | 4131.278 | 4230.490 | [tq] |    1.2899
1747.595 | 4130.375 | 4231.193 | [q5] |    0.3520

I think the convergence is reasonable - it's clear the results (even with HF/[dt]) are not complete junk.

psi4/driver/driver.py
```diff
@@ -1746,18 +1747,19 @@ def frequency(name, **kwargs):
...
+    # if hasattr(name, '__call__'):
+    #     raise ValidationError("Frequency: Cannot use custom function")
+    # 
+    # lowername = name.lower()
+    # 
+    # if "/" in lowername:
+    #     return driver_cbs._cbs_gufunc(frequency, name, ptype='frequency', **kwargs)
+    # 
+    # if kwargs.get('bsse_type', None) is not None:
+    #     raise ValdiationError("Frequency: Does not currently support 'bsse_type' arguements")
```
PeterKraus: I've kept this piece of code for review, but it can be deleted later.

psi4/driver/driver_cbs.py
```diff
@@ -1681,6 +1488,7 @@ def cbs(func, label, **kwargs):
             finalquantity.print_out()
     elif ptype == 'hessian':
         finalquantity = finalhessian
+        wfn.set_gradient(finalgradient)
```
PeterKraus: If there is a good reason for not doing this with all `frequency(cbs, ...)` calculations, let me know - otherwise the frequency calculations fail (as the gradient is not set properly).

tests/cbs-xtpl-freq/input.dat
```diff
@@ -3,42 +3,61 @@ import numpy as np
...
 nucenergy_ref = 9.16819326039
 
-molecule h2o {
+molecule h2o_dz {
   O
   H 1 0.96
   H 1 0.96 2 104.5
 }
 
+molecule h2o_tz {
```
PeterKraus: These two were optimized using `g_convergence gau_verytight`

PeterKraus:
So the windows test fails. I am assuming it's because it's using (implicit) `dertype=2` for the scf frequency - is this a known fail on Windows? @raimis, do you have any idea?

raimis:
Also `cbs-parser` and `cbs-xtpl-func` are still failing, but I haven't yet tried to debug them. So, at the moment, no idea what might be wrong.

psi4/driver/driver_cbs.py
```diff
@@ -49,20 +49,31 @@
...
-    out-of-order zeta specification (e.g., [qtd]) and numeral for number
-    (e.g., [23]) but not skipped zetas (e.g., [dq]) or zetas outside [2,
-    8] or non-Dunning or non-Ahlrichs or non-Jensen sets or 
-    non-findable .gbs sets.
-
+    """Function to transform and validate basis series specification for cbs(). 
+    
+    Parameters
+    ----------
+    basisstring: string
```
loriab: space before `:`. sphinx is very picky

loriab: also, probably `str` rather than `string`

psi4/driver/driver_cbs.py
```diff
@@ -703,6 +893,149 @@ def return_energy_components():
...
+                else:
+                    metadata[-1]["wfn"] = item.get("wfn")
+                if "basis" not in item:
+                    raise ValidationError("Basis set has to be supplied for every stage.")
+                else:
+                    metadata[-1]["basis"] = _expand_bracketed_basis(item.get("basis").lower())
+                metadata[-1]["treatment"] = item.get("treatment", "scf" if len(metadata) == 1 else "corl")
+                metadata[-1]["scheme"] = item.get("scheme", _get_default_xtpl(len(metadata[-1]["basis"][1]),
+                                                                                 metadata[-1]["treatment"])) 
+                if len(metadata) > 1:
```
loriab: doesn't this only kick in for deltas at `len(metadata) > 2:`?

psi4/driver/driver_cbs.py
```diff
@@ -703,6 +893,149 @@ def return_energy_components():
...
+                    metadata[-1]["wfn_lo"] = item.get("wfn_lo", metadata[-2].get("wfn"))
+                    metadata[-1]["basis_lo"] = _expand_bracketed_basis(item.get("basis_lo", 
+                                                                       item.get("basis")).lower())
+                    if len(metadata[-1]["basis"][0]) != len(metadata[-1]["basis_lo"][0]):
```
loriab: I wonder if it'd be a bit cleaner to have a `trial_stage = {}` at the start, fill that instead of `metadata[-1]`, then `metadata.append(trial_stage)` at the end if all successful. Open for debate.

psi4/driver/driver_cbs.py
```diff
@@ -703,6 +893,149 @@ def return_energy_components():
...
+                        raise ValidationError("""Number of basis sets inconsistent 
+                                             between high ({0:d}) and low ({1:d}) levels.""".format(len(metadata[-1]["basis"][0]),len(metadata[-1]["basis_lo"][0])))
+                metadata[-1]["stage"] = item.get("stage", None)
+                if metadata[-1]["stage"] == None:
+                    if len(metadata) == 1:
+                        metadata[-1]["stage"] = "scf"
+                    elif len(metadata) == 2:
+                        metadata[-1]["stage"] = "corl"
+                    else:
+                        metadata[-1]["stage"] = "delta{0:d}".format(len(metadata) - 2)
```
loriab: If you're doing the `0` in `delta{0:d}` in order to specify the item in the format string to format, that was only necessary in py26 or if you want to go out-of-order. 

You may have a better reason -- I'm not totally up on format string syntax.

psi4/driver/driver_cbs.py
```diff
@@ -703,6 +893,149 @@ def return_energy_components():
...
+            metadata[-1]["alpha"] = item.get("alpha", None)
+            metadata[-1]["options"] = item.get("options", False)
```
loriab: This'll be great. I've long considered lack of stage-directed options as one of cbs()'s greatest weaknesses.

psi4/driver/driver_cbs.py
```diff
@@ -1076,23 +1365,21 @@ def cbs(func, label, **kwargs):
...
     verbose = kwargs.pop('verbose', 0)
     ptype = kwargs.pop('ptype')
-    cbs_alpha = {
-        'scf': kwargs.get('cbs_scf_alpha', kwargs.get('scf_alpha', None)),
-        'corl': kwargs.get('cbs_corl_alpha', kwargs.get('corl_alpha', None)),
-        'delta': kwargs.get('cbs_delta_alpha', kwargs.get('delta_alpha', None)),
-        'delta2': kwargs.get('cbs_delta2_alpha', kwargs.get('delta2_alpha', None)),
-    }
 
     # Establish function to call (only energy makes sense for cbs)
```
JonathonMisiewicz: This comment, and the error message two lines below, are inaccurate now.

JonathonMisiewicz:
Two comments about lines that weren't changed:
https://github.com/psi4/psi4/blob/a658c8e50e6d5d7bde2b8058b33e78f98f800db3/psi4/driver/driver_cbs.py#L1910-L1912
We shouldn't be dropping out for frequencies anymore.

https://github.com/psi4/psi4/blob/a658c8e50e6d5d7bde2b8058b33e78f98f800db3/psi4/driver/driver_cbs.py#L1940-L1944
It would be good to put a comment here explicitly saying that `_parse_cbs_gufunc_string` should enforce that `len(method_list)` can only be 1 or 2. The logic confused me for a few minutes, on first read.

psi4/driver/driver_cbs.py
```diff
@@ -703,6 +893,149 @@ def return_energy_components():
...
+            if "wfn" in item:
+                if "wfn" not in item:
+                    raise ValidationError("Wavefunction type has to be supplied for every stage.")
+                else:
+                    metadata[-1]["wfn"] = item.get("wfn")
+                if "basis" not in item:
+                    raise ValidationError("Basis set has to be supplied for every stage.")
+                else:
+                    metadata[-1]["basis"] = _expand_bracketed_basis(item.get("basis").lower())
+                metadata[-1]["treatment"] = item.get("treatment", "scf" if len(metadata) == 1 else "corl")
```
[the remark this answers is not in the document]
PeterKraus: I'm not sure I understand. The old behaviour of just specifying "corl_wfn" and higher, without specifying "scf_wfn" still works. For "cbs_metadata", the first item in the array is by default a "scf" stage, which means a scf extrapolation method will be used (as opposed to corl extrapolation) - it can (as before) be an extrapolated, or highest_1 stage, but is not restricted to "scf"/"hf"/"c4-" anymore. The extrapolation (scf vs corl) is overridable by providing the "treatment" key, the name of the stage is overridable by the "stage" key.

psi4/driver/driver_cbs.py
PeterKraus: No, the treatment of correlation (index == 1) and following delta-correlations (index >1) is essentially identical.

psi4/driver/driver_cbs.py
PeterKraus: Fine by me.

psi4/driver/driver_cbs.py
PeterKraus: Yeah - it's what motivated me to do this in the first place.

psi4/driver/driver_cbs.py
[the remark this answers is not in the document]
loriab: But is it correct that `energy(cbs, cbs_metadata=[{'wfn': 'mp2', 'basis': 'cc-pv[dt]z'}])` is no longer valid b/c no scf stage? What did you have in mind as something ppl would want other than HF (in various flavors and sources)?

PeterKraus:
> Two comments about lines that weren't changed:
> 
> [psi4/psi4/driver/driver_cbs.py](https://github.com/psi4/psi4/blob/a658c8e50e6d5d7bde2b8058b33e78f98f800db3/psi4/driver/driver_cbs.py#L1910-L1912)
> 
> Lines 1910 to 1912 in [a658c8e](/psi4/psi4/commit/a658c8e50e6d5d7bde2b8058b33e78f98f800db3)
>  # Drop out for props and freqs 
>  if ptype in ["properties", "frequency"]: 
>      raise ValidationError("%s: Cannot extrapolate or delta correct %s yet." % (ptype.title(), ptype)) 
> 
> We shouldn't be dropping out for frequencies anymore.

Fair point. I changed the raw 'frequency' function too, to be a true wrapper around 'hessian', that code is just a left-over. Changed.

> [psi4/psi4/driver/driver_cbs.py](https://github.com/psi4/psi4/blob/a658c8e50e6d5d7bde2b8058b33e78f98f800db3/psi4/driver/driver_cbs.py#L1940-L1944)
> 
> Lines 1940 to 1944 in [a658c8e](/psi4/psi4/commit/a658c8e50e6d5d7bde2b8058b33e78f98f800db3)
>  if len(method_list) > 1: 
>      cbs_kwargs['delta_wfn'] = method_list[1] 
>      cbs_kwargs['delta_basis'] = basis_list[1] 
>      if 'delta_scheme' in kwargs: 
>          cbs_kwargs['delta_scheme'] = kwargs['delta_scheme'] 
> 
> It would be good to put a comment here explicitly saying that `_parse_cbs_gufunc_string` should enforce that `len(method_list)` can only be 1 or 2. The logic confused me for a few minutes, on first read.

I'll edit the docs. The idea is that one usually only specifies one "delta correction" in a string, ie:
```
mp2/cc-pv[tq]z+D:ccsd(t)/cc-pvtz
```
as opposed to trying to do two delta corrections in a "method/basis" syntax (which is not supported):
```
hf/cc-pv[q5]z+D:mp2/cc-pv[tq]z+D2:ccsd(t)/cc-pvtz
```
Folks that want to pile on the deltas can use the old interface (with scf, corl, delta and delta2 stages) or the cbs_metadata interface, which doesn't have a limit.

psi4/driver/driver_cbs.py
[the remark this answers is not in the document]
PeterKraus: So there are a few options:
```
[1] energy("mp2/cc-pv[dt]z+d:ccsd(t)/cc-pvdz")
[2] energy(cbs, corl_wfn="mp2", corl_basis="cc-pv[dt]z", 
                delta_wfn="ccsd(t)", delta_basis="cc-pvdz")
[3] energy(cbs, cbs_metadata=[{"wfn": "hf", "basis": "cc-pvtz"},
                              {"wfn": "mp2", "basis": "cc-pv[dt]z"},
                              {"wfn": "ccsd(t)", "basis": "cc-pvdz"}])
E = hf/cc-pvtz + (mp2 - hf)/cc-pv[dt]z + (ccsd(t) - mp2)/cc-pvtz
E = -76.35375057

[4] energy(cbs, cbs_metadata=[{"wfn": "mp2", "basis": "cc-pv[dt]z"},
                              {"wfn": "ccsd(t)", "basis": "cc-pvdz"}])
E = mp2/cc-pv[dt]z + (ccsd(t) - mp2)/cc-pvdz
E = -76.35033786

[5] energy(cbs, cbs_metadata=[{"wfn": "mp2", "basis": "cc-pv[dt]z", "treatment": "corl"},
                              {"wfn": "ccsd(t)", "basis": "cc-pvdz"}])
E = mp2/cc-pv[dt]z + (ccsd(t) - mp2)/cc-pvdz
E = -76.36620899

```
The first three calls are equivalent and are the default behaviour. The last two differ from one another by calling "scf_xtpl_helgaker_2" in [4] and "corl_xtpl_helgaker_2" in [5] directly over the mp2 energies.

loriab:
I got that appveyor py27 error, too. Rebasing to bring in #1278 fixed it. No urgency, just when you're ready.

PeterKraus:
I'll rebase when I work on this next time. When you find some time, could you go through my comments in my own reviews and in yours? Also, I'm not sure what the conclusion of the "scf stage" debate above was - is it fine as it is, does it need more docs, or shall it be changed?

psi4/driver/driver_cbs.py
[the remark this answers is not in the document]
loriab: I'm separately confused by your `[1--5]`. By my common practice (and hence cbs's), `mp2/cc-pv[dt]z == hf/cc-pvtz + (mp2 - hf)/cc-pv[dt]z` with the last term using scf_xtpl_helgaker_2. So `[1--5]` should all be the same. Are you reasoning differently?

If `[4]` is, in fact, the same as `[1--3]`, then that resolves my original concern. That is, is it necessary to explicitly have SCF as the first item in the `cbs_metadata` list? Apparently not, and all's well.

If it clarifies anything, I don't think one should ever apply a correlation energy extrapolation equation to a total energy, unless compensating by subtracting out the reference energy subjected to the same (unnatural b/c eqn designed for correlation) equation is the next step.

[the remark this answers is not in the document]
PeterKraus: Yes, currently `[3]` and `[4]` are different, as in  `[1,2,3]` the energy is taken as:
```
  -------------------------------------------------------------------------------------
      Stage      Method / Basis             Wt      Energy [Eh]   Scheme
  -------------------------------------------------------------------------------------
        scf          hf / cc-pvtz            1     -76.05098620   xtpl_highest_1
       corl         mp2 / cc-pv[dt]z         1     -76.35320638   corl_xtpl_helgaker_2
       corl          hf / cc-pv[dt]z        -1     -76.06344462   corl_xtpl_helgaker_2
     delta1     ccsd(t) / cc-pvdz            1     -76.23906022   xtpl_highest_1
     delta1         mp2 / cc-pvdz           -1     -76.22605761   xtpl_highest_1
  -------------------------------------------------------------------------------------
```
In the fourth case it is:
```
  ------------------------------------------------------------------------------------
      Stage     Method / Basis             Wt      Energy [Eh]   Scheme
  ------------------------------------------------------------------------------------
        scf        mp2 / cc-pv[dt]z         1     -76.33733525   scf_xtpl_helgaker_2
       corl    ccsd(t) / cc-pvdz            1     -76.23906022   xtpl_highest_1
       corl        mp2 / cc-pvdz           -1     -76.22605761   xtpl_highest_1
  ------------------------------------------------------------------------------------
```
and in the fifth case:
```
  ------------------------------------------------------------------------------------
      Stage     Method / Basis             Wt      Energy [Eh]   Scheme
  ------------------------------------------------------------------------------------
        scf        mp2 / cc-pv[dt]z         1     -76.35320638   *corl_xtpl_helgaker_2*
       corl    ccsd(t) / cc-pvdz            1     -76.23906022   xtpl_highest_1
       corl        mp2 / cc-pvdz           -1     -76.22605761   xtpl_highest_1
  ------------------------------------------------------------------------------------
```
If you think allowing `[4,5]` is not a good idea, I can validate the `wfn` key of the first item and raise a validation error if it's not one of `HF/SCF/C4-HF/C4-SCF`.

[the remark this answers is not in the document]
loriab: I _do_ think allowing `[4,5]` is a Bad Idea. Though I'd relent if everyone disagrees.

Validating `wfn` of the first item if not `HF/C4-HF` (fine by me if the vague `SCF` gets dropped) is a good plan. My notion of a better plan would be if first item not `HF/C4-HF`, just create `{'wfn': "hf", "treatment": highest_1, 'basis': copy_corl_basis_str}` such that the first iteration through `cbs_metadata` creates _two_ entries, the scf and corl stages. Whichever you prefer.

PeterKraus:
@psi4/editors: this one should be ready for a final review & merge.

psi4/driver/driver_cbs.py
```diff
@@ -701,13 +936,170 @@ def return_energy_components():
...
+        cbs_metadata = []
+        possible_stages = ["scf", "corl"]
+        while len(possible_stages) > 0:
+            sn = possible_stages.pop(0)
+            # either both *_wfn and *_basis have to be specified
+            if "{:s}_wfn".format(sn) in kwargs and "{:s}_basis".format(sn) in kwargs:
+                stage = {"wfn": kwargs["{:s}_wfn".format(sn)], "basis": kwargs["{:s}_basis".format(sn)]}
+            # or we're at a scf stage which can be implied with a provided scf_basis
+            elif sn == "scf" and "{:s}_basis".format(sn) in kwargs:
+                stage = {"wfn": "hf", "basis": kwargs["{:s}_basis".format(sn)]}
```
loriab: Can this handle the internal setup for SCF stage right `energy(cbs, corl_wfn='mp2', corl_basis='cc-pv[dt]z')`? I'm not seeing it, but maybe I'm wrong. You've got the running code handy.

psi4/driver/driver_cbs.py
```diff
@@ -798,88 +1189,52 @@ def cbs(func, label, **kwargs):
...
-
-        Indicates the (superior) energy method for which a fifth delta correction
-        to the correlation energy is to be obtained.
-
-    :type delta5_wfn_lesser: string
-    :param delta5_wfn_lesser: |dl| ``delta4_wfn`` |dr| || ``'ccsd(t)'`` || etc.
-
-        Indicates the inferior energy method for which a fifth delta correction
-        to the correlation energy is to be obtained.
-
```
loriab: yay, deletions :-)

psi4/driver/driver_cbs.py
```diff
@@ -701,13 +936,170 @@ def return_energy_components():
...
+        # 1a) all items must have wfn
+        if "wfn" not in item:
+            raise ValidationError("Stage {d} doesn't have defined level of theory!".format(cbs_metadata.index(item)))
+    # 1b) all items must have basis set
+        if "basis" not in item:
+            raise ValidationError("Stage {d} doesn't have defined basis sets!".format(cbs_metadata.index(item)))
+    # 2a) process required stage parameters and assign defaults
+        stage = {}
+        stage["wfn"] = item["wfn"].lower()
+        stage["basis"] = _expand_bracketed_basis(item["basis"].lower())
```
loriab: doesn't this need `molecule=molstr` as second arg to fn call? Can probably replace the str with dict nowadays.

psi4/driver/driver_cbs.py
loriab: good addition

psi4/driver/driver_cbs.py
```diff
@@ -1839,6 +1651,9 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
     cbs_kwargs['return_wfn'] = True
     cbs_kwargs['molecule'] = molecule
     cbs_kwargs['verbose'] = cbs_verbose
+    
+    if user_dertype != None:
+        cbs_kwargs['dertype'] = user_dertype
```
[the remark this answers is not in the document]
loriab: Not sure. If it was deliberate, recursiveness or sow/reap would be my suspects.

psi4/driver/driver_cbs.py
```diff
@@ -1839,6 +1651,9 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
 
     # Find method and basis
     if method_list[0] in ['scf', 'hf', 'c4-scf', 'c4-hf']:
```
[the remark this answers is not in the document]
loriab: Sure, I think this is an excellent use for cbs_metadata,

doc/sphinxman/source/cbs.rst
```diff
@@ -181,3 +191,24 @@ existing examples are below.
...
+
+.. autofunction:: psi4.driver.driver_cbs._expand_scheme_orders
+
+.. autofunction:: psi4.driver.driver_cbs._contract_scheme_orders
+
+.. autofunction:: psi4.driver.driver_cbs._cbs_wrapper_methods
+
+.. autofunction:: psi4.driver.driver_cbs._parse_cbs_gufunc_string
+
+.. autofunction:: psi4.driver.driver_cbs._cbs_gufunc
```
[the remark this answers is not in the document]
PeterKraus: I wasn't sure what the policy is - whether we want to document each function, or just the user-facing ones. My thought was: if I spend the time writing the docs, I'll want them rendered in the manual. I can remove the section, if you prefer to keep the manual concise.

psi4/driver/driver_cbs.py
```diff
@@ -701,13 +936,170 @@ def return_energy_components():
...
+    Parameters
+    ----------
+    cbs_metadata : array
+        array of dicts containing CBS stage keywords.
+
+    Returns
+    -------
+    array
+        Array of dictionaries, with each item consisting of an extrapolation
+        stage. All validation takes place here.
```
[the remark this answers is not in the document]
PeterKraus: I've replaced most occurrences of the word `array` with `list`.

psi4/driver/driver_cbs.py
```diff
@@ -701,13 +936,170 @@ def return_energy_components():
...
+    """
+
+    metadata = []
+    for item in cbs_metadata:
+        # 1a) all items must have wfn
+        if "wfn" not in item:
+            raise ValidationError("Stage {d} doesn't have defined level of theory!".format(cbs_metadata.index(item)))
```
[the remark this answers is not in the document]
PeterKraus: I meant to use {:d}, actually. I'd prefer the expected type to be shown.

psi4/driver/driver_cbs.py
PeterKraus: It doesn't need it (as it builds the basis for H2 when no molecule is provided), but it might be safer to pass it on.

psi4/driver/driver_cbs.py
PeterKraus: Yup, exactly the same behaviour as in 1.3a1:
[cbs_dict](https://github.com/psi4/psi4/files/2519138/post.txt), [1.3a1](https://github.com/psi4/psi4/files/2519139/pre.txt)

doc/sphinxman/source/cbs.rst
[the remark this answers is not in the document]
loriab: I worry users will be confused and think they should make use of the internal fns. Go ahead and remove them, please. In a better version of the API docs for devs, they'll show up.

psi4/driver/driver_cbs.py
```diff
@@ -701,13 +936,170 @@ def return_energy_components():
...
+            scf["basis"] = ([stage["basis"][0][-1]], [stage["basis"][1][-1]])
+            scf["treatment"] = "scf"
+            scf["stage"] = "scf"
+            scf["scheme"] = _get_default_xtpl(len(scf["basis"][1]), scf["treatment"])
+            scf["alpha"] = None
+            scf["options"] = False
+            metadata.append(scf)
+    # 2c) keep processing current stage
+        stage["treatment"] = item.get("treatment", "scf" if len(metadata) == 0 else "corl")
+        stage["stage"] = item.get("stage", False)
```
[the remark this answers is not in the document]
loriab: I agree on `stage` -- it'll be nice to have variable labels. Is that in a test somewhere?

Not so sure about `treatment`. It's switching btwn `scf` and `corl`, right? I wouldn't have expected past code to handle a non-scf-or-corl value correctly. Great if it can now, just not seeing it from a quick search.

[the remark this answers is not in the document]
loriab: Because of GitHub's infernal new "helpfullness" in hiding conversations, I can't find the `basis_lo` thread. I understand your use-case now and agree. And perhaps if there were several alternate formulas to be applied to `"AE - FC"` then that would constitute a new `treatment` class in your view? For `basis_lo`, since that's a new feature, I'd feel better if it had a test to make sure we don't break it.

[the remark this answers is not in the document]
PeterKraus: In the past code, you could only ever pass in one `scf` stage. The new code, due to `_get_default_xtpl` requires each stage to have either `corl` or `scf` set using the `treatment` tag. One can, in principle, pass two `scf` stages using the new code. I can't think of why someone would want to do that, perhaps with the exception of relativistic corrections over scf values from extrapolated basis sets. It's a very niche use case (the xtpl formula for scf and corl for 1 basis set is the same - xtpl_highest_1). I can of course have the code ignore user input.

There is no test for `stage`; it is really just a print label - I didn't think it was necessary.

The `AE - FC` class is still correction (or extrapolation with 2 basis sets) of correlation; it doesn't require a new class of `treatment`.

I'll add a test for `basis_lo`.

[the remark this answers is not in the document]
loriab: Thanks. I think the `basis_lo` is the important test. Throw in an alternate `stage` label, just so the line of code gets hit and we may notice if something breaks.

For `treatment`, pass `treatment='leeches'` and if a sensible error happens, I'm good with having it user-settable. If error is obscure, best to take that knob away from the user. (If calculation works, I'm backing away slowly.)

psi4/driver/driver_cbs.py
PeterKraus: Fixed now.

psi4/driver/driver_cbs.py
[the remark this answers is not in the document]
PeterKraus: Leeches now throw a validation error ;)

PeterKraus:
@JonathonMisiewicz , as you also reviewed this PR, any comments?

PR #1244, opened 2018-09-20
fixes amplitude memory estimate not recognizing frozen core

### LINKED ISSUE #898: mem requirement calculation in the dfocc module
I noticed that there is some kind of overflow happening in the printing of the memory (see below).
Looks like all variables are doubles so idk why. https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dfocc/manager.cc#L2493

The other thing is that this (and similar lines)
`cost_ampAA = nocc2AA * nvir2AA;` should maybe be `cost_ampAA = naocc2AA * nvir2AA;`
Otherwise frozen-core is ignored for the memory consideration. Or I misunderstand something.

If you could have a look @bozkaya ?

```
                    DF-MP2.5  
              Program Written by Ugur Bozkaya
              Latest Revision September 9, 2017

 ============================================================================== 
 ============================================================================== 
 ============================================================================== 

        MO spaces... 

         FC   OCC   VIR   FV 
        ----------------------
         19   49   827    0

        Number of basis functions in the DF-CC basis: 2173

        Available memory                      :  55313.11 MB 
        Memory requirement for 3-index ints   :  12050.26 MB 
        Memory requirement for DF-CC int trans:  36889.52 MB 
        Memory requirement for CC contractions: -34560.53 MB 
        Total memory requirement for DF+CC int: -22510.28 MB 
        Memory requirement for Wabef term     :  39520.86 MB 



```

hokru:
## Description
Memory estimation for the amplitudes now recognize frozen_core settings.

## Todos
- [x] fixes #898 
- [ ]  check code for same problem in CD algorithm variant.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfocc/manager.cc
+++ b/psi4/src/psi4/dfocc/manager.cc
@@ -64,7 +64,7 @@ void DFOCC::omp2_manager() {
     if (reference_ == "RESTRICTED") {
         // mem for amplitudes
         cost_ampAA = 0.0;
-        cost_ampAA = nocc2AA * nvir2AA;
+        cost_ampAA = naocc2AA * nvir2AA;
         cost_ampAA /= 1024.0 * 1024.0;
         cost_ampAA *= sizeof(double);
         cost_amp = 3.0 * cost_ampAA;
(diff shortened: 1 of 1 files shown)


hokru:
All "MB" units seem to be "MiB" units if one wants to be pedantic. Since there are so many, scripting it would be better. @bozkaya do you remember if you stuck to binary units throughout?

bozkaya:
I think present changes are okay. I missed FC case when I first wrote it, and I did not pay attention to it too much. I did not stuck to binary units. We may safely change 1024 to 1000 to make it canonical MB.  @hokru @dgasmith

Issue #1247, opened 2018-09-20
Coupled Cluster Rewrite


amjames:
We (@amjames and @robertodr) propose a partial rewrite of the coupled cluster and DPD functionalities in Psi4.
The main motivation is to make the pending [PR #1061, Expose wavefunction and amplitudes to python](https://github.com/psi4/psi4/pull/1061) available satisfying the constraints and comments offered by @dgasmith and @amjames.
Exposing the T and Lambda amplitudes to the Python layer will have a series of beneficial consequences:
 - It will be possible to experiment with these quantities easily.
- Writing and reading amplitudes to file ([HDF5](https://portal.hdfgroup.org/display/support), [Zarr](https://zarr.readthedocs.io/en/stable/), or any other format) is more easily managed from Python, especially if we need to preserve compatibility with Windows.
 - Restart capabilities.
### cc* modules
- Create a single `cc` folder where all the other `cc*` folders will live as subfolders.
- Provide a top-level `CCWavefunction` class, partly modeled on the existing `CIWavefunction` class to overtake the whole implementation of the CC functionality.
 - The class will manage `MOInfo`,` Params`, `DPD` as local data members. This will ensure that the lifetime of these objects is tied to a particular instance of a CC calculation.
 - The class will contain the various steps in a CC calculation as methods. 
 - Move from tstart-tstop to the new timers timer_on-timer_off
### DPD Library 
 - Remove use of static global machinery from DPD member functions. 
 - Implement proper copy/move assign/construct for the DPD object. 
 - Implement internal structs with proper constructors as needed.
 - Replace c-style arrays/raw pointers with appropriate smart pointer/containers 
 - Refactor shared state (shared by multiple DPD instances, currently dpd_gbl struct) into singleton(s). 
    - **Note**: Memory tracking in its current form is not thread safe and relies on the client code properly locking calls to function that need to perform allocations with a mutex.
 - Insert a layer of abstraction between DPD and psio. (slated for deprecation/replacement)

For visibility: @lothian @CDSherrill @andysim @jturney @fevangelista @dgasmith

CDSherrill:
This sounds nice!  Python accessibility of amplitudes would be a great new feature.  And cleanup from C structs towards Wavefunction object would also be welcome.


[...]

PR #1206, opened 2018-08-31
fmt: clang-format the cclambda module

andysim:
## Description
clang-format the cclambda module

## Questions
- [ ] If merging this PR will hinder your research b/c you have local edits, protest here or in #1144

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1253, opened 2018-09-21
Creating a CC wavefunction (part 0 of N)

robertodr:
## Description

The main motivation for these changes is to make the pending #1061 available satisfying the constraints and comments offered by @dgasmith and @amjames.
This mainly amounts to localizing state to a suitable CC wavefunction object and we tried doing that within the scope of #1061, but quickly recognized that some form of rewrite of the CC infrastructure would be more helpful in general. See also #1247 

This PR makes the first step of moving the various `cc*` folders inside a top-level `cc` folder which contains the `ccwave.h` and top-level implementation files. This is thus mostly aesthetics, but avoids drowning the (hopefully!) good stuff to come in a large diff.

~Hold off merging until #1203, #1206, #1251, and #1252 are in. I'll keep rebasing this one accordingly.~

## Todos
- [x] Moved `cc{energy,hbar,lambda,density,eom,response}` under `cc` folder.
- [x] Moved `ccwave.h` header file to top-level `cc` folder.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/CMakeLists.txt
+++ b/psi4/src/psi4/cc/CMakeLists.txt
@@ -1,5 +1,28 @@
 add_library(cc STATIC "")
 
+# FIXME Remove once done
+list(APPEND _Clang_debug_flags
+  "-Wall"
+  "-O0"
+  "-g3"
+  "-Wextra"
+  "-Winit-self"
+  "-Woverloaded-virtual"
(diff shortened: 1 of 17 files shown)


robertodr:
@loriab these were all `bin`-type modules, so the `install` rule was not added.

PR #1261, opened 2018-09-25
[WIP] support for single-precision (FP32) linear algebra

hokru:
## Description
**work in progress**
Enables basic linear algebra in single-precision (FP32).
For now rather a rough proof-of-principle implementation for perturbative triples correction in single-precision. Any help is welcome and perhaps even needed for clean C++ and/or smart implementations.

## Todos
- [x] Most of BLAS in single-precision
- [x] SharedTensor analogues for FP32 in `occ` module
- [x] basic version of (T)-correction in FP32  for `occ` module working. (50% faster on my workstation over FP64, error below 1e-11 for benzene-methane complex at VTZ)
- [ ] tidy up
- [ ] template for double-to-float conversion of SharedTensor (currently manual loops)
- [ ] documentation
- [ ] ...

## Questions/Notes
 - general comments and directions welcome! Also on general conventions. 

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


andysim:
This is a fantastic addition, and a very welcome feature.  Just skimming through it, it looks like you cloned Ugur's tensor code and converted it to use floats.  That's a good strategy for getting started, but it effectively leaves two different codes to be maintained.  It may be worth considering modifying that tensor code to take the floating point type as a template parameter instead.  That way, only one tensor code and one (T) code needs to be written / maintained, and the appropriate version can easily be instantiated based on user input.  I'm happy to add more detail if you like, but just wanted to get the conversation started.  Either way - this will be a fantastic feature.

hokru:
Yes, I started by copying Ugur's Tensor code. His code/syntax with these constructions is easily understandable. But then I also had to make changes in what seems quite fundamental code pieces to psi4 (libciomr) about block matrices and arrays.
With my limited C++ understanding I was forced to make sometimes copies with `_float` added:
https://github.com/psi4/psi4/pull/1261/files#diff-4345139288e2ce7d0f6fea4a51fb2ea1R140
 which perhaps can be solved smarter as well.

In these single-precision algorithms one will need some mixing of doubles with floats as it is best for the accuracy-performance compromise to selectively change the precision for only specific terms. Not sure if this is entirely solvable without cloning some blocks of physics code.
But not needing to clone his entire tensor class would be good!

hokru:
Actually no. I played around with it in ORCA for CCSD and (T) during my PhD years ago for GPUs. (back then the GPUs were still slow and it wasn't too impressive). We also used FP32 codes in the group of Prof. Grimme for MP2 and CIS(D) since forever. At least my motivation for this is quite old.

hokru:
I rebased, but it still used MSVC? The appveyor log does not show the beginning, not sure how BLAS is linked and why C_SDOT is missing.

loriab:
There's some links to display the whole log. Link line advisor recommends normal `mkl_rt.lib`, which is there. Another project mentions adding `double-conversion.lib`. I'm really not sure.

hokru:
@andysim (or others)
>  It may be worth considering modifying that tensor code to take the floating point type as a template parameter instead.

Could you be more specific? Perhaps an example?

I have a longer train ride coming up and might continue working on this.

robertodr:
@hokru a simple example. Rather than having two functions:
```
float add(float x, float y);
double add(double x, double y);
```
you can implement one leaving the type as a template parameter:
```
template <typename T> T add(T x, T y);
```
then at call site one would write:
```
float a = add(1.0, 2.0);
```
and the compiler will instantiate the correct function for the types passed.

susilehtola:
I think you'd still have to declare the instantiations into ```float``` and ```double```. Without it, it would be
```float a = add<float>(1.0f, 2.0f)```.

amjames:
@susilehtola You should not have to explicitly provide the function template in the example you gave. Provided `T` can be deduced from the function arguments,  `add<float>` will be instantiated by the compiler.

Class templates on the other hand would *always* need to have their template arguments fully specified even if instantiations of the template have been declared. Prior to c++17, there is no template parameter deduction performed on class templates, only function templates.

susilehtola:
I stand corrected. But since the present case is class related, explicit instantiations are needed, right? That is, the CC solver class should probably be templated as a whole.

PR #1262, opened 2018-09-25
Replace O1 and O3 with CH2s and CH2t to prevent orbital ordering noise.

PeterKraus:
## Description
Thanks @raimis for reporting and @hokru and @andysim for debugging!

## Checklist
- [x] `cubeprop-frontier` passes, and orbital energies are non-degenerate

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Yes, by tradition still storing output.ref files. But cc is only thing that uses them. Missing ones checked for about once a year. And updating them done systematically even less frequently.

andysim:
Thanks for the reminders.  Even though the tests don't use `output.ref`, it can be handy for quickly diagnosing failures, so thanks for adding it, @PeterKraus .

PR #1234, opened 2018-09-08
qcelemental connection for physconst and periodictable

loriab:
## Description
This follows (and includes #1213), so that one should be merged and this one rebased for a smaller PR to examine. Whereas that one switches out physconst py-side and c-side (big effect) and periodictable c-side (no effect), this one deletes all physconst and periodictable sources py-side and starts pulling from qcelemental.

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] driver-side physconst --> no change
- [x] driver-side psifiles --> `psi4.driver.constants.PSIF_` --> `psi4.driver.psif.PSIF_`
- [x] psiapi --> can't `from psi4.driver.constants import c`, instead `psi4.driver.constants.c` or `c = psi4.driver.constants.c`
- [x] psithon --> no change
- [x] qcdb --> `psi_c` --> `qcel.constants.c`
- [x] put in a basic connection to cmake. it'll clone it for you and install next to psi (which means pythonpath picks it up)

## Checklist
- [ ] ~Tests added for any new features~
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1275, opened 2018-10-01
Use qcelemental molparse

loriab:
## Description
Third in the qcelemental series, following #1234 (no use reviewing this one until that one goes through and this one's rebased. This moves molecule parsing out of `driver/qcdb/molparse` and into `qcelemental.molparse`. Since the latter are files copied from the former last week, no real concern about different behavior.

## Checklist
- [x] full test suite run

## Status
- [ ] Ready for review
- [ ] Ready for merge


psi4/driver/molutil.py
```diff
@@ -102,156 +103,154 @@ def molecule_from_arrays(cls,
...
+            #missing_enabled_return=missing_enabled_return,
+            #tooclose=tooclose,
+            #zero_ghost_fragments=zero_ghost_fragments,
+            #nonphysical=nonphysical,
+            #mtol=mtol,
+            verbose=verbose)
+
+    else:
+        raise ValidationError("""Schema not recognized, schema_name/schema_version: {}/{} """.format(
+            molschema.get('schema_name', 'NA'), molschema.get('schema_version', 'NA')))
```
dgasmith: Instead of `NA` could we do a `None` or `null` which might indicate a bit better that we are missing those elements.

psi4/driver/molutil.py
```diff
@@ -102,156 +103,154 @@ def molecule_from_arrays(cls,
...
 
-        if return_dict:
-            return core.Molecule.from_dict(molrec), molrec
+        if 'fragments' in ms:
+            frag_pattern = ms['fragments']
         else:
-            return core.Molecule.from_dict(molrec)
+            frag_pattern = [np.arange(len(ms['symbols']))]
+
+        dcontig = qcdb.Molecule.contiguize_from_fragment_pattern(
```
dgasmith: Should this go into `qcelemental` as well?

psi4/driver/qcdb/molecule.py
```diff
@@ -566,7 +559,7 @@ def format_molecule_for_orca(self):
         """
         options = collections.defaultdict(lambda: collections.defaultdict(dict))
         self.update_geometry()
-        factor = 1.0 if self.PYunits == 'Angstrom' else psi_bohr2angstroms
+        factor = 1.0 if self.PYunits == 'Angstrom' else qcel.constants.bohr2angstroms
```
dgasmith: Is there any reason the units are case sensitive?

psi4/driver/molutil.py
loriab: Yes, on my radar. There's no reason it needs to be in Molecule rather than molparse. I even have tests. I think I just wrote it hastily for json_wrapper, so it was more direct to stuff in Molecule at the time.

psi4/driver/qcdb/molecule.py
loriab: The _storage_ is case sensitive, like an enum. I think the _input_ has been rendered insensitive. Ping me if you find otherwise.

psi4/driver/molutil.py
```diff
@@ -311,7 +249,9 @@ def geometry(geom, name="default"):
         try:
             import pylibefp
         except ImportError as e:  # py36 ModuleNotFoundError
```
andysim: Do we need to also worry about `ModuleNotFoundError` here?  I suspect not, because I think it derives from `ImportError`, but just want to check that this is the correct way for us to support py3.6+

loriab: I'm pretty sure this is the right way to support 2.7+. then when we switch to 3.6+ can replace with the more specific `ModuleNotFoundError`

PR #1287, opened 2018-10-12
Replace only the last match when generating outfile name from infile. 

### LINKED ISSUE #1286: possible bug in output file name
Hi!
It seems that when the input file is e.g. `some_name.dat.xyz.dat` , psi4 creates the 
output file named `some_name.out.xyz.out`  instead of just `some_name.dat.xyz.out`.
My guess would be that enything except for the file extension (the last `.dat` in this case)
should be untouched...

amjames:
## Description
There is no `rreplace` function which would be handy, so split on the first match from the end, and join on the replacement. does the job. 

fixes #1286 

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/run_psi4.py
+++ b/psi4/run_psi4.py
@@ -111,7 +111,8 @@ if args["inplace"]:
     import sysconfig
     core_location = os.path.dirname(os.path.abspath(__file__)) + os.path.sep + "core" + sysconfig.get_config_var("SO")
     if not os.path.isfile(core_location):
-        raise ImportError("A compiled Psi4 core{} needs to be symlinked to the {} folder".format(sysconfig.get_config_var("SO"), os.path.dirname(__file__)))
+        raise ImportError("A compiled Psi4 core{} needs to be symlinked to the {} folder".format(
+            sysconfig.get_config_var("SO"), os.path.dirname(__file__)))
 
     lib_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
     if ("PSIDATADIR" not in os.environ.keys()) and (not args["psidatadir"]):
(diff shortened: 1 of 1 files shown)


psi4/run_psi4.py
```diff
@@ -134,9 +135,9 @@
     if args["input"] == "input.dat":
         args["output"] = "output.dat"
     elif args["input"].endswith(".in"):
-        args["output"] = args["input"].replace(".in", ".out")
+        args["output"] = ".out".join(args["input"].rsplit(".in", 1))
     elif args["input"].endswith(".dat"):
-        args["output"] = args["input"].replace(".dat", ".out")
+        args["output"] = ".out".join(args["input"].rsplit(".dat", 1))
```
loriab: Bikeshedding, I know, but I'd mildly prefer

`args["output"] = args["input"][:-3] + "out"`

or

```
inp = pathlib.Path(args["input"])
otp = inp.stem + '.out'
```

amjames: The first one should work fine, but the second will have other consequences I think.

```python
p1 = Path("input.dat")
p2 = Path("/home/a/b/c/input.dat")
p3 = Path("./subdir1/input.dat")
p1.stem == p2.stem == p3.stem
# True
```
pathlib.Path.stem just returns the filename without extensions, So any directories will be lost. 
`psi4 -i "./calculation_a/name.in"` and `psi4 -i "./calculation_b/name.in"`  would overwrite each others outfiles.

loriab: you're right
```
>>> from pathlib import Path
>>> aa = Path('mydir/snddir/asdfasdf.in.xyz.in')
>>> str(aa.parent / aa.stem) + '.out'
'mydir/snddir/asdfasdf.in.xyz.out'
```

Issue #1148, opened 2018-08-23
Convert py-side findif from lists to dict


loriab:
In the spirit of _not_ keeping issues open for years just because there's something more that could be done, this is a new ticket for some enhancements to the finite difference code, now that we can play with it py-side.

- [ ] instead of lists to track all the geometries and corresponding energies, use lists with a bit of metadata to facilitate workflows. discussion started [here](https://github.com/psi4/psi4/issues/884#issuecomment-360953535)
- [ ] (adding mostly so we don't lose track) make sure the findif can handle a qcdb.Mol as well as a psi4.core.Mol

I believe @JonathonMisiewicz is on this one.

JonathonMisiewicz:
First, let's make sure we're agreed on what this metadata should look like. I'm envisioning the following skeleton:

```
{
    "step": { # Read from here if a displacement doesn't have its own step field
        "units": str, # "bohr" or "angstroms"
        "size": float
    },
    "stencil_size": int, # 3 or 5, currently  
    "displacement_space": str, # "SALC", "normal", "internal", "cartesian"... always "SALC" for us
    "project_translations": bool,
    "project_rotations": bool,
    "geoms": collections.OrderedDict([
        disp_label: {
            "geometry": np.ndarray,
            "energy": float,
            "gradient": np.ndarray, # May not be present!
            "step": step dictionary defined above, # May not be present!
        }
    ])
}
```

Feedback greatly appreciated.

disp_label is going to be a bit tricky. `GRAD_0_X_+2` is a reasonable way to say the double displacement of atom 0 in the X coordinate, but if you're displacing CdSALCs, it won't do. We can either give labels so abstract they don't care about the coordinate_space, or we can give each coordinate_space its own labels. Either way, we need to specify which basis vector(s) in displacement space are being displaced and how much. Perhaps a tuple of ints? `(2, 0, 0, 0, 0, 0, 0, 0, 0)` would be the analogue of `GRAD_0_X_+2`. I like the abstraction, but typing all those extra 0's gets obnoxious, and it does not scale well with system size. A frozen dictionary might work, which would instead be `{0: 2}`, but frozen dictionaries aren't in the Python standard library.

dgasmith:
Also for JSON output you cannot have OrderedDicts, nor can you have anything besides a string as a key. In many ways the key doesnt matter, it would be good to be human readable. However, it does not have to be.

Not very helpful, but a few things to think about.

JonathonMisiewicz:
[...]

PR #1289, opened 2018-10-14
Findif Metadata

JonathonMisiewicz:
## Description
Change of FINDIF to be metadata based per #1148, task one, and qcdb molecule compatibility per #1148, task two.

## Todos
- [x] Change finite difference code to use metadata.
- [x] Fixes bugs #1296 and #1306.
- [x] Reap/Sow code removed.

## Questions
- [x] How should I document that a functions takes as an argument or returns the metadata dictionary? That needs to be specified somewhere.
- [x] Are we all agreed on the metadata style? The things I've left undone are going to depend heavily on any additional metadata changes.

## Checklist
- [x] Implement the changes to all four finite difference functions: empirical dispersion, hessian by gradients, hessian by energies, gradient by energies.
- [x] Polish the resulting code. In particular, there are a few variables taken from the initialization `data` that should probably be taken from `metadict` instead. For the compute functions, it should only need the barest skeleton of the current initialization call. Depending on other revisions, I may want to create a separate initialization function for the compute functions.
- [x] Modify the reap/sow cookbooks to work.
- [x] Make this compatible with QCDB molecules. To do this _correctly_, we need CdSalc tech reproduced in QCDB. Better to just coerce the QCDB molecule for now.
- [x] Remove the optimization side reap/sow code.

## Status
- [x] Ready for review
- [x] Ready for merge.

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -139,8 +139,9 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
     print(""" %d""" % (n), end=('\n' if (n == ndisp) else ''))
     sys.stdout.flush()
 
-    # Load in displacement into the active molecule
-    molecule.set_geometry(core.Matrix.from_array(displacement["geometry"]))
+    # Load in displacement (flat list) into the active molecule
+    geom_array = np.reshape(displacement["geometry"], (-1, 3))
+    molecule.set_geometry(core.Matrix.from_array(geom_array))
 
     # Perform the derivative calculation
     derivative, wfn = derivfunc(method, return_wfn=True, molecule=molecule, **kwargs)
@@ -148,10 +149,10 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
 
     # If we computed a first or higher order derivative, set it.
     if derivfunc == gradient:
-        displacement["gradient"] = wfn.gradient().to_array()
+        displacement["gradient"] = wfn.gradient().np.ravel().tolist()
 
     # clean may be necessary when changing irreps of displacements
-    core.clean()
+    #core.clean()
 
     return wfn
 
@@ -680,7 +681,7 @@ def gradient(name, **kwargs):
 
         # Compute the gradient
         core.set_local_option('FINDIF', 'GRADIENT_WRITE', True)
-        G = driver_findif.compute_gradient_from_energy(molecule, findif_meta_dict)
+        G = driver_findif.compute_gradient_from_energy(findif_meta_dict)
         grad_psi_matrix = core.Matrix.from_array(G)
         grad_psi_matrix.print_out()
         wfn.set_gradient(grad_psi_matrix)
@@ -1030,7 +1031,8 @@ def optimize(name, **kwargs):
         if (n > 1) and (core.get_option('OPTKING', 'OPT_TYPE') == 'IRC'):
             old_thisenergy = core.get_variable('CURRENT ENERGY')
 
-        # Compute the gradient
+        # Compute the gradient - preserve opt data despite core.clean calls in gradient
+        #core.IOManager.shared_object().set_specific_retention(1, True)
         G, wfn = gradient(lowername, return_wfn=True, molecule=moleculeclone, **kwargs)
         thisenergy = core.get_variable('CURRENT ENERGY')
 
@@ -1300,13 +1302,14 @@ def hessian(name, **kwargs):
 
         for n, displacement in enumerate(findif_meta_dict["displacements"].values(), start=1):
             _process_displacement(gradient, lowername, moleculeclone, displacement, n, ndisp, **kwargs)
+            core.clean()
 
         wfn = _process_displacement(gradient, lowername, moleculeclone, findif_meta_dict["reference"], ndisp, ndisp,
                                     **kwargs)
 
         # Assemble Hessian from gradients
         #   Final disp is undisp, so wfn has mol, G, H general to freq calc
-        H = driver_findif.compute_hessian_from_gradient(molecule, findif_meta_dict, irrep)  # TODO or moleculeclone?
+        H = driver_findif.compute_hessian_from_gradient(findif_meta_dict, irrep)  # TODO or moleculeclone?
         wfn.set_hessian(core.Matrix.from_array(H))
         wfn.set_gradient(G0)
 
@@ -1342,12 +1345,13 @@ def hessian(name, **kwargs):
 
         for n, displacement in enumerate(findif_meta_dict["displacements"].values(), start=1):
             _process_displacement(energy, lowername, moleculeclone, displacement, n, ndisp, **kwargs)
+            core.clean()
 
         wfn = _process_displacement(energy, lowername, moleculeclone, findif_meta_dict["reference"], ndisp, ndisp,
                                     **kwargs)
 
         # Assemble Hessian from energies
-        H = driver_findif.compute_hessian_from_energy(molecule, findif_meta_dict, irrep)
+        H = driver_findif.compute_hessian_from_energy(findif_meta_dict, irrep)
         wfn.set_hessian(core.Matrix.from_array(H))
         wfn.set_gradient(G0)
 
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -151,7 +151,7 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
 
        Returns
        -------
-       wfn: psi4.core.wfn
+       wfn: :py:class:`~psi4.core.Wavefunction`
            The wavefunction computed.
     """
 
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -868,17 +868,6 @@ def optimize(name, **kwargs):
         If a nested series of python functions is intended (see :ref:`sec:intercalls`),
         use keyword ``opt_func`` instead of ``func``.
 
-    :type mode: string
-    :param mode: |dl| ``'continuous'`` |dr| || ``'sow'`` || ``'reap'``
-
-        For a finite difference of energies optimization, indicates whether
-        the calculations required to complete the
-        optimization are to be run in one file (``'continuous'``) or are to be
-        farmed out in an embarrassingly parallel fashion
-        (``'sow'``/``'reap'``). For the latter, run an initial job with
-        ``'sow'`` and follow instructions in its output file. For maximum
-        flexibility, ``return_wfn`` is always on in ``'reap'`` mode.
-
     :type dertype: :ref:`dertype <op_py_dertype>`
     :param dertype: ``'gradient'`` || ``'energy'``
 
@@ -954,19 +943,15 @@ def optimize(name, **kwargs):
     >>> e, wfn = opt('mp5', return_wfn='yes')
     >>> wfn.gradient().print_out()
 
-    >>> # [3] Forced finite difference hf optimization run in
-    >>> #     embarrassingly parallel fashion
-    >>> optimize('hf', dertype='energy', mode='sow')
-
-    >>> # [4] Can automatically perform complete basis set extrapolations
+    >>> # [3] Can automatically perform complete basis set extrapolations
     >>> optimize('MP2/cc-pV([D,T]+d)Z')
 
-    >>> # [5] Can automatically perform delta corrections that include extrapolations
+    >>> # [4] Can automatically perform delta corrections that include extrapolations
     >>> # even with a user-defined extrapolation formula. See sample inputs named
     >>> # cbs-xtpl* for more examples of this input style
     >>> optimize("MP2/aug-cc-pv([d,t]+d)z + d:ccsd(t)/cc-pvdz", corl_scheme=myxtplfn_2)
 
-    >>> # [6] Get info like geometry, gradient, energy back after an
+    >>> # [5] Get info like geometry, gradient, energy back after an
     >>> #     optimization fails. Note that the energy and gradient
     >>> #     correspond to the last optimization cycle, whereas the
     >>> #     geometry (by default) is the anticipated *next* optimization step.
@@ -1009,11 +994,6 @@ def optimize(name, **kwargs):
     else:
         hessian_with_method = kwargs.get('hessian_with', lowername)
 
-    # are we in sow/reap mode?
-    opt_mode = kwargs.get('mode', 'continuous').lower()
-    if opt_mode not in ['continuous', 'sow', 'reap']:
-        raise ValidationError("""Optimize execution mode '%s' not valid.""" % (opt_mode))
-
     optstash = p4util.OptionsState(
         ['OPTKING', 'INTRAFRAG_STEP_LIMIT'],
         ['FINDIF', 'HESSIAN_WRITE'],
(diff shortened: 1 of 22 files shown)


dgasmith:
Can you provide an example of the metadata after generation and after computation (in a `gist` if its large please).

JonathonMisiewicz:
You can see the two versions [here](https://gist.github.com/JonathonMisiewicz/ac6389c6104603dff0fc2c953bcb4066). Now that I look at it, I imagine we'll want both the geometry and the gradient converted to nested lists for JSON purposes?

dgasmith:
Look good. Psi4 matrices should be convert to flat list `matrix.np.ravel().tolist()`. You should also consider adding a Molecule JSON object to the generated data as you do not carry enough data to reproduce the molecule at the moment. I don't know the exact command from a Psi4 molecule, but @loriab should have it.

JonathonMisiewicz:
I assume the gradient arrays are to be converted to lists as well?
Any reason reason to prefer flat lists to nested?

dgasmith:
Flat list have lower serialization costs, and yes both gradient and geometries to lists.

loriab:
`a_qcdb_or_libmints_molecule_instance.to_dict()` will get you the psi4 internal representation, which has rather more information than `.to_schema(dtype=1)`. Save the latter for final dump in json_runner. 

Second the flat over nested lists.

For an internal representation, it's fine (and in my experience a lot more convenient) to leave arrays as `np.array`. Then for potentially external fns and for to_schema, have a `np_out` argument that calls https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/util/misc.py#L126 to ndarray-->list as needed (e.g., https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/molparse/from_arrays.py#L402-L403)

JonathonMisiewicz:
What's `to_schema_`, and what is an example of a "potentially external function"? I think I get the _purpose_ of QCDB by now, but the rest is vague.

loriab:
Just for background, this is the structure of `qcdb.molparse` (or `qcel.molparse`) which now handles Molecule I/O
```
def from_arrays(**arrays):
    validate_and_type ...
    construct_molrec_dictionary ...
    return molrec

def from_string(str):
    parse ...
    return from_arrays(**arrays)

def from_schema(schema_dict):
    process ...
    return from_arrays(**arrays)

def to_string(molrec, dtype):
    format_by_dtype ...
    return str

def to_schema(molrec, dtype):
    depending_on_schema_flavor_dtype_rearrange_and_remove_fields ...
    return dict
```

then proper objects like `qcdb.Molecule` and `psi4.core.Molecule` can hook in through
```
class Molecule():
    def to_dict(self):
        convert_object_structure_into_molrec ...
        return molrec

    @classmethod
    def from_dict(cls, molrec):
        construct_object_from_molrec_contents ...
        return mol

    @staticmethod
    def from_string(str):
        molrec = molparse.from_string(str)
        return Molecule.from_dict(molrec)

    def to_string(self, dtype):
        return molparse.to_string(self.to_dict(), dtype=dtype)

    @staticmethod
    def from_arrays(**arrays):
        molrec = molparse.from_arrays(**arrays)
        return Molecule.from_dict(molrec)

    def to_schema(self, dtype):
        return molparse.to_schema(self.to_dict(), dtype=dtype)
```

The above is more detailed than you need, but it's about time I wrote it down.

By external function, I just meant one that returns your findif dictionary. If you want to keep arrays as numpy structures, then can have a flag that does or does not call `unnp` (or similar) and things that are destined for a JSON file (like a `to_schema()` call) can be sure to use `to_dict(np_out=False)` while internal calls can do `to_dict(np_out=True)`. I suggest it merely as a way to avoid lots of `np.array(inp)` and `to_list`s.

And keep in mind that the above structure is necessary b/c the "molrec" psi4 internal repr dict isn't quite the same as the QCSchema dict, so some translation necessary. Possibly you can avoid that complication.

This answer is more background than specifics. Glad to talk the latter instead.

JonathonMisiewicz:
Don't be alarmed if this fails tests. I'm still working out the correct sequence of `core.clean()` and `set_specific_retention` so nothing breaks.

I'm currently tracking down a bug with the `full_hess_every` keyword set to something not zero, which slipped through test coverage. If it wasn't introduced by this PR, I'll post more details about it.

loriab:
On "How should I document that a functions takes as an argument or returns the metadata dictionary? That needs to be specified somewhere.", I'd say:
* in some fn (doesn't much matter where at this point), define each of the fields the metadata dict can or could include very carefully. e.g., https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_arrays.py#L216-L251
* choose a unique name for the variable that effectively signals its type, e.g., `findifrec`
* Use `findifrec` (or whatever you choose) as if it was a proper type when functions take or return the object. e.g., https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_string.py#L64-L67

JonathonMisiewicz:
I have now run almost the _entire_ test suite with 100% pass rate. (DFT benchmarks and fsapt-1 excluded as they always time out and are not relevant to these changes.) If there are still concerns about the metadata, tell me sooner rather than later. Next step is documenting the metadata.

Also, am I correct in assuming that we want the optimization sow/reap code ripped up as well? I assume that wouldn't be very useful, without the sow/reap code in hessian and gradient.

dgasmith:
The current metadata looks great. I would run with it.

As per removing sow/reap, I would hold off a bit until we have a firm understanding of how it will be replaced. @loriab might have ideas there.

loriab:
As for sow/reap, I don't know what it's getting replaced with exactly, but I'm determined to do away with it, and the current tech doesn't work in py3 (and won't work w/pyoptking), so I'd say clear it away (from `driver.py` -- leave the stuff in p4util).

psi4/driver/driver_findif.py
```diff
@@ -238,13 +245,72 @@ def _geom_generator(mol, freq_irrep_only, mode):
...
+        The reference molecule, in MolSSI schema. See
+        https://molssi-qc-schema.readthedocs.io/en/latest/auto_topology.html
+
+    displacements : dict
+        A dictionary mapping labels specifying the displacement to data about
+        the geometry. Labels are of the form "A: B, C: D" where A and B index the
+        basis vector in displacement space and are ordered, and C and D index the
+        step magnitude. For instance, "0: 1, 1: -1" specifies displacing +1 in
+        displacement vector 0 and -1 in displacement vector 1. "1: -1, 0: 1" is
+        forbidden for breaking ordering. Generalizes to arbitrary numbers of
```
loriab: dictionaries are inherently unordered. how enforce this?

psi4/driver/driver_findif.py
```diff
@@ -238,13 +245,72 @@ def _geom_generator(mol, freq_irrep_only, mode):
...
+        Whether rotations are to be projected out of the displacements.
+
+    molecule : dict
+        The reference molecule, in MolSSI schema. See
+        https://molssi-qc-schema.readthedocs.io/en/latest/auto_topology.html
+
+    displacements : dict
+        A dictionary mapping labels specifying the displacement to data about
+        the geometry. Labels are of the form "A: B, C: D" where A and B index the
+        basis vector in displacement space and are ordered, and C and D index the
```
loriab: sure you don't mean A & C index in displacement space and B & D in magnitude? might want to use `A: a, B: b`

psi4/driver/driver_findif.py
JonathonMisiewicz: I'm not enforcing what you think I am. The comma-separated parts of the label are ordered _within the string_. The labels are not ordered within the dictionary. I'll rewrite for clarity when you're done.

JonathonMisiewicz:
And, that's all I've got planned for this PR! Unless tests fail (which they shouldn't), this should be good to go.

psi4/driver/driver.py
```diff
@@ -130,6 +130,56 @@ def _energy_is_invariant(gradient, stationary_criterion=1.e-2):
...
+          A dictionary containing the necessary information for the displacement.
+          See driver_findif/_geom_generator.py docstring for details.
+       n : int
+          The number of the displacement being computed, for print purposes.
+       ndisp : int
+           The total number of geometries, for print purposes.
+
+       Returns
+       -------
+       wfn: psi4.core.wfn
```
loriab: If you spec classes like the below in docstrings, they'll show up [linked in the manual](http://psicode.org/psi4manual/master/api/psi4.driver.vibanal_wfn.html). Not required, but it is good to spell them out and capitalize, like `psi4.core.Wavefunction`.

```
:py:class:`~psi4.core.Molecule`"
```

psi4/driver/driver_findif.py
```diff
@@ -238,13 +245,72 @@ def _geom_generator(mol, freq_irrep_only, mode):
...
+            2. Displacement specific overrides for globals will not be
+               available unless the user specified the overrides.
+               (Such overrides are not implemented at time of writing. An example
+               is giving a displacement its own step dict.)
+
+    step : dict
+        A descriptor for the finite difference step.
+        In future, this can be overriden by step fields for individual displacements.
+
+        units : {'bohr', 'angstrom'}
```
loriab: for cohesion with everything else that started from libmints, make this `{'Bohr', 'Angstrom'}`

psi4/driver/driver.py
JonathonMisiewicz: Is the quote mark supposed to be there?

loriab: no

JonathonMisiewicz:
Okay, that should be everything. Running `ctest -E bench` as requested ended with all passes except one. The test was fixed, and now it passes.

psi4/driver/driver_findif.py
```diff
@@ -460,59 +549,59 @@ def _process_hessian(H_blocks, B_blocks, massweighter, print_lvl):
...
+        core.print_out("\n{}\n".format(np.array_str(Hx)))
 
     if print_lvl:
         core.print_out("\n-------------------------------------------------------------\n")
 
     return Hx
 
 
-def compute_hessian_from_gradient(mol, G, freq_irrep_only):
+def compute_hessian_from_gradient(metadict, freq_irrep_only):
```
loriab: should arg (and code) be `findifrec` to match docstring?

psi4/driver/driver_findif.py
```diff
@@ -643,115 +734,113 @@ def recursive_gradients(n):
...
+                        ) / (12.0 * metadict["step"]["size"])
 
         H_pi[-1] = _process_hessian_symmetry_block(H_pi[-1], B_pi[-1], massweighter, irrep_lbls[h], data["print_lvl"])
 
     # All blocks of the Hessian are now constructed!
     return _process_hessian(H_pi, B_pi, massweighter, data["print_lvl"])
 
 
-def compute_hessian_from_energy(mol, E, freq_irrep_only):
+def compute_hessian_from_energy(metadict, freq_irrep_only):
```
loriab: metadict-->findifrec?

psi4/driver/driver_findif.py
```diff
@@ -643,115 +734,113 @@ def recursive_gradients(n):
...
-                    fc = (-offdiag_row[0] - offdiag_row[1] + 9 * offdiag_row[2] - offdiag_row[3] - offdiag_row[4] +
-                          9 * offdiag_row[5] - offdiag_row[6] - offdiag_row[7] + energies[i][0] - 7 * energies[i][1] -
-                          7 * energies[i][2] + energies[i][3] + energies[j][0] - 7 * energies[j][1] -
-                          7 * energies[j][2] + energies[j][3] + 12 * ref_energy) / (12 * data["disp_size"]**2)
+        # We need off-diagonal energies, diagonal energies, AND the reference energy
+        # Grabbing off-diagonal energies is a pain, so once we know our SALCs...
+        # ...define offdiag_en to do that for us.
+        for i, salc in enumerate(salc_indices):
+            for j, salc2 in enumerate(salc_indices[:i]):
+                offdiag_en = lambda index: displacements["{j}: {}, {i}: {}".format(i=salc, j=salc2, *data["disps"]["off"][index])]["energy"]
```
loriab: Should index variables be reused here (i, j)? And note that you may need to "capture" external variables if you want their values frozen into the lambda (e.g., https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/nucleus.py#L180) otherwise the lambda will use the values as eval-time not definition-time.

JonathonMisiewicz:
I'll respond to the other points (although slowly), but what failing test are you talking about? Travis and AppVeyor both passed, and `ctest -E bench` on my machine did too.

tests/cubeprop-esp/input.dat
```diff
@@ -14,8 +14,8 @@ h2o.R = 0.52917721067 / 0.52917720859
 
 set basis cc-pvqz
 set cubeprop_tasks ['esp']
-set e_convergence 9
-set d_convergence 8
+set e_convergence 12
+set d_convergence 12
```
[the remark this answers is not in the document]
JonathonMisiewicz: Do you still want the reference cubes re-generated? It's no trouble to me.

tests/cubeprop-frontier/input.dat
```diff
@@ -23,8 +23,9 @@ set df_scf_guess  false
 set scf_type pk
 set cubeprop_tasks ['frontier_orbitals']
 set cubic_grid_overage [1.0,1.0,1.0]
-set e_convergence 9
-set d_convergence 8
+set e_convergence 12
+set d_convergence 12
+set DOCC [3, 0, 0, 1]
```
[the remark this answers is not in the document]
JonathonMisiewicz: Nope, no ref files needed to be added.

psi4/driver/driver_findif.py
```diff
@@ -37,17 +37,19 @@
 # _pi at the end of a variable name is short for "per irrep."
 # h is the index of an irrep.
 
+np.set_printoptions(formatter={'float': '{: 0.10f}'.format})
```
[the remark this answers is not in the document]
JonathonMisiewicz: Actively control printing.

I'll define a format dict at the top of the file and change all matrix print statements to `print(np.array_str(matrix, **format))`, then.

psi4/driver/driver_findif.py
```diff
@@ -238,13 +245,72 @@ def _geom_generator(mol, freq_irrep_only, mode):
...
+               available unless the user specified the overrides.
+               (Such overrides are not implemented at time of writing. An example
+               is giving a displacement its own step dict.)
+
+    step : dict
+        A descriptor for the finite difference step.
+        In future, this can be overriden by step fields for individual displacements.
+
+        units : {'Bohr', 'Angstrom'}
+            The units for the displacement. The code currently assumes "bohr".
```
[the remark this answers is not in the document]
JonathonMisiewicz: That only Bohr works. That was the case in C-findif, and Psi doesn't have a way to specify disp size in angstroms at present. I could make angstroms work by converting the disp size to bohr in the initialization step and passing that to the data dict.

psi4/driver/driver_findif.py
```diff
@@ -266,22 +332,40 @@ def init_string(data):
                 "    Generating geometries for use with {:d}-point formula.\n"
                 "    Displacement size will be {:6.2e}.\n".format(print_msg, data["num_pts"], data["disp_size"]))
 
+    # Genuine support for qcdb molecules would be nice. But that requires qcdb CdSalc tech.
+    # Until then, silently swap the qcdb molecule out for a psi4.core.molecule.
+    if isinstance(mol, molecule.Molecule):
```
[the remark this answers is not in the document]
loriab: yikes, `molecule` is the file containing the `Molecule` class. I get it now.

Properly, one would `import qcdb`, then its `__init__.py` exposes `Molecule`, so you would `qcdb.Molecule` here. But some of the driver's import structure is ... strange. Leave it as it -- might get a circular import if tried to unwind.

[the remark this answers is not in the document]
JonathonMisiewicz: I tried replacing with `from psi4.driver import qcdb`, and the one fd test I tried passed. I'll make sure all fd tests pass before the next commit.

psi4/driver/driver_findif.py
loriab: Is this one still in the works?

JonathonMisiewicz: Already handled. I swapped i and j in the lambda for k and l.

PR #1291, opened 2018-10-15
Parallelized grid construction with OpenMP

zachglick:
## Description
Parallelism was added to the construction of the DFT Grid with OpenMP. 

The grid construction's asymptotic complexity is a function of the number of atoms, so it can be significant for small-to-medium sized molecules, especially when a smaller basis set is used.

## Todos
- [ x ] Parallelized DFT grid construction

## Questions
- [ ] Is this necessary? I ran some scaling tests, and grid construction accounted for up to one third of the total wall time on 6 cores for molecules with between ~75 and ~225 atoms
- [ ] Are there best practices for OpenMP specific to Psi4? I used the static scheduler since it seems that each atom will result in the same amount of work, but I'm open to other opinions. The best scheduler could also be determined empirically on a set of test calculations, but I don't know if the differences between scheduler performance is significant enough to warrant this.

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


zachglick:
Some benchmarking results from timer.dat walltimes:

<details><summary>Before</summary>
<p>

before-par

</p>
</details>

<details><summary>After</summary>
<p>

after-par

</p>
</details>

psi4/src/psi4/libfock/cubature.cc
```diff
@@ -3632,7 +3649,11 @@ void MolecularGrid::buildGridFromOptions(MolecularGridOptions const& opt,
     radial_grids_.clear();
     spherical_grids_.clear();
 
+    radial_grids_.resize(molecule_->natom());
+    spherical_grids_.resize(molecule_->natom());
```
loriab: Any need to handle `opt.namedGrid` differently here?

zachglick: It looks like opt.namedGrid can be:

- -1 (Default) if not using an explicitly named grid (uses radial_grids_ and spherical_grids_)
- 0 to use the SG0 grid (doesn't touch radial_grids_ or spherical_grids_)
- 1 to use the SG1 grid (doesn't touch radial_grids_ or spherical_grids_)

I think that in this overloaded version of buildGridFromOptions(), the fact that parameters passed are passed in (rs, ws, Ls) precludes the grid from being SG0 or SG1. I'll admit that I'm learning the details of DFT alongside looking at this code so I could be wrong about how this implementation works.

susilehtola: SG2 and SG3 aren't implemented in Psi4 yet?

zachglick: It looks that way to me:

https://github.com/psi4/psi4/blob/f687f496d84aeb72162db36d7da4b5577ed7a4cb/psi4/src/psi4/libfock/cubature.cc#L2465-L2474

loriab: @zachglick sounds plausible, thanks. @susilehtola, actually none of the SG grids are in Psi4; SG0 and 1 are just placeholders.

Issue #1282, opened 2018-10-08
full_hess_every N with N > 0 does not work as expected


tgaillard:
Dear Psi4 developers,

The 'full_hess_every N' functionality, used to compute the full Hessian
every N steps in an optimization, does not work as expected for N > 0:
the Hessian is always recalculated on the initial structure instead of
the current one.

This can be tested as follows.

First, we only calculate the Hessian initially (full_hess_every 0). The
minimization converges in 7 steps to an R=0.95, B=104.6 geometry. The
frequencies obtained are 1776, 4112, 4211.

#input: hess0.in
molecule h2o {
  symmetry c1
  O
  H 1 R
  H 1 R 2 B
  R = 1.2
  B = 120
}
set basis cc-pVDZ
set full_hess_every 0
optimize('scf', dertype='gradient')
freq('scf', dertype='gradient')

#grep '~' hess0.out
      1     -75.93350508   -7.59e+01      1.35e-01      1.14e-01 o    3.44e-01      2.89e-01 o  ~
      2     -76.01434847   -8.08e-02      6.88e-02      5.91e-02 o    2.89e-01      2.51e-01 o  ~
      3     -76.00936003    4.99e-03      1.27e-01      1.04e-01 o    1.83e-01      1.51e-01 o  ~
      4     -76.02647334   -1.71e-02      1.62e-02      1.35e-02 o    4.78e-02      4.32e-02 o  ~
      5     -76.02694403   -4.71e-04      6.79e-03      5.64e-03 o    1.47e-02      1.11e-02 o  ~
      6     -76.02702784   -8.38e-05      1.73e-03      1.43e-03 o    2.89e-03      2.48e-03 o  ~
      7     -76.02703276   -4.93e-06      1.05e-04 *    8.74e-05 o    1.63e-04 *    1.45e-04 o  ~

#grep -A1 'post-proj  all' hess0.out
  post-proj  all modes:['0.0000i' '0.0000i' '0.0000i' '1103.2728' '1246.1842' '1310.1099'
 '1334.3071' '1578.1703' '1760.1248']
  post-proj  all modes:['0.0000i' '0.0000i' '0.0000i' '0.0000i' '0.0000' '0.0001' '1776.3032'
 '4112.4185' '4210.8898']

Now, we calculate the Hessian at every step (full_hess_every 1). The
minimization does not converge after 15 steps. The energy printed at
each step is always the initial energy. The frequencies obtained at each
step are always the initial frequencies.

[...]

PR #1299, opened 2018-10-18
Fix #1298, full_hess_every taking hessians at the starting geometry.

### LINKED ISSUE #1298: full_hess_every optimization corruption
A clean version of 287272b showed that when `full_hess_every` is enabled in an optimization, the data is corrupted. Every time a hessian is taken, it appears to be taken from the starting geometry, not the current geometry in the optimization. optking thinks the "reference" of the hessian it gets should be the starting point for the next step for purposes of convergence criteria output, but not for purposes of the internal coordinate step. In this example, every odd displacement has exactly the same energy. To me, this screams that `intco.dat` is being handled correctly, but some other file is not.

Input file:
```
molecule h2o { 
     symmetry c1
     O    
     H 1 1.0 
     H 1 1.0 2 104.5 
}

set {
  basis sto-3g
  e_convergence 10
  d_convergence 10
  points 3
  geom_maxiter 50
  full_hess_every 2 
}

optimize('scf', dertype = 0)
```

We direly need `full_hess_every` coverage in the test suite. `grep` says the only test covering it is `opt10`, which isn't even built when I build all the tests, for some reason.

The exact commit number is probably not integral to this problem.

JonathonMisiewicz:
## Description
Fixes and closes #1298. `full_hess_every` now takes hessians at the latest geometry rather than the reference geometry.

Because this bug was not caught by the test suite, I've added a test to catch this bug, and it clocks in at just over 6 seconds on my Mac.

## Checklist
- [x] Tests added for newly working features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/opt-full-hess-every/input.dat
+++ b/tests/opt-full-hess-every/input.dat
@@ -1,0 +1,20 @@
+# SCF/sto-3g optimization with a hessian every step
+
+molecule h2o {
+     O
+     H 1 1.0
+     H 1 1.0 2 104.5
+}
+
+set {
+  basis sto-3g
+  e_convergence 10
+  d_convergence 10
+  geom_maxiter 5
+  full_hess_every 1
+  g_convergence gau_verytight
+}
+
+opt_energy = optimize('scf')
+
+compare_values(opt_energy, -74.965990122534, 6, "SCF Optimized energy") #TEST
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1151,7 +1151,7 @@ def optimize(name, **kwargs):
             G = core.get_gradient()  # TODO
             core.IOManager.shared_object().set_specific_retention(1, True)
             core.IOManager.shared_object().set_specific_path(1, './')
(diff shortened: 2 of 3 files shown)


JonathonMisiewicz:
Also fixes and closes #1282, of which my #1298 was an unintentional duplicate. Sorry about that!

tests/opt-full-hess-every/input.dat
```diff
@@ -0,0 +1,20 @@
...
+     H 1 1.0 2 104.5
+}
+
+set {
+  basis sto-3g
+  e_convergence 10
+  d_convergence 10
+  geom_maxiter 5
+  full_hess_every 1
+  g_convergence gau_verytight
```
dgasmith: Any reason to use very tight here? Be good if the test could execute quickly.

JonathonMisiewicz: When I tried with `gau_tight`, it took exactly as many iterations: 3. I can change to `gau_tight` if you want, but it would be a purely cosmetic change unless something changes with the optimizer. And in that case, better safe than sorry.

PR #1304, opened 2018-10-22
Serialize Wavefunction 

### LINKED ISSUE #887: Serializing Wavefunctions
It would be very handy to have the ability to read/write entire Wavefunctions. We do this somewhat awkwardly for the SCF `read` guess functionality on [this](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1384) line. 

Some issues and solutions:
 - Serializing basis sets could be problematic. Saving the Python dictionary representation of the basis is the likely solution see [here](https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L43) to get started.
- Molecule can be serialized to/from a string (future JSON) `core.Molecule.create_molecule_from_string(mol_string)`.
- Initializing a new Wavefunction could be done with a long signature call that looks something like `Wavefunction(std::map<str::string, SharedMatrix>, std::map<str::string, SharedVector>, std::map<str::string, Dimension>, ...`. This could help reduce the number of inputs required. Curious if anyone has suggestions to simplify the population of private data.

Use cases:
 - Write our RHF wavefunction, read in for many post-SCF methods.
 - Save initial guesses for tricky to converge molecules.

zachglick:
## Description
This is a first pass at implementing serialization of the Wavefunction object as suggested by @dgasmith in #887 . This functionality allows a wavefunction to be read to or written from a file:

closes #887 

```
# get a wavefunction from some calculation
e, wfn_old = energy('mp2', return_wfn=True)

# write the wavefunction to file
Wavefunction.to_file(wfn_old, filename)

# ...
# other calculations may occur
# ...

# read the wavefunction from file
wfn_new = Wavefunction.from_file(filename)
```

The wavefunction can also be stored directly in memory as a python dictionary:
```
# write the wavefunction to file
wfn_dict = Wavefunction.to_file(wfn_old)

# ...
# other calculations may occur
# ...

# read the wavefunction from file
wfn_new = Wavefunction.from_file(wfn_dict)
```

In this code snippet, `wfn_new` and `wfn_old` should be identical (by value, not reference). A test case was added that corresponds to this example.

## Todos
- [x] Read/Write Wavefunction object to file
- [x] Serialize a few more variables (PCM related stuff, etc.)
- [x] ~~Deal with Wavefunction inheritance~~ (future PR)

## Questions
- [x] Does anyone have thoughts on the file naming scheme? Currently, the user specifies the filename in the call and a `.npy` file is generated in the working directory. This allows the user to have an arbitrary number of saved wavefunctions that persist between jobs. Alternatively, using a reserved filename (similar the `scf guess` procedure) might be better because users wouldn't have to work with filenames at all, but they would be limited to a single wavefunction.
- [x] ~~How should classes that are derived from Wavefunction (like CIWavefunction) be dealt with?~~
- [ ] Is everything implemented in a reasonable location? 
- [ ] Any inconsistencies in format/style?

## Checklist
- [x] Added Test `serial-wfn`

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -126,8 +126,14 @@ core.Wavefunction.build = pybuild_wavefunction
 
 
 @staticmethod
-def pyread_wavefunction(filename):
-    wfn_data = np.load(filename + '.npy').item()
+def pyread_wavefunction(wfn_data):
+
+    # load the wavefunction from file
+    if isinstance(wfn_data, str):
+        wfn_data = np.load(wfn_data + '.npy').item()
+    # otherwise a dictionary was passed in
+    else:
+        pass
     
     # variable type specific dictionaries to be passed into C++ constructor
     wfn_matrix = wfn_data['matrix']
@@ -185,9 +191,13 @@ core.Wavefunction.from_file = pyread_wavefunction
 
 
 @staticmethod
-def pywrite_wavefunction(wfn, filename):
+def pywrite_wavefunction(wfn, filename=None):
     # collect the wavefunction's variables in a dictionary indexed by varaible type
     # some of the data types have to be made numpy-friendly first
+
+    if wfn.basisset().name().startswith("anonymous"):
+        raise ValidationError("Cannot serialize wavefunction with custom basissets.")
+
     wfn_data = {
         'molecule' : wfn.molecule().to_dict(),
         'matrix' : {
@@ -208,8 +218,6 @@ def pywrite_wavefunction(wfn, filename):
             'epsilon_a'            : wfn.epsilon_a().to_array()   if wfn.epsilon_a() else None,
             'epsilon_b'            : wfn.epsilon_b().to_array()   if wfn.epsilon_b() else None,
             'frequencies'          : wfn.frequencies().to_array() if wfn.frequencies() else None
-            #'atomic_point_charges' : wfn.atomic_point_charges().to_array() if wfn.atomic_point_charges() else None,
-            #'esp_at_nuclei'        : wfn.esp_at_nuclei().to_array() if wfn.esp_at_nuclei() else None,
             },
         'dimension' : {
             'doccpi'   : wfn.doccpi().to_tuple(),
@@ -252,7 +260,10 @@ def pywrite_wavefunction(wfn, filename):
         'matrixarr' : {k: v.to_array() for k, v in wfn.arrays().items()}
         }
 
-    np.save(filename,wfn_data)
+    if filename is not None:
+        np.save(filename,wfn_data)
+    else:
+        return wfn_data
 
 core.Wavefunction.to_file = pywrite_wavefunction
 
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -137,6 +137,8 @@ def pyread_wavefunction(filename):
     wfn_string = wfn_data['string']
     wfn_boolean = wfn_data['boolean']
     wfn_float = wfn_data['float']
+    wfn_floatvar = wfn_data['floatvar']
+    wfn_matrixarr = wfn_data['matrixarr']
 
     # reconstruct molecule from dictionary representation
     wfn_molecule = wfn_data['molecule']
@@ -149,14 +151,13 @@ def pyread_wavefunction(filename):
 
     basis_puream = wfn_boolean['basispuream']
     basisset = core.BasisSet.build(molecule, 'ORBITAL', basis_name, puream=basis_puream)
-    core.print_out('printing type of basis set\n')
-    core.print_out(str(type(basisset)))
 
     # change some variables to psi4 specific data types (Matrix, Vector, Dimension)
-    # TODO: Make sure python's 'None' translates sensibly to C++
     for label in wfn_matrix:
         array = wfn_matrix[label]
-        wfn_matrix[label] = core.Matrix.from_array(array,name=label) if array else None
+        #core.print_out(label + '\n')
+        #core.print_out(str(type(array))+'\n')
+        wfn_matrix[label] = core.Matrix.from_array(array,name=label) if array is not None else None
     
     for label in wfn_vector:
         array = wfn_vector[label]
@@ -166,8 +167,20 @@ def pyread_wavefunction(filename):
         tup = wfn_dimension[label]
         wfn_dimension[label] = core.Dimension.from_list(tup,name=label) if tup else None
 
+    for label in wfn_matrixarr:
+        array = wfn_dimension[label]
+        wfn_dimension[label] = core.Matrix.from_array(array,name=label) if array else None
+
+    # make the wavefunction 
     wfn = core.Wavefunction(molecule, basisset, wfn_matrix, wfn_vector, wfn_dimension, wfn_int,
                             wfn_string, wfn_boolean, wfn_float)
+
+    # some of the wavefunction's variables can be changed directly
+    for k, v in wfn_floatvar.items():
+        wfn.set_variable(k,v)
+    for k, v in wfn_matrixarr.items():
+        wfn.set_array(k,v)
+
     return wfn
 
 core.Wavefunction.from_file = pyread_wavefunction
@@ -216,22 +229,26 @@ def pywrite_wavefunction(wfn, filename):
             'nfrzc'  : wfn.nfrzc(),
             'nirrep' : wfn.nirrep(),
             'nmo'    : wfn.nmo(),
-            'nso'    : wfn.nso()
+            'nso'    : wfn.nso(),
+            'print'  : wfn.get_print(),
             },
         'string' : {
             'name' : wfn.name(),
             'basisname' : wfn.basisset().name()
             },
         'boolean' : {
-            'PCM_enabled'   : wfn.PCM_enabled(),
-            'same_a_b_dens' : wfn.same_a_b_dens(),
-            'same_a_b_orbs' : wfn.same_a_b_orbs(),
-            'basispuream' : wfn.basisset().has_puream()
+            'PCM_enabled'    : wfn.PCM_enabled(),
+            'same_a_b_dens'  : wfn.same_a_b_dens(),
+            'same_a_b_orbs'  : wfn.same_a_b_orbs(),
+            'density_fitted' : wfn.density_fitted(),
+            'basispuream'    : wfn.basisset().has_puream()
             },
         'float' : {
             'energy' : wfn.energy(),
-            #'efzc' : wfn.efzc()
-            }
+            'efzc' : wfn.efzc()
+            },
+        'floatvar' : wfn.variables(),
+        'matrixarr' : {k: v.to_array() for k, v in wfn.arrays().items()}
         }
 
     np.save(filename,wfn_data)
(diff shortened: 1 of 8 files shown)


psi4/driver/p4util/python_helpers.py
```diff
@@ -123,6 +123,120 @@ def pybuild_wavefunction(mol, basis=None):
...
+            'nalpha' : wfn.nalpha(),
+            'nbeta'  : wfn.nbeta(),
+            'nfrzc'  : wfn.nfrzc(),
+            'nirrep' : wfn.nirrep(),
+            'nmo'    : wfn.nmo(),
+            'nso'    : wfn.nso()
+            },
+        'string' : {
+            'name' : wfn.name(),
+            'basisname' : wfn.basisset().name()
```
dgasmith: @loriab I think `wfn.basisset().something` should contain the python basis dict so that we can serialize custom basis sets?

psi4/driver/p4util/python_helpers.py
```diff
@@ -123,6 +123,120 @@ def pybuild_wavefunction(mol, basis=None):
...
+    if ".gbs" in basis_name:
+        basis_name = basis_name.split('/')[-1].replace('.gbs', '')
+
+    basis_puream = wfn_boolean['basispuream']
+    basisset = core.BasisSet.build(molecule, 'ORBITAL', basis_name, puream=basis_puream)
+    core.print_out('printing type of basis set\n')
+    core.print_out(str(type(basisset)))
+
+    # change some variables to psi4 specific data types (Matrix, Vector, Dimension)
+    # TODO: Make sure python's 'None' translates sensibly to C++
```
dgasmith: I think None should translate to `nullptr` always here? Worth checking.

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -85,6 +85,70 @@ Wavefunction::Wavefunction(std::shared_ptr<Molecule> molecule, std::shared_ptr<B
     common_init();
 }
 
+Wavefunction::Wavefunction(std::shared_ptr<Molecule> molecule, std::shared_ptr<BasisSet> basisset, 
+                           std::map<std::string, std::shared_ptr<Matrix>> matrices,
+                           std::map<std::string, std::shared_ptr<Vector>> vectors,
+                           std::map<std::string, Dimension> dimensions, std::map<std::string, int> ints, 
+                           std::map<std::string, std::string> strings, std::map<std::string, bool> booleans, 
+                           std::map<std::string, float> floats)
+    : options_(Process::environment.options),
```
dgasmith: Might want to add a point here that it would be good to change this in the future when we can pass around option objects rather than globals.

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -85,6 +85,70 @@ Wavefunction::Wavefunction(std::shared_ptr<Molecule> molecule, std::shared_ptr<B
...
+      basisset_(basisset),
```
dgasmith: Does this build the sobasisset as well?

psi4/driver/p4util/python_helpers.py
```diff
@@ -123,6 +123,120 @@ def pybuild_wavefunction(mol, basis=None):
...
+        'boolean' : {
+            'PCM_enabled'   : wfn.PCM_enabled(),
+            'same_a_b_dens' : wfn.same_a_b_dens(),
+            'same_a_b_orbs' : wfn.same_a_b_orbs(),
+            'basispuream' : wfn.basisset().has_puream()
+            },
+        'float' : {
+            'energy' : wfn.energy(),
+            #'efzc' : wfn.efzc()
+            }
```
dgasmith: Need to add variables/arrays here:

```
"variables": wfn.variables()
"arrays": {k: v.to_array() for k, v in wfn.arrays().items()}
```

The can be set by `set_variable/array`.

psi4/driver/p4util/python_helpers.py
```diff
@@ -123,6 +123,120 @@ def pybuild_wavefunction(mol, basis=None):
...
+            'nsopi'    : wfn.nsopi().to_tuple(),   
+            'soccpi'   : wfn.soccpi().to_tuple()  
+            },
+        'int' : { 
+            'nalpha' : wfn.nalpha(),
+            'nbeta'  : wfn.nbeta(),
+            'nfrzc'  : wfn.nfrzc(),
+            'nirrep' : wfn.nirrep(),
+            'nmo'    : wfn.nmo(),
+            'nso'    : wfn.nso()
```
dgasmith: `print_lvl_`,

psi4/driver/p4util/python_helpers.py
```diff
@@ -123,6 +123,120 @@ def pybuild_wavefunction(mol, basis=None):
...
+            },
+        'string' : {
+            'name' : wfn.name(),
+            'basisname' : wfn.basisset().name()
+            },
+        'boolean' : {
+            'PCM_enabled'   : wfn.PCM_enabled(),
+            'same_a_b_dens' : wfn.same_a_b_dens(),
+            'same_a_b_orbs' : wfn.same_a_b_orbs(),
+            'basispuream' : wfn.basisset().has_puream()
```
dgasmith: `density_fitted` here.

psi4/driver/p4util/python_helpers.py
```diff
@@ -123,6 +123,120 @@ def pybuild_wavefunction(mol, basis=None):
...
+            },
+        'float' : {
+            'energy' : wfn.energy(),
+            #'efzc' : wfn.efzc()
```
dgasmith: Why comment out `efzc`? We might consider setting this to `0` on C++ init if you hit a segfault here when calling.

psi4/driver/p4util/python_helpers.py
loriab: Will probably be important to decide and enforce what the minimal expected fields are for a serialized Wfn. There's some "dummy wfn"s out there with only mol & basis set contents.

psi4/driver/p4util/python_helpers.py
loriab: Yeah, I had hoped to blithely ignore BasisSet serialization until we jumped directly to Ben's schema. Preferences for the py basis dict vs the QCSchema route, @dgasmith?

And aux basis also should be attached, methinks. use the basis dictionary, and then pluck off "orbital" to be the `Wfn.basisset()`.

tests/serial-wfn/input.dat
```diff
@@ -0,0 +1,47 @@
...
+if wfn_old.epsilon_b():   compare_vectors(wfn_old.epsilon_b(), wfn_new.epsilon_b(), 5, 'compare epsilon_b') #TEST
+if wfn_old.frequencies(): compare_vectors(wfn_old.frequencies(), wfn_new.frequencies(), 5, 'compare frequencies') #TEST
+compare_integers(wfn_old.nalpha(), wfn_new.nalpha(), 'compare nalpha') #TEST
+compare_integers(wfn_old.nbeta(), wfn_new.nbeta(), 'compare nbeta') #TEST
+compare_integers(wfn_old.nfrzc(), wfn_new.nfrzc(), 'compare nfrzc') #TEST
+compare_integers(wfn_old.nirrep(), wfn_new.nirrep(), 'compare nirrep') #TEST
+compare_integers(wfn_old.nmo(), wfn_new.nmo(), 'compare nirrep') #TEST
+compare_integers(wfn_old.nso(), wfn_new.nso(), 'compare nso') #TEST
+compare_strings(wfn_old.name(), wfn_new.name(), 'compare name') #TEST
+compare_values(wfn_old.energy(), wfn_new.energy(), 5, 'compare energy') #TEST
```
loriab: Probably worth writing a `def compare_wavefunction()` to join the `compare_*` suite (live in driver/p4utils somewhere). Better to compare the numpy form, rather than the Matrix form and call `compare_dict` with required fields. Sound ok, @dgasmith ?

dgasmith: Sounds good, probably best to just compare base wavefunction quantities.

psi4/driver/p4util/python_helpers.py
dgasmith: If we can get QCSchema in relatively soon, I would say keep just name for now (perhaps with a check if custom) and do QCSchema before 1.3?

psi4/driver/p4util/python_helpers.py
dgasmith: mol & basis wavefunctions should initialize all of the above IIRC, if not we might want to consider doing so.

psi4/driver/p4util/python_helpers.py
zachglick: Yeah, it does exactly that

psi4/driver/p4util/python_helpers.py
zachglick: Will do. That's a nice way of iterating through dictionaries.

By the way, I noticed that the wavefunction's `arrays` is always empty. This was true for both different methods (scf, mp2, ccsd, b3lyp) and different kinds of calculations (energy, gradient, hessian). Is that a bug, or have a just not encountered a calculation that actually uses it?

dgasmith: It is not very common, but it does happen in a number of places where arrays are set. This is also behavior that we are trying to encourage to give users more programmatic access to internal quantities so it should certainly trend in that direction.

psi4/driver/p4util/python_helpers.py
zachglick: I meant to update the comment with more info: for some reason, there isn't an ` wfn.efzc()` python method, although `efzc()` is defined on the C++ side. It wouldn't be any trouble to add this in, but I didn't know if this was an oversight or an intentional decision.

Should I add the function then?

tests/serial-wfn/input.dat
zachglick: I like that idea, I can work on it alongside this PR

psi4/src/psi4/libmints/wavefunction.cc
zachglick: Will do

psi4/driver/p4util/python_helpers.py
dgasmith: Yes, please. Any Wavefunction attribute not bound to Python is an oversight.

psi4/src/psi4/libmints/wavefunction.cc
zachglick: Not yet, about to add. Do you think it would be better to try and save the SO basis or recreate it in the same way that the normal constructor does?

https://github.com/psi4/psi4/blob/3c26e8eb81c20bce3fe0223d44606c5f9be59b81/psi4/src/psi4/libmints/wavefunction.cc#L376-L379

psi4/driver/p4util/python_helpers.py
zachglick: I left this out for the same reason as `efzc()`: it appears that this boolean isn't accessible through the python interface. Should that be changed?

psi4/src/psi4/libmints/wavefunction.cc
dgasmith: The above two quantities can be deterministically created from a basisset, no need to save.

PR #1314, opened 2018-10-26
Replace calls to deprecated LAPACK subroutines

### LINKED ISSUE #1290: Replace deprecated `dtzrqf` with alternative
**psi4** uses the LAPACK routine `dtzrqf`. However, at least in the current LAPACK 3.8.0, the routine `dtzrqf` is marked as deprecated, the [suggested alternative](http://www.netlib.org/lapack/explore-html/da/dba/group__double_o_t_h_e_rcomputational_ga2d309addf953bcdd497d58608a7f7935.html) being `dtzrzf`.

As a result, the default build of LAPACK 3.8.0 (<del>as well as of OpenBLAS 0.3.3</del> EDITED) does not contain `dtzrqf`anymore and fails to provide the necessary dependency. Perhaps **psi4** could use the alternative routine to avoid this problem?

robertodr:
## Description
Fixes #1290 by wrapping calls to deprecated (and removed in Netlib's LAPACK 3.8.0) with appropriate replacements.

**UPDATE**
I have added a `PSI_DEPRECATED` macro in `pragma.h` to be used to deprecate functions, classes, variables. It will emit a warning at compile-time.

## Todos
- [x] Added a `PSI_DEPRECATED` macro
- [x] Wrap call to `DGEGS` with `DGGES`. Deprecate `C_DGEGS`.
- [x] Wrap call to `DTZRQF` with `DTZRZF`. Deprecate `C_DTZRQF`.
- [x] Wrap call to `DGEGV` with `DGGEV`. Deprecate `C_DGEGV`
- [x] Wrap call to `DGELSX` with `DGELSY`. Deprecate `C_DGELSX`
- [x] Wrap call to `DGEQPF` with `DGEQP3`. Deprecate `C_DGEQPF`
- [x] Check whether `DGGSVD3` and `DGGSVP3` are available in `tgt::lapack`. Fence off their use accordingly based on the `LAPACK_HAS_DGGSVD3` and `LAPACK_HAS_DGGSVP3` preprocessor variables, respectively. 
- [x] Wrap call to `DGGSVD` with `DGGSVD3`, if the latter is available. Deprecate `C_DGGSVD`.
- [x] Wrap call to `DGGSVP` with `DGGSVP3`, if the latter is available. Deprecate `C_DGGSVP`

## Checklist
None of these subroutines were used in Psi4, so I only ran smoketests locally.

## Status
- [x] Ready for review

### DIFF
--- a/psi4/include/psi4/pragma.h
+++ b/psi4/include/psi4/pragma.h
@@ -155,4 +155,24 @@
 #define PSI_API PSI_HELPER_SO_EXPORT
 #define PSI_LOCAL PSI_HELPER_SO_LOCAL
 
+// Use in the header file as follows:
+// PSI_DEPRECATED("extremely unsafe, use 'combust' instead!!!") void explode(void);
+// will produce this kind output when compiling:
+//    warning: 'explode' is deprecated: extremely unsafe, use 'combust' instead!!!
+// The macro can similarly be used to deprecate variables.
+// The implementation uses the standard attribute if C++14 available, falling back
+// to compiler extensions if C++11 is used.
+#if __cplusplus >= 201402L
+#define PSI_DEPRECATED(msg) [[deprecated(msg)]]
+#else
+#if defined(__GNUC__) || defined(__clang__)
+#define PSI_DEPRECATED(msg) __attribute__((deprecated(msg)))
+#elif defined(_MSC_VER)
+#define PSI_DEPRECATED(msg) __declspec(deprecated(msg))
+#else
+#pragma message("WARNING: You need to implement PSI_DEPRECATED for this compiler")
+#define PSI_DEPRECATED
+#endif
+#endif
+
 #endif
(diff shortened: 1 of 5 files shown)


robertodr:
None of these subroutines are used anywhere in Psi4, so this PR will only be useful in the off chance that the bindings are being used by some plugin.

andysim:
You're right about the potential for plugins.  My vote is to ping @fevangelista to see if any of these are needed in Forte.  If not, I suggest we remove them and encourage plugin writers to file issues for missing bindings.

loriab:
Pity it's the new versions not in travis rather than the old version. Can use conda mkl for the lot, of course, but I thought it nice that this one non-mkl psi got built and checked regularly. Don't suppose there's lapack version macros to `#ifdef`?

susilehtola:
CMake has ```CHECK_FORTRAN_FUNCTION_EXISTS(FUNCTION VARIABLE)``` which menas that it shouldn't be too difficult: check if the new one exists, otherwise use the old one. Right?
https://cmake.org/cmake/help/v3.0/module/CheckFortranFunctionExists.html

loriab:
* I think the old versions of routines we want to outright delete (when some cutting edge LAPACK comes out w/o those symbols, we want to be able to link to it).
* Agree that there's no way we want to get math detection involved.
* I think the new versions of routines should be added, checked that they link properly against a `DGGSVD3`-capable LAPACK, then commented out, and a note left in developers wish list to check travis periodically for reactivation.

PR #1316, opened 2018-10-28
Molecule Provenance

loriab:
## Description
QCSchema has comment and provenance fields. With this PR (and molssi/qcelemental#7), the `qcdb.Molecule` and `psi4.core.Molecule` classes can passthrough these objects.

## Checklist
- [x] full tests run

## Status
In good shape, but I'll probably add more.
- [x] Ready for review
- [x] Ready for merge


loriab:
This has been rebased and updated for qcschema compatibility. It goes with https://github.com/MolSSI/QCElemental/pull/12 (and won't work without it). All quicktests pass, but I wouldn't say it's been closely looked at. Particularly uncertain of how to deepcopy these structures in c++. @dgasmith

loriab:
https://github.com/psi4/psi4/pull/1316/files#diff-ed72addaa8414fc1c5e4b58b14f3b1d1R3 will make sure builds with conda pkg 0.1.2 don't use that and instead build and use an internal 0.1.3. But b/c that's an ATLEAST not EXACT, ppl with 0.1.3 in conda and building pre-this-PR psi4 will get errors at runtime. unfortunate, but that's what we get for violating semantic versioning. hopefully far fewer of these in future as approach 1.

PR #1312, opened 2018-10-25
Export compile commands for psi4-core

robertodr:
## Description
This is a trivial one. It just sets the `CMAKE_EXPORT_COMPILE_COMMANDS` option to `ON` when setting up the `psi4-core` external project.
Upon generation, CMake will dump the compile options for each file in `psi4-core` to a `compile_commands.json` file in `<build-dir>/psi4-core-prefix/src/psi4-core-build`. This file can be used by `clang-tidy` to perform various static checks and apply corresponding fixes. For the latter, PRs are incoming.

## Checklist
- [ ] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1319, opened 2018-10-30
clang-tidy apply modernize-use-override fixes

robertodr:
## Description
Uses `clang-tidy` to find and fix uses of `virtual` that can be replaced by the `override` keyword introduced in C++11 and later standards. Fixes applied with:
```
cd <build-dir>/psi4-core-prefix/src/psi4-core-build
run-clang-tidy.py -header-filter='.*' -checks='-*,modernize-use-override' -fix
```
Based on #1312 

## Todos
- [x] `clang-tidy` find and fix with `modernize-use-override`

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/prog_style_c.rst
+++ b/doc/sphinxman/source/prog_style_c.rst
@@ -46,7 +46,7 @@ use ``nullptr`` to signal the null pointer. The correct overload/template
 parameter will then be deduced. Using ``nullptr`` also makes the code more
 readable, especially if ``auto`` is used consistently throughout.
 
-*Reference:* Item 8 in `[Effective Modern C++] <https://edisciplinas.usp.br/pluginfile.php/1995323/mod_resource/content/1/Effective%20Modern%20C%2B%2B%202014.pdf>`_
+*Reference:* Item 8 in `[Effective Modern C++] <https://isbnsearch.org/isbn/9781491903995>`_
 
 
 .. _`faq:automakeshared`:
@@ -77,7 +77,7 @@ Using ``std::make_shared``:
     // Performs ONE allocation
     auto F = std::make_shared<Matrix>("Fock matrix", nso, nso);
 
-*Reference:* Item 21 in `[Effective Modern C++] <https://edisciplinas.usp.br/pluginfile.php/1995323/mod_resource/content/1/Effective%20Modern%20C%2B%2B%202014.pdf>`_
+*Reference:* Item 21 in `[Effective Modern C++] <https://isbnsearch.org/isbn/9781491903995>`_
 
 
 .. _`faq:autodecl`:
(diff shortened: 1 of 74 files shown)


andysim:
I'm really glad you figured out clang-tidy, @robertodr .  This will be really useful for getting rid of all the warnings on appveyor.  I agree with @dgasmith and will issue a PR to this branch soon, documenting the `override` in the C++ section of the dev manual.

loriab:
Yes, this `clang-tidy` is really nifty. A series of short lessons reprogramming my notions of good C++ syntax.

doc/sphinxman/source/prog_style_c.rst
```diff
@@ -108,8 +108,15 @@ Using ``auto`` reduces and/or avoids:
...
     auto size = v.size();  // size is ALWAYS of the correct type
 
-*Reference:* Items 2 and 5 in `[Effective Modern C++] <https://edisciplinas.usp.br/pluginfile.php/1995323/mod_resource/content/1/Effective%20Modern%20C%2B%2B%202014.pdf>`_
+*Reference:* Items 2 and 5 in `[Effective Modern C++] <https://isbnsearch.org/isbn/9781491903995>`_
 
+Mark virtual functions in derived classes with override
+-------------------------------------------------------
+
+The ``override`` keyword introduced in C++11 is used to mark a function in a
+derived class and guarantee that it is overloadinga function *with the same
```
loriab: --> "overloading a"

Can catch this on another PR.

psi4/src/psi4/libmints/pseudospectral.h
```diff
@@ -95,7 +95,7 @@ class PseudospectralInt : public OneBodyAOInt {
     void use_omega(bool yes) { use_omega_ = yes; }
 
     /// Does the method provide first derivatives?
-    bool has_deriv1() { return false; }
+    bool has_deriv1() override { return false; }
```
loriab: Does it make sense to have a default implementation for the parent class and also to demand (I think, through `override`) that the derived implement it?

robertodr: Having these virtual methods with implementation in the base class means that all subclasses will conform to the same interface. For the particular case of `has_deriv1` and `has_deriv2` only subclasses that have 1st and 2nd derivative must override the default implementation (so that it returns `true`). For all other subclasses there might be a slight performance penalty.

loriab: got it. `override` demands derived have same interface, not that derived reimplement.

robertodr: That's a demand coming from `virtual` in the base class, actually. `override` is there to declare that the method in the derived class is overriding the one in the base class, such that the compiler can actually check we're doing it correctly.

andysim: Just repeating stuff i'm sure, but my understanding is that it's to avoid something like the following.  If we have some function in the class Base `virtual Base::func(float arg)`.  If there's a class Derived that derives from Base with a function `Derived::func(int arg)` it could be something the programmer intended to implement to overload the base class but accidentally got the wrong arg type; it would compile (probably with warnings) because the int can be converted to a float.  The problem is that a `Base*` pointer would then call the version from Base, while a `Derived*` pointer would call the version from Derived because `Derived::func` doesn't overload `Base::func` -- it's a different function -- when both pointers should result in a call to `Derived::func`.  The `overload` descriptor just guarantees that similarly named functions have the same interface to avoid this confusion.

PR #1330, opened 2018-11-04
expand pytest and qcelemental

loriab:
## Description

## Todos
- [x] switch to qcel for covalent radii. require v0.1.2
- [x] to try out covalent radii, add a test for translating to MOL format. add to psi4.core.Mol.
- [x] fix the compare_vals in threading_test
- [x] clean up all but json_out files from ctest run
- [x] add levels to pytest. while `psi4 --test` still does smoke only, other access modes have smoke/quick/default/long and we can start expanding the pytest suite

## Checklist
- [x] added MOL writer test

## Status
- [x] Ready for review
- [x] Ready for merge


tests/pytest/addons.py
```diff
@@ -58,6 +58,9 @@ def is_nvidia_gpu_present():
 using_memory_profiler = pytest.mark.skipif(_plugin_import('memory_profiler') is False,
                                 reason='Not detecting module memory_profiler. Install package if necessary and add to envvar PYTHONPATH')
 
+using_psi4 = pytest.mark.skipif(False,
+                                reason='Note detecting module psi4. Install package if necessary and add to envvar PYTHONPATH')
```
andysim:
```suggestion
                                reason='Not detecting module psi4. Install package if necessary and add to envvar PYTHONPATH')
```

loriab: Good catch. And since you're probably wondering why `@using_psi4` is defined for a test in the psi4 test suite, the answer is so that I can use the same test file in another repo, where psi4 isn't mandatory.

andysim: I suspected that may be the case, given your propensity to write reusable components.  I've never tried the suggested change thingy, but apparently it will let you click a button to accept the diff.  More cool stuff from GH :)

loriab: ooh, I've never hit that button before. very nifty.

PR #1336, opened 2018-11-06
Fix 1288: ODC Compatible with Other OO Code

### LINKED ISSUE #1288: PSIO Error Using ODC after OMP2/OLCCD on UHF Reference
The following input file
```
molecule mol {
  0 1 
  units bohr
 O 0 0 -0.143225816552
 H 0 1.638036840407 1.136548822547
 H 0 -1.638036840407 1.136548822547
}

set {
  reference uhf 
  basis sto-3g
  scf_type pk
  mp2_type conv
  dcft_functional odc-12
}

energy('omp2')
energy('dcft')
```
crashes with error
```
PSIO_ERROR: Attempt to write into next entry: 61, MO Ints (OO|OV)
PSIO_ERROR: unit = 61, errval = 18
PSIO_ERROR: 18 (Incorrect block end address)
  File "/Users/jonathonmisiewicz/psi4/objdir/stage/bin/psi4", line 271, in <module>
    exec(content)
  File "<string>", line 33, in <module>
  File "/Users/jonathonmisiewicz/psi4/objdir/stage/lib/psi4/driver/driver.py", line 493, in energy
  File "/Users/jonathonmisiewicz/psi4/objdir/stage/lib/psi4/driver/procrouting/proc.py", line 1479, in run_dcft
    dcft_wfn = core.dcft(ref_wfn)

RuntimeError: 
Fatal Error: PSIO Error
Error occurred in file: /Users/jonathonmisiewicz/psi4/psi4/src/psi4/libpsio/error.cc on line: 128
```
Modifying the input file showed the following:

- The error resolves when changing the reference from UHF to RHF
- Both ODC-06 and ODC-12 will trigger the error, but neither DC-06 nor DC-12 will
- Both OMP2 and OLCCD before the ODC will trigger the error, but neither MP2 nor LCCD will

JonathonMisiewicz:
## Description
Fixes #1288. The trouble was that Bozkaya's code used libtrans to transform integrals, and those were written to disk in their antisymmetrized, redundancy-free form, as libtrans does. The DCT code gets different integrals from libtrans but then sorts those to get the same blocks as Bozkaya wrote earlier in the redundant form. Because the DCT-sorted integral block had antisymmetry redundancies, it was larger than the libtrans block that did not have those redundancies.

Fixed by changing the code to use a block without those redundancies.

I imagine this is of interest to @amjames and @ssh2.

## Todos
- [x] Fixes 1288

## Checklist
- [x] DCFT tests run

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dcft/dcft_gradient_UHF.cc
+++ b/psi4/src/psi4/dcft/dcft_gradient_UHF.cc
@@ -1466,7 +1466,7 @@ double DCFTSolver::compute_response_coupling() {
     // zI_IJ = (IJ|KC) z_KC
     global_dpd_->file2_init(&zI, PSIF_DCFT_DPD, 0, ID('O'), ID('O'), "zI <O|O>");
     global_dpd_->file2_init(&z, PSIF_DCFT_DPD, 0, ID('O'), ID('V'), "z <O|V>");
-    global_dpd_->buf4_init(&I, PSIF_LIBTRANS_DPD, 0, ID("[O,O]"), ID("[O,V]"), ID("[O,O]"), ID("[O,V]"), 0,
+    global_dpd_->buf4_init(&I, PSIF_LIBTRANS_DPD, 0, ID("[O,O]"), ID("[O,V]"), ID("[O>=O]+"), ID("[O,V]"), 0,
                            "MO Ints (OO|OV)");
     global_dpd_->contract422(&I, &z, &zI, 0, 0, 1.0, 0.0);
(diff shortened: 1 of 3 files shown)


ssh2:
Great, thanks, Jonathon. The fix looks good to me.

ssh2:
Pretty sure this was me @andysim, but, to my defense, O^3 V integrals don't usually cost much storage. Anyway, great job locating the problem, Jonathon.

PR #1346, opened 2018-11-09
adds nZaPa-NR 

hokru:
## Description
Adds the nZaPa basis sets that converge faster for SCF and correlation energies than many others.

## Todos
- [x] nZaPa-NR basis sets for n=2,3,4,5,6,7
- [x] RI and JKFIT definitions for nZaPa and cc-pVnZ-F12
- [x] `[23]ZaPa-NR` bracket detection for CBS

## Checklist
- [x] relevant fraction of full tests run

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/qcdb/basislistother.py
```diff
@@ -505,3 +505,56 @@ def load_basfam_other():
...
+    basis_5zapa_nr.add_rifit('aug-cc-pv5z-ri')
+    basis_5zapa_nr.add_rifit('aug-cc-pv5z-ri')
+    basis_6zapa_nr.add_rifit('aug-cc-pv6z-ri')
+
+    basisfamily_list.append(basis_2zapa_nr)
+    basisfamily_list.append(basis_3zapa_nr)
+    basisfamily_list.append(basis_4zapa_nr)
+    basisfamily_list.append(basis_5zapa_nr)
+    basisfamily_list.append(basis_6zapa_nr)
+    basisfamily_list.append(basis_7zapa_nr)
```
loriab: try running a 7zapa calc density fitted and make sure it throws an error for insufficient jkfit, rather than defaulting to def2-qzvpp, which is the universal default and definitely insufficient.

psi4/driver/qcdb/basislistother.py
```diff
@@ -505,3 +505,56 @@ def load_basfam_other():
...
+    basis_cc_pvqz_f12.add_jkfit('cc-pv5z-jkfit')
+
+    basis_cc_pvdz_f12.add_rifit('cc-pvtz-jkfit')
+    basis_cc_pvtz_f12.add_rifit('cc-pvqz-jkfit')
+    basis_cc_pvqz_f12.add_rifit('cc-pv5z-jkfit')
+
+    basisfamily_list.append(basis_cc_pvqz_f12)
+    basisfamily_list.append(basis_cc_pvtz_f12)
+    basisfamily_list.append(basis_cc_pvqz_f12)
+    # basisfamily_list.append(basis_cc_pv5z_f12)
```
loriab: Could you add a table of the new basisfamily to https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/basissets_tables.rst?

psi4/driver/qcdb/basislistother.py
```diff
@@ -505,3 +505,56 @@ def load_basfam_other():
     basis_6311ppg_3df_2p_.add_rifit('aug-cc-pvtz-ri')
     basis_6311ppg_3df_2pd_.add_rifit('aug-cc-pvtz-ri')
     basis_6311ppg_3df_3pd_.add_rifit('aug-cc-pvtz-ri')
+
+    # Peterson's nZaPa-NR basis sets
+    basis_2zapa_nr = BasisFamily('3zapa-nr',zeta=2)
```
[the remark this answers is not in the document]
hokru: yes, fixed it in next commit

psi4/driver/qcdb/basislistother.py
hokru: ok, added.

hokru:
turns out the downloaded basis sets files from the SI don't comply with the expected angular momentum notation. E.g. there is a J where a K should be. (psi4 does G-H-I-K-L etc.)

susilehtola:
> turns out the downloaded basis sets files from the SI don't comply with the expected angular momentum notation. E.g. there is a J where a K should be. (psi4 does G-H-I-K-L etc.)

Great, a third standard.

psi4/driver/qcdb/basislistother.py
hokru: It defaults to def2/J. Is there already somewhere such an error handler?

doc/sphinxman/source/basissets_tables.rst
```diff
@@ -301,6 +302,22 @@ Appendix :ref:`apdx:basisElement`.
 |
 |
 
+.. _`table:basisPeterson`:
+
+.. table:: Summary of (other) Peterson orbital basis sets available in |PSIfour|
```
susilehtola: It's not Peterson, it's Petersson

doc/sphinxman/source/basissets_tables.rst
```diff
@@ -40,6 +40,7 @@ Tables :ref:`Pople <table:basisPopleOrbital>`,
 :ref:`Dunning (Douglas-Kroll) <table:basisDunningDK>`, 
 :ref:`Karlsruhe <table:basisKarlsruhe>`,
 :ref:`Jensen <table:basisJensen>`,
+:ref:`Peterson <table:basisPeterson>`,
```
susilehtola: Petersson

psi4/driver/qcdb/basislistother.py
```diff
@@ -505,3 +505,65 @@ def load_basfam_other():
     basis_6311ppg_3df_2p_.add_rifit('aug-cc-pvtz-ri')
     basis_6311ppg_3df_2pd_.add_rifit('aug-cc-pvtz-ri')
     basis_6311ppg_3df_3pd_.add_rifit('aug-cc-pvtz-ri')
+
+    # Peterson's nZaPa-NR basis sets
```
susilehtola: Petersson

doc/sphinxman/source/basissets_tables.rst
hokru: oh right, I had K.A Petersson in the back of my mind, but it is G.A Peterson!

hokru: that also means the table is wrong. Since the F12 are from K.A Petersson.

loriab:
On the test suite error, it's a good idea to turn convergences way down (e.g., 10 or 12) to compute the reference values, then remove conv settings for the final test case and do the compare_values check to the number or fewer decimals than the default for the calc type (energy, gradient, etc.). You're checking the latter to 6, so probably better ref vals will do the trick.

psi4/driver/qcdb/basislistother.py
loriab: To clarify, the below with 3-zeta will run just fine because even though no JKfit is available for that element and basis, it can use def2-qzvpp (much better periodic table coverage) as a universal default b/c 3<=4. For the 6-zeta, Dunning JKfits similarly unavailable but no suitable def2 parachute, so it throws an error. The logic is happening [here](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/libmintsbasisset.py#L812-L813).

I'm just wanting to make sure that [567]zapa are behaving like cc-pv6z and erroring, rather than running with a way-too-small aux basis.

```
molecule {
1 1
Li
}

energy('hf/cc-pvtz')  # fine
energy('hf/cc-pv6z')  # error
```

3-zeta
```
  Basis Set: CC-PVTZ
    Blend: CC-PVTZ
    Number of shells: 10
    Number of basis function: 30
    Number of Cartesian functions: 35
    Spherical Harmonics?: true
    Max angular momentum: 3

   => Loading Basis Set <=

    Name: (CC-PVTZ AUX)
    Role: JKFIT
    Keyword: DF_BASIS_SCF
    atoms 1 entry LI         line    59 file /home/psilocaluser/gits/hrw-release/objdir37/stage/share/psi4/basis/def2-qzvpp-jkfit.gbs
```

6-zeta
```
BasisSetNotFound: BasisSet::construct: Unable to find a basis set for atom 1 for key BASIS among:
  Shell Entries: ['LI']
  Basis Sets: [('CC-PV6Z', 'CC-PV6Z', None)]
  File Path: /home/psilocaluser/gits/hrw-release/objdir37, /home/psilocaluser/gits/hrw-release/objdir37, /home/psilocaluser/gits/hrw-release/objdir37/stage/share/psi4/basis
```

hokru:
I think this is ready. I assigned auxiliary basis sets that I deem reliable and which I used for calculations already.

PR #1347, opened 2018-11-09
Move math bindings to CBLAS and LAPACKE

robertodr:
## Description
What it says on the tin: use standardized C bindings to BLAS and LAPACK.

## Todos
- [x] BLAS level 1, 2, and 3 moved to CBLAS
- [x] Conda install `mkl-include` on Travis
- [x] LAPACK moved to LAPACKE
- [x] Remove use of `check_fortran_function_exists` to check for newere LAPACK subroutines
- [x] Remove duplicated name-mangling headers

## Questions
- Should the bindings be someplace more central, rather than in `libqt`? It could encourage _not rewriting them_ in other modules...

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


andysim:
@jturney should've written this and called it JetPACK.  I'm not sure there's a more central location; the libqt library (math routines from the original 'quantum trio') is about as central a math library as we have.  I agree that avoiding duplication would be very good though.

robertodr:
The reason for the tests failing seems to be inconsistencies in the way LAPACK functions are used. To get this fully working I'd need to inspect _all call points_ to BLAS and LAPACK in the code. Given we agreed on transitioning to xtensor, this seems quite the waste of effort.

loriab:
So much good stuff -- should anything be salvaged before closing in favor of xtensor?

robertodr:
Some stuff can probably be salvaged, but even if this is closed I will still be able to browse the diffs.

PR #1350, opened 2018-11-10
Charge-aware frozen core

### LINKED ISSUE #1271: Calcium Basis Set Error?
Reposted from forum with modification.

The following attempt at open-shell SAPT fails with segfault `forrtl: severe (174): SIGSEGV, segmentation fault occurred`. Original reporter concluded it was a problem with Calcium, as modifying it to Chlorine (along with the charges) resulted in a successful computation. The example molecule is reduced from the original report, for ease of reproduction. 40 GB is (hopefully) enough to indicate this is not a problem of needing more memory.

```memory 40 gb
molecule {
0 1
H -0.803001 1.098737 0.000360
H -0.919454 -1.089722 0.000521
O -0.035537 -0.067631 0.000218
--
2 1
Ca 2.281447 -0.004247 -0.000154
units angstrom
no_reorient
symmetry c1
}

set globals {
scf_type df
FREEZE_CORE true
}

basis {
assign aug-cc-pVTZ
assign Ca def2-TZVPP
}

df_basis_scf {
assign aug-cc-pVTZ-jkfit
assign Ca def2-TZVPP-jkfit
}

df_basis_sapt {
assign aug-cc-pVTZ-RI
assign Ca def2-TZVPP-ri
}

energy('sapt0')```

### PR DESCRIPTION
**Pull request by PeterKraus**
## Description
This is my attempt to resolve #1271 . I would like to use this PR to hash out the FC functionality more thoroughly. The questions I have are:

- what is the ideal expected behaviour of `freeze_core True`
- do we want to implement "previous shells" freezing, using eg. `freeze_core {1,2,...}`
- how about "ridiculous" cases such as Li2+ or F2-, how much hand-holding do we want the code to do and at which point we give up?

## Todos
- [X] Resolves #1271
- [ ] Clarify 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### CODE REVIEW

**Comment by PeterKraus** [comment]
To give my 2 cents to the questions above:
- `freeze_core True` should be a best-effort solution, always freezing as much as possible, but ensuring there are valence electrons (ie. for Ca2+, [Ne] core is frozen instead of [Ar], while for Ca0, [Ar] is frozen; similarly for Ar- the whole [Ar] shell should be frozen). This one is charge aware. This option should never crash the code.
- `freeze_core N` where `N = [1,2,3...]` will always freeze the N-th previous full shell, ie for `N=1` the current behaviour in master. Validation error when no valence electrons remain, to avoid the non-helpful segfault in #1271.
- @JonathonMisiewicz's idea of per-atom frozen cores (or in my view even better: per-fragment frozen orbitals) is a good one, and I'd be keen to get that sorted, but it'd probably need changes to `molecule {}` or somewhere else.

**Comment by JonathonMisiewicz** [comment]
I agree with Peter on all points, so far. I have an unexpectedly long to-do list today, so I'll refrain from looking over the code until we're agreed on Peter's questions.

**Comment by PeterKraus** [comment]
Just to prevent things falling in between the cracks, @loriab and I have discussed this issue on Slack a while ago, and the consensus was that while adding FC spec to `Molecule` might be attractive because it's easier to do, it really ought to belong to the `Wavefunction` object. 

To paraphrase @loriab's comments, as the `Wavefunction` owns a `Basisset` which owns a `Molecule` anyway, it is the most logical place to plug this logic in. It should process options like `frozen_docc Int` or `freeze_core True` (or any possible `freeze_core N` spec where `N` is the N-th previous noble gas, or potential per-fragment `frozen_docc`). For the latter case (`freeze_core True`), it should take into account the fragment charge, or any electrons already frozen by ECP's, so that atoms such as Ca2+ with FC have some valence electrons.

**Comment by andysim** [comment]
I'm late to look at this one - sorry about that.  I fully agree with the design ideas behind this.  Yes, `True` should just err on the side of freezing questionable electrons (and, yes, it should make sure it leaves some!).  Exiting gracefully when too many electrons have been frozen by the user is also important.  I really like the idea of the fragment based approach, but I think it's too involved to be considered in the current PR.

**Comment by PeterKraus** [comment]
I've not noticed that this is meant for the 1.3 release. I'll have another look at it and will implement both `freeze_core True/False` and `freeze_core 0/-1/-2/-3` options, where `False == 0` witout FC, `True` is the "smart handling" (previous rare gas of each atom, taking into account charge), and `-1/-2/-3` would be "strict" N-th previous rare gas, ignoring charge but throwing a reasonable error when all electrons are frozen. I will have it done by Monday, if that's OK?

**Comment by andysim** [comment]
That would be excellent, if you have time.  I think a little bit of documentation and testing would be needed in that case, to alert users of the new syntax.

**Comment by loriab** [comment]
Side note, as I haven’t looked at this yet today: I was going to add some qcelemental calls to enable lookup of levels of frozen core per element. Useful for this or not so much since c-side or other reasons?

File: psi4/src/psi4/libmints/basisset.cc [hunk 1]
```diff
@@ -193,20 +196,35 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
             if (Z > 0) {
                 // Add ECPs to Z, the number of electrons less ECP-treated electrons
                 double ECP = n_ecp_core(mymol->label(A));
```
susilehtola: Why is this a ```double```?

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -193,20 +196,35 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
...
-                if (Z + ECP > 54) nfzc += 9;
-                if (Z + ECP > 86) nfzc += 16;
+                int delta = 0;
+                if (Z + ECP > 2)  delta = 2;
+                if (Z + ECP > 10) delta = 10;
+                if (Z + ECP > 18) delta = 18;
+                if (Z + ECP > 36) delta = 36;
+                if (Z + ECP > 54) delta = 54;
+                if (Z + ECP > 86) delta = 86;
                 if (Z + ECP > 108) {
```
susilehtola: This set of if statements and duplicated numbers could be replaced with a ```std::vector<int>``` and a ```lower_bound``` call, like in #1452 .

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -193,20 +196,35 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
...
-                // If this center has an ECP, some pairs are already frozen
-                if (ECP > 0) nfzc -= ECP / 2;
+                // Keep track of the largest frozen shell, in case its a cationic species
+                if (lg_delta < delta) {
+                    if (delta == 2) lg_shell = 2;
+                    if (delta == 10) lg_shell = 8;
+                    if (delta == 18) lg_shell = 8;
+                    if (delta == 36) lg_shell = 18;
+                    if (delta == 54) lg_shell = 18;
+                    if (delta == 86) lg_shell = 32;
```
susilehtola: Same here.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -182,7 +182,10 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
     if (local == "FALSE") {
         return 0;
     } else if (local == "TRUE") {
-        int nfzc = 0;
+        int nfzel = 0;
+        int valence = -1 * mymol->molecular_charge();
+        int lg_delta = 0;
+        int lg_shell = 0;
```
susilehtola: You could comment what ```lg_shell``` and ```lg_delta``` are, as it is not instantly clear from the code.

psi4/src/psi4/libmints/basisset.cc
PeterKraus: Neat, thanks!

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -173,16 +173,35 @@ int BasisSet::n_ecp_core() const {
...
     if (depth.empty()) local = Process::environment.options.get_str("FREEZE_CORE");
 
     SharedMolecule mymol = mol ? mol : molecule_;
 
-    if (local == "FALSE") {
+    if (local == "FALSE" or local == "0") {
         return 0;
-    } else if (local == "TRUE") {
-        int nfzc = 0;
+    } else if (local == "TRUE" or local == "1") {
```
andysim: Could there be a problem if the user specifies "1" intending to freeze only 1 orbital, but this line interprets it as true?

andysim: It'd be a weird corner case, admittedly.  I need more coffee to be able to think it through

PeterKraus: I think you're looking for `FROZEN_DOCC int` to freeze only `int` orbitals.

andysim: You're completely correct.  I just got down to the `read_options` and upon seeing the allowed values I realized how stupid my comment was.  Coffee time!

psi4/src/read_options.cc
```diff
@@ -109,7 +109,7 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     of orbitals to freeze, program picks the lowest-energy orbitals)
     or |globals__frozen_docc| (gives the number of orbitals to freeze per
     irreducible representation) -*/
```
andysim: This string is parsed and included in the manual to document the keyword, so this is the most important place to describe the behavior of the different values, in case you didn't know that already

tests/dfmp2-ecp/input.dat
```diff
@@ -63,3 +63,45 @@ set guess sad
...
+0 1
+Kr 3 0 0 
+no_reorient
+no_com
+}
+
+set globals {
+scf_type df
+mp2_type df
+}
```
andysim: An ECP + frozen core syntax test.  Fantastic!

psi4/src/psi4/libmints/basisset.h
```diff
@@ -281,6 +281,10 @@ class PSI_API BasisSet {
     /// Return the number of core electrons associated with this (ECP) basisset, for the specified label.
     int n_ecp_core(const std::string &label) const { return ncore_.count(label) ? ncore_.at(label) : 0; }
 
+    /// Helper functions for frozen core to reduce LOC
+    int _atom_to_period(int Z);
```
loriab: These are internal functions for the class? If so, what's signaled in py with the underscore prefix should in c++ remove the underscore and move to the private: section of the header.

JonathonMisiewicz:
History contamination aside, the need to "fix" the new frozen core and ECP test indicates that some other commit in the last few days has **changed** the computed energies for this case. I strongly recommend we figure out what the cause of this change is before pulling in.

Oddly enough, it looks like #1456 was merged before Peter's last commit that passed started running. I'm assuming that's a perfect alibi for #1456, which was going to be my prime suspect otherwise.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -173,16 +173,35 @@ int BasisSet::n_ecp_core() const {
...
+    static const std::vector<int> full_shell_values = {0, 2, 10, 18, 36, 54, 86, 118};
+    auto period = std::lower_bound(full_shell_values.begin(), full_shell_values.end(), Z) - full_shell_values.begin();
+    return period;
+}
+
+int BasisSet::period_to_full_shell(int p) {
+    if (p > 7) {
+        throw PSIEXCEPTION("Atomic number beyond Oganesson");
+    }
+    static const std::vector<int> full_shell_values = {0, 2, 10, 18, 36, 54, 86, 118};
```
susilehtola: Since full_shell_values is already a ```static const``` array, you could declare it outside the functions so you avoid having it in two places.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -192,23 +211,51 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
...
-        throw std::invalid_argument("Frozen core spec is not supported, options are {true, false}.");
+        // Options are filtered in read_options.cc; allowed strings are:
+        // TRUE, FALSE, -1, -2, -3
+        int req_shell = -std::stoi(local, nullptr, 10);
+        int num_frozen_el = 0;
+        int mol_valence = -1 * mymol->molecular_charge();
+        // Freeze the number of core electrons strictly corresponding to the
+        // requested previous n-th shell.
+        for (int A = 0; A < mymol->natom(); A++) {
+            double Z = mymol->Z(A);
```
susilehtola: Isn't this an int?

JonathonMisiewicz:
> @JonathonMisiewicz the test in question is `dfmp2-ecp` which has been unreliable for some time, way before #1456 was merged. See #1433.

I can't tell if we're disagreeing about facts or best practices.

Peter added new tests of SCF energies to `dfmp2-ecp`. A few days ago, his new tests passed. Before the force-push, his tests failed. This tipped us off that _some_ other PR changed the energy to which frozen core ECP computations converge. Peter was able to reproduce the change in energies on a non-PR branch, confirming it.

The first problem is that unless I missed something, there _shouldn't_ have been a PR that would change those energies. Apparently, there is one, and we don't know how many computations it afflicts. Probably just frozen core ECP, but there is a question mark on that one.

Second, one of these two answers has to be wrong. I cannot believe this is a case of computations being insufficiently converged or landing on different states. If Psi is giving or was giving wrong answers, then at the absolute minimum, we have an obligation to let people know. And if this test is just meant to detect changes in energies without worrying about why, that needs to be said in the test itself.

susilehtola:
What I mean is that ECPs appear to have been broken for a while in Psi4. Even a month ago, before any of the SAD guess stuff was included, the ```dfmp2-ecp``` job was mysterious. Namely, it failed in the initial version of the new SAD guess, but mysteriously was working again with the reorganized but functionally equivalent version. However, at the same time, the ```dfmp2-ecp``` job fails to converge altogether when started from the ```CORE``` guess, which really makes no sense since it's dealing with a very weakly interacting noble gas dimer, which should converge straight away.

Are the failures/differences related to the presence of ECPs, or do they also occur in the case of no ECPs?

PeterKraus:
We can argue about whether the `dfmp2-ecp` test was or wasn't broken before PR #1350 was made, but the following point remains:

- the original part of the test did is not touched by this PR, and by looking at `git blame` for that file, hasn't been touched in quite some time.
- as part of this PR, i have added a completely new section of tests, that are Kr--Kr interaction energies with DF-MP2/3-21G (no ECP's).
- a previous version of this PR, from around 2019-01-04, passed both the new and the old test.
- a later version of this PR, from until and excluding 2019-01-09, did not pass the new part of the test, but still passed the old test
- the current version of this PR has modified the new part of the test, after rebasing from master; it passes both the old test and the new test to 7 DP (I need to tighten convergence on the energy to pass it to 8 DP).

A good diff to look at the issue is here:
https://github.com/psi4/psi4/pull/1350/commits/9e6329abb493ed79dbb6913ac4c1df957e9839ae#diff-32d6050d16bb770a64f7f94a67f26101
which compares the `output.ref` of the test for the version that passed on 2019-01-04 when the test was originally written, and passes the test now (locally) on 2019-01-09 with updated values.

The puzzling thing is: what caused the change of the Kr--Kr interaction energies (and Kr MP2/3-21G energies converged to 1e-9)? What change would cause that, but not affect the first part of the test?

psi4/src/psi4/libmints/basisset.cc
PeterKraus: Doesn't look like that to me:

https://github.com/psi4/psi4/blob/9e6329abb493ed79dbb6913ac4c1df957e9839ae/psi4/src/psi4/libmints/molecule.h#L255-L256

susilehtola: Huh... why is that

susilehtola:
This is very odd, indeed. With Q-Chem I get the following results

|  | HF | MP2 |
|--|--|--|
|Kr2 dimer | -5478.3777659931 | -5478.43678797 |
|Kr atom | -2739.1975667825 | -2739.22713486 |
|interaction energy | 0.017368 | 0.017482 |

which are close to the original df values, so the new values are wrong.

susilehtola:
Rerunning the Kr --- Kr 3-21G calculation in Psi4 master gives me 0.017482 with exact integrals, and 0.017535 with density fitting.

susilehtola:
Btw since the 3-21G case doesn't even have ECPs, it should go in its own test.
Where you could check both the conventional SCF/MP2 as well as the density-fitted ones.

PeterKraus:
To summarise, it seems that running DF-MP2/3-21G is a bad idea for these small systems. The value of the interaction energy one gets, even with `E_CONVERGENCE` and `D_CONVERGENCE` at 1.0e-12, differs in the 4th DP based on which `DF_FITTING_CONDITION` is used. This caused the change in the energy during this PR (`df_fitting_condition 1.0e-10 -> 1.0e-12`)

The def2-svp basis set is not affected nearly as much, as its fitting basis is better suited for the purpose. Switching the basis set in the test and tightening the energy/density convergence criteria to 1e-10 passes agreement with 1e-12 reference to 8DP's.

psi4/src/psi4/libmints/basisset.cc
loriab: Because in the beginning there was the thought that `Z` would be effective nuclear charge and float in case any MM atoms were around. At least that's my understanding.

psi4/src/read_options.cc
```diff
@@ -101,15 +101,19 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+    shell may be unfrozen, to ensure there are valence electrons in each 
+    fragment. With ``FALSE`` or ``0``, no electrons are frozen (with the 
+    exception of electrons treated by an ECP). With ``-1``, ``-2``, and ``-3``,
+    the user might request strict freezing of the previous first/second/third
+    noble gas shell on every atom. In this case, when there are no valence 
+    electrons, the code raises an exception. More precise control over the 
+    number of frozen orbitals can be attained by using the keywords 
+    |globals__num_frozen_docc| (gives the total number of orbitals to freeze,
+    program picks the lowest-energy orbitals) or |globals__frozen_docc| (gives
+    the number of orbitals to freeze per irreducible representation) -*/
```
loriab: nice

susilehtola:
Doesn't seem to be working too great; I get on my machine
```
$ psi4 -n 8 
	SAPT0 energy with Ca2+, without ECP: computed value (-0.095615678) does not match (-0.095616042) to 8 digits.
  File "/home/work/psi4/install.susi/bin/psi4", line 287, in <module>
    exec(content)
  File "<string>", line 38, in <module>
  File "/home/work/psi4/install.susi/lib/psi4/driver/p4util/util.py", line 230, in compare_values
    raise TestComparisonError(message)

TestComparisonError: 	SAPT0 energy with Ca2+, without ECP: computed value (-0.095615678) does not match (-0.095616042) to 8 digits.
```
and the result is the same regardless of the guess...

loriab:
Thanks, your fix does the trick. I also get the above, but I'm guessing that the reference values just weren't constructed defensively. Varying convergence now ...

susilehtola:
I bet this is since #1472 was merged.

Is there no non-df version of SAPT?

loriab:
I should know this. My best recollection is that there is SAPT with conventional integrals but there might be terms that are always constructed through DF. Unless I'm mixing that up with Molpro's CCSD-F12.

Yes, ref energies were probably constructed with default ~6 convergence but checking to 8.

PeterKraus:
Apologies. I could swear I ran all references at 1.0e-12, but clearly I must have messed them up for sapt-ecp.

loriab:
No worries. Thanks for checking up on the refs. As I recall, you ran a couple dozen refs to examine the fitting metric issue.

PR #1351, opened 2018-11-10
Refactor Recursive Driver Into Distributed Driver

alenaizan:
## Description
Starting PR. Change the Psi4 driver from being recursive to being based on compute classes that have planning, computing and querying functions. Then, integrate the new driver with the QCArchive project for parallelizing nbody, CBS and findif.

## Deconstruction Progress
99 changed files, and 5,314 additions and 2,946 deletions. c. dgasmith:recursive
- #2512
- #2515
- #2518
- #2526 
- #2530 
- #2532
- #2536 
- #2544

59 changed files, and 3,408 additions and 1,885 deletions. c. loriab:recursive62
- #2552 
- #2557
- [ ] #2516

## Todos
- [ ] Create compute classes for nbody, CBS and findif
- [ ] Integrate QCArchive

## Release Notes
- `msgpack-python` added as req'd dependency. this keeps numpy arrays serialized when communicating in schema
- python 3.8 requires pint 0.10
- add a logging file, presently `file.log` that currently is continuously appended
- any fns mirroring `sherrill_gold_standard` or `allen_focal_point` need to be reformatted and registered with `register_composite_function`
- extrapolation functions for composite need to be registered with `register_xtpl_function`
- CBS and xtpl functions need to be called as strings, not objects. So `energy('cbs', scf_wfn='scf', scf_basis='cc-pV[DTQ]Z', scf_scheme='scf_xtpl_helgaker_3')` not `energy(cbs, scf_wfn='scf', scf_basis='cc-pV[DTQ]Z', scf_scheme=scf_xtpl_helgaker_3)`.
- no longer always evaluating gradient before Hessian to test safety of projecting rotations. Instead assuming unsafe. Can pass `ref_gradient` array to test on or `set findif fd_project T/F` explcitly to control.
- `psi4.QMMM()` object replaced by `embedding_charges` kwarg in Bohr. See extern examples for details.
- `json_ret = psi4.schema_wrapper.run_qcschema(json_input)` not `json_ret = psi4.json_wrapper.run_json(json_input)`. `json_ret` is now an object, so `.` access, not dictionary.
- CBS wrapper issues a `clean()` btwn calcs. Plain string modelchem calcs like `energy('hf/cc-pvdz')` were getting caught and also being cleaned, meaning their behavior was slightly different than `set basis cc-pvdz \n energy('hf')`. This is no longer happening, so for occasional string modelchem calcs, you may need to add a clean in the input.
- extern object units are bohr, not whatever the Molecule was.
- `psi4 --module` returns the path for loading via `import psi4`. opposite of `psi4.executable`
- `psi4.set_output_file` now wraps `psi4.core.set_output_file` and additionally sets up a truncating python logging file with `.log` extension at the same time as the `.out` file.
- Besides the input.dat/output.dat exception, default output naming simplified to always `.out` of input file name. If input has out or log extension (weird), extra extension added.
- Output file in continuous mode now looks much as it normally does in the master branch, except there's a few more headers printed and the individual cbs/findif/nbody `AtomicResult` outputs aren't added in real time. Running through QCFractal, the output file is expected to break down, but this hasn't been tried.
- A Logging file has been introduced. All its contents should be considered experimental and subject to change.

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


dgasmith:
This is pretty early days on this PR. The short is that a `Computer` can be `CBS/FindIf/N-Body/Single/etc` (?) so that you can nest them and where a `Single` is the only object that calls quantum chemistry. For example, a `N-Body` can be made up of `Single` computations or `CBS` computations where a CBS computation will be made up for two or more `Single` computations. In this case, calling `compute` on `N-Body` will call `compute` on all of its children which if is a `Single` will call a quantum chemistry computation or if `CBS` will call 2 or more single computations (or FindIf!) and then construct that data into an extrapolated energy which will be given to `N-Body`.

The key here is that the `Single` can have three modes 1) a continuous computation on a single node 2) a sow/reap mode using files and 3) a sow/reap mode using [QCFractal](https://qcfractal.readthedocs.io/en/latest/). There will assuredly be many more docs.

psi4/driver/driver_nbody.py
```diff
@@ -778,15 +779,134 @@ def assemble_nbody_components(metadata, component_results):
...
+    def compute(self):
+        for k, v in self.task_list.items():
+            self.results_list[k] = v.compute()
+            print(self.results_list[k]["return_result"])
+
+    def get_results(self):
+        energies = {k: v['properties']["return_energy"] for k, v in self.results_list.items()}
+        ptype = None
+        if self.driver == 'gradient':
+            ptype = {k: np.array(v["return_result"]).reshape((len(v["return_result"])//3, 3))
```
loriab: why not just do `reshape((-1, 3))` here? error checking?

psi4/driver/driver_nbody.py
```diff
@@ -900,15 +899,66 @@ def get_results(self, results={}):
...
+            'molecule': self.molecule.to_schema(dtype=1)['molecule'],
+            'properties': {
+                'calcinfo_natom': self.molecule.natom(),
+                'nuclear_repulsion_energy': self.molecule.nuclear_repulsion_energy(),
+                'return_energy': ret_energy,
+            },
+            'provenance': p4util.provenance_stamp(__name__),
+            'psi4:qcvars': qcvars,
+            #'raw_output': None,
+            'return_result': ret_energy if self.driver == 'energy' else ret_ptype.tolist(),
```
loriab: @alenaizan All this casting between list, ndarray, and Matrix can get excessive. As long as a thing doesn't need to get written to json, I'm defaulting (for cbs) to leaving it as ndarray (non-flat). Then there's a `unnp` fn (in task_base.py) for now to take a nested dictionary and convert any ndarray in it to flat lists just before declaring the dict ready for json. And there's a `plump` fn (still in driver_cbs.py) that can re-inflate lists to ndarray with sensible shape.

alenaizan: I fixed this and moved ```plump``` to task_base.py

jeffschriber:
I have some small changes to make the multilevel driver work with QCFractal. Could I get push permission, or should I open a PR to the branch?

dgasmith:
Invite in your mailbox.

loriab:
hooray, lgtm PR analysis is working! this may be a landslide.

loriab:
right now nbody->cbs->singleresult is the only multilevel path implemented. it has the routing split btwn `SingleResult` and `CBSComputer` in the main task_planner.py https://github.com/dgasmith/psi4/blob/recursive/psi4/driver/task_planner.py#L149-L154 . nbody also has a separate processing step in its computer so that it first inits, takes a break for some logic, then spins up the next level (single or cbs) in `build_tasks`, which is a dummy fn in all other computers.

I've got working locally the findif->cbs->singleresult multilevel path. this doesn't require intermediate logic, so the spinning up of the next step (cbs or singleresult) could be done internally to the FinDifComputer based on whether `method == 'cbs'` or not. This requires FinDifComputer to be aware that computers other than SingleResult exist.

Organizationally, is it better to (a) keep the logic of CBSComputer vs SingleResult in task_planner and pass it to the Findifcomputer (possibly at the need to add build_tasks to FinDifComputer) or to (b) tuck it away within FinDifComputer? I'm thinking (a).

Secondly, with using actual pydantic objects instead of dicts and turning on no-extras, Computers are objecting to formerly passed-through kwargs (e.g., `irrep` through cbs). At a first pass, I've been popping them in the Computer and explicitly passing them on where relevant. I'm thinking this is the wrong approach (again, why should cbs admit existence of findif). So my next approach would be to split up kwargs in task_planner and pass them on as relevant only within that file. (Leftover kwargs still passed into the computers -- pydantic will probably object, but the user meant something by them.)

thoughts on any of this? @dgasmith

loriab:
right now nbody->cbs->singleresult is the only multilevel path implemented. it has the routing split btwn `SingleResult` and `CBSComputer` in the main task_planner.py https://github.com/dgasmith/psi4/blob/recursive/psi4/driver/task_planner.py#L149-L154 . nbody also has a separate processing step in its computer so that it first inits, takes a break for some logic, then spins up the next level (single or cbs) in `build_tasks`, which is a dummy fn in all other computers.

I've got working locally the findif->cbs->singleresult multilevel path. this doesn't require intermediate logic, so the spinning up of the next step (cbs or singleresult) could be done internally to the FinDifComputer based on whether `method == 'cbs'` or not. This requires FinDifComputer to be aware that computers other than SingleResult exist.

Organizationally, is it better to (a) keep the logic of CBSComputer vs SingleResult in task_planner and pass it to the Findifcomputer (possibly at the need to add build_tasks to FinDifComputer) or to (b) tuck it away within FinDifComputer? I'm thinking (a).

Secondly, with using actual pydantic objects instead of dicts and turning on no-extras, Computers are objecting to formerly passed-through kwargs (e.g., `irrep` through cbs). At a first pass, I've been popping them in the Computer and explicitly passing them on where relevant. I'm thinking this is the wrong approach (again, why should cbs admit existence of findif). So my next approach would be to split up kwargs in task_planner and pass them on as relevant only within that file. (Leftover kwargs still passed into the computers -- pydantic will probably object, but the user meant something by them.)

thoughts on any of this? @dgasmith

dgasmith:
Definitely a) I think. One requirement that we discussed early on was the ability to do something complex mixing CBS and levels of theory per N-body computation (for example). Having the logic centrally would make this easier.

The pydantic solution seems good as well. The `no-extras` is quite useful in complex system to help keep it understandable. Working vs future tweak able is always hard.

loriab:
tks. I've got (a) mostly implemented. I've backed off on forbiding extras for now -- will try that again when everything's working, so I know I'm not breaking things.

loriab:
* I started the renaming, so you'll need master qcel and possibly this qcng https://github.com/MolSSI/QCEngine/pull/172
* pretty much at same tests failing as before rebase. charge embedding for nbody needs refixing
* logging is in place -- always to `file.log` that grows.

loriab:
I believe we have clean DDD test suite.

loriab:
This branch could still use field testing and polishing. However, its main blocks of (1) all tests and Win CI not working and (2) scattered or swallowed output are now fixed. Ready for review.

psi4/__init__.py
```diff
@@ -104,3 +97,7 @@
     sys.path.insert(1, "@cppe_PYMOD@")
 if "@ENABLE_libefp@".upper() in ["1", "ON", "YES", "TRUE", "Y"]:
     sys.path.insert(1, "@pylibefp_PYMOD@")
+
+# Create a custom logger
+import logging
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
loriab: The logger is very much in its infancy. It's contents are a free-for-all. :-)

psi4/driver/driver.py
```diff
@@ -501,38 +439,56 @@ def energy(name, **kwargs):
...
+    # Are we planning?
+    plan = task_planner.task_planner("energy", lowername, molecule, **kwargs)
+    logger.debug('ENERGY PLAN')
+    logger.debug(pp.pformat(plan.dict()))
+
+    if kwargs.get("return_plan", False):
+        return plan
+
+    elif isinstance(plan, AtomicComputer):
+        # We have unpacked to a AtomicResult
```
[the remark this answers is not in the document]
loriab: True, I remember staring at this logic longer than I should have needed to.

psi4/driver/driver.py
```diff
@@ -501,38 +439,56 @@ def energy(name, **kwargs):
...
+    # Make sure the molecule the user provided is the active one
+    molecule = kwargs.pop('molecule', core.get_active_molecule())
+    molecule.update_geometry()
 
-    # Allow specification of methods to arbitrary order
-    lowername = name.lower()
+    ## Pre-planning interventions
+
+    # * Trip on function or alias as name
+    lowername = driver_util.upgrade_interventions(name)
```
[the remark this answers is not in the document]
loriab: May as well. I think they're being tested, so should keep working.

psi4/driver/driver_util.py
```diff
@@ -58,67 +45,68 @@ def _set_convergence_criterion(ptype, method_name, scf_Ec, pscf_Ec, scf_Dc, pscf
...
+    try:
+        lowermethod = method.lower()
+    except AttributeError as e:
+        if method.__name__ == 'cbs':
+            raise UpgradeHelper(method, repr(method.__name__), 1.4,
+                                ' Replace cbs or complete_basis_set function with cbs string.')
+        elif method.__name__ in ['sherrill_gold_standard', 'allen_focal_point']:
+            raise UpgradeHelper(
+                'argument ' + method.__name__, 'function ' + method.__name__, 1.4,
+                ' Replace `energy(sherrill_gold_standard)` with `energy("sherrill_gold_standard")` or similar.')
```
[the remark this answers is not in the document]
loriab: keeping the `or similar` in case it's gradient, but I've fixed up the methodname otherwise.

psi4/driver/driver_cbs_helper.py
```diff
@@ -0,0 +1,649 @@
...
+    """
+
+    if type(valueLO) != type(valueHI):
+        raise ValidationError(
+            f"scf_xtpl_karton_2: Inputs must be of the same datatype! ({type(valueLO)}, {type(valueHI)})")
+
+    if alpha is None:
+        alpha = 6.30
+
+    beta_division = 1 / (math.exp(-1 * alpha) * (math.exp(math.sqrt(zHI)) - math.exp(math.sqrt(zLO))))
```
PeterKraus, line 346: This should be `1 / (math.exp(-1 * alpha * math.sqrt(zHI)) - math.exp(-1 * alpha * math.sqrt(zLO)))` as exponents don't factorize out like that. I'm not sure how long this bug has been in `cbs()`, whoops!

psi4/driver/task_base.py
```diff
@@ -0,0 +1,197 @@
...
+    def compute(self, client=None):
+        from psi4.driver import pp
+
+        if self.computed:
+            return
+
+        if client:
+
+            self.computed = True
+            from qcfractal.interface.models import KeywordSet
```
dgasmith, line 118: This should be qcportal not fractal- greatly lowers dependancies.

dgasmith (commented):
Been awhile, but looks solid 🚀

psi4/driver/task_base.py
loriab: even snowflake needs only portal?

psi4/driver/aliases.py
```diff
@@ -140,40 +141,39 @@ def allen_focal_point(func, label, **kwargs):
...
+        'basis': kwargs.pop('scf_basis', 'cc-pV[Q56]Z'),
+        'scheme': kwargs.pop('scf_scheme', 'scf_xtpl_helgaker_3'),
+    }
+    corl = {  # MP2 - HF
+        'wfn': kwargs.pop('corl_wfn', 'mp2'),
+        'basis': kwargs.pop('corl_basis', 'cc-pV[56]Z'),
+        'scheme': kwargs.pop('corl_scheme', 'corl_xtpl_helgaker_2'),
+    }
+    delta = {  # CCSD - MP2
+        'wfn': kwargs.pop('delta_wfn', 'mrccsd'),
```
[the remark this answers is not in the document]
loriab: The existing error is something like "can't find dmrcc", but I've added an interception.

JonathonMisiewicz:
See #2575.

PR #1355, opened 2018-11-11
Ghost atom mass and Z: Round 2

bgpeyton:
## Description
New PR for [PR#990](https://github.com/psi4/psi4/pull/990) to avoid a mess of merge conflicts. Thanks to @robertodr for the help.

Provides a remedy for [#909](https://github.com/psi4/psi4/issues/909). `Molecule::mass` and `Molecule::Z` show inconsistent behavior for ghosts. This PR requires a new argument `bool zero_ghost` on the C++ side and optionally in the Python layer which gives a choice for behavior.

## Todos
* **Developer Interest**
  - [x] Required C++ argument `bool zero_ghost` for `Molecule::mass`
  - [x] Optional python argument `bool zero_ghost` for `Molecule::mass` (default false)
  - [x] Required argument `bool zero_ghost` for `Molecule::Z`
  - [x] Optional python argument `bool zero_ghost` for `Molecule::Z` (default true)
  - [x] Optional C++ argument `bool zero_ghost` for `Molecule::center_of_mass`
  - [x] Optional C++ argument `bool zero_ghost` for `Molecule::inertia_tensor`
  - [x] Check all C++ calls for "intended behavior"
  - [x] New optional python arguments for `Molecule.mass()` and `Molecule.Z()`

## Questions
- For clarity: "previous behavior" is that `mass` returns non-zero for ghost atoms (equivalent to `zero_ghost = false`), and `Z` returns zero for ghost atoms (`zero_ghost = true`).
- `Molecule::mass` and `Molecule::Z` on the C++ side now need to be passed either `True` or `False`, for whether or not to return zero for ghost atoms. I'm defaulting to previous behavior in most cases, though there are some (such as `optrot.cc`) that likely need to be changed. See the discussion in [#909](https://github.com/psi4/psi4/issues/909).
- May also need a logic pass of `center_of_mass` and `inertia_tensor` in a separate PR?
- Suggestions for changes C++ side or py-side are greatly encouraged. 

## Checklist
- [x] Add argument for `mass`
- [x] Add argument for `Z`
- [x] Tests
- [x] Docs

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -175,7 +175,7 @@ std::shared_ptr<BasisSet> construct_basisset_from_pydict(const std::shared_ptr<M
     // Modify the nuclear charges, to account for the ECP.
     if (totalncore) {
         for (int atom = 0; atom < mol->natom(); ++atom) {
-            if (mol->Z(atom) > 0) {
+            if (mol->Z(atom, true) > 0) {
                 const std::string& basis = mol->basis_on_atom(atom);
                 const std::string& label = mol->label(atom);
                 int ncore = basis_atom_ncore[basis][label];
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1231,7 +1231,8 @@ void export_mints(py::module& m) {
         .def("save_string_xyz_file", &Molecule::save_string_xyz_file, "Saves an XYZ file to arg2")
         .def("save_string_xyz", &Molecule::save_string_xyz, "Saves the string of an XYZ file to arg2")
         .def("Z", &Molecule::Z, py::return_value_policy::copy,
-             "Nuclear charge of atom arg0 (0-indexed without dummies)")
+             "Nuclear charge of atom arg0 (0-indexed without dummies), by default ghost atoms return zero nuclear charge",
+             py::arg("atom"), py::arg("zero_ghost") = true)
         .def("mass_number", &Molecule::mass_number, py::return_value_policy::copy, "Mass number (A) of atom if known, else -1")
         .def("x", &Molecule::x, "x position [Bohr] of atom arg0 (0-indexed without dummies)")
         .def("y", &Molecule::y, "y position [Bohr] of atom arg0 (0-indexed without dummies)")
@@ -1250,7 +1251,7 @@ void export_mints(py::module& m) {
              "Computes center of mass of molecule (does not translate molecule)")
         .def("translate", &Molecule::translate, "Translates molecule by arg0")
         .def("move_to_com", &Molecule::move_to_com, "Moves molecule to center of mass")
-        .def("mass", &Molecule::mass, "Returns mass of *atom* (0-indexed), by default ghost atoms return zero mass",
+        .def("mass", &Molecule::mass, "Returns mass of *atom* (0-indexed), by default ghost atoms return non-zero mass",
              py::arg("atom"), py::arg("zero_ghost") = false)
         .def("set_mass", &Molecule::set_mass,
              "Sets mass of *atom* (0-indexed) to *mass* (good for isotopic substitutions)",
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1250,7 +1250,8 @@ void export_mints(py::module& m) {
              "Computes center of mass of molecule (does not translate molecule)")
         .def("translate", &Molecule::translate, "Translates molecule by arg0")
         .def("move_to_com", &Molecule::move_to_com, "Moves molecule to center of mass")
-        .def("mass", &Molecule::mass, "Returns mass of *atom* (0-indexed)", py::arg("atom"))
+        .def("mass", &Molecule::mass, "Returns mass of *atom* (0-indexed), by default ghost atoms return zero mass",
+             py::arg("atom"), py::arg("zero_ghost") = false)
         .def("set_mass", &Molecule::set_mass,
              "Sets mass of *atom* (0-indexed) to *mass* (good for isotopic substitutions)",
              py::arg("atom"), py::arg("mass"))
--- a/psi4/src/psi4/cc/ccresponse/optrot.cc
+++ b/psi4/src/psi4/cc/ccresponse/optrot.cc
@@ -406,7 +406,7 @@ void optrot(std::shared_ptr<Molecule> molecule) {
             }
 
         /* compute the specific rotation */
-        for (j = 0, M = 0.0; j < moinfo.natom; j++) M += molecule->mass(j); /* amu */
+        for (j = 0, M = 0.0; j < moinfo.natom; j++) M += molecule->mass(j,true); /* amu */
         nu = params.omega[i];                                               /* hartree */
         bohr2a4 = pc_bohr2angstroms * pc_bohr2angstroms * pc_bohr2angstroms * pc_bohr2angstroms;
         m2a = pc_bohr2angstroms * 1.0e-10;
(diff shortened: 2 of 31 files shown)


psi4/src/export_mints.cc
```diff
@@ -1231,7 +1231,8 @@ void export_mints(py::module& m) {
         .def("save_string_xyz_file", &Molecule::save_string_xyz_file, "Saves an XYZ file to arg2")
         .def("save_string_xyz", &Molecule::save_string_xyz, "Saves the string of an XYZ file to arg2")
         .def("Z", &Molecule::Z, py::return_value_policy::copy,
-             "Nuclear charge of atom arg0 (0-indexed without dummies)")
+             "Nuclear charge of atom arg0 (0-indexed without dummies), by default ghost atoms return zero nuclear charge",
+             py::arg("atom"), py::arg("zero_ghost") = true)
```
bgpeyton: I see that `py::arg("atom")` has been removed upstream; is this intentional? If so I can also remove it. @loriab

psi4/src/psi4/cc/ccresponse/optrot.cc
```diff
@@ -406,7 +406,7 @@ void optrot(std::shared_ptr<Molecule> molecule) {
             }
 
         /* compute the specific rotation */
-        for (j = 0, M = 0.0; j < moinfo.natom; j++) M += molecule->mass(j); /* amu */
+        for (j = 0, M = 0.0; j < moinfo.natom; j++) M += molecule->mass(j,true); /* amu */
```
bgpeyton: Definitely needs to be `true` for mass-weighting the specific rotation.

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -1064,9 +1066,9 @@ void Molecule::print() const {
 
         for (int i = 0; i < natom(); ++i) {
             Vector3 geom = atoms_[i]->compute();
-            outfile->Printf("      %3s%-7s ", Z(i) ? "" : "Gh(", (symbol(i) + (Z(i) ? "" : ")")).c_str());
+            outfile->Printf("      %3s%-7s ", Z(i, true) ? "" : "Gh(", (symbol(i) + (Z(i, true) ? "" : ")")).c_str());
             for (int j = 0; j < 3; j++) outfile->Printf("  %17.12f", geom[j]);
-            outfile->Printf("  %17.12f", mass(i));
+            outfile->Printf("  %17.12f", mass(i, true));
```
bgpeyton: Printed masses of ghost atoms should probably be 0, but it is of little consequence.

psi4/src/psi4/cc/ccresponse/scatter.cc
```diff
@@ -291,14 +291,14 @@ void scatter(std::shared_ptr<Molecule> molecule, Options &options, double step,
     // Mass-weighting the co-ordinates //
     // double massi[natom];
     outfile->Printf("\tAtomic Masses for Raman Computation:\n");
-    for (i = 0; i < natom; i++) outfile->Printf("\t%d %12.8f\n", i, molecule->mass(i));
+    for (i = 0; i < natom; i++) outfile->Printf("\t%d %12.8f\n", i, molecule->mass(i, false));
```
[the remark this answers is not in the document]
amjames: From my understanding a ghost atom is just a point in space where we are putting basis functions as if an atom were there, but no atom is really there. 

Assuming that is correct mass weighting and computing the center of mass should use zero mass for ghosts.

psi4/src/psi4/cc/ccresponse/scatter.cc
```diff
@@ -408,7 +408,7 @@ void scatter(std::shared_ptr<Molecule> molecule, Options &options, double step,
     for (i = 0; i < natom; i++) {
         for (j = 0; j < 3; j++) {
             // M->set((i*3+j),(i*3+j),1/sqrt(an2mass[(int)zvals[i]]/_au2amu));
-            M->set((i * 3 + j), (i * 3 + j), 1 / sqrt((molecule->mass(i)) / pc_au2amu));
+            M->set((i * 3 + j), (i * 3 + j), 1 / sqrt((molecule->mass(i, false)) / pc_au2amu));
```
amjames: We will have to be careful, if atom `i` here is a ghost and the mass is zero this would cause a divide by zero error. I don't think that coordinates for ghost atoms should be included in the cartesian hessian? We might need @lothian to chime in here.

psi4/src/psi4/cc/ccresponse/scatter.cc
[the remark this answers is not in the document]
bgpeyton: That's exactly right. You might want a non-zero mass if you want consistent frames between real and ghosted fragments (RE:[909 discussion](https://github.com/psi4/psi4/issues/909)). I can see people wanting an un-changing `center_of_mass` during, say, a many body expansion just for `reorient`-like purposes, but for actual property calculations I don't _think_ the ghost masses should contribute to `center_of_mass` or otherwise.

psi4/src/psi4/cc/ccresponse/scatter.cc
bgpeyton: I believe we're in agreement with @lothian that the cartesian hessian should not include ghost atoms. However, a quick test of `psi4.hessian` on Helium with and without adding a Gh(He) reveals that it does indeed include the ghost atom (goes from size 3x3 to 6x6); so, skipping ghosts here would create a mass-weighting matrix of the "wrong" size (really a hessian with the wrong size) which should throw an error below at the `gemm` when  mass-weighting.

The question is- should we fix the hessian and `scatter.cc` here and now, should we only fix `scatter.cc` and leave the error at the `gemm` and fix the hessian in a separate PR, or let sleeping ghosts lie? I'm not particularly fond of the latter.

lothian: The Hessian should not include ghost atoms at all, in my opinion.

loriab:
Since most calls are c-side (w/o) arg names, I still find it counterintuitive that `mB->Z(B, true)` _excludes_ ghosts, rather than True<-->Inclusive. `tangible_ghosts`, `undead_ghosts` to switch the logic, anyone?

But that's an old complaint. More to the point, it'd be good to use `PSI_DEPRECATED` and `warnings.warn` to require the ghost arg in v1.4 both py-side and c-side. This PR is right that uses w/i our control should be switched out in v1.3.

bgpeyton:
@loriab I'm happy to switch the logic, and I would love to name it `undead_ghosts`, but the name should probably be strikingly clear at a glance since it's a required arg c-side. `include_ghosts` or `count_ghosts` maybe? 

As for uses w/i our control, I'm not convinced we've switched everything that needs it, but the ones I _have_ switched from default I am confident in (current work on `scatter.cc` and the Hessian notwithstanding). If there's others we could ping for the other cases I'm all for it.

loriab:
@bgpeyton, I think the following points for the deprecation hold

* c-side in molecule.h have ```PSI_DEPRECATED("Using `Molecule.Z(atom)` instead of `Molecule.Z(atom, include_ghosts)` (where `include_ghosts=True` has been the default behavior) is deprecated, and in 1.4 it will stop working") fn_name(atom_arg)```
* c-side in molecule.h have `fn_name(atom_arg, ghosts_arg)`
* c-side in molecule.cc have `fn_name(atom_arg, ghosts_arg) {...}`
* c-side in molecule.cc have `fn_name(atom_arg) { fn_name(atom_arg, True) }`

I don't vouch for my c++ syntax, but there's some models in libmints/wavefunction.

PR #1358, opened 2018-11-15
Light CMake clean-up

robertodr:
## Description
Cleans up CMake so that external dependencies can be linked to libraries in `psi4/src/psi4` and then inherited by `core` without invoking `target_link_libraries`/`target_include_directories` again.

## Todos
- [x] Rewrote `psi4_add_module` CMake macro. Now it takes `bin` or `lib` argument, name of module, and list of sources. Note that the macro:
     - Now defines **two** global lists of libraries `core` will link to, one for `bin`s and one for `lib`s. This is needed because `lib`s need to appear last in the link line to avoid missing symbols.
     - Sets source file properties to generate position-independent code with symbols hidden by default.
- [x] Rewrote downstream `CMakeLists.txt` so that: 
     - Sources are in a list, with one source file per line, for easier diff inspection.
     - External dependencies are linked where they are used first.
- [x] Reordered `include`s in _some_ files according to the hierarchy: 
     - For class implementation sources: header corresponding to current source file first (_.i.e._ `#include "myself.h"` in `myself.cc`)
     - C++ standard headers, including `<omp.h>`
     - Psi4 higher-level headers, _i.e._ `psi4/psifiles.h` &co.
     - Psi4 headers in subdirectories _i.e._ `psi4/libmints/matrix.h` &co.
     - Headers in current directory _without full path_.
- [x]  `psi4/src/psi4/liboptions/liboptions_python.h` moved to `psi4/src/python_data_type.h` and 
`psi4/src/psi4/liboptions/python.cc` moved to `psi4/src/python_data_type.cc`. These are now compiled and linked as part of `core`, rather than `liboptions`.

## Questions
I note that `mrcc.cc` and `civect.h` include `pybind11.h`. It might be a good idea to move those bindings one level up, directly into `core` without spreading the dependency too deep in the code. Opinions?

## Checklist

## Status
- [x] Ready for review


external/upstream/pybind11/CMakeLists.txt
```diff
@@ -16,7 +16,6 @@ else()
         UPDATE_COMMAND ""
         CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-                   -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}  # unused, but needs working compiler
```
loriab: This one came about when there's a C compiler available (say, `/usr/bin/gcc` 4.6) but not up to compiling anything serious. Errors arise when cmake tries to compile a simple test prog, even though a better C compiler has been passed in through `CMAKE_C_COMPILER`.

cmake/psi4OptionsTools.cmake
```diff
@@ -197,21 +197,6 @@ macro(option_with_flags option msg default)
...
-macro(optional_plugin plugin_name)
-string(TOUPPER ${plugin_name} PLUGIN_NAME)
-if(${ENABLE_${PLUGIN_NAME}})
-   find_package(${plugin_name} REQUIRED)
-   set_property(GLOBAL APPEND PROPERTY PSI4_MODULES ${${PLUGIN_NAME}_LIBRARIES})
-   add_definitions(-DENABLE_${PLUGIN_NAME})
-else()
-   add_library(${plugin_name} INTERFACE)
-endif()
-endmacro(optional_plugin plugin_name test_names)
```
loriab: check with @amjames just to make sure this wasn't the pasture plugin treatment. Not that I think any code is coming back from the pasture.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -26,31 +26,33 @@
...
 #include "psi4/libmints/molecule.h"
 #include "psi4/libmints/potential.h"
-#include "psi4/libmints/integral.h"
+#include "psi4/libmints/vector.h"
 #include "psi4/liboptions/liboptions.h"
-#include "psi4/libcubeprop/csg.h"
 #include "psi4/libpsi4util/process.h"
-#include "psi4/lib3index/dfhelper.h"
-
-#include <ctime>
```
jturney: I think this header is needed to fix the appveyor compilation.

external/upstream/pybind11/CMakeLists.txt
robertodr: I removed it because of the annoying "Passed variable not used by project" warning. The culprit for the need of a C compiler seems to be the check for `PThreads` in pybind11 own `CMakeLists.txt`: https://github.com/pybind/pybind11/blob/master/CMakeLists.txt#L15-L20
According to the comment there, problem is solved with CMake higher than 3.4.0 That would explain why I never saw the problem myself. Mind if I bump the CMake requirement to 3.8?

psi4/src/psi4/fisapt/fisapt.cc
robertodr: :man_facepalming:

external/upstream/pybind11/CMakeLists.txt
loriab: I think we're safe, policy-wise https://cmake.org/cmake/help/v3.8/manual/cmake-policies.7.html#policies-introduced-by-cmake-3-4

cmake/psi4OptionsTools.cmake
amjames: Nope, I set up the pasture plugins to build as if they were standard plugins. They just use the `add_psi4_plugin` macro. They are built as downstream and loaded in the driver by specific request, if they have been installed into the psi4 install. 

All that said I just took look a quick look at the pasture CMake, I have no idea if those plugins are even still working. We have been relying on people to report a problem if one comes up, so either they are working fine or no one is using them. There is no build testing done on that repo as far as I know, and they are not included in conda pkgs or anything.

psi4/src/psi4/libmints/CMakeLists.txt
```diff
@@ -1,89 +1,114 @@
...
+# Always linked-to external projects
+target_link_libraries(mints
+  PUBLIC
+    Libint::libint
+  )
+
+# Conditional sources
 if(TARGET simint::simint)
-    list(APPEND sources_list siminteri.cc)
-    add_definitions("-DUSING_simint")
```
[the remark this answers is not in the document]
robertodr: 1) No, I didn't try that.
2) If you mean with all addons enabled, then I haven't tried that either.
3) Yes, it got lost. I assumed it was in the library interface.
I'll try all of the above tomorrow first thing in the morning.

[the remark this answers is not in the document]
robertodr: 1. `-j16` worked like a charm.
2. Building addons worked.

PR #1379, opened 2018-11-27
CoordValue/CoordEntry Cleanup and 1325 Fix

### LINKED ISSUE #1325: Backwards Incompatibility: Coordinates Have Been Set in the Wrong Order
Discovered on a build of 8db12d6 while investigating [a forum post](http://forum.psicode.org/t/ri-mp2-def2-tzvppd-convergence-issue-for-calculations-with-iodide/1106/5). The user's input file crashes during the MP2 computation with traceback:

```
  File "/opt/vulcan/opt/vulcan/linux-x86_64/intel-16.0.1/psi4-master-avducsgnonqrcpftiesq432gagal6y7r/bin/psi4", line 272, in <module>
    exec(content)
  File "<string>", line 100, in <module>
  File "/opt/vulcan/opt/vulcan/linux-x86_64/intel-16.0.1/psi4-master-avducsgnonqrcpftiesq432gagal6y7r/lib/psi4/driver/driver.py", line 431, in energy
    return driver_nbody.nbody_gufunc(energy, name, ptype='energy', **kwargs)
  File "/opt/vulcan/opt/vulcan/linux-x86_64/intel-16.0.1/psi4-master-avducsgnonqrcpftiesq432gagal6y7r/lib/psi4/driver/driver_nbody.py", line 225, in nbody_gufunc
    component_results = compute_nbody_components(func, method_string, metadata)
  File "/opt/vulcan/opt/vulcan/linux-x86_64/intel-16.0.1/psi4-master-avducsgnonqrcpftiesq432gagal6y7r/lib/psi4/driver/driver_nbody.py", line 411, in compute_nbody_components
    current_mol = molecule.extract_subsets(list(pair[0]), ghost)

RuntimeError: 
Fatal Error: Coordinates have been set in the wrong order
Error occurred in file: /home/vulcan/vadmin/programs/psi4/psi4/src/psi4/libmints/coordentry.cc on line: 239 

psi::Molecule::operator=(psi::Molecule const&)
psi::Molecule::Molecule(psi::Molecule const&)
```

Given that the user's problem is failure for the SCF to converge, I can only assume some change since the 1.2.1 release is at fault. I've been able to get the crash with the following simplified input file:
```
molecule dimer {
1 1
 C1      0.000000    0.513942    0.000000
 N2      0.628754    1.703580    0.000000
 H3      0.109407    2.571793    0.000000
 H4      1.638743    1.762759    0.000000
 N5     -1.342220    0.473050    0.000000
 H6     -1.890262    1.323784    0.000000
 H7     -1.839824   -0.407373    0.000000
 N8      0.700723   -0.623756    0.000000
 H9      1.712010   -0.555355    0.000000
 C10     0.085273   -1.957243    0.000000
 H11     0.892679   -2.697454    0.000000
 H12    -0.522596   -2.104233    0.905528
 H13    -0.522596   -2.104233   -0.905528
--
-1 1
 I   N8   3.6   H9   90.0   C1   90.0
}

basis {

   assign H   def2-tzvpd
   assign C   def2-tzvpd
   assign N   def2-tzvpd

   assign I   def2-tzvpd
}

set guess sad
set scf_type df
set maxiter 500
set soscf true
set soscf_max_iter 15
set soscf_conv 1.e-4
set freeze_core false   ## false for ions

energy("mp2", bsse_type = "cp")
```
This can probably be simplified further. It would be worth investigating whether this is a consequence of mixing Cartesian and ZMAT coordinates.

JonathonMisiewicz:
## Description
My debugging efforts indicate the problem with #1325 is that when attempting to clone a molecule with an atom with ZMAT and Cartesian coordinates, when that ZMatrixEntry atom is cloned, its `rto_`, `ato_`, `dto_` CoordEntries are Cartesian. This somehow leads to the parent ZMatrixEntry having `computed_` true, but not for `rto_` and friends. So, we somehow have a bug in molecule creation.

To avoid creating new bugs, I want to understand exactly what is going on before trying to implement a fix. Unfortunately, this part of the code is not in great shape. I'll hopefully include a fix for the bug in here, but in the meantime, I'm submitting the code cleanup I'm doing as I try to follow this part of the code. We do not want `libmints` changes to escape a thorough vetting process.

## Todos
- [x] Cleanup `CoordValue` and `CoordEntry`
- [x] Document `add_atom` and `add_unsettled_atom`
- [x] Actually fix #1325 

## Checklist
- [ ] Is it worth adding a test for this bugfix? This should be a quick one to test, but it looks like we're seeing timeouts due to test overload again.
- [x] Tested this fixes #1325 
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests) obsessively - everything in ctest except the dft benchmarks and python-vibanalysis

## Status
- [x] Ready for review
- [ ] Ready for merge. (After the question of whether to add a test.)

### DIFF
--- a/psi4/src/psi4/libmints/molecule.cc
+++ b/psi4/src/psi4/libmints/molecule.cc
@@ -335,6 +335,10 @@ void Molecule::add_unsettled_atom(double Z, std::vector<std::string> anchor, std
     } else {
         throw PSIEXCEPTION("Illegal geometry specification (neither Cartesian nor Z-Matrix)");
     }
+
+    if ((label != "X") && (label != "x")) {
+        atoms_.push_back(full_atoms_.back());
+    }
 }
 
 double Molecule::mass(int atom) const {
--- a/psi4/driver/qcdb/libmintscoordentry.py
+++ b/psi4/driver/qcdb/libmintscoordentry.py
@@ -52,10 +52,8 @@ class CoordValue(object):
     """
 
     def __init__(self, fixed=False, computed=False):
-        # Fixed coordinate?
+        # Fixed coordinate? For a fixed value, the reset method does nothing.
         self.PYfixed = fixed
-        # Whether the current value is up to date or not
-        self.computed = computed
 
     def set_fixed(self, fixed):
         """Set whether the coordinate value is fixed or not"""
@@ -64,10 +62,6 @@ class CoordValue(object):
     def fixed(self):
         """Get whether the coordinate value is fixed or not"""
         return self.PYfixed
-
-    def invalidate(self):
-        """Flag the current value as outdated"""
-        self.computed = False
 
     def everything(self):
         print('\nCoordValue\n  Fixed = %s\n  Computed = %s\n\n' % (self.PYfixed, self.computed))
@@ -103,8 +97,8 @@ class NumberValue(CoordValue):
         return "%*.*f" % (precision + 5, precision, self.compute())
 
     def everything(self):
-        print('\nNumberValue\n  Fixed = %s\n  Computed = %s\n  Type = %s\n  Value = %f\n  FValue = %s\n\n' %
-            (self.PYfixed, self.computed, self.type(), self.value, self.variable_to_string(4)))
+        print('\nNumberValue\n  Fixed = %s\n  Type = %s\n  Value = %f\n  FValue = %s\n\n' %
+            (self.PYfixed, self.type(), self.value, self.variable_to_string(4)))
 
 
 class VariableValue(CoordValue):
@@ -163,8 +157,8 @@ class VariableValue(CoordValue):
             return self.PYname
 
     def everything(self):
-        print('\nVariableValue\n  Fixed = %s\n  Computed = %s\n  Type = %s\n  Value = %f\n  FValue = %s\n  Name = %s\n  Negated = %s\n  Map = %s\n\n' %
-            (self.PYfixed, self.computed, self.type(), self.compute(), self.variable_to_string(4), self.name(), self.negated(), self.geometryVariables))
+        print('\nVariableValue\n  Fixed = %s\n  Type = %s\n  Value = %f\n  FValue = %s\n  Name = %s\n  Negated = %s\n  Map = %s\n\n' %
+            (self.PYfixed, self.type(), self.compute(), self.variable_to_string(4), self.name(), self.negated(), self.geometryVariables))
 
 
 class CoordEntry(object):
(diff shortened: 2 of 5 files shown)


psi4/src/psi4/libmints/molecule.cc
```diff
@@ -321,6 +321,10 @@ void Molecule::add_unsettled_atom(double Z, std::vector<std::string> anchor, std
     } else {
         throw PSIEXCEPTION("Illegal geometry specification (neither Cartesian nor Z-Matrix)");
     }
+
+    if ((label != "X") && (label != "x")) {
```
loriab: did you hit this directly? I really thought `label` should be properly cased by this point.

JonathonMisiewicz: I didn't check the label casing. I noticed that `atoms_` wasn't populated after an `add_unsettled_atoms` call and just copied https://github.com/psi4/psi4/blob/68af15a18c9abf8e1b803c38b1e8d2c89408eb23/psi4/src/psi4/libmints/molecule.cc#L243-L245 from `add_atoms`. I can check if proper casing is enforced by this point and edit the relevant lines from both `add_atoms` and `add_unsettled_atoms` if it is.

loriab: I think you're in the right

JonathonMisiewicz:
I can now finally explain why this bug is happening:
Assume you're cloning a ZMAT atom. [This line](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/coordentry.h#L372) I added in #1108 enforces that if ZMAT coordinates were computed for the parent, they be computed for the clone atom. When giving a ZMAT atom coordinates in space, the reference atoms need to already have coordinates in space computed. Psi checks this, naturally enough, with`CoordEntry.computed_` (via a getter). Because of that line in #1108, `computed_` will always be set correctly for ZMAT atoms, so if you only ever use ZMAT atoms, all is well. However, there is no mechanism in place to compute coordinates for Cartesians as well, so those are listed as not computed when Psi checks `computed_`. The fix seems clear enough to me: add a `compute` call in `CartesianEntry::clone` to mirror the one in `ZMatrixEntry::clone`. Because changes in `molecule` can have wide-reaching repercussions, I'll hold off on testing the fix until @loriab is done fixing the full test suite, so I can use it to safety check this PR.

And with that realization, I am quite hopeful to get this in Psi 1.3 after all.

JonathonMisiewicz:
For context, the last commit is the one that fixes the bug. It is one line of code. The rest is all just cleanup.

Again, should I add a test for this? It's difficult for me to think of what an appropriate test would be.

Issue #1375, opened 2018-11-26
PsiPEP 002


loriab:
# PsiPEP2: Rationalize Variable Access

Pursuant to #883, common driver migration, the War Against Globals, and [QCSchema](https://github.com/molssi/qcschema), we want clear programmatic access to QC quantities, and we want them on the Wfn/JSONrec, not off `Process.environment.globals`. First step toward this is duplication of `P::e.globals` onto `Wfn`. Unfortunately, the syntax and methods don't align (see table below). Also unfortunately, we have many redundant `get_`s. This PsiPEP proposes the following syntax changes and additions.

### Final Proposal (dynamic)

* **W-C** means C-side access to `psi4.core.Wavefunction`
* **W-Py** means Py-side access to `psi4.core.Wavefunction`
* **G-Py** means Py-side access to `psi4.core` globals

| Now `Wavefunction` | Now `psi4.core` | Proposed Both | W-C | W-Py | G-Py | doc |
|-----|----|----|---|---|---|---|
| | has_variable | has_scalar_variable |y|y|y| is float var set ? |
| get_variable | get_variable | scalar_variable |y|y|y| return float var value |
| set_variable | set_variable | set_scalar_variable |y|y|y | set float var value |
| | | del_scalar_variable |y|y|y | unset float var |
| variables | get_variables | scalar_variables |y|y|y| return dict of all float vars |
| |
| | | has_array_variable |y|y|y| is array var set ? |
| get_array | get_array_variable | array_variable |y|y|y| return array var value |
| set_array | set_array_variable | set_array_variable |y|y|y| set array var value |
| | | del_array_variable |y|y|y| unset array var |
| arrays | get_array_variables | array_variables |y|y|y| return dict of all array vars |
| |
| | | has_variable |–|y|y| is any float or array var set ?|
| | | variable |–|y|y| return float or array var value |
| | | set_variable |–|y|y| set float/Matrix/ndarray var value |
| | | del_variable |–|y|y| unset float or array var |
| | | variables |–|y|y| return dict of all float or array vars |
| | clean_variables | clean_variables |–|–|y| unset all float or array vars |

### Original Proposal (static)

| Now `Wavefunction` | Now `psi4.core` | Proposed Both | doc |
|-----|----|----|---|
| | has_variable | has_variable | is float var set ? |
| get_variable | get_variable | variable | return float var value |
| set_variable | set_variable | set_variable | set float var value |
| | | del_variable | unset float var |
| |
[...]

Issue #883, opened 2017-12-17
Localization of Psi4 variables to Wavefunction objects


dgasmith:
Psi4 modules writing their variable to globals should be moved to Wavefunction local variables. This is part of the general "localization" of Psi4 to assist in complex Psi4 runs involving multiple computations and to further promote the API. 

The general plan is as follows:
1) Add a bit of code that dumps all local Wavefunction variables to global variables at the end of each *driver* computation `energy/gradient/Hessian/etc`. The small code snippet looks like the following:

```python
scf_e, wfn = psi4.energy("HF", return_wfn=True)
for k, v in wfn.variables().items():
    core.set_variable(k, v)
```
Thats it! Just need to be called at the right places in the driver to ensure we do not miss a crucial export. This will keep the global variables working as expected.

2) Replace all `Process::environment.globals["name"] = variable` calls with `Wavefunction->set_variable("name", variable)` for modules.
3) Keep in mind some modules may grab variables from these global options. Some thought will be needed in these cases so this is not a blanket find and replace.
4) @loriab Are there going to be doc issues when we do this?

I would not do this all in one go due to point 3. So I would recommend a series of PR's:
 - PR 1: Do (1) and apply (2) to adc, dfep2, dfmp2, detci, and libsapt_solver modules (in that order, testing in between!).
 - PR 2 (may be multiple):  Apply (2) to major module blocks one at a time CC*, occ, dfocc, etc.
 - PR 3: Clean up any remaining module global exports.

loriab:
Note that I've done parts of this (but by no means all) in another branch, so ping me if you'd like to take this up.

dgasmith:
@loriab Will #1445 close this, or is this still an ongoing project that will last until 1.4?

loriab:
Ongoing, regrettably. 1445 accomplishes scf, mp2,mp2.5 across the modules.

dgasmith:
Moving to 1.4 in this case.

loriab:
### Storing data on `Wavefunction` instances

#### Choices (dipole gradient used as example)

- (a) store in dedicated slot, `wfn.dipole_gradient_`. access through `wfn.dipole_gradient()`
- (b) store in `wfn.arrays_`. access through `wfn.variable('CURRENT DIPOLE GRADIENT')` or `wfn.variable('MTD DIPOLE GRADIENT')`
- (c) do (b) but also accessible through `wfn.dipole_gradient()` which just performs `wfn.arrays_` lookup
[...]

PR #1381, opened 2018-11-28
qcvar handling on wfn

loriab:
## Description
A step toward #883 by improving qcvar capabilities on Wfn and localizing some modules.

## Todos
- [x] removes or py3-izes scripts
- [x] addresses the floats part of #1375 by
   - expanding the variable operations on `Wavefunction`
   - making their storage always uppercase and their access case-insensitive
   - dev alert: any code that wrote to wfn array directly in non-UC (e.g.,  `wfn.variables_["lower or mixed-case var"] = val`) needed to be changed to `wfn.set_variable`
- [x] adds runtime FutureWarning to renamed fns: `psi4.core.get_variable`, psi4.core.get_variables`, `Wavefunction.get_variable`
- [x] improves the wfn localization of qcvars for DFT, scf_helper, DFMP2, OEPROP. trying to set on `wfn` c-side, then just copy into `P::e.globals` in proc.py.
- [x] moves DETCI, DFMP2 entirely to qcvar passing on Wfn. OCC, DFOCC, FNOCC do so for mp2-like only.
- [x] adds some functional-including qcvars for DFT
- [x] BREAKS8
   - remove `SCS-MP2 OPPOSITE-SPIN CORRELATION ENERGY` and `SCS-MP2 SAME-SPIN CORRELATION ENERGY` from dfmp2 `Wavefunction.variables_`
  - remove MP2 psivars from double-hybrid DFT final return
  - `SCS-MP2 * ENERGY` now are defined (6/5, 1/3) quantities. For the user-influenceable vars, it's `CUSTOM SCS-MP2 * ENERGY`
- `core.variable` now throws when you access a key not set. This is what `Wavefunction.variable` did before, but `core.variable` used to just create the key and set to 0.

## Questions
- [ ] FutureWarning provides the offending line. However, this is complicated by our promoting the get_variable namespace and the PsiAPI and Psithon calling routes. It's going to be hard to get this offending line printing universally right. It's hit and miss now.

## Checklist
- [x] full tests

## Status
- [x] Ready for review
- [ ] Ready for merge


loriab:
Ok, this one's passing and ready to review. There'll be two more PRs atop this: one to mend all the internal deprecation warnings (passing of this PR proves that they're just warnings, not breaks) and one (the real intent) with MP2D.

psi4/driver/p4util/python_helpers.py
```diff
@@ -520,3 +522,27 @@ def py_psi_set_global_option_python(key, EXTERN):
 
 
 core.set_global_option_python = py_psi_set_global_option_python
+
+
+def core_get_variables():
+    warnings.warn("Using `psi4.core.get_variables` instead of `psi4.core.variables` is deprecated, and in 1.4 it will stop working\n", category=FutureWarning, stacklevel=2)
```
dgasmith: Shouldn't these be `DeprecationWarning`s? Might want to YAPF this file as well.

psi4/driver/procrouting/proc.py
```diff
@@ -3045,12 +3093,16 @@ def run_detci(name, **kwargs):
 
     ciwfn = core.detci(ref_wfn)
 
+    # Shove variables into global space
+    for k, v in ciwfn.variables().items():
+        core.set_variable(k, v)
```
dgasmith: Is it possible to throw locals back into globals one level up, this seems like it could add a lot of code overall and make procreating even more complex.

psi4/driver/procrouting/proc.py
```diff
@@ -1924,6 +1939,11 @@ def run_occ(name, **kwargs):
 
     occ_wfn = core.occ(ref_wfn)
 
+    # Shove variables into global space
+    if name in ['mp2', 'omp2', 'scs-omp2', 'scs(n)-omp2', 'scs-omp2-vdw', 'sos-omp2']:
```
dgasmith: I don't quite see why we need a `if` in these?

psi4/driver/p4util/python_helpers.py
loriab: I agree that everyone else is using `DeprecationWarning`, and I'm glad to change. But reading [this table and the note at the end](https://docs.python.org/3/library/warnings.html#warning-categories) suggested that `FutureWarning` was right (for targeting users' input files).

psi4/driver/procrouting/proc.py
loriab: This and your previous comment are connected, so I'll answer together.

Indeed, I blithely though the `if ` was unnecessary, too! It turns out that if replacing (`P::e.globals` --> `wfn_->set_variabl`) and shoving instead of duplicating every set_variable to globals with one to wfn, one must either do the whole module (which I did for DFMP2 and DETCI) or prescribed bits (*mp2 for DFOCC, OCC, FNOCC). Otherwise what happens is that the SCF vars set ofnthe reference wfn are what get shoved back into globals, thereby replacing the desired `P::e.globals` set during the correlated calc.

Interestingly, there was trouble with `ci-multi` that might pop up again for other modules. That's a test with two rasci jobs with the wfn of the former starting the latter. Now that we're starting to set `wfn.energy()` on post-scf, the second job was taking the rasci energy as the HF energy. I changed how detci gets its ref energy, but we'll have to watch out.

So in answer to your previous comment, absolutely that's how it should be done, but it's unsafe for now.

loriab:
Don't bother reviewing, as I'm breaking this up into morsels, first of which is #1393.

loriab:
succeeded by #1393, #1394, #1395, #1444, #1445

PR #1393, opened 2018-12-02
Regularize Wfn and P::e psivar access for scalars and arrays

loriab:
An implementation for #1375 PsiPEP 002

## Todos
- [x] Standardizes function names surrounding scalar and array variables in globals & Wfn. Fills in missing methods.
- [x] C-side deprecates old names for Wfn variable access (compile-time for C-API downstream)
- [x] Py-side deprecates old names for global and Wfn variable access (run-time for Py-API downstream, incl. Psithon inputs)
- [x] New set of functions (reuse of many of the old fn names) for uniform access
- [x] Wfn variable access now case insensitive and storage always uppercase (as globals always has)
- [x] array variable getters/setters now return _copies_ of the arrays
- [x] globals now throws instead of autovivifying to 0 when getter called for non-existant key (as Wfn always has)
- [x] Py-side (and hence all globals access) resists giving the same key to a scalar and an array variable. This check is not in place for C-side Wfn, but shouldn't be a problem given our habit of labeling with E/G/H, etc.
- **ALERT** this generates oodles of `FutureWarning`s b/c to keep this PR focussed on the substantive changes, I haven't updated the fn names w/i the codebase. That will be a separate PR hard after this gains approval, and should probably be merged in sequence, lest users go batty. I've worked on other topics beyond this one, and I'm confident the deprecation warnings are allowing the unhindered use of old syntax.
- [x] Deprecation Warnings look like the below in PsiAPI and PSIthon modes, respectively. For the latter, I had to suppress printing of offending line b/c it was very wrong (usually part of the LGPL license).
  ```
   psiapi.py:8: FutureWarning: Using `psi4.core.get_variable` instead of `psi4.core.variable` (or 
   `psi4.core.scalar_variable` for scalar variables only) is deprecated, and in 1.4 it will stop working

    print(psi4.get_variable('current energy'))
   ```
   ```
    Using `psi4.core.get_variable` instead of `psi4.core.variable` (or `psi4.core.scalar_variable` for scalar variables only) is deprecated, and in 1.4 it will stop working

   ```

## Breaking Changes
- Not many, as there's heavy use of deprecation
- Any access that expected `Wfn.variables` to return only scalars will now be surprised to get all types. Easy to replace with `Wfn.get_scalar_variables`.
- Any code that wrote to Wfn array directly in non-UC (e.g., `wfn.variables_["lower or mixed-case var"] = val`) needs to be changed to `"UPPER-CASE VAR"` or to `wfn.set_scalar_variables`
- Any code that tries to access a non-existent global will now throw, hence the fixes to the driver and test cases you see here.

## Checklist
- [x] quick tests pass and I've run a full tests with this+other commits

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/p4util/python_helpers.py
```diff
@@ -30,12 +30,15 @@
...
 import uuid
+import warnings
+
 import numpy as np
 import qcelemental as qcel
 
 from psi4 import core
 from psi4.driver import qcdb
 from . import optproc
+from .exceptions import *
```
andysim: I think it would be nicer to enumerate the exceptions, rather than the wildcard.  It's kinda verbose, but it's also immediately obvious where each function came from.  You're far more updated on all the Pythonic conventions, so feel free to disregard.

loriab: You're right -- I've enumerated them in more recent projects. And most of the time in psi it's only `ValidationError` anyways.

psi4/driver/p4util/python_helpers.py
```diff
@@ -520,3 +523,184 @@ def py_psi_set_global_option_python(key, EXTERN):
...
+
+def _core_has_variable(key):
+    return core.has_scalar_variable(key) or core.has_array_variable(key)
+
+
+def _core_wavefunction_has_variable(cls, key):
+    return cls.has_scalar_variable(key) or cls.has_array_variable(key)
+
+
+def _core_variable(key, np_out=False):
```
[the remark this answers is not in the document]
loriab: good catch. forgot to remove. eventually, I'd like it to toggle between returning arrays as `psi4.core.Matrix` (default) and `ndarray`. Easy to program, but I haven't added the tests. And then, should they be stacked (same dims as Matrix) or flat (what all the schema's want)?

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -647,15 +647,30 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction> {
...
-    double get_variable(const std::string key);
-    void set_variable(const std::string key, double value) { variables_[key] = value; }
-    std::map<std::string, double> variables() { return variables_; }
-
-    /// Get and set arrays dictionary
-    SharedMatrix get_array(const std::string key);
-    void set_array(const std::string key, SharedMatrix value) { arrays_[key] = value; }
-    std::map<std::string, SharedMatrix> arrays() { return arrays_; }
+    /// Get and set variables and arrays dictionaries
+    int has_scalar_variable(const std::string& key);
```
andysim: I think the intent of the `has_` functions would be  more apparent if they were `bool` instead of `int`

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -1356,23 +1356,78 @@ std::vector<std::vector<std::tuple<double, int, int>>> Wavefunction::get_no_occu
...
+int Wavefunction::has_scalar_variable(const std::string &key) {
+    std::string uc_key = key;
+    std::transform(uc_key.begin(), uc_key.end(), uc_key.begin(), ::toupper);
+    return variables_.count(uc_key);
+}
+
+int Wavefunction::has_array_variable(const std::string &key) {
+    std::string uc_key = key;
+    std::transform(uc_key.begin(), uc_key.end(), uc_key.begin(), ::toupper);
+    return arrays_.count(uc_key);
```
loriab: @andysim ah, cast this to `bool` and change the return type? sounds good to me.

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -1356,23 +1356,78 @@ std::vector<std::vector<std::tuple<double, int, int>>> Wavefunction::get_no_occu
...
 }
 
-double Wavefunction::get_variable(std::string label) {
-    std::string uc_label = label;
 
-    if (variables_.count(uc_label) == 0) {
-        throw PSIEXCEPTION("Wavefunction::get_variable: Requested variable " + label + " was not set!\n");
+int Wavefunction::has_scalar_variable(const std::string &key) {
+    std::string uc_key = key;
+    std::transform(uc_key.begin(), uc_key.end(), uc_key.begin(), ::toupper);
```
robertodr: `psi4/libpsi4util/stl_string.cc` has a `to_upper_copy` function, so there's no need to do it "by hand" here.

psi4/src/psi4/libmints/wavefunction.cc
andysim: If the return type is bool, the compiler will handle the cast for you.  It makes no difference, of course, but it means at first glance new programmers know that this is saying whether a var exists or not, rather than _how many_ exists (which makes no sense, of course).

robertodr: Or even just `return (arrays_.find(uc_key) != arrays_.end());`

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -1356,23 +1356,78 @@ std::vector<std::vector<std::tuple<double, int, int>>> Wavefunction::get_no_occu
...
+    return arrays_.count(uc_key);
+}
+
+double Wavefunction::scalar_variable(const std::string &key) {
+    std::string uc_key = key;
+    std::transform(uc_key.begin(), uc_key.end(), uc_key.begin(), ::toupper);
+
+    if (variables_.count(uc_key) == 0) {
```
robertodr: Probably more intuitive for readers to use `find`:
```
if (variables_.find(uc_key) != variables_.end()) {
... 
} else {
}
```

psi4/src/psi4/libmints/wavefunction.cc
loriab: Thanks, this solved a mystery for me.. I had tried to add a `to_upper` fn into `wavefunction.cc` last week, and it kept complaining about bad `const`. It mush have been accessing the `stl_sting.cc` one. Since your cmake renovation, libpsi4util isn't accessible, hence the added header.

PR #1394, opened 2018-12-03
p4util deprecations, findif fn renaming, legacy gradient, dfmp2 timers

loriab:
## Description
Miscellaneous substantive changes short of moving vars from globals to wfn.

## Todos
- [x] for FINDIF, renamed some functions for easier comprehension (e.g., `compute_gradient_from_energy` --> `compute_gradient_from_energies`). speak up, @JonathonMisiewicz if I've misinterpreted anything (it won't count in GH's reckoning, but you're always welcome to review).
- [x] `core.get_gradient`/`core.set_gradient` aren't regularly set, even in v1.2, being used only in internal communication for OPTKING. I consider this effectively a silent deprecation in 1.2, so made that official (in case downstream using it), and renamed to `legacy_gradient` internally, until the demise of C-optking.
- [x] de-py2-adapted `psi4.driver.p4util`
- [x] fixed bug where in `driver_fn(..., kwargvar=None)`, `kwargvar` turned into `False`
- [x] deprecated many old fns in p4util -- see if you agree. Most for v1.4 but a few for v1.3.
- [x] added tolerance field to `compare_wavefunctions`
- [x] renamed and `_`-prefixed many functions in python_helpers.py so connection to demangled fn name easier to track and the `import *` doesn't contaminate namespace
- [x] for dfmp2, energy was starting the timer _after_ `scf_helper`, gradient was starting timer _before_ SCF, and properties wasn't timing at all. changed the latter two to match the first.
- [x] added `set_energy` to Wfn
- [x] `psi4 --fulltest -n NTHREAD` available to run full pytest test suite. Parallel only available if `pytest-xdist` installed (`conda install pytest-xdist`)

## Checklist
- [x] quick tests is fine at this PR. full tests of this + other commits is fine.

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
I really couldn't care less about pluralization in function names, but this is going to break the three findif related tests in `cookbook`.

Expected timeline on getting this in? The changes to read/write wavefunction syntax will impact the updated orbital guess code I should be submitting tomorrow.

I also never thought I'd see you _remove_ hand-formatting. See the read/write wavefunction code.

loriab:
Thanks for the reminder. I'll patch up the `cookbook`.

I'm going to push to get this weekend's PRs in soon so that my other non-separable changes can get posted. All I really wanted to do was add in MP2-D, but there turned out to be a lot of opportunities to either work toward future plans or work to put in hacks to maintain existing plans, and I chose the former. Go ahead and post your PR as-is (or copy over formatting changes). I don't care how much I have to rebase, and if the two PRs merges are going to be close in time, I'll see that yours goes in first.

Yeah, I `yapf`ed it, then made the minimal changes toward readability.

psi4/driver/p4util/procutil.py
```diff
@@ -25,23 +25,19 @@
...
-from __future__ import print_function
 import os
 import ast
 import sys
 import pickle
 import inspect
+import warnings
 import collections
 
 from .exceptions import *
```
andysim: I know this isn't part of your PR, but could you remove this wildcard also if you make any more changes to this file, please?

loriab: fixed all the exception wildcards in #1393

andysim: Great, thanks!

tests/pytest/addons.py
```diff
@@ -1,6 +1,29 @@
+import os
+import subprocess
+
 import pytest
 
 
+def _which(command):
```
dgasmith: Can we use [shutil.which](https://docs.python.org/3/library/shutil.html#shutil.which) here?

JonathonMisiewicz:
And if we're talking about pulling this in, reminder that the `cookbook` tests still need to be fixed.

tests/pytest/addons.py
loriab: tested and switched out

PR #1397, opened 2018-12-03
clang-format libfock module

loriab:
## Description
Relevant to #1144. I think I caught all the hand-formatted periodic tables.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/CDJK.cc
+++ b/psi4/src/psi4/libfock/CDJK.cc
@@ -52,57 +52,48 @@ using namespace psi;
 
 namespace psi {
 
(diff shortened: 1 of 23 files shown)


JonathonMisiewicz:
We're investigating a `libmints` bug here at UGA that's blocking my symmetry casting PR. `libfock` is likely highly affected, so best to get approval from @jturney for this one.

loriab:
Good to know. Jet can put a block on it if it's long term. For future reference, `cubature.cc` is the file that has clang-format directives to preserve hand formatting.

PR #1398, opened 2018-12-03
Moves to C++14 and fixes a SAPT memory bug

robertodr:
## Description
Set the minimum required C++ standard to C++14. Lower bounds on compilers can be found [here](https://en.cppreference.com/w/cpp/compiler_support#cpp14).  Also addresses a bug in the SAPT code that resulted in a double free.

## About the SAPT bug
The Iterator class in SAPT contained an `int *` array that was cleaned up by the class's destructor.  Iterator objects are constructed by a builder routine that first creates a local Iterator object, fills it, then returns it by value.  The C++98 behavior of this design is problematic; a copy of the local temp Iterator is made and that copy is returned to the caller.  When that copy is made, both the local temp and its copy have `int *` pointers that point to the same pool of memory because no deep-copy copy constructor exists for Iterator.  Upon returning, the local temp object is destroyed, triggering the memory pointed to by the `int *` to be freed, causing the returned object to point to freed memory which is undefined behavior.  When that returned object eventually goes out of scope, it will try to free the memory again, leading to the double free memory error we observed.  Because we use C++11 most compilers appear to be able to correctly elide the copy, by implementing move semantics, so we haven't seen this before.  The switch to C++14 with GCC5.4 caused consistent segfaults, revealing the issue.  The fix is simple; don't use raw `int *`, but `std::vector<int>` instead; the lifetime of these is correctly managed automatically and the various move constructor/copy constructor/destructor can be generated correctly by the compiler.  RAII for the win!

## Todos
- [x] Fix C++14 compliance CMake-side
- [x] Fix memory bug in SAPT code
- [x] Move Travis to Xenial

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.scripts/miniconda.sh
+++ b/.scripts/miniconda.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+echo "-- Installing latest Miniconda"
+if [ -d "$HOME/Deps/miniconda/bin" ]; then
+    echo "-- Miniconda latest version FOUND in cache"
+else
+    cd "$HOME"/Downloads
+    echo "-- Miniconda latest version NOT FOUND in cache"
+    if [ "$TRAVIS_OS_NAME" = "linux" ]; then
+        curl -Ls https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > miniconda.sh
+    elif [ "$TRAVIS_OS_NAME" = "osx" ]; then
+        curl -Ls https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh > miniconda.sh
+    fi
+    # Travis creates the cached directories for us.
+    # This is problematic when wanting to install Anaconda for the first time...
+    rm -rf "$HOME"/Deps/miniconda
+    bash miniconda.sh -b -p "$HOME"/Deps/miniconda > /dev/null 2>&1
+    touch "$HOME"/Deps/miniconda/conda-meta/pinned
+    PATH=$HOME/Deps/miniconda/bin${PATH:+:$PATH}
+    hash -r
+    conda config --set show_channel_urls True > /dev/null 2>&1
+    conda config --set always_yes yes > /dev/null 2>&1
+    conda config --set changeps1 no > /dev/null 2>&1
+    conda update --all --yes > /dev/null 2>&1
+    conda clean -tipy > /dev/null 2>&1
+    conda info -a
+    cd "$TRAVIS_BUILD_DIR"
+    rm -f "$HOME"/Downloads/miniconda.sh
+fi
+echo "-- Done with latest Miniconda"
--- a/.scripts/miniconda.sh
+++ b/.scripts/miniconda.sh
@@ -1,32 +1,47 @@
-#!/usr/bin/env bash
-
 set -eu -o pipefail
 
+# Temporarily change directory to $HOME to install software
+pushd .
+cd "$HOME"
+
+# Install Miniconda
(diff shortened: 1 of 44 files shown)


.scripts/miniconda.sh
```diff
@@ -0,0 +1,32 @@
...
+    conda config --set show_channel_urls True > /dev/null 2>&1
+    conda config --set always_yes yes > /dev/null 2>&1
+    conda config --set changeps1 no > /dev/null 2>&1
+    conda update --all --yes > /dev/null 2>&1
+    conda clean -tipy > /dev/null 2>&1
+    conda info -a
+    cd "$TRAVIS_BUILD_DIR"
+    rm -f "$HOME"/Downloads/miniconda.sh
+fi
+echo "-- Done with latest Miniconda"
```
loriab: has this done a bout with the latest molssi conda downloader to see who advances? https://github.com/MolSSI/QCFractal/blob/master/devtools/travis-ci/before_install.sh

loriab: may as well, unless you see something that should be merged into it

robertodr: The cache check is missing from the MolSSI script. I'll merge the two.

loriab:
Well, SAPT (and maybe SAPT Disp) is the common theme to the errors. Strange!

robertodr:
Annoying :rage: Any idea what to try out?

andysim:
It looks like those changes did the trick.  The tests passed fairly quickly on travis, but i'm not sure what the timings were like before so i don't know how much impact the header refactoring had.  When you think this one's good, could you ping us on Slack, please? The seg fault seems to be hitting a few people so it would be good to get it in quickly.

robertodr:
I'm still figuring out the caching of Conda-installed dependencies.

andysim:
I updated the title and posted some blurb describing the issue.  I approve the PR, but I'm a part author so maybe another reviewer would be good.

robertodr:
Thanks @andysim! The caching of conda is currently a bit idiotic: I am nuking `p4env` every time after a build succeeds. Probably we want to keep the conda environment in cache and just activate and update it. @loriab does that sound like a reasonable proposition? I'll tackle it in a subsequent PR.

loriab:
For cacheing, keeping “ miniconda/packages” dir around is what I expect to be the key part. So long as the mkl dir or tarball is in there, making the conda env is trivial.

robertodr:
As I said, my current take is idiotic. The only thing kept around is the Conda installation itself, not the environment.

loriab:
What I meant was that that may be just fine. If what's being cached is `$HOME/miniconda/bin` and `$HOME/miniconda/packages`, then you've preserved all the heavy stuff (mkl w/i the packages dir). A new `conda create -n p4dev` that goes into `$HOME/miniconda/envs/p4dev` is trivial compared to downloading all the packages. I'd prefer the env to be re-solved each time, so I'm not advocating caching p4dev env.

.scripts/miniconda.sh
loriab: @dgasmith is that a MKL or no-MKL DL/install?

loriab: oh, and when we switch to AM8 libint, that's also an issue (though probably could arrange AM6 for CI).

dgasmith: MKL install which does take ~40 seconds on its own.

dgasmith: Historically caching has provided an array of issues, but it might be worth seeing if the above causes any instabilities as these things do evolve quickly. So I would leave it in for now and simply be watchful.

dgasmith:
Let's wait on a review from @jturney before emerging just to give it a final once over.

PR #1400, opened 2018-12-04
Expose good chunks of libmints and libciomr to plugins

fevangelista:
## Description
This PR exposes all of libmints and libciomr to plugins via the PSI_API macro.

## Todos
- [x] Expose libmints
- [x] Expose libciomr
- [x] Expose JK and Denominator classes
- [x] Expose Options print functions

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libciomr/libciomr.h
```diff
@@ -44,77 +44,80 @@
...
+PSI_API void zero_long_int_matrix(long int **array, int rows, int cols);
+PSI_API void print_long_int_mat(long int **a, int m, int n, std::string out);
 
 /* Functions in block_matrix.c */
 PSI_API double **block_matrix(size_t n, size_t m, bool mlock = false);
-void free_block(double **array);
+PSI_API void free_block(double **array);
 
 /* Functions in fndcor */
+<<<<<<< HEAD
```
robertodr: Merge conflict marker left-over

fevangelista: Got it!

fevangelista:
@jturney: I concur in theory, but practically that's going to take a bit of time to execute. Is hiding symbols the right way to prevent a user from constructing an object from a derived class?

fevangelista:
I posted some obvious fixes. @jturney: Can you please advise on whether or not to expose the following?
- Classes derived from `PotentialInt `, e.g. `class PSI_API ElectrostaticInt : public PotentialInt {`
- Anything derived from `TwoBodyAOInt`, `GaussianFundamental`, `Fjt`
- Classes derived from `CoordEntry` or `CoordValue`

jturney:
- No need to export classes that derive from `PotentialInt` they all derive from `OneBodyAOInt`. 
- Don't export anything that derives from `TwoBodyAOInt`.  I don't think even `GaussianFundamental` and `Fjt` or their derived classes need to be exported as they are implementation specific.
- I'm not sure about those derived from `CoordEntry` and `CoordValue`. With so much molecule construction happening Py-side I'm not sure if these classes will exist long term. @loriab might be better to answer this one.

fevangelista:
Thanks! It looks like I will have to move the function `SharedVector DipoleInt::nuclear_contribution` out of the class `DipoleInt::` otherwise it won't be exposed since `DipoleInt` is a `class DipoleInt : public OneBodyAOInt`. Any idea of where to put it?

loriab:
Even c-side Mol manipulation has very little (no?) need to access `CoordEntry`/`CoordValue` directly. I'd leave them hidden unless a need arises.

fevangelista:
Actually, it's a bit of a mess, since several of the classes derived from `OneBodyAOInt` have different variations on the same function with different signatures:
```
    // in class ElectricFieldInt : public OneBodyAOInt 
    static Vector3 nuclear_contribution(const Vector3 &origin, std::shared_ptr<Molecule> mol);
    static SharedMatrix nuclear_contribution_to_gradient(const Vector3 &origin, std::shared_ptr<Molecule> mol);

    // in class MultipoleInt : public OneBodyAOInt
    static SharedVector nuclear_contribution(std::shared_ptr<Molecule> mol, int order, const Vector3 &origin);

    // in class QuadrupoleInt : public OneBodyAOInt
    static SharedVector nuclear_contribution(std::shared_ptr<Molecule> mol, const Vector3 &origin);
```

fevangelista:
Ok, I addressed the classes that are derived mentioned in the post above. For the `nuclear_contribution` function I was thinking of adding a virtual function to the base class and standardizing all functions to use SharedVector/SharedMatrix. How does that sound to you @jturney?

jturney:
Yeah, that could work. The other changes are looking good to me, too.

fevangelista:
I found a copy of the function to computer the nuclear contribution to the dipole moment in the Molecule class. So I am going to leave that part of the refactoring for another PR. I think this is ready to go once all tests pass.

jturney:
Looks good to me. I'll let someone else perform the actual merge after they look it over again.

PR #1401, opened 2018-12-04
Polarizable Embedding model

maxscheurer:
## Description
This PR's purpose is to add the Polarizable Embedding (PE) model to Psi4.

In the [PE](https://pubs.acs.org/doi/10.1021/ct1003803) model, the environment is modeled by a multi-center multipole expansion with dipole-dipole polarizabilities placed at the expansion sites (similar to EFP). A tutorial on PE in general can be found [here](https://onlinelibrary.wiley.com/doi/full/10.1002/qua.25717).

I added the PE model through my [CPPE](https://github.com/maxscheurer/cppe) library on the Python level.

## Todos
- [x] Ground-state (HF and DFT) SCF calculations with PE
~~- [ ] Perturbative state-specific (ptSS) corrections for EOM-CC calculations~~ can be done in user code on the Python layer
- [x] rebase once #1657 is merged
- [x] include `cppe` on the Python layer
- [x] remove C++ code
- [x] cleanup CMake
- [x] pin a released CPPE version
- [x] bump pybind to v2.3.0
- [x] simple documentation

## Questions
- [x] The tests need a _potential_ input file (`*.pot`), currently residing in the `tests/cppe/scf-hf` folder. The path specification I have to put in the input file is currently a bit weird because the tests are not run from inside the test directory. Is there any nice solution to this issue?

## Checklist
- [x] Tests added for any new features (SCF)

## Status
- [x] Ready for review
- [x] Ready for merge


tests/cppe/scf-hf/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(cppe-scf-hf "psi;cppe;addon;scf")
```
loriab: Use commands like https://github.com/psi4/psi4/blob/master/tests/cubeprop/CMakeLists.txt here to get your `*.pot` file available at test-time.

CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.3 FATAL_ERROR)  # 3.3: EP_add(DEPENDS interface
                                                  # 3.2: continue()
                                                  # 3.1: CMAKE_CXX_STANDARD, BUILD_ALWAYS
 project(psi4
-        LANGUAGES C CXX)
+	LANGUAGES C CXX Fortran)
```
loriab: Don't add `Fortran` here. Instead add your pkg to https://github.com/psi4/psi4/pull/1401/files#diff-af3b638bc2a3e6c650974192a53c7291R106

CMakeLists.txt
```diff
@@ -85,6 +85,7 @@ option(ENABLE_ambit "Enables the ambit tensor library" OFF)
 option(ENABLE_CheMPS2 "Enables CheMPS2 for DMRG (requires HDF5)" OFF)
 option(ENABLE_dkh "Enables DKH integrals (requires Fortran)" OFF)
 option(ENABLE_libefp "Enables LIBEFP and PylibEFP for fragments" OFF)
+option(ENABLE_cppe "Enables CPPE for Polarizable Embedding" OFF)
```
loriab: alphabetize, please, ignoring "lib"

external/upstream/cppe/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+        include(ExternalProject)
+        message(STATUS "Suitable cppe could not be located, ${Magenta}Building cppe${ColourReset} instead.")
+        ExternalProject_Add(cppe_external
+            GIT_REPOSITORY https://github.com/maxscheurer/cppe
+            GIT_TAG master
+            # UPDATE_COMMAND ""
+            CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
+                       -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+                       -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
```
loriab: Fortran compiler needed?

psi4/driver/procrouting/proc.py
```diff
@@ -1371,6 +1371,42 @@ def scf_helper(name, post_scf=True, **kwargs):
...
+        if pol_embed_options.pe_border:
+            pol_embed_border_options = core.PeBorderOptions()
+            pe_btype = core.get_local_option('PE', 'BORDER_TYPE').upper()
+            if pe_btype == "REMOVE":
+                pol_embed_border_options.border_type = core.PeBorderOptions.BorderType.rem
+            elif pe_btype == "REDIST":
+                pol_embed_border_options.border_type = core.PeBorderOptions.BorderType.redist
+            pol_embed_border_options.rmin = core.get_local_option('PE', 'BORDER_RMIN')
+            if core.get_local_option('PE', 'BORDER_RMIN_UNIT').upper() == "AA":
+                pol_embed_border_options.rmin *= 1.8897261246
```
loriab: `constants.bohr2angstroms`

psi4/src/psi4/libpsipe/pe_calc_handler_i.hh
```diff
@@ -0,0 +1,31 @@
+#ifndef LIBPSIPE_PE_CALC_HANDLER_H
```
loriab: Please switch to `.h` extension. And avoid "psi" in dirname -- we've got so many already :-)

psi4/src/read_options.cc
```diff
@@ -248,6 +248,40 @@ int read_options(const std::string &name, Options & options, bool suppress_print
...
   }
+  
+  /*- PE boolean for polarizable embedding module -*/
+  options.add_bool("PE", false);
+  if (name == "PE" || options.read_globals()) {
+    /*- MODULEDESCRIPTION Performs polarizable embedding model (PE) computations. -*/
+
+    /*- Name of the potential file -*/
+    options.add_str_i("POTFILE", "potfile.pot");
+    /* - Use DIIS acceleration to obtain induced moments -*/
```
loriab: Make sure the comment delimiters are exactly `/*-` and `-*/`. The scraper is picky

psi4/src/read_options.cc
```diff
@@ -248,6 +248,40 @@ int read_options(const std::string &name, Options & options, bool suppress_print
...
+    options.add_bool("DIIS", true);
+    /* - Threshold for induced moments convergence -*/
+    options.add_int("CONVERGENCE_INDUCED", 8);
+    /* - Maximum number of iterations for induced moments -*/
+    options.add_int("MAXITER", 50);
+    
+    /* - Activate border options for sites in proximity to the QM/MM border -*/
+    options.add_bool("BORDER", false);
+    /* - border type, either remove or redistribute moments/polarizabilities -*/
+    options.add_str_i("BORDER_TYPE", "REMOVE", "REMOVE REDIST");
```
loriab: Avoid the case-preserving `str_i` except where it's really needed, like filenames.

psi4/src/psi4/libpsipe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+#ifdef USING_cppe
+
+#include <iomanip>
+
+#include <armadillo>
```
loriab: Where is armadillo coming from in the build process? How onerous would it be to switch out for LAPACK calls or `psi4.core.Matrix`?

tests/cppe/scf-hf/CMakeLists.txt
maxscheurer: Addressed in 1930d51b585da776121e44a700b166c00765a99d.
Thanks!

CMakeLists.txt
maxscheurer: Oops, I don't remember why I even put this Fortran flag there...
Resolved in a030a01c6.

CMakeLists.txt
maxscheurer: Resolved in a030a01c6.

psi4/src/psi4/libpsipe/pe_calc_handler_i.hh
maxscheurer: Resolved in 893f823f7 and 01ae0e5ac.
The header file you mentioned was not being used and I just forgot to remove it...
I renamed the `libpsipe` directory to `libpe`.

psi4/src/psi4/libpsipe/psipe.cc
maxscheurer: When the CPPE library is built, cmake searches for armadillo on the system, and builds it from source if it's not found ([here is the code in my repo](https://github.com/maxscheurer/cppe/blob/master/external/upstream/armadillo/CMakeLists.txt)). The stand-alone CPPE build works fine on Travis CI at least.

I think at the moment it is very hard to replace armadillo since this would require re-writing a major part of the library, but I understand that having yet another linear algebra library
is annoying...

Just some additional information: I originally developed the CPPE library for PE in Q-Chem where armadillo commonly used, so it was kind of convenient to use it for CPPE, too.

psi4/src/export_cppe.cc
```diff
@@ -0,0 +1,73 @@
...
+#include "psi4/pybind11.h"
+
+#include "psi4/libmints/basisset.h"
+#include "psi4/libmints/matrix.h"
+#include "psi4/libpsipe/psipe.h"
+
+using namespace psi;
+#ifdef USING_cppe
+
+void export_cppe(py::module& m) {
```
dgasmith: @loriab What do you think about name spacing such extensions under `core.cppe.*`?

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+#include <cppe/utils/pot_manipulation.hh>
+#include <cppe/core/multipole_expansion.hh>
+
+
+#include "psipe.h"
+
+// TODO: implement output stream option in CPPE for printing
+
+namespace psi {
+
```
dgasmith: Can you add a `cppe` namespace or similar here as well?

dgasmith:
Can you expand a bit on "not really header-only"? Checking the source code, I only see `hpp` files.

maxscheurer:
> Can you expand a bit on "not really header-only"? Checking the source code, I only see `hpp` files.

Sure. If the armadillo runtime library is to be used, `src/wrapper.cpp` is compiled and
one can simply link against `armadillo`. However, this wrapper can also be switched off. In this case, one needs to link directly against BLAS and LAPACK. The [documentation](http://arma.sourceforge.net/faq.html) explains this in a little more detail in the _Linking_ section.

I have never tried to use armadillo without the wrapper to be honest... I'll give it a try.

psi4/src/export_cppe.cc
loriab: I haven't thought about such matter, but it feels sensible. So cautious yes. Though I don't think we want to go explicit subpackage, `import psi4.core.cppe` with separate compiled lib, as @robertodr was having some problems with structuring that.

robertodr: @loriab I managed to solve the problem, so it can be done without too much effort.

external/upstream/cppe/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+                       -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                       -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
+                       -DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}
+                       -DCMAKE_INSTALL_DATADIR=${CMAKE_INSTALL_DATADIR}
+                       -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                       -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+                       -DENABLE_OPENMP=${ENABLE_OPENMP}
+                       -DENABLE_PYTHON_INTERFFACE=OFF
```
robertodr: I think you have a typo in this option.

psi4/driver/procrouting/proc.py
```diff
@@ -1371,6 +1371,42 @@ def scf_helper(name, post_scf=True, **kwargs):
         core.print_out("""  PCM does not make use of molecular symmetry: """
                        """further calculations in C1 point group.\n""")
         use_c1 = True
+
```
robertodr: Since a 3-layer QM/MM/PCM is currently not possible, we need to check that both options are not set at the same time and abort if that's the case.

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+    return std::pair<double, SharedMatrix>(energy_result, V_ind);
+}
+
+void PeState::print_energy_summary() {
+    cppe_state_.print_summary();
+}
+
+SharedMatrix PeIntegralHelper::compute_multipole_potential_integrals(Vector3 site, int order, std::vector<double>& moments) {
+    auto integrals = std::make_shared<IntegralFactory>(basisset_, basisset_, basisset_, basisset_);
+    std::shared_ptr<OneBodyAOInt> multipole_integrals = static_cast<std::shared_ptr<OneBodyAOInt>>(integrals->ao_efp_multipole_potential());
```
robertodr: I think `auto multipole_integrals = std::make_shared<OneBodyAOInt>(integrals->ao_efp_multipole_potential());` should suffice here.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -826,6 +827,39 @@ void OEProp::compute() {
     if (tasks_.count("NO_OCCUPATIONS")) compute_no_occupations();
     if (tasks_.count("GRID_FIELD")) compute_field_over_grid();
     if (tasks_.count("GRID_ESP")) compute_esp_over_grid();
+    if (tasks_.count("PE_CORR")) compute_pe();
+}
+
+PePropCalc::PeOutputType PePropCalc::compute_pe_prop(bool transition, bool print_output, bool verbose) {
```
robertodr: I don't think this belongs here as a one-electron property. It also seems the only thing you need to compute the perturbative correction is the AO density matrix, hence this can possibly become a free function in the Python layer of the code?

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+}
+
+SharedMatrix PeIntegralHelper::compute_field_integrals(Vector3 site, arma::vec moment) {
+    std::vector<SharedMatrix> fields;
+    fields.push_back(std::make_shared<Matrix>("AO Field X", basisset_->nbf(), basisset_->nbf()));
+    fields.push_back(std::make_shared<Matrix>("AO Field Y", basisset_->nbf(), basisset_->nbf()));
+    fields.push_back(std::make_shared<Matrix>("AO Field Z", basisset_->nbf(), basisset_->nbf()));
+    
+    auto integrals = std::make_shared<IntegralFactory>(basisset_, basisset_, basisset_, basisset_);
+    std::shared_ptr<OneBodyAOInt> field_integrals_ = static_cast<std::shared_ptr<OneBodyAOInt>>(integrals->electric_field());
```
robertodr: I think `auto field_integrals_ = std::make_shared<OneBodyAOInt>(integrals->electric_field());` should suffice here too.

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+            Vector3 site(p.m_x, p.m_y, p.m_z);
+            std::vector<double> moments;
+            for (auto& m : p.get_multipoles()) {
+                m.remove_trace();
+                if (m.m_k > 2) throw std::runtime_error("PE is only implemented up to quadrupoles.");
+                for (auto d : m.get_values()) moments.push_back(d);
+            }
+            V_es_->add(int_helper_.compute_multipole_potential_integrals(site, 2, moments));
+        }
+        cppe_state_.set_es_operator(arma::mat(V_es_->get_pointer(), V_es_->nrow(), V_es_->ncol(), true));
```
robertodr: Would it be possible to add an overload for this function in CPPE that accepts the raw pointer and the sizes and reconstructs the Armadillo matrix internally? It could also take care of saving the matrix to text file. 

I guess this could be a good strategy to avoid Armadillo spilling into Psi4 (or other QC programs, for that matter)

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+        cppe_state_.es_operator_copy().save("V_es_psi4.txt",  arma::raw_ascii);
+    }
+    
+    SharedMatrix D = Dm;
```
robertodr: I am not positive you need to keep a copy of the SCF density matrix at hand here. For example, in CC you can pass in the correlation-only density matrix and set the `CalcType` to `electronic_only` to achieve the same effect (I think there's an outdated PR of mine showing that for PCM). And if that's not feasible, the wavefunction itself carries the current density matrix, so you can do the subtraction _outside_ of here and simplify the state of the `PeState` object a bit.

psi4/src/psi4/libpe/psipe.cc
robertodr: Also, why do you need to keep track of the electrostatics operator inside the library?

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+    if (subtract_scf_density && type == PeState::CalcType::electronic_only) {
+        D->subtract(D_scf_);
+    }
+    cppe_state_.update_energies(arma::mat(D->get_pointer(), D->nrow(), D->ncol(), true));
```
robertodr: Having not read the CPPE code, I think the naming of this function is a bit confusing. If it updates energies, why does it take the density matrix as argument? Is it because you trace the operator matrices with the density matrix internally to get the embedding energy? Couldn't that be done Psi4-side?

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+    }
+    
+    if (type == PeState::CalcType::total) {
+        V_ind->add(V_es_);
+        D_scf_ = D;
+    }
+    // TODO: only for debugging
+    // cppe_state_.print_summary();
+    
+    if (type == PeState::CalcType::total) iteration++;
```
robertodr: The iteration number could be passed from outside when calling this function, rather than being kept track of here. Why does it only increment when `CalcType` is `total`? In some CC models with polarization you also need to iterate until self-consistency, but `CalcType` would be `electronic_only`.

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+    fields.push_back(std::make_shared<Matrix>("AO Field X", basisset_->nbf(), basisset_->nbf()));
+    fields.push_back(std::make_shared<Matrix>("AO Field Y", basisset_->nbf(), basisset_->nbf()));
+    fields.push_back(std::make_shared<Matrix>("AO Field Z", basisset_->nbf(), basisset_->nbf()));
+    
+    auto integrals = std::make_shared<IntegralFactory>(basisset_, basisset_, basisset_, basisset_);
+    std::shared_ptr<OneBodyAOInt> field_integrals_ = static_cast<std::shared_ptr<OneBodyAOInt>>(integrals->electric_field());
+    field_integrals_->set_origin(site);
+    field_integrals_->compute(fields);
+    
+    fields[0]->scale(-1.0*moment[0]);
```
robertodr: A bit more compact:
```
for (int i = 0; i < 3; ++i) {
   fields[i]->scale(-1.0*moment[i]);
}
```

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+    fields.push_back(std::make_shared<Matrix>("AO Field Y", basisset_->nbf(), basisset_->nbf()));
+    fields.push_back(std::make_shared<Matrix>("AO Field Z", basisset_->nbf(), basisset_->nbf()));
+    
+    auto integrals = std::make_shared<IntegralFactory>(basisset_, basisset_, basisset_, basisset_);
+    std::shared_ptr<OneBodyAOInt> field_integrals_ = static_cast<std::shared_ptr<OneBodyAOInt>>(integrals->electric_field());
+    field_integrals_->set_origin(site);
+    field_integrals_->compute(fields);
+    
+    Vector efields("efields", 3);
+    efields.set(0, D->vector_dot(fields[0]));
```
robertodr: A bit more compact:
```
for (int i = 0; i < 3; ++i) {
  efields[i] = D->vector_dot(fields[i]);
}
```

psi4/src/psi4/libpe/psipe.h
```diff
@@ -0,0 +1,110 @@
...
+
+    /* \brief prints the summary table of PE energy contributions to the Psi4 output file
+    */
+    void print_energy_summary();
+
+    // void calculate_fock_contribution(arma::mat& Ptot, const libqints::multi_array<double>& out, double* energy);
+    // void calculate_excited_state_energy_correction(double* p_exc, double* energy, bool is_tdm);
+
+  private:
+      std::shared_ptr<BasisSet> basisset_;
```
robertodr: I guess once the `PeIntegralHelper` is initialized, it will hold the `basisset_` and there's probably no need for a redundant instance in this class?

psi4/src/psi4/libpe/psipe.cc
```diff
@@ -0,0 +1,259 @@
...
+      int_helper_(PeIntegralHelper(basisset)) {
+
+    potentials_ = cppe_state_.get_potentials();
+    cppe_state_.calculate_static_energies_and_fields();
+}
+
+
+std::pair<double, SharedMatrix> PeState::compute_pe_contribution(const SharedMatrix &Dm, CalcType type, bool subtract_scf_density) {
+    // build the electrostatics operator only once!
+    if (!iteration && (type == PeState::CalcType::total)) {
```
robertodr: If this is built once and stored, it's probalby better to do it in the CTOR.

psi4/src/psi4/libpe/psipe.h
```diff
@@ -0,0 +1,110 @@
...
+      private:
+          std::shared_ptr<BasisSet> basisset_;
+  };
+
+  class PeState {
+
+  public:
+    enum CalcType { total, electronic_only };
+    PeState() = default;
+    PeState(libcppe::PeOptions options, std::shared_ptr<BasisSet> basisset);
```
robertodr: This CTOR can be marked `explicit`.

psi4/src/psi4/libpe/psipe.h
```diff
@@ -0,0 +1,110 @@
...
+    // void calculate_excited_state_energy_correction(double* p_exc, double* energy, bool is_tdm);
+
+  private:
+      std::shared_ptr<BasisSet> basisset_;
+      std::vector<libcppe::Potential> potentials_;
+      libcppe::CppeState cppe_state_;
+      PeIntegralHelper int_helper_;
+      
+      SharedMatrix V_es_;
+      SharedMatrix D_scf_;
```
robertodr: As per comments above, `D_scf_` and `iteration` are most likely redundant.

susilehtola:
Armadillo can be used without the wrapper without any problems. Personally I have problems understanding why the wrapper was implemented in the first place, it usually just is in the way. All that the wrapper does is link to the BLAS and LAPACK library.

maxscheurer:
@robertodr Thanks for the detailed code review. I have to admit that I obviously did not clean up the code properly, sorry for that. Also, my CPPE library is still in an early stage and under heavy development, so thanks for pointing out some design/naming flaws. I'll try to address your points as soon as I find the time.

psi4/driver/procrouting/proc.py
maxscheurer: Resolved in c3cd34779.

psi4/src/psi4/libpe/psipe.h
maxscheurer: Done in b9742ad4d.

psi4/src/psi4/libpe/psipe.cc
maxscheurer: Done in b9742ad4d.

psi4/src/psi4/libpe/psipe.cc
maxscheurer: Done in b9742ad4d.

psi4/src/psi4/libpe/psipe.cc
maxscheurer: Resolved in b9742ad4d.

psi4/src/psi4/libpe/psipe.cc
maxscheurer: Done in b9742ad4d.

external/upstream/cppe/CMakeLists.txt
maxscheurer: Corrected in e8ab7d0b0e153ff326918521bd26cc13d738b481.

psi4/src/psi4/libpe/psipe.cc
maxscheurer: I agree. Removed it in 22bb0665c.

psi4/src/psi4/libpe/psipe.cc
maxscheurer: Resolved in be3006403.

dgasmith:
@robertodr @loriab Can we get a fresh look here when you have a chance?

psi4/src/psi4/libmints/multipolepotential.cc
```diff
@@ -0,0 +1,221 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+#include "psi4/libmints/multipolepotential.h"
+#include "psi4/libmints/basisset.h"
+#include "psi4/libmints/integral.h"
+
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
```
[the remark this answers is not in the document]
jturney: I can't remember. Feel free to change to the `std::` versions.

maxscheurer:
Since `cppe` has its own Python interface now (which it did not when I started this PR), I guess there is no reason to interface with psi4 on the C++ layer?

I guess it would be much simpler wrt maintenance of the code to refactor everything and use the `cppe` Python interface directly. What do you think, @loriab and @dgasmith ?

dgasmith:
I think that is likely the way to go, I believe you will still need to merge some of the new integral code. It may be worth splitting that off into a another PR to get a digestible component there.

We usually can merge small chunks quicker and expanding functionality through python is usually quite preferred if that level of integration is sufficient.

maxscheurer:
Okay, then I'll go for it.

> It may be worth splitting that off into a another PR to get a digestible component there.
Alright.

maxscheurer:
I added some experimental changes to `electricfield.{cc,h}` in order to compute
field integrals (for the induction operator) and electric field expectation values (needed to solve for induced moments) for multiple sites/coordinates at once.

Even though I don't like the code I implemented, it gives massive performance improvements.
In my tests (Python code is uploaded [here](https://gist.github.com/maxscheurer/4736a100ff5de12190ef27f7bc2eaade)), I computed integrals at 2000 random points for Water/cc-pVQZ, for which I get a 25x speed-up compared to the existing implementation. EFP would also benefit from these performance improvements.

Since the code is rather ugly at the moment, I'm thankful for any suggestions on how to proceed from here.

dgasmith:
The integral speedups would look awesome, but would be good in another PR. The more digestible the better! If you could split this off we could continue this conversation in the new PR.

maxscheurer:
> The integral speedups would look awesome, but would be good in another PR. The more digestible the better! If you could split this off we could continue this conversation in the new PR.

Sounds like a good plan.

dgasmith:
Awesome, if you pull that out. Is this PR ready for review and merge?

maxscheurer:
No, I have to refactor the code such that the PE stuff is Python-side only (see ToDo list above).
In the meantime, I'll hopefully a) have a first "release" of `cppe` ready that can be pinned in the `ExternalProject` and b) have the PR with performance improvements for electric field integrals done.

Is there a deadline for the Psi4 1.4 release?

dgasmith:
I think we have a general target of this fall for Psi4 1.4

maxscheurer:
> I think we have a general target of this fall for Psi4 1.4

Alright, that sounds feasible :blush:

maxscheurer:
**Update**

I've completely replaced the C++-based implementation of PE with a light-weight Python-side version.
Further, I had to bump pybind11 to `v.2.2.4` for the CPPE library to work correctly, and I pinned a release version of CPPE in the `ExternalProject`.

So I guess this PR will be ready for merge soon. I'm going to add some simple documentation now.

psi4/driver/procrouting/proc.py
```diff
@@ -54,6 +54,15 @@
 from . import mcscf
 from . import response
 
+_have_cppe = False
```
robertodr: Doesn't the same portion of code in `solvent/__init__.py` suffice?

maxscheurer: Yes. Done.

maxscheurer:
Basic documentation is now added.

I think the only thing missing is to push a recent version of CPPE (0.0.8) to conda, @loriab .
Other than that, this is ready for merge, in my opinion.

external/upstream/cppe/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+            CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
+                       -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+                       -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+                       -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                       -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
+                       -DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}
+                       -DCMAKE_INSTALL_DATADIR=${CMAKE_INSTALL_DATADIR}
+                       -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                       -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+                       -DENABLE_OPENMP=${ENABLE_OPENMP}
```
loriab: How much of a performance hit would turning off OpenMP (in cppe, not psi4) be? Not permanently but as a temporary, let's-get-this-integrated measure.

I suggest this because https://github.com/maxscheurer/cppe/blob/master/cppe/CMakeLists.txt#L12-L18 is fine for standalone, but it's going to choose a threading library based only on compiler, not taking LAPACK threading or `OpenMP_LIBRARY_DIRS` into account. Psi uses the crazy [`MathOpenMP`](https://github.com/psi4/psi4/blob/master/external/common/lapack/FindMathOpenMP.cmake) to get this right, a scheme I've no affection for but guardedly trust. We can find a way to hand the correct openmp flags to cppe (which is how we do lapack for other addons), but I don't want that detail to hold up this PR.

external/upstream/pybind11/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ else()
     include(ExternalProject)
     message(STATUS "Suitable pybind11 could not be located, ${Magenta}Building pybind11${ColourReset} instead.")
     ExternalProject_Add(pybind11_external
-        URL https://github.com/pybind/pybind11/archive/v2.2.3.tar.gz
+        URL https://github.com/pybind/pybind11/archive/v2.2.4.tar.gz
```
loriab: for consistency, bump it [here](https://github.com/psi4/psi4/pull/1401/files#diff-e8cf7e78ee17d1826f8f9cef1a04c7bfR60), too, please.

psi4/__init__.py
```diff
@@ -100,5 +100,7 @@
 import sys
 if "@ENABLE_PCMSolver@".upper() in ["1", "ON", "YES", "TRUE", "Y"]:
     sys.path.insert(1, "@PCMSolver_PYMOD@")
+if "@ENABLE_cppe@".upper() in ["1", "ON", "YES", "TRUE", "Y"]:
+    sys.path.insert(1, "@cppe_PYMOD@")
```
loriab: this `_PYMOD` variable is provided by you, akin to https://github.com/PCMSolver/pcmsolver/blob/master/cmake/PCMSolverConfig.cmake.in#L143. I don't think that's happening?

psi4/driver/procrouting/solvent/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+# @END LICENSE
+#
+
+"""
+A helper folder for solvent models
+"""
+
+_have_pe = False
+try:
+    from . import pol_embed
```
loriab: won't the `pol_embed` file always be present?

psi4/driver/procrouting/solvent/pol_embed.py
```diff
@@ -0,0 +1,120 @@
...
+import cppe
+
+from psi4 import core
+
+
+def psi4mol_to_cppemol(psi4mol):
+    mol = cppe.Molecule()
+    geom = psi4mol.geometry().np
+    for i, c in enumerate(geom):
+        a = cppe.Atom((int(psi4mol.Z(i))), *c)
```
loriab: be good to try this on something with a ghost atom (`Z(i)` will be 0.0) to make sure nothing breaks and result is sane.

psi4/src/read_options.cc
```diff
@@ -260,6 +260,42 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+    options.add_bool("PE", false);
+    if (name == "PE" || options.read_globals()) {
+        /*- MODULEDESCRIPTION Performs polarizable embedding model (PE) computations. -*/
+
+        /*- Name of the potential file -*/
+        options.add_str_i("POTFILE", "potfile.pot");
+        /*- Use DIIS acceleration to obtain induced moments -*/
+        options.add_bool("DIIS", true);
+        /*- Threshold for induced moments convergence -*/
+        options.add_int("CONVERGENCE_INDUCED", 8);
```
loriab: ok to reverse this to `INDUCED_CONVERGENCE` to be more predictable wrt `D_CONVERGENCE`, etc.?

tests/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ foreach(test_name adc1 adc2 casscf-fzc-sp casscf-semi casscf-sa-sp ao-casscf-sp
                   cc50 cc51 cc52 cc53 cc54 cc55 cc5a cc6 cc7 cc8 cc8a cc8b cc8c
                   cc9 cc9a cdomp2-1 cdomp2-2 cepa0-grad1 cepa0-grad2 cepa1
                   cepa2 cepa3 cepa4 cepa-module ci-multi cisd-h2o+-0 cisd-h2o+-1
-                  cisd-h2o+-2 cisd-h2o-clpse cisd-opt-fd cisd-sp cisd-sp-2
+                  cisd-h2o+-2 cisd-h2o-clpse cisd-opt-fd cisd-sp cisd-sp-2 cppe
```
loriab: ah, this doesn't go here but instead the `cppe/` dir gets added like https://github.com/psi4/psi4/pull/1401/files#diff-3dea97423f83302e647a3bab62bf1157L157

psi4/driver/procrouting/proc.py
```diff
@@ -53,6 +53,7 @@
 from . import dft
 from . import mcscf
 from . import response
+from . import solvent
```
loriab: note for a second review pass: need to find a spot (scf_helper?) where we look if `PE` is on and if so, set c1/noreorient so user doesn't need to handle this.

psi4/driver/procrouting/solvent/__init__.py
maxscheurer: Yes, but inside `pol_embed`, `cppe` is imported. So if the top-level `pol_embed` import fails, PE won't be available.

tests/CMakeLists.txt
maxscheurer: Don't know what line the link refers to :slightly_frowning_face: 
Could you specify a line number?

external/upstream/cppe/CMakeLists.txt
maxscheurer: Won't be an issue I guess. One can always provide a self-built `cppe` via `PYTHONPATH` that is built with OpenMP for the time being.
Will turn it off.

tests/CMakeLists.txt
loriab: https://github.com/psi4/psi4/blob/master/tests/CMakeLists.txt#L155-L161

CMakeLists.txt
```diff
@@ -187,6 +188,7 @@ ExternalProject_Add(psi4-core
            chemps2_external
            dkh_external
            libefp_external
+           cppe_external
```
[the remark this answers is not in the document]
maxscheurer: Done.

external/upstream/cppe/CMakeLists.txt
maxscheurer: Done.

external/upstream/pybind11/CMakeLists.txt
maxscheurer: Done.

psi4/src/read_options.cc
maxscheurer: Done.

tests/CMakeLists.txt
maxscheurer: Done.

psi4/driver/procrouting/proc.py
maxscheurer: I'll try to implement a clean way of doing this... the problem is that PE needs information about the molecule and basis set to construct the `MintsHelper` downstream... So one would have to do it [here](https://github.com/psi4/psi4/blob/083fb200b45aa3e0390b934b23627d83d9abf3bc/psi4/driver/procrouting/proc.py#L1500), where the molecule/basisset are already transformed to C1.

maxscheurer: Okay, I give up on this for today...

psi4/driver/procrouting/solvent/pol_embed.py
maxscheurer: Result looks good with a ghost atom.

psi4/driver/procrouting/proc.py
maxscheurer: I guess the only way to do this is through the user input. The molecule must not be moved or rotated... and there is no way of knowing if PE is enabled at the moment the molecule is constructed. Any suggestions?

loriab: Unless PE uses are peculiar, don’t need to know c1ness at mol creation time, only need to set before computation time. Mol class stores original and applies transformations.

psi4/__init__.py
maxscheurer: Will be done in cppe 0.0.9.

psi4/driver/procrouting/proc.py
maxscheurer: I'm giving up on this... probably you can help, @loriab ☹️ 
I played around with code from [here](https://github.com/psi4/psi4/blob/a8b9036e49d01f385fa3ccfb4b3eb48c13785899/psi4/driver/procrouting/proc.py#L1461), but could not make it work after all...

doc/sphinxman/source/cppe.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Installation
+~~~~~~~~~~~~
+
+**Binary**
+
+* .. image:: https://anaconda.org/psi4/cppe/badges/version.svg
+     :target: https://anaconda.org/psi4/cppe
+
+* CPPE is available as a conda package for Linux and macOS (and Windows, through the Ubuntu shell).
```
dgasmith: Any chance of this compiling on native Windows? We can likely help with that and we might have some appveyor/Azure tips for you for the CI.

maxscheurer: I guess it should compile on native Windows since `cppe` ships with Q-Chem, and Q-Chem is available for Windows... though I have no idea how to setup the CI for Windows. Help would be much appreciated, @dgasmith, probably in a PR for `cppe`?

dgasmith: A good place to start might be [here](https://github.com/MolSSI/cookiecutter-cms/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/appveyor.yml) which is the MolSSI cookie cutter example. This does not include compiled code, but will get you setup with Appveyor. Another example is [gau2grid](https://github.com/dgasmith/gau2grid/blob/master/appveyor.yml) which has some compiled code. The two together should give you enough information. If you can get a PR together I can likely comment on it on CPPE git.

psi4/__init__.py
maxscheurer: Done.

dgasmith:
Ready to go I think, lets get the reviews in.

psi4/driver/procrouting/solvent/pol_embed.py
```diff
@@ -0,0 +1,153 @@
...
+        raise ValidationError("""Error: 3-layer QM/MM/PCM not implemented.\n""")
+    potfile_name = core.get_local_option('PE', 'POTFILE')
+    pol_embed_options = cppe.PeOptions()
+    pol_embed_options.potfile = potfile_name
+    pol_embed_options.induced_thresh = core.get_local_option('PE', 'INDUCED_CONVERGENCE')
+    pol_embed_options.iso_pol = core.get_local_option('PE', 'ISOTROPIC_POL')
+
+    pol_embed_options.do_diis = core.get_local_option('PE', 'DIIS')
+    pol_embed_options.maxiter = core.get_local_option('PE', 'MAXITER')
+    pol_embed_options.pe_border = core.get_local_option('PE', 'BORDER')
```
loriab: I think you want `get_option` (same args) in these lines and below rather than `get_local_option`. Details why upon request, but you're better off not knowing. Any tests that previously worked with `set cppe {...}` should now work the same with `set {...}` (provided the option doesn't overlap and contradict a setting in a different used module). It's just a little more flexible if users aren't required to specify the local scope.

If you have experience that contradicts this, let me know.

psi4/driver/procrouting/solvent/pol_embed.py
```diff
@@ -0,0 +1,153 @@
...
+    def get_pe_contribution(self, density_matrix, elec_only=False):
+        # build electrostatics operator
+        if self.V_es is None and not elec_only:
+            self.build_electrostatics_operator()
+
+        n_bas = self.basisset.nbf()
+        V_pe = np.zeros((n_bas, n_bas))
+        if self._enable_induction:
+            # obtain expectation values of elec. field at polarizable sites
+            elec_fields = self.mints.electric_field_value(
```
loriab: If you're dedicated to this formatting, I suspect you'll want a `# yapf disable` http://psicode.org/psi4manual/master/prog_style_py.html?highlight=yapf to preserve it. Otherwise, you may want to run `yapf` over the code now to better preserve attribution.

psi4/src/read_options.cc
```diff
@@ -260,6 +260,42 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+    options.add_bool("PE", false);
+    if (name == "PE" || options.read_globals()) {
+        /*- MODULEDESCRIPTION Performs polarizable embedding model (PE) computations. -*/
+
+        /*- Name of the potential file -*/
+        options.add_str_i("POTFILE", "potfile.pot");
+        /*- Use DIIS acceleration to obtain induced moments -*/
+        options.add_bool("DIIS", true);
+        /*- Threshold for induced moments convergence -*/
+        options.add_int("INDUCED_CONVERGENCE", 8);
```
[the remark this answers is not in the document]
maxscheurer: Will be changed upstream in `cppe 0.1.1`, which I will push later today and pin in `psi4`.

psi4/driver/procrouting/solvent/pol_embed.py
maxscheurer: I ran `yapf` in the latest commit.

psi4/driver/procrouting/solvent/pol_embed.py
```diff
@@ -0,0 +1,153 @@
...
+import cppe
+from qcelemental import constants
+
+from psi4 import core
+from psi4.driver.p4util.exceptions import ValidationError
+
+
+def get_pe_options():
+    if core.get_option('SCF', 'PCM'):
+        raise ValidationError("""Error: 3-layer QM/MM/PCM not implemented.\n""")
```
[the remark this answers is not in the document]
maxscheurer: Done.

dgasmith:
@loriab can you give this a quick glance?

maxscheurer:
@dgasmith
I need to make some minor changes (see @loriab’s comments) before merge.

dgasmith:
Ah ok, I thought the latest changes had included those. Just eager to get this in :)

maxscheurer:
Me too... I‘ll finish up ASAP (when I get back home tomorrow evening) :smile:

psi4/driver/procrouting/solvent/pol_embed.py
maxscheurer: I'd prefer to have the separate input section. In this manner, the input is similar to the other solvent models, e.g., PCM.

psi4/src/read_options.cc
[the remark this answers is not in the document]
maxscheurer: Done.

maxscheurer:
Ready to go IMHO 😄

maxscheurer:
Now that all the tests pass (travis working again), I think this can be merged. ping @loriab

dgasmith:
@maxscheurer Awesome! Thank you for working to get this PR in. It was a new type of plugin for us, so thank you for bearing with the comments and changes.

psi4/driver/procrouting/solvent/pol_embed.py
loriab: A `set pe { ... }` block will still work as expected if switch to `get_option`. Right now, `set potfile pna_6w.pot` is going to do nothing, and I worry that will be confusing. But if you really want to try it, ok.

psi4/src/read_options.cc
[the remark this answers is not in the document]
maxscheurer: Done.

psi4/driver/procrouting/solvent/pol_embed.py
maxscheurer: Done.

maxscheurer:
I've addressed @loriab's final comments. Once the CI is finished, this can be merged.

dgasmith:
In, we go. Thanks again @maxscheurer!

PR #1403, opened 2018-12-05
dft: "B97" conceded to Grimme

loriab:
## Description
So the reason there haven't been conda packages for a while is `dft1-alt` is non-deterministically broken on py35. Further investigation shows that the `energy('b97-d')` call was sometimes Grimme GGA for XC and sometimes Becke hybrid but always Grimme dispersion (unordered dicts, remember). Probably introduced by #1151 when I stopped listing `b97-d` as an XC of its own and special casing the processing. The proposed sol'n -- conceding "B97" to Grimme and calling the original hyb "B97-0" -- isn't great or historically sound. But I think it's workable.

## Checklist
- [x] a ctest command in py35 that failed 2/6 runs before failed 0/6 runs after

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/dft/libxc_functionals.py
+++ b/psi4/driver/procrouting/dft/libxc_functionals.py
@@ -73,7 +73,7 @@ funcs.append({"name": "B3P86"          , "xc_functionals": {"HYB_GGA_XC_B3P86"  
 funcs.append({"name": "O3LYP"          , "xc_functionals": {"HYB_GGA_XC_O3LYP"          : {}}})
 funcs.append({"name": "mPW1K"          , "xc_functionals": {"HYB_GGA_XC_mPW1K"          : {}}})
 funcs.append({"name": "PBE0"           , "xc_functionals": {"HYB_GGA_XC_PBEH"           : {}}, "alias": ["PBEH"]})
-funcs.append({"name": "B97"            , "xc_functionals": {"HYB_GGA_XC_B97"            : {}}, "alias": ["B97-0"]})
+funcs.append({"name": "B97-0"          , "xc_functionals": {"HYB_GGA_XC_B97"            : {}}})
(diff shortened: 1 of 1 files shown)


PeterKraus:
I'm fine with this. We should mention it in release notes, and perhaps make sure the "Description" and "Citation" elements are filled in properly for both new B97 and B97-0 (not sure they currently are).

loriab:
Definitely agree on RN. I think the metadata are getting filled in ok. Below was how I came to diagnose the problem.

```
  ==> DFT Potential <==								  ==> DFT Potential <==

   => Composite Functional: B97-D2 <= 					  |	   => B97-D2P4 Composite Functional <= 

    B97-d2 Hyb-GGA Exchange-Correlation Functional			  |	    B97-D Pure-GGA Exchange-Correlation Functional

    A. D. Becke, J. Chem. Phys. 107, 8554 (1997)			  |	    S. Grimme, J. Comput. Chem., 27, 1787-1799, 2006

									  >	    Points   =           5000
    Deriv               =              1					    Deriv    =              1
    GGA                 =           TRUE					    GGA      =           TRUE
    Meta                =          FALSE					    Meta     =          FALSE

    Exchange Hybrid     =           TRUE				  |	    X_LRC        =          FALSE
    MP2 Hybrid          =          FALSE				  |	    X_Hybrid     =          FALSE
									  >	    X_Alpha      =   0.000000E+00
```

susilehtola:
Wouldn't it make sense to name the functionals B97-D2 and B97-D2P4 as above?

loriab:
@susilehtola, there's a variety of [aliases for dispersion](https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/intf_dftd3/dashparam.py#L46), so `B97-D` will show up as explicitly `B97-D2` in the output. `D2P4` is an old extension in psi that combined what to run (D2) with what should run it (psi4 internally rather than Grimme `dftd3`). Only reason it appears above is because I was diff-ing against an old reference output.

PR #1409, opened 2018-12-06
Add opt-full-hess-every test only when MAX_AM_ERI is high enough

jturney:
## Description
The default `MAX_AM_ERI` isn't high enough to perform this test.  Add it only when it is configured high enough.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/integral.cc
+++ b/psi4/src/psi4/libmints/integral.cc
@@ -216,7 +216,7 @@ TwoBodyAOInt* IntegralFactory::erd_eri(int deriv, bool use_shell_pairs) {
 #ifdef USING_erd
     if (deriv == 0 && integral_package == "ERD") return new ERDERI(this, deriv, use_shell_pairs);
 #endif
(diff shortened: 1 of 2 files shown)


jturney:
Oh, I see.  Is there another variable I should be checking?

loriab:
`Libint_MAX_AM_ERI` from https://github.com/psi4/psi4/blob/master/external/upstream/libint/CMakeLists.txt#L5 might be available. I have no confidence in my CMake scope predictions. May need to global it.

jturney:
I'll test and commit, if it works.

jturney:
It didn't work. Must be a scope issue. I'm not confident in my CMake abilities to use do the scoping.

loriab:
Can try some of [these solutions](https://stackoverflow.com/questions/19345930/cmake-lost-in-the-concept-of-global-variables-and-parent-scope-or-add-subdirec), but probably not worth it. I wonder if there's a psi4 call that can return Libint's AM (thinking ahead to pytest `@using_am_atleast_6`)? My checkmark still stands as-is.

Issue #1362, opened 2018-11-19
Gaussian Primitive Overlap Screening


zachglick:
As suggested by @andysim on slack, the 2D array of ShellPair objects initialized in eribase.cc (linked below) could be optimized by screening out negligibly small terms. 
https://github.com/psi4/psi4/blob/206420b5c9765140f5ffd86777f694a2547c8002/psi4/src/psi4/libmints/eribase.cc#L1905
The two ideas he proposed are:
1) Shell pair (block) screening: ignore entire shell pairs only when all of their primitive pairs have small enough overlap, otherwise keep the entire shell pair
2) Primitive pair (fine-grained) screening: individually ignore primitive pairs with small enough overlap, allowing part (or all) of a shell pair to be discarded

For both approaches, I counted how many primitive pairs would be removed in this function over a range of systems/basis sets:
![image](https://user-images.githubusercontent.com/16376046/48686241-ee3f3700-eb88-11e8-8e69-04ffabca9b39.png)
Note that this considers a pair of primitives negligible if the overlap term (shown below) is less than a cutoff of 10e-10, which will likely have to be tuned
https://github.com/psi4/psi4/blob/206420b5c9765140f5ffd86777f694a2547c8002/psi4/src/psi4/libmints/eribase.cc#L2026

Block screening would definitely be easier to implement, but it doesn't remove as many primitive pairs as the fine-grained screening. However, the discrepancy between the two methods becomes smaller with both the size of the system and basis set. 

Is this discrepancy small enough to justify block screening, or should we go with fine-grained screening? (Also, should I change up the overlap-cutoff / basis sets / test molecules?)

susilehtola:
Hm, does Psi4 not employ Schwarz screening for the shell pairs? This would be a much better approach, that is also easier to implement.

Namely, the trick is to just calculate the diagonal ERIs ```(ij|ij)``` which only scales as ```N^2```, and then store those shell pairs that satisfy a cutoff ```sqrt(thresh)```.

andysim:
It does for the [direct integrals](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/DirectJK.cc#L228), but not for one-electron integrals, or the disk-based TEIs.  The one-electron integrals being slow was not a problem before we started adding external point charge fields like PCM.

PR #1411, opened 2018-12-06
ShellPair Overhaul and Gaussian Primitive Screening

zachglick:
## Description
Addresses #1362 

This PR will modify the ShellPair struct used throughout the ERI code. This object is an intermediate in integral calculations that describes the overlap between two basis set shells. Each shell pair is composed of gaussian primitives pairs. Many of these primitive pairs don't make a significant contribution to the integral calculations and can be screened via the overlap term.

This PR will also modernize the ShellPair object by replacing C++ arrays with `std::vector` and remove explicit memory-related operations.

## Todos
- [x] Functional `std::vector` based ShellPair
- [x] Primitive Screening in new ShellPair
- [x] Remove old ShellPair struct

## Questions
- [ ] What should we use for the overlap threshold? It's currently set at ~~10e-10~~ 10e-12. This doesn't affect the SCF energy in the handful of molecules I've tested. I may have to investigate other codes.

## Checklist
- [x] [All ~~or relevant fraction of full~~ tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/eri.h
+++ b/psi4/src/psi4/libmints/eri.h
@@ -71,7 +71,7 @@ typedef struct ShellPair_typ {
  * \ingroup MINTS
  * Structure to hold precomputed gaussian product information
  */
-typedef struct PrimPairScreen_typ {
+struct PrimPairScreen {
     //!x, y, z coordinate of gaussian
     double P[3];
     //! Distance between P and shell i center
@@ -86,20 +86,20 @@ typedef struct PrimPairScreen_typ {
     double ci, cj;
     //! Overlap between primitives
     double overlap;
-} PrimPairScreen;
+} ;
 
 /**
  * \ingroup MINTS
  * Structure to hold precomputed shell pair information
  */
-typedef struct ShellPairScreen_typ {
+struct ShellPairScreen {
     //! Shells for this information.
     int i, j;
     //! Distance between shell i and shell j centers
     double AB[3];
     //! Vector of significant primitive pairs that define this shell pair
     std::vector<PrimPairScreen> nonzeroPrimPairs;
-} ShellPairScreen;
+} ;
 
 
 /*! \ingroup MINTS
--- a/psi4/src/psi4/libmints/eri.h
+++ b/psi4/src/psi4/libmints/eri.h
@@ -67,6 +67,41 @@ typedef struct ShellPair_typ {
     double** overlap;
 } ShellPair;
 
+/**
+ * \ingroup MINTS
+ * Structure to hold precomputed gaussian product information
+ */
+typedef struct PrimPairScreen_typ {
+    //!x, y, z coordinate of gaussian
+    double P[3];
+    //! Distance between P and shell i center
+    double PA[3];
+    //! Distance between P and shell j center
+    double PB[3];
+    //! alphas for both centers
+    double ai, aj;
+    //! gamma (ai + aj)
+    double gamma;
+    //! Contraction coefficients
+    double ci, cj;
+    //! Overlap between primitives
+    double overlap;
+} PrimPairScreen;
+
+/**
+ * \ingroup MINTS
+ * Structure to hold precomputed shell pair information
+ */
+typedef struct ShellPairScreen_typ {
+    //! Shells for this information.
+    int i, j;
+    //! Distance between shell i and shell j centers
+    double AB[3];
+    //! Vector of significant primitive pairs that define this shell pair
+    std::vector<PrimPairScreen> nonzeroPrimPairs;
+} ShellPairScreen;
+
+
 /*! \ingroup MINTS
  *  \class ERI
  *  \brief Capable of computing two-electron repulsion integrals.
@@ -109,6 +144,9 @@ class TwoElectronInt : public TwoBodyAOInt {
 
     //! Shell pair information
     ShellPair **pairs12_, **pairs34_;
+    
+    //! Shell pair information, same as **pairs12_ and **pairs34_ but using std::vec instead of new [] for memory
+    std::vector<std::vector<ShellPairScreen>> screenpairs12_, screenpairs34_;
 
     //! Evaluates how much memory (in doubles) is needed to store shell pair data
     size_t memory_to_store_shell_pairs(const std::shared_ptr<BasisSet>&, const std::shared_ptr<BasisSet>&);
--- a/psi4/src/psi4/libmints/eri.h
+++ b/psi4/src/psi4/libmints/eri.h
@@ -145,9 +145,9 @@ class TwoElectronInt : public TwoBodyAOInt {
     //! Shell pair information
     ShellPair **pairs12_, **pairs34_;
     
-    //! Shell pair information, same as **pairs12_ and **pairs34_ but using std::vec instead of new [] for memory
-    std::vector<std::vector<ShellPairScreen>> screenpairs12_, screenpairs34_;
-
+    //! Shell pair information, same as **pairs12_ and **pairs34_ but using std::vec instead of new [] for memory and applies screening
+    std::shared_ptr<std::vector<std::vector<ShellPairScreen>>> screenpairs12_, screenpairs34_;
+   
     //! Evaluates how much memory (in doubles) is needed to store shell pair data
     size_t memory_to_store_shell_pairs(const std::shared_ptr<BasisSet>&, const std::shared_ptr<BasisSet>&);
 
--- a/psi4/src/psi4/libmints/eri.h
+++ b/psi4/src/psi4/libmints/eri.h
@@ -44,35 +44,10 @@ class CorrelationFactor;
 
 /**
  * \ingroup MINTS
- * Structure to hold precomputed shell pair information
- */
-typedef struct ShellPair_typ {
-    //! Shells for this information.
-    int i, j;
-    //! Matrix over primitives with x, y, z coordinate of average Gaussian
-    double*** P;
-    //! Distance between shell i and shell j centers
-    double AB[3];
-    //! Distance between P and shell i center
-    double*** PA;
-    //! Distance between P and shell j center
-    double*** PB;
-    //! Array of alphas for both centers
-    double *ai, *aj;
-    //! Array of the gammas (ai + aj)
-    double** gamma;
-    //! Contraction coefficients
-    double *ci, *cj;
-    //! Overlap between primitives on i and j
-    double** overlap;
-} ShellPair;
-
-/**
- * \ingroup MINTS
  * Structure to hold precomputed gaussian product information
  */
-struct PrimPairScreen {
-    //!x, y, z coordinate of gaussian
+struct PrimPair {
+    //! x, y, z coordinate of gaussian
     double P[3];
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/libmints/eribase.cc
```diff
@@ -1779,6 +1779,104 @@ static size_t fill_primitive_data(prim_data *PrimQuartet, Fjt *fjt, const ShellP
...
     return nprim;
 }
+static size_t fill_primitive_data_screen(prim_data *PrimQuartet, Fjt *fjt, ShellPairScreen sp12, ShellPairScreen sp34, int am,
+                                  bool sh1eqsh2, bool sh3eqsh4, int deriv_lvl) {
+    double zeta, eta, ooze, rho, poz, coef1, PQx, PQy, PQz, PQ2, Wx, Wy, Wz, o12, o34, T, *F;
+    double a1, a2, a3, a4;
+    int p12, p34, i;
+    size_t nprim = 0L;
+    for (p12 = 0; p12 < sp12.nonzeroPrimPairs.size(); ++p12) {
+        PrimPairScreen pp12 = sp12.nonzeroPrimPairs[p12];
```
jturney: I'd make this `const PrimPairScreen& pp12 = ...` to avoid a copy.

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -1779,6 +1779,104 @@ static size_t fill_primitive_data(prim_data *PrimQuartet, Fjt *fjt, const ShellP
...
+        double PAy = pp12.PA[1];
+        double PAz = pp12.PA[2];
+        double PBx = pp12.PB[0];
+        double PBy = pp12.PB[1];
+        double PBz = pp12.PB[2];
+        double PABx = pp12.P[0];
+        double PABy = pp12.P[1];
+        double PABz = pp12.P[2];
+        for (p34 = 0; p34 < sp34.nonzeroPrimPairs.size(); ++p34) {
+            PrimPairScreen pp34 = sp34.nonzeroPrimPairs[p34];
```
jturney: Likewise, I'd make this `const PrimPairScreen& pp34 = ...`

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -1779,6 +1779,104 @@ static size_t fill_primitive_data(prim_data *PrimQuartet, Fjt *fjt, const ShellP
     }
     return nprim;
 }
+static size_t fill_primitive_data_screen(prim_data *PrimQuartet, Fjt *fjt, ShellPairScreen sp12, ShellPairScreen sp34, int am,
```
jturney: `const ShellPairScreen&` to avoid a copy.

psi4/src/psi4/libmints/eri.h
```diff
@@ -67,6 +67,41 @@ typedef struct ShellPair_typ {
...
+    double PB[3];
+    //! alphas for both centers
+    double ai, aj;
+    //! gamma (ai + aj)
+    double gamma;
+    //! Contraction coefficients
+    double ci, cj;
+    //! Overlap between primitives
+    double overlap;
+} PrimPairScreen;
```
andysim: There's nothing wrong with the C style struct here, but I think writing it as 
```
struct PrimPairScreen {
....
};
```
is a little more canonical for our C++ style.

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -1921,6 +2019,19 @@ void TwoElectronInt::init_shell_pairs12() {
...
+    //  Store screened shell pairs in a vector of vectors
+    for (i = 0; i < basis1()->nshell(); ++i) {
+        std::vector<ShellPairScreen> row;
+        screenpairs12_.push_back(row);
+    }
+
+    // Total number of primitive pairs in all shell pairs (before screening) [FOR PRINING]
+    int total_prim_pairs = 0;
+    // Criteria for determining of overlap between primitives is significant for screening [FOR PRINTING]
+    double overlap_cutoff = 1.0e-12;
```
jturney: This is probably something we would allow the user to change in their input.

andysim: +1

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -2024,10 +2143,45 @@ void TwoElectronInt::init_shell_pairs12() {
...
+
+                        pps.ai = a1;
+                        pps.aj = a2;
+                        pps.gamma = gam;
+                        pps.ci = c1;
+                        pps.cj = c2;
+                        pps.overlap = sp->overlap[i][j];
+
+                        //Store this primitive pair in the shell pair
+                        spsc.nonzeroPrimPairs.push_back(pps);
```
jturney: Probably should pre-allocate some space for the vector (you know what the maximum size is going to be).  `push_back` may perform multiple allocations and allocate too much.

psi4/src/psi4/libmints/eribase.cc
andysim: I think (in a later PR) we should set the default to 1e-10, which will require changing some test cases it seems

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -2427,9 +2581,15 @@ size_t TwoElectronInt::compute_quartet(int sh1, int sh2, int sh3, int sh4) {
         // 1234 -> 1234 no change
         p12 = &(pairs12_[sh1][sh2]);
         p34 = &(pairs34_[sh3][sh4]);
+        ShellPairScreen sp12 = screenpairs12_[sh1][sh2];
+        ShellPairScreen sp34 = screenpairs34_[sh3][sh4];
```
jturney: `const ShellPairScreen&` for both of these.

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -2036,6 +2190,7 @@ void TwoElectronInt::init_shell_pairs34() {
     if (use_shell_pairs_ == true) {
         // This assumes init_shell_pairs12 was called and precomputed the values.
         pairs34_ = pairs12_;
+        screenpairs34_ = screenpairs12_;
```
jturney: Is there a way to get around this full copy of `screenpairs12_` to `screenpairs34_`?  The old `pairs12_` and `pairs12_` were pointers and thus free.  The new way could be costly.

psi4/src/psi4/libmints/eribase.cc
```diff
@@ -1921,6 +2019,19 @@ void TwoElectronInt::init_shell_pairs12() {
     pairs12_ = new ShellPair *[basis1()->nshell()];
     for (i = 0; i < basis1()->nshell(); ++i) pairs12_[i] = new ShellPair[basis2()->nshell()];
 
+    //  Store screened shell pairs in a vector of vectors
+    for (i = 0; i < basis1()->nshell(); ++i) {
+        std::vector<ShellPairScreen> row;
+        screenpairs12_.push_back(row);
+    }
+
```
jturney: I think this entire for loop can be replaced with: `screenpairs12_.resize(basis1()->nshell());`

psi4/src/psi4/libmints/eribase.cc
andysim: They could both be defined as shared_ptrs that point to the same object in this case, right?

psi4/src/psi4/libmints/eribase.cc
zachglick: Thanks!

psi4/src/psi4/libmints/eri.h
zachglick: Got it, thanks!

psi4/src/psi4/libmints/eribase.cc
zachglick: That's a good idea. I changed the `nonzeroPrimPairs` vector so that memory is reserved for `np_i*np_j` elements (the maximum number of primitive pairs). However, this still uses `push_back`. 

I'm a little hesitant to resize the vector to its maximum length, since  screening will ideally make the vector significantly shorter. Do you think that's okay, or should we use `resize`?

jturney: You should be able to shrink it down to what it needs to be afterwards.

psi4/src/psi4/libmints/eribase.cc
zachglick: I initially set the cutoff to 10e-10, but a few of the quick tests failed when checking SCF energy to 10 digits

psi4/src/psi4/libmints/eribase.cc
zachglick: Definitely, I replaced the loop with a single initialization statement that creates `screenpairs12_` with the correct number of elements

psi4/src/psi4/libmints/eribase.cc
andysim: Yes, defining them as shared pointers to some object that holds the list of shell pairs (similar to how the vector does now) should do it, i think.  Looking great so far!

psi4/src/psi4/libmints/eribase.cc
zachglick: It looks like there's already something like this already,:
https://github.com/psi4/psi4/blob/637dbab34d0313a8cc81ea2d4c3740d07ff54328/psi4/src/read_options.cc#L1231
Although it's not clear to me how `INTS_TOLERANCE` should be used for screening

psi4/src/psi4/libmints/eribase.cc
zachglick: Thanks! I've changed `pairs12_` to be a pointer to a vector of vectors. The initialization is pretty ugly, but it works:
    ```pairs12_ = std::make_shared<std::vector<std::vector<ShellPair>>>(std::vector<std::vector<ShellPair>>(basis2()->nshell(), std::vector<ShellPair>(basis2()->nshell())));```
Elements of the object are accessed by doing:
```(*pairs12_)[i][j]```
Are you all ok with this? I could alternatively put the vector of vectors in another struct. This would allow me to use `->` instead of `*` (which seems more in line with Psi4's style). The initialization of `pairs12_` could also be made to look a little cleaner.
Let me know what you all think!

psi4/src/psi4/libmints/eribase.cc
zachglick: That's perfect!
The `shrink_to_fit()` function reduces the capacity of the vector to match its size. Right now, I'm doing the following:

1. Make vector and resize to maximum size
2. Fill vector with primitive pairs via `[]`
3. Resize to final size (less than maximum because of screening)
4. Shrink to fit

Is this any better than doing the following, since `push_back()` is amortized constant time?

1. Make empty vector
2. Fill vector with primitive pairs via `push_back`
3. Shrink to fit

hokru:
Happy this is getting in :-)

andysim:
Replaced with #1721

PR #1412, opened 2018-12-07
Fix bug that prevents setting a local string/int option from the py-side if module != current_module

fevangelista:
## Description
Small bug fix that prevents adding local options to a module. This error shows only for `string` and `int` type options, but I potentially also in `array` and `python` data types. This PR addresses all four cases.

## Todos
- [x] Fix option reading bug

## Questions
- In the following part of the code one can potentially encounter the same problem. This can be fixed by adding `module` to the list of parameters passed to this function. Any reason why there is no `std::string module` in the parameter list? Adding it would also make this function consistent with other `set_local_x` functions.
https://github.com/psi4/psi4/blob/396b4c51d6902301baeed41386c49ccb9099a30b/psi4/src/core.cc#L705-L707

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -614,10 +614,14 @@ bool py_psi_set_global_option_double(std::string const& key, double value) {
 bool py_psi_set_local_option_array(std::string const& module, std::string const& key, const py::list& values,
                                    DataType* entry = nullptr) {
     std::string nonconst_key = to_upper(key);
+
     // Assign a new head entry on the first time around only
     if (entry == nullptr) {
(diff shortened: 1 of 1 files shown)


fevangelista:
@loriab: Yes, the state must change somewhere and the error comes when the input is parsed (say after running SCF). My solution is to temporarily switch to the module passes by the user and then go back to the one set before call these functions. In the future, I would like to see if we could find a way to rely less on the set_/get_module. I see the benefits, but it makes the options model quite obscure.

PR #1382, opened 2018-11-28
Add analytic dipole derivatives for RHF

andysim:
## Description
Adds analytic dipole derivatives for RHF, allowing IR intensities to be computed; @loriab, @dsirianni and crew are working on the finite difference version for cases like DF-MP2, where finite differences of gradients are used for frequencies.

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] Adds dipole derivative code, and a test to verify that they match the CCQC programming project values

## Questions
- [ ] Does anybody object to wavefunction holding these?
- [ ] Does anybody object to the various `set_` methods of wavefunction taking SharedMatrix by value, not by reference?

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1414, opened 2018-12-07
parallelize longest ctests

loriab:
## Description
move dft-bench and vibanalysis tests to pytest

## Todos
- [x] move the 4 dft-bench tests to pytest with parameterization
- [x] move the vibanalysis test to pytest
- [x] add IR intensity benchmarks from Cfour so #1382 has something to test against
- [ ] delete the ctest versions
- [ ] add in `pytest.mark.long` to avoid running some bench routinely

## Questions
- [ ] Question1

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


tests/pytest/test_dft_benchmarks.py
```diff
@@ -0,0 +1,347 @@
...
+
+    psi4.set_options({
+       'dft_radial_points': 200,
+       'dft_spherical_points': 590,
+       'guess': 'sad',
+       'e_convergence': 9,
+       'd_convergence': 9,
+    })
+
+    return {'h2o': h2o, 'h2o_plus': h2o_plus, 'h2o_dimer': h2o_dimer}
```
dgasmith: Maybe `yield` these then set the above `set_options` as an opt stash to be cleared after the yield returns?

tests/pytest/test_dft_benchmarks.py
```diff
@@ -0,0 +1,347 @@
...
+    pytest.param(          'PBE50',    0.4509652861, '6-31G'),  # Q-Chem
+    pytest.param(            'N12',    0.4479283792, '6-31G'),  # Q-Chem
+    pytest.param(           'VSXC',    0.4547894146, '6-31G'),  # Q-Chem
+    pytest.param(      'CAM-B3LYP',    0.4568003604, '6-31G'),  # Q-Chem
+    pytest.param(           'VV10',    0.4551366594, '6-31G'),  # Q-Chem
+    pytest.param(         'B97M-V',    0.4561660762, '6-31G'),  # Q-Chem
+    pytest.param(        'LC-VV10',    0.4556872545, '6-31G'),  # Q-Chem
+    pytest.param(        'wB97M-V',    0.4544676075, '6-31G'),  # Q-Chem
+    pytest.param(        'wB97X-V',    0.455302602,  '6-31G'),  # Q-Chem
+], ids=name_dft_test)
```
dgasmith: `# yap: disable`

tests/pytest/test_dft_benchmarks.py
loriab: Not quite following. All options get cleared btwn tests via conftest and a fixture gets rerun for every parametrize, so what aspect are you seeking to change?

dgasmith: Oh, if options are always nuked between tests then no comment from me. I was thinking about manually handling that via fixture cleanups.

loriab: yup, https://github.com/psi4/psi4/blob/master/tests/pytest/conftest.py#L9-L14. iirc, `clean_options` was invented for pytest, as we were leaking options madly.

loriab:
just a note that two VV10 dft-bench-ionization tests take 10 min to run on azure, but the whole pytest suite takes 10 min on travis, so win build is wip. removed these tests from "quick" for now.

loriab:
@hokru, fyi, the tests you added in #1320 got moved around here. Because they use dftd3, which isn't always available, they need `using_dftd3` decorator (for pytest) or to be in `tests/dftd3/` (for ctest).

PR #1421, opened 2018-12-10
QCSchema update and SCF properties

dgasmith:
## Description
Updates the SCF one-electron properties computation to allow users to specify which properties should be computed. This applies to all computations besides property computations.

## Todos
- [x] Allows arbitrary one-electron properties to be computed upon SCF energy calls.
- [x] Moves QCSchema to a more flexible metadata approach, this section of code will certainly evolve over time.
- [x] Removes Psi4 v1.1 JSON specification for 1.3 as per the deprecation warning.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/json_wrapper.py
```diff
@@ -52,11 +55,126 @@
...
+    # SCF variables
+    "scf": {
+        "scf_one_electron_energy": {"variables": "ONE-ELECTRON ENERGY"},
+        "scf_two_electron_energy": {"variables": "TWO-ELECTRON ENERGY"},
+        "scf_dipole_moment": {"variables": ["SCF DIPOLE X", "SCF DIPOLE Y", "SCF DIPOLE Z"]},
+        "scf_iterations": {"variables": "SCF ITERATIONS", "cast": int},
+        "scf_total_energy": {"variables": "SCF TOTAL ENERGY"},
+        "scf_vv10_energy": {"variables": "DFT VV10 ENERGY", "skip_zero": True},
+        "scf_xc_energy": {"variables": "DFT XC ENERGY", "skip_zero": True},
+        "scf_dispersion_correction_energy": {"variables": "DISPERSION CORRECTION ENERGY", "skip_zero": True},
```
loriab: I like this systematic translation layer.

psi4/driver/json_wrapper.py
```diff
@@ -52,11 +55,126 @@
...
+    },
+
+    # MP2 variables
+    "mp2": {
+        "mp2_same_spin_correlation_energy": {"variables": "MP2 SAME-SPIN CORRELATION ENERGY"},
+        "mp2_opposite_spin_correlation_energy": {"variables": "MP2 OPPOSITE-SPIN CORRELATION ENERGY"},
+        "mp2_singles_energy": {"variables": "NYI", "default": 0.0},
+        "mp2_doubles_energy": {"variables": "MP2 CORRELATION ENERGY"},
+        "mp2_total_correlation_energy": {"variables": "MP2 CORRELATION ENERGY"},
+        "mp2_total_energy": {"variables": "MP2 TOTAL ENERGY"},
```
loriab: This mp2 section currently runs afoul of OCC, as I noticed when running CBS through json b/c some variables not set. I think this'll be fixed by another hunk of #1381, so just a heads-up.

psi4/driver/json_wrapper.py
```diff
@@ -52,11 +55,126 @@
...
+        if (value is None) and ("default" in var):
+            value = var["default"]
+
+        # Cast if called
+        if "cast" in var:
+            value = var["cast"](value)
+
+        ret[key] = _json_translation(value)
+
+    return ret
```
loriab: nice. will have to eventually hook it into the [validator](https://github.com/qcdb/qcdb/blob/master/qcdb/qcvars/psivardefs.py) and glossary that ensures psivars are consistent and meet normal QC definitions (ref + corl = total; same-spin + opposite-spin + singles = total).

psi4/driver/json_wrapper.py
```diff
@@ -52,11 +55,126 @@
...
+    if isinstance(value, (psi4.core.Matrix, psi4.core.Vector)):
+        value = value.np.ravel().tolist()
+    elif isinstance(value, (psi4.core.Dimension)):
+        value = value.to_tuple()
+    elif isinstance(value, np.ndarray):
+        value = value.ravel().tolist()
+
+    return value
+
+def _convert_variables(data, include=None):
```
loriab: maybe `required` instead of `include`

psi4/driver/json_wrapper.py
dgasmith: That is mostly my thought. We slowly push all of this into metadata so that we can write other manipulators and validators with ease.

psi4/driver/json_wrapper.py
dgasmith: Trying out `context`.

psi4/driver/procrouting/proc.py
```diff
@@ -1377,11 +1377,21 @@ def scf_helper(name, post_scf=True, **kwargs):
...
-    # We always would like to print a little dipole information
-    if kwargs.get('scf_do_dipole', True):
+    # We always would like to print a little property information
+    if kwargs.get('scf_do_properties', True):
         oeprop = core.OEProp(scf_wfn)
         oeprop.set_title("SCF")
-        oeprop.add("DIPOLE")
+
+        # Figure our properties, if empty do dipole
+        props = [x.upper() for x in core.get_option("SCF", "SCF_PROPERTIES")]
```
andysim: My knowledge of this machinery is a little outdated.  Is there a reason to do this via options, rather than just pass the `props` array directly in place of the `scf_do_properties` bool?

Issue #1296, opened 2018-10-18
Findif incompatible with full_hess_every


JonathonMisiewicz:
I wondered how the pre-#1289 driver handled core cleaning when `optimize` called for a `hessian`. I discovered the handling was bugged. The following input file crashes when I try to run it on my control version of Psi, 287272b. The initial Hessian is fine because `GUESS` hasn't yet been set to `READ`. (I assume that happens somewhere in OPTKING.) After the first step, `GUESS` is set to read, causing the seventh displacement for the hessian to crash with `ValidationError: Cannot compute projection of different symmetries.` I had been under the impression that https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L1454-L1455 existed precisely to avoid such things, so maybe those lines do something else?

```
molecule h2o { 
     O    
     H 1 1.0 
     H 1 1.0 2 104.5 
}

set {
  basis sto-3g
  e_convergence 10
  d_convergence 10
  points 3
  geom_maxiter 7
  full_hess_every 1 
}

optimize('scf', dertype = 0)
```

EDIT: The conditions are more general than originally reported. Any findif hessian with `GUESS` set to `READ` will trigger the bug. The hessians computed by `full_hess_every` are just one particular instance.

JonathonMisiewicz:
Per Slack discussion: Let's just disable writing orbitals from displaced geometries. We trust the undisplaced orbitals more for `GUESS=READ` purposes, which is the only reason we need to touch file 180, so this will give us better orbitals as well. Will be included in #1289, because this looks like a job for `_process_displacement`.

PR #1424, opened 2018-12-11
Symmetry-Breaking Findif, and Reference First

### LINKED ISSUE #1052: Displacements Breaking Explicit Symmetry Crash Psi
If a user sets the symmetry explicitly and then attempts to perform a finite difference calculation, Psi raises `Fatal Error: User specified point group is not a subgroup of the highest detected point group.` This problem does not occur if the user does not explicitly provide the symmetry. The problem also occurs if `reset_point_group` is used. Example input below.

I don't know enough about Psi's symmetry and molecule tech to tell if this is a problem that needs to be fixed findif side or molecule side. If this needs to be fixed findif side, I'll take it for obvious reasons. If not, then if this isn't fixed by the time findif is done, I'll take this anyways. I've had a vendetta against this bug since November.

```
molecule h2o {
  symmetry c2v
  O
  H 1 0.9894093
  H 1 0.9894093 2 100.02688
}

set {
  basis sto-3g
  d_convergence 11
  scf_type pk
}

frequencies('scf', dertype=1, return_wfn=True)
```

JonathonMisiewicz:
## Description
Fixes #1052 at long, long last: when the user explicitly declares a symmetry, findif is now smart enough to put displacements in the proper subgroup.

As part of the fix, control over molecule cloning for finite difference has passed to `_process_displacement`. It seemed easier (and more appropriate) than adding code to make a new clone in each for loop. The clone would be necessary so that each displacement can have its own symmetry.

Since I was touching that part of the code anyways, I have reorganized so that the reference is guaranteed to be the first displacement. This has the advantage that if orbital reading is enabled, the following displacements will be able to read from the reference, as intended.

## Todos
- [x] Fixes #1052
- [x] Exposes more C functions to Python
- [x] Move the reference geometry first for ease of orbital reads

## Checklist
- [x] Tests modified to account for bug fix
- [x] All findif tests passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -140,7 +140,7 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
        method : str
           A string specifying the method to be used for the computation.
        molecule: psi4.core.molecule or qcdb.molecule
-          The molecule for the computation. Geometry setting is handled internally.
+          The molecule for the computation. All processing is handled internally.
        displacement : dict
           A dictionary containing the necessary information for the displacement.
           See driver_findif/_geom_generator.py docstring for details.
@@ -161,15 +161,25 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
     print(""" %d""" % (n), end=('\n' if (n == ndisp) else ''))
     sys.stdout.flush()
 
+    parent_group = molecule.point_group()
+    clone = molecule.clone()
+    clone.reinterpret_coordentry(False)
+    clone.fix_orientation(True)
+
     # Load in displacement (flat list) into the active molecule
     geom_array = np.reshape(displacement["geometry"], (-1, 3))
-    molecule.set_geometry(core.Matrix.from_array(geom_array))
+    clone.set_geometry(core.Matrix.from_array(geom_array))
 
+    # If we lost symmetry, use the highest subgroup of the parent this displacement has..
+    disp_group = clone.find_highest_point_group()
+    new_bits = parent_group.bits() & disp_group.bits()
+    new_symm_string = qcdb.PointGroup.bits_to_full_name(new_bits)
+    clone.reset_point_group(new_symm_string)
     # clean possibly necessary for n=1 if its irrep (unsorted in displacement list) different from initial G0 for freq
     core.clean()
 
     # Perform the derivative calculation
-    derivative, wfn = derivfunc(method, return_wfn=True, molecule=molecule, **kwargs)
+    derivative, wfn = derivfunc(method, return_wfn=True, molecule=clone, **kwargs)
     displacement["energy"] = core.variable('CURRENT ENERGY')
 
     # If we computed a first or higher order derivative, set it.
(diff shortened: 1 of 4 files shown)


psi4/driver/driver.py
```diff
@@ -1269,13 +1275,8 @@ def hessian(name, **kwargs):
...
             """hessian() will perform frequency computation by finite difference of analytic gradients.\n""")
 
-        # Shifting the geometry so need to copy the active molecule
-        moleculeclone = molecule.clone()
-
         # Obtain list of displacements
-        findif_meta_dict = driver_findif.hessian_from_gradient_geometries(moleculeclone, irrep)
-        moleculeclone.reinterpret_coordentry(False)
-        moleculeclone.fix_orientation(True)
+        findif_meta_dict = driver_findif.hessian_from_gradient_geometries(molecule, irrep)
```
dgasmith: Should this be `clone` here?

andysim:
This looks great!  Thanks for figuring out this very tricky problem.  A long time ago, @fevangelista and I figured out the machinery to map symmetry-dependent arrays like `DOCC` and `SOCC` into the appropriate subgroups.  Would you mind quickly testing to see if `DOCC` can be specified in c2v symmetry with your current version?  If not, we'll have to open a follow up PR to make sure this mapping is done (I think it's best left out of this PR to avoid scope creep).

psi4/driver/driver.py
```diff
@@ -689,23 +699,19 @@ def gradient(name, **kwargs):
...
 
-        # This version is pretty dependent on the reference geometry being last (as it is now)
         print(""" %d displacements needed ...""" % (ndisp), end='')
 
-        for n, displacement in enumerate(findif_meta_dict["displacements"].values(), start=1):
-            _process_displacement(
-                energy, lowername, moleculeclone, displacement, n, ndisp, write_orbitals=False, **kwargs)
-
-        wfn = _process_displacement(energy, lowername, moleculeclone, findif_meta_dict["reference"], ndisp, ndisp,
+        wfn = _process_displacement(energy, lowername, molecule, findif_meta_dict["reference"], 1, ndisp,
```
dgasmith: Can you elaborate on the new looping structure here?

psi4/driver/driver.py
JonathonMisiewicz: No.

None of the geometry generators change the molecule, so there's no need to have a clone for purposes of generating the metadict. Cloning is only needed for the computations and is now done in `_process_displacement`, because different displacements need different clones for symmetry purposes.

psi4/driver/driver.py
JonathonMisiewicz: This isn't a new looping structure, just a change in the order in which we compute displacements. The reference displacement, previously done last, is now moved to be first. The argument to `_process_displacement` specifying the displacement number for print purposes has been changed accordingly.

The point of the change is so the reference is guaranteed to be available for use as an orbital guess, per https://github.com/psi4/psi4/issues/1296#issuecomment-431018568.

JonathonMisiewicz:
@andysim I checked DOCC compatibility just before I submitted this, and it worked. I'll add a simple DOCC set to the findif test when I fix whatever is causing tests to fail.

It'll likely be a few days before I have the time to get those fixed.

JonathonMisiewicz:
* Fixed mp2-module failure. Changed the symmetry reset to only occur if the user provided a default symmetry. My changes overrode Bozkaya's DF code running computations in C1, and the computational symmetry change resulted in BH3 landing on different UHF solutions.
* Fixed pywrap-all failure. Per discussion in Slack, findif now saves and resets the core variables from the reference, after all displacements are finished.
* Addressed concerns of @andysim. I now set DOCC explicitly for my test case that you can set symmetry.

JonathonMisiewicz:
I know the code used `P::e` before I touched it, but I don't think any of the code I added uses `P::e`. If that's not the case, let me know, and I'll see if I can remove the offending code, to help eliminate globals.

I'm holding off on documentation update until I know if I need to make other changes while I'm at it.

loriab:
The driver portion you're touching with findif is reactionary. It can't modernize until the c-side and proc.py improves. So you're good.

JonathonMisiewicz:
And, warning against modifying molecule in `_process_displacement` added.

PR #1443, opened 2018-12-21
[WIP] Moving to xtensor

robertodr:
## Description
It was agreed at PsiCon that the `Matrix` and `Vector` class be rewritten to use xtensor as storage backend. This is my work on this so far. It is not complete - indeed `Matrix` and `Vector` are still used everywhere - but I think it's better to put this up early because it will take time to transition to the new classes. **We do need a strategy to orchestrate the transition**

## Todos
- [x] Use `_a` literal in pybind11 export sources rather that `py::arg`
- [x] Basic templated `Tensor` class with Python bindings.
- [x] `doublet` for `Tensor<double, 2>` aka a `Matrix` of `double`-s working.

## Questions
- With @amjames we discussed that it might be better to have linear algebra operations as free functions (BLAS/LAPACK- and NumPy-like) Should an `LA` namespace be introduced? Possibly also create a Python submodule with pybind11?
- The Python bindings for the `Tensor` class allow to do: `a_tensor[h]` to get a block (binding `operator[]` to `__getitem__`, but not `a_tensor[h] = np.random.randn(*a_tensor[h].shape)`. For the latter I tried binding `set_block` to `__setitem__` but I get horrible, horrible compiler errors. I think this should be made to work to make the Python interface feel as native as possible.
- As a quick way to get `xtensor-python` up and running, I've copied their `FindNumPy.cmake` file. @loriab your input on this and how to make avoid duplication with `FindPythonModule.cmake` wanted.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


fevangelista:
@robertodr: this looks to me like two PRs: i) major improvements to the Vector/Matrix classes (very happy to see those) and ii) introducing xtensor. I just gave it a quick look but I was wondering: Do you actually use xtensor anywhere in the code? Before we go with this major overhaul I think it would be important to show what we could do with it.

cmake/FindNumPy.cmake
```diff
@@ -0,0 +1,89 @@
...
+# OTHER DEALINGS IN THE SOFTWARE.
+#
+#============================================================================
+
+# Finding NumPy involves calling the Python interpreter
+if(NumPy_FIND_REQUIRED)
+    find_package(PythonInterp REQUIRED)
+else()
+    find_package(PythonInterp)
+endif()
```
loriab: I've no particular grievance with this file, but could `find_python_module` be used instead? If it's just `NUMPY_INCLUDE_DIRS`, I'd slightly prefer to avoid calling `find_package(PythonInterp)`, and do build up a proper target with the numpy include dirs on it.

psi4/src/export_fock.cc
```diff
@@ -155,8 +157,7 @@ void export_fock(py::module &m) {
         .def("add_space", &DFHelper::add_space)
         .def("initialize", &DFHelper::initialize)
         .def("print_header", &DFHelper::print_header)
-        .def("add_transformation", &DFHelper::add_transformation, py::arg("name"), py::arg("key1"), py::arg("key2"),
-             py::arg("order") = "Qpq")
+        .def("add_transformation", &DFHelper::add_transformation, "name"_a, "key1"_a, "key2"_a, "order"_a = "Qpq")
```
loriab: Having done some I'm-really-trying-to-add-a-feature-but-ended-up-modernizing-cleaning-formatting-fixing-along-the-way projects of late, I agree with @fevangelista that it's a good idea to `git reset` and do a PR of the uncontroversial changes like `py::arg` --> `_a` and de-c++11 so that those go in sooner and the foundations for your bigger changes aren't shifting.

jturney:
Great progress!

I agree with @fevangelista and @loriab about splitting this into multiple PRs. Certainly the updating pybind from `py::arg` to `_a` is one. 

I'm on board with create a namespace of linear algebra functions. I'd preferred it named `la` lowercase though, or follow numpy's lead and use `linalg`. I'm assuming the `doublet, triplet, horzcat, create`, etc would all go into there.

psi4/CMakeLists.txt
```diff
@@ -146,10 +146,36 @@ else()
 endif()
 
 find_package(Libxc 4.0.2 CONFIG REQUIRED)
-get_property(_loc TARGET Libxc::xc PROPERTY LOCATION)
+get_property(_loc TARGET Libxc::xc PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
```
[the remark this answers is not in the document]
robertodr: I don't remember why I did this, actually...

robertodr:
Thanks all for the feedback. I'll split this one in three: i) Vector/Matrix polish ii) C++14 constructs for pybind11 iii) Actual xtensor stuff (CI doesn't seem very happy about those :weary:) 

@fevangelista xtensor is used only in the test for the `doublet`. I initially intended to swap the storage in `Vector` and `Matrix` to use xtensor, but that turned out to be quite hard (due to the extensive use of `pointer` basically everywhere in the code) The strategy now is to rewrite the storage object (I think that's largely done, though copy CTORs, assignment etc are missing) and the symmetry-aware operations (multiplies, diagonalization, etc) unit testing them on the way. The latter will take a bit of time, but it's also something that more people can work on concurrently, I think.

fevangelista:
@robertodr: I assume that the only reason no to use straightforward xtensor objects is symmetry, no? It might be easier to transition if we started to use xtensor and slowly deprecated Matrix and Vector out of the code (if even possible).

robertodr:
I've removed the work-in-progress marker as this first step is reasonably complete. The `Tensor` class as storage object has an acceptable interface both on the C++ and the Python side (I think!) I can use this to work further on the `linalg` namespace, add BLAS/LAPACK wrappers, and eventually transition the SCF to xtensor.

I have tried to fix Azure CI, to no avail. I can't quite figure out whether I need to change the xtensor source or the CI setup.

robertodr:
**Update on the Azure CI failures**
The problem seems _not_ to be entirely controllable through flags. Parts of the xtensor stack code base (xtensor itself, xsimd, possibly xtensor-python) need to be changed and I am working on it. Once I have PR submitted to all these projects, they are hopefully merged and new releases minted.

jturney:
This is looking really nice.

robertodr:
Sadly, I can't get it to build on Azure :cry:

cmake/FindNumPy.cmake
```diff
@@ -0,0 +1,80 @@
...
+      "(${NUMPY_VERSION_MAJOR} * 10000) + (${NUMPY_VERSION_MINOR} * 100) + ${NUMPY_VERSION_PATCH}")
+
+  add_library(NumPy INTERFACE)
+  target_include_directories(NumPy
+    INTERFACE
+      ${NUMPY_INCLUDE_DIRS}
+    )
+
+  set(NUMPY_FOUND TRUE)
+endif()
```
loriab: so will psi now be built against a specific numpy API (akin to building against a specific python API presently)? looking at https://anaconda.org/conda-forge/xtensor-python/files it seems not, as there's one for ancient numpy and one for everything >=1.11. just checking.

robertodr: No, I am not asking for a specific numpy version when doing `find_package` (but we could, if necessary) And as you said, xtensor-pyhon doesn't force one specific version of numpy.

external/upstream/xtl/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+else()
+  if(${CMAKE_INSIST_FIND_PACKAGE_xtl})
+    message(FATAL_ERROR "Suitable xtl could not be externally located as user insists")
+  endif()
+
+  include(ExternalProject)
+  message(STATUS "Suitable xtl could not be located, ${Magenta}Building xtl${ColourReset} instead.")
+  ExternalProject_Add(xtl_external
+    URL
+      https://github.com/QuantStack/xtl/archive/0.6.4.zip
```
loriab: 0.6.4/5 mismatch?

external/upstream/xtensor/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+  endif()
+
+  include(ExternalProject)
+  message(STATUS "Suitable xtensor could not be located, ${Magenta}Building xtensor${ColourReset} instead.")
+  ExternalProject_Add(xtensor_external
+    DEPENDS
+      xtl_external
+      xsimd_external
+    URL
+      https://github.com/QuantStack/xtensor/archive/0.20.7.zip
```
loriab: mismatch here, too

psi4/CMakeLists.txt
```diff
@@ -157,6 +157,32 @@ get_property(_loc TARGET Libxc::xc PROPERTY LOCATION)
...
+message(STATUS "${Cyan}Using xtensor${ColourReset}: ${_loc} (version ${xtensor_VERSION})")
+
+find_package(xtensor-blas 0.16.1 CONFIG REQUIRED)
+message(STATUS "${Cyan}Using xtensor-blas${ColourReset}: ${_loc} (version ${xtensor-blas_VERSION})")
+
+find_package(xtensor-python 0.23.1 CONFIG REQUIRED)
+get_property(_loc TARGET xtensor-python PROPERTY INTERFACE_INCLUDE_DIRECTORIES)
+list(APPEND _addons ${_loc})
+message(STATUS "${Cyan}Using xtensor-python${ColourReset}: ${_loc} (version ${xtensor-python_VERSION})")
+find_package(NumPy REQUIRED)
```
loriab: may want to double-check versions in this block, too

external/upstream/xtl/CMakeLists.txt
robertodr: Yup, thanks for catching it.

psi4/src/psi4/libmints/linalg.h
```diff
@@ -0,0 +1,494 @@
...
+            ax_B.push_back(1);
+            label += B->label() + "H";
+            break;
+    }
+
+    for (size_t hA = 0; hA < A->nirrep(); ++hA) {
+        size_t hB = hA ^ (transA ? 0 : A->symmetry()) ^ (transB ? B->symmetry() : 0);
+        size_t hC = hA ^ (transA ? A->symmetry() : 0);
+
+        // FIXME Currently NOT handling transpose-conjugate case
```
loriab: should there be a NYI error somewhere for TransposeConj?

robertodr: Sounds good! I should probably also limit scalar types to non-complex.

robertodr: I have added support for `std::complex<double>` and the function can now handle transpose conjugation. The tests issue a warning, see MolSSI/QCElemental#77

psi4/src/psi4/libmints/linalg.h
```diff
@@ -0,0 +1,494 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2018 The Psi4 Developers.
```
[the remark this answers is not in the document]
robertodr: Done, thanks for catching this.

psi4/CMakeLists.txt
robertodr: Done.

Issue #1431, opened 2018-12-17
FCIDUMP format


SebWouters:
Dear developers,

I've received an issue from a user of chemps2. I traced it back to incompatible fcidump and molden files. In particular, the number of orbitals per irrep as indicated in "ORBSYM" of the FCIDUMP file and as written per orbital in the the molden file were different. I might be mistaken, but seems that you write out a symmetry in psi4 numbering convention plus one, from the line:

https://github.com/psi4/psi4/blob/c0b8eb86430673c9e09a180dfe91bf6980a1f9c1/psi4/driver/p4util/fcidump.py#L99

This is not the convention for FCIDUMP. The irrep numbering should be the one originally used in molcas. For d2h, for example, this should be:

    psi4 irrep "0" = Ag  = molcas irrep "1";
    psi4 irrep "1" = B1g = molcas irrep "4";
    psi4 irrep "2" = B2g = molcas irrep "6";
    psi4 irrep "3" = B3g = molcas irrep "7";
    psi4 irrep "4" = Au  = molcas irrep "8";
    psi4 irrep "5" = B1u = molcas irrep "5";
    psi4 irrep "6" = B2u = molcas irrep "3";
    psi4 irrep "7" = B3u = molcas irrep "2";

For a complete mapping of all abelian real-valued point group irreps, see e.g.

https://github.com/SebWouters/CheMPS2/blob/master/CheMPS2/Irreps.cpp#L184

Thanks!

Sebastian

dgasmith:
@andysim I believe this is correct, do you concur?

@SebWouters You seem to have a pretty good handle on exactly where this error occurs, would you mind making a PR to patch this up?

SebWouters:
@dgasmith 
@andysim 
@sunqm

I saw it occurs in 

    psi4/psi4/driver/p4util/fcidump.py

which may be fixed via

    def fcidump_helper(wfn):
    
        nirrep = wfn.nirrep()
        symm   = wfn->molecule().sym_label()
        psi2dump = np.zeros([nirrep], dtype='int')
    
        if ( 'c1' == symm ):
            psi2dump[0] = 1    # A
        if ( 'ci' == symm ):
            psi2dump[0] = 1    # Ag
            psi2dump[1] = 2    # Au
        if ( 'c2' == symm ):
            psi2dump[0] = 1    # A
            psi2dump[1] = 2    # B
        if ( 'cs' == symm ):
            psi2dump[0] = 1    # A'
[...]

PR #1453, opened 2019-01-03
Fix FCIDUMP irrep ordering convention

zachglick:
## Description
Addresses #1431 

## Todos
- [x] Changes ORBSYM (the symmetry of each orbital) from Psi4's internal convention of ordering irreps to the traditional FCIDUMP ordering
- [x] Adds ISYM (the symmetry of the overall wavefunction)
- [x] Updates the `fcidump` test to match new ordering

## Questions
- [x] Does this change in convention break anything in Psi4? (Or anything that uses Psi4?)
- [x] Since I altered the tests to match the output, how to verify correctness of this change?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


CDSherrill:
Wfn irrep logic for socc not correct... (Around line 91).  Need to take
direct product of each socc orbital irrep.  Can do this by XOR of each
irrep in the socc list.

zachglick:
@robertodr and @andysim 
I could certainly make a PsiAPI test for energies, although I think the leading underscore of `_energies_from_fcidump` will have to be removed, since it denotes that the function is private.

@CDSherrill 
I intended for this bit of code to do exactly what you describe:

``` wfn_irrep = 0
    for h, n_socc in enumerate(active_socc):
        if n_socc % 2 == 1:
            wfn_irrep ^= h
```

The line `wfn_irrep ^= h` accumulates the XOR of all irreps `h` in `wfn_irrep`. I could rewrite it as `wfn_irrep = wfn_irrep^h` to make it more intelligible.

I think that the format of `active_socc` is not a list of irreps, instead it's a list of orbital counts per irrep, which is why I XOR `h` and not `n_socc`. The `if n_socc % 2 == 1` check takes advantage of the fact that a number XORed with itself is 0, so only irreps with an odd number of orbitals matter.

robertodr:
I have no objection in making `_energies_from_fcidump` public, it was in an earlier iteration of the FCIDUMP functionality.

robertodr:
:+1: for adding the test and thanks again for taking care of this issue!

CDSherrill:
Ah, I quickly read this on my phone and missed that the XOR was in a loop.  Yes, looks good!

PR #1460, opened 2019-01-05
Implement DIIS absolute error criterion

susilehtola:
## Description
Psi4 is currently using the RMS norm to check for wave function convergence. However, it's well known that the RMS norm isn't as robust as the absolute error in some pathological cases: even though the RMS error is small, you may still have significant gradients for a few orbitals.

This PR implements the maximum absolute norm for DIIS in SCF calculations, which can be toggled by setting ```DIIS_RMS_ERROR false```

## Todos
- [x] Implemented absolute error criterion for DIIS

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -252,16 +252,17 @@ def scf_iterate(self, e_conv=None, d_conv=None):
     soscf_enabled = _validate_soscf()
     frac_enabled = _validate_frac()
     efp_enabled = hasattr(self.molecule(), 'EFP')
+    diis_rms = core.get_option('SCF', 'DIIS_RMS_ERROR')
 
     if self.iteration_ < 2:
         core.print_out("  ==> Iterations <==\n\n")
-        core.print_out("%s                        Total Energy        Delta E     RMS |[F,P]|\n\n" % ("   "
-                                                                                                      if is_dfjk else ""))
+        core.print_out("%s                        Total Energy        Delta E     %s |[F,P]|\n\n" % ("   "
+                                                                                                     if is_dfjk else "", "RMS" if diis_rms else "MAX"))
 
     # SCF iterations!
     SCFE_old = 0.0
     SCFE = 0.0
-    Drms = 0.0
+    Dnorm = 0.0
     while True:
         self.iteration_ += 1
 
@@ -331,15 +332,15 @@ def scf_iterate(self, e_conv=None, d_conv=None):
         status = []
 
         # Check if we are doing SOSCF
-        if (soscf_enabled and (self.iteration_ >= 3) and (Drms < core.get_option('SCF', 'SOSCF_START_CONVERGENCE'))):
-            Drms = self.compute_orbital_gradient(False, core.get_option('SCF', 'DIIS_MAX_VECS'))
+        if (soscf_enabled and (self.iteration_ >= 3) and (Dnorm < core.get_option('SCF', 'SOSCF_START_CONVERGENCE'))):
+            Dnorm = self.compute_orbital_gradient(False, core.get_option('SCF', 'DIIS_MAX_VECS'))
             diis_performed = False
             if self.functional().needs_xc():
                 base_name = "SOKS, nmicro="
             else:
                 base_name = "SOSCF, nmicro="
 
-            if not _converged(Ediff, Drms, e_conv=e_conv, d_conv=d_conv):
+            if not _converged(Ediff, Dnorm, e_conv=e_conv, d_conv=d_conv):
                 nmicro = self.soscf_update(
                     core.get_option('SCF', 'SOSCF_CONV'),
                     core.get_option('SCF', 'SOSCF_MIN_ITER'),
@@ -379,7 +380,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
                 diis_performed = False
                 add_to_diis_subspace = self.diis_enabled_ and self.iteration_ >= self.diis_start_
 
-                Drms = self.compute_orbital_gradient(add_to_diis_subspace, core.get_option('SCF', 'DIIS_MAX_VECS'))
+                Dnorm = self.compute_orbital_gradient(add_to_diis_subspace, core.get_option('SCF', 'DIIS_MAX_VECS'))
 
                 if (add_to_diis_subspace and core.get_option('SCF', 'DIIS_MIN_VECS') - 1):
                     diis_performed = self.diis()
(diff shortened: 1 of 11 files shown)


psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -252,16 +252,21 @@ def scf_iterate(self, e_conv=None, d_conv=None):
     soscf_enabled = _validate_soscf()
     frac_enabled = _validate_frac()
     efp_enabled = hasattr(self.molecule(), 'EFP')
+    diis_rms = core.get_option('SCF', 'DIIS_RMS_ERROR')
 
     if self.iteration_ < 2:
         core.print_out("  ==> Iterations <==\n\n")
-        core.print_out("%s                        Total Energy        Delta E     RMS |[F,P]|\n\n" % ("   "
-                                                                                                      if is_dfjk else ""))
+        if diis_rms:
```
dgasmith: Better to put this if similar to `is_dfjk` so we are not duplicating lines.

psi4/src/psi4/libscf_solver/cuhf.cc
```diff
@@ -392,7 +387,11 @@ double CUHF::compute_orbital_gradient(bool save_diis, int max_diis_vectors) {
 
         diis_manager_->add_entry(4, grad_a.get(), grad_b.get(), Fa_.get(), Fb_.get());
     }
-    return Drms;
+
+    if (options_.get_bool("DIIS_RMS_ERROR"))
+        return sqrt(0.5 * (std::pow(grad_a->rms(), 2) + std::pow(grad_b->rms(), 2)));
```
dgasmith: `std::sqrt` everywhere.

dgasmith:
Can you provide examples and literature on these pathological cases?

susilehtola:
@dgasmith well, it's quite a simple matter: if you have a large matrix where a few elements are large while the rest are numerically zero, the RMS norm may be small since it's only looking at the average. The maximum absolute norm, however, picks up this case. Q-Chem appears to have switched to the maximum absolute norm from the RMS norm in version 3.0. It's also the default in GAMESS'94, probably other codes as well.

This means that the bigger system you have, the more likely you're going to see problems with the RMS norm, as it will say the wave function has converged even if you have significant residual gradients for a limited set of orbitals.

The extreme case is, of course, if you e.g. use ```guess read``` with ```guess_mix true``` with a previously converged calculation. Then only the HOMO-LUMO pair isn't converged, and even in cases where the ground state wave function may have significant symmetry breaking, you might not pick this up since the orbital gradient only has two large elements.

I've been running calculations on some small transition metal complexes lately, and there it's quite common to see differences of 1 or 2 orders of magnitude between the RMS and MAX error.

psi4/driver/procrouting/scf_proc/scf_iterator.py
susilehtola: Thanks for the hint, still learning Python...

psi4/src/psi4/libscf_solver/cuhf.cc
susilehtola: Found a few entries elsewhere as well.

susilehtola:
Damn it. The history has somehow gotten screwed up again, even though I started off with a clean tree and **I** haven't pulled anything into the branch from master.... Ugh

susilehtola:
Every time this has happened, the problems have started when I've tried to push changes to the branch from my local git repo to find out that the branch has changed on github. When I pull the changes into my branch, I get conflicts, that I then try to resolve. The end result appears to be a big mess every time. Why is this happening and how can I avoid it??

dgasmith:
```bash
git push --force origin feature_branch
```

Kind of looks like you did:
```
git pull origin feature_branch
git push
```

susilehtola:
But the thing is, I haven't done the latter in the branch. For some reason, it had changed on github, which resulted in git push failing. Then I ```git pull origin feature_branch``` and ```git push origin feature_branch```, which caused the mess!

Now, rebasing isn't working anymore since the history is screwed up. Ugh.

susilehtola:
Thanks @loriab. Does that come as a printable PDF?

susilehtola:
Here's one sample case
```
molecule {
0 2
V 0.000000 0.000000 0.000000 
N -0.009654 1.561722 1.062138 
N 0.009654 -1.561722 1.062138 
N -1.561722 -0.009654 -1.062138 
N 1.561722 0.009654 -1.062138 
C -1.219978 2.113945 1.655979 
C 1.219978 -2.113945 1.655979 
C -2.113945 -1.219978 -1.655979 
C 2.113945 1.219978 -1.655979 
C 1.188278 2.080402 1.709075 
C -1.188278 -2.080402 1.709075 
C -2.080402 1.188278 -1.709075 
C 2.080402 -1.188278 -1.709075 
H -2.112996 1.691813 1.177356 
H 2.112996 -1.691813 1.177356 
H -1.691813 -2.112996 -1.177356 
H 1.691813 2.112996 -1.177356 
H 2.089441 1.627028 1.275983 
H -2.089441 -1.627028 1.275983 
H -1.627028 2.089441 -1.275983 
H 1.627028 -2.089441 -1.275983 
H -1.254235 3.214541 1.536136 
H 1.254235 -3.214541 1.536136 
H -1.905317 -1.281383 -2.742242 
H 1.905317 1.281383 -2.742242 
H 1.192045 1.877111 2.798018 
H -1.192045 -1.877111 2.798018 
H -3.178469 1.263762 -1.584921 
H 3.178469 -1.263762 -1.584921 
H -1.281383 1.905317 2.742242 
H 1.281383 -1.905317 2.742242 
H -3.214541 -1.254235 -1.536136 
H 3.214541 1.254235 -1.536136 
H 1.263762 3.178469 1.584921 
H -1.263762 -3.178469 1.584921 
H -1.877111 1.192045 -2.798018 
H 1.877111 -1.192045 -2.798018
}

memory 24 gb
set basis aug-pcseg-1
set reference uhf
set guess sad
set scf_type direct
set df_scf_guess false
set diis_rms_error true
set ints_tolerance 1e-14
energy('scf')

set guess read
set diis_rms_error false
energy('scf')
```

The first step converges as
```
   @UHF iter  19: -1477.30511799303667   -5.87622e-07   5.94954e-07 DIIS
  Energy converged.
```
where the second starts out as
```
                        Total Energy        Delta E     MAX |[F,P]|

   @UHF iter   0: -1477.30511823973779   -1.47731e+03   3.66064e-05 
   @UHF iter   1: -1477.30511827087457   -3.11368e-08   2.33361e-05 DIIS
```
so the absolute norm is larger by two orders of magnitude. And this is still a relatively small system.

loriab:
@susilehtola afraid it's jpeg, but it's the "prflow" files [here](https://github.com/psi4/psi4/tree/master/doc/sphinxman/source/resources)

susilehtola:
I think this must also be the default in The Code That Shall Not Be Mentioned. At least the 1999 paper by Rabuck and Scuseria, "Improving self-consistent field convergence by varying occupation numbers" states that

> The DIIS error factor is used to determine the status of the SCF iteration process and reaches zero  at convergence. This factor is the maximum element in the commutator of the density matrix and the Fock matrix, (FPS-SPF), where F is the Fock matrix, P is the density matrix and S is the overlap matrix.

Issue #1462, opened 2019-01-05
SAPT test suite failures on big-endian architectures


mbanck:
Psi4-1.2.1 on Debian shows several test suite failures on non-Intel hardware in the SAPT code that go like this:

```
90/140 Test #315: sapt1 ............................***Failed    7.40 sec
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/builddir/stage/usr/bin/psi4", line 269, in <module>
    exec(content)
  File "<string>", line 53, in <module>
  File "/<<PKGBUILDDIR>>/builddir/stage/usr/lib/s390x-linux-gnu/psi4/driver/driver.py", line 492, in energy
    wfn = procedures['energy'][lowername](lowername, molecule=molecule, **kwargs)
  File "/<<PKGBUILDDIR>>/builddir/stage/usr/lib/s390x-linux-gnu/psi4/driver/procrouting/proc.py", line 3438, in run_sapt
    e_sapt = core.sapt(dimer_wfn, monomerA_wfn, monomerB_wfn)

RuntimeError: 
Fatal Error: Laplace Quadrature requested for (E_HUMO - E_LOMO)/(E_LUMO-E_HOMO) > 7.0 * 10^12, quadratures are not designed for this range.
Error occurred in file: /<<PKGBUILDDIR>>/psi4/src/psi4/lib3index/denominator.cc on line: 609
The most recent 5 function calls were:

psi::PsiException::PsiException(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, char const*, int)

Exit Status: infile ( 1 ); autotest ( None ); sowreap ( None ); overall ( 1 )
```
See https://buildd.debian.org/status/fetch.php?pkg=psi4&arch=s390x&ver=1%3A1.2.1-2&stamp=1546705364&raw=0 for the full log.

Failing test cases from `quicktests` are `sapt1`, `sapt-dft1`, `tu5-sapt` and `pywrap-db1`, failing architectures are (so far) `s390x` and `powerpc`.

Anything I can do to investigate?

loriab:
This came to mind for failing SAPT tests: https://github.com/psi4/psi4/pull/1398/commits/05a8eacd9d3a1b34e008f4078787dc6f4cb2c64f . Main action is in the SAPT iterator destructor. Fixed a double free. Sorry for the formatting mixed in.

[...]

PR #1464, opened 2019-01-07
big endian quadrature file

loriab:
## Description
Solve #1462 by creating both big-endian and little-endian quadrature data files, then installing the proper one via cmake. Will probably fail cross-compile, but I don't know of such cases.

## Checklist
- [ ] no big endian machine to test on

## Status
- [x] Ready for review
- [x] Ready for merge


susilehtola:
What is even contained in these quadrature files? The Python source looks very short... Wouldn't it make sense to generate this on-the-fly?

PR #1470, opened 2019-01-09
Fix for issue #1356.

### LINKED ISSUE #1356: PBEh-3c: wrong energy passed into findif/gradient
PBEh-3c passes the wrong energy into Optking or Findif, sometimes, for some molecules. It happens for O2(s) but not for C3H8. See the O2(s) log posted here:

https://psi4.slack.com/files/U97H5UTPA/FE1KSGTT4/-.txt

As @hokru pointed out, it's the D3 dispersion energy of O2 that gets forwarded instead of the total energy, the most likely culprit (as identified by @andysim) is here:

https://github.com/psi4/psi4/blob/8cb0765c2735b0735fc3468d210d258de476a72b/psi4/driver/qcdb/intf_dftd3/runner.py#L88

PeterKraus:
## Description
Stops overwriting DFT energy by D3 energy only. I am not sure how to test for this, as the gradients as well as geometries with and without the patch agree to 8 DP's. I only noticed it when using the `vibrational_analysis(wfn)`, as there the current total energy plays a role...

## Todos
- [x] Resolves #1356

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/qcdb/intf_dftd3/runner.py
+++ b/psi4/driver/qcdb/intf_dftd3/runner.py
@@ -85,8 +85,7 @@ def run_dftd3(name, molecule, options, **kwargs):
         jobrec['error'] += repr(err)
     else:
         jobrec['success'] = True
-        jobrec['qcvars']['CURRENT ENERGY'] = copy.deepcopy(jobrec['qcvars']['DISPERSION CORRECTION ENERGY'])
-
+        
     return jobrec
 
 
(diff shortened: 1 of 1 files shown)


dgasmith:
Thanks, I think this is something for Lori to look into to double check.

dgasmith:
Here, I kind of feel like for the returned `jobrec` the dispersion energy is the correct thing for the current energy. I would hazard that this is an issue in reintegration of this energy down the line.

PeterKraus:
I had another look and I believe the problem is, that in a `ptype == "gradient"` calculation, we first do a normal energy calculation (where 

```jobrec['qcvars']['CURRENT ENERGY'] = copy.deepcopy(jobrec['qcvars']['DISPERSION CORRECTION ENERGY'])```

might be the correct thing to do), and then we do a gradient, where i don't think it is...

Issue #1272, opened 2018-10-01
Gaussian FCHK Errors and Inefficiencies


JonathonMisiewicz:
Reposted from the forum. I can't investigate this myself, due to being banned by Gaussian.

1. The MO and Density blocks are being mislabeled, for all levels of theory as far as I can tell. This prevents 3rd party apps from parsing the files correctly. The labels for the MO orbital energy and coefficients blocks should always be as follows:

Alpha Orbital Energies
Alpha MO Coefficients
Beta Orbital Energies
Beta MO Coefficients

Title case is important.

The Density blocks should be labeled according to the level of theory as follows:

Total SCF Density (this should be the label for both HF and DFT)
Total MP2 Density (don’t prepend ‘DF’ to any methods when using density-fitting)
Total MP3 Density
Total MP4 Density
Total CC Density
Total CI Density

Spin Density should be labeled similarly:

Spin SCF Density (this should be the label for HF, DFT, and CASSCF)
Spin MP2 Density
Spin MP3 Density
Spin MP4 Density
Spin CC Density
Spin CI Density

2. Don’t write the Beta blocks for RHF jobs, as their presence makes 3rd party apps think that the level of theory is UHF, and that in turn leads to unnecessarily expensive calcs

3. If canonical orbitals have been removed from the basis due to S_TOLERANCE, then the “number of independent functions” in the .fchk file should be the resulting number of canonical orbitals. Currently the “number of independent functions” always seems to be equal to the “number of basis functions”, which screws up parsing because the matrices aren’t nbasis x nbasis in this case.

4. This wasn’t happening in v1.1, but in v1.2, writing the .fchk file triggers printing of the MOs to the output file, which bloats the file tremendously. I can’t figure out how to turn this off, but it is undesirable as a default behavior.

andysim:
I think the problem is the MOs appearing in the output file in addition to the FCHK file.  Should be an easy one to track down, I hope.

susilehtola:
> I think the problem is the MOs appearing in the output file in addition to the FCHK file. Should be an easy one to track down, I hope.

Whoops, that's right..

susilehtola:
#1475 now includes fixes to everything else but the spin-restriction; this would be easy to fix as well, I just need a way to check if the calculation is RHF or not.

JonathonMisiewicz:
And, reopening. Let me quote the user:
```
Still not working in this latest version.

[...]

PR #1475, opened 2019-01-14
Fix inconsistencies in formatted checkpoint files

susilehtola:
## Description
The formatted checkpoint files currently produced by Psi4 misname the orbital coefficient and the density matrix fields. This PR (re)establishes the expected behavior, as well as fixes other bugs in the code.

## Todos
- [x] Fix labels of MO energies and coefficients
- [x] Fix labels of density matrices
- [x] Fix printout of number of linearly independent functions
- [x] Don't print the MO coefficients into the log file
- [x] Only print out beta orbs and spin density if wave function is not spin-restricted.
- [ ] Identify if density is correlated; if it is, print out both SCF and correlated density. Otherwise only print out SCF density.

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/writer.cc
+++ b/psi4/src/psi4/libmints/writer.cc
@@ -318,7 +318,6 @@ void FCHKWriter::write(const std::string &filename) {
     chk_ = fopen(filename.c_str(), "w");
     std::shared_ptr<BasisSet> basis = wavefunction_->basisset();
     int maxam = basis->max_am();
-    if (maxam > 4) throw PSIEXCEPTION("The Psi4 FCHK writer only supports up to G functions");
     std::shared_ptr<Molecule> mol = wavefunction_->molecule();
     SharedMatrix Ca_ao = wavefunction_->Ca_subset("AO");
     SharedMatrix Cb_ao = wavefunction_->Cb_subset("AO");
@@ -469,18 +468,29 @@ void FCHKWriter::write(const std::string &filename) {
                     for (int col = 0; col < 3; ++col) transmat->set(offset + row, offset + col, pureP[row][col]);
             }
         } else {
-            // Cartesians - P orbitals are fine, but higher terms need reordering
-            if (am == 2) {
-                for (int row = 0; row < 6; ++row)
-                    for (int col = 0; col < 6; ++col) transmat->set(offset + row, offset + col, cartD[row][col]);
-            }
-            if (am == 3) {
-                for (int row = 0; row < 10; ++row)
-                    for (int col = 0; col < 10; ++col) transmat->set(offset + row, offset + col, cartF[row][col]);
-            }
-            if (am == 4) {
-                for (int row = 0; row < 15; ++row)
-                    for (int col = 0; col < 15; ++col) transmat->set(offset + row, offset + col, cartG[row][col]);
+            // Cartesians - S and P orbitals are fine, but higher terms need reordering
+            switch (am) {
+                case (0):
+                case (1):
+                    break;
+
+                case (2):
+                    for (int row = 0; row < 6; ++row)
+                        for (int col = 0; col < 6; ++col) transmat->set(offset + row, offset + col, cartD[row][col]);
+                    break;
+
+                case (3):
+                    for (int row = 0; row < 10; ++row)
+                        for (int col = 0; col < 10; ++col) transmat->set(offset + row, offset + col, cartF[row][col]);
+                    break;
+
+                case (4):
+                    for (int row = 0; row < 15; ++row)
+                        for (int col = 0; col < 15; ++col) transmat->set(offset + row, offset + col, cartG[row][col]);
+                    break;
+
+                default:
+                    throw PSIEXCEPTION("The Psi4 FCHK writer only supports up to G shell (l=4) cartesian functions");
             }
         }
         offset += nfunc;
--- a/psi4/src/psi4/libmints/writer.cc
+++ b/psi4/src/psi4/libmints/writer.cc
@@ -563,21 +563,15 @@ void FCHKWriter::write(const std::string &filename) {
     write_matrix("Contraction coefficients", coefficients);
     write_matrix("Coordinates of each shell", shell_coords);
     write_number("Total Energy", wavefunction_->energy());
-    // write_matrix("Alpha Orbital Energies", wavefunction_->epsilon_a_subset("AO"));
-    write_matrix(wavefunction_->epsilon_a()->name().c_str(), wavefunction_->epsilon_a_subset("AO"));
-    // write_matrix("Alpha MO coefficients", reorderedCa);
-    write_matrix(wavefunction_->Ca()->name().c_str(), reorderedCa);
-    // write_matrix("Beta Orbital Energies", wavefunction_->epsilon_b_subset("AO"));
-    write_matrix(wavefunction_->epsilon_b()->name().c_str(), wavefunction_->epsilon_b_subset("AO"));
-    // write_matrix("Beta MO coefficients", reorderedCb);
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz:
This is a partial fix to #1272.

susilehtola:
@JonathonMisiewicz thanks for pointing that out; I had a vague recollection of other problems existing in the code.

JonathonMisiewicz:
Are the densities supposed to always be SCF? The original reporter for #1272 was talking about densities for other methods, as well. Otherwise, I agree with your assessment that this will be a full fix to #1272. I'll check with the original reporter before closing.

susilehtola:
@JonathonMisiewicz AFAIK it's only printing out SCF densities. If there's a way to get unrelaxed MP2, CC, etc density matrices to print out in Psi4, that'd be great. In that case, one would just need to add in the correlated density matrices to print out *on top* of the SCF quantities.

susilehtola:
Nevermind:
```
molecule {
0 1
H          0.43774        1.07383       -0.88985
O         -0.04657       -0.75634        0.00000
H          0.86440       -1.05180        0.00000
}

set basis pcseg-1
set reference rhf
set s_tolerance 1e-5
set guess sad
scf_en, scf_wfn = energy('scf', return_wfn=True)
fchk(scf_wfn, "scf.fchk")
ccsd_en, ccsd_wfn = properties('ccsd', properties=['dipole'], return_wfn=True, ref_wfn=scf_wfn)
fchk(ccsd_wfn, "ccsd.fchk")
```
shows that the density matrix **does** change from SCF to CCSD.

Now, I would need a function in the Wavefunction class that would get me the SCF density matrix.

susilehtola:
... as well a way to check if the density matrix is correlated or not.

loriab:
@susilehtola there's https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.h#L352 but consider it API subject to change, as reference_wavefunction never really fulfilled its promise. On the other hand, if there's a test that uses that functionality, we can make sure it gets replaced with something that serves the same purpose.

Probably any wfn that has a ref_wfn is correlated.

susilehtola:
Well, the code is there, but it appears that reference_wavefunction() is returning me the correlated wave function. At least, the density matrix I get from refwfn coincides with the one from SCF, even though I know this should NOT be the case.

JonathonMisiewicz:
Apparently, the handling of correlated/non-correlated densities has been [inconsistent](http://forum.psicode.org/t/problems-with-wavefunction-object-and-fchk-interface/532/9) for some time.

susilehtola:
I think this could be merged.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -469,27 +507,48 @@ void FCHKWriter::write(const std::string &filename) {
...
-                    for (int col = 0; col < 6; ++col) transmat->set(offset + row, offset + col, cartD[row][col]);
-            }
-            if (am == 3) {
-                for (int row = 0; row < 10; ++row)
-                    for (int col = 0; col < 10; ++col) transmat->set(offset + row, offset + col, cartF[row][col]);
-            }
-            if (am == 4) {
-                for (int row = 0; row < 15; ++row)
-                    for (int col = 0; col < 15; ++col) transmat->set(offset + row, offset + col, cartG[row][col]);
+            // Cartesians - S and P orbitals are fine, but higher terms need reordering
```
dgasmith: Why change this? It does not seem that you have altered the effect of the code but to use `case` which Psi4 stays away from.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -318,19 +316,59 @@ void FCHKWriter::write(const std::string &filename) {
...
+        DPHFtot_ao = DPHFa_ao->clone();
+        DPHFtot_ao->add(DPHFb_ao);
+        DPHFspin_ao = DPHFa_ao->clone();
+        DPHFspin_ao->subtract(DPHFb_ao);
+    }
+
+    const std::string &name = wavefunction_->name();
+    const std::string &basisname = basis->name();
     int nbf = basis->nbf();
+    assert(Ca_ao->nrow() == nbf);
```
dgasmith: Any reason for the assert here, by definition within Psi4 this is true.

susilehtola: I guess I wanted to make sure in case Ca_ao was symmetry blocked. Removed

psi4/src/psi4/libmints/writer.cc
susilehtola: The separate if clauses weren't aesthetic, so I wanted to use the syntax that exists in the language for this case.

Reverted to if else if.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -25,6 +25,7 @@
  *
  * @END LICENSE
  */
+#include <cassert>
```
dgasmith: Not needed I believe.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -563,21 +610,23 @@ void FCHKWriter::write(const std::string &filename) {
...
-    delete[] label;
+
+    // This is the format expected in the formatted checkpoint file
+    write_matrix("Alpha Orbital Energies", wavefunction_->epsilon_a_subset("AO"));
+    write_matrix("Alpha MO coefficients", reorderedCa);
+    write_sym_matrix("Total SCF Density", reorderedDt);
+    // In theory, the correlated density should be printed out with a
+    // legend depending on the method (MP2, MP3, MP4, CI, CC) but this
+    // is probably the most common anyhow
+    if (pHF) write_sym_matrix("Total CC Density", reorderedDPHFt);
```
dgasmith: This isn't just the CC density? Could be many theories.

loriab: [This](http://wild.life.nctu.edu.tw/~jsyu/compchem/g09/g09ur/f_formchk.htm) looks like a description of the fchk format that might have been in an earlier version of the Gaussian manual. I can't find similar in current docs.

The headers for `Total <mtd> Density` are so general (SCF, MP2, CI, CC) that CC may be reasonable. Or, since this is called from py, can pull `name` off wfn, regex it into SCF/MP2/CI/CC and pass that label into the fchk writer. In any case, the 2nd line of the fchk, `Type, Method, Basis`, provide a place to put the proper, e.g., `CISD` method.

Other than tidying up the name, lgtm! oh, and should probably check that gdma is content with the changes. Thanks for fixing this up.

psi4/src/psi4/libmints/writer.cc
susilehtola: Removed.

psi4/src/psi4/libmints/writer.cc
susilehtola: As discussed elsewhere, this is just an aesthetic issue. Since the checkpoint never includes other than the SCF and one unrelaxed post-HF density, it doesn't really matter how the post-HF density is named - as long as the name doesn't violate the standard. Since >99% of use cases are likely CC, I don't see why this would need to be changed here.

susilehtola:
“
        At present, only SCF quantities will appear in the FCHK.”

That’s not true - both the SCF and post-HF quantities are printed out. The code just assumes that if ref_wfn exists, the main density is a correlated one.


Susi Lehtola
Sent from my phone so excuse my brevity.


On 2 Feb 2019, at 18.20, Lori A. Burns <notifications@github.com<mailto:notifications@github.com>> wrote:


@loriab approved this pull request.

I added a commit with some notes. Very glad to have this patch up. Thanks!

PR #1477, opened 2019-01-15
Implement SAD_SCF_TYPE = CD and PK

susilehtola:
## Description
This PR implements  ```SAD_SCF_TYPE``` options that aren't currently allowed as ```SAD_SCF_TYPE``` options, as well as adds tests to check that the computed SAD guess is exact for closed-shell atoms.

## Todos
- [x] SAD solution is fully converged for noble gases when ```SAD_SCF_TYPE = SCF_TYPE```

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/sad.cc
+++ b/psi4/src/psi4/libscf_solver/sad.cc
@@ -517,13 +517,16 @@ void SADGuess::get_uhf_atomic_density(std::shared_ptr<BasisSet> bas, std::shared
             directjk->set_df_ints_num_threads(options_.get_int("DF_INTS_NUM_THREADS"));
         jk = std::unique_ptr<JK>(directjk);
     } else if (options_.get_str("SAD_SCF_TYPE") == "CD") {
-        CDJK* cdjk(new CDJK(bas,options_.get_double("CHOLESKY_TOLERANCE")));
+        CDJK* cdjk(new CDJK(bas, options_.get_double("CHOLESKY_TOLERANCE")));
         if (options_["DF_INTS_NUM_THREADS"].has_changed())
             cdjk->set_df_ints_num_threads(options_.get_int("DF_INTS_NUM_THREADS"));
         jk = std::unique_ptr<JK>(cdjk);
     } else if (options_.get_str("SAD_SCF_TYPE") == "PK") {
-        PKJK* pkjk(new PKJK(bas,options_));
+        PKJK* pkjk(new PKJK(bas, options_));
         jk = std::unique_ptr<JK>(pkjk);
+    } else if (options_.get_str("SAD_SCF_TYPE") == "OUT_OF_CORE") {
+        DiskJK* diskjk = new DiskJK(bas, options_);
+        jk = std::unique_ptr<JK>(diskjk);
     } else {
         std::stringstream msg;
         msg << "SAD: JK type of " << options_.get_str("SAD_SCF_TYPE") << " not understood.\n";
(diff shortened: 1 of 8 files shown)


dgasmith:
This gets pretty tricky as some integrals need to be written to disk which would clash with the integrals generated for the full SCF computations. MemDFJK and Direct were both picked as they do not have this particular problem.

Going to push this to 1.4 as I bet this will take quite a bit to get working.

susilehtola:
Hm, maybe it would make sense to just allow a full set of options to SAD_SCF_TYPE, but only use MemDFJK or Direct in the background?

psi4/src/psi4/libscf_solver/sad.cc
```diff
@@ -488,21 +488,33 @@ void SADGuess::get_uhf_atomic_density(std::shared_ptr<BasisSet> bas, std::shared
     // Setup JK
     std::unique_ptr<JK> jk;
 
+    std::string jk_type(options_.get_str("SAD_SCF_TYPE"));
+
+    // Handle default cases for compatibility
+    if ((jk_type == "PK") || (jk_type == "OUT_OF_CORE")) {
+        jk_type = "DIRECT";
```
dgasmith: This new strategy looks good to me. It hits the user a little bit, but overall makes sense I think as it gives them effectively what they asked for.

psi4/src/psi4/libscf_solver/sad.cc
```diff
@@ -488,21 +488,33 @@ void SADGuess::get_uhf_atomic_density(std::shared_ptr<BasisSet> bas, std::shared
...
             dfjk->set_df_ints_num_threads(options_.get_int("DF_INTS_NUM_THREADS"));
         dfjk->dfh()->set_print_lvl(0);
         jk = std::unique_ptr<JK>(dfjk);
-    } else if (options_.get_str("SAD_SCF_TYPE") == "DIRECT") {
+    } else if (jk_type == "DIRECT") {
         DirectJK* directjk(new DirectJK(bas));
         if (options_["DF_INTS_NUM_THREADS"].has_changed())
             directjk->set_df_ints_num_threads(options_.get_int("DF_INTS_NUM_THREADS"));
         jk = std::unique_ptr<JK>(directjk);
+    } else if (jk_type == "CD") {
```
dgasmith: Not sure what to do about CD however as I believe this will use disk and could cause issues.

susilehtola: Oh it's not in-core?

Well, change it to direct as well I guess :P

susilehtola: Unfortunately in that case the guess isn't exact, unless the Cholesky tolerance is very small.

loriab:
All three are consistent -- the `scf-guess` test is unhappy.

susilehtola:
I don't understand - `scf-guess` runs without problems on my copy.

susilehtola:
OK, I had a copy-paste error.

dgasmith:
@psi4/admins This is ready for review.

susilehtola:
> I'd protest against code duplication in options, but I hope to see liboptions retirement, so not worth dev time for consolidation.

Yeah I don't like it either, but it looks like it already has plenty of it going on and I couldn't think of a way to avoid it... is there a way to write both unsafe and safe versions of a function without code duplication?

PR #1483, opened 2019-01-18
Symmetrize DFMP2 gradient to largest Abelian group of the molecule

andysim:
## Description
The DFMP2 gradient was not symmetrized to the molecule's (Abelian) point group before, so optimizations could potentially break symmetry, as pointed out [on the forums](http://forum.psicode.org/t/losing-symmetry-on-first-step-of-geometry-optimization/1201).  This PR enforces the Abelian symmetry, preventing problems with symmetry breaking in optimizations.

## Todos
- [x] Fixes a bug in DFMP2 gradients that led to slight noise in which, in some cases, could break the symmetry of the molecule

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -313,6 +313,9 @@ SharedMatrix DFMP2::compute_gradient() {
     print_energies();
     energy_ = variables_["MP2 TOTAL ENERGY"];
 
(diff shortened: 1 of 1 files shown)


JonathonMisiewicz:
Thanks for the swift investigation! This will be a good bug fix to have in 1.3.

Issue #1439, opened 2018-12-20
ECPs Break Basis Guess


JonathonMisiewicz:
The following input file gives absurd results using the current dev version of Psi:
```
memory 1000 mb

molecule {
-1 1
 I
}

basis {
   assign I def2-tzvppd
}

set basis_guess true

energy('scf')
```

The small basis computation ends with
```
    Final Occupation by Irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     9,    2,    2,    2,    0,    4,    4,    4 ] 

  @DF-RHF Final Energy: -6892.35157729932507
```

Here's how the large basis computation went:
```
   @DF-RHF iter   0:   590.08005432721836    5.90080e+02   2.80084e-01
    Occupation by irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     5,    2,    2,    2,    1,    5,    5,    5 ] 

   @DF-RHF iter   1:   -67.13552163426122   -6.57216e+02   3.44500e-01 
    Occupation by irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     7,    2,    2,    2,    0,    5,    4,    5 ] 

   @DF-RHF iter   2:    -5.42420113211885    6.17113e+01   6.71875e-01 DIIS
   @DF-RHF iter   3:  -208.00917694080209   -2.02585e+02   3.53265e-01 DIIS
    Occupation by irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     8,    2,    2,    3,    0,    4,    4,    4 ] 

   @DF-RHF iter   4:  -213.73836050769307   -5.72918e+00   3.05852e-01 DIIS
   @DF-RHF iter   5:  -229.29046486569237   -1.55521e+01   1.60752e-01 DIIS
   @DF-RHF iter   6:  -231.83352362698204   -2.54306e+00   8.54643e-02 DIIS
   @DF-RHF iter   7:  -233.14739556085581   -1.31387e+00   1.00459e-02 DIIS
   @DF-RHF iter   8:  -233.26115993293257   -1.13764e-01   6.01984e-03 DIIS
   @DF-RHF iter   9:  -233.30679760470645   -4.56377e-02   2.25206e-03 DIIS
   @DF-RHF iter  10:  -233.31593146369286   -9.13386e-03   6.80191e-04 DIIS
[...]

PR #1484, opened 2019-01-19
prototype basis-dependent basis guess

loriab:
## Description
https://github.com/psi4/psi4/issues/1439#issuecomment-455799123 is a good idea prototyped here. Instead of always defaulting the cast-up basis to 3-21g, choose from within family. Here, only def2 is rolled out. Also, not integrated with basis blocks, so active for `set basis def2-tzvpp`, not `basis {assign I def2-tzvppd}` in original example.

## Todos
- [x] basis guess default Dunning, Petersson (zapa), and Jensen families to `pcseg-0`, which is 3-21g sized. none of these are ECP
- [x] basis guess default Karlsruhe to `def2-sv(p)`, so that ECP present for both orbital and fitting
- [x] remainder (incl. Pople and any custom) default to 3-21g. EDIT: Pople explicitly defaulting to 3-21g
- [x] EDIT: replaced DUAL in [this table](http://psicode.org/psi4manual/master/basissets_byfamily.html) with GUESS

- [x] I got `str not a callable fn` in error handling, so fixed up the error/trackback handling a bit. now prints to screen, too. EDIT: I think this got added elsewhere

## Testing
- [x] full ctest and pytest passes (except for dcft8, but that's a known issue)

## Status
- [x] Ready for review
- [x] Ready for merge


susilehtola:
Cool. I guess Pople sets (other than STO-3G) could start from 3-21G, all Dunning sets from cc-pVDZ (or cc-pVDZ-DK), and Jensen sets from pcseg-0.

PeterKraus:
While you're at this, could the ECP's (and heavier post-Kr elements) from def2-svp be included in def2-msvp (which is the expected behaviour)?

psi4/driver/qcdb/basislistdunning.py
```diff
@@ -620,3 +620,108 @@ def load_basfam_dunning():
     basis_augccpvqz.add_dualfit('aug-cc-pvqz-dual')
     basis_heavyaugccpvtz.add_dualfit('heavy-aug-cc-pvtz-dual')
     basis_heavyaugccpvqz.add_dualfit('heavy-aug-cc-pvqz-dual')
+
+    basis_ccpvdz.add_guess('pcseg-0')
```
PeterKraus: What's the rationale behind using `pcseg-0` as a guess for the `cc` family (as well as for `zapa` below)? Are they both ecp-free?

loriab: Yes, trying to match up ECP, designed for SCF vs. correlated, and puream characteristics better between guess orbital and target orbital basis sets. It was ECP that was the main problem (there's a trend there ...) when this was suggested. At the moment, the calc is stopping if guess/target ECP characteristics don't match up. This'll make that case less likely. IIRC, Susi suggested the pcseg as a modern ecp-free, correlation-designed family. I haven't used them yet myself.

susilehtola: No, pcseg-0 was picked because it's a modern minimal polarization consistent basis that's similar in size to 3-21G but more accurate. It also covers the first four periods of the periodic table, where ECPs aren't necessary.

psi4/src/export_mints.cc
```diff
@@ -174,15 +174,16 @@ std::shared_ptr<BasisSet> construct_basisset_from_pydict(const std::shared_ptr<M
     auto basisset = std::make_shared<BasisSet>(key, mol, basis_atom_shell, basis_atom_ecpshell);
 
     // Modify the nuclear charges, to account for the ECP.
-    if (totalncore) {
+    if (key == "BASIS") {
```
PeterKraus: This makes sure Z is not modified for fitting basis sets? Should it not be applied to `"GUESS"` too?

psi4/driver/qcdb/basislist.py
```diff
@@ -142,6 +144,13 @@ def add_rifit_default(self, fit):
         """
         self.ridef = sanitize_basisname(fit)
 
+    def add_guess(self, fit):
```
PeterKraus: is `fit` really an appropriate name of the variable here?

psi4/driver/procrouting/proc.py
```diff
@@ -1151,7 +1152,9 @@ def scf_helper(name, post_scf=True, **kwargs):
 
         # A user can set "BASIS_GUESS" to True and we default to 3-21G
         if cast is True:
-            guessbasis = '3-21G'
+            guessbasis = corresponding_basis(core.get_global_option('BASIS'), 'GUESS')[0]
+            if guessbasis is None:
+                guessbasis = '3-21G'
```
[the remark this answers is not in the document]
loriab: `pcseg-0` is new default for Dunning, pcseg, and zapa (and misc. correlation-consistent maybe), but `3-21G` is still the default for Pople and the default of last resort. This bit of code is the default of last resort, and it's a judgement call on whether `pcseg-0` or `3-21g` would be better. Glad of opinions.

That this was confusing, and I had to poke around on the PR to remind myself probably means I should add some docs.

PeterKraus:
@loriab this one needs a rebase.

psi4/src/export_mints.cc
loriab: Well spotted. GUESS is a little different in that cast-up is handled with the 1st SCF / 2nd SCF sequence in scf_helper in proc.py, rather than being brought along as a fitting basis companion to the orbital. The orbital/basis, jkfit, rifit are the `key`s here and in https://github.com/psi4/psi4/blob/master/psi4/driver/qcdb/libmintsbasisset.py#L605. Whereas in the cast-up, the GUESS basis arrives here as a BASIS in its own right, possibly with JKFIT of its own.

I'm pretty sure of this, but if it's not clear or you still suspect I'm wrong, please say so.

PeterKraus: It's clear now that you explained it. Thanks!

dgasmith:
Need conflicts fixed before merging.

PR #1494, opened 2019-01-24
ecp and df_fitting_condition fixes

loriab:
## Description
heals up the remaining broken test cases except for cookbook-rohf-rot

## Todos
- [x] ecp: now assert Z on molecule when ordinary basis applied, not just E…  …
- [x] df: df_fitting_condition was defaulting to 1e-12 unless explicitly set
- [x]better format error and also print to terminal

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1333,6 +1333,10 @@ def scf_helper(name, post_scf=True, **kwargs):
 
     if cast:
         core.print_out("\n  Computing basis projection from %s to %s\n\n" % (ref_wfn.basisset().name(), base_wfn.basisset().name()))
+        if ref_wfn.basisset().n_ecp_core() != base_wfn.basisset().n_ecp_core():
+            raise ValidationError("Projecting from basis ({}) with ({}) ECP electrons to basis ({}) with ({}) ECP electrons will be a disaster. Select a compatible cast-up basis with `set guess_basis YOUR_BASIS_HERE`.".format(
+                                  ref_wfn.basisset().name(), ref_wfn.basisset().n_ecp_core(), base_wfn.basisset().name(), base_wfn.basisset().n_ecp_core()))
+
         pCa = ref_wfn.basis_projection(ref_wfn.Ca(), ref_wfn.nalphapi(), ref_wfn.basisset(), scf_wfn.basisset())
         pCb = ref_wfn.basis_projection(ref_wfn.Cb(), ref_wfn.nbetapi(), ref_wfn.basisset(), scf_wfn.basisset())
         scf_wfn.guess_Ca(pCa)
--- a/tests/basis-ecp/input.dat
+++ b/tests/basis-ecp/input.dat
@@ -1,4 +1,4 @@
-memory 1000 mb
+#! check mixing ECP and non-ECP orbital/fitting basis sets in a session
 
 set df_fitting_condition 1e-12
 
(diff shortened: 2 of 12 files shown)


tests/basis-ecp/input.dat
```diff
@@ -0,0 +1,56 @@
+memory 1000 mb
```
andysim: Are we still adding `#! description` type headers?

tests/basis-ecp/input.dat
```diff
@@ -0,0 +1,56 @@
...
+
+
+clean()
+
+set basis_guess 3-21g
+
+try:
+    energy('scf')
+except ValidationError as e:
+    compare_integers(True, 'ECP electrons will be a disaster' in str(e), 'caught bad cast-up basis')  #TEST
```
andysim: This is a very clever approach.  We should add an `assert_throws` method to our comparison helpers that will do this (in another PR).  Something like
``` python
def assert_throws(fn, args, exception_type=Exception, expected_message=""):
    try:
        fn(*args)
    except exception_type as e:
        if expected_message in str(e):
            return True
        else:
            print("Exception did not contain expected message")
            return False
    except Exception as e:
        print("Expected exception did not get raised")
        return False
```
but without the errors that are surely present in this code that I just made up

andysim: oh yeah, if we do this it'd need an extra statement after the `try/except` to handle the case where it doesn't throw at all, of course

psi4/driver/procrouting/proc.py
```diff
@@ -1333,6 +1333,10 @@ def scf_helper(name, post_scf=True, **kwargs):
 
     if cast:
         core.print_out("\n  Computing basis projection from %s to %s\n\n" % (ref_wfn.basisset().name(), base_wfn.basisset().name()))
+        if ref_wfn.basisset().n_ecp_core() != base_wfn.basisset().n_ecp_core():
+            raise ValidationError("Projecting from basis ({}) with ({}) ECP electrons to basis ({}) with ({}) ECP electrons will be a disaster. Select a compatible cast-up basis with `set guess_basis YOUR_BASIS_HERE`.".format(
+                                  ref_wfn.basisset().name(), ref_wfn.basisset().n_ecp_core(), base_wfn.basisset().name(), base_wfn.basisset().n_ecp_core()))
+
```
susilehtola: Yes. Even better would be to check if the ECPs match, but that's for sure a smaller problem.

tests/basis-ecp/input.dat
loriab: yes. I usually just wait to catch the missing ones at some supervised docs build, when it prints a list of offenders.

andysim: okey dokey - just wanted to check in case it was accidentally forgotten

tests/basis-ecp/input.dat
loriab: I entirely agree with that strategy in general. But I think it'd be better to do it in `pytest`. That already has `with pytest.raises(exceptiontype):` and `with pytest.warns(warningtype:` constructs, and I turn all our `compare_*` functions into T/F for use with `assert` https://github.com/psi4/psi4/blob/master/tests/pytest/utils.py .

loriab:
Just a general note, if the lesson hadn't been hammered in already by @PeterKraus's experiences with sapt-ecp: small basis sets are _very_ susceptible to `DF_FITTING_CONDITION`. The opt-irc-1 test, which was healed by propagating the condition value, was taking ~15 opt cycles to converge each IRC point with 1e-12 (and finally failed on one) but takes ~4 cycles with 1e-10.

PR #1499, opened 2019-01-27
Add CSAM screening to ERISieve

zachglick:
## Description
This PR implements The Combined Schwarz Approximation Maximum (CSAM) estimate for prescreening two-electron integrals. CSAM was described in this paper by Thompson and Ochsenfeld: https://aip.scitation.org/doi/10.1063/1.4994190

In Psi4, TEIs are screened using the well-known Cauchy-Schwarz Inequality:
`(pq|rs)^2 <= (pq|pq)*(rs|rs)`
This type of screening is effective when the overlap between orbitals p and q (or r and s) is sufficiently small. However, it doesn't take into account the scenario in which the two orbital products pq and rs are individually significant, but very far apart. CSAM factors in this distance dependence. 

The following benchmarks show modest improvements in efficiency with negligible loss in accuracy:
Screen Shot 2019-05-28 at 1 59 57 PM

## Todos
- [x] Adds CSAM screening to ERISieve  
- [x] Adds python binding for the ERISieve class
- [x] Adds a ERISieve PsiAPI test

## Questions
- [ ] The `ERISieve` constructor now takes a boolean `do_csam` (which defaults to `false`). What's the best way to hook this up to the user interface?
- [x] The `ERISieve` class contains a partially-finished attempt at implementing a similar type of screening (QQR). Are their plans to finish this? Or implement other integral-related optimizations?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


zachglick:
The timings were with the direct algorithm, although it looks like this will speed up PK too. Unfortunately CSAM only works for 4-index integrals; there's much less literature on screening with density fitting.

On the bright side, CSAM outperforms QQR according to the paper (and is a lot simpler to implement), so that code might not need to be fixed.

dgasmith:
Density fitting can "look" like a 4-index integral as `(0p|qr)` which is how we compute them. CASM-style screening is unlikely to be able to be used in the contractions as it would make the GEMM's sparse rather than dense; however, the construction of the integrals can take some time and can benefit from this screening. The impact may be less, but still a useful thing to do!

zachglick:
That makes sense, you could compute overlap-like terms between all pairs of primary and auxiliary orbitals. It would be interesting to see if that cost is even worth the additional sparsity

psi4/src/psi4/libmints/sieve.cc
```diff
@@ -307,6 +309,45 @@ void ERISieve::integrals() {
 #endif
 }
 
+void ERISieve::csam_integrals() {
+    function_sqrt_.resize((size_t)nbf_);
+    shell_pair_exchange_values_.resize(nshell_ * (size_t)nshell_);
+    ::memset(&function_sqrt_[0], '\0', sizeof(double) * nbf_);
+    ::memset(&shell_pair_exchange_values_[0], '\0', sizeof(double) * nshell_ * nshell_);
```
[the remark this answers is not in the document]
zachglick: Done!

susilehtola:
The problem is that in density fitting the 0s are non-normalizable, and the auxiliary functions have unit norm, so you don't get anything from straightforward Schwarz screening...

psi4/src/psi4/libmints/sieve.cc
```diff
@@ -307,6 +309,45 @@ void ERISieve::integrals() {
...
+            int nQ = primary_->shell(Q).nfunction();
+            int oP = primary_->shell(P).function_index();
+            int oQ = primary_->shell(Q).function_index();
+            eri->compute_shell(P, P, Q, Q);
+
+            // Computing Q_mu_mu (for denominator of Eq.9)
+            if (Q == P) {
+                int oP = primary_->shell(P).function_index();
+                for (int p = 0; p < nP; ++p) {
+                    function_sqrt_[oP + p] = std::sqrt(std::fabs(buffer[p * (nP * nP * nP + nP) + p * (nP * nP + 1)]));
```
susilehtola: What's the Psi4 style: ```std::abs``` or ```std::fabs```?

IMHO it's a bit silly to stick to the ansi C name, since ```std::abs``` is polymorphic.

psi4/src/psi4/libmints/sieve.cc
```diff
@@ -329,5 +370,25 @@ bool ERISieve::shell_significant_qqr(int M, int N, int R, int S) {
...
+bool ERISieve::shell_significant_csam(int M, int N, int R, int S) {
+    // Square of standard Cauchy-Schwarz Q_mu_nu terms (Eq. 1)
+    double mn_mn = shell_pair_values_[N * (size_t)nshell_ + M];
+    double rs_rs = shell_pair_values_[S * (size_t)nshell_ + R];
+
+    // Square of M~_mu_nu terms (Eq. 9)
+    double mm_rr = shell_pair_exchange_values_[R * (size_t)nshell_ + M];
+    double nn_ss = shell_pair_exchange_values_[S * (size_t)nshell_ + N];
+    double mm_ss = shell_pair_exchange_values_[S * (size_t)nshell_ + M];
+    double nn_rr = shell_pair_exchange_values_[R * (size_t)nshell_ + N];
```
susilehtola: Would it make sense to define these as ```size_t``` from the start to avoid the casts?

susilehtola:
I would recommend enabling the combined Schwarz bound (CSB) even if the approximate CSAM bound isn't used.

susilehtola:
Also, Psi4 is probably not going to get much help from the additional screening, since IIRC range-separated functionals aren't evaluated using the short-range kernel that leads to efficient screening, but the long-range kernels that **don't** become sparse even the exchange is very short-range. :disappointed:

psi4/src/psi4/libmints/sieve.cc
zachglick: Good point, `std::fabs` is probably present for historical reasons

zachglick:
I temporarily put in the more rigorous CSB from the paper, but it was virtually the same as regular Schwarz screening, unlike the tight but non-rigorous CSAM. The CSB would definitely be useful if applied to other types of 2e integrals that contain faster decaying functions than 1/r. However, I don't think the `ERISieve` class was designed to do this.

Would any of these range-separated functions fit this criteria? It might be useful to make a general integral screening object that takes arbitrary functionals (and/or numbers of centers).

susilehtola:
Well, the paper says both CSB and CSAM are like Schwarz screening for the usual Coulomb interaction, except if you have a short-range kernel which is not used in Psi4 at present, so that's not surprising :)

psi4/src/psi4/libmints/sieve.cc
```diff
@@ -56,10 +56,12 @@ void ERISieve::common_init() {
 
     Options &options = Process::environment.options;
     do_qqr_ = false;  // Code below for QQR was/is utterly broken.
+    do_csam_ = false;
```
dgasmith: How do we switch this option on/off? I think we should write an accessor on the ERISieve objects and then have a setting so that this can be turned on/off at the JK (example) algorithm layer.

psi4/src/psi4/libmints/sieve.h
```diff
@@ -200,6 +212,9 @@ class PSI_API ERISieve {
             bool res = shell_significant_qqr(M, N, R, S);
             // std::cout << "QQR prune: " << res << "\n";
             return res;
+        } else if (do_csam_ && schwarz_bound) {
+            bool res = shell_significant_csam(M, N, R, S);
```
dgasmith: Just `return shell...`?

psi4/src/psi4/libmints/sieve.cc
zachglick: Done

psi4/src/psi4/libmints/sieve.cc
zachglick: I added a `do_csam` boolean as a constructor for the sieve object and an accessor method for the variable

psi4/src/psi4/libmints/sieve.h
zachglick: Done

tests/pytests/test_erisieve.py
```diff
@@ -0,0 +1,98 @@
...
+
+    shell_inds = range(basis.nshell())
+    quartets = itertools.product(shell_inds, shell_inds, shell_inds, shell_inds)
+
+    screen_count = 0
+    for m, n, r, s in quartets:
+        if not sieve_schwarz.shell_significant(m, n, r, s):
+            screen_count += 1
+
+    assert psi4.compare_values(0, screen_count, 'Quartets Schwarz Screened, Cutoff 0')  #TEST
```
loriab: `tests/pytests/` is an odd case where you want to `assert compare_fn` directly instead of `psi4.compare_fn`. At the top, do `from .utils import compare`. `compare` (or `compare_integers`) are used when floating-point comparison not needed. The former is a new qcel fn and you can feed it integer arrays, too.

the reason for using compare functions from utils instead of psi4 is that the latter returns True/raise for pass/fail, whereas for `assert`, we want True/False.

zachglick: Thanks, that explanation makes a lot of sense!

tests/pytests/test_erisieve.py
```diff
@@ -92,7 +93,7 @@ def test_schwarz_vs_csam():
...
             screen_count_none += 1
 
-    assert psi4.compare_values(75792, screen_count_both, 'Quartets CSAM Screened, Cutoff 0')  #TEST
-    assert psi4.compare_values(1344, screen_count_csam, 'Quartets CSAM Screened, Cutoff 0')  #TEST
-    assert psi4.compare_values(0, screen_count_schwarz, 'Quartets CSAM Screened, Cutoff 0')  #TEST
-    assert psi4.compare_values(27840, screen_count_none, 'Quartets CSAM Screened, Cutoff 0')  #TEST
+    assert psi4.compare_integers(75792, screen_count_both, 'Schwarz vs CSAM Screening, Cutoff 1.0e-12')  #TEST
+    assert psi4.compare_integers(1344, screen_count_csam, 'Schwarz vs CSAM Screening, Cutoff 1.0e-12')  #TEST
+    assert psi4.compare_integers(0, screen_count_schwarz, 'Schwarz vs CSAM Screening, Cutoff 1.0e-12')  #TEST
+    assert psi4.compare_integers(27840, screen_count_none, 'Schwarz vs CSAM Screening, Cutoff 1.0e-12')  #TEST
```
loriab: you want to drop the `psi4.` here. `psi4.compare_integers` is the True/raise, while `compare_integers` (from tests/pytests/utils is the True/False. Also, no need for `#TEST`. (The `#TEST` flag is for lines to remove when generating `samples/` from `tests/` and that doesn't happen for the pytests.

zachglick:
> LGTM, awesome python side testing!
> 
> Do we have a plan to add this keyword to globals so that Psi4 users can start accessing this screening?

Would adding a global, user-specified boolean (something like `CSAM_SCREENING`, default false) be the best way to do this? It could be accessed by the JK object, much like `INTS_TOLERANCE` already is

dgasmith:
Yes, I believe that is the best option. This would require some changes to the `jk.cc` factory.

loriab:
I think either now or soon, we want this on by default, as it's more filtering at negligible accuracy cost. Right?

What do we think of a screening enum (current options `SCHWARZ`, `CSAM` with former as present default, latter as soon default) that gets moderated by `INTS_TOLERANCE` obeyed by both?

zachglick:
I added a global `SCREENING` that can either be `SCHWARZ` or `CSAM` and is recognized by the JK objects. There's also an extra test to compare scf energies.

loriab:
Thanks much. I think it should be `INTS_SCREENING` but no use running CI again for that minor change. I'm trying to collect opinions on when should be turned on but RTG imo. Have you run a full test suite with `SCREENING=CSAM`?

susilehtola:
How much work would be involved in switching from use of long-range exchange to short-range exchange which can be screened more efficiently?

zachglick:
@loriab The ctest results are unaffected when running with `SCREENING=CSAM`

@susilehtola I'm not familiar with the difference between short and long range exchange. Would that be a different functional form of the two-electron integral?

susilehtola:
@zachglick this is just about how you calculate a range-separated functional. The point is that when you have a functional that has A% exact exchange at the short-range and B% exact exchange in the long-range, you can evaluate it either as B% full Hartree-Fock exchange and (A-B)% short-range exchange, or as A% full Hartree-Fock exchange and (B-A)% long-range exchange. Psi4 is currently doing it the latter way.

My intuition is that the former way of using short-range integrals would be more efficient for screening, since fewer terms can contribute at the short range and the more numerous long-range terms are killed by the range separation, but it might also be that since exchange integrals die off more rapidly in the long range than Coulomb integrals, the evaluating the long-range part is faster than the short-range part.

Of course, if you have both long- and short-range exchange, then you anyway have to evaluate the full exact exchange and the speedup isn't going to be mind-blowing. But, some functionals like HSE are short-range-only, and there it might make a difference if you don't have to evaluate the long-range part at all.

I guess the optimal implementation would include both and pick on a case by case basis: e.g. if you only have short-range exchange, then just do the short-range part and skip the full exact exchange.

PR #1348, opened 2018-11-10
Fix OpenMP config for clang-cl on Windonws

raimis:
## Description
This is part of *Psi4* porting to Windows (#933).

*FindOpenMP* doesn't yet support *clang-cl*, so the config has to be done manually.

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] Fix OpenMP config for clang-cl on Windonws
- [x] Check if OpenMP config hadn't failed silently

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #1520, opened 2019-02-05
openmp inheritance for plugins

loriab:
## Description
OpenMP detection is very finely balanced to compensate for LAPACK demands, `libgomp` self aggrandizement, and psi4's need to sanely and transparently impose the psi4-compile-omp status on any derived plugins. Unfortunately, #1348 disturbed some of that for the particular case of plugins with fewer active languages than their parent psi4.

A plugin (dfmp2) wasn't inheriting openmp from psi4 and thus was giving unknown openmp pragma warnings upon compile. This wasn't getting caught on linux because I was only checking that iomp5/gomp/omp weren't mixing, not that at least one was present. By happenstance, this did trigger an error on Mac in the course of building `psi4-dev`. This seems to patch things up on Linux. Will use this branch to try out Mac again.

## Status
- [x] Ready for review
- [x] Ready for merge


raimis:
Looks good! I don't see any problems with Windows.

PR #1427, opened 2018-12-12
Fix a memory error in psi::MOInfo::read_mo_spaces

raimis:
## Description
This is part of *Psi4* porting to Windows (#933).

The problem was already reported in #1255

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] Fix a memory error in `psi::MOInfo::read_mo_spaces`
- [x] Fix `psimrcc-fd-freq2` test on Windows

## Checklist
- [x] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #1521, opened 2019-02-05
Fix a memory error in psi::MOInfo::read_mo_spaces

raimis:
## Description
This is a part of *Psi4* porting to Windows (#933).

The size `ref_wfn.frzcpi()` is not `nirreps_ref` in `psimrcc-fd-freq2` test:
https://github.com/psi4/psi4/blob/5c0e495bdd3042795d8e39fe250ec153c4d236e9/psi4/src/psi4/libmoinfo/moinfo.cc#L257
It makes `convert_int_array_to_vector` to access invalid memory.

In the past, I replaced `nirreps` with `nirreps_ref` to solve another problem #1427 and this one became visible after fixing #1480.

## Todos
- [x] Demonstrate a memory error in `psi::MOInfo::read_mo_spaces`
- [x] Fix the memory error in `psi::MOInfo::read_mo_spaces`

## Questions
- [x] I'm lacking enough understanding of the code to fix, so just highlighting the problem. Who could help?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -101,15 +101,6 @@ jobs:
         displayName: "Test Psi4 (OpenMP)"
         workingDirectory: $(Build.BinariesDirectory)/install/lib
 
-      # Test (psimrcc-fd-freq2)
-      - script: |
-          ctest --build-config Debug ^
-                -R psimrcc-fd-freq2 ^
-                --output-on-failure ^
-                --parallel %NUMBER_OF_PROCESSORS%
-        displayName: "Test Psi4 (quick ctest)"
-        workingDirectory: $(Build.BinariesDirectory)/build
-
       # Test (quick ctest)
       - script: |
           ctest --build-config Debug ^
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -101,6 +101,15 @@ jobs:
         displayName: "Test Psi4 (OpenMP)"
         workingDirectory: $(Build.BinariesDirectory)/install/lib
 
+      # Test (psimrcc-fd-freq2)
+      - script: |
+          ctest --build-config Debug ^
+                -R psimrcc-fd-freq2 ^
+                --output-on-failure ^
+                --parallel %NUMBER_OF_PROCESSORS%
+        displayName: "Test Psi4 (quick ctest)"
+        workingDirectory: $(Build.BinariesDirectory)/build
+
       # Test (quick ctest)
       - script: |
           ctest --build-config Debug ^
--- a/psi4/src/psi4/libmoinfo/moinfo.cc
+++ b/psi4/src/psi4/libmoinfo/moinfo.cc
@@ -254,6 +254,8 @@ void MOInfo::read_mo_spaces() {
         intvec fvir_ref;
         //        intvec actv_docc_ref;
 
+        if (ref_wfn.frzcpi().n() != nirreps_ref) exit(42);
+
         focc_ref = convert_int_array_to_vector(nirreps_ref, ref_wfn.frzcpi());
         docc_ref = convert_int_array_to_vector(nirreps_ref, ref_wfn.doccpi());
         actv_ref = convert_int_array_to_vector(nirreps_ref, ref_wfn.soccpi());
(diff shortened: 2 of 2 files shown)


loriab:
This error may be of concern to @fevangelista.

andysim:
More problems from legacy code.  Doh!  Thanks for the report - I will hopefully get a chance to dig around and fix this one later today.

fevangelista:
Let me look into it.

andysim:
I cannot reproduce this one.  I pulled your branch and rebased it to master (which has not introduced any changes to the moinfo code) and it ran cleanly with address and memory sanitizers on my mac.  I modified your changes to read:
```c++
        if (ref_wfn.frzcpi().n() != nirreps_ref) exit(42);
        std::cout << "DIMENSIONS " << ref_wfn.frzcpi().n() << "  " << nirreps_ref << std::endl;
```
and it gave the following output:
```
(base) QuickSilver:~/programming/psi4/obj_asserts (fix_nirreps) > DYLD_INSERT_LIBRARIES=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/10.0.0/lib/darwin/libclang_rt.asan_osx_dynamic.dylib psi4
Performing finite difference calculations
 5 displacements needed ... 1 2 3 4 5
 8 displacements needed.
 1DIMENSIONS 4  4
 2DIMENSIONS 4  4
 3DIMENSIONS 4  4
 4DIMENSIONS 4  4
 5DIMENSIONS 4  4
 6DIMENSIONS 4  4
 7DIMENSIONS 4  4
 8
```
I tried the psimrcc-fd-freq1 test also and it runs just fine.

raimis:
@andysim Thanks for looking, but the problem is with psimrcc-fd-freq**2** test (not psimrcc-fd-freq**1**). After rebasing, the problem persists for me.

andysim:
I tried both `freq1` and `freq2` tests, but could not find any problems with GCC8 (with sanitizers) on my Mac.  My linux machines have a kernel that [doesn't allow the use of sanitizers](https://github.com/google/sanitizers/issues/856) so I can't test those.  However, if I put the same print statement into an Intel linux build, I see the exact same output that I posted above - *i.e.* that both dimensions are the same.  It appears that the mismatch in dimensions is only happening on windows, which is very strange.  Can anybody else confirm the behavior on Linux and/or Mac?

raimis:
@andysim I have add the print statement and made *Azure* to run the test.

It produces, the following output:
```
Performing finite difference calculations
 5 displacements needed ... 1 2 3 4 5
 8 displacements needed.
 1DIMENSIONS 4  4
 2DIMENSIONS 4  4
 3DIMENSIONS 4  4
 4DIMENSIONS 4  4
 5DIMENSIONS 4  4
 6DIMENSIONS 4  4
 7DIMENSIONS 4  4
 8
DIMENSIONS 2  4
```

The same output I get on *Linux* with GCC 7.3.0.

andysim:
I'm an idiot!  I put the print statement after the exit, so it didn't fire for the final displacement.  Sorry for wasting your time - I can now reproduce the problem.  I will see if I can figure out the cause...

andysim:
Ok, I have a PR to this branch that fixes it.  Here's some information for those reviewing.  The code that's causing the issue is the stuff that handles symmetry-dependent input arrays, such as DOCC, SOCC, making sure that they are mapped correctly if the molecule is displaced to a lower symmetry point group.  The dimension `nirreps` is the current (displaced) symmetry and may be lower than the parent symmetry `nirreps_ref`.  The problem came from re-use of the arrays that handle docc and frozen orbitals.  In the first step they are read from wfn (which relays the info in the reduced symmetry so no mapping is needed).  Then, after accumulating the info from wfn, the same arrays are used to query the input file, which has the full dimensions so the arrays should be resized.  @raimis correctly identified the mismatch in the dimensioning and his idea of increasing the dimensions to `nirreps_ref` is correct; the conversion routine just doesn't do this correctly without reading past the end of the wfn's reduced dimension arrays.  This fix just handles the wfn quantities first, then resizes for the mrcc parsing step.  There are a few comments in there to help clarify what's happening.  I think this section just failed to keep up with the many changes to how wfn handles things over the last decade.  Thanks very much for your help, @raimis .

raimis:
Confirmed! The problem is solved.

PR #1525, opened 2019-02-06
DFT: Adds OctreeGridBlocker -> BlockOPoints pruning check.

jturney:
## Description
`BlockOPoints` performs additional pruning but the result is never checked.  We recently hit a case where this pruning resulted in a `BlockOPoints` object with zero points. This would cause a segfault here: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/v.cc#L262 because `ncols` would be zero.

Adds a check to `OctreeGridBlocker` where the list of `BlockOPoints` is populated, if the number of points is zero then don't add it to the list.

## Checklist
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)  All DFT test cases pass on my machine.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/cubature.cc
+++ b/psi4/src/psi4/libfock/cubature.cc
@@ -4402,8 +4402,10 @@ void OctreeGridBlocker::block() {
     for (size_t A = 0; A < completed_tree.size(); A++) {
         std::vector<int> block = completed_tree[A];
         if (!block.size()) continue;
-        blocks_.push_back(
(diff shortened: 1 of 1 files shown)


jturney:
Don't merge this just yet. I'm checking to see if https://github.com/psi4/psi4/pull/1525/files#diff-0699760b6ecbe2dbea8a9ca91f1979b1R4412 needs to be executed regardless.

jturney:
All the tests still pass when the index is still offset.  Looking at the code it seems logical to continue offsetting the index as that is what is used when creating the next block of points.

PR #1531, opened 2019-02-13
Add 6-311G(2df,X) basis sets for second row atoms

JonathonMisiewicz:
## Description
Does exactly what it says.

I tested for correctness not only by comparison to basis sets that are on EMSL, but also by reproducing the G2 energy of SO2 reported by Brown and Barber (DOI: 10.1021/j100020a034) within 2E-5 hartrees. The G2 method requires this basis set. I get the same error margin when I try to reproduce their CO2 energies, so I attribute this to them not having converged their geometry tightly.

Many thanks to @loriab for guidance with Psi's basis set parser.

## Status
- [x] Ready for review
- [X] Ready for merge


JonathonMisiewicz:
Hold off on merging. My Pople basis set "wish" list has just grown quite a lot bigger.

JonathonMisiewicz:
That _should_ be all, again. I'll post back if I need to add still more Pople basis sets.

JonathonMisiewicz:
Assuming tests pass (which they should), ready for merge!

Issue #1532, opened 2019-02-13
Memory Leaks - DETCI, Others?


JonathonMisiewicz:
Carrying forward from [a forum post](http://forum.psicode.org/t/memory-leak-in-psi4-energy/1230/). 

A user reported memory leaks in Psi4. Terminating the following sample code prematurely
```
import psi4
psi4.set_memory('1000 MB')

mol = psi4.geometry('''
  O
  H 1 0.96
  H 1 0.96 2 104.5
''')

psi4.set_options({'basis' : 'cc-pVDZ'})
EHF, wfn = psi4.energy('scf', return_wfn='on')
for i in range(10000):
    psi4.energy('detci', ref_wfn=wfn)
```
You get the following memory usage graph, which spans several iterations:

![detci_mem](https://user-images.githubusercontent.com/29777889/52713199-aac84f00-2f64-11e9-9069-0bdda9ee6ac6.png)

User reported another, smaller memory leak in the `scf` code. That, I have not attempted to reproduce. There may be _other_ memory leaks, and it would be worthwhile to check for them.

andysim:
I bet this is an issue with mismatched `new`/`malloc` and `free`/`delete` in the legacy parts of the code.  The newer code uses smart pointers to avoid this.  The mismatched free and deletes can be tracked down using valgrind, which will also help to identify where the largest leaks are.  Anybody who tackles this should consider replacing the various explicit allocations with smart pointers to avoid this mess.  I'll take a pass through when I get chance.

hokru:
The MKL can also cause small memory leaks, iirc.

susilehtola:
cppcheck can also find mismatched new/malloc free/delete statements.

dgasmith:
Not surprising that DETCI leaks as it was created before the days of RAII and the DETCI module was expected to be run as a system process so all small leaks were non-issues then. I moved some of it to RAII, but there is a lot there and it is unlikely it will have a complete overhaul in the near future. I remember this was an issue before and we cleaned at least the most egregious leaks.

bgobbi:
While trying to optimize the geometry for conformers of the ligand in the 2vx0 pdb structure I am also running into memory issues. Our cluster is running with cgroups and I ask for 10GB, but tell psi4 to use only 8gb. In that case most of the calculations fail >374 out of 500 conformers. To make it work I have to ask for 20GB and use only 4GB in psi4.

[...]

PR #1534, opened 2019-02-14
Fix memory leaks in the code

andysim:
## Description
#1532 identified problem with memory leaks.  Using cppcheck, I've found a number of possible candidates.  My idea is to slowly work through all of them and replace all alloc/dealloc pairs with an RAII solution.  Feel free to chime in if you disagree with the and design decisions.  I'll do one object at a time so allow easy reverting, so this will eventually need to be squashed.

## Todos
- [ ] Fix leaks identified in #1532 by avoiding explicit allocation and deallocation.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/petitelist.h
+++ b/psi4/src/psi4/libmints/petitelist.h
@@ -50,6 +50,9 @@ class IntegralFactory;
 class Matrix;
 class Dimension;
 
+using ShellMapType = std::vector<std::array<int, 8>>;
+static std::array<int, 8> pgZeros{0, 0, 0, 0, 0, 0, 0, 0};
+
 inline int64_t ij_offset64(int64_t i, int64_t j) {
     return (i > j) ? (((i * (i + 1)) >> 1) + j) : (((j * (j + 1)) >> 1) + i);
 }
@@ -80,8 +83,11 @@ void delete_atom_map(int **atom_map, const std::shared_ptr<Molecule> &mol);
 void delete_atom_map(int **atom_map, const Molecule *mol);
 /// @}
 
-int **compute_shell_map(int **atom_map, const std::shared_ptr<BasisSet> &);
-void delete_shell_map(int **shell_map, const std::shared_ptr<BasisSet> &);
+ShellMapType compute_shell_map(int **atom_map, const std::shared_ptr<BasisSet> &);
+PSI_DEPRECATED(
+    "Calling `PetiteList.delete_shell_map` is no longer necessary.  In the version "
+    " of Psi4 following 1.4 this function will no longer be provided.")
+void delete_shell_map(ShellMapType shell_map, const std::shared_ptr<BasisSet> &);
 
 /////////////////////////////////////////////////////////////////////////////
 
@@ -161,7 +167,7 @@ class PSI_API PetiteList {
 
     char *p1_;
     int **atom_map_;
-    int **shell_map_;
+    ShellMapType shell_map_;
     int **unique_shell_map_;
     char *lamij_;
     int *nbf_in_ir_;
--- a/psi4/src/psi4/libmints/petitelist.h
+++ b/psi4/src/psi4/libmints/petitelist.h
@@ -51,7 +51,7 @@ class Matrix;
 class Dimension;
 
 using ShellMapType = std::vector<std::array<int, 8>>;
-static std::array<int, 8> pgZeros{0, 0, 0, 0, 0, 0, 0, 0};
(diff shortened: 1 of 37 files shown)


psi4/src/psi4/libmints/petitelist.h
```diff
@@ -50,6 +50,9 @@ class IntegralFactory;
 class Matrix;
 class Dimension;
 
+using ShellMapType = std::vector<std::array<int, 8>>;
+static std::array<int, 8> pgZeros{0, 0, 0, 0, 0, 0, 0, 0};
```
susilehtola: Should this be const?

susilehtola:
Cool!

psi4/src/psi4/libmints/petitelist.h
andysim: it sure should - thanks!

dgasmith:
This is heroes' work, Andy!

jturney:
Nice!

dgasmith:
@andysim I know this isn't done. But could we think about getting this PR in and keep working on this in subsequent PRs?

andysim:
Thanks for the prompt - it had slipped off my radar, true to form.  I have a few hours to bash at this a bit more this morning so I'll see how far I can get before we pull it in.  I still have the cppcheck output handy, so I should be able to make some pretty decent progress.

andysim:
Ok, so I've addressed everything that cppcheck found.  There are a number of false positives in that list, but I still fixed them regardless.  The full tests pass for me (except `nbody-convergence`, which needs a higher angmom build and `dcft8`, which segfaults).  Is anybody else seeing issues with `dcft8`?  I'm using the 2019 Intel compilers, so it's possible that there's an issue in my toolchain.  If others are seeing no problems in `dcft8` I'll try to debug it later today.

Unfortunately, cppcheck isn't too helpful with regards to the memory leaks that this PR was originally intended to address.  After it's pulled in I'll try and find some other valgrind-like tools to address those issues.

JonathonMisiewicz:
`dcft8` passes on my master. Let me know what you find for memory leak tools, or if `dcft8` continues giving you trouble.

andysim:
Thanks, @JonathonMisiewicz .  I fixed the bug and it appears to be a genuine problem - could you review my last commit please?  The DCFT tests all pass with it, so I assume it's fine.  Strangely enough, cppcheck didn't pick up on this one and we appear to have been getting lucky with our usual compilers accidentally zeroing out the memory.

All, this one's ready for review and merge.  I'm sure there's some overlap with the many LGTM commits, so I strongly suggest that this goes in after those PRs are all merged: that way I can take care of handling any merge conflicts as penance for my tardiness in finishing this cppcheck pass.  Also, I deliberately pushed each commit separately here in case we need to roll back any individual changes, so maybe squashing isn't the best way to go here.

JonathonMisiewicz:
@andysim That isn't "luck". That's the troublesome code being only three days old. The problematic arrays had been constructed by `init_int_array`, but @robertodr [had me](https://github.com/psi4/psi4/pull/1624#discussion_r280180193) [switch that out](https://github.com/psi4/psi4/pull/1624/commits/a6dc9f0c602ad874ca96fc7db3afa97e90c3af12). Apparently I hadn't done it correctly.

Do you have a reference on when this kind of initialization is necessary? Memory management in C++ continues to elude me.

andysim:
Ah, got it.  I don't have a good reference unfortunately, but a good rule of thumb is to assume that C++ is not initializing memory for you.  Assigning to an entire array can be quite expensive, so it's left to the programmer.  Moreover, you may be assigning values to it later, which would render any compiler-assigned initial value moot.  The old `init_int_array` and friends were inherited from Psi3: they take care of allocating *and* zeroing the memory for you (using `malloc` and then `memset`) so they made things a little more convenient for the programmer.

The modern C++ equivalent is to use `std::vector<int> myVec(nElements);`  That simply makes a vector of length `nElements`, which can be passed to any function expecting an `int *` by passing `myVec.data()` or `&myVec[0]`.  Note that even this way of constructing a vector doesn't initialize the data at all: to do that you need to either `std::fill(myVec.begin(), myVec.end(), 0);` or, better still, tell the compiler which value to use to construct the vector with `std::vector<int> myVec(nElements, 0);`  

You can see in my commit that I only zeroed the first (index 0) element.  The reason for that is because the loop over irreps starts from 1 in that code, so indices 1+ are correctly assigned.  Index 0 was not being touched, so it retained whatever value resulted from the call to `new` that created the array.  I hope this helps - please let me know if there's anything I can clarify at all.

PR #1536, opened 2019-02-14
Use before_cache to solve caching troubles on Travis

robertodr:
## Description
Lets packages update even though we're caching on Travis CI

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1538, opened 2019-02-15
Try stages to warm up cache

robertodr:
## Description
Trying out a different strategy to get the cache working. Supposedly this gets the cache warmed up for all the build lanes before they start. See https://docs.travis-ci.com/user/build-stages/warm-cache/

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,11 @@ jobs:
   include:
     - stage: Warm up cache
       script:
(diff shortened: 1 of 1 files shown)


robertodr:
This is more trouble than it's worth... #1536 is good enough

loriab:
too bad, it was a nice idea

PR #1551, opened 2019-02-21
dertype and gradient passing

JonathonMisiewicz:
## Description
Previously, the `dertype` specified in a frequency or optimization-with-hessian computaiton wouldn't get passed on to the pre-hessian gradient. This is now fixed. For optimizations, we can do one better and pass in the last gradient, so we don't need to recompute it. While solving this, I uncovered that supplying a dertype that was "too high" for the derivative level would lead to the default case of energy-only computations. That is also fixed.

**Although this is an enhancement, this is also a bug fix.** Without this PR, I have no way to get a frozen core MP2 optimization with a `full_hess_every` hessian. The user is responsible for supplying the `dertype`, and they currently have no way to get the `dertype` argument to the pre-hessian gradient.

## Checklist
- [x] Tested this solved my fc-MP2 use case
- [x] Tested this solved the case of a dertype=2 RHF hessian

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -58,7 +58,9 @@ def _find_derivative_type(ptype, method_name, user_dertype):
     a given method.
     """
 
-    if ptype not in ['gradient', 'hessian']:
+    derivatives = {"gradient": 1, "hessian": 2}
+
+    if ptype not in derivatives:
         raise ValidationError("_find_derivative_type: ptype must either be gradient or hessian.")
 
     dertype = "(auto)"
@@ -105,6 +107,8 @@ def _find_derivative_type(ptype, method_name, user_dertype):
 
         raise ValidationError("""Derivative method 'name' %s and derivative level 'dertype' %s are not available.%s"""
                               % (method_name, str(dertype), alternatives))
+
+    dertype = min(dertype, derivatives[ptype])
 
     return dertype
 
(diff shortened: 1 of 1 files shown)


JonathonMisiewicz:
This also needs to be marked "bug fix". Without this PR, `full_hess_every` is broken for fc-MP2 optimizations because `dertype` isn't propagating correctly.

psi4/driver/driver.py
```diff
@@ -106,6 +108,8 @@ def _find_derivative_type(ptype, method_name, user_dertype):
         raise ValidationError("""Derivative method 'name' %s and derivative level 'dertype' %s are not available.%s"""
                               % (method_name, str(dertype), alternatives))
 
+    dertype = min(dertype, derivatives[ptype])
```
loriab: this is in case Hess available but only grad needed?

I can believe this fn has some edge cases. I ended up refactoring it https://github.com/psi4/psi4/pull/1351/files#diff-a7c88e038c7c4c8ab84033e68c2b7c38R227

JonathonMisiewicz: Right. After the first commit, I realized there was an edge case where we were failing: if `dertype=2` for a method with hessians (currently RHF), the gradient code would see that `dertype!=1` and default to getting the gradient by energies, which is rather silly. Hence the second commit, where I added this line.

What do you recommend I do - proceed as-is but realize this is a short-term fix until #1351, or revert the second commit? Only the first commit is essential.

loriab: This is good for now. #1570 only adds the fn -- it won't be hooked up until #1351. just be aware that you'll have to test again that the path you want works under new scheme. there's a test_dertype can use to confirm.

PR #1553, opened 2019-02-22
json_wrapper uses qcschema_input

loriab:
## Description
Name is changing slightly. Both versions accepted.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/build_faq.rst
+++ b/doc/sphinxman/source/build_faq.rst
@@ -117,6 +117,7 @@ Runtime Setup
 #. :ref:`faq:psi4scratch`
 #. :ref:`faq:psidatadir`
 #. :ref:`faq:psi4psiapipath`
+#. :ref:`faq:modulenotfounderror`
 
 Running |PSIfour|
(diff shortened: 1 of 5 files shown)


loriab:
trouble is that "qcschema_input" only passes with a later qcel than the v0.1.3 that psi requires. will also need changes to qcel to allow "qcschema_molecule". can either
* mint a 0.1.4 with qcschema and qcschema_molecule changes. v1.3 would have to pin to 0.1.4 _or_ >=0.2.6
* mint a 0.2.6 and required pydantic unnecessarily for v1.3.

dgasmith:
Can we just autocorrect the name baed off QCElemental version?

loriab:
Very likely yes for the "qcschema_input" by standardizing down to the old string. The "qcschema_molecule" will be extra qcel code, though. I'll make the qcel changes and then revisit pinning strategy.

loriab:
This is RTG on the original issue. "qcschema_molecule" can be separate.

PR #1558, opened 2019-02-25
Update to include an auxiliary basis for def2-mSVP

carlosborca:
## Description
I realized that when running PBEh-3c, the calculation would take longer than expected. This was due to the use of a large basis during the density fitting procedure. I have just added the information required to make the code use the same auxiliary basis for the def2-mSVP basis as the ones used for def2-SVP. I did not test this modifications using the test cases.

## Todos
- [x] Improve performance of PBEh-3c when used with the default def2-mSVP basis set.

## Questions
- [ ] None. 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/qcdb/basislistother.py
+++ b/psi4/driver/qcdb/basislistother.py
@@ -338,7 +338,7 @@ def load_basfam_other():
     basisfamily_list.append(basis_pcsseg4)
 
     # Minix
-    basis_minix = BasisFamily('Minix', zeta=2)
+    basis_minix = BasisFamily('minix', zeta=2)
     
     basis_minix.add_jfit('def2-SVP-JFIT')    
     basis_minix.add_jkfit('def2-SVP-JKFIT')    
--- a/psi4/driver/qcdb/basislistother.py
+++ b/psi4/driver/qcdb/basislistother.py
@@ -338,13 +338,13 @@ def load_basfam_other():
     basisfamily_list.append(basis_pcsseg4)
 
     # Minix
-    basis_def2msvp = BasisFamily('def2-mSVP', zeta=2)
+    basis_minix = BasisFamily('Minix', zeta=2)
     
-    basis_def2msvp.add_jfit('def2-SVP-JFIT')    
-    basis_def2msvp.add_jkfit('def2-SVP-JKFIT')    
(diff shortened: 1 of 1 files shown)


loriab:
Great! How would you like to add MINIX with the same fitting basis sets?

psi4/driver/qcdb/basislistother.py
```diff
@@ -332,6 +337,15 @@ def load_basfam_other():
     basisfamily_list.append(basis_pcsseg3)
     basisfamily_list.append(basis_pcsseg4)
 
+    # Minix
+    basis_def2msvp = BasisFamily('def2-mSVP', zeta=2)
```
loriab: not quite following. RHS isn't MINIX, and `basis_def2msvp` on LHS has already been used above.

carlosborca:
I have just added information for MINIX. I hope it is acceptable to put in its own family.

psi4/driver/qcdb/basislistother.py
carlosborca: Oops.

hokru:
Aren't all those def2/JK fitting basis sets the same? (similar to def2/J)

dgasmith:
The key observation was that J can be the same for many basis sets, but K still must change on a per-basis setting.

loriab:
Psi never actually uses J-only fitting bases.

hokru:
For the same basis set family, a `QZ-JKFIT` will generally fit well enough for a `TZ` orbital basis. And after checking the def2/JK are all the same, regardless of zeta.

loriab:
@hokru is right. of the JKFIT below, the large ones (T, Q) are the same and the small ones (SV) are the same, and the large and small differ by post-La missing.
```
def2-qzvp-jkfit.gbs
def2-qzvpp-jkfit.gbs
def2-sv_p_-jkfit.gbs
def2-svp-jkfit.gbs
def2-tzvp-jkfit.gbs
def2-tzvpp-jkfit.gbs
```

The def2 RI sets _are_ different, so that aux family isn't entirely foolish. Goes to show that I was worn out of composing basis sets after Dunning. Whoops. :-)

PR #1568, opened 2019-03-05
Makes corrections suggested by lgtm.

obrien951:
## Description
Fixed a few lgtm errors

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/p4util/procutil.py
+++ b/psi4/driver/p4util/procutil.py
@@ -280,19 +280,14 @@ def extract_sowreap_from_output(sowout, quantity, sownum, linkage, allvital=Fals
         if allvital:
             raise ValidationError('Aborting upon output file \'%s.out\' not found.\n' % (sowout))
         else:
-            ValidationError('Aborting upon output file \'%s.out\' not found.\n' % (sowout))
             return 0.0
     else:
         while True:
             line = freagent.readline()
             if not line:
                 if E == 0.0:
-                    if allvital:
-                        raise ValidationError(
-                            'Aborting upon output file \'%s.out\' has no %s RESULT line.\n' % (sowout, quantity))
-                    else:
-                        ValidationError(
-                            'Aborting upon output file \'%s.out\' has no %s RESULT line.\n' % (sowout, quantity))
+                    raise ValidationError(
+                        'Aborting upon output file \'%s.out\' has no %s RESULT line.\n' % (sowout, quantity))
                 break
             s = line.strip().split(None, 10)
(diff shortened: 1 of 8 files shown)


psi4/driver/procrouting/dft/superfunctionals.py
```diff
@@ -221,7 +221,7 @@ def test_ccl_functional(functional, ccl_functional):
         gamma_bb[index] = point['gamma_bb']
         index = index + 1
 
-    super = build_superfunctional(functional, N, 1)
+    super = build_superfunctional(functional, True, N, 1)
```
dgasmith: Can you explicitly pass kwargs here? `(functional, True, npoints=N, deriv=1)`

loriab: @obrien951, please edit when you get a chance.

psi4/driver/p4util/procutil.py
```diff
@@ -280,19 +280,14 @@ def extract_sowreap_from_output(sowout, quantity, sownum, linkage, allvital=Fals
...
             if not line:
                 if E == 0.0:
-                    if allvital:
-                        raise ValidationError(
-                            'Aborting upon output file \'%s.out\' has no %s RESULT line.\n' % (sowout, quantity))
-                    else:
-                        ValidationError(
-                            'Aborting upon output file \'%s.out\' has no %s RESULT line.\n' % (sowout, quantity))
+                    raise ValidationError(
+                        'Aborting upon output file \'%s.out\' has no %s RESULT line.\n' % (sowout, quantity))
```
andysim: Not really a problem with what you've done in this PR, but I think we should try to settle on a standard for `"xxx"%(yyy)` vs `"xxx".format(yyy)`.  I think it's fair to say that the latter is becoming more canonical, and I'd prefer that be used for the two instances in this PR where the `%` version is used.  The `.format` is present up above, I noticed.

andysim: I welcome @loriab 's thoughts on that matter, but I suspect I already know that answer

dgasmith: I might have been the last Python `%` holdout and have switched to format for the QCArchive ecosystem.

loriab: I'm definitely in favor of f-string > `.format` > `%s` in general. Been gleefully moving to f-strings whenever it'll shorten the code. But these fns in particular are dead code for sow/reap that last worked in py2, so I was only willing to edit enough to silence lgtm.

andysim: If this is dead code, there's no need to change anything then. Just wanted to bring it up

PR #1576, opened 2019-03-15
JSON: Moves output data to tempfile

dgasmith:
## Description
Moves JSON output handle to tempfile to remove odd atexit issues.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/json_wrapper.py
+++ b/psi4/driver/json_wrapper.py
@@ -101,7 +101,7 @@ _qcschema_translation = {
         "mp2_opposite_spin_correlation_energy": {"variables": "MP2 OPPOSITE-SPIN CORRELATION ENERGY"},
         "mp2_singles_energy": {"variables": "NYI", "default": 0.0},
         "mp2_doubles_energy": {"variables": "MP2 CORRELATION ENERGY"},
-        "mp2_total_correlation_energy": {"variables": "MP2 CORRELATION ENERGY"},
+        "mp2_correlation_energy": {"variables": "MP2 CORRELATION ENERGY"},
         "mp2_total_energy": {"variables": "MP2 TOTAL ENERGY"},
     },
 
--- a/psi4/driver/json_wrapper.py
+++ b/psi4/driver/json_wrapper.py
@@ -34,7 +34,7 @@ import sys
 import copy
 import json
 import uuid
-import tempfile
+import atexit
 import traceback
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


loriab:
Ugg, Windows isn't even finding it to open, much less open-so-can't-close

```
      Start 419: json-schema-1-gradient
141/142 Test #419: json-schema-1-gradient ...........***Failed    1.97 sec
  File "D:/a/1/s/tests/json/schema-1-gradient/input.py", line 62, in <module>
    json_ret = psi4.json_wrapper.run_json(json_data)
  File "D:\a\1\b\build\stage\lib\psi4\driver\json_wrapper.py", line 191, in run_json
    core.set_output_file(outfile.name, False)
RuntimeError: 
Fatal Error: PsiOutStream: Failed to open file.
Error occurred in file: D:\a\1\s\psi4\src\psi4\libpsi4util\PsiOutStream.cc on line: 46
```
maybe it should be `core.set_output_file(outfile)` or `core.set_output_file(outfile.file)`.

also what do you think of putting the whole 190--215 in a `with tempfile.NamedTemporaryFile as handle:` to try to clean up closing issues?

dgasmith:
The tempfile will delete when it falls out of scope. `outfile` is Python object and will not work `outfile.file` is a handler and will not work.

Great:
```
Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).
```

We could also make a temporary directory and write to a file there that Psi creates would also work.

loriab:
| whose file attribute is the underlying true file object

looked so promising as something psi would blindly accept

loriab:
I think you solved this another way, so close, @dgasmith?

dgasmith:
Perhaps, this is still a problem however where `atext` is used rather than opportunistic cleaning which would be preferable. We might want to move this to a named temp dir + old UUID name instead which would have a similar effect.

dgasmith:
Well, some small effect here ensuring that Psi4 does not crash trying to read this file. This is ready for review.

psi4/driver/json_wrapper.py
```diff
@@ -178,6 +178,14 @@ def _clean_psi_environ(do_clean):
         psi4.core.clean_options()
         psi4.core.clean()
 
+def _read_output(outfile):
+    try:
+        with open(outfile, 'r') as f:
+            output = f.read()
+
+        return output
```
loriab: should be w/i context manager? or `return f.read()` since probable error in the `open` line

dgasmith: Unclear what happens in this case if closing the file fails? This seems safer.

loriab: try/except/finally, then, to minimize lines in `try`? approving regardless

dgasmith: We could, but why try to minimize lines in try here? We want something that cannot fail.

PeterKraus: If anything, I guess `try/except/else` would work (`finally` gets executed even on exceptions, which I guess we don't want), but I don't see a problem with this structure.

PR #1578, opened 2019-03-20
update run_json and wfn qcvars

loriab:
## Todos
- [x] modernizes `run_json` to the level of the ddd branch
  - sets all options at once and allows module-level options (`scf__reference` --> `set scf reference`)
  - shifts `json['psi4:qcvars']` --> `json['extras']['qcvars']`
  - `json['extras']['wfn_qcvars_only']` collects qcvars only from wfn, not from `P::e`
  - uses molecule dtype=2, qcschema one-word, and new error format
- [x] adds more psivars
  - `HF TOTAL HESSIAN`
  - removes scs-mp2 ss/os psivars in dfmp2 that were getting autovivified. thereby also corrects printing of 0s
  - in cc* modules, defines psivars on wfn that I'm confident will endure

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/json_wrapper.py
+++ b/psi4/driver/json_wrapper.py
@@ -205,13 +205,14 @@ def run_json(json_data, clean=True):
 
     # Attempt to run the computer
     try:
-        # qc_schema should be copied
-        json_data = run_json_qc_schema(copy.deepcopy(json_data), clean)
+        # qcschema should be copied
+        json_data = run_json_qcschema(copy.deepcopy(json_data), clean)
 
-    except Exception as error:
-        exc_type, exc_value, exc_traceback = sys.exc_info()
-        json_data["error"] = repr(traceback.format_exception(exc_type, exc_value,
-                                          exc_traceback))
+    except Exception as exc:
+        json_data["error"] = {
+            'error_type': type(exc).__name__,
+            'error_message': ''.join(traceback.format_exception(*sys.exc_info())),
+        }
         json_data["success"] = False
 
         with open(outfile, 'r') as f:
@@ -225,7 +226,7 @@ def run_json(json_data, clean=True):
     return json_data
 
 
-def run_json_qc_schema(json_data, clean):
+def run_json_qcschema(json_data, clean):
     """
     An implementation of the QC JSON Schema (molssi-qc-schema.readthedocs.io/en/latest/index.html#) implementation in Psi4.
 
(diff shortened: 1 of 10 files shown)


psi4/driver/json_wrapper.py
```diff
@@ -205,13 +205,14 @@ def run_json(json_data, clean=True):
...
+        json_data = run_json_qcschema(copy.deepcopy(json_data), clean)
+
+    except Exception as exc:
+        json_data["error"] = {
+            'error_type': type(exc).__name__,
+            'error_message': ''.join(traceback.format_exception(*sys.exc_info())),
+        }
         json_data["success"] = False
 
         with open(outfile, 'r') as f:
```
dgasmith: While we are here could we move the output file to the scratch directory rather than `tmp` in an effort to remove some odd supercomputer bugs we have seen.

loriab: Not following -- the only file written to here presently in scratch directory https://github.com/psi4/psi4/pull/1578/files#diff-f0c7d86da302944bdf61cc5d48e3701cR190

dgasmith: Ah, ok so we are doing that. That explains an issue on my end, thanks!

psi4/driver/procrouting/proc.py
```diff
@@ -2217,6 +2217,9 @@ def run_scf_hessian(name, **kwargs):
     H = core.scfhess(ref_wfn)
     ref_wfn.set_hessian(H)
 
+    # Clearly, add some logic when the reach of this fn expands
+    ref_wfn.set_variable('HF TOTAL HESSIAN', H)
```
andysim: It's `HF` here, and `SCF` in a different file below.  Is that intentional?  Also, is the intent of the `TOTAL` qualifier to eventually allow reporting of correlation contributions separately?

loriab: Yes, the idea is that `SCF TOTAL HESSIAN` gets set where the code knows is an SCF (in scfhess) and `HF TOTAL HESSIAN` gets set where the code knows it's Hartree–Fock. Then `B3LYP TOTAL HESSIAN` would get set for that method.

The "total" is in analogy to `MP2 TOTAL ENERGY`, where ref and corl separation is widespread. Not expecting any such for grad/hess, unless `HF TOTAL HESSIAN` was separately available.

andysim: Ok, just wanted to check to make sure about the SCF/HF terminology.  I have similar reservations to you about the separation of grad/hess components, but I like the idea of consistent key for the data arrays.  Thanks for the clarification.

Issue #1579, opened 2019-03-20
DF DCFT Properties Incorrect


JonathonMisiewicz:
An error which is very high on my to-fix list, but where the fix is very involved.

DF-DCT properties as present are not correct. I've tested against my own Python code and the Psi code and found disagreements in the fifth and sixth decimal place between finite difference and analytic properties. I've spoken with the code authors, and we are aware of two sources of disagreement.
1. For gradients, the RDMs are contracted against the conventional derivative integrals, not the density fitted derivative integrals.
2. Due to a subtle point in the derivation, the DF-DCT cumulant stationarity conditions are not variational minimization of the energy, as they are in conventional DCT. Accordingly, you're missing a cumulant response term for gradients and other properties.

The fix for point 1 will hopefully just be marking the wavefunction as density fitted and leave the contractions to the `Deriv` object. However, DCT at present interferes with RDM processing in `libtrans`, so the fix will be likely be more involved.

The fix for point 2 is not clear. I'm aware of three options:
* Change the DF-DCFT stationarity conditions to be variational minimization of the energy
* Change which terms are to be contracted against JKFIT integrals and which against RIFIT integrals, so the stationarity conditions are "fixed" automatically
* Add a cumulant response term. (Please don't make me do this.)

I favor option B, but it'll take time to see what B does to energies.

ssh2:
I agree that option B is the best. The effect of the fix will be very minor. This can be easily tested by comparing to DCT code with conventional integrals (and we have done this in the past). Also, current version of the code should recover the exact DCT energy and gradient when using a very large auxiliary basis set.

JonathonMisiewicz:
Status update on this, including developments from half an hour ago.

Numerical tests indicate the difference in reaction energies between the current DF scheme and the new one are negligible. The Excel file is attached if anybody wants to look over the numbers themselves. @ssh2 and I are agreed that between that and some additional theoretical points, we'll just switch the density fitting scheme to fix Point 2. In the new scheme, the cumulant gets contracted against RIFIT integrals as before, and the _entire_ rest of the 2RDM gets contracted against JKFIT integrals.

[df_dcft_comparison.xlsx](https://github.com/psi4/psi4/files/3023032/df_dcft_comparison.xlsx)

[...]

PR #1580, opened 2019-03-20
Add DF-DCFT warning messages.

JonathonMisiewicz:
## Description
Adding some warnings regarding #1579.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dcft/dcft_compute.cc
+++ b/psi4/src/psi4/dcft/dcft_compute.cc
@@ -56,6 +56,9 @@ double DCFTSolver::compute_energy() {
     }
 
     if (options_.get_str("DCFT_TYPE") == "DF") {
+        outfile->Printf(
+            "\n\n\t**** Warning: The density-fitted DCFT cumulant is only approximately variational.\n"
+            "\t     The stationarity conditions, and thus DF-DCFT energy, are subject to change. ****\n");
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/dcft/dcft_compute.cc
```diff
@@ -56,6 +56,9 @@ double DCFTSolver::compute_energy() {
     }
 
     if (options_.get_str("DCFT_TYPE") == "DF") {
+        outfile->Printf(
+            "\n\n\t**** Warning: The density-fitted DCFT cumulant is only approximately variational.\n"
+            "\t     The stationarity conditions, and thus DF-DCFT energy, are subject to change. ****\n");
```
loriab: stationary, stationality, stationarity?

JonathonMisiewicz: I still prefer "stationarity", but it's not something I'll argue about it if anybody has strong opinions to the contrary.

loriab: I don't have strong feelings -- just didn't recognize the word. Ok by me so long as intended, not typo.

susilehtola: It should IMHO indeed be "stationarity condition", i.e. "condition for a stationary wave function"

PR #1581, opened 2019-03-20
quadrature weights cutoff

hokru:
## Description
XC quadrature weights can become very small and thus their contributions are negligible. This PR introduces a cutoff for small weights. The grid points with small weights are not included in the final grid. Such a cutoff is a standard procedure in many programs.

## Todos
- [x] expert option `DFT_WEIGHTS_TOLERANCE` keyword (current default `1e-14`).
- [x] small error overview added
- [x] together with pruning ca. 40% faster calculation for C60 compared to `v1.3` :-)
- [x] simple test added

## Questions
- [x] which default value? -> `1e-15` conservative or `1e-14` like in ORCA

## Checklist
- [x] all `ctest -L dft` pass

## Status
- [x] Ready for review
- [x] Ready for merge

------
water dimer (RKS total energy): `PBE/aug-cc-pVTZ (99,520)` econv=1e-10 dconv=1e-8

 | tolerance | error | #points |
 | --- | --- | ---|
 | off | 0.0 | 350460 |                                                 
 | 1e-30 | 1.36e-12 | 349417 |
 | 1e-25 | 1.65e-12 | 348670 |
 | 1e-20 | 1.65e-12 | 346918 |
 | 1e-18 | 1.68e-12 | 345747 |
 | 1e-16 | 1.63e-11 | 340646 |
 | 1e-15 | 2.74e-11 | 339647|
 | 1e-14 | 5.73e-09  |  335089 |
 | 1e-13 | 3.65e-08  |  331285 |
 | 1e-12 | 3.18e-07 | 326462 |

c2 sym. C60 isomer  (RKS total energy): `PBE/aug-cc-pVDZ (75,302)` econv=1e-10 dconv=1e-8

 | tolerance | error | #points |
 | --- | --- | ---|
 | off | 0.0 |  1359000 |    
 | 1e-30 -| 3.17e-10| 1229846|
 | 1e-25 | -5.34e-10| 1209622|
 | 1e-20 | 5.94e-10 |1183674|
 | 1e-18 | 5.41e-10 |1171102|
 | 1e-16 | -6.25e-10 |1156776|
 | 1e-15 | 2.48e-09| 1131228|
 | 1e-14 | 3.54e-09| 1123018|
 | 1e-13 | 1.25e-07 |1110668 |
 | 1e-12 |1.07e-06 |1085824 |

-----

*further timings*
C60 fullerene (C2h) :PBE/aug-cc-pVDZ with grid (434,75); 16 Threads; comparison to `v1.3.1`
* pruning `ROBUST`=1.33 // `1361880/1953000` grid points // dE=2E-11
* pruning `TREUTLER=1.36`// `1064520/1953000` grid points // dE=-4E-05
* pruning `ROBUST` +  `screening(1e-14)` = 1.39; `1070388/1953000` grid points //dE=-2E-09
* pruning `TREUTLER` +  `screening(1e-14)` = 1.44; `773028/1953000` grid points //dE=-4E-05

### DIFF
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -1494,8 +1494,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("DFT_BS_RADIUS_ALPHA", 1.0);
         /*- DFT basis cutoff. -*/
         options.add_double("DFT_BASIS_TOLERANCE", 1.0E-12);
-        /*- grid weight cutoff. !expert -*/
-        options.add_double("DFT_WEIGHTS_TOLERANCE", 1.0E-14);
+        /*- grid weight cutoff. Disable with -1.0. !expert -*/
+        options.add_double("DFT_WEIGHTS_TOLERANCE", 1.0E-15);
         /*- The DFT grid specification, such as SG1.!expert -*/
         options.add_str("DFT_GRID_NAME", "", "SG0 SG1");
         /*- Select approach for pruning. Options ``ROBUST`` and ``TREUTLER`` prune based on regions (proximity to nucleus) while
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -1494,6 +1494,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("DFT_BS_RADIUS_ALPHA", 1.0);
         /*- DFT basis cutoff. -*/
         options.add_double("DFT_BASIS_TOLERANCE", 1.0E-12);
+        /*- grid weight cutoff. !expert -*/
+        options.add_double("DFT_WEIGHTS_TOLERANCE", 1.0E-14);
         /*- The DFT grid specification, such as SG1.!expert -*/
         options.add_str("DFT_GRID_NAME", "", "SG0 SG1");
         /*- Select approach for pruning. Options ``ROBUST`` and ``TREUTLER`` prune based on regions (proximity to nucleus) while
--- a/tests/dft-pruning/input.dat
+++ b/tests/dft-pruning/input.dat
@@ -12,32 +12,29 @@ set maxiter 1
 set FAIL_ON_MAXITER false
 set DFT_WEIGHTS_TOLERANCE -1.0
 
-SCHEMES={
-"NONE" : 66989, #TEST
-"FLAT" : 66989, #TEST
-"TREUTLER" : 37217, #TEST
-"ROBUST" : 45929, #TEST
-"P_SLATER" : 21465, #TEST
-"D_SLATER" : 15681, #TEST
-"LOG_SLATER" : 13377, #TEST
-"P_GAUSSIAN" : 16059, #TEST
-"D_GAUSSIAN" : 11727, #TEST
-"LOG_GAUSSIAN" : 15948, #TEST
+
+SCHEMES = {
+    "NONE": 66989,  #TEST
+    "FLAT": 66989,  #TEST
+    "TREUTLER": 37217,  #TEST
+    "ROBUST": 45929,  #TEST
+    "P_SLATER": 21465,  #TEST
+    "D_SLATER": 15681,  #TEST
+    "LOG_SLATER": 13377,  #TEST
+    "P_GAUSSIAN": 16059,  #TEST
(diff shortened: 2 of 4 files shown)


susilehtola:
Looks reasonable, other than what happens near the nucleus: the weight has r^2 in the Jacobian, but tight functions can have a large value there. Screening far away should be fine.

psi4/src/read_options.cc
```diff
@@ -1484,6 +1484,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("DFT_BS_RADIUS_ALPHA", 1.0);
         /*- DFT basis cutoff. -*/
         options.add_double("DFT_BASIS_TOLERANCE", 1.0E-12);
+        /*- grid weight cutoff. !expert -*/
+        options.add_double("DFT_WEIGHTS_TOLERANCE", 1.0E-14);
```
PeterKraus: I suppose the default should be the largest value that still passes full tests. I had a look at the `dft-smoke` test, the first change I see is on the "total points" line, where your test has ~2300 fewer grid points. The SAD energy is a full Hartree off...

susilehtola: @PeterKraus before or after this change..?

hokru:
I didn't factor in that the weights can be negative. Somehow I thought they must be positive.
There are many 'big' negative weights

IMO it should remain an expert option and somewhat hidden.

psi4/src/read_options.cc
PeterKraus: After this change, there are fewer points.

susilehtola: Yeah I was wondering about the Hartree difference in the SAD energy. But yes, since the quadrature weights can be negative, which was not taken into account in the code, it's quite possible to get non-reasonable energies...

tests/dft-smoke/input.dat
```diff
@@ -16,7 +16,7 @@ H 1 1.0
 H 1 1.0 2 104.5
 }
 
-
+set dft_weights_tolerance -100.0
```
PeterKraus: Hey, that's cheating! :stuck_out_tongue_winking_eye:

hokru: "This is not the line you are looking for"

hokru:
short study of the total error in the absolute energies for two cases added.

hokru:
on hold until PR #1591 is finished.

dgasmith:
Can we rebase and patch up this PR now that #1591 is in as well. Be good to look at migrating this change in as well.

hokru:
rebased and test added.

tests/dft-pruning/input.dat
```diff
@@ -27,4 +28,16 @@ SCHEMES={
...
-  compare_integers(SCHEMES[scheme], int(variable('XC GRID TOTAL POINTS')), 'grid porints for '+scheme) #TEST
+  compare_integers(SCHEMES[scheme], int(variable('XC GRID TOTAL POINTS')), 'grid points for '+scheme) #TEST
+
+# screening based on small weights
+set DFT_PRUNING_SCHEME ROBUST
+screening={
+1e-12:45198,
+1e-14:45429,
+1e-16:45679
+}
```
dgasmith: Can you format this a bit more inline with general Python formatting?

dgasmith:
@psi4/admins This could use a review round.

hokru:
> tol=14 or 15, everyone?

I have the impression PSI4 always leans on the conservative side of thresholds, instead of tuning things for speed. Also, perhaps for floppy frequencies or similarly sensitive calculations one wants to be as safe as possible. Then tol=15 might be the safe choice.

But..on the other hand it is DFT and the error from the quadrature itself is larger except for impractically large grids, I suppose. The often tight gradients and frequencies tests of psi4 pass with the current tolerance of 14.

A dynamic adjustment of the threshold could be envisioned, but I think that would be overkill.

jturney:
tol=15 sounds good to me.

tests/dft-pruning/input.dat
hokru: I ran yapf over some parts

hokru:
I increased the default threshold to 1e-15 as this seems slightly favoured. (not re-doing the timing benchmark, should be very similar)

PR #1583, opened 2019-03-22
Fix Database for Open Shells; Rip out Reap/Sow

JonathonMisiewicz:
## Description
Fix the ability to use database computations when open-shell molecules, and thus reference changes, are involved. Because the time at which `molecule` exists, the fix I implemented would be incompatible with reap and sow, so those have been removed. I've simplified the code accordingly.

## Checklist
- [x] `pywrap-db1` and `pywrab-db3` pass. There doesn't seem to be a Ctest tag for wrapper_database

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/wrapper_database.py
+++ b/psi4/driver/wrapper_database.py
@@ -38,6 +38,7 @@ import pickle
 import collections
 
 from psi4.driver import constants
+from psi4.driver import p4util
 from psi4.driver.driver import *
 # never import aliases into this file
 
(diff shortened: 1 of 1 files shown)


psi4/driver/wrapper_database.py
```diff
@@ -276,7 +277,7 @@ def database(name, db_name, **kwargs):
         except AttributeError:
             pass
         else:
-            if yes.match(str(database.isOS)):
+            if p4util.yes.match(str(database.isOS)):
```
loriab: This is the bug fix line for v1.3.1?

JonathonMisiewicz: No, that is _a_ bug fix line.

If you look at the code that orders the RHF -> UHF switch, there's a check on `molecule.multiplicity()`. `molecule` wasn't defined until the new input file ran, before this PR. By getting rid of reap and sow, I could move the definition of molecule up, so that `molecule.multiplicity()` is defined when the code assumed it was.

This is a large enough change that I understand it not being a 1.3.1 candidate.

loriab: thanks. yes, if the bugfix is most of this PR, I think it can wait in master.

PR #1595, opened 2019-04-10
Miscellaneous minor changes

robertodr:
## Description
Extracts some miscellaneous, uncontroversial and minor, changes from #1443 

## Todos
- [x] Use `pybind11_add_module` to declare the Python module at the CMake level. This follows suggestions from  the pybind11 project. Note that I had to deactivate the link-time optimization (the `NO_EXTRAS` flag). I'd get `Aborted (core dumped)` otherwise with GCC (tested with versions 5, 6, and 7)
- [x] More `auto`-izing and switch to the free functions in the `linalg` namespace. 

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libmints/vector.cc
```diff
@@ -265,10 +265,6 @@ void Vector::add(const SharedVector &other) { axpy(1.0, *other.get()); }
 
 void Vector::add(const Vector &other) { axpy(1.0, other); }
 
-void Vector::subtract(const SharedVector &other) { axpy(-1.0, *other.get()); }
-
-void Vector::subtract(const Vector &other) { axpy(-1.0, other); }
-
```
loriab: I don't remember the whole story, but but I'm willing to give up subtract so long as it doesn't break v2rdm. https://github.com/edeprince3/v2rdm_casscf/blob/4375855616ca440b093b5cde20e45542ece6681d/v2rdm_solver.cc#L1761

robertodr: Mistake on my part, thanks for catching it!

PR #1602, opened 2019-04-18
Fix Bugs in NBOWriter

### LINKED ISSUE #1566: Symmetry Can Mangle NBO Output
The following input 
```
molecule {
   0 1 
 S          0.0904282062       -0.4095883553        0.0000000000
 O          1.3547506908        0.3023925212        0.0000000000
 O         -0.7450974916        0.1927054620        1.2404690507
 H         -0.3631668951        1.0551025998        1.4663407894
 O         -0.7450974916        0.1927054620       -1.2404690507
 H         -0.3631668951        1.0551025998       -1.4663407894
}


set basis cc-pV(D+d)Z

en, wfn = energy('b3lyp', return_wfn = True)
nbo = NBOWriter(wfn)
nbo.write("input.47")
```
produces an absolute nonsense NBO analysis, with heavily delocalized core orbitals, orbital energies over 350, and 1000+ strong donor-acceptor interactions. When you add `symmetry c1`, all is normal again.

I haven't been successful in finding the minimal possible example input that displays the problem. Water, for instance, seems to be just fine.

JonathonMisiewicz:
## Description
This PR fixes three bugs in Psi's NBOWriter.
1. The AO->MO matrix was previously put in the transpose way to how NBO expected it.
2. MO coefficients for non-totally symmetric irrep were transformed from SOs to AOs incorrectly, leading to #1566. Closes #1566.
3. The map from angular momentum functions in Psi to labels for NBO was wrong. For instance, p0 was previously px instead of pz.

Due to point 3, **all previous NBO computations from Psi-generated NBO files are suspect.** I very nearly published wrong results due to that, so I advocate backporting this to 1.3.

**NEW**
Linear dependencies and angular momentum up to _h_ is now handled properly.

## Todos
- [x] Fix critical errors in NBOWriter

## Tests of Correctness
I compared the overlap, density, and Fock matrices for nearly-cis diazene from file47's from Psi and QChem. After accounting for the fact that they order the AOs differently, they match almost perfectly. The density matrix was off in the last decimal place for a few entries. The NAO energies, NAO occupations, and perturbative analysis matched as well.

I repeated for my instigating case, cis-sulfurous acid. QChem matches Psi with symmetry matches Psi without symmetry for NAO energies, NAO occupations, and perturbative analysis. Finally, to verify the molecular orbitals, I had both QChem and Psi report the MOs in the basis of NBOs. Apart from ordering the NBOs differently, the two match to within the last decimal place.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Fixes look so simple in retrospect :-)

Could you take one of your examples and make a test case, please? Not parsing the NBO output and checking the numbers or anything, just running the `NBOWriter` (we've no examples at present) and getting a `reference.dat` of the compliant psi nbo output and a copy of the qchem nbo output into the repository.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -930,42 +930,40 @@ void NBOWriter::write(const std::string &filename) {
...
+    SharedMatrix alphac = wavefunction_->Ca_subset("AO", "ALL");
 
     printer->Printf("\n $LCAOMO");
+
+    // Now write out the coefficients. NBO expects all coefficients
+    // for one MO before writing coefficients for the next MO.
+
     if (wavefunction_->same_a_b_orbs()) {
         for (int i = 0; i < nbf; i++) {
             for (int j = 0; j < nbf; j++) {
```
susilehtola: This code assumes there are no linearly dependent functions!

psi4/src/psi4/libmints/writer.cc
```diff
@@ -930,42 +930,40 @@ void NBOWriter::write(const std::string &filename) {
...
-                printer->Printf("%15.6E", alphac->get(0, i, j));
+                printer->Printf("%15.6E", alphac->get(0, j, i));
             }
         }
         for (int i = 0; i < nbf; i++) {
             for (int j = 0; j < nbf; j++) {
                 count++;
                 if (count % 5 == 1) printer->Printf("\n  ");
-                printer->Printf("%15.6E ", betac->get(0, i, j));
+                printer->Printf("%15.6E ", betac->get(0, j, i));
```
susilehtola: Same here

psi4/src/psi4/libmints/writer.cc
JonathonMisiewicz: Do you have an example offhand? I can try to find one if not.

susilehtola: Pretty easy to find, just pick a molecule and a large, augmented basis set. Here's an example
```
molecule {
0 1
H          0.62810        0.62810        0.62810
C          0.00000        0.00000        0.00000
H         -0.62810       -0.62810        0.62810
H         -0.62810        0.62810       -0.62810
H          0.62810       -0.62810       -0.62810
}

set basis aug-pcseg-3
set s_tolerance 1e-5
energy('scf')
```
output
```
  Minimum eigenvalue in the overlap matrix is 3.7150323933E-07.
  Using Canonical Orthogonalization with cutoff of 1.0000000000E-05.
  Overall, 3 of 281 possible MOs eliminated.
```

susilehtola: For some reason, this is a **really** common bug...

JonathonMisiewicz: When I tried your example, Psi says NBO can't handle g basis functions, while QChem gave NBO the g basis functions _without_ eliminating linear dependencies. NBO handles the g functions and the dependencies just fine, eliminating the dependencies itself.

I'm inclined to let NBO sort out the linear dependencies itself, but teach Psi-NBO how to deal with g functions. Either way, I won't be getting this until tomorrow.

susilehtola: You can remove the g functions from the basis, or compile a bigger version of Psi4. If you run in Q-Chem, you need to set the linear dependence threshold like in Psi4, since it's 1e-7 by default.

Currently, the code is just going to produce nonsense, since if you have linear dependencies, C is Nbf x Nmo, and so the last (Nbf-Nmo) columns of the matrix are going to run over the buffer and give you a segfault, or run and give garbage in the output. The linear dependence is **built in C**.

susilehtola: The issue is pretty simple, just replace the column loop over nbf with a loop over the true number of columns.

JonathonMisiewicz: Thanks for the advice. Everything you said was clear after sleeping on it.

I tried the fix and am running into some problems with NBO syntax... I'll get back to this when the NBO devs get back to me.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -636,15 +636,19 @@ void FCHKWriter::write(const std::string &filename) {
...
+        {1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},        // s
+        {103, 101, 102, 0, 0, 0, 0, 0, 0, 0, 0},  // p
         // z2  xz   yz  x2-y2 xy
-        {255, 252, 253, 254, 251, 0, 0},  // d
+        {255, 252, 253, 254, 251, 0, 0, 0, 0, 0, 0},  // d
         // z(z2-r2), x(z2-r2), y(z2-r2) z(x2-y2), xyz, x(x2-y2), y(x2-y2)
-        {351, 352, 353, 354, 355, 356, 357}  // f
+        {351, 352, 353, 354, 355, 356, 357, 0, 0, 0, 0},  // f
+        {451, 452, 453, 454, 455, 456, 457, 458, 459, 0, 0},  // g
+        {551, 552, 553, 554, 555, 556, 557, 558, 559, 560, 561}  // h
```
susilehtola: Switch to ```std::vector<int>```?

JonathonMisiewicz: Python spoiled me, so my C++ typing is weak, but wouldn't it be`const std::vector<std::vector<int> >`? Or am I misunderstanding C++ typing? Either way, I agree that the zeros are avoidable ugliness.

susilehtola: Right, this is a two-dimensional array, so yes: ```std::vector< std::vector<int> >```. There's also the `pure_order` array that should be changed.

Similarly, you could change the 
```
map = new int[nmo];
auto *skip = new bool[nmo];
```
to using `std::vector` to avoid manual delete statements.

JonathonMisiewicz: I'm fine with the std::vector switch, but changes to MOWriter are outside the scope of this PR.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -795,6 +799,9 @@ void NBOWriter::write(const std::string &filename) {
         }
         printer->Printf("%15.6E", exponents.get(0, i));
     }
+    // TODO: Rewrite to eliminate code duplication. The only thing changing between
```
PeterKraus: Rather than adding a TODO, I feel like it's not that much work to roll this up into a loop; who knows how long the C++ version of NBOWriter will remain around.

I suppose the iteration over the names can be done using something like:
```
char coeff_labels[] = { 'S', 'P', 'D', 'F', 'G', 'H' }; 
for (int am = 0; i < basisset.max_am(); am++) {
    ...
    printer->Printf("\n      C%s =", coeff_labels[am]);
    ...
}
```

psi4/src/psi4/libmints/writer.cc
```diff
@@ -636,15 +636,19 @@ void FCHKWriter::write(const std::string &filename) {
 NBOWriter::NBOWriter(std::shared_ptr<Wavefunction> wavefunction) : wavefunction_(wavefunction) {}
 
 void NBOWriter::write(const std::string &filename) {
-    int pure_order[][7] = {
-        {1, 0, 0, 0, 0, 0, 0},        // s
-        {101, 102, 103, 0, 0, 0, 0},  // p
+    const std::vector<std::vector<int>> pure_order = {
```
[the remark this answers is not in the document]
JonathonMisiewicz: They're how NBO reads in orbital angular momentum, see p. 112-3 http://nbo7.chem.wisc.edu/nboman.pdf

First number = l
Second number = Cartesian or angular?
Third number = What projections of the spherical harmonics are we adding together?

psi4/src/psi4/libmints/writer.cc
```diff
@@ -799,68 +799,21 @@ void NBOWriter::write(const std::string &filename) {
...
+    std::string coeff_labels[] = {"S", "P", "D", "F", "G", "H"};
+    for (int am = 0; am <= basisset.max_am(); am++) {
+        for (int i = 0; i < nprim; i++) {
+            if ((i + 1) % 4 == 1) {
+                if (i == 0)
+                    printer->Printf("\n      C%s =", coeff_labels[am].c_str());
+                else
+                    printer->Printf("\n          ");
+            }
+            printer->Printf("%15.6E", coefficient.get(0, am, i));
```
susilehtola: Talk about prettier and more compact code!

PeterKraus: @JonathonMisiewicz just to double check before merging - as the `nbo` test is not ran by CI, it still passes after these changes, right?

psi4/src/psi4/libmints/writer.cc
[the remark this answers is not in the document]
susilehtola: I think you're missing a space, it should be `> >`.

psi4/src/psi4/libmints/writer.cc
susilehtola: I was wondering the same thing - I can see a changed input

psi4/src/psi4/libmints/writer.cc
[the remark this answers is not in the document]
andysim: @susilehtola not any more: that requirement was relaxed in C++11

psi4/src/psi4/libmints/writer.cc
JonathonMisiewicz: It passes in the sense that it reproduces the `output.nbo` file, except for the last line about execution time, which is variable between runs. That is technically a different thing than the test passing. `nbo` just tests that Psi does not crash after `NBOWriter.write()`, per @loriab's instructions.

PeterKraus: Great, thanks! Just wanted to make sure we didn't forget anything.

PR #1607, opened 2019-04-22
explicitly np.load(..., allow_pickle=True) since we're building objects

loriab:
Need to explicitly allow pickles (which we need to since building objects) in NumPy 16.3. See https://github.com/numpy/numpy/issues/12759 . @dgasmith, needed [here](https://github.com/psi4/psi4/blob/fbb2ff444490bf6b43cb6e027637d8fd857adcee/psi4/driver/p4util/numpy_helper.py#L385), too? quicktests passes w/o. This should fix travis.

## Checklist
- [x] quicktests passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -153,10 +153,10 @@ def _core_wavefunction_from_file(wfn_data):
     elif isinstance(wfn_data, str):
         if not wfn_data.endswith(".npy"):
             wfn_data = wfn_data + ".npy"
-        wfn_data = np.load(wfn_data).item()
+        wfn_data = np.load(wfn_data, allow_pickle=True).item()
     else:
         # Could be path-like or file-like, let `np.load` handle it
-        wfn_data = np.load(wfn_data).item()
+        wfn_data = np.load(wfn_data, allow_pickle=True).item()
 
     # variable type specific dictionaries to be passed into C++ constructor
     wfn_matrix = wfn_data['matrix']


PeterKraus:
The way I read the `_np_read()` function, that really is just numpy arrays (not classes or etc) - not a pickle file; perhaps try adding `allow_pickle=False` and see what breaks?

JonathonMisiewicz:
> The way I read the `_np_read()` function, that really is just numpy arrays (not classes or etc) - not a pickle file; perhaps try adding `allow_pickle=False` and see what breaks?

`allow_pickle=True` was the NumPy default behavior, so this PR just maintains that for Psi. `allow_pickle=False` is the new default behavior, which breaks Psi. See this Travis to log to see the test cases that fail: https://travis-ci.org/psi4/psi4/jobs/523119052

PeterKraus:
> > The way I read the `_np_read()` function, that really is just numpy arrays (not classes or etc) - not a pickle file; perhaps try adding `allow_pickle=False` and see what breaks?
> 
> `allow_pickle=True` was the NumPy default behavior, so this PR just maintains that for Psi. `allow_pickle=False` is the new default behavior, which breaks Psi. See this Travis to log to see the test cases that fail: https://travis-ci.org/psi4/psi4/jobs/523119052

Yes, but that's crashing because of `_core_wavefunction_from_file()`, not `_np_read()`, right? Raw log shows that 14 tests fail because of the former, and none because of the latter?

loriab:
Yes, I think you're both right. the numpy_helpers `np.load` is serializing ordinary floats for Matrix/Vector, so np machinery, not pickle, used. Then the python_helpers `np.load` is forming the custom Matrix/Vector objects themselves, so they do need pickle True, and >=16.3 need it explicitly.

PeterKraus:
I think the `np-array-interface` should be checking for this already: it is hitting `Matrix.np_read()` after writing it into file.

JonathonMisiewicz:
I'm having a hard time following all this, so let me try to make sure I'm understanding your concerns about `driver/p4util/numpy_helper.py:_np_read`. I think you're suggesting we explicitly mark `allow_pickle=False` for that function because
1. security reasons mean `allow_pickle` should be `False` unless needed to be `True`
2. `allow_pickle` is not needed to be `True` for that function (as demonstrated by `np-array-interface` passing, even with Travis passing in `False`)
3. some users may compile with pre-16.3 NumPy, where default definitions would make `allow_pickle=True`, which is bad by (1) and (2)

Did I get that right?

PR #1490, opened 2019-01-22
add total_disp function to fisapt.cc

jwaldrop107:
Add the option for SAPT to compute the second order exchange-dispersion term without the Single-Exchange Approximation.

## Todos
* __User-Facing for Release Notes__
    - [ ] Add Sinf option to second order exchange-dispersion

PR #1608, opened 2019-04-26
Fisapt extern

chapincavender:
## Description
Include external potentials in the FISAPT module

## Todos
- [x] Add interaction of external potential with nuclei to nuclear repulsion
- [x] Add interaction of external potential with basis functions to one-electron Hamiltonian for SCF calculations and calculation of HF interaction energy
- [x] Fixed minor typo in proc_util.py

## Questions
- [ ] I added a method to the Wavefunction class to access the protected external_pot_ object. Was there some other way I could have accessed this object?
- [ ] Adding the external potential directly to the one-electron potential matrices for each fragment leads to this potential being double-counted in some terms used to calculate the HF interaction energy. So I stored the external potential as an additional matrix, matrices_["VE"], that is then added to the one-electron potential for the monomer SCF calculations and the HF term. Is there a better way to handle this problem that doesn't require storing an additional matrix?
- [ ] I compared total SAPT0 energy for a 2-fragment F-SAPT calculation to the SAPT module. Is there a good way to check a 3-fragment I-SAPT energy with the SAPT module?

## Checklist
- [x] For a 2-fragment system, total SAPT0 energies with and without an external potential in FISAPT module agree with SAPT module
- [x] For a 3-fragment system, total SAPT0 energies with and without an external potential give reasonable energies
- [x] Passed included smoke and sapt tests

## Status
- [x] Ready for review
- [ ] Ready for merge


CDSherrill:
Very happy to see this, it will be a nice addition!

Maybe @andysim and/or @robertodr can check some of the external potential
handling stuff.

My group and I will look into some of the subtleties of FSAPT and ISAPT ...
off the top of my head, I think there may be some additional places
(induction) where the external potential needs to go.  Maybe also the
dressing on the two-electron integrals.

Best,
David

andysim:
This looks good to me - great work figuring this stuff out.  I think the long-term plan will be to get all of the external perturbations into `MintsHelper` to avoid this kind of problem.  However that's beyond the scope of this pull request.  It would be nice to have a test case to prevent any regressions of this behavior: could you add a small molecule test (like the details you put on the forum post), by doing the following, please?

1)  Make a copy of an existing test folder, such as [this](https://github.com/psi4/psi4/tree/master/tests/fsapt1) and call it something like `fisapt-ext`.

2)  In the CMakeLists.txt in the newly created `fisapt-ext`,  edit the name `fsapt1` to read `fsapt-ext` in the two locations where it is present.

3) Modify the `input.dat` to represent a small system with a small basis set and populate the `Eref` dictionary with values from running that system with your updated, working psi4 version: these will be checked against the computed values when the test is run

4) If the test takes less that about 30 seconds (I doubt it will, but that depends on the size of system) then please add `quicktests` to the list of labels on the third line of `CMakeLists.txt` and remove the `PROPERTIES COST 400` from the following line: this will allow the test to be run every pull request.  If not, it's too expensive to run every time so it'll just be run during our nightly tests instead.

5)  When you have a working test case, please run it and generate the output file as `output.ref`, so that we have something to look at if any changes break this feature in the future.

6) Add the new `fsapt-ext` test name [here](https://github.com/psi4/psi4/blob/master/tests/CMakeLists.txt#L60) so that our testing framework knows about it.

I don't think you need the README or the `fsapt-ref` folder that are in that test directory, but I'm happy to be correct by other devs who know more about SAPT than I do.  Thanks again for this change - it's a very important addition.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -1096,6 +1158,7 @@ void FISAPT::dHF() {
     LH_BC->copy(T);
     LH_BC->add(V_B);
     LH_BC->add(V_C);
+    if (reference_->external_pot()) LH_BC->add(matrices_["VE"]);
```
dgasmith: Ah, this kind of arithmetic makes the external potential tricky. If we added it to `V` we would need the following extra matrices:
```
V_AB
V_AC
V_BC
```

I am actually leaning a bit towards this as future perturbations would be simpler. Any thoughts here?

chapincavender:
I've added a new test called fsapt-ext that uses a water trimer, performing F-SAPT on two waters and representing the third as an external potential with TIP3P charges. The test runs in less than a minute on my machine. I've left the fsapt-ref folder as this is part of the output of the FISAPT module. I also verified that the test breaks without my revisions.

psi4/src/psi4/fisapt/fisapt.cc
chapincavender: This makes a lot of sense to me. It avoids a lot of unnecessary if statements. Should I change to doing things this way?

loriab:
You have hit The Rebase Error! Fear not, we can solve this. 

Follow the advice in https://github.com/psi4/psi4/pull/1490#issuecomment-457765390. In your case, you'll only want the pure-you commits: 1e5cf3a, b9c1a00, 11307b7, 50d44d6 in that order. Post here if you've any questions.

psi4/src/psi4/fisapt/fisapt.cc
dgasmith: I think so, lets do it this way unless there are any blockers. @psi4/admins

chapincavender:
Sorry about that! Looks like everything is back to normal. Thanks for clear instructions on the fix.

loriab:
Looks great again -- thanks for healing the branch. Glad to know those instructions are transferable. And no worries -- `git` has led everyone astray.

psi4/src/psi4/fisapt/fisapt.cc
chapincavender: On second thought, one problem with this approach is that the functions that compute the terms in the SAPT energy, e.g. elst(), also use the VA and VB matrices. If we add the external potential to VA and VB, it will be included in those SAPT energies. This seems to disagree with how the external potential is handled in Ed's SAPT module, which accounts for it in the HF energy through the SCF module but uses the molecule's one-electron potential without the external potential to compute the SAPT terms. Indeed, after making this change (I haven't committed it yet), the FISAPT module's total SAPT0 energy disagrees with the SAPT module's total SAPT0 energy. If including the external potential in the SAPT terms is the right thing to do, we also need to change this in the SAPT module. What do you think?

dgasmith: I think you are right. It would be a good to have a central place, but this would generate even more intermediates. I don't suppose you see a better solution than the current code here?

chapincavender: One option is to define V_AB, V_AC, and V_BC locally within the dHF() function. This would still require keeping matrices_["VE"] as a global matrix that's added to the appropriate potential matrix in the call to FISAPTSCF and within dHF(), but at least the change would be in one place in dHF().

PR #1550, opened 2019-02-20
Cleanup response code

andysim:
## Description
The Hessian code can be cleaned up by using the SCF class's hessian-vector products.  To avoid the need to cast pointers inside the orbital response code, this PR specializes the response class for each reference type, and removes some junk symmetry code.

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] Specialize response code by reference
- [x] Simplify dipole derivative handling
- [x] Remove symmetry code from response
- [x] Use the `Hx` functions from the SCF class to solve CPHF

## Checklist
- [ ] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #1609, opened 2019-04-26
Hessian CPHF options

loriab:
## Description
#1550 switched the CPHF solver for analytic Hessians from libfock to the wfn one. That switched the CPHF from 1e-6 to 1e-4 and the HOOH_TS-analytic test case in test_vibanalysis took exception. This PR sends the relevant `CPHF` module (aka libfock) options to the Hessian CPHF call. So returns the default to 6. Thanks to @andysim for difference-hunting.

## Questions
- [ ] user set-able via `set solver_convergence 5`. but `set cphf solver_convergence 5` has no effect b/c "SCF" module is the active options set at this time. I think that calls for a greater overhaul than this fix.

## Checklist
- [x] full tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
whoops, copy-paste fail. SQUASH, please.

PR #1612, opened 2019-04-28
ambit v0.3, parallel pytest, binder start

loriab:
## Todos
- [x] bump ambit to v0.3
- [x] enable parallel pytest for CI
- [x] `environment.yaml` to build binder from
- [x] install qcng scripts to the right place

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,7 @@ install:
       pylibefp
       snsmp2
       mp2d
+      pytest-xdist
       -c psi4/label/dev
       # * even if name package removed later, this installs their deps
   - >
@@ -110,7 +111,7 @@ script:
 - ./stage/bin/psi4 ../tests/tu1-h2o-energy/input.dat
 - python ../.scripts/travis_run_test.py
 - python ../.scripts/travis_print_failing.py
-- PYTHONPATH=stage/lib/ pytest -v -rws --durations=15 -m 'quick' stage/lib/psi4/tests/
+- PYTHONPATH=stage/lib/ pytest -v -rws --durations=15 -n 2 -m 'quick' stage/lib/psi4/tests/
 
 # safelist
 branches:
--- a/external/upstream/qcengine/CMakeLists.txt
+++ b/external/upstream/qcengine/CMakeLists.txt
@@ -29,6 +29,6 @@ else()
         INSTALL_COMMAND ${PYTHON_EXECUTABLE} setup.py install
                                                       --record=record.txt
                                                       --single-version-externally-managed
-                                                      --install-bin=${_install_bin}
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -110,7 +111,7 @@ script:
 - ./stage/bin/psi4 ../tests/tu1-h2o-energy/input.dat
 - python ../.scripts/travis_run_test.py
 - python ../.scripts/travis_print_failing.py
-- PYTHONPATH=stage/lib/ pytest -v -rws --durations=15 -m 'quick' stage/lib/psi4/tests/
+- PYTHONPATH=stage/lib/ pytest -v -rws --durations=15 -n 2 -m 'quick' stage/lib/psi4/tests/
```
PeterKraus: Why a hardcoded `-n 2` here and an autodetect `-n $NUMBER_OF_PROCESSORS` for azure below?

andysim: That might be left over from the Travis stuff that I put in place in the early days of the GitHub workflow.  At the time my reasoning was that we know the [number of cores available on Travis is 2](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) but I think you're right. If there's an easy way to detect the cores available that would future proof the CI stuff and allow us to benefit from any upgrades to the CI runners.

loriab: Following https://github.com/psi4/psi4/blob/master/.scripts/travis_build.sh#L5 for the former. I just looked it up and travis isn't providing any additional guidance. Azure is more modern, and Raimondas set it up nicely.

The big win for safe `pytest -nN` is in the conda build, where `-n12` takes the full pytest down to 10 min.

andysim: It looks like there is no equivalent default [environmental variable on Travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) though :(  There is an entertaining `HAS_JOSH_K_SEAL_OF_APPROVAL` variable that's always true: apparently Josh K was a co-founder of Travis and this is a way to detect whether the current build is running on Travis infrastructure.

external/upstream/qcengine/CMakeLists.txt
```diff
@@ -28,5 +29,6 @@ else()
         INSTALL_COMMAND ${PYTHON_EXECUTABLE} setup.py install
                                                       --record=record.txt
                                                       --single-version-externally-managed
+                                                      --install-scripts=${_install_bin}
```
dgasmith: Ah, nice!

.travis.yml
robertodr: Maybe `-n $(nproc)` could work?

PR #1617, opened 2019-04-30
Fix lgtm warnings in libmoinfo

fevangelista:
## Description
Fix lgtm warnings in libmoinfo.

## Todos
- [x] Fixed printing and casting to `size_t`

## Status
- [x] Ready for review
- [x] Ready for merge


dgasmith:
I think it's slightly safer as these are evaluated left to right. It shouldn't be a problem, but good practice to be consistent throughout the whole code base to cast on the left.

PR #1619, opened 2019-04-30
LGTM fixes for libqt

jeffschriber:
Fixed timer.cc and blas_intfc.cc

- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libqt/timer.cc
+++ b/psi4/src/psi4/libqt/timer.cc
@@ -1017,6 +1017,7 @@ void timer_done() {
     printer->Printf("Host: %s\n", host);
     free(host);
     printer->Printf("\n");
+    // NOTE: ctime is not thread-safe and could potentially cause overwriting of the return strings. 
     printer->Printf("Timers On : %s", ctime(&timer_start)); // lgtm[cpp/potentially-dangerous-function] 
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/libqt/timer.cc
```diff
@@ -1017,9 +1017,9 @@ void timer_done() {
     printer->Printf("Host: %s\n", host);
     free(host);
     printer->Printf("\n");
-    printer->Printf("Timers On : %s", ctime(&timer_start));
+    printer->Printf("Timers On : %s", ctime(&timer_start)); // lgtm[cpp/potentially-dangerous-function]
```
dgasmith: In any idea how this is potentially dangerous?

jeffschriber: LGTM flagged this because `ctime` is not thread safe. Their suggestion was to use `ctime_r`,  but that failed for the windows builds.

dgasmith: Interesting, if you have a moment could you add a comment to that effect in the code base?

jeffschriber: Sure thing

PR #1623, opened 2019-04-30
LGTM for libsapt_solver

dsirianni:
## Description
Efforts to LGTM-ize `libsapt_solver`.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/fisapt/fisapt.cc
+++ b/psi4/src/psi4/fisapt/fisapt.cc
@@ -3739,8 +3739,8 @@ void FISAPT::find() {
         // => JK Object <= //
 
         // TODO: Account for 2-index overhead in memory
-        long int nso = primary_->nbf();
-        long int jk_memory = (long int)doubles_;
+        auto nso = primary_->nbf();
+        auto jk_memory = (long int)doubles_;
         jk_memory -= 24 * nso * nso;
         jk_memory -= 4 * na * nso;
         jk_memory -= 4 * nb * nso;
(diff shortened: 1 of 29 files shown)


psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -194,9 +194,8 @@ void FISAPT::partition() {
         double delta = options_.get_double("FISAPT_CHARGE_COMPLETENESS");
         outfile->Printf("    Charge Completeness = %5.3f\n\n", delta);
         for (int a = 0; a < na; a++) {
-            if (QFp[0][a] > delta) {
-            } else if (QFp[1][a] > delta) {
-            } else if (QFp[2][a] > delta) {
+            if (QFp[0][a] > delta || QFp[1][a] > delta || QFp[2][a] > delta) {
+                continue;
```
loriab: I think you have to be careful using `continue;` like the python `pass`. It works here b/c end of `for` loop is right after end of `if` sequence but if there had been anything in between, it would have been unintentionally skipped.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -3739,7 +3739,7 @@ void FISAPT::find() {
         // => JK Object <= //
 
         // TODO: Account for 2-index overhead in memory
-        int nso = primary_->nbf();
+        long int nso = primary_->nbf();
```
loriab: `size_t` not `long int`, right?

andysim: I agree with lori here.  Better still:
```suggestion
        auto nso = primary_->nbf();
```

Issue #1615, opened 2019-04-30
LGTM Day modules


loriab:
Sign up with name, check off when complete
- [ ] adc
- [ ] ambit_interface
- [x] cc/ccdensity @jeffschriber #1632
- [x] cc/ccenergy - @fevangelista #1622
- [x] cc/cceom - @fevangelista #1630
- [x] cc/cchbar @jeffschriber #1626
- [x] cc/cclambda @jeffschriber #1634
- [ ] cc/ccresponse
- [ ] cc/cctriples
- [ ] cctransort
- [x] dcft - @JonathonMisiewicz partial #1624
- [x] detci - @loriab #1631
- [x] dfep2
- [x] dfmp2 @obrien951 @derekmetcalf #1633 #1628
- [ ] ~dfocc~
- [x] dmrg
- [ ] fisapt - @zachglick 
- [ ] fnocc
- [x] gdma_interface
- [ ] lib3index @derekmetcalf #1633 #1628
- [x] libciomr #1636
- [x] libcubeprop
- [x] libdiis @derekmetcalf #1633 #1628
- [x] libdisp
- [x] libdpd @jeffschriber #1625
- [x] libfilesystem
- [ ] libfock @obrien951 #1635 
- [x] libfunctional @obrien951 #1633 #1628
- [x] libiwl
- [ ] libmints - @jturney #1616 
- [x] libmoinfo - @fevangelista #1617
- [ ] liboptions - @loriab #1621
- [x] libplugin
- [x] libpsi4util
- [x] libpsio - @fevangelista #1618
- [x] libpsipcm
- [x] libqt @jeffschriber #1619
- [x] libsapt_solver @zachglick @dsirianni #1623
- [x] libscf_solver @JonathonMisiewicz #1629
- [x] libtrans @JonathonMisiewicz #1627
- [ ] ~mcscf~
- [x] mrcc
- [ ] ~occ~
- [ ] ~optking~
- [x] psimrcc
- [x] sapt @zachglick 
- [x] scfgrad @alenaizan #1620
- [x] core #1636

changing type to `size_t`
* `BasisSet.nbf_` int --> size_t
* `Wavefunction.nso_` int --> size_t

JonathonMisiewicz:
Example of fixing the dreaded "Multiplication result may overflow 'int' before it is converted to 'size_t'."
Replace
```
    nElements += R.params->coltot[h] * R.params->rowtot[h];
```
with
```
[...]

PR #1624, opened 2019-04-30
dcft LGTM: Pt. I

JonathonMisiewicz:
Part 1 of the dcft segment of #1615. `ctest -R dcft -j4` passes. Most of the LGTM issues in `dcft` are _not_ addressed in this PR, as the fixes belong elsewhere:
* Changing the type of basisset.nbf and wavefunction.nso from ints to size_t
* Changing Dimension from std::vector<int> to std::vector<size_t>
* Changing rowtot and coltot in dpd.h to be arrays of size_t instead of ints. This one may be more controversial.

### DIFF
--- a/psi4/src/psi4/dcft/dcft_scf_UHF.cc
+++ b/psi4/src/psi4/dcft/dcft_scf_UHF.cc
@@ -238,8 +238,8 @@ void DCFTSolver::print_orbital_energies() {
     std::sort(aPairs.begin(), aPairs.end());
     std::sort(bPairs.begin(), bPairs.end());
 
-    int *aIrrepCount = init_int_array(nirrep_);
-    int *bIrrepCount = init_int_array(nirrep_);
+    auto aIrrepCount = std::vector<int>(nirrep_, 0);
+    auto bIrrepCount = std::vector<int>(nirrep_, 0);
     std::vector<std::string> irrepLabels = molecule_->irrep_labels();
 
     outfile->Printf("\n\tOrbital energies (a.u.):\n\t\tAlpha occupied orbitals\n\t\t");
@@ -290,8 +290,6 @@ void DCFTSolver::print_orbital_energies() {
         Ca_->print();
         Cb_->print();
     }
-    free(aIrrepCount);
-    free(bIrrepCount);
 }
 
 /**
--- a/psi4/src/psi4/dcft/dcft_scf_UHF.cc
+++ b/psi4/src/psi4/dcft/dcft_scf_UHF.cc
@@ -290,8 +290,8 @@ void DCFTSolver::print_orbital_energies() {
         Ca_->print();
         Cb_->print();
     }
-    delete[] aIrrepCount;
-    delete[] bIrrepCount;
+    free(aIrrepCount);
+    free(bIrrepCount);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


psi4/src/psi4/dcft/dcft_scf_UHF.cc
```diff
@@ -290,8 +290,8 @@ void DCFTSolver::print_orbital_energies() {
         Ca_->print();
         Cb_->print();
     }
-    delete[] aIrrepCount;
-    delete[] bIrrepCount;
+    free(aIrrepCount);
+    free(bIrrepCount);
```
loriab: strange, it was already `free` in two other files. wonder how this one was missed.

robertodr: I am pretty sure we want `delete[]` rather than `free`... I'd rather focus on getting these arrays using `new` or `std::vector` rather that `init_int_array`

JonathonMisiewicz: I don't follow. I think you're saying, "We want to use delete[] instead of free wherever possible because Reason X. Therefore, it would be better to change aIrrepCount and bIrrepCount to being created through new or std::vector rather than through init_int_array, so that delete[] is the correct call." Do I have that right?

robertodr: The arrays are allocated using `malloc` in `init_int_array`. LGTM is right in complaining because the corresponding deallocation should be done using `free`. However, these arrays can be just as easily created either with:
- `std::vector`, which would free resources automatically, or 
- `auto aIrrepCount = new  int[nIrrep_];`, which is the idiomatic C++ construct. The deallocation would then be done with `delete[] aIrrepCount;`.

For primitive types there is no difference, that I know of, between using the `malloc`/`free` couple and the `new[]/delete[]` couple. Just a question of being idiomatic in C++.

`init_int_array` also zeros out the newly allocated vector. This achieves the same: `auto aIrrepCount = std::vector<int>(nIrrep_, 0);`

JonathonMisiewicz: Okay, so why not replace _all_ the `init_int_array` calls in `dcft`? By my count, there are 22 total. (And it looks like some of them are never freed. This includes `aocc_off_` at least and probably others.)

robertodr: I have nothing against the change you propose. Doesn't LGTM complain about the arrays that don't get freed?

JonathonMisiewicz: Apparently not.

The `init_int_array` calls fell into three classes.
1. `aIrrepCount` and friends. Changed to `std::vector`.
2. `aocc_off_` and friends. These are needed by libdpd, which cannot take a `std::vector`, so they are now called by `new int[nirrep_]` syntax. They are also properly deleted.
3. The `dcft_sort_mo_tpdm.cc` functions. This is a complicated enough function that I didn't want to touch it. Especially as I have plans to eliminate this in the near future. I cannot imagine why RHF-DCT needs its own TPDM sort for gradients.

This should satisfy the changes?

robertodr: Thanks for your thorough work!

PR #1625, opened 2019-04-30
LGTM for libdpd

jeffschriber:
Remove LGTM errors from libdpd

- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libdpd/T3_AAA.cc
+++ b/psi4/src/psi4/libdpd/T3_AAA.cc
@@ -167,7 +167,7 @@ void DPD::T3_AAA(double ***W1, int nirreps, int I, int Gi, int J, int Gj, int K,
         W2[Gab] = dpd_block_matrix(F->params->coltot[Gab], virtpi[Gc]);
 
         if (F->params->coltot[Gab] && virtpi[Gc]) {
-            ::memset(W1[Gab][0], 0, F->params->coltot[Gab] * virtpi[Gc] * sizeof(double));
+            ::memset(W1[Gab][0], 0, sizeof(double) * F->params->coltot[Gab] * virtpi[Gc]);
         }
     }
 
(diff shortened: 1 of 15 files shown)


psi4/src/psi4/libdpd/T3_AAA.cc
```diff
@@ -167,7 +167,7 @@ void DPD::T3_AAA(double ***W1, int nirreps, int I, int Gi, int J, int Gj, int K,
         W2[Gab] = dpd_block_matrix(F->params->coltot[Gab], virtpi[Gc]);
 
         if (F->params->coltot[Gab] && virtpi[Gc]) {
-            ::memset(W1[Gab][0], 0, F->params->coltot[Gab] * virtpi[Gc] * sizeof(double));
+            ::memset(W1[Gab][0], 0, sizeof(double) * F->params->coltot[Gab] * virtpi[Gc]);
```
robertodr: I don't understand why this swap would make a difference?

psi4/src/psi4/libdpd/pairnum.cc
```diff
@@ -54,7 +55,7 @@ int DPD::pairnum(string pair) {
             if (v[2] == moSpaces[i]) right = i;
         }
         if (left != right) {
-            throw;
+            throw PSIEXCEPTION("");
```
robertodr: Maybe put a message here?

psi4/src/psi4/libdpd/T3_AAA.cc
loriab: the hypothesis is that lgtm works from language rules and implicit casts multiplication left-to-right, rather than compiler practice, which doesn't care which position the big type is at. we'll have to merge one of these to see how lgtm responds.

psi4/src/psi4/libdpd/pairnum.cc
andysim:
```suggestion
            throw PSIEXCEPTION("Invalid pair " + pair + " encountered in DPD::pairnum().");
```

psi4/src/psi4/libdpd/pairnum.cc
```diff
@@ -66,7 +67,7 @@ int DPD::pairnum(string pair) {
             if (v[3] == moSpaces[i]) right = i;
         }
         if (left != right) {
-            throw;
+            throw PSIEXCEPTION("");
```
andysim:
```suggestion
            throw PSIEXCEPTION("Invalid pair " + pair + " encountered in DPD::pairnum().");
```

PR #1626, opened 2019-04-30
LGTM for cchbar

jeffschriber:
Add a few header guards

- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/cchbar/globals.h
+++ b/psi4/src/psi4/cc/cchbar/globals.h
@@ -26,6 +26,9 @@
  * @END LICENSE
  */
 
+#ifndef PSI4_CCHBAR_GLOBALS_H
+#define PSI4_CCHBAR_GLOBALS_H
+
 /*! \file
     \ingroup CCHBAR
     \brief Enter brief description of file here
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/cc/cchbar/globals.h
```diff
@@ -26,6 +26,9 @@
  * @END LICENSE
  */
 
+#ifndef PSI4_CCHBAR_GLOBALS_H
+#define PSI4_CCHBAR_GLOBALS_H
```
loriab: do we have a pattern for when gets the `psi4` prefix? these are the only ones I see:

```
../psi4/include/psi4/pybind11.h:#ifndef PSI4_CORE_PYBIND11_H_H
../psi4/src/psi4/dfep2/dfep2.h:#ifndef PSI4_GUARD_EPMP2_H
../psi4/src/psi4/libfilesystem/path.h:#ifndef PSI4_CORE_PATH_H
```

robertodr: I do prefer using just `#pragma once` at the top, easier to settle on a name for the `define` when there is not name to provide.

loriab: `#pragma once` discussion for anyone else who wasn't aware of an alternative to header guards https://en.wikipedia.org/wiki/Pragma_once . compilers we're interested in (except cuda) support it.

PR #1628, opened 2019-04-30
Adds lgtm changes to these files.

obrien951:
## Description

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -1158,7 +1158,7 @@ void RDFMP2::form_Pab() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= navir * navir;
+    doubles -= static_cast<unsigned long> (navir) * navir;
     double C = -(double)doubles;
     double B = 4.0 * navir * naux;
     double A = 2.0 * navir * (double)navir;
@@ -1167,7 +1167,6 @@ void RDFMP2::form_Pab() {
     if (max_i <= 0) {
         throw PSIEXCEPTION("Not enough memory in DFMP2");
     }
-    max_i = (max_i <= 0 ? 1 : max_i);
     max_i = (max_i > naocc ? naocc : max_i);
 
     // Blocks
@@ -1350,7 +1349,7 @@ void RDFMP2::form_Pij() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= naocc * naocc;
+    doubles -= static_cast<unsigned long> (naocc) * naocc;
     double C = -(double)doubles;
     double B = 2.0 * naocc * naux;
     double A = 2.0 * naocc * (double)naocc;
@@ -1359,7 +1358,6 @@ void RDFMP2::form_Pij() {
     if (max_a <= 0) {
         throw PSIEXCEPTION("Not enough memory in DFMP2");
     }
-    max_a = (max_a <= 0 ? 1 : max_a);
     max_a = (max_a > navir ? navir : max_a);
 
     // Blocks
@@ -1728,7 +1726,7 @@ void RDFMP2::form_Amn_x_terms() {
 
 // > Integrals < //
 #pragma omp parallel for schedule(dynamic) num_threads(num_threads)
-        for (long int PMN = 0L; PMN < NP * npairs; PMN++) {
+        for (long int PMN = 0L; PMN < static_cast<long> (NP) * npairs; PMN++) {
             int thread = 0;
 #ifdef _OPENMP
             thread = omp_get_thread_num();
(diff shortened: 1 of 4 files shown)


psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1158,7 +1158,7 @@ void RDFMP2::form_Pab() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= navir * navir;
+    doubles -= static_cast<unsigned long> (navir) * navir;
```
loriab: `navir` is an `int`. do we want to settle on `size_t` in this case?

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1167,7 +1167,6 @@ void RDFMP2::form_Pab() {
     if (max_i <= 0) {
         throw PSIEXCEPTION("Not enough memory in DFMP2");
     }
-    max_i = (max_i <= 0 ? 1 : max_i);
```
loriab: everyone agree that this line is unneeded after the `throw` above, or am I missing something?

psi4/src/psi4/dfmp2/mp2.cc
robertodr: I think subtracting from `size_t` could lead either to undefined behavior or nasty surprises. Probably use a `long int`?

psi4/src/psi4/dfmp2/mp2.cc
andysim: I agree!

psi4/src/psi4/dfmp2/mp2.cc
andysim: @robertodr I don't think that helps here - it would have to be the type of `doubles` that gets changed to a signed quantity, right?  I think the best approach is to check that the size of `navir*navir` is less than doubles, before doing the subtraction, to avoid surprises.  Thoughts?

robertodr: I agree there should be a check that `navir*navir` isn't larger than `doubles`. Maybe `std::ptrdiff_t` is the appropriate type for `doubles` here? https://en.cppreference.com/w/cpp/types/ptrdiff_t

loriab: Stepping back, oughtn't Wfn's `memory_` be a `size_t`?https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.h#L120

Considering that Rob's checking below if the memory appears negative, rearranging types may be interfering.

dgasmith:
Gentle reminder here. Could use a rebase as well.

psi4/src/psi4/dfmp2/mp2.cc
obrien951: What change should be made to this code in the end? I had a hard time drawing a conclusion from this conversation.

psi4/src/psi4/dfmp2/mp2.cc
obrien951: Line 1170 doesn't appear here as it does in the code. Reviewers should take a look at the file to see what's going on.

psi4/src/psi4/dfmp2/mp2.cc
dgasmith:
```C++
size_t navir2 = (size_t)navir * navir;
if ( (navir2) > double) { throw ...}

doubles -= navir2
```

psi4/src/psi4/dfmp2/mp2.cc
dgasmith: Can you explain what you mean by that? "It doesn't appear here as it does in the code" Are changes not checked in?

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1350,7 +1349,7 @@ void RDFMP2::form_Pij() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= naocc * naocc;
+    doubles -= static_cast<unsigned long> (naocc) * naocc;
```
dgasmith: Similar issues as to the above.

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1830,9 +1828,9 @@ void RDFMP2::form_L() {
     // => Memory Constraints <= //
 
     size_t memory = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    memory -= naocc * nso;
-    memory -= navir * nso;
-    memory -= naocc * navir;
+    memory -= static_cast<unsigned long> (naocc) * nso;
+    memory -= static_cast<unsigned long> (navir) * nso;
+    memory -= static_cast<unsigned long> (naocc) * navir;
```
dgasmith: Ditto. Maybe best here is to sum them all together before doing the check.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -238,7 +238,7 @@ void DFHelper::print_header() {
     outfile->Printf("    Schwarz Cutoff:          %11.0E\n", cutoff_);
     outfile->Printf("    Mask sparsity (%%):       %11.0f\n", 100. * ao_sparsity());
     outfile->Printf("    DFH Avail. Memory [GiB]: %11.3f\n", (memory_ * 8L) / ((double)(1024L * 1024L * 1024L)));
-    outfile->Printf("    OpenMP threads:          %11d\n", nthreads_);
+    outfile->Printf("    OpenMP threads:          %11d\n", static_cast<int> (nthreads_));
```
dgasmith: `(int)nthreads_` is fine.

dgasmith:
@obrien951 Can you patch this up for 1.4?

dgasmith:
@obrien951 Lets patch this up or close this.

JonathonMisiewicz:
This PR should be closed. It's superseded by #1635.

PR #1629, opened 2019-04-30
libscf_solver LGTM

JonathonMisiewicz:
The libscf_solver segment of #1615 in what should be its entirety. Mostly changing format strings, but also setting a header guard back to guarding the header, rather than a blank line. Passes quicktests.

### DIFF
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -769,8 +769,7 @@ void HF::form_Shalf() {
             // descending order
             int start_index = 0;
             for (int i = 0; i < dimpi[h]; ++i) {
-                if (S_cutoff < eigval->get(h, i)) { // Include this empty block by analogy of prev. block
-                } else {
+                if (S_cutoff >= eigval->get(h, i)) {
                     start_index++;
                 }
             }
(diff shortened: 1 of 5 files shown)


psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -769,7 +769,7 @@ void HF::form_Shalf() {
             // descending order
             int start_index = 0;
             for (int i = 0; i < dimpi[h]; ++i) {
-                if (S_cutoff < eigval->get(h, i)) {
+                if (S_cutoff < eigval->get(h, i)) { // Include this empty block by analogy of prev. block
```
loriab: what do others think? (a) comment (like above change), (b) `continue`, (c) suppress lgtm, (d) `<` --> `>=`. I'm thinking (d)

andysim: Yup, d) gets my vote

andysim:
```suggestion
                if (S_cutoff >= eigval->get(h, i)) {
```

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -769,7 +769,7 @@ void HF::form_Shalf() {
...
                 } else {
```
andysim:
```suggestion
```

PR #1630, opened 2019-04-30
Fix lgtm issues in cceom

fevangelista:
## Description
Removes lgtm issues with the cceom module.

## Questions
- [ ] Someone from the Crawford group should review the changes to `overlap.cc` to make sure that the information printed is correct.

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/cceom/diag.cc
+++ b/psi4/src/psi4/cc/cceom/diag.cc
@@ -930,10 +930,12 @@ void diag() {
 
         outfile->Printf("\nProcedure converged for %d root(s).\n", num_converged);
         if (num_converged == eom_params.cs_per_irrep[C_irr]) {
+            // all good
         } else if (iter == eom_params.max_iter) {
             outfile->Printf("\nMaximum number of iterations exceeded, ");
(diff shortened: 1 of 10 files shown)


psi4/src/psi4/cc/cceom/diag.cc
```diff
@@ -930,10 +930,12 @@ void diag() {
 
         outfile->Printf("\nProcedure converged for %d root(s).\n", num_converged);
         if (num_converged == eom_params.cs_per_irrep[C_irr]) {
+            // all good
         } else if (iter == eom_params.max_iter) {
             outfile->Printf("\nMaximum number of iterations exceeded, ");
             outfile->Printf("so not all roots converged!\n\n");
         } else if ((params.wfn == "EOM_CC3") && (num_converged == 1)) {
+            // all good
```
dgasmith: What are these comments for?

loriab: lgtm doesn't like blank `if {}` blocks. not sure what the recc. systematic fix should be.

PR #1635, opened 2019-04-30
Lgtm libfock

obrien951:
## Description
Makes changes suggested by lgtm
ATTENTION: Changes local variable names that overlapped with parameter names in 
libfock/hamiltonian.cc

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -1157,8 +1157,8 @@ void RDFMP2::form_Pab() {
 #endif
 
     // Memory
-    size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= static_cast<unsigned long> (navir) * navir;
+    size_t doubles = static_cast<size_t> (options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L);
+    doubles -= static_cast<size_t> (navir) * navir;
     double C = -(double)doubles;
     double B = 4.0 * navir * naux;
     double A = 2.0 * navir * (double)navir;
@@ -1828,9 +1828,9 @@ void RDFMP2::form_L() {
     // => Memory Constraints <= //
 
     size_t memory = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    memory -= static_cast<unsigned long> (naocc) * nso;
-    memory -= static_cast<unsigned long> (navir) * nso;
-    memory -= static_cast<unsigned long> (naocc) * navir;
+    memory -= static_cast<size_t> (naocc) * nso;
+    memory -= static_cast<size_t> (navir) * nso;
+    memory -= static_cast<size_t> (naocc) * navir;
     int max_rows;
     int maxP = ribasis_->max_function_per_shell();
     size_t row_cost = 0L;
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -1158,7 +1158,7 @@ void RDFMP2::form_Pab() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= navir * navir;
+    doubles -= static_cast<unsigned long> (navir) * navir;
     double C = -(double)doubles;
     double B = 4.0 * navir * naux;
     double A = 2.0 * navir * (double)navir;
@@ -1167,7 +1167,6 @@ void RDFMP2::form_Pab() {
     if (max_i <= 0) {
         throw PSIEXCEPTION("Not enough memory in DFMP2");
     }
-    max_i = (max_i <= 0 ? 1 : max_i);
     max_i = (max_i > naocc ? naocc : max_i);
 
     // Blocks
@@ -1350,7 +1349,7 @@ void RDFMP2::form_Pij() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= naocc * naocc;
+    doubles -= static_cast<unsigned long> (naocc) * naocc;
     double C = -(double)doubles;
     double B = 2.0 * naocc * naux;
     double A = 2.0 * naocc * (double)naocc;
@@ -1359,7 +1358,6 @@ void RDFMP2::form_Pij() {
     if (max_a <= 0) {
         throw PSIEXCEPTION("Not enough memory in DFMP2");
     }
-    max_a = (max_a <= 0 ? 1 : max_a);
     max_a = (max_a > navir ? navir : max_a);
 
     // Blocks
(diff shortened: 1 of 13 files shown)


psi4/src/psi4/lib3index/cholesky.cc
```diff
@@ -163,7 +163,7 @@ CholeskyERI::CholeskyERI(std::shared_ptr<TwoBodyAOInt> integral, double schwarz,
     basisset_ = integral_->basis();
 }
 CholeskyERI::~CholeskyERI() {}
-size_t CholeskyERI::N() { return basisset_->nbf() * basisset_->nbf(); }
+size_t CholeskyERI::N() { return ( size_t ) basisset_->nbf() * basisset_->nbf(); }
```
loriab: I think we're going with the `static_cast<type>(tgt)` style of casting. But can hold off b/c this type is changing more fundamentally in libmints.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -238,7 +238,7 @@ void DFHelper::print_header() {
     outfile->Printf("    Schwarz Cutoff:          %11.0E\n", cutoff_);
     outfile->Printf("    Mask sparsity (%%):       %11.0f\n", 100. * ao_sparsity());
     outfile->Printf("    DFH Avail. Memory [GiB]: %11.3f\n", (memory_ * 8L) / ((double)(1024L * 1024L * 1024L)));
-    outfile->Printf("    OpenMP threads:          %11d\n", nthreads_);
+    outfile->Printf("    OpenMP threads:          %11d\n", static_cast<int> (nthreads_));
```
loriab: Better to change the printf code to `%11zu` than cast here.

psi4/src/psi4/libfock/hamiltonian.cc
```diff
@@ -248,20 +248,20 @@ void CISRHamiltonian::product(const std::vector<std::shared_ptr<Vector> >& x,
                 int nvir = Cavir_->colspi()[h ^ symm];
                 for (int i = 0; i < nocc; i++) {
                     for (int a = 0; a < nvir; a++) {
-                        double x = xp[i * nvir + a + offset];
-                        if (x != 0.0 && x != 1.0) {
+                        double local_x = xp[i * nvir + a + offset];
+                        if (local_x != 0.0 && local_x != 1.0) {
```
loriab: So this is where name `x` (a `const` arg to the fn) was being reused. Anything clever that I'm missing by suggesting `local_x` as more readable. And appeasing lgtm

psi4/src/psi4/libfunctional/LibXCfunctional.cc
```diff
@@ -766,7 +766,7 @@ void LibXCFunctional::compute_functional(const std::map<std::string, SharedVecto
                 }
             }
         }
-        if (deriv > 2) {
+        if (deriv > 2) { // lgtm[cpp/constant-comparison]
             throw PSIEXCEPTION("TRYING TO COPMUTE DERIV > 3 ");
```
loriab: may as well fix the `COPMUTE` typo here, too

psi4/src/psi4/libfock/DiskJK.cc
```diff
@@ -75,8 +75,8 @@ void DiskJK::preiterations() {
 
     std::shared_ptr<SOBasisSet> bas = mints->sobasisset();
 
-    so2symblk_ = new int[primary_->nbf()];
-    so2index_ = new int[primary_->nbf()];
+    so2symblk_ = new int[primary_->nbf()]; // lgtm [cpp/resource-not-released-in-destructor]
+    so2index_ = new int[primary_->nbf()]; // lgtm [cpp/resource-not-released-in-destructor]
```
loriab: these are freed in `postiterations()`. would that be a double free if called from the destructor to be sure?

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1158,7 +1158,7 @@ void RDFMP2::form_Pab() {
 
     // Memory
     size_t doubles = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    doubles -= navir * navir;
+    doubles -= static_cast<unsigned long> (navir) * navir;
```
dgasmith: Normally we should cast to `size_t` here and below. I would actually vote to change `doubles` to type double. We would need to check this below, but this will make things safer and prevent some issues with looping around in some severe low memory cases.

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1830,9 +1828,9 @@ void RDFMP2::form_L() {
     // => Memory Constraints <= //
 
     size_t memory = ((size_t)(options_.get_double("DFMP2_MEM_FACTOR") * memory_ / 8L));
-    memory -= naocc * nso;
-    memory -= navir * nso;
-    memory -= naocc * navir;
+    memory -= static_cast<unsigned long> (naocc) * nso;
```
dgasmith: Same here, at the very least `size_t`.

dgasmith:
Gentle poke here, looks almost ready to go with a few comments.

psi4/src/psi4/libfock/DiskJK.cc
obrien951: If the method is called in both, it would be a double-free based on how I can tell the code is written.

Having looked into some of the code that calls these methods, `postiterations()` is a function belonging to the jk object, overridden by each class, that gets called at the end of each scf calculation. We could move the `delete` calls to the destructor to appease, but my guess is that whoever wrote this code did it to match the other methods more closely. 

As far as I can tell, someone with more knowledge than I have should decide whether we want to move the `delete` calls to the destructor to appease LGTM or leave it as is, trusting those who wrote this code to have done it best.

obrien951:
Thanks for the poke. Minimal recommended changes are made, and a question is addressed. Progress is to the point where someone with better knowledge of the code-base should recommend further changes.

psi4/src/psi4/libfunctional/LibXCfunctional.cc
```diff
@@ -767,7 +767,7 @@ void LibXCFunctional::compute_functional(const std::map<std::string, SharedVecto
             }
         }
         if (deriv > 2) { // lgtm[cpp/constant-comparison]
-            throw PSIEXCEPTION("TRYING TO COPMUTE DERIV > 3 ");
+            throw PSIEXCEPTION("TRYING TO COMPUTE DERIV > 3 ");
```
susilehtola: The message is also wrong, it should be `>=3`!

obrien951:
I think that last change should be in there.

dgasmith:
@obrien951 Can you fix tests here so we can merge this in?

psi4/src/psi4/libfock/DiskJK.cc
obrien951: Do we want to move the delete or keep it in post-iterations?

obrien951: @loriab I'm summoning you because I don't expect you to be checking this

PR #1640, opened 2019-05-03
Revert changes to cmake that cause Forte compilation problems on Mac

fevangelista:
## Description
Recent changes to the cmake build systems introduced an issue when compiling psi4 plugins and Forte on a Mac. Compilation fails at the linking stage with the error
```
ld: can't link with bundle (MH_BUNDLE) only dylibs (MH_DYLIB)
```

This PR reverts two changes.

## Questions
- [ ] Is there a better solution that still uses `pybind11_add_module(core MODULE NO_EXTRAS core.cc)`?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/CMakeLists.txt
+++ b/psi4/src/CMakeLists.txt
@@ -1,7 +1,6 @@
 ### >> Create core target, so we may link modules to it directly <<
 ### >> This allow to propagate dependencies upwards sanely
 add_library(core SHARED core.cc)
-#pybind11_add_module(core MODULE NO_EXTRAS core.cc)
 
 ### >> Go into psi4 subdirectory to compile libraries and modules <<
 add_subdirectory(psi4)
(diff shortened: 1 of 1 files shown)


fevangelista:
No problem, this is a simple enough hack that we can apply locally for now. Perhaps the PR will evolve into a better solution once we investigate more.

PeterKraus:
I guess if there are near-future plans to find a better solution, then there's no point merging, right?

dgasmith:
Probably want a @loriab review on this as well before merging.

loriab:
@robertodr, was using `MODULE` instead of `SHARED` deliberate? https://github.com/pybind/pybind11/blob/master/tools/pybind11Tools.cmake#L116

I confirm that v2rdm won't build in psi4's current state with same module error.

@fevangelista, there's a new ambit 0.3 conda pkg available for Mac now.

robertodr:
I was following the CMake manual: "SHARED libraries are linked dynamically and loaded at runtime. MODULE libraries are plugins that are not linked into other targets but may be loaded dynamically at runtime using dlopen-like functionality." Guess I was wrong? Maybe @fevangelista can test that out?

loriab:
testing the `MODULE` --> `SHARED` on v2rdm now

PeterKraus:
Closing as #1642 resolves the issue.

PR #1644, opened 2019-05-13
qcelemental for alignment, which, testing

loriab:
## Description
Rebase through of qcel v0.4.0 (not yet released)

## Todos
- [x] moved `tests/pytest` to `tests/pytests` so latter can be a module w/o confusing "import pytest"
- [x] migrated to use which, which_import, parse_version from qcl
- [x] removed alignment code that migrated to QCElemental

- [x] derived testing fns from qcel, rather than scattering and duplicating in psi4/qcdb
  * there's two conflicting APIs going on:
    * Psi4: `compare_something(expected, computed, significant_digits, label, verbose=1)`
    * QCEl: `compare_something(expected, computed, label=None, atol=1.e-6, quiet=False)`
  * and there's two sets of fn names:
    * Psi4: `compare_integers`, `compare_strings`, `compare_arrays`, `compare_values`, `compare_matrices`
    * QCEl: `compare` (for exact comparison scalars or np-castable arrays), `compare_values` (for floating point comparison scalaurs or np-castable arrays), `compare_recursive` (for iterables containing any of the previous)
  * Basically, now in psi4 or qcdb, you can use the qcel fns and the psi4 fns and the psi4 API or the qcel API. And their failure response is localized in a fn that can be switched out. 
  * A few highly specialized comp fns (like `compare_molrecs` have been deprecated or psi4 API deprecated, but there should be enough UpgradeHelper guides to raise a problem and suggest solution.
  * EDIT: whether using psi4 or qcel APIs, `label` is no longer required -- it'll default to the calling fn name (or `<module>` in a script)

- [x] added link to binder image and fix tutorial cmd that didn't work in binder

## Questions
- [ ] Someone should look at the old/new `compare_matrices` fns to check that the values comparison is still catching any detailed uses of the `^` operator

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


psi4/driver/p4util/fcidump.py
```diff
@@ -342,26 +342,16 @@ def compare_fcidumps(expected, computed, label):
...
-            """Python module deepdiff not found. Solve by installing it: `conda install deepdiff -c conda-forge` or `pip install deepdiff`"""
-        )
-
     # Grab expected header and integrals
     ref_intdump = fcidump_from_file(expected)
     intdump = fcidump_from_file(computed)
 
     # Compare headers
-    header_diff = DeepDiff(
+    compare_recursive(
```
robertodr: Great!

psi4/driver/qcdb/align.py
```diff
@@ -26,148 +26,10 @@
...
-
-    Zidx = list(set(sorted(Zhash)))
-    pZ = [Zidx.index(z) for z in Zhash]
-    return _nre(pZ, geom)
-
-
-#         1         2         3         4         5         6         7         8         9         10
-
 
 def B787(cgeom,
```
PeterKraus: Didn't even realise we had this module available. Great. Why is this fn called "B787" - anything to do with the airplane?

psi4/driver/qcdb/molecule.py
```diff
@@ -1473,7 +1477,7 @@ def to_dict(self, force_c1=False, force_units=False, np_out=True):
             validated_molrec = qcel.molparse.from_arrays(speclabel=False, verbose=0, domain='qm', **molrec)
             forgive.append('fragment_charges')
             forgive.append('fragment_multiplicities')
-        compare_molrecs(validated_molrec, molrec, 6, 'to_dict', forgive=forgive, verbose=0)
+        compare_molrecs(validated_molrec, molrec, 'to_dict', atol=1.e-6, forgive=forgive, verbose=0)
```
PeterKraus: Are we still planning to define "global" tolerances for tests?

loriab: When you use the qcel API (`fn(expected, computed, label=None, kwargs)`), then, yes, there's a global tol of 1.e-6. I actually really like the explicit choice that the psi4 API demands, so I'll keep adding them. Most all fns you can use either API but I figured molrecs was local enough to me to convert.

doc/sphinxman/source/build_planning.rst
```diff
@@ -218,19 +218,24 @@ build system will automatically download and build.
...
 * NumPy http://www.numpy.org/
 
-* networkx https://github.com/networkx/networkx
+* networkx https://github.com/networkx/networkx (transitive dependency of QCElemental)
 
-* deepdiff https://github.com/seperman/deepdiff
+* pint https://pint.readthedocs.io/en/latest/ (transitive dependency of QCElemental)
 
-* pint https://pint.readthedocs.io/en/latest/
+* pydantic https://pydantic-docs.helpmanual.io/# (transitive dependency of QCElemental)
```
dgasmith: `pint` and `pedantic` are required. Only networkx is not.

dgasmith: Going to merge this and pin for a TODO as its not too crucial.

psi4/driver/qcdb/align.py
loriab: Yep, it can be handy. Now lives [here](https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molutil/align.py#L33) but you'll probably want to call it through `psi4.core.Molecule.B787`, `qcdb.Molecule.B787`, `qcel.models.Molecule.align` (I was persuaded to rename the last on the findability grounds). Its definitely good through coupled cluster sized mols. Yes, the airplane -- the function's nickname in the lab is "dreamaligner."

doc/sphinxman/source/build_planning.rst
loriab: true, but psi uses capabilites that demand it. I'll clarify.

PR #1654, opened 2019-05-30
QCAspect --> qcel.Datum in vibrational analysis

loriab:
## Description
Ditch `QCAspect` in favor of its code descendant, `qcel.Datum`. Because vib involves strings and complex numbers, need qcel `v0.4.1` (not released at time of posting).

## Checklist
- [x] full tests run

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/driver.py
```diff
@@ -1556,7 +1556,7 @@ def vibanal_wfn(wfn, hess=None, irrep=None, molecule=None, project_trans=True, p
 
     vibinfo, vibtext = qcdb.vib.harmonic_analysis(
         nmwhess, geom, m, wfn.basisset(), irrep_labels, dipder=dipder, project_trans=project_trans, project_rot=project_rot)
-    vibrec.update({k: qca.to_dict() for k, qca in vibinfo.items()})
+    vibrec.update({k: qca.json() for k, qca in vibinfo.items()})
```
dgasmith: This builds serialized JSON blobs?

loriab: yes, read back in [example here](https://github.com/psi4/psi4/pull/1654/files#diff-ef77f359ab40beab06fac4b16aa64518R110). it's awkward and not final vib schema form, but a step in the right direction.

psi4/driver/qcdb/vib.py
```diff
@@ -58,7 +57,7 @@ def _success(label):
         """Function to print a '*label*...PASSED' line to screen.
         Used by :py:func:`util.compare_values` family when functions pass.
         """
-        msg = '\t{0:.<66}PASSED'.format(label)
+        msg = f'\t{label:.<66}PASSED'
```
PeterKraus: Is this going to be the python string formatting standard we're agreeing on?

psi4/driver/qcdb/vib.py
```diff
@@ -424,13 +423,14 @@ def vec_in_space(vec, space, tol=1.0e-4):
...
-        project_trans, project_rot, nrt, nrt_expected))
+    text.append(
+        f'  projection of translations ({project_trans}) and rotations ({project_rot}) removed {nrt} degrees of freedom ({nrt_expected})'
+    )
 
     P = np.identity(3 * nat)
     for irt in TRspace:
         P -= np.outer(irt, irt)
-    text.append(mat_symm_info(P, lbl='total projector') + ' ({})'.format(nrt))
+    text.append(mat_symm_info(P, lbl='total projector') + f' ({nrt})')
```
PeterKraus: I suppose if we're doing away with `lbl` in `harmonic_analysis`, it wouldn't hurt to `lbl -> label` here too. Perhaps a ToDo item?

psi4/driver/qcdb/vib.py
```diff
@@ -696,81 +696,119 @@ def grouper(iterable, n, fillvalue=None):
             text += """{:^{width}}{:{colsp}}""".format(val, '', width=width, colsp=colsp)
         text += '\n'
 
-        text += """{:{presp}}{:{prewidth}}""".format(
-            '', 'Reduced mass ' + _br(vibinfo['mu'].units), prewidth=prewidth, presp=presp)
+        text += """{:{presp}}{:{prewidth}}""".format('',
```
PeterKraus: Is it impossible to convert this to the PEP498 format? or is `f"...."` incompatible with the `"""...."""` syntax?

psi4/driver/qcdb/vib.py
loriab: There hasn't been a formatting standard btwn `%` and `.format`, partly because the core devs were in two camps. But everyone loves f-strings, so now that we're py36, things are migrating as opportunity allows. Personally, I do f-strings unless the insides are too complicated such that I think format looks cleaner. And I'd still do a simple % if the string involved a lot of curly brackets.

psi4/driver/qcdb/vib.py
loriab: Sure. The other `lbl` was a field in the namedtuple QCAspect, where its counterpart in Datum was `label`, so it _had_ to change. This one's a fairly local argument, but I agree `label` is more readable.

psi4/driver/qcdb/vib.py
loriab: I don't think it's impossible. This is just complicated formatting, and I didn't want to test it again with f-strings. :-)

PR #1663, opened 2019-06-10
IRC: Displace from initial geom correctly

JonathonMisiewicz:
## Description
Last summer, I discovered that the energy _increased_ along the IRC for my complex formation coordinate. I eventually tracked this down to a bug where the molecule did not displace the initial geometry correctly, which this PR fixes. Since that time, those of us at UGA have found this fix essential to follow modes <200i. I discuss this in more detail [here](https://github.com/psi-rking/optking/issues/19).

I updated a test to reflect the fact that the IRC can go a bit further when it displaces correctly. I didn't bother updating the reference files, as I expect it'll be invalidated soon anyways. The whole reason I'm submitting this PR now is to eliminate this bug from consideration when I send @psi-rking the _latest_ IRC bug I found.

## Todos
- [x] Fix IRCs

## Checklist
- [x] All IRC tests run

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/optking/molecule_irc_step.cc
+++ b/psi4/src/psi4/optking/molecule_irc_step.cc
@@ -290,7 +290,7 @@ void MOLECULE::irc_step()
       double *v_m = lowest_evector(H_m, Nintco);
       v = init_array(Nintco);
 
-      // Transform the gradient to internals
+      // Transform the displacement to pivot point to internals
       for(int i=0; i<Nintco; i++)
           v[i] = array_dot(rootG_inv[i], v_m, Nintco);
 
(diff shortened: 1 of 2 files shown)


psi-rking:
Thanks for fixing this.  Please do send any other IRC comments to me ASAP.  I'm working hard on the Python version right now.

psi4/src/psi4/optking/molecule_irc_step.cc
```diff
@@ -283,12 +283,16 @@ void MOLECULE::irc_step()
...
     double *v;
     if (at_TS) {
-      //v = lowest_evector(H_m, Nintco);
-      v = lowest_evector(H, Nintco);
+      double *v_m = lowest_evector(H_m, Nintco);
+      v = init_array(Nintco);
+
+      // Transform the gradient to internals
+      for(int i=0; i<Nintco; i++)
+          v[i] = array_dot(rootG_inv[i], v_m, Nintco);
```
andysim: Shouldn't the gradient be zero, by definition, if we're at a TS?

JonathonMisiewicz:
This should be marked bug fix rather than enhancement. There are transition states that I cannot get an IRC for, without that fix.

Also, corrected a comment per @andysim's observation.

PR #1664, opened 2019-06-14
LSDA analytic Hessians

andysim:
## Description
This is a first step towards DFT Hessians.  The code seems to be correct based on initial tests, but this needs a lot more stress testing.  The emphasis here is correctness over efficiency, and I just want to get all of the nasty CPKS infrastructure in place before worrying about making it fast and memory efficient.  The symmetric stretch of water is off by 2.8cm-1 (less than 1%), but there are a number of possible causes: 1) noise in the finite difference reference, which is unlikely given my experiments with 5 point FDs and different step sizes 2) the lack of grid weight derivatives, also unlikely because the error doesn't decrease when I use a huge grid 3) the filtering of the functional derivatives in areas of small density.  I think number 3) is most likely, but it's also possible that there's something wrong in the code still.

## Todos
- [x] Adds analytic DFT hessians for LSDA functionals (RKS wavefunctions only)

## Questions
- [x] How should we handle the procrouting?  We could explicitly list all LSDA type functionals, but that would not be good for custom functionals.  I'm sure there's a better way that I'm just not familiar with.

## Checklist
- [x] [All or relevant fraction of full tests run]
## Status
- [x] Ready for review
- [x] Ready for merge


andysim:
Also, note the presence of merge commits.  I hit a few conflicts when rebase merging, so I figured it's just easier to squash merge this PR.

loriab:
I think we can add by class like https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc_table.py#L216 . Just have to be careful to use `not` logic because `is_gga` is True for all, rather than being the final classification, as we think of it. https://github.com/psi4/psi4/blob/master/tests/docs-dft/input.dat

psi4/src/psi4/libfock/v.cc
```diff
@@ -714,6 +718,247 @@ void RV::compute_V(std::vector<SharedMatrix> ret) {
...
+                //       /  | ∂^2 F
+                // T <- | ɸ | ----- ρ_z
+                //       \  | ∂ ρ^2
+                std::fill(Tp[P], Tp[P] + nlocal, 0);
+                C_DAXPY(nlocal, -v_rho_aa[P] * w[P] * rho_zP, phi[P], 1, Tp[P], 1);
+                //
+                //       /    | ∂ F
+                // T <- | ɸ_z | ---
+                //       \    | ∂ ρ
+                C_DAXPY(nfuncs, -0.5*v_rho_a[P] * w[P], &phi_z[P][first_func_addr], 1, &Tp[P][first_func_addr], 1);
```
dgasmith: As we move to GGA's we the meta-GGA UKS gradient architecture would be a useful template to shorten this code.

JonathonMisiewicz:
Thirding the enthusiasm for this.

A couple notes:
* As discussed on conference call, `scfgrad` is just fine on its own and shouldn't be affected by the gradient refactoring elsewhere
* Let me know if you continue investigating noise in the FD. I have a lot of experience testing that code.

andysim:
Alrighty, I've figured it out; I was missing yet another pesky CPKS term.  After running this test, all looks great:
```
molecule {
    O            0.000000000000     0.000000000000    -0.076532381406
    H            0.000000000000    -0.769857650400     0.607312215449
    H           -0.000000000000     0.769857650400     0.607312215449
}

dft_grids = [ (75, 302), (90, 434), (99, 590) ]
for radial,spherical in dft_grids:
    set {
      scf_type pk
      d_convergence 12
      #dft_v2_rho_cutoff 1e-16
      dft_spherical_points $spherical
      dft_radial_points    $radial
      points 5
      disp_size 1e-4
      g_convergence gau_tight
    }

    method = 'SVWN/def2-svp'

    optimize(method)
    Eener, Ewfn = freq(method, dertype=0, return_wfn=True)
    Egrad, Gwfn = freq(method, dertype=1, return_wfn=True)
    Ehess, Hwfn = freq(method, dertype=2, return_wfn=True)

    Ewfn.frequencies().print_out()
    Gwfn.frequencies().print_out()
    Hwfn.frequencies().print_out()
```
This tests the frequencies by 5 point finite differences of energies (which includes full grid weight derivatives, as a result of the displacements each rebuilding their grid), by gradients (which includes only the first order grid weight derivative term) and by analytic Hessians (which has no grid weight derivative term coded up).  The results below confirm that the implementation is now correct (well, for water at least):

|Grid    | FD type  |  mode 1   |  mode 2   |  mode 3  |
|--------|----------|-----------|-----------|----------|
|        |   Ener   | 1559.6255 | 3735.3708 | 3841.9257|       
| 75,302 |   Grad   | 1560.1514 | 3735.4746 | 3841.8555|       
|        |   Hess   | 1560.1712 | 3735.5120 | 3841.8152|       
|        |          |           |           |          |  
|        |   Ener   | 1560.2564 | 3735.4425 | 3841.8996|       
| 90,434 |   Grad   | 1560.1980 | 3735.4777 | 3841.7622|       
|        |   Hess   | 1560.2152 | 3735.5349 | 3841.8257|       
|        |          |           |           |          |
|        |   Ener   | 1559.9800 | 3735.4104 | 3841.5086|       
| 99,590 |   Grad   | 1560.2142 | 3735.4799 | 3841.7807|       
|        |   Hess   | 1560.2242 | 3735.5373 | 3841.8279|  

It's really not worth adding the grid weight derivatives, on the basis of this simple test.  We just need to ensure that users don't run analytic Hessians with tiny grids.  There could be a problem with noise if people want to do things like quartic force fields from analytic second derivatives, but those kinds of jobs should be using huge grids anyway.

dgasmith:
Awesome! What can we review here?

andysim:
Realistically I'm not going to have the bandwidth to get full (i.e. production quality) DFT hessians up and running in one shot, so I'm in favor of incremental PRs.  This one does two things: 1) sets up the CPKS stuff we'll need for all restricted DFT methods and 2) paves the way for GGA Hessians, by adding all the appropriate hooks for the various derivative terms.  It currently needs 3 x Natom x nbf^2 memory for the Fock derivatives, but that will be easy to fix in the future.  Therefore, my preference is to get this PR in as-is, so that I (and, hopefully, others) can build on it by adding the GGA terms and making it more memory efficient.  I'd be very happy to guide any interested people on the refactoring front, but will do it myself if needed; either way, it's something that's much easier to do when there's at least something that works in place already.  The main review questions I have were already sorted out in the great feedback I've had so far:  @loriab told me exactly where, and how, to hook into the driver.  You (@dgasmith) have pointed out a good source of inspiration for the refactoring pass, so I think my questions are answered at this point.  Once this is merged, I'll continue to work on the restricted GGA terms so that we're well poised for a brainstorming session to make the code more efficient.

psi4/driver/procrouting/proc_table.py
```diff
@@ -217,7 +217,7 @@
         procedures['gradient'][key] = proc.run_scf_gradient
 
     # Hessians
-    if not ssuper.needs_xc():
+    if not ssuper.is_gga(): # N.B. that meta-GGA classed as GGA and mGGA
```
dgasmith: More that a mGGA is effectively a GGA as well?

psi4/src/psi4/libfock/v.cc
```diff
@@ -714,6 +718,247 @@ void RV::compute_V(std::vector<SharedMatrix> ret) {
...
+            // T = ɸ D, remembering that only the bfs centered on the current atom of interest contribute to the derivative
+            C_DGEMM('N', 'N', npoints, nfuncs, nlocal, 1.0, phi[0], coll_funcs, &Dp[0][first_func_addr], max_functions, 0.0, Tp[0], max_functions);
+            for (int P = 0; P < npoints; P++) {
+                // ρ_y  = T ɸ_y^t
+                double rho_yP = C_DDOT(nfuncs, Tp[P], 1, &phi_y[P][first_func_addr], 1);
+                // Now redefine the intermediate T:
+                //
+                //       /  | ∂^2 F
+                // T <- | ɸ | ----- ρ_y
+                //       \  | ∂ ρ^2
```
dgasmith: Nice diagrams!

psi4/src/psi4/libfock/v.cc
```diff
@@ -1289,53 +1517,140 @@ SharedMatrix RV::compute_hessian() {
...
+                int B = primary_->function_to_center(function_map[nl]);
+                Hp[3 * A + 0][3 * B + 0] += pHXX[ml][nl];
+                Hp[3 * A + 1][3 * B + 0] += pHYX[ml][nl];
+                Hp[3 * A + 2][3 * B + 0] += pHZX[ml][nl];
+                Hp[3 * A + 0][3 * B + 1] += pHXY[ml][nl];
+                Hp[3 * A + 1][3 * B + 1] += pHYY[ml][nl];
+                Hp[3 * A + 2][3 * B + 1] += pHZY[ml][nl];
+                Hp[3 * A + 0][3 * B + 2] += pHXZ[ml][nl];
+                Hp[3 * A + 1][3 * B + 2] += pHYZ[ml][nl];
+                Hp[3 * A + 2][3 * B + 2] += pHZZ[ml][nl];
```
dgasmith: Can we add a small blurb that this is *not* symmetric? Not sure why, but that caught me for a bit.

psi4/src/psi4/scfgrad/response.cc
```diff
@@ -420,8 +420,30 @@ std::shared_ptr<Matrix> RSCFDeriv::hessian_response()
             psio_->write(PSIF_HESS,"Vpi^A",(char*)Vpip[0], static_cast<size_t> (nmo) * nocc * sizeof(double),next_Vpi,&next_Vpi);
         }
     }
+#if 0
```
loriab: debug? will the ACS compile?

andysim: excellent catch!  this is something i stuck in there while debugging (I usually grep for ACS when done).  I completely forgot in this case, but that code should not be in there: those terms are already handled [here](https://github.com/psi4/psi4/blob/42ecb0859b36112f17f021981f4e7d8e18ad86ef/psi4/src/psi4/scfgrad/scf_grad.cc#L1020-L1034).  I'll nuke the offending lines in my cleanup pass

PR #1657, opened 2019-06-05
Computing multipole potential integrals up to desired order

maxscheurer:
## Description
This PR aims to implement computation of multipole potential integrals up to a desired
order. Such integrals are used for EFP and PE (#1401), where for EFP, the integrals
are already implemented, but are always computed through third order (octupoles).

For PE, often parameters are only available up to quadrupoles, so I'd like to avoid computation
of higher order integrals all the time.

~~The code is copied from `efpmultipolepotential`, so probably one should decide on whether
to keep `efpmultipolepotential` or to use the new code to avoid code duplication.~~

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] enable octupole moments as well
- [x] change ordering of cartesian components to Cartesian
- [x] use `multipolepotential` for EFP (`ao_efp_multipole_potential` calls `ao_multipole_potential` and does the remapping of components in `mintshelper.cc`)
- [x] remove `efpmultipolepotential.{hh,cc}`
- [x] tests of the actual integrals using finite differences

## Questions
- [x] ~~Switch to `multipolepotentials` for EFP in another PR?~~

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1671, opened 2019-06-22
Speed up electric field integrals

maxscheurer:
## Description
This PR aims to improve the performance of electric field integrals for multiple
coordinates.

These are needed, e.g., in EFP and Polarizable Embedding (PE, #1401) calculations, where the
integrals need to be evaluated at all sites in the environment. For large environments, the electric field integrals can become the bottleneck of the calculation.

The code I've already implemented is quite ugly and mainly serves as a proof of principle.
Exposure to the Python layer is also not well done at the moment, but it made testing
super easy. In my current [test script](https://gist.github.com/maxscheurer/4736a100ff5de12190ef27f7bc2eaade), I computed integrals at 2000 random points for Water/cc-pVQZ, for which I get a 25x speed-up compared to the existing implementation.

Tasks the implementation enables:
- compute field integrals on multiple points
- contract the integrals with a density matrix (for expectation values of the electric field at all points)
- contract the integrals with a dipole moment (for building the induction operator for PE/EFP)

## Todos
- [x] hackish, preliminary version (proof of principle)
- [x] cleanup `mintshelper` etc.
- [x] proper naming of new functions

## Checklist
- [x] Tests added for any new features (`pytests/test_elec_fields.py`)

## Status
- [x] Ready for review
- [x] Ready for merge


maxscheurer:
> This is awesome! Less code and faster, always nice to see.

Thanks! 😊 
I'll polish the code a little and make the EFP driver code use the new routines.

Another question: The current implementation (both the "new" and "old" field integral code) does not seem to make use of OpenMP parallelism. Is this correct or did I mess something up?

andysim:
The field integrals are not OMP parallel, no (we only ever evaluated them at nuclei before, so they were never a bottleneck).  The easiest way to go is to make `buffer_` into a std::vector of buffers instead (one for each thread) and then use the appropriate one for each thread.  That would probably be best done on a separate PR, but whatever is easiest for you is fine with me

maxscheurer:
> The easiest way to go is to make buffer_ into a std::vector of buffers instead 

Cool. I'm happy to make another PR for this once this PR is polished and merged.
Any place in `libmints` where this is already done with the `std::vector` of buffers (I quickly checked, but could not find anything of the sort...) ?

andysim:
It looks like we do something similar in `MintsHelper::one_body_ao_computer` but, instead of making a list of buffers inside the integral object, a list of integral objects (one for each thread) is expected.  Each integral object has little storage apart from the buffer, so there isn't much extra memory overhead from this approach.  I don't have a strong opinion, but perhaps the vector of integral objects is more consistent than my original suggestion.

maxscheurer:
> It looks like we do something similar in `MintsHelper::one_body_ao_computer`

Thanks @andysim, that's the code I also just found (see my edited comment) 👍 
I'll play around with this a little and see what is a good solution, also with respect to the computation of the electric fields on multiple sites.

psi4/src/psi4/libmints/electricfield.cc
```diff
@@ -214,6 +214,138 @@ void ElectricFieldInt::compute_pair(const GaussianShell& s1, const GaussianShell
...
+
+            double*** ex = efield_recur_.x();
+            double*** ey = efield_recur_.y();
+            double*** ez = efield_recur_.z();
+
+            double** xyz = coords->pointer();
+            int nsites = coords->rowspi()[0];
+
+            for (int site = 0; site < nsites; ++site) {
+                memset(buffer_, 0, 3 * size * sizeof(double));
```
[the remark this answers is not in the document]
maxscheurer: Done.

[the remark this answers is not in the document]
susilehtola: `std::assign` is another possibility, right?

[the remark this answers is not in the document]
maxscheurer: I'm using `std::fill_n` now, if that's okay.

susilehtola:
Stupid question: what is actually the difference to the earlier code? It appears you have rewritten some of the Obara-Saika recursion routines..? Where do the speedups come from?

maxscheurer:
> Stupid question: what is actually the difference to the earlier code? It appears you have rewritten some of the Obara-Saika recursion routines..? Where do the speedups come from?

I did not rewrite any recursion routines (just removed `ObaraSaikaTwoCenterEFPRecursion`, which became obsolete since my recent PR #1657). However, the overhead of having to call `mints.electric_field` several hundred times in a row for each site is eliminated by the new implementation, since the coordinates are directly "injected" into the collection of the integrals.

dgasmith:
The `MintsHelper::one_body_ao_computer` function is a helper routine for all one-body integrals to parallelize them. All of the overlap, potential, etc run through them.

psi4/src/psi4/libmints/electricfield.h
```diff
@@ -68,6 +71,49 @@ class ElectricFieldInt : public OneBodyAOInt {
...
+    double **pF_;
+    /// The array of dipoles
+    double **dipoles_;
+
+   public:
+    ContractOverDipolesFunctor(SharedMatrix dipoles, SharedMatrix F) : pF_(F->pointer()), dipoles_(dipoles->pointer()) {
+        if (F->rowdim() != F->coldim()) throw PSIEXCEPTION("Invalid Fock matrix in ContractOverCharges");
+        if (dipoles->coldim() != 3) throw PSIEXCEPTION("Dipole matrix must have 3 columns.");
+        int nbf = F->rowdim();
+        std::fill_n(pF_[0], nbf * nbf, 0.0);
```
jturney: What if the user wants to add this contribution to an existing matrix? Maybe this function shouldn't zero it out first. Let it be up to the caller to zero it out, if desired.

I say remove it and let the user decide what they want to do. Maybe add documentation saying contributions are added to the `F`.

psi4/src/psi4/libmints/electricfield.h
```diff
@@ -68,6 +71,49 @@ class ElectricFieldInt : public OneBodyAOInt {
...
+   protected:
+    /// The array of the density matrix
+    double **pD_;
+    /// The array of electric fields
+    double **field_;
+
+   public:
+    ContractOverDensityFieldFunctor(SharedMatrix field, SharedMatrix D) : pD_(D->pointer()), field_(field->pointer()) {
+        if (D->rowdim() != D->coldim()) throw PSIEXCEPTION("Invalid density matrix in ContractOverDensityFieldFunctor");
+        if (field->coldim() != 3) throw PSIEXCEPTION("Field matrix must have 3 columns.");
```
jturney: Here you didn't use `std::fill_n` on `field_` which is how I think it should be.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1562,6 +1563,46 @@ std::vector<SharedMatrix> MintsHelper::electric_field(const std::vector<double>
     return field;
 }
 
+SharedMatrix MintsHelper::induction_operator(SharedMatrix coords, SharedMatrix moments) {
+    SharedMatrix mat = std::make_shared<Matrix>("Induction operator", basisset_->nao(), basisset_->nao());
```
jturney: This call automatically zeroes out the matrix so the `std::fill_n` above would be redundant in this case.

psi4/src/psi4/libmints/electricfield.h
maxscheurer: Very good point! Resolved.

psi4/src/psi4/libmints/electricfield.h
maxscheurer: Done.

psi4/src/psi4/libmints/electricfield.cc
```diff
@@ -214,6 +214,138 @@ void ElectricFieldInt::compute_pair(const GaussianShell& s1, const GaussianShell
...
+                    }  // loop over primitives of shell 2
+                }      // loop over primitives of shell 1
+            }          // loop over points
+            bf2_offset += nj;
+        }  // loop over shell 2
+        bf1_offset += ni;
+    }  // loop over shell 1
+}
+
+template void ElectricFieldInt::compute_with_functor(ContractOverDipolesFunctor&, SharedMatrix);
```
robertodr: Shouldn't `template` be followed by `<>` for this full specialization? https://en.cppreference.com/w/cpp/language/template_specialization

maxscheurer: I thought the same, but then got `Symbol not found` errors when loading the `core` library from Python...

maxscheurer: Found this: https://en.cppreference.com/w/cpp/language/function_template
The variant I used seems to be OK (case (2) in 'Function template instantiation'), but I really don't know why `template<>` does not work...

maxscheurer:
Ready for merge, if there are no further concerns 😉

psi4/src/psi4/libmints/electricfield.h
```diff
@@ -108,6 +118,7 @@ class ContractOverDensityFieldFunctor {
     ContractOverDensityFieldFunctor(SharedMatrix field, SharedMatrix D) : pD_(D->pointer()), field_(field->pointer()) {
         if (D->rowdim() != D->coldim()) throw PSIEXCEPTION("Invalid density matrix in ContractOverDensityFieldFunctor");
         if (field->coldim() != 3) throw PSIEXCEPTION("Field matrix must have 3 columns.");
+        field->zero();
```
jturney: Sorry, I meant to not zero this variable out....as you had it before.

My original comment was comparing this one to the dipole functor: the dipole one you used to zero it out and this one you didn't. So I wanted the dipole one to behave like this one was.

Once this is fixed, someone else can feel free to merge it. I'll be away from a computer for most of the day.

maxscheurer: Now I'm confused...

Just to clarify: The suggestion is *not* to zero out anything?

jturney: Right. You weren't before which allowed the caller to do what they wanted (add to an existing matrix if they desired). My original comment was trying to say that it was correct.

maxscheurer: Alright, got it now... sorry.

dgasmith:
@maxscheurer Thanks for this *awesome* PR! Merging it in,

PR #1676, opened 2019-07-11
Enable interoperability via The MolSSI Driver Interface

taylor-a-barnes:
## Description
This PR enables Psi4 to interoperate on-the-fly with other codes via [The MolSSI Driver Interface (MDI)](https://github.com/MolSSI/MDI_Library).  MDI is an effort of [The MolSSI](https://molssi.org/) to improve and standardize the process of interoperating codes within the computational molecular sciences domain.

In particular, MDI supports interoperability within a driver-engine paradigm, in which a driver code controls the high-level operations of one or more engine codes, orchestrating complex calculations like QM/MM or advanced sampling.  This PR provides all functionality required for Psi4 to act as an MDI engine.  This makes it possible for external codes to exercise control over Psi4 in a manner similar to what is possible using Psi4’s existing API (albeit with a much smaller subset of the functionality), but with the advantage of cross-code generality (all MDI commands are defined by [The MDI Standard](https://molssi.github.io/MDI_Library/html/mdi_standard.html)) and the option of fast inter-code communication via the MPI or TCP/IP methods (selected by the user at runtime).

The PR adds [The MDI Library](https://molssi.github.io/MDI_Library/html/index.html), which handles the details of inter-code communication, as an external dependency.  The MDI Library compiles with CMake and has no additional requirements, although it will link to an MPI library if one is found.

## Todos
- [x] Include the MDI Library as an external dependency
- [x] Add functionality for Psi4 to respond to MDI commands as defined by [The MDI Standard](https://molssi.github.io/MDI_Library/html/mdi_standard.html)
- [x] Add the --mdi command-line option, which causes Psi4 to run as an MDI engine.
- [ ] Document the use of Psi4 as an MDI Engine.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,6 +202,7 @@ ExternalProject_Add(psi4-core
            qcengine_external
            simint_external
            libxc_external
+	   mdi_external
    SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/psi4
    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
               -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,7 @@ ExternalProject_Add(psi4-core
            qcengine_external
            simint_external
            libxc_external
-	   mdi_external
+           mdi_external
    SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/psi4
    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
               -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/external/upstream/mdi/CMakeLists.txt
+++ b/external/upstream/mdi/CMakeLists.txt
@@ -1,0 +1,39 @@
+##if(${ENABLE_mdi})
+    find_package(mdi CONFIG QUIET)
+
+    if(${mdi_FOUND})
+        get_property(_loc TARGET mdi::mdi PROPERTY LOCATION)
+        message(STATUS "${Cyan}Found mdi${ColourReset}: ${_loc} (found version ${mdi_VERSION})")
+        add_library(mdi_external INTERFACE)  # dummy
+    else()
+        if(${CMAKE_INSIST_FIND_PACKAGE_mdi})
+            message(FATAL_ERROR "Suitable mdi could not be externally located as user insists")
+        endif()
+
+        include(ExternalProject)
+        message(STATUS "Suitable mdi could not be located, ${Magenta}Building mdi${ColourReset} instead.")
+        ExternalProject_Add(mdi_external
+            URL https://github.com/MolSSI/MDI_Library/archive/v0.6.1.tar.gz
+            UPDATE_COMMAND ""
+            CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
+                       -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+                       -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                       -DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
+                       -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
+                       -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                       -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+                       -DENABLE_OPENMP=${ENABLE_OPENMP}  # relevant for thread safety
+                       -DENABLE_XHOST=${ENABLE_XHOST}
+                       -DBUILD_FPIC=${BUILD_FPIC}
+                       -DENABLE_GENERIC=${ENABLE_GENERIC}
+                       -DLIBC_INTERJECT=${LIBC_INTERJECT}
+		       -Dlanguage=Python
+            CMAKE_CACHE_ARGS -DCMAKE_Fortran_FLAGS:STRING=${CMAKE_Fortran_FLAGS}
+                             -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS})
+
+        set(mdi_DIR ${STAGED_INSTALL_PREFIX}/share/cmake/mdi CACHE PATH "path to internally built mdiConfig.cmake" FORCE)
+    endif()
+##else()
+##    add_library(mdi_external INTERFACE)  # dummy
+##endif()
+
--- a/external/upstream/mdi/CMakeLists.txt
+++ b/external/upstream/mdi/CMakeLists.txt
@@ -13,7 +13,7 @@
         include(ExternalProject)
         message(STATUS "Suitable mdi could not be located, ${Magenta}Building mdi${ColourReset} instead.")
         ExternalProject_Add(mdi_external
-            URL https://github.com/MolSSI/MDI_Library/archive/v0.6.3.tar.gz
+            URL https://github.com/MolSSI/MDI_Library/archive/v0.6.4.tar.gz
             UPDATE_COMMAND ""
             CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
                        -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/external/upstream/mdi/CMakeLists.txt
+++ b/external/upstream/mdi/CMakeLists.txt
@@ -18,7 +18,6 @@ else()
                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
                    -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
                    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
-                   -DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
                    -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
                    -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
                    -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
@@ -28,8 +27,7 @@ else()
                    -DENABLE_GENERIC=${ENABLE_GENERIC}
                    -DLIBC_INTERJECT=${LIBC_INTERJECT}
                    -Dlanguage=Python
-        CMAKE_CACHE_ARGS -DCMAKE_Fortran_FLAGS:STRING=${CMAKE_Fortran_FLAGS}
-                         -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}
+        CMAKE_CACHE_ARGS -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}
                          -DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS})
 
     set(mdi_DIR ${STAGED_INSTALL_PREFIX}/share/cmake/mdi CACHE PATH "path to internally built mdiConfig.cmake" FORCE)
--- a/external/upstream/mdi/CMakeLists.txt
+++ b/external/upstream/mdi/CMakeLists.txt
@@ -1,41 +1,36 @@
-##if(${ENABLE_mdi})
-    find_package(mdi CONFIG QUIET)
+find_package(mdi CONFIG QUIET)
 
-    if(${mdi_FOUND})
-        get_property(_loc TARGET mdi::mdi PROPERTY LOCATION)
-        message(STATUS "${Cyan}Found mdi${ColourReset}: ${_loc} (found version ${mdi_VERSION})")
-        add_library(mdi_external INTERFACE)  # dummy
-    else()
-        if(${CMAKE_INSIST_FIND_PACKAGE_mdi})
-            message(FATAL_ERROR "Suitable mdi could not be externally located as user insists")
-        endif()
-
-        include(ExternalProject)
-        message(STATUS "Suitable mdi could not be located, ${Magenta}Building mdi${ColourReset} instead.")
-        ExternalProject_Add(mdi_external
-            URL https://github.com/MolSSI/MDI_Library/archive/v0.6.4.tar.gz
-            UPDATE_COMMAND ""
-            CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
-                       -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-                       -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-                       -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
-                       -DCMAKE_Fortran_COMPILER=${CMAKE_Fortran_COMPILER}
-                       -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
-                       -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
-                       -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
-                       -DENABLE_OPENMP=${ENABLE_OPENMP}  # relevant for thread safety
-                       -DENABLE_XHOST=${ENABLE_XHOST}
-                       -DBUILD_FPIC=${BUILD_FPIC}
-                       -DENABLE_GENERIC=${ENABLE_GENERIC}
-                       -DLIBC_INTERJECT=${LIBC_INTERJECT}
-		       -Dlanguage=Python
-            CMAKE_CACHE_ARGS -DCMAKE_Fortran_FLAGS:STRING=${CMAKE_Fortran_FLAGS}
-                             -DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}
-                             -DCMAKE_CXX_FLAGS:STRING=${CMAKE_CXX_FLAGS})
-
-        set(mdi_DIR ${STAGED_INSTALL_PREFIX}/share/cmake/mdi CACHE PATH "path to internally built mdiConfig.cmake" FORCE)
+if(${mdi_FOUND})
+    get_property(_loc TARGET mdi::mdi PROPERTY LOCATION)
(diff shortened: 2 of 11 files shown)


psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,415 @@
...
+
+try:
+    from mpi4py import MPI
+    use_mpi4py = True
+except ImportError:
+    use_mpi4py = False
+
+class MDIEngine():
+
+    def __init__(self, scf_method):
```
dgasmith: It would be good to have a `molecule = None` here and a `self.molecule = molecule or psi4.core.get_active_molecule()` below.

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,415 @@
...
+        self.molecule.fix_orientation(True)
+        self.molecule.fix_com(True)
+        self.molecule.reinterpret_coordentry(False)
+        self.molecule.update_geometry()
+
+        # Flag to stop listening for MDI commands
+        self.stop_listening = False
+
+        # List of all supported MDI commands
+        self.commands = { }
```
dgasmith: Consider writing this as a dict literal:
```
self.commands = {
    "<NATOMS": self.send_natoms,
   ...
}
```

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,415 @@
...
+        natom = self.molecule.natom()
+        MDI_Send(natom, 1, MDI_INT, self.comm)
+        return natom
+
+    # respond to <COORDS command
+    def send_coords(self):
+        """ Send the nuclear coordinates through MDI
+        """
+        natom = self.molecule.natom()
+        coords = [ self.molecule.xyz(iatom)[icoord] for iatom in range(natom) for icoord in range(3) ]
```
dgasmith: `coords = self.molecule.geometry().np.ravel()`
`natom = len(coords)`

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,415 @@
...
+        return self.energy
+
+    # respond to <FORCES command
+    def send_forces(self):
+        """ Send the nuclear forces through MDI
+        """
+        natom = self.molecule.natom()
+        force_matrix = psi4.driver.gradient(self.scf_method)
+        unit_conv = 1.0
+        forces = [ force_matrix.get(iatom,icoord)*unit_conv for iatom in range(natom) for icoord in range(3) ]
```
dgasmith: `(force_matrix.np * unit_conv).ravel().tolist()`

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,415 @@
...
+    def recv_coords(self, coords = None):
+        """ Receive a set of nuclear coordinates through MDI and assign them to the atoms in the current molecule
+
+        Arguments:
+            coords: New nuclear coordinates. If None, receive through MDI.
+        """
+        natom = self.molecule.natom()
+        if coords is None:
+            coords = MDI_Recv(3*natom, MDI_DOUBLE, self.comm)
+        matrix = psi4.core.Matrix(natom, 3)
```
dgasmith: `matrix = psi4.core.Matrix.from_array(np.array(coords).reshape(-1, 3))`

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,415 @@
...
+            self.lattice = MDI_Recv(self.nlattice, MDI_DOUBLE, self.comm)
+        else:
+            self.lattice = lattice
+        self.set_lattice_field()
+
+    # respond to SCF command
+    def run_scf(self):
+        """ Run an energy calculation
+        """
+        self.energy = psi4.energy(self.scf_method)
```
dgasmith: `psi4.energy(self.scf_method, molecule=self.molecule)`

dgasmith:
@taylor-a-barnes I don't see any special issues in our Azure DevOps internals. Can you post the build log here?

@loriab @raimis Taylor is trying to reproduce Psi builds on his own Azure DevOps dashboard to debug the above issue, but even the simple Azure steps like check python version fail. Any ideas what this could be?

raimis:
@taylor-a-barnes I suspect, you are missing some pipeline variables.

Check if your pipeline is configured like this:
azure_config

psi4/run_psi4.py
```diff
@@ -74,6 +74,8 @@
                     help="Runs a JSON input file. !Warning! experimental option.")
 parser.add_argument("-t", "--test", nargs='?', const='smoke', default=None,
                     help="Runs pytest tests. If `pytest-xdist` installed, parallel with `--nthread`.")
+parser.add_argument("--mdi", default=None,
```
loriab: should this mirror `--json` as much as possible?

tests/pytests/test_mdi.py
```diff
@@ -0,0 +1,141 @@
+import pytest
+from qcelemental.testing import compare, compare_recursive, compare_values, tnm
+
+import psi4
+
+
```
loriab: find the marker for "quick" tests and add it here, so it at least tries to run. is mdi ready as a conda package so we can test it in CI?

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,417 @@
...
+        # Get correct intra-code MPI communicator
+        if use_mpi4py:
+            self.mpi_world = MDI_Get_Intra_Code_MPI_Comm()
+            self.world_rank = self.mpi_world.Get_rank()
+
+        # Accept a communicator to the driver code
+        self.comm = MDI_Accept_Communicator()
+
+        # Ensure that the molecule is using c1 symmetry
+        self.molecule.reset_point_group('c1')
```
loriab: if you're altering the molecule, better to clone it first.

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,417 @@
...
+            ">ELEC_MULT": self.recv_multiplicity,
+            "EXIT": self.exit
+        }
+
+    def length_conversion(self):
+        """ Obtain the conversion factor between the geometry specification units and bohr
+        """
+        unit_name = self.molecule.units()
+        if unit_name == "Angstrom":
+            unit_conv = MDI_Conversion_Factor("bohr", "angstrom")
```
loriab: for Angstrom, do you want the value by which psi is converting the stored coords (Ang) to run the method (Bohr)? that's from qcel and is stored in the psi4.core.Mol class. or do you really want the MDI conversion factor?

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,417 @@
...
+        charges = [self.molecule.charge(iatom) for iatom in range(natom)]
+        MDI_Send(charges, natom, MDI_DOUBLE, self.comm)
+        return charges
+
+    # Respond to the <MASSES command
+    def send_masses(self):
+        """ Send the nuclear masses through MDI
+        """
+        natom = self.molecule.natom()
+        masses = [self.molecule.mass(iatom) for iatom in range(natom)]
```
loriab: if you ever have a really big mol, note that you can `molecule.to_dict()` and get these back already in numpy form.

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,417 @@
...
+
+    # Respond to the <ENERGY command
+    def send_energy(self):
+        """ Send the total energy through MDI
+        """
+        MDI_Send(self.energy, 1, MDI_DOUBLE, self.comm)
+        return self.energy
+
+    # Respond to the <FORCES command
+    def send_forces(self):
```
loriab: this and send_energy definitely asymmetric? this runs a gradient QC job, while other returns precomputed? energy.

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,417 @@
...
+            masses: New nuclear masses. If None, receive through MDI.
+        """
+        natom = self.molecule.natom()
+        if masses is None:
+            masses = MDI_Recv(natom, MDI_DOUBLE, self.comm)
+
+        # Assign the mass of all atoms, taking care to avoid ghost atoms
+        jatom = 0
+        for iatom in range(natom):
+            while self.molecule.fcharge(jatom) == 0.0 and jatom < self.molecule.nallatom():
```
loriab: strangely, Z=0 or atom in "ghost" fragment is the best test of ghostedness, rather than charge. Why work around ghosts here for mass but not in above fn for charges?

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,417 @@
...
+        """ Send the total energy through MDI
+        """
+        MDI_Send(self.energy, 1, MDI_DOUBLE, self.comm)
+        return self.energy
+
+    # Respond to the <FORCES command
+    def send_forces(self):
+        """ Send the nuclear forces through MDI
+        """
+        force_matrix = psi4.driver.gradient(self.scf_method)
```
loriab: also, pass along the `molecule=self.molecule` here

taylor-a-barnes:
@raimis Thanks very much - this is exactly the information I needed, and I've been able to reproduce the Azure DevOps results in my own tests.

@dgasmith The Azure DevOps failures were associated with how the MDI Library compiles and installs DLL libraries.  I've fixed the problems with the library, and all the tests are finishing successfully now.  The relevant commits are [here](https://github.com/MolSSI/MDI_Library/commit/25cdcb30abfb4e13ced726249f619ef8acc19582) and [here](https://github.com/MolSSI/MDI_Library/commit/2ea5621ff3ee2020276c588a9962d756cc2cfb18).  Thanks for your comments; I've incorporated them in the latest commits.

@loriab Thanks very much for these comments.  I'll try to address them soon.

psi4/driver/mdi_engine.py
PeterKraus: The manual says the following:
```
<ENERGY

The engine sends its most recently calculated energy to the driver. The MD_INIT, SCF, and ATOM_STEP commands can be used to cause the engine to calculate a new energy.

[...]

<FORCES

The engine calculates and sends a set of atomic forces to the driver. These forces include all force contributions, including the force contributions associated with any constraint algorithm (e.g. SHAKE, RATTLE, etc.).
```
```

File: psi4/run_psi4.py (hunk 7 above)
Comment (PeterKraus) [reply psi4/run_psi4.py]: I'd rewrite to say this creates and launches a MDI listener (as opposed to just "configuring" one). Also, the `scf_method` argument should be passed here, correct?

Comment (PeterKraus) [reply psi4/run_psi4.py]: Actually, upon re-reading this, I have no idea how the `--mdi` command line option works. The test only covers setting/getting values out of/into psi4 using the python interface. Do you have an example snippet of what it looks like in practice?

File: psi4/driver/mdi_engine.py [hunk 15]
```diff
@@ -0,0 +1,417 @@
...
+        """ Receive commands through MDI and respond to them as defined by the MDI Standard
+        """
+
+        while not self.stop_listening:
+            if self.world_rank == 0:
+                command = MDI_Recv_Command(self.comm)
+            else:
+                command = None
+            if use_mpi4py:
+                command = self.mpi_world.bcast(command, root=0)
```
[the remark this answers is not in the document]
taylor-a-barnes: These lines send the same `command` to every MPI instance.  That having been said, I haven't done anything to add MPI parallelization to Psi4, and there should only ever be a single MPI instance of Psi4.  I've added a check to the `MDIEngine` initialization to ensure that this is the case.

When MDI connects codes using MPI, it does so in the MPMD regime, in which a single `mpiexec` command launches all MPI-connected codes.  For example:

```suggestion
mpiexec -n 4 driver -mdi "-role DRIVER -name driver -method MPI" : \
    -n 8 engine1 -mdi "-role ENGINE -name MM -method MPI" : \
    -n 1 psi4 -n 16 -i psi4.dat --mdi "-role ENGINE -name QM -method MPI"
```

This means that MPI.COMM_WORLD may intercommunicate between many MPI ranks (13 in this example), but there should only be a single Psi4 MPI rank.  The `self.mpi_world` object is obtained from the MDI library through the `MDI_Get_Intra_Code_MPI_Comm` function, which returns an inter-communicator that only contains the Psi4 MPI ranks.

In short, I'm doing the broadcast as part of future-proofing efforts, but it doesn't actually matter currently.

taylor-a-barnes:
@loriab @PeterKraus Thanks very much for these comments, and my apologies for the slow turnover time.  I believe that I've addressed everything in the latest commits, but please let me know if any further edits or clarifications are required.

loriab:
please either rebase or [change this one character](https://github.com/psi4/psi4/pull/1712/files) to get travis passing. sorry for the trouble.

dgasmith:
@loriab @PeterKraus This should be ready to go.

loriab:
Great, I’ll look it over again late tonight.

tests/pytests/test_mdi.py
loriab: Before or by the time MDI is available as a conda package, it'd be good to add a test to test_addons.py, too, as those are the "smoke" tests for the psi4 interfaces.

external/upstream/mdi/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
+find_package(mdi CONFIG QUIET)
```
loriab: probably want to add `0.6.4` (or whatever is the minimal compatible version) here as 2nd argument.

often there's a similar `find_package` section in psi4/CMakeLists.txt. not req'd for pure-py but can help in locating.

psi4/run_psi4.py
loriab: The `mdi('hf/cc-pvdz')` is a little scary, as we try to keep the top-level commands constant (energy/gradient/optimize/hessian/frequency) even as their capabilities increase. A `energy('hf/cc-pvdz', mdi=True)` would be better from the perspective of normal Psi4 behavior. Can we revisit this after the distributed driver merge or before you start advertising the mdi/psi interface?

psi4/driver/mdi_engine.py
```diff
@@ -0,0 +1,465 @@
...
+    def send_commands(self):
+        """ Send the supported MDI commands through MDI
+
+        :returns: *command_string* String containing the name of each supported command
+        """
+        command_string = ""
+        for command in self.commands:
+            command_string += command
+            for i in range(len(command), MDI_COMMAND_LENGTH):
+                command_string += " "
```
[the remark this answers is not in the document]
taylor-a-barnes: Thanks for the suggestion!  I've made the change.

external/upstream/mdi/CMakeLists.txt
taylor-a-barnes: I've made both changes.

taylor-a-barnes:
@loriab Thanks for these comments!

I now have some simple documentation [for running Psi4 with MDI](https://molssi.github.io/MDI_Library/html/mdi_ecosystem.html#ecosystem_psi4) as part of the MDI_Library documentation pages.  Its still very basic, but thankfully there isn't much more to the process than what is described in the [general MDI launch documentation](https://molssi.github.io/MDI_Library/html/library_page.html#library_launching_sec).

I've tried to work on adding some how-to-use documentation to Psi4 proper, but I can't seem to get my Sphinx environment set up correctly:

```
Building html documentation ...
Running Sphinx v2.2.0
making output directory... done
/Users/tbarnes/anaconda3/envs/psi4/lib/python3.7/site-packages/sphinx/events.py:103: RemovedInSphinx30Warning: autodoc_default_flags is now deprecated. Please use autodoc_default_options instead.
  results.append(callback(self.app, *args))

Theme error:
no theme named 'cloud' found, inherited by 'sphinx_psi_theme'
make[3]: *** [sphinxman] Error 2
make[2]: *** [doc/sphinxman/CMakeFiles/sphinxman.dir/all] Error 2
make[1]: *** [doc/sphinxman/CMakeFiles/sphinxman.dir/rule] Error 2
make: *** [sphinxman] Error 2
```

This is despite having a conda installation of `cloud_sptheme`.

Regarding the `mdi('hf/cc-pvdz')` command, I would be fine with something more like your suggestion of `energy('hf/cc-pvdz', mdi=True)`.  One alternative possibility would be to say that if the `--mdi` runtime option is provided, the mdi code is just automatically called after everything in the input file has executed; this approach would still require some mechanism of setting the method used, of course, and I have no idea whether this would fit into the broader Psi4 design philosophy.

dgasmith:
@taylor-a-barnes Can you update this PR for the MDI 1.0 release? We are looking to get Psi4 1.4 out soon and love to get this in!

taylor-a-barnes:
@dgasmith I updated to support MDI 1.0 in commit 736ac2c.  The build failed due to reasons that I believe are independent of this PR.  I'll rebase and see if that fixes things.

psi4/run_psi4.py
PeterKraus: I do agree with @loriab on this. I'd rather have the `energy` function be `mdi`-aware than add a top-level command.  I think once we add `psi4.mdi()` it'll be hard to get rid of it...

Actually, the only thing the `psi4.mdi()` function does that cannot be done on init is setting the method string. I'm not sure whether the MDI standard provides a way to set this using a command? Or does the standard require a `mdi` function?

CMakeLists.txt
```diff
@@ -202,6 +202,7 @@ ExternalProject_Add(psi4-core
            qcengine_external
            simint_external
            libxc_external
+	   mdi_external
```
PeterKraus, line 205: The indent here is off.

psi4/driver/mdi_engine.py
```diff
@@ -34,10 +34,17 @@
...
+try:
+    # Attempt to use an internal build of MDI
+    from MDI_Library.mdi import MDI_Init, MDI_Get_Intra_Code_MPI_Comm, MDI_Accept_Communicator, \
+        MDI_Send, MDI_Recv, MDI_Recv_Command, MDI_INT, MDI_DOUBLE, \
+        MDI_Register_Node, MDI_Register_Command
+except ImportError:
+    # Attempt to use an installed MDI package
+    from mdi import MDI_Init, MDI_Get_Intra_Code_MPI_Comm, MDI_Accept_Communicator, \
+        MDI_Send, MDI_Recv, MDI_Recv_Command, MDI_INT, MDI_DOUBLE, \
+        MDI_Register_Node, MDI_Register_Command
```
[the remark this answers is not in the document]
taylor-a-barnes: You're correct.  There was an inconsistency with the MDI install directories that I've fixed.  I've removed the unnecessary lines.

CMakeLists.txt
taylor-a-barnes: Fixed in commit 403e742.

taylor-a-barnes:
@loriab @PeterKraus I agree that the `energy('hf/cc-pvdz', mdi=True)` input format is the way to go.  The MDI Standard doesn't include commands for sending information about method type or related parameters, since there are so many possible ways of representing this information in various codes (i.e., Gaussian codes, plane wave codes, molecular dynamics codes, etc.).  Calling energy with an `mdi` argument seems like the most sensible way for the user to set this information in Psi4.  I've made the change in 403e742.

psi4/driver/driver.py
```diff
@@ -499,6 +500,11 @@ def energy(name, **kwargs):
     """
     kwargs = p4util.kwargs_lower(kwargs)
 
+    # Bounce to MDI if mdi kwarg
+    use_mdi = kwargs.pop('mdi', False)
+    if use_mdi:
+        return mdi_run(name, **kwargs)
```
loriab, line 506: Thanks! I'll rebase the distributed driver branch once this is in and see how it breaks :-)

PR #1680, opened 2019-07-24
Fix SCF plugin python file

andysim:
## Description
Simple fix for the SCF plugin template, to work with the new `set_variable` syntax.

## Questions
- [x] What about [these](https://github.com/psi4/psi4/blob/master/psi4/driver/driver_nbody.py#L295-L298)?  I only discovered them by grepping the codebase so I have no evidence to suggest they should be changed - it's probably the difference between the `wfn` or `core` namespaces, vs. the `psi4` namespace being used in the plugin snippet.

## Status
- [x] Ready for review
- [x] Ready for merge


andysim:
Thanks for the info, @loriab

PR #1687, opened 2019-07-26
Close input file in basis set parser once it's not needed anymore

susilehtola:
# Description
Fixes warning message described on slack by iwatobipen:
```
/home/user/anaconda3/envs/chemo37/lib/python3.7/site-packages/psi4/driver/qcdb/libmintsbasisset.py:854: ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/user/anaconda3/envs/chemo37/share/psi4/basis/aug-cc-pvdz.gbs' mode='r' encoding='UTF-8'>
  names[index] = parser.load_file(fullfilename)
```
as suggested by @loriab 

## Todos
- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/qcdb/libmintsbasissetparser.py
+++ b/psi4/driver/qcdb/libmintsbasissetparser.py
@@ -64,13 +64,14 @@ class Gaussian94BasisSetParser(object):
         basis_separator = re.compile(r'^\s*\[\s*(.*?)\s*\]\s*$')
 
         # Loads an entire file.
-        if os.stat(filename).st_size == 0:
-            raise ValidationError("""BasisSetParser::parse: given filename '%s' is blank.""" % (filename))
         try:
-            with open(filename, 'r') as infile:
-                contents = infile.readlines()
+            infile = open(filename, 'r')
         except IOError:
             raise BasisSetFileNotFound("""BasisSetParser::parse: Unable to open basis set file: %s""" % (filename))
+        if os.stat(filename).st_size == 0:
+            raise ValidationError("""BasisSetParser::parse: given filename '%s' is blank.""" % (filename))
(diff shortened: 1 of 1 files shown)


loriab:
The Azure build log not available, so I can't see its exact problem, but I recall Windows to have very strict ideas about file handlers and closing files. So while I agree that `with open` constructs are better than `open`/`close`, I think go with the simple solution that pleases Windows. Plus, I hope to retire this function in the next few months when we switch over to the new MolSSI BSE.

Thanks for following up on the slack question/fix.

susilehtola:
OK, reverted.

dgasmith:
What about:

```python
try:
    infile = open(filename, 'r')
    if os.stat(filename).st_size == 0:
        raise ValidationError("""BasisSetParser::parse: given filename '%s' is blank.""" % (filename))
    contents = infile.readlines()
except IOError:
    raise BasisSetFileNotFound("""BasisSetParser::parse: Unable to open basis set file: %s""" % (filename))
finally:
    infile.close()
```

susilehtola:
@dgasmith why? It's functionally equivalent. And as @loriab said, this code is going to be replaced anyhow in a few months' time...

dgasmith:
@susilehtola They are not quite equivalent as the `close()` will not be called if an exception is raised. Granted an edge case. Happy to merge this either way as the code will be removed soon.

PeterKraus:
I am reasonably sure that Azure failed due to the issue with libxc download location (while tddft.org was down), and it wasn't a problem with the `with ... as ` way. Anyway, merging.

Issue #1694, opened 2019-08-13
Calculations fail in unconventional basis sets


susilehtola:
A segfaulting test case
```
molecule {
0 2
H
}
set reference uhf
set df_scf_guess false
set scf_type direct

basis {
assign test

[test]
spherical
****
H 0
P 1 1.0
 1.0 1.0
****
}
energy('scf')
```

I have found one issue and fixed it in hf.cc
```
diff --git a/psi4/src/psi4/libscf_solver/hf.cc b/psi4/src/psi4/libscf_solver/hf.cc
index fec5efaa4..c462addab 100644
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -705,9 +705,15 @@ void HF::form_Shalf() {
 
     // Convert the eigenvales to 1/sqrt(eigenvalues)
     const Dimension& dimpi = eigval->dimpi();
-    double min_S = std::fabs(eigval->get(0, 0));
+    // Cannot assume that (0,0) is a valid reference
+    bool min_S_initialized = false;
+    double min_S;
     for (int h = 0; h < nirrep_; ++h) {
         for (int i = 0; i < dimpi[h]; ++i) {
+            if(!min_S_initialized) {
+              min_S = eigval->get(h, i);
+              min_S_initialized = true;
+            }
```
but now the code crashes somewhere in pybind11.

susilehtola:
Ah, nevermind, the pybind11 issue is because it had been updated and I just needed a recompile.

JonathonMisiewicz:
Should #1695 close this?

loriab:
good call, @JonathonMisiewicz

PR #1695, opened 2019-08-13
Fix segfault with unconventional basis sets.

susilehtola:
## Description
Fixes the issue #1694 .

## Todos
- [x] Calculations work also when the zeroth symmetry block has no basis functions

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -722,8 +722,7 @@ void HF::form_Shalf() {
             if (!min_S_initialized) {
                 min_S = eigval->get(h, i);
                 min_S_initialized = true;
-            }
-            if (min_S > eigval->get(h, i)) {
+            } else if (min_S > eigval->get(h, i)) {
                 min_S = eigval->get(h, i);
             }
             double scale = 1.0 / std::sqrt(eigval->get(h, i));
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -705,10 +705,18 @@ void HF::form_Shalf() {
 
     // Convert the eigenvales to 1/sqrt(eigenvalues)
     const Dimension& dimpi = eigval->dimpi();
-    double min_S = std::fabs(eigval->get(0, 0));
+    // Cannot assume that (0,0) is a valid reference
+    bool min_S_initialized = false;
+    double min_S;
     for (int h = 0; h < nirrep_; ++h) {
         for (int i = 0; i < dimpi[h]; ++i) {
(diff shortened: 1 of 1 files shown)


psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -705,10 +714,18 @@ void HF::form_Shalf() {
...
+    bool min_S_initialized = false;
+    double min_S;
     for (int h = 0; h < nirrep_; ++h) {
         for (int i = 0; i < dimpi[h]; ++i) {
-            if (min_S > eigval->get(h, i)) min_S = eigval->get(h, i);
+            if (!min_S_initialized) {
+                min_S = eigval->get(h, i);
+                min_S_initialized = true;
+            }
+            if (min_S > eigval->get(h, i)) {
```
PeterKraus: This could probably be an `elif`, saves an `if` check on initialisation...

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -705,10 +714,18 @@ void HF::form_Shalf() {
...
     const Dimension& dimpi = eigval->dimpi();
-    double min_S = std::fabs(eigval->get(0, 0));
+    // Cannot assume that (0,0) is a valid reference
+    bool min_S_initialized = false;
+    double min_S;
     for (int h = 0; h < nirrep_; ++h) {
         for (int i = 0; i < dimpi[h]; ++i) {
-            if (min_S > eigval->get(h, i)) min_S = eigval->get(h, i);
+            if (!min_S_initialized) {
+                min_S = eigval->get(h, i);
```
PeterKraus: I'm wondering, above we used to initialise using `abs(eigval[0,0])`, here it's just `eigval[h,i]`. Is that really correct? I guess values in `eigval` must be positive, otherwise the `sqrt(eigval[h,i])` below wouldn't work, but I'm just checking.

susilehtola: Good catch. The old version was actually wrong. In the theoretical case (I'm not sure if it even exists?) where you would have a negative eigenvalue but no eigenvalues close to zero, the old version would have just used symmetric orthogonalization, whereas the correct way is, of course, to only take the positive definite part.

But, this is just an academic discussion: the overlap matrix is supposed to be positive definite. Sometimes you do see small negative eigenvalues.

susilehtola: Also note that the square root is only taken from eigenvalues that are bigger than the cutoff that is positive.

psi4/src/psi4/libscf_solver/hf.cc
susilehtola: Done, although I would assume the compiler is smart enough to do this.

PR #1699, opened 2019-08-13
XDM calculations in psi4

aoterodelaroza:
This PR contains the implementation of XDM-enabled density functional theory calculations in psi4. XDM is a dispersion correction in which the dispersion coefficients that enter the asymptotic dispersion formula are calculated post-SCF from the self-consistent electron density and kinetic energy density. XDM maintains the simplicity and accuracy of a damped asymptotic dispersion correction while, at the same time, it calculates the dispersion coefficients from the self-consistent wavefunction. 

The XDM energy and gradient are calculated by the [postg](https://github.com/aoterodelaroza/postg)  external program, to which the new code interfaces. The implementation is very similar to the way in which psi4 currently interfaces to gCP. In addition, a few XDM-enabled functionals are provided to facilitate its use. A few tests, samples, and documentation are also included in the package. That said, I recommend that you check the XDM/postg documentation meets psi4 standards as I have no way to verify what it looks like in HTML form.


CDSherrill:
This will be a nice interface to have!

loriab:
Thanks much!

* at first read-through functional integration parts look good
* we'll want to move the running of postg over to qcengine to imitate `dftd3` more than `gcp`. this'll integrate better with the hookup-disp-to-fctls machinery to avoid `## hack ##`. sound ok?
* is https://github.com/aoterodelaroza/postg the official repository home for `postg` where development occurs? PRs considered?
* we'll probably want to switch some logic so that XDM doesn't just turn on in the background when dft_xdm parameters change without the driver's knowledge as this could lead to wrongly labeled variables.
* any objection to a `postg` conda package distributed off the `psi4` or `conda-forge` channels? I have bullets if you're not familiar.

Looking forward to getting this integrated. thanks again!

aoterodelaroza:
Sure, sounds good. 
- OK with moving to qcengine (sorry about that; I didn't want to touch multiple repos at the same time, hence the `## hack ##`).
- That github repo is where postg development occurs, yes. PRs and requests for features are welcome.
- OK with changing the logic in any way that is more appropriate. Let me know and I can make an attempt.
- Never used conda. I'm OK with making a postg package, though, but I'll probably need some guidance.

doc/sphinxman/source/postg.rst
```diff
@@ -0,0 +1,109 @@
...
+
+.. codeauthor:: Alberto Otero de la Roza
+.. sectionauthor:: Alberto Otero de la Roza
+
+*Module:* :ref:`Samples <apdx:testSuitepostg>`
+
+The `exchange-hole dipole moment (XDM) model <http://schooner.chem.dal.ca/wiki/XDM>`_ 
+[Becke:2007:154108]_ is a dispersion correction used to get accurate intermolecular
+interaction energies in density-functional theory calculations. The
+XDM energy is calculated post-SCF using an external program called
```
susilehtola: This is an empirical correction, not a post-SCF a.k.a. single-shot correction, right?

susilehtola: Nevermind, I see that this is ab initio but the calculation is done by postg via a Molden file, not Psi4...

psi4/driver/procrouting/dft/xdm_functionals.py
```diff
@@ -0,0 +1,283 @@
...
+    '    B86bPBE Exchange-Correlation Functional plus XDM dispersion (GGA).\n',
+    "dispersion": {
+        "type": "xdm",
+        "params": {
+            'a1': 0.7839,
+            'a2': 1.2544,
+            'vol': '0.0',
+        }
+    }
+})
```
dgasmith: IIRC XDM is not done self-consistently. Could we try to add this more likely canonical -D terms where we treat it as a post-processing step.

loriab: In the beginning (in q-chem at least), XDM was done self-consistently. But I think they decided it didn't matter.

I was thinking of setting it up as dftd3-like (fall back on nl-like) as possible. Most unwieldy item seems to be these `vol`s where there are a periodic table's-worth for each of 9 fctls (https://github.com/aoterodelaroza/postg/blob/master/param.f90#L220-L317), then a fallback to %HF scale of a generic data block. So that `vol` really is a string or float.

aoterodelaroza: The XDM implementation in q-chem is indeed self-consistent but, as you point out, it does not matter very much (or at all) and post-SCF is much simpler. The atomic volumes are there because optimal results are obtained when the atom-in-molecule volume calculated from your system goes to the "correct" in-vacuo atomic volume if the molecule is taken apart into atoms. If the functional in question is not known, then the fraction of exact exchange is used instead, which more or less solves the problem.

dgasmith:
Gentle poke here, let us know if you need help migrating to QCEngine!

aoterodelaroza:
Oh, sorry - I thought someone from the psi4 team was going to do
it. Sure, I can put it wherever you want but it'll have to wait until
I have some free time. Probably one or two weeks.

dgasmith:
Thanks, you can hop on the QCArchive [slack](https://join.slack.com/t/qcarchive/shared_invite/enQtNDIzNTQ2OTExODk0LTE3MWI0YzBjNzVhNzczNDM0ZTA5MmQ1ODcxYTc0YTA1ZDQ2MTk1NDhlMjhjMmQ0YWYwOGMzYzJkZTM2NDlmOGM) and we can help you from there.

Issue #1708, opened 2019-08-28
Futureproofing ctest?


rxk900:
Have been going through the psi4 build in preparation for our new supercomputer and found a "glitch" in that all the psi4numpy tests failed with:
ImportError: No module named numpy
which puzzled me momentarily.
Chased it down to:
tests/runtest.py:    pyexitcode = backtick(["python", infile, " > ", outfile])
Our python3 doesn't equivalence to python, and alias (which I usually use to get round this) didn't carry through though it may have if I put it in my dot files. Anyway the Q&D of python-> python3 did the trick but our system guys have pointed out:
[https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8/](https://developers.redhat.com/blog/2019/05/07/what-no-python-in-red-hat-enterprise-linux-8/)
Something to keep in mind?

JonathonMisiewicz:
The action item that remains here is transitioning fully to `pytest`.

PR #1709, opened 2019-08-30
shebang and travis fixes

loriab:
## Todos
- [x] since some OSes will only be supplying `python2` and `python3` and not `python`, we'd better work
- [x] below is what was failing at `atol=1.e-8`. I venture it was tipped over the threshold, so loosened to 2.e-8. should fix travis.
```
AAAAA 1e-08
[[ 0.         0.        -0.0669792]
 [ 0.         0.         0.0669792]]
[[ 0.          0.         -0.06697921]
 [ 0.          0.          0.06697921]]
    cc-pVDZ tot grad..................................................FAILED
```

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/document_cfour.py
+++ b/doc/sphinxman/document_cfour.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
--- a/doc/sphinxman/document_databases.py
+++ b/doc/sphinxman/document_databases.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
--- a/doc/sphinxman/document_driver.py
+++ b/doc/sphinxman/document_driver.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
--- a/doc/sphinxman/document_efpfrag.py
+++ b/doc/sphinxman/document_efpfrag.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
--- a/doc/sphinxman/document_plugins.py
+++ b/doc/sphinxman/document_plugins.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
--- a/doc/sphinxman/document_psifiles.py
+++ b/doc/sphinxman/document_psifiles.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
--- a/doc/sphinxman/extract_changeset.py
+++ b/doc/sphinxman/extract_changeset.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 #
 # @BEGIN LICENSE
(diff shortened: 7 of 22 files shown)


doc/sphinxman/document_cfour.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

doc/sphinxman/document_databases.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

doc/sphinxman/document_driver.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

doc/sphinxman/document_efpfrag.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

doc/sphinxman/document_plugins.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

doc/sphinxman/document_psifiles.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

doc/sphinxman/extract_changeset.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
robertodr: `#!/usr/bin/env python3` for consistency with the rest, plus some systems do not have `python3` under `/usr/bin`.

loriab:
the consensus from [SO](https://stackoverflow.com/questions/34212036/python-3-installation-on-windows-running-from-command-line) and azure is that Windows has no `python3` cmd. if from #1708, redhat has no `python`, then we're in a pickle.

andysim:
A couple of options: 1) use [find_executable()]( https://docs.python.org/3.5/distutils/apiref.html#module-distutils.spawn) to figure it out or 2) have CMake inject the python executable path that it detected, using something like [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html).

loriab:
The difficulty with those normally good solutions, @andysim, is that testing writes its output to `objdir/` but all the testing files and `runtest.py` itself stay in source.

My last commit was stupid -- ignore that one.

These "special" `ctest` dirs `python/`, `json/`, `psi4numpy/` that `python input.py` (instead of `psi4 input.dat` like the others) were constructed when there was no `pytest`. Since we want to kill off `ctest` anyways, any objections to just moving these over to pytest now? CI covers pytest tests already. Only loss in coverage will be among users/devs who don't know to run `pytest` as well as `ctest`.

loriab:
This PR is a good idea, but we just use python in so many different ways that it's risky w/o having OSes to test it on. OS packagers may just need to patch in python3. Leaving PR open for guidance.

Issue #1711, opened 2019-09-04
print Initial orbital occupation?


hokru:
It can be very helpful to know the orbital occupations per IRREP before the SCF.
Otherwise one was has to wait until the end of the SCF or when it changes during the iterations.
I suggest to add a printout of the occupations after the SCF guess. 

This is a small change I think, just a print statement, and if this is wanted I can make the PR myself.

dgasmith:
When we print out the original `nom` matrix, maybe we can show guess occupations there instead of 0's?

hokru:
That would be a good location. Easy to move the printout to after the guess stage.

Though I didn't take into account that the `SAD` guess does not give orbitals and the occupations look strange. Not sure what it being printed.

susilehtola:
> Though I didn't take into account that the `SAD` guess does not give orbitals and the occupations look strange. Not sure what it being printed.

Garbage? :smile:

dgasmith:
With SAD, canonical operations are built and then munged. We store the canonical occupations in another temporary array... err forgot the name, should be at the bottom of `sad.cc`.

hokru:
water in Cs symmetry:
SAD
```   -------------------------------------------------------
    Irrep   Nso     Nmo     Nalpha   Nbeta   Ndocc  Nsocc
   -------------------------------------------------------
     A'        28      28       6       6       6       0
     A"        11      11       1       1       1       0
   -------------------------------------------------------
    Total      39      39       7       7       7       0
   -------------------------------------------------------
```

SADNO (correct occ)
```   -------------------------------------------------------
    Irrep   Nso     Nmo     Nalpha   Nbeta   Ndocc  Nsocc
   -------------------------------------------------------
     A'        28      28       4       4       4       0
     A"        11      11       1       1       1       0
   -------------------------------------------------------
    Total      39      39       5       5       5       0
   -------------------------------------------------------
```

dgasmith:
Right there are temporary arrays where you will have to change the default printing from `nalpha/nbeta` to `_nalpha_holder` or whatever for SAD.

susilehtola:
For SAD you should only print out the occupations after the 1st iteration, since that's when it determines the real occupations.

hokru:
SAD is annoying :-(

Maybe it is best to keep printing zeros for SAD (somehow) and additional a note saying that no occupations are available yet. For difficult open-shell systems, where this table becomes interesting, people might switch the guess anyway.

susilehtola:
[...]

PR #1713, opened 2019-09-05
printout of initial orbital occupations

hokru:
## Description
Suggestion for #1711
Moves the `nmo` table to a place after the orbital guess, so that it actual contains all the information. Adds a reduced `nmo` table for SAD together with a note that no occupations are available.

_Formatting details up to debate_. Perhaps especially the placement of the `Pre-Iterations` line.

Edit: See updated examples in thread below

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -166,11 +166,6 @@ def scf_initialize(self):
     self.memory_jk_ = int(total_memory - collocation_memory)
     self.memory_collocation_ = int(collocation_memory)
 
-    # Print out initial docc/socc/etc data
-    if self.get_print():
-        core.print_out("  ==> Pre-Iterations <==\n\n")
-        self.print_preiterations()
-
     if self.get_print():
         core.print_out("  ==> Integral Setup <==\n\n")
 
@@ -198,7 +193,6 @@ def scf_initialize(self):
         self.initialize_jk(self.memory_jk_, jk=jk)
         if self.V_potential():
             self.V_potential().build_collocation_cache(self.memory_collocation_)
-
         core.timer_on("HF: Form core H")
         self.form_H()
         core.timer_off("HF: Form core H")
@@ -219,9 +213,19 @@ def scf_initialize(self):
         self.form_Shalf()
         core.timer_off("HF: Form S/X")
 
+        core.print_out("\n  ==> Pre-Iterations <==\n\n")
+
         core.timer_on("HF: Guess")
         self.guess()
         core.timer_off("HF: Guess")
+        # Print out initial docc/socc/etc data
+        if self.get_print():                    
+            lack_occupancy = core.get_local_option('SCF', 'GUESS') in ['SAD']
+            if core.get_global_option('GUESS') in ['SAD']:
+                lack_occupancy = core.get_local_option('SCF', 'GUESS') in ['AUTO']
+                self.print_preiterations(small=lack_occupancy)
+            else:
+                self.print_preiterations(small=lack_occupancy)
 
     else:
         # We're reading the orbitals from the previous set of iterations.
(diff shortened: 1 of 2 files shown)


psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -687,6 +691,26 @@ def scf_print_preiterations(self):
     )
     core.print_out("   -------------------------------------------------------\n\n")
 
+def scf_print_preiterations_small(self):
```
dgasmith: Would this be cleaner as part of the `scf_print_preiterations` function with `small=True`?

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -222,6 +217,15 @@ def scf_initialize(self):
         core.timer_on("HF: Guess")
         self.guess()
         core.timer_off("HF: Guess")
+        # Print out initial docc/socc/etc data
```
dgasmith: Interesting that this will put the pre-iterations after two-electron integrals which will change the input file substantially. Can you you post a before/after comparison?

psi4/driver/procrouting/scf_proc/scf_iterator.py
hokru: Yes, i think so. I wasn't sure if I need to make changes somewhere else if I modify the function and chose the saver approach for now.

hokru:
For comparison the normal output
[default.out.txt](https://github.com/psi4/psi4/files/3580080/default.out.txt)

psi4/driver/procrouting/scf_proc/scf_iterator.py
hokru: added a comparison.

dgasmith:
I wonder if it would be better if we moved the pre-iterations down so that it generated the one-electron quantities in the "pre-iterations" and then the the occupations. 

We would have a `==> JK Setup <==` area where the JK builds would be moved down to after the occupation printing.

hokru:
The default guess is `AUTO`. If condition needs adjustment.

hokru:
Small update on this. Also moved `==> Pre-Iterations <==` to before the SCF guess:

```
  Minimum eigenvalue in the overlap matrix is 3.2955701764E-02.

  ==> Pre-Iterations <==

   Note: current guess does not provide
         initial orbital occupations

   -------------------------
    Irrep   Nso     Nmo    
   -------------------------
     A         48      48 
   -------------------------
    Total      48      48
   -------------------------

  ==> Iterations <==
```

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -166,11 +166,6 @@ def scf_initialize(self):
     self.memory_jk_ = int(total_memory - collocation_memory)
     self.memory_collocation_ = int(collocation_memory)
 
-    # Print out initial docc/socc/etc data
-    if self.get_print():
-        core.print_out("  ==> Pre-Iterations <==\n\n")
-        self.print_preiterations()
-
     if self.get_print():
         core.print_out("  ==> Integral Setup <==\n\n")
```
dgasmith: I guess what I am getting at here is that we can move `initialize_jk` to below all of the one-electron stuff. Where the one-electron stuff would become the `Pre-iterations` in total. `Integral Setup` would then become something specific to JK. I think it is useful to have some sort of information before JK is built which takes often takes more time than a few SCF iterations.

@psi4/editors This could use a look by everyone. 

@hokru Hits on a pretty important topic here! Apologies for convolving it with other issues, but this should be looked at holistically I think.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -669,23 +677,44 @@ def scf_print_energies(self):
     self.set_variable('SCF ITERATIONS', self.iteration_)
 
 
-def scf_print_preiterations(self):
+def scf_print_preiterations(self,small=False):
```
dgasmith: Nice! Thanks.

psi4/driver/procrouting/scf_proc/scf_iterator.py
hokru: No worries, I agree on carefully crafting how the output is ordered. Though I am not yet sure I fully understand where you want everything :-)  Let's see other people's comments.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -669,23 +677,44 @@ def scf_print_energies(self):
...
+        core.print_out("   -------------------------------------------------------\n")
         core.print_out(
-            f"     {ct.gamma(h).symbol():<3s}   {self.nsopi()[h]:6d}  {self.nmopi()[h]:6d}  {self.nalphapi()[h]:6d}  {self.nbetapi()[h]:6d}  {self.doccpi()[h]:6d}  {self.soccpi()[h]:6d}\n"
+            f"    Total  {self.nso():6d}  {self.nmo():6d}  {self.nalpha():6d}  {self.nbeta():6d}  {self.nbeta():6d}  {self.nalpha() - self.nbeta():6d}\n"
         )
+        core.print_out("   -------------------------------------------------------\n\n")
+    else:
+
+        core.print_out("   Note: current guess does not provide\n")
+        core.print_out("         initial orbital occupations\n\n")
```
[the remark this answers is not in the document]
hokru: Fine by me. My intention was to make it generic in case another guess is added without orbital occupations.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -219,9 +214,22 @@ def scf_initialize(self):
...
+        if self.get_print():
+            if not core.has_global_option_changed('GUESS'):
+                if core.get_local_option('SCF', 'GUESS') in ['SAD']:
+                    self.print_preiterations(small=True)
+                else:
+                    self.print_preiterations()
+            elif core.get_global_option('GUESS') in ['SAD']:
+                    self.print_preiterations(small=True)
+            else:
+                    self.print_preiterations(small=False)
```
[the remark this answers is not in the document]
hokru: My troubles were that `get_global_option('GUESS')` is and remains `AUTO` by default and driver only changes the `local` option, e.g. here:
https://github.com/psi4/psi4/blob/ab68a2666287dc9df25ee1e13370ad92d446ccbf/psi4/driver/procrouting/proc.py#L1237

I believe I need to catch the cases `global=AUTO & local=SAD/read` and `global=SAD & local=SAD/read`.

hokru:
Updated according to suggestions. 
```
=> Pre-Iterations <==

  Minimum eigenvalue in the overlap matrix is 3.9724095377E-02.

  SCF Guess: Superposition of Atomic Densities via on-the-fly atomic UHF (no occupation info).

   -------------------------
    Irrep   Nso     Nmo
   -------------------------
     A'        18      18
     A"         6       6
   -------------------------
    Total      24      24
   -------------------------

  ==> 2e-Integral Setup <==

  DFHelper Memory: AOs need 0.001 GiB; user supplied 0.321 GiB. Using in-core AOs.

  ==> MemDFJK: Density-Fitted J/K Matrices <==

    J tasked:                   Yes
```

psi4/driver/procrouting/scf_proc/scf_iterator.py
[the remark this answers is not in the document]
loriab: Ah, got it. Could you check what `get_option('SCF', 'GUESS')` in the cases you want to catch?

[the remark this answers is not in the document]
hokru: `get_option('SCF', 'GUESS')` is `AUTO` unless changed by user input, also for optimization cycles etc.

I now see I don't catch the `global=SAD` & `local=read` case.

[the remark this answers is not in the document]
hokru: I re-worked the if-else and added some debug prints to monitor. So far it looks alright.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -1069,7 +1069,7 @@ void HF::guess() {
         guess_E = compute_initial_E();
 
     } else if (guess_type == "SAD") {
-        if (print_) outfile->Printf("  SCF Guess: Superposition of Atomic Densities via on-the-fly atomic UHF.\n\n");
+        if (print_) outfile->Printf("  SCF Guess: Superposition of Atomic Densities via on-the-fly atomic UHF (no occupation info).\n\n");
```
dgasmith: "no occupation information"

hokru: done.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -222,18 +221,29 @@ def scf_initialize(self):
         core.timer_on("HF: Guess")
         self.guess()
         core.timer_off("HF: Guess")
+        # Print out initial docc/socc/etc data
+        global_=core.get_global_option('GUESS') # DEBUG REMOVE
```
PeterKraus: Don't forget to remove these before merging.

PeterKraus:
@hokru Perhaps better to resolve the conflicts sooner than later.

hokru:
Updated. Had some rebase issues to solve. A vimdiff of the old and new output below

---
vimdiff of the output between 1.4a2.dev215 (left) and this PR (right)
Screenshot 2020-03-03 at 23 18 02

dgasmith:
Thanks for the update! Looks good.

@psi4/admins Can we get one more review here?

robertodr:
I think another rebase is necessary to fix the build.

PR #1714, opened 2019-09-07
Issue cleanup: fix various crashes.

### LINKED ISSUE #1613: Relativistic calculations referring to non-standard basis fails
This problem was already mentioned in passing here: https://github.com/psi4/psi4/issues/662#issuecomment-475861476

This is a simple example input file showing the problem:

```
molecule {
  H
}

set {
    scf_type pk
    basis ano-rcc
    relativistic x2c
    reference uhf
}

energy('hf')
```

I have set my `$PSIPATH` to a directory containing the basis set file (downloaded from EMSL):

```bash
ll $PSIPATH
total 2.1M
drwxr-xr-x. 2 toon toon 4.0K Apr 29 17:13 .
drwxr-xr-x. 7 toon toon 4.0K Apr 29 16:34 ..
-rw-r--r--. 1 toon toon 2.1M Apr 29 15:49 ano-rcc.gbs
```

When running Psi4 on this input, I get the following error message:

```
  File "/home/toon/miniconda3/envs/p4dev/bin/psi4", line 287, in <module>
    exec(content)
  File "<string>", line 26, in <module>
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/driver.py", line 561, in energy
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/procrouting/proc.py", line 2045, in run_scf
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/procrouting/proc.py", line 1281, in scf_helper
    scf_wfn = scf_wavefunction_factory(name, base_wfn, core.get_option('SCF', 'REFERENCE'), **kwargs)
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/procrouting/proc.py", line 1056, in scf_wavefunction_factory
    puream=wfn.basisset().has_puream())
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/p4util/python_helpers.py", line 80, in _pybuild_basis
    mol.to_dict(), key, resolved_target, fitrole, other, return_dict=True, return_atomlist=return_atomlist)
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/qcdb/libmintsbasisset.py", line 679, in pyconstruct
    verbose=verbose)
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/qcdb/libmintsbasisset.py", line 838, in construct
    filename = cls.make_filename(basgbs)
  File "/home/toon/miniconda3/envs/p4dev/lib/psi4/driver/qcdb/libmintsbasisset.py", line 1396, in make_filename
    basisname = basisname.lower()

AttributeError: 'NoneType' object has no attribute 'lower'

Printing out the relevant lines from the Psithon --> Python processed input file:
    core.IO.set_default_namespace("")
    core.set_global_option("SCF_TYPE", "pk")
    core.set_global_option("BASIS", "ano-rcc")
    core.set_global_option("RELATIVISTIC", "x2c")
    core.set_global_option("REFERENCE", "uhf")
--> energy('hf')
```

This problem was previously also brought up on the forum: http://forum.psicode.org/t/scf-problems-please-help/765/1

### LINKED ISSUE #1675: Molecule xyz getter segfault
Currently the `Molecule.xyz()` method does not check the length of the array that it is getting.

```
>>> import psi4
>>> mol  = psi4.geometry("He 0 0 0\nHe 0 0 5")
>>> mol.xyz(3)
[1]    26521 segmentation fault  python
```

This should be capped and provide a nice error message to the user. This should be a straightforward fix, the relevant file is [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/molecule.cc).

### LINKED ISSUE #1694: Calculations fail in unconventional basis sets

PeterKraus:
## Description
Resolves #1694 by throwing an exception when active molecule has only 1 fragment.
Resolves #1675 by throwing an exception when requested atom is out of bounds.
Resolves #1613 by changing the lookup path for AUX/DECON bases from `libraryPath` to `basisPath`.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/qcdb/libmintsbasisset.py
+++ b/psi4/driver/qcdb/libmintsbasisset.py
@@ -795,11 +795,11 @@ class BasisSet(object):
                     raise BasisSetNotDefined("""BasisSet::construct: No basis set specified for %s and %s.""" %
                         (symbol, key))
                 else:
-                    # No auxiliary basis set for atom, so try darnedest to find one.
+                    # No auxiliary / decon basis set for atom, so try darnedest to find one.
                     #   This involves querying the BasisFamily for default and
                     #   default-default and finally the universal default (defined
                     #   in this function). Since user hasn't indicated any specifics,
-                    #   look only in Psi4's library and for symbol only, not label.
+                    #   look for symbol only, not label.
                     tmp = []
                     tmp.append(corresponding_basis(basdict['BASIS'], deffit))
                     #NYI#tmp.append(corresponding_basis(basdict['BASIS'], deffit + '-DEFAULT'))
@@ -808,7 +808,7 @@ class BasisSet(object):
                         tmp.append(univdef[deffit])
                     seek['basis'] = [item for item in tmp if item != (None, None, None)]
                     seek['entry'] = [symbol]
-                    seek['path'] = libraryPath
+                    seek['path'] = basisPath
                     seek['strings'] = ''
             else:
                 # User (I hope ... dratted has_changed) has set basis for atom,
--- a/psi4/src/psi4/libmints/molecule.cc
+++ b/psi4/src/psi4/libmints/molecule.cc
@@ -2221,100 +2221,109 @@ std::vector<std::string> Molecule::irrep_labels() {
     return irreplabel;
 }
 
-Vector3 Molecule::xyz(int atom) const { 
-    if (atom >= atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
-    return input_units_to_au_ * atoms_[atom]->compute(); 
+void Molecule::check_atom_(int atom, bool full) const  {
+    if (full && atom >= full_atoms_.size()) {
+        throw std::runtime_error("Requested atom doesn't exist in full atoms list.");
+    }
+    if (not full && atom >= atoms_.size()) {
+        throw std::runtime_error("Requested atom doesn't exist in atoms list.");
+    }
+}
+
+Vector3 Molecule::xyz(int atom) const {
+    check_atom_(atom, false);
+    return input_units_to_au_ * atoms_[atom]->compute();
 }
 
 Vector3 Molecule::fxyz(int atom) const {
-    if (atom >= full_atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
+    check_atom_(atom, true);
     return input_units_to_au_ * full_atoms_[atom]->compute();
 }
 
 double Molecule::xyz(int atom, int _xyz) const {
-    if (atom >= atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
+    check_atom_(atom, false);
     return input_units_to_au_ * atoms_[atom]->compute()[_xyz];
 }
 
 const double &Molecule::Z(int atom) const {
-    if (atom >= atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
-    return atoms_[atom]->Z(); 
+    check_atom_(atom, false);
+    return atoms_[atom]->Z();
 }
 
 double Molecule::fZ(int atom) const {
-    if (atom >= full_atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
+    check_atom_(atom, true);
     return full_atoms_[atom]->Z();
 }
 
 double Molecule::x(int atom) const {
-    if (atom >= atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
+    check_atom_(atom, false);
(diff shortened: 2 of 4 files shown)


psi4/src/psi4/libmints/molecule.cc
```diff
@@ -2221,47 +2221,100 @@ std::vector<std::string> Molecule::irrep_labels() {
...
 
-void Molecule::set_nuclear_charge(int atom, double newZ) { atoms_[atom]->set_nuclear_charge(newZ); }
+void Molecule::set_nuclear_charge(int atom, double newZ) {
+    if (atom >= atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
+    atoms_[atom]->set_nuclear_charge(newZ);
+}
 
-const std::string &Molecule::basis_on_atom(int atom) const { return atoms_[atom]->basisset(); }
+const std::string &Molecule::basis_on_atom(int atom) const {
+    if (atom >= atoms_.size()) throw std::runtime_error("Requested atom doesn't exist.");
```
dgasmith: Would it make more sense if this was a utility function `void check_atom(int atom)`?

Might make things slightly simpler.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -808,7 +808,7 @@ def construct(cls, parser, mol, key, deffit=None, basstrings=None, return_atomli
                         tmp.append(univdef[deffit])
                     seek['basis'] = [item for item in tmp if item != (None, None, None)]
                     seek['entry'] = [symbol]
-                    seek['path'] = libraryPath
+                    seek['path'] = basisPath
```
dgasmith: I think this right, but defiantly needs a @loriab thumbs up.

loriab: When the issue came up, I couldn't decide for several days if I was comfortable with the switch. Now ok with it.

psi4/src/psi4/libmints/molecule.h
```diff
@@ -277,6 +277,8 @@ class PSI_API Molecule {
     double xyz(int atom, int _xyz) const;
     /// Returns mass atom atom
     double mass(int atom) const;
+    /// Checks whether atom is within bounds of atom_ or full_atom_
+    void check_atom_(int atom, bool full) const;
```
loriab: By my understanding, because `check_atom_` is just a helper function local to the `.cc` file, it doesn't need to be declared in the header, and the definition in the `.cc` can drop the `Molecule::` part.

A rebase will fix Azure now, but we can also simply merge after the 3rd review since it'll heal in master.

PeterKraus: I don't think that works, as the function needs to access `atoms_` and `full_atoms_` which are only defined if the function belongs to the `Molecule::` class.

At least, with L281 commented and https://github.com/psi4/psi4/blob/7a892760e628f2d54ba02d6d09da71acf4c42c8c/psi4/src/psi4/libmints/molecule.cc#L2224 changed to:
```
void check_atom_(int atom, bool full) const  {
```

the compilation crashes with `error: 'full_atoms_' was not declared in this scope`.

loriab: Sounds about like the depth of my C++ experience. Try for `protected:`, then, like https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/molecule.h#L155

PR #1717, opened 2019-09-14
Restore DCFT Guess

### LINKED ISSUE #1309: DCT Estimate of MP2 Inconsistent
During an optimization where an initial hessian by finite difference is requested, the gradient is requested at the starting geometry _twice_. Once by the optimizer and once to determine what modes should be projected in the finite difference treatment. The SCF and final energies agree, but for some reason, the estimated MP2 energies disagree, and the RMS lambda errors don't match either. In this example, the MP2 correlation energy the first time is -0.018978215515226, but the MP2 correlation energy the second time is -0.017979235543976. An MP2 computation agreed with the first correlation energy.

Minimal input file below. I've seen this bug on systems with UHF and systems of C1 symmetry. I want this fixed ASAP, just in case the bug does affect energies in some way I haven't seen, so I'll take this myself.
```
molecule mol {
  0 1 
  units bohr
  H
  H 1 1.8 
}

set {
  reference rhf 
  basis sto-3g
  scf_type pk
  dcft_functional odc-12
  full_hess_every 0
}

optimize('dcft')
````

### PR DESCRIPTION
**Pull request by JonathonMisiewicz**
## Description
While working on a _different_ PR for the dcft code, I stumbled across a bug that I had accidentally introduced in #1310. DCFT previously allowed you to use the amplitudes from a previous computation as your guess for the next one by setting `dcft_guess dcft`. Key to this mechanism was the fact that the LibDPD files persisted after the computation... which is precisely what I changed in #1310 to fix #1309, where intermediate files would interfere with the amplitude guess.

The primary purpose of this PR is to both keep the guess feature working and keep issue #1309 fixed. This is accomplished by checking if `dcft_guess` is `dcft` at the time of `compute_energy` and choosing whether or not to nuke previous files. So past intermediate files will interfere only when you want them to.

There are also some related changes: adding a helpful error message to replace the PSIO error that appears when the past files don't exist, adding a test case to prevent a repeat of this little mess, removing some dead code, and per discussions with @jturney, moving some tensors from the libtrans libdpd to the dcft libdpd if they were more than integrals.

The other `dcft` PR will be coming as soon as this one is merged in: another bugfix, and (finally!) the transition from `dcft` to `dct`.

Obligatory @ssh2 notification.

## Todos
- [x] Restore `dcft_guess dcft` functionality
- [x] Move tensors (DCT's amplitudes and "generalized" Fock matrix) from the `libtrans` dpd buffer to the `dcft` buffer
- [x] Add a test for `dcft_guess`
- [x] Eliminate one obnoxious PSIO_ERROR

## Checklist
- [x] Tests added for any newly working features
- [x] All 14 `dcft` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dcft/dcft_compute.cc
+++ b/psi4/src/psi4/dcft/dcft_compute.cc
@@ -35,6 +35,7 @@
 #include "psi4/libtrans/integraltransform.h"
 #include "psi4/libdiis/diismanager.h"
 #include "psi4/libpsio/psio.hpp"
+#include "psi4/libpsio/psio.h"
 
 #include <cmath>
 
@@ -46,6 +47,19 @@ namespace dcft {
  */
 double DCFTSolver::compute_energy() {
     double total_energy = 0.0;
+
+    if (options_.get_str("DCFT_GUESS") == "DCFT") {
+		// We're reusing DCFT files from a previous computation. Let's check that one exists.
+		if (!((same_a_b_orbs_ == true && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda SF <OO|VV>")) ||
+			(same_a_b_orbs_ == false && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda <Oo|Vv>")))) {
+			throw PSIEXCEPTION("Could not find a previous DCFT computation as the DCFT_GUESS=DCFT guess.");
+		}
+	}
+	else {
+		// Obliviate all old DCFT files.
+		psio_->close(PSIF_DCFT_DPD, 0);
+		psio_->open(PSIF_DCFT_DPD, PSIO_OPEN_OLD);
+	}
 
     if ((options_.get_str("SCF_TYPE").find("DF") != std::string::npos) || options_.get_str("SCF_TYPE") == "CD" ||
         options_.get_str("SCF_TYPE") == "DIRECT") {
--- a/psi4/src/psi4/dcft/dcft_compute.cc
+++ b/psi4/src/psi4/dcft/dcft_compute.cc
@@ -49,17 +49,16 @@ double DCFTSolver::compute_energy() {
     double total_energy = 0.0;
 
     if (options_.get_str("DCFT_GUESS") == "DCFT") {
-		// We're reusing DCFT files from a previous computation. Let's check that one exists.
-		if (!((same_a_b_orbs_ == true && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda SF <OO|VV>")) ||
-			(same_a_b_orbs_ == false && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda <Oo|Vv>")))) {
-			throw PSIEXCEPTION("Could not find a previous DCFT computation as the DCFT_GUESS=DCFT guess.");
-		}
-	}
-	else {
-		// Obliviate all old DCFT files.
-		psio_->close(PSIF_DCFT_DPD, 0);
-		psio_->open(PSIF_DCFT_DPD, PSIO_OPEN_OLD);
-	}
+        // We're reusing DCFT files from a previous computation. Let's check that one exists.
+        if (!((same_a_b_orbs_ == true && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda SF <OO|VV>")) ||
+              (same_a_b_orbs_ == false && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda <Oo|Vv>")))) {
+            throw PSIEXCEPTION("Could not find a previous DCFT computation as the DCFT_GUESS=DCFT guess.");
+        }
+    } else {
+        // Obliviate all old DCFT files.
+        psio_->close(PSIF_DCFT_DPD, 0);
+        psio_->open(PSIF_DCFT_DPD, PSIO_OPEN_OLD);
+    }
 
     if ((options_.get_str("SCF_TYPE").find("DF") != std::string::npos) || options_.get_str("SCF_TYPE") == "CD" ||
         options_.get_str("SCF_TYPE") == "DIRECT") {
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: psi4/src/psi4/dcft/dcft_compute.cc [hunk 1]
```diff
@@ -47,6 +48,19 @@ namespace dcft {
...
+		if (!((same_a_b_orbs_ == true && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda SF <OO|VV>")) ||
+			(same_a_b_orbs_ == false && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda <Oo|Vv>")))) {
+			throw PSIEXCEPTION("Could not find a previous DCFT computation as the DCFT_GUESS=DCFT guess.");
+		}
+	}
+	else {
+		// Obliviate all old DCFT files.
+		psio_->close(PSIF_DCFT_DPD, 0);
+		psio_->open(PSIF_DCFT_DPD, PSIO_OPEN_OLD);
+	}
```
loriab: spacing a little funny here. if it's set up, many want to `clang-format` it.

psi4/src/psi4/dcft/dcft_compute_RHF.cc
```diff
@@ -152,11 +152,11 @@ void DCFTSolver::run_simult_dcft_RHF() {
 
     // DIIS on orbitals (AA and BB) and cumulants (AA, AB, BB)
     dpdbuf4 Laa, Lab, Lbb;
-    global_dpd_->buf4_init(&Laa, PSIF_LIBTRANS_DPD, 0, ID("[O,O]"), ID("[V,V]"), ID("[O,O]"), ID("[V,V]"), 0,
+    global_dpd_->buf4_init(&Laa, PSIF_DCFT_DPD, 0, ID("[O,O]"), ID("[V,V]"), ID("[O,O]"), ID("[V,V]"), 0,
```
loriab: this doesn't need a `psifiles.h` edit?

tests/dcft10/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(dcft1 "psi;dcft")
```
loriab: dcft10, not dcft1

psi4/src/psi4/dcft/dcft_compute_RHF.cc
JonathonMisiewicz: `PSIF_DCFT_DPD` [already exists](https://github.com/psi4/psi4/blob/master/psi4/include/psi4/psifiles.h#L297), so I wouldn't think so.

loriab:
I think pull this in after 3rd review -- the Azure error has nothing to do with the PR

tests/dcft10/input.dat
```diff
@@ -0,0 +1,34 @@
...
+energy('dcft')
+compare_values(refodc12, variable("DCFT TOTAL ENERGY"), 10, "ODC-12 Energy");                #TEST
+
+set dcft_guess cc
+energy('dcft')
+compare_values(refodc12, variable("DCFT TOTAL ENERGY"), 10, "ODC-12 Energy");                #TEST
+
+set dcft_guess bcc
+wfn = energy('dcft', return_wfn=True)[1]
+compare_values(refodc12, variable("DCFT TOTAL ENERGY"), 10, "ODC-12 Energy");                #TEST
```
andysim: Is it possible / advisable to limit the number of iterations to just a couple here?  That way we'd be testing not only that the run converges to the correct solution, but that the starting guess is correctly read in.

psi4/src/psi4/dcft/dcft_compute.cc
```diff
@@ -47,6 +48,18 @@ namespace dcft {
...
+    if (options_.get_str("DCFT_GUESS") == "DCFT") {
+        // We're reusing DCFT files from a previous computation. Let's check that one exists.
+        if (!((same_a_b_orbs_ == true && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda SF <OO|VV>")) ||
+              (same_a_b_orbs_ == false && psio_tocentry_exists(PSIF_DCFT_DPD, "Lambda <Oo|Vv>")))) {
+            throw PSIEXCEPTION("Could not find a previous DCFT computation as the DCFT_GUESS=DCFT guess.");
+        }
+    } else {
+        // Obliviate all old DCFT files.
+        psio_->close(PSIF_DCFT_DPD, 0);
+        psio_->open(PSIF_DCFT_DPD, PSIO_OPEN_OLD);
```
PeterKraus: I'm not sure how this works. If the `close( ..., 0)` should remove the file, then `open( ..., ` should be on a file that is new, not old. Or `close(..., 0)` truncates to 0 instead of removing? If the latter is true, then the following should be corrected:

https://github.com/psi4/psi4/blob/fbb2ff444490bf6b43cb6e027637d8fd857adcee/psi4/src/psi4/libpsio/psio.hpp#L229-L230

tests/dcft10/input.dat
JonathonMisiewicz: No; the `cc` and `bcc` starting guesses are _worse_ than the MP2 guess for this system, at about 5 times the RMS error in the amplitudes. Each takes 12 iterations.

When you read in the amplitudes and orbitals (`dcft_guess dcft` with `ref_wfn` passing), the guess is perfect, so I already set the number of iterations to 3 there.

andysim: Ah, my bad!  I missed the line where you set the number of iterations to 3

psi4/src/psi4/dcft/dcft_compute.cc
JonathonMisiewicz: If I had to guess, `psio_->open` silently creates the file if it's been removed.

Making `libpsio` comprehensible to mere mortals is far, far, far outside the scope of this PR.

PR #1718, opened 2019-09-15
fixing g2g ci on win

loriab:
## Todos
- [x] internal builds on pybind11 with Intel atop non-system compilers would think compiler broken b/c can't find gcc guts. so we'll pass along the flags. way overkill considering that "build" is just a `cp`.
- [x] fix the Azure g2g header trouble
- [x] in the process, update to the new g2g 2.0 (unreleased) interface
- [x] all tests pass
- [x] CPPE testing -- fix ctest and add pytest (latter needed for`psi4-rt`)
- [x] qz2p basis from CFOUR

- Q: imminence of g2g 2?

## Status
- [x] Ready for review
- [x] Ready for merge (except for no packaged g2g)


loriab:
@dgasmith, so this gets passed the mm_malloc header the other PRs were having. Unfortunately, by building off g2g master rather than Dec release, it hits all the ordering changes you made. The current state compiles and runs but gives nonsensical results for DFT. You perhaps had other plans for psi's transition on this matter. Basically, is it better to patch the ifdef rerouting onto 1.3.1 and keep gg_collocation stable, or is now as good a time as any to update the psi interface to g2g?

psi4/src/psi4/libfock/points.cc
```diff
@@ -763,12 +767,12 @@ void BasisFunctions::compute_functions(std::shared_ptr<BlockOPoints> block) {
 
         // Copmute collocation
         if (deriv_ == 0) {
-            gg_collocation(L, npoints, x, y, z, nprim, norm, alpha, center.data(), (int)puream_, phi_start);
+            gg_collocation(L, npoints, xyz, 1, nprim, norm, alpha, center.data(), GG_SPHERICAL_GAUSSIAN, phi_start);
```
dgasmith: This should be be `GG_SPHERICAL_GAUSSIAN` or `GG_CARTESIAN_ROW` I believe depending on `puream_`

psi4/CMakeLists.txt
```diff
@@ -115,7 +115,7 @@ else()
     message(STATUS "Disabled erd")
 endif()
 
-find_package(gau2grid 1.2 CONFIG REQUIRED COMPONENTS gaussian)
+find_package(gau2grid 1.3.1 CONFIG REQUIRED)
```
dgasmith: Just an FYI, next release will be a 2.0 due to API breaking changes.

psi4/src/psi4/libfock/points.cc
```diff
@@ -702,6 +702,10 @@ void BasisFunctions::compute_functions(std::shared_ptr<BlockOPoints> block) {
     double* x = block->x();
     double* y = block->y();
     double* z = block->z();
+    double* xyz = new double[npoints * 3];
+    memcpy(xyz, x, npoints);
+    memcpy(xyz + npoints, y, npoints);
+    memcpy(xyz + 2 * npoints, z, npoints);
```
dgasmith: Cool, simpler than I thought it would be.

external/upstream/gau2grid/CMakeLists.txt
```diff
@@ -13,7 +13,8 @@ else()
     message(STATUS "Suitable gau2grid could not be located, ${Magenta}Building gau2grid${ColourReset} instead.")
     ExternalProject_Add(gau2grid_external
         DEPENDS pybind11_external
-        URL https://github.com/dgasmith/gau2grid/archive/v1.3.1.tar.gz
+        #URL https://github.com/dgasmith/gau2grid/archive/v1.3.1.tar.gz
+        URL https://github.com/dgasmith/gau2grid/archive/loriab-patch-1.tar.gz
```
dgasmith: I deleted this branch btw.

loriab:
ok, folks, this now gives correct answers and fixes the azure error.

* @dgasmith, what's the prospect of a g2g 2.0 release
* all those who consider pointers their allies should check that my formation of `xyz` is equivalent to https://github.com/dgasmith/gau2grid/compare/v1.3.1...master#diff-f8f6cc08ea43fcbc6d2e07d960e118c2L69-R94

external/upstream/gau2grid/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ else()
     message(STATUS "Suitable gau2grid could not be located, ${Magenta}Building gau2grid${ColourReset} instead.")
     ExternalProject_Add(gau2grid_external
         DEPENDS pybind11_external
-        URL https://github.com/dgasmith/gau2grid/archive/v1.3.1.tar.gz
+        URL https://github.com/dgasmith/gau2grid/archive/v2.0.0.tar.gz
```
dgasmith:
```suggestion
        URL https://github.com/dgasmith/gau2grid/archive/v2.0.1.tar.gz
```

psi4/src/psi4/libfock/points.cc
```diff
@@ -702,6 +702,10 @@ void BasisFunctions::compute_functions(std::shared_ptr<BlockOPoints> block) {
     double* x = block->x();
     double* y = block->y();
     double* z = block->z();
+    double* xyz = new double[npoints * 3];
+    ::memcpy(xyz, x, sizeof(double) * npoints);
+    ::memcpy(xyz + npoints, y, sizeof(double) * npoints);
+    ::memcpy(xyz + 2 * npoints, z, sizeof(double) * npoints);
```
andysim: The C++ way to do this is with either `std::copy` or `std::copy_n`.  Let's leave this as-is for now and I'll update it when i reconcile with the GGA hessian changes

psi4/share/psi4/basis/NOTES
```diff
@@ -1,6 +1,13 @@
 *******************************************************************************
 *******************************************************************************
 
+Sunday, September 15, 2019 -- LAB
+
+[45] Added qz2p basis translated from CFOUR's GENBAS.
```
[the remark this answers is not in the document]
loriab: Cfour uses qz2p in a lot of their examples, so I wanted it for matching values. Others maybe for a MolSSI pass.

psi4/src/psi4/libfock/points.cc
loriab: Right, I couldn't find that branch you mentioned. I've had my learning exercise -- glad to see the modern way.

PR #1719, opened 2019-09-17
Rename DCFT to DCT; Make DF-DCT Custom Basis Sets Findable

JonathonMisiewicz:
## Description
@andysim added the `dcft` module to Psi in 2010, as its pilot implementation. As the method was developed, @ssh2 decided that it was best to drop the `f` and name it `dct`. The abbreviation `dcft` hasn't been used in papers since 2012, and DCT papers since last year have just called the method density cumulant theory, cutting out the `f` entirely. This PR brings Psi up-to-the-present and calls the method `dct` everywhere.

Now that we know what the method _is called_, this PR fixes a bug reported on the forums, where the method's density fitting basis set wasn't picked up by the input parser. This was the motivation for this massive PR, and it took one whole line of code.

## Todos
- [x] Replaces `dcft` and `DCFT` with `dct` and `DCT`, throughout all C- and Py- side code
- [x] Added UpgradeHelper errors for users who try `energy('dcft')` and some such
- [x] Added PsiException errors for users who try a dcft option block
- [x] Added PsiException errors for users who try a dcft option
- [x] Updates the tests
- [x] Updates the samples
- [x] Updates the documentation (abbreviations kept when they are part of a paper title)
- [x] Removed debug options which the code didn't even use anymore
- [x] Adds the one line of code needed to fix the instigating bug

## Checklist
- [x] Every `ctest` test calling DCT is run and passes (14 dedicated DCT tests plus four additional tests)
- [x] I've confirmed that you encounter the various errors when expected
- [x] I've manually checked the documentation. The only thing not working is the link to the `dct` folder, which won't exist until _after_ the PR is accepted.

## Status
- [x] Ready for review
- [x] Ready for merge


dgasmith:
Are there particular areas that should be reviewed here?

JonathonMisiewicz:
The main things worth looking at:
* psi4/driver/driver.py, psi4/src/core.cc
	Do we have any recommended changes to the error messages? Any error messages I should add but didn't?
* Are there any files I didn't edit that I should have, just based on the fact that I am renaming an entire module?
* I also removed some options which the code stopped using. (The MO_RELAX and IGNORE_TAU options, both of which were marked to be used as debug tools.) Is there anything I should do for that other than remove mention of them in psi4/src/read_options.cc and anywhere they appeared in the DCT code?

Beyond that, the changes should be search and replace. Particularly zealous reviewers can double-check that the only instances of `dcft` left are those that refer to specific citations or say that `dcft` was the old name.

JonathonMisiewicz:
That's a great idea. Do you have an example of where pytest catches a _C-side_ error? The error handling in renamed options needs to be C-side, and I don't know how to refer to C's PsiException Py-side.

psi4/src/core.cc
```diff
@@ -501,11 +501,23 @@ bool specifies_convergence(std::string const& key) {
     return ((key.find("CONV") != key.npos) || (key.find("TOL") != key.npos));
 }
 
+// DCFT deprecation errors first added in 1.4. Feel free to retire after "enough" time. 
+void throw_deprecation_errors(std::string const& key, std::string const& module = "") {
+    if (module == "DCFT") {
+		throw PsiException("Rename local options block. All instances of `dcft` should be replaced with `dct`.", __FILE__, __LINE__);
```
dgasmith: "DCFT local options block was renamed in v1.4"?

loriab:
Sorry for the really slow response. No, haven't caught c-side errors before. You can see if catching general `Exception` works and if can scan traceback for text. Bu don't expend too much effort. It's the placement of driver logic that I'm more worried about breaking.

Also, if you rebase, Azure should heal.

JonathonMisiewicz:
All comments addressed. Unfortunately, I need to filter out renamed methods before any checks of the derivative type, since that raises a validation error if the method is invalid. Hence I need to catch the error in multiple branches of `gradient`. Isn't recursive driver wonderful?

PR #1721, opened 2019-09-19
Libint2 and shell screening

loriab:
## Description
Libint2 and shell screening

## Todos
- [x] initial cmake integration of fixed AM Libint2

## Questions
- [ ] Question1

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


loriab:
@jturney @andysim @fevangelista @zachglick 
FYI, both libint1 and libint2 can now be built internally from this PR.

* libint1 pulls a branch where the `Libint_t` has already become `Libint1_t`. Because the compiler looks everywhere in includes directories for headers, easiest to still build this branch in a conda env _without_ `libint` in `conda list`. (If you have `psi4-dev` installed, `conda remove libint --force` is what you need.)
* libint2 pulls a tarball, then configures with spherical `gaussian` (looks like the quotes were the problem last week).
* if using `psi4-dev`, may want to dial down `-DMAX_AM_ERI` to 6 or so.

If you've already built the specialized libint1 and/or libint2 externally, the below are what you want to tell psi4 to find them (it's the location of the Config.cmake file).
```
-DLibint2_DIR=/programming/libint-2.6.0/installdir/lib/cmake/libint2 
-DLibint_DIR=/programming/libint1/installdir/share/cmake/Libint
```

andysim:
Ok, I just pushed a (still slightly broken) version that changes the libint2 interface very significantly, and almost has gradients working.  The problematic derivatives are those where the angular momentum is greater in the bra than the ket, so perhaps that's a limitation that I need to resolve by ensuring loops satisfy ang mom constraints.  Here's how I built the code in case anybody wants to replicate...

1. Clone the libint2 repo and checkout the v2.6 branch

2. `LIBS=-L/u/andysim/anaconda3/lib CXXFLAGS="-std=c++14 -I/u/andysim/anaconda3/include" ../configure --enable-eri=2 --enable-eri3=2 --enable-eri2=2 --with-eri-max-am=5 --with-eri2-max-am=6 --with-eri3-max-am=6 --with-shgauss-ordering=gaussian --prefix=/u/andysim/programming/libint/installdir3 --enable-eri2-pure-sh --enable-eri3-pure-sh`

3. `make install` (this will take a *very* long time)

4. `make export` - this will generate a tarball that can be used to build a version for Psi4 called `libint-2.7.0-beta.1.tgz`

5. Untar that tarball somewhere else and modify the file `include/libint2_params.h` to read
```
#ifndef LIBINT2_MAX_AM_default
# define LIBINT2_MAX_AM_default 4
#endif

#ifndef LIBINT2_MAX_AM_default1
# define LIBINT2_MAX_AM_default1 LIBINT2_MAX_AM_default
#endif

#ifndef LIBINT2_MAX_AM_default2
# define LIBINT2_MAX_AM_default2 LIBINT2_MAX_AM_default
#endif
```
 from line 81 onwards

6. Make a `build` subdir and from there `cmake .. -DCMAKE_INSTALL_PREFIX=/u/andysim/programming/libint/installdir_2.6 -DLIBINT2_SHGAUSS_ORDERING=gaussian -DCMAKE_CXX_FLAGS="-fPIC"`  The fPIC flag is needed, or you'll get some nasty link errors.

7.  Now you can build Psi4.  Yay!  `cmake .. -DCMAKE_BUILD_TYPE=Debug -DLibint2_DIR=/u/andysim/programming/libint/installdir_2.6/lib/cmake/libint2 -DLibint_DIR=/u/andysim/programming/libint1/installdir/share/cmake/Libint -DCMAKE_CXX_FLAGS="-g -O0 -gxx-name=/v/apps/gcc/5.4.0/bin/g++  $CXXFLAGS"`  The libint1 should be the hacked version with the `Libint_t` changed to `Libint1_t`.  Also I strongly recommend adding the `-O0` flag to the compiler flags, because our CMake debug build doesn't do this (we should change that) and the code takes a _very_ long time to compile with optimization.

A lot of this will be streamlined in the not so distant future, and much of the complication will be eased when we nuke libint1.  Until then, the above command should be sufficient to get you up and running with a development build.  Please try it out - I'd love to have some help with getting the derivatives working :)

andysim:
I should add that the reason I point to conda in the first step of the above is because I installed boost, Eigen and gmp in there.  Those are all needed for libint2, and are easily obtained via conda.

andysim:
Also, here's a useful test case for anyone interested in getting their hands dirty.  Toggling between the (correct) libint1 and engine and libint2 version in C1 / cartesians should yield identical results.
```
molecule h2o {
0 1
O
H 1 R
H 1 R 2 A

R = 1.0
A = 104.5
}

set {
   integral_package libint1
   puream false
   scf_type direct
   df_scf_guess false
   guess core
   reference rhf
   basis sto-3g
   e_convergence 12
   d_convergence 9
}

analytic = gradient('scf')

set findif points 3
findif = gradient('scf', dertype=0)

compare_matrices(analytic, findif, 5, "RHF SCF finite-diff (3-pt) vs. analytic gradient to 10^-5") #TEST

set findif points 5
findif = gradient('scf', dertype=0)

compare_matrices(analytic, findif, 7, "RHF SCF finite-diff (5-pt) vs. analytic gradient to 10^-7") #TEST
```

loriab:
Alternatively to the [separate libtool build of libint2](https://github.com/psi4/psi4/pull/1721#issuecomment-537945417), you can build a Libint2 externally with cmake or internal to psi4.

1. Get some extra dependencies `conda install boost mpfr eigen ninja`. Only header-only parts of Boost are used. Eigen is needed for the l2 c++ interface. Ninja is because l2 now makes files with weird enough names that when the get written to `Makefile`s, `make` can't process them. EFV is strongly recommending Ninja, too. 

2. Make sure a Libint1 package is **not** in your conda environment. Even if you're properly pointing to a `Libint1_t`-modified l1 via `-DLibint_DIR`, the compiler will still find those conventional Libint1 headers in `$CONDA_PREFIX`. If you have `psi4-dev` installed, `conda remove libint --force` to remove libint(1) but keep `psi4-path-advisor`.

3. In https://github.com/loriab/psi4/blob/libint2/external/upstream/libint2/CMakeLists.txt and https://github.com/loriab/psi4/blob/libint2/psi4/CMakeLists.txt, there are three `CHOOSE!` toggles between detect-prebuilt-libtool-l2 or detect-or-build-pure-cmake-l2. Comment out the one in each pair you don't need.

4. Normal cmake configuration of Psi4 into a new objdir. I recommend tacking on the following `-DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS="-O0" -DMAX_AM_ERI=5 -DBOOST_ROOT=$CONDA_PREFIX -DMPFR_ROOT=$CONDA_PREFIX -DEigen3_DIR=$CONDA_PREFIX/share/eigen3/cmake/`. The `MAX_AM_ERI` only controls Libint1. The Libint2 MAX_AMs are set [here](https://github.com/psi4/psi4/pull/1721/files#diff-3cbeb175ac8e4122ee11133c203e49b4R32). On command-line, modify by e.g., `-DENABLE_ERI=1 -DERI_MAX_AM="5;4" -DENABLE_ERI3=0 -DERI3_MAX_AM=5`

5. Build. The downloading source for library step will take a loooong time because that's when it's running the generator. Don't disrupt it during this step or it'll start over. The first time you build through this procedure, you may want to turn Hessians off and AM down to threes or fours to test the whole procedure. I'm pretty confident that the libraries itself are built right, but I wouldn't be surprised at can't-find-dependencies-or-headers problems on other computers.

6. Once libint2 is finished building and psi4 starts, psi4 will error with `error: 'LIBINT2_MAX_AM_default1' was not declared in this scope`. The fix is the same as [Andy's (5)](https://github.com/psi4/psi4/pull/1721#issuecomment-537945417), editing `objdir/stage/include/libint2/libint2_params.h`. Fix that up and resume compiling psi4.

7. If you fiddle with the external/upstream/libint2/CM file and the generator starts rebuilding, you're toast. `rm -rf objdir/external/upstream/libint2` and `make` again from objdir to restart. Details: The way EFV does the AM defaulting is not a way CMake wants to act. I'm still working on a fix. You'll see "exponent has no digits" errors. In any case, if the generator has to run again (which is probably correct -- it's not the dependency structure that's wrong, it's the generated config.h), there's no time to be saved.

CMakeLists.txt
```diff
@@ -243,6 +244,8 @@ ExternalProject_Add(psi4-core
               -Dgau2grid_DIR=${gau2grid_DIR}
               -Dgdma_DIR=${gdma_DIR}
               -DLibint_DIR=${Libint_DIR}
+              -DLibint2_DIR=${Libint2_DIR}
+              -Dlibint2_DIR=${Libint2_DIR}
```
jturney: Line 248 repeats 247.

loriab: sadly, not quite. one's `L` and one's `l` b/c we're dealing with two ppl's cmake at the moment. you can use the EFV's `make export` that's gnumake+cmake or my pure cmake. definitely there will be only one in the end.

cmake/psi4OptionsTools.cmake
```diff
@@ -61,6 +61,7 @@ macro(psi4_add_module binlib libname sources)
   target_link_libraries(${libname}
     PRIVATE
       tgt::lapack
+      Libint2::cxx
```
loriab: for those reviewing, this is linking L2 into _every_ p4 module. this contrasts with L1 which only linked into mints. Reason is b/c using L2 shell directly in basisset.h, and that's included everywhere. worth a revisit to check absolutely necessary.

CMakeLists.txt
jturney: Oh, right.... I didn't see the difference in capitalization the first go round.

cmake/psi4OptionsTools.cmake
andysim: Excellent news!  I changed the design a lot, and now only eri.cc and eribase.cc include the libint2 headers; it's forward declared elsewhere.  Let's change the behavior back to the old libint1 way and save a lot of compile time!

loriab: great, kill this line, then back to mints only! let me know if you're pausing this evening or weekend, then I'll bully azure for a bit.

andysim: have at it!  the lastest pushed version is rebased to master and passes full tests and pytests.  i'm just writing some docs right now, so won't conflict with any of your stuff

psi4/src/psi4/dfmp2/corr_grad.cc
```diff
@@ -656,12 +656,13 @@ void DFCorrGrad::build_AB_x_terms() {
         int oQ = auxiliary_->shell(Q).function_index();
 
         int ncart = cP * cQ;
-        const double* Px = buffer + 0 * ncart;
-        const double* Py = buffer + 1 * ncart;
-        const double* Pz = buffer + 2 * ncart;
-        const double* Qx = buffer + 3 * ncart;
-        const double* Qy = buffer + 4 * ncart;
-        const double* Qz = buffer + 5 * ncart;
+        bool libint1 = Process::environment.options.get_str("INTEGRAL_PACKAGE") != "LIBINT2";
```
susilehtola: Should this rather be `== "LIBINT"`?

susilehtola: Or switch it around so that the variable is `libint2`.

susilehtola: Also, since the indexing is the same, you could define `const double *bufp = libint2 ? Jint[thread]->buffers() : buffer` and then the 6 lines of assignment wouldn't need the conditionals.

psi4/src/psi4/dfmp2/corr_grad.cc
```diff
@@ -941,15 +942,16 @@ void DFCorrGrad::build_Amn_x_terms() {
...
+            bool libint1 = Process::environment.options.get_str("INTEGRAL_PACKAGE") != "LIBINT2";
+            const double* Px = libint1 ? buffer + 0 * ncart : eri[thread]->buffers()[0];
+            const double* Py = libint1 ? buffer + 1 * ncart : eri[thread]->buffers()[1];
+            const double* Pz = libint1 ? buffer + 2 * ncart : eri[thread]->buffers()[2];
+            const double* Mx = libint1 ? buffer + 3 * ncart : eri[thread]->buffers()[3];
+            const double* My = libint1 ? buffer + 4 * ncart : eri[thread]->buffers()[4];
+            const double* Mz = libint1 ? buffer + 5 * ncart : eri[thread]->buffers()[5];
+            const double* Nx = libint1 ? buffer + 6 * ncart : eri[thread]->buffers()[6];
+            const double* Ny = libint1 ? buffer + 7 * ncart : eri[thread]->buffers()[7];
+            const double* Nz = libint1 ? buffer + 8 * ncart : eri[thread]->buffers()[8];
```
susilehtola: Same thing here.

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1573,12 +1574,14 @@ void RDFMP2::form_AB_x_terms() {
...
-        const double* Qy = buffer + 4 * ncart;
-        const double* Qz = buffer + 5 * ncart;
+        bool libint1 = Process::environment.options.get_str("INTEGRAL_PACKAGE") != "LIBINT2";
+        auto& buffers = Jint[thread]->buffers();
+        const double* Px = libint1 ? buffer + 0 * ncart : buffers[0];
+        const double* Py = libint1 ? buffer + 1 * ncart : buffers[1];
+        const double* Pz = libint1 ? buffer + 2 * ncart : buffers[2];
+        const double* Qx = libint1 ? buffer + 3 * ncart : buffers[3];
+        const double* Qy = libint1 ? buffer + 4 * ncart : buffers[4];
+        const double* Qz = libint1 ? buffer + 5 * ncart : buffers[5];
```
susilehtola: and here

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1759,15 +1763,16 @@ void RDFMP2::form_Amn_x_terms() {
...
+            bool libint1 = Process::environment.options.get_str("INTEGRAL_PACKAGE") != "LIBINT2";
+            const double* Px = libint1 ? buffer + 0 * ncart : buffers[0];
+            const double* Py = libint1 ? buffer + 1 * ncart : buffers[1];
+            const double* Pz = libint1 ? buffer + 2 * ncart : buffers[2];
+            const double* Mx = libint1 ? buffer + 3 * ncart : buffers[3];
+            const double* My = libint1 ? buffer + 4 * ncart : buffers[4];
+            const double* Mz = libint1 ? buffer + 5 * ncart : buffers[5];
+            const double* Nx = libint1 ? buffer + 6 * ncart : buffers[6];
+            const double* Ny = libint1 ? buffer + 7 * ncart : buffers[7];
+            const double* Nz = libint1 ? buffer + 8 * ncart : buffers[8];
```
susilehtola: and here

psi4/src/psi4/dfocc/tei_grad_corr.cc
```diff
@@ -133,12 +135,12 @@ void DFOCC::tei_grad_corr() {
...
-        const double *Pz = buffer + 2 * ncart;
-        const double *Qx = buffer + 3 * ncart;
-        const double *Qy = buffer + 4 * ncart;
-        const double *Qz = buffer + 5 * ncart;
+        const double *Px = libint1 ? buffer + 0 * ncart : buffers[0];
+        const double *Py = libint1 ? buffer + 1 * ncart : buffers[1];
+        const double *Pz = libint1 ? buffer + 2 * ncart : buffers[2];
+        const double *Qx = libint1 ? buffer + 3 * ncart : buffers[3];
+        const double *Qy = libint1 ? buffer + 4 * ncart : buffers[4];
+        const double *Qz = libint1 ? buffer + 5 * ncart : buffers[5];
```
susilehtola: and here

psi4/src/psi4/dfocc/tei_grad_corr.cc
```diff
@@ -278,15 +281,15 @@ void DFOCC::tei_grad_corr() {
...
-            const double *Nz = buffer + 8 * ncart;
+            const double *Px = libint1 ? buffer + 0 * ncart : buffers[0];
+            const double *Py = libint1 ? buffer + 1 * ncart : buffers[1];
+            const double *Pz = libint1 ? buffer + 2 * ncart : buffers[2];
+            const double *Mx = libint1 ? buffer + 3 * ncart : buffers[3];
+            const double *My = libint1 ? buffer + 4 * ncart : buffers[4];
+            const double *Mz = libint1 ? buffer + 5 * ncart : buffers[5];
+            const double *Nx = libint1 ? buffer + 6 * ncart : buffers[6];
+            const double *Ny = libint1 ? buffer + 7 * ncart : buffers[7];
+            const double *Nz = libint1 ? buffer + 8 * ncart : buffers[8];
```
susilehtola: and here

psi4/src/psi4/dfocc/tei_grad_ref.cc
```diff
@@ -133,12 +134,14 @@ void DFOCC::tei_grad_ref() {
...
-        const double *Qy = buffer + 4 * ncart;
-        const double *Qz = buffer + 5 * ncart;
+
+        bool libint1 = Process::environment.options.get_str("INTEGRAL_PACKAGE") != "LIBINT2";
+        const double *Px = libint1 ? buffer + 0 * ncart : buffers[0];
+        const double *Py = libint1 ? buffer + 1 * ncart : buffers[1];
+        const double *Pz = libint1 ? buffer + 2 * ncart : buffers[2];
+        const double *Qx = libint1 ? buffer + 3 * ncart : buffers[3];
+        const double *Qy = libint1 ? buffer + 4 * ncart : buffers[4];
+        const double *Qz = libint1 ? buffer + 5 * ncart : buffers[5];
```
susilehtola: and here

psi4/src/psi4/dfocc/tei_grad_ref.cc
```diff
@@ -278,15 +283,15 @@ void DFOCC::tei_grad_ref() {
...
-            const double *Nz = buffer + 8 * ncart;
+            const double *Px = libint1 ? buffer + 0 * ncart : buffers[0];
+            const double *Py = libint1 ? buffer + 1 * ncart : buffers[1];
+            const double *Pz = libint1 ? buffer + 2 * ncart : buffers[2];
+            const double *Mx = libint1 ? buffer + 3 * ncart : buffers[3];
+            const double *My = libint1 ? buffer + 4 * ncart : buffers[4];
+            const double *Mz = libint1 ? buffer + 5 * ncart : buffers[5];
+            const double *Nx = libint1 ? buffer + 6 * ncart : buffers[6];
+            const double *Ny = libint1 ? buffer + 7 * ncart : buffers[7];
+            const double *Nz = libint1 ? buffer + 8 * ncart : buffers[8];
```
susilehtola: and here

psi4/src/psi4/libmints/sointegral_twobody.h
```diff
@@ -686,6 +686,21 @@ void TwoBodySOInt::compute_shell_deriv1(int uish, int ujsh, int uksh, int ulsh,
...
+        const double *pAz = libint1 ? aobuffer + 2 * nao : buffers[2];
+        const double *pBx = libint1 ? aobuffer + 0 * nao : buffers[3];
+        const double *pBy = libint1 ? aobuffer + 0 * nao : buffers[4];
+        const double *pBz = libint1 ? aobuffer + 0 * nao : buffers[5];
+        const double *pCx = libint1 ? aobuffer + 3 * nao : buffers[6];
+        const double *pCy = libint1 ? aobuffer + 4 * nao : buffers[7];
+        const double *pCz = libint1 ? aobuffer + 5 * nao : buffers[8];
+        const double *pDx = libint1 ? aobuffer + 6 * nao : buffers[9];
+        const double *pDy = libint1 ? aobuffer + 7 * nao : buffers[10];
+        const double *pDz = libint1 ? aobuffer + 8 * nao : buffers[11];
```
susilehtola: and here

psi4/src/psi4/dfmp2/corr_grad.cc
andysim: These are all going to be removed when Libint1 goes away.  I just needed to hack something in temporarily to get both libint1 and libint2 to compile simultaneously, for debugging and testing.

susilehtola:
I forget if I've already asked this, so I'll ask it here again.

Libint2 has native support for evaluating range-separated repulsion integrals. Especially, in addition to the common `erfc` kernel, it also can evaluate the integrals for the Yukawa kernel.

It would be awesome if this was also enabled in Psi4. Libxc has a few range-separated functionals that assume the Yukawa kernel.

andysim:
@susilehtola it does, and we do plan to interface all available integrals.  I also have plans to generalize the Erfc integrals to a form like `(a + b erf(omega r)/r` so we don't have to compute two sets of integrals for the range separated functionals

cmake/psi4OptionsTools.cmake
```diff
@@ -62,6 +62,12 @@ macro(psi4_add_module binlib libname sources)
     PRIVATE
       tgt::lapack
     )
+  if(MSVC)
```
jturney, line 65: Was there a reason my MSVC had a special case here? I know it’s been a while since it was added. Just curious.

susilehtola:
Hm, is there a requirement on the ordering used in Libint2? This may complicate packaging...

susilehtola:
Assuming non-default ordering in libint2 will mean that distributions will either have to ship a separate, incompatible copy of libint2 just for psi4.

It would be better if the order used in Psi4 could be customized. At the worst case, this would just mean reshuffling the integrals.

loriab:
> Assuming non-default ordering in libint2 will mean that distributions will either have to ship a separate, incompatible copy of libint2 just for psi4.

> It would be better if the order used in Psi4 could be customized. At the worst case, this would just mean reshuffling the integrals.

If it was only the spherical, then I imagine some selector could be done like in gau2grid. But reshuffling probably isn't a near-term solution for psi4. But wouldn't that still be solving only a third of the problem? Between cart, sph, and shell_set, there's at least five combinations in known use of the twenty possible. https://github.com/evaleev/libint/pull/148/files#diff-4cfbbb105ca3b371b7251a6e37a56f74R31-R41 Isn't the integral library the sensible point for customization? I guess the default distributed libint2 is only mpqc-compatible?

psi4/src/psi4/libmints/eri.cc
```diff
@@ -179,10 +179,31 @@ Libint2ERI::Libint2ERI(const IntegralFactory *integral, double screening_thresho
...
+        braket_ = libint2::BraKet::xx_xx;
+    } else if (!dummy1 && dummy2 && !dummy3 && !dummy4) {
+        braket_ = libint2::BraKet::xs_xx;
+    } else if (!dummy1 && !dummy2 && !dummy3 && dummy4) {
+        braket_ = libint2::BraKet::xx_xs;
+    } else if (!dummy1 && dummy2 && !dummy3 && dummy4) {
+        braket_ = libint2::BraKet::xs_xs;
+    } else {
+        throw PSIEXCEPTION("Bad BraKet type in Libint2TwoElectronInt");
+    }
```
andysim, line 198: This block doesn't hurt, but I don't think it's necessary - the base class's `common_init` should have been called by this time, so `braket_` should already be set by now.

psi4/src/psi4/libmints/eri.cc
```diff
@@ -179,10 +179,31 @@ Libint2ERI::Libint2ERI(const IntegralFactory *integral, double screening_thresho
...
+
     for (int der = 0; der <= deriv; ++der) {
-        engines_.emplace_back(libint2::Operator::coulomb, max_nprim, max_am, der, engine_precision);
+        // set braket upon engine construction so particular LIBINT2_MAX_AM_eri[|2|3] limit governs validity
+        engines_.emplace_back(libint2::Operator::coulomb, max_nprim, max_am, der, engine_precision, libint2::operator_traits<libint2::Operator::coulomb>::default_params(), braket_);
```
andysim, line 202: Neat!  A good way to preserve the `emplace_back` machinery

psi4/src/psi4/libmints/eri.cc
```diff
@@ -179,10 +179,31 @@ Libint2ERI::Libint2ERI(const IntegralFactory *integral, double screening_thresho
...
     }
-    schwarz_engine_ = libint2::Engine(libint2::Operator::coulomb, max_nprim, max_am, 0, max_precision);
+    // set max_am for primary basis to be sieved, not all basis1234
+    max_am = std::max(std::min(basis1()->max_am(), basis2()->max_am()), std::min(basis3()->max_am(), basis4()->max_am()));
```
andysim, line 205: I think this can be simplified to `max_am = bra_same_ ? basis1()->max_am() : ket_same_ ? basis3()->max_am() : 0;`. Those variables are defined upstream in twobody.cc; essentially if we find a matching pair the bra, that's what will be used for the sieve, else if we find a matching pair in the ket, the ket basis will be used.  If neither bra nor ket match, the integral object doesn't do any sieving - there's some specialized sieve code in the DF JK object than handles things like (A0|B0) for aux integrals (where 0 is the empty basis).

susilehtola: @andysim I'm not sure that's simpler :laughing:

andysim: Oops, I was misreading.  @loriab's solution gives the exact same results and, as you point out, is probably easier to read.  My concern was that it would end up generating high angular momentum requirements for DF integrals but it does indeed yield 0.  Probably best to leave it as-is

psi4/src/export_pcm.cc
```diff
@@ -26,6 +26,7 @@
  * @END LICENSE
  */
 
+#include <libint2/shell.h>
```
robertodr, line 29: Why is this include needed here?

psi4/src/psi4/dfocc/tei_grad_ref.cc
```diff
@@ -0,0 +1,335 @@
+/*
```
JonathonMisiewicz, line 1: This file is not used in `dfocc` as of #1953 and can be deleted.

Congratulations on passing tests.

andysim: Thanks for the tip! It'll be gone in the next commit.

psi4/src/psi4/libmints/CMakeLists.txt
```diff
@@ -74,10 +74,17 @@ psi4_add_module(lib mints sources)
...
+    Libint2::cxx
   )
 
 # Conditionally linked-to external projects
+if(TARGET Libint::libint)
+  target_link_libraries(mints
+    PUBLIC
+      Libint::libint
+    )
+endif()
```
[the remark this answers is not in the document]
loriab: it is -- remove in L1 pass

cmake/psi4OptionsTools.cmake
loriab: I tested again and win fails without.

psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1672,15 +1669,15 @@ void RDFMP2::form_Amn_x_terms() {
             int oN = basisset_->shell(N).function_index();
 
             int ncart = cP * cM * cN;
```
hokru, line 1671: `ncart` still needed?

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -1119,15 +1083,16 @@ void DFJKGrad::build_Amn_x_terms()
             int oN = primary_->shell(N).function_index();
 
             int ncart = cP * cM * cN;
```
hokru, line 1085: also here the `ncart` question.

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -1524,27 +1482,37 @@ void DFJKGrad::compute_hessian()
                 int ny = 3 * Ncenter + 1;
                 int nz = 3 * Ncenter + 2;
 
-                size_t stride = static_cast<size_t> (Pncart) * Mncart * Nncart;
+                size_t stride = static_cast<size_t>(Pncart) * Mncart * Nncart;
```
hokru, line 1485: `stride` still needed?

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -1660,67 +1627,71 @@ void DFJKGrad::compute_hessian()
             int Qy = 3 * Qcenter + 1;
             int Qz = 3 * Qcenter + 2;
 
-            size_t stride = static_cast<size_t> (Pncart) * Qncart;
+            size_t stride = static_cast<size_t>(Pncart) * Qncart;
```
hokru, line 1630: 'stride` still needed?

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -1660,67 +1627,71 @@ void DFJKGrad::compute_hessian()
...
-                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, ptr+4*stride, nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Qy][oP*na*na], na*na);
-                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, ptr+5*stride, nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Qz][oP*na*na], na*na);
+                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, const_cast<double*>(Pxbuf), nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Px][oP*na*na], na*na);
+                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, const_cast<double*>(Pybuf), nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Py][oP*na*na], na*na);
+                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, const_cast<double*>(Pzbuf), nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Pz][oP*na*na], na*na);
+                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, const_cast<double*>(Qxbuf), nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Qx][oP*na*na], na*na);
+                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, const_cast<double*>(Qybuf), nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Qy][oP*na*na], na*na);
+                C_DGEMM('n', 'n', nP, na*na, nQ, 1.0, const_cast<double*>(Qzbuf), nQ, Ba_ijp[oQ], na*na, 1.0, &dea_ijp[Qz][oP*na*na], na*na);
```
hokru, line 1664: batman!

psi4/src/psi4/dfocc/tei_grad.cc
```diff
@@ -199,15 +198,15 @@ void DFOCC::tei_grad(std::string aux_type) {
             int oN = primary_->shell(N).function_index();
 
             int ncart = cP * cM * cN;
```
hokru, line 200: `ncart` needed?

hokru (approved):
Few questions/comments, else lgtm.
Caveat is that I am neither an expert on the integral machinery nor C++ magic.
I also tested the local compilation of the branch last week on my mac, including libint2 from scratch.

psi4/src/psi4/libmints/integral.cc
```diff
@@ -84,7 +82,7 @@ void IntegralFactory::set_basis(std::shared_ptr<BasisSet> bs1, std::shared_ptr<B
     bs4_ = bs4;
 
     // Use the max am from libint
-    init_spherical_harmonics(LIBINT_MAX_AM + 1);
+    init_spherical_harmonics(7);
```
jturney, line 85: Does libint2 provide a macro for the max am?

psi4/src/psi4/libmints/integral.cc
```diff
@@ -203,65 +201,93 @@ OneBodyAOInt* IntegralFactory::electric_field(int deriv) {
...
     if (integral_package == "SIMINT" || integral_package == "ERD")
         outfile->Printf("Chosen integral package " + integral_package +
                         " unavailable.\nRecompile with the appropriate option set.\nFalling back to Libint");
+#ifdef ENABLE_Libint1t
     return new ERI(this, deriv, use_shell_pairs);
+#endif
 }
 
-TwoBodyAOInt* IntegralFactory::eri(int deriv, bool use_shell_pairs) {
+TwoBodyAOInt* IntegralFactory::eri(int deriv, bool use_shell_pairs, bool needs_exchange) {
```
jturney, line 224: I think some of these functions need to be cleaned up.

psi4/src/psi4/libmints/integral.cc
loriab: I think the difficulty is that l2 isn't loaded at this point. I did reference this in https://github.com/psi4/psi4/pull/1721/files#diff-f35a0bcb0ee227178fc9064649dd02291985af93a5995f2b4bb08e507e0d8985R24. We could probably boost this a couple more for safety (after all, it would have been 9, I think, for l1), but 12 shows performance problems.

jturney: I remember that now.  Yeah, maybe bump it up ... but once everything is switched to libint2 I'm not sure this will even be needed any more.

susilehtola: I'm pretty sure that's a compile-time option declared in the libint2 headers?

psi4/src/psi4/libmints/integral.cc
loriab: was it anything in particular to `needs_exchange` that needed cleaning? I think the follow-up pass to drop ERD and L1 will help here.

jturney: Yeah, I think you're right that the follow up pass to drop ERD and L1 will do it.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -56,24 +60,277 @@ TwoBodyAOInt::TwoBodyAOInt(const IntegralFactory *intsfactory, int deriv)
...
+     * in bra or ket.  So DF integrals like (A0|RS) with 0 denoting the dummy index will have a sieve
+     * constructed from the RS pair in the ket that can be used to build dense indexing of the RS target.
+     * The other type of DF integral (A0|B0) will not have any sieve associated with it.  Where dummy indices
+     * are present they must occupy the 2nd and/or 4th index.
+     */
+
+    // Different approaches are possible, so we use a function pointer and set it once, to avoid logic later on
+    switch (screening_type_) {
+        case ScreeningType::CSAM:
+            sieve_impl_ = [=](int M, int N, int R, int S) { return this->shell_significant_csam(M, N, R, S); };
```
jturney, line 153: Nice use of lambda functions! 👍

psi4/src/psi4/libmints/eri.h
```diff
@@ -84,6 +110,9 @@ class TwoElectronInt : public TwoBodyAOInt {
     //! Computes the fundamental
     Fjt* fjt_;
 
+    //! The number of integrals in the current shell quartet
+    int batchsize_;
```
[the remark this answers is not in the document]
loriab: My guess is that's `batchsize_` is there to be comparable to siminteri.h. There it is a `size_t`, so I changed it here as well.

psi4/src/psi4/libmints/integral.cc
loriab: I've changed it to 8. 9 sometimes added 20s to the total time of a tu1 * 50. I think we tried to pull it in directly from l2 once , but there were problems. Sounds like this may be unnecessary soon.

psi4/src/psi4/dfocc/tei_grad.cc
loriab: occ and dfocc we live alone on small matters until we can reconcile with Ugur.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -56,24 +60,277 @@ TwoBodyAOInt::TwoBodyAOInt(const IntegralFactory *intsfactory, int deriv)
...
+    screening_threshold_ = Process::environment.options.get_double("INTS_TOLERANCE");
+    auto screentype = Process::environment.options.get_str("SCREENING");
+    if (screentype == "SCHWARZ")
+        screening_type_ = ScreeningType::Schwarz;
+    else if (screentype == "CSAM")
+        screening_type_ = ScreeningType::CSAM;
+    else
+        throw PSIEXCEPTION("Unknown screening type " + screentype + " in TwoBodyAOInt()");
+    
+    if (screening_threshold_ == 0.0) screening_type_ = ScreeningType::None;
```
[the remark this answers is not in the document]
loriab: I'm not sure how much of a problem this is. I think we are seeking to test for the repr of float 0. And l2 is the realm where 1.e-34 vs 1.e-150 can mean something.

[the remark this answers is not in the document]
hokru: Not sure either. Allowing super small numbers for the screening value might make it difficult working with epsilon functions. A safer option is checking for negative numbers (to my believe) instead of equal zero. If you or Andy  think it's fine as it is then it's ok for me, too.

jturney (approved):
Monumental work!

psi4/src/psi4/libfock/PKmanagers.cc
```diff
@@ -219,7 +219,7 @@ void PKManager::compute_integrals(bool wK) {
     }
 
     size_t nshqu = 0;
-#pragma omp parallel for num_threads(nthreads_) schedule(dynamic) reduction(+ : nshqu)
+//#pragma omp parallel for num_threads(nthreads_) schedule(dynamic) reduction(+ : nshqu)
```
andysim, line 222:
```suggestion
#pragma omp parallel for num_threads(nthreads_) schedule(dynamic) reduction(+ : nshqu)
```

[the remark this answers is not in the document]
andysim: Nope! Good catch - I was debugging, and wanted to eliminate any possibility of threading causing issues.

psi4/src/psi4/scfgrad/jk_grad.cc
andysim: Good catch!  I think there will be quite a few of these kinds of loop variables that are not needed, now that the new buffer structure is in place.  They will be optimized away so they're harmless.  If I see a lot of compiler warnings I'll go take a pass through and get rid of them.

andysim (approved):
I shouldn't really review this one, but a *big* thanks to those who weighed in with code and/or reviews.  There's some cleanup that will be needed, but this PR is waaaaaaay too big already, so I vote for it being merged as-is, then we can take ERD and Libint1 out in separate stages, as well as reinstating any F12 integrals.

psi4/src/psi4/scfgrad/jk_grad.cc
hokru: Lori removed all those I found already, most superfluous `stride`/`ncart` should be gone. I did a quick 'find all' search while reviewing.

hokru:
I agree with Andy about merge-as-is. Also more people than just the listed reviewers looked at (parts of) the code according to github.

Before we merge, just wanting to confirm if the right `conda install libint2 -c psi4/label/dev` available?

loriab:
> Before we merge, just wanting to confirm if the right conda install libint2 -c psi4/label/dev is available?

I hope so -- they should be: https://anaconda.org/psi4/libint2/files

PR #1728, opened 2019-09-28
QCSchema Wavefunction Quantities

dgasmith:
## Description
Adds the new QCSchema Wavefunction properties to the Psi4 schema runner.

## Todos
- [x] Creates a schema BasisSet object
- [ ] ECP Basis quantities
- [x] Maps Psi4 arrays to schema `return_result` quantities
- [x] Reorders Psi4 arrays to CCA ordering for both spherical and cartesian basis functions.
- [ ] Provides support for post-SCF `return_result`s

Not entirely sure how we can correctly map post-SCF quantities. It becomes difficult to understand when new orbitals/densities are created. Thinking through a few options here, but if anyone has suggestions let me know!

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -197,6 +197,81 @@ def _convert_variables(data, context=None, json=False):
     return ret
 
 
+def _convert_basis(basis):
+    """Converts a Psi4 basis object to a QCElemental basis.
+    """
+    centers = []
+    symbols = []
+
+    # Loop over centers
+    for c in range(basis.molecule().natom()):
+        center_shells = []
+        symbols.append(basis.molecule().symbol(c).title())
+
+        # Loop over shells *on* a center
+        for s in range(basis.nshell_on_center(c)):
+            shell = basis.shell(basis.shell_on_center(c, s))
+            if shell.is_pure():
+                dtype = "spherical"
+            else:
+                dtype = "cartesian"
+
+            # Build the shell
+            coefs = [[shell.coef(x) for x in range(shell.nprimitive)]]
+            exps = [shell.exp(x) for x in range(shell.nprimitive)]
+            qshell = qcel.models.basis.ElectronShell(angular_momentum=[shell.am],
+                                                     harmonic_type=dtype,
+                                                     exponents=exps,
+                                                     coefficients=coefs)
+            center_shells.append(qshell)
+
+        centers.append(qcel.models.basis.BasisCenter(electron_shells=center_shells))
+
+    # Take unique to prevent duplicate data, doesn't matter too much
+    hashes = [hash(json.dumps(centers[x].dict(), sort_keys=True)) for x in range(len(centers))]
+
+    uniques = {k: v for k, v in zip(hashes, centers)}
+    name_map = {}
+    counter = {}
+
+    # Generate reasonable names
+    for symbol, h in zip(symbols, hashes):
+        if h in name_map:
+            continue
+
+        if symbol in counter:
+            counter[symbol] += 1
+        else:
+            counter[symbol] = 1
+
+        name_map[h] = f"{basis.name()}_{symbol}{counter[symbol]}"
+
+    center_data = {name_map[k]: v for k, v in uniques.items()}
+    atom_map = [name_map[x] for x in hashes]
+
+    ret = qcel.models.BasisSet(name=basis.name(), center_data=center_data, atom_map=atom_map)
+    return ret
+
+
+def _convert_wavefunction(wfn, context=None):
+
+    # Build the correct translation units
+    if context is None:
+        needed_vars = {}
+        for v in _qcschema_translation.values():
+            needed_vars.update(v)
+    else:
+        needed_vars = _qcschema_translation[context]
+
+    ret = {"basis": _conver_basis(wfn.basis()), "restricted": (wfn.same_a_b_orbs() and wfn.same_a_b_dens())}
+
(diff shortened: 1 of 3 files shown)


dgasmith:
@loriab Ready for review.

psi4/driver/schema_wrapper.py
```diff
@@ -85,7 +86,7 @@
     "scf": {
         "scf_one_electron_energy": {"variables": "ONE-ELECTRON ENERGY"},
         "scf_two_electron_energy": {"variables": "TWO-ELECTRON ENERGY"},
-        "scf_dipole_moment": {"variables": ["SCF DIPOLE X", "SCF DIPOLE Y", "SCF DIPOLE Z"]},
+        "scf_dipole_moment": {"variables": ["SCF DIPOLE X", "SCF DIPOLE Y", "SCF DIPOLE Z"], "conversion_factor": (1 / 1 / qcel.constants.dipmom_au2debye)},
```
loriab: `1 / 1`?

psi4/driver/schema_wrapper.py
```diff
@@ -167,13 +169,22 @@ def _convert_variables(data, context=None, json=False):
...
+        conversion_factor = var.get("conversion_factor", 1.0)
+
         # Get the actual variables
         if isinstance(var["variables"], str):
             value = data.get(var["variables"], None)
+            if value:
+                if isinstance(value, (int, float)):
+                    value *= conversion_factor
+                elif isinstance(value, (core.Matrix, core.Vector)):
+                    value.scale(conversion_factor)
```
loriab: always multiply rather than `conversion_factor` devault to None, then `if conversion_factor: convert`?

psi4/driver/schema_wrapper.py
```diff
@@ -197,6 +208,145 @@ def _convert_variables(data, context=None, json=False):
...
+    # Loop over centers
+    for c in range(basis.molecule().natom()):
+        center_shells = []
+        symbols.append(basis.molecule().symbol(c).title())
+
+        # Loop over shells *on* a center
+        for s in range(basis.nshell_on_center(c)):
+            shell = basis.shell(basis.shell_on_center(c, s))
+            if shell.is_pure():
+                dtype = "spherical"
```
loriab: "dtype" not in data type role. perhaps "htype"

psi4/driver/schema_wrapper.py
```diff
@@ -197,6 +208,145 @@ def _convert_variables(data, context=None, json=False):
...
+
+    # Generate reasonable names
+    for symbol, h in zip(symbols, hashes):
+        if h in name_map:
+            continue
+
+        if symbol in counter:
+            counter[symbol] += 1
+        else:
+            counter[symbol] = 1
```
loriab:
```
counter = defaultdict(int)
for symbol, h ...:
    counter[symbol] += 1
```

psi4/driver/schema_wrapper.py
```diff
@@ -197,6 +208,145 @@ def _convert_variables(data, context=None, json=False):
...
+
+        name_map[h] = f"{basis.name()}_{symbol}{counter[symbol]}"
```
loriab: because psi doesn't allow setting basis by atom number, you could use `atom.symbol() + atom.label()` for uniqueness. but this is more general anyways.

dgasmith: This would always generate a unique label for each atom though? We want to create only unique labels per unique atom/basis combo.

psi4/driver/schema_wrapper.py
dgasmith: I thought it would be a lot more logic, but isn't too bad.

psi4/driver/schema_wrapper.py
loriab: Bad diction. I meant what you have is more general anyways. And your scheme will generate fewer labels than my comment.

dgasmith:
@psi4/admins Please do review when possible.

psi4/src/export_mints.cc
```diff
@@ -1411,7 +1411,7 @@ void export_mints(py::module& m) {
     typedef int (BasisSet::*ncore_no_args)() const;
     typedef int (BasisSet::*ncore_one_arg)(const std::string&) const;
 
-    py::class_<BasisSet, std::shared_ptr<BasisSet>>(m, "BasisSet", "Contains basis set information")
+    py::class_<BasisSet, std::shared_ptr<BasisSet>>(m, "BasisSet", "Contains basis set information", py::dynamic_attr())
```
jturney: I didn't know about `dynamic_attr` might be useful in other places.

psi4/driver/schema_wrapper.py
```diff
@@ -197,6 +209,142 @@ def _convert_variables(data, context=None, json=False):
...
+            else:
+                htype = "cartesian"
+
+            # Build the shell
+            coefs = [[shell.coef(x) for x in range(shell.nprimitive)]]
+            exps = [shell.exp(x) for x in range(shell.nprimitive)]
+            qshell = qcel.models.basis.ElectronShell(angular_momentum=[shell.am],
+                                                     harmonic_type=htype,
+                                                     exponents=exps,
+                                                     coefficients=coefs)
```
andysim: Just a comment here - the coefs you're grabbing are those that are normalized according to the CCA spec, not those found in the basis set spec.  Many (most?) other codes expect a different normalization convention, which is why we have the erd_coefficients around also.

PR #1738, opened 2019-10-21
Dipder

stvogt:
## Description

assemble_dipder_from_dipole function to compute dipole derivatives using finite differences.
Modified one test to not forgive IR intensities when dertype is 1.


loriab:
the `git` problem in is hand ...

loriab:
replaced by #1740

PR #1740, opened 2019-10-23
Dipder6

stvogt:
## Description

assemble_dipder_from_dipole function to compute dipole derivatives using finite differences.
Modified test to not forgive IR intensities when dertype is 1.

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -190,15 +190,14 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
     derivative, wfn = derivfunc(method, return_wfn=True, molecule=clone, **kwargs)
     displacement["energy"] = core.variable('CURRENT ENERGY')
 
-    # Getting the dipole moment and saving it a numpy array.
-    displacement["dipole"] = np.array(
-        [core.variable('CURRENT DIPOLE X'),
-         core.variable('CURRENT DIPOLE Y'),
-         core.variable('CURRENT DIPOLE Z')]) / constants.dipmom_au2debye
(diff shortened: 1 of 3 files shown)


psi4/driver/driver.py
```diff
@@ -188,6 +189,11 @@ def _process_displacement(derivfunc, method, molecule, displacement, n, ndisp, *
     derivative, wfn = derivfunc(method, return_wfn=True, molecule=clone, **kwargs)
     displacement["energy"] = core.variable('CURRENT ENERGY')
 
+    # Getting the dipole moment and saving it a numpy array.
+    displacement["dipole"] = np.array([core.variable('CURRENT DIPOLE X'),
+                                      core.variable('CURRENT DIPOLE Y'),
+                                      core.variable('CURRENT DIPOLE Z')])/constants.dipmom_au2debye
```
dgasmith: Please YAPF this file when you have a moment.

stvogt: Should I also YAPF the driver_findif.py, which is the other file I modified?  Also should I YAPF only portions I modifed or the whole file? Because YAPF suggest changes to code I didn't write.

loriab: if it's a wall of changes, no; if its under a dozen, yes, imo.

loriab:
finite difference got rearranged so much in #1351 that this would have been a weird rebase if added to master, so changes have been committed directly to DDD branch with credit to @stvogt in https://github.com/psi4/psi4/pull/1351/commits/ac0cd27470ae1dabf00f37473c9a8dcd74fd368a .

Issue #1033, opened 2018-06-06
Problems with ADC(2) in Psi4?


ssh2:
Hi everyone,

I was testing the ADC(2) code in Psi4 on small molecules and noticed that the results I obtained were significantly different from Q-Chem. The SCF and MP2 energies match very well.

For N2 (r = 1.2 A) with cc-pVDZ basis, here's a summary of the energies for the 8 lowest-lying singlet states from Q-Chem 4.4.1 compared to Psi4 1.1:

Q-Chem 4.4.1:
MP2 ground-state energy: -109.2579498247
  Excitation energy:                                               8.385211 eV
  Excitation energy:                                               8.385211 eV
  Excitation energy:                                               8.832813 eV
  Excitation energy:                                               9.441957 eV
  Excitation energy:                                               9.441957 eV
  Excitation energy:                                              13.868446 eV
  Excitation energy:                                              13.868446 eV
  Excitation energy:                                              14.791430 eV

Psi4 1.1: 
MP2 ground-state energy: -109.25794982519506
->	1  A state   :  0.3113289 (a.u.),  8.4716885 (eV)
->	2  A state   :  0.3113289 (a.u.),  8.4716885 (eV)
->	3  A state   :  0.3343892 (a.u.),  9.0991916 (eV)
->	4  A state   :  0.3565394 (a.u.),  9.7019279 (eV)
->	5  A state   :  0.3565394 (a.u.),  9.7019279 (eV)
->	6  A state   :  0.5069575 (a.u.), 13.7950140 (eV)
->	7  A state   :  0.5069575 (a.u.), 13.7950140 (eV)
->	8  A state   :  0.5127312 (a.u.), 13.9521243 (eV)

The Q-Chem energies agree with my standalone code. For N2 with STO-3G basis I simply get zero excitation energies with Psi4:

    ==> ADC(2) Computation <==                       
                                                     
->  1  A state   :  0.0000000 (a.u.),  0.0000000 (eV)
    Non-iterative:  0.3864160 (a.u.), 10.5149136 (eV)
       ┆   ┆ Occ Vir        Coefficient              
    ---------------------------------------------    
                                                     
    Converged in   2 iteration.                      
    Squared norm of the S component:  1.0000000      
    The S vector is rotated up to 90.000 (deg.)      
    #WARNING: Strongly rotated from the CIS state!   
                                                     
->  2  A state   :  0.0000000 (a.u.),  0.0000000 (eV)
[...]

PR #1744, opened 2019-10-27
Integrating adcc with Psi4

mfherbst:
## Description
The purpose of this PR is to integrate with [adc-connect/adcc](https://github.com/adc-connect/adcc/) inside Psi4 and allow Psi4 users to directly run the ADC calculations supported inside adcc. The opposite direction (i.e. using Psi4 SCF results in adcc) is already implemented and working, see the [adcc manual](https://adc-connect.org/calculations.html#general-adc-n-calculations), so this should mainly be an effort of connecting the strings together on the Psi4 side.

Currently I am still very much in the planning phase of this PR, thankful for any early pointers or feedback to get me moving in a sensible direction. Since this is my first PR here I still feel a little lost inside the Psi4 source :smile:.

My implementation so far hooks into the driver python code of Psi4 and branches out to either `adcc` or the built-in ADC module of Psi4.

## Todos
- [x] Get ADC(2) calculation triggered in adcc from Psi4 (see `adcc_examples` subfolder).
- [x] Allow all ADC variants to be run in Psi4
- [x] Proper error handling of Psi4 SCF references not supported in adcc (symmetry, density-fitting, ...)
- [x] Integration of adcc installation into Psi4 build system
    - Currently adcc can only be installed via `pip`, but we would love to get `conda` support working on our end.
- [x] Tests and examples
- [x] Test install process
- [x] Documentation

## Questions
- [x] How to best branch between the builtin module and adcc? The two implementations share some features, but use different approaches to solve ADC, which can in some cases lead to differing results (related to #1033 )
- [x] How to best direct options and configurables from Psi4 and adcc? The current solution I see is to add basically everything to the big `read_options.cc`, which is pretty inflexible and could lead to a lot of duplicated information being necessary. Is there some better way to do this?
- [x] What to do with the calculation results? Right now only the MP2 ground state energy is returned to the python world in ADC calculations of Ps4 if I understand this right. Our [ExcitedStates](https://adc-connect.org/reference.html#adcc.ExcitedStates) object, however, is capable of returning more results, most importantly oscillators strengths, or the complete excitation vector. I tried to understand what was done in the eom-cc case and as far as I can judge, the `CCEnergyWavefunction` seems at least related. In principle it should be possible to do something similar for ADC inside the Psi4 code base. Another option is also to just blindly reduce the object we produce in `adcc`, which of course could be considered leaking out interface details from adcc into Psi4.
-  Are the tests copied to samples automatically?

## Checklist
    - [x] ADC(1)
    - [x] ADC(2)
    - [x] ADC(2)-x
    - [x] ADC(3)
    - [x] CVS-ADC(1)
    - [x] CVS-ADC(2)
    - [x] CVS-ADC(2)-x
    - [x] CVS-ADC(3)
    - [x] Frozen core variant
    - [x] Frozen virtual variant
    - [x] singlet and triplet states

## Status
- [x] Ready for review
- [x] Ready for merge

Thanks in advance for your help :smile:


hokru:
This would be wonderful addition! Make sure to join also the slack channel (see README for link) for more immediate help and questions.

There are different levels of integration. An independent plugin, a deep C-layer integration and a python-level integration. Some of the questions depends on what you want to aim for (or need). 

Should `adcc` be distributed/build with psi4 in the end? (conda, cmake, etc.)

mfherbst:
Thanks for pointing me to the slack channel. I usually don't use slack, but I'll try to get something set up.

> There are different levels of integration. An independent plugin, a deep C-layer integration and a
> python-level integration. Some of the questions depends on what you want to aim for (or need).

Well, also depends a bit what you guys have in mind and how deep we want the integration to be. My thoughts:
- A **C-layer integration** is possible, but might be quite hard, since our iterative solvers are written in `python`. Given one falls back to solvers in C from psi4, in theory one can integrate adcc into the C core as well at the moment, but our long-term plan is to shuffle more and more to the python layer, so it's most likely not going to be a very sustainable idea. On the upside it would lead to a deep integration of the codes and allow to re-use parts of adcc elsewhere, e.g. the tensor library we are using (I know you are working on xtensor, but just an example).
- **Independent plugin** or **python-level integration**. I'm not sure what the difference is. Maybe you can detail?
What I primarily had in mind was to be able to use adcc from psi4 integrating with the psi4-way to set up and run calculations. My ideas was to be able to use the psi4 interface to setup ADC-related settings for adcc and transfer defaults or preferences from the psi4 setup to adcc. Right now you can already use psi4 SCF results in adcc (see e.g. [this example](https://github.com/adc-connect/adcc/blob/master/examples/water/psi4_ccpvdz_adc2_spectrum.py)), but some parameters (like convergence tolerances) need to be manually tweaked to agreement in both programs. That's an annoyance I'd like to see resolved.
In turn the results from adcc are right now in our own python datastructure. From  this they could be easily transformed in some other datastructure more useful / agreeing with the rest of psi4. So far I was, however, unable to deduce what this would be from the psi4 code base. Of course one could also just return it to the user as is. Again the idea being that a psi4 user can use adcc implicitly for calculations, without needing to worry that he/she does.

> Should adcc be distributed/build with psi4 in the end? (conda, cmake, etc.)

Not sure what will work best. We are working on getting adcc distributed via conda ourselves, so I guess one could just pull it from there as part of the usual psi4 installation procedure. If the interfacing is done on the python level as I described above and as I did in the PR so far, this should not introduce big changes. One could even easily make the dependency optional. I am happy to hear other suggestions, too :smile:. Since we have a mandatory binary blob (in the form of the `adccore` library) for the moment, building the code completely as part of the psi4 setup is not possible unfortunately.

hokru:
Thanks. I mainly wanted to get the discussion going. Slack can be used also in the browser.

psi4 is very modular and having `adcc` as a fully optional dependency, c-layer or python-layer, is fine.

The `psi4 plugins` (see http://psicode.org/psi4manual/master/prog_newcode.html) are independent of the psi4 code base, though can still use psi4's libraries. Examples would be `sns-mp2` or `v2rdm_casscf`. There you can basically do whatever you want.

With python-layer integrations I mean that you add to the proper psi4 python code base. This is what you already did. http://psicode.org/psi4manual/master/proc_py.html
Check the managed method section for how multiple implementations of the same method are usually handled.

You only need the `wfn` object?

hokru:
The `scf_helper` can automatically convert an SCF calculation done with symmetry into C1. Most post-scf DF methods require this already.

mfherbst:
Thanks for the pointers, I'll take a look.

> You only need the `wfn` object?

Yes and no. To bootstrap the calculation in adcc that's all what is needed. Down the line we also need the `MintsHelper` (see our [psi4 backend code](https://github.com/adc-connect/adcc/blob/master/adcc/backends/psi4.py)) to get electronic dipoles and ERI integrals in MO.


> The `scf_helper` can automatically convert an SCF calculation done with symmetry into C1. Most post-scf DF methods require this already.

Perfect!

dgasmith:
As a note our general take it to shuffle more up to the Python layer as well, especially [solvers](https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/solvers.py). The plugin approach is defiantly the best solution I think. It has been a bit, but we should be able to register codes with Psi4 automatically to gain functionality through `psi4.energy/gradient/etc`.

@loriab Can make the final call, but I suspect if adcc is on conda, we will just add it to our meta package of `psi4-rt` that will pull down all the run time dependancies (including this one) so there shouldn't be too much of a need for close coupling. 

--

The Wavefunction object has the ability to store arrays, matrices, and values in a key:value fashion. Can you give a bit of a description of how complex your return objects are? You may want to consider adding to the MolSSI [QCSchema](https://github.com/MolSSI/QCSchema) project the properties that you are returning so that we can have a schema-based definition of these objects that other codes can use and conform to.

mfherbst:
Thanks for the pointers from all of you, I feel I have a much better idea what to do. Btw: Great plugin system, but after thinking about it for a bit I am no longer too sure it is the best solution for this case.

--

On first sight [snsmp2](https://github.com/DEShawResearch/sns-mp2/blob/master/snsmp2/) seems to be a good example project to orient along. So I would make another python module, `psi2adcc` say, and in it implement the routing between psi4 and adcc. Then adcc is one of its `setup.py` dependencies. In this python module I will need to adapt `psi4.driver.procedures['energy']['adc2']` and similar and add my energy function. So far, so good.

Now the problem I see is that in a vanilla Psi4 version without adcc-support, `psi4.driver.procedures['energy']['adc2']` is assigned to `run_adc`, which rightfully runs the ADC(2) implementation in Psi4. If my plugin is loaded, however, this should become a `select_adc2` method, which either picks the `run_adc` in my plugin or the `run_adc` in Psi4, depending on some user settings.

One option to resolve this is to give the adcc methods a different name, e.g. `adcc-adc(2)`. This is roughly what is done e.g. in the `gpu_dfcc` plugin to distinguish against the CPU versions if I understand correctly. For me this solution does not feel very satisfying, however, because now it's again the user's responsibility to select the right ADC implementation.
Another option is to implement a dummy `select_adc` in Psi4, which does no branching and always falls to the built-in implementation. This could then be monkey-patched in my plugin by wrapping it somehow ... but that is not clean either. 

Alternative is to integrate deeper into the python layer and subject to the availablity of the `adcc` module either have two branches in `select_adc` or just one. That's what I will pursue now if there are no other thoughts.

--

About the wavefunction object: I have not looked at the Psi4 implementation details (or the schema) so far, but roughly speaking it's (for each excited state) the excitation energy, a few property vectors and AO density matrices. Theoretically one could also export the full excitation vector, but that involves copying a rank-4 tensor, so I would not do that. But one can export important amplitudes, which is a mapping from a pair / quadruple of integers to a real value.

hokru:
Would it make sense to add a `qc_module adcc` in this case?
Then a new `select_adc` in proc.py could manage both the  (old) `run_adc` and `run_adcc`. 

Though having an optional/external `qc_module` is unprecedented I think.

--

Isn't the `mints.mo_eri` routine becoming quickly a bottleneck?

mfherbst:
That's indeed what I had in mind. Add a `qc_module` and a `select_adc` branching between `adcc` and the builtin. I guess one can just use `psi4.addons` in `select_adc` to query the availability of `adcc` before going into the branching. That was at least my idea so for.

--

> Isn't the `mints.mo_eri` routine becoming quickly a bottleneck?

You mean because it's not parallel at the moment? Yes, that's indeed a problem and was a deal breaker for some of the things we wanted to try in the past ... I have already discussed this with @maxscheurer and he promised to have a look once we have gotten a few other things out of the way.

hokru:
`mints.ao_eri` (which is called underneath for the AOs) is not parallel yes, but also the AO-MO transform is done all in-core.

mfherbst:
I'm not so much worried about that, because adcc is completely in-core as well at the moment :smile:.

dgasmith:
There are suggestions that our ADC code is currently incorrect and we have marked it, we could consider deprecating our own ADC module in favor of this approach. Otherwise the select adc approach is the correct one. For now I would inject a name of `adcc-adc` or similar so that it is currently a unique identifier for testing and we can correctly wire up the select functionality later.

Dictionary objects are fine to return to the user, the Wavefunction is a PyBind11 dynamic object so you can set additional objects to it `wfn.mydata = {"hello": "world"}`. I would take this approach for the first pass.

mfherbst:
@dgasmith Just that I understand correctly. You suggest to:
- Develop a standalone `psi42adcc` python plugin for Psi4 with `adcc` as a dependency in the `setup.py`. In this all the mapping psi4 <-> adcc happens.
- Inside its `__init__.py` inject a `run_adcc_adc`  method under unique method keys such as `adcc-adc(2)` into the proc mapping.
- Expose calculation results via the `wfn` using dynamic attributes

dgasmith:
Yes on the first two items with a few tweaks. We can add you as an endorsed plugin [here](https://github.com/psi4/psi4/blob/fbb2ff444490bf6b43cb6e027637d8fd857adcee/psi4/driver/endorsed_plugins.py). I wouldn't write a new python module just for the binding, as long as the API is stable we could consider adding it directly to the endorsed plugin file. You could also consider reverse hooks where you have a method like `adcc.register_psi4()` which would hook into Psi4, but be called from the endorsed plugin file. 

A few ways to do it, if your API is still in a bit of flux then it probably makes more sense for the code to live in your current repo.

Yes to the third item :)

mfherbst:
I made a little progress and briefly want to summarise` briefly:

- `ADC(2)` is denoted as `adc` inside the Psi4 source code. This does not really make sense to me, since there are now plenty of more ADC methods, so I renamed it to `adc(2)`. `adc` now no longer exists as a method.
- I'm a bit confused about the `mtd_type`. If I get it right, neither the builtin (nor adcc) supports anything but `CONV`(entional) ERIs. But I might be wrong about the builtin code. I copied the value from `MP_TYPE`, since after all the energies returned by the `energy(adcN)` calls are `MP(N)` energies (ADC(n) is constructed for excited states building on top of an MP(n) ground state)). Still, one could introduce `ADC_TYPE` ... you're call.
- I have not registered `run_adcc` for MP(2) and MP(3), but I could because we can compute these energies (and a few properties), too. Not sure that's worth the effort, however.
- For determining frozen core and frozen virtual orbital count I used `Wavefunction.frzcpi` and `Wavefunction.frzvpi`, but it feels a bit strange to do it this way. Is there a better method?
- I added quite a couple of extra options. Is the only way to register them, to add them to the `C++` `read_options.cc`?
- adcc prints by default to `sys.stdout`, but can be easily redirected such that all output goes to another python stream. Psi4 wants to print via `core.print_out`, which takes only a string, so I'm forced to redirect a python stream to a text string. That could become messy unless there's a better option on the Psi4 end.

Again thanks for your help :smile:

hokru:
You are correct that the `adc` is only available as `conv`. Thus `mtd_type` is not necessarily needed for now. Unless `adcc` will support it sometime soon ;-) 

You mentioned an inconsistent density being returned. This is OK but should be mentioned in the manual. Currently within psi4, e.g. simple MP2 single points also just return the SCF density.

maxscheurer:
With @loriab's help, I managed to deploy `adcc` to conda.
There is a dev/temporary version available via
```conda install -c adcc/label/dev_mkl adcc```
for testing purposes. Don't rely on it, though, it may disappear... :smile:
Make sure to install it in an existing `psi4` conda environment.

**Comment by mfherbst** [comment]
A few updates on my progress:
- Most features I'd like to see available from adcc in Psi4 are now from my point of view. We're still working on UHF support for Psi4 (see adc-connect/adcc#20), which only involves adcc-side changes. With PR 20 merged, I'd consider the adcc-Psi4 interface feature complete for now as well.
- Apart from the tests requiring UHF, all tests I have committed are passing on my machine. The testdata is generated using adcc and pyscf, so no overlap with the Psi4 interface code I add here.
- Installation of adcc from the Psi4 buildsystem is untested, since it requires adcc `v0.13.2` to be tagged on our end. We'll do that after PR 20 is merged on our end.

With that said: Feel free to be nitty on my code if you find something :smile:.

**Comment by dgasmith** [comment]
What is left for this PR and what level of integration are you shooting for in this first pass?

File: psi4/driver/procrouting/proc.py [hunk 1]
```diff
@@ -3061,6 +3098,269 @@ def run_adc(name, **kwargs):
...
+    as a wavefunction.
+
+    """
+    # TODO Maybe it would improve readability if this function was spilt
+    #      up and the whole thing went to a separate file (like for sapt,
+    #      interface_cfour.py, ...
+
+    if not extras.addons("adcc"):
+        # TODO or PsiImportError ???
+        raise ValidationError("adcc extras module not available")
```
dgasmith: ""adcc extras module not available, please install via ..."

```
try:
    import adcc
except ModuleNotFoundError:
    raise ...

Comment (mfherbst) [reply psi4/driver/procrouting/proc.py]: Good point. Thanks.

**Comment by mfherbst** [comment]
> What is left for this PR and what level of integration are you shooting for in this first pass?

Running the aforementioned methods from Psi4 and being able to "see" the results in the output file and get the most important lot of them (dipole moments, oscillator strength, energies) from the `Wavefunction` object. Support for restricted and unrestricted references. Smooth selection between builtin ADC (only for ADC(2) on RHF and singlet excitations) and adcc (for all the rest). That's not fully there (see my TODO). If you are ok with using more options than in other `select_` cases for the `select_adc2` method, than I'll go ahead and implement that.

On that note: I noticed that already for small and simple systems such as cc-pvdz water the adcc ADC(2) and the builtin ADC(2) differ. Since we compare and test our code very thoroughly to reference data from Q-Chem/adcman including test cases of such complexity and derived properties, I would be quite surprised if the error was on our end. I'm not sure what to do with that. At best this is confusing for a user and one should definitely make a statement about in the manual, but still ... obviously one implementation is wrong.

I have not worked on the ADC documentation inside Psi4 so far. I am not sure what kind of detail you want in the manual, I'd be very brief, explaining only the Psi4-option for ADC and refer to [adc-connect.org](https://adc-connect.org) for the details on ADC itself. If I understand correctly the main file I have to work on is `doc/sphinxman/source/adc.rst`.

Our PR 20 is now merged. We'll proceed with the tagging of a release soon.

File: psi4/driver/procrouting/proc.py (hunk 1 above)
Comment (mfherbst) [reply psi4/driver/procrouting/proc.py]: Fixed now.

**Comment by mfherbst** [comment]
For note: I have enabled the installation of adcc (via pip) on Travis via `-DENABLE_adcc=ON` for testing purposes.

**Comment by hokru** [comment]
It would be nice for the users to have a memory estimate and check at some point.
Unless `adcc` does the checking itself.

----


> On that note: I noticed that already for small and simple systems such as cc-pvdz water the adcc ADC(2) and the builtin ADC(2) differ.

You already mentioned #1033. I'd agree that psi4's `adc` should be disabled for the time being. Perhaps a point to discuss at the upcoming psicon @dgasmith 
Would't necessarily affect this PR though.

**Comment by mfherbst** [comment]
>It would be nice for the users to have a memory estimate and check at some point.
Unless adcc does the checking itself.

I agree, but so far we do not have this feature in adcc and we don't check. If you run low on memory adcc will more or less just fail. It's a little tricky to estimate that beforehand in adcc due to the lazy import and computational nature of adcc, importing things only as needed. I could be very crude and come up with something rough for the workflows exposed to psi4, but I'd rather think about this more properly and do it at a later point, unless you think it's a major blocker.

**Comment by hokru** [comment]
A rough check is better than none. Especially because people won't expect a memory demanding in-core implementation. Remember that no one reads manuals :-) 
I think your `adcc` could get quite popular and people want to try it out.  Good idea to make it a smooth experience from the start.

**Comment by mfherbst** [comment]
With the last sentence I agree 100% :smile: ... I'll come up with something.

**Comment by mfherbst** [comment]
**Update**: Tests now pass in Azure Linux builds including a few quicktests for adcc. I'll proceed to add documentation when I get round to it (I'm a bit busy at the moment).

**Comment by mfherbst** [comment]
@hokru, respective the memory estimate: I did some experiments and came up with a very rough estimate for the lower bound of a computation. I'll add a check and a warning in the next adcc release if the estimate is a substantial fraction of the main memory. Thanks for your suggestion in the first place. The respective change in adcc will be transparent to the interface needed for this PR.

**Comment by mfherbst** [comment]
I think this is now ready for review.

File: psi4/driver/procrouting/proc_table.py [hunk 2]
```diff
@@ -113,7 +113,6 @@
         'fci'           : proc.run_detci,
         'casscf'        : proc.run_detcas,
         'rasscf'        : proc.run_detcas,
-        'adc'           : proc.run_adc,
```
dgasmith: To prevent a deprecation can you add something like:

```python
'adc'           : proc.adc_deprecation,

def adc_deprecation(*args, **kwargs):
     warnings.warn("The method 'adc' has been deprecated please use .... The method 'adc' will be deprecated in Psi4 1.6.", DeprecationWarning)
     return proc.select_adc2(*args, **kwargs)
```

@loriab Do we wrap deprecation warnings these days?

mfherbst: Thanks I fixed that.

maxscheurer:
Can we get two more reviews for this PR, please? 🎆

dgasmith:
@psi4/editors

psi4/driver/procrouting/proc.py
```diff
@@ -988,6 +989,46 @@ def select_mp4(name, **kwargs):
         return func(name, **kwargs)
 
 
+def select_adc2(name, **kwargs):
+    """Function selecting the algorithm for ADC(2) excited state energy
+    call and directing to specified or best-performance default modules.
+
+    """
+    reference = core.get_option('SCF', 'REFERENCE')
+    mtd_type = core.get_global_option('MP_TYPE')  # TODO or ADC_TYPE or CC_TYPE or ...
```
loriab: different names are mostly for enough knobs to influence single jobs with complex formulation like cbs. mp3 and adc seem pretty orthogonal, so `MP_TYPE` sounds good for now.

psi4/driver/procrouting/proc.py
```diff
@@ -988,6 +989,46 @@ def select_mp4(name, **kwargs):
...
+    module = core.get_global_option('QC_MODULE')
+    # Considering only adcc/adc
+
+    # TODO Actually one should do selection on a couple of other options here
+    #      as well, e.g. adcc supports frozen-core and frozen-virtual,
+    #      spin-specific states or spin-flip methods.
```
loriab: frozen-core as a selector is on my near-term todo list, so if you'd sketch in the logic, even in comments, that'd be great.

psi4/driver/procrouting/proc.py
```diff
@@ -988,6 +989,46 @@ def select_mp4(name, **kwargs):
...
+
+    func = None
+    if reference == 'RHF':
+        if mtd_type == 'CONV':
+            if module == 'ADCC' and extras.addons("adcc"):
+                func = run_adcc
+            elif module in ['', 'BUILTIN']:
+                func = run_adc
+        # TODO Does the builtin support any other MP type?
```
loriab: may switch `BUILTIN` to `ADC`, but leave for now.

No, I'm pretty sure it's adc2 only.

psi4/driver/procrouting/proc.py
```diff
@@ -988,6 +989,46 @@ def select_mp4(name, **kwargs):
...
+    # Note: ROHF is theoretically available in adcc, but are not fully tested
+    #       ... so will be added later.
+
+    if func is None:
+        raise ManagedMethodError(['select_adc2', name, 'MP_TYPE', mtd_type, reference, module])
+
+    if kwargs.pop('probe', False):
+        return
+    else:
+        return func(name, **kwargs)
```
loriab: thanks for the select fn. logic looks good. can add a table section https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/notes_c.rst if you like.

psi4/driver/procrouting/proc.py
```diff
@@ -3061,6 +3108,271 @@ def run_adc(name, **kwargs):
...
+    # Build kwargs for adcc
+    #
+    kwargs.pop("molecule", None)
+
+    if ref_wfn.frzcpi()[0] > 0:
+        kwargs["frozen_core"] = ref_wfn.frzcpi()[0]
+    if ref_wfn.frzvpi()[0] > 0:
+        kwargs["frozen_virtual"] = ref_wfn.frzvpi()[0]
+    if core.get_option("ADC", "NUM_CORE_ORBITALS"):
+        kwargs["core_orbitals"] = core.get_option("ADC", "NUM_CORE_ORBITALS")
```
loriab: All: should we try to make this defaultable from quantities in wfn and existing electron spaces options?

tests/adcc/cn_adc2/input.dat
```diff
@@ -0,0 +1,31 @@
+#! UADC(2)/cc-pvdz calculation of 5 states for the cyanide radical
```
loriab: exceedingly trivial point, but `_` have been banned from test names in favor of `-`. I can always fix it up later.

psi4/driver/procrouting/proc.py
JonathonMisiewicz: Because, as discussed Wed., I'm currently changing a lot of the existing electron space logic...

What's the difference between frozen core and core orbitals?

mfherbst: > What's the difference between frozen core and core orbitals?

Core orbitals is specific to ADC variants with the core-valence separation approximation applied (used for X-ray spectroscopy for example) and defines how many orbitals to put in this "special" core space. On top of that you may always do classic frozen core as well, so you need two things.

JonathonMisiewicz: Okay, a "wavefunciton" specific orbital space. (I've heard of CVS, but every presentation of linear response I've seen has gone over my head.)

I cannot imagine a sensible default from existing wavefunction quantities/spaces.

For what it's worth, a PR I'm working on will allow wavefunctions to define custom orbital spaces to override the current frozen core / occupied / virtual / frozen virtual which is imposed on all wavefunctions, fitting or not. But that space still needs to be filled somehow, and the PR won't give any new ways to do that.

tests/adcc/cn_adc2/input.dat
mfherbst: Sure, no problem. I fixed that.

psi4/driver/procrouting/proc.py
mfherbst: Well, it's pretty simple: If I get it right adcc has frozen-core and frozen virtual, but the builtin has not, so one would go for adcc if it is available if any frozen core is desired.

psi4/driver/procrouting/proc.py
mfherbst: I'm not sure if switching `BULTIN` to `ADC` is such a good idea, because then you have `module` being either `ADC` or `ADCC`, which is pretty easy to mix up in my opinion.

Ah, sorry, what I meant was whether the builtin supports anything else, but `CONV`, but that has been answered before ... it only supports `CONV`.

psi4/driver/procrouting/proc.py
mfherbst: No problem.

psi4/driver/procrouting/proc.py
mfherbst: If you want to think of it like so, yes. 

Core orbitals is not really defaultable, because it depends on what you want to compute. See https://github.com/mfherbst/psi4/blob/integrate-adcc/doc/sphinxman/source/adc.rst under **Using the core-valence separation**.

By the way: CVS is not just a concept of linear response or propagator theory ... people in CI use it, too ... they just use different names for it (like restricted channel or something similar). It just means that you neglect some couplings in the matrix describing your states and thus are able to focus on state, which would otherwise come much higher in energy, embedded in the valence continuum.

psi4/driver/procrouting/proc.py
```diff
@@ -3061,6 +3108,271 @@ def run_adc(name, **kwargs):
...
+    # TODO Maybe it would improve readability if this function was spilt
+    #      up and the whole thing went to a separate file (like for sapt,
+    #      interface_cfour.py, ...
+
+    try:
+        import adcc
+        from adcc.backends import InvalidReference
+    except ModuleNotFoundError:
+        raise ValidationError("adcc extras qc_module not available. Try installing "
+            "via 'pip install adcc' or 'conda install -c adcc adcc'.")
```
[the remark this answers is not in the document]
andysim: Sorry, I tried to review this over a week ago but I just discovered that comments are not visible to you until I `finalize review` - I thought that button was for final acceptance.  Time for ☕️ 

Anyway, this is fantastic work and please don't mistake my inability to use GitHub reviews correctly with a lack of interest.  This will add a number of great methods and you've done a great job with documentation and tests 😄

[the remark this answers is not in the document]
mfherbst: Yes that's currently still needed to pull adcc from conda.

@maxscheurer Correct me if I am wrong.

tests/adcc/cn-adc1/input.dat
```diff
@@ -0,0 +1,31 @@
...
+
+# Alternative: Run computation of properties as well
+properties('adc(1)', properties=["oscillator_strength", "dipole"])
+
+ref_gs = -92.21289215243166                                                            #TEST
+ref_energies = [0.1597543246960989, 0.15975432469610037, 0.2391697500304988,           #TEST
+                0.2744917588674049]                                                    #TEST
+ref_oscillator = [0.007747129119166235, 0.007747129119166262, 3.8625889729943975e-06,  #TEST
+                  4.999193313301502e-10]                                               #TEST
+compare_values(ref_gs, energy_adc, 7, "ADC(1) GS energy")                              #TEST
```
loriab: Are you sure energies will be converged to 1.e-7 w/o tightening convergences? We sometimes get problems when switching architectures. If uncertain, there's some guidance in last paragraph of http://psicode.org/psi4manual/master/add_tests.html#faq-add-tests.

doc/sphinxman/source/adc.rst
```diff
@@ -38,38 +38,276 @@
...
+* |adc__max_num_vecs|: Specifies the maximal number of subspace vectors
+  in the Jacobi-Davidson scheme before a restart occurs. The defaults are usually
+  good, but do not be shy to increase this value if you encounter convergence problems.
+* |adc__num_guesses|: By default adcc uses twice as many guess vectors as
+  states to be computed. Sometimes increasing this value by a few vectors can be helpful.
+  If you encounter a convergence to zero eigenvalues, than decreasing this parameter might
+  solve the problems.
+
+.. _`sec:adcbuiltin`:
+
```
loriab: Many thanks for good docs with both theoretical background and practical tips.

psi4/driver/procrouting/proc.py
```diff
@@ -3061,6 +3109,271 @@ def run_adc(name, **kwargs):
...
+        lines += [ind + "Hartree-Fock (HF)"]
+        lines += [ind + ind + format_vector("Dipole moment (in a.u.)", hf.dipole_moment)]
+
+        if state.method.level > 1:
+            lines += [ind + "Møller Plesset 2nd order (MP2)"]
+            lines += [ind + ind + format_vector("Dipole moment (in a.u.)", mp.dipole_moment(2))]
+            for i, cart in enumerate(["X", "Y", "Z"]):
+                adc_wfn.set_variable("MP2 dipole " + cart, mp.dipole_moment(2)[i])
+                core.set_variable("MP2 dipole " + cart, mp.dipole_moment(2)[i])
+                core.set_variable("current dipole " + cart, mp.dipole_moment(2)[i])
```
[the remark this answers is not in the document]
mfherbst: Good idea. That sounds simpler and more extensible.

psi4/driver/procrouting/proc.py
```diff
@@ -3061,6 +3109,271 @@ def run_adc(name, **kwargs):
...
+
+    if "OSCILLATOR_STRENGTH" in properties:
+        data = state.oscillator_strengths.reshape(-1, 1)
+        computed["Oscillator strength (length gauge)"] = data
+        adc_wfn.set_variable("oscillator strengths", core.Matrix.from_array(data))
+
+    if "DIPOLE" in properties:
+        data = state.state_dipole_moments
+        computed["State dipole moment (in a.u.)"] = data
+        adc_wfn.set_variable("dipoles", core.Matrix.from_array(data))
```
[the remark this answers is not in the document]
mfherbst: Not sure I understand what you are referring to. What do you mean, for example, in this precise case?

psi4/driver/procrouting/proc.py
```diff
@@ -3061,6 +3109,271 @@ def run_adc(name, **kwargs):
...
+        core.print_out("Ground state energy breakdown:\n")
+        core.print_out("    Energy             SCF   {0:15.8g} Ha\n".format(ref_wfn.energy()))
+        for level in range(2, state.method.level + 1):
+            if level >= 3 and is_cvs_adc3:
+                continue
+            energy = mp.energy_correction(level)
+            mp_corr += energy
+            adc_wfn.set_variable(f"MP{level} correlation energy", energy)
+            adc_wfn.set_variable(f"MP{level} total energy", mp.energy(level))
+            core.print_out(f"    Energy correlation MP{level}   {energy:15.8g} Ha\n")
```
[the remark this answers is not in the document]
mfherbst: Not really ... I just did something. I'm happy to comply to the Psi4 standard, however.

doc/sphinxman/source/adc.rst
mfherbst: Well, not all is my work in this file, so I can't take all the credit ...

tests/adcc/cn-adc1/input.dat
mfherbst: Good point ... I changed it to 6 digits only.

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
maxscheurer: `conda install -c adcc/label/dev_mkl adcc` would be more suitable at the moment I guess...

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
maxscheurer: I think the variables should be somehow labeled with "ADC" in this case like it's done for CC methods [here](https://github.com/psi4/psi4/blob/5f9977460bf45cc8a7aa69ac86cd0ef9fcf76edb/psi4/driver/procrouting/proc.py#L2810), e.g. `adc_wfn.set_variable("ADC state dipoles", core.Matrix.from_array(data))` (just a suggestion...).
Please correct me if I'm wrong, @loriab.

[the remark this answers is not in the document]
loriab: Yes, like “ADC(3) DIPOLE”, please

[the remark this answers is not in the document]
mfherbst: Ok. I'll use something like `name + " state dipoles" if that's fine.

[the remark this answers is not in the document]
mfherbst: ... and updated.

mfherbst:
I rebased this branch due to conflicts with master ... 

On that note: Could we try to get this into master at some point soon? Of course let me know if there is still work to be done.

dgasmith:
Pinging @psi4/editors again. Please review!

psi4/driver/procrouting/proc_table.py
```diff
@@ -179,31 +187,39 @@
...
+        'cisdt'        : proc.run_detci_property,
+        'cisdtq'       : proc.run_detci_property,
+        'ci'           : proc.run_detci_property,  # arbitrary order ci(n)
+        'fci'          : proc.run_detci_property,
+        'rasscf'       : proc.run_detci_property,
+        'casscf'       : proc.run_detci_property,
+        'omp2'         : proc.select_omp2_property,
+        'omp2.5'       : proc.select_omp2p5_property,
+        'omp3'         : proc.select_omp3_property,
+        'olccd'        : proc.select_olccd_property
```
andysim:
```suggestion
        'olccd'        : proc.select_olccd_property,
```

andysim: This change should fix the CI error

mfherbst: Thanks very much! I did not manage to investigate ;smile:

mfherbst:
Thanks for the merge @loriab and thanks everyone for the help and comments along the way of this PR :smile:.

Issue #1736, opened 2019-10-16
Correlated Derivative Refactor: Planning


JonathonMisiewicz:
I'm posting this for discussion before I start coding anything. I was hoping to have some more details on prefactors, but I've both fallen ill and been given a shift in priorities, so I'm posting what I have in the interests of not wasting more time. I'm hoping to have more to say on the prefactor problem when I've recovered.

@kcpearce's plans for MP2 hessians are related, but not the part of the project that I'll be focusing on.

## Motivation
Several different groups of Psi developers are dissatisfied with Psi's current technology for correlated derivatives.
* At Emory, @fevangelista has recently implemented gradients for DSRG-PT2 via a standalone code interfaced to Psi. Getting those working required consulting three(!) other experienced Psi developers and borrowing code from a fourth because the assumptions of the `libmints/libtrans` machinery are undocumented and confusing.
* At University of Georgia, the DCT code currently has to go through several contortions to be compatible with the gradient code. It has to [rewrite part of libtrans for RHF gradients](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dct/dct_sort_mo_tpdm.cc) because the developer could not figure out what to send libtrans otherwise. It has to [convert LibDPD buffers into LibIWL](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dct/dct_gradient_UHF.cc#L4815) buffers so `libtrans` can [convert those back to LibDPD buffers](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc), but sorted differently. It also has to [re-index](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dct/dct_gradient_UHF.cc#L3839) its own 1PDM and energy-weighted density matrix, and merge [density matrix blocks into a single density matrix](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dct/dct_gradient_UHF.cc#L3789). These tasks shouldn't be necessary.
* @edeprince3 has also [rewritten parts of libtrans](https://github.com/edeprince3/v2rdm_casscf/blob/master/backtransform_tpdm.cc) to get V2RDM CASSCF gradients. As best as I can tell, the rewrite just changes orbital indexing conventions, but I could be mistaken.
[...]

PR #1745, opened 2019-10-30
Gradient Refactor

JonathonMisiewicz:
## Description
Described in great detail [here](https://github.com/psi4/psi4/issues/1736).

## Todos
I see this PR occurring in six major stages.
- [x] Introducing the OPDM/Lagrangian Tech
- [ ] Introducing the Conventional TPDM Tech
- [ ] Migrating Psi to new-style gradients
- [ ] Adding deprecation warnings for old-style gradients
- [ ] Miscellaneous revision corrections

### Step One: OPDM/Lagrangian Tech
This PR introduces a new file, `integraltransform_opdm_lagrangian.cc`, to perform this backtransformation. This was previously done by `integraltransform_oei.cc:trans_one`. As this function has other responsibilities, `trans_one` is preserved. Compared to the previous interface, there are three key changes.
1. Callers can pass in blocks of the OPDM/Lagrangian. This gives callers flexibility to pass in `oo` and `vv` and `ov` and `vo` blocks separately if available, and to not pass in a block that is conjugate to another block already passed in. This is frequently the case.
2. Callers pass in these blocks as a map from orbital space labels to a SharedMatrix containing the relevant block and are responsible for attaching this to the wavefunction object via the `Da`, `Db`, and `X` member variables. This is the most important point for this section! All codes except `cc` already store the OPDM/Lagrangian as `SharedMatrix` objects, so this is the path of least resistance. The one holdout predates `libmints` and instead uses a `libmints` `Matrix`-like structure. (See discussion in Step Four.) Existing code had to perform many an astounding number of type conversions to ultimately convert back to the type we had originally. `dct`, for instance, went `SharedMatrix` -> `file2` -> QTOrdered Block Matrix -> PSIO Entry -> Lower Triangle -> Block Matrix -> Lower Triangle -> PSIO Entry -> `SharedMatrix`. We now stay in a `SharedMatrix` the entire time. This type change is why `integraltransform_oei.cc:trans_one` is not used in the new version. When you use a `Matrix`, the transformation reduces to a `linalg::triplet` call.
3. Lastly, prefactors. Because it's ambiguous what you call the Lagrangian, I'm going to use as a point of reference "the thing you contract against the overlap derivatives", because that is pefectly unambiguous. In the old code, you'd send -2 times that thing. Now, you send -1 times that thing. This is necessary for consistency with how the SCF code determines the value of the Lagrangian on a wavefunction object.

If you want an example for how this works, see the newly added code in `dct_gradient_UHF.cc:compute_ewdm_odc`. With these changes, 30 lines of code replaces about 230. I call this a good start.

### Step Two: Conventional TPDM Tech
I don't see a way around still using `dpdbuf4` for this, even though this isn't the most plugin-friendly. The main simplification here will be eliminating the need to dump to `libiwl`.

## Questions
- [ ] There is a goof in `libtrans`: There is a [pre-defined frozen core space `O`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libtrans/mospace.h#L51-L57), and also a [pre-defined occupied space `o`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libtrans/mospace.h#L58-L68), and likewise for virtual orbitals. These spaces have conflicting labels. 
[Uppercase letters mean the alpha orbitals](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libtrans/integraltransform_moinfo.cc#L306) and [lower case letters mean the beta orbitals](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libtrans/integraltransform_moinfo.cc#L505). We need to change the labels of the frozen core and frozen virtual spaces to break the conflict. We currently have five votes (Lori, Andy, Susi, Jet, and I) for `C` for frozen core and `W` for frozen virtual. Anybody opposed? Also, how on earth do I deprecate _this_? I need this for some frozen core gradient plans, so I would appreciate being able to change this sooner rather than later. Thankfully, neither space seems to be used anywhere in Psi.
- [ ] I made some internal changes to `libtrans`. There is now a case sensitive `MOCoefficients_` map, and there are case-insensitive `aMOCoefficients` and `bMOCoefficients` functions. I'd like to remove the old `aMOCoefficients_` and `bMOCoefficients_` maps, now that we have these new functions. Do I need to formally deprecate them? These were never exposed via PyBind.

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/dct/dct_gradient_UHF.cc
+++ b/psi4/src/psi4/dct/dct_gradient_UHF.cc
@@ -4803,6 +4803,37 @@ void DCTSolver::compute_ewdm_odc() {
         offset += nmopi_[h];
     }
 
+    // Here be example code of the new interface.
+    // auto a_opdm_oo = std::make_shared<Matrix>("MO-basis Alpha OPDM <O|O>", nirrep_, naoccpi_, naoccpi_);
+    // a_opdm_oo->identity();  // The kappa term
+    // a_opdm_oo->add(aocc_tau_);
+    // auto a_opdm_vv = avir_tau_->clone();
+    // a_opdm_vv->set_name("MO-basis Alpha OPDM <V|V>");
+    // std::map<std::array<char, 2>, SharedMatrix> alpha_density = {{{'O', 'O'}, a_opdm_oo}, {{'V', 'V'}, a_opdm_vv}};
+    // Da_ = _ints->backtransform_two_index(alpha_density, "SO-basis Alpha OPDM");
+
+    // auto b_opdm_oo = std::make_shared<Matrix>("MO-basis Beta OPDM <O|O>", nirrep_, nboccpi_, nboccpi_);
+    // b_opdm_oo->identity();  // The kappa term
+    // b_opdm_oo->add(bocc_tau_);
+    // auto b_opdm_vv = bvir_tau_->clone();
+    // b_opdm_vv->set_name("MO-basis Beta OPDM <V|V>");
+    // std::map<std::array<char, 2>, SharedMatrix> beta_density = {{{'o', 'o'}, b_opdm_oo}, {{'v', 'v'}, b_opdm_vv}};
+    // Db_ = _ints->backtransform_two_index(beta_density, "SO-basis Beta OPDM");
+
+    //// Rather than send over the Lagrangian, 0.5(X_pq+X_qp), let's send X_pq and leave the symmetrizing to libtrans.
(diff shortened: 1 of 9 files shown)


JonathonMisiewicz:
Status update for those who are watching this: my original attempt at the four-index TPDM elements using pre-existing `libdpd` capabilities proved slower than the existing `libiwl` algorithm, so I've had to write new `libdpd` algorithms. The new algorithm working on just the `VVVV` block is beating the `libiwl` algorithm for speed. I'll need to modify the algorithm to handle more general blocks, make sure the speed is still good, and then I can commit the next step.

The current plan is to replace the libiwl and presorting steps, but leave the rest of the TPDM logic as-is. This may change.

PR #1755, opened 2019-11-21
Fix some memory leaks

JonathonMisiewicz:
## Description
I was testing a trick I had heard of to clean up valgrind output (which alas, didn't work). I got carried away fixing the memory leaks it reported before deciding that continuing was not the best way to use my time. Don't expect this work to be continued in the immediate future. A "Valgrind Day" just like "LGTM Day" or "Documentation Day" would be very useful.

## Checklist
- [x] Because this was just fixing memory leaks, I'll leave testing to Azure. I can do more if anybody thinks there would be use in it?

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dct/dct_gradient_RHF.cc
+++ b/psi4/src/psi4/dct/dct_gradient_RHF.cc
@@ -179,6 +179,9 @@ void DCTSolver::compute_lagrangian_OO_RHF() {
             }
         }
         global_dpd_->file2_mat_wrt(&X);
+        global_dpd_->file2_mat_close(&pT);
+        global_dpd_->file2_mat_close(&H);
+        global_dpd_->file2_mat_close(&X);
         global_dpd_->file2_close(&pT);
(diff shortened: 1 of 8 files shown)


susilehtola:
Has the code been run extensively through cppcheck? It is also pretty good at finding leaks.

JonathonMisiewicz:
No, there has been no extensive or systematic cppcheck sweep of Psi code in my memory. I think Lori encouraged us to valgrind new code at a PsiCon several years back? I saw the slide.

susilehtola:
Valgrinding is runtime and terribly slow. Cppcheck is static and quite fast.

andysim:
I ran the code through cppcheck in [this PR](https://github.com/psi4/psi4/pull/1534).  It found some memory leaks, but not as many as I'd expected.  I second the call to have a valgrind/cleanup day at some point - there are still a number of mismatched new/malloc/free/delete pairs in the code that should be replaced with a RAII solution.

PR #1756, opened 2019-11-22
Let the user select the type of derivative computation in the Deriv class

fevangelista:
## Description
This PR introduces an option in the call to `Deriv::compute()` so that the user can specify the type of gradient computation. Currently, this class determines which procedure to follow using an internal logic that is not sufficiently flexible to accommodate plugins.

## Todos
- [x] Add deric_calc_type option to `Deriv::compute()`
- [x] Define enum and python interface

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -647,10 +647,10 @@ void export_mints(py::module& m) {
           "B"_a, "C"_a, "transA"_a = false, "transB"_a = false, "transC"_a = false);
 
     py::enum_<DerivCalcType>(m, "DerivCalcType")
-        .value("Default", DerivCalcType::Default)
-        .value("SCF", DerivCalcType::SCF)
-        .value("SCFandDF", DerivCalcType::SCFandDF)
-        .value("Correlated", DerivCalcType::Correlated);
+        .value("Default", DerivCalcType::Default, "Use internal logic.")
+        .value("SCF", DerivCalcType::SCF, "SCF methods.")
+        .value("SCFandDF", DerivCalcType::SCFandDF, "Correlated methods using DF (no reference contribution).")
+        .value("Correlated", DerivCalcType::Correlated, "Correlated methods that write RDMs and Lagrangian to disk.");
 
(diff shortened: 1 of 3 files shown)


psi4/src/export_mints.cc
```diff
@@ -644,6 +646,12 @@ void export_mints(py::module& m) {
           "Returns the multiplication of three matrics A, B, and C, with options to transpose each beforehand", "A"_a,
           "B"_a, "C"_a, "transA"_a = false, "transB"_a = false, "transC"_a = false);
 
+    py::enum_<DerivCalcType>(m, "DerivCalcType")
+        .value("Default", DerivCalcType::Default)
+        .value("SCF", DerivCalcType::SCF)
+        .value("SCFandDF", DerivCalcType::SCFandDF)
+        .value("Correlated", DerivCalcType::Correlated);
```
loriab: fyi, starting fairly recently, one can finally add docstrings to enums in pb11. https://github.com/loriab/pylibefp/blob/master/pylibefp/core.cc#L493

JonathonMisiewicz:
I mentioned this on Slack, but at present, only the Correlated procedure is used in Psi, and the others are legacy code. I have no idea what plugins are doing, but I assumed those all used Correlated as well. Let me know if you're going to change that. I imagine that you're using the SCF procedure for CASSCF purposes?

fevangelista:
We need the correlated procedure to run CASSCF and  may adapt the code to use DF once the new Deriv code is available. In putting together this PR, my intention was to make very limited changes and only improve the functionality of the current code. Since I am not sure what plugins do, I just reproduced the old functionality as it was before. I'd be happy to delete some of the legacy code if we are certain it is not used.

psi4/src/export_mints.cc
fevangelista: Done.

JonathonMisiewicz:
I don't think we _can_ be certain what `Psi_API` code is being used and what isn't... We probably should leave it for now, with the possibility of deprecating that old system entirely when I get the new one working properly.

fevangelista:
@loriab: I indeed forgot to pass the default to python. Pushed a commit and waiting for testing. In the long run the default should disappear.

PR #1759, opened 2019-11-25
Move basis set orthogonalization to libmints/orthog.cc

susilehtola:
## Description
This PR removes duplicated code in Psi4 so that orthogonalization of the basis is handled in one place.

## Todos
- [x] No code duplication
- [x] Everything works as before

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/dimension.h
+++ b/psi4/src/psi4/libmints/dimension.h
@@ -46,7 +46,6 @@ class PSI_API Dimension {
     Dimension();
     Dimension(int n, const std::string& name = "");
     Dimension(const std::vector<int>& other);
-    ~Dimension();
 
     /// Assignment operator, this one can be very dangerous
     Dimension& operator=(const int* other);
--- a/psi4/src/psi4/libmints/dimension.h
+++ b/psi4/src/psi4/libmints/dimension.h
@@ -44,13 +44,9 @@ class PSI_API Dimension {
 
    public:
     Dimension();
-    Dimension(const Dimension& other);
     Dimension(int n, const std::string& name = "");
     Dimension(const std::vector<int>& other);
     ~Dimension();
-
-    /// Assignment operator
-    Dimension& operator=(const Dimension& other);
 
     /// Assignment operator, this one can be very dangerous
     Dimension& operator=(const int* other);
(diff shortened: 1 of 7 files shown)


fevangelista:
@susilehtola: Lately I have been thinking that the best way to do canonical orthogonalization is not the typical one where you select an eigenvalue threshold, but instead one with a guaranteed condition number. This could be easy to implement, it just requires comparing the ratio of the current and largest eigenvalue of S, something like this
```
            if (eigval_->get(h, i) / eigval_max > condnum_tol_) {
```
Have you ever considered implementing something like this? I am not sure if it would make a big difference, but this definition in my opinion is more rigorous.

susilehtola:
@fevangelista yes that would be easy, but it would still be numerically unstable.

Wait for a few hours for the next arXiv update to see the correct solution ;)

psi4/src/psi4/libmints/orthog.cc
```diff
@@ -29,205 +29,165 @@
...
 #include <iterator>
+#include <algorithm>
 
 #include "psi4/psi4-dec.h"
+#include "psi4/libpsi4util/PsiOutStream.h"
 
-#include "matrix.h"
-#include "vector.h"
-#include "orthog.h"
+#include "psi4/libmints/matrix.h"
```
robertodr: These includes were actually fine.

loriab: ok, my fault

susilehtola: Should I revert?

robertodr: I don't think we have strict rules in place and both forms are fine. I've been cleaning up includes in the files I've opened lately and this jumped to my eyes.

susilehtola: I guess this is resolved then?

psi4/src/psi4/libmints/dimension.h
```diff
@@ -44,14 +44,10 @@ class PSI_API Dimension {
 
    public:
     Dimension();
-    Dimension(const Dimension& other);
```
loriab: guess it's not used at present, but this is ordinary copy constructor, so by rule of three should keep this and copy assignment. or maybe since destructor is empty should kill off all three?

susilehtola: I think I killed off all non-trivial ones?

susilehtola: OK, I guess you learn something new every day. Explicit destructor removed.

PR #1760, opened 2019-11-26
Implement partial Cholesky orthogonalization.

susilehtola:
## Description
This PR implements the partial Cholesky orthogonalization procedure I've recently described in https://arxiv.org/abs/1911.10372, which can be used to avoid pathological linear dependencies in the basis set. This is the simple version of the algorithm which just hides the pathological functions; the more sophisticated one is already available in ERKALE but it is limited to fixed geometries.

## Todos
- [x] Partial Cholesky decomposition implemented
- [x] Calculation of basis function extents via `r^2` value implemented
- [x] Basis functions sorted so that steep ones get treated first by the decomposition

## Questions
- [x] How to handle the sorting of basis functions in the presence of symmetry? AFAIK each SO basis functions is a superposition of identical functions on different atoms; this means that the corresponding symmetry-blocked `r^2` values could just be one of these.
- [x] Although matrix.cc already implements pivoted Cholesky, it's also available in LAPACK as `dpstrf()`. Maybe the implementation should be replaced with a call to LAPACK?
- [ ] Would it be possible to make Psi4 drop the irrelevant shells altogether from the basis set to make calculations more efficient? This would need to be done at every geometry...

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libmints/matrix.cc
```diff
@@ -1802,19 +1797,16 @@ void Matrix::svd_a(SharedMatrix &U, SharedVector &S, SharedMatrix &V) {
...
 
-            double *work = new double[(int)lwork];
+            std::vector<double> work((int)lwork);
 
             // SVD
-            info = C_DGESDD('A', n, m, Ap[0], n, Sp, Vp[0], n, Up[0], m, work, (int)lwork, iwork);
-
-            delete[] work;
-            delete[] iwork;
+            info = C_DGESDD('A', n, m, Ap[0], n, Sp, Vp[0], n, Up[0], m, work.data(), (int)lwork, iwork.data());
```
dgasmith: Solid changes here, thank you!

psi4/src/psi4/libmints/orthog.cc
```diff
@@ -40,18 +41,58 @@
 
 namespace psi {
 
-OverlapOrthog::OverlapOrthog(OrthogMethod method, SharedMatrix overlap, double lindep_tolerance, int print)
-    : orthog_method_(method), overlap_(overlap), lindep_tol_(lindep_tolerance), print_(print) {
+OverlapOrthog::OverlapOrthog(OrthogMethod method, SharedMatrix overlap, SharedVector rsq, double lindep_tolerance,
```
dgasmith: Can we write this out as `OverlapOrthogonalization`? Not too much longer and messy it more clean (IMO).

psi4/src/psi4/libmints/matrix.cc
jturney: Agreed!

jturney:
Glad to see the old codes I wrote years ago are brought back to life.

psi4/src/psi4/libmints/orthog.cc
susilehtola: I hope @jturney agrees :)

susilehtola:
> Interesting idea to drop the duplicate shells. We generally consider that the total drop is pretty small and haven't worried about optimizing for this case much. For example coronene aug-cc-pVTZ duplication is only about 15% which is a pretty extreme case. Do you have example where very large numbers of AO's are dropped?

See the paper on arXiv. I think the usual use case would be multiply augmented basis sets that some people use. But the current implementation will go a long way to make sure the program works even for these pathological cases.

loriab:
if you'll rebase now, the Linux `using_mp2d` CI error should go away. there'll also be a Windows spot check section, https://github.com/psi4/psi4/blob/master/.azure-pipelines/azure-pipelines-windows.yml#L191-L197 . If you add your failing Win test there, we should get better output for debugging.

susilehtola:
Looks like the azure failure was caused by something else in psi4, which has now been fixed by a rebase. Ready to merge.

doc/sphinxman/source/scf.rst
```diff
@@ -337,31 +337,57 @@ Orthogonalization
...
+combinations of AO basis functions may be discarded, the number of
+canonical-orthogonalized OSOs and MOs may be slightly smaller than the
+number of AOs.
+
+When the basis set is too overcomplete, the eigendecomposition of the
+overlap matrix is no longer numerically stable. In this case the
+partial Cholesky decomposition can be used to pick a subset of basis
+functions that span a sufficiently complete set, see
+[Lehtola:2019:cholesky]. This subset can then be orthonormalized as
+usual, and the rest of the functions hidden from the calculation.
```
loriab: yay, docs!

loriab:
error is a Windows install of conda that chocolatey seems to be failing connection on today. It was passing CI earlier.

.azure-pipelines/azure-pipelines-windows.yml
```diff
@@ -196,6 +196,14 @@ jobs:
         displayName: 'Spot Test'
         workingDirectory: $(Build.BinariesDirectory)/build
 
+      # Spot Test
+      - script: |
+          setlocal EnableDelayedExpansion
+          set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
+          psi4 $(Build.SourcesDirectory)\tests\scf-cholesky-basis\input.dat
+        displayName: 'Cholesky basis test'
+        workingDirectory: $(Build.BinariesDirectory)/build
```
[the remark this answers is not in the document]
susilehtola: Already removed, dropping the two commits altogether.

tests/scf-cholesky-basis/input.ref
```diff
@@ -0,0 +1,2262 @@
+
```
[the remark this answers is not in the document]
susilehtola: Whoops

susilehtola:
CI passed again

psi4/src/psi4/libmints/orthog.h
```diff
@@ -34,30 +34,56 @@
 
 namespace psi {
 
-class PSI_API OverlapOrthog {
+class PSI_API OverlapOrthogonalization {
```
fevangelista: I was thinking that even though the class name is somewhat self-explanatory, it might be a good idea to add a comment that explains what this class does and how/where it is used. Just thinking about the future programmers.

loriab:
oh dear, the scf_cholesky_basis test is failing again on Windows. That it's intermittent suggests that there's a reference/number-of-decimals-checked/runtime-convergence that's borderline out of whack. Maybe try collecting the ref vals from a very tight run and/or loosening the decimals checked.

doc/sphinxman/source/scf.rst
```diff
@@ -337,31 +337,57 @@ Orthogonalization
...
+the smallest overlap eigenvalue falls below the user-supplied double
+option |scf__s_tolerance|, which defaults to 1E-7. If the smallest
+eigenvalue is below this cutoff, canonical orthogonalization is
+forced, and all eigenvectors corresponding to eigenvalues below the
+cutoff are eliminated.
+
+If the eigendecomposition is detected to be numerically unstable - the
+reciprocal condition number of the overlap matrix to be smaller than
+the machine epsilon - the partial Cholesky decomposition is undertaken
+until |scf__s_cholesky_tolerance|, which defaults to 1E-8.
```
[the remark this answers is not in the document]
andysim: Sorry, I sent this review a few days ago but didn't realize I have to hit "finish review" if not just adding a single comment.  My bad.  The PR looks fantastic overall, but I'm just not sure about this one keyword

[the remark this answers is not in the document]
susilehtola: No, because the Cholesky decomposition threshold controls a wholly different aspect of the algorithm. Having a separate keyword is important, since the Cholesky decomposition method is hiding AO basis functions altogether. `scf__s_tolerance` is used after the Cholesky procedure to orthonormalize the reduced basis set.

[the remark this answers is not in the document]
andysim: Great - thanks for the clarification and for the PR :)

loriab:
This has a record number of approvals, so just need to get over the Windows block. May be time to add it as a Spot test again to see output (`-o stdout` like below is my latest refinement to Windows debugging).

```
      # Spot Test
      - script: |
          set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
          psi4 -o stdout $(Build.SourcesDirectory)\tests\pywrap-cbs1\input.dat
        displayName: 'MP5 Spot Test'
        workingDirectory: $(Build.BinariesDirectory)/build
```

loriab:
Mild reminder to throw in an [extra Windows test](https://github.com/psi4/psi4/pull/1760#issuecomment-568071614) to debug what I hope is a borderline tolerance instability.

loriab:
Some [large ints](https://dev.azure.com/psi4/psi4/_build/results?buildId=1567&view=logs&j=f31c9f97-4411-58e7-49ac-fc73f645e6b6&t=168f406a-25eb-5f8a-7d4d-4c979d542d09&l=728) in the Cholesky decomp matrix that apparently the lower tolerances can manage but the 1e-6 can't

susilehtola:
Lol wtf. That's not possible: it's memory corruption. There's bugs on the Windows platform (gasp).

susilehtola:
On my computer, there's no problem with huge integers. Also, that bug is unrelated to the algorithm: I don't use the Cholesky decomposed overlap, the only thing that matter are the pivot indices.

So, looks like the problem is that in the 5th symmetry block, the Windows build picks up the 3rd SO basis function, while my build picks up the 12th basis function, thus leading to a different energy. The other blocks are the same.

susilehtola:
OK, let's see whether going from double to triple augmentation fixes the discrepancy.

andysim:
If that doesn't work, I have Windows VM that I've been using for Libint2 troubleshooting.  I should be able to find some time to look into it if needed

susilehtola:
@andysim great.

Since the weird thing is that I can't find any errors in the build log that @loriab linked:
```
2020-01-06T17:54:37.8471030Z     Energy with Cholesky 1e-4.........................................PASSED
2020-01-06T17:54:37.8471436Z     Energy with Cholesky 1e-4.........................................PASSED
2020-01-06T17:55:04.0834277Z     Energy with Cholesky 1e-5.........................................PASSED
2020-01-06T17:55:04.0834591Z     Energy with Cholesky 1e-5.........................................PASSED
```
The crash comes suddenly
```
2020-01-06T17:55:21.8639362Z 
2020-01-06T17:55:21.8712539Z     Using partial Cholesky orthogonalization (doi:10.1063/1.5139948).
2020-01-06T17:55:30.8500954Z ##[error]Cmd.exe exited with code '-1073741819'.
2020-01-06T17:55:30.9285461Z ##[section]Finishing: CHolesky basis Spot Test
```

susilehtola:
Ugh. Going to a larger basis set doesn't help: now the energies mismatch already at the 1e-4 Cholesky threshold.

If anyone has any good ideas on how to make the algorithm less susceptible to numerical noise switching the pivot indices, I'm all ears...

susilehtola:
So... now the VS2015 test crashes even before it anything is printed out....

tests/scf-cholesky-basis/input.dat
```diff
@@ -0,0 +1,159 @@
+# Reference energies
+e_cholesky_4  = -230.64594709545503
+e_cholesky_5  = -230.64614462290461
+e_cholesky_6  = -230.64617987370553
+e_cholesky_7  = -230.64620544189239
+e_cholesky_8  = -230.64620513216849
+e_cholesky_9  = -230.64620507251573
+e_cholesky_10 = -230.64620507295609
+e_cholesky_11 = -230.64620504920356
+e_canonical   = -230.64620504921135
```
susilehtola: This is actually quite interesting - but not surprising: the energy goes down going down from large thresholds at the beginning, but then goes up starting from the 1e-7 threshold.

susilehtola:
@andysim please have a look. The Windows test crashes even when the checks are turned off!

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -1975,6 +1967,71 @@ void Matrix::cholesky_factorize() {
...
+    L->zero();
+    for (int h = 0; h < nirrep_; ++h) {
+        for (int m = 0; m < rowspi_[h]; ++m) {
+            for (int n = 0; n < std::min(m + 1, nchol[h]); ++n) {
+              // Psi4 stores matrices as row major, LAPACK as column major
+              L->set(h, m, n, get(h, n, m));
+            }
+        }
+    }
+    // Switch to the properly sized matrix
```
andysim:
```suggestion
    // Switch to the properly sized matrix
    colspi_ = nchol;
```
This should fix the memory bug - I found it using address sanitizer on linux.  There's another error that's flagged after this fix, but occurs in the DFJK code and is likely a false positive.

susilehtola: OK, this is a bug in `matrix.cc`.

susilehtola: Fixed in the copy assignment operator.

andysim: Yes, your fix is better than mine.  I blame a lack of :coffee:

susilehtola:
All checks pass :tada:
Ready to be squashed!

dgasmith:
@andysim should do the honors.

PR #1766, opened 2019-12-12
Remove functions rendered obsolete by matrix_subset_helper

JonathonMisiewicz:
## Description
Discovered while poking around `wavefunction.h` for the ongoing `libtrans` refactor. 150 less lines of code with no effort sounds good to me.

## Todos
- [x] Removed dead code

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
The functions weren't exposed to plugins or Python, so the only downstreams we could possibly break are using locally modified Psi. I'm not worried about them.

PR #1767, opened 2019-12-12
Fix CI

loriab:
## Todos
- [x] bump qcel and qcng
- [x] new `using('<prog>')` API instead of `using_<prog>` from `qcng.testing`
- there's still a mix of selectors right now. will do a simplification pass after some other branches where I've edited pytests are merged in.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/pytests/addons.py
+++ b/tests/pytests/addons.py
@@ -1,7 +1,20 @@
-import subprocess
-
 import pytest
+
 from qcelemental.util import parse_version, which, which_import
+from qcengine.testing import using
+
+__all__ = [
+    'hardware_nvidia_gpu',
+    'using_cppe',
+    'using_dftd3',
+    'using_dftd3_321',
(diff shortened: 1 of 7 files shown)


tests/pytests/addons.py
```diff
@@ -50,7 +45,7 @@ def is_nvidia_gpu_present():
             return ngpu > 0
 
 
-hardware_nvidia_gpu = pytest.mark.skipif(is_nvidia_gpu_present() is False,
+hardware_nvidia_gpu = pytest.mark.skipif(True, #is_nvidia_gpu_present() is False,
```
fevangelista: I assume this is intentional?

loriab: Yes, the thing is broken, though line by line it seems to work. It’s vexed me one time too many.

Issue #1758, opened 2019-11-25
perturb_h has no effect in combination with X2C relativistic corrections


tovrstra:
Possibly the following could (also) be due to my misunderstanding of how the input is supposed to work. My apologies if that is the case. In the following example, the perturbation with the external field seems to be ignored:

```
molecule {
  0 1
  He 0.0 0.0 0.0
  symmetry c1  # field breaks symmetry
}

set {
  reference   rhf
  basis  aug-cc-pvtz
  relativistic x2c
  perturb_h true
  perturb_with dipole
  perturb_dipole [0.0, 0.0, 0.1]
}

energy('scf')
```

I've attached a series of examples, with non-relativistic, dkh and x2c calculations on a dummy test case: [reltest.zip](https://github.com/psi4/psi4/files/3884841/reltest.zip)

```
grep "Total Energy =" *.out
dkh_0.out:    Total Energy =                         -2.8613053380694344
dkh_p.out:    Total Energy =                         -2.8679760471040372
nrl_0.out:    Total Energy =                         -2.8612063084748343
nrl_p.out:    Total Energy =                         -2.8678772073375063
x2c_0.out:    Total Energy =                         -2.8613053511779820
x2c_p.out:    Total Energy =                         -2.8613053511779745
```

The suffix `_0` and `_p` stand for "no field" and "dipolar field" respectively.

loriab:
I'd hazard we need something like 
```        } else if (options_.get_str("RELATIVISTIC") == "X2C") {
            outfile->Printf(" OEINTS: Using relativistic (X2C) overlap, kinetic, and potential integrals.\n");

            if (!rel_basisset_) {
                throw PSIEXCEPTION("OEINTS: X2C requested, but relativistic basis was not set.");
            }    
            X2CInt x2cint;
            SharedMatrix so_overlap_x2c = so_overlap();
            SharedMatrix so_kinetic_x2c = so_kinetic();
            SharedMatrix so_potential_x2c = so_potential();
            x2cint.compute(basisset_, rel_basisset_, so_overlap_x2c, so_kinetic_x2c, so_potential_x2c);

            potential_mat->add(so_potential_x2c);
        }    
```
about [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/mintshelper.cc#L1338) but that's not a literal fix. @fevangelista, is that on the right track?

susilehtola:
[...]

PR #1770, opened 2019-12-13
Keep one-electron integrals in core (MintsHelper)

fevangelista:
## Description
The goal of this PR is to save one-electron integrals in core and cache them.

## Todos
- [x] One-electron integrals are no longer stored on disk. The `Wavefunction` base class holds a pointer to a `MintsHelper` object.
- [x] T, V, and S integrals are cached.
- [x] Address issue #1758 (no effect of perturbation on X2C energy)
- [x] Fixes issue #1794
- [x] May fix issue #1791

## For developers
- [x] Added a `MintsHelper` object to `Wavefunction`.

## Checklist
- [x] Tests added for X2C finite field contributions
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests) except for eight test cases that are linked to #1793

## Status
- [x] Working code
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1215,57 +1163,57 @@ SharedMatrix MintsHelper::mo_spin_eri_helper(SharedMatrix Iso, int n1, int n2) {
...
-        overlap_mat->apply_symmetry(ao_overlap(), petite_list()->aotoso());
-        return overlap_mat;
+        SharedMatrix so_ints_sym(factory_->create_matrix(label));
+        so_ints_sym->apply_symmetry(ao_ints, petite_list()->aotoso());
+        cached_oe_ints_[label] = so_ints_sym;
     }
 }
 
-SharedMatrix MintsHelper::so_kinetic() {
+SharedMatrix MintsHelper::so_overlap_nr() {
```
dgasmith: Should the `_nr` functions be options to the non-`nr`functions?

```
SharedMatrix MintsHelper::so_kinetic(std::string relativistic = "auto")
```

tests/CMakeLists.txt
```diff
@@ -93,7 +93,7 @@ foreach(test_name adc1 adc2 casscf-fzc-sp casscf-semi casscf-sa-sp ao-casscf-sp
                   scf2 scf3 scf4 scf5 scf6 scf7 scf-property serial-wfn soscf-large soscf-ref
                   soscf-dft stability1 dfep2-1 dfep2-2 sapt-dft1 sapt-dft2 sapt-compare sapt-sf1 dft-custom dft-reference
                   stability2 tu1-h2o-energy tu2-ch2-energy tu3-h2o-opt scf-response1
-                  tu4-h2o-freq tu5-sapt tu6-cp-ne2 x2c1 x2c2 x2c3 zaptn-nh2
+                  tu4-h2o-freq tu5-sapt tu6-cp-ne2 x2c1 x2c2 x2c3 x2c-perturb_h zaptn-nh2
```
loriab: no underscores in testcase names, please.

tests/x2c-perturb_h/input.dat
```diff
@@ -0,0 +1,28 @@
+#! Test SFX2C-1e with a static electric field on He aug-cc-pVTZ
+
+eref_x2c_0 = -2.8613053511779594
+eref_x2c_p = -2.8679759489083452
+
+molecule {
+  0 1
+  He 0.0 0.0 0.0
+  symmetry c1  # field breaks symmetry
```
loriab: should we arrange for the presence of a field to impose c1/no_com/no_reorient on the mol? does the field ever not break mol symmetry and we want to preserve symm?

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -192,9 +192,9 @@ def scf_initialize(self):
     if self.attempt_number_ == 1:
         mints = core.MintsHelper(self.basisset())
         if core.get_global_option('RELATIVISTIC') in ['X2C', 'DKH']:
-            mints.set_rel_basisset(self.get_basisset('BASIS_RELATIVISTIC'))
+            mintshelper = self.mintshelper()
+            mintshelper.set_rel_basisset(self.get_basisset('BASIS_RELATIVISTIC'))
```
loriab: I think stashing the rel_basis on wfn.basissets() is pretty general/robust. Is it the interaction with mintshelper that's troubling?

tests/CMakeLists.txt
fevangelista: Fixed it!

psi4/src/psi4/libmints/mintshelper.cc
fevangelista: I found it hard to come up with a good solution here. The problem is that in X2C one has to compute S, T, and V all at once and this computation needs the non-relativistic S, T, and V.  I used this somewhat redundant form to clearly separate the function that handles both relativistic and non-relativist cases from the actual computation of non-relativistic overlap. I can adopt the solution that you suggest, but that will introduce a recursive calling to the `so_kinetic` function. If that's not considered too bad then I can give it a try. Let me know.

tests/x2c-perturb_h/input.dat
fevangelista: I think there are cases when you might want the user to determine how to lower the symmetry. And in some cases fields do not break symmetry (e.g. filed in the z direction applied to water, C2v). There is some code to make sure that the computation is setup correctly, e.g. (around line 1321 of `mintshelper.cc`)
```
    if (lambda[0] != 0.0) {
        if (msymm.component_symmetry(0) != 0) {
            outfile->Printf("  WARNING: Requested mu(x) perturbation, but mu(x) is not symmetric.\n");
        } else {
            outfile->Printf("  Perturbing V by %f mu(x).\n", lambda[0]);
            dipoles[0]->scale(lambda[0]);
            potential_mat->add(dipoles[0]);
        }
    }
```

psi4/driver/procrouting/scf_proc/scf_iterator.py
fevangelista: This is a good point. `MintsHelper` does not store a wavefunction object, so that's why I am storing a copy of the relativistic basis in it. So the alternative is storing a `Wavefunction` object in `MintsHelper`. I am not sure which options is more sound. Any suggestion?

loriab: I think my original comment was in response to your 

> Passing relativistic_basis is done in an awkward way, by piping it into MintsHelper in the SCF code (py-side). What would be a good way to deal with it that is more general/robust?

and me trying to figure out what problem you were raising. Wfn owning a mintshelper owning a Wfn sounds like the old "which comes first? the molecule or the basisset" problem. Maybe mintshelper can get a basissets dict like Wfn https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.h#L90-L91 that gets initialized from Wfn so that each aux basis doesn't need its own slot.

fevangelista: Good suggestion!

PR #1771, opened 2019-12-13
fixes bug in df-cc when reading wave functions with scf_type cd.

edeprince3:
## Description
My DFCC code was relying on Process::environment.globals() for the number of auxiliary functions in the case of scf_type = cd, which failed to work correctly when passing wavefunctions into the energy call.  This PR fixes that bug.

## Checklist
- [x] Quicktests pass. 

## Status
- [x] Ready for review
- [x] Ready for merge


edeprince3:
I think this is good now?

dgasmith:
Yup, thank you!

PR #1772, opened 2019-12-16
Much Healing of DFOCC

### LINKED ISSUE #1143: OCC/DFOCC convergence evaluated unintuitively
In the OCC & DFOCC modules, setting values of 
`E_CONVERGENCE`,`MAX_MOGRAD_CONVERGENCE`, `RMS_MOGRAD_CONVERGENCE` does not produce the expected behavior. 

Rather than requiring `RMS_MOGRAD <= RMS_MOGRAD_CONVERGENCE`, `MAX_MOGRAD <= MAX_MOGRAD_CONVERGENCE`, `DE <= E_CONVERGENCE` to ALL evaluate to `True`, convergence is declared if EITHER `DE <= E_CONVERGENCE` or the MOGRAD convergence is met. 

This leads to situations where tightening `MAX_MOGRAD_CONVERGENCE` has no effect on the result of the calculation, and convergence is declared when `MAX_MOGRAD` is higher than the convergence criterion.

As far as I can tell, the offending lines of code are in occ_iterations.cc L279:282:
```
       if (wfn_type_ != "DF-OLCCD") {
           if (rms_wog < tol_grad && biggest_mograd < mograd_max) break;
           if (std::fabs(DE) <= tol_Eod) break;
       }
```

Removing these lines fixes my issue, but as I am unclear about the purpose of this code block perhaps that is not the best fix.

With original code:
```
 ==============================================================================
 ================ Performing DF-OMP2 iterations... ============================
 ==============================================================================
                    Minimizing DF-MP2-L Functional
                    ------------------------------
 Iter       E_total           DE           RMS MO Grad      MAX MO Grad      RMS T2
 ----    ---------------    ----------     -----------      -----------     ----------
   1     -76.2055179040     -1.60e-03       3.84e-04         1.53e-03        5.38e-05
   2     -76.2055243130     -6.41e-06       3.88e-05         2.04e-04        5.14e-06

 ==============================================================================
 ======================== DF-OMP2 ITERATIONS ARE CONVERGED ====================
 ==============================================================================
```

With lines removed/commented (EXECUTING LOOP print statements added by me during debugging, ignore them):
```
 ==============================================================================
 ================ Performing DF-OMP2 iterations... ============================
 ==============================================================================
                    Minimizing DF-MP2-L Functional
                    ------------------------------
 Iter       E_total           DE           RMS MO Grad      MAX MO Grad      RMS T2
 ----    ---------------    ----------     -----------      -----------     ----------
        EXECUTING LOOP
   1     -76.2055179040     -1.60e-03       3.84e-04         1.53e-03        5.38e-05
        EXECUTING LOOP
   2     -76.2055243130     -6.41e-06       3.88e-05         2.04e-04        5.14e-06
        EXECUTING LOOP
   3     -76.2055243922     -7.91e-08       4.89e-06         2.97e-05        6.34e-07
        EXECUTING LOOP
   4     -76.2055243935     -1.38e-09       6.15e-07         4.23e-06        8.53e-08
        EXECUTING LOOP
   5     -76.2055243936     -2.45e-11       1.61e-07         7.61e-07        1.16e-08
        EXECUTING LOOP
   6     -76.2055243936     -9.95e-13       1.87e-08         1.49e-07        1.92e-09

 ==============================================================================
 ======================== DF-OMP2 ITERATIONS ARE CONVERGED ====================
 ==============================================================================
```

### LINKED ISSUE #1509: DFOCC's Final Energy Dependent on Whether User Energy Convergence Criteria Met
A rather important issue for @bozkaya.

The test suite documents that energies computed with DFOCC can be internally inconsistent. `dfomp2p5-2` demonstrates the problem. The final energy computed via Lagrangian is [-75.7907](https://github.com/psi4/psi4/blob/master/tests/dfomp2p5-2/output.ref#L498). The final energy is reported as [-75.7914](https://github.com/psi4/psi4/blob/master/tests/dfomp2p5-2/output.ref#L530) hartrees, which are inconsistent. The discrepancy disappeared upon elimination of [the lines in dfocc that allow for the user-specified energy convergence criteria to be overridden](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dfocc/occ_iterations.cc#L278-L281), per issue #1143.

The relevant part of the output file is:

```
 ============================================================================== 
 ================ Performing DF-OMP2.5 iterations... ========================== 
 ============================================================================== 
                Minimizing DF-MP2.5-L Functional 
                ------------------------------ 
 Iter       E_total           DE           RMS MO Grad      MAX MO Grad      RMS T2                                                                                 
 ----    ---------------    ----------     -----------      -----------     ---------- 
   1     -75.7906384498     -7.38e-04       1.89e-03         7.57e-03        5.95e-06 
   2     -75.7907030487     -6.46e-05       6.80e-04         3.45e-03        1.29e-06 
   3     -75.7907190240     -1.60e-05       4.26e-04         1.69e-03        5.72e-07 
   4     -75.7907244154     -5.39e-06       2.25e-04         1.39e-03        2.96e-07 
   5     -75.7907266815     -2.27e-06       1.20e-04         1.13e-03        1.64e-07 
   6     -75.7907291572     -2.48e-06       9.37e-05         4.54e-04        2.34e-07 
   7     -75.7907295719     -4.15e-07       2.53e-05         2.18e-04        7.57e-08 

 ============================================================================== 
 ======================== DF-OMP2.5 ITERATIONS ARE CONVERGED ================== 
 ============================================================================== 

    Computing DF-MP2.5 energy using optimized MOs... 
    ======================================================================= 
    Nuclear Repulsion Energy (a.u.)    :     9.18738645758982
    SCF Energy (a.u.)                  :   -75.63188585944289
    REF Energy (a.u.)                  :   -75.63118855740093
    Alpha-Alpha Contribution (a.u.)    :    -0.02415964277065
    Alpha-Beta Contribution (a.u.)     :    -0.12430830985037
    Beta-Beta Contribution (a.u.)      :    -0.01107306313769
    DF-MP3 Correlation Energy (a.u.)   :    -0.16666563305500
    DF-MP3 Total Energy (a.u.)         :   -75.79855149249789
    DF-MP2.5 Correlation Energy (a.u.) :    -0.15954101575872
    DF-MP2.5 Total Energy (a.u.)       :   -75.79072957315965
    ======================================================================= 

    ======================================================================= 
    ================ DF-OMP2.5 FINAL RESULTS ============================== 
    ======================================================================= 
    Nuclear Repulsion Energy (a.u.)    :     9.18738645758982
    DF-HF Energy (a.u.)                :   -75.63188585944289
    REF Energy (a.u.)                  :   -75.63118855740093
    DF-OMP2.5 Correlation Energy (a.u.):    -0.15884371245450
    Edfomp2.5 - Eref (a.u.)            :    -0.15954101449645
    DF-OMP2.5 Total Energy (a.u.)      :   -75.79072957189739
    ======================================================================= 
```

The Lagrangian energy and the final energy now agree perfectly, despite only a tiny change in the Lagrangian energy. However, the key is to note that the lines
```
	Orbitals are optimized now.
	Transforming MOs to the semicanonical basis... 
        Switching to the standard DF-MP2 computation... 
```
are present in the reference output file but not in the output file that I posted. This is due to the if clause [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dfocc/manager.cc#L2648), which requires `std::fabs(DE) >= tol_Eod`, i.e., that the observed delta E is _greater than_ the user-specified convergence criteria. This situation is precisely what the fix to #1143 no longer allows for, so I must conclude that the code in that if suite is what is responsible for the discrepancy.

It looks like this code in this if suite is re-computing the energy using the orbitals and T-amplitudes computed during the Lagrangian optimization. Hence the discrepancy is that when the user's energy convergence criteria isn't met, the Lagrangian information was overwritten with the orbital-and-amplitude formalism. Is this correct? Why does this code only run if the user's energy convergence criteria isn't met? Shouldn't we warn the user that these kinds of discrepancies are possible?

JonathonMisiewicz:
## Description
@hokru recently reported that when he attempted to use `dfomp2` through `dfocc` to compute dipoles, he was not able to reproduce ORCA's RI-OMP2 to within even 0.01 a.u. I determined by comparison to my own Psi4Numpy implementation of DF-OMP2 that this was an error in the `dfocc` module and eventually tracked it down to `dfocc` sending the wrong orbitals to `OEProp`. The primary purpose of this PR was to fix that and get density-fitted orbital optimized properties hooked up to the driver.

...Unfortunately, adding a test that the remaining dipoles are correct required me to fix some other standing `dfocc` issues that prevent users from setting tight convergence (#1143). That in turn required quite a few test changes. While working through that, I also solved #1509. The 0.7 mHartree difference I was seeing was due to another bug in `dfocc`. It computed the MP2.5/MP3 energy by adding to the _SCF reference_ and not the current determinant. Those are different things in orbital-optimized methods. When you add in the correct reference energy, the disagreement reported in #1509 goes away. In brief, this PR fixes #1143 and fixed #1509.

## Todos
- [x] Fix a bug where `dfocc` properties were incorrect
- [x] Add a test that the `dfocc` properties are now correct
- [x] Hook up `dfocc` properties to the driver
- [x] Fix a bug where DFOMP2.5/DFOMP3 energies could be incorrect depending on convergence settings
- [x] Simplified/Sanitized handling of convergence criteria in `dfocc`
- [x] Modified `dfocc` tests due to above

## Checklist
- [x] `ctest -L dfo` and `ctest -L cdo` pass
- [x] `test_dipole.py` passes

## Status
- [x] Ready for review
- [x] Ready for merge


bozkaya:
Hi All,

Thank you for fixes!  

As far as I remember in the earlier version of DFOCC, I was computed oeprop by my functions; hence, I even did not need to save MOs. However, then my code is replaced by OEPROP class, and at that point, it has been forgotten to call the save_mo function before oeprop. @JonathonMisiewicz @hokru

psi4/driver/procrouting/proc.py
```diff
@@ -1781,9 +1862,8 @@ def run_dfocc_property(name, **kwargs):
     dfocc_wfn = core.dfocc(ref_wfn)
 
     # Shove variables into global space
-    if name in ['mp2', 'omp2']:
-        for k, v in dfocc_wfn.variables().items():
-            core.set_variable(k, v)
+    for k, v in dfocc_wfn.variables().items():
```
dgasmith: I forget why this was nested, can you comment on it?

tests/pytests/test_dipoles.py
```diff
@@ -0,0 +1,41 @@
...
+    pytest.param({'name': 'omp2', 'options': {'mp2_type': 'df', 'max_mograd_convergence': 6}, 'varname': 'DF-OMP2'}, id='df-omp2 ae'),
+    pytest.param({'name': 'omp2', 'options': {'mp2_type': 'df', 'freeze_core': 'true', 'max_mograd_convergence': 6}, 'varname': 'DF-OMP2'}, id='df-omp2 fc'),
+    pytest.param({'name': 'omp3', 'options': {'mp_type': 'df', 'max_mograd_convergence': 6}, 'varname': 'DF-OMP3'}, id='df-omp3 ae'),
+    pytest.param({'name': 'omp3', 'options': {'mp_type': 'df', 'freeze_core': 'true', 'max_mograd_convergence': 6}, 'varname': 'DF-OMP3'}, id='df-omp3 fc'),
+    pytest.param({'name': 'omp2.5', 'options': {'mp_type': 'df', 'max_mograd_convergence': 6}, 'varname': 'DF-OMP2.5'}, id='df-omp2.5 ae'),
+    pytest.param({'name': 'omp2.5', 'options': {'mp_type': 'df', 'freeze_core': 'true', 'max_mograd_convergence': 6}, 'varname': 'DF-OMP2.5'}, id='df-omp2.5 fc'),
+    pytest.param({'name': 'olccd', 'options': {'cc_type': 'df', 'max_mograd_convergence': 6}, 'varname': 'DF-OLCCD'}, id='df-olccd ae'),
+    pytest.param({'name': 'olccd', 'options': {'cc_type': 'df', 'freeze_core': 'true', 'max_mograd_convergence': 6}, 'varname': 'DF-OLCCD'}, id='df-olccd fc')
```
dgasmith: Do we need to mark this as slow?

tests/pytests/test_dipoles.py
```diff
@@ -0,0 +1,41 @@
+import pytest
+
+import qcelemental as qcel
+import psi4
+
+from .utils import compare_values
+
+perturbation_strength = 0.001
+
+# TODO: That "true" needs to be a string is silly. Convert it to a boolean when you can do that ithout incurring a NaN energy.
```
dgasmith: Typo.

psi4/src/psi4/dfocc/dfocc.cc
```diff
@@ -109,40 +109,8 @@ void DFOCC::common_init() {
...
-            double temp2;
-            temp2 = (-0.8 * std::log10(tol_grad)) - 0.5;
-            if (temp2 < 3.0) {
-                temp2 = 3.0;
-            }
-            mograd_max = pow(10.0, -temp2);
-            // mograd_max = 10.0*tol_grad;
-            outfile->Printf("\tMAX orbital gradient is changed to : %12.2e\n", mograd_max);
-        }
-    }  // end if (orb_opt_ == "TRUE")
```
dgasmith: @bozkaya Can you review this in light of the issue?

psi4/driver/procrouting/proc.py
JonathonMisiewicz: @loriab would know for sure, since she wrote this code [a year ago](https://github.com/psi4/psi4/commit/880f87f5b2765ed00efaf6dd63c406b574c40157). My read of the situation was that she was only checking that MP2 worked at the time and wasn't concerned about the rest.

tests/pytests/test_dipoles.py
JonathonMisiewicz: It timed in somewhere between 60-100 seconds when I ran it on my laptop. I can rerun it for the exact number. I want dipoles on _every_ property-producing method tested, so if it doesn't count as slow yet, it probably will before long. What's the cutoff for "slow", and how do you mark a pytest as slow?

JonathonMisiewicz: Updated estimate: 80 seconds at present, and hopefully we'll make the test cover more ground in future PRs.

psi4/driver/procrouting/proc.py
```diff
@@ -281,6 +281,81 @@ def select_omp2_property(name, **kwargs):
...
+    # Considering only [df]occ
+
+    func = None
+    if reference in ['RHF', 'UHF', 'ROHF', 'RKS', 'UKS']:
+        if mtd_type == 'DF':
+            if module in ['', 'OCC']:
+                func = run_dfocc_property
+
+    if func is None:
+        raise ManagedMethodError(['select_omp3_property', name, 'CC_TYPE', mtd_type, reference, module])
```
loriab: 'omp3' --> 'olccd'

psi4/driver/procrouting/proc.py
loriab: Yeah, as discussed at ccqc, this should go back to mp2/omp2 only since other mtds haven't been transitioned from globals to wfn qcvars.

psi4/src/psi4/dfocc/dfocc.cc
loriab: Was this resetting of the mograd_conv keywords causing issues, @JonathonMisiewicz? It's admittedly strange, but I don't think it's harmful. The read_options comment should probably note that the defaults are tied to e_convergence.

psi4/src/psi4/dfocc/manager.cc
```diff
@@ -329,6 +329,9 @@ void DFOCC::omp2_manager() {
         // if (natorb == "TRUE") nbo();
         // if (occ_orb_energy == "TRUE") semi_canonic();
 
+        // Save MOs to wfn
+        save_mo_to_wfn();
```
loriab: 🎉

tests/pytests/test_dipoles.py
```diff
@@ -0,0 +1,41 @@
+import pytest
```
loriab: need to add this file to https://github.com/psi4/psi4/blob/master/psi4/CMakeLists.txt#L257 to get it installed.

JonathonMisiewicz: Reminder that we need the procedure for adding pytests documented. I'm surprised that's the only thing that needed changing.

psi4/src/psi4/dfocc/dfocc.cc
JonathonMisiewicz: It was causing an issue, but this was probably an issue of expected behavior. I assumed that since [we bothered to set default convergence criteria specifically for the dfocc module](https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L2758-L2762) that these were supposed to be honored, and that when dfocc said it was _changing_ the convergence criteria, it meant that it was overriding the defaults which were actually supposed to be the defaults.

It sounds like the default behavior is instead supposed to be handled internally by `dfocc` and `occ`, not by the defaults in `read_options.cc`. Is that correct? If so, I can restore the code and make this much clearer. The options need to be better documented, the code in occ and dfocc needs to be documented better, and we should just set the `read_options` defaults that are actually placeholders to values that are obviously placeholders, and not things that look like actual defaults.

By the way, the `dfomp-1` and `dfomp2-2` tests don't actually converge energy to e-6 with the defaults `dfocc` sets.

loriab: dfocc is doing the right thing by the user by obeying the [user's specific directions](https://github.com/psi4/psi4/pull/1772/files#diff-3f4d38c9abf7b05dc3f5c0296fb1dac5L133-L134) wrt mograd. liboptions is limited (not to its discredit) in that it can only set up independent defaults for options, whereas sometimes [modules need tied options](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/detci/params.cc#L145-L184). This particular tying is peculiar, but so long as it's respecting the user's override, it's ok imo.

What you mention about tests not converging to 1.e-6 is bad. You fixed it through the deletion https://github.com/psi4/psi4/pull/1772/files#diff-550445b03966d0041b9fa2c2fb58e4c8L278-L281, right?

JonathonMisiewicz: Okay, I'll restore the code in question but make it clearer that is what it's actually doing.

No, [deleting those lines](https://github.com/psi4/psi4/pull/1772/files#diff-550445b03966d0041b9fa2c2fb58e4c8L278-L281) didn't fix it. The problem is that the orbital steps converge so slowly in `dfocc` that even though subsequent steps differ by less than 1.e-6, we're still 2.e-6 or so away from the converged result. I just amped up the convergence criteria so the tests passed.

Any further comments on the test convergence before I commit review corrections?

loriab: Ah, I think I see what you mean. I tried out the deletion and it "worked", but I was only looking at `DE` at face value. In SCF, `D_CONVERGENCE` is harder to achieve than `DE`, so the extra iterations to get the former, effectively make sure the latter is not just nominally satisfied but is w/i that delta of the converged energy. Whereas the tied gradient and energy convcrit that dfocc sets up is so loose on the former that the latter is the harder to converge even to nominal satisfaction and doesn't get w/i the DE of the actual answer. Does this sound like your assessment? Can we just add 1 to the grad power that the linear regression spits out? I think actually converging is more important than deceptive minimization of niter here.

JonathonMisiewicz: Yeah, for the default convergence settings, it's harder to get the nominal energy convergence than the `max_mograd_convergence`, so we may or may not get within the `DE` of the actual energy.

Just increasing the gradient power should work fine for now. I have future plans to improve `dfocc` orbital convergence (my Psi4Numpy script doesn't have the orbital convergence problems `dfocc` has).

loriab: Sounds good. My vote's for increasing the gradient power.

JonathonMisiewicz: _siiiiiigh_ And it turns out that `occ` and `dfocc` have completely different maps to go from energy convergence to orbital convergence. Any objections to me making `occ` use `dfocc` settings as well? DF should not change convergence thresholds.

loriab: No objections -- hack away.

JonathonMisiewicz:
All issues should be resolved now.

psi4/src/psi4/dfocc/dfocc.cc
```diff
@@ -126,7 +126,7 @@ void DFOCC::common_init() {
             }
             tol_grad = pow(10.0, -temp);
             // tol_grad = 100.0*tol_Eod;
-            outfile->Printf("\tRMS orbital gradient is changed to : %12.2e\n", tol_grad);
+            outfile->Printf("\tFor this energy convergence, default RMS orbital gradient is: %12.2e\n", mograd_max);
```
loriab: isn't the var `tol_grad` still?

JonathonMisiewicz: Clearly, my attempts to be careful when copy-pasting are not careful enough... Fixing.

tests/dfomp2-4/input.dat
```diff
@@ -18,10 +18,11 @@ set {
   guess sad
   reference uhf
   mp2_type df
+  rms_mograd_convergence 5
```
loriab: can do without this addition now?

JonathonMisiewicz: Oh, _this_ test. Message me on Slack when you want to talk about this one. GitHub isn't good for the back-and-forth that discussing this one will require.

JonathonMisiewicz: For future reference: this test is pathological, so we can't be sure if we have the energy to six decimal places or not. Lori and I agreed to lower the convergence check to 5 decimals.

JonathonMisiewicz:
> @bozkaya should have an opportunity to look over the changes to reference energies in the tests.

For his reference, most of the changes are going to be just in the sixth decimal place due to tighter convergence. The one exception are DF-OMP3 tests, where a bug caused fourth decimal place changes as described in the first post.

bozkaya:
@JonathonMisiewicz I think new ref energies are okay.

tests/pytests/test_dipoles.py
```diff
@@ -0,0 +1,42 @@
...
+
+import qcelemental as qcel
+import psi4
+
+from .utils import compare_values
+
+perturbation_strength = 0.001
+
+@pytest.mark.slow
+# TODO: That "true" needs to be a string is silly. Convert it to a boolean when you can do that without incurring a NaN energy.
```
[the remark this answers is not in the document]
PeterKraus: Now that we allow integer values to `FREEZE_CORE` (1, 0, -1, -2, ...) I think the possible future change should be `'true' -> 1` as opposed to `'true' -> True`. I'll try to remember this for the future.

tests/pytests/test_dipoles.py
```diff
@@ -0,0 +1,42 @@
...
+        H 1 1.0 2 101.5
+    """)
+
+    psi4.set_options({'perturb_h': True, 'perturb_with': 'dipole', 'basis': 'cc-pvdz'})
+    psi4.set_options(inp['options'])
+    energies = []
+    for l in [1, -1, 2, -2]:
+        psi4.set_options({'perturb_dipole': [0, 0, l * perturbation_strength]})
+        energies.append(psi4.energy(inp['name']))
+    findif_dipole = (8 * energies[0] - 8 * energies[1] - energies[2] + energies[3]) / (12 * perturbation_strength) * qcel.constants.dipmom_au2debye
```
PeterKraus: So is this just `(8 * ((l = +1) - (l = -1)) - 1 * ((l = +2) - (l = -2)))`? The indices to `energies[]` got me confused.

psi4/src/read_options.cc
```diff
@@ -2626,10 +2626,13 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Convergence criterion for amplitudes (residuals). -*/
         options.add_double("R_CONVERGENCE", 1e-5);
         /*- Convergence criterion for RMS orbital gradient. Default adjusts
-        depending on |occ__e_convergence|. -*/
-        options.add_double("RMS_MOGRAD_CONVERGENCE", 1e-6);
-        /*- Convergence criterion for maximum orbital gradient -*/
-        options.add_double("MAX_MOGRAD_CONVERGENCE", 1e-3);
+        depending on |occ__e_convergence|. Default represents the default
```
PeterKraus: There is a lot of `default`'s in these strings. I don't really follow what which default means.

psi4/src/psi4/dfocc/occ_iterations.cc
```diff
@@ -275,11 +275,6 @@ void DFOCC::occ_iterations() {
             break;
         }
 
-        if (wfn_type_ != "DF-OLCCD") {
-            if (rms_wog < tol_grad && biggest_mograd < mograd_max) break;
```
PeterKraus: I only had a quick look at the list of fixes, but just out of curiosity, why was this removed?

JonathonMisiewicz: To fix #1143, which was interfering with my tests.

tests/pytests/test_dipoles.py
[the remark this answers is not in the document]
JonathonMisiewicz: Turning it into an integer just makes it harder to read.

tests/pytests/test_dipoles.py
JonathonMisiewicz: Correct. I'll be changing `energies` to a dict and using `(8 * energies[1] - 8 * energies[-1] - energies[2] + energies[-2])` in the next commit.

tests/pytests/test_dipoles.py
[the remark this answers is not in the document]
loriab: Agree that `true` needs to still be accepted. Options revamp can handle this sanely or can patch read_options if it becomes vexing.

PR #1776, opened 2020-01-07
python 3.8

loriab:
## Description
Make psi safe for py38. The basic conda pkg does build, but I haven't done full stack yet. For practical reasons, no use turning on a CI lane for 38 until conda pkg available.

## Todos
- [x] req'd mods for python 3.8
- [x] avoid "DeprecationWarning: In future, it will be an error for 'np.bool_' scalars to be interpreted as an index" in `Molecule.from_dict`
- [x] more killing off travis

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


dgasmith:
A single service is nice. But Azure is both much faster and provides more concurrent resources. We originally didn't dive too much into Azure due to its relative complexity, but switching over only took an afternoon of trying.

psi4/src/export_mints.cc
```diff
@@ -237,7 +237,7 @@ std::shared_ptr<Molecule> from_dict(py::dict molrec) {
     std::vector<double> elez = molrec["elez"].cast<std::vector<double>>();
     std::vector<std::string> elem = molrec["elem"].cast<std::vector<std::string>>();
     std::vector<double> mass = molrec["mass"].cast<std::vector<double>>();
-    std::vector<int> real = molrec["real"].cast<std::vector<int>>();
+    std::vector<bool> real = molrec["real"].cast<std::vector<bool>>();
```
andysim: Just a note for the future here - `std::vector<bool>` is usually implemented by storing each element as a single bit, so pointer arithmetic and thread safety do not work as expected.  I don't see any usage in this file that would be a problem, but some really evil bugs can possibly result from this kind of change (I found out the hard way, hence this note!)

andysim:
@PeterKraus - the time limits are more generous on Azure than Travis also, which has been an issue for us in the past

psi4/src/export_mints.cc
loriab: That's good to know. The change here is only to get rid of the `np.bool_` as index warning. I don't see the use of the `real` as becoming more extensive, so your safety analysis will probably continue sound. The only alternative I see is adding a python wrapper fn to do the casting before it heads c-side.

PR #1781, opened 2020-01-15
Fix issues in ROHF and CUHF

susilehtola:
## Description
The orthogonalization changes in the Cholesky pull #1760 broke `scf-guess-read2`. The problem is that I started to renormalize the overlap matrix before doing the eigendecomposition. However, it turns out that there's no (simple?) way to renormalize **and** still have a symmetric orthogonalizing matrix **X**.  

Namely, the natural way to get the right normalization out is that you

1. normalize `S`: `Snorm = diag(1.0/sqrt(S)) *  S * diag(1.0/sqrt(S))`
2. form `Xnorm` for `Snorm`
3. form `X = diag(1.0/sqrt(S)) * Xnorm`

Trying to do step 3 in a way that maintains symmetricity of `X` would require much deeper linear algebra, as you need to know how the eigenvalues and eigenvectors change when you balance S to have a unit diagonal. (It may be that some old linear algebra textbooks show how this is done.)

Anyway, IMHO there's nothing wrong with having an asymmetric `X`: if you have a linearly dependent basis, this is bound to happen, and so there's no reason to assume `X` is symmetric. All that matters in the end is the number of molecular orbitals you get from `X`, since the SCF (or even the core Hamiltonian) will give you a basis that is different from `X`.

The asymmetric `X` revealed a problem in the initialization of the ROHF code, which is now fixed. ROHF should now also work for the case of canonical orthogonalization, I'm not sure why there ever was a stopping clause.

A problem was also found in CUHF, which wasn't tested before. For some reason doing the natural orbitals in the alpha orbital basis made the calculation unstable; simply changing to using `X` to find the natural orbitals suddenly fixed everything.

## Todos
- [x] ROHF orbital guess read-in works
- [x] CUHF orbital guess read-in works
- [x] Orbital orthonormality is tested in the calculation; if the off-diagonal elements of the MO overlap sum to >= 1e-10 then the calculation crashes.

## Questions
- [x] Is this necessary? (It's not, and is removed by the present PR) https://github.com/psi4/psi4/blob/c871e6cc2bff3e4252a19a82c5b49b6e6c6e05de/psi4/src/psi4/libscf_solver/rohf.cc#L113 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/orthog.cc
+++ b/psi4/src/psi4/libmints/orthog.cc
@@ -60,7 +60,7 @@ BasisSetOrthogonalization::BasisSetOrthogonalization(OrthogonalizationMethod met
 void BasisSetOrthogonalization::normalize() {
     // Compute normalization coefficients
     if (normalization_ != nullptr)
-        throw PSIEXCEPTION("BasisSetOrthogonalization::normalizate: normalization_ should be nullptr");
+        throw PSIEXCEPTION("BasisSetOrthogonalization::normalize: normalization_ should be nullptr");
     normalization_ = std::make_shared<Vector>(overlap_->rowspi());
     normalization_->set_name("SO normalization factors");
     for (int h = 0; h < overlap_->nirrep(); h++)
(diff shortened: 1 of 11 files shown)


dgasmith:
I think this all makes sense and the fact that this didn't break tests is pretty impressive (or says something about our test?). I don't thin these is a reason to assume that X is symmetric either. We seem to assume this in some places of the code however, but this should be restricted to the SCF code.

It would be good for @andysim/ @jturney to comment on this as well.

psi4/src/psi4/libmints/orthog.cc
```diff
@@ -137,34 +140,56 @@ void BasisSetOrthogonalization::compute_inverse() {
...
+      require tracking the dependence of the eigenvectors and
+      eigenvalues on the scaling transform.
+
+      If one omits the normalization with the commented code below, X
+      becomes symmetric. However, the symmetricity of X or the lack of
+      it makes no difference at the end, since the orbitals anyway arise
+      from diagonalizing the Fock matrix or orbital optimization, and
+      any code in Psi4 should also work with canonical orthogonalization
+      which is asymmetric by force.
+    */
```
andysim: Thanks for this note - it will be really helpful in a few years when we inevitably revisit this code and wonder what's going on

psi4/src/psi4/libscf_solver/cuhf.cc
```diff
@@ -252,7 +251,7 @@ void CUHF::form_F() {
         outfile->Printf("CUHF Natural Orbital Occupations:\n");
         No_->print();
     }
-    Cno_->gemm(false, false, 1.0, Ca_, Cno_temp_, 0.0);
+    Cno_->gemm(false, false, 1.0, X_, Cno_temp_, 0.0);
```
andysim: Excellent catch - this one is a subtle problem

andysim:
The ROHF guess==read error is indeed outdated

PR #1783, opened 2020-01-17
Remove Custom DIIS: OCC's Open-Heart Surgery I

JonathonMisiewicz:
## Big Picture of the Project
For those who haven't been following the saga, orbital convergence in occ has been unreliable for quite some time. In #1772, I changed the defaults for orbital convergence to be the same as for dfocc. Tests started failing because the orbitals couldn't be converged. My investigation has indicated the problem relates back to the details of the DIIS implementation: the orbital DIIS needs to be coupled to the DIIS for T2 amplitudes. My own Psi4Numpy implementations of orbital-optimized methods show no convergence problems when I do this, but show quite poor convergence when I uncouple them.

...Unfortunately, getting the DIIS coupled together requires open-heart surgery on OCC due to a number of long-standing issues with the module, which I would have needed to fix anyways for a current research project. Per Lori's request, I'm breaking up what would be a monolithic PR into several.

## Description of the PR
The objective of the current PR is to switch orbital convergence from using a custom DIIS implementation of Bozkaya's to using Andy's libdiis. Andy's is general enough to accomodate combined DIIS on a dpdbuf4 and Vector/Matrix quantities. Bozkaya's is not.

## Infrastructure Changes
* Less code duplication!
* I'm introducing a strategy to eliminate the custom tensor classes in OCC over time. When a function uses a custom tensor class, use a constructor to convert it to a libmints standard class. If necessary, write it back at the end of the function call. This allows us to take the task of rewriting all of OCC (and its coupled variables) into going function-by-function. When we can, we'll then just change the variable in the header and eliminate the relevant converters.
* To eliminate code duplication, I added some new objects to the wavefunction that "contain" alpha/beta specific quantities. So instead of needing to write the orbital rotation code once for closed-shell, once for open-shell alpha, and once for open-shell beta, you only write it once. I'm not sure how well it generalizes to other functions, but it sure cleaned up update_mo!

## Capability Loss/Gain
As a result of this switch, we lose the ability to "level shift" in OCC. "Level shift" is a misnomer and actually refers to scaling the diagonal elements of the DIIS B-matrix. When I'm done, the orbital convergence will be so much better that we won't notice it's gone. In the short term, this does cause more test failures.

## Questions
* Because we're losing "level shifting", we're losing some keywords. I don't know how to deprecate these. If they're provided as module-level keywords, then I don't know the technical details of how you would raise a warning message. The bigger problem is when they're global keywords. DFOCC still uses "level shifting", so I can only raise a warning when I know what module the user is in. Thoughts on how to handle this?
* Should we be losing "level shifting"? We can add _damping_ to `libdiis` if there's demand for it. I think there are better things to do with my time.

## Quick Test Status
* Switching out the DIIS version caused the `omp2p5-2` test's OMP2.5 energy to fall outside convergence criteria, again emphasizing how unstable orbital convergence in OCC is. We'll find out what the actual value is by the time I'm done.
* Losing "level shifting" was enough to break the `ocepa-grad2` test. This test asks for a higher-than-normal convergence from OCC, so that's no surprise. (I imagine Psi ups the energy convergence default when the user requests a gradient.) As a temp fix, I lowered the demanded energy convergence from 8 to 7. Even after that, we're outside the test tolerance, so I changed the value to boot. If anybody is dying to know the correct answer, I can run it on my Python code.
* `omp2-2` encountered oscillatory convergence, and I responded by lowering the convergence we seek. Again, coupling amplitude and orbital DIIS will break the oscillations and render this test hack unnecessary.
* We may have lost other full tests, but we'll get them back at the end.

## So Lori Knows What to Put in Changelog
- [x] Overhauled orbital convergence in the OCC module. As a result, the `level_shift` and `do_level_shift` keywords have been removed. Contact the developers if you've found a case where the new convergence is worse than old!

## Checklist
- [x] [Quick tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
* I think CDS would like to see levelshifting as a widely available convergence strategy. But no need to tack it onto this project.
* The disappearance of the level_shift keyword (for reference, used by dfocc, occ, mcscf) doesn't trouble me much. If `throw_deprecation_errors` can't catch set as global, only way I can think of to emit a warning is to add it back to read_options, then probe in the occ module itself and throw an exception there.

JonathonMisiewicz:
For your first point, do you mean damping or level shifting? Remember that "level shifting" in `occ` is a misnomer. It has nothing to do with increasing the energy gap between occupied and virtual orbitals but is a _damping_ scheme that scales the diagonal elements of the DIIS B matrix. I was sloppy about this distinction when I first posted this.

andysim:
We can still scale the B matrix elements if needed, can't we?  I agree with all that you say - it's not likely to be needed with the two sets of iterations coupled, but do we need to remove the keyword?

JonathonMisiewicz:
If we (practically meaning I) added that functionality to `libdiis`, we could. Unless I'm badly misreading the code, it doesn't exist currently.

Between the options of removing the keyword and expanding the scope of the project to involve `libdiis` improvements, I vote for the former.

andysim:
Ah, yes, it would require changes to libDIIS.  It's not clear to me that it would help, so I agree with removal too - thanks for the clarification

tests/omp2p5-2/input.dat
```diff
@@ -2,7 +2,7 @@
 refnuc      =  25.59060766929188 #TEST
 refscf      = -129.25245637508348 #TEST
 refmp3      = -129.58077576726413 #TEST
-refomp3     = -129.58887635080728 #TEST
+refomp3     = -129.58890163328363 #TEST
```
andysim: Is this change, and the gradient in `grad2` due to different tolerances?

JonathonMisiewicz: This PR doesn't change tolerance. The `occ` convergence is so poor that even though the energy changes may no longer be within `e_convergence`, you may be outside `e_convergence` of the converged result. The `occ` convergence is hyper-sensitive enough that switching the DIIS (probably losing the DIIS damping) caused the energy to stabilize at values outside `e_convergence` tolerance, even though the tolerance is exactly the same! In some cases, DIIS damping helps, and in others, it hurts, so Psi can't tell me which one is the _real_ energy/gradient.
The long-term fix is to make `occ` convergence _not_ poor. If there's doubt as to which value is the real one, I have my own Psi4Numpy OCEPA gradient code for `ocepa-grad2`. I don't have an OMP2.5 code, but I have OMP2, so I can add it if necessary.
But even if the old version is right, all I'd so is loosen the convergence tolerance so this value passes until I get the convergence in proper working order.

andysim: Got it, thanks!

susilehtola:
Would it be possible to decouple amplitude and orbital DIIS, i.e. so that the amplitudes get solved at fixed orbitals, and then the orbitals are updated? This would be useful for cases that are hard to converge, as well as active space calculations in large basis sets.

JonathonMisiewicz:
@susilehtola 
You're talking about a different orbital-optimization algorithm. I can organize the code so that it's relatively easy to reuse the "key pieces", including DIIS, for different algorithms, but new OCC algorithms are outside the scope of this PR. Maybe after I get frozen core energies/gradients working.
I'll also add that the orbital-optimization code isn't centralized, so active space calculations aren't handled by OCC.

PR #1785, opened 2020-01-19
Clean manager.cc Variables: OCC's Open-Heart Surgery II

JonathonMisiewicz:
## Big Picture of the Project
See #1783. tl;dr: The goal is to improve orbital convergence in `occ`, but I need the module cleaned up to make it easier to think about the code I'm needing to mess with. I'm breaking a monolith PR into several.

## Description of the PR
The objective of the current PR is to clean up `occ/manager.cc`. All duplicated printing and variable setting code is now located in `postprocessing.cc`.

## Infrastructure Changes
* Less code duplication and less code!
* The important `occ` variables are now set in `variables_` C-side so they can be set Py-side.
* A new file has been created, `postprocessing.cc`, for all `occ`'s repetitive printng and Psi variable-setting needs

## Questions
* @loriab, please check that there's nothing obvious missing. In particular, there is no `test_omp.py`, so I just made a .txt file for you.

## Checklist
- [x] [Quick tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/pytests/occ_methods.txt
+++ b/tests/pytests/occ_methods.txt
@@ -1,0 +1,9 @@
+Candidate Structure for other method-tests
+
+_ref_module = {scftype: {ref: {frz: {mp2type: {} for mp2type in ['conv', 'df', 'cd']} for frz in ['true', 'false']} for ref in ['rhf', 'uhf']} for scftype in ['pk', 'df']}
+
+* The above will work for mp2, mp2.5, mp3, cepa, and the orbital-optimized variants thereof. MP2 has ROHF, but that's special.
+* The dct module can also supply non-FC, conventional integral cepa.
+* At present, occ has no frozen core energies for orbital optimized methods or frozen core gradients for anything. dfocc does. Jonathon will be changing this in a future
+  project, as well as adding some frozen virtual. ETA is Spring 2021, because that depends on the Gradient Rewrite.
+* dfocc has all DF gradients and CD energies for all options.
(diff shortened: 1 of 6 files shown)


JonathonMisiewicz:
The diffs come up clean for every section of variable printing in OCEPA, CEPA, OMP2, OMP3, OMP2.5, MP3, MP2.5, MP2... Numbers vary, unsurprisingly. There's no printing difference depending on closed vs open-shell except for ROHF-MP2. I can't find an ROHF-MP2 test anywhere, so I compared a UHF-MP2 and an ROHF-MP2 calculation. After fixing one silly mistake (see last commit), the diff was exactly what I expected.

tests/pytests/occ_methods.txt
```diff
@@ -0,0 +1,9 @@
+Candidate Structure for other method-tests
+
+_ref_module = {scftype: {ref: {frz: {mp2type: {} for mp2type in ['conv', 'df', 'cd']} for frz in ['true', 'false']} for ref in ['rhf', 'uhf']} for scftype in ['pk', 'df']}
```
loriab: I didn't convey clearly that what I had in mind was filling in the ref values for mp2.5, mp3, lccd, if you have them handy. If not, ok.

JonathonMisiewicz: I'll probably do this for all 8 `occ` methods at the end. I'm going to want to examine every single `occ` test before I call the project done, whether the method is orbital-optimized or not. Again, `occ` has a way of introducing unexpected bugs.

PR #1789, opened 2020-01-24
Condense MP2, Sanitize Spin Scaling: OCC's Open-Heart Surgery III

JonathonMisiewicz:
## Big Picture of the Project
See #1783. tl;dr: The goal is to improve orbital convergence in `occ`, but I need the module cleaned up to make it easier to think about the code I'm needing to mess with. I'm breaking a monolith PR into several.

## Description of the PR
The current PR has two objectives. First, condense the MP2 amplitude functions, MP2 energy functions, and MP2 1RDM functions into one function each. Although several methods needed these functions with minimal changes, there were two or three functions created for each of those three. Dealing with how he handled spin-scaling forced me to take a second objective: to clean OCC's handling of spin-scaling variables. OCC previously allowed the user to overwrite the definitions of opposite-spin scaling and spin-component-scaling schemes and used far too many variables. The current scheme still allows users to make their own definitions, but those are now sent to custom variables.  In addition, many keywords have been eliminated. OCC's handling of spin-scaling now differs from DFOCC's, but I am not fool enough to take on that module.

Obligatory ping of @susilehtola and @hokru, who expressed opinions about this on Slack, and @loriab, who expressed many opinions about this on Slack.

## Todos
- [x] We have shed another 800+ lines of code.
- [x] Major refactor of spin-scaling in OCC. All keywords for manual handling of spin scaling in OCC are removed. Please use `os_scale` and `ss_scale` for custom scaling. Select the spin scaling method that goes to the current energy by the method string you pass to the energy call.
~~- [ ] Update documentation accordingly.~~ Deferred to Lori PR.

## Checklist
- [x] Quick tests, occ related tests, manual diff checks, and `test_mp2.py` all pass

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/occ/postprocessing.cc
```diff
@@ -151,6 +150,9 @@ void OCCWave::mp3_postprocessing() {
     Process::environment.globals["SCSN-MP3 CORRELATION ENERGY"] = Escsnmp3 - Escf;
     Process::environment.globals["SCS-MP3-VDW CORRELATION ENERGY"] = Escsmp3vdw - Escf;
     Process::environment.globals["SOS-PI-MP3 CORRELATION ENERGY"] = Esospimp3 - Escf;
```
hokru: Are SOS-PI-MP3 and the other additional flavors defined anywhere in the literature? Would'nt the E3_SCALE be different each time?

JonathonMisiewicz: As best as I can tell, they are not defined. Section 2.1 of [Bozkaya's second OMP3 benchmark paper](https://pubs.acs.org/doi/10.1021/ct301078q) defines SCS-OMP3 and SOS-OMP3, and the [first OMP3 benchmark paper](https://aip.scitation.org/doi/10.1063/1.3665134) is where SOS-OMP3 was first defined. See the last paragraph of section III for the opposite spin scaling factor he settled on. Grimme defined [defined SCS-MP3](https://onlinelibrary.wiley.com/doi/full/10.1002/jcc.10320)

As far as I can tell, there's been no examination of what the optimal E3_SCALE is for methods other than SCS-MP3. Of the MP3 series, the only spin scaled variants I can find defined in the literature are SCS-OMP3, SOS-OMP3, and SCS-MP3.

hokru: That agrees with my knowledge about spin-scaled MP3 methods. So essentially the remaining variants are unpublished. I don't know how others feel about this, but I would suggest to comment them out. To avoid users mistakenly using the resulting numbers.

JonathonMisiewicz: ...This is going to create _another_ disagreement with DFOCC, but I think this is the right call, scientifically. I want a core dev's approval (probably @loriab) before I strip out all the unpublished spin-scaling schemes from `occ`... Which there are probably a lot of.

loriab: Removing unpublished is good by me and in the spirit of the P::e->wfn pass.

JonathonMisiewicz: I'll get on it then, unless some other core dev objects. Yak shaving intensifies.

JonathonMisiewicz:
Forewarning that the [documentation in occ/dfocc](http://www.psicode.org/psi4manual/master/occ.html#) needed reworking even _before_ this commit, and it certainly will afterwards. I'll hold off on rewriting that (and updating the `driver.energy` docstring) until the rest of the PR looks good. How I decide to rewrite the manual will depend on if I need to make other changes, and the docstring updates will depend on the manual changes...

Aren't dependencies fun?

psi4/src/psi4/occ/postprocessing.cc
```diff
@@ -138,19 +133,14 @@ void OCCWave::mp3_postprocessing() {
...
     variables_["MP2.5 TOTAL ENERGY"] = 0.5 * (Emp3 + Emp2);
     variables_["MP3 CORRELATION ENERGY"] = Emp3 - Escf;
     Process::environment.globals["SCS-MP3 CORRELATION ENERGY"] = Escsmp3 - Escf;
-    Process::environment.globals["SOS-MP3 CORRELATION ENERGY"] = Esosmp3 - Escf;
-    Process::environment.globals["SCSN-MP3 CORRELATION ENERGY"] = Escsnmp3 - Escf;
-    Process::environment.globals["SCS-MP3-VDW CORRELATION ENERGY"] = Escsmp3vdw - Escf;
-    Process::environment.globals["SOS-PI-MP3 CORRELATION ENERGY"] = Esospimp3 - Escf;
+
+    variables_["CUSTOM SCS-MP3 CORRELATION ENERGY"] = os_scale * Emp3AB + ss_scale * (Emp3AA + Emp3BB);
+    variables_["CUSTOM SCS-MP3 TOTAL ENERGY"] = Escf + variables_["CUSTOM SCS-MP3 CORRELATION ENERGY"];
```
loriab: I see why you always need to define the `CUSTOM ` vars c-side for your neat current energy lookups. And I guess it's worth keeping the current energy setting c-side in case we can ever apply scaling to the gradient. So could you add something like the below (untested) to proc.py so that CUSTOM variables aren't cluttering up unnecessarily?

```
if not((core.option_has_changed("OCC", "SS_SCALE") or core.option_has_changed("OCC", "SS_SCALE")):
    for qcv in wfn.variables():
        if qcv.startswith("CUSTOM SCS"):    
            wfn.del_variable(qcv)
```

psi4/src/psi4/occ/postprocessing.cc
```diff
@@ -113,22 +109,21 @@ void OCCWave::mp2_postprocessing(bool include_singles) {
     // Wavefunctions saved to variables are set on the wavefunction Py-side.
     // Other variables will be set Py-side via QCDB formulas, per a future Lori project.
     variables_["MP2 CORRELATION ENERGY"] = Emp2 - Escf;
-    variables_["SCS-MP2 CORRELATION ENERGY"] = 6.0/5.0 * Emp2AB + 1.0/3.0 * (Emp2AA + Emp2BB) + Emp2_t1;
+    variables_["SCS-MP2 CORRELATION ENERGY"] = 6.0 / 5.0 * Emp2AB + 1.0 / 3.0 * (Emp2AA + Emp2BB) + Emp2_t1;
```
loriab: Isn't the 6/5, 1/3 already covered by the likes of https://github.com/psi4/psi4/pull/1789/files#diff-fd2a6d2dc2bc31f90b49a7bdcff983d6R181 ?

psi4/src/psi4/occ/cc_energy.cc
```diff
@@ -110,90 +110,80 @@ void OCCWave::omp2_mp2_energy() {
...
                                "MO Ints <OO|VV>");
-        Emp2AA = 0.5 * global_dpd_->buf4_dot(&Tss, &K);
-        global_dpd_->buf4_close(&Tss);
+        Emp2AA = 0.5 * global_dpd_->buf4_dot(&T, &K);
+        global_dpd_->buf4_close(&T);
 
-        Escsmp2AA = ss_scale * Emp2AA;
+        Escsmp2AA = 1.0 / 3.0 * Emp2AA;
         Escsnmp2AA = 1.76 * Emp2AA;
         Escsmimp2AA = 1.29 * Emp2AA;
```
loriab: SCS(MI)-MP2 not in `energy()` list or printed out? I don't blame you -- it's basis zeta dependent (and even tq is different from t or q) -- I just want to check not an oversight.

psi4/src/psi4/occ/manager.cc
```diff
@@ -165,9 +162,6 @@ void OCCWave::omp2_manager() {
         Process::environment.globals["OMP2 TOTAL ENERGY"] = Emp2L;
         Process::environment.globals["SCS-OMP2 TOTAL ENERGY"] = Escsmp2;
         Process::environment.globals["SOS-OMP2 TOTAL ENERGY"] = Esosmp2;
```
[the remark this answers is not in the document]
JonathonMisiewicz: I thought you wanted SCS/SOS not copied to `variables_`, because that should be qcbd's responsibility? I'm fine bringing SCS/SOS to variables_ if you are.

psi4/src/psi4/occ/cc_energy.cc
JonathonMisiewicz: It was probably an oversight in the original `occ` code. `Escsmimp2` variables are set, but aren't used. Nothing in `occ` accounts for basis-set dependence of these quantities beyond the basis-set dependence of the spin components of the correlation energy.

If that's not correct, I'll just rip them out.

psi4/src/psi4/occ/manager.cc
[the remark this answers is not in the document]
loriab: At the point that it’s qcdb’s responsibility, the only things that psi needs to set are same-spin, opposite-spin, and singles energies. Going that far would necessitate rewriting the occ printing, which seemed a step too far for this round. So yes, please go for variables_ only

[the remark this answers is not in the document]
loriab: At the point that it’s qcdb’s responsibility, the only things that psi needs to set are same-spin, opposite-spin, and singles energies. Going that far would necessitate rewriting the occ printing, which seemed a step too far for this round. So yes, please go for variables_ only

JonathonMisiewicz:
Are the "notes for pre-1.4 follow-up" on my to-do list or yours?

psi4/src/psi4/occ/postprocessing.cc
JonathonMisiewicz: They are. To make things even more "fun", the two Escsmp2 formulas aren't even the same. They differ on whether you include the T1 contribution in the SCS-MP2 formula.

I can either replace the wavefunction variables (advantage: less cluttered wavefunction) or replace this call with said variables (advantage: we don't need to construct the spin-component energy multiple times). I'll think over which approach seems better.

loriab:
My todo list. It’s the least I can do if you can fix convergence.

psi4/src/psi4/occ/manager.cc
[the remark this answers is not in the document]
JonathonMisiewicz: So is there anything for me to do on this point? `OMP2 TOTAL ENERGY` is set in `variables_` on the line above, and you don't want me to set `SCS-OMP2 TOTAL ENERGY` or `SOS-OMP2 TOTAL ENERGY`.

psi4/driver/procrouting/proc.py
```diff
@@ -1959,107 +1959,124 @@ def run_occ(name, **kwargs):
...
+        core.set_local_option('OCC', 'SPIN_SCALE_TYPE', 'CUSTOM')
+
     elif name == 'omp2.5':
         core.set_local_option('OCC', 'WFN_TYPE', 'OMP2.5')
         core.set_local_option('OCC', 'ORB_OPT', 'TRUE')
-        core.set_local_option('OCC', 'DO_SCS', 'FALSE')
-        core.set_local_option('OCC', 'DO_SOS', 'FALSE')
+        core.set_local_option('OCC', 'SPIN_SCALE_TYPE', 'NONE')
+    elif name == 'custom-scs-omp2.5':
+        core.set_local_option('OCC', 'WFN_TYPE', 'OCEPA')
```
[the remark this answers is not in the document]
JonathonMisiewicz: The sad part is that this mistake is _after_ I specifically looked for copy-paste errors on `WFN_TYPE`. Will be fixed in the next commit.

psi4/src/psi4/occ/postprocessing.cc
loriab: fwiw, I went with including singles in qcdb https://github.com/qcdb/qcdb/blob/master/qcdb/qcvars/identities.py#L87

psi4/src/psi4/occ/cc_energy.cc
loriab: it won't pain me if scsmi is ripped out.

psi4/src/psi4/occ/manager.cc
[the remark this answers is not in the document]
loriab: I think kill off all the remaining `P::e.globals`. Then the copy in proc.py will populate them from `variables_` and keep it in sync.

psi4/src/psi4/occ/postprocessing.cc
JonathonMisiewicz: After checking the original Grimme paper, I _suspect_ that SCS-MP2 is defined to exclude singles contribution, but it isn't explicit. Because QCDB, I'll include singles energy provisionally. Grimme is stopping by UGA in a couple weeks, so I'll see if I can ask him directly.

loriab: iirc, Sherrill's SCS-CCSD paper _did_ include singles, hence qcdb's bias

JonathonMisiewicz:
One final logistical point before I head off for the night: my current plan is to make these changes and commit those. If you are willing to add `occ`/`dfocc` documentation and docstring changes to your to-do list, I say can call this PR a wrap so I can slay the behemoth that is orbital convergence. If you don't have time for that, I can take it on, but that will delay getting orbital convergence fixed.

JonathonMisiewicz:
To confirm this explicitly:
* With the changes to method names, the energy docstring needs to be updated, as does the occ/dfocc manual page, which needed to be updated anyways. That is shifting to your responsibility in a future, but pre-1.4 PR.
* In another future PR of yours, QCDB will handle all spin-component quantities. At that time, we'll save singles contributions and spin components to `variables_`. For now, it's enough to remove the spin component globals in preparation for that and as part of the war on globals. We're losing many spin-component-related Psivars in the interim, but this is an acceptable loss.

Also, will the future migration of spin-component treatment to QCDB mean `occ` no longer prints spin-component quantities? You implied that it would ("Going that far would necessitate rewriting the occ printing"). If so, then the correct solution to [the spin-redundancy problem](https://github.com/psi4/psi4/pull/1789#discussion_r373811576) is to eliminate the spin-component variables from `occ`. At that time, `occ`'s only use for spin-component scaled quantities is if spin scaling is being used in orbital optimization, which shouldn't require those variables anyways. We might as well get rid of them now, since the repeat typing will be gone before long.

loriab:
In my view, the most important variables that get set in OCC are the `[SAME-SPIN|OPPOSITE-SPIN|SINGLES] [|O][MP2|MP3|MP2.5|LCCD] CORRELATION ENERGY`] ones both now and in future because all others can be constructed from these.

yes, I think in future the main printing from OCC could be moved py-side, especially for quantities like SCS that OCC does nothing further with internally. I broadly plan that OCC would set the above variables and any totals that it likes, that OCC would print c-side the main energies that it's dealing with, then in proc.py, the return variables get sent to qcdb to be composed into `SCS(N)-MP2 TOTAL ENERGY` and anything else for which input pieces are present. the key thing about qcdb qcvar composition is that it also checks to make sure the SCS(N) built from components equals any SCS(N) passed in whole so ensures that OCC and qcdb are using the same definitions. I don't want to delete OCC's definitions before this check is active. So that plus leaving Ugur's printing mostly undisrupted is why I figured this could be put off.

> If so, then the correct solution to the spin-redundancy problem is to eliminate the spin-component variables from occ

Are you meaning SCS(N) or SAME-/OPPOSITE-SPIN variables?

JonathonMisiewicz:
I _was_ meaning SCS(N) variables, but after reading your comments, I've changed my mind again. While the number of spin-related variables on OCC's wavefunction object is _bad_, defining the same quantities multiple times throughout OCC is _worse_. That risks inconsistencies in the definitions and is another thing you need to check for when moving to qcdb. So the current plan (which is what I'll go with pending your objections) is to just use the variables already saved to OCC rather than re-computing them.

JonathonMisiewicz:
...And unfortunately, one more wrinkle. Upon deleting the remaining `P::e.globals`, we lose  the`SCS-OMP3 TOTAL ENERGY` variable, causing a quick test failure.

When you said "kill off all the remaining P::e.globals", did you mean _turn them into variables_, or did you mean _delete the lines, and remove any reference to them in tests_? There is nothing in`proc.py` to reconstruct the missing variables.

loriab:
Turn into wfn variables_

JonathonMisiewicz:
That should be everything now, thanks. Quick tests still pass.

JonathonMisiewicz:
`mp2_test.py` changed by no longer looking for custom spin variables when custom spin wasn't set, but that was all.

Now that this is in, time for the Big One...

PR #1790, opened 2020-01-29
NBO Fix, Py-Movement

### LINKED ISSUE #1698: NBOWriter Wrong for Cartesian Basis Sets
Consider the following input file:
```
molecule {
O
H 1 1.0 
H 1 1.0 2 104.5
}

set scf_type pk
set basis 6-31+G(d)
set r_convergence 12

energy, wfn = energy('hf', return_wfn = True)
NBOWriter(wfn).write('input.47')
```

This is the NBO test case, except I have replaced the basis set with a Pople basis. When I run the produced input file through NBO, the result I get does not match QChem. The fault is with Psi. Just look at the unphysical energy gap in the d NAOs:
```
  15    O  1  dxy    Ryd( 3d)     0.00000       0.22545
  16    O  1  dxz    Ryd( 3d)     0.00196       0.22862
  17    O  1  dyz    Ryd( 3d)     0.00518       0.52493
  18    O  1  dx2y2  Ryd( 3d)     0.00125       2.36883
  19    O  1  dz2    Ryd( 3d)     0.00227       2.02191
```

For comparison, QChem gives
```
  15    O  1  dxy    Ryd( 3d)     0.00000       2.02907
  16    O  1  dxz    Ryd( 3d)     0.00518       2.71710
  17    O  1  dyz    Ryd( 3d)     0.00196       2.05762
  18    O  1  dx2y2  Ryd( 3d)     0.00125       2.36883
  19    O  1  dz2    Ryd( 3d)     0.00227       2.02191
```

I suspect the error has to do with Pople basis sets being defined in cartesians and not spherical. This is the only reason I can think of for agreement with QChem being basis-set dependent, and it also agrees nicely with the fact that "Input basis functions are not normalized; fixed" appears in the Psi-NBO output file.

While I'll likely end up fixing this, this is very low on my priority list. If anybody else wants to take this, feel free. Just post here.

JonathonMisiewicz:
## Description
Fixes #1698. Psi wasn't sending AOs normalized to unity, and NBO was correcting it by transforming the Fock matrix and the density matrix both _inversely_ to the overlap matrix. That's not how you want to do it! The Fock matrix transforms the same way, but the density matrix (and the AO index of C) transform inversely. Because NBO can't be trusted to do the transformation correctly, Psi does it itself. To make coding that easier, the NBO writing was moved to Python. The C-code still exists until 1.5, where it shall be ripped out of `libmints`. Deprecation warnings have been added.

## Todos
- [x] Fixes a bug causing incorrect Psi to produce incorrect NBO input files involving Cartesian basis sets with angular momentum of d or more (#1698).
- [x] Moved NBO Writing python-side.
- [x] Please use `wfn.write_nbo(name)` to write NBO files from now on.

## Checklist
- [x] Confirmed that the old and new NBO interfaces produce the same input.47 files for the current test case and a test case with linear dependencies.
- [x] Confirmed that the NAO populations and energies match with the 6-31+G(d) basis between QChem and Psi4. This confirms the accuracy of the overlap, AO density, and AO Fock matrices. The disagreements are either in the last decimal place or due to setting up coordinate systems.
- [x] Confirmed that the MOs in terms of AOs, as reported by NBO, match with the 6-31+G(d) basis between QChem and Psi4. This additionally checks the accuracy of the C matrix. Again, the order of the d matrices shifts due to different orientation systems in my test cases.

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/p4util/__init__.py
```diff
@@ -37,3 +37,4 @@
 from .p4regex import *
 from .python_helpers import *
 from .solvers import *
+from .writer import *
```
loriab: try `from . import writer` since you don't actually need any functions, just need the code executed. I think that'll appease lgtm.

tests/nbo/input.dat
```diff
@@ -11,5 +11,5 @@ set basis cc-pvdz
 set r_convergence 12
 
 energy, wfn = energy('hf', return_wfn = True)
-NBOWriter(wfn).write('input.47')
+wfn.write_nbo('input.47')
```
loriab: Do the nbo 47 or analysis files in the test need updating?

JonathonMisiewicz: Those used a spherical basis set, so the existing ones are fine. I'll update them to include a Cartesian example later this evening.

PR #1799, opened 2020-02-04
Export Electric Dipole Derivative Integrals to Python

kcpearce:
## Description
Exports electric dipole derivative integrals to Python side for use in calculating VCD

## Todos
- [x] Adds `.def("ao_elec_dip_deriv1")` and `.def("mo_elec_dip_deriv1")` binding to `psi4/src/export_mints.cc`
- [x] Adds `ao_elec_dip_deriv1()` and `mo_elec_dip_deriv1()` definition/declaration to `psi4/src/psi4/libmints/mintshelper.cc(h)`
- [x] Adds `ao_elec_dip_deriv1_helper()` helper function to `psi4/src/psi4/libmints/mintshelper.cc(h)`

## Checklist
- [x] Add `/tests/pytests/test_mints.py`

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
FYI, the build is going to fail because of the semicolons in your `export_mints.cc` changes. The semicolon must be after the last and only last `def` for a given class.

kcpearce:
@dgasmith I finally got the export working and tested it on the python side by contracting the exported integrals with the ao density and comparing against the already implemented and exported `dipole_grad()` function. How/where do you want me to implement a test for this?

dgasmith:
@loriab might be a better place to recommend. But my first thought it `tests/pytests/test_mints.py`. This would be a new file, but a good place for libmints unit tests.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2597,6 +2597,230 @@ std::vector<SharedMatrix> MintsHelper::ao_overlap_kinetic_deriv2_helper(const st
...
+//                for (int atom_cart = 0; atom_cart < 3; atom_cart++) {
+//                    int offset = (6 * mu_cart * nP * nQ) + (atom_cart * nP * nQ) + (3 * (atom == aQ) * nP * nQ);
+//                    for (int p = 0; p < nP; p++) {
+//                        for (int q = 0; q < nQ; q++) {
+//                            grad[3 * mu_cart + atom_cart]->add(p + oP, q + oQ, buffer[p * nQ + q + offset]);
+//                            if (aP == aQ) grad[3 * mu_cart + atom_cart]->add(p + oP, q + oQ, buffer[p * nQ + q + offset - (3 * nP * nQ)]);
+//                        }
+//                    }
+//                }
+//            }
```
kcpearce: @dgasmith @loriab Quick question: I wrote the code in two equivalent forms - 1. the above commented code that exhibits a more compact loop structure, and 2. the code below this that mirrors a similar structure to other surrounding mintshelper functions. Is one of these preferred over the other? I tend to lean toward fewer lines of code being better, but perhaps it's better to match the style of existing similar functions. Thoughts?

dgasmith: @jturney @andysim Probably have more preferences here.

jturney: I'm good with the compact code.

kcpearce:
> @loriab might be a better place to recommend. But my first thought it `tests/pytests/test_mints.py`. This would be a new file, but a good place for libmints unit tests.
 
Would it be best to still test this function by contracting the exported dipole derivates with a density and then comparing results to the `dipole_grad()` function in mintshelper that contracts the dipole derivatives with an AO density before exporting to python side?

loriab:
Test failures are everywhere, even on master. I think the MKL constraints of some of the psi packages and adcc are interacting badly.

kcpearce:
@loriab I ran a full set of tests and seem to be failing only the following two ctests:

tests/pytests/test_mints.py
```diff
@@ -0,0 +1,45 @@
...
+        deriv1_mat["MU_" + str(atom)] = mints.ao_elec_dip_deriv1(atom)
+        for mu_cart in range(3):
+            for atom_cart in range(3):
+                map_key = "MU" + cart[mu_cart] + "_" + str(atom) + cart[atom_cart]
+                deriv1_np[map_key] = np.asarray(deriv1_mat["MU_" + str(atom)][3 * mu_cart + atom_cart])
+                MU_Gradient[3 * atom + atom_cart, mu_cart] += np.einsum("uv,uv->", deriv1_np[map_key], D_np)
+
+    PSI4_MU_Grad = mints.dipole_grad(D)
+    G_python_MU_mat = psi4.core.Matrix.from_array(MU_Gradient)
+    psi4.compare_matrices(PSI4_MU_Grad, G_python_MU_mat, 10, "DIPOLE_GRADIENT_TEST")  # TEST
```
loriab: needs an assert to trigger pass/fail

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -3137,6 +3195,38 @@ std::vector<SharedMatrix> MintsHelper::mo_oei_deriv2(const std::string &oei_type
...
+    std::vector<SharedMatrix> ao_grad = ao_elec_dip_deriv1_helper(atom);
+
+    return ao_grad;
+}
+
+std::vector<SharedMatrix> MintsHelper::mo_elec_dip_deriv1(int atom, SharedMatrix C1, SharedMatrix C2) {
+    std::vector<std::string> cartcomp;
+    cartcomp.push_back("X");
+    cartcomp.push_back("Y");
+    cartcomp.push_back("Z");
```
loriab: is this a place for one of those initializer lists?

kcpearce: I'm not sure I'm familiar with those. Can you expand?

JonathonMisiewicz: See "Initializing like arrays" example [here](https://www.geeksforgeeks.org/initialize-a-vector-in-cpp-different-ways/). It'll save you some lines of code and improve readability.

kcpearce:
@loriab All tests are passing! Ready for review on my end.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2682,6 +2682,61 @@ std::vector<SharedMatrix> MintsHelper::ao_overlap_kinetic_deriv2_helper(const st
...
+        for (int Q = 0; Q < bs2->nshell(); Q++) {
+            int nP = basisset_->shell(P).nfunction();
+            int oP = basisset_->shell(P).function_index();
+            int aP = basisset_->shell(P).ncenter();
+
+            int nQ = basisset_->shell(Q).nfunction();
+            int oQ = basisset_->shell(Q).function_index();
+            int aQ = basisset_->shell(Q).ncenter();
+
+            if (aP != atom && aQ != atom) continue;
```
jturney: Is this right, the atoms of P and Q have to be the same?

kcpearce: Not exactly. This line checks to make sure that the atom passed is equal to at least one of them before computing any derivatives

jturney: Doh, you're right.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2684,10 +2684,7 @@ std::vector<SharedMatrix> MintsHelper::ao_overlap_kinetic_deriv2_helper(const st
 
 /* 1st derivatives of electric dipole integrals in the AO basis */
 std::vector<SharedMatrix> MintsHelper::ao_elec_dip_deriv1_helper(int atom) {
-    std::vector<std::string> cartcomp;
-    cartcomp.push_back("X");
-    cartcomp.push_back("Y");
-    cartcomp.push_back("Z");
+    std::vector<std::string> cartcomp{ "X", "Y", "Z" };
```
kcpearce: @loriab Is this the change that you were hoping for? Or something else?

loriab: Line 3208 was where I had in mind. Did you get the cmake change? That’s the vital one before merge. I wouldn’t go all out on mintsgelper — just initializer list the xyz occurrences in your code.

loriab: Line 3208 was where I had in mind. Did you get the cmake change? That’s the vital one before merge. I wouldn’t go all out on mintsgelper — just initializer list the xyz occurrences in your code.

kcpearce: Ah, yes. My mistake. I made the change for my `ao_elec_dip_deriv1_helper` first, but missed the `mo_elec_dip_deriv1` function. I fixed that one as well now. And added the line with my test to CMakeLists.txt as well. Thanks.

PR #1800, opened 2020-02-04
dipole and quadrupole arrays and units

loriab:
## Description

## Todos
- [x] deprecate DIPOLE X and QUADRUPOLE XX qcvars that were anomolously in Debye-Ang units
- [x] add DIPOLE and QUADRUPOLE array qcvars in atomic units
- [x] edit all the tests so that they check both for the interim
- [x] the `P::e.globals` and `Wfn.variables_` maps work in `core.Matrix`. Want to keep this storage for continuity and the preservation of symmetry, but 2D is not what one expects for dipole or charges (much less what one expects for octupole). So selected properties vars are returned as ndarray from the `variable()` query fns. This means you can't assume the return is a Matrix.
- [x] schema handling addressed accordingly and tested for run_json and run_qcschema setups. Note that return units now au
- [x] handling >quadrupole

**EDIT** change of plans from first post. in order not to have to go from unique multipole elements (10 for octupole) to complete array (27) c-side and flat since P::e.globals restricted to Matrix, now only the compressed array is stored in globals/variables_ and that gets turned into (and back) the complete multidimensional ndarray when queried py-side. quadrupole has been turned back c-side to `(6)`, rather than `(3,3)` for consistency with the other poles.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -604,6 +604,19 @@ core.set_global_option_python = _core_set_global_option_python
 
 ## QCvar helps
 
+def _qcvar_warnings(key):
+    if any([key.upper().endswith(" DIPOLE " + cart) for cart in ["X", "Y", "Z"]]):
+        warnings.warn(
+            f"Using scalar QCVariable `{key.upper()}` [D] instead of array `{key.upper()[:-2]}` [e a0] is deprecated, and in 1.5 it will stop working\n",
+        category=FutureWarning,
+        stacklevel=3)
+
+    if any([key.upper().endswith(" QUADRUPOLE " + cart) for cart in ["XX", "YY", "ZZ", "XY", "XZ", "YZ"]]):
+        warnings.warn(
+            f"Using scalar QCVariable `{key.upper()}` [D A] instead of array `{key.upper()[:-3]}` [e a0^2] is deprecated, and in 1.5 it will stop working\n",
+        category=FutureWarning,
+        stacklevel=3)
+
 
 def _core_has_variable(key):
     return core.has_scalar_variable(key) or core.has_array_variable(key)
@@ -614,6 +627,8 @@ def _core_wavefunction_has_variable(cls, key):
 
 
 def _core_variable(key):
+    _qcvar_warnings(key)
+
     if core.has_scalar_variable(key):
         return core.scalar_variable(key)
     elif core.has_array_variable(key):
@@ -623,6 +638,8 @@ def _core_variable(key):
 
 
 def _core_wavefunction_variable(cls, key):
+    _qcvar_warnings(key)
+
     if cls.has_scalar_variable(key):
         return cls.scalar_variable(key)
     elif cls.has_array_variable(key):
@@ -632,6 +649,8 @@ def _core_wavefunction_variable(cls, key):
 
 
 def _core_set_variable(key, val):
+    _qcvar_warnings(key)
+
     if isinstance(val, core.Matrix):
         if core.has_scalar_variable(key):
             raise ValidationError("psi4.core.set_variable: Target variable " + key + " already a scalar variable!")
@@ -650,6 +669,8 @@ def _core_set_variable(key, val):
 
 
 def _core_wavefunction_set_variable(cls, key, val):
+    _qcvar_warnings(key)
+
     if isinstance(val, core.Matrix):
         if cls.has_scalar_variable(key):
             raise ValidationError("psi4.core.Wavefunction.set_variable: Target variable " + key +
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -31,6 +31,8 @@ import re
 import sys
 import uuid
 import warnings
+from collections import Counter
+from itertools import product
 
 import numpy as np
 
@@ -76,8 +78,13 @@ def _pybuild_basis(mol,
     # if a string, they search for a gbs file with that name.
     # if a function, it needs to apply a basis to each atom.
 
-    bs, basisdict = qcdb.BasisSet.pyconstruct(
-        mol.to_dict(), key, resolved_target, fitrole, other, return_dict=True, return_atomlist=return_atomlist)
+    bs, basisdict = qcdb.BasisSet.pyconstruct(mol.to_dict(),
+                                              key,
+                                              resolved_target,
+                                              fitrole,
+                                              other,
+                                              return_dict=True,
+                                              return_atomlist=return_atomlist)
 
     if return_atomlist:
         atom_basis_list = []
@@ -123,6 +130,7 @@ def _core_wavefunction_build(mol, basis=None):
 
 core.Wavefunction.build = _core_wavefunction_build
 
+
 def _core_wavefunction_get_scratch_filename(self, filenumber):
     """ Given a wavefunction and a scratch file number, canonicalizes the name
         so that files can be consistently written and read """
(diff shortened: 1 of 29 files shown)


psi4/driver/p4util/python_helpers.py
```diff
@@ -604,6 +604,54 @@ def _core_set_global_option_python(key, EXTERN):
 
 ## QCvar helps
 
+def _qcvar_warnings(key):
+    if any([key.upper().endswith(" DIPOLE " + cart) for cart in ["X", "Y", "Z"]]):
+        warnings.warn(
+            f"Using scalar QCVariable `{key.upper()}` [D] instead of array `{key.upper()[:-2]}` [e a0] is deprecated, and in 1.5 it will stop working\n",
+        category=FutureWarning,
```
PeterKraus: Why is this a `FutureWarning` and not a `DeprecationWarning`? I am not quite sure how the warning machinery works. After this PR (and in 1.4), when someone accesses the dipole/quadrupole components, this warning is triggered, but in the tests due to `warnings.simplefilter("ignore")` it's silenced. In other use cases, I assume a warning will be printed.

What happens in 1.5? You remove the whole component section eg. in proc.py at line 2880?

https://github.com/psi4/psi4/blob/47bdf344cb500e87e799deed95701a87d9e888af/psi4/driver/procrouting/proc.py#L2880-L2885

psi4/driver/schema_wrapper.py
```diff
@@ -87,7 +87,7 @@
     "scf": {
         "scf_one_electron_energy": {"variables": "ONE-ELECTRON ENERGY"},
         "scf_two_electron_energy": {"variables": "TWO-ELECTRON ENERGY"},
-        "scf_dipole_moment": {"variables": ["SCF DIPOLE X", "SCF DIPOLE Y", "SCF DIPOLE Z"], "conversion_factor": (1 / qcel.constants.dipmom_au2debye)},
+        "scf_dipole_moment": {"variables": "SCF DIPOLE"},
```
PeterKraus: Just out of interest, why not uncomment the quadrupole moment lines below? Or is it not part of the schema, yet?

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1194,6 +1194,17 @@ void OEProp::compute_dipole(bool transition) {
     s << title_ << " DIPOLE Z";
     Process::environment.globals[s.str()] = dipole->get(2);
     wfn_->set_scalar_variable(s.str(), dipole->get(2));
+
+    // Dipole array in au
```
PeterKraus: So `DIPOLE X` etc. are in Debyes and stored in `dipole`, but the `dipole_array` should be in au's, hence the conversion?

psi4/driver/schema_wrapper.py
loriab: quadrupole isn't in the schema yet. but the real reason is add the `mtd +` bit to the qcvar string in the exceptions section https://github.com/psi4/psi4/pull/1800/files#diff-a54f9d4707cbe64f08c3df0dfc0c8376R567

psi4/src/psi4/libmints/oeprop.cc
loriab: Yes, the Debye was a bad choice years ago. So for scalar qcvars, dipole & quadrupole are in D and D A, whereas octupole components are in au. Now that array qcvars exist and aren't a pain to use in qcschema (thanks to Daniel's msgpack, pydantic, and shape setup), this converts both the shape and the units.

Possibly I should have edited the called fn https://github.com/psi4/psi4/pull/1800/files#diff-91bf9e90eb569eec33cd0e5edae2e2b4R1352 to instead return au instead of un-debye-ing the values here, but with other modules calling oeprop, I didn't want to risk it.

psi4/driver/p4util/python_helpers.py
loriab: Python seems to still be sorting out the difference btwn those warnings, too. [this](https://docs.python.org/3/library/warnings.html#warning-categories) is the text from which I settled on FutureWarning. 

Usually I try to fix all the uses of the old syntax within psi4 itself so that users can look for the problems in their own code or input. But since we're changing both units and shape, I think it's important to keep testing both. There _are_ a few lingering warnings in the test suite from triplet and run_json, but the dipole warnings were so thick, I had to suppress them.

Yes, in 1.5 we can delete all the ` X`, ` XX`, ` XXX` qcvars. In practice, we haven't culled any "in 1.4 will stop working" ones yet, but they're fair game.

loriab:
this PR now has higher order multipole access and is complete (imo)

psi4/driver/schema_wrapper.py
PeterKraus: OK maybe this is a stupid question, but if there's an "SCF DIPOLE" why isn't there a "SCF QUADRUPOLE"?

psi4/driver/p4util/python_helpers.py
```diff
@@ -661,14 +817,16 @@ def _core_wavefunction_set_variable(cls, key, val):
...
-            cls.set_array_variable(key, core.Matrix.from_array(val))
+            cls.set_array_variable(key, core.Matrix.from_array(_qcvar_reshape_set(key, val)))
     else:
         if cls.has_array_variable(key):
             raise ValidationError("psi4.core.Wavefunction.set_variable: Target variable " + key +
                                   " already an array variable!")
         else:
             cls.set_scalar_variable(key, val)
 
+    # TODO _qcvar_warnings(key)
```
PeterKraus: Are these todos still todos?

psi4/driver/schema_wrapper.py
loriab: good question. qcvar-wise, there now is, as of this PR. lhs-wise, that's dictated by qcschema, and I've just petitioned for a first quadrupole, https://github.com/MolSSI/QCElemental/pull/207/files#diff-f178293574dc070baaac6713e52806beR54 . probably, no one had made a decision about complete/compressed and the shape of the array. that is, `(6,)` vs. `(3, 3)` for quadrupole.

psi4/src/psi4/libmints/oeprop.h
```diff
@@ -278,7 +278,7 @@ class MultipolePropCalc : public Prop {
     /// Common initialization
     MultipolePropCalc(std::shared_ptr<Wavefunction> wfn, Vector3 const& origin);
     // Output Type of multipole function, name, elec, nuc, tot
```
PeterKraus: The int below is undocumented.

loriab: good point -- I'll stick `and order` into the next commit.

psi4/driver/p4util/python_helpers.py
loriab: regrettably, yes. I'd like to have the qcvar_warnings active to fend off downstream API setting of the QUAD XY vars. unfortunately, there's some internal call I can't find, so I can't make psi4 itself clean, hence the todo.

loriab:
This PR was actually developed over in DDD and is needed there to fix a test after adding Stefan's findif dipder_by_dipole. I'll give this another day to accumulate reviews, but if everyone's busy, I'll merge the commits at the tip of DDD, and they'll get rebased out later. Rest assured, DDD test suite still clean (though only on my localbox). @jeffschriber

loriab:
Let's not merge this until some details of interaction with QCFractal are ironed out. Resulting changes to this PR would be in the schema_wrapper.py file, not psivars.

PR #1805, opened 2020-02-06
Fix performance issue in DF SCF Hessians

### LINKED ISSUE #1145: stuck in analytic frequency computation with HF (tags/v1.2.1)
If I do analytic frequency calculation with HF it seems eternally stuck (with full cpu load) - for this particular molecule at least.

attached log with optimization and frequency calculation that is stuck when using analytical and full successful log with dertype="gradient"

```
E, wfn = optimize('HF',return_wfn=True)
#E, wfn = frequencies('HF',return_wfn=True, dertype='gradient')
E, wfn = frequencies('HF',return_wfn=True)

```
[2202-HF-analytic.log](https://github.com/psi4/psi4/files/2307871/2202-HF-low-analytic.log)
[2202-HF-gradient.log](https://github.com/psi4/psi4/files/2307873/2202-HF-low.log)

andysim:
## Description
Some MO transformations were being performed many times instead of just once in the DF SCF analytic Hessians, as a result of some closed braces being in the wrong place.  The results were correct before but the code was *very* slow.  Sorry for the mistake - my bad.  Fixes #1145

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/scfgrad/jk_grad.cc
+++ b/psi4/src/psi4/scfgrad/jk_grad.cc
@@ -1529,22 +1529,21 @@ void DFJKGrad::compute_hessian()
                 for (int p = oP; p < oP+nP; p++) {
                     for (int m = oM; m < oM+nM; m++) {
                         for (int n = oN; n < oN+nN; n++) {
-                            Amnp[p][m*nso+n] += (*buffer++);
+                            Amnp[p][m*nso+n] = (*buffer++);
                         }
                     }
                 }
-                // c[A] = (A|mn) D[m][n]
-                C_DGEMV('N', np, nso*(size_t)nso, 1.0, Amnp[0], nso*(size_t)nso, Dtp[0], 1, 0.0, cp, 1);
-                // (A|mj) = (A|mn) C[n][j]
-                C_DGEMM('N','N',np*(size_t)nso,na,nso,1.0,Amnp[0],nso,Cap[0],na,0.0,Amip[0],na);
-                // (A|ij) = (A|mj) C[m][i]
-                #pragma omp parallel for
-                for (int p = 0; p < np; p++) {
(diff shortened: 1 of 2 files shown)


andysim:
Many thanks to @JonathonMisiewicz and some very diligent forum members for pointing out the performance problem :)

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -1529,23 +1529,22 @@ void DFJKGrad::compute_hessian()
...
-                // (A|ij) = (A|mj) C[m][i]
-                #pragma omp parallel for
-                for (int p = 0; p < np; p++) {
-                    C_DGEMM('T','N',na,na,nso,1.0,Amip[p],na,Cap[0],na,0.0,&Aijp[0][p * (size_t) na * na],na);
-                }
-
             }
         }
     }
+    // c[A] = (A|mn) D[m][n]
```
dgasmith: I don't quite follow how this is the same, the top looks incorrect as you were repeatedly summing a continuously larger non-zero part of the `Amn` matrix which should give bad results (and fail tests). The new one looks correct however.

andysim: It's initially transforming a matrix of mostly zeros and *assigning* the result (not accumulating, note the 0.0 arg for beta in the DGEMM and DGEMV calls).  Then during each pass through the integrals, that Amn array becomes more populated, until the final iteration when it's the full TEI batch.  Only then should it be transformed, but the transform was accidentally being performed after each shell computation, overwriting the previously computed transformed ints.  So only the final iteration's transform is actually recorded and the code was correct (but very bad).

andysim: I just closed the braces in the wrong place, but didn't pick up on it because the answer is still correct and I haven't done any profiling to optimize this code (note it's still in core and has only minimal threading)

dgasmith: Makes sense, I see now!

Issue #1793, opened 2020-01-31
Broken tests


fevangelista:
Commits prior to PR 65c0921213e6fe2979733f6c316915f79051f558 seem to have broken some tests that I suspect are not checked by our CI system:
```
	255 - ocepa-freq1 (Failed)
	259 - ocepa2 (Failed)
	260 - ocepa3 (Failed)
	273 - omp3-2 (Failed)
	278 - omp3-grad2 (Failed)
	320 - pywrap-checkrun-rhf (Failed)
	321 - pywrap-checkrun-rohf (Failed)
	322 - pywrap-checkrun-uhf (Failed)
```

The following are known to fail due to poor convergence of the occ module
```
	260 - ocepa3 (Failed)
	263 - omp2-3 (Failed)
	274 - omp3-3 (Failed)
```

The error message for `pywrap-checkrun-rhf` is also consistent with convergence issues:
```
  45     -75.0144633211      0.00e+00       1.85e-05         1.42e-04        3.01e-15
  46     -75.0144633211      0.00e+00       1.85e-05         1.42e-04        1.73e-15
  47     -75.0144633211     -1.42e-14       1.85e-05         1.42e-04        9.95e-16
  48     -75.0144633211      0.00e+00       1.85e-05         1.42e-04        5.65e-16
  49     -75.0144633211      0.00e+00       1.85e-05         1.42e-04        3.27e-16
  50     -75.0144633211      0.00e+00       1.85e-05         1.42e-04        1.88e-16
 ======================== OCEPA IS NOT CONVERGED IN 50 ITERATIONS ============
Traceback (most recent call last):
  File "/Users/fevange/Source/psi4/objdir-Debug/stage/bin/psi4", line 331, in <module>
    exec(content)
  File "<string>", line 54, in <module>
  File "/Users/fevange/Source/psi4/objdir-Debug/stage/lib/psi4/driver/driver.py", line 711, in gradient
    wfn = procedures['gradient'][lowername](lowername, molecule=molecule, **kwargs)
  File "/Users/fevange/Source/psi4/objdir-Debug/stage/lib/psi4/driver/procrouting/proc.py", line 752, in select_olccd_gradient
    return func(name, **kwargs)
  File "/Users/fevange/Source/psi4/objdir-Debug/stage/lib/psi4/driver/procrouting/proc.py", line 2147, in run_occ_gradient
    occ_wfn = core.occ(ref_wfn)
RuntimeError:
Fatal Error: OCC iterations did not converge
[...]

Issue #1777, opened 2020-01-07
Level Shifting in OCC


JonathonMisiewicz:
I've been investigating orbital convergence problems in the `occ` module, such as in [this test case](https://github.com/psi4/psi4/blob/master/tests/omp2-3/input.dat). I find that convergence massively improves when I turn level shifting _off_, as it is [on by default](https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L2695). `grep -inr 'level_shift' .` indicates that the level shifting is currently being used exclusively [to scale the diagonal elements of the B matrix in the orbital optimization step of DIIS](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/occ/diis.cc#L74-L77).

@bozkaya, I know this is old code, but can you comment on why you did the level shifting this way? This doesn't seem similar to any level shifting scheme I've ever seen. My quick numerical tests indicate that this makes convergence _much worse_, and getting rid of this heals some tests that @loriab is currently concerned about. Comments from @susilehtola are also welcome for this one.

bozkaya:
I scale the DIIS B matrix diagonals based on the suggestion made by Pulay (JCP, 84, 5728, 1986). It is expected to improve convergence in problematic cases, where B matrix is not diagonally dominant. However, it slightly slow down the diis. @JonathonMisiewicz

loriab:
@bozkaya, should we turn level-shifting off by default then, so it can still be invoked for problem cases but doesn't impede ordinary diis convergence?

JonathonMisiewicz:
This is damping, not level shifting. This keyword should be renamed.

To give you some perspective on how severe these convergence problems are, for the test case I gave, which is in the Psi test suite:
No damping: 16 iterations
0.01 damping: 38 iterations
0.02 damping (default): 53 iterations

So we've made convergence **three times harder** to achieve in this example. Have you seen many cases where this damping is necessary or helpful to converge? Otherwise, I agree with Lori about turning this off by default.

For future reference:
Eq. 5 of the paper has a sign error: the - needs to be a + in order for minimizing to favor a solution where c_i^2 B_ii small, which should prevent individual c coefficients blowing up. The (1+d) rule in the paper is correct, as is the Psi implementation.

loriab:
[...]

PR #1809, opened 2020-02-10
Condense OMP, Fix Correctness Bug: OCC's Open-Heart Surgery IV

### LINKED ISSUE #1808: OMP2.5 Amplitudes Not Converging
Upon examining the reference output from tests, none of the four OMP2.5 tests show convergence of the T2 amplitudes. [The T2 residuals stay at the same order of magnitude for every single iteration.](https://github.com/psi4/psi4/blob/master/tests/omp2p5-2/output.ref#L352-L360) Neither the OMP3 tests nor the DF-OMP2.5 tests show this behavior, which suggests the T2 amplitude non-convergence is a bug in the `occ` modification of OMP3 to produce OMP2.5.

This issue blocks improving orbital convergence in OCC by coupling it to amplitude convergence, issue #1777. Due to this issue, the proposed scheme instead causes non-convergence of both quantities. I will fix this issue as the next PR in the "OCC's Open-Heart Surgery" series. Core developers, please mark this issue as a 1.4 target.

JonathonMisiewicz:
## Description
While working on a fix to #1777, I encountered bug #1808, which blocks my fix. This PR fixes #1808 and condenses some of the relevant code.

The bug is discussed in detail at https://github.com/psi4/psi4/issues/1808. Briefly, OMP2.5 theory has two sets of amplitudes, T2(1) and T2(2). Compared to the very similar OMP3 theory, it halves certain terms. OCC implemented this by halving the T2(2) amplitudes instead of halving intermediates that use them. While the intermediates will be correct, this has the side-effect of changing the T2(2) amplitudes from their current values, so they do not converge. This PR changes the implementation so the necessary terms are halved, but not the T2(2) amplitudes. Now they converge properly and agree better with my Psi4Numpy implementation of OMP2.5.

Disclaimer that @bozkaya has yet to comment. While I'm confident I've implemented the theory correctly, he is more familiar with OMP2.5 than I am.

## Todos
- [x] Fixes a bug causing small (~fifth decimal place) errors in OMP2.5 gradients with conventional integrals.
- [x] Condenses OMP2 and OMP2.5/OMP3's functions for the first order T2 amplitudes. 

## Checklist
- [x] Quick tests pass. `^omp2p5` passes. `^omp3` passes with the exceptions noted in #1793. 

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/occ/omp3_g_int.cc
```diff
@@ -274,59 +257,23 @@ void OCCWave::omp3_g_int() {
...
-        global_dpd_->file2_mat_init(&G);
-        global_dpd_->file2_mat_rd(&G);
-#pragma omp parallel for
-        for (int h = 0; h < nirrep_; ++h) {
-            for (int i = 0; i < aoccpiA[h]; ++i) {
-                for (int j = 0; j < aoccpiA[h]; ++j) {
-                    GooA->set(h, i, j, G.matrix[h][i][j]);
-                }
-            }
-        }
```
dgasmith: Nice!

JonathonMisiewicz:
Re-pinging @bozkaya. Even an estimate on when you can look at this would be helpful.

bozkaya:
You may verify your changes with respect to DFOCC CD-OMP2.5 code. In the case of energy with CD tolerance of 10^-14, CD becomes exact. Hence, you may verify your changes with respect to DFOCC module. As long as you can reproduce my results in OMP2.5 paper (with CDS) or you can get the same results with CD-OMP2.5, you are welcome to make any change in OCC module. Frankly speaking, I have focused on DFOCC since 2014. @JonathonMisiewicz @loriab

JonathonMisiewicz:
The CD test passes, so we're clear to merge once pending reviews are in.

bozkaya:
Then, your changes are okay. Thank you for your effort! @JonathonMisiewicz

PR #1813, opened 2020-02-17
GeomeTRIC Interface

zachglick:
## Description
Allows for the use of the GeomeTRIC optimizer within a Psi4 input. The desired optimization engine, either `geometric` or `optking` (default), can now specified with an `engine` argument to the `optimize()` function. In addition, a dictionary of GeomeTRIC-specific keywords and options (like constraints) may be passed to the optimizer. The `test_h2o_constrained` pytest demonstrates how this is done. 

```
e = optimize(..., engine=`geometric`, geometric_opts={...})
```
Output is consistent with Psi4's default geometry optimization:
<details><summary>Example result </summary>
<p>

```
>>> grep "~" output.dat


  ==> GeomeTRIC Optimizer <==                                                                   ~
  Psi4 convergence criteria QCHEM  not recognized by GeomeTRIC, switching to GAU_TIGHT          ~
  Measures of convergence in internal coordinates in au.                                        ~
  Criteria marked as inactive (o), active & met (*), and active & unmet ( ).                    ~
  --------------------------------------------------------------------------------------------- ~
   Step     Total Energy     Delta E     MAX Force     RMS Force      MAX Disp      RMS Disp    ~
  --------------------------------------------------------------------------------------------- ~
    Convergence Criteria    1.00e-06      1.50e-05      1.00e-05      6.00e-05      4.00e-05    ~
  --------------------------------------------------------------------------------------------- ~
      0  -7.64427364e+01    --------      5.01e-02      4.03e-02      --------      --------    ~
      1  -7.64446505e+01   -1.91e-03      2.68e-03      1.95e-03      3.06e-02      2.16e-02    ~
      2  -7.64446681e+01   -1.77e-05      5.27e-04      4.17e-04      4.22e-03      3.98e-03    ~
      3  -7.64446684e+01   -3.06e-07 *    2.27e-05      2.03e-05      4.11e-04      2.93e-04    ~
      4  -7.64446684e+01    6.91e-10 *    3.28e-06 *    2.74e-06 *    1.78e-05 *    1.49e-05 *  ~
  Optimization converged!                                                                       ~

```

</p>
</details>

## Todos
- [x] Working GeomeTRIC interface
- [x] Improved printing and error handling
- [x] Pytest(s)
- [x] Constrained optimizations
- [x] Process GeomeTRIC keywords in Psi4 input
- [x] Composite energy calls (CBS, etc.)
- [x] Documentation

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/driver.py
```diff
@@ -872,6 +872,70 @@ def properties(*args, **kwargs):
     else:
         return core.variable('CURRENT ENERGY')
 
+def optimize_geometric(name, **kwargs):
+
+    # will move imports to top of file when i'm sure where this method goes
+    # TODO: check if GeomeTRIC is installed at the start
```
loriab: can do something like this to check geometric's presence https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molutil/align.py#L416-L419

psi4/driver/driver.py
```diff
@@ -872,6 +872,70 @@ def properties(*args, **kwargs):
...
+    import qcengine as qcng
+    import qcelemental as qcel
+    from qcelemental.models import OptimizationInput
+
+    core.print_out('\n  Optimizing system with GeomeTRIC through QCEngine\n')
+
+    # TODO: return dummy wavefunction? Or run gradient at optimized geometry to get a wfn?
+    return_wfn = kwargs.pop('return_wfn', False)
+    if return_wfn:
+        raise ValidationError(f"return_wfn=True is not supported with GeomeTRIC.")
```
loriab: of the two choices, I slightly favor running a final grad at the optimized geom to get a wfn

psi4/driver/driver.py
loriab: actually, nevermind this. once qcng is imported, just check if geometric is `.found(raise_error=True)` and that'll handle the warning, I think.

psi4/driver/driver.py
```diff
@@ -872,6 +872,70 @@ def properties(*args, **kwargs):
...
+    })
+
+    # TODO: return history if requested
+    ret = qcng.compute_procedure(input_data, "geometric", raise_error=True, local_options={})
+    last_iteration = ret.trajectory[-1]
+    return_energy = last_iteration.properties.return_energy
+
+    # Update molecule with optimized geometry
+    # is there a better way to do this?
+    molecule_opt = last_iteration.molecule # qcelemental.models.molecule.Molecule
```
loriab: you could pull the `molecule_opt.geometry` and pass the return to `molecule.set_geometry`, I think.

dgasmith:
You can pull geometric apart like so: https://github.com/leeping/geomeTRIC/blob/master/geometric/tests/test_batch_opt.py

This is a bit complex compared to what you wish to do, but it may give a better product.

Here is everything unwound:

```python
import qcengine as qcng
import qcelemental as qcel
import geometric
import pkg_resources
import logging.config
import sys

mol_uc2 = qcel.models.Molecule.from_data(
    """
O 0 0 0
H 0 0 1
H 0 1 0
"""
)

input_data = {
    "keywords": {
        "convergence_set": "GAU_LOOSE",
        "coordsys": "tric",
        "maxiter": 25,
        "enforce": 0.1,
        "constraints": {
            "set": [
                {"type": "distance", "indices": [0, 1], "value": 1},
            ]
        },
#        "program": "psi4",
        "program": "mopac",
    },
    "input_specification": {
        "driver": "gradient",
        "model": {"method": "pm6-d3"},
#        "model": {"method": "b3lyp-d3", "basis": "sto-3g"},
    },
    "initial_molecule": mol_uc2.dict(),
}



# Set a temporary logger to capture output
log_stream = geometric.nifty.RawStreamHandler(stream=sys.stdout)
#log_stream = geometric.nifty.RawStreamHandler(stream=StringIO())
logger = geometric.nifty.logger
logger.addHandler(log_stream)


# Parse JSON
input_opts = geometric.run_json.parse_input_json_dict(input_data)
M, engine = geometric.optimize.get_molecule_engine(**input_opts)

# Handle constraints
constraints_dict = input_opts.get('constraints', {})
constraints_string = geometric.run_json.make_constraints_string(constraints_dict)
Cons, CVals = None, None
if constraints_string:
    if 'scan' in constraints_dict:
        raise ValueError("No scan!")
    Cons, CVals = geometric.optimize.ParseConstraints(M, constraints_string)


# Set up the internal coordinate system
coordsys = input_opts.get('coordsys', 'tric')
CoordSysDict = {
    'cart': (geometric.internal.CartesianCoordinates, False, False),
    'prim': (geometric.internal.PrimitiveInternalCoordinates, True, False),
    'dlc': (geometric.internal.DelocalizedInternalCoordinates, True, False),
    'hdlc': (geometric.internal.DelocalizedInternalCoordinates, False, True),
    'tric': (geometric.internal.DelocalizedInternalCoordinates, False, False)
}

# Build internal coordinates
CoordClass, connect, addcart = CoordSysDict[coordsys.lower()]
IC = CoordClass(
    M,
    build=True,
    connect=connect,
    addcart=addcart,
    constraints=Cons,
    cvals=CVals[0] if CVals is not None else None)



# Get initial coordinates in bohr
coords = M.xyzs[0].flatten() * geometric.nifty.ang2bohr

# Setup an optimizer object
params = geometric.optimize.OptParams(**input_opts)
optimizer = geometric.optimize.Optimizer(coords, M, IC, engine, None, params)

# Print
IC.printConstraints(coords, thre=-1)

def compute(coords, opt):
    mol_dict = mol_uc2.dict()
    mol_dict['geometry'] = coords

    inpmodel = {
        "molecule": mol_dict,
        "driver": "gradient",
        "model": {"method": "pm6"}
    }
    ret = qcng.compute(inpmodel, "mopac")
    opt.E = ret.properties.return_energy
    opt.gradx = ret.return_result
    return ret


optimizer.calcEnergyForce()
optimizer.prepareFirstStep()
logger.info("[AU]: e=%.5f bl=%.5f,%.5f g=%.4f" % (
        optimizer.E, optimizer.X[0],optimizer.X[3], optimizer.gradx[0]))

while True:
    if optimizer.state in [geometric.optimize.OPT_STATE.CONVERGED, geometric.optimize.OPT_STATE.FAILED]:
        logger.info("Optmization convereged!")
        break

    optimizer.step()
    optimizer.calcEnergyForce()
    optimizer.evaluateStep()
    logger.info("[AU]: e=%.5f bl=%.5f,%.5f g=%.4f" % (
            optimizer.E, optimizer.X[0],optimizer.X[3], optimizer.gradx[0]))
```

You may want to build another [Engine](https://github.com/leeping/geomeTRIC/blob/master/geometric/engine.py).

psi4/driver/driver.py
zachglick: very useful to know. geometric is now explicitly imported, so i guess we can just let psi throw an import error now

psi4/driver/driver.py
zachglick: I wonder if we should always run a final gradient (with or without a wavefunction request) in order to populate qcvars?

zachglick:
> You can pull geometric apart like so: https://github.com/leeping/geomeTRIC/blob/master/geometric/tests/test_batch_opt.py
> 
> This is a bit complex compared to what you wish to do, but it may give a better product.
> 
> Here is everything unwound:
> 
> ```python
> import qcengine as qcng
> import qcelemental as qcel
> import geometric
> import pkg_resources
> import logging.config
> import sys
> 
> mol_uc2 = qcel.models.Molecule.from_data(
>     """
> O 0 0 0
> H 0 0 1
> H 0 1 0
> """
> )
> 
> input_data = {
>     "keywords": {
>         "convergence_set": "GAU_LOOSE",
>         "coordsys": "tric",
>         "maxiter": 25,
>         "enforce": 0.1,
>         "constraints": {
>             "set": [
>                 {"type": "distance", "indices": [0, 1], "value": 1},
>             ]
>         },
> #        "program": "psi4",
>         "program": "mopac",
>     },
>     "input_specification": {
>         "driver": "gradient",
>         "model": {"method": "pm6-d3"},
> #        "model": {"method": "b3lyp-d3", "basis": "sto-3g"},
>     },
>     "initial_molecule": mol_uc2.dict(),
> }
> 
> 
> 
> # Set a temporary logger to capture output
> log_stream = geometric.nifty.RawStreamHandler(stream=sys.stdout)
> #log_stream = geometric.nifty.RawStreamHandler(stream=StringIO())
> logger = geometric.nifty.logger
> logger.addHandler(log_stream)
> 
> 
> # Parse JSON
> input_opts = geometric.run_json.parse_input_json_dict(input_data)
> M, engine = geometric.optimize.get_molecule_engine(**input_opts)
> 
> # Handle constraints
> constraints_dict = input_opts.get('constraints', {})
> constraints_string = geometric.run_json.make_constraints_string(constraints_dict)
> Cons, CVals = None, None
> if constraints_string:
>     if 'scan' in constraints_dict:
>         raise ValueError("No scan!")
>     Cons, CVals = geometric.optimize.ParseConstraints(M, constraints_string)
> 
> 
> # Set up the internal coordinate system
> coordsys = input_opts.get('coordsys', 'tric')
> CoordSysDict = {
>     'cart': (geometric.internal.CartesianCoordinates, False, False),
>     'prim': (geometric.internal.PrimitiveInternalCoordinates, True, False),
>     'dlc': (geometric.internal.DelocalizedInternalCoordinates, True, False),
>     'hdlc': (geometric.internal.DelocalizedInternalCoordinates, False, True),
>     'tric': (geometric.internal.DelocalizedInternalCoordinates, False, False)
> }
> 
> # Build internal coordinates
> CoordClass, connect, addcart = CoordSysDict[coordsys.lower()]
> IC = CoordClass(
>     M,
>     build=True,
>     connect=connect,
>     addcart=addcart,
>     constraints=Cons,
>     cvals=CVals[0] if CVals is not None else None)
> 
> 
> 
> # Get initial coordinates in bohr
> coords = M.xyzs[0].flatten() * geometric.nifty.ang2bohr
> 
> # Setup an optimizer object
> params = geometric.optimize.OptParams(**input_opts)
> optimizer = geometric.optimize.Optimizer(coords, M, IC, engine, None, params)
> 
> # Print
> IC.printConstraints(coords, thre=-1)
> 
> def compute(coords, opt):
>     mol_dict = mol_uc2.dict()
>     mol_dict['geometry'] = coords
> 
>     inpmodel = {
>         "molecule": mol_dict,
>         "driver": "gradient",
>         "model": {"method": "pm6"}
>     }
>     ret = qcng.compute(inpmodel, "mopac")
>     opt.E = ret.properties.return_energy
>     opt.gradx = ret.return_result
>     return ret
> 
> 
> optimizer.calcEnergyForce()
> optimizer.prepareFirstStep()
> logger.info("[AU]: e=%.5f bl=%.5f,%.5f g=%.4f" % (
>         optimizer.E, optimizer.X[0],optimizer.X[3], optimizer.gradx[0]))
> 
> while True:
>     if optimizer.state in [geometric.optimize.OPT_STATE.CONVERGED, geometric.optimize.OPT_STATE.FAILED]:
>         logger.info("Optmization convereged!")
>         break
> 
>     optimizer.step()
>     optimizer.calcEnergyForce()
>     optimizer.evaluateStep()
>     logger.info("[AU]: e=%.5f bl=%.5f,%.5f g=%.4f" % (
>             optimizer.E, optimizer.X[0],optimizer.X[3], optimizer.gradx[0]))
> ```
> 
> You may want to build another [Engine](https://github.com/leeping/geomeTRIC/blob/master/geometric/engine.py).

This is super helpful, thanks!

tests/pytests/test_geometric.py
```diff
@@ -0,0 +1,38 @@
...
+
+from .utils import *
+
+import psi4
+from qcengine.testing import using
+
+
+@using('geometric')
+@pytest.mark.parametrize('opt', [
+    pytest.param({'engine' : 'optking'}, id='optking'),
```
dgasmith: Dict's are likely not optimal here a simple `['optking', 'geometric']` is likely better.

psi4/driver/driver.py
```diff
@@ -873,6 +873,150 @@ def properties(*args, **kwargs):
...
+        molecule.fix_com(True)
+    molecule.update_geometry()
+
+    input_data = {
+        "keywords": {
+            "convergence_set": core.get_global_option("G_CONVERGENCE"),
+            "maxiter": core.get_global_option("GEOM_MAXITER"),
+            "coordsys": "tric",
+            "enforce": 0.1,
+            "constraints": {},
```
dgasmith: Constraints are probably the #1 most vital item here.

psi4/driver/driver.py
```diff
@@ -873,6 +873,150 @@ def properties(*args, **kwargs):
...
+        },
+        "initial_molecule": molecule.to_schema(dtype=2),
+    }
+
+    if input_data["keywords"]["convergence_set"] in ["CFOUR", "QCHEM", "MOLPRO"]:
+        input_data["keywords"]["convergence_set"] = "GAU_TIGHT"
+
+    # Parse JSON
+    input_opts = geometric.run_json.parse_input_json_dict(input_data)
+    M, engine = geometric.optimize.get_molecule_engine(**input_opts)
```
dgasmith: You will want to build your own engine, see [here](https://github.com/leeping/geomeTRIC/blob/master/geometric/engine.py#L846). In addition you can build your own custom Molecule conversion class which is commended I think.

psi4/driver/driver.py
loriab: Excellent, Lee-Ping does have a language: https://github.com/leeping/geomeTRIC/blob/master/examples/constraints.txt . Only other one I know was Gaussian's ModRedundant ("Options" tab off https://gaussian.com/opt/). Psi4 has a not-great-one due to constraints of the options class - https://github.com/psi4/psi4/blob/master/tests/opt-freeze-coords/input.dat#L21-L22 - and I know Rollin's wanting to get away from it.

dgasmith: I would use the JSON version and not the string versions.

```python
{
    "constraints": {
        "set": [{"indices": [0, 1], "type": "distance", "value": 1.4}],
        "freeze": [{"indices": [0, 1, 2], "type": "angle"}],
    }
}
```

zachglick: You can now specify constraints in the JSON format (which geometric internally converts to strings). I added a pytest demonstrating how to do this.

psi4/driver/driver.py
zachglick: Thanks, using a custom engine simplified things a lot

psi4/driver/driver.py
```diff
@@ -873,6 +873,158 @@ def properties(*args, **kwargs):
...
+            molecule.build_bonds()
+                                 
+            super(Psi4NativeEngine, self).__init__(molecule)
+    
+        def calc(self, coords, dirname):
+            self.p4_mol.set_geometry(core.Matrix.from_array(coords.reshape(-1,3)))
+            self.p4_mol.update_geometry()
+            g, wfn = gradient(self.p4_name, return_wfn=True, molecule=self.p4_mol, **self.p4_kwargs)
+            e = core.variable('CURRENT ENERGY')
+            return {'energy': e, 'gradient': (g.np).reshape(-1)}
```
PeterKraus, line 904: Wouldn't `(g.np).ravel()` be a bit clearer? Do we have a style policy on `reshape`s and `ravel`s?

psi4/driver/driver.py
```diff
@@ -873,6 +873,158 @@ def properties(*args, **kwargs):
...
+    engine = Psi4NativeEngine(name, molecule, **kwargs)
+    M = engine.M
+    
+    # Handle constraints
+    constraints_dict = geometric_opts.get('constraints', {})
+    constraints_string = geometric.run_json.make_constraints_string(constraints_dict)
+    Cons, CVals = None, None
+    if constraints_string:
+        if 'scan' in constraints_dict:
+            raise ValueError("No scan!")
```
PeterKraus, line 943: A bit more verbose, like `"scan not yet implemented"` would be better. Why doesn't it work? Not implemented in GeomeTRIC, or in the Psi4-GeomeTRIC interface?

psi4/driver/driver.py
```diff
@@ -873,6 +873,158 @@ def properties(*args, **kwargs):
...
+    core.print_out("\n  Criteria marked as inactive (o), active & met (*), and active & unmet ( ).")
+    core.print_out("\n  ---------------------------------------------------------------------------------------------")
+    core.print_out("\n   Step     Total Energy     Delta E     MAX Force     RMS Force      MAX Disp      RMS Disp   ")
+    core.print_out("\n  ---------------------------------------------------------------------------------------------")
+    core.print_out((f"\n    Convergence Criteria  {params.Convergence_energy:10.2e}    "
+                    f"{params.Convergence_gmax:10.2e}    {params.Convergence_grms:10.2e}    "
+                    f"{params.Convergence_dmax:10.2e}    {params.Convergence_drms:10.2e}"))
+    core.print_out("\n  ---------------------------------------------------------------------------------------------")
+
+    core.print_out((f"\n   {optimizer.Iteration:4d} {optimizer.E:16.8e}    -------- *  "
```
PeterKraus, line 992: I really like the fact that you can grep OPTKING steps using `~`. Could we do that here too?

psi4/driver/driver.py
```diff
@@ -873,6 +873,158 @@ def properties(*args, **kwargs):
...
+                    f"{gmax:10.2e} {conv_gmax}  {grms:10.2e} {conv_grms}    -------- *    -------- *"))
+    while True:
+        if optimizer.state in [geometric.optimize.OPT_STATE.CONVERGED, geometric.optimize.OPT_STATE.FAILED]:
+            core.print_out("\n\n  Optmization converged!\n")
```
PeterKraus, line 996: Wrong spelling. Should we be printing "Optimization converged" in case the `OPT_STATE.FAILED` case?

psi4/driver/driver.py
```diff
@@ -873,6 +873,158 @@ def properties(*args, **kwargs):
...
+            molecule.elem = [p4_mol.symbol(i) for i in range(p4_mol.natom())]
+            molecule.xyzs = [p4_mol.geometry().np * qcel.constants.bohr2angstroms] # reshape?
+            molecule.build_bonds()
+                                 
+            super(Psi4NativeEngine, self).__init__(molecule)
+    
+        def calc(self, coords, dirname):
+            self.p4_mol.set_geometry(core.Matrix.from_array(coords.reshape(-1,3)))
+            self.p4_mol.update_geometry()
+            g, wfn = gradient(self.p4_name, return_wfn=True, molecule=self.p4_mol, **self.p4_kwargs)
```
PeterKraus, line 902: Regarding the fact that CBS calls apparently don't work: I didn't try it out, but I don't see why a `cbs` call with `cbs_metadata` in kwargs shouldn't work here... If anything is missing, let me know, maybe I can help.

psi4/driver/driver.py
zachglick: I wasn't aware of `.ravel()`, it's definitely clearer

psi4/driver/driver.py
zachglick: You're totally right. CBS might not have worked before DGAS's suggestion to use a custom engine, but now it works fine without any modification

psi4/driver/driver.py
zachglick: As far as I know the scan feature works in GeomeTRIC, but I think that hooking it up in Psi4 would require additional accommodations to the geometry optimization loop and print out (since you're iterating over additional variables). I changed the print out to be a little more clear.

zachglick:
> This looks great already and I'm looking forward to trying it out. I have a couple of minor comments, though. Also, can GeomeTRIC do TS searches? If yes, perhaps a test would be good.

Thanks Peter, your comments are very helpful! I don't think transition state searches are implemented yet, although I see that there is a PR in the GeomeTRIC repo:
https://github.com/leeping/geomeTRIC/pull/107

tests/pytests/test_geometric.py
```diff
@@ -0,0 +1,73 @@
...
+    assert compare_values(inp['ref_nuc'], h2o.nuclear_repulsion_energy(), 3)
+
+
+@using('geometric')
+@pytest.mark.parametrize('inp', [
+    pytest.param({'name': 'hf', 'options': {'scf_type': 'df'}, 'ref_ene' : -76.02079629252714, 'ref_nuc': 9.265341708725257}, id='rhf(df)'),
+    pytest.param({'name': 'hf', 'options': {'scf_type': 'pk'}, 'ref_ene' : -76.02082389228, 'ref_nuc': 9.26528625744628}, id='rhf(pk)'),
+    pytest.param({'name': 'mp2', 'options': {'mp2_type': 'df'}, 'ref_ene' : -76.22711819393223, 'ref_nuc': 9.09137805747361}, id='mp2(df)'),
+    pytest.param({'name': 'mp2', 'options': {'mp2_type': 'conv'}, 'ref_ene' : -76.2271678506303, 'ref_nuc': 9.091178486990861}, id='mp2(conv)'),
+    pytest.param({'name': 'b3lyp', 'options': {'scf_type': 'df'}, 'ref_ene' : -76.41632755714534, 'ref_nuc': 9.04535641436914}, id='b3lyp'),
```
loriab, line 43: How (step & opt convergence conditions) did you get the ref E and nre? It can be difficult to make opt tests checks that are robust to the optimizer taking an extra step (either b/c of platform or noise or algorithm changes). It can be good to crank up opt crit to verytight and all the e/d/r_convergences to 10 to gather the ref values. Then, at least when the test fails, runtime opt convergence and compare_values atol are rationally adjustable.

psi4/driver/driver.py
```diff
@@ -873,6 +873,190 @@ def properties(*args, **kwargs):
...
+    if return_history:
+        history = {
+            'energy': step_energies,
+            'gradient': step_gradients,
+            'coordinates': step_coordinates,
+        }
+
+    # run gradient at optimized geometry to get a wfn
+    if return_wfn:
+        g, wfn = gradient(name, return_wfn=True, **kwargs)
```
loriab, line 1046: I got nervous about different P::e.globals or other settings depending on whether return_wfn = T/F. Even if it's fine now, may change btwn psithon/psiapi, if qcng is involved whether extras['psiapi'] = T/F, and with ddd or not. I don't particularly want to waste cycles by always running the extra gradient calc, but may be worth it to avert routing bugs.

psi4/driver/driver.py
```diff
@@ -873,6 +873,190 @@ def properties(*args, **kwargs):
...
+    import qcelemental as qcel
+
+    class Psi4NativeEngine(geometric.engine.Engine):
+        """
+        Internally run an energy and gradient calculation for geometric 
+        """
+        def __init__(self, p4_name, p4_mol, **p4_kwargs):
+    
+            self.p4_name = p4_name
+            self.p4_mol = p4_mol
```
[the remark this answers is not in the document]
zachglick: I tested it out, dummy atoms in the geometry are fine

tests/pytests/test_geometric.py
```diff
@@ -0,0 +1,73 @@
+import pytest
+
+from .utils import *
+
+import psi4
+from qcengine.testing import using
+
+
+@using('geometric')
+@pytest.mark.parametrize('engine', ['optking', 'geometric'])
```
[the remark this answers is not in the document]
zachglick: Good to know that's possible, thanks!

tests/pytests/test_geometric.py
zachglick: I got the reference energies exactly as you described: `e_convergence` : 10, `d_convergence` : 8, `g_convergence` : `GAU_TIGHT`. I wouldn't have any problem with relaxing the comparison criteria, since I imagine any future error that would cause a test to fail will do so in a major way.

psi4/driver/driver.py
zachglick: Are you saying we should run that gradient calculation regardless of the value of `return_wfn`? I think that would be a pretty reasonable decision. 

I wouldn't be too worried about the extra cycle. In my limited experience, GeomeTRIC usually requires many more optimization cycles than Optking, so if a user has made the decision to use GeomeTRIC, a single additional cycle would not be that much of a problem.

loriab: Good point on number of iterations. In that case, yes, go for final gradient() call under all circumstances.

zachglick: Even better, I did away with the final gradient call. I can stick the wfn to the GeomeTRIC Engine object during the optimization, so there's no need to make any additional calls after the optimization loop.

tests/pytests/test_geometric.py
loriab: Sounds workable. It can be good to get the ref nre from a verytight opt convergence, then run at plain tight opt convergence so that an extra opt step moves closer to ref value rather than away from it. But existing will be fine for now.

psi4/driver/driver.py
```diff
@@ -873,6 +873,194 @@ def properties(*args, **kwargs):
...
+    core.print_out((f"\n    Convergence Criteria  {params.Convergence_energy:10.2e}    "
+                    f"{params.Convergence_gmax:10.2e}    {params.Convergence_grms:10.2e}    "
+                    f"{params.Convergence_dmax:10.2e}    {params.Convergence_drms:10.2e}    ~"))
+    core.print_out("\n  --------------------------------------------------------------------------------------------- ~")
+
+    core.print_out((f"\n   {optimizer.Iteration:4d} {optimizer.E:16.8e}    --------    "
+                    f"{gmax:10.2e} {conv_gmax}  {grms:10.2e} {conv_grms}    --------      --------    ~"))
+    while True:
+        if optimizer.state == geometric.optimize.OPT_STATE.CONVERGED:
+            core.print_out("\n\n  Optimization converged!                                                                       ~\n")
```
PeterKraus, line 1010: You're missing a closing `)` here and also below.

doc/sphinxman/source/optking.rst
```diff
@@ -441,6 +441,18 @@ achieve convergence. To avoid this revokation, turn on keyword |optking__flexibl
...
+
+The GeomeTRIC optimizer developed by Wang and Song [Wang:2016:214108]_ may be used in place of
+Psi4's native Optking optimizer. GeomeTRIC uses a translation-rotation-internal coordinate (TRIC)
+system that works well for optimizing geometries of systems containing noncovalent interactions.
+
+Use of the GeomeTRIC optimizer is specified with the ``engine`` argument to
+:py:func:`~psi4.optimize`. The optimization will respect the keywords |optking__g_convergence|
+and |optking__geom_maxiter|. Any other GeomeTRIC-specific options (including constraints)
+may be specified with the ``geometric_opts`` argument to :py:func:`~psi4.optimize`.
+
```
dgasmith, line 455: I think we need a few constrained optimization examples here.

doc/sphinxman/source/bibliography.rst
```diff
@@ -459,3 +459,7 @@ Bibliography
    M. F. Herbst, M. Scheurer, T. Fransson, D. R. Rehn, and A. Dreuw.
    *WIREs Comput. Mol. Sci.*, **DOI: 10.1002/wcms.1462**, Preprint https://adc-connect.org/q/publications
 
+.. [Wang:2016:214108]
+   L. Wang and C. Song,
```
andysim, line 463:
```suggestion
   L.-P. Wang and C. Song,
```

psi4/driver/driver.py
```diff
@@ -873,6 +873,194 @@ def properties(*args, **kwargs):
...
 
 
+
+def optimize_geometric(name, **kwargs):
+
+    import qcelemental as qcel
+    from qcelemental.util import which_import
+
+    if not which_import('geometric', return_bool=True):
+        raise ModuleNotFoundError('Python module geometric not found. Solve by installing it: `conda install -c conda-forge geometric` or `pip install geometric`')
```
andysim, line 883: Excellent!  This kind of error message really helps

psi4/driver/driver.py
```diff
@@ -873,6 +873,194 @@ def properties(*args, **kwargs):
...
+    molecule.update_geometry()
+
+    # Get geometric-specific options
+    geometric_opts = kwargs.get('geometric_opts', {})
+
+    core.print_out('\n')
+    core.print_out("\n  ==> GeomeTRIC Optimizer <==                                                                   ~\n")
+                                 
+    # Default to Psi4 maxiter unless overridden
+    if 'maxiter' not in geometric_opts:
```
andysim, line 938: Do we need to worry about this being mixed / upper case at all?  I know @loriab sanitizes these things at various points, so maybe not

psi4/driver/driver.py
```diff
@@ -873,6 +873,194 @@ def properties(*args, **kwargs):
...
+
+    engine = Psi4NativeEngine(name, molecule, return_wfn, **kwargs)
+    M = engine.M
+    
+    # Handle constraints
+    constraints_dict = geometric_opts.get('constraints', {})
+    constraints_string = geometric.run_json.make_constraints_string(constraints_dict)
+    Cons, CVals = None, None
+    if constraints_string:
+        if 'scan' in constraints_dict:
```
andysim, line 958: Same here - just want to check that we know `scan` is always lower case

leeping:
Hi there, thanks a lot for including me on this.  :)  

Transition state optimization is implemented, but we have not tested it extensively against other codes.  It does work quite well for in-house applications containing 50+ atoms.

I'm very interested to see how you run these optimizations directly in Psi4.  It should be a lot more efficient than calling Psi4 repeatedly on the command line.

Also happy to provide examples of constrained optimization.  Let me know if you need any.

susilehtola:
Is there a performance benefit over running the program in the command line? Nuclear forces / hessians i/o is inconsequential compared to the quantum chemistry part. Any savings would come from reusing checkpoint information for the Fock / density matrices... right?

leeping:
I'm just referring to the overhead associated with setup & teardown of the calculation.  It could slow things down if the calculations are fast (which might be the case with semiempirical methods or minimal basis sets).

dgasmith:
It can a bit, Psi's startup time is ~0.4 seconds or so with all of Python loading in. In general QC will dwarf this time so it isn't much of an issue. Your right though with XTB and DFTB it gets more interesting.

psi4/driver/driver.py
loriab: Good point. kwargs are sanitized to lowercase keys [here](https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/procutil.py#L47), but the geometric_opts dictionary won't be affected. Can fix with something like: `optimizer_keywords = {k.lower(): v for k, v in kwargs.get("optimizer_keywords", {}).items()}`

Is "geometric_opts" a name used in geometric? If not, may want something like "optimizer_keywords" to fall in line with qcsk and plan for future optimizers.

zachglick: Very helpful, thanks. `optimizer_keywords` does make more sense.

psi4/driver/driver.py
zachglick: Good catch!

doc/sphinxman/source/optking.rst
zachglick: Added

PR #1814, opened 2020-02-18
TDDFT

robertodr:
## Description
Small improvements to the TDDFT solver. This one should be merged after #1885

## Todos
- [X] Document
- [x] Handle computation of triplets properly
- [x] Implement transition moments and spectroscopic observables:
     - [x] Electric dipole moment, length representation
     - [x] Electric dipole moment, velocity representation
     - [x] Magnetic dipole moment
     - [x] Oscillator strengths, length representation
     - [x] Oscillator strengths, velocity representation
     - [x] Rotatory strengths, length representation
     - [x] Rotatory strengths, velocity representation
- [x] Robustness fix: `states_per_irrep` sometimes fails with an obscure out-of-bounds error.
- [x] Usability fix: verbosity should be slightly higher, to let the user know stuff is being done (currently, only printout is at convergence)
- [x] Implement functions to plot electronic absorption (EAS) and eletronic circular dichroism (ECD) spectra
- [x] Remove warning on potential inaccuracy of results.
- [X] Check convergence only against the norm of the residual (`r_tol`). The `e_tol` parameter has been completely removed.

## Follow-up
- The solver stalls when very tight convergence (`r_tol = 1e-8`) is requested. I am not quite sure if  this is a fundamental issue with the algorithm. That convergence threshold is insanely tight, I am not even sure if this should be considered a real problem...

## Checklist

## Status
- [x] Ready for review
- [X] Ready for merge


psi4/driver/p4util/solvers.py
```diff
@@ -126,8 +124,8 @@ def cg_solver(rhs_vec, hx_function, preconditioner, guess=None, printer=None, pr
         resid = printer(0, x_vec, r_vec)
     elif printlvl:
         # core.print_out('         CG Iteration Guess:    Rel. RMS = %1.5e\n' %  np.mean(resid))
-        core.print_out("    %5s %14.3e %12.3e %7d %9d\n" % ("Guess", np.mean(resid), np.max(resid), len(z_vec),
-                                                            time.time() - tstart))
+        core.print_out("    %5s %14.3e %12.3e %7d %9d\n" %
```
dgasmith: @loriab Can we use strings yet?

loriab: f-strings? yes.

only f-string debugging (https://tirkarthi.github.io/programming/2019/05/08/f-string-debugging.html) like `f"{resid=}` is forbidden to us.

robertodr: I'm moving printing to use f-strings while a move through the code.

robertodr:
Unit tests fail because the velocity-gauge electric dipole transition moments are not quite the same as the reference.

robertodr:
This is ready for review. With my latest changes:
- The user specifies a number of roots to seek to converge. If the system has symmetry, we spread them around among the irreps. For example, 10 roots in C2v means `[3, 3, 2, 2]` in whatever order the irreps are for the given system. To have more granular control over what symmetries to seek roots in, the user can pass a list: `[10, 0, 0, 0]` will seek the 10 roots in the totally symmetric irrep.
- Option for triplets is now passed down to the solver correctly. `"none"` is the default (singlet roots only); `"also"` will compute as many singlet roots as triplet roots, so this makes the calculation cost twice as much (at least).

Notable limitations:
   - restricted triplets: RPA or TDA, Hartree-Fock only. Notes in code state that the XC kernel is only spin-adapted for singlets.
   - unresctricted: RPA or TDA, Hartre-Fock and LDA only. Notes in code state that there is a bug for UKS beyond LDA, but aren't more specific than that.

I am waiting on #1885 to be in to write docs.

An example of PsiAPI usage:
```python
import psi4

from psi4.driver.procrouting.response.scf_response import tdscf_excitations
from psi4.driver.p4util import spectrum

psi4.core.set_output_file("moxy.out")

# structure from Pederson et al., CPL, submitted
moxy = psi4.geometry("""0 1
C  0.152133 -0.035800  0.485797
C -1.039475  0.615938 -0.061249
C  1.507144  0.097806 -0.148460
O -0.828215 -0.788248 -0.239431
H  0.153725 -0.249258  1.552136
H -1.863178  0.881921  0.593333
H -0.949807  1.214210 -0.962771
H  2.076806 -0.826189 -0.036671
H  2.074465  0.901788  0.325106
H  1.414895  0.315852 -1.212218
""", name="(S)-methyloxirane")

psi4.set_options({
    'save_jk': True,
})

method = 'HF'
basis = 'cc-pVDZ'
e, wfn = psi4.energy(f"{method}/{basis}", return_wfn=True, molecule=moxy)
res = tdscf_excitations(wfn, states=8, triplets="also")

from typing import Tuple, Dict

import numpy as np
import pandas as pd
import altair as alt

def plot_spectrum(data: Dict,
               *,
               title: str = "",
               x_title: Tuple[str, str] = ("ω", "au"),
               y_title: Tuple[str, str] = ("ε", "L⋅mol⁻¹⋅cm⁻¹"),
               offset: int = 0):
    hover = alt.selection_single(
      fields=["x"],
      nearest=True,
      on="mouseover",
      empty="none",
      clear="mouseout"
    )

    s1 = pd.DataFrame(data["convolution"])
    lines = alt.Chart(s1).mark_line(size=1.5).encode(
       x=alt.X("x", axis=alt.Axis(title=f"{x_title[0]} [{x_title[1]}]", offset=offset)),
       y=alt.Y("y", axis=alt.Axis(title=f"{y_title[0]} [{y_title[1]}]")),
       )

    points = lines.transform_filter(hover).mark_circle()

    tooltips = alt.Chart(s1).mark_rule().encode(
      x='x:Q',
      opacity=alt.condition(hover, alt.value(0.3), alt.value(0)),
      tooltip=[alt.Tooltip("x:Q", format=".4f", title=f"{x_title[0]}"), alt.Tooltip("y:Q", format=".1f", title=f"{y_title[0]}")]
      ).add_selection(
        hover
        )

    s2 = pd.DataFrame(data["sticks"])
    sticks = alt.Chart(s2).mark_bar(size=2, opacity=0.2, color="red").encode(
        x="poles:Q",
        y="residues:Q",
        )

    # Put the layers into a chart and bind the data
    plot = alt.layer(
      lines, points, tooltips, sticks,
      ).properties(
        title=title,
        )

    return plot

# get poles and residues to plot OPA and ECD spectra
poles = [r["EXCITATION ENERGY"] for r in res]
opa_residues = [np.linalg.norm(r["LENGTH-GAUGE ELECTRIC DIPOLE TRANSITION MOMENT"])**2 for r in res]
ecd_residues = [r["LENGTH-GAUGE ROTATORY STRENGTH"] for r in res]

opa_spectrum = spectrum(poles=poles, residues=opa_residues, gamma=0.01, out_units="nm")
opa_plot = plot_spectrum(opa_spectrum,
                         title="OPA (Gaussian broadening)",
                         x_title=("λ", "nm"))

ecd_spectrum = spectrum(poles=poles, residues=ecd_residues, kind="ECD", gamma=0.01, out_units="nm")
ecd_plot = plot_spectrum(ecd_spectrum,
                         title="ECD (Gaussian broadening)",
                         x_title=("λ", "nm"),
                         y_title=("Δε", "L⋅mol⁻¹⋅cm⁻¹"))

(opa_plot & ecd_plot).save("moxy.html")
```
and obtain these plots:
visualization

The plotting function itself (`plot_spectrum`) is not in Psi4. The `spectrum` function will compute the convolution with the selected lineshape and then with that data any plotting library will work. Here I am using [Altair](https://altair-viz.github.io/)

The output file contains the following summary:
```
******************************************************************************************
**********                               WARNING                                **********
**********  Length-gauge rotatory strengths are **NOT** gauge-origin invariant  **********
******************************************************************************************

                                    Excitation Energy         Total Energy        Oscillator Strength             Rotatory Strength       
     #   Sym: GS->ES (Trans)        au              eV              au          au (length)    au (velocity)    au (length)    au (velocity) 
    ---- -------------------- --------------- --------------- --------------- --------------- --------------- --------------- ---------------
     1        A->A (3 A)       0.29232         7.95436        -191.62963       0.0000          0.0000          0.0000         -0.0000        
     2        A->A (3 A)       0.32745         8.91025        -191.59450       0.0000          0.0000          0.0000         -0.0000        
     3        A->A (3 A)       0.33091         9.00446        -191.59104       0.0000          0.0000          0.0000         -0.0000        
     4        A->A (3 A)       0.34771         9.46180        -191.57423       0.0000          0.0000          0.0000         -0.0000        
     5        A->A (3 A)       0.36617         9.96389        -191.55578       0.0000          0.0000          0.0000         -0.0000        
     6        A->A (1 A)       0.36723         9.99278        -191.55472       0.0030          0.0073         -0.0070         -0.0076        
     7        A->A (3 A)       0.38101         10.36789       -191.54093       0.0000          0.0000          0.0000         -0.0000        
     8        A->A (3 A)       0.38434         10.45833       -191.53761       0.0000          0.0000          0.0000         -0.0000        
     9        A->A (1 A)       0.38603         10.50444       -191.53591       0.0057          0.0118          0.0207          0.0211        
     10       A->A (3 A)       0.39165         10.65733       -191.53030       0.0000          0.0000          0.0000         -0.0000        
     11       A->A (1 A)       0.39470         10.74043       -191.52724       0.0255          0.0287          0.0509          0.0617        
     12       A->A (1 A)       0.40636         11.05757       -191.51559       0.1813          0.1877          0.0663          0.0563        
     13       A->A (1 A)       0.41678         11.34118       -191.50516       0.1340          0.1330         -0.0054         -0.0118        
     14       A->A (1 A)       0.41767         11.36536       -191.50428       0.1386          0.1429          0.0673          0.0764        
     15       A->A (1 A)       0.42651         11.60586       -191.49544       0.1515          0.1547          0.0072          0.0178        
     16       A->A (1 A)       0.43710         11.89422       -191.48484       0.0668          0.0802         -0.0746         -0.0822        
```

psi4/driver/p4util/solvers.py
```diff
@@ -946,15 +976,16 @@ def hamiltonian_solver(engine,
...
         #    - User would probably not expect this
         # 3. Perform Stability update and restart with new reference
         if np.any(H2_ss_val < 0.0):
-            raise Exception("H2 is not Positive Definite")
+            msg = ("The H2 matrix is not Positive Definite. " "This means the reference state is not stable.")
+            raise RuntimeError(msg)
 
         # Build H2^(1/2)
-        H2_ss_half = np.dot(H2_ss_vec, np.diag(np.sqrt(H2_ss_val))).dot(H2_ss_vec.T)
+        H2_ss_half = np.einsum("ik,k,jk->ij", H2_ss_vec, np.sqrt(H2_ss_val), H2_ss_vec)
```
loriab, line 988: hooray for einsum notation and readability

psi4/driver/p4util/solvers.py
```diff
@@ -606,48 +623,63 @@ def vector_copy(X):
...
+       least one "large" dimension. See :class:`SolverEngine` for requirements
     guess : list {engine dependent}
        At least `nroot` initial expansion vectors
-    e_tol : float
-        Convergence tolerance for eigenvalues
-    r_tol : float
-        Convergence tolerance for residual vectors
     nroot : int
         Number of roots desired
+    r_tol : float
```
loriab, line 675: * Is this close enough to the general option `r_convergence` that we should call it that (maybe with a TDSCF prefix) for the user?
* If so, do you think it's worth doing a wholesale `r_tol` --> `r_convergence` in the solver code? Or maybe `r_conv` as compromise?
* If r_tol is not touched by user, should its default be (in exponents) SCF `e_convergence` - 2 (or `d_convergence`)?

psi4/driver/p4util/spectrum.py
```diff
@@ -0,0 +1,376 @@
...
+        k = \frac{4\pi^{2}N_{\mathrm{A}}}{3\times 1000\times \ln(10) (4 \pi \epsilon_{0}) n \hbar c}
+
+    The prefactor is computed in SI units and then adjusted for the fact that
+    we use atomic units to express microscopic observables: excitation energies
+    and transition dipole moments.
+    The refractive index :math:`n` is, in general, frequency-dependent. We
+    assume it to be constant and equal to 1.
+    """
+
+    N_A = qcel.constants.get("Avogadro constant")
```
loriab, line 166: It's a little safer to `from psi4.driver import constants` and then use as usual. At the moment there's no difference. But that could change if qcel shifts default to codata2018 and psi stays with 2014. At the driver layer, psi can pull an explicit 2014, even when qcel updates default.

psi4/driver/p4util/spectrum.py
```diff
@@ -0,0 +1,376 @@
...
+    .. math::
+
+        k = \frac{16\pi^{2}N_{\mathrm{A}}}{3\times 1000\times \ln(10) (4 \pi \epsilon_{0}) n \hbar c^{2}}
+
+    The prefactor is computed in SI units and then adjusted for the fact that
+    we use atomic units to express microscopic observables: excitation energies
+    and transition dipole moments.
+    The refractive index :math:`n` is, in general, frequency-dependent. We
+    assume it to be constant and equal to 1.
+    """
```
loriab, line 200: Lovely docstrings throughout. I'm looking into if we can get sphinx to pull types from annotations and skip the duplication in docstring.

psi4/driver/p4util/spectrum.py
```diff
@@ -0,0 +1,376 @@
...
+    au_to_Joule_inverse_Tesla = 2.0 * qcel.constants.get("Bohr magneton") * qcel.constants.conversion_factor("m", "cm")
+    conversion = au_to_Coulomb_centimeter * au_to_Joule_inverse_Tesla
+
+    numerator = 16.0 * np.pi**2 * N_A
+    denominator = 3 * 1000 * np.log(10) * (4 * np.pi * e_0) * hbar * c**2
+
+    return (numerator / denominator) * conversion
+
+
+def spectrum(*,
```
loriab, line 218: I'm a fan of the preventative keyword-only.

tests/pytests/test_tdscf_excitations.py
```diff
@@ -24,1262 +23,218 @@
...
+
+# Canonical restricted system
+h2o = psi4.geometry("""0 1
+O           0.000000    0.000000    0.135446
+H          -0.000000    0.866812   -0.541782
+H          -0.000000   -0.866812   -0.541782
+no_reorient
+no_com
+""",
+                    name="H2O")
```
loriab, line 48: If any test has a `h2o` and optimizes it, I worry about the state of this local variable. May want to play it safe and fixture it.

```
@pytest.fixture
def mols():
    smols = {
        "H": """
0 2
H
""",
        "H2": """
1 2
H
H 1 1.0
""",
        "HeH": """
1 1
He
H 1 1.0
""",
    }

    return {k: psi4.core.Molecule.from_string(v) for k, v in smols.items()}
```

```
def test_weird_basis(subject, bas, ans, mols, request):
    """non-consecutive angular momentum basis set patterns"""

    mol = mols[subject]
```

tests/pytests/test_tdscf_excitations.py
```diff
@@ -24,1262 +23,218 @@
...
+    pytest.param(  h2o, 'RHF-3',     'wB97X',  'RPA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_triplet, RPA]),
+    pytest.param(  h2o, 'RHF-3',     'wB97X',  'TDA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_triplet, TDA]),
+    pytest.param( h2o2, 'RHF-1',     'wB97X',  'RPA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_singlet, RPA]),
+    pytest.param( h2o2, 'RHF-1',     'wB97X',  'TDA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_singlet, TDA]),
+    pytest.param( h2o2, 'RHF-3',     'wB97X',  'RPA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_triplet, RPA]),
+    pytest.param( h2o2, 'RHF-3',     'wB97X',  'TDA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_triplet, TDA]),
+    pytest.param( moxy, 'RHF-1',     'wB97X',  'RPA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_singlet, RPA]),
+    pytest.param( moxy, 'RHF-1',     'wB97X',  'TDA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_singlet, TDA]),
+    pytest.param( moxy, 'RHF-3',     'wB97X',  'RPA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_triplet, RPA]),
+    pytest.param( moxy, 'RHF-3',     'wB97X',  'TDA',  'cc-pvdz', marks=[hyb_gga_lrc, RHF_triplet, TDA]),
```
loriab, line 177: guess these aren't regular enough for a multi-level parametrize?

tests/pytests/test_tdscf_excitations.py
```diff
@@ -24,1262 +23,218 @@
...
-        schmidt_tol=1.0e-12)
-
-    test_vals = out[0]
-    stats = out[-1]
-    assert stats[-1]['done'], "Solver did not converge"
-
-    for i, my_v in enumerate(test_vals):
-        ref_v = exp_energy_sorted[i]
-        assert compare_values(ref_v, my_v, 4, "{}-{}-{}-ROOT-{}".format(ref, func, ptype, i + 1))
+    psi4.core.clean()
```
loriab, line 186: this is called between each test anyways. is it misbehaving that needed again here?

psi4/driver/p4util/solvers.py
dgasmith: `np.einsum(..., optimize=True)`

psi4/driver/p4util/solvers.py
robertodr: * It is similar to the `r_convergence` option in CC, so I agree to rename to `r_convergence` in all TDDFT-related functions.
* User-facing, I agree on `TDSCF_R_CONVERGENCE`. But maybe this is something for #1885 
* `r_convergence` can be set by the user, but we can come up with a pretty safe default: `min(1e-4, d_convergence*1e2)` this way the user has to explicitly tell Psi4 to seek an insanely tight residual convergence on tight `d_convergence`. I am unsure how to achieve this though. Is there a way I can probe defaults from a PsiAPI function?

tests/pytests/test_tdscf_excitations.py
robertodr: I'm unfamiliar with a multi-level parametrize. How does it look?

tests/pytests/test_tdscf_excitations.py
robertodr: :+1:

tests/pytests/test_tdscf_excitations.py
robertodr: copy-paste from previous iteration on these tests.

robertodr:
Anyone has any idea why LGTM is failing on JavaScript? Is it because the reference data for TDDFT in JSON format?

hokru:
Following https://lgtm.com/help/lgtm/javascript-extraction I'd also assume Javascript analysis is triggered by the presence of a *.json file.

robertodr:
Let's see whether my last commit helps...

loriab:
If it’s not rebased lately, can also try that. I’ve asked them to remove js analysis, but it’s hard to tell. Any change would have been late March.

robertodr:
I've rebased in the past couple of days...

loriab:
Fallback soln is rename the double id fields in codemeta.json, I think.

robertodr:
@loriab not sure I follow :thinking:

loriab:
My hypothesis is that the double `@id` is causing the trouble, and maybe the GH needs to be replaced by `identifier`. https://github.com/psi4/psi4/blob/master/codemeta.json#L19-L20

robertodr:
@loriab still no luck with JavaScript...

loriab:
[here's](https://github.com/psi4/psi4/pull/1850/files#diff-bb7d2d2847635701fd31b1a59cddb683R64-R114) an example of a multi-level parametrize.

we'll see if #1905 helps the lgtm analysis

robertodr:
> [here's](https://github.com/psi4/psi4/pull/1850/files#diff-bb7d2d2847635701fd31b1a59cddb683R64-R114) an example of a multi-level parametrize.

mmmm I think we're fine as-is for the moment being...

> we'll see if #1905 helps the lgtm analysis

LGTM doesn't seem to be appeased.

loriab:
I think JS analysis is squashed now, so ready for another try.

No need to clear up history imo.

I think it has to be `identifier`, not `@identifier` since latter not in JSON-LD (unless you're understanding the schema differently). I hadn't seem that you tried the id->identifier, or I wouldn't have starting in on #1905, sorry.

robertodr:
@loriab JavaScript is gone and the builds are green. I'm stashing the converged response vectors in the `Wavefunction` too, this should enable the computation of other transition quantities, if desired.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -534,54 +559,68 @@ def tdscf_excitations(wfn,
...
-            "SYMMETRY": irrep_trans,
-            "SPIN": spin_mult,
+            "EXCITATION ENERGY": x.E_ex_au,
+            "LENGTH-GAUGE ELECTRIC DIPOLE TRANSITION MOMENT": x.edtm_length,
+            "LENGTH-GAUGE OSCILLATOR STRENGTH": x.f_length,
+            "VELOCITY-GAUGE ELECTRIC DIPOLE TRANSITION MOMENT": x.edtm_velocity,
+            "VELOCITY-GAUGE OSCILLATOR STRENGTH": x.f_velocity,
+            "MAGNETIC DIPOLE TRANSITION MOMENT": x.mdtm,
+            "LENGTH-GAUGE ROTATORY STRENGTH": x.R_length,
+            "VELOCITY-GAUGE ROTATORY STRENGTH": x.R_velocity,
```
loriab, line 578: I like the dataclass. What do you think for following cc* for the gauge labels?

```
../psi4/src/psi4/cc/ccresponse/optrot.cc:/*- Process::environment.globals["CCSD SPECIFIC ROTATION (VEL) @ xNM"] -*/
../psi4/src/psi4/cc/ccresponse/optrot.cc:/*- Process::environment.globals["CCSD SPECIFIC ROTATION (LEN) @ xNM"] -*/
```

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -534,54 +559,68 @@ def tdscf_excitations(wfn,
...
         # stash in psivars/wfnvars
         ssuper_name = wfn.functional().name()
-        wfn.set_variable(f"TD-{ssuper_name} ROOT {i+1} TOTAL ENERGY - {irrep_ES} SYMMETRY", E_tot_au)
-        wfn.set_variable(f"TD-{ssuper_name} ROOT 0 -> ROOT {i+1} EXCITATION ENERGY - {irrep_ES} SYMMETRY", E_ex_au)
+        wfn.set_variable(f"TD-{ssuper_name} ROOT {i+1} TOTAL ENERGY - {x.irrep_ES} SYMMETRY", E_tot_au)
+        wfn.set_variable(f"TD-{ssuper_name} ROOT 0 -> ROOT {i+1} EXCITATION ENERGY - {x.irrep_ES} SYMMETRY", x.E_ex_au)
         wfn.set_variable(
-            f"TD-{ssuper_name} ROOT 0 -> ROOT {i+1} LENGTH-GAUGE OSCILLATOR STRENGTH - {irrep_ES} SYMMETRY", f_length)
+            f"TD-{ssuper_name} ROOT 0 -> ROOT {i+1} LENGTH-GAUGE OSCILLATOR STRENGTH - {x.irrep_ES} SYMMETRY",
+            x.f_length)
```
loriab, line 593: are all these in atomic units?

psi4/driver/p4util/solvers.py
loriab: is something like the below what you're after? upper path will only trigger if user has touched the specific tdscf option. lower path d_conv value already reflects general user setting or the driver adjustment for derivative type a la http://psicode.org/psi4manual/master/scf.html#convergence-and-algorithm-defaults (essentially 1.e-6).
```
if core.has_option_changed('SCF', 'TDSCF_R_CONVERGENCE'):
    r_convergence = core.get_option('SCF', 'TDSCF_R_CONVERGENCE')
else:
    r_convergence = min(1.e-4, core.get_option('SCF', 'D_CONVERGENCE') * 1.e2)
```

psi4/driver/procrouting/response/scf_response.py
robertodr: Much better, since there's less to type.

psi4/driver/procrouting/response/scf_response.py
robertodr: Everything is in atomic units. I'll be more specific with that.

robertodr: Now I'm wondering, should the units be part of the variables name? Or is it generally assumed that whatever Psi4 returns is in atomic units?

psi4/driver/p4util/solvers.py
robertodr: I think that's what I am after.

psi4/driver/p4util/spectrum.py
```diff
@@ -0,0 +1,376 @@
...
+from typing import Union, List, Callable, Dict
+from dataclasses import dataclass
+from abc import abstractmethod
+
+import numpy as np
+
+from psi4.driver import constants
+
+
+@dataclass
```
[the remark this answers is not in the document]
robertodr: I knew I was getting ahead of myself here. Where should I look for an example of pydantic use? QCElemental?

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -256,162 +259,375 @@ def _print_output(complete_dict, output):
...
+    if triplets == "also":
+        for nstate, state_sym in state_info:
+            core.print_out(f"      {nstate} singlet states with {state_sym} symmetry\n")
+            core.print_out(f"      {nstate} triplet states with {state_sym} symmetry\n")
+    elif triplets == "only":
+        for nstate, state_sym in state_info:
+            core.print_out(f"      {nstate} triplet states with {state_sym} symmetry\n")
+    elif triplets == "none":
+        for nstate, state_sym in state_info:
+            core.print_out(f"      {nstate} singlet states with {state_sym} symmetry\n")
```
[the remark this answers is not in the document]
robertodr: No strong opinions, but the certain program's behavior is certainly more familiar to the public. My only concern if how to do the 50-50 split when the user asks for `[3, 3, 3, 3]` roots (in C2v)

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -256,162 +259,375 @@ def _print_output(complete_dict, output):
...
+    if guess != "denominators":
+        raise ValidationError(f"Guess type {guess} is not valid")
 
-        #TODO: psivars/wfnvars
+    # TODO Switch to this form when Jeff's PR is merged
+    #if core.has_option_changed('SCF', 'TDSCF_R_CONVERGENCE'):
+    #    r_convergence = core.get_option('SCF', 'TDSCF_R_CONVERGENCE')
+    #else:
+    #    r_convergence = min(1.e-4, core.get_option('SCF', 'D_CONVERGENCE') * 1.e2)
+    r_convergence = min(1.e-4, core.get_option('SCF', 'D_CONVERGENCE') * 1.e2)
```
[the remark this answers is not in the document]
robertodr: Not sure I follow here? Maybe we should postpone until Jeff's PR is in.

psi4/driver/p4util/spectrum.py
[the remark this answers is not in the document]
loriab: pydantic may have a drop-in replacement for `dataclass` https://pydantic-docs.helpmanual.io/usage/dataclasses/

psi4/driver/procrouting/response/scf_response.py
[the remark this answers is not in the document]
loriab: maybe favor singlets in each irrep split so it ends up as singlets 2,2,2,2 and triplets 1,1,1,1 and put a note in the nstates docstring?

psi4/driver/procrouting/response/scf_response.py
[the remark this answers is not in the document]
loriab: Jeff's interface PR is passing. should we just merge that as-is and patch the light remaining issues here?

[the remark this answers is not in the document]
robertodr: Yes, let's merge Jeff's first and then clean this up.

robertodr:
I've rebased on top of Jeff's (#1885) changes. I need to implement the handling of triplets _à la_ famous program. Is there a way to have the `TDSCF_STATES` option be either an array or an integer? It would be good to move the logic for apportioning states per irrep (and triplets) further up from `tdscf_excitations`, I think.

psi4/driver/procrouting/response/scf_response.py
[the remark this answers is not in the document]
robertodr: Done with the last commit.

robertodr:
This is ready for more reviews while I write documentation for it.

doc/sphinxman/source/tdscf.rst
```diff
@@ -0,0 +1,988 @@
+.. #
+.. # @BEGIN LICENSE
+.. #
+.. # Psi4: an open-source quantum chemistry software package
+.. #
+.. # Copyright (c) 2007-2019 The Psi4 Developers.
```
andysim, line 6:
```suggestion
.. # Copyright (c) 2007-2020 The Psi4 Developers.
```

psi4/driver/p4util/spectrum.py
```diff
@@ -0,0 +1,376 @@
+#
+# @BEGIN LICENSE
+#
+# Psi4: an open-source quantum chemistry software package
+#
+# Copyright (c) 2007-2019 The Psi4 Developers.
```
andysim, line 6:
```suggestion
# Copyright (c) 2007-2020 The Psi4 Developers.
```

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -256,164 +262,398 @@ def _print_output(complete_dict, output):
...
-                   '         {:^57}\n'.format('TDSCF excitation energies') +
-                   '         {:^57}\n'.format('by Andrew M. James and Daniel G. A. Smith') +
-                   '         ---------------------------------------------------------\n')
-
-    core.print_out("\n  ==> Requested Excitations <==\n\n")
-    state_info = options.pop('states')
-    for nstate, state_sym in state_info:
-        core.print_out("      {} states with {} symmetry\n".format(nstate, state_sym))
+    core.print_out("\n\n         ---------------------------------------------------------\n"
+                   f"         {'TDSCF excitation energies':^57}\n" +
```
andysim, line 266: Just checking, are those ^57 characters supposed to be there?

andysim:
So far I think this looks great - type safe python with a lot of testing, and a great modular design for the solvers.  I really like the line broadening plot generation.  If there's some kind of provenance info that can be added to the json reference data file that would be handy.

psi4/driver/procrouting/response/scf_response.py
robertodr: They make sure both lines are centered in the box.

andysim: Ah, I see that now in my [go-to source](https://pyformat.info/#number_padding) for all things pyformat.  Thanks!

robertodr:
> So far I think this looks great - type safe python with a lot of testing, and a great modular design for the solvers. 

All AJ's work!

> I really like the line broadening plot generation. 

Thanks, it took me an embarrassingly long amount of time to check that units of measure all made sense...

> If there's some kind of provenance info that can be added to the json reference data file that would be handy.

I have the raw G09 input and output files, together with the notebook that extracts the data from the formatted checkpoint files. Should I include that in the `pytests` folder?

@loriab I'm seeing some funny behavior (or most likely I don't understand what the normal behavior actually is)
```
molecule {
0 1
O        0.000000    0.695000   -0.092486
O       -0.000000   -0.695000   -0.092486
H       -0.388142    0.895249    0.739888
H        0.388142   -0.895249    0.739888
}

set {
tdscf_states 10
scf_type pk
}

energy('td-scf/cc-pvdz')
```
I'd expect this to converge the wavefunction to `D_CONVERGENCE` 1e-6 and then compute the 10 states to `TDSCF_R_CONVERGENCE` 1e-4 The thresholds seems however to be 1e-8 (and 1e-6) respectively. I can't figure out where `D_CONVERGENCE` gets tightened from what stated in the [table](http://psicode.org/psi4manual/master/scf.html#convergence-and-algorithm-defaults)

loriab:
I've lost your `D_CONVERGENCE` comment, so answering here. Looks like expanding the functions that classify to SCF [here](https://github.com/psi4/psi4/blob/master/psi4/driver/driver_util.py#L96) will get the right 1.e-6 value.

robertodr:
> I've lost your `D_CONVERGENCE` comment, so answering here. Looks like expanding the functions that classify to SCF [here](https://github.com/psi4/psi4/blob/master/psi4/driver/driver_util.py#L96) will get the right 1.e-6 value.

Thanks for the pointer. Now fixed with the following contraption:
```python
        if (procedures['energy'][method_name] == proc.run_scf
                or procedures['energy'][method_name] == proc.run_tdscf_energy):
```
so that TDSCF closely follows SCF. Probably there's a better way?

I'm also slightly puzzled at the following. TDSCF will fail to converge for non totally symmetric roots when running with symmetry _and_ density-fitting (which is by and large the default behavior for small-ish molecules, I believe) I thought this was a fluke for one particular system I was testing, but it does indeed show up consistently for irrep >= 1. The solver will either stall and die exceeding the maximum number of iterations (the subspace never grows after a certain, early iteration) or it will (falsely, I believe) identify a negative eigenvalue in the subspace representation of the Hessian (instability) and consequently fail.
Is there something in JK that I am misusing? @andysim @dgasmith

andysim:
The contraption you show above could also be written as `if procedures['energy'][method_name] in [proc.run_scf, proc.run_tdscf_energy]:`.  Regarding the nirrep>1 problem, it sounds like the backtransformation is either not numerically sound or is not being applied correctly.  I didn't see anything wrong in the first pass through but I'll look again.

robertodr:
> The contraption you show above could also be written as `if procedures['energy'][method_name] in [proc.run_scf, proc.run_tdscf_energy]:`. 

Yep, much cleaner syntax.

> Regarding the nirrep>1 problem, it sounds like the backtransformation is either not numerically sound or is not being applied correctly. I didn't see anything wrong in the first pass through but I'll look again.

The code should be in `twoel_Hx` (and deeper) which was not touched in this PR.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -255,165 +261,398 @@ def _print_output(complete_dict, output):
...
+            raise TDSCFConvergenceError(maxiter, wfn, f"singlet excitations in irrep {irrep_ES}", ret["stats"][-1])
+
+        # flatten dictionary: helps with sorting by energy
+        # also append state symmetry to return value
+        for e, (R, L) in zip(ret["eigvals"], ret["eigvecs"]):
+            irrep_trans = wfn.molecule().irrep_labels()[engine.G_gs ^ state_sym]
+
+            # length-gauge electric dipole transition moment
+            edtm_length = engine.residue(R, mints.so_dipole())
+            # lenght-gauge oscillator strength
```
loriab, line 355: lenght typo

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -255,165 +261,398 @@ def _print_output(complete_dict, output):
...
+    for i in range(states % nirrep):
+        spi[i] += 1
+
+    return spi
+
+
+def _validate_tdscf(*, wfn, states, triplets, guess) -> None:
+
+    # validate states
+    if not (isinstance(states, int) or isinstance(states, list)):
```
loriab, line 390: `isinstance(states, (int, list))`

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -255,165 +261,398 @@ def _print_output(complete_dict, output):
...
+    _validate_tdscf(wfn=wfn, states=states, triplets=triplets, guess=guess)
 
-    #TODO: check/handle convergence failures
+    restricted = wfn.same_a_b_orbs()
+
+    # determine how many states per irrep to seek and apportion them between singlets/triplets and irreps.
+    singlets_per_irrep = []
+    triplets_per_irrep = []
+    if isinstance(states, list):
+        if triplets == "ONLY":
```
loriab, line 524: `triplets` has been sanity-checked for case? rhs is going to be lowercase if it passes through `energy(..., **kwargs)`.

tests/pytests/test_tdscf_excitations.py
```diff
@@ -26,1260 +25,219 @@
...
-    for i, my_v in enumerate(test_vals):
-        ref_v = exp_energy_sorted[i]
-        assert compare_values(ref_v, my_v, 4, "{}-{}-{}-ROOT-{}".format(ref, func, ptype, i + 1))
+        pytest.xfail("UKS Vx kernel bug for non-LDA")
+
+    molecule = molecules[mol]
+    psi4.set_options({'scf_type': 'pk', 'e_convergence': 8, 'd_convergence': 8, 'save_jk': True})
+    if ref == "UHF":
+        psi4.set_options({'reference': 'UHF'})
+    molecule.reset_point_group('c1')
```
loriab, line 193: goal is to remove this line when backtransformation issue resolved?

doc/sphinxman/source/tdscf.rst
```diff
@@ -0,0 +1,1024 @@
...
+Psithon usage
+~~~~~~~~~~~~~
+
+PsiAPI usage
+~~~~~~~~~~~~
+
+Plotting one-photon absorption and electronic circular dichroism spectra
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
```
loriab, line 89: all below is placeholder, I think -- this file was easier to review than I thought :-)

psi4/driver/procrouting/response/scf_response.py
robertodr: I really can't write this word without making a mistake...

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -255,165 +261,398 @@ def _print_output(complete_dict, output):
...
     # construct the engine
-    if restricted:
-        engine = TDRSCFEngine(wfn, triplet=triplet, ptype=ptype)
-    else:
-        engine = TDUSCFEngine(wfn, ptype=ptype)
-
-    # just energies for now
-    solver_results = []
+    engine = TDRSCFEngine(wfn, ptype=ptype.lower(), triplet=True if spin_mult == "triplet" else
+                          False) if restricted else TDUSCFEngine(wfn, ptype=ptype.lower())
```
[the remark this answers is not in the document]
robertodr: I tend to get carried away by one-liners...

doc/sphinxman/source/tdscf.rst
robertodr: Yes, placeholder. I started writing when I noticed that the example I chose (H2O2 with symmetry and DF) didn't actually run...

tests/pytests/test_tdscf_excitations.py
robertodr: The tests are all run with PK, so that can be removed. AJ had left a comment saying that running with symmetry was "slow" and I just kept the tests as they were.

robertodr: Just tried removing that and there's two problems:
1. The reference values were obtained without symmetry, so the ordering of states will differ. I can regenerate this data easily in a few days. However, this requires seeking 8 states at least (to make sure we cover all irreps reasonably well) and the tests become accordingly longer.
2. UHF+PK+Symmetry lead to convergence failures too. I thought I had tested this but apparently not...

robertodr: Issue no. 2 is actually a too tight `r_convergence`: the solver stalls for the first irrep already. Not sure how to solve it...

robertodr:
This is ready for (hopefully!) final review and merge. I've documented the currently known issues.

psi4/driver/p4util/exceptions.py
```diff
@@ -148,6 +150,44 @@ def __init__(self, eqn_description, iteration, wfn, e_conv, d_conv):
...
+        for e, diff, r_norm in zip(stats["val"], stats["delta_val"], stats["res_norm"]):
+            conv_info += f"      {e:.6f}         {diff:-11.5e}    {r_norm:12.5e}\n"
+        ConvergenceError.__init__(self,
+                                  eqn_description=f"""TDSCF solver ({what})""",
+                                  iteration=iteration,
+                                  additional_info=conv_info)
+        self.wfn = wfn
+        self.what = what
+        self.stats = stats
+
```
andysim, line 189: This is awesome!  Great to have a clear picture of how things stand and how far from convergence the calculation is

doc/sphinxman/source/tdscf.rst
```diff
@@ -0,0 +1,342 @@
...
+Known limitations
+~~~~~~~~~~~~~~~~~
+
+.. warning:: The implementation cannot currently handle the following cases:
+             - The use of symmetry with density-fitted two-electron integrals.
+             - Excited states of triplet symmetry from a restricted DFT reference.
+             - Excited states from an unrestricted reference other than HF or LDA.
+
+.. warning:: The length-gauge rotatory strengths |PSIfour| computes are
+             currently **not** gauge-origin invariant.
```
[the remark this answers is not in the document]
robertodr: It is for SCF when using London orbitals.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -255,165 +261,411 @@ def _print_output(complete_dict, output):
...
+        ptype = "TDA"
+        solve_function = lambda e, n, g, m: solvers.davidson_solver(engine=e,
+                                                                    nroot=n,
+                                                                    guess=g,
+                                                                    r_convergence=r_convergence,
+                                                                    max_ss_size=max_vecs_per_root * n,
+                                                                    verbose=verbose)
+    else:
+        ptype = "RPA"
+        solve_function = lambda e, n, g, m: solvers.hamiltonian_solver(engine=e,
```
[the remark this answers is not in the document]
robertodr: I went for two local functions rather than two lambdas. It might make the code slightly more readable.

PR #1821, opened 2020-02-26
exposes useful functions from detci and libpsio classes to python

SinaMostafanejad:
## Description
This PR aims to expose useful functions from _detci_ and _libpsio_ to the python side:
- libpsio needs **read()**, **write()**, **read_entry()**, and **write_entry()** functions available in python to be actually useful to the user who intends to use Psi4 intermediate files, containers and buffers.
- There are publicly accessible _get_opdm()_ and _get_tpdm()_ functions in _detci_ but for the active-space only. This PR tries to bring the **opdm_add_inactive()** function to the public scope as well and expose it to the python side.
- Members of the **psio_address** structure are also exposed that can be adopted as arguments in _libsio_'s _read()/write()_ function calls.

## Todos

*  Having a spatial symmetry-imposed bock-structured sparse TPDM as a result of get_tpdm() instead of the dense format only is desirable.
* Adding _opdm_add_inactive()_'s counterpart in order to complete the full structure of TPDM [such as **tpdm_add_inactive()**] will be extremely useful.

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/export_psio.cc
```diff
@@ -56,6 +62,12 @@ void export_psio(py::module &m) {
         .def("tocscan", &PSIO::tocscan,
              "Seek string in binary file. This export is only good for catching None, as returned success object not "
              "exported.")
+	.def("read", &PSIO::read, "Reads data from within a TOC entry from a PSI file", "unit"_a,
+	     "key"_a, "buffer"_a, "size"_a, "start"_a, "end"_a)
+	.def("write", &PSIO::write, "Writes data to a TOC entry in a PSI file", "unit"_a,
+	     "key"_a, "buffer"_a, "size"_a, "start"_a, "end"_a)
+	.def("read_entry", &PSIO::read_entry, "docstring", "unit"_a, "key"_a, "buffer"_a, "size"_a)
+	.def("write_entry", &PSIO::write_entry, "docstring", "unit"_a, "key"_a, "buffer"_a, "size"_a)
```
dgasmith: Can you add a test for this? It would be very useful just to demonstrate how this would work to users as well.

SinaMostafanejad: Absolutely. I am working on it to show its use in the context of combining Psi4's CASSCF with OpenRDM's MCPDFT module. Once I get it done, I will add a test to its corresponding directory in Psi4.

SinaMostafanejad: I might also design a much simpler test specifically for PSIO class as well. I need to think more about it.

Thanks for the suggestion.

loriab:
Also, you'll want to edit the top of your PR so that you're trying to merge into `psi4:master` not `psi4:1.3.x`.

SinaMostafanejad:
> Also, you'll want to edit the top of your PR so that you're trying to merge into `psi4:master` not `psi4:1.3.x`.

I think that would lead to conflicts in many basic files which are not there in the master branch anymore (such as .travis.yml) or are duplicated by several branches. This was the most relevant and cleanest branch that I could merge with. I did it so that you guys could possibly look into it before merging it into master yourselves, if that is possible.

PeterKraus:
It'll be a lot easier for us if you'll rebase from `master` as opposed to `1.3.2` ...

SinaMostafanejad:
> It'll be a lot easier for us if you'll rebase from `master` as opposed to `1.3.2` ...

 have changed the base to psi4:master as you and Lori suggested.

loriab:
Thanks for switching the base from `1.3.x` to `master`. Unfortunately, it didn't go as cleanly as we need. The thing to do is on your clone, make sure you have `origin` and `upstream` remotes at `git remote -v`. Then checkout the upstream master branch, check that its hash matches the upstream psi4 master (that is, the hash next to "Latest commit" right off the https://github.com/psi4/psi4/ page). Then checkout a new branch, e.g., `git checkout -b libpsioexport`. Then apply your one commit, `git cherry-pick 9788333289e4c48d57a9bf7e453fb8855fbc8ae9`. Then push that new branch to the branch associated with this PR. It will have to be force-pushed b/c overwriting history, `git push origin libpsioexport:psi4-dev --force`. Then this PR should have only one commit in it, yours, and be with respect to the development, master branch.

SinaMostafanejad:
> Thanks for switching the base from `1.3.x` to `master`. Unfortunately, it didn't go as cleanly as we need. The thing to do is on your clone, make sure you have `origin` and `upstream` remotes at `git remote -v`. Then checkout the upstream master branch, check that its hash matches the upstream psi4 master (that is, the hash next to "Latest commit" right off the https://github.com/psi4/psi4/ page). Then checkout a new branch, e.g., `git checkout -b libpsioexport`. Then apply your one commit, `git cherry-pick 9788333289e4c48d57a9bf7e453fb8855fbc8ae9`. Then push that new branch to the branch associated with this PR. It will have to be force-pushed b/c overwriting history, `git push origin libpsioexport:psi4-dev --force`. Then this PR should have only one commit in it, yours, and be with respect to the development, master branch.

Done.

psi4/src/export_psio.cc
SinaMostafanejad: Dear Daniel,

I have prepared a very simple test case for the exposed read/write libpsio functions. I could not attach the input file here for some reason. Please find the input file at

https://github.com/SinaMostafanejad/v2rdm_casscf/blob/OpenRDM-interface/libpsio_test.dat

The result however shows that the exposed functions do not work the way they are expected to, possibly because some of the function arguments (such as **char * buffer**) on the C++ side need to be pointers to characters or strings while on the python side, they become immutable objects (strings) and cannot be changed when we try to write in them as buffers.

I am not sure if I am correct in my interpretation of the test result or doing something wrong, but could you please take a look into it and let me know what do you think about the problem?

Thanks. 

Do you have any simple way around this?

dgasmith: I don't have anything specific off the top of my head for this, but this kind of issue is one of the main reasons I encouraged a test. See [here](https://pybind11.readthedocs.io/en/stable/advanced/cast/strings.html) for more information, see if they have anything on this.

SinaMostafanejad: I agree with you that any new functionality should be accompanied by a test case to show its use. However, this is libpsio's function that is being exposed to python and therefore, there has been no similar test about it in Psi4 last time i checked. So, I think based on your comment, adding a test would be really useful and I will update you guys on that if I find a solution to this problem.

mattwelborn: @loriab Do you understand why this test does not work? (https://github.com/SinaMostafanejad/v2rdm_casscf/blob/OpenRDM-interface/libpsio_test.dat)

SinaMostafanejad: I have hijacked @edeprince3's idea and tried to get around the problem of immutability of the strings in python for these read/write() functions in libpsio by wrapping libpsio's functions inside another which does not use **char *** for buffer arguments

`void PSIO::py_read(size_t unit, const char *key, double* buffer, size_t size, psio_address start, psio_address end) {
    read(unit, key, (char *) buffer, size, start, (psio_address *) &end);
}
`
or 

`void PSIO::py_read(size_t unit, const char *key, double buffer, size_t size, psio_address start, psio_address end) {
    read(unit, key, (char *) &buffer, size, start, (psio_address *) &end);
}
`

It does not solve the problem after exposing these new functions to python. So, the issue might be something else.

SinaMostafanejad: I can confirm that using **ctypes** module and **ctypes.create_string_buffer()**  and other functions from this module in the above test, results in the same response from libpsio read()/write() functions by not updating the buffer values.

Codes like

```
opdm_c = c_char_p(b'120')
psioh.write_entry(PSIF_V2RDM_D1A, "D1a", repr(byref(opdm_c)), sizeof(opdm_c))

opdm2_c = create_string_buffer(sizeof(opdm_c))
psioh.read_entry(PSIF_V2RDM_D1A, "D1a", repr(opdm2_c), sizeof(opdm2_c))
```
or similar variations like

```
opdm_c = c_char_p(b'120')
psioh.write_entry(PSIF_V2RDM_D1A, "D1a", string_at(addressof(opdm_c)), sizeof(opdm_c))

opdm2_c = create_string_buffer(sizeof(opdm_c))
psioh.read_entry(PSIF_V2RDM_D1A, "D1a", string_at(addressof(opdm2_c)), sizeof(opdm2_c))
```

would not work.

dgasmith:
@SinaMostafanejad Were you able to get this to work? I think the only way in which we could do this would be to write a wrapper code around these functions (similar to the __array_interface__ on Matrix objects) so that we can correctly return the python values instead of updating them in place.

SinaMostafanejad:
> @SinaMostafanejad Were you able to get this to work? I think the only way in which we could do this would be to write a wrapper code around these functions (similar to the **array_interface** on Matrix objects) so that we can correctly return the python values instead of updating them in place.

Hi @dgasmith,

Unfortunately not. I had a lot in my plate last month because of my dissertation defense.

I agree with you about this possibility as I have already discussed it with @mattwelborn and Sam although I am not aware of specifics you are referring to about the array_interface in Psi4.

Since C extension in python is out of my comfort zone in programming, it might take me some time to work it out considering a few other tasks w/ priorities on my list, in case if you want me to do that. If so, please point me to the files and directions in Psi4 that I need to look into before I begin. Otherwise, please tell me how can I be of any help.

Thank you.

PR #1825, opened 2020-03-02
main channel adcc

loriab:
## Description
main channel adcc uses MKL and subchannel is gone. try to fix CI.
## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -96,7 +96,7 @@ jobs:
         qcengine \
         pybind11 \
         -c conda-forge
-      conda install adcc -c adcc/label/dev_mkl -c conda-forge
+      conda install adcc -c adcc -c conda-forge
       conda list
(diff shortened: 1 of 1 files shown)


mfherbst:
Yes sorry about that ... in the build it downloads an old adcc version from github instead of recognising the installed version from conda it seems. I'll take a look.

loriab:
@mfherbst, looks like the adcc of main is working nicely. If you could build py3.6, that should fix the remaining CI lane. And building py3.8 would future-proof the lineup. Thanks!

mfherbst:
Yes it's py 3.6. We do not have this at the moment. I made an amendment in #1827 to disable adcc testes if pyversion != 3.7. We'll look into producing builds for 3.6 and 3.8 as well.

mfherbst:
Py 3.6 is now available on anaconda as well, see #1827.

loriab:
closed in favor of #1827

PR #1390, opened 2018-12-02
FNO support in DFOCC (take 2)

loriab:
## Description
Replaces #824: Frozen Natural Orbitals were implemented for DF-CCSD and DF-CCSD(T) energies.

## Todos
- [x] FNOs are implemented for MP2.5, MP3, LCCD, CCD, CCSD, CCSD(T), and CCSD(AT).
- [x] Both DF and CD options are available (only for RHF ref for now).

## Checklist
- [ ] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [ ] Ready for review
- [ ] Ready for merge

PR #1829, opened 2020-03-04
Debug and Quarantine HF MO Hessian: OCC's Open-Heart Surgery V

JonathonMisiewicz:
## Description
Think of the default orbital optimization algorithm in OCC with the following pseudocode:
```
orbital_gradient = compute_orbital_gradient()
try:
    orbital_step = compute_NR_step_via_PCG_with_HF_MO_hessian(orbital_gradient)
except ConvergenceError:
    orbital_step = compute_step_with_crude_diagonal_MO_hessian(orbital_gradient)
diis_step()
```

I traced a test failure in an upcoming PR (ocepa-grad2) to a failure in the approximate NR step. Five failures, in fact.
1. The HF MO Hessian couples alpha and beta IDPs. You need a single alpha, not a different alpha for each spin.
2. For the same reason, you need a single beta.
3. The Fock term assumed semicanonical orbitals. OCC does not use semicanonical orbitals.
4. The Fock matrix was acting on the wrong vector.
5. One of the tensor contractions involving TEI was incorrect.

After fixing those five bugs, I observed three things:
1. After accounting for whether you index by (v,o) or (o, v), the matrix-vector products of the new occ implementation match those of the SOUHF P4N tutorials,
2. The X (kappa) converged to does satisfy Ax=b
3. The orbital equations refuse to converge

My conclusion is that the implementation of the algorithm is now correct, but it's not a good way to generate orbital steps. The off-diagonal terms can end up being significant for the open-shell systems that would lead you to use an orbital-optimized method. This didn't cause obvious problems before because the previous errors rendered the PCG procedure non-convergent, so you would always fall back to the crude diagonal step _unless_ your orbital gradient was so low that you didn't need to iterate at all. So what we expect to see from this bug is difficulties converging occ tightly, which has been my experience.

To remedy the situation, the default orbital step is now the crude semicanonical step.

The next PR should be the final one in the series. I'll just need to rebase it onto this one, repair the two tests I mentioned on broken tests, and confirm the test suite is clean.

Pinging the usual suspects: @loriab, especially @dgasmith for PCG, and @jturney. I think Jet is sick today, so a different third reviewer would be welcome.

## Todos
- [x] Debugged HF MO hessian based orbital step
- [x] Made HF MO hessian based orbital step opt-in rather than default; Not recommended

## Checklist
Can't run tests if the tests were already broken!
- [x] Confirmed b-Ax approximately zero for the first step of the `ocepa-grad2` test case
- [x] Confirmed Ax reproduces the Psi4Numpy reference SOUHF Ax by acting on a test alpha orbital and a test beta orbital

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/occ/kappa_orb_resp_iter.cc
+++ b/psi4/src/psi4/occ/kappa_orb_resp_iter.cc
@@ -64,10 +64,10 @@ void OCCWave::kappa_orb_resp_iter() {
         global_dpd_->buf4_sort(&K, PSIF_LIBTRANS_DPD, qpsr, ID("[V,O]"), ID("[V,O]"), "MO Ints (VO|VO)");
         global_dpd_->buf4_close(&K);
 
-        // (ai|bj) -> (aj|bi)
-        global_dpd_->buf4_init(&K, PSIF_LIBTRANS_DPD, 0, ID("[V,O]"), ID("[V,O]"), ID("[V,O]"), ID("[V,O]"), 0,
-                               "MO Ints (VO|VO)");
-        global_dpd_->buf4_sort(&K, PSIF_LIBTRANS_DPD, psrq, ID("[V,O]"), ID("[V,O]"), "MO Ints (aj|bi)");
+        // <OO|VV> -> <OV|VO>
+        global_dpd_->buf4_init(&K, PSIF_LIBTRANS_DPD, 0, ID("[O,O]"), ID("[V,V]"), ID("[O,O]"), ID("[V,V]"), 0,
+                               "MO Ints <OO|VV>");
+        global_dpd_->buf4_sort(&K, PSIF_LIBTRANS_DPD, psrq, ID("[O,V]"), ID("[V,O]"), "MO Ints <OV|VO>");
         global_dpd_->buf4_close(&K);
 
         // <OV|OV> -> <VO|VO>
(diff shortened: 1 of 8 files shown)


psi4/src/psi4/occ/kappa_orb_resp_iter.cc
```diff
@@ -854,29 +509,189 @@ void OCCWave::orb_resp_pcg_uhf() {
...
+    if (reference_ == "RESTRICTED") {
         // Build sigma = A * p
-        // sigma_AI = 4 \sum_{BJ} (AI|BJ) P_BJ
+
+        // sigma_AI = 2 \sum_{B} F_AB P_BJ
         global_dpd_->file2_init(&S, PSIF_OCC_DPD, 0, ID('V'), ID('O'), "Sigma <V|O>");
         global_dpd_->file2_init(&P, PSIF_OCC_DPD, 0, ID('V'), ID('O'), "P <V|O>");
+        // CAUTION! "Fock" libdpd entries are the Fock elements, as you expect, defined in
+        // fock_[alpha/beta]a. "F" is defined in trans_ints_[r/u]hf and is off-diagonal Fock elements.
+        // We need all Fock elements, so we use "Fock", not "F".
```
loriab: the text caution speaks of "Fock" vs "F", but all code uses of the former are gone and new code is "FD" and "F". just checking that the caution is in sync.

JonathonMisiewicz:
`dfocc` was better off. `dfocc` doesn't have problems 1, 2, or 4. Not sure about 5 because I'm rusty on converting DF integrals to conventional, but if it did have that problem, it would show up in the gradients. 3 is a problem in `dfocc`. I can deal with it if you'd like, but I assumed nobody would use that algorithm for orbital optimization, and it wasn't worth adding to the #1390 nightmare.

Good catch on the comment. I'll update that.

susilehtola:
Approximating the Hessian with the semicanonical rule H_{ia,ia} ~ e_{a} - e{i} probably makes the most sense, since the CC hessian is anyway not going to be the same as the HF one due to the correlation effects. I'm not sure if computing the exact diagonal orbital Hessian is tractable, but it might make the convergence faster.

JonathonMisiewicz:
I have the expanded OMP2/OCEPA hessian formulas in a Jupyter notebook, and I know they're implemented in DFOCC. Seeing how that choice affects convergence would be an interesting test some time. (Well, assuming DFOCC orbital convergence gets straightened out...)

PR #1847, opened 2020-03-10
Fix bug in CASSCF semi-canonicalization

fevangelista:
## Description
Fixes a bug in the semicanonicalization procedure cause by an uninitialized matrix.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/detci/ciwave.cc
```diff
@@ -747,6 +747,8 @@ void CIWavefunction::semicanonical_orbs() {
 
     // Allocate unitary transformation (only for DOCC + ACTV + VIR orbs)
     auto U = std::make_shared<Matrix>("U to semi", nrotpi, nrotpi);
+    // zero U otherwise we get small errors
+    U->zero();
```
dgasmith: Can we extend one of the tests to ensure the the root value and final value match exactly?

fevangelista: Ok, I added the case that fails and checked for the energy.

fevangelista:
@dgasmith: The problems was with modifying the orbitals. In the old code I created a unique unitary U and used it to rotate the space "ROT", but that caused the energy to differ. In this solution I rotate each space separately ("DOCC", "ACT", "VIR"). However, I am not sure why "ROT" did not work. How is this space defined? Any idea?

dgasmith:
ROT is defined by DOCC + ACT + VIR removed (inactive docc/inactive air). Perhaps rotating all three simultaneously causes some accidental mixing?

susilehtola:
> ROT is defined by DOCC + ACT + VIR removed (inactive docc/inactive air). Perhaps rotating all three simultaneously causes some accidental mixing?

This would sound extremely weird: if the rotation parameter matrix is block-diagonal, the result of the exponentiation should be the same as a block-diagonal rotation matrix?

dgasmith:
You're right, I was thinking of this as the canonical rotation problem between valid pairs which forces mixing by choice while this is the block diagonal form. The previous code should work as advertised. Any properties of the Pade approximant which would make this invalid? Doesn't seem like it, but I have been out of the loop on this kind of tech for awhile.

fevangelista:
I am not really sure the Pade matters. However I am still a bit confused by the wording: "ROT is defined by DOCC + ACT + VIR removed", what is removed? I have not tried to compare the ROT vs individual blocks of DOCC + ACT + VIR, maybe I should do that to figure out what's wrong. The PR is now failing on a molden test (which apparently I contributed to :) but I am wondering if such test is meaningful. After all the orbital phase is arbitrary so the test could fail without anything being wrong.

dgasmith:
I think in other orbital test we attempted to take magnitudes or norms to remove these issues.

ROT is DOCC + ACT + VIR with the inactive pieces removed. I can see how the previous message was confusing.

fevangelista:
@dgasmith: I realized that there was a typo in the fix that I suggested and when I corrected it the issue persisted. I think the current (and the previous) procedures were correct and that's because if I take the MOs from psi4's CASSCF and run FCI in forte I get the correct energy and the orbitals are semi-canonical. So I suspect the problem is in this snippet of code where we compute the energy after semi-canonicalization:
https://github.com/psi4/psi4/blob/1662c5d7669e80eb32864e62e73df41d929621eb/psi4/driver/procrouting/mcscf/mcscf_solver.py#L358-L372
Am I missing some step here?

dgasmith:
No, I think that is all that is required. Could be rotating far enough that we are finding a new minima in the CI steps?

PR #1854, opened 2020-03-30
Adapt PSIMRCC to compute_energy standard

JonathonMisiewicz:
## Description
I need to give `psimrcc` a wavefunction for the upcoming `libtrans` refactor. Step one is to take the existing `compute_energy`-like functions and make them conform to the normal `Wavefunction::compute_energy` API. No arguments. Returns a double.

A follow-up PR will be needed to actually give `psimrcc` a wavefunction. I'm hopeful this will not be another long series like the `occ` one.

I don't care whether this is in 1.4 or not.

Pinging @fevangelista and @andysim, the two people who have some chance of remembering how this code works.

## Todos
- [x] Conform `psimrcc`'s myriad `Wavefunction::compute_energy`-like functions to the `Wavefunction` interface, as prep for the next PR.
- [x] Removed unpublished "perturbative CBS" capabilities from `psimrcc` and adds appropriate warnings.

## Questions
- [x] @fevangelista, can you explain what the "perturbative CBS" is? Those should be added to the current energy somehow, but I can't tell which of the three flavors to add, or even what this is.

## Checklist
- [x] `ctest -R psimrcc` passes

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/psimrcc/mp2_ccsd.cc
```diff
@@ -152,10 +152,14 @@ void MP2_CCSD::compute_mp2_ccsd_energy() {
     outfile->Printf("\n\n   * MP2-CCSD total energy = %25.15f\n", current_energy);
 
     compute_mp2_ccsd_components();
+
+    ref_wfn_->set_scalar_variable("CURRENT ENERGY", current_energy);
+    ref_wfn_->set_scalar_variable("MP2-CCSD TOTAL ENERGY", current_energy);
+    return ref_wfn_->scalar_variable("CURRENT ENERGY");
```
loriab: For anyone worried about this using qcvars instead of member data, be assured that the `CURRENT`s have been tied into the fundamental e/g/h_ quantities: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.cc#L1262-L1273

JonathonMisiewicz: Except for where `perturb_cbs` is concerned, but I'll hopefully be fixing that.

fevangelista:
@JonathonMisiewicz: I didn't remember the perturbative CBS was in psi4. This code was part of an old project that was never published. It's all about single-reference CCSD. I think the code has never been used and so it can be removed.

JonathonMisiewicz:
Thanks for the response! This PR now removes perturbative CBS entirely from Psi.

loriab:
switch the two options to warnings, perhaps. https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L2668-L2669

JonathonMisiewicz:
Good catch! Now I need to know if this is a 1.4 or 1.5 target so I can write the warnings.

loriab:
Make it a deprecated 1.4/ will fail 1.5, I think.

JonathonMisiewicz:
Done.

psi4/src/psi4/psimrcc/mp2_ccsd.cc
loriab: You definitely mean `ref_wfn_` here, not the wfn itself? Usually the reference wfn wouldn't have CC quality. I can believe psimrcc is the exception, so just checking.

JonathonMisiewicz: `psimrcc` doesn't have a wavefunction yet! The next PR should give it one, and all genuinely `psimrcc` quantities (like this one) will then go on `psimrcc`'s wavefunction.

JonathonMisiewicz:
> @JonathonMisiewicz: Thank you for making these changes! All looks good. Where is `updater_` initialized now?

In the constructors of the relevant `CCManyBody` subclass.

IDMRPT2:
https://github.com/psi4/psi4/pull/1854/files#diff-71fa3415381b67d965d3a4528925ce08R52

CCMRCC:
https://github.com/psi4/psi4/pull/1854/files#diff-73facf13685af2d1bd4f7796befb33e6R77-R80

PR #1860, opened 2020-04-07
Enables LibXC density screening

hokru:
## Description
Enables screening of small densities using LibXC's built-in functionality. Enables user to set custom values. 
This fixes, e.g., issues with calculating the H-atom with many functionals.

## Todos
- [x] introduces `xc_func_set_dens_threshold`
- [x] `DFT_DENSITY_TOLERANCE` as a (expert) user option
- [x] related PsiAPI options for Functional/LibXCFunctional/Superfunctional

## Questions
- [ ] `lsda/meta_cutoff` routines seem unused. Remove?
- [x] default is **off** 

## Checklist
- [x] documentation added

## Status
- [x] Ready for review
- [x] Ready for merge

----

Python API notes (as one would use in `superfunctionals.py`)
It is recommend to set the threshold via the SuperFunctional class
e.g. `sup[0].set_density_tolerance(1e-10) `

Alternatively one can set the threshold also via the LibXCFunctional class, and potentially
set individual thresholds for the components. Below sets the same threshold for all XC kernels.
```
    dens_cut=core.get_option("SCF","DFT_DENSITY_TOLERANCE")
    if sup[0].is_libxc_func():
        sup[0].c_functionals()[0].set_density_cutoff(dens_cut)
        # print("screening python:",sup[0].c_functionals()[0].density_tolerance())
    else:
        for x_func in sup[0].x_functionals():
            x_func.set_density_cutoff(dens_cut)
        for c_func in sup[0].c_functionals():
            c_func.set_density_cutoff(dens_cut)
```
---

Tests concerning the threshold:
`1e-12` passes all tests without issues
`1e-10` fails only at `pcmsolver-dft` else everything passes

Testing meta-GGA TPSS/aug-pcseg-1 on the A24 benchmark set /w ghost-atoms (simple organic molecules):
[cut10_tpss.apc1.out.txt](https://github.com/psi4/psi4/files/4475320/cut10_tpss.apc1.out.txt)
[cut12_tpss.apc1.out.txt](https://github.com/psi4/psi4/files/4475322/cut12_tpss.apc1.out.txt)
[master_tpss.apc1.out.txt](https://github.com/psi4/psi4/files/4475323/master_tpss.apc1.out.txt)

Looking at the files with vimdiff even `1e-10` shows no practical differences, but offers a better stability (see test cases).  Dipole moments unchanged within accuracy of printout. It is maybe the best choice, but I cannot do exhaustive testing.
`1e-12` is the conservative choice to keep closest to current master.

**Why stay with LibXC defaults:**
However, LibXC defaults (see https://gist.github.com/hokru/546307545c7bf0d96de5cf867e036cae) are sometimes crude out of necessity (< 1e-10) and any custom value will be set recursively for all components.
This can lead to confusion especially if it is a functional definition provided by LibXC itself.
For example for CAM-B3LYP a LibXC query will report a default of `1e-32`, which is misleading because `XC_GGA_X_ITYH` has a default of only `1e-8`. As discussed in https://gitlab.com/libxc/libxc/-/issues/99#note_324078822 this can cause numerical instabilities for density thresholds lower than `1e-10`.
(affects  CAM-B3LYP in `dft-grad-lr2`)
Thus I suggest we stay with LibXC defaults and only in the (hopefully rare) cases a user is facing numerical issues the use of `set dft_density_tolerance 1e-10` is recommended.

### DIFF
--- a/doc/sphinxman/source/dft.rst
+++ b/doc/sphinxman/source/dft.rst
@@ -574,6 +574,12 @@ and the basis cutoffs, but it is likely that significant speed gains may be
 realized by relaxing the basis cutoff tolerance, with negligible decrease in
 accuracy.
 
+Small density values can introduce numerical instabilities with some functionals that
+can result in trailing SCF convergence issues or even numerical failures (NaNs).
+If the default settings of the LibXC library are insufficient, a custom value can be 
+request by setting |scf__dft_density_tolerance|. For notorious cases a value of 1E-10
+is sensible.
+
 An example of a fully specified grid is as follows::
 
     molecule {
@@ -584,12 +590,13 @@ An example of a fully specified grid is as follows::
     set {
     basis cc-pvdz
     scf_type df
-    dft_spherical_points 590     # Often needed
-    dft_radial_points 99         # Often needed
-    dft_radial_scheme treutler   # Rarely needed
-    dft_nuclear_scheme treutler  # Rarely needed
-    dft_basis_tolerance 1.0E-11  # Can speed things up, but benchmark the error
-    dft_pruning_scheme robust    # generally safe and will speed things up
+    dft_spherical_points 590      # Often needed
+    dft_radial_points 99          # Often needed
+    dft_radial_scheme treutler    # Rarely needed
+    dft_nuclear_scheme treutler   # Rarely needed
+    dft_density_tolerance 1.0E-10 # Rarely needed
+    dft_basis_tolerance 1.0E-11   # Can speed things up, but benchmark the error
+    dft_pruning_scheme robust     # Generally safe and will speed things up
     }
 
     energy('b3lyp')
--- a/psi4/driver/procrouting/dft/superfunctionals.py
+++ b/psi4/driver/procrouting/dft/superfunctionals.py
@@ -77,6 +77,11 @@ def build_superfunctional(name, restricted, npoints=None, deriv=1):
     # Lock and unlock the functional
     sup[0].set_lock(False)
 
+    # set LibXC density screening
+    dens_tol = core.get_option("SCF", "DFT_DENSITY_TOLERANCE")
+    if (dens_tol > 0.0):
+        sup[0].set_density_tolerance(dens_tol)
+
     # Set options
     if core.has_option_changed("SCF", "DFT_OMEGA") and sup[0].is_x_lrc():
         omega = core.get_option("SCF", "DFT_OMEGA")
--- a/psi4/src/export_functional.cc
+++ b/psi4/src/export_functional.cc
@@ -83,6 +83,7 @@ void export_functional(py::module &m) {
         .def("grac_shift", &SuperFunctional::grac_shift, "Shift of the bulk potenital.")
         .def("grac_alpha", &SuperFunctional::grac_alpha, "GRAC Alpha.")
         .def("grac_beta", &SuperFunctional::grac_beta, "GRAC Beta.")
+        .def("density_tolerance", &SuperFunctional::density_tolerance, "Density threshold for LibXC.")
         .def("is_gga", &SuperFunctional::is_gga, "Is this a GGA?")
         .def("is_meta", &SuperFunctional::is_meta, "Is this a MGGA?")
         .def("is_x_lrc", &SuperFunctional::is_x_lrc, "Contains range-seperated exchange?")
@@ -111,6 +112,9 @@ void export_functional(py::module &m) {
         .def("set_grac_shift", &SuperFunctional::set_grac_shift, "Sets the GRAC bulk shift value.")
         .def("set_grac_alpha", &SuperFunctional::set_grac_alpha, "Sets the GRAC alpha parameter.")
         .def("set_grac_beta", &SuperFunctional::set_grac_beta, "Sets the GRAC beta parameter.")
+        .def("set_density_tolerance", &SuperFunctional::set_density_tolerance, "Sets the density threshold for LibXC.")
+        .def("copy_density_tolerance", &SuperFunctional::copy_density_tolerance, "Copy the density threshold variable to the SuperFunctional layer.")
+        .def("print_density_threshold", &SuperFunctional::py_print_density_threshold, "Queries the LibXCFunctionals for their density threshold values")
         .def("needs_xc", &SuperFunctional::needs_xc, "Does this functional need XC quantities.")
         .def("needs_vv10", &SuperFunctional::needs_vv10, "Does this functional need VV10 dispersion.")
         .def("needs_grac", &SuperFunctional::needs_grac, "Does this functional need GRAC.")
@@ -127,6 +131,7 @@ void export_functional(py::module &m) {
         .def("omega", &Functional::omega, "docstring")
         .def("lsda_cutoff", &Functional::lsda_cutoff, "docstring")
         .def("meta_cutoff", &Functional::meta_cutoff, "docstring")
+        .def("density_cutoff", &Functional::density_cutoff, "docstring")
         .def("is_gga", &Functional::is_gga, "docstring")
         .def("is_meta", &Functional::is_meta, "docstring")
         .def("is_lrc", &Functional::is_lrc, "docstring")
@@ -139,6 +144,7 @@ void export_functional(py::module &m) {
         .def("set_omega", &Functional::set_omega, "docstring")
         .def("set_lsda_cutoff", &Functional::set_lsda_cutoff, "docstring")
         .def("set_meta_cutoff", &Functional::set_meta_cutoff, "docstring")
+        .def("set_density_cutoff", &Functional::set_density_cutoff, "docstring")
         .def("set_parameter", &Functional::set_parameter, "docstring")
         .def("print_out", &Functional::py_print, "docstring")
         .def("print_detail", &Functional::py_print_detail, "docstring");
@@ -148,6 +154,8 @@ void export_functional(py::module &m) {
         .def("get_mix_data", &LibXCFunctional::get_mix_data, "docstring")
         .def("set_tweak", &LibXCFunctional::set_tweak, "docstring")
         .def("set_omega", &LibXCFunctional::set_omega, "docstring")
+        .def("set_density_cutoff", &LibXCFunctional::set_density_cutoff, "docstring")
+        .def("density_cutoff", &LibXCFunctional::density_cutoff, "docstring")
         .def("query_libxc", &LibXCFunctional::query_libxc, "query libxc regarding functional parameters.");
 
     py::class_<VBase, std::shared_ptr<VBase>>(m, "VBase", "docstring")
--- a/psi4/src/export_functional.cc
+++ b/psi4/src/export_functional.cc
@@ -113,7 +113,6 @@ void export_functional(py::module &m) {
         .def("set_grac_alpha", &SuperFunctional::set_grac_alpha, "Sets the GRAC alpha parameter.")
         .def("set_grac_beta", &SuperFunctional::set_grac_beta, "Sets the GRAC beta parameter.")
         .def("set_density_tolerance", &SuperFunctional::set_density_tolerance, "Sets the density threshold for LibXC.")
-        .def("copy_density_tolerance", &SuperFunctional::copy_density_tolerance, "Copy the density threshold variable to the SuperFunctional layer.")
         .def("print_density_threshold", &SuperFunctional::py_print_density_threshold, "Queries the LibXCFunctionals for their density threshold values")
         .def("needs_xc", &SuperFunctional::needs_xc, "Does this functional need XC quantities.")
         .def("needs_vv10", &SuperFunctional::needs_vv10, "Does this functional need VV10 dispersion.")
--- a/psi4/src/psi4/libfunctional/LibXCfunctional.cc
+++ b/psi4/src/psi4/libfunctional/LibXCfunctional.cc
@@ -161,6 +161,7 @@ LibXCFunctional::LibXCFunctional(std::string xc_name, bool unpolarized) {
 
     // Set any other parameters
     user_omega_ = false;
+    density_cutoff_ = -1.0;
     exc_ = xc_functional_->info->flags & XC_FLAGS_HAVE_EXC;
     vxc_ = xc_functional_->info->flags & XC_FLAGS_HAVE_VXC;
     fxc_ = xc_functional_->info->flags & XC_FLAGS_HAVE_FXC;
@@ -194,11 +195,21 @@ std::shared_ptr<Functional> LibXCFunctional::build_worker() {
     func->set_meta(meta_);
     func->set_lsda_cutoff(lsda_cutoff_);
     func->set_meta_cutoff(meta_cutoff_);
+    func->set_density_cutoff(density_cutoff_);
     func->exc_ = exc_;
     func->vxc_ = vxc_;
     func->fxc_ = fxc_;
 
     return static_cast<std::shared_ptr<Functional>>(func);
+}
+void LibXCFunctional::set_density_cutoff(double cut) {
+    density_cutoff_ = cut;
+    if (density_cutoff_ > 0){
+        xc_func_set_dens_threshold(xc_functional_.get(),cut);
+    }
+}
+double LibXCFunctional::query_density_cutoff() {
+    return xc_functional_->dens_threshold ;
 }
 void LibXCFunctional::set_omega(double omega) {
     omega_ = omega;
--- a/psi4/src/psi4/libfunctional/superfunctional.cc
+++ b/psi4/src/psi4/libfunctional/superfunctional.cc
@@ -384,37 +384,20 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
     can_edit();
     c_functionals_.push_back(fun);
 }
-void SuperFunctional::copy_density_tolerance(double cut) {
-    can_edit();
-    density_tolerance_ = cut;
-}
 void SuperFunctional::set_density_tolerance(double cut) {
-    // modify both LibXCFunctional and Functional
     density_tolerance_ = cut;
-    if (libxc_xc_func_) {
-        for (int Q = 0; Q < c_functionals_.size(); Q++) {
-            c_functionals_[Q]->set_density_cutoff(density_tolerance_);
-        }
-    } else {
         for (int Q = 0; Q < c_functionals_.size(); Q++) {
             c_functionals_[Q]->set_density_cutoff(density_tolerance_);
         };
         for (int Q = 0; Q < x_functionals_.size(); Q++) {
             x_functionals_[Q]->set_density_cutoff(density_tolerance_);
         };
-    };
 }
 void SuperFunctional::print_density_threshold(std::string out, int level) const {
     if (level < 1) return;
     std::shared_ptr<psi::PsiOutStream> printer = (out == "outfile" ? outfile : std::make_shared<PsiOutStream>(out));
     printer->Printf("   => LibXC Density Thresholds  <==\n\n");
     double val = 0.0;
-    if (libxc_xc_func_) {
-        for (int Q = 0; Q < c_functionals_.size(); Q++) {
-            val = c_functionals_[Q]->query_density_cutoff();
-            printer->Printf("    %s:  %6.2E \n", c_functionals_[Q]->name().c_str(), val);
-        }
-    } else {
         for (int Q = 0; Q < c_functionals_.size(); Q++) {
             val = c_functionals_[Q]->query_density_cutoff();
             printer->Printf("    %s:  %6.2E \n", c_functionals_[Q]->name().c_str(), val);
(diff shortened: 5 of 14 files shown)


hokru:
Biggest issue is moving the variables between the Functional/LibXCFunctional/Superfunctional
 objects. Some of the C++ magic is completely lost to me.

This line [here](https://github.com/hokru/psi4/blob/rho_cutoff/psi4/src/psi4/libfunctional/superfunctional.cc#L98)
is supposed to set the density cutoff to the superfunctional. But this [print out](https://github.com/hokru/psi4/blob/rho_cutoff/psi4/src/psi4/libfunctional/superfunctional.cc#L287) shows me it does not work. I can set this value using python [superfunctional](https://github.com/hokru/psi4/blob/rho_cutoff/psi4/driver/procrouting/dft/superfunctionals.py#L80) but that doesn't seem consistent.

I spend a lot of time trying to fix this, but for now I am too confused. Need to revisit later. I welcome any pointers to what could be wrong :)

susilehtola:
A smarter way would be to skip libxc and quadrature altogether for the points that have too small density, e.g. by pruning the grid. (But this would have to be re-done at every iteration.)

susilehtola:
`DFT_DENSITY_TOLERANCE` is a very unintuitive keyword. Use `DFT_DENSITY_CUTOFF` instead.

hokru:
"TOLERANCE" is used in most ofter cutoff-type keywords, so I kept to that. 
Not the biggest fan of it either, but consistency among the keywords is useful.

psi4/src/export_functional.cc
```diff
@@ -83,6 +83,7 @@ void export_functional(py::module &m) {
         .def("grac_shift", &SuperFunctional::grac_shift, "Shift of the bulk potenital.")
         .def("grac_alpha", &SuperFunctional::grac_alpha, "GRAC Alpha.")
         .def("grac_beta", &SuperFunctional::grac_beta, "GRAC Beta.")
+        .def("dens_tolerance", &SuperFunctional::dens_tolerance, "density threshold for LibXC")
```
dgasmith, line 86: `density_tolerance` I wouldn't abbreviate here (I don't think we do elsewhere) and I have been trying to discourage shortenings. Also "Density threshold for LibXC." (note capital/period.)

psi4/src/export_functional.cc
```diff
@@ -111,6 +112,7 @@ void export_functional(py::module &m) {
         .def("set_grac_shift", &SuperFunctional::set_grac_shift, "Sets the GRAC bulk shift value.")
         .def("set_grac_alpha", &SuperFunctional::set_grac_alpha, "Sets the GRAC alpha parameter.")
         .def("set_grac_beta", &SuperFunctional::set_grac_beta, "Sets the GRAC beta parameter.")
+        .def("set_dens_tolerance", &SuperFunctional::set_dens_tolerance, "Sets the density threshold for LibXC")
```
dgasmith, line 115: Ends in a period.

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -381,6 +390,9 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
     can_edit();
     c_functionals_.push_back(fun);
 }
+void SuperFunctional::set_dens_tolerance(double cut) {
+    dens_tolerance_=  cut;
+}
```
dgasmith, line 395: The problem here is that all of the other parameters that you can set do not need to tweak the LibXCFunctional objects and can be handled at the super functional layer which mixed LibXC quantities together. For this function you will need to loop through every functional that is owned by the Superfunctional and call their `set_dens_cutoff` function.

psi4/src/export_functional.cc
hokru: OK. I switched to full length names.

psi4/src/psi4/libfunctional/superfunctional.cc
hokru: Thanks! Setting the cutoff value through SuperFunctional class via python works now.

hokru:
Still having trouble getting the `density_tolerance_` variable in SuperFunctional updated when I set the cutoff value through the LibXCFunctional class.

hokru:
> Still having trouble getting the `density_tolerance_` variable in SuperFunctional updated when I set the cutoff value through the LibXCFunctional class.

Is this necessary at all? I mean, do we want to set the density cutoff _also_ through the LibXCFunctional class? How much flexibility is wanted?

dgasmith:
> > Still having trouble getting the `density_tolerance_` variable in SuperFunctional updated when I set the cutoff value through the LibXCFunctional class.
> 
> Is this necessary at all? I mean, do we want to set the density cutoff _also_ through the LibXCFunctional class? How much flexibility is wanted?

Where are you trying to set this at? The first thing that comes to mind is the functional can be "locked" so that when a threaded V matrix is computed we can correctly replicate the `SuperFunctional` to compute on each thread. It may be worth double checking that you have the correct locking guards around this like other setters.

hokru:
> Where are you trying to set this at?

Too many places...but for example here: https://github.com/hokru/psi4/blob/rho_cutoff/psi4/src/psi4/libfunctional/superfunctional.cc#L98
with `sup->set_density_tolerance(xc_func->density_cutoff())` where `set_density_tolerance` used to set just `density_tolerance_ = <input value>`. But `xc_func->density_cutoff()` only gave me the value from its `init` function.
Though I may have lost track at some point with libxc vs composite functionals..

I solved this now by querrying the first LibXCFunctional correlation functional : https://github.com/hokru/psi4/blob/rho_cutoff/psi4/src/psi4/libfunctional/superfunctional.cc#L417
Same here with explanation:
https://github.com/hokru/psi4/blob/rho_cutoff/psi4/src/psi4/libfunctional/superfunctional.cc#L285

It won't work for a "correlation-less"  XC, which is maybe an too odd edge case?

hokru:
Maybe I should mention that the current version works and I am running tests.
That's also why debug statement are still included.

susilehtola:
Well, it really should work for exchange-only calculations, so ”correlation-less” is important.

hokru:
OK yes I agree.

I already worked something out that will be ready tomorrow.

susilehtola:
Is there a way to control the grid sparsity cutoffs, i.e. the basis function screening thresholds?

The cutoff for the density should be similar to that of the basis functions to be consistent. Smaller densities than this are not evaluated accurately, which may be why you’ve seen converge problems for CAM-B3LYP....

hokru:
> Is there a way to control the grid sparsity cutoffs, i.e. the basis function screening thresholds?

Yes,  i believe `dft_basis_tolerance` does that job. This and my new `dft_density_tolerance` are at `1e-12` for my testing.

hokru:
OK. I added a test that also servers as an example on some details. Combinations of empty, X-only, C-only and multiple X,C kernels are being tested. Also check out the updated OP.
I am happy with the functionality.

This would be ready for review.

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -381,6 +386,51 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
...
+    };
+}
+void SuperFunctional::print_LibXC_dens_threshold(std::string out, int level) const {
+    if (level < 1) return;
+    std::shared_ptr<psi::PsiOutStream> printer = (out == "outfile" ? outfile : std::make_shared<PsiOutStream>(out));
+    printer->Printf("   => LibXC Densities Thresholds  <==\n\n");
+    double val = 0.0;
+    if (libxc_xc_func_) {
+        for (int Q = 0; Q < c_functionals_.size(); Q++) {
+            val = dynamic_cast<LibXCFunctional*>(c_functionals_[Q].get())->querry_density_cutoff();
```
loriab, line 419: querry -> query

multiple occurrences, incl in export_functional.cc

doc/sphinxman/source/dft.rst
```diff
@@ -584,12 +589,13 @@ An example of a fully specified grid is as follows::
...
-    dft_nuclear_scheme treutler  # Rarely needed
-    dft_basis_tolerance 1.0E-11  # Can speed things up, but benchmark the error
-    dft_pruning_scheme robust    # generally safe and will speed things up
+    dft_spherical_points 590      # Often needed
+    dft_radial_points 99          # Often needed
+    dft_radial_scheme treutler    # Rarely needed
+    dft_nuclear_scheme treutler   # Rarely needed
+    dft_density_tolerance 1.0E-10 # Rarely needed
+    dft_basis_tolerance 1.0E-11   # Can speed things up, but benchmark the error
+    dft_pruning_scheme robust     # generally safe and will speed things up
```
dgasmith, line 598: Capitalize the `g` in `generally` for consistency please.

psi4/driver/procrouting/dft/superfunctionals.py
```diff
@@ -77,6 +77,13 @@ def build_superfunctional(name, restricted, npoints=None, deriv=1):
     # Lock and unlock the functional
     sup[0].set_lock(False)
 
+    # set LibXC density screening
+    dens_tol = core.get_option("SCF","DFT_DENSITY_TOLERANCE")
```
dgasmith, line 81: Please format this file.

psi4/src/export_functional.cc
```diff
@@ -111,6 +112,9 @@ void export_functional(py::module &m) {
         .def("set_grac_shift", &SuperFunctional::set_grac_shift, "Sets the GRAC bulk shift value.")
         .def("set_grac_alpha", &SuperFunctional::set_grac_alpha, "Sets the GRAC alpha parameter.")
         .def("set_grac_beta", &SuperFunctional::set_grac_beta, "Sets the GRAC beta parameter.")
+        .def("set_density_tolerance", &SuperFunctional::set_density_tolerance, "Sets the density threshold for LibXC.")
+        .def("copy_density_tolerance", &SuperFunctional::copy_density_tolerance, "Copy the density threshold variable to the SuperFunctional layer.")
+        .def("print_LibXC_dens_threshold", &SuperFunctional::py_print_LibXC_dens_threshold, "Querries the LibXCFunctionals for their density threshold values")
```
dgasmith, line 117: `_libxc_` should be lowercase and dens -> density. This particular function does start to have a leaky abstraction, is there any way this could be `print_density_threshold`?

hokru: What does leaky abstraction mean? I can rename this, yes. I have no strong opinion about the naming of any functions.

dgasmith: The idea here is to tuck all of the detail of LibXC behind the Functional/SuperFunctional classes so if we need to add another library it is easy to do. Here SuperFunctional now mentions LibXC when the practicalities should be hidden.

hokru:
Addressed the first review comments so far. Also added more notes on the default cutoff value in the OP.

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -381,6 +386,51 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
...
+    };
+}
+void SuperFunctional::print_density_threshold(std::string out, int level) const {
+    if (level < 1) return;
+    std::shared_ptr<psi::PsiOutStream> printer = (out == "outfile" ? outfile : std::make_shared<PsiOutStream>(out));
+    printer->Printf("   => LibXC Density Thresholds  <==\n\n");
+    double val = 0.0;
+    if (libxc_xc_func_) {
+        for (int Q = 0; Q < c_functionals_.size(); Q++) {
+            val = dynamic_cast<LibXCFunctional*>(c_functionals_[Q].get())->query_density_cutoff();
```
dgasmith, line 419: Could we have getters/setters for the Functional class which raise a not implemented error and then put the specific implementation in LibXCFunctional so that we do not need to dynamic cast here?

hokru: Do you mean just here or also in the function above?

hokru: OK, today I learned about what a virtual function is and what those `override` commands do. Very helpful. Will remove all those dynamic casts I introduced.

hokru:
The introduction of `dens_threshold` has implications for PSI4 and CAM-B3LYP: https://gitlab.com/libxc/libxc/-/issues/99#note_324078822
(possibly other functionals, too? @susilehtola )

Either the default threshold will to be larger than `1e-12` or perhaps functional specific threshold are an option.
It seems `2.5E-11` still passes the cam-b3lyp test, but `1E-11` not anymore.
For users CAM-B3LYP should work out of the box

susilehtola:
@hokru yes, the threshold will override whatever is the default in libxc. The way I did it in ERKALE was by excluding the points where the density was too small, i.e. they weren't even fed in to libxc. This was before libxc got the capability to set the density threshold.

hokru:
> the threshold will override whatever is the default in libxc

They are usually super small, I only saw values between 1e-20 to 1e-32, so the 1e-8 in a component of cam-b3lyp is therefore surprising.

Pruning the grid points from the beginning is a bigger project, I think. And current PSI4 can fail at H atoms, which should be fixed more quickly.

susilehtola:
Yeah, I'm not saying my way was the best :stuck_out_tongue_closed_eyes:

hokru:
The following list is likely incomplete, just a basic grep for certain numbers, but here is
a list of functional files with a default threshold `< 1e-12` that can cause problems:

```
gga_c_gapc.c:  1e-10,
gga_c_q2d.c:  1e-10,
gga_c_wi.c:  1e-10,
gga_c_zvpbeint.c:  1e-10,
gga_c_zvpbeint.c:  1e-10,
gga_c_zvpbeloc.c:  1e-10,
gga_c_zvpbeloc.c:  1e-10,
gga_c_zvpbeloc.c:  1e-10,
gga_x_hjs.c:  1e-10,
mgga_x_mbeefvdw.c:  1e-10,
gga_c_pbeloc.c:  1e-11,
mgga_x_m11.c:  1e-11,
mgga_x_m11.c:  1e-11,
mgga_x_rscan.c:  1e-11,
gga_x_hjs.c:  1e-7, /* densities smaller than 1e-7 yield NaNs */
gga_x_ityh.c:  1e-8,
gga_x_lcgau.c:  1e-8,
gga_x_lcgau.c:  1e-8,
gga_x_lcgau.c:  1e-8,
lda_c_2d_amgb.c:  1e-9,
mgga_c_tpssloc.c:  1e-9, /* densities smaller than 1e-26 give NaNs */
```

And for `= 1e-12`:
```
gga_c_lm.c:  1e-12,
gga_c_optc.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe.c:  1e-12,
gga_c_pbe_vwn.c:  1e-12,
gga_c_pw91.c:  1e-12,
gga_c_sg4.c:  1e-12,
gga_c_wl.c:  1e-12,
gga_c_zpbeint.c:  1e-12,
gga_c_zpbeint.c:  1e-12,
gga_x_beefvdw.c:  1e-12,
gga_x_pbeint.c:  1e-12,
```

hokru:
Full overview of default cutoffs and generating python file can be found here: https://gist.github.com/hokru/546307545c7bf0d96de5cf867e036cae

Currently the best way forward, I think, is leaving density screening **off** by default to avoid setting too tight thresholds for certain components leading to unintuitive behaviour. 
I will update the PR accordingly unless another direction is wanted.

psi4/src/export_functional.cc
```diff
@@ -111,6 +112,9 @@ void export_functional(py::module &m) {
         .def("set_grac_shift", &SuperFunctional::set_grac_shift, "Sets the GRAC bulk shift value.")
         .def("set_grac_alpha", &SuperFunctional::set_grac_alpha, "Sets the GRAC alpha parameter.")
         .def("set_grac_beta", &SuperFunctional::set_grac_beta, "Sets the GRAC beta parameter.")
+        .def("set_density_tolerance", &SuperFunctional::set_density_tolerance, "Sets the density threshold for LibXC.")
+        .def("copy_density_tolerance", &SuperFunctional::copy_density_tolerance, "Copy the density threshold variable to the SuperFunctional layer.")
+        .def("print_density_threshold", &SuperFunctional::py_print_density_threshold, "Querries the LibXCFunctionals for their density threshold values")
```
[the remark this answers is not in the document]
hokru: I followed the other `::py` print functions e.g. `.def("print_out", &Functional::py_print, "docstring")`.
Those call the C++ print function with` py_` with `"outfile"` as `PsiOutStream`. Then one connects automatically to the right output when used at the python layer. At least that's how I make sense of it.

The wrong spelling of query is burned into my mind :(

hokru:
I ran clang-format over some of the files and also looked manually for odd empty lines.

There is the remaining question if the unused (as far as I see) `lsda/meta_cutoff` routines should stay or be remove?

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -381,6 +384,48 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
     can_edit();
     c_functionals_.push_back(fun);
 }
+void SuperFunctional::copy_density_tolerance(double cut) {
```
dgasmith, line 387: This isn't used as far as I can tell? Should be handled by the `set` function.

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -381,6 +384,48 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
...
+    can_edit();
+    density_tolerance_ = cut;
+}
+void SuperFunctional::set_density_tolerance(double cut) {
+    // modify both LibXCFunctional and Functional
+    density_tolerance_ = cut;
+    if (libxc_xc_func_) {
```
dgasmith, line 394: I don't think this `if` needs to be here. `x_functionals_.size()` is simply zero.

psi4/src/psi4/libfunctional/LibXCfunctional.cc
```diff
@@ -194,12 +195,22 @@ std::shared_ptr<Functional> LibXCFunctional::build_worker() {
...
 
     return static_cast<std::shared_ptr<Functional>>(func);
 }
+void LibXCFunctional::set_density_cutoff(double cut) {
+    density_cutoff_ = cut;
+    if (density_cutoff_ > 0){
+        xc_func_set_dens_threshold(xc_functional_.get(),cut);
+    }
+}
+double LibXCFunctional::query_density_cutoff() {
```
dgasmith, line 211: As far as I can see the only reason its here is if you don't trust the interface so this is a testing function which is better handled by an actual test (setting the density tolerance).

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -381,6 +384,48 @@ void SuperFunctional::add_c_functional(std::shared_ptr<Functional> fun) {
...
+            x_functionals_[Q]->set_density_cutoff(density_tolerance_);
+        };
+    };
+}
+void SuperFunctional::print_density_threshold(std::string out, int level) const {
+    if (level < 1) return;
+    std::shared_ptr<psi::PsiOutStream> printer = (out == "outfile" ? outfile : std::make_shared<PsiOutStream>(out));
+    printer->Printf("   => LibXC Density Thresholds  <==\n\n");
+    double val = 0.0;
+    if (libxc_xc_func_) {
```
dgasmith, line 412: Ditto with the if statement here.

psi4/src/psi4/libfunctional/LibXCfunctional.cc
hokru: The main use is in the `SuperFunctional::print_density_threshold` function https://github.com/psi4/psi4/pull/1860/files#diff-7c91ae12434f3595779f6de77e652235R407  to also print the internal defaults used by LibXC and not just the user specified values. (with the caveat that `libxc_xc_func` functionals do not report threshold for their functional components , see CAM-B3LYP issue in the main post)

psi4/src/psi4/libfunctional/superfunctional.cc
hokru: correct, will be removed in next push.

psi4/src/psi4/libfunctional/superfunctional.cc
hokru: right, I think it was needed before when I did things incorrectly but can be removed now.

hokru:
If you agree with the review changes @dgasmith, this would be ready for merging.

dgasmith:
Looks good, thanks for the changes!

PR #1862, opened 2020-04-09
enhances the visibility of psi4's error messages

hokru:
## Description
Extracts the `Exception`-based error message and prints it out in an attention-seeking box.

## Checklist
- [x] few manual tests

## Status
- [x] Ready for review
- [x] Ready for merge

----
Example:
```
        Cleaning optimization helper files.

PsiException: Could not converge geometry optimization in 1 iterations.


  File "/Users/kruse/qc/psi4.bin/bin/psi4", line 331, in <module>
    exec(content)
  File "<string>", line 29, in <module>
  File "/Users/kruse/qc/psi4.bin/lib/psi4/driver/driver.py", line 1201, in optimize
    raise OptimizationConvergenceError("""geometry optimization""", n - 1, wfn)

psi4.driver.p4util.exceptions.OptimizationConvergenceError: Could not converge geometry optimization in 1 iterations.


Printing out the relevant lines from the Psithon --> Python processed input file:
    ""","LiF2m")
    core.IO.set_default_namespace("LiF2m")
    core.set_global_option("BASIS", "minix")
    core.set_global_option("SCF_TYPE", "pk")
    core.set_global_option("GEOM_MAXITER", 1)
--> optimize('hf')

!----------------------------------------------------------- !
!                                                            !
!  Could not converge geometry optimization in 1 iterations. !
!                                                            !
!----------------------------------------------------------- !

    Psi4 stopped on: Thursday, 09 April 2020 10:17AM
    Psi4 wall time for execution: 0:00:00.77

*** Psi4 encountered an error. Buy a developer more coffee!
*** Resources and help at github.com/psi4/psi4.
```

### DIFF
--- a/psi4/extras.py
+++ b/psi4/extras.py
@@ -204,38 +204,4 @@ def test(extent='full', extras=None):
 
     retcode = pytest.main(command)
     return retcode
-
-def message_box(message: str = None, max_width: int = 80):
-    """ put a message string into a box for extra attention
-
-    Parameters
-    -----------
-    message
-       message string to be boxed
-
-    max_width
-        maximal character width of the box
-
-    Returns
-    --------
-    str
-       box containing the message as a multiline string
-    """
-    from textwrap import wrap
     
-    is_long = (len(message) > max_width)
-    box_width = max_width if is_long else len(message)
-
-    message = " ".join(message.split("\n"))
-    message = wrap(message, box_width, break_long_words=False)
-    error_str = []
-    error_str.append('\n!' + '-' * box_width + '--!\n')
-
-    fmt = "! {:" + str(box_width) + "} !\n"
-    for line in message:
-        error_str.append(fmt.format(line))
-
-    error_str.append('!' + '-' * box_width + '--!\n')
-    error_str = ''.join(error_str)
-
-    return error_str
--- a/psi4/extras.py
+++ b/psi4/extras.py
@@ -204,3 +204,23 @@ def test(extent='full', extras=None):
 
     retcode = pytest.main(command)
     return retcode
+
+def error_box(message: str = None):
+    """ put a message string into a box for extra attention
+
+    Parameters
+    -----------
+    message
+       message string to be boxed
+
+    Returns
+    --------
+    str
+       box containing the message as a multiline string
+    """
+    error_str = '\n!' + '-' * len(message) + ' !\n'
+    error_str += '!' + ' ' * len(message) + ' !\n'
+    error_str += '! ' + message.rstrip() + ' !' + '\n'
+    error_str += '!' + ' ' * len(message) + ' !\n'
+    error_str += '!' + '-' * len(message) + ' !\n'
+    return error_str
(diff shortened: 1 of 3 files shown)


hokru:
The extraction of the error message looks complicated, but I want to catch the case where the message contains other ":" characters.

JonathonMisiewicz:
I'd suggest trying to identify what the "header" to remove in a given error message is for a large variety of error messages, those raised C-side and those raised Py-side. Once you have that, look into the source to confirm the error messages follow that pattern, and when it's confirmed, remove what matches that pattern.

Then you should be able to handle the case that an error message contains a colon.

hokru:
I may have been unclear. It already works with multiple ":" characters. Not sure if there is already such an error message somewhere.

Format always seems: `<exception name>: <exception message>`, so the splitting the string at the first ":"  in 2 parts and then taking the second (=last) part should work well.

JonathonMisiewicz:
Oh, I misread - my bad. I'm mostly fine with the style, but the righthand line is off. There should be two dashes between the end of text and the start of the line, to mirror the lefthand side. Right now, you have one space for some reason.

psi4/extras.py
```diff
@@ -204,3 +204,23 @@ def test(extent='full', extras=None):
...
+    Parameters
+    -----------
+    message
+       message string to be boxed
+
+    Returns
+    --------
+    str
+       box containing the message as a multiline string
+    """
```
dgasmith, line 220: I think we need something like:

```python
box_width = 120
message = " ".join(message.split("\n"))
from textwrap import wrap
message = wrap(message, box_width, break_long_words=False)

error_str = []
error_str.append('\n!' + '-' * len(box_width) + ' !\n')

fmt = "{:" + str(box_width) + "}"
for line in message:
    error_str.append(fmt.format(line))
...

```

This will handle long lines gracefully and will correctly pad the width like the rest of the code. @loriab might have a box style built in.

loriab: there are the banners https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/text.py#L157 but they're pretty old and could probably benefit from `wrap` themselves.

hokru: Good point. I assumed only a short, single line error message needs to be boxed.

hokru:
Thanks for the code snippet @dgasmith ! I added the text wrapping. It adapts to the length of the message up to a maximum value of 80.  One can set a different maximum via optional argument. I renamed the function to a more general `message_box` in case one wants to use it for something else.

 For testing: [wrap.py.txt](https://github.com/psi4/psi4/files/4457324/wrap.py.txt)

hokru:
```
!------------------------------------------------------!
!                                                      !
!  Could not converge SCF iterations in 1 iterations.  !
!                                                      !
!------------------------------------------------------!
```
or
```
!------------------------------------------------------!
!  Could not converge SCF iterations in 1 iterations.  !
!------------------------------------------------------!
```

I'd prefer the additional blank lines to make it stand out more.

loriab:
Since this is for ppl who can't pick a nice natural language sentence repeated twice in a couple dozen lines of helpful context, I'm with you on the extra space to be as obvious as possible. It might even be time for red (first 8 and last line of https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/primitives/diff_gbs.py#L45-L100)

PR #1879, opened 2020-05-05
Eliminate `psimrcc` Debugging and Timer

JonathonMisiewicz:
## Description
This PR eliminates the old debugging and timer system from `psimrcc`. All debugger code is gone for good, since we have `lldb` and such nowadays. All timer code now uses the more standard `libqt` timer system. This eliminates two globals from psimrcc and several hundreds of lines of code. This also eliminates some functions that were never completely coded. They had no reason to be in the code.

This is part of the yak shaving needed for the gradient refactor. Yes, seriously.

## Todos
- [x] Eliminated `psimrcc` globals and standardized timer behavior

## Checklist
- [x] `ctest -R psimrcc -j4` passes

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/psimrcc/operation_compute.cc
```diff
@@ -188,13 +182,6 @@ void CCOperation::tensor_product() {
 }
 
 void CCOperation::contract() {
-    if (compatible_contract() && (reindexing.size() == 0)) {
-        // Same indexing contract, let BLAS directly handle this (although we guide it)
-        DEBUGGING(4, outfile->Printf("\n...same indexing for the target and the output of this operation"););
-    } else {
-        // Different indexing contract, we work this by hand at first
-        DEBUGGING(4, outfile->Printf("\n...different indexing for the target and the output of this operation"););
```
loriab, line 196: should any of these be warnings or exits?

JonathonMisiewicz: My read of the code is that these are **not** warnings or exits. They just indicate whether or not the algorithm needs to account for complicated tensor re-indexing. Which is exactly the kind of thing you'd want to know if you were hunting for a bug in this function.

JonathonMisiewicz:
> Looks fine, but why are we keeping `debbuging.{cc,h}` around?

That was supposed to be part of the last commit, but didn't get added for some reason. My blind guess is that the `git add *` command I used doesn't stage file deletion.

PR #1882, opened 2020-05-06
Update Ca_subset/Cb_subset docs.

JonathonMisiewicz:
## Description/Todos
- [x] Updated documentation for the `Ca_subset` and `Cb_subset` functions to explicitly show the accepted argument values.

## Checklist
- [x] Built the docs

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_wavefunction.cc
+++ b/psi4/src/export_wavefunction.cc
@@ -105,10 +105,38 @@ void export_wavefunction(py::module& m) {
         .def("efzc", &Wavefunction::efzc, "Returns the frozen-core energy")
         .def("Ca", &Wavefunction::Ca, "Returns the Alpha Orbitals.")
         .def("Cb", &Wavefunction::Cb, "Returns the Beta Orbitals.")
-        .def("Ca_subset", &Wavefunction::Ca_subset, py::return_value_policy::take_ownership,
-             "Returns the requested Alpha Orbital subset.")
-        .def("Cb_subset", &Wavefunction::Cb_subset, py::return_value_policy::take_ownership,
-             "Returns the requested Beta Orbital subset.")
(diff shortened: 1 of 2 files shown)


psi4/src/export_wavefunction.cc
```diff
@@ -105,10 +105,38 @@ void export_wavefunction(py::module& m) {
...
+                  A Pitzer-ordered matrix of the orbitals, (# basis functions, # orbitals in the subset).
+         )pbdoc")
+        .def("Cb_subset", &Wavefunction::Cb_subset, py::return_value_policy::take_ownership, R"pbdoc(
+              Returns the requested Beta orbital subset.
+
+              Parameters
+              ----------
+              basis : string
+                  Should the subset be of symmetry orbitals or atomic orbitals, SO or AO?
+              subset : string
```
loriab, line 131: Can do `subset : {'ALL', 'ACTIVE'}` etc. for enum-like (bottom of 4 in https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)

JonathonMisiewicz: Ah, I was looking for syntax like that but must have missed it. sphinx needs to rebuild, but I'll commit that once that's done.

loriab: they don't make it easy to find. but sadly, I have read that document carefully. I'm fine with merging as-is, as there'll probably be other tweaks, too, but commit if you like. Now that psi4-dev is built again, I'll try to rehabilitate psi4-docs (docs & cov) locally.

loriab:
try https://github.com/pybind/pybind11/issues/2171 for Win build fail

PR #1884, opened 2020-05-08
improvements to the FCHK writer

### LINKED ISSUE #2021: Psi4 thinks Iodine is Maganese
[_5-iodouracil-dma.psi4.txt](https://github.com/psi4/psi4/files/5410680/_5-iodouracil-dma.psi4.txt)

If you look at the fchk file produced, the atom number for Iodine will say 25, not 53 for Iodine!

hokru:
## Description
Tries to set the correct density label and solves some bugs with writing the incorrect densities.

Relates issue: #1272 

## Todos
- [x] deducts density label (`|SCF|MP2|CC|CI|DCT`) for the FCHK file from the method.
- [x] allows PSI4 method names as density label with option `strict_label=False`
- [x] related issue [1886](https://github.com/psi4/psi4/issues/1886) is handled within the FCHK driver.
- [x] new pytests for FCHK files that check if the correct SCF density and SCF orbital coefficients is set.
- [x] No densities available from `occ` gradients. FCHK driver exists with a warning.
- [x] DCT, DFOCC densities available
- [x] exposes `Vector.copy()` to python
- [x] better ECP support. Fixes #2021.
- [x] adds a 'compare_fchkfiles` function.

## Checklist
- [x] docs changed to reflect that the high-level `fchk()` should be called (instead of `FCHKWriter()` directly)
- [x] new test passes
- [x] manual inspect of various fchk files using this (ugly) script: https://gist.github.com/hokru/f1b00a87665ac345415b4ad3e0006f1d

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1931,7 +1931,7 @@ def gdma(wfn, datafile=""):
     if not datafile:
         os.remove(commands)
 
-def fchk(wfn, filename, debug=False, strict_label=True):
+def fchk(wfn, filename, *, debug=False, strict_label=True):
     """Function to write wavefunction information in *wfn* to *filename* in
     Gaussian FCHK format.
 
@@ -1949,7 +1949,8 @@ def fchk(wfn, filename, debug=False, strict_label=True):
     :param debug: returns a dictionary to aid with debugging
 
     :type strict_label: boolean
-    :param strict_label: Set a standard label if possible. A warning will be printed if this is not possible.
+    :param strict_label: If true set a density label compliant with what Gaussian would write. A warning will be printed if this is not possible.
+                         Otherwise set the density label according to the method name.
 
     Notes
     -----
@@ -1977,7 +1978,7 @@ def fchk(wfn, filename, debug=False, strict_label=True):
     """
     # * Known limitations and notes *
     #
-    # OCC: (occ theory module only, not dfocc) is turned off as densities are not correctly set. 
+    # OCC: (occ theory module only, not dfocc) is turned off as densities are not correctly set.
     # DFMP2: Contains natural orbitals in wfn.C() and wfn.epsilon() data. This is fixed to contain respective HF data.
 
     allowed = ['DFMP2', 'SCF', 'CCENERGY', 'DCT', 'DFOCC']
@@ -2013,10 +2014,8 @@ def fchk(wfn, filename, debug=False, strict_label=True):
     current = varlist['CURRENT ENERGY']
 
     # delete problematic entries
-    delete_list = ['CURRENT ENERGY', 'CURRENT REFERENCE ENERGY']
-    for key in delete_list:
-        if key in varlist:
-            del varlist[key]
+    for key in ['CURRENT ENERGY', 'CURRENT REFERENCE ENERGY']:
+        varlist.pop(key, None)
 
     # find closest matching energy
     for (key, val) in varlist.items():
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1931,7 +1931,7 @@ def gdma(wfn, datafile=""):
     if not datafile:
         os.remove(commands)
 
-def fchk(wfn, filename, debug=False):
+def fchk(wfn, filename, debug=False, strict_label=True):
     """Function to write wavefunction information in *wfn* to *filename* in
     Gaussian FCHK format.
 
@@ -1948,11 +1948,14 @@ def fchk(wfn, filename, debug=False):
     :type debug: boolean
     :param debug: returns a dictionary to aid with debugging
 
+    :type strict_label: boolean
+    :param strict_label: Set a standard label if possible. A warning will be printed if this is not possible.
+
     Notes
     -----
     * A description of the FCHK format is http://wild.life.nctu.edu.tw/~jsyu/compchem/g09/g09ur/f_formchk.htm
     * The allowed headers for methods are general and limited, i.e., "Total SCF|MP2|CI|CC Density",
-      PSI4 will try to find the right one for the current calculation.
+      PSI4 will try to find the right one for the current calculation. If `strict_label=False` the PSI4 method name will be used as label.
     * Not all theory modules in PSI4 are compatible with the FCHK writer.
       A warning will be printed if a theory module is not supported.
     * Caution! For orbital-optimized correlated methods (e.g. DCT, OMP2) the 'Orbital Energy' field contains ambiguous data.
@@ -1967,12 +1970,15 @@ def fchk(wfn, filename, debug=False):
     >>> E, wfn = gradient('ccsd', return_wfn=True)
     >>> fchk(wfn, 'mycalc.fchk')
 
+    >>> # [2] Write FCHK file with non-standard label.
+    >>> E, wfn = gradient('mp2.5', return_wfn=True)
+    >>> fchk(wfn, 'mycalc.fchk', strict_label=False)
+
     """
     # * Known limitations and notes *
     #
-    # OCC: (occ theory module only, not dfocc) is turned off as it densities are not set. 
-    # DFMP2: Contains natural orbitals in wfn.C() and wfn.epsilon() data.
-    # This is fixed to contain respective HF data.
+    # OCC: (occ theory module only, not dfocc) is turned off as densities are not correctly set. 
+    # DFMP2: Contains natural orbitals in wfn.C() and wfn.epsilon() data. This is fixed to contain respective HF data.
 
     allowed = ['DFMP2', 'SCF', 'CCENERGY', 'DCT', 'DFOCC']
     module_ = wfn.module().upper()
@@ -1999,7 +2005,7 @@ def fchk(wfn, filename, debug=False):
 
     if (module_ in ['DCT', 'DFOCC']):
         core.print_out("""FCHKWriter: Caution! For orbital-optimized correlated methods
-            the 'Orbital Energy' field contains ambiguous data.""")
+            the 'Orbital Energy' field contains ambiguous data. \n""")
 
     # At this point we don't know the method name, so we try to search for it.
     # idea: get the method from the variable matching closely the 'current energy'
@@ -2014,11 +2020,13 @@ def fchk(wfn, filename, debug=False):
 
     # find closest matching energy
     for (key, val) in varlist.items():
-        if (np.isclose(val, current, 1e-10)):
-            method = key
+        print(key,val)
+        if (np.isclose(val, current, 1e-12)):
+            method = key.split()[0]
 
-    # The labels this functions knows. Extend if needed.
-    labels = {
+    # The 'official' list of labels for compatibility.
+    # OMP2,MP2.5,OCCD, etc get reduced to MP2,CC.
+    allowed_labels = {
         "HF": " SCF Density",
         "SCF": " SCF Density",
         "DFT": " SCF Density",
@@ -2027,15 +2035,18 @@ def fchk(wfn, filename, debug=False):
         "MP4": " MP4 Density",
         "CI": " CI Density",
         "CC": " CC Density",
-        "DCT": " DCT Density",
     }
     # assign label from method name
-    fchk_label = 'UNKNOWN'
-    for key in labels:
-        if key in method:
-            fchk_label = labels[key]
-    if fchk_label == 'UNKNOWN':
-        raise ValidationError('Unknown density in FCHKWriter')
+    fchk_label = f" {method} Density"
+    if strict_label:
+        in_list = False
+        for key in allowed_labels:
+            if key in method:
+                fchk_label = allowed_labels[key]
+                in_list = True
+        if not in_list:
+            core.print_out(f"FCHKWriter: !WARNING! {method} is not recognized. Using non-standard label.\n")
+    core.print_out(f"FCHKWriter: Writing {filename} with label '{fchk_label}'.\n")
     fw.set_postscf_density_label(fchk_label)
 
     fw.write(filename)
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1982,7 +1982,7 @@ def fchk(wfn, filename, debug=False, strict_label=True):
 
     allowed = ['DFMP2', 'SCF', 'CCENERGY', 'DCT', 'DFOCC']
     module_ = wfn.module().upper()
-    if (module_ not in allowed):
+    if module_ not in allowed:
         core.print_out(f"FCHKWriter: Theory module {module_} is currently not supported by the FCHK writer.")
         return None
 
@@ -1990,7 +1990,7 @@ def fchk(wfn, filename, debug=False, strict_label=True):
         core.print_out(f"FCHKWriter: Limited ECP support! No ECP data will be written to the FCHK file.")
 
     # fix orbital coefficients and energies for DFMP2
-    if (module_ in ['DFMP2']):
+    if module_ in ['DFMP2']:
         wfn_ = core.Wavefunction.build(wfn.molecule(), core.get_global_option('BASIS'))
         wfn_.deep_copy(wfn)
         refwfn = wfn.reference_wavefunction()
@@ -2003,7 +2003,7 @@ def fchk(wfn, filename, debug=False, strict_label=True):
     else:
         fw = core.FCHKWriter(wfn)
 
-    if (module_ in ['DCT', 'DFOCC']):
+    if module_ in ['DCT', 'DFOCC']:
         core.print_out("""FCHKWriter: Caution! For orbital-optimized correlated methods
             the 'Orbital Energy' field contains ambiguous data. \n""")
 
@@ -2020,9 +2020,9 @@ def fchk(wfn, filename, debug=False, strict_label=True):
 
     # find closest matching energy
     for (key, val) in varlist.items():
-        print(key,val)
         if (np.isclose(val, current, 1e-12)):
             method = key.split()[0]
+            break
 
     # The 'official' list of labels for compatibility.
     # OMP2,MP2.5,OCCD, etc get reduced to MP2,CC.
@@ -2042,6 +2042,8 @@ def fchk(wfn, filename, debug=False, strict_label=True):
         in_list = False
         for key in allowed_labels:
             if key in method:
+                if key is not method:
+                    core.print_out(f"FCHKWriter: !WARNING! method '{method}'' renamed to label '{key}'.\n")
                 fchk_label = allowed_labels[key]
                 in_list = True
         if not in_list:
@@ -2051,7 +2053,7 @@ def fchk(wfn, filename, debug=False, strict_label=True):
 
     fw.write(filename)
     # needed for the pytest. The SCF density below follows PSI4 ordering not FCHK ordering.
-    if (debug):
+    if debug:
         ret = {
             "filename": filename,
             "detected energy": method,
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1980,6 +1980,9 @@ def fchk(wfn, filename, debug=False):
         core.print_out(f"FCHKWriter: Theory module {module_} is currently not supported by the FCHK writer.")
         return None
 
+    if (wfn.basisset().has_ECP()):
+        core.print_out(f"FCHKWriter: Limited ECP support! No ECP data will be written to the FCHK file.")
+
     # fix orbital coefficients and energies for DFMP2
     if (module_ in ['DFMP2']):
         wfn_ = core.Wavefunction.build(wfn.molecule(), core.get_global_option('BASIS'))
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -1931,8 +1931,7 @@ def gdma(wfn, datafile=""):
     if not datafile:
         os.remove(commands)
 
-
-def fchk(wfn, filename):
+def fchk(wfn, filename, debug=False):
     """Function to write wavefunction information in *wfn* to *filename* in
     Gaussian FCHK format.
 
@@ -1940,17 +1939,23 @@ def fchk(wfn, filename):
 
     :returns: None
 
+    :type wfn: :py:class:`~psi4.core.Wavefunction`
+    :param wfn: set of molecule, basis, orbitals from which to generate fchk file
+
     :type filename: string
     :param filename: destination file name for FCHK file
 
-    :type wfn: :py:class:`~psi4.core.Wavefunction`
-    :param wfn: set of molecule, basis, orbitals from which to generate fchk file
+    :type debug: boolean
+    :param debug: returns a dictionary to aid with debugging
 
     Notes
     -----
     * A description of the FCHK format is http://wild.life.nctu.edu.tw/~jsyu/compchem/g09/g09ur/f_formchk.htm
     * The allowed headers for methods are general and limited, i.e., "Total SCF|MP2|CI|CC Density",
-      so "CC" is always used for the post-HF case.
+      PSI4 will try to find the right one for the current calculation.
+    * Not all theory modules in PSI4 are compatible with the FCHK writer.
+      A warning will be printed if a theory module is not supported.
+    * Caution! For orbital-optimized correlated methods (e.g. DCT, OMP2) the 'Orbital Energy' field contains ambiguous data.
 
     :examples:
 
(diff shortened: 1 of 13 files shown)


hokru:
`wfn.name` seems not the most reliable identifier. Investigating.

psi4/driver/driver.py
```diff
@@ -1948,6 +1948,45 @@ def fchk(wfn, filename):
...
+        fchk_label = ' CC Density'
+    elif (re.search("MP2", method) is not None):
+        fchk_label = ' MP2 Density'
+    elif (re.search("MP3", method) is not None):
+        fchk_label = ' MP3 Density'
+    elif (re.search("MP4", method) is not None):
+        fchk_label = ' MP4 Density'
+    elif (re.search("CI", method) is not None):
+        fchk_label = ' CI Density'
+    elif (re.search("DC", method) is not None):
```
JonathonMisiewicz, line 1986: All DCT psivars use the letters "[DCT](http://www.psicode.org/psi4manual/master/autodir_psivariables/module__dct.html#apdx-dct-psivar)", so you can do that for a more specific search.

Also, is there a reason you're using a regex here? All of your `re.search(search_string, method) is not None` can be written as `search_string in method`. Less regex, less code.

hokru: thx. Initially I thought it would be more difficult to map `method` to the `label`. The `wfn.name()` could really have been anything. Now with the psivars also that got simpler and regex is likely not needed, true.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -617,13 +622,13 @@ void FCHKWriter::write(const std::string &filename) {
     // In theory, the correlated density should be printed out with a
     // legend depending on the method (MP2, MP3, MP4, CI, CC) but this
     // is probably the most common anyhow
```
susilehtola, line 624: You can change this comment

susilehtola:
Since Psi4 I guess is assuming the Aufbau principle anyway, you can recreate the SCF density from the orbitals in the code.

hokru:
> Since Psi4 I guess is assuming the Aufbau principle anyway, you can recreate the SCF density from the orbitals in the code.

That's a good idea. The cost of recomputing is small enough in practical use cases.

Then it wouldn't matter what densities are written to the reference wfn, though one still would want to have a harmonised behaviour throughout psi4 and not guess which density one gets, long term at least.

hokru:
Added the SCF density recomputation. `grad('ccsd')` (via `ccenergy`) now gives the correct SCF densities in the FCHK file.

Code not yet cleaned up. I also want to add some form of minimal testing.

hokru:
Uff, I don't think I will manage to patch up the writer to treat the symptoms from how various modules set the wavefunction data when running `gradient`.

One way forward would be to disallow certain module/method combinations or print warnings, which is tricky as `QC_MODULE` is not rigorously set either.
Currently I went with some warnings since users will want DF-MP2 densities from `dfmp2`.

hokru:
The scope of the PR changed. Took longer than expected, but hopefully does help to prevent users from getting a wrong FCHK file.

Ready for a first round of review/feedback, i think. I will still need to run auto-formatting.

psi4/driver/procrouting/proc.py
```diff
@@ -1814,6 +1814,7 @@ def run_dfocc_gradient(name, **kwargs):
         core.set_local_option('DFOCC', 'ORB_OPT', 'TRUE')
 
     core.set_global_option('DERTYPE', 'FIRST')
+    core.set_global_option('QC_MODULE', 'OCC')
```
hokru, line 1817: This seems to have been a bad idea as it breaks the default-pathway in the module selection for composite approaches. The default-pathways expect an empty `QC_MODULE` variable. Not sure how else to get module information back to the driver, though.

psi4/driver/driver.py
```diff
@@ -1938,17 +1938,111 @@ def fchk(wfn, filename):
...
+    if (qc_module in ['DFMP2']):
+        print('False')
+        fw.recompute_scf_density(False)
+    else:
+        fw.recompute_scf_density(True)
+
+    if 'DCT' in fchk_label:
+        # The Ca/Cb are the optimized orbitals from DCT. 
+        warn = " !! FCHK: In DCT calculations the SCF Density and DCT density entries are switched. !!"
+        core.print_out(warn)
```
JonathonMisiewicz, line 2032: Obligatory comment that I'm not happy about this, but fine. As discussed in Slack, the C and D fields need to be standardized.

psi4/driver/procrouting/proc.py
JonathonMisiewicz: Getting module information back to the driver this way via options is a bad idea. Consider the following example, which I can see myself doing:
```
ocepa_wfn = gradient("ocepa")
dct_wfn = gradient("dct")
fchk(ocepa_wfn, "ocepa")
fchk(dct_wfn, "dct")
```
Your approach cannot possibly determine the module correctly for both of those.

This needs to be determined from the wavefunction somehow. If there isn't a way to do this at present, the first idea I can think of for adding a way is to use the wavefunction name attribute, adding it in OCC if needed.

psi4/driver/driver.py
```diff
@@ -1938,17 +1938,111 @@ def fchk(wfn, filename):
...
     fw = core.FCHKWriter(wfn)
+
+    # At this point we don't the the method name, so we try to search for it.
+    # idea: get the method from the variable matching closely the 'current energy'
+    varlist = core.variables()
+    current = varlist['CURRENT ENERGY']
+    del varlist['CURRENT ENERGY']
+
+    # delete problematic entries
+    delete_list = ['CURRENT ENERGY', 'CURRENT GRADIENT', 'CURRENT REFERENCE ENERGY']
```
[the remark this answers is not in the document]
hokru: Stems from my inability to deal with a `psi4.core.Matrix` object in the dictionary. Didn't know `hasattr` can also check this.

psi4/driver/driver.py
```diff
@@ -1938,17 +1938,111 @@ def fchk(wfn, filename):
...
+    for key in delete_list[:]:
+        if key in varlist:
+            del varlist[key]
+
+    # find closest match, skip arrays
+    for (key, val) in varlist.items():
+        if hasattr(val, '__len__'): continue
```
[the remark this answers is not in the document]
hokru: It looks like all ENERGY variables are scalar currently, so `endswith` is a good idea. Best would be if one can check if the entry is a scalar, instead of excluding arrays, various psi4.core objects, etc.

psi4/driver/driver.py
hokru: I don't insist on enabling DCT. It's more a "one might as well since it's there".

psi4/driver/procrouting/proc.py
hokru: Perhaps one could make a `wfn.set_variable` equivalent for strings? Possibly a standard variable to set the parent module of a `wfn`? Could also be useful for gettings notes and names to a `wfn` object for users, instead of being limited to what the filename allows.

psi4/driver/driver.py
JonathonMisiewicz: Rewording: "I'm not happy that we need such a system of hacks to get DCT enabled at all."

psi4/driver/driver.py
```diff
@@ -1938,17 +1938,111 @@ def fchk(wfn, filename):
     -----
     * A description of the FCHK format is http://wild.life.nctu.edu.tw/~jsyu/compchem/g09/g09ur/f_formchk.htm
     * The allowed headers for methods are general and limited, i.e., "Total SCF|MP2|CI|CC Density",
-      so "CC" is always used for the post-HF case.
+      PSI4 will try to find the right one for the current calculation.
+    * Not all implementations in PSI4 are compatible with the FCHK writer. Currently available and tested are:
+      All SCF/DFT variants, DF-RHF-MP2 (dfmp2 module), CONV-RHF/UHF-CCSD (ccenergy module). No UHF-MP2 densities are available.
```
hokru, line 1943: Note to myself: This list is incomplete and needs rewording. All densities obtained from the `ccenergy` module work.

psi4/driver/driver.py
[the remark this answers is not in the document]
hokru: oh, using `core.scalar_variables()` directly will avoid much pain... (It's `core.` variables until the transition to `wfn.` variables is finished.)

psi4/driver/procrouting/proc.py
loriab: if you add QC_MODULE to the OptionsState at the start of the fn, it'll reset when the fn finishes

hokru: hmm, but that doesn't help with getting the information back to the driver since it resets.

Is `wfn.name()` used by any internal logic, or is it free to modify? Hard to search the code base for it.
At least for `occ` it is inconsistent (aka often missing) anyway. `dfocc` is better.

Something like `name = DF-MP2 [OCC]` could work. Best to avoid round brackets.

JonathonMisiewicz: `wfn.name` should be free to modify. When I gave DCT wavefunctions names, nothing broke.

psi4/driver/driver.py
```diff
@@ -2040,15 +2043,13 @@ def fchk(wfn, filename, debug=False):
...
-        print('False')
         fw.recompute_scf_density(False)
     else:
         fw.recompute_scf_density(True)
 
     if 'DCT' in fchk_label:
         # The Ca/Cb are the optimized orbitals from DCT. 
-        warn = " !! FCHK: In DCT calculations the SCF Density and DCT density entries are switched. !!"
-        core.print_out(warn)
+        core.print_out(" !! FCHKWriter: In DCT calculations the SCF Density and DCT density entries are switched. !!")
```
susilehtola, line 2052: Wouldn't it be better to switch these around in the fchk code?

psi4/driver/driver.py
```diff
@@ -1932,35 +1932,137 @@ def gdma(wfn, datafile=""):
...
+    # find closest matching energy
+    for (key, val) in varlist.items():
+        if (np.isclose(val, current,1e-10)):
+            method = key
+
+    # The labels this functions knows. Extend if needed.
+    labels = {
+    "HF" :" SCF Density",
+    "SCF":" SCF Density",
+    "DFT":" SCF Density",
```
susilehtola, line 2026: ROHF is missing, right?

susilehtola: CUHF, too.

psi4/driver/procrouting/proc.py
```diff
@@ -2441,7 +2442,7 @@ def run_dfmp2_gradient(name, **kwargs):
         ['DF_BASIS_SCF'],
         ['DF_BASIS_MP2'],
         ['SCF_TYPE'])
-
+
```
susilehtola, line 2445: Run clang-format

psi4/src/export_mints.cc
```diff
@@ -1542,7 +1542,10 @@ void export_mints(py::module& m) {
                                                         "Extracts information from a wavefunction object, \
                                                                           and writes it to an FCHK file")
         .def(py::init<std::shared_ptr<Wavefunction>>())
-        .def("write", &FCHKWriter::write, "Write wavefunction information to file", "filename"_a);
+        .def("write", &FCHKWriter::write, "Write wavefunction information to file", "filename"_a)
+        .def("SCF_Dtot",&FCHKWriter::SCF_Dtot,py::return_value_policy::reference_internal)
+        .def("recompute_scf_density", &FCHKWriter::recompute_scf_density, "Re-compute the SCF density from wfn.Ca/Cb if True, or use SCF density from reference_wavefunction if False")
+        .def("set_postscf_density_label", &FCHKWriter::set_pHF_density_label, "Set base label for post-SCF density, e.g. ' CC Density'.", "label"_a);
```
susilehtola, line 1548: Why do two out of four string constants have a `_a` suffix? What does it even do?

psi4/src/psi4/libmints/writer.cc
```diff
@@ -322,20 +328,37 @@ void FCHKWriter::write(const std::string &filename) {
...
+    int nalpha = wavefunction_->nalpha();
+    int nbeta = wavefunction_->nbeta();
+    int natoms = mol->natom();
+    int nprimitive = basis->nprimitive();
+
     // SCF density matrices
-    SharedMatrix Da_ao;
-    SharedMatrix Db_ao;
+    auto Da_ao = std::make_shared<Matrix>("Da(AO)", nmo, nmo);
+    auto Db_ao = std::make_shared<Matrix>("Db(AO)", nmo, nmo);
```
susilehtola, line 342: Pretty sure these should be `nbf x nbf`

psi4/src/export_mints.cc
hokru: It's a shorthand to inform about the name of the argument:
https://pybind11.readthedocs.io/en/stable/basics.html#keyword-arguments

And for `recompute_scf_density` I missed to add an argument name.

psi4/driver/procrouting/proc.py
hokru: there shouldn't be any changes in proc.py at all :(

psi4/driver/driver.py
hokru: Not sure what you mean. As a label? All `reference` options use the same `SCF TOTAL ENERGY` variable for the detection.
The "HF" line is currently not used but also does not hurt to have in case `UHF TOTAL ENERGY`or something like that is introduced.

psi4/driver/driver.py
hokru: maybe? Not sure about what approach would be good. Shouldn't be too intrusive.
This patch is a short-term solution and once the underlying issues are solved, this whole thing needs to be re-written again.

psi4/src/psi4/libmints/writer.cc
hokru: Yes, I think you are right.

dgasmith:
To kick this along, is there anything else that needs to be done here?

psi4/src/psi4/libmints/writer.cc
hokru: changed accordingly.

psi4/src/export_mints.cc
hokru: added missing argument name

psi4/driver/driver.py
hokru: Added with latest commit a python-based fix for DCT. Not 100% sure this is a good idea. Else I would maybe turn of DCT.

hokru:
Thanks for reminding me.
Now freshly rebased and with a python-patch for DCT. UHF-CCSD re-enabled in pytest as well. 
Ready for review.

psi4/driver/driver.py
```diff
@@ -1932,35 +1932,164 @@ def gdma(wfn, datafile=""):
...
+
+    # handle special cases
+    if (module_ == 'CCENERGY'):
+        fw.recompute_scf_density(True)
+    elif (module_ == 'DCT'):
+        # tasks:
+        # i)   compute DCT density from wfn.C coefficients and write result to wfn.D
+        # ii)  overwrite wfn.C with SCF orbitals from reference wfn.
+        # iii) let FCHKWriter recompute SCF density
+        wfn_ = wfn
```
JonathonMisiewicz, line 2012: In Python, this doesn't make `wfn_` a _copy_ of `wfn`. This makes `wfn_` another name for the same object. So when you mutate `wfn_`, you are changing the input `wfn`. Which is bad. I think you want the cloning method for the Wavefunction.

hokru: right, I had a feeling this isn't right in python. I meant to clone and preserve the original `wfn`.

psi4/driver/driver.py
```diff
@@ -1932,35 +1932,165 @@ def gdma(wfn, datafile=""):
...
+    module_ = wfn.module().upper()
+    if (module_ not in allowed):
+        core.print_out(f"Theory module {module_} is currently not supported by the FCHK writer.")
+        return None
+
+    # handle special cases
+    if (module_ == 'CCENERGY'):
+        fw = core.FCHKWriter(wfn)
+        fw.recompute_scf_density(True)
+    elif (module_ == 'DCT'):
```
JonathonMisiewicz, line 2006: Looking at the code, I don't understand how DCT FCHK works. Just to confirm, have you tested DCT FCHK? I don't see that anyway in your tests. The reasons for my skepticism are:
1. You're re-computing the SCF density and then writing that back as the density matrix of `wfn_`, not its reference wavefunction. How does the correlated density matrix get passed to the FCHKWriter?
2. The orbitals you're using to compute the reference density come from the input wavefunction's `Ca` and `Cb`, not the reference wavefunction's. Are those the same? I would have thought that the input wavefunction's orbitals would have been the optimized orbitals computed during the iterations.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -333,9 +348,17 @@ void FCHKWriter::write(const std::string &filename) {
     bool pHF = (refwf != NULL);
 
     if (pHF) {
-        // Level of theory may be correlated
-        Da_ao = refwf->Da_subset("AO");
-        Db_ao = refwf->Db_subset("AO");
+        // regenerate SCF densities from coefficients matrix
+        // since one currently cannot be sure Da_subset("AO")
```
JonathonMisiewicz, line 352: Wait, what? The reference wavefunction's density matrices aren't actually the reference densities? If you can give me an example, I can go ahead and fix this for DCT. No guarantees about my ability to fix this for `ccenergy`, but I can at least give a look.

hokru: For DCT i think `refwfn.D` is correct (=SCF), but for `ccenergy` it is the CC density.

JonathonMisiewicz: Then why is it that for DCT, we [send in `recompute_` true](https://github.com/psi4/psi4/pull/1884/files#diff-c4d6e39b37f0195e3587302260658d00R2039)? The `else` clause should do what we want for DCT in that case, right?

hokru: hmm, oversight from me. It's no longer needed. I checked and setting it to `False` gives the same fchk file.

psi4/driver/driver.py
hokru: I tested by comparing against a ccsd fchk file, though CC and DCT density are different, they are both different to the SCF density. I can add a regression test. Not sure if I can get a DCT density matrix out of the `dct` module?

1. `wfn_.Da().np[:] = Da` writes the new density to `wfn_` and from there FCHKWriter takes it.
2. I calculate the correlated density from the input wfn Ca/Cbs, which are the optimized DCT orbitals.

The writer needs `wfn.D()` be the DCT density, but it's the SCF density on input. I dont think orbital-optimized methods are covered in FCHK format and the orbital coefficient field needs come from the SCF. So there is this confusing switching around happening.

JonathonMisiewicz: Okay, I see.

That formula for going from the orbitals to the density is correct for mean-field theories (HF and DFT), but **wrong** for anything else. What you need is the DCT one-particle density matrix (OPDM) available on the wavefunction. I can get a PR fixing that (for any ODC-06, ODC-12, and ODC-13, and also DC-06 with a gradient computation) tomorrow. I'd need it anyways for Gradient Refactor. Getting OPDM for any other DCT methods would require deriving analytic gradient expressions, and that isn't worthwhile.

hokru: right! Clearly not paying enough attention to all this :( 
If you find no time for a PR I can also disable DCT for now.

JonathonMisiewicz: Unless a miracle happens, this is actually the perfect time for me to work on a PR like this.

JonathonMisiewicz:
Between my last three PRs, there shouldn't be a need for these strange hacks.

Ping me if you encounter anything strange in the densities.

hokru:
> Between my last three PRs, there shouldn't be a need for these strange hacks

Yes, many thanks for that!

hokru:
I think this is again in a good shape and also writes now the correct orbital coeff/energies in the FCHK sense, meaning SCF ones.
Please check the updated main post.

I am not 100% sure what I did with the `Vector.copy()` is fully correct, please have a close look. It works for my use case.
(https://github.com/psi4/psi4/pull/1884/files#diff-ce944be82d781507cd2bfa56fdb26af1R323)

susilehtola:
> I think this is again in a good shape and also writes now the correct orbital coeff/energies in the FCHK sense, meaning SCF ones.
> Please check the updated main post.

What do you mean? If you run OO-CC, does the checkpoint file contain the original HF orbitals, or the OO-CC orbitals?

hokru:
The FCHK file will contain SCF orbitals for OO-CC. As far as I know this is what the format expects. Happy to be convinced otherwise.

susilehtola:
What are "SCF" orbitals?

susilehtola:
There is no FCHK standard - the orbitals should be whatever the method gives. Even in TCTSNBN you can modify the orbitals in the checkpoint file e.g. with `Pop=SaveNBOs`, `Pop=SaveNLMOs` or `Pop=SaveMixed`...

hokru:
If that is what users will expect then I will change it back so one gets the DCT and OO-MP/CC orbital data.

JonathonMisiewicz:
Warning that orbital energies then become ambiguous as well. Do you want the "orbital energies" at the correlated level of theory or the reference?

If there's no standard either way, then it may be best to let the user toggle which one they want in this corner case.

susilehtola:
> Warning that orbital energies then become ambiguous as well. Do you want the "orbital energies" at the correlated level of theory or the reference?
> 
> If there's no standard either way, then it may be best to let the user toggle which one they want in this corner case.

I don't know what behavior is expected; energy expectation value might be one option but there might be different definitions for the Fock operator. Most of all, you can't reorder the orbitals...

hokru:
> > Warning that orbital energies then become ambiguous as well. Do you want the "orbital energies" at the correlated level of theory or the reference?
> > If there's no standard either way, then it may be best to let the user toggle which one they want in this corner case.
> 
> I don't know what behavior is expected; energy expectation value might be one option but there might be different definitions for the Fock operator. Most of all, you can't reorder the orbitals...

This seems like a difficult issue to get right.
At this point I'd prefer to give a warning in the manual and output about the orbital energy field and print whatever the respective theory module puts there.

psi4/driver/driver.py
```diff
@@ -1931,37 +1931,121 @@ def gdma(wfn, datafile=""):
     if not datafile:
         os.remove(commands)
 
-
-def fchk(wfn, filename):
+def fchk(wfn, filename, debug=False):
```
JonathonMisiewicz, line 1934: `NBOWriter` currently is a wavefunction method, but `fchk` and `molden` are top-level driver functions. Should we try to make these consistent, and if so, how? All we need for this PR is to decide if we want to make all of these wavefunction methods.

psi4/driver/driver.py
```diff
@@ -1931,37 +1931,121 @@ def gdma(wfn, datafile=""):
...
+            the 'Orbital Energy' field contains ambiguous data.""")
+
+    # At this point we don't know the method name, so we try to search for it.
+    # idea: get the method from the variable matching closely the 'current energy'
+    varlist = core.scalar_variables()
+    current = varlist['CURRENT ENERGY']
+
+    # delete problematic entries
+    delete_list = ['CURRENT ENERGY', 'CURRENT REFERENCE ENERGY']
+    for key in delete_list[:]:
```
JonathonMisiewicz, line 2008: No [:] needed here.

psi4/driver/driver.py
```diff
@@ -1931,37 +1931,121 @@ def gdma(wfn, datafile=""):
...
+        "MP3": " MP3 Density",
+        "MP4": " MP4 Density",
+        "CI": " CI Density",
+        "CC": " CC Density",
+        "DCT": " DCT Density",
+    }
+    # assign label from method name
+    fchk_label = 'UNKNOWN'
+    for key in labels:
+        if key in method:
```
JonathonMisiewicz, line 2032: Run a `dfocc` OMP2.5 or OLCCD and see what happens. OLCCD won't be recognized, and OMP2.5 will get mistaken for OMP2. This is the big problem.

susilehtola: But that's the great fun of the fchkpt format: TCTSNBN doesn't have these methods, and that's why the standard doesn't have such entries either. So, the choice is either to throw an error in such cases (comply to standard), or write something else than what the header says (status quo which the PR is trying to change)... :rofl:

hokru: From an old forum post (http://forum.psicode.org/t/fchkwriter-writing-wrongly-formatted-fchk-files/1082) I am under the impression one needs to stick to certain labels to not break 3rd-party programs.

Thus I went with putting "MP2 Density" and "CC Density" for the cases you mention. "DCT Density" admittedly breaks with this idea. The whole fchk formatting is just somewhat.. unfortunate.

psi4/driver/driver.py
loriab: I'm guessing we want to make them all wavefunction methods (attached to the class py-side if need be) as they're proliferating (fchk, molden, nbo, qcsk wfn). But `fchk(wfn)` and `molden(wfn)` don't get removed for years, if ever. No strong feelings, though.

hokru:
might as well add fix for https://github.com/psi4/psi4/issues/2021

psi4/driver/driver.py
PeterKraus: This might be too much to ask, but how about having a "strict" `fchk` writer (default) which would strictly conform to format and always produce valid `fchk` files (ie. SCF, MP2, MP3, MP4, CC and CI, if available, only) and add a permissive tag which would print things like DCT, OMP2.5 with non-standard headings and let users handle the renaming of labels if it breaks 3rd party programs?

hokru: That sounds like a good compromise and is a simple change.

psi4/driver/driver.py
```diff
@@ -1931,37 +1931,135 @@ def gdma(wfn, datafile=""):
...
+        "CI": " CI Density",
+        "CC": " CC Density",
+    }
+    # assign label from method name
+    fchk_label = f" {method} Density"
+    if strict_label:
+        in_list = False
+        for key in allowed_labels:
+            if key in method:
+                fchk_label = allowed_labels[key]
```
PeterKraus, line 2045: How about a warning in case `key in method and key is not method`:
`core.print_out(f"FCHKWriter: !WARNING! {method} renamed to {key}.\n")`

psi4/src/export_mints.cc
```diff
@@ -1544,7 +1546,9 @@ void export_mints(py::module& m) {
                                                         "Extracts information from a wavefunction object, \
                                                                           and writes it to an FCHK file")
         .def(py::init<std::shared_ptr<Wavefunction>>())
-        .def("write", &FCHKWriter::write, "Write wavefunction information to file", "filename"_a);
+        .def("write", &FCHKWriter::write, "Write wavefunction information to file", "filename"_a)
+        .def("SCF_Dtot",&FCHKWriter::SCF_Dtot,py::return_value_policy::reference_internal)
+        .def("set_postscf_density_label", &FCHKWriter::set_pHF_density_label, "Set base label for post-SCF density, e.g. ' CC Density'.", "label"_a);
```
PeterKraus, line 1551: I'm not sure what's happening with the semicolons here.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -363,30 +375,18 @@ void FCHKWriter::write(const std::string &filename) {
...
-        auto intmass = static_cast<int>(mass);
-        atomic_weights.push_back(mass);
-        int_atomic_weights.push_back(intmass);
-        nuc_charges.push_back(Z);
-        atomic_numbers.push_back(intZ);
+        int intZ = static_cast<int>(mol->Z(atom));
+        atomic_weights.push_back(mol->mass(atom));
+        int_atomic_weights.push_back(mol->mass_number(atom));
+        nuc_charges.push_back(mol->Z(atom));
+        atomic_numbers.push_back(intZ > 0 ? mol->true_atomic_number(atom) : intZ); // care about ECP & ghosts!
```
PeterKraus, line 389: Neat.

psi4/driver/driver.py
```diff
@@ -1931,37 +1931,135 @@ def gdma(wfn, datafile=""):
...
+                in_list = True
+        if not in_list:
+            core.print_out(f"FCHKWriter: !WARNING! {method} is not recognized. Using non-standard label.\n")
+    core.print_out(f"FCHKWriter: Writing {filename} with label '{fchk_label}'.\n")
+    fw.set_postscf_density_label(fchk_label)
 
+    fw.write(filename)
+    # needed for the pytest. The SCF density below follows PSI4 ordering not FCHK ordering.
```
PeterKraus, line 2053: I think the test should ideally test the files as written, to make sure the format is also right.

psi4/src/export_mints.cc
hokru: The semicolon is needed at the end of the class.

psi4/driver/driver.py
hokru: good idea

psi4/driver/driver.py
hokru: You mean like a line-by-line comparison and check formatting? That works for integer fields but not floats.
Not sure how one would go best about this and if the effort is really needed.

hokru: Added a comparison against stored files in the pytest using a comparison function added to psi4. The function should be robust to fchk format variants but I have no way of testing it extensively. For PSI4 FCHK files it should work fine.

PeterKraus (approved):
Looks great!

loriab (commented):
lgtm, though I want to read again

psi4/driver/p4util/fchk.py
```diff
@@ -0,0 +1,128 @@
...
+# @END LICENSE
+#
+"""Module with utility functions for FCHK files."""
+
+import numpy as np
+from psi4.driver.p4util.testing import compare_strings, compare_arrays, compare_values, compare_integers
+from psi4 import core
+from .exceptions import ValidationError
+
+__all__ = ['fchkfile_to_string','compare_fchkfiles']
```
PeterKraus, line 35: Is this necessary? Or is it to stop some import clash elsewhere?

hokru: Not having it does not break anything but LGTM complained about polluting the namespace. These are the only two functions that should be called directly by users.

PeterKraus: I see, so it's good practice to declare `__all__` whenever the module is imported via `from module import *`. You learn something every day!

loriab (approved):
Investigation into and method specific warnings are great. And tests will help keep this in line. Thanks!

dgasmith (approved):
LGTM, thanks for wading through this format!

PR #1885, opened 2020-05-08
TDSCF psithon interface

jeffschriber:
## Description
Allows TDSCF code to be called from any SCF (HF,DFT) computation. 

## Todos
- [X] Implement functions in proc.py
- [X] Add options handling
- [x] Add tests

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -2069,11 +2069,7 @@ def molden(wfn, filename=None, density_a=None, density_b=None, dovirtual=None):
         mw.write(filename, wfn.Ca(), wfn.Cb(), wfn.epsilon_a(), wfn.epsilon_b(), occa, occb, dovirt)
 
 def tdscf(name=None, **kwargs):
-    if name is not None:
-        e, wfn = energy(name, return_wfn=True)
-        return proc.run_tdscf_energy(name = name,ref_wfn = wfn, **kwargs)
-    else:
-        return proc.run_tdscf_energy(name = name,**kwargs)
+    return proc.run_tdscf_energy(name = name,**kwargs)
 
 # Aliases
 opt = optimize
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -2068,6 +2068,12 @@ def molden(wfn, filename=None, density_a=None, density_b=None, dovirtual=None):
         mw = core.MoldenWriter(wfn)
         mw.write(filename, wfn.Ca(), wfn.Cb(), wfn.epsilon_a(), wfn.epsilon_b(), occa, occb, dovirt)
 
+def tdscf(name=None, **kwargs):
+    if name is not None:
+        e, wfn = energy(name, return_wfn=True)
+        return proc.run_tdscf_energy(name = name,ref_wfn = wfn, **kwargs)
+    else:
+        return proc.run_tdscf_energy(name = name,**kwargs)
 
 # Aliases
 opt = optimize
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -2068,8 +2068,8 @@ def molden(wfn, filename=None, density_a=None, density_b=None, dovirtual=None):
         mw = core.MoldenWriter(wfn)
         mw.write(filename, wfn.Ca(), wfn.Cb(), wfn.epsilon_a(), wfn.epsilon_b(), occa, occb, dovirt)
 
-def tdscf(name=None, **kwargs):
-    return proc.run_tdscf_energy(name = name,**kwargs)
+def tdscf(wfn, **kwargs):
+    return proc.run_tdscf_excitations(wfn,**kwargs)
 
 # Aliases
 opt = optimize
--- a/psi4/driver/p4util/util.py
+++ b/psi4/driver/p4util/util.py
@@ -36,9 +36,6 @@ from psi4 import core
 from psi4.driver.procrouting import *
 from .exceptions import ValidationError
 
-def tdscf(wfn,**kwargs):
-    return proc.run_tdscf_energy(ref_wfn = wfn)        
-
 def oeprop(wfn, *args, **kwargs):
     """Evaluate one-electron properties.
 
--- a/psi4/driver/p4util/util.py
+++ b/psi4/driver/p4util/util.py
@@ -33,8 +33,11 @@ import sys
 import warnings
 
 from psi4 import core
+from psi4.driver.procrouting import *
 from .exceptions import ValidationError
 
+def tdscf(wfn,**kwargs):
+    return proc.run_tdscf_energy(ref_wfn = wfn)        
 
 def oeprop(wfn, *args, **kwargs):
     """Evaluate one-electron properties.
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -2732,6 +2732,7 @@ def run_tdscf_energy(**kwargs):
 
     ssuper_name = ref_wfn.functional().name()
 
+    # Do we need this return value? 
     ret = response.scf_response.tdscf_excitations(ref_wfn, states_per_irrep = states,
                                                   triplet = core.get_global_option("TDSCF_TRIPLETS"),
                                                   tda = core.get_global_option("TDSCF_TDA"),
@@ -2739,8 +2740,6 @@ def run_tdscf_energy(**kwargs):
                                                   r_tol = core.get_global_option("TDSCF_R_TOL"),
                                                   guess = core.get_global_option("TDSCF_GUESS"))
 
-    ref_wfn.set_array_variable(f"TD-{ssuper_name} EXCITATION ENERGIES", core.Matrix.from_array(np.asarray(ret)[:,0].reshape(-1,1)))
-    ref_wfn.set_array_variable(f"TD-{ssuper_name} SYMMETRY LABELS", core.Matrix.from_array(np.asarray(ret)[:,1].reshape(-1,1)))
 
     return ref_wfn
     
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -2714,7 +2714,34 @@ def run_bccd(name, **kwargs):
     optstash.restore()
     return ref_wfn
 
-def run_tdscf_energy(name=None,**kwargs):
+def run_tdscf_excitations(wfn,**kwargs):
+
+    states = core.get_option("SCF","TDSCF_STATES_PER_IRREP")
+
+    # some sanity checks
+    if sum(states) == 0:
+        raise ValidationError("TDSCF: No states requested in TDSCF_STATES_PER_IRREP") 
+
+    if len(states) != wfn.nirrep():
+        raise ValidationError("TDSCF: Requested TDSCF_STATES_PER_IRREP do not match wave function symmetry") 
+
+    ssuper_name = wfn.functional().name()
+
+    # Do we need this return value? 
+    ret = response.scf_response.tdscf_excitations(wfn, states_per_irrep = states,
+                                                  triplet = core.get_option("SCF","TDSCF_TRIPLETS"),
+                                                  tda = core.get_option("SCF","TDSCF_TDA"),
+                                                  e_tol = core.get_option("SCF","TDSCF_E_TOL"),
+                                                  r_tol = core.get_option("SCF","TDSCF_R_TOL"),
+                                                  guess = core.get_option("SCF","TDSCF_GUESS"))
+
+    # Shove variables into global space
+    for k, v in wfn.variables().items():
+        core.set_variable(k, v)
+
+    return wfn
+
+def run_tdscf_energy(name,**kwargs):
 
     # Get a wfn in case we aren't given one
     ref_wfn = kwargs.get('ref_wfn', None)
@@ -2725,31 +2752,7 @@ def run_tdscf_energy(name=None,**kwargs):
         else:
             ref_wfn = run_scf(name.strip('td-'), **kwargs)
 
-    states = core.get_option("SCF","TDSCF_STATES_PER_IRREP")
-
-    # some sanity checks
-    if sum(states) == 0:
-        raise ValidationError("TDSCF: No states requested in TDSCF_STATES_PER_IRREP") 
-
-    if len(states) != ref_wfn.nirrep():
-        raise ValidationError("TDSCF: Requested TDSCF_STATES_PER_IRREP do not match wave function symmetry") 
-
-    ssuper_name = ref_wfn.functional().name()
-
-    # Do we need this return value? 
-    ret = response.scf_response.tdscf_excitations(ref_wfn, states_per_irrep = states,
-                                                  triplet = core.get_option("SCF","TDSCF_TRIPLETS"),
-                                                  tda = core.get_option("SCF","TDSCF_TDA"),
-                                                  e_tol = core.get_option("SCF","TDSCF_E_TOL"),
-                                                  r_tol = core.get_option("SCF","TDSCF_R_TOL"),
-                                                  guess = core.get_option("SCF","TDSCF_GUESS"))
-
-    # Shove variables into global space
-    for k, v in ref_wfn.variables().items():
-        core.set_variable(k, v)
-
-    return ref_wfn
-    
+    return run_tdscf_excitations(ref_wfn, **kwargs)
 
 def run_scf_property(name, **kwargs):
     """Function encoding sequence of PSI module calls for
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -2316,7 +2316,6 @@ def run_scf(name, **kwargs):
     # optionally get excited states
     if do_tdscf:
         excited_states = run_tdscf_energy(name, ref_wfn=scf_wfn)
-        scf_wfn.set_array_variable("TDSCF EXCITATION ENERGIES", core.Matrix.from_array(np.asarray(excited_states)))
             
     optstash_scf.restore()
     optstash_mp2.restore()
@@ -2723,10 +2722,12 @@ def run_bccd(name, **kwargs):
     return ref_wfn
 
 def run_tdscf_energy(name, **kwargs):
+#def run_tdscf_energy(**kwargs):
 
     # Get a wfn in case we aren't given one
     ref_wfn = kwargs.get('ref_wfn', None)
     if ref_wfn is None:
+        name = kwargs.get('name',None)
         ref_wfn = scf_helper(name, **kwargs)
 
     states = core.get_global_option("TDSCF_STATES_PER_IRREP")
@@ -2737,14 +2738,18 @@ def run_tdscf_energy(name, **kwargs):
 
     if len(states) != ref_wfn.nirrep():
         raise ValidationError("TDSCF: Requested TDSCF_STATES_PER_IRREP do not match wave function symmetry") 
-         
+
+    ssuper_name = ref_wfn.functional().name()
 
     ret = response.scf_response.tdscf_excitations(ref_wfn, states_per_irrep = states,
-                                                  triplets = core.get_global_option("TDSCF_TRIPLETS"),
+                                                  triplet = core.get_global_option("TDSCF_TRIPLETS"),
                                                   tda = core.get_global_option("TDSCF_TDA"),
                                                   e_tol = core.get_global_option("TDSCF_E_TOL"),
                                                   r_tol = core.get_global_option("TDSCF_R_TOL"),
                                                   guess = core.get_global_option("TDSCF_GUESS"))
+
+    ref_wfn.set_array_variable(f"TD-{ssuper_name} EXCITATION ENERGIES", core.Matrix.from_array(np.asarray(ret)[:,0].reshape(-1,1)))
+    ref_wfn.set_array_variable(f"TD-{ssuper_name} SYMMETRY LABELS", core.Matrix.from_array(np.asarray(ret)[:,1].reshape(-1,1)))
 
     return ret
     
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -2720,7 +2720,10 @@ def run_tdscf_energy(name=None,**kwargs):
     ref_wfn = kwargs.get('ref_wfn', None)
 
     if ref_wfn is None:
-        raise ValidationError("TDSCF: No reference wave function!")
+        if name is None:
+            raise ValidationError("TDSCF: No reference wave function!")
+        else:
+            ref_wfn = run_scf(name.strip('td-'), **kwargs)
 
     states = core.get_option("SCF","TDSCF_STATES_PER_IRREP")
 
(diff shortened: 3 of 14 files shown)


dgasmith:
This looks great, anything we need to start working on in the review?

jeffschriber:
I think this is mostly good to go, I'll add a test or two soon. I would like a little feedback on if there's a better way to store the excitation energies. I kind of just throw them in Wavefunction one at a time. I'm sure all of that will change anyway with @robertodr 's changes.

dgasmith:
I was originally thinking the interface would be more like:
```
tddft("B3LYP/cc-pVDZ")

or 

ref_e, ref_wfn = psi4.energy("B3LYP/cc-pVDZ", return_wfn=True)
tddft(ref_wfn)
```

I would be curious as to others thoughts on this.

hokru:
I'd prefer sticking to the `energy` call though I can see the appeal of using `tddft(ref_wfn)`.

`energy("TD-B3LYP/c-pVDZ")` and `energy("TDA-B3LYP/cc-pVDZ")` would be cool.

psi4/driver/procrouting/proc.py
```diff
@@ -2291,6 +2297,17 @@ def run_scf(name, **kwargs):
...
+
+    # optionally get excited states
+    if do_tdscf:
+        excited_states = run_tdscf_energy(name, ref_wfn=scf_wfn)
+    
+        # This way of storing energies and states seems ... not ideal
+        for n, ex_pair in enumerate(excited_states): 
+            ex_en, ex_state = ex_pair
+            scf_wfn.set_variable(f'STATE {n + 1} EXCITATION ENERGY', ex_en) 
+            scf_wfn.set_variable(f'STATE {n + 1} IRREP', ex_state)
```
hokru, line 2309: You could use `set_array_variable` like e.g. in dipole moments?

robertodr:
I'd prefer @dgasmith option. As for storage of excitations, I was thinking to modify the return value of the solver to a data structure collecting energy, response vectors, and transition moments for each excitation. `Wavefunction` could hold a dictionary of these, with labeling by irrep (`f"STATE {n} in IRREP {m}"`) or by energy (`f"STATE {n}"`) I think the latter could be more intuitive.

psi4/driver/procrouting/proc.py
jeffschriber: Agreed, that's a better option

psi4/src/read_options.cc
```diff
@@ -1576,6 +1576,29 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+            - none : No triplets computed
+            - also : lowest-energy triplets and singlets included
+            - only : Only triplet states computed
+             -*/
+        options.add_str("TDSCF_TRIPLETS", "none");
+        /*- Run with Tamm-Dancoff approximation, uses RPA when false -*/ 
+        options.add_bool("TDSCF_TDA", false);
+        /*- Convergence threshold for excitation energies -*/
+        // NOTE: This option may do nothing
+        options.add_double("TDSCF_E_TOL", 1E-6);
```
loriab, line 1594: Is this and below playing the `E/R_CONVERGENCE` roles? Need to have those names, then, with TD* prefix. In practice, should take TD_E/R_CONVERGENCE if has_changed, else E/R_CONVERGENCE values.

psi4/src/read_options.cc
```diff
@@ -1576,6 +1576,29 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        /*- Convergence threshold for the norm of the residual vector -*/
+        options.add_double("TDSCF_R_TOL", 1E-4);
+        /*- Guess type, only 'denominators' currently supported -*/
+        options.add_str("TDSCF_GUESS", "denominators");
+        /*- Max number of vectors to store before collapsing -*/
+        options.add_int("TDSCF_MAX_SS_VECTORS",50);
```
loriab, line 1600: what do others think about TDSCF or TDDFT wrt future-proofing? we'll want a term to cover both I think.

robertodr: TDSCF would be preferable. Though it could be confused with the explicit time propagation of the SCF state.

psi4/src/read_options.cc
robertodr: I plan on entirely phasing out the use of `e_tol` in the solver with #1814 Only `r_tol` should remain as the main convergence check. And the way convergence thresholds propagate should go the other way around: converging the residual of the response equations to 10^{-N} requires convergence of SCF to 10^{-(N+2)} (as for gradients, in practice).

There are other ways to tweak convergence of the response solver: either by starting with more trial vectors in the subspace (in practice ask for more roots than you actually need) or by allowing the subspace to become "arbitrarily" large during the iterations (`TDSCF_MAX_SS_VECTORS`)

robertodr:
As for tests, you can have a look at `tests/pytests/test_tdscf_excitations.py` and pick a couple of the combinations for which AJ already computed reference values.

jeffschriber:
Great thanks! I'll add the tests right away. 

And for `energy()` vs `tdscf()`, I don't feel too strongly either way. I like the simplicity of `energy()`, particularly since tdscf is only returning energies for now. `tdscf()` may be the way to go eventually since it is more flexible. One option is to have both work, easy to do since `energy()` is just calling a `tdscf()` function in `run_scf`.

psi4/src/read_options.cc
hokru: Language can be a bit community dependent. But TD-SCF is usually used to explicitly include both TD-HF and TD-DFT, I think.

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
jeffschriber: Should these be set here in `proc.py`, or does it make more sense for them to be set in the tdscf code itself? The latter would allow storage of more things in wfn (eg absolute energies of excited states) without requiring a complex return type from `tdscf_excitations`.

tests/tdscf-4/output.ref
```diff
@@ -0,0 +1,381 @@
...
+
+
+         ---------------------------------------------------------
+                         TDSCF excitation energies                
+                 by Andrew M. James and Daniel G. A. Smith        
+         ---------------------------------------------------------
+********************************************************************************
+**********                          WARNING                           **********
+**********      TDSCF is experimental results may be inaccurate       **********
+********************************************************************************
```
robertodr, line 335: Could you remove these lines of warning from the code?

psi4/src/read_options.cc
```diff
@@ -1586,7 +1586,9 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
             - also : lowest-energy triplets and singlets included
             - only : Only triplet states computed
              -*/
-        options.add_str("TDSCF_TRIPLETS", "none");
```
robertodr: Why the change?

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
robertodr: I agree it makes more sense to set these on `Wavefunction` directly in the TDSCF code.

psi4/src/read_options.cc
jeffschriber: I don't the the string is ever interpreted. In `tdscf_excitations`, the `triplet` variable is passed to `TDRSCCFEngine` as a bool. I like the idea of the none/some/all specifications but I couldn't tell if that was implemented.

robertodr: You're right. This is not properly in place yet.

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
jeffschriber: Done, I also like Lori's suggestion for the variable keys.

psi4/driver/p4util/util.py
```diff
@@ -33,8 +33,11 @@
 import warnings
 
 from psi4 import core
+from psi4.driver.procrouting import *
 from .exceptions import ValidationError
 
+def tdscf(wfn,**kwargs):
+    return proc.run_tdscf_energy(ref_wfn = wfn)
```
loriab, line 40: calling a high-level fn (from `proc`) from a menial file (like `util`) seems like risking a circular import now or in the future. I'd have no problem with a two-liner like this is driver.py or proc.py, wherever more convenient.

psi4/driver/procrouting/proc.py
```diff
@@ -2209,6 +2209,10 @@ def run_scf(name, **kwargs):
     if "dft_functional" in kwargs:
         dft_func = True
 
+    # See if we're doing TDSCF after, keep JK if so 
+    if sum(core.get_global_option("TDSCF_STATES_PER_IRREP")) > 0:
+        core.set_global_option("SAVE_JK", True)
```
loriab, line 2214: need to add to the `OptionsState` above via `['SCF', 'SAVE_JK']`. this resets the option when the run_scf exits, so that options look like how user set them, not with interference from driver. any reason not to do `get_option("SCF", "TDSCF_STATES_PER_IRREP")` and `set_local_option("SAVE_JK")`? probably as-is your routine wouldn't work if input was `set scf tdscf_states_per_irrep [1]`.

psi4/driver/procrouting/proc.py
```diff
@@ -2691,6 +2695,35 @@ def run_bccd(name, **kwargs):
...
     return ref_wfn
 
+def run_tdscf_energy(**kwargs):
+
+    # Get a wfn in case we aren't given one
+    ref_wfn = kwargs.get('ref_wfn', None)
+    if ref_wfn is None:
+        raise ValidationError("TDSCF: No reference wave function passed!")
+
+    states = core.get_global_option("TDSCF_STATES_PER_IRREP")
```
loriab, line 2705: get_option("SCF", "TDSCF...

psi4/driver/procrouting/proc.py
```diff
@@ -2691,6 +2695,35 @@ def run_bccd(name, **kwargs):
...
+
+    # some sanity checks
+    if sum(states) == 0:
+        raise ValidationError("TDSCF: No states requested in TDSCF_STATES_PER_IRREP") 
+
+    if len(states) != ref_wfn.nirrep():
+        raise ValidationError("TDSCF: Requested TDSCF_STATES_PER_IRREP do not match wave function symmetry")
```
loriab, line 2712: JCP must be proud of your spelling of "wavefunction" :-)

psi4/driver/procrouting/proc.py
```diff
@@ -2691,6 +2695,35 @@ def run_bccd(name, **kwargs):
...
+
+    if len(states) != ref_wfn.nirrep():
+        raise ValidationError("TDSCF: Requested TDSCF_STATES_PER_IRREP do not match wave function symmetry") 
+
+    ssuper_name = ref_wfn.functional().name()
+
+    # Do we need this return value? 
+    ret = response.scf_response.tdscf_excitations(ref_wfn, states_per_irrep = states,
+                                                  triplet = core.get_global_option("TDSCF_TRIPLETS"),
+                                                  tda = core.get_global_option("TDSCF_TDA"),
```
loriab, line 2719: get_option for the triplets, tda, guess. E/R_tol are still in negotiations, I guess.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -261,11 +261,6 @@ def _print_tdscf_header(**options):
                    '         {:^57}\n'.format('by Andrew M. James and Daniel G. A. Smith') +
                    '         ---------------------------------------------------------\n')
 
-    core.print_out("{}\n".format("*"*80) +
-                   "{}{:^60}{}\n".format("*"*10, "WARNING", "*"*10) +
-                   "{}{:^60}{}\n".format("*"*10, "TDSCF is experimental results may be inaccurate", "*"*10) +
-                   "{}\n".format("*"*80)) #yapf: disable
```
loriab, line 267: bold :-)

psi4/src/read_options.cc
```diff
@@ -1576,6 +1576,31 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_str("UHF_NOONS", "3");
         /*- Save the UHF NOs -*/
         options.add_bool("SAVE_UHF_NOS", false);
+
+        /*- SUB-SUBSECTION TDDFT -*/
```
loriab, line 1580: must it be a sub-sub section? ye olde perl script won't understand it.

tests/tdscf-2/input.dat
```diff
@@ -0,0 +1,38 @@
...
+
+molecule ch2 {
+0 3
+
+C           0.000000    0.000000    0.159693
+H          -0.000000    0.895527   -0.479080
+H          -0.000000   -0.895527   -0.479080
+no_reorient
+no_com
+symmetry c1
```
loriab, line 19: are the symm c1 and no_* required? should integrate into driver rather than demanding from user.

tests/tdscf-3/input.dat
```diff
@@ -0,0 +1,43 @@
...
+    reference rhf
+    scf_type pk
+    e_convergence 8
+    d_convergence 8
+    TDSCF_STATES_PER_IRREP [10]
+}
+
+
+e, wfn = energy('wb97x/cc-pvdz', return_wfn=True)
+tdscf(wfn)
```
loriab, line 38: be good to have a test that's one-line `energy('td-wb97x/cc-pvdz')`

tests/tdscf-3/input.dat
```diff
@@ -0,0 +1,43 @@
...
+
+for n, ref in enumerate(rhf_wB97X_RPA_cc_pvdz):
+    ex_en = wfn.variable(f"TD-wB97X ROOT 0 -> ROOT {n+1} EXCITATION ENERGY - A SYMMETRY")
```
loriab, line 41: check `psi4.variable(f"TD-w...`, too, while you're at it, please.

psi4/driver/procrouting/proc.py
```diff
@@ -2691,6 +2695,35 @@ def run_bccd(name, **kwargs):
...
+
+    # some sanity checks
+    if sum(states) == 0:
+        raise ValidationError("TDSCF: No states requested in TDSCF_STATES_PER_IRREP") 
+
+    if len(states) != ref_wfn.nirrep():
+        raise ValidationError("TDSCF: Requested TDSCF_STATES_PER_IRREP do not match wave function symmetry") 
+
+    ssuper_name = ref_wfn.functional().name()
+
```
hokru, line 2715: We need a way to detect double-hybrid functionals. Possibly `SuperFunctional.c_alpha > 0`.
As a reminder, Grimme/Neese suggested to replace the MP2 part in the ground state with CIS(D) in the excited state.

We need to decide if we still calculate the TD-DFT part or exist with a warning right away. To avoid misunderstandings, I'd suggest to exist before the calculation starts with a warning that excited states with double-hybrids is not implemented.

tests/tdscf-1/output.ref
```diff
@@ -0,0 +1,352 @@
...
+
+  ==> Options <==
+
+     etol      :              1e-05
+     rtol      :              0.0001
+     guess_type:              DENOMINATORS
+     restricted:              False
+     triplet   :              None
+     ptype     :              rpa
+
```
hokru, line 322: It would be good to have the printed options resemble the input options as closely as possible (or vice versa).
E.g. `r_tol` vs `rtol` (or `r_tolerance` if this is subject to change)  and in particular `pytpe` vs `tda True/False`. Though having the term `rpa` explicitly printed out is valuable for clarity.

psi4/driver/procrouting/response/scf_response.py
robertodr: The numbers are correct and have been tested against a couple of other codes, so no reason to warn about inaccuracy :)

tests/tdscf-2/input.dat
robertodr: Jeff noticed that the solver chokes with symmetry, though it shouldn't. I'm looking at it and hope to lift the restriction soon.

psi4/driver/procrouting/proc.py
robertodr: Exiting is the safest course of action, I think.

loriab: https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc_table.py#L246 is a model for DH check

tests/tdscf-2/input.dat
jeffschriber: no_* isn't required, initially took it from the pytest but it passes without. I'll remove them.

tests/tdscf-3/input.dat
jeffschriber: So I've got `tdscf(wfn)` and `tdscf('wb97x/cc-pvdz')` working. I don't have any handling of functionals with "td-" prefixes, nor is there any one-liner to get tdscf via an `energy` call (the one-liner is a `tddft()` call). Should I add this?

loriab: So of the three sequences below, I'm in favor of first and third and don't care for second.
* `e, wfn = energy(wb97x/cc-pvdz', return_wfn=True)  \n tdscf(wfn)`
* `tdscf('wb97x/cc-pvdz')`
* `energy('td-wb97x/cc-pvdz')`

In order to get the third, I'd go into proc_table.py and add to the `procedures` dict akin to how the DFT fctls themselves are added, so after any excluding logic for DH or whatever, `procedures['energy']['td-' + key] = proc.run_tdscf` . And then in proc.py, define `def run_tdscf` as essentially the `run_scf()` followed by tdscf(wfn)`

psi4/driver/procrouting/proc.py
```diff
@@ -2691,6 +2695,42 @@ def run_bccd(name, **kwargs):
     optstash.restore()
     return ref_wfn
 
+def run_tdscf_energy(name=None,**kwargs):
```
loriab, line 2698: now `e, wfn = energy('pbe', return_wfn=True) \n tddft(wfn)` doesn't work. do people agree that that's what we want for the piecewise access rather than `tddft(ref_wfn=wfn)`?

Maybe this needs two very short functions and one workhorse fn that both call. The inconsistent signatures with `name=None` here and in tddft() make me a little uneasy.

tests/tdscf-3/input.dat
dgasmith: I'm not a huge fan of the third. I think its overloading `energy` too much. I think the first two are sufficient.

How do you handle `tdscf(wfn)` with regard to JK saving?

hokru: Not wanting to overload `energy` too much is a fair point.
Although a `TD-` prefix would go along with the `EOM-` prefix we already have for CCSD, CC2/3. Also how would we want to run a TD-DFT gradient in the future? If we would do `gradient('td-pbe')` then `energy('td-pbe')` seems more natural.

jeffschriber: The `SAVE_JK` option is set to true in `run_scf` if the `TDSCF_STATES_PER_IRREP` option is set to compute any number of states. I suppose this would break if the states option was set after the initial scf call.

loriab: I think we must have the third. Most users have no notion how tddft internally different from cis or eom-ccsd, and it would seem strange to them that there's a psi method not accessed through `energy()`. Good point that we have mp2 defined as a fctl -- if it wasn't already excluded from `td-`ing by being a DH, I agree it should be manually removed.

psi4/driver/procrouting/proc_table.py
```diff
@@ -245,6 +245,7 @@
     # Gradients
     if not (ssuper.is_c_hybrid() or ssuper.is_c_lrc() or ssuper.needs_vv10()):
         procedures['gradient'][key] = proc.run_scf_gradient
+        procedures['energy']['td-' + key] = proc.run_tdscf_energy
```
loriab, line 248: just in case ppl pay more attention to the comment than the key, maybe move this out of `# Gradients`.

psi4/src/read_options.cc
loriab: I think kill off `TDSCF_E_TOL` and replace `TDSCF_R_TOL` with `TDSCF_R_CONVERGENCE`.

psi4/src/read_options.cc
```diff
@@ -1576,6 +1576,30 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        /*- Get the number of desired states per irrep -*/
+        options.add("TDSCF_STATES_PER_IRREP", new ArrayType());
+        /*- Controls inclusion of triplet states, valid options:
+            - none : No triplets computed
+            - also : lowest-energy triplets and singlets included
+            - only : Only triplet states computed
+             -*/
+        //options.add_str("TDSCF_TRIPLETS", "none");
+        //*- Computes spin-adapted products for triplets -*/
+        options.add_bool("TDSCF_TRIPLETS", false);
```
loriab, line 1590: this becomes `add_str` with `NONE ONLY ALSO` depending on result of https://github.com/psi4/psi4/pull/1814/files#diff-341c62f24f732e5bdaaef3c16b0fa970R270

psi4/src/read_options.cc
```diff
@@ -1576,6 +1576,30 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        /*- Run with Tamm-Dancoff approximation, uses RPA when false -*/ 
+        options.add_bool("TDSCF_TDA", false);
+        /*- Convergence threshold for excitation energies -*/
+        // NOTE: This option may do nothing
+        options.add_double("TDSCF_E_TOL", 1E-6);
+        /*- Convergence threshold for the norm of the residual vector -*/
+        options.add_double("TDSCF_R_TOL", 1E-4);
+        /*- Guess type, only 'denominators' currently supported -*/
+        options.add_str("TDSCF_GUESS", "denominators");
```
loriab, line 1599: uppercase denominators unless there's something special I'm not seeing. psi is only case insensitive for users, not for programmers :-)

tests/tdscf-2/input.dat
```diff
@@ -0,0 +1,36 @@
...
+    e_convergence 8
+    d_convergence 8
+    TDSCF_E_TOL 5
+    TDSCF_STATES_PER_IRREP [9]
+    TDSCF_TRIPLETS true
+    TDSCF_TDA true
+}
+
+e, wfn = energy('hf/cc-pvdz', return_wfn=True)
+tdscf(ref_wfn = wfn)
```
loriab, line 32: in my notions, this (and in `tdscf-3`) would be plain `tdscf(wfn)`

tests/tdscf-1/input.dat
```diff
@@ -0,0 +1,37 @@
+UHF_RPA_cc_pvdz = [0.2445704160468683,
+                   0.2878574429978692,
```
loriab, line 2: all the tests need entering in tests/CM, each test needs its own e.g., tests/tdscf-N/CM, and a quick first line `#! quick description of test` for each would be appreciated.

psi4/driver/driver.py
```diff
@@ -2062,6 +2062,8 @@ def molden(wfn, filename=None, density_a=None, density_b=None, dovirtual=None):
         mw = core.MoldenWriter(wfn)
         mw.write(filename, wfn.Ca(), wfn.Cb(), wfn.epsilon_a(), wfn.epsilon_b(), occa, occb, dovirt)
 
+def tdscf(name=None, **kwargs):
+    return proc.run_tdscf_energy(name = name,**kwargs)
```
loriab, line 2066: in my notions, this would become the below and `run_tdscf_excitations` is a fn that live in `proc.py` and has contents https://github.com/psi4/psi4/pull/1885/files#diff-dbf22cd80d16f6d8ef8250145a6d91bfR2709-R2732 .

```
def tdscf(wfn, **kwargs):
    return run_tdscf_excitations(wfn, **kwargs)
```

Alternately, `run_tdscf_excitations` in `proc.py` is named `tdscf` in `proc.py` and has the same contents. Then would also need to import tdscf into higher namespace for `psi4.tdscf` to work.

psi4/driver/procrouting/proc.py
loriab: in my notions, `run_tdscf_energy(name, **kwargs)` has just that signature (no default for `name`) and has contents https://github.com/psi4/psi4/pull/1885/files#diff-dbf22cd80d16f6d8ef8250145a6d91bfR2698-R2707 and then returns `tdscf(ref_wfn)`.

robertodr:
@loriab I can handle the reconciliation/kill off of options when rebasing #1814, so no need of letting this in a partially broken state.

psi4/driver/p4util/util.py
```diff
@@ -33,9 +33,9 @@
 import warnings
 
 from psi4 import core
+from psi4.driver.procrouting import *
```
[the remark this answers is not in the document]
andysim: I agree with nuking this - polluting the namespace with the `*` import makes it hard to figure out where things come from in the code

loriab:
I think you'll need to rebase @jeffschriber to get the code that goes with the latest cppe package so tests can pass.

jeffschriber:
Sounds good. I'll rebase, and I can also do a little cleanup of the function signatures.

PR #1893, opened 2020-05-19
Support for the BrianQC GPU module

kis-gergely-dzsi:
## Description
Added core BrianQC functionality
 - psi4 can now be compiled with the USING_BrianQC switch
 - when compiled with BrianQC, several operations can be computed on the available GPUs through the BrianQC module, including Fock building and diagonalization

## Status
- [x] Ready for review
- [ ] Ready for merge


hokru:
I am afraid this will require a rebase to get rid of other people's commits.

kis-gergely-dzsi:
Rebased brian branch onto psi4:master.

psi4/CMakeLists.txt
```diff
@@ -161,6 +161,24 @@ else()
...
+        include_directories("${BrianQC_DIR}/src/brian_module/static_wrapper/")
+        link_directories("${BrianQC_DIR}/build/lib/")
+    else()
+        message(STATUS "${Cyan}Using installed BrianQC${ColourReset}: ${BrianQC_DIR}")
+        include_directories("${BrianQC_DIR}/include/")
+        link_directories("${BrianQC_DIR}/lib/")
+    endif()
+else()
+    message(STATUS "Disabled BrianQC")
+endif()
```
loriab, line 180: Is CMake BrianQC's build system (or one of them)? Is it exporting a `BrianQCConfig.cmake` file? We like that setup because it can export a "target" with link and include properties attached. This keeps the info about how to use a project near the people who know how to use it (you) :-)

Even though as a private project you're a little different, I think for pattern's sake, it'd still be good for you to have a file like https://github.com/psi4/psi4/blob/master/external/upstream/gau2grid/CMakeLists.txt . Apart from `gau2grid` -> `BrianQC`, I think the main change is that the middle section https://github.com/psi4/psi4/blob/master/external/upstream/gau2grid/CMakeLists.txt#L12-L30 becomes just a `message` command to contact you and arrange for a license.

Let me know if I can point you to code models.

psi4/src/core.cc
```diff
@@ -979,13 +996,36 @@ bool psi4_python_module_initialize() {
     static char* argv = (char*)"";
     for_rtl_init_(&argc, &argv);
 #endif
+    
+#ifdef USING_BrianQC
+    const char* brianEnableEnv = getenv("BRIANQC_ENABLE");
```
loriab, line 1001: is environment variables how the user customarily activates BrianQC? In psi, `cmake -DENABLE_<external>=ON` is a signal to detect-or-build-or-die-trying, and runtime activation would normally be through an option (read_options.cc). Pretty much no envvar control. But if BrianQC users with other programs are accustomed to envvar control, there's something to be said for keeping that consistency.

psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -81,8 +94,149 @@ void DirectJK::print_header() const {
         outfile->Printf("    Schwarz Cutoff:    %11.0E\n\n", cutoff_);
     }
 }
-void DirectJK::preiterations() { sieve_ = std::make_shared<ERISieve>(primary_, cutoff_, do_csam_); }
+void DirectJK::preiterations() {
+    sieve_ = std::make_shared<ERISieve>(primary_, cutoff_, do_csam_);
+    
+#ifdef USING_BrianQC
+    if (brianCookie != 0) {
+        double threshold = cutoff_ * (brianCPHFFlag ? 1e-3 : 1e-0); // CPHF needs higher precision
```
loriab, line 102: Are these two only reasonable values? Or is this threshold something we could turn into a runtime option?

psi4/src/psi4/libfock/CMakeLists.txt
```diff
@@ -32,3 +32,10 @@ target_link_libraries(fock
   PRIVATE
     gau2grid::gg
   )
+
+if(ENABLE_BrianQC)
+  target_compile_definitions(fock
+    PUBLIC
+      USING_BrianQC
+    )
+endif()
```
loriab, line 41: When the CMake target is available, should be able to do

```
if(TARGET BrianQC::BrianQC)
  target_link_libraries(fock
    PUBLIC
      BrianQC::BrianQC
    )
endif()
```

loriab:
oh, and don't worry about the Azure fail. another rebase (to include #1894) will fix it.

psi4/src/core.cc
kis-gergely-dzsi: I will have to talk this over with the other developers. For the sake of consistency between host programs, we would probably like to keep the env var, but we can create an option as well. One advantage of the env var over an input file option is that is allows the user (and us) to use the same input verbatim with and without Brian, for the sake of comparing performace, precision, etc.

psi4/CMakeLists.txt
kis-gergely-dzsi: Thanks for the suggestion! As you wrote, since Brian is closed-source, there are a few differences, such as Psi4 not being able to automatically build us. Also, since Brian is not normally installed from source or a package manager, the location on the machine is much less predictable, so writing a FindBrianQC.cmake (which, if I understood correctly, is required by Psi4's 3rdparty dependecy cmakes, such as the gau2grid one) might be difficult. That's why originally we opted for letting the user specify everything explicitly. But you are right that we should streamline the build process somehow, so we are going to look into this.

psi4/src/psi4/libfock/DirectJK.cc
kis-gergely-dzsi: Due to certain features of our implementation (such as using GPU arithmetic), we usually have to use slightly stricter precision thresholds for Brian's calculations that the ones used by the host. On the other hand, we usually try to make Brian's operation "transparent"; that is, if the user specifies a certain computation with certain precision parameters, then s/he should get the same results with host+Brian as with the host itself. Thus, these multipliers are not normally visible to the user; s/he instead sets the usual parameters like the integral precision threshold, and the required multipliers are determined by us, from several precision measurements.
The two values here are because we observed that the results of a vibrational analysis (using Fock builds for CPHF) are much more precision-sensitive than the single point energy (using Fock builds for SCF). Over time, if the number of calculation types supported by Brian grows, the number of these fitted parameters might increase, and eventually, it might need to be refactored.

psi4/src/psi4/libfock/CMakeLists.txt
kis-gergely-dzsi: You are right, I will rewrite the conditional linking to match the CMake style/pattern used in Psi4.

psi4/CMakeLists.txt
loriab: The nice thing about the `<project>Config.cmake` + `find_package(<project> CONFIG)` setup is that it works with package managers, from source, or in-between since it's separating the internal install layout (known in project's build system) from the location of that layout (needed by psi's build/detect system). Here's some links, in case helpful. 

* https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages
* https://gitlab.com/libxc/libxc/-/blob/master/CMakeLists.txt#L485-506
* https://github.com/dgasmith/gau2grid/blob/master/cmake/gau2gridConfig.cmake.in

psi4/src/core.cc
loriab: I expect having both envvar and option will work (though we might try to capture envvar and route the instruction through an option). QC programs are already awful at exposing their state, and I worry about envvar leaving no trace about what was running, except maybe a line in text output. Agree about utility for testing.

psi4/src/psi4/libfock/DirectJK.cc
loriab: ah, multipliers, I see now! good, they're already effectively tunable by user.

loriab:
Rebase the next time you push, please, to include #1896. I think we finally have the inconsistency under control.

psi4/src/psi4/libfock/CMakeLists.txt
kis-gergely-dzsi: I have rewritten the conditionals to follow Psi4's conventions, please let me know if anything else is amiss.

psi4/CMakeLists.txt
kis-gergely-dzsi: I have created a BrianQCConfig.cmake, and I have rewritten the Psi4-BrianQC linking to more closely match the conventions. Please let me know if there is anything else I should change.

kis-gergely-dzsi: You are right, the global include_directories is not necessary. I've moved the includes into the target-specific conditional blocks.

CMakeLists.txt
```diff
@@ -98,6 +98,7 @@ option(ENABLE_PCMSolver "Enables PCMSolver library (requires Fortran)" OFF)
 option(ENABLE_snsmp2 "Enables SNSMP2 plugin (can also be added at runtime)" OFF)
 option(ENABLE_v2rdm_casscf "Enables V2RDM_CASSCF plugin (requires Fortran; can also be added at runtime)" OFF)
 option(ENABLE_gpu_dfcc "Enables GPU_DFCC plugin for gpu-accelerated df-cc (requires CUDA; can also be added at runtime)" OFF)
+option(ENABLE_BrianQC "Enables the BrianQC GPU module" OFF)
```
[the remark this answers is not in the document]
kis-gergely-dzsi: I've added a note to the option.

external/upstream/brianqc/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+find_package(BrianQC 1.1 CONFIG QUIET)
+
+if(${BrianQC_FOUND})
+    get_property(_loc TARGET BrianQC::static_wrapper PROPERTY LOCATION)
+    if(${BrianQC_DEV})
+        message(STATUS "${Cyan}Found BrianQC${ColourReset}: ${_loc} (found development version ${BrianQC_VERSION})")
+    else(${BrianQC_DEV})
+        message(STATUS "${Cyan}Found BrianQC${ColourReset}: ${_loc} (found installed version ${BrianQC_VERSION})")
+    endif(${BrianQC_DEV})
```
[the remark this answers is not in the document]
kis-gergely-dzsi: Yes, the BrianQC_DEV is set in BrianQCConfig.cmake, and it is actually no longer really necessary, since it is only used to tweak the CMake message depending on the BrianQC version. I've also removed the superfluous conditions from the else() and endif().

psi4/src/psi4/libfock/CMakeLists.txt
kis-gergely-dzsi: As far as I know, the build speed is not affected. However, most of the brian-specific code blocks use one or two of BrianQC's own datatypes (for example, BrianCookie and brianInt), declared in our headers. Thus, when building Psi4 without BrianQC, almost all of our code has to be disabled, otherwise it will not compile at all (referring to unavailable types). A possible workaround would be declaring a dummy version of those types when building without BrianQC (adding an #else to the #ifdefs that include BrianQC's headers), however, I'm not sure that would be a nicer solution than leaving the #ifdefs in. What is your opinion, which method introduces less ugliness into the code?

external/upstream/brianqc/CMakeLists.txt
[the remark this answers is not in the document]
kis-gergely-dzsi: In fact, after reviewing the CMake, I realized that it is probably better to remove that variable altogether, since it served only a cosmetic purpose.

kis-gergely-dzsi:
Currently, the automated tests fail due to an automatically not resolvable merge conflict. @loriab, is only a rebase an acceptable solution for such conflicts, or could I also just merge with the newest psi4/master?

loriab:
> Currently, the automated tests fail due to an automatically not resolvable merge conflict. @loriab, is only a rebase an acceptable solution for such conflicts, or could I also just merge with the newest psi4/master?

Rebase preferred, but other `git` operations allowed so long as all the commits are still by you. (What can't work are PR branches where submitter's commits are repeated or interspersed with commits already in master.)

psi4/src/psi4/libfock/CMakeLists.txt
```diff
@@ -32,3 +32,14 @@ target_link_libraries(fock
...
   )
+
+if(TARGET BrianQC::static_wrapper)
+  target_compile_definitions(fock
+    PUBLIC
+      USING_BrianQC
+    )
+  target_include_directories(fock
+    PUBLIC
+      ${BrianQC_INCLUDE_DIRS}
```
loriab, line 43:
```suggestion
      $<TARGET_PROPERTY:BrianQC::static_wrapper,INTERFACE_INCLUDE_DIRECTORIES>
```

psi4/src/psi4/libfock/CMakeLists.txt
loriab: The above suggestion is just what I was thinking of. It compiles fine with BrianQC disabled for me. It's a fairly minor detail, so at your choice.

kis-gergely-dzsi:
@loriab I have looked into your suggestion about enabling and disabling BrianQC with an input file option instead of (or in addition to) an env var, and I think there is a minor issue here. Normally, BrianQC's initialization should happen once, preferably during Psi4's own initialization. But, due to the way Psi4 is normally used from Python (first doing `import psi4`, and then calling `psi4.set_options()`), the input options are not yet available when `psi4_python_module_initialize()` is called (from which `brianAPIInit()` had been called until now).

As far as I could tell, there is no function that normally runs between setting the options and performing a calculation. So, to enable BrianQC through an input option, I have two choices: either I add the check-and-initialize to the options setter method itself (`py_psi_set_global_option_int()`), or I add it to every high-level computation function (like `energy()` and `optimize()`) that could potentially use BrianQC. Which way would be better in your opinion?

kis-gergely-dzsi:
@loriab I've started writing some documentation for BrianQC within Psi4's Sphinx documentation system. However, I can't get `make sphinxman` or `make sphinxpdf` to work on the newest psi4/master (without BrianQC's modifications); I'm getting a Python exception (AttributeError: 'pybind11_static_property' object has no attribute 'expandtabs'). Could there be something I'm doing wrong?

loriab:
@kis-gergely-dzsi this patch to sphinx fixes that https://github.com/loriab/sphinx/commit/fa5fa5ce56aaa7f5d45c5923c6da34382ee4bce2 . I haven't fully investigated to see if it's been fixed in pybind11 or if I need to submit that to sphinx itself.

psi4/src/psi4/libfock/CMakeLists.txt
kis-gergely-dzsi: I have changed the include dir syntax to the one you proposed.

loriab:
@kis-gergely-dzsi 

> I have looked into your suggestion about enabling and disabling BrianQC with an input file option instead of (or in addition to) an env var, and I think there is a minor issue here. Normally, BrianQC's initialization should happen once, preferably during Psi4's own initialization. But, due to the way Psi4 is normally used from Python (first doing import psi4, and then calling psi4.set_options()), the input options are not yet available when psi4_python_module_initialize() is called (from which brianAPIInit() had been called until now).

> As far as I could tell, there is no function that normally runs between setting the options and performing a calculation. So, to enable BrianQC through an input option, I have two choices: either I add the check-and-initialize to the options setter method itself (py_psi_set_global_option_int()), or I add it to every high-level computation function (like energy() and optimize()) that could potentially use BrianQC. Which way would be better in your opinion?

Sorry, I only noticed the docs query and not this one. Looking through the code you've added I think I understand your difficulty. It looks like right now, initialization brianAPIInit() sets briancookie boolean, and that setting governs in the SCF code and otherwise whether brianqc is used or bypassed. So in the course of an input file with several togglings of brianqc on/off, then brianqc gets initiated and torn down multiple times. Is this correct? And is the multiple teardowns wanted? It's fine to add a hook to each of the major fns like energy/gradient/hessian/properties. But could there instead be a single initiation, then checking the option `options.get_bool("brianqc");` as needed in the scf, grid, etc.?

```
import psi4  # brianqc init happens alongside psi4 init

psi4.geometry(A)

psi4.set_options({'brianqc': False})

energy('b3lyp/cc-pvdz')  # runs all CPU

psi4.set_options({'brianqc': True})

energy('b3lyp/cc-pvdz')  # runs w/GPU
```

psi4/src/core.cc
```diff
@@ -70,6 +70,66 @@ using namespace psi;
...
+        outfile->Printf("BRIANQC_ENABLE option found, but overridden by BRIANQC_ENABLE environment variable\n");
+    } else {
+        outfile->Printf("BRIANQC_ENABLE option found, checking value\n");
+        if (value && (brianCookie == 0)) {
+            outfile->Printf("BRIANQC_ENABLE option set to true, initializing BrianQC\n");
+            brianInit();
+        } else if (!value && (brianCookie != 0)) {
+            outfile->Printf("BRIANQC_ENABLE option set to false, releasing BrianQC\n");
+            brianRelease();
+        }
```
loriab, line 128: I think toggling brianqc/pure-cpu control still needs working out (and if you want a conference call with me to discuss this easier, I'm glad to do so), but since there's no immediate users to disrupt, that can be in a follow-up PR.

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -2332,6 +2350,61 @@ void DirectJKGrad::compute_gradient()
...
+        
+        if (do_J_) {
+            gradients_["Coulomb"] = Jgrad;
+        }
+        
+        if (do_K_) {
+            gradients_["Exchange"] = Kgrada;
+            
+            if (do_wK_) {
+                gradients_["Exchange,LR"] = std::make_shared<Matrix>("Exchange,LR Gradient", primary_->molecule()->natom(), 3);
```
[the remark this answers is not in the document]
kis-gergely-dzsi: Yes, it's getting initialized to zero. If there is both a "normal" and a range-separated exchange (do_K_ and do_wK_ both true), then BrianQC only computes their sum. (Also, it factors in the DFT functional's exchange weights internally, so Psi4's scaling later is disabled when BrianQC is enabled.) So, I wanted to ensure two things: that all matrices that Psi4 would otherwise initialize are still initialized (since I wasn't really sure if any later code depends on them existing), and that the sum of K and wK are always correct. So that's why in that branch, we put BrianQC's exchange (the sum of K and wK) into K, and we initialize wK, but leave it zero. It would be more accurate if BrianQC computed the two contributions separately, but the current API doesn't support that.

doc/sphinxman/source/brianqc.rst
```diff
@@ -0,0 +1,164 @@
...
+
+To achieve peak performance, the following settings are recommended when enabling
+BrianQC:
+
+* By default, |PSIfour| uses density-fitted preiterations for SCF, which BrianQC
+  doesn't handle yet. To ensure that no non-BrianQC-accelerated iterations are
+  performed, disable the preiterations by setting |scf__df_scf_guess| to `False`.
+* By default, |PSIfour| uses a disk-based Fock building, but BrianQC currently only
+  accelerates direct Fock builds. To ensure that Fock building is accelerated by
+  BrianQC, set |globals__scf_type| to `direct`.
```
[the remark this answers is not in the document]
kis-gergely-dzsi: You're right, I will add a note about Psi4+BrianQC still respecting Psi4's usual precision parameters.

tests/brianqc/linear-dependence/input.dat
```diff
@@ -0,0 +1,64 @@
+#! Water with modified sto-3g to test linearly-dependent basis
```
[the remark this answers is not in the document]
kis-gergely-dzsi: Sure, I will change the number of threads. How many threads would be preferable? Maybe 4 or 2? Or should this also be a single-thread testcase like the others?

My only concern is that when actually running the test suite with BrianQC (as opposed to running them as "dummy" tests with pure Psi4), it would be very unusual for the machine to have enough GPUs for multiple Psi4+BrianQC instances to run in parallel. And even then, the user would need to configure the BrianQC instances to only use one GPU per instance (by default, BrianQC tries to allocate all available GPUs it can handle). So I've originally set the testcase to use 8 threads because BrianQC's tests usually need to be run serially anyway.

psi4/src/core.cc
kis-gergely-dzsi: It would be good to discuss this issue over a video call, since I'm still not sure which potential solution is preferable. In the meantime, I will create a separate internal variable to enable and disable BrianQC, so that enabling is not tied to whether the cookie is initialized.

loriab: sounds good. will switch to email, but I can do today or M–W of next week.

psi4/src/psi4/scfgrad/jk_grad.cc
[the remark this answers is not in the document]
loriab: got it. thanks for the explanation.

tests/brianqc/linear-dependence/input.dat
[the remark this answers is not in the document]
loriab: good point. I would say it should be single thread to fit with running the whole suite. but the `tests/brianqc/` directory will only be added if brianqc is detected at build-time, at which point it's reasonable for the user to manage two-phase testing of `ctest -L brianqc -j1` and `ctest -LE brianqc -j12`. then may as well leave the 8 threads as is so the former runs in reasonable time.

psi4/src/core.cc
loriab: From discussions today, we're thinking of two controls:
* an envvar to toggle whether BrianQC library (and GPUs) even gets initialized (takes about 10 s). 
* a psi4 option to toggle whether to use BrianQC for next jobs so that in an input one can run DFT w/ and w/o GPUs.
* BrianQC devs' choice about whether initialization is either once per `import psi4` at the same time as psi4 initialization or per-job where BrianQC is to be used.

They're going to talk this over and it can either go into this PR or a subsequent one, depending on review timings.

PR #1895, opened 2020-05-20
Overlap Half-derivative Integrals

kcpearce:
## Description
Adds capabilities for overlap half-derivative integrals and exports them to Python side for use in calculating VCD

## Todos
- [x] Adds `.def("ao_overlap_half_deriv1")` and `.def("mo_overlap_half_deriv1")` binding to `psi4/src/export_mints.cc`
- [x] Adds `ao_overlap_half_deriv1()` and `mo_overlap_half_deriv1()` definition/declaration to `psi4/src/psi4/libmints/mintshelper.cc(h)`
- [x] Adds `ao_overlap_half_deriv1_helper()` helper function to `psi4/src/psi4/libmints/mintshelper.cc(h)`

## Questions
- None

## Checklist
- [x] Add test in `/tests/pytests/test_mints.py`

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1055,32 +1055,37 @@ void export_mints(py::module& m) {
 
         // First and second derivatives of one and two electron integrals in AO and MO basis.
         .def("ao_oei_deriv1", &MintsHelper::ao_oei_deriv1,
-             "Gradient of AO basis OEI integrals: returns (3 * natoms) matrices")
+             "Gradient of AO basis OEI integrals: returns (3 * natoms) matrices", "oei_type"_a, "atom"_a)
         .def("ao_oei_deriv2", &MintsHelper::ao_oei_deriv2,
-             "Hessian  of AO basis OEI integrals: returns (3 * natoms)^2 matrices")
+             "Hessian  of AO basis OEI integrals: returns (3 * natoms)^2 matrices", "oei_type"_a, "atom1"_a, "atom2"_a)
         .def("ao_overlap_half_deriv1", &MintsHelper::ao_overlap_half_deriv1,
-             "Half-derivative of AO basis overlap integrals: returns (3 * natoms) matrices")
+             "Half-derivative of AO basis overlap integrals: returns (3 * natoms) matrices","side"_a, "atom"_a)
         .def("ao_tei_deriv1", &MintsHelper::ao_tei_deriv1,
-             "Gradient of AO basis TEI integrals: returns (3 * natoms) matrices", "atom"_a, "omega"_a = 0.0,
-             "factory"_a = nullptr)
+             "Gradient of AO basis TEI integrals: returns (3 * natoms) matrices",
+             "atom"_a, "omega"_a = 0.0, "factory"_a = nullptr)
         .def("ao_tei_deriv2", &MintsHelper::ao_tei_deriv2,
-             "Hessian  of AO basis TEI integrals: returns (3 * natoms)^2 matrices")
+             "Hessian  of AO basis TEI integrals: returns (3 * natoms)^2 matrices", "atom1"_a, "atom2"_a)
         .def("mo_oei_deriv1", &MintsHelper::mo_oei_deriv1,
-             "Gradient of MO basis OEI integrals: returns (3 * natoms) matrices")
+             "Gradient of MO basis OEI integrals: returns (3 * natoms) matrices",
+             "oei_type"_a, "atom"_a, "C1"_a, "C2"_a)
         .def("mo_oei_deriv2", &MintsHelper::mo_oei_deriv2,
-             "Hessian  of MO basis OEI integrals: returns (3 * natoms)^2 matrices")
+             "Hessian  of MO basis OEI integrals: returns (3 * natoms)^2 matrices",
+             "oei_type"_a, "atom1"_a, "atom2"_a, "C1"_a, "C2"_a)
         .def("mo_overlap_half_deriv1", &MintsHelper::mo_overlap_half_deriv1,
-             "Half-derivative of MO basis overlap integrals: returns (3 * natoms) matrices")
+             "Half-derivative of MO basis overlap integrals: returns (3 * natoms) matrices",
+             "side"_a, "atom"_a, "C1"_a, "C2"_a)
         .def("mo_tei_deriv1", &MintsHelper::mo_tei_deriv1,
-             "Gradient of MO basis TEI integrals: returns (3 * natoms) matrices")
+             "Gradient of MO basis TEI integrals: returns (3 * natoms) matrices",
+             "atom"_a, "C1"_a, "C2"_a, "C3"_a, "C4"_a)
         .def("mo_tei_deriv2", &MintsHelper::mo_tei_deriv2,
-             "Hessian  of MO basis TEI integrals: returns (3 * natoms)^2 matrices")
+             "Hessian  of MO basis TEI integrals: returns (3 * natoms)^2 matrices",
+             "atom1"_a, "atom2"_a, "C1"_a, "C2"_a, "C3"_a, "C4"_a)
         
         // First derivatives of electric dipole integrals in AO and MO basis.
         .def("ao_elec_dip_deriv1", &MintsHelper::ao_elec_dip_deriv1,
-             "Gradient of AO basis electric dipole integrals: returns (3 * natoms) matrices")
+             "Gradient of AO basis electric dipole integrals: returns (3 * natoms) matrices", "atom"_a)
         .def("mo_elec_dip_deriv1", &MintsHelper::mo_elec_dip_deriv1,
-             "Gradient of MO basis electric dipole integrals: returns (3 * natoms) matrices");
+             "Gradient of MO basis electric dipole integrals: returns (3 * natoms) matrices", "atom"_a, "C1"_a, "C2"_a);
 
     py::class_<Vector3>(m, "Vector3",
                         "Class for vectors of length three, often Cartesian coordinate vectors, "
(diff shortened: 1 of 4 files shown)


susilehtola:
Aren't these half-overlap derivative integrals? :stuck_out_tongue_closed_eyes:

kcpearce:
@susilehtola tomato, tomato? 🍅

tests/pytests/test_mints.py
```diff
@@ -43,3 +43,56 @@ def test_export_ao_elec_dip_deriv():
...
+        O
+        H 1 1.0
+        H 1 1.0 2 101.5
+        symmetry c1
+    """)
+
+    psi4.core.set_active_molecule(h2o)
+    psi4.set_options({'basis': 'STO-3G'})
+
+    rhf_e, wfn = psi4.energy('SCF', return_wfn=True)
```
loriab, line 58: fyi, can do these three lines with `rhf_e, wfn = psi4.energy('scf/sto-3g')`. or can add `molecule=h2o` to energy call.

susilehtola:
> @susilehtola tomato, tomato? tomato

Not really, more like tomato, matoto. Half-derivative is confusing due to https://en.wikipedia.org/wiki/Fractional_calculus

kcpearce:
> tests will fix with a rebase

@loriab Looks like I'm still failing the Cholesky convergence test after a rebase

loriab:
> Looks like I'm still failing the Cholesky convergence test after a rebase

Well, you're not hitting the same error. This one is down the sequence at 1e-10, not 1e-7. And always by about same amt. 

```  Energy with Cholesky 1e-10: computed value (-230.6415254219) does not    !
! match (-230.6415254333) to atol=1e-08 by difference (0.0000000114).
```

I wonder if this is associated with the new 2020.1 mkl set that's ~9 days old.

loriab:
rebase next time, please, to include #1896

kcpearce:
Ready for review!

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2337,6 +2337,106 @@ std::vector<SharedMatrix> MintsHelper::ao_potential_deriv1_helper(int atom) {
...
+    std::shared_ptr<BasisSet> bs2 = GInt->basis2();
+
+    int nbf1 = bs1->nbf();
+    int nbf2 = bs2->nbf();
+
+    std::vector<SharedMatrix> grad;
+    for (int p = 0; p < 3; p++) {
+        std::stringstream sstream;
+        sstream << "ao_overlap_half_deriv1_" << atom << cartcomp[p];
+        grad.push_back(SharedMatrix(new Matrix(sstream.str(), nbf1, nbf2)));
```
robertodr, line 2358: I think:
```
grad.push_back(std::make_shared<Matrix>(sstream.str(), nbf1, nbf2));
```
is the preferred idiom.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -3193,6 +3293,19 @@ std::vector<SharedMatrix> MintsHelper::ao_oei_deriv1(const std::string &oei_type
...
 
+std::vector<SharedMatrix> MintsHelper::ao_overlap_half_deriv1(const std::string &half_der_side, int atom) {
+    std::vector<SharedMatrix> ao_grad;
+
+    if (half_der_side == "LEFT")
+        ao_grad = ao_overlap_half_deriv1_helper("LEFT", atom);
+    else if (half_der_side == "RIGHT")
+        ao_grad = ao_overlap_half_deriv1_helper("RIGHT", atom);
+    else
+        throw PSIEXCEPTION("Not a valid choice of half derivative side: must be LEFT or RIGHT");
```
robertodr, line 3304: What about the following?
```c++
if (half_der_side == "LEFT" || half_der_side == "RIGHT") {
   ao_grad = ao_overlap_half_deriv1_helper(half_der_side, atom);
} else {
   throw PSIEXCEPTION("Not a valid choice of half derivative side: must be LEFT or RIGHT");
}
```

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2337,6 +2337,106 @@ std::vector<SharedMatrix> MintsHelper::ao_potential_deriv1_helper(int atom) {
     return grad;
 }
 
+std::vector<SharedMatrix> MintsHelper::ao_overlap_half_deriv1_helper(const std::string &half_der_side, int atom) {
+    std::vector<std::string> cartcomp;
```
robertodr, line 2341: A fixed-size `std::array<std::string, 3>` should suffice here: `std::array<int, 3> cartcomp{ {"X", "Y", "Z"} };`

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -3277,6 +3390,30 @@ std::vector<SharedMatrix> MintsHelper::mo_oei_deriv2(const std::string &oei_type
     return mo_grad;
 }
 
+std::vector<SharedMatrix> MintsHelper::mo_overlap_half_deriv1(const std::string &half_der_side, int atom, SharedMatrix C1, SharedMatrix C2) {
+    std::vector<std::string> cartcomp;
```
robertodr, line 3394: Ditto. `std::array` suffices, I think.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -3277,6 +3390,30 @@ std::vector<SharedMatrix> MintsHelper::mo_oei_deriv2(const std::string &oei_type
...
+
+    // Assuming C1 symmetry
+    int nbf1 = ao_grad[0]->rowdim();
+    int nbf2 = ao_grad[0]->coldim();
+
+    std::vector<SharedMatrix> mo_grad;
+    for (int p = 0; p < 3; p++) {
+        std::stringstream sstream;
+        sstream << "mo_overlap_half_deriv1_" << atom << cartcomp[p];
+        SharedMatrix temp(new Matrix(sstream.str(), nbf1, nbf2));
```
robertodr, line 3410: Ditto, the preferred idiom is `auto temp = std::make_shared<Matrix>(sstream.str(), nbf1, nbf2));`

tests/pytests/test_mints.py
```diff
@@ -43,3 +43,53 @@ def test_export_ao_elec_dip_deriv():
...
+        for atom_cart in range(3):
+            map_key1 = "S_LEFT_HALF_" + str(atom) + cart[atom_cart]
+            map_key2 = "S_RIGHT_HALF_" + str(atom) + cart[atom_cart]
+            map_key3 = "S_" + str(atom) + cart[atom_cart]
+            deriv1_np[map_key1] = np.asarray(deriv1_mat["S_LEFT_HALF_" + str(atom)][atom_cart])
+            deriv1_np[map_key2] = np.asarray(deriv1_mat["S_RIGHT_HALF_" + str(atom)][atom_cart])
+            deriv1_np[map_key3] = np.asarray(deriv1_mat["S_" + str(atom)][atom_cart])
+
+            # Test (S_ii)^x = 2 * < i^x | i >
+            assert np.allclose(deriv1_np[map_key1].diagonal(), (deriv1_np[map_key3].diagonal())/2)
```
robertodr, line 80: You can probably use the `compare_arrays` function here.

psi4/src/psi4/libmints/mintshelper.cc
kcpearce: Made this change at every instance I could find in mintshelper.cc

psi4/src/psi4/libmints/mintshelper.cc
kcpearce: Ditto. Made this change throughout the file.

kcpearce: NOTE: I was making this change throughout mintshelper.cc (even in functions that existed before this PR) and found out the hard way that two functions require that the `"x", "y", "z"` in this vector be lowercase, viz. [ao_potential_deriv2_helper](https://github.com/psi4/psi4/blob/ea5b5eb71ab65469629b345d78fe6e7b1a5a34fa/psi4/src/psi4/libmints/mintshelper.cc#L2340-L2344) and [ao_tei_deriv2](https://github.com/psi4/psi4/blob/ea5b5eb71ab65469629b345d78fe6e7b1a5a34fa/psi4/src/psi4/libmints/mintshelper.cc#L2892-L2896). This has to do with the fact that later on in these two functions, they reference a `hess_map` with hard coded key values, e.g. `hess_key["CxAx"]` [here](https://github.com/psi4/psi4/blob/ea5b5eb71ab65469629b345d78fe6e7b1a5a34fa/psi4/src/psi4/libmints/mintshelper.cc#L2451) in which the `"x"` portions of the string are lowercase. If the `"x", "y", "z"` of the `cartomp` vector are uppercase, this becomes an issue [here]() where we build the `key` variable from our `cartcomp` vector and then use it to access different keys of `hess_map`. Using the earlier example, this would build the `key` variable as `"CXAX"`, but `hess_map` only has an entry for the key `"CxAx"`. This is a very minor issue, but it took me a while to hunt down.

kcpearce:
Any takers for additional review?

psi4/src/export_mints.cc
```diff
@@ -1067,6 +1069,8 @@ void export_mints(py::module& m) {
              "Gradient of MO basis OEI integrals: returns (3 * natoms) matrices")
         .def("mo_oei_deriv2", &MintsHelper::mo_oei_deriv2,
              "Hessian  of MO basis OEI integrals: returns (3 * natoms)^2 matrices")
+        .def("mo_overlap_half_deriv1", &MintsHelper::mo_overlap_half_deriv1,
+             "Half-derivative of MO basis overlap integrals: returns (3 * natoms) matrices")
```
loriab, line 1073: consider naming vars. gives user option of remembering var names rather than order. `"side"_a, "atom"_a, "C1"_a, "C2"_a`

kcpearce: Done. For this instance and surrounding ones.

kcpearce:
Ready for merge!

PR #1896, opened 2020-05-21
accommodate mkl 2020.1 and cholesky-basis, take 5

loriab:
## Description
MKL 2020.1 often selects a different basis fn set than 2019.4 or 2020.0, so answers often differ by 1.14e-8. This relaxes tolerance to 2.e-8. Alternative is to add `mkl!=2020.1` to `conda create` line, but this would be disruptive to conda packaging.

* **Note** A definitive test of above statement that returns to 1.e-8 but bans 2020.1 is running at https://dev.azure.com/psi4/psi4/_build/results?buildId=1969&view=logs&j=96451287-da82-57a8-2c11-8da7db9ab71a , but I'll get a head start and submit this PR.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/scf-cholesky-basis/input.dat
+++ b/tests/scf-cholesky-basis/input.dat
@@ -15,16 +15,16 @@ e_canonical   = -230.64152543331906
 # indices. Since there's no guarantee that the same functions get
 # picked every time, we need to allow for changes in the energy that
 # go to zero when the tolerance is tightened.
-# MKL 2020.1 seems to often pick different fns, so need to relax the tolerance from 1.e-8 to 2.e-8
+# MKL 2020.1 can use alternate code path, necessitating relaxing the tolerance from 1.e-8 to 2.e-8, incl. canonical
 
 tol_4 = 3
 tol_5 = 4
 tol_6 = 5
-tol_7 = 2.e-8
-tol_8 = 2.e-8
-tol_9 = 2.e-8
-tol_10 = 2.e-8
-tol_11 = 2.e-8
+tol_7 = 8
+tol_8 = 8
+tol_9 = 8
+tol_10 = 8
+tol_11 = 8
 tol_can = 8
 
 molecule {
--- a/tests/scf-cholesky-basis/input.dat
+++ b/tests/scf-cholesky-basis/input.dat
@@ -15,6 +15,7 @@ e_canonical   = -230.64152543331906
 # indices. Since there's no guarantee that the same functions get
 # picked every time, we need to allow for changes in the energy that
 # go to zero when the tolerance is tightened.
+# MKL 2020.1 seems to often pick different fns, so need to relax the tolerance from 1.e-8 to 2.e-8
 
 tol_4 = 3
 tol_5 = 4
@@ -22,8 +23,8 @@ tol_6 = 5
 tol_7 = 2.e-8
 tol_8 = 2.e-8
 tol_9 = 2.e-8
-tol_10 = 8
-tol_11 = 8
+tol_10 = 2.e-8
+tol_11 = 2.e-8
 tol_can = 8
 
 molecule {
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -7,6 +7,7 @@ jobs:
   timeoutInMinutes: 120
   variables:
     CTEST_OUTPUT_ON_FAILURE: 1
+    MKL_CBWR: AUTO
   strategy:
     maxParallel: 8
     matrix:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,7 @@ option_with_print(ENABLE_OPENMP "Enables OpenMP parallelization" ON)
 option_with_print(ENABLE_AUTO_BLAS "Enables CMake to auto-detect BLAS" ON)
 option_with_print(ENABLE_AUTO_LAPACK "Enables CMake to auto-detect LAPACK" ON)
 option_with_print(ENABLE_PLUGIN_TESTING "Test the plugin templates build and run" OFF)
-option_with_print(ENABLE_CYTHONIZE "Compile each python file rather than plaintext" OFF)
+option_with_print(ENABLE_CYTHONIZE "Compile each python file rather than plaintext (requires cython) !experimental!" OFF)
 option_with_flags(ENABLE_XHOST "Enables processor-specific optimization (with MSVC, it enables AVX2 instructions)" ON
                   "-xHost" "-march=native" "/arch:AVX2")
 option_with_flags(ENABLE_CODE_COVERAGE "Enables details on code coverage" OFF


tests/scf-cholesky-basis/input.dat
```diff
@@ -15,16 +15,17 @@ e_canonical   = -230.64152543331906
...
 tol_6 = 5
-tol_7 = 7
-tol_8 = 7
-tol_9 = 7
-tol_10 = 8
-tol_11 = 8
-tol_can = 8
+tol_7 = 2.e-8
+tol_8 = 2.e-8
+tol_9 = 2.e-8
```
susilehtola, line 25: Shouldn't these be 2e-7?

tests/scf-cholesky-basis/input.dat
```diff
@@ -15,16 +15,17 @@ e_canonical   = -230.64152543331906
...
+tol_10 = 2.e-8
+tol_11 = 2.e-8
+tol_can = 2.e-8
```
susilehtola, line 28: I'm not sure you should change the canonical threshold

tests/scf-cholesky-basis/input.dat
loriab: They were all 1.e-8 before yesterday.
https://github.com/psi4/psi4/commit/ac1bd292bc711b78ff593a73027c8071a42d86f6#diff-db0b071835a8a6a10a315bb623c173cb  All the differences are about 1.14e-8. Jonathon relaxed tol_7/8/9 to 1.e-7 yesterday, but I think the lot works fine with the more conservative 2.e-8. So I think as-is is right.

tests/scf-cholesky-basis/input.dat
loriab: good point

loriab: nope, canonical, too: `Energy with canonical orthogonalization: computed value                  !
! (-230.6415254219) does not match (-230.6415254333) to atol=1e-08 by difference   !
! (0.0000000114). ` 
does this change the analysis of the situation, @susilehtola?

susilehtola: @loriab then this is not related to Cholesky but mkl changing linear algebra...

loriab: The canonical calc is dependent on the Cholesky calcs through `guess read`. I haven't been able to get the canonical calc to produce the reference value when run alone (off by about 0.001857 [E_h]) with 2019.4 or 2020.1. Add back the Cholesky 1e-4 and then you get the canonical passing for 2019.4 for the 1.1e-8 error for 2020.1 .

```
<     One-Electron Energy =                -715.6736276673161683
<     Two-Electron Energy =                 281.5139810571366183
<     Total Energy =                       -230.6415254410109696
---
>     One-Electron Energy =                -715.6736277971549498
>     Two-Electron Energy =                 281.5139811795980904
>     Total Energy =                       -230.6415254483882791
```

But overall, MKL values are changing results. Here's two other basis sets besides the native one in the test (fitting basis clearly not appropriate for datz). They show differences, but the test basis is by far the largest magnitude.

```
basis   mkl     DF-UHF                  Norb    Naux
-----   ---     ------                  ----    ----
 adz    2019.4  -230.6918 9098 9512     192     708
 adz    2020.1  -230.6918 9098 845158

test    2019.4  -230.6415 2544 83882791 270     558
test    2020.1  -230.6415 2544 10109696

datz    2019.4  -230.7736 4428 528847   564     558
datz    2020.1  -230.7736 4428 529125
```

Here's the [RN for MKL](https://software.intel.com/content/www/us/en/develop/articles/intel-math-kernel-library-release-notes-and-new-features.html). Not seeing anything likely. Psi should be more reliable than this (all with d_conv=10), so I'm thinking we'll need to ban 2020.1. Glad to hear views.

susilehtola: Can you get the same value from 2019 and 2020 from a basic canonical calc?

TiborGY: > 
> Here's the [RN for MKL](https://software.intel.com/content/www/us/en/develop/articles/intel-math-kernel-library-release-notes-and-new-features.html). Not seeing anything likely. Psi should be more reliable than this (all with d_conv=10), so I'm thinking we'll need to ban 2020.1. Glad to hear views.

My 2 cents is that banning specific versions of specific linalg packages, with no other reason than "it changes results", is something that makes me feel uneasy. It feels like a bandaid over a bigger problem. For example what if all future versions of MKL start behaving like 2020.1?

With that said, I do like consistent results. Personally, a few weeks ago I spent about an entire day trying to make sense of why a certain Molpro calculation fails to converge for someone else, while the same exact input works for me.
As it turns out, the calculation only converges with an older version of Molpro, and only with an AMD CPU. No, I am serious, for some unknown reason (probably related to CPU dispatching), that input only converges on AMD CPUs running a given older version.

loriab: problem hits for charged/uhf (haven't disentagled) systems between pk/tz and df/atz. Neutral singlet water shows no problems matching to 1e-10 for any reference, any mkl, any scf_type.

```
memory 3 gb

molecule {
-1 2
C          0.00000        0.00000        1.39495
C          0.00000        1.20806        0.69748
C          0.00000        1.20806       -0.69748
C          0.00000        0.00000       -1.39495
C          0.00000       -1.20806       -0.69748
C          0.00000       -1.20806        0.69748
H          0.00000        0.00000        2.47753
H          0.00000        2.14560        1.23877
H          0.00000        2.14560       -1.23877
H          0.00000        0.00000       -2.47753
H          0.00000       -2.14560       -1.23877
H          0.00000       -2.14560        1.23877
}

#set basis aug-cc-pVTZ
set basis cc-pVTZ
#set basis cc-pVQZ
#set basis aug-cc-pVQZ
set scf_type pk
#set scf_type df
set d_convergence 10
set e_convergence 11
set df_scf_guess false
set guess core
#set reference rohf
set reference uhf
e = energy('scf')
print(e)

pk_qz_2019  = -76.06462080005818
pk_aqz_2019 = -76.06578901994756
df_qz_2019 =  -76.06462015458904
df_aqz_2019 = -76.06578796456354

pk_tz_rohf_2019 = -230.674287472723
pk_tz_uhf_2019 = -230.684552455135      # passes 2019.4, passes 2020.1
df_atz_uhf_2019 = -230.6966011258728    # passes 2019.4, fails 2020.1
ref = pk_tz_uhf_2019
compare_values(ref, variable('SCF TOTAL ENERGY'), 10, 'SCF energy')  #TEST
```

JonathonMisiewicz: I'm echoing, "[B]anning specific versions of specific linalg packages, with no other reason than "it changes results", is something that makes me feel uneasy. It feels like a bandaid over a bigger problem. For example what if all future versions of MKL start behaving like 2020.1?"

loriab:
> [B]anning specific versions of specific linalg packages, with no other reason than "it changes results", is something that makes me feel uneasy.

It's definitely not something I take lightly either and is only plausible if it's a bug in MKL or as temporary in how psi is using MKL. The reason I'm suspicious of MKL is that the answer changes between two values between runs (sometimes have to run 8 times to trigger).
* affects df (both mem and disk) but pk is fine
* affects any basis (currently set as cc-pvdz for efficiency)
* nothing to do with SAD guess
* in test suite (run through only once), only the scf-cholesky-basis and an IRC test are affected.
* even by running plain water in df/aqz/uhf, I can't make it act unstable, as the df/dz/uhf benzene doublet does.

```
memory 3 gb

molecule {
-1 2
C          0.00000        0.00000        1.39495
C          0.00000        1.20806        0.69748
C          0.00000        1.20806       -0.69748
C          0.00000        0.00000       -1.39495
C          0.00000       -1.20806       -0.69748
C          0.00000       -1.20806        0.69748
H          0.00000        0.00000        2.47753
H          0.00000        2.14560        1.23877
H          0.00000        2.14560       -1.23877
H          0.00000        0.00000       -2.47753
H          0.00000       -2.14560       -1.23877
H          0.00000       -2.14560        1.23877
}

set basis cc-pVDZ
#set basis aug-cc-pVDZ
#set basis cc-pVTZ
#set basis aug-cc-pVTZ
#set basis cc-pVQZ
#set basis aug-cc-pVQZ
#set scf_type pk
#set scf_type df
set scf_type disk_df
#set scf_type mem_df
set d_convergence 10
set e_convergence 11
set df_scf_guess false
set guess core
#set reference rohf
set reference uhf
e = energy('scf')
print(e)

pk_qz_2019  = -76.06462080005818
pk_aqz_2019 = -76.06578901994756
df_qz_2019 =  -76.06462015458904
df_aqz_2019 = -76.06578796456354

pk_tz_rohf_2019 = -230.674287472723

pk_dz_uhf_2019 = -230.61792846279093    #                passes
pk_adz_uhf_2019 = -230.64203999528007   #                passes
pk_tz_uhf_2019 = -230.684552455135      # passes 2019.4, passes 2020.1
df_dz_uhf_2019 = -230.61751709208824    #                fails           0.0000 0000 0102
df_adz_uhf_2019 = -230.64167627416992   #                fails           0.0000 0000 1137
df_tz_uhf_2019 = -230.68445301909452    #                fails          -0.0000 0000 0161
df_atz_uhf_2019 = -230.6966011258728    # passes 2019.4, fails 2020.1   -0.0000 0000 2270

ref = df_dz_uhf_2019
compare_values(ref, variable('SCF TOTAL ENERGY'), 10, 'SCF energy')  #TEST
```

loriab:
New info is that http://d3f8ykwhia686p.cloudfront.net/1live/intel/Intel_PUMag_Issue11_CBWR.pdf is relevant. 

Test case is dz/df/uhf via `for i in {1..20}; do MKL_CBWR=AVX stage/bin/psi4 chgbasic.in ; done`
* always passes: `COMPATIBLE`, `SSE2`, `AVX`, `AUTO`
* mixed pass/fail: `MKL_CBWR` unset
* always fails: `SSE4_2`

susilehtola:
> New info is that http://d3f8ykwhia686p.cloudfront.net/1live/intel/Intel_PUMag_Issue11_CBWR.pdf is relevant.
> 
> Test case is dz/df/uhf via `for i in {1..20}; do MKL_CBWR=AVX stage/bin/psi4 chgbasic.in ; done`
> 
>     * always passes: `COMPATIBLE`, `SSE2`, `AVX`, `AUTO`
> 
>     * mixed pass/fail: `MKL_CBWR` unset
> 
>     * always fails: `SSE4_2`

New Intel feature: get the wrong result faster!

hokru:
https://software.intel.com/content/www/us/en/develop/documentation/mkl-windows-developer-guide/top/obtaining-numerically-reproducible-results/specifying-code-branches.html

Kinda funny they have a `STRICT` mode even for this..

hokru:
FWIW, I cannot make the input fail with mkl2020.1 on my Ryzen (gen1) workstation.
It has instructions till AVX2.

TiborGY:
> 
> 
> FWIW, I cannot make the input fail with mkl2020.1 on my Ryzen (gen1) workstation.
> It has instructions till AVX2.

MKL and any binaries generated by Intel compilers are almost always detecting CPU vendors based on vendor first and instruction support second. Something like this:
```
if ( CPU_VENDOR == Intel){
    if ( CPU_HAS_AVX_512 ) <call AVX-512 version of the function>
    if ( CPU_HAS_AVX2 ) <call AVX2 version of the function>
    ...
    else <call SSE2 version of the function>
} else { //non-Intel CPU
    <call SSE2 version of the function>
}
```
So this means that unless one add some some unofficial hacks before compiling the binaries, all AMD CPUs, regardless of their AVX support, will be running SSE2 (or similarly unoptimized) codepaths only.
This unfair CPU dispatch function can be overridden at runtime for the MKL using an unofficial debug environmental variable, but if the binary was compiled with an Intel compiler, then all parts touched by the auto-vectorizer are probably still going to execute the SSE2 (or a similarly unoptimized) path on AMD CPUs.

TiborGY:
To be clear, if the MKL_DEBUG_CPU_TYPE=5 envvar trick is used, all MKL functions will execute the AVX2 path, but if there are CPU dispatch routines inserted into the program by the Intel compiler, then code that is outside the MKL functions will probably still only use SSE2 when run on AMD CPUs.

hokru:
I used the MKL DEBUG flag.  The compiler penalties you mentioned were removed at one point (https://github.com/jimenezrick/patch-AuthenticAMD), currently I don't worry about them unless there is a new report on this.

TiborGY:
According to Agner Fog ([https://www.agner.org/forum/viewtopic.php?f=1&t=6](https://www.agner.org/forum/viewtopic.php?f=1&t=6)), these unfair dispatch functions are still present and active in the 2019 Intel compilers.

Benchmarks from Puget Systems also support that this is still happening inside MKL: [https://www.pugetsystems.com/labs/hpc/How-To-Use-MKL-with-AMD-Ryzen-and-Threadripper-CPU-s-Effectively-for-Python-Numpy-And-Other-Applications-1637/](https://www.pugetsystems.com/labs/hpc/How-To-Use-MKL-with-AMD-Ryzen-and-Threadripper-CPU-s-Effectively-for-Python-Numpy-And-Other-Applications-1637/)

loriab:
what should be happening here in the CI is that we've got a mixture of
* Libint, from conda psi4, built with Intel 2019 compilers, not calling MKL
* numpy, from conda defaults, built with GCC, calling MKL
* psi4, built on Azure with GCC or Clang, calling MKL
* MKL, distributed from conda but built by Intel with ? compilers (we'll assume Intel)

My local tests with the reported `MKL_CBWR` are the same as the bullets except psi4 was build with Intel 2019.4 compilers. To test, I've solely been switching out MKL package with `conda install mkl=2019.4` or `=2020.1`

hokru:
Setting MKL_CBWR=AUTO for Intel machines (CI runs on intel for now, right?) seems reasonable. 

I am all for moving forward with this as it blocks other PRs.
We could make an issue about MKL instabilities and keep monitoring it for now?

With supercomputers getting build with AMD Epyc we might need to consider supporting an AMD-oriented build in the future and switch to another BLAS library for it. (i'm not up to date on BLAS stuff there, but things have been improving i think)

tests/scf-cholesky-basis/input.dat
```diff
@@ -15,13 +15,14 @@ e_canonical   = -230.64152543331906
 # indices. Since there's no guarantee that the same functions get
 # picked every time, we need to allow for changes in the energy that
 # go to zero when the tolerance is tightened.
+# MKL 2020.1 can use alternate code path, necessitating relaxing the tolerance from 1.e-8 to 2.e-8, incl. canonical
```
hokru, line 19: Will need an update if it stays like this.

loriab:
> I am all for moving forward with this as it blocks other PRs.
We could make an issue about MKL instabilities and keep monitoring it for now?

I agree. We know how to coerce any MKL into a correct answer. Poor cholesky-basis is exonerated and instead lauded as the coal mine canary. Time to get PRs moving again.

> (CI runs on intel for now, right?)

I don't think we have any guarantee of hardware brand or instruction set. Agree that should try threaded non-MKL again someday.

PR #1900, opened 2020-05-22
Oeprop eval fix

### LINKED ISSUE #1899: Multithreaded execution of compute_esp_over_grid_in_memory produces invalid data
When calling 
oeprop.cc: 1057 compute_esp_over_grid_in_memory for more than one thread garbage is produced.

The reason is most probably that
epot->compute(ints, origin);
is not threadsafe, because it uses a single non thread-local internal buffer.

The solution for the moment should be to remove the 
#pragma omp parallel for
in front of this section.

I will send a pull request with a unit test in a few seconds, which can be used to reproduce and fix this issue. The unit test will be called python-mints14.

timostrunk:
## Description
This pull request fixes #1899 . Multithreaded computation of the ESP potential was invalid, because a buffer_ object was not thread-safe. This pull request disables multithreaded calculation of the ESP on grid for the moment and also adds a unit test to check.
## Todos
No Todos, this is really just a trivial disable.

## Questions
No question

## Checklist
- [X] Tests added for any new features. 
- [X] I ran the libmints tests (http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [X] Ready for review
- [X] Ready for merge


tests/python/mints14/input.py
```diff
@@ -0,0 +1,74 @@
...
+                   2.57027134e-03,  3.22234880e-03,  3.25925100e-03, -3.45599629e-04,
+                   1.12615341e-03,  2.20372521e-03,  2.71976417e-03,  2.79365593e-03,
+                  -1.83114358e-05,  1.00066735e-03,  1.76948007e-03,  2.18338059e-03,
+                   2.29358229e-03]
+
+# Comparison
+for i in range(0, len(points)):
+    psi4.compare_values(esps_1threads[i],esps_4threads[i],3,"Reference value for ESP calculation, 1 thread vs. 4 threads.")
+    psi4.compare_values(esps_1threads[i],reference_esps[i],3,"Reference value for ESP calculation, 1 thread vs. reference calculation.")
+
```
hokru, line 74: It's cleaner with array tools:
`psi4.compare_arrays(esps_1threads,esps_4threads,...`

timostrunk: Thank you, didn't know about this function.
I added a commit.

hokru:
Great to have test for this. 

Shared pointers like [epot](https://github.com/psi4/psi4/pull/1900/files#diff-91bf9e90eb569eec33cd0e5edae2e2b4L1070) in this case are probably a bad idea in OpenMP.
Unfortunately I cannot test threading things on my mac.

@timostrunk how important is threading here, performance wise? Is re-introducing openmp quickly crucial?

timostrunk:
As it never worked before, it is also not crucial to add this again. For large grids the OpenMP speedup at this place would be nearly perfect, because it parallelizes quite trivially. It would lead to a linear speedup in something like an ESP charge fit (the sample code does thankfully not use this function yet, but a rather very slow write to disk, read from disk cycle, so fallout should be quite minimal).

It would be quite simple for me to add threading back to this function if required. The only thing necessary (in my opinion, need to check the code better) would be to make the buffer object in OneBodyAOInt thread-local.

tests/python/mints14/input.py
```diff
@@ -0,0 +1,73 @@
...
+# Number of points to evaluate the ESP on per dimension (reference values below were generated with 5)
+numpoints_per_dim = 5
+points = np.array([ [float(x),float(y),float(z) ] for x in range(0,numpoints_per_dim) for y in range(0,numpoints_per_dim) for z in range(0,numpoints_per_dim) ])
+psi4_matrix = p4c.Matrix.from_array(points)
+
+# We calculate the ESPs with 1 thread
+psi4.set_num_threads(1, quiet = True)
+esps_1threads = np.array(myepc.compute_esp_over_grid_in_memory(psi4_matrix))
+# We calculate the ESPs with 4 threads
+psi4.set_num_threads(4, quiet = True)
```
loriab, line 33: A complication is that the whole test suite is often run `-n<num_cores>`, so extra threads aren't available. But since this is focused on launching the threads, not checking for speedup, I think it'll be ok.

timostrunk:
> please rebase to include #1896 to fix CI.
[x] Done

PR #1909, opened 2020-05-27
Standardize CEPA/LCCD

### LINKED ISSUE #1872: Inconsistent CEPA(0) Definitions Across Psi
For the `fnocc` modules, [`cepa(0)` means `lccsd`](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L4655-L4660). I have only ever heard of `cepa(0)` meaning `lccd`. It is `lccd` in the `dct` module, it is `lccd` in the `occ` module, and it is `lccd` in the `dfocc` module. Can we make this consistent, or at least print a warning?

CEPA(0) is very important to my own research, and I had only ever heard of it as `lccd`. After doing some more digging, I've come across some sources that equate CEPA(0) and `lccsd`, but it seems to me to be a minority position. 

Making this even more complicated, all requests for `cepa(0)` go to `fnocc`. `occ` and `dfocc` require `lccd`. `dct` also requires `cepa0`, but as a keyword rather than the energy name.

Obligatory ping of @edeprince3 for `fnocc` issues and @loriab for defining methods consistently across modules and @jturney because this has been causing problems at CCQC.

JonathonMisiewicz:
## Description
Closes #1872. 

CEPA Conventions Now:
`dct`: Does not know about CEPA(0), per #1873 
`occ`: Every use of CEPA or CEPA(0) is now converted to LCCD. The section header and the final results section include [CEPA(0)] in parentheses. The same applies to the orbital optimized methods. The string "OCEPA" is still used internally as a wavefunction type when referring to LCCD, but that part of `occ` could use refactoring anyways, so I'm not going to spend the time to change it for something that relatively minor.
`dfocc`: Unchanged, because it always said `lccd` anyways. Bozkaya changed his preferred name for the method from CEPA to LCCD between `occ` and `dfocc`
`fnocc`: For consistency with the other CEPA methods, this still uses CEPA(0) to refer to both LCCD and LCCSD for the final results, but the start of the computation writes out Linearized CCD or Linearized CCSD.
Driver: This is unchanged. `cepa(0)` means LCCSD.

Obligatory ping to notify @edeprince3.

## Todos
- [x] Standardizes CEPA terminology in Psi output

## Checklist
- [x] Quick tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


hokru:
Would there be tests (names/descriptions) that need to be adapted?

JonathonMisiewicz:
Oh, I had forgotten. Yes, quite a few `occ` tests. I'm renaming anything from `occ` cepa to `lccd`, changing the `ocepa` test marker to `olccd`, changing the `cepa` marker for occ tests to `lccd`, and adding `lccd` to `cepa-module`, which tests all of them.

However, even after renaming the redirectories, updating `tests/CMakeLists.txt`, and making a new `objdir` from scratch, I'm _still_ getting `ocepa*` test folders. Any ideas why?

loriab:
Definitely in `tests/`, not `samples/`? if the `ocepa*` tests are in objdir, it's because you'd have to rename inside the `tests/formerly-ocepa/CMakeLists.txt` add_test fn.

JonathonMisiewicz:
As usual, Lori, thanks. Test renaming pushed.

JonathonMisiewicz:
For those who have no idea what Lori and I are talking about, we're switching topics from standardizing method names to the fact that she's going to be standardizing method tests, and I have a reference implementation that can be used to confirm those tests are correct. Test standardization is not directly relevant to this PR.

I didn't run comparisons on the LCCD code in `occ`, only the OLCCD code. That said, if you tell me what comparisons you want, I can run them by just commenting out the line in my code that rotates the orbitals. I can do all-electron or frozen core LCCD with conventional integrals. It'll take me a bit of work to get DF-LCCD, but that is something I need to do sooner or later, and this is a decent time for me to do it.

loriab:
* these four mols https://github.com/MolSSI/QCEngine/blob/master/qcengine/programs/tests/standard_suite_ref.py#L12-L40
* rhf for hf & h2o. uhf/rohf for bh3p & nh2.
* cc-pvdz for hf & bh2p. aug-cc-pvdz & cfour-qz2p for h2o & nh2.
* conv, df, cd, though conv only is a fine start
* all-electron (ae) and frozen-core (fc)
* lccd, olccd, lccsd all appreciated, though maybe your code doesn't cover the last
* energies and energy components of first priority, but should you have dipoles, gradients, etc., glad to collect them now

You don't have to do any of this -- I can take values from existing psi modules -- but an independent reference implementation is not to be sniffed at, and you can ensure psi will fail if it doesn't match you from here out.

* either send me numbers (non-redundant; that is, corl, not corl and tot) or PR directly to qcengine (just that one standard_suite_ref.py file)
* I'm only testing to 1.e-6 but refs generated with https://github.com/psi4/psi4/blob/master/tests/pytests/standard_suite_runner.py#L60-L63

I'll merge this PR

PR #1910, opened 2020-05-27
More TDMs from DETCI

### LINKED ISSUE #1890: Select TDMs Inaccessible
Carried forward from [forum post](http://forum.psicode.org/t/calculating-transition-and-state-density-matrices-and-dipole-moment-matrices/1784).

The API to get transition density matrix elements out of a `detci` computation currently fails unless the first state is zero. This is [not expected behavior](https://github.com/psi4/psi4/issues/1638#issuecomment-489080431). Doing `grep` for a minute leads me to think the culprit is [these lines](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/detci/opdm.cc#L113-L117). This looks like an extremely easy fix: just loop over another variable to change the first element in the tuple as well.

Can a developer more experienced in the ways of `detci` (probably @dgasmith) confirm this?

JonathonMisiewicz:
## Description
Closes #1890

## Todos
- [x] When DETCI computes TDMs, it does so for all excitations to higher roots, not just excitations from the ground state.
- [x] Added OPDM and TPDM tests.

## Questions
- [x] ~~Is it worth adding the i > j case, or is it i < j the standard convention for multistate problems? If so, rather than have these be explicitly constructed, it's probably best to just take the adjoint of the i < j case.~~ Added.
- [x] ~~We currently have zero test coverage of OPDMs from `detci`. We could check that Matrices are equal for a same-state OPDM and a TDM. This is a bit larger than usual for a Matrix comparison, but I think it'll be fine with H2/cc-pVDZ. I assume this would go in pytests?~~ Added.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/detci/opdm.cc
+++ b/psi4/src/psi4/detci/opdm.cc
@@ -476,12 +476,31 @@ std::vector<std::vector<SharedMatrix> > CIWavefunction::opdm(SharedCIVector Ivec
             offset += CalcInfo_->ci_orbs[h];
         }
 
-        std::vector<SharedMatrix> opdm_root_vec;
-        opdm_root_vec.push_back(new_OPDM_a);
-        opdm_root_vec.push_back(new_OPDM_b);
-        opdm_root_vec.push_back(new_OPDM);
+        std::vector<SharedMatrix> opdm_root_vec = {new_OPDM_a, new_OPDM_b, new_OPDM};
 
         opdm_list.push_back(opdm_root_vec);
+
+        // Now for the other order.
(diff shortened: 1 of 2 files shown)


CDSherrill:
I'm not sure what the conventions are, if any, for computing transition density matrices in other codes.  Might see what a code like Q-Chem does.  

I agree it'd be nice to have some OPDM test.  Maybe some small test like H2 would suffice.

JonathonMisiewicz:
The outstanding questions are settled, so this is now ready for merge, once reviews come in.

PR #1911, opened 2020-05-29
Omega Combine

obrien951:
## Description
Calculates exchange terms in a single tensor rather than contracting and adding. Reduces the cost of contractions against SCF coefficients in range separated DFT.

Changes the working equation for calculating the coulomb matrix in range separated DFT calculations. This lowers the number of tensors that Psi4 needs to store, reducing memory costs and extending the number of cases operable by DFHelper.

## Todos
- [ ] Extend formulations to DiskDFJK

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -120,8 +120,12 @@ def initialize_jk(self, memory, jk=None):
     jk.set_memory(memory)
     jk.set_do_K(functional.is_x_hybrid())
     jk.set_do_wK(functional.is_x_lrc())
+    jk.set_wcombine(core.get_option("SCF","WCOMBINE"))
```
loriab, line 123: can this be set upon initializing the class instance?

psi4/src/psi4/lib3index/dfhelper.h
```diff
@@ -151,7 +159,21 @@ class PSI_API DFHelper {
     /// @param omega double indicating parameter for other type
     ///
     void set_omega(double omega) { omega_ = omega; }
-    size_t get_omega() { return omega_; }
+    double get_omega() { return omega_; }
```
loriab, line 162: omegas can be negative?

psi4/src/export_fock.cc
```diff
@@ -67,6 +67,9 @@ void export_fock(py::module &m) {
         .def("set_do_K", &JK::set_do_K)
         .def("set_do_wK", &JK::set_do_wK)
         .def("set_omega", &JK::set_omega)
+        .def("set_omega_alpha", &JK::set_omega_alpha)
+        .def("set_omega_beta", &JK::set_omega_beta)
```
loriab, line 71: add a quick docstring and then an extra argument `"beta"_a`. this will mean that in python ppl can call `set_omega_beta(beta=0.4)`. that in itself is uncommon, but handy for docstring.

psi4/src/psi4/libfock/MemDFJK.cc
```diff
@@ -92,7 +102,7 @@ void MemDFJK::compute_JK() {
     if (lr_symmetric_) {
         if (do_wK_) {
             for (size_t N = 0; N < wK_ao_.size(); N++) {
-                wK_ao_[N]->hermitivitize();
+                 wK_ao_[N]->hermitivitize();
```
loriab, line 105: extra space

psi4/src/psi4/libfock/jk.cc
```diff
@@ -103,7 +103,9 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
 
     } else if (jk_type == "MEM_DF") {
         MemDFJK* jk = new MemDFJK(primary, auxiliary);
+        jk->set_wcombine(true);
         _set_dfjk_options<MemDFJK>(jk, options);
+        if (options["WCOMBINE"].has_changed()) { jk->set_wcombine(options.get_bool("WCOMBINE")); }
```
loriab, line 108: I think this should be plain `jk->set_wcombine(options.get_bool("WCOMBINE")` w/o the `if` clause. if you switch that here, do you still need the options line in the driver?

psi4/src/psi4/libscf_solver/rhf.cc
```diff
@@ -406,6 +416,9 @@ std::vector<SharedMatrix> RHF::twoel_Hx(std::vector<SharedMatrix> x_vec, bool co
         throw PSIEXCEPTION("RHF::twoel_Hx: JK object is not initialized, please set option SAVE_JK to True.");
     }
 
+    printf("jk is of type %s\n", jk_->name().c_str());
+    printf("jk_->wcombine() is %s\n", (jk_->get_wcombine() ? "true" : "false" ) );
```
loriab, line 420: mark for removal before final merge

psi4/src/read_options.cc
```diff
@@ -1576,6 +1576,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_str("UHF_NOONS", "3");
         /*- Save the UHF NOs -*/
         options.add_bool("SAVE_UHF_NOS", false);
+        /* combine omega exchange and HF exchange into one matrix */
```
loriab, line 1579: needs to be `/*-` and `-*/` to appease the perl scraper

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -604,31 +611,48 @@ void DFHelper::prepare_AO_wK_core() {
...
-        timer_off("DFH: wAO Copy");
+            // contract half metric inverse
+            timer_on("DFH: AO-Met. Contraction");
+            contract_metric_AO_core_symm(M1p, ppq, metp, begin, end);
+            timer_off("DFH: AO-Met. Contraction");
+    
+            // contract full metric inverse computes  [J^{-1.0}](Q|mn)
+            timer_on("DFH: AO-Met. Contraction");
+            contract_metric_AO_core_symm(M1p, m1ppq, met1p, begin, end);
+            timer_off("DFH: AO-Met. Contraction");
```
loriab, line 644: is there an adv to switching order of half and full in non-wcombine case?

psi4/src/psi4/libfock/jk.h
```diff
@@ -1064,6 +1087,10 @@ class PSI_API MemDFJK : public JK {
     */
     void print_header() const override;
 
+    void set_omega_alpha(double alpha);
+    void set_omega_beta(double beta);
+    void set_wcombine(bool wcombine);
```
loriab, line 1092: these aren't inherited from JK?

obrien951: 2 reasons why these have to be done differently:

1. The actual math is done by the dfhelper object in MemDFJK, so inheriting a function won't change the values there, and I can't program altering values in a dfhelper object in a generic JK object because it doesn't have one.

2. wcombine is only available on MemDFJK right now, so I need to the JK object to throw an error if the user tries to use wcombine on another JK daughter class.

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: None that informed its change

susilehtola:
btw some functionals use several range separation constants, could this be already supported in this PR or should it wait?

obrien951:
@susilehtola Multiple coefficients, exponential constants, and weights is not supported in this PR. You should wait.

psi4/src/psi4/lib3index/dfhelper.h
obrien951: The `size_t` was a typo. The double valued function is the one currently in the code.

psi4/src/psi4/libfock/jk.h
loriab: Maybe I'm really missing the connection. There's two definitions -- one in header for JK class and one in dfhelper header -- but the one for MemDFJK class has only the declaration. where am I missing its implementation?

psi4/driver/procrouting/scf_proc/scf_iterator.py
obrien951: Pytests pass locally with that line removed, but I saw line 122 (`jk.set_do_wK(functional.is_x_lrc())`) in the code before starting this project, so I thought it would make sense to leave the option here out of an abundance of caution and to match the pre existing protocol.

loriab: it's good to keep all the options handling in one spot (back c-side in this case). `jk`, `functional`, `memory` are all local to the function or local to the class; whereas, the `get_options` call to to globals.

psi4/src/psi4/libfock/jk.h
obrien951: Oversight on my part

psi4/src/psi4/libfock/jk.cc
obrien951: I'm sure your suggestion will work, but I'm hesitant to implement it because all the other options in `jk.cc` are handled as line 108 above.

To be completely honest, I don't know whether or not I need the options line in the driver to begin with. I put it there to match the convention with handling of `do_wK`

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: In fact, there's only one contraction in wcombine now, so that shouldn't matter at all! I forgot to delete the code I'd commented out.

psi4/src/psi4/libfock/jk.h
obrien951: Next commit should have the definitions added in `MemDFJK.cc`

psi4/src/psi4/libfock/jk.cc
loriab: the other options in jk.cc are that way b/c the true default is set https://github.com/psi4/psi4/pull/1911/files#diff-822f8557d151459bb012892a642889b8R189-R207, not from read_options.

do_wK in the scf_iterations is querying `functional`, not globals. I'm not seeing a case for this line in the driver.

psi4/src/psi4/libfock/jk.h
loriab: My worry then would be why was this working w/o the definitions. You've run the combinations of wcombine=T/F, and scf_type=mem_df/disk_df to make sure they're taking the code paths you expect?

obrien951: The values were getting set to the dfhelper object in `jk.preiterations()` which is called python side in `jk.initialize()`

psi4/src/psi4/libfock/jk.cc
obrien951: If you don't see the case, the change can be made. My aesthetic preferences may be what they are, but I can go with what you think is less fragile or easier to follow.

obrien951: If I get rid of that conditional, then the default value for `wcombine` in MemDFJK will not be `true` because the global default is `false` . This default was chosen because the other JK classes don't have machinery to combine AO tensors.

psi4/driver/procrouting/scf_proc/scf_iterator.py
obrien951: Latest commit makes the suggested change

psi4/src/psi4/libfock/jk.cc
loriab: below is the goal, right? the read_options summary will need language that the default is most efficient available for the `scf_type` algorithm. earlier opposition to `if has_changed` above retracted. thanks for clearing away the driver-side options querying.

user `wcombine` | eff. `scf_type` | outcome
----|----|----
true | mem_df | omega combined
false | mem_df | omega separate
(unset) | mem_df | omega combined (b/c more efficient)
true | disk_df | NYI error
false | disk_df | omega separate
(unset) | disk_df | omega separate

psi4/src/psi4/lib3index/dfhelper.cc
loriab: Perhaps switch half and full back then so it doesn't look like a meaningful change.

obrien951: I didn't read your comment carefully enough. my bad.

psi4/src/psi4/libfock/jk.cc
obrien951: That is the expected, desired, and achieved behavior with this PR, yes.

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: One of my latest commits fixes this issue

psi4/src/psi4/libfock/jk.cc
```diff
@@ -618,5 +623,13 @@ void JK::compute() {
...
     }
 }
+void JK::set_wcombine(bool wcombine) {
+    wcombine_ = wcombine;
+    if (wcombine) {
+        throw PSIEXCEPTION("To combine exchange terms, use MemDFJK\n");
+    } else {
+        wcombine_ = false;
+    }
+}
```
loriab, line 633: the purpose of this fn is to forbid setting `set_wcombine(true)` unless a derived class overrides it?

can kill off the `else` clause, I think.

obrien951: Yes. That is exactly the purpose.

obrien951: I killed off the else. It's showing up in files changed but not here for whatever reason.

psi4/src/psi4/libfock/MemDFJK.cc
```diff
@@ -125,5 +133,19 @@ int MemDFJK::max_nocc() const {
...
+}
+void MemDFJK::set_omega_beta(double beta){
+    omega_beta_ = beta;
+    dfh_->set_omega_beta(omega_beta_);
+}
 void MemDFJK::set_do_wK(bool tf) { do_wK_ = tf; dfh_->set_do_wK(tf); }
+void MemDFJK::set_wcombine(bool wcombine) { 
+    wcombine_ = wcombine;
+    if (dfh_) {
+        dfh_->set_wcombine(wcombine);
```
loriab, line 148: Is there a reason `dfh_->set_wcombine` is conditional (`if (dfh_) {`) but `set_omega_alpha/beta` above are not?

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -2942,7 +3054,14 @@ void DFHelper::compute_JK(std::vector<SharedMatrix> Cleft, std::vector<SharedMat
...
         Mp = M.get();
     } else
-        Mp = Ppq_.get();
+        if (!wcombine_) Mp = Ppq_.get();
+
+    double* M1p;
+    if (wcombine_) {
+        M1p = m1Ppq_.get();
+    } else {
+        //M1p = nullptr;
```
loriab, line 3063: what's supposed to be happening in this `else {}`?

some safety brackets around the `} else` above probably advisable.

maybe evaluate whether the added/removed newlines were intentional.

psi4/src/psi4/libfock/jk.h
```diff
@@ -430,12 +439,33 @@ class PSI_API JK {
...
+    * Set the alpha value for w exchange: weight for HF Term                
+    * @param omega_alpha HF-Exchange weight
+    */
+    virtual void set_omega_alpha(double alpha) { omega_alpha_ = alpha; }
+
+    /**
+    * Set the alpha value for w exchange: weight for dampened Term                
+    * @param omega_beta Dampened Exchange weight
+    */
+    virtual void set_omega_beta(double beta) { omega_beta_ = beta; }
```
loriab, line 467: You know more about inheritance and virtual conventions than I do, but should the definitions be removed here since they're irrelevant if wcombine _not_ implemented and insufficient if wcombine _is_ implemented?

should omega_alpha_/beta_ have getters?

obrien951: I'd like to leave the setter in there because I plan on putting wcombine in DiskDFJK. You need at least a dummy definition of the function if you don't want an implementation for every subclass, so I thought it made more sense to put this function in here because it's a choice between `{}` and the implementation in the file.

The answer to the second question is a matter of taste. Some schools of thought say that you should always have getters and setters. Others say that you should only use them if they're called in the code.

I haven't implemented a getter for either because they aren't called.

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: Agreed. Removal is probably better.

psi4/src/psi4/libfock/MemDFJK.cc
obrien951: Yes.

I want the function to be able to be called before the JK object is initialized. The DFHelper object is constructed and initialized when the JK object is initialized, so an attempt to use this function without the conditional throws a seg fault.

loriab: aha, thanks!

psi4/src/psi4/libfock/jk.h
loriab: Thanks for the explanation. I guess I like `{}` b/c you never want that code executed (DiskDFJK needs the add'l `dfh_` line like MemDFJK, right?). But ok as-is.

On the second question, good to know the schools of thought. I agree with no getters if it was pure c-side (and if you wanted to access them you'd add getters and recompile anyways). But since there's py-side, it's nice to export them so ppl can probe the object. Up to you.

omega_ and omega_alpha_/beta_ are independent tracks, right? is there any sense in consolidating to omega_alpha_/beta_ only for the member data and preparing for a future of  g/set_omega+omega_alpha or g/set_omega_alpha+omega_beta?

obrien951: I think python side is a good enough reason to add getters. 

omega_alpha_ and omega_beta_ debend on omega_ but not the other way around. Are you suggesting to input a tuple that sets alpha and beta at once or something else?

loriab: no, I wasn't thinking a tuple, just a setup so that two omega* set commands would work for both DiskDFJK and MemDFJK, rather than three non-independent set commands. but may be worth waiting to see how the multi-omegas pan out.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -570,15 +570,22 @@ void DFHelper::prepare_AO_wK_core() {
...
     wPpq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
-    Ppq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
     m1Ppq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
 
+    if (!wcombine_) Ppq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
+
+
     double* wppq = wPpq_.get();
-    double* ppq = Ppq_.get();
+    double* ppq;
```
jturney, line 579: Set this to `nullptr`.  Setting to a known value will help if we ever have to debug this in the future.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -570,15 +570,22 @@ void DFHelper::prepare_AO_wK_core() {
...
+    if (!wcombine_) ppq = Ppq_.get();
+
     double* m1ppq = m1Ppq_.get();
 
     std::unique_ptr<double[]> Qpq(new double[std::get<0>(plargest)]);
+    std::unique_ptr<double[]> Qpq2(new double[std::get<0>(plargest)]);
     double* M1p = Qpq.get();
+    double* M2p = Qpq2.get();
     std::unique_ptr<double[]> metric1;
     double* met1p;
```
jturney, line 590: To be safe, do the same here `nullptr`.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -3107,6 +3223,71 @@ void DFHelper::compute_J(std::vector<SharedMatrix> D, std::vector<SharedMatrix>
         }
     }
 }
+void DFHelper::compute_J_combined(std::vector<SharedMatrix> D, std::vector<SharedMatrix> J, double* Mp, double* T1p, double* T2p, std::vector<std::vector<double>>& D_buffers, size_t bcount, size_t block_size) {
+//    double* coul_metp = metric_prep_core(1.0);
+    double ZERO[16] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
```
jturney, line 3228: Why is this hard-coded to 16 elements and why is it needed?

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -3107,6 +3223,71 @@ void DFHelper::compute_J(std::vector<SharedMatrix> D, std::vector<SharedMatrix>
...
+        get_tensor_(std::get<0>(files_[filename]), coul_metp, 0, naux_ - 1, 0, naux_ - 1);
+    } else
+        coul_metp = metric_prep_core(1.0);
+
+
+    for (size_t i = 0; i < J.size(); i++) {
+        double* Dp = D[i]->pointer()[0];
+        double* Jp = J[i]->pointer()[0];
+        C_DCOPY(naux_ *  nthreads_, zerop, 0, T1p, 1);
+        C_DCOPY(naux_, zerop, 0, T2p, 1);
```
jturney, line 3247: In these two lines you're copying `naux_ * nthreads_` and `naux_` elements from the **16** element ZERO array.  Are you simply trying to zero out `T1p` and `T2p`? If so, use `memset(T1p, 0, sizeof(double) * naux_ * nthreads_` for the first one and something similar for `T2p`.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -3107,6 +3223,71 @@ void DFHelper::compute_J(std::vector<SharedMatrix> D, std::vector<SharedMatrix>
...
+            C_DGEMV('N', block_size, sp_size, 1.0, &Mp[jump], sp_size, &D_buffers[rank][0], 1, 1.0, &T1p[rank * naux_], 1);
+        }
+        //reduce
+        for (size_t k = 1; k < nthreads_; k++) {
+            for (size_t l = 0; l < naux_; l++) { T1p[l] += T1p[ k * naux_ + l ]; }
+        }
+        //metric contraction
+        C_DGEMV('N', naux_, naux_, 1.0, coul_metp, naux_, T1p, 1, 0.0, T2p, 1);
+
+        C_DCOPY(nbf_ * nbf_, zerop, 0, T1p, 1);
```
jturney, line 3272: Same question here as on 3246 and 3247.

psi4/src/psi4/libfock/jk.h
obrien951: This function definition should be sufficient for DiskDFJK. Would it be ok to just leave this one in there for that?

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: ok

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: ok

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: I need to zero out some arrays, and I use C_DCOPY to do so.  A fast copying code might try to copy one cache-length at a time, so I wanted one cache-length worth of memory past an address to be set to zeros in case a C_DCOPY implementation does that.

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: Yes exactly, and ok.

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: I saw your comment below, so I'll be getting rid of this.

psi4/src/psi4/lib3index/dfhelper.cc
obrien951: memset looks like the right thing to do here as well

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -570,19 +570,26 @@ void DFHelper::prepare_AO_wK_core() {
     std::pair<size_t, size_t> plargest = pshell_blocks_for_AO_build(memory_, 1, psteps);
 
     wPpq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
-    Ppq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
     m1Ppq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
 
+    if (!wcombine_) Ppq_ = std::unique_ptr<double[]>(new double[big_skips_[nbf_]]);
```
robertodr, line 575: Small nitpick: could you switch these constructors to use `make_unique` instead?
```
Ppq_ = std::make_unique<double[]>(big_skips_[nbf_]);
```

obrien951: I will be making the changes. Someone might want to take a look at the rest of dfhelper.cc to make this change to another group of `unique_ptr` 's. I don't know what all the rest of the code does, so I don't think I should submit those changes in this PR.

obrien951: I ended up making changes only to functions touched by this PR. I'll leave the other functions to people who know what to do with them.

psi4/src/export_fock.cc
```diff
@@ -67,9 +67,13 @@ void export_fock(py::module &m) {
...
         .def("set_do_wK", &JK::set_do_wK)
         .def("set_omega", &JK::set_omega)
+        .def("get_omega", &JK::get_omega)
+        .def("set_wcombine", &JK::set_wcombine)
+        .def("get_wcombine", &JK::get_wcombine)
         .def("set_omega_alpha", &JK::set_omega_alpha, "Weight for HF exchange term in range-separated DFT", "alpha"_a)
+        .def("get_omega_alpha", &JK::get_omega)
         .def("set_omega_beta", &JK::set_omega_beta, "Weight for dampened exchange term in range-separated DFT", "beta"_a)
-        .def("set_wcombine", &JK::set_wcombine)
+        .def("get_omega_beta", &JK::get_omega)
```
loriab, line 76: get_omega_alpha/beta need to have get_omega_alpha/beta as 2nd arg, not get_omega. may as well add docstring, too.

obrien951: Thank you. I'm sorry I let the typo slip in. I must have gotten carried away with vim tricks

loriab: np. at least you didn't create a cubic black hole with ~radius 1 bohr (see #1938)

loriab:
rebase will fix CI

psi4/src/psi4/libscf_solver/uhf.cc
```diff
@@ -208,7 +207,7 @@ void UHF::form_G() {
     double alpha = functional_->x_alpha();
     double beta = functional_->x_beta();
 
-    if (alpha != 0.0) {
+    if ( alpha != 0.0 && !(functional_->is_x_lrc() && jk_->name() == "MemDFJK" && jk_->get_wcombine()) ){
```
dgasmith, line 210: It might be best if `jk_->wcombine()` or similar is for every JK object which normally returns false. This could be useful for Direct as well.

obrien951: The memory checker does make the correct estimation based on this new operation.

As far as testing it goes, the only way would be to set the memory in the test script and to make sure that psi gives back the right value, but that could be problematic if someone messes with the fudge factor and gibi vs. giga bytes.

The complicated logic in that if statement is necessary because there are conditions that can be taken for granted without wcombine that can't without it. 

It would be useful for direct, but I wanted to wait on the libint2 branch to get sorted before I did anything with that.

dgasmith: I'm not suggesting that you implement it into Direct at the moment, but more of the point that `jk_->name() == "MemDFJK" && jk_->get_wcombine()` is a leaky abstraction and creates a headache throughout the code. This `get_wcombine` should be native to the JK base object and return false for all classes where this isn't implemented.

Otherwise if you add this to direct we will need to add additional logic to every instance where you wrote this.

obrien951: `jk_->get_wcombine()` will already return false for all classes that aren't `MemDFJK` as you've suggested. I included the condition `jk_->name() == "MemDFJK"` because I didn't think I would need the `!jk_->get_wcombine()` case when I started working on the code.

I'm going to try getting rid of the `jk_->name() == "MemDFJK"` condition. If I understand the code correctly, it should work.

obrien951: I made the change and tested it. It seems to be working just fine.

obrien951: fyi, commit message would more precisely say "condition" than "conditional"

dgasmith:
Looks good, thanks for the changes.

PR #1919, opened 2020-06-03
CGRSolver Minor Commenting, Bug Fix

### LINKED ISSUE #1918: `CGRSolver` "Subspace" Preconditioner Broken
The following input file
```
molecule {
O
H 1 1.0 
H 1 1.0 2 104.5
}

set solver_precondition subspace
set solver_n_guess 3

gradient('mp2/cc-pvdz')
```
segfaults due to a bad access exception. While the segfault occurred on [this line](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L290), the real problem is earlier. The loop structure [assumes](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L280-L285) that `x_` and `b_` have the same number of elements. But the last method we ran [was `setup`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L163-L164), which [clears `x_`](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L276).

Even then, `setup` appears to have its own, unrelated problem. The preconditioner submatrix `A_` is [populated using the results of a `products_x()` call](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L259-L272). But `x_` [consists of zero vectors](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L251), so the submatrix guess procedure is just an overcomplicated way to write zero at present. That is probably very bad for the algorithm.

Making matters even worse, I have not been able to find a standard conjugate gradients preconditioner with a name similar to "SUBSPACE," so I cannot figure out how this algorithm was supposed to work.

In short, this code is broken, and I would say beyond repair. Can I eliminate all the code for the "SUBSPACE" preconditioner from the `CGRSolver`? We'd just fall back to the "JACOBI" preconditioner, which is the default anyways.

JonathonMisiewicz:
## Description
Fixes #1918 by removing the subspace preconditioner, per @robertodr's recommendation.

This also adds some more comments to the code, crucially, the fact that this solves the problem `Hx=b` as opposed to `Hx+b=0`. Signs are important.

## Todos
- [x] Disables the broken subspace preconditioner in the CGRSolver.

## Checklist
- [x] Tested that a helpful warning runs if the user specifies a subspace preconditioner for the CGRSolver.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/solver.cc
+++ b/psi4/src/psi4/libfock/solver.cc
@@ -90,7 +90,14 @@ std::shared_ptr<CGRSolver> CGRSolver::build_solver(Options& options, std::shared
         solver->set_bench(options.get_int("BENCH"));
     }
     if (options["SOLVER_PRECONDITION"].has_changed()) {
-        solver->set_precondition(options.get_str("SOLVER_PRECONDITION"));
+        if (options.get_str("SOLVER_PRECONDITION") == "SUBSPACE") {
+            outfile->Printf("  !!!Warning!!!\n")
+            outfile->Printf("  The subspace preconditioner has been broken for some time and was removed in Psi4 1.4.\n");
+            outfile->Printf("  Setting preconditioner to Jacobi instead.\n\n");
+            solver->set_precondition("JACOBI");
+        } else {
+            solver->set_precondition(options.get_str("SOLVER_PRECONDITION"));
+        }
     } else if (options["SOLVER_MAXITER"].has_changed()) {
         solver->set_maxiter(options.get_int("SOLVER_MAXITER"));
     }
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz:
Deprecating it there would have been my first choice, but isn't a viable solution to this one. Rob created many other solvers which also use that keyword, such as [this one](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/solver.cc#L1060). Blocking there would disable the keyword for those solvers, too. I am already several rabbit holes deep, so I am not going to check whether subspace works in his _other_ solvers.

PR #1922, opened 2020-06-06
Update CPPE

maxscheurer:
## Description
This PR updates the `cppe` interface code (including option handling) to the newest version `v0.2.1` (major performance improvement 🚀). 

## Todos
- [x] Updated interface, options
- [x] Add options for Thole damping

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1517,8 +1517,9 @@ def scf_helper(name, post_scf=True, **kwargs):
                        """further calculations in C1 point group.\n""")
         # PE needs information about molecule and basis set
         pol_embed_options = solvent.pol_embed.get_pe_options()
-        core.print_out(""" Using potential file {} for Polarizable Embedding
-                       calculation.\n""".format(pol_embed_options["potfile"]))
+        core.print_out(f""" Using potential file
+                       {pol_embed_options["potfile"]}
+                       for Polarizable Embedding calculation.\n""")
         scf_wfn.pe_state = solvent.pol_embed.CppeInterface(
             molecule=scf_molecule, options=pol_embed_options,
             basisset=scf_wfn.basisset()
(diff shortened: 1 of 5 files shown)


psi4/driver/procrouting/proc.py
```diff
@@ -1518,7 +1518,7 @@ def scf_helper(name, post_scf=True, **kwargs):
         # PE needs information about molecule and basis set
         pol_embed_options = solvent.pol_embed.get_pe_options()
         core.print_out(""" Using potential file {} for Polarizable Embedding
-                       calculation.\n""".format(pol_embed_options.potfile))
+                       calculation.\n""".format(pol_embed_options["potfile"]))
```
robertodr, line 1521: Would you mind changing this to an f-string?

maxscheurer: Done with pleasure 😄

andysim:
Just out of interest, can the power or R used in the Thole damping exponential be changed?

maxscheurer:
> Just out of interest, can the power or R used in the Thole damping exponential be changed?

I guess anything could in principle be customized... (probably not in the current version) 😄 Anything specific you have in mind?

andysim:
Thanks for the reply.  I don't have any immediate needs, but it might be something for the future.  The AMOEBA force field uses R^3 decay in the exponent, but the CHARMM Drude force field uses R instead.  I'm interested in using CPPE as we move towards QM/MM for these, so it would be nice to be able to set that eventually.

maxscheurer:
For Amoeba, we have the Python code to generate the interaction tensors, so that's in principle easy. It's not in there yet because I almost never use damping myself 😄 Let me know if you need anything (Slack, issue on [CPPE repo](https://github.com/maxscheurer/cppe)...)

PR #1925, opened 2020-06-09
Moves import of qcelemental in main Psi4 script to be later in the file.

jturney:
## Description
The main script (`run_psi4.py` in the source tree or `psi4` when installed) was importing qcelemental before it had a chance to modify the Python search path.  If you didn't already have a copy of qcelemental in your Python search path would cause an error. Encountered this on our cluster while trying to run the tests.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/run_psi4.py
+++ b/psi4/run_psi4.py
@@ -36,7 +36,6 @@ import datetime
 import argparse
 from argparse import RawTextHelpFormatter
 from pathlib import Path
-import qcelemental as qcel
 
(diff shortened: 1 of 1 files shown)


jturney:
@loriab Exactly right.

andysim:
I sense you're just trying to one-up @JonathonMisiewicz and his few character diff to fix frozen virtual gradients, by adding a 0 character change bugfix

PR #1926, opened 2020-06-10
close detci files so independent psiapi calcs possible

loriab:
## Description
now we can have >1 detci test in the whole pytest suite

## Todos
- [x] cepa directed to fnocc was running conv when user requested (unavailable) df. now stops. thanks, @JonathonMisiewicz 
- [x] close detci files so independent psiapi calcs possible

## Questions
- [x] @CDSherrill should `CI_FILE_START` be shifted so others aren't confused by seeming overlap with https://github.com/psi4/psi4/blob/master/psi4/include/psi4/psifiles.h#L77-L81 ? maybe 350

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -573,7 +573,7 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
 
         /*- What file do we start at for hd/c/s/d CIvects? Should be 50 for normal
         CI calculations and 54 if we are going to do a second monomer. !expert -*/
-        options.add_int("CI_FILE_START", 50);
+        options.add_int("CI_FILE_START", 350);
 
         /*- Guess vector type.  Accepted values are ``UNIT`` for a unit vector
         guess (|detci__num_roots| and |detci__num_init_vecs| must both be 1); ``H0_BLOCK`` to use
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -571,8 +571,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
 
         /*- SUBSECTION Guess Vectors -*/
 
-        /*- What file do we start at for hd/c/s/d CIvects? Should be 50 for normal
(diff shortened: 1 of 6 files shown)


psi4/src/read_options.cc
```diff
@@ -573,7 +573,7 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
 
         /*- What file do we start at for hd/c/s/d CIvects? Should be 50 for normal
         CI calculations and 54 if we are going to do a second monomer. !expert -*/
```
PeterKraus, line 575: So was this overwriting the `MRCC` files before? Anyway, could you modify the comment too before merging, this way it looks like 350 is a typo.

loriab: good point on editing the comment. no, it wasn't interfering with psimrcc files. what would happen is if you did a larger calc after a smaller, like a frozen-core followed by all-electron, it would error on the second with the below:

```
PSIO_ERROR: Attempt to write into next entry: 51, buffer_ 0
PSIO_ERROR: unit = 51, errval = 18
PSIO_ERROR: 18 (Incorrect block end address)
```

The detci.1 file mentions file50 and file51, but those numbers in `psifiles.h` have nothing to do with the calc. And there is no file51 physically present on disk that I've been able to see -- it's all in psio's imagination. But what does happen (before PR) is that `CI_FILE_START` sets the index and then detci claims 4 files starting with that number. They are managed with normal psio commands but aren't closed, so for succeeding jobs, https://github.com/psi4/psi4/blob/master/psi4/src/psi4/detci/civect.cc#L1721 fails and fresh files aren't triggered for independent calculations. We thought this might have been a hangover from the psi3-->psi4 translation, but using an option for the starting file number actually came about between beta5 and v1.0. It may have something to do with fci-sapt, in which case @dgasmith would know more.

On the whole, this is just a patch-up job. But apart from tests, CI is unlikely to be run en masse. Hopefully UHF CC gradients, which @hokru and I have also seen fail at successive calcs, can be similarly tricked.

PR #1929, opened 2020-06-13
convert MDI to optional dependency

loriab:
## Description
as it says on the tin

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Me neither. Oh well, easier for me to patch up than outsider anyway. Plus, I finally figured out why `@using_snsmp2`, `@using_cppe`, etc. weren't enabling/skipping tests as they ought -- will be fixed in next qcel.

PR #1930, opened 2020-06-13
Fixed ECP and f-electron combinations.

PeterKraus:
## Description
In some basis sets that contain ECPs, the "valence" f-shell is included in the effective core. This seems true for all post-lanthanide/actinide atoms. 

In current master, those 60 ECP electrons are always frozen, but in a non-ECP basis with frozen core we'd freeze only 54 of those. The simple solution is to always freeze the 4f electrons (for 6th period) if the shell is full.

## Checklist
- [x] `ctest -L ecp` passes.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -250,6 +250,10 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
                 if (largest_shell < current_shell) {
                     largest_shell = current_shell;
                 }
+                // If center is a post-lanthanide in > 6th period, freeze its f electrons too
+                if (current_shell > 5) {
+                    if ((Z + ECP - delta) >= 18 ) delta += 14;
+                }
                 // If this center has an ECP, some electrons are already frozen
                 if (ECP > 0) delta -= ECP;
                 // Keep track of current valence electrons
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz:
CCQC is currently going through a heavy metals phase, so I want to be clear on how this PR will affect us.

1. If we freeze the 54 electrons of Xenon and then the 4f electrons, wouldn't we be freezing 68 electrons rather than 60?

2. I think [the comment](https://github.com/psi4/psi4/pull/1930/files#diff-19caeae917feb299f8f5e72bc1e28436R253) would be clearer as `If center is after the inner transition metals, freeze those f electrons too`. As written, the comment doesn't apply to the actinides, and `> 6th period` doesn't apply to the example of Bismuth, which is 6th period. Am I misunderstanding the comment?

susilehtola:
IMHO "lanthanides and actinides" is clearer than "inner transition metals".

PeterKraus:
Yes, this is a good point. In my opinion, `freeze_core = 1` is meant to be best-effort behaviour, and freezing 4f for post-lanthanides and 5f for post-actinides will have the additional benefit of freezing all electrons contained in most ECPs, and the 5s5p / 6s6p shell for lanthanides/actinides. This means `freeze_core = 1` will get you the same active occupied orbitals whether you have an ECP or not.

The previous behaviour is still accessible using `freeze_core = -1` which would freeze [Xe] strictly for all elements in 6th period and [Rn] for all in 7th period. The behaviour here with a 60-core ECP is clearly nonsensical, but as it's an "advanced option" I'm fine with that.

PR #1933, opened 2020-06-18
Keep _default_psio_manager_ alive for Process::Environment

bennybp:
The global `_default_psio_manager_` shared_ptr can destruct
before Process::Environment. However, some destructors called
from Process::Environment may require it. So have Process::Environment
keep a copy of the shared_ptr

## Description
Some tasks being run by QCArchive/QCEngine would hang indefinitely, even though the calculation finished. GDB showed memory corruption when destructing the Process::Environment.

The cause was related to the destructor for DIISManager calling PSIO::close

https://github.com/psi4/psi4/blob/8d1a8787504a481f61956fb22b282b6f30005ebc/psi4/src/psi4/libpsio/close.cc#L82 

`PSIOManager::shared_object()` returns a copy of the global shared_ptr `_default_psio_manager_`. The main issue is that `_default_psio_manager_` can be destructed before Process::Environment, as order of destruction is not well-defined for globals.

The easiest solution is for Process::Environment to hold a copy of the shared_ptr for the global psio manager, even if it is not used. This prevents premature destruction.

How this issue caused indefinite hanging/locking isn't very clear, but crazy stuff can happen with memory corruptions.

## Todos
- [X] Fix crashing/hanging in certain instances due to using already destructed global objects

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -955,10 +955,6 @@ bool psi4_python_module_initialize() {
         return true;
     }
 
-    // Setup the environment
-    Process::environment.initialize();  // Defaults to obtaining the environment from the global environ variable
-    Process::environment.set_memory(524288000);
-
     // There should only be one of these in Psi4
     Wavefunction::initialize_singletons();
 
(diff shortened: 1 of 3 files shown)


bennybp:
> Thanks for fix! Shocking that it hadn’t caused ill effects before.
> 
> Any reason the manager keepalive is a preceding underscore rather then trailing underscore?

Pure muscle memory, and I guess because `_default_psio_manager_` is what I was mirroring. I can change it if you'd like

loriab:
> Pure muscle memory, and I guess because _default_psio_manager_ is what I was mirroring. I can change it if you'd like

ok, I was just curious about conventions. restarted the single azure job, so no need for change commit.

PR #1935, opened 2020-06-22
module provenance of Wavefunction

loriab:
## Description
Some info about module-level provenance of a Wfn would be handy.

## Todos
- [x] add `Wavefunction.module_` member data and export
- [x] add setting to most modules (not psimrcc or fockci)
- [x] incorporate into stdsuite testing

## Questions
- [x] @dgasmith I've given mcscf its own name. prefer "detci"?
- [x] @maxscheurer I edited a couple error messages to ADCC from ADC. ok?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfocc/dfocc.cc
+++ b/psi4/src/psi4/dfocc/dfocc.cc
@@ -51,6 +51,8 @@ void DFOCC::common_init() {
     print_ = options_.get_int("PRINT");
     if (print_ > 0) options_.print();
 
+    module_ = "dfocc";
+
     cc_maxiter = options_.get_int("CC_MAXITER");
     mo_maxiter = options_.get_int("MO_MAXITER");
     num_vecs = options_.get_int("MO_DIIS_NUM_VECS");
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -288,6 +288,7 @@ def _core_wavefunction_to_file(wfn, filename=None):
         },
         'string': {
             'name': wfn.name(),
+            'module': wfn.module(),
             'basisname': wfn.basisset().name()
         },
         'boolean': {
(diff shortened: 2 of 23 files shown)


maxscheurer:
I would favor "adcc" instead of "ADCC". You agree, @mfherbst?

dgasmith:
I think best to keep it `detci` since the `mcscf` bit is agnostic to the underlying CI code.

psi4/src/psi4/dfocc/dfocc.cc
```diff
@@ -51,6 +51,8 @@ void DFOCC::common_init() {
     print_ = options_.get_int("PRINT");
     if (print_ > 0) options_.print();
 
+    module_ = "dfocc";
```
dgasmith, line 54: Just to double check:
- Inherits from Wavefunciton uses `module_ = ...`
- Does not inherit from Wavefunction uses `->set_module()`?

loriab: or, `.set_module()` py-side, yes. I used whatever was closest to the class. No strong opinions, though.

mfherbst:
I agree "adcc" is better ;). Thanks for the cleanup @loriab!

loriab:
> Is SAPT(DFT)/SF-SAPT included with the sapt.cc changes? Since they are mainly python procedures.

I just fixed plain SAPT in the latest commit, as it has to be py-side. I'm not sure what to do about sf-sapt and dft-sapt at this point. Leaving those for now unless other have ideas.

> How are plugins handled? Especially endorsed ones like sns-mp2, v2rdm_cassscf etc.

They'll have to set their own name like https://github.com/piecuch-group/cct3/pull/2/files#diff-dd599d9e2da08b736cbf0b96add9a840R414 or in the pymodule.py file. I've done it for cct3 and adcc (whose py is in psi4/driver) and caspt2 but will have to adjust the other ones you mentioned. And some like pcmsolver, pe, efp/scf won't get recorded. When plugins move to cookiecutter or similar, the module stamping should get baked into the template.

loriab:
> Could go py-sapt (py + procrouting name) perhaps.

sounds reasonable. I'm inclined to leave it blank until needed.

Issue #1903, opened 2020-05-24
frequencies with nonphysical isotopes


hokru:
Example:
```
import psi4
weird = psi4.core.Molecule.from_arrays(
    geom=[0, 0, 0, 2, 0, 0, 0, 2, 0],
    elbl=['O', 'H@1000.0', 'H'], nonphysical=True,
)
ene = frequencies('hf/cc-pvdz', molecule=weird,dertype=0)
masses = [weird.mass(iat) for iat in range(weird.natom())]
print(ene, masses)
```

Needed changes (by @loriab):
[edit: no longer needed]
```
diff --git a/psi4/driver/qcdb/molecule.py b/psi4/driver/qcdb/molecule.py
index f8f7ea6..4ea64c1 100644
--- a/psi4/driver/qcdb/molecule.py
+++ b/psi4/driver/qcdb/molecule.py
@@ -1470,11 +1470,11 @@ class Molecule(LibmintsMolecule):
             # * this can legitimately happen if total chg or mult has been set
             #   independently b/c fragment chg/mult not reset. so try again.
             print(
-                """Following warning is harmless if you've altered chgmult through `set_molecular_change` or `set_multiplicity`. Such alterations are an expert feature. Specifying in the original molecule string is preferred."""
+                """Following warning is harmless if you've altered chgmult through `set_molecular_change` or `set_multiplicity`. Such alterations are an expert feature. Specifying in the original molecule string is preferred. Nonphysical masses may also trigger the warning."""
             )
             molrec['fragment_charges'] = [None] * len(fragments)
             molrec['fragment_multiplicities'] = [None] * len(fragments)
-            validated_molrec = qcel.molparse.from_arrays(speclabel=False, verbose=0, domain='qm', **molrec)
+            validated_molrec = qcel.molparse.from_arrays(speclabel=False, nonphysical=True, verbose=0, domain='qm', **molrec)
             forgive.append('fragment_charges')
             forgive.append('fragment_multiplicities')
         compare_molrecs(validated_molrec, molrec, 'to_dict', atol=1.e-6, forgive=forgive, verbose=0)

```

Uncomment:
https://github.com/MolSSI/QCElemental/blob/master/qcelemental/molparse/from_schema.py#L85
For numerical hessians (FINDIF) one needs to set `nonphysical=True,` above as a work around or figure out the following error:

```
File "/Users/kruse/qc/psi4.bin/lib/psi4/driver/driver_findif.py", line 417, in assemble_gradient_from_energies
[...]

PR #1937, opened 2020-06-23
stdsuite roundup for lccd

loriab:
## Description
finish off low-cepa testing and API standardization

## Todos
- [x] df-lccd from dfocc (tested against @JonathonMisiewicz's p4n impl)
- [x] cd-lccd from dfocc (tested against itself)
- [x] added ability to do freq with nonphysical masses #1903 (needs qcel 0.15.1)
- [x] bumped qcel to v0.15.1
-  changes to test_qcel_molparse_to_string are formatting, and updating against expanded test file in qcel. no need for review.
- [x] bumped qcng to v0.15.0
- [x] activated stdsuite. only a couple min extra tests are run in quick

## Questions
- [x] @bozkaya, you weren't printing spin components for CD-LCCD as you do for DF-LCCD, but the variables `ElccdAA`, etc. were present and contained plausible numbers for UHF. I went ahead and stored the variables so CD could be as capable as DF. Is this sound? (I'll repeat this Q in a separate comment b/c I'm not sure of GH notifications upon edit.)

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
@bozkaya, you weren't printing spin components for CD-LCCD as you do for DF-LCCD, but the variables `ElccdAA`, etc. were present and contained plausible numbers for UHF. I went ahead and stored the variables so CD could be as capable as DF. Is this sound?

bozkaya:
Sounds good Lori! @loriab

psi4/src/psi4/dfocc/manager.cc
```diff
@@ -3485,23 +3485,25 @@ void DFOCC::lccd_manager() {
...
     outfile->Printf("\t======================================================================= \n");
 
-    Process::environment.globals["MP2 TOTAL ENERGY"] = Emp2;
+    variables_["MP2 TOTAL ENERGY"] = Emp2;
     Process::environment.globals["SCS-MP2 TOTAL ENERGY"] = Escsmp2;
     Process::environment.globals["SOS-MP2 TOTAL ENERGY"] = Esosmp2;
     Process::environment.globals["SCS(N)-MP2 TOTAL ENERGY"] = Escsnmp2;
-    Process::environment.globals["MP2 CORRELATION ENERGY"] = Emp2 - Escf;
+
+    variables_["MP2 CORRELATION ENERGY"] = Emp2 - Escf;
```
PeterKraus, line 3493: What's the reasoning behind this change? I don't quite get why we're shifting some variables into `variables_` out of the global namespace, but not others. Is it to attach them to wfn?

loriab: Exactly on the attaching qcvars to wfn. `variables_` sets on wfn, then in proc.py, wfn vars get copied into `P::e`. That's for the variable names I can vouch for. My current focus is totals and components. SCS variations are another pass, and I may kill off some of them so that `Wfn.variables()` is a fresh start from `core.variables()`. Someday `P::e.globals` will be only a pseudo-global that gets written to at the end of `energy()`, etc. for user convenience.

PeterKraus: That makes sense. Thanks.

tests/pytests/test_qcel_molparse_to_string.py
```diff
@@ -92,87 +90,104 @@
...
+        (("subject2", {"dtype": "orca", "units": "bohr"}), "ans2_orca_au"),
+        (("subject2", {"dtype": "orca", "units": "angstrom"}), "ans2_orca_ang"),
+        (("subject2", {"dtype": "turbomole", "units": "bohr"}), "ans2_turbomole_au"),
+        (("subject2", {"dtype": "nglview-sdf"}), "ans2_ngslviewsdf"),
+        (("subject2", {"dtype": "qchem", "units": "bohr"}), "ans2_qchem_au"),
+    ],
+)  # yapf: disable
 def test_to_string_xyz(subjects, inp, expected):
-    if subjects == 'pmol':
+    if subjects == "pmol":
```
andysim, line 131: Just out of interest is there a reason to prefer the double quote in the python layer?  Obviously it matters in the C++ layer but I tend to use a mixture in python.

loriab: No reason. In fact I personally prefer the single quote. But QCA has switched to `black` from `yapf`, and that automatically double-quotes things. I don't care for the best-thing-since-sliced-bread that black adherents exhibit, but I do have to disable formatting less with black than with yapf. This particular file is a partial copy of a test from qcel, so it made sense to reformat it.

PR #1946, opened 2020-07-08
Threaded external charge potential

chapincavender:
## Description
This PR threads the calculation of one-electron potential integrals from external point charges. It replaces a call to `OneBodyAOInt.compute()`, which is not threaded and computes the full N-by-N matrix, with a new threaded `for` loop over pairs of shells so that only the lower triangle of the symmetric matrix is calculated. Here's a quick benchmark of times for a SAPT0/jun-cc-pVDZ calculation of a nucleic acid base pair (144 electrons, 425 basis functions, 1879 auxiliary basis functions) with 7393 external point charges on an Intel Xeon E5-2695v4 with 18 cores in each of 2 sockets:

Cores | Old code, no charges | Old code, with charges | New code, no charges | New code, with charges
----- | ------- | ------- | ------- | -------
1 | 1.00 | 1.00 | 1.00 | 1.00
2 | 1.65 | 1.08 | 1.71 | 1.68
4 | 2.43 | 1.14 | 2.64 | 2.61
8 | 3.25 | 1.20 | 3.08 | 4.01
16 | 5.29 | 1.24 | 2.61 | 2.90
32 | 5.36 | 1.23 | 4.24 | 7.04

Each column shows the speedup relative to the calculation with 1 core (i.e. t_1 / t_c). With charges, the performance of the old code scales poorly with additional cores; the performance of the new code matches or exceeds the scaling of the calculation without charges.

## Questions
- [x] The resulting matrix `V_charge` is always symmetric in my tests. Are there any cases where this would not be true?
- [ ] I opted to replace `force_cartesian_ ? shell(i).ncartesian() : shell(i).nfunction()` from `OneBodyAOInt.compute()` with `shell(i).nfunction()`. Are there any cases where `force_cartesian_` needs to be checked for this potential?

## Checklist
- [x] Passes smoke tests and `fsapt-ext` test added in #1608

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/extern.cc
+++ b/psi4/src/psi4/libmints/extern.cc
@@ -97,28 +97,73 @@ SharedMatrix ExternalPotential::computePotentialMatrix(std::shared_ptr<BasisSet>
     auto V = std::make_shared<Matrix>("External Potential", n, n);
     auto fact = std::make_shared<IntegralFactory>(basis, basis, basis, basis);
 
+    // Thread count
+    int nthreads = 1;
+#ifdef _OPENMP
+    nthreads = Process::environment.get_n_threads();
+#endif
+
     double convfac = 1.0;
     if (basis->molecule()->units() == Molecule::Angstrom) convfac /= pc_bohr2angstroms;
 
     // Monopoles
     auto V_charge = std::make_shared<Matrix>("External Potential (Charges)", n, n);
+    double **V_charge_p = V_charge->pointer();
 
     auto Zxyz = std::make_shared<Matrix>("Charges (Z,x,y,z)", charges_.size(), 4);
     double **Zxyzp = Zxyz->pointer();
-    for (size_t i = 0; i < charges_.size(); i++) {
+    for (size_t i = 0; i < charges_.size(); ++i) {
         Zxyzp[i][0] = std::get<0>(charges_[i]);
         Zxyzp[i][1] = convfac * std::get<1>(charges_[i]);
         Zxyzp[i][2] = convfac * std::get<2>(charges_[i]);
         Zxyzp[i][3] = convfac * std::get<3>(charges_[i]);
     }
 
-    std::shared_ptr<PotentialInt> pot(static_cast<PotentialInt *>(fact->ao_potential()));
-    pot->set_charge_field(Zxyz);
-    pot->compute(V_charge);
+    std::vector<std::shared_ptr<PotentialInt> > pot;
+    for (size_t t = 0; t < nthreads; ++t) {
+        pot.push_back(std::shared_ptr<PotentialInt>(static_cast<PotentialInt *>(fact->ao_potential())));
+        pot[t]->set_charge_field(Zxyz);
+    }
+
+    // Monopole potential is symmetric, so generate unique pairs of shells
+    std::vector<std::pair<size_t, size_t> > ij_pairs;
+    for (size_t i = 0; i < basis->nshell(); ++i) {
+        for (size_t j = 0; j <= i; ++j) {
+            ij_pairs.push_back(std::pair<size_t, size_t>(i, j));
+        }
+    }
+
+    // Calculate monopole potential
+#pragma omp parallel for schedule(dynamic) num_threads(nthreads)
+    for (size_t p = 0; p < ij_pairs.size(); ++p) {
+        size_t i = ij_pairs[p].first;
+        size_t j = ij_pairs[p].second;
+        size_t ni = basis->shell(i).nfunction();
+        size_t nj = basis->shell(j).nfunction();
+        size_t index_i = basis->shell(i).function_index();
+        size_t index_j = basis->shell(j).function_index();
+
+        size_t rank = 0;
+#ifdef _OPENMP
+        rank = omp_get_thread_num();
+#endif
+
+        const double *buffer = pot[rank]->buffer();
+        pot[rank]->compute_shell(i, j);
+
+        size_t index = 0;
+        for (size_t ii = index_i; ii < (index_i + ni); ++ii) {
+            for (size_t jj = index_j; jj < (index_j + nj); ++jj) {
+                V_charge_p[ii][jj] = V_charge_p[jj][ii] = buffer[index++];
+            }
+        }
+    } // p
 
     V->add(V_charge);
     V_charge.reset();
-    pot.reset();
+    for (size_t t = 0; t < nthreads; ++t) {
+        pot[t].reset();
+    }
 
     // Diffuse Bases
     for (size_t ind = 0; ind < bases_.size(); ind++) {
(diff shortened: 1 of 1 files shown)


susilehtola:
The matrix is always going to be symmetric, since the basis functions are real.

TiborGY:
If I am interpreting the data correctly, the new code has worse parallel scaling at high core counts if there are no external charges. (5.36 vs 4.24)
And in general, the parallel scalings of the new code do not seem to follow Amdahl's law, note the drop in performance going from 8-->16 cores, then the sudden jump between 16 and 32. Even with charges, the running on 16 cores only yields a 2.9x speedup, which is better than the old code, but much worse than using 8 cores.
My first guess would be either some sort of NUMA related issue, or false sharing.

psi4/src/psi4/libmints/extern.cc
```diff
@@ -97,28 +97,73 @@ SharedMatrix ExternalPotential::computePotentialMatrix(std::shared_ptr<BasisSet>
...
+    // Monopole potential is symmetric, so generate unique pairs of shells
+    std::vector<std::pair<size_t, size_t> > ij_pairs;
+    for (size_t i = 0; i < basis->nshell(); ++i) {
+        for (size_t j = 0; j <= i; ++j) {
+            ij_pairs.push_back(std::pair<size_t, size_t>(i, j));
+        }
+    }
+
+    // Calculate monopole potential
+#pragma omp parallel for schedule(dynamic) num_threads(nthreads)
```
susilehtola, line 137: I think Psi4 prefers `guided`

chapincavender:
Thanks for your comments. I've changed `schedule(dynamic)` to `schedule(guided)` in the OpenMP loop.
I think @TiborGY is right that there is false sharing. I changed the code to create a copy of the output matrix `V_charge` for each thread and then sum them at the end. New timings are:

Cores | New code, no charges | New code, with charges
---- | ---- | ----
1 | 1.00 | 1.00
2 | 1.67 | 1.29
4 | 2.72 | 2.18
8 | 3.92 | 3.57
16 | 4.87 | 5.04
32 | 5.80 | 7.21

psi4/src/psi4/libmints/extern.cc
```diff
@@ -97,28 +97,74 @@ SharedMatrix ExternalPotential::computePotentialMatrix(std::shared_ptr<BasisSet>
...
+    // Monopole potential is symmetric, so generate unique pairs of shells
+    std::vector<std::pair<size_t, size_t> > ij_pairs;
+    for (size_t i = 0; i < basis->nshell(); ++i) {
+        for (size_t j = 0; j <= i; ++j) {
+            ij_pairs.push_back(std::pair<size_t, size_t>(i, j));
+        }
+    }
+
+    // Calculate monopole potential
+#pragma omp parallel for schedule(guided) num_threads(nthreads)
```
hokru, line 137: The size of each ij-pair is unknown to the scheduler so it's hardly possible to determine a good chunk-size.
Not sure how they are ordered but I'd assume they increase in angular momentum and thus the later loops become more expensive than the earlier.
Probably worth experimenting which scheduler is best.  `dynamic,1` has overhead in creating threads but there would be no waiting on expensive loops.

hokru: I know you changed the scheduler with the new table, but you also made other changes.

susilehtola:
I'm not sure I understand how the numbers are different for the case of no charges. If this routine is only computing the potential from external point charges (i.e. not the nuclei), then the routine shouldn't be run at all if there are no external point charges!

chapincavender:
I agree that with no charges the function that I changed should not be called. I am running the benchmark on a shared cluster, so the differences in the speedup with no charges reflect the uncertainty in the benchmark. Although the pairs of columns with and without charges are run on the same node in the same job.

I ran the benchmark again using `schedule(dynamic)` to directly compare to `schedule(guided)` copied from my last comment.

Cores | Guided, no charges | Guided, with charges | Dynamic, no charges | Dynamic, with charges
---- | ---- | ---- | ---- | ----
1 | 1.00 | 1.00 | 1.00 | 1.00
2 | 1.67 | 1.29 | 1.60 | 1.26
4 | 2.72 | 2.18 | 2.66 | 2.07
8 | 3.92 | 3.57 | 3.74 | 3.40
16 | 4.87 | 5.04 | 4.68 | 5.06
32 | 5.80 | 7.21 | 5.11 | 7.16

It looks like `dynamic` is slightly lower speedup than `guided`, but the two results are probably indistinguishable given the precision of my benchmark. This might also depend on the system and basis set used. The heaviest atom in my system is an oxygen with a jun-cc-pVDZ basis.

andysim (approved):
Ok, with Libint2 changes stable we can pull this one in - thanks so much for taking the time to code this up and for you patience as we overhauled the rest of the integrals.  There may be further changes we apply as we get the one electron integrals from Libint2, but this one looks good :)

jturney (approved):

loriab:
I have retested this atop a Libint2-based branch, and all's well.

cores |      old, no charges |  old, with charges | PR, no charges | PR, with charges
---- | ---- | ---- | ---- | ----
1   |    344    |     600    |         351   |      479
6   |    117    |     386     |        114  |       160

```
memory 20 gb

molecule {
0 1
N     0.9350155   -0.0279801   -0.3788916
C     1.6739638   -0.0357766    0.7424316
C     3.0747955   -0.0094480    0.5994562
C     3.5646109    0.0195446   -0.7059872
N     2.8531510    0.0258031   -1.8409596
C     1.5490760    0.0012569   -1.5808009
N     4.0885824   -0.0054429    1.5289786
C     5.1829921    0.0253971    0.7872176
N     4.9294871    0.0412404   -0.5567274
N     1.0716177   -0.0765366    1.9391390
H     0.8794435    0.0050260   -2.4315709
H     6.1882591    0.0375542    1.1738824
H     5.6035368    0.0648755   -1.3036811
H     0.0586915   -0.0423765    2.0039181
H     1.6443796   -0.0347395    2.7619159
--
0 1
N    -3.9211729   -0.0009646   -1.5163659
C    -4.6136833    0.0169051   -0.3336520
C    -3.9917387    0.0219348    0.8663338
C    -2.5361367    0.0074651    0.8766724
N    -1.9256484   -0.0110593   -0.3638948
C    -2.5395897   -0.0149474   -1.5962357
C    -4.7106131    0.0413373    2.1738637
O    -1.8674730    0.0112093    1.9120833
O    -1.9416783   -0.0291878   -2.6573783
H    -4.4017172   -0.0036078   -2.4004924
H    -0.8838255   -0.0216168   -0.3784269
H    -5.6909220    0.0269347   -0.4227183
H    -4.4439282   -0.8302573    2.7695655
H    -4.4267056    0.9186178    2.7530256
H    -5.7883971    0.0505530    2.0247280
units angstrom
}

Chrgfield = QMMM()

for i in range(25):
    for j in range(25):
        Chrgfield.extern.addCharge(-0.834, i * 3.0 + 1.649232019048, 0.0, j * 3.0 + -2.356023604706)
        Chrgfield.extern.addCharge( 0.417, i * 3.0 + 0.544757019107, 0.0, j * 3.0 + -3.799961446760)
        Chrgfield.extern.addCharge( 0.417, i * 3.0 + 0.544757019107, 0.0, j * 3.0 + -0.912085762652)

psi4.set_global_option_python('EXTERN', Chrgfield.extern)

energy("sapt0/jun-cc-pvdz")
```

loriab (approved):
lgtm! thank you for the speedup and testing.

Issue #1830, opened 2020-03-04
seg faults in usapt0 solver 


PhillCli:
Hi,
for odd cases where there is no electron in either alpha or beta block, sapt0 solver for UHF will result in segfault due to ```zero``` dimension passed to some of the procedures, e.g. ```C_DGEMM``` calls.

minimal working example
```Python
import psi4

psi4.set_memory("1024 MB")
geo_string = """
0 2
Li 0.000 0.000 -1.000
--
0 2
H  0.000 0.000 +1.000

no_reorient
no_com
units angstrom
symmetry c1
""".strip()

BASIS = "aug-cc-pvdz"
REFERENCE = "uhf"
BASIS_GUESS = "3-21G"
GUESS = "sad"
dimer = psi4.geometry(geo_string)

psi4.set_options(
    {
        "basis": BASIS,
        "reference": REFERENCE,
        "basis_guess": BASIS_GUESS,
        "guess": GUESS,
        "scf_type": "mem_df",
        "e_convergence": 1e-12,
        "d_convergence": 1e-12,
    }
)

psi4.energy("sapt0", return_wfn=True, molecule=dimer)
```

result:
```
    Role: ORBITAL
    Keyword: BASIS
    atoms 1 entry LI         line    35 file /home/filip/software/psi4-1.4rc/share/psi4/basis/3-21g.gbs 
    atoms 2 entry H          line    21 file /home/filip/software/psi4-1.4rc/share/psi4/basis/3-21g.gbs 


         ---------------------------------------------------------
                           SCF Castup computation                  
         ---------------------------------------------------------
                                   SCF
               by Justin Turney, Rob Parrish, Andy Simmonett
                          and Daniel G. A. Smith
                              UHF Reference
                        1 Threads,    976 MiB Core
         ---------------------------------------------------------

  ==> Geometry <==

    Molecular point group: c1
    Full point group: C_inf_v

    Geometry (in Angstrom), charge = 0, multiplicity = 2:

       Center              X                  Y                   Z               Mass       
    ------------   -----------------  -----------------  -----------------  -----------------
      Gh(LI)          0.000000000000     0.000000000000    -1.000000000000     7.016003436600
         H            0.000000000000     0.000000000000     1.000000000000     1.007825032230

  Running in c1 symmetry.

  Rotational constants: A = ************  B =      2.10095  C =      2.10095 [cm^-1]
  Rotational constants: A = ************  B =  62984.77227  C =  62984.77227 [MHz]
[...]

PR #1948, opened 2020-07-09
Usapt spin fix

jgonthier:
## Description
This PR (finally) fixes issue #1830 with USAPT0 by introducing a bunch of conditionals. I also uncovered and corrected a bug in the Exch-Disp term in USAPT0 while working on this.

## Questions
- [ ] Does anybody have acces to another UHF-based SAPT0 code that we could test against ?

## Checklist
- [x] Tests need to be run
- [ ] There should probably be some kind of announcement about the USAPT0 Exch-Disp term

## Status
- [x ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libsapt_solver/usapt0.cc
+++ b/psi4/src/psi4/libsapt_solver/usapt0.cc
@@ -1602,18 +1602,24 @@ void CPKS_USAPT0::compute_cpks() {
             double** rp = ra_A->pointer();
             double** pp = pa_A->pointer();
             double** sp = sa_A->pointer();
-            C_DAXPY(no * nv, alpha, pp[0], 1, xp[0], 1);
-            C_DAXPY(no * nv, -alpha, sp[0], 1, rp[0], 1);
-            r2A = C_DDOT(no * nv, rp[0], 1, rp[0], 1);
+            if (no > 0 && nv > 0) {
+                C_DAXPY(no * nv, alpha, pp[0], 1, xp[0], 1);
+                C_DAXPY(no * nv, -alpha, sp[0], 1, rp[0], 1);
+                r2A = C_DDOT(no * nv, rp[0], 1, rp[0], 1);
+            } else {
+                r2A = 0.0;
+            }
             no = xb_A_->nrow();
             nv = xb_A_->ncol();
             xp = xb_A_->pointer();
             rp = rb_A->pointer();
             pp = pb_A->pointer();
             sp = sb_A->pointer();
-            C_DAXPY(no * nv, alpha, pp[0], 1, xp[0], 1);
-            C_DAXPY(no * nv, -alpha, sp[0], 1, rp[0], 1);
-            r2A += C_DDOT(no * nv, rp[0], 1, rp[0], 1);
+            if (no > 0 && nv > 0) {
+                C_DAXPY(no * nv, alpha, pp[0], 1, xp[0], 1);
+                C_DAXPY(no * nv, -alpha, sp[0], 1, rp[0], 1);
+                r2A += C_DDOT(no * nv, rp[0], 1, rp[0], 1);
+            }
             r2A = sqrt(r2A) / b2A;
         }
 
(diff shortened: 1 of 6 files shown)


JonathonMisiewicz:
Should the code in this fork be correct now? I've published some USAPT0 results, so I'd like to test how severe the bug in the old code was (at least for my results).

jgonthier:
Yes, the code in this PR is corrected. In my test, which was Li interacting with H, the Exch-Disp term was wrong by about 25%. I haven't experimented with other systems but I'd be curious to know.

JonathonMisiewicz:
Having rerun my systems, I'm getting the same answers as before. Here's a sample input file.
```
molecule {
   0 3
 S         -0.3387469593        0.0000000000       -1.1908558477
 O          0.8847472382        0.0000000000       -0.3445279161
   --
   0 1
 H          0.4210342108        0.0000000000        1.7275935987
 O         -0.2480570158        0.0000000000        2.4179317954
 H          0.2472257748        0.0000000000        3.2385953098
     symmetry c1
     no_reorient
     no_com
}

set {
reference uhf
scf_type     direct
guess        sad
basis         jun-cc-pV(D+d)Z
df_basis_sapt jun-cc-pV(D+d)Z-ri
}

energy('sapt0')
```

The output file confirms that I'm running on `2d342fe`, and the SAPT output file is displaying `Open-shell: Jérôme Gonthier`, so this should be running your new code... but it's getting the same answers. Is there a reason for this? Perhaps this bug will only appear for two open-shell systems, but not for one open-shell and one closed-shell?

PhillCli:
I think I could get the access to Michał Hapka's (JCP 2012) dev molpro code with SAPT0(UKS), which can be force to run CPUHF, but it would be a comparison with code without density fitting.

I had these results laying around, with larger than usual fitting basis set.

```python
import time
import psi4

geometry = """
 0 1
O 0.0000   0.00000   3.175060
H 0.0000   0.75750   2.587994
H 0.0000  -0.75750   2.587994
 --
 0 2
O 0.0000   0.00000   0.000000
O 0.0000   1.32501   0.000000
H 0.0000  -0.22470  -0.973420

units angstrom
no_reorient
no_com
symmetry c1
"""

BASIS = "aug-cc-pvdz"
MEMORY = "20000 MB"
THREADS = 4

psi4.set_memory(MEMORY)
psi4.core.set_num_threads(THREADS)
psi4.set_options(
    {
        "reference": "uhf",
        "scf_type": "mem_df",
        "stability_analysis": "follow",
        "d_convergence": 1e-10,
        "e_convergence": 1e-10,
        "guess": "sad",
        "basis": BASIS,
        "df_basis_sapt": "aug-cc-pv5z-ri",
        "df_basis_scf": "aug-cc-pv5z-jkfit",
    }
)

def compute_s2():
    time_start = time.time()
    dimer = psi4.geometry(geometry)
    psi4.energy("sapt0", molecule=dimer)

if __name__ == "__main__":
    compute_s2()
```

run on top of `2d342fe`

```shell
    SAPT Results
  --------------------------------------------------------------------------------------------------------
    Electrostatics                 -3.76447859 [mEh]      -2.36224598 [kcal/mol]      -9.88363717 [kJ/mol]
      Elst10,r                     -3.76447859 [mEh]      -2.36224598 [kcal/mol]      -9.88363717 [kJ/mol]

    Exchange                        2.19095680 [mEh]       1.37484615 [kcal/mol]       5.75235628 [kJ/mol]
      Exch10                        2.19095680 [mEh]       1.37484615 [kcal/mol]       5.75235628 [kJ/mol]
      Exch10(S^2)                   2.18871598 [mEh]       1.37344002 [kcal/mol]       5.74647303 [kJ/mol]

    Induction                      -0.46899705 [mEh]      -0.29430009 [kcal/mol]      -1.23135159 [kJ/mol]
      Ind20,r                      -0.81555155 [mEh]      -0.51176632 [kcal/mol]      -2.14123029 [kJ/mol]
      Exch-Ind20,r                  0.51152243 [mEh]       0.32098517 [kcal/mol]       1.34300196 [kJ/mol]
      delta HF,r (2)               -0.16496793 [mEh]      -0.10351894 [kcal/mol]      -0.43312325 [kJ/mol]

    Dispersion                     -1.62383455 [mEh]      -1.01897156 [kcal/mol]      -4.26337702 [kJ/mol]
      Disp20                       -1.79467532 [mEh]      -1.12617576 [kcal/mol]      -4.71191939 [kJ/mol]
      Exch-Disp20                   0.17084077 [mEh]       0.10720420 [kcal/mol]       0.44854237 [kJ/mol]

  Total HF                         -2.04251884 [mEh]      -1.28169992 [kcal/mol]      -5.36263248 [kJ/mol]
  Total SAPT0                      -3.66635339 [mEh]      -2.30067149 [kcal/mol]      -9.62600950 [kJ/mol]
```

`JCP 2012 doi: 10.1063/1.4758455` molpro-dev

```shell
Elst10,r     -3.76435858088087
Exch10        2.19085113274701
Ind20,r      -0.815540698087809
Exch-Ind20,r  0.511512600752067
Disp20       -1.79467641180960
Exch-Disp20   0.170852132052007
```

Can probably get the results for other modest systems, with relatively small basis sets.
Li is problematic as it has only universal ```JK``` and it is not good to compare with non-density fitted results.

jgonthier:
@JonathonMisiewicz Indeed, I also ran your input and the energies are the same. The issue I found was that the `Ka_O` and `Kb_O` matrices were saved *after* computation of the coupled induction term, and since coupled induction uses the same `jk` object, it changed these matrices. I switched the code to save a copy of the `Ka_O` and `Kb_O` matrices before computing coupled induction and it solved my issue. However, looking at your case, it appears these matrices are not overwritten even by the old code. Honestly, I don't understand why right now. I may look at that more over the weekend.

@PhillCli Thank you so much for this data! The agreement seems to be pretty good. I guess we only need more data if the core developers think it's necessary before merging.

loriab:
@jgonthier, @PhillCli, I've added a test from the original issue. It's not independently confirmed results, but the above discussion reasonably covers that. Provided you two are satisfied and I can scrounge up another review, I'll merge.

PhillCli:
@loriab I'm fine with the test case from the original issue. Thanks for putting it there!

jgonthier:
Thanks everyone and thanks @loriab for adding the test! We could also add a test based on the Molpro results above but that would imply using high angular momentum functions so I guess it's not really essential.

PR #1950, opened 2020-07-09
Print pairs

sjrl:
## Description
I would like access to the CCSD pair energies from the FNOCC code in Psi4. I made a small modification to the FNOCC code so the pair energies are calculated (using an existing for loop) and I've added the new variable "CCSD PAIR ENERGIES" to store the quantity. 

This is my first PR with Psi4 so I apologize in advance if I've missed any coding guidelines. 

## Questions
- [ ] Should I add a test for this new variable?

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


loriab:
@sjrl see if the test I added is satisfactory. If so, I'll merge.

PR #1951, opened 2020-07-10
Add point group information to FCIDUMP interface (2)

fevangelista:
## Description
This PR perfects the `fcidump` interface by adding a field that stores the point group symmetry (`PNTGRP`, just to keep it in the FORTRAN spirit of "Thou shall use no unnecessary vowels in variable names"). For example:
```
ORBSYM=1,1,1,1,4,4,4,3,3,3,2,2,2,
ISYM=1,
PNTGRP=D2H,
```
Why is this change necessary? In essence because with the current information it is impossible to restart a computation that uses symmetry. Without the point group information, one cannot determine the psi4 irrep labels of orbitals from the standard information contained in the FCIDUMP file (see issue #1431) because when the file is written the irreps are resorted in a way that depends on the point group.

One potential issue: codes that consume FCIDUMP files produced by psi4 could encounter problems parsing the line that starts with `PNTGRP`. So I made this extra line optional via an argument passed to the `fcidump()` python function

## Todos
- [x] Add point group information to fcidump files (enabled via bool argument)
- [x] Fix a bug in the function that reads fcidump files when no orbital energies are present.

## Checklist
- [x] Updated `fcidump` test

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/p4util/fcidump.py
+++ b/psi4/driver/p4util/fcidump.py
@@ -36,23 +36,29 @@ from psi4.driver.p4util.testing import compare_integers, compare_values, compare
 from psi4.driver.procrouting.proc_util import check_iwl_file_from_scf_type
 
 from psi4 import core
-from .exceptions import ValidationError, TestComparisonError
 
 
-def fcidump(wfn, fname='INTDUMP', oe_ints=None):
+def fcidump(wfn, fname='INTDUMP', oe_ints=None, write_pntgrp=False):
     """Save integrals to file in FCIDUMP format as defined in Comp. Phys. Commun. 54 75 (1989)
     Additional one-electron integrals, including orbital energies, can also be saved.
     This latter format can be used with the HANDE QMC code but is not standard.
+    This function converts the irrep labels from Cotton's order (used in psi4)
+    to molpro ordering. Since the latter is ambiguous unless the point group is
+    specificied, this function has an option to write the point group label in the
+    FCIDUMP file. This, however, is not part of the standard.
 
     :returns: None
 
-    :raises: ValidationError when SCF wavefunction is not RHF
+    :raises: ValidationError when SCF wavefunction is not RHF, ROHF, or UHF
 
     :type wfn: :py:class:`~psi4.core.Wavefunction`
     :param wfn: set of molecule, basis, orbitals from which to generate cube files
     :param fname: name of the integrals file, defaults to INTDUMP
     :param oe_ints: list of additional one-electron integrals to save to file.
     So far only EIGENVALUES is a valid option.
+    :param write_pntgrp: write the point group to file. This information is not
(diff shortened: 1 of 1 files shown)


psi4/driver/p4util/fcidump.py
```diff
@@ -36,23 +36,29 @@
...
 
 
-def fcidump(wfn, fname='INTDUMP', oe_ints=None):
+def fcidump(wfn, fname='INTDUMP', oe_ints=None, write_pntgrp=False):
     """Save integrals to file in FCIDUMP format as defined in Comp. Phys. Commun. 54 75 (1989)
     Additional one-electron integrals, including orbital energies, can also be saved.
     This latter format can be used with the HANDE QMC code but is not standard.
+    This function converts the irrep labels from Cotton's order (used in psi4)
+    to molpro ordering. Since the latter is ambiguous unless the point group is
+    specificied, this function has an option to write the point group label in the
```
robertodr, line 47: Small typo
```suggestion
    specified, this function has an option to write the point group label in the
```

psi4/driver/p4util/fcidump.py
```diff
@@ -36,23 +36,29 @@
...
+    FCIDUMP file. This, however, is not part of the standard.
 
     :returns: None
 
-    :raises: ValidationError when SCF wavefunction is not RHF
+    :raises: ValidationError when SCF wavefunction is not RHF, ROHF, or UHF
```
robertodr, line 52: I am not sure the original format specification allows ROHF.

psi4/driver/p4util/fcidump.py
```diff
@@ -155,8 +164,11 @@ def fcidump(wfn, fname='INTDUMP', oe_ints=None):
             # Orbital energies
             core.print_out('Writing orbital energies in FCIDUMP format to ' + fname + '\n')
             if 'EIGENVALUES' in oe_ints:
-                eigs_dump = write_eigenvalues(wfn.epsilon_a().get_block(mo_slice).to_array(), mo_idx)
-                intdump.write(eigs_dump)
+                if reference == 'RHF'
```
JonathonMisiewicz, line 167:
```suggestion
                if reference == 'RHF':
```

psi4/driver/p4util/fcidump.py
fevangelista: I am not sure either. So in this case I avoid saving the eigenvalues. But the 1-/2-e integrals are still well defined. It might be difficult to reconstruct a ROHF wave function without specifying the orbital energies, but still it can be used to exchange integrals. The more I use FCIDUMP the more it is clear that this format is not designed to be general enough.

fevangelista:
@JonathonMisiewicz : thanks for catching the issue. I tested the `fcidump` test and it runs fine now.

fevangelista:
@dgasmith: ok, let me work on a test case

PR #1956, opened 2020-07-15
PSIMRCC Finishing Details

### LINKED ISSUE #1877: Inconsistent `psimrcc` Warnings
For historical reasons, Mk-MRCC was only implemented in Psi4 assuming small active spaces, and `psimrcc` rightly gives you [two](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/psimrcc/main.cc#L114-L124) [warnings](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/psimrcc/main.cc#L130-L143) that the equations it is using are not the exact Mk-MRCC equations if your active space is any larger.

If you use Mk-MRPT2, the first warning fires, but the second does not. Further, the error message is specific to Mk-MRCC and says nothing about Mk-MRPT2. Is the Mk-MRPT2 procedure implemented in `psimrcc` exact or approximate? If it's exact, we need to stop the first warning from firing for Mk-MRPT2. If it's approximate, we need to make sure the second warning does fire for Mk-MRPT2. I checked the original paper, but couldn't find a clear statement either way, and I'd rather not learn Mk-MRCC theory just to fix this error message.

I'm going to get back to giving `psimrcc` a wavefunction, so I'll fix this as part of that PR. I need to touch this part of the logic anyways.

@fevangelista? @andysim?

JonathonMisiewicz:
## Description
This PR covers some miscellanea needed to fully finish the `psimrcc` series, which will close #1877. Ready for review, but not yet merge. I need @fevangelista to answer the questions, and then to run clang format. And then I can assess where I'm at for  the libtrans refactor preliminary to the gradient refactor. It'll end one of these days...

## Todos
- [x] Removes some dead code
- [x] `free_memory_` is now actually used
- [x] Clang formatting

## Questions
- [x] special_matrices.cc::MatrixBase and derived subclasses currently do not update the free memory. While I can make it update memory, that would require giving each Matrix a wavefunction. Should I add memory costs for this?
- [x] I assumed that the memory cost of the index.cc classes was negligible, as these are three-index or smaller. Should I add memory costs for this?1

## Checklist
- [x] psirmrcc tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/psimrcc/mrcc_t2_amps.cc
```diff
@@ -42,6 +42,10 @@
 namespace psi {
 namespace psimrcc {
 
+// Many functions in this file have six-index local variables.
```
loriab, line 45: wow

fevangelista: I think these functions are never called. It must have been an attempt to implement Mk-MRCCSDT.

fevangelista:
@JonathonMisiewicz : I would not add matrices and indices to free_memory_.

JonathonMisiewicz:
Great, this can come in as soon as the tests pass. (And if clang format breaks anything, we should be concerned.)

PR #1958, opened 2020-07-15
testing label for c1-incapable tests

loriab:
## Description
BrianQC in #1893 only runs in c1 point group, so they'd like to have a sub test suite that can work in c1. This provides a label `noc1` for tests that _need_ symmetry (occasionally for convergence but mostly b/c keywords include DOCC or STATES_PER_IRREP). It was determined by forcing c1 by adding `mol->reset_point_group("c1");` here https://github.com/psi4/psi4/blob/master/psi4/src/export_mints.cc#L235 . That line *is not* in this PR, but the BrianQC ppl may want to consider it when their engine is enabled. With that line in place, `ctest -LE noc1 -j<N>` runs cleanly. It can also be combined, so `ctest -L quick -LE noc1` and `ctest -R scf -LE noc1` also valid.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


tests/dft-grad-lr2/input.dat
```diff
@@ -25,7 +25,7 @@ set {
     dft_radial_points     120
     dft_spherical_points  434
     e_convergence         12
-    d_convergence         12
+    d_convergence         10
```
hokru, line 28: if the test is turned off for C1, what motivated the change?

loriab: good point. I fixed it with the 10 (last single-point of last gradient was failing to converge w/i iterations), then suppressed it anyways. Unsuppressing soon.

PR #1961, opened 2020-07-19
Added MBIS Charge Partitioning Scheme

andyj10224:
## Authors
Andy Jiang, Zachary L. Glick (Sherrill Group, Georgia Institute of Technology)

## Description
Adds the MBIS Charge Partitioning Feature to Psi4, from the associated paper by Toon Verstraelen et al. (2016)
(https://pubs.acs.org/doi/10.1021/acs.jctc.6b00456)

## Todos
- [x] Calculation of MBIS Atomic Charges
- [x] Calculation of MBIS Atomic Multipoles

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libfock/points.h
```diff
@@ -245,7 +245,7 @@ class RKSFunctions : public PointFunctions {
     size_t block_index() { return block_index_; }
 };
 
-class UKSFunctions : public PointFunctions {
+class PSI_API UKSFunctions : public PointFunctions {
```
loriab, line 248: Were these psi_api leftover from the plugin implementation?

tests/mbis/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(props1 "psi;properties")
```
loriab, line 3: Props1-> mbis

loriab, line 3: Props1-> mbis

psi4/src/read_options.cc
```diff
@@ -247,6 +247,13 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     /*- How many NOONS to print -- used in libscf_solver/uhf.cc and libmints/oeprop.cc -*/
     options.add_str("PRINT_NOONS", "3");
 
+    ///MBIS Options (libmints/oeprop.cc)
+
+    /*- How many MBIS Iterations to run -*/
+    options.add_int("MBIS_MAX_ITER", 200);
```
loriab, line 253: One word maxiter

loriab, line 253: One word maxiter

psi4/src/psi4/libfock/points.h
andyj10224: Yes, they were. I will fix it

tests/mbis/CMakeLists.txt
andyj10224: Fixed in commit 17

psi4/src/read_options.cc
andyj10224: Fixed

dgasmith:
Can you link the paper?

andyj10224:
> Can you link the paper?

Yes, https://pubs.acs.org/doi/10.1021/acs.jctc.6b00456

zachglick:
It would be good to add that paper to the manual bibliography (psi4/doc/sphinxman/source/bibliography.rst) and cite it within the mbis documentation.

andyj10224:
> It would be good to add that paper to the manual bibliography (psi4/doc/sphinxman/source/bibliography.rst) and cite it within the mbis documentation.

I added the citations

doc/sphinxman/source/oeprop.rst
```diff
@@ -166,3 +168,18 @@ grid point, in the order that the grid was specified, while the *Exvals*,
...
+Minimal Basis Iterative Stockholder
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Minimal Basis Iterative Stockholder (MBIS) partitioning is one of many procedures
+that produce atomic one-electron densities from the molecular one-electron density.
+Calling MBIS in Psi4 will calculate atomic charges, as well as dipoles, quadrupoles, and
+octupoles. The allowed number of iterations and convergence criteria for the stockholder 
+algorithm is controlled by |globals__mbis_maxiter| and |globals__mbis_convergence|. Note 
+that the density is partitioned on a molecular grid, so any settings related to the number
+or type of grid points applies. (Associated Paper: T. Verstraelen et al. "Minimal Basis Iterative Stockholder: Atoms in Molecules for Force-Field Development". *J. Chem. Theory and Comput.* doi: 10.1021/acs.jctc.6b00456.)
```
zachglick, line 185: Instead of writing out the paper information, you can directly link to the citation that you added to the bibliography. The syntax looks something like [Verstraelen:2016]_. You can look at other files in doc/sphinxman/source/ to get an idea

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1788,445 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+PopulationAnalysisCalc::compute_mbis_multipoles(bool print_output) {
+
+    if (print_output) outfile->Printf("  ==> Computing MBIS Charges <==\n\n");
+    timer_on("MBIS");
+    
+    // => Setup 1RDM on DFTGrid <= //
+
+    Options& options = Process::environment.options;
+    const int MAX_ITER = options.get_int("MBIS_MAXITER");
+    const double CONV = options.get_double("MBIS_CONVERGENCE");
```
loriab, line 1849: all caps vars tend to be reserved for file level constants.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1788,445 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+
+        outfile->Printf("\n  MBIS Dipoles: (a.u.)\n");
+        outfile->Printf("   Center  Symbol  Z        X           Y           Z\n");
+
+        for (int a = 0; a < num_atoms; a++) {
+            outfile->Printf("  %5d      %2s %4d   %9.6f   %9.6f   %9.6f\n", a+1, mol->label(a).c_str(), 
+                    (int)mol->Z(a), dpole_flat->get(a, 0), dpole_flat->get(a, 1), dpole_flat->get(a, 2));
+        }
+
+        outfile->Printf("\n  MBIS Quadrupoles: (a.u.)\n");
```
loriab, line 2204: good to label units specifically: `[e a0]` for Dipole, `[e a0^2]` for Quad, etc.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1788,445 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+
+    // Cartesian atomic multipoles (Horton convention)
+    std::vector<double> monopoles(num_atoms, 0.0);
+    std::vector<std::vector<double>> dipoles(num_atoms, std::vector<double>(3, 0.0));
+    std::vector<std::vector<std::vector<double>>> quadrupoles(num_atoms, std::vector<std::vector<double>>(3, std::vector<double>(3, 0.0)));
+    std::vector<std::vector<std::vector<std::vector<double>>>> octupoles(num_atoms, std::vector<std::vector<std::vector<double>>>(3, std::vector<std::vector<double>>(3, std::vector<double>(3, 0.0))));
+
+    // Cartesian atomic multipoles (convention from Anthony Stone's "The Theory of Intermolecular Forces, Second Edition")
+    std::vector<std::vector<std::vector<double>>> stone_quadrupoles(num_atoms, std::vector<std::vector<double>>(3, std::vector<double>(3, 0.0)));
+    std::vector<std::vector<std::vector<std::vector<double>>>> stone_octupoles(num_atoms, std::vector<std::vector<std::vector<double>>>(3, std::vector<std::vector<double>>(3, std::vector<double>(3, 0.0))));
```
loriab, line 2093: we should probably chat about flat/shaped, unique/redundant arrays. Psi and qcsk are committed to returning full, sensibly-shaped arrays to the user upon `variable("* OCTUPOLE")`, etc. that can be mindlessly indexed as `[0][1][1]`. But I'm sure you've met the restriction of `set_array_variable` only taking 2d arrays. The scheme I started for the molecular multipoles is to compute flat+unique here in oeprop, then let https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L638-L670 and other plumper/compressor machinery trigger human-friendly forms. This case is of course different b/c per-atom. 

So we should talk to make sure user gets redundant arrays and to see if both forms needed c-side.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1788,445 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+    int running_points = 0;
+
+    // Coordinates, weights, and rho (molecular electron density) at each grid point
+    std::vector<double> x_points(total_points, 0.0);
+    std::vector<double> y_points(total_points, 0.0);
+    std::vector<double> z_points(total_points, 0.0);
+    std::vector<double> weights(total_points, 0.0);
+    std::vector<double> rho(total_points, 0.0);
+
+    for (int b = 0; b < num_blocks; b++) {
```
loriab, line 1884: won't change fn a bit. but since `num_blocks` only used once, probably doesn't need to be separate var.

tests/mbis/input.dat
```diff
@@ -0,0 +1,56 @@
...
+octupoles_ref = psi4.Matrix.from_list([                                                                                                    #TEST
+       [ 0.00000000,  0.00000000, -0.15703089, 0.00000000, 0.00000000, 0.00000000,   0.00000000, -0.45651683,  0.00000000, -0.64988960],   #TEST
+       [ 0.00000000, -0.01049888,  0.00088680, 0.00000000, 0.00000000, 0.00000000,  -0.03241728, -0.00186054, -0.01442813, -0.00608710],   #TEST
+       [-0.00000000,  0.01049888,  0.00088680, 0.00000000, 0.00000000, 0.00000000,   0.03241728, -0.00186054,  0.01442813, -0.00608710]])  #TEST
+
+molecule mol {
+  0 1
+  O
+  H 1 1.0
+  H 1 1.0 2 104.5
```
loriab, line 27: since there's a lot of electron and basisfn counting, having another test with an anion or even a Na+ would be a good precaution.

psi4/src/read_options.cc
```diff
@@ -247,6 +247,13 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     /*- How many NOONS to print -- used in libscf_solver/uhf.cc and libmints/oeprop.cc -*/
     options.add_str("PRINT_NOONS", "3");
 
+    ///MBIS Options (libmints/oeprop.cc)
+
+    /*- How many MBIS Iterations to run -*/
+    options.add_int("MBIS_MAXITER", 200);
+    /*- MBIS Convergence Criteria -*/
+    options.add_double("MBIS_CONVERGENCE", 1.0e-8);
```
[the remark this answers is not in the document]
andyj10224: There is a sqrt involved

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1788,445 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+    // Next iteration populations and widths
+    std::vector<std::vector<double>> Nai_next(num_atoms, std::vector<double>(4, 0.0));
+    std::vector<std::vector<double>> Sai_next(num_atoms, std::vector<double>(4, 0.0));
+
+    // Population and width guesses
+    for (int atom = 0; atom < num_atoms; atom++) {
+        int atomic_num = (int) mol->Z(atom);
+
+        for (int m = 0; m < mA[atom]; m++) {
+            Nai[atom][m] = (double) PopulationAnalysisCalc::get_nai(atomic_num, m+1);
```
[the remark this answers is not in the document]
andyj10224: get_nai sets the initial guess for the atomic population in each shell, while mA gives the number of shells in an atom

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1788,445 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+                    sum_n += weights[point] * rho[point] * rho_ai_0_point / rho_0_points[point];
+                    sum_s += weights[point] * distances[atom][point] * rho[point] * rho_ai_0_point / rho_0_points[point];
+                }
+
+                Nai_next[atom][m] = sum_n;
+                Sai_next[atom][m] = sum_s / (3 * Nai_next[atom][m]);
+            }
+        }
+
+        // Convergence check (Equation 20 in Verstraelen et al.)
```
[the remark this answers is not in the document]
andyj10224: I already defined the Verstraelen paper in the .h file

tests/mbis/PhOH.dat
```diff
@@ -0,0 +1,110 @@
+#! MBIS calculation on H2O
+
+charges_ref = psi4.Matrix.from_list([
```
loriab, line 3: PhOH.dat and .out is a separate test, yes? Tests need to be one per directory and no variation in file names. So probably your original test directory becoming mbis1 and then a separate directory for the this phenol as mbis2 is a good plan. (Remember to change the `#!` line at the top -- this gets slurped up into the docs.)

psi4/src/read_options.cc
```diff
@@ -247,6 +247,18 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+
+    /*- Maximum Number of MBIS Iterations -*/
+    options.add_int("MBIS_MAXITER", 500);
+    /*- MBIS Convergence Criteria -*/
+    options.add_double("MBIS_D_CONVERGENCE", 1.0e-8);
+    /*- MBIS Number of Radial Points -*/
+    /*- Additional Radial Points may be needed for Heavier Atoms (200-300) like Zinc -*/
+    options.add_int("MBIS_RADIAL_POINTS", 99);
+    /*- MBIS Number of Spherical Points -*/
+    options.add_int("MBIS_SPHERICAL_POINTS", 350);
```
loriab, line 260: wasn't the plan to use normal dft grid by default (75, 302), then warn for nonphysical integration so user could retry with tighter grid?

tests/mbis-2/input.dat
```diff
@@ -0,0 +1,52 @@
...
+}
+
+e, wfn = energy('hf/cc-pvdz', return_wfn=True)
+oeprop(wfn, 'MBIS_CHARGES', title='OH- SCF')
+
+#NOTE: wfn.array_variable gives you flattened atomic multipole arrays; for expanded arrays, use wfn.variable
+charges = wfn.array_variable('MBIS_CHARGES')
+dipoles = wfn.array_variable('MBIS_DIPOLES')
+quadrupoles = wfn.array_variable('MBIS_QUADRUPOLES')
+octupoles = wfn.array_variable('MBIS_OCTUPOLES')
```
loriab, line 47: for one test case, you'll want to test your python_helpers changes by calling `wfn.variable('MBIS *POLE')`, etc. do it on a 2-atom test like this, so the ref arrays aren't so big. and you can use 0 in the ref arrays for all the noise number.

can use `compare_values(ref, vals, "MBIS *POLE", atol=1.e-7)` for np arrays

tests/mbis-4/input.dat
```diff
@@ -0,0 +1,52 @@
...
+  -1.61387749e-13, -6.44818499e-14,  4.70577367e-13,  2.25696361e+00,
+   2.06112555e-13,  4.56955103e+00]]) #TEST
+
+molecule mol {
+  0 1
+  Na 0.00 0.00 0.00
+  Cl 0.00 0.00 2.36
+  symmetry c1
+  no_reorient
+  no_com
```
loriab, line 31: need the c1, no_com, no_reorient?

psi4/driver/p4util/python_helpers.py
```diff
@@ -657,7 +657,23 @@ def _qcvar_reshape_get(key, val):
...
+        for atom in range(len(val)):
+            val_new.append(_multipole_plumper(val[atom], 2))
+        val = np.array(val_new)
+        return val
+    elif key.upper() == "MBIS_OCTUPOLES":
+        val = val.np.reshape(-1, 10)
+        val_new = []
+        for atom in range(len(val)):
+            val_new.append(_multipole_plumper(val[atom], 3))
+        val = np.array(val_new)
```
loriab, line 674: `val = np.array([_multipole_plumper(val[iat], 3) for iat in range(len(val))])`

list comprehensions are fun

loriab:
looks like gcc 5.2 may not like your initializer lists https://dev.azure.com/psi4/psi4/_build/results?buildId=2276&view=logs&j=96451287-da82-57a8-2c11-8da7db9ab71a&t=5d474b6c-8b79-5932-1eb0-049666067a4b&l=1075

tests/mbis-4/input.dat
zachglick: I think we want `no_reorient`, since the atomic dipoles and quadrupoles change with orientation

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,598 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+        mol_nuc_charge += mol->fZ(a);
+    }
+    double mol_electron_count = mol_nuc_charge - mol_charge;
+
+    double electron_diff = fabs(mol_electron_count - rho_num_electrons);
+
+    if (print_output) outfile->Printf("  ELECTRON DIFFERENCE: %8.5f\n\n", electron_diff);
+
+    if (electron_diff > 0.001) {
+        throw PsiException("Number of electrons calculated using the grid (" + std::to_string(round(100000.0 * rho_num_electrons) / 100000.0) + ") does not match the number of electrons in molecule. Try increasing the number of radial or spherical points (mbis_radial_points and mbis_spherical_points options).", __FILE__, __LINE__);
```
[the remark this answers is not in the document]
zachglick: My concern with a warning is that it would get buried in the output file. On the other hand, I can see how throwing an error might be too disruptive, particularly if the user is okay with kind of wrong charges.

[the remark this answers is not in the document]
andyj10224: A good idea would be to let the user set the electron_difference_tolerance as an option, default will be 0.001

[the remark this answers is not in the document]
andyj10224: We could set that option as an expert option, and give a warning/error if the tolerance is too loose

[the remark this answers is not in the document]
loriab: I'm coming around to thinking of this as like the UHF spin-contamination or t1 amplitudes indicator. The integration must get printed so users can judge the quality of their calc, but it doesn't error the job. (or maybe only errors if outrageous.) It'd be good to know how missing electrons correlate with bad charges, which Andy's current grid testing will help with, too.

psi4/src/read_options.cc
andyj10224: Based on tests I ran on the Benzene-Naphthalene Series, I discovered that 99/350 is indeed a better default than 75/302 or even 99/302, based on looking at the error growth

Electron Errors:

75/302:
Benzene: 2E-5
Naphthalene: 7E-5
Anthracene: 9E-5
Tetracene: 12E-5
Pentacene: 14E-5

99/302:
Benzene: 2E-5
Naphthalene: 7E-5
Anthracene: 10E-5
Tetracene: 12E-5
Pentacene: 14E-5

99/350:
Benzene: 1E-5
Naphthalene: 3E-5
Anthracene: 3E-5
Tetracene: 2E-5
Pentacene: 2E-5

loriab: smaller grid errors are larger but not troublesomely so -- less than an order of magnitude. do the charges change appreciably (>0.02) over the different grid sizes? (75, 302) looking good to me.

andyj10224: All atomic charges are within 0.001 electrons of each other on tetracene when compared on a 75/302 grid and a 99/350 grid

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,597 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+// See PopulationAnalysisCalc::compute_mbis_multipoles
+void OEProp::compute_mbis_multipoles() {
+
+    SharedMatrix mpole, dpole, qpole, opole;
+    std::tie(mpole, dpole, qpole, opole) = pac_.compute_mbis_multipoles(true);
+
+    wfn_->set_array_variable("MBIS CHARGES", mpole);
+    wfn_->set_array_variable("MBIS DIPOLES", dpole);
+    wfn_->set_array_variable("MBIS QUADRUPOLES", qpole);
+    wfn_->set_array_variable("MBIS OCTUPOLES", opole);
```
loriab, line 1804: what do other reviewers think about plural here? plain (molecular) multipoles are singular, but then these are atomic.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,597 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+
+    // Sets up MBIS Grid using MBIS_RAIDAL_POINTS and MBIS_SPHERICAL_POINTS options
+    std::map<std::string, int> mbis_grid_options_int;
+
+    mbis_grid_options_int["DFT_RADIAL_POINTS"] = options.get_int("MBIS_RADIAL_POINTS");
+    mbis_grid_options_int["DFT_SPHERICAL_POINTS"] = options.get_int("MBIS_SPHERICAL_POINTS");
+
+    std::map<std::string, std::string> mbis_grid_options_str;
+
+    mbis_grid_options_str["DFT_PRUNING_SCHEME"] = "NONE";
```
hokru, line 1973: Does is not work well with pruning? ("ROBUST" or "TREUTLER") Did you test it?

How about `DFT_WEIGHTS_TOLERANCE`?

andyj10224: @hokru You bring up a good point, it may be a good idea to create a set of grid parameters exclusively for the MBIS grid, as we need to keep the MBIS grid independent of the DFT grid in case we are running DFT, and then MBIS on the molecule. I have not done much testing with pruning schemes or the `DFT_WEIGHTS_TOLERANCE` option. My teammate Zach suggested that I set `DFT_PRUNING_SCHEME` to NONE. What kinds of tests do you recommend?

hokru: `DFT_WEIGHTS_TOLERANCE=-1` will turns off the screening of small weights, the current default (1e-15) is conservative, but for a test I'd try turning it off.

If it is sensible to seek computational efficiency for MBIS charges, then I'd try at least the `ROBUST` scheme.

andyj10224: @hokru I tested the `ROBUST` and `TREUTLER` schemes on a tetracene molecule (6 cores), there were no significant time savings, while the quality of the integration remained the same (same electron discrepancy of 12E-5). I also tested setting `DFT_WEIGHTS_TOLERANCE -1.0`, the time was slightly worse while the quality of the integration did not change. (The MBIS charges were all within 0.001 electrons for each atom in all of the tests)

Timings:

Pruning schemes:
Treutler: ~50 seconds
Robust: ~52 seconds
None: ~52 seconds

DFT_WEIGHTS_TOLERANCE -1.0: ~55-65 seconds

Do you suggest that I add an `MBIS_PRUNING_SCHEME` option to MBIS, or any other options to the MBIS grid that can be set on a DFT grid (like `DFT_NUCLEAR_SCHEME` and `DFT_BASIS_TOLERANCE` http://www.psicode.org/psi4manual/master/dft.html#grid-selection)

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,597 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+    if (print_output) outfile->Printf("  Electron Count from Grid (Expected Number): %8.5f (%8.5f)\n", rho_num_electrons, mol_electron_count);
+    if (print_output) outfile->Printf("  Difference: %8.5f\n\n", electron_diff);
+
+    if (fabs(electron_diff) > 0.001) {
+        throw PsiException("Number of electrons calculated using the grid (" + std::to_string(round(100000.0 * rho_num_electrons) / 100000.0) + ") does not match the number of electrons in molecule. Try increasing the number of radial or spherical points (mbis_radial_points and mbis_spherical_points options).", __FILE__, __LINE__);
+    }
+
+    // Distances and displacements between grid points and nuclei
+    std::vector<std::vector<double>> distances(num_atoms, std::vector<double>(total_points, 0.0));
+    std::vector<std::vector<std::vector<double>>> disps(num_atoms, std::vector<std::vector<double>>(total_points, std::vector<double>(3, 0.0)));
```
hokru, line 2065: Is this style of nesting `std:vector` here and elsewhere OK for a large `total_points` (e.g millions of grid points)?

zachglick: That's an interesting question. A nested `std::vector` does require more overhead than a flattened `std:vector`, but is a little more readable (to me at least). Since this routine shouldn't be rate-limiting, I'd lean towards leaving as is.

hokru: My intention was to double check if a possibly large nested `std::vector` was thought about. While this part might not be rate-limiting other parts could. I don't have principle objections if you think it's OK.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,597 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+    while (iter < max_iter) {
+
+        // Calculate proatom and promolecule density at all points
+#pragma omp parallel for
+        for (int point = 0; point < total_points; point++) {
+            rho_0_points[point] = 0.0;
+            for (int atom = 0; atom < num_atoms; atom++) {
+                rho_a_0_points[atom][point] = 0.0;
+                for (int m = 0; m < mA[atom]; m++) {
+                    rho_a_0_points[atom][point] += rho_ai_0(Nai[atom][m], Sai[atom][m], distances[atom][point]);
```
susilehtola, line 2144: The array indices are in the wrong order, so the memory access here is very inefficient.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,597 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+                Sai_next[atom][m] = sum_s / (3 * Nai_next[atom][m]);
+            }
+        }
+
+        // Convergence check (Equation 20 in Verstraelen et al.)
+        std::vector<double> delta_rho_atoms_0(num_atoms, 0.0);
+#pragma omp parallel for
+        for (int atom = 0; atom < num_atoms; atom++) {
+
+            for (int point = 0; point < total_points; point++) {
```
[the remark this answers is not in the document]
susilehtola: But that ruins the memory access.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1791,597 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+        }
+
+        iter += 1;
+    }
+
+    if (!is_converged) throw ConvergenceError<int>("MBIS", max_iter, conv, delta_rho_max_0, __FILE__, __LINE__);
+
+    // => Post-Processing <= //
+
+    // Final density calculation afiter convergence
```
susilehtola, line 2216: Typo

psi4/src/psi4/libmints/oeprop.cc
andyj10224: Turns out, the pruning was not activated the last time I tested it

I tested pruning again on tetracene (6 cores), and here are my results:

| **Scheme** | **Time (s)** | **Accurate Atomic Charges?** | **Electron Difference** |
| ------------ | ------------ | ------------------------------ | ----------------------- |
|     Robust      |     34.8         |                  Yes                             |              12E-5                 |
|     None         |     49.7         |                  Yes                             |              12E-5                 |
|     Treutler     |     27.3          |                  Yes                             |               2E-5                 |
|     None*       |     54.6          |                  Yes                             |               12E-5                |

Last Row = `DFT_WEIGHTS_TOLERANCE -1.0`


Seeing the "improvement" in both electron difference and time on the treutler scheme, I decided to test it on other molecules

Molecule, Difference (Treutler), Difference (None)
Benzene, 4E-5, 2E-5
Naphthalene, 4E-5, 7E-5
Anthracene, 3E-5, 9E-5
Tetracene, 2E-5, 12E-5
Pentacene, 1E-5, 14E-5

Therefore, I added `MBIS_PRUNING_SCHEME` as an option and set the default to `NONE`, should I instead set the default to `ROBUST` due to no difference in error and the relative safety of the robust scheme?

psi4/src/psi4/libmints/oeprop.cc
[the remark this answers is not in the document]
andyj10224: I decided to only throw an exception if the electron difference is egregious (> 0.1 electrons), else (> 0.001 electrons), I throw a warning

psi4/src/psi4/libmints/oeprop.cc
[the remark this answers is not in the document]
zachglick: This is a tricky situation. Holger is definitely right that parallel scaling would be better if the loops were switched. Unfortunately, the `atom` loop has to be outside of the `point` loop here because we're summing over all points to compute an atomic error. In other places the opposite is true, and the `point` loop has to be outer.

As Susi points out, storing data as `[atom][point]` is bad for memory access if the outer for loop is over `point`. Since we're forced to use a mix of loop orderings (sometime `atom` is outer, other times `point`), it's not clear which memory pattern is better, so I timed both patterns on a large system:

|                        | 1 core time (s) | 6 core time (s)  |
| ------------- |:---------:|:-----:|
| [atom][point] | 274 | 62.7 |
| [point][atom] | 300 |  65.3 |

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1784,6 +1792,550 @@ PopulationAnalysisCalc::compute_lowdin_charges(bool print_output) {
...
+                                                {1, 2, 2}, 
+                                                {2, 2, 2}};
+
+    // Non-redundant cartesian multipoles
+    auto mpole = std::make_shared<Matrix>("MBIS Charges: (a.u.)", num_atoms, 1);
+    auto dpole = std::make_shared<Matrix>("MBIS Dipoles: (a.u.)", num_atoms, 3);
+    auto qpole = std::make_shared<Matrix>("MBIS Quadrupoles: (a.u.)", num_atoms, 6);
+    auto opole = std::make_shared<Matrix>("MBIS Octupoles: (a.u.)", num_atoms, 10);
+    //auto qpole_stone = std::make_shared<Matrix>("MBIS Quadrupoles: (a.u.)", num_atoms, 6);
+    //auto opole_stone = std::make_shared<Matrix>("MBIS Octupoles: (a.u.)", num_atoms, 10);
```
hokru, line 2259: Were the `_stone` variants used for testing? I see them in the loop below. If they are correct and useful in the future a short comment about that and what they are would help. Otherwise better remove.

hokru (approved):
lgtm. Only a question about commented out code.
Thanks for doing the effort of benchmarking loop variants.

psi4/src/psi4/libmints/oeprop.cc
hokru: Oh I didn't notice this. I would say "yes" to `ROBUST`. Not sure how others think about this.

psi4/src/read_options.cc
```diff
@@ -247,6 +247,20 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+    options.add_int("MBIS_MAXITER", 500);
+    /*- MBIS Convergence Criteria -*/
+    options.add_double("MBIS_D_CONVERGENCE", 1.0e-8);
+    /*- MBIS Number of Radial Points -*/
+    /*- Additional Radial and/or Spherical Points may be needed for Heavier Atoms (200-300) like Zinc -*/
+    options.add_int("MBIS_RADIAL_POINTS", 75);
+    /*- MBIS Number of Spherical Points -*/
+    options.add_int("MBIS_SPHERICAL_POINTS", 302);
+    /*- Pruning scheme for MBIS Grid -*/
+    options.add_str("MBIS_PRUNING_SCHEME", "NONE");
```
loriab, line 262: * you're going to want this option to look more like the `DFT_PRUNING_SCHEME` entry, particularly by giving it a third argument for allowed values so it functions like an enum. a string w/o third arg, like at present, can take any value, so `set mbis_pruning_scheme bulldog` allowed.
* may as well make the default value (second arg) `ROBUST` as Holger suggests

psi4/src/psi4/libmints/oeprop.cc
andyj10224: The current multipoles are given by the convention in the Horton software. `_stone` represents multipoles as given by the convention in _The Theory of Intermolecular Forces, 2nd Edition_ by Anthony Stone.

psi4/src/read_options.cc
andyj10224: @hokru I ran `mbis_pruning_scheme robust` on ZnO, and I got an egregious error in the number of electrons as calculated by the grid (37.784030), whereas the number of electrons is fine when I run with all of the same options using `mbis_pruning_scheme none` (38.00079). 

For reference, here are the other grid options

scf_type df
d_convergence 8
e_convergence 10
mbis_radial_points 300
mbis_spherical_points 350

`mbis_pruning_scheme robust` worked perfectly fine for my other 4 test systems.

Is this to be expected? If so, should I leave a comment in the ZnO test system, and put a suggestion to turn off pruning in the error message?

hokru: hmm, that deviation looks suspiciously large actually. With PBE/def2-TZVP and standard grids i get much smaller deviations in ORCA and Turbomole (N=37.999..).

loriab: Andy, try adding a print grid call to the mbis https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/cubature.cc#L4490 to make quite sure you're getting what you ordered up.

hokru: I got similar results for the normal DFT grid. My feeling that something is wrong was right, bug fix submitted. Zn should behave normal after that.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -65,6 +68,10 @@
 #include <regex>
 #include <tuple>
 #include <functional>
+#include <memory>
+#include <map>
+#include <vector>
+#include <omp.h>
```
jturney, line 74: This include needs to be wrapped with `#ifdef _OPENMP` and `#endif` otherwise it won't build on a standard Mac environment.

Actually, if no specific omp functions are being used (I didn't find any during a quick scan) then the include should be removed.

psi4/src/read_options.cc
andyj10224: After implementing the bug fix, I got a much better electron count for ZnO (error of 4E-5, with robust pruning). I updated the test for ZnO, changing the reference values to match those given after the bug fix. I didn't commit the bug fix into MBIS for this PR, as it is a separate PR. Thank you for catching the error @hokru

psi4/src/psi4/libmints/oeprop.cc
andyj10224: Thank you for letting me know. I removed the `#include <omp.h>`

PR #1964, opened 2020-07-20
DFOCC Density to Wavefunction

### LINKED ISSUE #1952: DF-CCSD(T) Densities
It would extremely useful to have density fitted coupled cluster densities. I know these already exist for conventional coupled cluster, but the density fitted version is so much faster! Would love to see this feature in Psi4!

JonathonMisiewicz:
## Description
Yet another PR to get densities properly accessible on the wavefunction. This one sets the alpha and beta densities for the DF and CD methods in `dfocc` that support gradients - we should be able to close #1952. Because this is `dfocc`, I also got in some code cleanup as well.

Some of the CD methods claim to have working one-electron properties, but not gradients, but there's no elaboration. To be cautious, I haven't set densities for those methods.

## Todos
- [x] Exposes `dfocc` densities for methods with gradients.

## Checklist
- [x] `ctest -L ^dfo -j4`, `ctest -L ^dfm -j4`, and `ctest -L ^dfc -j4` all pass. I'm using that as a proxy for running all dfocc tests. Sadly, those aren't known.

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/dfocc/properties.cc
```diff
@@ -42,6 +42,10 @@ void DFOCC::oeprop() {
     outfile->Printf("\tComputing one-electron properties...\n");
 
     timer_on("oeprop");
+    // This density finagling won't be necessary if you just set the density on the wavefunction.
+    // However, CD-OMP2/OLCCD have oeprop enabled but gradients disabled for undocumented reasons.
+    // Accordingly, we'll play it safe and not set the density for those cases just yet.
+    // ...Which means we can't assume the density to be set in general. Sad. - JPM 07/2020
     SharedMatrix Da_ = SharedMatrix(new Matrix("MO-basis alpha OPDM", nmo_, nmo_));
```
robertodr, line 49: Could you change these to use `std::make_shared`?
```c++
auto Da_ =  std::make_shared<Matrix>("MO-basis alpha OPDM", nmo_, nmo_);
```

JonathonMisiewicz: Sure, general `properties.cc` cleanup done.

JonathonMisiewicz:
`pytest test_standard_suite.py` passes (after xfail and skips).

psi4/src/psi4/dfocc/prepare4grad.cc
```diff
@@ -954,5 +955,21 @@ void DFOCC::effective_pdm_gfm() {
...
+    if (reference_ == "RESTRICTED") {
+        G1->to_shared_matrix(Da);
+        Da->scale(0.5);
+        Da_ = linalg::triplet(Ca_, Da, Ca_, false, false, true);
+        Db_ = Da_;
+    } else {
+        G1A->to_shared_matrix(Da);
+        G1B->to_shared_matrix(Db);
+        Da_ = linalg::triplet(Ca_, Da, Ca_, false, false, true);
+        Db_ = linalg::triplet(Cb_, Db, Cb_, false, false, true);
```
loriab, line 970: this is taking into account that ROHF is UNRESTRICTED in dfocc? https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dfocc/dfocc.cc#L156-L157

JonathonMisiewicz: To work nicely with orbital optimization, `dfocc` is built to work with whatever orbitals your put in. The difference between UHF and ROHF is hidden in terms in the densities. That said, `dfocc` doesn't have ROHF densities anyways.

The MO densities are shamelessly stolen from formula [here](https://github.com/psi4/psi4/blob/eb2336bce2cb0c38fa6aeee35c8b3130cc5a5656/psi4/src/psi4/dfocc/properties.cc#L51), so those are right. The only question left is if the AO backtransformation is correct. And as long as Da and Db are the alpha and beta MO density matrices (see previous sentence), then those are correct too.

JoshRackers:
Wow- fast work! Thanks for doing this.

JoshRackers:
Overdue, but I have verified that this fix works for DF-CCSD(T) densities!

PR #1970, opened 2020-07-27
Delegate `basissets_` Management to `wavefunction.mintshelper()`

JonathonMisiewicz:
## Description
In master Psi, both the `Wavefunction` class and the `MintsHelper` class have their own copy of `basissets_`, a map from basis name to the actual `BasisSet`. There are `get_basisset`, `set_basisset`, and `basisset_exists` methods to manipulate `basissets_`, which are identical between the two classes. So `basissets_` and associated methods are functionally identical on the two classes.

The problem is that every `Wavefunction` object is already guaranteed to have a `MintsHelper` object, there is no reason for the `Wavefunction` object to ever have a different `basissets_` from its `MintsHelper` (because `Wavefunction`’s only use for `basissets_` is aforementioned methods), but there is no mechanism to synchronize the `basissets_` variables of a `Wavefunction` and its `MintsHelper`. So we can have two sources disagreeing about the same information.

This PR solves the problem by moving the responsibility for managing `basissets_` onto `MintsHelper`. The `Wavefunction` methods now just call the relevant `MintsHelper` methods. ~~I’m in favor of deprecating the offending `Wavefunction` methods, but I’ll open this to core developer discussion before adding that to the PR because I expect changing `Wavefunction` API will be controversial.~~

I’ve also marked some functions as override to silence annoying compiler warnings.

Obligatory Pings: @jturney for `libmints` changes, @loriab for API changes.

## Todos
- [x] Silences compiler warnings
- [x] Eliminates `wavefunction.basissets_` so that `wavefunction.mintshelper().basissets_` is the source of truth

## Questions
~~Do we want to deprecate `wavefunction.basissets()`, `wavefunction.get_basisset()`, `wavefunction.set_basisset()`, and `wavefunction.basisset_exists()` in favor of `wavefunction.mintshelper().method_name_here()`?~~

## Checklist
- [x] Quick tests, dkh tests, and x2c tests all pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -189,9 +189,6 @@ def scf_initialize(self):
     # Initilize all integratals and perform the first guess
     if self.attempt_number_ == 1:
         mints = core.MintsHelper(self.basisset())
-        if core.get_global_option('RELATIVISTIC') in ['X2C', 'DKH']:
-            mintshelper = self.mintshelper()
-            mintshelper.set_basisset('BASIS_RELATIVISTIC',self.get_basisset('BASIS_RELATIVISTIC'))
 
         self.initialize_jk(self.memory_jk_, jk=jk)
         if self.V_potential():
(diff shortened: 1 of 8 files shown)


robertodr:
I don't think there's a need to deprecate those. Users might depend on it and it's less typing if they stay in `Wavefunction`.

JonathonMisiewicz:
Just updated the last commit to fix `mintshelper_` copying. For the shallow copy, the copy and the original use the same `mintshelper_`, so all the `set_basisset` I was doing was wasted lines. For the deep copy, I just needed to copy from the original `mintshelper_`, not the new one. Follow-up PR should have a test for this, as I need some other infrastructure in place.

JonathonMisiewicz:
@loriab Can we bring this in, for the sake of the follow-up PR?

fevangelista:
Thank you for this PR. I think it was me who put a `mintshelper` object in `Wavefunction` and this PR makes so much sense.

PR #1977, opened 2020-08-07
Include ghost atoms in PCMsolver interface 

### LINKED ISSUE #1656: ghost atoms and pcmsolver
The error can appear in form of various crashes (e.g. bad alloc, weird print outs) but it seems ultimately the problem is that ghost atoms do not have an atomic radius for pcmsolver. I finally got a proper error message:

```
PCMSolver fatal error.
 In function initMolecule at line 588 of file /Users/github/builds/conda-builds/pcmsolver_1532565726046/work/src/interface/Meddle.cpp
In the molecule:
                 Geometry (in Angstrom)
   Center            X             Y             Z
------------   ------------  ------------  ------------
           N-5.29177e-11 -0.0550546 0
           H-0.477098 0.254982 -0.826358
           H-0.477098 0.254982 0.826358
           H0.954196 0.254982 0
            -5.29177e-11 -0.0550546 5.29177
            -0.477098 0.254982 4.46541
            -0.477098 0.254982 6.11813
            0.954196 0.254982 5.29177
Rotor type: Asymmetric
Some atoms do not have a radius attached.
Please specify a radius for all atoms!
```

input for testing: [ghost_pcm.txt](https://github.com/psi4/psi4/files/3256204/ghost_pcm.txt)

---

A manual specification is possible as a work-around but automatic would be nice.

hokru:
## Description
The PCMsolver interface takes as input nuclear charges (`mol.Z(i)`) to identify the elements, but did not account for ghost atoms.
Change asks now for the atomic number and thus also avoids`mol.fZ(i)=0`.

**This forms the cavity also around the ghost atoms**, which I think is the expected behaviour.

## Todos
- [x] fixes  #1656 

## Checklist
- [x ] `ctest -R pcm` passes
- [x] The `charges` array no longer contains zeros for ghost atoms (manual inspection)

## Status
- [x] Ready for review
- [ ] Ready for merge


hokru:
Can't check now but several Hartree. See  pcmsolver-scf value(slightly different basis).

hokru:
Hmm. Something is going wrong. I am getting confusing results running `nbody` calculations, even plain `nocp` ones.

hokru:
`ftrue_atomic_number` (ie including dummy atoms) gives **wrong** atomic numbers in the case of `nbody` calculations, leading non-sense energies.

I think this is because `ftrue_atomic_numbers` uses `nallatoms()` and not `natom()`, which equates to the dimer molecule dimension and _ordering_. That is likely how it should work,  @loriab can you confirm?

Thus we need to use `true_atomic_numbers` (without the leading f) here and it gives indeed correct values.

I'll modify the test to include `nbody` capacities.

psi4/src/psi4/libpsipcm/psipcm.cc
```diff
@@ -57,7 +57,7 @@ std::pair<std::vector<double>, std::vector<double>> collect_atoms(std::shared_pt
     int nat = molecule->natom();
     std::vector<double> charges(nat);
     for (int i = 0; i < nat; ++i) {
-        charges[i] = molecule->Z(i);
+        charges[i] = molecule->ftrue_atomic_number(i);
```
loriab, line 60: You plan to change this from `ftrue_atomic_number` -> `true_atomic_number`? That sounds good -- `full_atoms` at best refers back to the original pre-nbody molecule. And `natom`, `true_atomic_number`, `geometry` should be all the same with or without `f` prefix.

tests/pcmsolver/ghost/input.dat
```diff
@@ -0,0 +1,51 @@
...
+polenergy   =  -13.2757374726765178 #TEST
+totalenergy = -69.0429714614878094 #TEST
+
+molecule NH3 {
+symmetry c1
+N     -0.0000000001    -0.1040380466      0.0000000000
+H     -0.9015844116     0.4818470201     -1.5615900098
+H     -0.9015844116     0.4818470201      1.5615900098
+H      1.8031688251     0.4818470204      0.0000000000
+
```
loriab, line 13: still works with frag sep `--`, too?

hokru: yes, added `nbody` to cover this.

psi4/src/psi4/libpsipcm/psipcm.cc
hokru: change to `true_atomic_number`

hokru:
Turns out the PCM polarisation energy was way too large when using ghost atoms. I compared with what ORCA gives and I found I had to change back to 'ghost atoms' here: https://github.com/psi4/psi4/pull/1977/files#diff-acd2a9822c706c971498ea1a2669bb7eR175
Didn't try to reproduce ORCA exactly (not sure what PCM settings are needed in detail, if possible at all), but it looks similar enough.

@robertodr, can you have another look? static casting for the int<->double conversion OK for visibility?
Are there similar cases inside pcmsolver where one needs to differentiate between cavity-building-atoms and charge-carrying atoms?


I also manually checked the PEDRA outputs and the ghost atom-system uses the correct dimer-sized cavity.

tests/pcmsolver/ghost/input.dat
PeterKraus: That looks much more reasonable.

robertodr:
> Didn't try to reproduce ORCA exactly (not sure what PCM settings are needed in detail, if possible at all), but it looks similar enough.
I am not familiar with ORCA's implementation, but you could try to set the same exact permittivity and a very fine discretization of the cavity to get as close as possible.
 
> @robertodr, can you have another look? static casting for the int<->double conversion OK for visibility?
Yes, `static_cast` is reasonable here.

> Are there similar cases inside pcmsolver where one needs to differentiate between cavity-building-atoms and charge-carrying atoms?
This is the only case where I can think a reasonable default can be devised.

PR #1980, opened 2020-08-11
shift mp2.5 and mp3 default from conv to df

loriab:
## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] conventional only (not orbital optimized) MP2.5 and MP3 are now density-fit by default, fixes #1801 
- [x] note that this bifurcates the role of `mp_type`, with the above cases imposed in proc.py, leaving the keyword unchanged for others -- omp2.5, omp3, mp4, adc2
- [x] in occ, LCCD and OLCCD energies were often failing (to match to 1.e-6) with default convcrit. though there's a `abs(DE) < e_conv` test for iterations, the actual progressing of DE through the iterations has a lot of energy increases, so often a small energy increase satisfies the test. I reduced the energy test by an order of magnitude to allow a little more settling of iterations. obligatory @JonathonMisiewicz ping, who may be rightly appalled. it does fix up energy and 5-point findif gradient by energy convergence across the std suite.
- [x] in dfocc, most DF analytic gradients from MP2 to CCSD were failing with default convcrit w/o tightening pcg_convergence by an order of magnitude. so, I tied pcg_convergence to e_convergence if not specified. again, Jonathon's input welcome.
- [x] I maybe should write a warning to output if these methods run w/o mp_type specified so the change isn't wholly unmarked (besides release notes and docs)
- [x] bump qcel and qcng to v0.16

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #1988, opened 2020-08-25
Fix `occ` RMS Orbital Gradient

JonathonMisiewicz:
## Description
The most important part of this PR is the changes to `occ/mograd.cc`. The previous formula for the RMS MO gradient was incorrect, causing it to be too large by a factor of `sqrt(occ*vir)`. Even for systems in the test suite, this can be as large as a factor of 10. **As a result, under this PR, orbital-optimized `occ` computations will take more iterations; it was not actually achieving the accuracy it claimed previously.** We may loosen the default convergence criteria as a result of this, but exploring that is deferred to the standard suite pass in #1980.

Also included in this PR is commenting and removing some unused (not even compiled!) `dfocc` code found while investigating issues tangential to #1980.

Obligatory @loriab ping for #1980. Obligatory @ssh2 ping for OMP convergence changes.

## Todos
- [x] Fix the RMS orbital residual calculation in `occ`
- [x] Remove 1100+ lines of dead code from `dfocc`

## Checklist
- [x] `ctest -R "^omp|^olccd"` and `pytest -k test_olccd` pass

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/occ/mograd.cc
```diff
@@ -129,12 +129,12 @@ void OCCWave::mograd() {
...
         for (int i = 0; i < nidpA; i++) rms_wogA += wogA->get(i) * wogA->get(i);
         norm_wogA = std::sqrt(rms_wogA);
-        rms_wogA = std::sqrt(rms_wogA) / nidpA;
+        rms_wogA = std::sqrt(rms_wogA / nidpA);
 
         rms_wogB = 0;
         for (int i = 0; i < nidpB; i++) rms_wogB += wogB->get(i) * wogB->get(i);
         norm_wogB = std::sqrt(rms_wogB);
-        rms_wogB = std::sqrt(rms_wogB) / nidpB;
+        rms_wogB = std::sqrt(rms_wogB / nidpB);
```
loriab, line 137: this may win "most consternation caused by misplaced parenthesis"

jturney: Oh wow.

PeterKraus (approved):

Issue #1985, opened 2020-08-24
Actual memory usage appears much larger than psi4.set_memory()


slochower:
I see memory usage significantly above `psi4.set_memory()` limit. I do see the note in the documentation that memory usage might be _slightly_ higher, but if I even request within 20% of a node's total memory (requesting 300 out of ~376 GB), I risk crashing the entire node. Even now, running with `psi4.set_memory("200 GB")`, I see persistent VIRT usage of 259 GB and RES 256 GB. I know that determining the _actual_ amount of [memory consumed by a process is tricky](https://stackoverflow.com/questions/131303/how-to-measure-actual-memory-usage-of-an-application-or-process), but this is important to me to prevent the scheduler from killing my jobs. After some troubleshooting, I am requesting 300 GB from the scheduler, but 200 GB from psi4, and end up with ~260 GB used. Is this expected? 

Relatedly, but separate, is there a way to estimate what the scratch file size may be? I was planning to use node-local storage until I found out my scratch files are 1.5+ TB. Any rule of thumb way of estimating this ahead of time (number of atoms × level of theory × some magic thing?)

CDSherrill:
You may have found a memory allocation/usage bug.  What kind of
computation is this?

Unfortunately we don't really have a way to estimate scratch usage at
present. It's a feature we'd like to add, but it needs to be added for
every module separately, so is a somewhat daunting task.  In principle, can
be known ahead of time based on number of basis functions, number of atoms,
etc., but is different for every method (and sometimes depends on the
particular algorithm used within that method).

hokru:
Would be good to know the psi4 version and the input options used.

I know that `scf_type pk` can get very greedy for larger calculations.

slochower:
I'm running a single point energy calculation with `psi4.energy("ccsd(t)/aug-cc-pvdz")`. 

The input is simple. Just:

```
import psi4
from psi4 import *

psi4.core.set_output_file('xxx.dat', False)

psi4.set_memory('200 GB')
psi4_io = core.IOManager.shared_object()
psi4_io.set_default_path("xxx")
psi4.set_num_threads(24)
```
followed by the molecule specification with `psi4.geometry()` and then
```
psi4.energy("ccsd(t)/aug-cc-pvdz")
```

(I know this will take a lot of memory and disk space, but I believe the memory limit should stand.)

- psi4 version 1.3.2 in a new python 3.7 conda environment
- RHEL 7.2

hokru:
[...]

PR #1990, opened 2020-08-26
PK memory fixes

hokru:
## Description
The PK(reordered, disk) algorithm was using twice as much memory as it should.
This is due to not deleting integral and disk-space buffers correctly:
 * During the pre-iterations the `batch size` buffer that prepares the disk space ("pre-striping", within `zero_disk`) and used during the SCF for processing the integrals on disk is not released. 
* The `buffer size` buffer that is used to calculate the integrals is not released after the computation is done

Both buffers are actually the same size and effectively lead to a doubling of the memory demand.

## Todos
- [x] memory leaks in PK(reordered, disk) algorithm addressed
- [x]  Likely solution to #1985 

## Questions
- [ ] Is there a better solution to `AIO->synchronize()` in-between the `zero_disk` operations?
- [ ] Will there be an issue with: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/PKmanagers.cc#L886  now that `buffer()` returns a reference?

## Checklist
- [x] ctest -L quick/dft pass
- [x] memory usage on `top` eyeballed like a hawk

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -99,7 +99,7 @@ jobs:
         pybind11 \
         pymdi \
         -c conda-forge
-      conda install adcc -c adcc -c conda-forge
+      #conda install adcc -c adcc -c conda-forge
       pip install git+https://github.com/i-pi/i-pi.git@master-py3
       conda list
     displayName: 'Build Environment'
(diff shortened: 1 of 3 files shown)


robertodr (approved):

loriab:
Looks like adcc and I have gotten out of sync, conda-wise. try commenting out https://github.com/psi4/psi4/blob/master/.azure-pipelines/azure-pipelines-linux.yml#L102 and OFFing https://github.com/psi4/psi4/blob/master/.azure-pipelines/azure-pipelines-linux.yml#L125

PeterKraus:
Wow, this would explain some random crashes I got when using `pk` while having no available swap.

PR #1991, opened 2020-08-26
OEI libtrans Cleanup

JonathonMisiewicz:
## Description
This PR does some miscellaneous cleanup of the OEI part of `libtrans` and will probably lead to another series of smaller PRs involving `libtrans`. The focus of this PR is to take away from `libtrans` responsibility for creating unused or barely used one-electron objects.

## Todos
- [x] Remove unused OEI functions from `libtrans`
- [x] Remove creation of Fock matrices from being the responsibility of `libtrans`
- [x] Add a readme file explaining why `libtrans` is bothering with a frozen core operator.

## Checklist
- [x] `pytest test_standard_suite.py` and all C-tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libtrans/README.txt
+++ b/psi4/src/psi4/libtrans/README.txt
@@ -1,0 +1,16 @@
+General Notes About the Code
+
+1. While the primary purpose of the code is transformations between the AO and MO bases, 
+   some codes (mrcc, detci, and especially cc) request that all tasks involving frozen core
+   orbitals be either done by libtrans or converted into effective quantities free of frozen
+   core orbitals - that way, they don't need to worry about frozen core orbtials at all, and
+   the calculation retains all the simplicity of one where core electrons simply don't exist.
+   In particular, libtrans has the following responsibilities:
+   * Computing all energy contributions involving frozen core orbitals only and putting that
+     result into frozen_core_energy_. Used to sanity-check the HF energy.
+   * Constructing the "frozen-core operator", which is the core hamiltonian for non-frozen orbitals
+     plus the Couloumb and exchange contributions terms arising from the electric field of
+     the frozen core orbitals. Used to sanity check the correlated energy by computing it from
+     density matrices.
+   * Constructing the contribution to the (one-electron) density matrix solely due to frozen
+     core orbitals. This may needed for property calculations later.
--- a/psi4/src/psi4/libtrans/README.txt
+++ b/psi4/src/psi4/libtrans/README.txt
@@ -5,12 +5,10 @@ General Notes About the Code
    orbitals be either done by libtrans or converted into effective quantities free of frozen
    core orbitals - that way, they don't need to worry about frozen core orbtials at all, and
    the calculation retains all the simplicity of one where core electrons simply don't exist.
-   In particular, libtrans has the following responsibilities:
+   In particular, libtrans:integraltransform_sort_so_tei.cc has the following responsibilities:
    * Computing all energy contributions involving frozen core orbitals only and putting that
      result into frozen_core_energy_. Used to sanity-check the HF energy.
    * Constructing the "frozen-core operator", which is the core hamiltonian for non-frozen orbitals
      plus the Couloumb and exchange contributions terms arising from the electric field of
      the frozen core orbitals. Used to sanity check the correlated energy by computing it from
      density matrices.
-   * Constructing the contribution to the (one-electron) density matrix solely due to frozen
-     core orbitals. This may needed for property calculations later.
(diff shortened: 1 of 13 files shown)


psi4/src/psi4/libtrans/README.txt
```diff
@@ -0,0 +1,13 @@
...
+1. While the primary purpose of the code is transformations between the AO and MO bases, 
+   some codes (mrcc, detci, and especially cc) request that tasks involving frozen core
+   orbitals be either done by libtrans or converted into effective quantities free of frozen
+   core orbitals - that way, they don't need to worry about frozen core orbitals at all, and
+   the calculation retains all the simplicity of one where core electrons simply don't exist.
+   In particular, libtrans:integraltransform_sort_so_tei.cc has the following responsibilities:
+   * Computing all energy contributions involving frozen core orbitals only and putting that
+     result into frozen_core_energy_. Used to sanity-check the HF energy.
+   * Constructing the "frozen-core operator", which is the core hamiltonian for non-frozen orbitals
+     plus the Couloumb and exchange contributions terms arising from the electric field of
```
andysim, line 12: Looking back, I should have added this kind of documentation when shoehorning the code into psi3/psi4.  Thanks for taking the time to help future devs with these notes.

hokru (approved):
lgtm. Will need a rebase to pass CI I assume.

JonathonMisiewicz:
No, I'm still getting `pywrap` timeouts on Windows. Any ideas? I can't debug this readily.

loriab:
I can't find it again (hopefully rebased away), but I think I saw Windows pywrap on the mp3 PR, too, once. So before starting time-consuming investigations, let's get #2002 in, then #1980 in, then try a rebase on this one. If that's not the sol'n, then the only thing that comes to mind is that those pywraps involve detci, which this PR touches.

JonathonMisiewicz:
Same timeout as before. `pywrap-alias` doesn't seem to enter `detci`, so I doubt that's the problem. Any other advice on diagnosing this?

loriab:
I think pywrap-alias does use some of detci's more obscure methods (below). I was also remembering a bad bout I had with ddd before last Christmas where Windows only hung up for detci mp5 run only by exe or by api.

```
(py38) >>> grep -e 'PASSED' -e 'D E T C I' ../tests/pywrap-alias/input.out 
    [1]  Au std...........................................................................PASSED
    [2]  MP2 sp...........................................................................PASSED
          MP2 route check.................................................................PASSED
    [3]  MP3 occ..........................................................................PASSED
          MP3 route check.................................................................PASSED
    [4]  MP3 detci (energy and route check)...............................................PASSED
    [5]  MP4 fnocc........................................................................PASSED
                            (a 'D E T C I' module)
    [6]  MP11 sp..........................................................................PASSED
                            (a 'D E T C I' module)
    [7]  ZAPT4 sp.........................................................................PASSED
                            (a 'D E T C I' module)
    [8]  FCI sp...........................................................................PASSED
    [9]  MP2.5 sp.........................................................................PASSED
    [10] cbs()............................................................................PASSED
    [11] db().............................................................................PASSED
    [12] CISD sp..........................................................................PASSED
```

psi4/driver/procrouting/proc.py
```diff
@@ -2138,6 +2138,7 @@ def run_occ(name, **kwargs):
         ref_wfn.semicanonicalize()
 
     occ_wfn = core.occ(ref_wfn)
+    raise Exception
```
PeterKraus, line 2141: Is this meant to be here?

JonathonMisiewicz: Anything in the commit marked "Debug" is my attempt to debug a very annoying Windows-specific bug and will not be part of the final PR.

JonathonMisiewicz:
We are _finally_ passing on Windows. The problem was quite subtle and involved treacherous raw pointers.

The old code used `trans_one` and would pass in a pointer to position N in a `*double`, where N was the number of orbitals from previous irreps. The relevant test cases had a final irrep with no orbitals, so the old code would pass in a pointer to one position _after_ the last value. Which was fine, because it was never used. That stopped being fine, according to the Windows compiler, when I switched from `*double` to `std::array`. To be fair, trying to access a non-existent index should never have worked in the first place.

The fix I adopted was to eliminate those `trans_one` calls with `libmints::Matrix` operations, with the result that we're using less specialized `libtrans` tech and fewer LoC. Unsurprisingly, I am okay with that.

@loriab, unless you want to Jet to double-check that last commit, I think we're clear to bring this in.

loriab:
Wow, congratulations on tracking that down

PR #1993, opened 2020-08-27
Skip Ghost-Nuclear interaction

ferchault:
## Description
Fixes calculation of nuclear-nuclear interaction energy if a QM/MM point charge and a Ghost atom are at the same position as in the example below. Since a Ghost atom has no nuclear charge, it should not contribute. At any other position apart from the nuclear centers, it would not contribute. But since at this position, the distance between the Ghost and the atom is zero such that the current code wrongly produces a NaN energy.

Setups like this are relevant for [alchemical calculations](https://dx.doi.org/10.1103/PhysRevResearch.2.023220) where fractional nuclear charges are required.

```
molecule mol {
@N 0. 0. 1.
N 0. 0. 0.
no_reorient
no_com
symmetry c1
-7 1
}

Chrgfield = QMMM()
Chrgfield.extern.addCharge(7.05, 0., 0., 1.)
psi4.set_global_option_python('EXTERN', Chrgfield.extern)

set basis 6-31G
set reference rhf
e, wfn = energy('pbe', return_wfn=True)
```

## Todos

## Checklist
- [x] [All or relevant fraction of full tests run on local machine](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/extern.cc
+++ b/psi4/src/psi4/libmints/extern.cc
@@ -315,6 +315,10 @@ double ExternalPotential::computeNuclearEnergy(std::shared_ptr<Molecule> mol) {
         double yA = mol->y(A);
         double zA = mol->z(A);
         double ZA = mol->Z(A);
+        // skip Ghost interaction
+        if (ZA == 0) {
+            continue;
+        }
 
         for (size_t B = 0; B < charges_.size(); B++) {
             double ZB = std::get<0>(charges_[B]);
--- a/psi4/src/psi4/libmints/extern.cc
+++ b/psi4/src/psi4/libmints/extern.cc
@@ -315,23 +315,21 @@ double ExternalPotential::computeNuclearEnergy(std::shared_ptr<Molecule> mol) {
         double yA = mol->y(A);
         double zA = mol->z(A);
         double ZA = mol->Z(A);
-        // skip Ghost interaction
-        if (ZA == 0) {
-            continue;
-        }
 
-        for (size_t B = 0; B < charges_.size(); B++) {
-            double ZB = std::get<0>(charges_[B]);
-            double xB = convfac * std::get<1>(charges_[B]);
-            double yB = convfac * std::get<2>(charges_[B]);
-            double zB = convfac * std::get<3>(charges_[B]);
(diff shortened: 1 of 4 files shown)


PeterKraus (approved):
Makes sense to me.

psi4/src/psi4/libmints/extern.cc
```diff
@@ -315,6 +315,10 @@ double ExternalPotential::computeNuclearEnergy(std::shared_ptr<Molecule> mol) {
         double yA = mol->y(A);
         double zA = mol->z(A);
         double ZA = mol->Z(A);
+        // skip Ghost interaction
+        if (ZA == 0) {
```
hokru, line 319: Because `ZA` is a double it is better to avoid `==` and go with a `if (ZA > 0) {` loop around the computations instead.

hokru:
Welcome to psi4 and thanks for the PR.

Could you add a regression test with your example? http://psicode.org/psi4manual/master/add_tests.html
E.g. a third `extern` version would do.

If anything is unclear, please ask here or join us on slacks.

---

I am wondering. In principle `mol.set_nuclear_charge(0,7.05)` should work to modify Z. However, the C molecule object doesn't get the changes only the python one.

```
molecule mol {
N 0. 0. 1.
N 0. 0. 0.
no_reorient
no_com
symmetry c1
0 1
}

set basis 6-31G
set reference rhf

mol.set_nuclear_charge(0,7.05)
print(mol.Z(0),mol.Z(1))
nn = mol.nuclear_repulsion_energy()
mol.update_geometry()
e, wfn = energy('scf',return_wfn=True)
print(nn)
print(wfn.variable('NUCLEAR REPULSION ENERGY'))
```

loriab:
Fix looks good, thanks!

For the test, could you edit it to trigger your original problem and test the solution? I think it's still like extern2 at the moment.

> In principle mol.set_nuclear_charge(0,7.05) should work to modify Z

Yeah, I tried for a couple hours once to see about setting fractional nuclear charges in Molecule that made a difference to a calculation; no luck.

hokru (approved):

PR #1994, opened 2020-08-27
Improvements to basis set handling.

### LINKED ISSUE #628: Possible g94 basis set format incompatibility
Trying to run the calculation in http://forum.psicode.org/t/problem-with-detci-and-7z-basis-set/423, I noticed that psi4 is erroneously using the buggy convention used by the basis set exchange where angular momentum shells are S, P, D, F, G, H, I, K, ... However, in the g94 basis set format l=7 is not K, but J.

Namely, back when I was allowed to use Gaussian, I got the following info from their help:
```
Another issue which arises with these high angular momentum 
functions is an inconsistency in how they are labeled.  Gaussian
labels the L=7 functions with the letter j rather than the letter
k.  Other authors omit j so that l=7 is k.  So if your input has

K n 1.0
...

you are actually getting an L=8 basis set.  Change your input to

L=7 n 1.0

instead.
```

susilehtola:
## Description
Improvements to basis set handling, including
- elimination of duplicate code and introducing array size checks via use of std::string in gshell
- sanity checks for angular momentum parsing, eliminating segfaults
- support for explicit definition of angular momentum in the basis set input (fixes #628)

## Todos
- [x] Eliminated duplicate code and introduced array size checks via std::string in gshell.
- [x] Introduced angular momentum checks in the basis set parser.
- [x] Added support for L=n format into the basis set parser.

## Questions
- [x] The basis set parser is doing regexp validation for the shell entry. Thanks to @PhillCli for supplying a working regexp.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


robertodr (approved):

hokru:
Linux CI fails because of adcc (see https://github.com/psi4/psi4/pull/1990#issuecomment-681047414) but not sure about Windows which complains about missing `ctype`.

susilehtola:
If someone would be so kind as to put in the missing fix to
https://github.com/psi4/psi4/blob/e326f61e7b292c5131a029ef2ef8c0ab7b6d8e6b/psi4/driver/qcdb/libmintsbasissetparser.py#L118
that is now tripping the new feature at 
https://github.com/psi4/psi4/blob/e326f61e7b292c5131a029ef2ef8c0ab7b6d8e6b/psi4/driver/qcdb/libmintsbasissetparser.py#L242

then this pull should be okay.

PhillCli:
> If someone would be so kind as to put in the missing fix to
> 
> https://github.com/psi4/psi4/blob/e326f61e7b292c5131a029ef2ef8c0ab7b6d8e6b/psi4/driver/qcdb/libmintsbasissetparser.py#L118
> 

This should work:
```^\s*(\w+|L=\d+)\s*(\d+)\s*(-?\d+\.\d+)```

**Comment by susilehtola** [comment]
Thanks @PhillCli, so it does!

**Comment by susilehtola** [comment]
@hokru yeah, now the fails are just due to adcc. #1990 could be merged now so I can rebase.

File: psi4/driver/qcdb/libmintsbasissetparser.py [hunk 1]
```diff
@@ -247,6 +247,8 @@ def parse(self, symbol, dataset):
 
                                 if len(shell_type) == 1:
                                     am = shell_to_am[shell_type[0]]
```
loriab, line 249:
```suggestion
                                if len(shell_type) == 1 or (len(shell_type) >= 3 and shell_type[:2] == 'L='):
                                    if len(shell_type) == 1:
                                        am = shell_to_am[shell_type[0]]
                                    else:
                                        am = int(shell_type[2:])
```
Would this do away with the new block below?

psi4/driver/qcdb/libmintsbasissetparser.py
```diff
@@ -311,6 +315,37 @@ def parse(self, symbol, dataset):
                                         gaussian_type, 0, center, 0, 'Unnormalized'))
                                     shell_list.append(ShellInfo(am2, contractions2, exponents,
                                         gaussian_type, 0, center, 0, 'Unnormalized'))
+
+                                elif len(shell_type) >= 3 and shell_type[:2] == 'L=':
```
PeterKraus, line 319: I'm wondering if it wouldn't be cleaner to combine `len(shell_type) == 1` and `len(shell_type) >=3 and shell_type[:2] == 'L='` cases, as the only line that differs is the am assignment. Then we don't have the same logic 3 times...

Also, is `L=am` a rigorous spec, or can it have spaces (eg. `L = 5`)? Because the latter won't be parsed properly (one could use `.startswith("L")` and then `.split("=")[1]` to get the `am`).

PeterKraus (changes requested):

psi4/driver/qcdb/libmintsbasissetparser.py
susilehtola: Done. And yes that should be rigorous.

psi4/driver/qcdb/libmintsbasissetparser.py
susilehtola: Resolved differently

PeterKraus (approved):

susilehtola:
I guess this can be merged now; the fix to the 'J' vs 'K' question can be addressed in another pull.

PR #2022, opened 2020-10-20
CC response vars to wfn: polar, optrot, quadrupole

bgpeyton:
## Description
Adds the tensors generated from the CC response code to the reference wave function, eventually to be included in the JSON output for a [QCEng](https://github.com/molssi/qcengine) run. Also patched CC-level dipoles and (static) quadrupoles in `proc.py` to be consistent with other naming conventions, allowing them to pass through `schema_wrapper.py`. This means that all (ground state, static) CC one-electron properties should follow the naming convention used on wfn. Old excited state (EOM) and dynamic property variables remain unchanged. 

Updates to the schema wrapper to allow the other tensors though are coming in a separate PR. 

## Todos
- [x] Electric dipole polarizability tensor
- [x] Optical rotation (electric-dipole/magnetic-dipole) tensor
- [x] Electric dipole-quadrupole polarizability tensor (stored as three 3x3 components) (calculated through `roa.cc`)
- [x] Patch CC dipole and (static) quadrupole QCVars
- [x] New pytest for all new QCVars
- [x] Update `cc46` to adhere to new ground state `psi4.core.variables()` names

## Questions
- [x] `tests/pytests/test_qcschema.py` doesn't include any properties (yet). Are the existing returns of dipoles / quadrupoles in `schema_wrapper` tested anywhere? @loriab 

## Checklist
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests) `ctest -L cc` comes clean.
- [x] New pytest `test_ccresponse.py` comes clean.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -3008,7 +3008,7 @@ def run_cc_property(name, **kwargs):
     if n_one > 0:
         # call oe prop for GS density
         oe = core.OEProp(ccwfn)
-        oe.set_title("CC")
+        oe.set_title(name.upper())
         for oe_name in one:
             oe.add(oe_name.upper())
         oe.compute()
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -589,7 +589,6 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
         mtd = json_data["model"]["method"].upper()
 
         # Dipole/quadrupole still special case
-        # CC* dipole/quadrupole use umbrella qcvars
         if "dipole" in kwargs["properties"]:
             ret["dipole"] = _serial_translation(psi_props[mtd + " DIPOLE"], json=json_serialization)
         if "quadrupole" in kwargs["properties"]:
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -589,10 +589,17 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
         mtd = json_data["model"]["method"].upper()
 
         # Dipole/quadrupole still special case
+        # CC* dipole/quadrupole use umbrella qcvars
         if "dipole" in kwargs["properties"]:
-            ret["dipole"] = _serial_translation(psi_props[mtd + " DIPOLE"], json=json_serialization)
+            if "CC" in mtd:
+                ret["dipole"] = _serial_translation(psi_props["CC DIPOLE"], json=json_serialization)
+            else:
+                ret["dipole"] = _serial_translation(psi_props[mtd + " DIPOLE"], json=json_serialization)
         if "quadrupole" in kwargs["properties"]:
-            ret["quadrupole"] = _serial_translation(psi_props[mtd + " QUADRUPOLE"], json=json_serialization)
+            if "CC" in mtd:
+                ret["quadrupole"] = _serial_translation(psi_props["CC QUADRUPOLE"], json=json_serialization)
+            else:
+                ret["quadrupole"] = _serial_translation(psi_props[mtd + " QUADRUPOLE"], json=json_serialization)
         ret.update(_convert_variables(wfn.variables(), context="properties", json=json_serialization))
 
         json_data["return_result"] = ret
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -591,15 +591,9 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
         # Dipole/quadrupole still special case
         # CC* dipole/quadrupole use umbrella qcvars
         if "dipole" in kwargs["properties"]:
-            if "CC" in mtd:
-                ret["dipole"] = _serial_translation(psi_props["CC DIPOLE"], json=json_serialization)
-            else:
-                ret["dipole"] = _serial_translation(psi_props[mtd + " DIPOLE"], json=json_serialization)
+            ret["dipole"] = _serial_translation(psi_props[mtd + " DIPOLE"], json=json_serialization)
         if "quadrupole" in kwargs["properties"]:
-            if "CC" in mtd:
-                ret["quadrupole"] = _serial_translation(psi_props["CC QUADRUPOLE"], json=json_serialization)
-            else:
-                ret["quadrupole"] = _serial_translation(psi_props[mtd + " QUADRUPOLE"], json=json_serialization)
+            ret["quadrupole"] = _serial_translation(psi_props[mtd + " QUADRUPOLE"], json=json_serialization)
         ret.update(_convert_variables(wfn.variables(), context="properties", json=json_serialization))
 
         json_data["return_result"] = ret
--- a/psi4/src/psi4/cc/ccresponse/optrot.cc
+++ b/psi4/src/psi4/cc/ccresponse/optrot.cc
@@ -78,6 +78,7 @@
 #include "psi4/libpsio/psio.h"
 #include "psi4/libqt/qt.h"
 #include "psi4/libmints/molecule.h"
+#include "psi4/libmints/wavefunction.h"
 #include "psi4/psi4-dec.h"
 
 #include "MOInfo.h"
@@ -95,7 +96,7 @@ void compute_X(const char *pert, int irrep, double omega);
 void linresp(double *tensor, double A, double B, const char *pert_x, int x_irrep, double omega_x, const char *pert_y,
              int y_irrep, double omega_y);
 
-void optrot(std::shared_ptr<Molecule> molecule) {
+void optrot(std::shared_ptr<Wavefunction> ref_wfn) {
     double ***tensor_rl, ***tensor_pl, ***tensor_rp, **tensor0;
     double **tensor_rl0, **tensor_rl1, **tensor_pl0, **tensor_pl1;
     char **cartcomp, pert[32], pert_x[32], pert_y[32];
@@ -104,6 +105,8 @@ void optrot(std::shared_ptr<Molecule> molecule) {
     double *rotation_rl, *rotation_pl, *rotation_rp, *rotation_mod, **delta;
     char lbl1[32], lbl2[32], lbl3[32];
     int compute_rl = 0, compute_pl = 0;
+    std::shared_ptr<Molecule> molecule;
+    molecule = ref_wfn->molecule();
 
     /* Booleans for convenience */
     if (params.gauge == "LENGTH" || params.gauge == "BOTH") compute_rl = 1;
--- a/psi4/src/psi4/cc/ccresponse/optrot.cc
+++ b/psi4/src/psi4/cc/ccresponse/optrot.cc
@@ -87,6 +87,7 @@
 #define EXTERN
 #include "globals.h"
 #include "psi4/physconst.h"
+#include "psi4/libmints/matrix.h"
 
 namespace psi {
 namespace ccresponse {
@@ -184,15 +185,27 @@ void optrot(std::shared_ptr<Wavefunction> ref_wfn) {
             psio_read_entry(PSIF_CC_INFO, lbl1, (char *)tensor0[0], 9 * sizeof(double));
         }
 
-        if (params.wfn == "CC2")
+        std::stringstream tag_tensor0;
+        if (params.wfn == "CC2") {
             outfile->Printf("\n     CC2 Optical Rotation Tensor (Velocity Gauge): %s\n", lbl1);
-        else if (params.wfn == "CCSD")
+            tag_tensor0 << "CC2 OPTICAL ROTATION TENSOR (VEL) @ 0";
+        }
+        else if (params.wfn == "CCSD") {
             outfile->Printf("\n    CCSD Optical Rotation Tensor (Velocity Gauge): %s\n", lbl1);
+            tag_tensor0 << "CCSD OPTICAL ROTATION TENSOR (VEL) @ 0";
+        }
 
         outfile->Printf("  -------------------------------------------------------------------------\n");
         outfile->Printf("   Evaluated at omega = 0.00 E_h (Inf nm, 0.0 eV, 0.0 cm-1)\n");
         outfile->Printf("  -------------------------------------------------------------------------\n");
         mat_print(tensor0, 3, 3, "outfile");
+        auto tensor_mat0 = std::make_shared<Matrix>(3, 3);
+        for (int m = 0; m < 3; m++) {
+            for (int n = 0; n < 3; n++) {
+                tensor_mat0->set(m,n,tensor0[m][n]);
+            }
+        }
+        ref_wfn->set_array_variable(tag_tensor0.str(),tensor_mat0);
     }
 
     for (i = 0; i < params.nomega; i++) {
@@ -419,10 +432,15 @@ void optrot(std::shared_ptr<Wavefunction> ref_wfn) {
         prefactor *= 288.0e-30 * pc_pi * pc_pi * pc_na * bohr2a4;
 
         if (compute_rl) {
-            if (params.wfn == "CC2")
+            std::stringstream tag_tensor_rl;
+            if (params.wfn == "CC2") {
                 outfile->Printf("\n            CC2 Optical Rotation Tensor (Length Gauge):\n");
-            else if (params.wfn == "CCSD")
+                tag_tensor_rl << "CC2 OPTICAL ROTATION TENSOR (LEN) @ " << params.omega[i];
+            }
+            else if (params.wfn == "CCSD") {
                 outfile->Printf("\n           CCSD Optical Rotation Tensor (Length Gauge):\n");
+                tag_tensor_rl << "CCSD OPTICAL ROTATION TENSOR (LEN) @ " << params.omega[i];
+            }
 
             outfile->Printf("  -------------------------------------------------------------------------\n");
             outfile->Printf("   Evaluated at omega = %8.6f E_h (%6.2f nm, %5.3f eV, %8.2f cm-1)\n", params.omega[i],
@@ -430,6 +448,13 @@ void optrot(std::shared_ptr<Wavefunction> ref_wfn) {
                             pc_hartree2wavenumbers * params.omega[i]);
             outfile->Printf("  -------------------------------------------------------------------------\n");
             mat_print(tensor_rl[i], 3, 3, "outfile");
+            auto tensor_mat_rl = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat_rl->set(m,n,tensor_rl[i][m][n]);
+                }
+            }
+            ref_wfn->set_array_variable(tag_tensor_rl.str(),tensor_mat_rl);
 
             TrG_rl = -(tensor_rl[i][0][0] + tensor_rl[i][1][1] + tensor_rl[i][2][2]) / (3.0 * params.omega[i]);
 
@@ -458,10 +483,15 @@ void optrot(std::shared_ptr<Wavefunction> ref_wfn) {
         }
 
         if (compute_pl) {
-            if (params.wfn == "CC2")
+            std::stringstream tag_tensor_pl;
+            if (params.wfn == "CC2") {
                 outfile->Printf("\n          CC2 Optical Rotation Tensor (Velocity Gauge):\n");
-            else if (params.wfn == "CCSD")
+                tag_tensor_pl << "CC2 OPTICAL ROTATION TENSOR (VEL) @ " << params.omega[i];
+            }
+            else if (params.wfn == "CCSD") {
                 outfile->Printf("\n         CCSD Optical Rotation Tensor (Velocity Gauge):\n");
+                tag_tensor_pl << "CCSD OPTICAL ROTATION TENSOR (VEL) @ " << params.omega[i];
+            }
 
             outfile->Printf("  -------------------------------------------------------------------------\n");
             outfile->Printf("   Evaluated at omega = %8.6f E_h (%6.2f nm, %5.3f eV, %8.2f cm-1)\n", params.omega[i],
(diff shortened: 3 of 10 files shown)


loriab (commented):
nice, thanks!

No, test_schema isn't hitting properties, but adding checks or new tests to prop or response ctest tests or pytests/test_psi4.py would work.

psi4/driver/schema_wrapper.py
```diff
@@ -589,10 +589,17 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
         mtd = json_data["model"]["method"].upper()
 
         # Dipole/quadrupole still special case
+        # CC* dipole/quadrupole use umbrella qcvars
         if "dipole" in kwargs["properties"]:
-            ret["dipole"] = _serial_translation(psi_props[mtd + " DIPOLE"], json=json_serialization)
+            if "CC" in mtd:
+                ret["dipole"] = _serial_translation(psi_props["CC DIPOLE"], json=json_serialization)
```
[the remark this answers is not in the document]
bgpeyton: I agree. Let me dig that up, shouldn't be hard to change. Maybe I can do it for static quadrupoles as well..

psi4/driver/procrouting/proc.py
```diff
@@ -3008,7 +3008,7 @@ def run_cc_property(name, **kwargs):
     if n_one > 0:
         # call oe prop for GS density
         oe = core.OEProp(ccwfn)
-        oe.set_title("CC")
+        oe.set_title(name.upper())
```
bgpeyton, line 3011: `CC DIPOLE` wfn var is set by OEProp, so just had to push the proper title through `prop.py`. I would have used the above-defined `this_name` variable, but EOM needs to keep the `-`. 

The `Process::environment` variables are a little more annoying to change. Shall we keep them as is to keep from breaking downstream, or insist on consistency btwn Proc::env and wfn? @loriab

loriab: agree, keep `P::e` as is. I'm regarding qcvars on wfn as the opportunity to clean up wrongs we did with `P::e`. eventually direct setting of `P::e` will be deprecated, and wfn vars will be copied up.

bgpeyton:
Changed back to `NM` on the QCVars strings and added a new test. Opted for a separate pytest, clocking in at around a minute and a half I don't think I can really call it a smoke or quick test. 

Pending checks and review, I think this is ready to go from my end.

psi4/src/psi4/cc/ccresponse/optrot.cc
```diff
@@ -104,6 +106,8 @@ void optrot(std::shared_ptr<Molecule> molecule) {
     double *rotation_rl, *rotation_pl, *rotation_rp, *rotation_mod, **delta;
     char lbl1[32], lbl2[32], lbl3[32];
     int compute_rl = 0, compute_pl = 0;
+    std::shared_ptr<Molecule> molecule;
+    molecule = ref_wfn->molecule();
```
robertodr, line 110: Would `auto molecule = ref_wfn->molecule()` work?

psi4/src/psi4/cc/ccresponse/optrot.cc
```diff
@@ -416,50 +432,80 @@ void optrot(std::shared_ptr<Molecule> molecule) {
...
                             (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                             pc_hartree2wavenumbers * params.omega[i]);
             outfile->Printf("  -------------------------------------------------------------------------\n");
             mat_print(tensor_rl[i], 3, 3, "outfile");
+            auto tensor_mat_rl = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat_rl->set(m,n,tensor_rl[i][m][n]);
+                }
+            }
```
robertodr, line 459: Same comment as above.

psi4/src/psi4/cc/ccresponse/optrot.cc
```diff
@@ -416,50 +432,80 @@ void optrot(std::shared_ptr<Molecule> molecule) {
...
             rotation_rl[i] = prefactor * TrG_rl * nu * nu / M;
             outfile->Printf("\n   Specific rotation using length-gauge electric-dipole Rosenfeld tensor.\n");
             outfile->Printf("\t[alpha]_(%5.3f) = %10.5f deg/[dm (g/cm^3)]\n", params.omega[i], rotation_rl[i]);
 
-            /* grab omega in nm, rounded to nearest int */
-            long om_nm = std::lround((pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]));
-
             if (params.wfn == "CC2") {
                 std::stringstream tag;
+                std::stringstream tag_au;
```
robertodr, line 470: Is this variable used?

psi4/src/psi4/cc/ccresponse/optrot.cc
```diff
@@ -468,34 +514,45 @@ void optrot(std::shared_ptr<Molecule> molecule) {
...
                             (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                             pc_hartree2wavenumbers * params.omega[i]);
             outfile->Printf("  -------------------------------------------------------------------------\n");
             mat_print(tensor_pl[i], 3, 3, "outfile");
+            auto tensor_mat_pl2 = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat_pl2->set(m,n,tensor_pl[i][m][n]);
+                }
+            }
```
robertodr, line 554: Same comment as above.

psi4/src/psi4/cc/ccresponse/polar.cc
```diff
@@ -141,12 +143,32 @@ void polar() {
...
             tag << "CC2 DIPOLE POLARIZABILITY @ " << omega_nm_rd << "NM";
             Process::environment.globals[tag.str()] = trace[i] / 3.0;
+            ref_wfn->set_scalar_variable(tag.str(),trace[i] / 3.0);
+            tag_tensor << "CC2 DIPOLE POLARIZABILITY TENSOR @ " << omega_nm_rd << "NM";
+            auto tensor_mat = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat->set(m,n,tensor[i][m][n]);
+                }
+            }
```
robertodr, line 156: Same comment as above.

psi4/src/psi4/cc/ccresponse/polar.cc
```diff
@@ -141,12 +143,32 @@ void polar() {
...
             tag << "CCSD DIPOLE POLARIZABILITY @ " << omega_nm_rd << "NM";
             Process::environment.globals[tag.str()] = trace[i] / 3.0;
+            ref_wfn->set_scalar_variable(tag.str(),trace[i] / 3.0);
+            tag_tensor << "CCSD DIPOLE POLARIZABILITY TENSOR @ " << omega_nm_rd << "NM";
+            auto tensor_mat = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat->set(m,n,tensor[i][m][n]);
+                }
+            }
```
robertodr, line 170: Same comment as above.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -139,15 +145,27 @@ void roa() {
...
         outfile->Printf("  -------------------------------------------------------------------------\n");
         outfile->Printf("   Evaluated at omega = 0.00 E_h (Inf nm, 0.0 eV, 0.0 cm-1)\n");
         outfile->Printf("  -------------------------------------------------------------------------\n");
         mat_print(tensor0, 3, 3, "outfile");
+        auto tensor_mat0 = std::make_shared<Matrix>(3, 3);
+        for (int m = 0; m < 3; m++) {
+            for (int n = 0; n < 3; n++) {
+                tensor_mat0->set(m,n,tensor0[m][n]);
+            }
+        }
```
robertodr, line 167: Same comment as above.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -426,72 +444,139 @@ void roa() {
...
                         (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                         pc_hartree2wavenumbers * params.omega[i]);
         outfile->Printf("  -------------------------------------------------------------------------\n");
         mat_print(tensor_rr[i], 3, 3, "outfile");
+        auto tensor_mat_rr = std::make_shared<Matrix>(3, 3);
+        for (int m = 0; m < 3; m++) {
+            for (int n = 0; n < 3; n++) {
+                tensor_mat_rr->set(m,n,tensor_rr[i][m][n]);
+            }
+        }
```
robertodr, line 468: Same comment as above.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -426,72 +444,139 @@ void roa() {
...
                             (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                             pc_hartree2wavenumbers * params.omega[i]);
             outfile->Printf("  -------------------------------------------------------------------------\n");
             mat_print(tensor_rl[i], 3, 3, "outfile");
+            auto tensor_mat_rl = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat_rl->set(m,n,tensor_rl[i][m][n]);
+                }
+            }
```
robertodr, line 493: Same comment as above.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -426,72 +444,139 @@ void roa() {
...
                             (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                             pc_hartree2wavenumbers * params.omega[i]);
             outfile->Printf("  -------------------------------------------------------------------------\n");
             mat_print(tensor_pl[i], 3, 3, "outfile");
+            auto tensor_mat_pl = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat_pl->set(m,n,tensor_pl[i][m][n]);
+                }
+            }
```
robertodr, line 519: Same comment as above.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -426,72 +444,139 @@ void roa() {
...
                             (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                             pc_hartree2wavenumbers * params.omega[i]);
             outfile->Printf("  -------------------------------------------------------------------------\n");
             mat_print(tensor_pl[i], 3, 3, "outfile");
+            auto tensor_mat_pl2 = std::make_shared<Matrix>(3, 3);
+            for (int m = 0; m < 3; m++) {
+                for (int n = 0; n < 3; n++) {
+                    tensor_mat_pl2->set(m,n,tensor_pl[i][m][n]);
+                }
+            }
```
robertodr, line 547: Same comment as above.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -426,72 +444,139 @@ void roa() {
...
+            tag_quad[1] << "CCSD QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ " << om_nm << "NM";
+            tag_quad[2] << "CCSD QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ " << om_nm << "NM";
+        }
         outfile->Printf("  -------------------------------------------------------------------------\n");
         outfile->Printf("   Evaluated at omega = %8.6f E_h (%6.2f nm, %5.3f eV, %8.2f cm-1)\n", params.omega[i],
                         (pc_c * pc_h * 1e9) / (pc_hartree2J * params.omega[i]), pc_hartree2ev * params.omega[i],
                         pc_hartree2wavenumbers * params.omega[i]);
         outfile->Printf("  -------------------------------------------------------------------------\n");
         for (alpha = 0; alpha < 3; alpha++) mat_print(tensor_rQ[i][alpha], 3, 3, "outfile");
+        auto quad_mat = new SharedMatrix[3];
```
robertodr, line 570: It's probably enough to do `auto tmp = std::make_shared<Matrix>(3, 3);` before filling the matrix. I don't think there's a need to use an array of `SharedMatrix`.

robertodr (approved):
I've left some style comments around.

psi4/src/psi4/cc/ccresponse/optrot.cc
bgpeyton: nope, leftover from storing in au. good catch!

psi4/src/psi4/cc/ccresponse/optrot.cc
```diff
@@ -181,15 +185,27 @@ void optrot(std::shared_ptr<Molecule> molecule) {
...
         outfile->Printf("  -------------------------------------------------------------------------\n");
         outfile->Printf("   Evaluated at omega = 0.00 E_h (Inf nm, 0.0 eV, 0.0 cm-1)\n");
         outfile->Printf("  -------------------------------------------------------------------------\n");
         mat_print(tensor0, 3, 3, "outfile");
+        auto tensor_mat0 = std::make_shared<Matrix>(3, 3);
+        for (int m = 0; m < 3; m++) {
+            for (int n = 0; n < 3; n++) {
+                tensor_mat0->set(m,n,tensor0[m][n]);
+            }
+        }
```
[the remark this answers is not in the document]
bgpeyton: didn't know this, this will clean things up considerably

bgpeyton:
thanks for the thorough review @robertodr ! all comments addressed. also removed the `cc` pytest marker, since it doesn't exist yet and we probably don't need a marker for a single test.

psi4/src/psi4/cc/ccresponse/roa.cc
```diff
@@ -139,15 +145,23 @@ void roa() {
...
-        if (params.wfn == "CC2")
+        std::stringstream tag_tensor0;
+        if (params.wfn == "CC2") {
             outfile->Printf("\n     CC2 Optical Rotation Tensor (Velocity Gauge): %s\n", lbl1);
-        else if (params.wfn == "CCSD")
+            tag_tensor0 << "CC2 OPTICAL ROTATION TENSOR (VEL) @ 0NM";
+        }
+        else if (params.wfn == "CCSD") {
             outfile->Printf("\n    CCSD Optical Rotation Tensor (Velocity Gauge): %s\n", lbl1);
+            tag_tensor0 << "CCSD OPTICAL ROTATION TENSOR (VEL) @ 0NM";
```
loriab, line 155: Just checking that this set twice: `CCSD OPTICAL ROTATION TENSOR (VEL) @ 0NM` https://github.com/psi4/psi4/pull/2022/files#diff-13ff4e13f97ab3afef025c04b5535b0d14561dfdf20e0c22960b5d1656a4394aR194

tests/pytests/test_ccresponse.py
```diff
@@ -0,0 +1,116 @@
+import numpy as np
```
loriab, line 1: be sure to add this file to the list of test_*.py files in <top-level>/psi4/CMakeLists.txt

psi4/driver/procrouting/proc.py
loriab, line 3011: Are the generic `CC DIPOLE` still getting set? Used anywhere in the tests?

loriab (commented):
Looks great, thanks for the tests!

psi4/src/psi4/cc/ccresponse/roa.cc
bgpeyton: yes, the structure of `roa.cc` isn't optimal in that it goes ahead and re-computes tensors you may already have. at some point we should split this up, but for now if you want to do all three properties it's easiest to just use `roa_tensor` and compute the rotation and polarizabilities from the tensors yourself.

psi4/driver/procrouting/proc.py
bgpeyton: Just caught this, it is used in ctest 46, and yes this is failing. I'll probably change `ccdensity.cc` to explicitly use the old `CC DIPOLE/QUADRUPOLE` for proc::env, as wrong as it feels. Generic `CC DIPOLE/QUADRUPOLE` are not being set on wfn.

bgpeyton: actually, this is done at in OEProp, e.g. for the dipole: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/oeprop.cc#L1226
OEProp is nicely set up to use a consistent naming structure. do I intentionally break this with an `if` statement for CC to preserve the old names, or let it use the full method name and update the only test that uses it? (which leaves the unfortunate issue of _response_ property core variables having generic `CC` names, while _OEProp_ properties have the full method title... might be okay to let this go if proc::env is not long for this world)

tests/pytests/test_ccresponse.py
bgpeyton: done, thanks for the reminder!

bgpeyton:
so I've changed the `cc46` test case to grab `EOM-CC2` variables for all ground-state quantities. Excited state quantities are set by the EOM code and so still follow `CC XPROPERTYX`. I think this is a reasonable move, considering anyone wanting to use these variables will need to check their names to get the syntax right in the first place.

loriab (approved):
lgtm, thanks! Is there any atomic unit for specific rotation?

bgpeyton:
> lgtm, thanks! Is there any atomic unit for specific rotation?

since specific rotation is normalized for path length and concentration in experimental settings, I'd say the use of deg/[dm (g/cm^3)] for specific rotation is even more ubiquitous than the use of NM for wavelength for dynamic properties. @lothian may have more to say on this topic.

lothian:
> > lgtm, thanks! Is there any atomic unit for specific rotation?
> 
> since specific rotation is normalized for path length and concentration in experimental settings, I'd say the use of deg/[dm (g/cm^3)] for specific rotation is even more ubiquitous than the use of NM for wavelength for dynamic properties. @lothian may have more to say on this topic.

Are we talking about storing the specific rotation or the G' tensor?  I would think we'd want to store the latter in atomic units (along with the field energy, also in a.u.), and not store the specific rotation, which can quickly be computed from the tensor and the energy.

bgpeyton:
> > > lgtm, thanks! Is there any atomic unit for specific rotation?
> > 
> > 
> > since specific rotation is normalized for path length and concentration in experimental settings, I'd say the use of deg/[dm (g/cm^3)] for specific rotation is even more ubiquitous than the use of NM for wavelength for dynamic properties. @lothian may have more to say on this topic.
> 
> Are we talking about storing the specific rotation or the G' tensor? I would think we'd want to store the latter in atomic units (along with the field energy, also in a.u.), and not store the specific rotation, which can quickly be computed from the tensor and the energy.

The G' tensor is stored in atomic units (as it's reported in the output file), while the specific rotation itself is stored in deg/[dm (g/cm^3)] since it is only one additional float. For now, since the syntax is `CCxx xPROPERTYx @ xxNM`, it's a bit easier to use NM. But you're right that the field strength in a.u. should be available at the python level to an arbitrary number of digits. perhaps the `omega` dict passed in (as `psi4.set_options({"omega":[x,x,unit]})` is available on the wfn already?

loriab:
>>> lgtm, thanks! Is there any atomic unit for specific rotation?

>> since specific rotation is normalized for path length and concentration in experimental settings, I'd say the use of deg/[dm (g/cm^3)] for specific rotation is even more ubiquitous than the use of NM for wavelength for dynamic properties. @lothian may have more to say on this topic.

> Are we talking about storing the specific rotation or the G' tensor? I would think we'd want to store the latter in atomic units (along with the field energy, also in a.u.), and not store the specific rotation, which can quickly be computed from the tensor and the energy.

Both tensors and scalars are saved to psivars in the PR. I think that's usual procedure as we try to be explicit (mp2 tot, mp2 corl, hf tot) even if the arithmetic is easy. We are trying to map psivars more closely to qcschema properties though, which are rigidly atomic units. Over there, it's easy to request the ubiquitous user units of deg/[dm (g/cm^3)] be returned from au storage. We can add such capability to the `psi4.variable()` query in psi4. So my question was whether we could store the qcvars data in au. The qcvar name with nm I think is a reasonable exception since that won't translate directly to qcschema.

bgpeyton:
> > > > lgtm, thanks! Is there any atomic unit for specific rotation?
> 
> > > since specific rotation is normalized for path length and concentration in experimental settings, I'd say the use of deg/[dm (g/cm^3)] for specific rotation is even more ubiquitous than the use of NM for wavelength for dynamic properties. @lothian may have more to say on this topic.
> 
> > Are we talking about storing the specific rotation or the G' tensor? I would think we'd want to store the latter in atomic units (along with the field energy, also in a.u.), and not store the specific rotation, which can quickly be computed from the tensor and the energy.
> 
> Both tensors and scalars are saved to psivars in the PR. I think that's usual procedure as we try to be explicit (mp2 tot, mp2 corl, hf tot) even if the arithmetic is easy. We are trying to map psivars more closely to qcschema properties though, which are rigidly atomic units. Over there, it's easy to request the ubiquitous user units of deg/[dm (g/cm^3)] be returned from au storage. We can add such capability to the `psi4.variable()` query in psi4. So my question was whether we could store the qcvars data in au. The qcvar name with nm I think is a reasonable exception since that won't translate directly to qcschema.

in that case, we could in theory store the specific rotation in "atomic units" if such a thing even exists. it's computed in `ccresponse` from the trace of the G' tensor (in a.u.) which is then normalized for path length, concentration, and molecular weight to give the "specific" rotation, so I suppose we'd need atomic units for those quantities.

bgpeyton:
@loriab after some consideration with @lothian , I think the simplest course of action would be to simply not store the scalar variable. the unit question is somewhat nontrivial, plus there is the issue of scaling the rotation by the molecular weight of the system (which is handled very differently for solvated molecules, but is not accounted for in `ccresponse`). therefore the best way to make sure the user knows what they're getting is to let them compute it themselves from the mixed electric dipole-magnetic dipole response tensor. 

following this logic, I also propose removing the scalar electric dipole polarizability, which is also easily computed using the trace of the stored tensor. I would update the new test to reflect this, just want to be sure this is an agreeable solution for everyone before I commit the changes.

loriab:
I'm not keen on not storing the scalars, but I can't argue with your logic. Perhaps keep the tests, and show the tensor->scalar processing within the tests.

bgpeyton:
done. pending auto tests, this should be ready to go.

jturney (approved):
Looks great!

PR #2027, opened 2020-11-03
solved memory bug in libtrans when bucketSize exceeds an int

psi-rking:
## Description
Fix memory bug in libtrans for large cases.

## Todos
- [X] bucketSize was overflowing int storage causing psio write error

## Status
- [X] Ready to merge

### DIFF
--- a/psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc
+++ b/psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc
@@ -40,6 +40,20 @@
 
 using namespace psi;
 
+
+size_t *init_size_t_array(int size) {
+    size_t *array;
+
+    if ((array = (size_t *)malloc(sizeof(size_t) * size)) == nullptr) {
+        outfile->Printf("init_size_t_array:  trouble allocating memory \n");
+        outfile->Printf("size = %d\n", size);
+        exit(PSI_RETURN_FAILURE);
+    }
+    memset(array, 0, sizeof(size_t) * size);
+    return (array);
+}
+
+
 /**
  * Presort the (restricted) MO TPDM into DPD buffers to prepare it
  * for the the transformation.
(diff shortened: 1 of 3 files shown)


psi-rking:
I successfully ran a large optimization (CC2, 322 bf) and runs fine now.

psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc
```diff
@@ -40,6 +40,20 @@
...
+    size_t *array;
+
+    if ((array = (size_t *)malloc(sizeof(size_t) * size)) == nullptr) {
+        outfile->Printf("init_size_t_array:  trouble allocating memory \n");
+        outfile->Printf("size = %d\n", size);
+        exit(PSI_RETURN_FAILURE);
+    }
+    memset(array, 0, sizeof(size_t) * size);
+    return (array);
+}
```
loriab, line 54: should this be in libciomr.h with its friends?

psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc
```diff
@@ -77,8 +91,11 @@ void IntegralTransform::presort_mo_tpdm_restricted() {
     bucketOffset[0] = init_int_array(nirreps_);
     int **bucketRowDim = (int **)malloc(sizeof(int *));
     bucketRowDim[0] = init_int_array(nirreps_);
-    int **bucketSize = (int **)malloc(sizeof(int *));
-    bucketSize[0] = init_int_array(nirreps_);
+    //int **bucketSize = (int **)malloc(sizeof(int *));
+    //bucketSize[0] = init_int_array(nirreps_);
```
loriab, line 95: for my part I like replace over comment and add

loriab (approved):
yay, a fix, thanks!

psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc
psi-rking: I generally agree. This habit is a relic of the pre-git days of code development.  However, in this particular case there are some other 'ints' in the function that could someday need converted to size_t (but not in the immediate future) so the relics of this change may make that future fix easier.

jturney (approved):

psi4/src/psi4/libtrans/integraltransform_sort_mo_tpdm.cc
psi-rking: Yeah, I'll do it.  Might be my first contribution to ye old libciomr.  Can't recall.  But don't ask me to change init_int_array()!  I don't want to know how many little ways it might be used.

andysim (approved):
Thanks for tracking down a very tricky bug - I'm sure it took a long time, given that it only kicks in at large system size.  The fix is good and is consistent with the previous implementation.  The more modern approach would be to use std::vector to avoid memory leaks, but I see that it's freed at the bottom, so all good in this case.   LGTM!

psi-rking:
@loriab , rebased yesterday to keep things as clean as possible.  Can you go ahead and do the merge now?

andysim:
It's in now; I would've done it yesterday, but the CI checks took a while to complete.  Thanks again!

PR #2028, opened 2020-11-05
pybind11 v2.6, cbs dedup, enable_xhost

loriab:
## Description
misc

## Todos

- [x] let plugins work with pybind11 v2.6.0
- [x] there was extra testing in the == of cbs() tasks when options was involved (e.g., delta(ae - fc)). but I don't understand it, and it was preventing deduplication. so fixed it
- [x] Susi noticed trouble with ENABLE_XHOST in that the two options directed at Intel and GCC are valid cross-options. this is a workaround

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -97,7 +97,7 @@ jobs:
         psi4/label/dev::snsmp2 \
         psi4/label/dev::fockci \
         psi4/label/dev::mp2d \
-        psi4/label/dev::pybind11-headers \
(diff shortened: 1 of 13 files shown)


robertodr (approved):

andysim (approved):
LGTM!

hokru (approved):

PeterKraus:
I'm reasonably sure there was a reason why that `"f_options" != False` check was necessary at some point, I just don't remember it right now. As long as all test pass, it should be fine, and I'm guessing DDD will make changes to the routines anyway.

PR #2032, opened 2020-11-13
Add Compute Electric Field in Memory

### LINKED ISSUE #2031: Compute Electric Field in Memory
Currently it seems that only the electrostatic potential can be computed directly in memory from a wavefunction (`ESPPropCalc::compute_esp_over_grid_in_memory`). For our use it case it would also be useful to have a corresponding function for computing the electric field in memory (e.g. `ESPPropCalc::compute_field_over_grid_in_memory`).

I'd be happy to open a PR for this if you would be happy to include this (I have a local version working).

SimonBoothroyd:
## Description
Closes #2031 

## Todos
- [X] Add `compute_field_over_grid_in_memory` to `oeprop.cc` and `oeprop.h`
- [X] Expose `compute_field_over_grid_in_memory` in `export_oeprop.cc`.
- [X] Update `mints14` test

## Questions
- N / A

## Checklist

## Status
- [X] Ready for review
- [x] Ready for merge


psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1158,6 +1158,56 @@ void ESPPropCalc::compute_field_over_grid(bool print_output) {
...
+    int const nbf = basisset_->nbf();
+
+    std::vector<SharedMatrix> intmats;
+    intmats.push_back(std::make_shared<Matrix>("Ex integrals", nbf, nbf));
+    intmats.push_back(std::make_shared<Matrix>("Ey integrals", nbf, nbf));
+    intmats.push_back(std::make_shared<Matrix>("Ez integrals", nbf, nbf));
+
+    bool convert = mol->units() == Molecule::Angstrom;
+
+    for (int i = 0; i < number_of_grid_points; ++i) {
```
maxscheurer, line 1193: Instead of looping over all grid points, you could adapt [this](https://github.com/psi4/psi4/blob/64915980a2f0e6ea2a65d3e2cf16b3ec4488fe0d/psi4/src/psi4/libmints/mintshelper.cc#L1719) method which computes electric field values at all given `coords` using a functor. Then the for-loop would just be needed to form the nuclear contribution of the electric field.

tests/python/mints14/input.py
```diff
@@ -71,3 +71,142 @@
...
+                 [ 9.79431484e-05, -1.09066464e-04, -5.63407299e-04],
+                 [ 2.94543076e-04,  9.13557097e-05, -4.90653651e-04],
+                 [ 3.99793079e-04,  2.27763976e-04, -3.13890974e-04],
+                 [ 4.03799551e-04,  2.74855411e-04, -1.29722537e-04],
+                 [ 3.48015799e-04,  2.59140058e-04,  2.65049640e-06],
+                 ]
+
+# Comparison
+psi4.compare_arrays(efs_1threads, efs_4threads, 3, "Reference value for EF calculation, 1 thread vs. 4 threads.")
+psi4.compare_arrays(efs_1threads, reference_efs, 3, "Reference value for EF calculation, 1 thread vs. reference calculation.")
```
loriab, line 212: Could the test tolerance be tightened up beyond 1.e-3? I worry that much of the reference array is smaller.

Thanks for this new functionality!

susilehtola: 1e-3 is a pretty significant error for an observable! I would thing this should be accurate to 1e-9 or 1e-10, since the integrals are analytic!

psi4/src/psi4/libmints/oeprop.cc
SimonBoothroyd: Thanks for the suggestion! This should now be changed in 3850718.

tests/python/mints14/input.py
SimonBoothroyd: Ah good call - I've dropped the tolerance down in 3850718 to 1e-9 and the test still seems to pass ok.

SimonBoothroyd:
Thanks for the feedback @maxscheurer , @loriab and @susilehtola ! I've tried to make the suggested changes in 3850718.

loriab (approved):
lgtm, thank you!

robertodr (approved):

jturney (approved):
Looks good !

PR #1923, opened 2020-06-06
DFGrad II: Comment and Correct `dfmp2`

JonathonMisiewicz:
## Description
This PR closes #1916 and corrects a bug causing third-decimal errors in gradients computed by `dfmp2` with the frozen virtual approximation. See the commit marked "Bugfix!" to see exactly what the issue was. There was a sign error, and Rob was looping over frozen core orbitals when he meant to loop over frozen virtual orbitals.

More importantly for me, this code inserts many, many comments into Rob's code, so I know what I'm doing when I modify it throughout the rest of this PR series. This includes uncovering a very devious computational trick, that as far as I know is unpublished, and also some errors in the original papers. Many thanks to Francesco and Shuhe for the DSRG-PT2 paper, which formed part of the Rosetta Stone for figuring out how this code works.

Several explicit types are now `auto`. Reviewers, double-check that I didn't accidentally remove a pass-by-reference or `const` declaration when doing so.

Lastly, `test_gradients.py` has been added to pytests to confirm the accuracy of `dfmp2` gradients. This PR series will culminate with the addition of DF-ODC-12 to `test_gradients.py`. Currently, DF-ODC-12 would fail due to #1579.

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] Corrects a critical bug causing errors in frozen virtual gradients computed with the `dfmp2` module
- [x] Adds many comments to `dfmp2`
- [x] Replaces a lot of explicit type with `auto`

## Checklist
- [x] Tests added for newly working features
- [x] `dfmp2` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2033, opened 2020-11-13
Fix bug in large dfmp2

JonathonMisiewicz:
Fixes a bug reported by @hokru, where PSIO errors would result for large DFMP2 gradient computations. I forgot to rename a PSIO entry during #1923. 

This bug wasn't any more serious because it would only trigger when a block-based algorithm requested multiple blocks. For DFMP2, this would be a large computation indeed.

## Status
- [x] Ready for review
- [x] Ready for merge


andysim (approved):
LGTM.  Thanks for figuring this out so quickly!

jturney (approved):

loriab (approved):
Glad you know the code so well -- makes these bugs look painless!

PR #2035, opened 2020-11-16
Fix MBIS docs and multipole shapes

zachglick:
## Description
This PR updates the MBIS documentation to match the state of the code. Missing logic to reshape the mbis multipoles was also added, which previous tests don't check for, but is necessary when running psi4's mbis through qcengine.

## Todos
- [x] Fixed docs
- [x] Fixed / cleaned up the handling of multipole matrix shapes

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -591,7 +591,7 @@ def basis_helper(block, name='', key='BASIS', set_option=True):
 
 core.OEProp.valid_methods = [
     'DIPOLE', 'QUADRUPOLE', 'MULLIKEN_CHARGES', 'LOWDIN_CHARGES', 'WIBERG_LOWDIN_INDICES', 'MAYER_INDICES',
-    'MAYER_INDICES', 'MO_EXTENTS', 'GRID_FIELD', 'GRID_ESP', 'ESP_AT_NUCLEI', 'NO_OCCUPATIONS'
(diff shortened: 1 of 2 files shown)


psi4/driver/p4util/python_helpers.py
```diff
@@ -661,18 +676,20 @@ def _qcvar_reshape_get(key, val):
...
+            reshaper = (-1, 3)
+            return val.np.reshape(reshaper)
+        elif key.upper().endswith("QUADRUPOLES"):
+            val = val.np.reshape(-1, 6)
+            val = np.array([_multipole_plumper(val[iat], 2) for iat in range(len(val))])
+            return val
+        elif key.upper().endswith("OCTUPOLES"):
+            val = val.np.reshape(-1, 10)
+            val = np.array([_multipole_plumper(val[iat], 3) for iat in range(len(val))])
+            return val
```
loriab, line 692: is mirroring the other structure by separating into two-level if-mbis/if-n-pole the only change here?

loriab (approved):
good catch -- I forgot we weren't testing roundtrip

psi4/driver/p4util/python_helpers.py
zachglick: yep the logic is unchanged here, just reorganized to be a bit more maintainable

PeterKraus (approved):

robertodr (approved):

Issue #1987, opened 2020-08-25
QCEngine wavefunction error


jthorton:
When requesting the wavefunction as part of the result via the protocol in QCEngine only the `orbitals_and_eigenvalues` option works successfully. I think this is due to an error in the returned dictionary here https://github.com/psi4/psi4/blob/5de566ed7d4bf51816c673c4ea1def449e4e8618/psi4/driver/schema_wrapper.py#L324.I think this line should say `"density_a": "scf_density_a",`.

QCEngine error:
```
QCEngine Unknown Error: Traceback (most recent call last):
  File "/Users/joshuahorton/miniconda3/envs/qcfractal/lib//python3.7/site-packages/psi4/driver/schema_wrapper.py", line 400, in run_qcschema
    ret = qcel.models.Result(**ret_data, stdout=_read_output(outfile))
  File "/Users/joshuahorton/miniconda3/envs/qcfractal/lib/python3.7/site-packages/qcelemental/models/results.py", line 520, in __init__
    super().__init__(*args, **kwargs)
  File "pydantic/main.py", line 344, in pydantic.main.BaseModel.__init__
  File "pydantic/main.py", line 900, in pydantic.main.validate_model
  File "pydantic/fields.py", line 556, in pydantic.fields.ModelField.validate
  File "pydantic/fields.py", line 732, in pydantic.fields.ModelField._apply_validators
  File "pydantic/class_validators.py", line 280, in pydantic.class_validators._generic_validator_cls.lambda4
  File "/Users/joshuahorton/miniconda3/envs/qcfractal/lib/python3.7/site-packages/qcelemental/models/results.py", line 463, in _wavefunction_protocol
    ret_wfn[key] = wfn[key]
KeyError: 'scf_density_ba'
```

hokru:
I'm not very familiar with the schema_wrapper but that does look like a typo. Thanks!

jthorton:
Thanks for fixing that one! I have found another however which comes up when I request to save all wavefunction properties via QCEngine using the latest dev version of psi4, here is the error message.
```
QCEngine Input Error: Traceback (most recent call last):
File "/Users/joshua/miniconda3/envs/psi4/lib//python3.8/site-packages/psi4/driver/schema_wrapper.py", line 410, in run_qcschema
ret = qcel.models.AtomicResult(**ret_data, stdout=_read_output(outfile))
[...]

PR #2036, opened 2020-11-18
add orbital occupations to qcschema wavefunction

zachglick:
## Description
SCF orbital occupations are a required field for the wavefunction qcschema. This PR makes sure Psi4 populates the field. Addresses #1987 

## Todos
- [x] Add orbital occupations

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -313,6 +313,16 @@ def _convert_wavefunction(wfn, context=None):
 
         return arr
 
+    # get occupations in orbital-energy ordering
+    def sort_occs(noccpi, epsilon):
+        occs = []
+        for irrep, nocc in enumerate(noccpi):
+            for i, e in enumerate(epsilon[irrep]):
+                occs.append((e, int(i < nocc)))
+
+        occs.sort(key = lambda x : x[0])
+        return np.array([occ[1] for occ in occs])
+
     # Map back out what we can
     ret = {
         "basis": basis,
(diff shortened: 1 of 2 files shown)


psi4/driver/schema_wrapper.py
```diff
@@ -341,8 +351,8 @@ def re1d(mat):
         "scf_fock_b": re2d(wfn.Fa_subset("AO")),
         "scf_eigenvalues_a": re1d(wfn.epsilon_a_subset("AO", "ALL")),
         "scf_eigenvalues_b": re1d(wfn.epsilon_b_subset("AO", "ALL")),
-        # "scf_occupations_a": np.hstack(wfn.occupation_a().nph),
-        # "scf_occupations_b": np.hstack(wfn.occupation_b().nph),
+        "scf_occupations_a": re1d(sort_occs((wfn.doccpi() + wfn.soccpi()).to_tuple(), wfn.epsilon_a().nph)),
+        "scf_occupations_b": re1d(sort_occs(wfn.doccpi().to_tuple(), wfn.epsilon_b().nph)),
```
loriab, line 355: background in case useful to others:

occupation_a/b was the only one of these that was SCF-only. and the current doccpi/soccpi will also be SCF-only. there is an occupations subset function https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.cc#L847 but it looks not so developed or parallel compared to all the other nice _subset fns. this may need to be developed to access NO occupations as qcschema.wfn expands.

psi4/driver/schema_wrapper.py
```diff
@@ -313,6 +313,16 @@ def re1d(mat):
...
 
+    # get occupations in orbital-energy ordering
+    def sort_occs(noccpi, epsilon):
+        occs = []
+        for irrep, nocc in enumerate(noccpi):
+            for i, e in enumerate(epsilon[irrep]):
+                occs.append((e, int(i < nocc)))
+
+        occs.sort(key = lambda x : x[0])
+        return np.array([occ[1] for occ in occs])
```
loriab, line 324: you tested this with a high-symmetry (e.g., c2v) molecule?

loriab (approved):
lgtm, thanks for investigating!

psi4/driver/schema_wrapper.py
dgasmith: Agreed, we need a test here to make sure say `water` works and the occupations match the water eigenvalue ordering.

zachglick: Yeah, I manually validated that the occupations lined up with the orbital energies for c2v water, for both ground state and arbitrary docc settings. I can put that in a test

jturney (approved):

dgasmith (approved):
Thanks for the tests!

PR #2037, opened 2020-11-19
auto apply fix_com when needed, qcschema smoke tests

### LINKED ISSUE #1939: PCM fails with symmetry group coordinates, need to add error message for this to help user
Hello,

I realize that there is a warning about PCM failing with symmetry point groups however, it would be nice to add like a 1-3 line error message that would go a long way to helping out any user that has not read all of the docs very carefully.

### LINKED ISSUE #1978: SCF C1 symmetry requirements not caught, e.g. for addons
Running e.g. `pcmsolver` calculations with symmetry will lead to a crash:
```
  File "/Users/kruse/qc/psi4/objdir/stage/lib/psi4/driver/procrouting/scf_proc/scf_iterator.py", line 300, in scf_iterate
    upcm, Vpcm = self.get_PCM().compute_PCM_terms(Dt, calc_type)

RuntimeError:
Fatal Error: Matrix::gemm error: Number of irreps do not equal.
Error occurred in file: /Users/kruse/qc/psi4/psi4/src/psi4/libmints/matrix.cc on line: 1349
The most recent 3 function calls were:
```

The PCM message suggests c1 symmetry is enforced:
https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L1520-L1522
but the `use_c1` option is for making post-SCF c1 wavefunction and not for the SCF itself.
Similar issue with the `cppe` addon for polarizable embedding.

Either a graceful exist or actually making a c1 wfn object is desirable.

### LINKED ISSUE #1986: Allow PCM to be set via schema
Currently, PCM settings can not be set via QCEngine using the schema interface, passing a task like this from QCEngine results in an error 
```python
{'schema_name': 'qcschema_input',
 'schema_version': 1,
 'molecule': Molecule(name='H2O', formula='H2O', hash='941bd29'),
 'driver': 'energy',
 'model': {'method': 'SCF', 'basis': 'sto-3g'},
 'protocols': {'wavefunction': 'orbitals_and_eigenvalues'},
 'keywords': {'scf_type': 'df',
  'pcm_scf_type': 'total',
  'pcm': {'Units': 'Angstrom',
   'Medium': {'SolverType': 'IEFPCM', 'Solvent': 'Water'},
   'Cavity': {'RadiiSet': 'UFF',
    'Type': 'GePol',
    'Scaling': False,
    'Area': 0.3,
    'Mode': 'Implicit'}}}}
```
as it tries to pass the dictionary of settings to the `core.set_global` function. 

Extending the `set_options` function [here](https://github.com/psi4/psi4/blob/5de566ed7d4bf51816c673c4ea1def449e4e8618/psi4/driver/p4util/python_helpers.py#L423) to handle the `pcm` keyword should fix this issue.

tagging @loriab following the slack discussion.

### LINKED ISSUE #2011: PCM fails if symmetry is not fixed in schema.
When running psi4 via QCEngine, PCM can only successfully run if the symmetry is fixed by including the flag `fix_symmetry="c1"` in the qcelemental molecule. As this is always required the Psi4 driver should probably fix the symmetry when it comes across any PCM settings in the keywords. This would also users to run PCM calculations on molecules which were submitted to a QCArchive instance previously without the fixed symmetry flag.

loriab:
## Description
Repercussions from trying to fix #1986 . I think this fixes #1939, fixes #1978, fixes #2011

## Todos
- [x] standardize pytests on `@using("<addon>")` rather than `@using_<addon>`. addon detection is more runtime than compiletime
- [x] start a smoke tests suite of running through `psi4 --qcschema` to live alongside psithon and psiapi
- [x] finally fix the need for pe/pcm to manually set fix_com/fix_reorient. I have a strategy locally.
- [ ] expand `Mol._initial_cartesian` to other ways to construct a Mol
- [x] snsmp2 1.0.4
- [x] pybind11 2.6.1
- [x] pcmsolver ~1.3.?~ 1.3 having some numerical issues with psi, so going back to a patched v1.2.1

## Questions
- [ ] Opening the discussion to how to deal with methods that needs no_com/no_reorient/symmetry c1 _from_the_beginning_. Many of our present symmetry tolerant methods just need their parts consistent (e.g., dimer, monoA, monoB); whereas, cppe needs the exact input Cart to match the potfile. (Fun fact: a Mol can go from fix_com T -> F but not the reverse.) The sol'n in this PR is to carry the input Cartesians along as `Mol._initial_cartesians` (so long as not a variable or zmat Mol), which can then be used to reset through `set_geometry()` if the method demands it. This member is akin to the `Mol.EFP` object that gets carried along. Presently this is implemented for just the molecule constructors that psithon/psiapi `psi4.geometry()` and qcsk `Mol.from_schema()` use, but should probably be expanded if adopted. Thoughts on whether this is the best way to solve the problem?
- [ ] @robertodr PCMSolver: I'm only imposing c1, not fix_com/orient based on your slack reply. ok? I've removed all those lines from your ctest and pytest examples (except for alpha that needed it for matching). There's a new pcm mode test in test_addons_qcschema.py. It's writing the input file for pcm input, which is a little questionable in the schema-running world, but it's working for now. Any concerns?
- [ ] @maxscheurer CPPE: when active, this is getting the full c1/fix_com/fix_orient treatment, so I've removed all those lines from the tests. There's a new cppe mode test in test_addons_qcschema.py. Since there isn't a mechanism for passing files through, I'm allowing the potfile keyword to be file name or contents. Any concerns?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


loriab:
needs scrutiny by @robertodr and @maxscheurer. I also learned about problems in the course of this that @hokru had already identified.

psi4/driver/procrouting/proc.py
```diff
@@ -1272,6 +1272,29 @@ def scf_helper(name, post_scf=True, **kwargs):
...
+        c1_molecule = scf_molecule.clone()
+        if getattr(scf_molecule, "_initial_cartesian", None) is not None:
+            c1_molecule._initial_cartesian = scf_molecule._initial_cartesian.clone()
+            c1_molecule.set_geometry(c1_molecule._initial_cartesian)
+            c1_molecule.reset_point_group('c1')
+            c1_molecule.fix_orientation(True)
+            c1_molecule.fix_com(True)
+            c1_molecule.update_geometry()
+        else:
+            raise ValidationError("Set no_com/no_reorient/symmetry c1 by hand for PCM or PE on non-Cartesian molecules.")
```
PeterKraus, line 1294: The validation error here is PE-specific but mentions PCM too.

psi4/driver/procrouting/solvent/pol_embed.py
```diff
@@ -42,8 +44,20 @@ def get_pe_options():
     rmin = core.get_option('PE', 'BORDER_RMIN')
     if core.get_option('PE', 'BORDER_RMIN_UNIT').upper() == "AA":
         rmin *= 1.0 / constants.bohr2angstroms
+
+    # potfile option can be filename or contents
+    potfile_keyword = core.get_option('PE', 'POTFILE')
+    if len(potfile_keyword) > 300:
```
PeterKraus, line 50: 300?

psi4/driver/procrouting/solvent/pol_embed.py
```diff
@@ -42,8 +44,20 @@ def get_pe_options():
...
+        core.get_option('PE', 'POTFILE')
```
PeterKraus, line 51: Why is this here? We just called that above.

PeterKraus (commented):
Looks great, especially the automatic enforcement of known symmetry/com/coord requirements is super user friendly. A couple of questions though.

psi4/driver/procrouting/proc.py
```diff
@@ -1272,6 +1272,29 @@ def scf_helper(name, post_scf=True, **kwargs):
     if ref_wfn is not None:
         raise ValidationError("Cannot seed an SCF calculation with a reference wavefunction ('ref_wfn' kwarg).")
 
+    # PCM needs to be run w/o symmetry
```
robertodr, line 1275: Could you add a warning or print message that these options have been set under the hood?

robertodr (approved):
Looks good! Thanks Lori!

maxscheurer (approved):
Thanks for this important and user-friendly improvement! 🎉 👍

psi4/driver/procrouting/solvent/pol_embed.py
loriab: it was an attempt to differentiate between a long file path and file contents. I'm switching to testing on `@COORDINATES`.

loriab:
This needs https://github.com/PCMSolver/pcmsolver/pull/184

@robertodr, I switched out some uuid calls for `NamedTemporaryFile`. Seems to work ok, but lmk if bad idea.

robertodr:
Thanks Lori, will check your PR ASAP and try to get a new release.

psi4/driver/molutil.py
```diff
@@ -252,6 +256,11 @@ def geometry(geom, name="default"):
         geom, enable_qm=True, missing_enabled_return_qm='minimal', enable_efp=True, missing_enabled_return_efp='none')
 
     molecule = core.Molecule.from_dict(molrec['qm'])
+    if "geom" in molrec["qm"]:
+        geom = np.array(molrec["qm"]["geom"]).reshape((-1, 3))
+        if molrec["qm"]["units"] == "Angstrom":
```
dgasmith, line 261: `.title() == "Angstrom"` to be safe or lower it?

psi4/driver/p4util/procutil.py
```diff
@@ -426,6 +427,75 @@ def prepare_options_for_modules(changedOnly=False, commandsInsteadDict=False):
...
+            # some options are globals section (not level) so not in any module
+            flat_options[opt] = core.get_global_option(opt)
+
+    return flat_options
+
+
+def state_to_atomicinput(*, driver, method, basis=None, molecule=None, function_kwargs=None) -> "AtomicInput":
+    """Form a QCSchema for job input from the current state of Psi4 settings."""
+
+    import qcelemental as qcel
```
dgasmith, line 469: Any need for a local import? `qcel` is required throughout I believe.

tests/pytests/test_addons.py
```diff
@@ -779,24 +730,21 @@ def _test_scf5():
     assert psi4.compare_values(Eref_rohf_df, E, 6, 'Triplet DF CUHF energy')
 
 
-@pytest.mark.smoke
-@using_erd
+@using("erd")
```
dgasmith, line 733: Glad you find this design pattern useful.

dgasmith (commented):
Very clean from a user point of view!

psi4/driver/molutil.py
loriab: thanks for the catch, but this is psi4 schema, and it's strictly `Angstrom|Bohr`

PeterKraus (approved):
LGTM.

loriab:
this needs a new pcmsolver release and a merge of https://github.com/DEShawResearch/sns-mp2/pull/7

robertodr:
> this needs a new pcmsolver release and a merge of [DEShawResearch/sns-mp2#7](https://github.com/DEShawResearch/sns-mp2/pull/7)

The PCMSolver release will most likely happen this evening.

robertodr:
New release of PCMSolver is up: https://github.com/PCMSolver/pcmsolver/releases/tag/v1.3.0

psi4/driver/molutil.py
dgasmith: I always get a little nervous around these since the data is “just” a dict and there is no code-ensured contract for title case.

dgasmith (approved):

loriab:
This has compatible pcmsolver and snsmp2 versions now, so I'll merge it after CI unless there are concerns.

There's still a problem sometimes running pcm by API and in parallel, `npz_save: header read error`, but this PR is still a step in the right direction, as one couldn't run sequential pcm by API before.

PR #2040, opened 2020-11-24
Directdfjk

obrien951:
## Todos

- [x] On-the-fly calculation of DF-ERI tensors for Fock builds
- [x] Works for RHF cases

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/DirectDFJK.cc
+++ b/psi4/src/psi4/libfock/DirectDFJK.cc
@@ -1,0 +1,1600 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2019 The Psi4 Developers.
+ *
+ * The copyrights for code used from other parties are included in
+ * the corresponding files.
+ *
+ * This file is part of Psi4.
+ *
+ * Psi4 is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, version 3.
+ *
+ * Psi4 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+#include "psi4/libfock/jk.h"
+
+#include "psi4/libmints/siminteri.h"
+#include "psi4/lib3index/3index.h"
+#include "psi4/lib3index/dftensor.h"
+#include "psi4/libmints/basisset.h"
+#include "psi4/libmints/integral.h"
+#include "psi4/libmints/molecule.h"
+#include "psi4/libmints/twobody.h"
+#include "psi4/libmints/matrix.h"
+#include "psi4/pragma.h"
+#include "psi4/libqt/qt.h"
+
+#include <cstdlib>
+#include <memory>
+#include <vector>
+#include <algorithm>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+#include "psi4/libpsio/psio.hpp"
+#include "psi4/libpsio/psio.h"
+#include "psi4/libpsio/aiohandler.h"
+#include <sstream>
+#include "psi4/libpsi4util/PsiOutStream.h"
+
+
+
+namespace psi {
+
+DirectDFJK::DirectDFJK(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary )
+    : JK(primary), auxiliary_(auxiliary)  {
+// Not much happens in common init, just that the number of elements in the
+//  basis sets gets saved as private member data in the DirectDFJK object.
+//  However, the JK() constructor gets called as well. Not to fear!
+//  In our case, all that does is set a couple booleans. 
+    common_init();
+}
+
+DirectDFJK::~DirectDFJK() {}
+
+//It's hard to say what does and doesn't belong in this function. I am
+// open to suggestions as to what should change. JSOB
+void DirectDFJK::common_init(){
+    nbf_ = primary_->nbf();
+    naux_ = auxiliary_->nbf();
+
+	p_shells_ = primary_->nshell();
+	Q_shells_ = auxiliary_->nshell();
+}
+
+//Used as testing machinery. It's nice to have a public function
+//to be able to look at private member data from pyside.
+//I plan on nuking this as soon as I get this project done,
+//don't you worry. JSOB
+
+//syntax copied and adapted from DiskDFJK.
+//I decided to forgo copying from MemDFJK because
+//that code seems more specialized. 
+void DirectDFJK::print_header() const {
+	if (print_)	{
+		outfile->Printf("==> DirectDFJK: Density-Fitted J/K Matrices <==\n\n");
+
+        outfile->Printf("    J tasked:          %11s\n", (do_J_ ? "Yes" : "No"));
+        outfile->Printf("    K tasked:          %11s\n", (do_K_ ? "Yes" : "No"));
+		outfile->Printf("    wK tasked:         %11s\n", (do_wK_ ? "Yes" : "No"));
+		if (do_wK_) outfile->Printf("Don't interpret these results.\nWe don't have w integrals in DirectDFJK\n");
+		outfile->Printf("    OpenMP threads:    %11d\n", omp_nthread_);
+		outfile->Printf("    Integrals threads: %11d\n", df_ints_num_threads_);
+		outfile->Printf("    Memory [MiB]:      %11ld\n", (memory_ * 8L) / (1024L*1024L));
+		outfile->Printf("    Algorithm:         %11s\n", "Direct");
+        outfile->Printf("    Schwarz Cutoff:    %11.0E\n", cutoff_);
+        outfile->Printf("    Fitting Condition: %11.0E\n\n", condition_);
+
+        outfile->Printf("   => Auxiliary Basis Set <=\n\n");
+        auxiliary_->print_by_level("outfile", print_);
+	}
+}
+
+// In the long term, I'm not sure what belongs here and what
+// belongs in common_init() JSOB
+
+// I think this should call the memory estimate and determine the number of blocks.
+// Memory will be allocated locally by the compute_JK() method.
+
+void DirectDFJK::preiterations() {
+
+    //cutoff_ = 1e-12;
+	// set_df_ints_num_threads(omp_nthread_);
+	sparsity_prep_pQq();
+
+	//prepare blocks tells us at which indices we start and stop each block.
+	// I'm making it a separate function because I've seen evidence that 
+	// DFHelper does some resizing, and I want to be able to match that.
+//dirty
+	biggest_shell_ = 0;
+	for (int i = 0; i < primary_->nshell(); i++){
+		if (primary_->shell(i).nfunction() > biggest_shell_) {
+			biggest_shell_ = primary_->shell(i).nfunction();
+		}
+	}
+//end dirty
+
+	// prepares the coulomb metric. Of course, our calculations depend on the
+	//  inverse of the coulomb metric and the (-0.5) power of the metric, so
+	//  we will prepare those immediately afterwards
+	//prepare_metric();
+	get_met();
+	
+	prepare_metric_power(-0.5);
+
+	//for what it's worth, we should have the density matrix
+	//prepare_D_symm();
+
+}
+
+void DirectDFJK::sparsity_prep_pQq(){
+	int procs = 1;
+#ifdef _OPENMP
+	procs = omp_nthread_;
+#endif
+
+	// variables to hold data
+	double global_max_int = 0.0;
+
+
+	std::unique_ptr<double[]> shel_maxes(new double[p_shells_*p_shells_]);
+	std::unique_ptr<double[]> integrals(new double[nbf_*nbf_]);
+
+
+	for (size_t shel_sq_iter = 0; shel_sq_iter < p_shells_*p_shells_; shel_sq_iter++) shel_maxes[shel_sq_iter] = 0.0;
+
+
+	for (size_t func_sq_iter = 0; func_sq_iter < nbf_*nbf_; func_sq_iter++) integrals[func_sq_iter] = 0.0;
+
+	// AO machinery
+	auto rifactory = std::make_shared<IntegralFactory>(primary_, primary_, primary_, primary_);
+
+	std::vector<std::shared_ptr<TwoBodyAOInt>> eri(procs);
+	std::vector<const double*> buffer(procs);
+
+
+#pragma omp parallel num_threads(procs)
+    {
+    int rank = 0;
+#ifdef _OPENMP
+    rank = omp_get_thread_num();
+#endif
+    eri[rank] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
+	buffer[rank] = eri[rank]->buffer();
+    }
+
+	// Sparsity member data
+	schwarz_func_ints_.resize(nbf_);
+
+	std::vector<size_t> dense_its(nbf_);
+
+	//AO shell metadata
+	size_t Shell_M_count;
+	size_t Shell_M_start;
+	size_t Shell_N_count;
+	size_t Shell_N_start;
+	double val = 0.0;
+
+#pragma omp parallel for schedule(guided) num_threads(procs) private( Shell_M_count, Shell_M_start, Shell_N_count, Shell_N_start, val) reduction(max:global_max_int)
+	for(size_t Shell_M = 0; Shell_M < p_shells_; Shell_M++) {
+		int rank = 0;
+#ifdef _OPENMP
+    	rank = omp_get_thread_num();
+#endif
+        Shell_M_start = primary_->shell(Shell_M).function_index();
+		Shell_M_count = primary_->shell(Shell_M).nfunction();
+		for (size_t Shell_N = 0; Shell_N < p_shells_; Shell_N++) {
+	        Shell_N_start = primary_->shell(Shell_N).function_index();
+			Shell_N_count = primary_->shell(Shell_N).nfunction();
+			eri[rank]->compute_shell(Shell_M, Shell_N, Shell_M, Shell_N);
+			for (size_t func_m = 0; func_m < Shell_M_count; func_m++){
+				for (size_t func_n = 0; func_n < Shell_N_count; func_n++){
+					integrals[ (Shell_M_start + func_m) * nbf_ + Shell_N_start + func_n ]
+					=
+//					integrals[ (Shell_N_start + func_n) * nbf_ + Shell_M_start + func_m ]
+//					=
+					val 
+					= 
+ fabs(buffer[rank][ func_m * Shell_N_count * Shell_M_count * Shell_N_count + 
+				    func_n * Shell_M_count * Shell_N_count + 
+				    func_m * Shell_N_count + 
+				    func_n]);
+					if (global_max_int < val) global_max_int = val;
+					
+					if (shel_maxes[Shell_M * p_shells_ + Shell_N] < val) {
+						shel_maxes[Shell_M * p_shells_ + Shell_N]
+//						=
+//						shel_maxes[Shell_N * p_shells_ + Shell_M]
+						= val;
+					}
+				}
+			}
+		}
+	}
+
+	tolerance_ = cutoff_ * cutoff_ / global_max_int;
+	schwarz_shell_mask_pQq_.resize(p_shells_);
+	schwarz_func_starts_pQq_.resize(p_shells_);
+	schwarz_dense_funcs_.resize(p_shells_);
+	size_t count;
+
+	for ( size_t shell_it_out = 0U; shell_it_out < p_shells_; shell_it_out++ ) {
+		count = 0U;
+		for (size_t shell_it_in = 0U; shell_it_in < p_shells_; shell_it_in++ ) {
+			if(shel_maxes[shell_it_out*p_shells_ + shell_it_in] >  tolerance_) {
+				schwarz_shell_mask_pQq_[shell_it_out].push_back(shell_it_in);
+				schwarz_func_starts_pQq_[shell_it_out].push_back(count);
+				count += primary_->shell(shell_it_in).nfunction();
+			}
+		}
+		schwarz_dense_funcs_[shell_it_out] = count;
+	}
+}
+
+// calculates a block of AO's then conducts the proper contractions to get them
+//   where they need to go. I don't think it makes sense to have seperate J and
+//   K function calls as J and K are never fully built in a single function.
+void DirectDFJK::compute_JK() {
+	BB_ = false;
+
+	if ( pQq_ ) {
+        if (ao_sparse_) {
+		    pQp_sparse();
+        } else {
+    	    pQp();
+        }
+
+	}  else {
+
+		if (num_blocks_ == 1) {
+			build_jk_CC_Qpq_direct();
+		} else {
+    		build_jk_CC_Qpq_blocks();
+		}
+
+	}
+    n_col_last_ = C_left_ao_[0]->ncol();
+}
+
+void DirectDFJK::postiterations() {
+}
+
+// Determines the maximum amount of memory that DirectDFJK will be using
+//   at any one time. To be sure, the memory used to construct AO blocks
+//   will only be allocated during the compute_JK() method, so it will
+//   be available during other procedures. 
+// Uses blocking and sparsity information to set total_needs which is 
+//   returned by memory_estimate.
+void DirectDFJK::our_needs(){
+	free_memory_ = memory_;
+	size_t blocks;
+
+	size_t our_needs = 0;
+	double charges_f = 0.0;
+	size_t charges_z;
+	std::shared_ptr<Molecule> mol_ptr = primary_->molecule();
+	
+	for (int atom_iter = 0; atom_iter < mol_ptr->natom(); atom_iter++)
+		charges_f+=mol_ptr->fZ(atom_iter);
+
+	charges_f = charges_f/2.0;
+
+	charges_z = static_cast<size_t>(charges_f);
+	++charges_z;
+
+	//D
+	our_needs += nbf_*nbf_;
+	free_memory_-=nbf_*nbf_;
+
+	//C_a
+	our_needs += nbf_*charges_z;
+	free_memory_-=nbf_*charges_z;
+	//C_b
+	our_needs += nbf_*charges_z;
+	free_memory_-=nbf_*charges_z;
+	//K
+   	our_needs += nbf_*nbf_;	
+	free_memory_-=nbf_*nbf_;
+   	//J
+   	our_needs += nbf_*nbf_;
+	free_memory_-=nbf_*nbf_;
+	//[ J^{-1} ]
+	our_needs += naux_*naux_;
+	free_memory_-=naux_*naux_;
+	//[ J^{-\frac{1}{2} ]
+	our_needs += naux_*naux_;
+	free_memory_-=naux_*naux_;
+	blocks = (2* (nbf_ * nbf_ *naux_))/free_memory_ +1;
+	ABX_block_size_ = ((naux_/blocks)+1)*nbf_*nbf_;
+
+    total_needs_ = our_needs;// + 2*ABX_block_size_;
+}
+size_t DirectDFJK::memory_estimate(){
+
+	if (Qpq_) {
+		prepare_Q_blocks();
+	}
+
+	our_needs();
+	return total_needs_;
+}
+
+void DirectDFJK::prepare_metric_power(double power){
+	SharedMatrix my_met;
+	auto coul_met = std::make_shared<FittingMetric>(auxiliary_, true);
+	coul_met->form_fitting_metric();
+	my_met = coul_met->get_metric();
+	if ( !(1e-13 > fabs(1.0 - power)) ){
+		my_met->power(power, condition_);
+	}
+	
+	met_cols_.push_back(my_met->ncol());
+	met_rows_.push_back(my_met->nrow());
+	metric_.push_back(my_met);
+	met_powers_.push_back(power);
+
+    while (power < 0.0) {
+        power = power + 1.0;
+    }
+}
+
+double* DirectDFJK::get_metric_power(double power){
+	bool pow_on = false;
+	double ret_power;
+	size_t met_ind;
+	for (size_t met_iter = 0; met_iter < metric_.size(); met_iter++){
+		if (std::fabs(power - met_powers_[met_iter]) < condition_ ){
+			met_ind = met_iter;
+			pow_on = true;
+		}
+	}
+	if (!pow_on) {
+		met_ind = metric_.size();
+		SharedMatrix my_met;
+		auto coul_met = std::make_shared<FittingMetric>(auxiliary_, true);
+		coul_met->form_fitting_metric();
+		my_met = coul_met->get_metric();
+		my_met->power(power, condition_);
+		metric_.push_back(my_met);
+		met_powers_.push_back(power);
+	}
+	return metric_[met_ind]->pointer()[0];
+}
+
+// Computes the LU decomposition of the coulomb metric and stores it in
+//   member vector CMPQ_LU_.
+//   pivots are storec in vector PERMUTE_
+void DirectDFJK::get_met(){
+	auto coul_met = std::make_shared<FittingMetric>(auxiliary_, true);
+    int info;
+
+	coul_met->form_fitting_metric();
+	SharedMatrix my_met;
+	my_met = coul_met->get_metric();
+	PERMUTE_.resize(naux_);
+	CMPQ_LU_.resize(naux_*naux_);
+
+    CMPQ_inv_.resize(naux_*naux_);
+
+    double ZERO[8] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
+    double* zerop = &ZERO[0];
+
+    double* work_p = (double*) malloc(sizeof(double)*naux_);
+    //C_DCOPY( naux_, zerop, 0, work_p, 1);
+
+    memset(work_p, 0, naux_ * 8);
+
+	double* cmpq_lup = &CMPQ_LU_.front();
+    double* cmpq_invp = &CMPQ_inv_.front();
+	double* metp = my_met->pointer()[0];
+	int* pert = &PERMUTE_.front();
+	
+	info = C_DGETRF( naux_, naux_, metp, naux_, pert);
+	C_DCOPY(naux_*naux_, metp, 1, cmpq_lup, 1);
+
+    C_DCOPY(naux_*naux_, cmpq_lup, 1, cmpq_invp, 1);
+
+    C_DGETRI( naux_, cmpq_invp, naux_, pert, work_p, naux_);
+
+    if (work_p) {free(work_p);}
+	
+}
+
+// called after sparsity_prep_pQq() which is called in the memory estimate
+// is called in the memory estimate because the maximum permissable 
+// block size is needed for an effective memory estimate, and determining
+// that is equivalent to estimating the total amount of memory
+// Determines the number of blocks and 
+void DirectDFJK::prepare_p_blocks() {
+	Block_funcs_.resize(0);
+	Shell_starts_.resize(0);
+	Shell_stops_.resize(0);
+	k_disps_.resize(0);
+
+	size_t current_costs = 0;
+	// block max functions
+	size_t bmf = 0;
+
+	size_t block_size = 0;
+	size_t shell_funcs;
+    size_t biggest_shell = 0;
+
+	// one of two things. Either nelectron/2 + 1 or C_left_ao->ncol()
+	size_t charges_z;
+
+	if (C_left_ao_.size() == 0) {
+	// atomic charges float
+	double charges_f;
+
+/*	for (size_t func_it = 0; func_it < schwarz_func_ints_.size(); func_it++) {
+		sparse_fpf_ += schwarz_func_ints_[func_it];
+	}	*/
+
+// We need to figure out the biggest shell in the primary basis set
+    // to allocate memory for AO construction
+    for (size_t i = 0; i < primary_->nshell(); i++) {
+        if (primary_->shell(i).nfunction() > biggest_shell ) {
+            biggest_shell = primary_->shell(i).nfunction();
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


psi4/src/psi4/libfock/jk.h
```diff
@@ -562,6 +562,9 @@ class PSI_API JK {
     * type on output file
     */
     virtual void print_header() const = 0;
+/*JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_*/
+    int n_col_last_ = -1;
+/*JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_JSOB_DEBUG_*/
```
loriab, line 567: remove?

psi4/src/psi4/libfock/jk.h
```diff
@@ -1108,6 +1111,242 @@ class PSI_API MemDFJK : public JK {
     std::shared_ptr<DFHelper> dfh() { return dfh_; }
 };
 
+//}
+//
+//#endif
+
+
+class PSI_API DirectDFJK : public JK {
```
loriab, line 1119: run this through a formatter, please

psi4/src/read_options.cc
```diff
@@ -303,13 +303,6 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Thole damping factor for multipole fields -*/
         options.add_double("DAMPING_FACTOR_MULTIPOLE", 2.1304);
 
-        /*- Summation scheme for field computations, can be direct or fmm -*/
-        options.add_str_i("SUMMATION_FIELDS", "DIRECT", "DIRECT FMM");
-        /*- Expansion order of the multipoles for FMM -*/
-        options.add_int("TREE_EXPANSION_ORDER", 5);
-        /*- Opening angle theta -*/
-        options.add_double("TREE_THETA", 0.5);
```
loriab, line 311: these are for PE, I think. why remove?

psi4/src/read_options.cc
```diff
@@ -1022,23 +1012,14 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
-        options.add_str("SAPT_DFT_GRAC_DETERMINATION", "INPUT", "INPUT");
-        /*- Enables the hybrid xc kernel in dispersion? !expert -*/
-        options.add_bool("SAPT_DFT_DO_HYBRID", true);
-        /*- Scheme for approximating exchange-dispersion for SAPT-DFT.
-        ``NONE`` Use unscaled ``Exch-Disp2,u`` .
-        ``FIXED`` Use a fixed factor |sapt__sapt_dft_exch_disp_fixed_scale| to scale ``Exch-Disp2,u`` .
-        ``DISP`` Use the ratio of ``Disp2,r`` and ``Disp2,u`` to scale ``Exch-Disp2,u`` . -*/
-        options.add_str("SAPT_DFT_EXCH_DISP_SCALE_SCHEME", "DISP", "NONE FIXED DISP");
-        /*- Exch-disp scaling factor for FIXED scheme for |sapt__sapt_dft_exch_disp_scale_scheme|. Default value of 0.686 suggested by Hesselmann and Korona, J. Chem. Phys. 141, 094107 (2014). !expert -*/
-        options.add_double("SAPT_DFT_EXCH_DISP_FIXED_SCALE", 0.686);
```
loriab, line 1041: these are for dft-sapt, why remove?

tests/scf5/input.dat
```diff
@@ -168,6 +192,10 @@ E = energy('scf')
...
 
-set scf_type df
+set scf_type disk_df
+E = energy('scf')
+compare_values(Eref_rohf_df, E, 6, 'Triplet Disk DF CUHF energy') #TEST
+
+set scf_type mem_df
 E = energy('scf')
-compare_values(Eref_rohf_df, E, 6, 'Triplet DF CUHF energy') #TEST
+compare_values(Eref_rohf_df, E, 6, 'Triplet Mem DF CUHF energy') #TEST
```
loriab, line 201: if you have time, would you edit the tests/pytests/test_addons.py accordingly?

then I'll expand stdsuite ref testing.

loriab (commented):
thanks for expanding the SCF stable

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+//I decided to forgo copying from MemDFJK because
+//that code seems more specialized. 
+void DirectDFJK::print_header() const {
+	if (print_)	{
+		outfile->Printf("==> DirectDFJK: Density-Fitted J/K Matrices <==\n\n");
+
+        outfile->Printf("    J tasked:          %11s\n", (do_J_ ? "Yes" : "No"));
+        outfile->Printf("    K tasked:          %11s\n", (do_K_ ? "Yes" : "No"));
+		outfile->Printf("    wK tasked:         %11s\n", (do_wK_ ? "Yes" : "No"));
+		if (do_wK_) outfile->Printf("Don't interpret these results.\nWe don't have w integrals in DirectDFJK\n");
```
susilehtola, line 97: Why not?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+#ifdef _OPENMP
+	procs = omp_nthread_;
+#endif
+
+	// variables to hold data
+	double global_max_int = 0.0;
+
+
+	std::unique_ptr<double[]> shel_maxes(new double[p_shells_*p_shells_]);
+	std::unique_ptr<double[]> integrals(new double[nbf_*nbf_]);
```
susilehtola, line 158: Why not `std::vector<double>`?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+		for (size_t Shell_N = 0; Shell_N < p_shells_; Shell_N++) {
+	        Shell_N_start = primary_->shell(Shell_N).function_index();
+			Shell_N_count = primary_->shell(Shell_N).nfunction();
+			eri[rank]->compute_shell(Shell_M, Shell_N, Shell_M, Shell_N);
+			for (size_t func_m = 0; func_m < Shell_M_count; func_m++){
+				for (size_t func_n = 0; func_n < Shell_N_count; func_n++){
+					integrals[ (Shell_M_start + func_m) * nbf_ + Shell_N_start + func_n ]
+					=
+//					integrals[ (Shell_N_start + func_n) * nbf_ + Shell_M_start + func_m ]
+//					=
```
susilehtola, line 212: Remove comment

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+ fabs(buffer[rank][ func_m * Shell_N_count * Shell_M_count * Shell_N_count + 
+				    func_n * Shell_M_count * Shell_N_count + 
+				    func_m * Shell_N_count + 
+				    func_n]);
+					if (global_max_int < val) global_max_int = val;
+					
+					if (shel_maxes[Shell_M * p_shells_ + Shell_N] < val) {
+						shel_maxes[Shell_M * p_shells_ + Shell_N]
+//						=
+//						shel_maxes[Shell_N * p_shells_ + Shell_M]
```
susilehtola, line 224: Same here

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+						= val;
+					}
+				}
+			}
+		}
+	}
```
susilehtola, line 230: So, `integrals` is just the Schwarz screening matrix? You should use a more descriptive name.

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+				schwarz_func_starts_pQq_[shell_it_out].push_back(count);
+				count += primary_->shell(shell_it_in).nfunction();
+			}
+		}
+		schwarz_dense_funcs_[shell_it_out] = count;
+	}
+}
+
+// calculates a block of AO's then conducts the proper contractions to get them
+//   where they need to go. I don't think it makes sense to have seperate J and
```
susilehtola, line 252: separate

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+	//D
+	our_needs += nbf_*nbf_;
+	free_memory_-=nbf_*nbf_;
+
+	//C_a
+	our_needs += nbf_*charges_z;
+	free_memory_-=nbf_*charges_z;
+	//C_b
+	our_needs += nbf_*charges_z;
+	free_memory_-=nbf_*charges_z;
```
susilehtola, line 311: So... you're assuming a neutral molecule and you can fit all half-transformed integrals in memory? Both are no-no's.

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+		my_met->power(power, condition_);
+		metric_.push_back(my_met);
+		met_powers_.push_back(power);
+	}
+	return metric_[met_ind]->pointer()[0];
+}
+
+// Computes the LU decomposition of the coulomb metric and stores it in
+//   member vector CMPQ_LU_.
+//   pivots are storec in vector PERMUTE_
```
susilehtola, line 383: What's this used for?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+	my_met = coul_met->get_metric();
+	PERMUTE_.resize(naux_);
+	CMPQ_LU_.resize(naux_*naux_);
+
+    CMPQ_inv_.resize(naux_*naux_);
+
+    double ZERO[8] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
+    double* zerop = &ZERO[0];
+
+    double* work_p = (double*) malloc(sizeof(double)*naux_);
```
susilehtola, line 399: Why not `std::vector<double>`?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1600 @@
...
+
+void DirectDFJK::compute_dense_AO_block_p_pQq(size_t shell, double* ao_block, std::vector<std::shared_ptr<TwoBodyAOInt>> eri){
+    int procs = 1;
+#ifdef _OPENMP
+	procs = omp_nthread_;
+#endif
+    
+
+    std::vector<const double*> buffer(procs);
+#pragma omp parallel num_threads(procs)
```
susilehtola, line 594: This should be within `#ifdef _OPENMP`

psi4/src/read_options.cc
obrien951: I can change that. I copied that file in from elsewhere, so it doesn't contain the changes from Yi

susilehtola:
It's sort of hard to follow the code.

You should remove all the stuff about charges; the algorithm needs to do the memory estimate based on the actual occupied orbitals.

You should separate the Coulomb and exchange bits, since Coulomb can be computed as a dot product, while exchange is a matrix multiplication. Range-separated exchange should be supported.

You also can't assume that you can fit all the integrals into memory in a big system; you need to batch over the occupied orbitals based on the amount of memory that's available. 

Please check Weigend's paper on how to do RI-HF.
https://pubs.rsc.org/en/content/articlelanding/2002/cp/b204199p

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: I didn't plan on resizing the data, so I thought unique_ptr would be sufficient.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: ok

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: ok

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: ok

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: Integrals are not stored between SCF iterations. This is only meant as a rough estimate so that the DFT grid doesn't take all the memory. The code in the scf iterations allocates space based on the size of the actual SCF coefficient matrix from iteration to iteration, so this isn't a problem.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: It's used for constructing the Coulomb Matrix

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: I will change that

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: Will change

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: Then use `std::array`. `new` and `malloc` are things to avoid, since there's no performance benefit but debugging is harder;  you just make it easier to shoot yourself in the foot.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: Hm, not sure I follow. If you have the half-inverse overlap already, you don't need anything more.

psi4/src/read_options.cc
maxscheurer: Yes, please don't remove those options 😄

obrien951:
@susilehtola Algorithm agrees with Weigend.

psi4/src/psi4/libfock/jk.h
obrien951: yes remove for sure.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: As far as I can tell, `std::array` requires knowing the size of the data at compile time which wouldn't be true about this case because we rely on the size of the basis set here. Otherwise, I would gladly use it.

I'm also happy to use the standard libraries to manage allocation and deallocation of memory in my code. That is what motivated my use of smart pointers as you can see in lines 157 and 158. I'm not a huge fan of how I have to put `new` in there myself, but I did a lot of validation to make sure this was right.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: ok

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: In line with Weigend's formulation, I'm constructing the coulomb matrix with a transform against the full inverse of the coulomb metric. I had problems with conditioning, so I went with the LU decomposition with partial pivoting to avoid having to rewrite the code more than once.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: If `std::array` doesn't work, then use `std::vector`. There is no performance hit; it just allows for cleaner code i.e. no `new` or `free`. This is already preferred elsewhere in Psi

susilehtola:
Can you reformat the code? It's annoying to read.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: `std::unique_ptr` also does not require use of free. There is also preference for `std::unique_ptr` elsewhere in psi

hokru:
This is a big task! Thanks for taking it on!

Just wondering why little use of existing code is made?

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: The formulation we chose for the Fock matrix components relies on a symmetry that gets broken in the DF approximation to the omega integrals. Adding this capability would represent a significant change and would be another PR in its own right.

susilehtola: Which symmetry..?

obrien951:
@hokru Integrals are done through libmints. I wrote this code when I was starting out, so I didn't know the codebase that well.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: B^Q_{i \mu } = B^Q_{i \mu } following Weigend's notation

susilehtola: That's an identity..?

obrien951: In the contraction

K^{\mu \nu} = B^Q_{i \mu} B^Q_{i \nu}

We need that identity to hold. In the omega eri case, that symmetry does not hold, so a different strategy than the one taken in this PR would need to be used.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: I'm still confused about why you're bringing up `new` and `delete` because the `std::unique_ptr` takes care of those operations.

Beyond the point of `std::unique_ptr` addressing concerns about memory management and smart pointers in general being useful, lightweight, and a part of several key pieces of code in Psi, I would have to change all the memory accessing in the code which could be problematic for a number of reasons:

1) It would add no enhancement to functionality.
2) It would require changes to several hundred lines of code which I would need to do between now and December 31 when I am off Georgia Tech's payroll.

psi4/src/psi4/libfock/DirectDFJK.cc
andysim: I *think* what's going on here is the handling of the metric inverse term is different.  The B notation is the three index tensor contracted against the inverse square root of the Coulomb metric.  It looks like the erfc terms are not handled this way; one of the three index tensors is just the Coulomb term, while the other is the Coulomb term contracted against the inverse erfc metric.  Both metrics should be positive semidefinite (I think) so it's not immediately obvious to me why this asymmetry is introduced.

obrien951: You're right. It wouldn't introduce the problem that I assumed it would. In any case, it would be a lot of work to extend the code to the omega case, and I graduated last week. 

When this code was written over a year ago, the omega extension wasn't seen as a high priority, so it wasn't included. I am trying to finish this PR in the next couple weeks, so in any case, I won't be able to add this feature.

obrien951: It also isn't immediately clear to me how to generalize the particular algorithm in this code to the omega case. Worst case scenario, we can ask the user to use disk.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: Actually I take that back. I've looked up the API, and your suggested change doesn't seem like it will change anything. I've doublechecked in other parts of psi, and the same syntax as I have here is used there as well.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: Congrats @obrien951 !

susilehtola: Generalizing to the omega case should be as simple as replacing all the integrals with their attenuated versions.

Of course, then J and K have be evaluated separately.

susilehtola: ... which may be what one wants anyhow, since for large systems forming K with DF may be slower than with exact integrals.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: Sorry, I just saw the `new` statements, I didn't realize you had `unique_ptr`.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: Well, it makes compilers warn about pragmas that have no effect.

obrien951: You're asking me to make changes in my code that will still leave users with hundreds of the same compiler warnings from other parts of the codebase. I don't see how your suggestion affects the end user.

obrien951:
@susilehtola 

Regarding your first suggestion about charges, I would like to make that change, however, that information is not available at the time of construction of the JK object. More general changes will need to be made to the codebase for that to be possible. 

Separating the Coulomb and Exchange bits in a way that does not require extra AO ERI determination is beyond me. If you have a suggestion as to how to do it, I will consider implementing it in the limited time I have left. Otherwise, I will regard the task as impossible.

I'm confused about the perceived assumption that all integrals fit into memory. No such assumption is made. In fact, this code can be seen as an extension of Weigend's work to the case in which they *don't* all fit in memory, and working with the assumption that integrals do *not* fit in memory is the reason for many of the strange choices made in writing this code.

susilehtola:
> I'm confused about the perceived assumption that all integrals fit into memory. No such assumption is made. In fact, this code can be seen as an extension of Weigend's work to the case in which they _don't_ all fit in memory, and working with the assumption that integrals do _not_ fit in memory is the reason for many of the strange choices made in writing this code.

@obrien951 My comment originates from the reason that like I said, it's hard to follow the code. There are **no comments** describing what the functions actually do.

Also, if you look at when I posted the comment, it was 24 days ago, on the week before Thanksgiving, way before our lengthy discussion on Slack where it took several days and dozens of messages to understand that your algorithm is nothing else but RI-HF as presented by Weigend in the 2002 paper, with only the difference that in case it is not possible to fit even a single orbital vector in memory (the `(ui|A)` intermediates), you do some sort of blocking, but what kind, I do not know.

These sorts of unclarities mean there should be a short, general description of what the code does in the header file. If it's RI-HF à la Weigend, a reference to the paper and a brief description on the differences is enough.

At the moment, the function names are opaque. Some functions have unintelligible descriptions like `// Line  7 algorithm  8` (line and algorithm of what?!) and there are dozens of functions that don't have any kind of description.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: code quality ftw :rofl:

obrien951:
@susilehtola Apologies for missing the date on those remarks. For some reason I thought I saw them after your remark about annoying code.

The blocking is honestly difficult to describe in comments. I will try to provide an illustration in the code, but the best remedy would be for someone trying to work on the code to consult my thesis which I will make more widely available.

As for your last point about clarity of operations in the code, that can be fixed. The code was written a while ago, and there was a set of notes that went along with the code that is no longer useful. Comments that are more relevant will be added.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: If it's part of a larger change in standards in psi to eliminate these messages from the compiler, I will be happy to make the change. I can understand the wish to enforce strict standards in a package where the entire world can see you as a core dev.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: @susilehtola Thank you for your well wishes

Because of the particulars of my blocking scheme, extending to the omega case actually will require a significant degree of work both in formulating the working equations and writing the code.  Someone would have to write out the equations to figure which block of integrals would be attenuated and which block would be contracted against the coulomb metric, assuming we're still using the working equation from DiskDFJK:

<sup>&omega;</sup> K<sub>&mu;&nu;</sub> = C<sub>&lambda;i</sub>(Q|&mu;&lambda;) (Q|&omega;|&nu;&sigma;) C<sub>&sigma;i</sub>

Someone would have to rewrite a loop over the blocking index to compute blocks of  C<sub>&lambda;i</sub>(Q|&mu;&lambda;) over some values of the outer blocking index and (Q|&omega;|&nu;&sigma;) C<sub>&sigma;i</sub> over others. I chose Q for both tensors here although only the non-omega tensor is contracted against the inverse coulomb metric. They would also have to see if this is better than dealing with having to have more blocks over the auxiliary blocking index as would be required by the larger number of tensors when dealing with attenuated integrals.

Luckily, in my experience with working with attenuated integrals, the extra time to calculate a tensor (A|&omega;|&nu;&sigma;)  of those integrals rather than a tensor (A|&mu;&lambda;) (index changed to A because this time it's uncontracted) is not the thing that's going to make the code takes longer.

Furthermore, while our testing shows that blocking over the primary basis set is faster for the non-attenuated case, it is not clear that this will hold for the attenuated case which will require more testing and which is a reason that I kept the code for blocking over the auxiliary basis set in this PR. This is all without mentioning coding up combined calculation of exchange terms.

Both formulating and testing that code would represent a significant endeavor that would require large additions to the code included in this PR.

susilehtola: I think all integrals should be attenuated, so then it really is as simple as replacing the integral call from `1/r12` to the range separated kernel `[a*erf(w r12) + b*erfc(w r12)/r12]`

obrien951: If I'm not mistaken, you run into problems because then you're applying the kernel twice in the formulation. Rather than once in the conventional (that is the four-index eri) case. That said, if there's a formulation of the working equations that works better, I might be able to work into it.

When I looked through the DiskDFJK code and when I wrote the MemDFJK wK code which is currently working, the working equations involved one tensor of coulomb-operator 3-index integrals contracted against the inverse coulomb metric and another tensor of attenuated integrals as in

<sup>&omega;</sup> K<sub>&mu;&nu;</sub> = C<sub>&lambda;i</sub>(Q|&mu;&lambda;) (Q|&omega;|&nu;&sigma;) C<sub>&sigma;i</sub>

obrien951: i.e. <sup>&omega;</sup> K<sub>&mu;&nu;</sub> = C<sub>&lambda;i</sub>(A|&mu;&lambda;)(A|Q)<sup>-1</sup> (Q|&omega;|&nu;&sigma;) C<sub>&sigma;i</sub>

susilehtola: Um, I don't think so; it's just that you define the fitting in terms of the fitted kernel. Like you say, the range-separated exchange is
 <sup>&omega;</sup> K<sub>&mu;&nu;</sub> = &sum;<sub>i occ</sub> C<sub>&lambda;i</sub> (&mu;&lambda;|&omega;|&nu;&sigma;) C<sub>&sigma;i</sub>

Now you just do RI employing the &omega; metric as
(&mu;&lambda;|&omega;|&nu;&sigma;) &approx; (&mu;&lambda;|&omega;|A) (A|&omega;|B)<sup>-1</sup> (B|&omega;|&nu;&sigma;)
where your fit functions are normalized (A|&omega;|A) = 1.

This is exactly the same as the usual global hybrid: 
(&mu;&lambda;|&nu;&sigma;) &approx; (&mu;&lambda;|A) (A|B)<sup>-1</sup> (B|&nu;&sigma;)  with (A|A) = 1 if you just redefine how the integrals are calculated.

obrien951: I'll see how that goes. It would be great to expand the capability of the code.

susilehtola: Yeah. I'm not sure where Psi4 normalizes the auxiliary functions. I think the code might work even if Coulomb normalization is used; the main thing is just that one does not want the norms to vary by orders of magnitude since then the eigensolver gets into problems. Also the eigenvalues will be affected; the cutoffs are always defined in terms of normalized matrices so that's another hiccup. (The code in `orthog.cc` handles this properly by normalizing the overlap matrix before getting its eigenvalues.)

obrien951: What do you mean when you say "normalize the auxiliary functions?" I haven't see that language to describe any of these processes before.

obrien951: But I can tell you for a fact that the working equations I wrote above are used in Psi because I'm the one who wrote that code.  I will, however, make myself look foolish if I go around telling people that that's the only way to do it. I haven't followed up on your working equation because I'm still working on cleaning up the rest of the code.

susilehtola: If you look up the literature on RI, they don't use the overlap normalization `<A|A>=1` but rather the Coulomb normalization `(A|A)=1`.

obrien951:
Latest changes should benefit readability

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: Sorry for taking so long. I had to break for the holiday.

I tried out your attenuated coulomb metric, but I had some serious issues with conditioning. Psi4 actually has software written to help deal with these issues, and I used them in my trial. My code has the steps in the proper order, but when I try to take the half-power coulomb metric, either DSYEV fails or the energies are completely wrong. I read the Ahlrichs paper. He gives some strategies for dealing with some of the conditioning and all those issues. Implementing them might best be done in the lib3index/fittingmetric.cc, and would likely require a separate PR.

Furthermore, Head-Gordon has shown that the coulomb metric is able to meet many of the needs sought by attenuated metrics 

https://doi.org/10.1073/pnas.0408475102

and I am told that the Coulomb metric is seen as a more conventional way of performing density fitted attenuated calculations largely for this reason.

I think what I'm failing to communicate here is that if the two tensors between the coefficient matrices are not the same then the code must be rewritten starting with a new algorithm. I think that the amount of work that will go into extending direct-df calculations to attenuated integrals in a way that will not be restricted by system size is enough for a separate PR, and it would not have been possible for me to do in the time frame that came about from when we reran the trials of this code at the beginning of this fall semester.

I really hope I'm not coming off as recalcitrant or lazy. I tried the attenuated metric, and it didn't work within the parameters I needed it to.

susilehtola: OK. Kállay also suggests using the Coulomb metric in https://pubs.acs.org/doi/10.1021/acs.jctc.9b00891, as the attenuated metric may be singular.

I still think it should be possible to use the attenuated metric to accomplish fitting, if one handles the inverse appropriately. The pivoted Cholesky scheme implemented in `orthog.cc` has no problems with singular basis sets.

obrien951: This brings us back to the original problem that using the inverted coulomb metric would require completely reformulating the blocking scheme for the case in which the doubly contracted eri tensors don't fit in memory. If we had known the efficacy of the code earlier, then I would have been happy to do it, but as of now, it's not a project I can undertake.

susilehtola: Yeah. Just waiting on reviews from core devs

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+        std::stringstream error_message;
+        error_message << "DirectDFJK: not enough memory for overhead! Overhead requires "
+                      << current_costs * 8 / (1024 * 1024) << " [MiB].";
+        throw PSIEXCEPTION(error_message.str().c_str());
+    }
+
+    Shell_starts_.push_back(0);
+    for (size_t shell_iter = 0; shell_iter < p_shells_; shell_iter++) {
+        shell_funcs = primary_->shell(shell_iter).nfunction();
+        // The 5 is where we're getting our fudge factor we're getting charges_z
```
hokru, line 490: What 5?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+        //   to stand in for n_occ, and there should only be a factor of 2 included for
+        //   the number of tensors with dimension occ. The extra 3 is a massive amount of
+        //   memory that should cover the overhead from the reast of the calculation.
+        //   charges_z also includes a fudge factor
```
hokru, line 494: Is this comment at the wrong place?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+        // to allocate memory for AO construction
+        for (size_t i = 0; i < primary_->nshell(); i++) {
+            if (primary_->shell(i).nfunction() > biggest_shell) {
+                biggest_shell = primary_->shell(i).nfunction();
+            }
+        }
+
+        std::shared_ptr<Molecule> mol_ptr = primary_->molecule();
+
+        for (int atom_iter = 0; atom_iter < mol_ptr->natom(); atom_iter++) charges_f += mol_ptr->fZ(atom_iter);
```
hokru, line 460: Do you have plans to account for non-neutral molecules? Please add a comment in the code if not.

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+    //	std::vector<size_t> stops;//stops_.get();
+    size_t current_costs = 0;
+    double charges_f = 0.0;
+    size_t charges_z = 0;
+
+    std::shared_ptr<Molecule> mol_ptr = primary_->molecule();
+
+    for (int atom_iter = 0; atom_iter < mol_ptr->natom(); atom_iter++) charges_f += mol_ptr->fZ(atom_iter);
+
+    charges_f = charges_f / 2.0;
```
hokru, line 543: repeat of the p-block section? Maybe code repetition could be avoided by moving both parts to the class init?

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+       x o x o               x x x o
+       x x x x               x x x x
+         1 2<----new_iter      1<------new_iter
+
+       final: copy lower to upper
+       Legend x: constructed o: empty
+
+       Details can be found by requesting
+       Joseph Senan O'Brien thesis from the Georgia Tech Library
+       In that document (Q|mi) <--> X^Q_{\mu a}
```
hokru, line 1317: Is the thesis digitally available? If so please add DOI (or whatever link is available)

hokru (commented):
Not a thorough review, but left a few comments.

Are users getting a 'not implemented' message? 

Could you add to the PR header a few practical information like timings and memory requirements compared to e.g. mem_df (since you know that best, i suppose)? Nothing extensive, just general guidance what to expect.

psi4/src/psi4/libfock/DirectDFJK.cc
susilehtola: @hokru ... and that's why porting `FittingMetric` to using `orthog.cc` would be good, since the pivoted Cholesky in `orthog.cc` has no problem with overcomplete sets.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: I'll change to comment to add the URL. It wasn't available when I added that comment

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: no, this code is distinct for the q-blocking case. It is called in the memory estimator which is itself called in class init iirc.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: Unfortunately, this is the best I could do here. I believe i had had a comment there explaining what was going on, but I'll double check. The issue is that there's no way to know the actual dimensions of the orbital coefficient matrices every time this function is called, so I have to do what you see here to make sure that I'm asking psi to give the JK object enough memory to run the calculation. I'm using the molecule's charge as a way of guessing at the dimension of the coefficient matrix, and there's enough of a fudge factor here to make sure it doesn't cause any problems.

When the code gets called during the actual SCF iterations, the object does interrogate the coefficient matrices, so the non-neutral case is indeed handled.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: yes. it's the comment I mentioned in my next comment.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: That comment is for an old version of the code. I will fix it.

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+
+    coul_met->form_fitting_metric();
+    SharedMatrix my_met;
+    my_met = coul_met->get_metric();
+    PERMUTE_.resize(naux_);
+    CMPQ_LU_.resize(naux_ * naux_);
+
+    CMPQ_inv_.resize(naux_ * naux_);
+
+//    double ZERO[8] = {0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0};
```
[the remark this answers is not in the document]
obrien951: sorry about that.

psi4/src/psi4/libfock/DirectDFJK.cc
```diff
@@ -0,0 +1,1785 @@
...
+}
+size_t DirectDFJK::memory_estimate() {
+    if (Qpq_) {
+        prepare_Q_blocks();
+    }
+
+    our_needs();
+    return total_needs_;
+}
+
```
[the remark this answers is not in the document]
obrien951: Perhaps I should have documented this better. All that's going on here is that the DirectDFJK object is making sure it has all the metrics it needs. Checking for rows and columns is probably me just being overly cautious about memory when I first wrote the code. Don't worry: I'm not trying to write my own metric building code at all.

psi4/src/psi4/libfock/DirectDFJK.cc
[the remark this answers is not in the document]
obrien951: Also I can't reply to your previous comment:

"Instead of LU you could try the FittingMetric::form_cholesky_inverse() routine. That is closer to what is done by Ahlrichs and also used e.g. in ORCA's RIJK implementation. With available optimized RI basis sets that works well and is fast.

Heuristically auto-generated (over-complete) RI basis sets as available in ORCA can lead to numerical problems and pivoting is needed"

so I will here. I did try the cholesky inverse when I was first writing the code, and conditioning errors persisted. That's why I went with the LU decomposition with partial pivoting.

psi4/src/psi4/libfock/DirectDFJK.cc
obrien951: The two bits of code are distinct, and they get called in the memory estimate which iirc is called when the object is initiated.

PR #2044, opened 2020-12-02
Build Treutler grid if asked to do so.

susilehtola:
## Description
Fixes a bug where a Becke grid is used instead of Treutler in case a Treutler grid was requested.

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/cubature.cc
+++ b/psi4/src/psi4/libfock/cubature.cc
@@ -4857,7 +4857,7 @@ std::shared_ptr<RadialGrid> RadialGrid::build(const std::string &scheme, int npo
     if (scheme == "BECKE") {
         return RadialGrid::build_becke(npoints, alpha, Z);
     } else if (scheme == "TREUTLER") {
-        return RadialGrid::build_becke(npoints, alpha, Z);
+        return RadialGrid::build_treutler(npoints, alpha, Z);
     } else {
         throw PSIEXCEPTION("RadialGrid::build: Unrecognized radial grid.");
     }


robertodr (approved):

loriab (approved):

susilehtola:
I can't really say I understand the structure of `cubature.cc`. There appears to be a wild amount of code duplication. For instance, the functions `ahlrichs_r` https://github.com/psi4/psi4/blob/2f085df4cab054b8b01c3e5c0bc1e1270c2e7bf2/psi4/src/psi4/libfock/cubature.cc#L2046 and `ahlrichs_dr` https://github.com/psi4/psi4/blob/2f085df4cab054b8b01c3e5c0bc1e1270c2e7bf2/psi4/src/psi4/libfock/cubature.cc#L2050 appear to be duplicated in the `build_treutler` function, albeit inconsistently: alpha is used half-properly and half hard-coded to 0.6.

Similarly, `build_treutler` also duplicates the `getChebychevRoots` function https://github.com/psi4/psi4/blob/2f085df4cab054b8b01c3e5c0bc1e1270c2e7bf2/psi4/src/psi4/libfock/cubature.cc#L2096.

But then, https://github.com/psi4/psi4/blob/2f085df4cab054b8b01c3e5c0bc1e1270c2e7bf2/psi4/src/psi4/libfock/cubature.cc#L2400 makes it seem like the `build_becke` and `build_treutler` functions aren't even used! What's the idea here?!

hokru:
So `DFT_RADIAL_SCHEME=TREUTLER` (the default) was building Becke grids?

susilehtola:
> So `DFT_RADIAL_SCHEME=TREUTLER` (the default) was building Becke grids?

I guess that's so. But like I said above, I don't understand the structure of the file: there's significant code duplication and overlap.

hokru:
Adding breaking calls to exceptions, I suspect the proper `build/build_becke/build_treutler` functions are never called?!
At this point I don't know either what is going on. The file is too long and confusing, including the `hack` routines that survived.

susilehtola:
Yes, it definitely should be restructured and any unused code eliminated. At the moment I can't compile Psi4 due to #2046, so maybe you want to have a stab at the cleanup, knowing the code already

hokru:
**user summary**:
This is not a bug that is affecting any DFT calculations currently run with PSI4.
The bug is found in unused code.

---

The radial grid is entirely build with https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/cubature.cc#L2433

The function supposed to call `build_becke` `build_treutler` is never called, instead a dummy function is called that only populates the `RadialGrid` class, which is never used in `cubature.cc` anyway (looks like it is nowhere used). 

On the plus side, the bug fixed with this PR is never encountered. The downside is that the TREUTLER scheme is not entirely correct because at least the eta mapping parameters (`TreutlerEta`) are missing.

 If `build_treutler` is correct or not, I cannot 100% say. In a quick & dirty fashion I enforced it. _Without_ eta (set to 1.0) all DFT `ctests` pass. That's good!
I added the missing eta parameters with updated BS radii a while ago, but didn't notice that the etas changed nothing. I thought the effect was just too small. But there is a very minor one. Including the etas has 11 of 37 tests fail. (Did not run the pytests)

I will refactor `cubature.cc` ..somehow.. in the near future.

---

I will approve & merge tomorrow.

hokru (approved):

PR #2048, opened 2020-12-02
windows conda package through azure

### LINKED ISSUE #2060: PSI4 cmake build does not seem to check if python development is installed.
Hello, 

When building psi4 from source and python development packages are not installed compilation fails (Linux Fedora). Once I installed python-dev all complete successfully.
The complaint from source files  was along the lines

#include<Pyhon.h> not found and then aborts during compilation,

Ideally (IMO)  cmake should check during the configuration phase if all relevant and required package are  installed and abort if not. Perhaps there are more dependencies, but I may already have had them already installed. 

I no longer have the build machine to replicate all the steps, but I suspect amending the root cmake adding find_package with python-dev as REQUIRED (or perhaps via pkg_config) 
and abort immediately during the cmake configuration phase if not found, not halfway through the compilation when running make.

Perhaps this will also help.

https://cmake.org/cmake/help/latest/module/FindPython.html

Cheers.

loriab:
- [x] bump cmake to 3.15 from 3.8 (to get decent FindPython with NumPy)
- [x] switch out Python detection to fix #2060. note that all the CM vars now start `Python_`, not `PYTHON_`
- [x] use prebuilt g2g for Windows Azure
- [x] bump g2g to 2.0.7
- [ ] ~added some fn redefs for Windows in psi4-dec.h . right place?~
- [x] get windows conda package building through azure

Ok, Windows conda package was the original aim of this PR, but that's also in the Libxc5 PR that uses prebuilt g2g & libxc & libint2. This PR switched to using the new Python detection advocated by pb11: https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode . This means we're not using the FindPythonLibsNew that psi4 has a copy of from pb11 and which in turn we install for the benefit of plugins. Of course that means we need to use cmake 3.15+ (released July 2019). That's quite a leap from the 3.8+ we had before, and lgtm doesn't have it, hence that failure. We can drop a lot of explicit properties on `core` since pb11 is providing them (see https://github.com/pybind/pybind11/blob/master/tools/pybind11NewTools.cmake and https://github.com/pybind/pybind11/blob/master/tools/pybind11Common.cmake), but I'm not sure if this is the right time to make the leap. I'm also a bit surprised that `add_library(... MODULE)` seems to be working with the plugin tests where before we used `SHARED` to appear like a linux lib. Glad of opinions here.


psi4/src/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 ### >> Create core target, so we may link modules to it directly <<
 ### >> This allow to propagate dependencies upwards sanely
-pybind11_add_module(core SHARED NO_EXTRAS core.cc)
```
robertodr, line 3: `SHARED NO_EXTRAS` might have been my doing at some point, but I don't recall why... This seems the correct way to go for this case, according to the latest pybind11 documentation.

robertodr (approved):
I see no problem in bumping to CMake 3.15+: it's quite easy to get a recent version of CMake.

psi4/src/CMakeLists.txt
loriab: ok, I had supposed `SHARED NOEXTRAS` was from our last pb11 overhaul when we were still pushing back from plain `add_library(core ...)`. I'm would still have thought it should be `SHARED`, not `MODULE`, for the sake of the plugins. But plugins seem to work, and I don't think we want to link to `pybind11::embed` (https://github.com/pybind/pybind11/blob/master/tools/pybind11NewTools.cmake#L203-L207), so `MODULE` it is.

loriab:
> I see no problem in bumping to CMake 3.15+: it's quite easy to get a recent version of CMake.

I don't either, for users. Anyone successfully updated cmake on lgtm?

psi4/src/CMakeLists.txt
robertodr: As long as no one needs to do `-lpsi4` to produce _e.g._ an executable, we should use `MODULE`. I'm puzzled why it wasn't so already though :man_shrugging:

jturney (approved):

andysim (approved):
Wow, a lot of PYTHON->Python for you to track down there.  Is that syntax change a recent thing?  LGTM!

loriab:
> Wow, a lot of PYTHON->Python for you to track down there. Is that syntax change a recent thing? LGTM!

I guess long, long ago, cmake's native python's detection had no notion of finding matching interpreter/library pairs. That reported its results as `PYTHON_FOUND` and friends. Hence numpy/pb11/us used the `FindPythonLibsNew` module that had mixed c++/py in mind and also reported as `PYTHON_*`. But the cmake native detection has been much revamped, reformed, and componentized, enough to [inspire pybind11 to switch over](https://github.com/pybind/pybind11/pull/2370/files). The new setup reports as `Python_*`. The long time it's taken to get cmake's new detection modules hammered out leads to the big jump to 3.15.

There may still be plugin and docs build follow-up items, but I think I've caught most of the `Python`s. Main thing ppl have to do is edit their setup scripts if they're hinting the `PYTHON_EXECUTABLE`.

loriab:
We have a problem resulting from this PR. There's three ways to `add_library` in CMake -- STATIC, SHARED, MODULE . SHARED is the usual linux `.so` for linking, while MODULE is the dl_open style of a Py module. Up until this merged PR we were doing SHARED.

### SHARED
* links `libpython.so` into psi4's `core.*.so` -- this is new in the revamped pb11 add_pybind11_module that calls [Python_add_library](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPython/Support.cmake#L3306-3313)
* sets `DT_SONAME` on the `core.*.so`. This shows up as `Library soname` below in `readelf -d` of core.*.so. *This is important* b/c w/o this set, the full path to `core.*.so` gets linked into consumers like plugins, rather than just the library name. [trouble](https://stackoverflow.com/questions/27261288/cmake-linking-shared-c-object-from-externalproject-produces-binaries-with-rel)
* designed for ordinary library linking like how the plugins are using symbols from psi4. The https://github.com/psi4/psi4/blob/master/psi4/psi4Config.cmake.in#L141 does the linking of core.*.so to the plugins

### MODULE
* does *not* link `libpython.so` into psi4
* does *not* set `DT_SONAME` on psi4. that's ok for the local plugins where the full path that's baked in is the same as the RPATH+SONAME resolved through `ld`, but this breaks down during the v2rdm conda build. I've been unable to force a `DT_SONAME` entry onto a `MODULE` library
* one can perhaps get away with it (no errors; current state of `master`), but it's [advised not to link to a MODULE](https://stackoverflow.com/a/4968940)


### printouts from SHARED
```
>>> readelf -d stage/lib/psi4/core.cpython-39-x86_64-linux-gnu.so 

Dynamic section at offset 0x21a4aa0 contains 40 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libpython3.9.so.1.0]
 0x0000000000000001 (NEEDED)             Shared library: [libgg.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libxc.so.9]
 0x0000000000000001 (NEEDED)             Shared library: [libint2.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libmkl_rt.so]
 0x0000000000000001 (NEEDED)             Shared library: [libiomp5.so]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [libimf.so]
 0x0000000000000001 (NEEDED)             Shared library: [libsvml.so]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libintlc.so.5]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x000000000000000e (SONAME)             Library soname: [core.cpython-39-x86_64-linux-gnu.so]
 0x000000000000000f (RPATH)              Library rpath: [/home/psilocaluser/toolchainconda/envs/py39/lib:$ORIGIN/../:/home/psilocaluser/toolchainconda/envs/py39/lib]
 0x000000000000000c (INIT)               0x1d7000
 0x000000000000000d (FINI)               0x1b03250
 0x0000000000000019 (INIT_ARRAY)         0x2193c00
 0x000000000000001b (INIT_ARRAYSZ)       7200 (bytes)
 0x0000000000000004 (HASH)               0x260
 0x000000006ffffef5 (GNU_HASH)           0xe010
 0x0000000000000005 (STRTAB)             0x5e368
 0x0000000000000006 (SYMTAB)             0x1d118
 0x000000000000000a (STRSZ)              851488 (bytes)
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000003 (PLTGOT)             0x21a5d60
 0x0000000000000002 (PLTRELSZ)           121032 (bytes)
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000017 (JMPREL)             0x1b94d0
 0x0000000000000007 (RELA)               0x133a18
 0x0000000000000008 (RELASZ)             547512 (bytes)
 0x0000000000000009 (RELAENT)            24 (bytes)
 0x0000000000000018 (BIND_NOW)           
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x000000006ffffffe (VERNEED)            0x133868
 0x000000006fffffff (VERNEEDNUM)         6
 0x000000006ffffff0 (VERSYM)             0x12e188
 0x000000006ffffff9 (RELACOUNT)          15654
 0x0000000000000000 (NULL)               0x0
```
```
>>> ldd stage/lib/psi4/core.cpython-39-x86_64-linux-gnu.so 
	linux-vdso.so.1 =>  (0x00007fffd195a000)
	libpython3.9.so.1.0 => /home/psilocaluser/toolchainconda/envs/py39/lib/libpython3.9.so.1.0 (0x00007f6527640000)
	libgg.so.2 => /home/psilocaluser/toolchainconda/envs/py39/lib/libgg.so.2 (0x00007f652719f000)
	libxc.so.9 => /home/psilocaluser/toolchainconda/envs/py39/lib/libxc.so.9 (0x00007f6525fdf000)
	libint2.so.2 => /home/psilocaluser/toolchainconda/envs/py39/lib/libint2.so.2 (0x00007f64cf7cf000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f64cf5a5000)
	libmkl_rt.so => /home/psilocaluser/toolchainconda/envs/py39/lib/libmkl_rt.so (0x00007f64cee55000)
	libiomp5.so => /home/psilocaluser/toolchainconda/envs/py39/lib/libiomp5.so (0x00007f64cea4e000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f64ce832000)
	libimf.so => /theoryfs2/common/software/intel2019/compilers_and_libraries_2019.4.243/linux/compiler/lib/intel64_lin/libimf.so (0x00007f64ce292000)
	libsvml.so => /theoryfs2/common/software/intel2019/compilers_and_libraries_2019.4.243/linux/compiler/lib/intel64_lin/libsvml.so (0x00007f64cc8ed000)
	libstdc++.so.6 => /home/psilocaluser/toolchainconda/envs/py39/lib/libstdc++.so.6 (0x00007f64cc779000)
	libgcc_s.so.1 => /home/psilocaluser/toolchainconda/envs/py39/lib/libgcc_s.so.1 (0x00007f64cc765000)
	libintlc.so.5 => /theoryfs2/common/software/intel2019/compilers_and_libraries_2019.4.243/linux/compiler/lib/intel64_lin/libintlc.so.5 (0x00007f64cc4f2000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f64cc131000)
	libutil.so.1 => /lib64/libutil.so.1 (0x00007f64cbf2e000)
	librt.so.1 => /lib64/librt.so.1 (0x00007f64cbd25000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f64cba23000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f652a4b3000)
```

loriab:
above issues ultimately resolved in #2104

PR #2049, opened 2020-12-03
Eliminate gradient globals from `dfocc`

JonathonMisiewicz:
## Description
Part of the DF Gradient refactor. It's become clear that Rob's implementation of the DF-MP2 gradients can't be generalized without incurring a performance hit, so the scope of the project now is to generalize what `dfocc` does out of `dfocc` so other modules (like `dct`) can use it. The current step of that is to make the procedure independent of `dfocc` globals.

While I was at it, I got rid of unneeded headers and did some other minor code cleanup.

## Todos
- [x] Some `dfocc` cleanup

## Checklist
- [x] `ctest -L df$` and `test_standard_suite.py` pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfocc/dfocc.h
+++ b/psi4/src/psi4/dfocc/dfocc.h
@@ -79,8 +79,8 @@ class DFOCC : public Wavefunction {
     void tpdm_tilde_cc();
     void back_trans_cc();
     void dfgrad();
-    void oei_grad();
-    void tei_grad(std::string aux_type);
+    void oei_grad(std::map<std::string, SharedMatrix>&);
+    void tei_grad(std::string aux_type, std::map<std::string, SharedMatrix>&);
     void gfock_oo();
     void gfock_vo();
     void gfock_ov();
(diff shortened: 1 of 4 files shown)


psi4/src/psi4/dfocc/dfocc.h
```diff
@@ -595,10 +595,6 @@ class DFOCC : public Wavefunction {
     std::shared_ptr<DIISManager> ccsdDiisManagerAB;
     std::shared_ptr<DIISManager> ccsdlDiisManager;
 
-    // Gradients
-    std::map<std::string, SharedMatrix> gradients;
-    std::vector<std::string> gradient_terms;
```
loriab, line 600: yay for dropping things from this header

loriab (approved):
looks like a cleaner pattern for dfocc to use

jturney (approved):

andysim (approved):
Nice cleanup - thanks!

PR #2051, opened 2020-12-04
Remove spaces when reading options, fixes #2004

### LINKED ISSUE #2004: Fix issues with spaces in option names and unknown option names (PSIAPI)
The following snippet of code fails because of the extra space at the end of `'SCF_TYPE '`

```python
import psi4
psi4.set_options({'SCF_TYPE ' : 'DIRECT'})
```

This is an innocent error that should be fixed by psi4, or reported back to the user with a readable error message. Simply stripping the option label of left and right spaces would prevent this issue (with no potential side effects). If the option is not found, then psi4 should throw an exception saying something like "Option `'...'` is not valid."

This is the error message:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-c64c7a4449ad> in <module>
      1 import psi4
      2 
----> 3 psi4.set_options({'SCF_TYPE ' : 'DIRECT'})

/usr/local/psi4/lib/psi4/driver/p4util/python_helpers.py in set_options(options_dict, verbose)
    444     for k, v, in options_dict.items():
    445         mobj = optionre.match(k)
--> 446         module = mobj.group('module').upper()[:-2] if mobj.group('module') else None
    447         option = mobj.group('option').upper()
    448 

AttributeError: 'NoneType' object has no attribute 'group'
```

jeffschriber:
## Description
Remove spaces when reading options, fixes #2004

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -443,9 +443,7 @@ def set_options(options_dict, verbose=1):
 
     for k, v, in options_dict.items():
 
-        k = k.strip()
-
-        mobj = optionre.match(k)
+        mobj = optionre.match(k.strip())
         module = mobj.group('module').upper()[:-2] if mobj.group('module') else None
         option = mobj.group('option').upper()
 
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -442,6 +442,9 @@ def set_options(options_dict, verbose=1):
     rejected = {}
 
     for k, v, in options_dict.items():
+
+        k = k.strip()
+
         mobj = optionre.match(k)
         module = mobj.group('module').upper()[:-2] if mobj.group('module') else None
         option = mobj.group('option').upper()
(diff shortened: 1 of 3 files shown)


psi4/driver/p4util/python_helpers.py
```diff
@@ -442,6 +442,9 @@ def set_options(options_dict, verbose=1):
     rejected = {}
 
     for k, v, in options_dict.items():
+
+        k = k.strip()
```
loriab, line 446: neat, maybe add to following line `optionre.match(k.strip())` so avert the reference/value issue FAE brought up.

jeffschriber: Thanks!

loriab (approved):
thanks for fixing a lingering issue!

andysim (approved):
This is a game changer!

fevangelista (approved):
Nice fix!

tests/pytests/test_option.py
```diff
@@ -0,0 +1,23 @@
+import numpy as np
+import pytest
+
+import psi4
+
+pytestmark = pytest.mark.quick
+
+
+@pytest.mark.smoke
+def test_psi4_basic():
```
loriab, line 10:
```suggestion
def test_spacious_option():
```

this is fine -- I just need to trigger CI, since Azure didn't pick up your last commit

PR #2053, opened 2020-12-04
Optimize linalg::triplet matrix multiply

andyj10224:
## Description
linalg::triplet currently does matrix multiplication in the same order (AB)C. Sometimes, A(BC) may be more efficient, per observation of @zachglick 

## Todos

- [x] More efficient linalg::triplet call

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -3404,26 +3404,43 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA, 
 SharedMatrix triplet(const SharedMatrix &A, const SharedMatrix &B, const SharedMatrix &C, bool transA, bool transB,
                      bool transC) {
 
-    int dim1 = transA ? A->ncol() : A->nrow();
-    int dim2 = transA ? A->nrow() : A->ncol();
-    int dim3 = transB ? B->ncol() : B->nrow();
-    int dim4 = transB ? B->nrow() : B->ncol();
-    int dim5 = transC ? C->ncol() : C->nrow();
-    int dim6 = transC ? C->nrow() : C->ncol();
+    int nirrep = A->nirrep();
+    
+    if (nirrep != B->nirrep() || nirrep != C->nirrep()) {
+        throw PsiException("Input Matrices A, B, and C do not have the same number of irreps.", __FILE__, __LINE__);
+    }
 
-    // Checks validity of Matrix Multiply, don't want calculation to suddenly fail halfway through
-    if (dim2 != dim3 || dim4 != dim5) {
-        throw PsiException("Input matrices are of invalid size", __FILE__, __LINE__);
+    // cost1 = cost of (AB)C
+    // cost2 = cost of A(BC)
+    int cost1 = 0;
+    int cost2 = 0;
+
+    for (int h = 0; h < nirrep; h++) {
+
+        int dim1 = transA ? A->colspi(h) : A->rowspi(h);
+        int dim2 = transA ? A->rowspi(h) : A->colspi(h);
+        int dim3 = transB ? B->colspi(h) : B->rowspi(h);
+        int dim4 = transB ? B->rowspi(h) : B->colspi(h);
+        int dim5 = transC ? C->colspi(h) : C->rowspi(h);
+        int dim6 = transC ? C->rowspi(h) : C->colspi(h);
+        // Checks validity of Matrix Multiply, don't want calculation to suddenly fail halfway through
+        if (dim2 != dim3 || dim4 != dim5) {
+            throw PsiException("Input matrices are of invalid size", __FILE__, __LINE__);
+        }
+
+        // Cost of (AB)C compared to A(BC) per irrep
+        // A = dim1 * dim2, B = dim3 * dim4, C = dim5 * dim6
+        // (AB)C cost = dim1 * (dim2 == dim3) * dim4 + dim1 * (dim4 == dim5) * dim6
+        // A(BC) cost = (dim3 == dim2) * (dim4 == dim5) * dim6 + dim1 * (dim2 == dim3) * dim6
+        cost1 += dim1 * dim4 * (dim2 + dim6);
+        cost2 += dim2 * dim6 * (dim1 + dim4);
+
     }
 
     SharedMatrix T;
     SharedMatrix S;
 
-    // Cost of (AB)C compared to A(BC)
-    // A = dim1 * dim2, B = dim3 * dim4, C = dim5 * dim6
-    // (AB)C cost = dim1 * (dim2 == dim3) * dim4 + dim1 * (dim4 == dim5) * dim6
-    // A(BC) cost = (dim3 == dim2) * (dim4 == dim5) * dim6 + dim1 * (dim2 == dim3) * dim6
-    if (dim1 * dim4 * (dim2 + dim6) <= dim2 * dim6 * (dim1 + dim4)) {
+    if (cost1 <= cost2) {
         T = doublet(A, B, transA, transB);
         S = doublet(T, C, false, transC);
     } else {
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -3420,7 +3420,7 @@ SharedMatrix triplet(const SharedMatrix &A, const SharedMatrix &B, const SharedM
     int cost1 = 0;
     int cost2 = 0;
 
-    for (int h = 0; h < nirrep; h++) {
+    for (int h = 0; h < A->nirrep(); h++) {
 
         int dim1 = transA ? A->colspi(h) : A->rowspi(h);
         int dim2 = transA ? A->rowspi(h) : A->colspi(h);
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -3404,6 +3404,17 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA, 
 SharedMatrix triplet(const SharedMatrix &A, const SharedMatrix &B, const SharedMatrix &C, bool transA, bool transB,
                      bool transC) {
 
+    bool same_symmetry = (A->symmetry() == B->symmetry() && A->symmetry() == C->symmetry());
+
+    SharedMatrix T;
+    SharedMatrix S;
+
+    if (!same_symmetry) {
+	T = doublet(A, B, transA, transB);
+        S = doublet(T, C, false, transC);
+	return S;
+    }
+
     int nirrep = A->nirrep();
     
     if (nirrep != B->nirrep() || nirrep != C->nirrep()) {
@@ -3436,9 +3447,6 @@ SharedMatrix triplet(const SharedMatrix &A, const SharedMatrix &B, const SharedM
         cost2 += dim2 * dim6 * (dim1 + dim4);
 
     }
-
-    SharedMatrix T;
-    SharedMatrix S;
 
     if (cost1 <= cost2) {
         T = doublet(A, B, transA, transB);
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3403,8 +3403,49 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA,
...
                      bool transC) {
-    SharedMatrix T = doublet(A, B, transA, transB);
-    SharedMatrix S = doublet(T, C, false, transC);
+
+    int dim1 = A->nrow();
+    int dim2 = A->ncol();
+    int dim3 = B->nrow();
+    int dim4 = B->ncol();
+    int dim5 = C->nrow();
+    int dim6 = C->ncol();
```
susilehtola, line 3412: So A is 1 x 2, B is 3 x 4, and C is 5 x 6. OK.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3403,8 +3403,49 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA,
...
+        temp = dim3;
+        dim3 = dim4;
+        dim4 = temp;
+    }
+
+    if (transC) {
+        temp = dim5;
+        dim5 = dim6;
+        dim6 = temp;
+    }
```
susilehtola, line 3432: Here you've flipped everything so that the operation is (1 x 2) x (3 x 4) = (5 x 6).

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3403,8 +3403,49 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA,
...
+
+    if (dim2 != dim3 || dim4 != dim5) {
+        throw new PsiException("Input matrices are of invalid size", __FILE__, __LINE__);
+    }
```
susilehtola, line 3436: This check is wrong: one must have 1==5, 2==3, and 4==6.

psi4/src/psi4/libmints/matrix.cc
susilehtola: ... scratch that,  (1 x 2) x (3 x 4) x (5 x 6).

psi4/src/psi4/libmints/matrix.cc
susilehtola: Scratch that, it's okay.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3403,8 +3403,49 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA,
...
+
+    SharedMatrix T;
+    SharedMatrix S;
+
+    if (dim1 * dim4 * (dim2 + dim6) <= dim2 * dim6 * (dim1 + dim4)) {
```
susilehtola, line 3441: It would be helpful to have comments / steps here, since it'd be easier to check the dimensions. So...
The cost of doing AB is dim1 x (dim2==dim3) x dim4. Then doing (AB)C is dim1 x (dim4==dim5) x dim6. OK.
The cost of doing BC is dim3 x dim6 x (dim4==dim5). Then doing A(BC) is dim1 x (dim2==dim3) x dim6. OK.

psi4/src/psi4/libmints/matrix.cc
zachglick, line 3412: That was fast Andy!

You might try using ternary operators here to avoid swapping variables, like this:
```
int dim1 = transA ? A->ncol() : A->nrow();
int dim2 = transB ? B->ncol() : B->nrow();
int dim3 = transB ? B->nrow() : B->ncol();
int dim4 = transC ? C->nrow() : C->ncol();
```

psi4/src/psi4/libmints/matrix.cc
andyj10224: I added some comments :)

psi4/src/psi4/libmints/matrix.cc
andyj10224: Thank you for the suggestion, this indeed makes the code more concise and readable

PeterKraus (approved):
Thanks for the comments and the patch. I am not sure comparing timing like this in the test makes much sense, as you cannot guarantee Psi4 is the only thing running on the machine, but I suppose the difference should be around 10-fold for this test case. LGTM!

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3403,8 +3403,34 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA,
...
                      bool transC) {
-    SharedMatrix T = doublet(A, B, transA, transB);
-    SharedMatrix S = doublet(T, C, false, transC);
+
+    int dim1 = transA ? A->ncol() : A->nrow();
+    int dim2 = transA ? A->nrow() : A->ncol();
+    int dim3 = transB ? B->ncol() : B->nrow();
+    int dim4 = transB ? B->nrow() : B->ncol();
+    int dim5 = transC ? C->ncol() : C->nrow();
+    int dim6 = transC ? C->nrow() : C->ncol();
```
loriab, line 3412: I suspect `doublet` is calling the symmetry-aware 6-arg form of `gemm` (https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/matrix.cc#L1347), so I wonder if this is ok and if should be calling `max_ncol` etc. instead (https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/matrix.h#L640). better consult @jturney

susilehtola: Good point: the optimized `triplet` should actually contain a loop over irreps and determine the evaluation sequence one irrep at a time, since the sizes of the submatrices may not match.

andyj10224: @susilehtola Awesome! I will work on it tomorrow!

andyj10224: I just made the changes :)

susilehtola: No, that's still just a partial solution. What you need to do is compute the symmetry blocks separately once at a time, since the optimal order `(AB)C` vs `A(BC)` depends on the symmetry block.

susilehtola: i.e. you have the correct cost for each `h`, now what you need is to call `CBLAS` to evaluate the blocks of the matrix transform in the right order.

andyj10224: @susilehtola Ah, I talked to Zach and Lori about this earlier today. We agreed that this would be too much of a hassle. Although I'm not sure about CBLAS... @zachglick @loriab Thoughts?

susilehtola: I don't think it'd be a hassle; the relevant part of `gemm()` is something like 6 lines of code.

loriab: I think it'd need much of https://github.com/psi4/psi4/blob/87b0117c345f257ea763de92d0462e93e82e62d9/psi4/src/psi4/libmints/matrix.cc#L1359-L1378 over single irrep, and we'd want to avoid duplication, I'd guess.

`double` and `triplet` were Rob/Ed (Parrish/Hohenstein) concoctions so the use is heavily c1 codes of scf, sapt, dfmp2. but there are a few mcscf, dct calls will possible symmetry. bottom line, existing scheme is an improvement and is the maximum improvement for most present usages. but if `triplet` is to be more than easy syntax, getting the full efficiency out of it may be worth implementing.

andyj10224: I'll take a stab at it, doesn't seem to be too bad

susilehtola: maybe the `gemm` code can be refactored so that you make a routine that does the `gemm` only for a single irrep, and change the current `gemm` function to calling that in a `for` loop.

andyj10224: I like the idea

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3403,8 +3403,51 @@ SharedMatrix doublet(const SharedMatrix &A, const SharedMatrix &B, bool transA,
...
+    int cost2 = 0;
+
+    for (int h = 0; h < nirrep; h++) {
+
+        int dim1 = transA ? A->colspi(h) : A->rowspi(h);
+        int dim2 = transA ? A->rowspi(h) : A->colspi(h);
+        int dim3 = transB ? B->colspi(h) : B->rowspi(h);
+        int dim4 = transB ? B->rowspi(h) : B->colspi(h);
+        int dim5 = transC ? C->colspi(h) : C->rowspi(h);
+        int dim6 = transC ? C->rowspi(h) : C->colspi(h);
```
jturney, line 3425: These checks still assume `A`, `B`, and `C` have the same symmetry ... determined from `XX->symmetry()`.  The underlying gemm code does know how to handle symmetry properly.  However, if it's too complicated to determine the cost analysis taking into account the different symmetries then I would wrap this check and the multiplications that follow in an if statement that checks for the symmetry of all the matrices.  Then In the case where the symmetries don't match I would default to the original triplet code and leave it to the caller to figure out.

What Lori had referenced [here](https://github.com/psi4/psi4/blob/87b0117c345f257ea763de92d0462e93e82e62d9/psi4/src/psi4/libmints/matrix.cc#L1359-L1378) is the style of what is needed to handle arbitrary symmetry.

tests/pytests/test_triplet.py
```diff
@@ -0,0 +1,44 @@
...
+
+    start = time.time()
+    S = psi4.core.doublet(A, B, False, False)
+    S = psi4.core.doublet(S, C, False, False)
+    time2 = time.time() - start
+
+    assert (time1 < time2)
+    assert psi4.compare_matrices(R, S, 8, "linalg::triplet speedup test")
+
+@pytest.mark.quick
```
dgasmith, line 29: Make sure to add test case with irreps. Especially cases where irrep sizes are zero.

dgasmith (commented):
And thus the start of `opt_einsum` in C++ is born!

PS: Joking!

PPS: I hope?

tests/pytests/test_triplet.py
loriab: I noticed opt_einsum is among the deps of the open-source formulation of adcc (https://anaconda.org/adcc/adcc/files), so now it's in the psi4 dependency stream.

psi4/src/psi4/libmints/matrix.cc
loriab: Andy, you may want to consider adding the extra symmetry check mentioned above and the extra tests mentioned below and wrapping up the PR there. It's still an improvement over the naive existing triplet, and per-irrep multiplication and reassembly could be revisited later.

andyj10224: Got it!

andyj10224: PR ready for review

tests/pytests/test_triplet.py
loriab: good to add an irreped case here, since that figures in to your (AB)C vs A(BC) switch. an example of making such a `Matrix` is https://github.com/psi4/psi4/blob/master/tests/numpy-array-interface/input.dat#L143 .

Also, add the name of this test file https://github.com/psi4/psi4/blob/master/psi4/CMakeLists.txt#L308 so it installs.

andyj10224: Done! :)

loriab (approved):
lgtm, thanks!

dgasmith (approved):

dgasmith:
Really cool, thanks for the PR!

PR #2056, opened 2020-12-09
[MBIS/Charge Partitioning Feature] Added The Calculation of Atomic Volumes

### LINKED ISSUE #2041: [feature request] MBIS radial moments
Taken from the forum: http://forum.psicode.org/t/mbis-radial-moments/1990

"Would it be possible to also compute/output the third radial moment of the atomic electron density, ie eq 24 from this paper?: https://pubs.acs.org/doi/10.1021/acs.jctc.6b00456

This would be really valuable for dispersion C6 coefficients, and should be readily accessible from the MBIS partitioned AIM densities."

@andyj10224 @zachglick  thoughts?

andyj10224:
## Description
Uses MBIS (and future charge partitioning schemes) atomic electron densities to calculate expected atomic volumes (or radial moments). Useful for computing quantities like Hirshfeld ratios.

Note: <r^n> possible, with <r^3> being the default

## Todos

- [x] Accurately calculates Expected Atomic Volumes
- [ ] Calculates an arbitrary <r^n> (Expected Atomic Hyper-Volumes, if <r^3> passes, maybe that won't be necessary, as a quick code check would work)

## Questions
- [x] @jeffschriber Is there an external way to compare the output for my code to some kind of reference to ensure that the answers I'm getting are valid or at least reasonable?

## Checklist
- [x] Preliminary tests passed, MBIS code undamaged (mbis-1, 2, 3, 4, and 5)
- [x] More tests comparing the validity of Expected Volumes

## Status
- [x] Further Information Needed
- [x] Ready for review
- [ ] Ready for merge

closes #2041


jeffschriber:
Yes, it turns out the both volumes and volume ratios are quite easy to get from Horton if a Molden-formatted density is available. I think it would be important to automate the generation of the volume ratios here, meaning that some number of free-atom calculations are required. I'm not sure what the defaults are in Horton, but we'd need to think about how exactly these free-atom calculations should be run (i.e. restricted or unrestricted g.s.).

zachglick:
Great! Table 2 of the MBIS paper contains some reference values that would be good to reproduce.

JonathonMisiewicz:
I'll defer a more thorough review until after GA Tech's pass, but I do want some input on this, to make sure that properties stay sane. A few preliminary comments.

1. "ATOMIC_VOLUMES" computations will fail unless the user has already run "MBIS_CHARGES", since grid_, rho_a_points_, rho_a_0_points_, and distances_ won't even be meaningfully defined. Good luck calling the `blocks` method of `nullptr`.
2. You are making `grid_`, `rho_a_points_`, `rho_a_0_points_`, and `distances_` new members of `PopulationAnalysisCalc`, but these are only needed for MBIS multipoles.

Both of these indicate that MBIS_CHARGES and ATOMIC_VOLUMES are more tightly coupled than can be captured with a model of independent property calculations used throughout the rest of the class. Might I suggest that you move atomic volumes to be computed during an MBIS charge calculation, rather than as an independent property?

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -2267,11 +2272,13 @@ PopulationAnalysisCalc::compute_mbis_multipoles(bool print_output) {
     // => Post-Processing <= //
  
     // Atomic density, as defined in Equation 5 of Verstraelen et al. (Negative to account for negative charge of electron)
```
JonathonMisiewicz, line 2274: Update this comment, since you falsified it.

andyj10224: @JonathonMisiewicz In the near future, I intend to add the BSISA charge partitioning scheme, which may potentially require the grid, etc. to calculate Hirshfeld properties

JonathonMisiewicz:
As a point of organization, keep responses _on_ comments on specific lines of code limited to responses _about_ those comments. Your point about BSISA didn't belong below my point about how a comment needed to be corrected.

I still don't like the idea of adding member variables to `PopulationAnalysisCalc` that are not required for all or even most population analyses. This is how you get the explosion of member variables polluting a certain Psi module. I recommend either making a subclass or demoting those member variables to arguments of a function call inside the parent population analysis. The latter option requires that any further requests for a particular population analysis be done at the time the population analysis is done, rather than reading a new OEProp request. This makes more sense to me.

andyj10224:
> As a point of organization, keep responses _on_ comments on specific lines of code limited to responses _about_ those comments. Your point about BSISA didn't belong below my point about how a comment needed to be corrected.
> 
> I still don't like the idea of adding member variables to `PopulationAnalysisCalc` that are not required for all or even most population analyses. This is how you get the explosion of member variables polluting a certain Psi module. I recommend either making a subclass or demoting those member variables to arguments of a function call inside the parent population analysis. The latter option requires that any further requests for a particular population analysis be done at the time the population analysis is done, rather than reading a new OEProp request. This makes more sense to me.

Actually, I do like your idea for the latter option. Let me ask others what they think.

andyj10224:
> As a point of organization, keep responses _on_ comments on specific lines of code limited to responses _about_ those comments. Your point about BSISA didn't belong below my point about how a comment needed to be corrected.
> 
> I still don't like the idea of adding member variables to `PopulationAnalysisCalc` that are not required for all or even most population analyses. This is how you get the explosion of member variables polluting a certain Psi module. I recommend either making a subclass or demoting those member variables to arguments of a function call inside the parent population analysis. The latter option requires that any further requests for a particular population analysis be done at the time the population analysis is done, rather than reading a new OEProp request. This makes more sense to me.

Okay, I refactored my code to match your suggested changes (passing the former member variables into a function)

psi4/src/read_options.cc
```diff
@@ -266,6 +266,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     /*- Pruning scheme for MBIS Grid -*/
     options.add_str("MBIS_PRUNING_SCHEME", "ROBUST", 
                     "ROBUST TREUTLER NONE FLAT P_GAUSSIAN D_GAUSSIAN P_SLATER D_SLATER LOG_GAUSSIAN LOG_SLATER NONE");
+    /*- Calculates an expected atomic hypervolume given an arbitrary exponent -*/
```
loriab, line 269: might was to explicitly hint "Three provides ordinary volume."

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1997,6 +1998,44 @@ std::tuple<double, double> get_mbis_params(int atomic_num, int shell_num) {
     return params[atomic_num][shell_num];
 }
 
+// A Helper Method That Calculates the Atomic Volume using Atomic Electron Densities derived from Charge Partitioning
+SharedMatrix compute_atomic_volumes(const std::shared_ptr<DFTGrid>& grid, const std::vector<double>& rho_a_points, const std::vector<double>& distances, int num_atoms) {
```
loriab, line 2002: Have you thought about how to route calcs through oeprop to avoid recalcuation and class member data, as Jonathon mentioned? would it be suitable to have both mbis charges and volumes routed through one function, and volumes do extra steps at the end? and if both present, only runs once.

andyj10224: I set it up this way so that future charge partitioning schemes that are implemented can call the same function, allowing for reusability. Nothing is recalculated or copied in this design, as constant references are passed into the function.

tests/mbis-5/input.dat
```diff
@@ -22,6 +22,22 @@ octupoles_ref = psi4.Matrix.from_list([ #TEST
   -2.23989073e-12, -4.92237390e-07, -1.97235542e-07,  4.45294729e-01,
   -9.80994678e-08,  1.24298633e+00]])   #TEST
 
+# Reference Atomic Volumes from Horton
+# Zn 63.44244885428372
+# O  27.11773534802673
+
+avols_ref = psi4.Matrix.from_list([  #TEST
+ [63.44245356],    #TEST
+ [27.11771735]])   #TEST
```
loriab, line 31: any ref data available to check hypervolume?

andyj10224: This may not be necessary if <r^3> code works, and code is written properly.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -2381,6 +2421,15 @@ PopulationAnalysisCalc::compute_mbis_multipoles(bool print_output) {
 
     }
 
+    auto avols = compute_atomic_volumes(grid, rho_a, distances, num_atoms);
+    wfn_->set_array_variable("MBIS ATOMIC VOLUMES", avols);
```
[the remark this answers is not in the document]
andyj10224: I interpreted this to not include the literal exponent in the variable name, and just let the variable name be a generic <r^n>, I don't think that is a good idea. Is that what you meant?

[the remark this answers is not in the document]
loriab: I meant it to be a `R^3` to avoid running successive oeprops with exp=2 then exp=3 and have two different sets of numbers saved to the same qcvar.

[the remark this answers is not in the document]
andyj10224: Ah okay, is there a way to easily allow for formatting in C++/Psi4?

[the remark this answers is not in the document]
andyj10224: I tried std::format, but that is a VERY new feature (C++ 20)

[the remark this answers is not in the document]
loriab: following https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/oeprop.cc#L860 should work. dealing with C++ printing and strings is one of the unkindest transitions from Python to c++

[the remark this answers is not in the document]
andyj10224: Ah, thank you Lori, have to change it tomorrow. AFK right now.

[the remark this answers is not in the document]
andyj10224: Made the changes, also changed names to "Radial Moments" rather than "Atomic Volumes", since this is the language used in the MBIS Paper

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -2381,6 +2421,24 @@ PopulationAnalysisCalc::compute_mbis_multipoles(bool print_output) {
...
+
+    sstream << "MBIS RADIAL MOMENTS <R^" << power << ">";
+
+    auto var_name = sstream.str();
+    
+    wfn_->set_array_variable(var_name, avols);
+
+    auto valence_widths = std::make_shared<Matrix>("MBIS Valence Widths", num_atoms, 1);
+    for (int atom = 0; atom < num_atoms; atom++) {
+        valence_widths->set(atom, 0, Sai[atom][mA[atom]-1]);
```
loriab, line 2438: It may be more convenient to make the first dimension the 1 (https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/oeprop.cc#L1749). then if these are per-atom quantities, can add to list https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L666 so the qcvar comes to user as 1D array. ping me if I'm not being clear.

andyj10224: Well, the MBIS charges are given in the num_atoms x 1 format. I think it's better if we keep everything consistent

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -2266,12 +2306,13 @@ PopulationAnalysisCalc::compute_mbis_multipoles(bool print_output) {
 
     // => Post-Processing <= //
  
-    // Atomic density, as defined in Equation 5 of Verstraelen et al. (Negative to account for negative charge of electron)
```
dgasmith, line 2269: Can you provide a small blurb of why we removed this convention?

dgasmith (approved):
LGTM, looks like a nice change for properties.

psi4/src/psi4/libmints/oeprop.cc
andyj10224: Yes, I removed this to make the code to calculate the radial moments more similar to the original equations (The minus sign may confuse some people). Would you like me to include this comment in the source code?

dgasmith: No, I don't think enough people mess with the code that it will interrupt others. I believe it is generally good to explain the change in PRs when moving conventions so that everyone is aware.

andyj10224: Great, thanks!

loriab (approved):
thanks for the changes and the new functionality! lgtm

zachglick (changes requested):
Looks great, I only have a minor suggestion:

From talking with @jeffschriber, I believe the only possible radial moments that one might need are the 2nd, 3rd, and 4th. As the code is currently written, a user who wanted all three of those values would have to run three redundant SCF and MBIS calculations, each with a different radial moment.

Instead, how about you compute all three of those radial moments (and store the psivars) by default? This doesn't affect the cost---the radial moments are basically free relative to the SCF and MBIS---and it eliminates the need for an extra setting for users to specify.

jeffschriber:
Could also do an option like `MAX_RADIAL_MOMENT`, where all orders up to a given max order are computed just in case someone wants something higher than 4.

andyj10224:
> Could also do an option like `MAX_RADIAL_MOMENT`, where all orders up to a given max order are computed just in case someone wants something higher than 4.

Great suggestions @zachglick @jeffschriber, I will get them changed ASAP

loriab:
>> Could also do an option like MAX_RADIAL_MOMENT, where all orders up to a given max order are computed just in case someone wants something higher than 4.

> Great suggestions @zachglick @jeffschriber, I will get them changed ASAP

If 2, 3, 4 are all that's ever wanted and there's trivial additional cost, I advocate always computing and storing them all. Fewer options, more cheap results always better. Good idea, @zachglick

andyj10224:
> > > Could also do an option like MAX_RADIAL_MOMENT, where all orders up to a given max order are computed just in case someone wants something higher than 4.
> 
> > Great suggestions @zachglick @jeffschriber, I will get them changed ASAP
> 
> If 2, 3, 4 are all that's ever wanted and there's trivial additional cost, I advocate always computing and storing them all. Fewer options, more cheap results always better. Good idea, @zachglick

Updated my code, tests all passed :)

loriab (approved):
lgtm! thanks for more MBIS functionality

jturney (approved):

andysim (approved):
LGTM.  We should probably run oeprop.cc through clang-format, but that can be done another time

PR #2062, opened 2020-12-11
[Improvement] Added Density Screening to Direct SCF Algorithm

andyj10224:
## Description
Implements density screening into the direct SCF Algorithm, no incremental Fock build currently, but incremental Fock build will be a future addition

Reference paper: https://onlinelibrary.wiley.com/doi/abs/10.1002/jcc.540100111

## Notable Features
- [x] Improve the efficiency (less time) of the direct SCF Algorithm without compromising accuracy

## Checklist
- [x] Passes Sanity Check for Water Molecule, will include a Google Sheets link as I include more tests
- [x] Passes Tests Using The Benzene, Naphthalene, etc... series
- [x] Code works with multithreading
- [x] Code works with reaction energies, interaction energies, conformation energies
- [x] Code works on open-shelled systems
- [x] Code tested on a diverse set of molecules

## Status
- [x] Rough Draft Phase
- [x] Ready for review
- [ ] Ready for merge


CDSherrill:
Point of clarification, I don't think density screening will allow "fewer
iterations", so I suggest deleting those words from the "Notable Features"
description.  But hopefully it will lead to "less time" because fewer
integrals will be computed.

susilehtola:
I think this pull needs two options:

1. density screening and
2. incremental formation of Fock matrix

Density screening is still useful even if you don't form the Fock matrix incrementally.

andyj10224:
> I think this pull needs two options:
> 
> 1. density screening and
> 2. incremental formation of Fock matrix
> 
> Density screening is still useful even if you don't form the Fock matrix incrementally.

Yes, I agree. I talked to @andysim and he said to include the iterative Fock build in a separate PR

hokru:
Is this meant additionally to what Libint2 with `update_density()` can provide?

zachglick:
A couple thoughts before a more thorough review:

Can you add some test cases to verify that this doesn't change the SCF energy? That set of test molecules from yesterday would be good, plus a few larger molecules

Could you do some timing to verify that this does in fact speed the SCF code up? A comparison of system size, density threshold, timing, and energy error would be good.

It appears that you're calculating integrals within the master loop and doing redundant work, possibly more than the original algorithm. I think you want to calculate all (NM|NM) shells outside of the loop and use those integrals within the loop.

andyj10224:
> A couple thoughts before a more thorough review:
> 
> Can you add some test cases to verify that this doesn't change the SCF energy? That set of test molecules from yesterday would be good, plus a few larger molecules
> 
> Could you do some timing to verify that this does in fact speed the SCF code up? A comparison of system size, density threshold, timing, and energy error would be good.
> 
> It appears that you're calculating integrals within the master loop and doing redundant work, possibly more than the original algorithm. I think you want to calculate all (NM|NM) shells outside of the loop and use those integrals within the loop.

I will run a variety of tests and include that as a Google Sheets link. And your point about redundancy is good. I will get that fixed.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,65 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
...
+            int num_m = bs1_->shell(M).nfunction();
+
+            int n_start = bs1_->shell(N).function_index();
+            int num_n = bs1_->shell(N).nfunction();
+            
+            double max_dens = 0.0;
+
+            for (int m = m_start; m < m_start + num_m; m++) {
+                for (int n = n_start; n < n_start + num_n; n++) {
+                    double val = fabs(D->get(m, n));
```
susilehtola, line 135:
```suggestion
                    double val = std::abs(D->get(m, n));
```

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,65 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
...
+
+}
+
+bool TwoBodyAOInt::shell_significant_density(int M, int N, int R, int S) {
+
+    Options& options = Process::environment.options;
+    double density_threshold = options.get_double("DENSITY_SCREENING_THRESHOLD");
+
+    double Q_MN = sqrt(shell_pair_values_[N * nshell_ + M]);
+    double Q_RS = sqrt(shell_pair_values_[S * nshell_ + R]);
```
susilehtola, line 152: IMHO the square root should already be taken when `shell_pair_values` is formed; computing square roots for the screening may get expensive.

This would also simplify the screening code elsewhere, as one doesn't need the threshold for the squared quantity.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -129,13 +192,31 @@ bool TwoBodyAOInt::shell_significant_csam(int M, int N, int R, int S) {
...
 }
 
 bool TwoBodyAOInt::shell_significant_schwarz(int M, int N, int R, int S) {
-    return shell_pair_values_[N * nshell_ + M] * shell_pair_values_[R * nshell_ + S] >= screening_threshold_squared_;
+
+    bool dens = true;
+
+    if (do_dens_screen_) dens = shell_significant_density(M, N, R, S);
+
+    return (shell_pair_values_[N * nshell_ + M] * shell_pair_values_[R * nshell_ + S] >= screening_threshold_squared_ && dens);
```
susilehtola, line 208: That is, here: this would be just `screening_threshold_` instead of `screening_threshold_squared_`

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,65 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
...
+                }
+            }
+            max_dens_shell_pair_[M][N] = max_dens;
+        }
+    }
+    timer_off("Density Screen");
+
+}
+
+bool TwoBodyAOInt::shell_significant_density(int M, int N, int R, int S) {
```
susilehtola, line 146:
```suggestion
 bool TwoBodyAOInt::is_shell_density_significant(int M, int N, int R, int S) {
```

psi4/src/psi4/libmints/twobody.h
```diff
@@ -106,6 +107,8 @@ class PSI_API TwoBodyAOInt {
     bool ket_same_;
     /// Are the basis sets in the bra and the ket all the same?
     bool braket_same_;
+    /// Do density screening?
+    bool do_dens_screen_;
```
susilehtola, line 111:
```suggestion
    bool density_screening;
```

psi4/src/psi4/libmints/twobody.h
```diff
@@ -198,6 +205,8 @@ class PSI_API TwoBodyAOInt {
     /*
      * Sieve information
      */
+    /// Update the Max Density Per Shell Pair given an updated Density Matrix
```
susilehtola, line 208: capitalization

andyj10224:
I have included a summary of timings for this project on a Google Sheet:

https://docs.google.com/spreadsheets/d/18-d75mzSILBYby9o3yG2MPRXP8pz9Ou1tLUNfwkc0O4/edit?usp=sharing

psi4/src/psi4/libmints/twobody.cc
andyj10224: I think you are right, a lot fewer integrals are being computed, but the time has not gone down

susilehtola: I mean once you account for the square root in the array

susilehtola:
You still need the square root, but it's enough to do it just once, when the list is formed, by adding
```
shell_max_val = sqrt(shell_max_val)
```
before https://github.com/psi4/psi4/blob/3c9399773ef07fa0ab5298f7028b2f6c82c424df/psi4/src/psi4/libmints/twobody.cc#L307

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -145,29 +147,17 @@ void TwoBodyAOInt::update_density(const SharedMatrix &D) {
...
-        if (!is_pair) val *= 0.25;
-        if (val > max_dens_factor) max_dens_factor = val;
-
-    }
+    double max_dens_factor = max_dens_shell_pair_[M][N];
+    max_dens_factor = std::max(max_dens_factor, max_dens_shell_pair_[R][S]);
+    max_dens_factor = std::max(max_dens_factor, 0.25 * max_dens_shell_pair_[M][R]);
+    max_dens_factor = std::max(max_dens_factor, 0.25 * max_dens_shell_pair_[M][S]);
+    max_dens_factor = std::max(max_dens_factor, 0.25 * max_dens_shell_pair_[N][R]);
+    max_dens_factor = std::max(max_dens_factor, 0.25 * max_dens_shell_pair_[N][S]);
```
susilehtola, line 160: Yeah that loop looked very suspicious to me... this is a compute heavy procedure so you need simple code

andyj10224:
Hey all, I have a master spreadsheet of timings using my new and improved code (1 core):

https://docs.google.com/spreadsheets/d/1Kxii6TauAYj_2rWV3L-iBzyQZUmXDyajAF_rBvVektQ/edit?usp=sharing

andyj10224:
I tested my code on the S22by7 dimer set (1 core, direct SCF only, no df guess) using different screening thresholds

https://docs.google.com/spreadsheets/d/1aja4HIT-jzz8f6vRTX8c1YDMPStpq2TW8lRM6-WiVBk/edit?usp=sharing

andyj10224:
Hey all,

I have compiled a folder of all the important density screening results, with interaction energies, conformation energies (in progress), and reaction energies:

https://drive.google.com/drive/folders/10gbJYR7pzrdkbLy1AcKgRigSikwYIYL4?usp=sharing

Hope this helps!

andyj10224:
> Hey all,
> 
> I have compiled a folder of all the important density screening results, with interaction energies, conformation energies (in progress), and reaction energies:
> 
> https://drive.google.com/drive/folders/10gbJYR7pzrdkbLy1AcKgRigSikwYIYL4?usp=sharing
> 
> Hope this helps!

For reference, the energy values for conformation and reaction energies are in kcal/mol

hokru:
What are your conclusions from the data?

andyj10224:
> What are your conclusions from the data?

Density Screening usually is okay up to 10^-8 screening. Have not found any cases in all my tests where it is not okay. The only place this would be problematic is in a CrystalLattE style application.

andyj10224:
Actually, after reviewing my code, I realized that this density screening only applies in the RHF case, I need to modify the code slightly to make it work with UHF. I derived an expression for the density contribution in the UHF case (the Schwarz contribution will be the same):

[Density Screening UHF.pdf](https://github.com/psi4/psi4/files/5804430/Density.Screening.UHF.pdf)

susilehtola:
> Actually, after reviewing my code, I realized that this density screening only applies in the RHF case, I need to modify the code slightly to make it work with UHF. I derived an expression for the density contribution in the UHF case (the Schwarz contribution will be the same):
> 
> [Density Screening UHF.pdf](https://github.com/psi4/psi4/files/5804430/Density.Screening.UHF.pdf)

Wait, you are screening with respect to the energy contribution? This leads to incorrect results, since the contribution to the Fock matrix can still be non-negligible even if the contribution to the exchange is small. The screening should therefore be based on

1. the Coulomb contribution
2. the alpha and beta exchange contributions

If I remember correctly, many authors suggest doing Coulomb separately from exchange, since the range of the exchange integrals is much smaller than that of the Coulomb, while forming the exchange matrix is more costly than the Coulomb matrix due to the larger amount of permutations involved.

The difference between RHF and UHF screening should be trivial; instead of considering the density matrix |D/2|, you just take max(|Dalpha|, |Dbeta|) as your proxy.

andyj10224:
> > Actually, after reviewing my code, I realized that this density screening only applies in the RHF case, I need to modify the code slightly to make it work with UHF. I derived an expression for the density contribution in the UHF case (the Schwarz contribution will be the same):
> > [Density Screening UHF.pdf](https://github.com/psi4/psi4/files/5804430/Density.Screening.UHF.pdf)
> 
> Wait, you are screening with respect to the energy contribution? This leads to incorrect results, since the contribution to the Fock matrix can still be non-negligible even if the contribution to the exchange energy is small. The screening should therefore be based on
> 
> 1. the Coulomb contribution
> 2. the alpha and beta exchange contributions
> 
> If I remember correctly, many authors suggest doing Coulomb separately from exchange, since the range of the exchange integrals is much smaller than that of the Coulomb, while forming the exchange matrix is more costly than the Coulomb matrix due to the larger amount of permutations involved.
> 
> The difference between RHF and UHF screening should be trivial; instead of considering the density matrix |D/2|, you just take max(|Dalpha|, |Dbeta|) as your proxy.

That makes sense, thank you Susi, I will modify the code (prob tomorrow), and rerun the UHF tests, as well as some RHF tests to make sure the RHF still works

andyj10224:
Hey all, I have updated the code to account for the two density matrices present in UHF.

I ran a couple of tests. UHF does just as well as RHF for similar test cases with respect to screening and parallel performance.

I have also discovered that density screening is not very effective for systems with diffuse basis sets (like aug-cc-pvdz)

andyj10224:
> Hey all, I have updated the code to account for the two density matrices present in UHF.
> 
> I ran a couple of tests. UHF does just as well as RHF for similar test cases with respect to screening and parallel performance.
> 
> I have also discovered that density screening is not very effective for systems with diffuse basis sets (like aug-cc-pvdz)

Link to my tests:

https://docs.google.com/spreadsheets/d/1PDt8LPHo_kWV-ukY3Zm24qJUc-QLf5Cv7CXCKkYzFvs/edit?usp=sharing

andyj10224:
This PR is outdated, new PR is at @2155

PR #2065, opened 2020-12-14
bump resp, add psixas

loriab:
## Todos
- [x] bump RESP interface from 0.8 to 1.0 (this will fix CI)
- [x] add psixas (https://github.com/Masterluke87/psixas) "testing". output is a spectrum file, so hard to test beyond "runs".

## Status
- [x] Ready for review
- [x] Ready for merge


hokru (approved):
Any reason for the detailed output printing in the resp tests? At least the charge difference vectors seem excessive.

andysim (approved):
LGTM!

andysim:
I think the verbose output is ok here, @hokru , because pytest suppresses the output by default.  If the test happens to fail, running that test with output enabled could help with debugging possibly.  I totally agree that the CTests inputs should stay minimal because they ultimately get converted into sample inputs for users to have as starting points for job inputs.  The Windows test failure looks like we need a `using_resp` like mechanism to conditionally activate resp;  I'm sure you have it under control, Lori.

tests/pytests/test_addons.py
```diff
@@ -1103,41 +1103,146 @@ def test_resp():
...
                                    0.803999, -0.661279,  0.453270, -0.600039])
-    assert np.allclose(charges2[0][1], reference_charges2, atol=5e-4)
+    print('Reference RESP Charges')
+    print(reference_charges2)
+    print('Difference')
+    print(charges2[1]-reference_charges2)
+
+    assert np.allclose(charges2[1], reference_charges2, atol=5e-4)
+
+
```
loriab, line 1158:
```suggestion
@pytest.mark.smoke
@using_resp
```

jturney (approved):

PR #2064, opened 2020-12-13
Add set_Lagrangian in Wavefunction

lcyyork:
## Description
Add set_Lagrangian in Wavefunction. Also expose `C_DGEQP3` function. 

## Checklist
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests) `ctest -L quick` all passed.

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2066, opened 2020-12-16
Mark Lagrangian and X as deprecated

lcyyork:
## Description
This PR follows the change of PR #2064.

To avoid compilation errors for any downstream packages, `Lagrangian()` and `X()` in the `Wavefunction` class have been marked deprecated instead of removing them.

Future code should use lowercase `lagrangian()` and `set_lagrangian()` to obtain and set the orbital Lagrangian matrix (both C++ and Python), respectively.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -752,7 +752,7 @@ def _multipole_plumper(compressed, order):
     -------
     complete : ndarray
         Multipole array, order-dimensional Cartesian array expanded to complete components.
-        
+
     """
     shape = tuple([3] * order)
     complete = np.zeros(shape)
@@ -987,6 +987,17 @@ def _core_wavefunction_set_frequencies(cls, val):
 core.Wavefunction.frequencies = _core_wavefunction_frequencies
 core.Wavefunction.legacy_frequencies = _core_wavefunction_legacy_frequencies
 core.Wavefunction.set_frequencies = _core_wavefunction_set_frequencies
+
+
+def _core_wavefunction_X(cls):
+    warnings.warn(
+        "Using `psi4.core.Wavefunction.X` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.4 it will stop working\n",
+        category=FutureWarning,
+        stacklevel=2)
+    return cls.lagrangian()
+
+
+core.Wavefunction.X = _core_wavefunction_X
 
 ## Psi4 v1.3 Export Deprecations
 
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -997,16 +997,7 @@ def _core_wavefunction_X(cls):
     return cls.lagrangian()
 
 
-def _core_wavefunction_Lagrangian(cls):
-    warnings.warn(
-        "Using `psi4.core.Wavefunction.Lagrangian` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.5 it will stop working\n",
-        category=FutureWarning,
-        stacklevel=2)
-    return cls.lagrangian()
-
-
 core.Wavefunction.X = _core_wavefunction_X
-core.Wavefunction.Lagrangian = _core_wavefunction_Lagrangian
 
 ## Psi4 v1.3 Export Deprecations
 
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -991,13 +991,22 @@ core.Wavefunction.set_frequencies = _core_wavefunction_set_frequencies
 
 def _core_wavefunction_X(cls):
     warnings.warn(
-        "Using `psi4.core.Wavefunction.X` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.4 it will stop working\n",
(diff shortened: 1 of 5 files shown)


robertodr (approved):

loriab (commented):
thanks, @lcyyork. 

in my opinion, Py should match C-side, we shouldn't have duplicate fns (`.X()` & `.Lagrangian()`), and (fact) `.hessian()` is already lowercase. so my solution to this setup is:
* delete `.X()` c-side and reattach it py-side redirected to `.lagrangian()` and with a deprecation warning, following this pattern: https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L979-L989
* return `lagrangian()` and `set_lagrangian(X)` to lowercase in wfn.h/cc
* deprecate getter `Lagrangian()` in wavefunction.h/cc following this pattern: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.h#L682-L685 and https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.cc#L1371 to redirect to `lagrangian()`

does this plan seem coherent and covering all principles?

lcyyork:
Hi @loriab, I am not sure why LGTM keeps running for more than an hour. But would you mind to check if this meets the standard?

psi4/driver/p4util/python_helpers.py
```diff
@@ -988,6 +988,17 @@ def _core_wavefunction_set_frequencies(cls, val):
 core.Wavefunction.legacy_frequencies = _core_wavefunction_legacy_frequencies
 core.Wavefunction.set_frequencies = _core_wavefunction_set_frequencies
 
+
+def _core_wavefunction_X(cls):
+    warnings.warn(
+        "Using `psi4.core.Wavefunction.X` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.4 it will stop working\n",
```
loriab, line 994:
```suggestion
        "Using `psi4.core.Wavefunction.X` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.5 it will stop working\n",
```

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -574,10 +574,13 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction> {
...
     SharedMatrix lagrangian() const;
-    /// Returns the Lagrangian in SO basis for the wavefunction
-    SharedMatrix X() const;
     /// Set Lagrangian matrix in SO basis
-    void set_lagrangian(SharedMatrix X) { Lagrangian_ = X; }
+    void set_lagrangian(SharedMatrix X);
+    /// Returns the SO basis Lagrangian
+    PSI_DEPRECATED(
+        "Using `Wavefunction.Lagrangian` instead of `Wavefunction.lagrangian` is deprecated,"
+        " and in 1.4 it will stop working")
```
loriab, line 582:
```suggestion
        " and in 1.5 it will stop working")
```

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -574,10 +574,13 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction> {
 
     /// Returns the SO basis Lagrangian
     SharedMatrix lagrangian() const;
-    /// Returns the Lagrangian in SO basis for the wavefunction
-    SharedMatrix X() const;
```
loriab, line 578: I can't do the code suggestion op on removed lines, I guess, but I think this needs to come back and have the `PSI_DEPRECATED(` too.

loriab (commented):
thanks for doing this. a couple 1.4->1.5 and another deprecation, and I think it's good.

lgtm on c++ always takes about 3h. it's not required for merge, though.

psi4/src/psi4/libmints/wavefunction.h
lcyyork: I see. I have it back. For consistency, I also add the `Lagrangian` to the Py-side marked as deprecated.

psi4/driver/p4util/python_helpers.py
```diff
@@ -991,13 +991,22 @@ def _core_wavefunction_set_frequencies(cls, val):
...
 def _core_wavefunction_X(cls):
     warnings.warn(
-        "Using `psi4.core.Wavefunction.X` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.4 it will stop working\n",
+        "Using `psi4.core.Wavefunction.X` instead of `psi4.core.Wavefunction.lagrangian` is deprecated, and in 1.5 it will stop working\n",
+        category=FutureWarning,
+        stacklevel=2)
+    return cls.lagrangian()
+
+
+def _core_wavefunction_Lagrangian(cls):
```
loriab, line 1000: thanks, but I don't think py-side Lagrangian deprecation is needed since the fn didn't exist two weeks ago: https://github.com/psi4/psi4/blob/v1.4a2/psi4/src/export_wavefunction.cc . am I missing something?

lcyyork: No, my bad. I removed it.

loriab (approved):
great, thanks much!

hokru (approved):

andysim (approved):
Sorry for leading you astray with my comments on the previous PR, and thanks very much for the fix!

PR #2069, opened 2020-12-16
Multipole external potential machinery

andysim:
## Description
In trying to make a good test case for #2068, I found some normalization problems that are fixed in this PR.  Also, the analytic gradients were not implemented, has been addressed.  The interface used here is terrible, but will eventually be merged with #1720 to give a better user experience; this test case will be helpful in making sure everything works correctly during that development.

## Todos

- [x] Add a new feature to allow arbitrary order, arbitrary exponent multipoles to be defined.
- [x] Implement analytic gradients for the above.
- [x] New test case to make sure it all works and to demonstrate the current (terrible and to be replaced) syntax.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


loriab:
if this has fixes, should it be 1.4 instead of 1.5?

andysim:
Oops, I missed your comment. The old `addBasis` still works the same way, and the normalization associated with it is appropriate for external basis functions derived from density fitting.  In QM/MM we normally want some Gaussians at different points in space, and in this case we want unit normalization so that the charges, dipoles, etc. specified don't need to be adjusted.  Essentially there are two use cases, and I just added a new one here.  Not really a bug fix, but also harmless if we want it in 1.4 - entirely up to you.  It gives us a backend that can be tested for the multipole interface at least :)

PR #2072, opened 2020-12-21
OCC Densities to Wavefunction

nlk36701:
## Description
 
Fixes a couple of minor issues for a research project of mine. See todos.

## Todos

- [x] Within the OCC module, updating the opdm to  the wavefunction
- [x] Exposed a function to update the num_frozen_uocc global

## Checklist

- [x] quick tests, occ module tests, standard suite tests

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_wavefunction.cc
+++ b/psi4/src/export_wavefunction.cc
@@ -252,6 +252,7 @@ void export_wavefunction(py::module& m) {
         .def("PCM_enabled", &Wavefunction::PCM_enabled, "Whether running a PCM calculation");
 
     py::class_<scf::HF, std::shared_ptr<scf::HF>, Wavefunction>(m, "HF", "docstring")
+        .def("compute_fvpi", &scf::HF::compute_fvpi, "Update number of frozen virtuals")
         .def("form_C", &scf::HF::form_C, "Forms the Orbital Matrices from the current Fock Matrices.")
         .def("form_initial_C", &scf::HF::form_initial_C,
              "Forms the initial Orbital Matrices from the current Fock Matrices.")
--- a/psi4/src/psi4/libscf_solver/hf.h
+++ b/psi4/src/psi4/libscf_solver/hf.h
@@ -192,7 +192,6 @@ class HF : public Wavefunction {
 
     /// Determine how many core and virtual orbitals to freeze
     void compute_fcpi();
-    void compute_fvpi();
 
     /// Prints the orbitals energies and symmetries (helper method)
     void print_orbital_pairs(const char* header, std::vector<std::pair<double, std::pair<std::string, int>>> orbs);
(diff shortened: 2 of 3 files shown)


JonathonMisiewicz (approved):
For the record, this is a CCQC project, and I was the consultant for adding densities to the wavefunction and changing occ - both things I've gotten experience with, recently.

hokru:
Could you add to the description how correctness was tested?

So writing densities for FCHK files should work with these changes?
So far OCC is turned off now: https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L1984
(Should then probably be enabled in another PR, perhaps with test)

loriab:
Add a test that checks something this fixes, then lgtm. Thanks for the contribution!

loriab:
Please do add or suggest a test for this when you get a chance. If fchk is appropriate to enable, that testing structure is in place.

JonathonMisiewicz:
Between the two of you, the two of us are confused about what kinds of tests you want.

* Holger asked about correctness tests. Lori hasn't raised the issue. This issue wasn't raised during my earlier "density to wavefunction" PRs, so I don't understand why we're doubting its correctness now. This is the same code as I used to bring densities over then. If there is a bug, that's the fault of `occ` and outside the scope of this PR. (And while we _could_ test correctness for the 1RDM, that requires adding occ properties. Which we could do, but is outside the scope of this PR.)
* Holger asked if FCHK can be enabled but said that would be for a different PR. Lori says that may be appropriate for this PR. This is a straight contradiction, and we're confused which you want.
* While it's also possible to test `compute_fvpi`, I cannot recall one other case of a simple Pybind requiring a test, so I'm not sure what we'd be testing. Correctness? That this function can be called Python side?

loriab:
Sorry for the confusion. Right, no testing needed for a pybinding. For the densities, I was looking for "is this code exercised in the test suite?" and "if not, how can we avert someone breaking this later?". If such a test is tricky to construct and you're regarding this as more a better-process than a new-numbers PR, then plausible to wait for any occ bugs to crawl out later.

JonathonMisiewicz:
This isn't a new numbers PR. This is a "make numbers that area already computed accessible" PR.

I know how to test this one: hook up `occ` properties, and add this to `test_dipoles`. I can do that, but that's involved enough to be a separate PR, and one for me, not for Nate.

loriab (approved):

hokru:
> * Holger asked about correctness tests

Sorry if I was unclear. I only wanted a sentence about how correctness was tested (or which other code was followed) for completeness of the PR body. Just in case one has to track something down.
Maybe in this more trivial case it's not needed, I don't think I looked at the code changes then.

JonathonMisiewicz:
Indeed, in this trivial case, it's not needed. All the density-setting side of the code does is take the same operations done on the OEProp object and move them to being done on the density-set wavefunction. If there was a bug in the density, the properties would already have been wrong.

hokru (approved):

andysim (approved):
Weird that the old code was setting the SO density matrix as the MO quantity in OEProp.  This looks much better to me - thanks!

Issue #2046, opened 2020-12-02
Build fails due to missing Eigen3


susilehtola:
The addition of libint2 causes Psi4 to not build. Libint2 is trying to compile the hartree-fock++ test, which depends on Eigen3, which it is not finding for some reason even though I have it installed as a system package.

Libint2 tests should be disabled in the Psi4 build, or the Eigen3 dependency added so that the build works even in the lack of a system package.

(Note: libint2 is also suffering from the -march issue; the libint2 CMake files need to be updated.)

loriab:
> The addition of libint2 causes Psi4 to not build. Libint2 is trying to compile the hartree-fock++ test, which depends on Eigen3, which it is not finding for some reason even though I have it installed as a system package.

Are you using the latest psi4-dev package (.dev22)? That adds the `mpfr` and `eigen` deps to the build environment. I definitely acknowledge that the `Libint2Config.cmake` is not handling `find_dependency` right for eigen and mpfr (really, gmp) yet. I do have cmake edits for those locally, but I have not yet bent cmake to my will. So having the conda packages in place patches the problem. Does your system package have a `Eigen3Config.cmake`? Maybe expand CMAKE_MODULE_PATH?

> Libint2 tests should be disabled in the Psi4 build, or the Eigen3 dependency added so that the build works even in the lack of a system package.

All of Psi4 master is now running Libint2 -- any references to Libint1 or Libint1_t (a hack to allow both to coexist) are only waiting for removal.

> (Note: libint2 is also suffering from the -march issue; the libint2 CMake files need to be updated.)

Thanks for the reminder. I've just edited it locally.

susilehtola:
I am using the latest git snapshot of the master branch.

susilehtola:
And yes, the system package comes with CMake files:
```
/usr/share/cmake/Modules/FindEigen3.cmake
/usr/share/eigen3/Eigen3Config.cmake
/usr/share/eigen3/Eigen3ConfigVersion.cmake
/usr/share/eigen3/Eigen3Targets.cmake
/usr/share/eigen3/UseEigen3.cmake
```

loriab:
> I am using the latest git snapshot of the master branch.

[...]

PR #2073, opened 2021-01-03
Enable single-shot calculations

susilehtola:
## Description
This PR enables easier single-shot calculations in Psi4 by short-circuiting the SCF solver if `maxiter == 0`.

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -339,6 +323,34 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
         self.set_energies("Total Energy", SCFE)
         core.set_variable("SCF ITERATION ENERGY", SCFE)
+
+        return SCFE
+
+    # Single-shot calculation
+    if core.get_option('SCF', 'MAXITER') == 0:
+        form_fock()
```
PeterKraus, line 331: I think perhaps a print_out here may be a good idea, perhaps something like " ==> Non-SCF calculation <==" or something more appropriate. Just so it's super obvious what's happening.

PeterKraus (commented):
A reference output file for the test would be great - I think we're still collecting those?

susilehtola:
> A reference output file for the test would be great - I think we're still collecting those?

I agree; I just am unable to build Psi4 #2046, so I can't get you one for the new feature.

It also looks like the CI system can't build Psi4 at the moment.

loriab (commented):
looks equivalent to me. the output.ref can be generated en masse, but I agree it'd be good to add extra printing when SCF isn't self-consistent field. I'll try to pull this and run the full suite just to make sure.

loriab:
I've tried this out, and tests are clean. However, the output is confusing with no iterations printed. Fortunately, the same effect can be had with `fail_on_maxiter=False`. Then the output is:

```

  ==> Iterations <==

   @RKS iter   0:  -128.84964605391272   -1.28850e+02   1.60374e-02

PsiException: Could not converge SCF iterations in 0 iterations.

  Energy and/or wave function did not converge, but proceeding anyway.
```

Function reorganization is good, imo. I'll suggest edits.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -339,6 +323,34 @@ def scf_iterate(self, e_conv=None, d_conv=None):
...
+        return
+
```
loriab, line 333:
```suggestion
```

tests/scf-singleshot/input.dat
```diff
@@ -0,0 +1,45 @@
...
+E_PBE_minimum = energy('pbe')
+
+# Run HF single-shot
+set maxiter 0
+E_HF_postpbe = energy('scf')
+
+compare_values(refE_HF_minimum,  E_HF_minimum,  8, "HF energy")           # TEST
+compare_values(refE_PBE_posthf,  E_PBE_posthf,  8, "PBE energy, post-HF") # TEST
+compare_values(refE_PBE_minimum, E_PBE_minimum, 8, "PBE energy")          # TEST
+compare_values(refE_HF_postpbe,  E_HF_postpbe,  8, "HF energy, post-PBE") # TEST
```
loriab, line 44:
```suggestion
E_HF_minimum = energy('scf')
compare_values(refE_HF_minimum,  E_HF_minimum,  8, "HF energy")           # TEST

# Run PBE single-shot
set guess read
set maxiter 0
set fail_on_maxiter false
E_PBE_posthf = energy('pbe')
compare_values(refE_PBE_posthf,  E_PBE_posthf,  8, "PBE energy, post-HF") # TEST

# Run PBE self-consistent energy
set maxiter 100
set fail_on_maxiter true
E_PBE_minimum = energy('pbe')
compare_values(refE_PBE_minimum, E_PBE_minimum, 8, "PBE energy")          # TEST

# Run HF single-shot
set maxiter 0
set fail_on_maxiter false
E_HF_postpbe = energy('scf')
compare_values(refE_HF_postpbe,  E_HF_postpbe,  8, "HF energy, post-PBE") # TEST
```

loriab:
[scf-singleshot.ref.txt](https://github.com/psi4/psi4/files/5874423/scf-singleshot.ref.txt)

here's an output.ref if you want to add it. (and check if the fail_on_maxiter trick does all that you needed from the proposed early exit.)

loriab:
Did you get a change to look at this and see if the alternative route does what you need?

susilehtola:
Oh, I see you suggested some code changes. I guess that would do the trick; however, the point of `set maxiter 0` is exactly that one would not need to `set fail_on_maxiter false` since taking 0 SCF iterations means that the orbitals aren't updated. `fail_on_maxiter` should only be checked if `maxiter > 0`!

loriab:
Hmm, the combo `maxiter=0, fail_on_maxiter=false` is averting the naive (or greedy) user who wants the convergence stamp _and_ "minimal" iterations, as opposed to someone who knows what he's doing and using maxiter as a declarative interface. If others like the exception for maxiter=0, I'll go along with it, but my inclination is for consistency with all other maxiter != 0.

susilehtola:
`maxiter = 0` is fundamentally different from `maxiter>0`. Again, in the former no SCF step is even taken; the code should just evaluate the energy.

I think there's a limit to the stupidity we can assume from the users...

loriab:
Ok on the fundamentally different, but I think the place to reroute is https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/scf_proc/scf_iterator.py#L88 by adding a `and core.get_option('SCF', 'MAXITER') > 0:`. That way the iteration printing still happens, and there's a non-converged warning in the output.

JonathonMisiewicz:
Is there any movement on this? If not, I can move this to a 1.5 target.

susilehtola:
@loriab so the issue is that I don't want the orbitals to be changed at all if I call the code with `maxiter=0`.

Otherwise when you evaluate a bunch of energies with various methods, those evaluations will change the orbitals. That's why I short-circuit the run.

JonathonMisiewicz:
This PR will have a merge conflict with my #2247. I wasn't comfortable touching the SCF code this close to 1.4 _then_, and I'm still not comfortable with it now. My vote is to keep this as a 1.5 target.

I would prefer a dedicated option or keyword argument to `energy` for the short-circuit. Yes, this is a _reasonable_ behavior, but:

1. Explicit is better than implicit.
2. The SCF code is too complicated as-it-is due to overloading of the meaning of the `iterations_` variable. It was reasonable in one or two places, but becomes a mess as different parts of the code use it slightly differently. I'm paranoid of variable overloading, and having a special case of `maxiter = 0` is too close to that for comfort.

PR #2076, opened 2021-01-05
[WIP] numerical PCM solver

hokru:
## Description
Adds the option to solve the potential integrals for point charges interacting with the density and the related Vpcm operator numerically.
-> anthracene HF/6-31+G*:  factor 3 faster and parallel. (Makes PCM no longer the bottleneck). delta E(total)=0.11 mEh

_This is work in progress and messy in some places. But general comments welcome. Specific benchmark requests are encouraged._

## Todos

- [ ] spherical and cartesian basis
- [ ] Feature2

## Questions
- [ ] where should this live? 
- [ ] 

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/v.h
+++ b/psi4/src/psi4/libfock/v.h
@@ -161,63 +161,6 @@ class SAP : public VBase {
     void print_header() const override;
 };
 
-// => Derived Classes <= //
-class Num1Int {
-   protected:
-    /// Debug flag
-    int debug_;
-    /// Print flag
-    int print_;
-    /// Number of threads
-    int num_threads_;
-    /// treshold for small densities
-    double density_tolerance_;
-    /// Number of basis functions;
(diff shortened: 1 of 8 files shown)


psi4/src/psi4/libfock/v.h
```diff
@@ -194,5 +194,50 @@ class UV : public VBase {
 
     void print_header() const override;
 };
+
+// => Separate Classes <= //
+class Num1Int {
```
loriab, line 199: I don't have a great feel for C++ style preferences, but I think this class could be a separate file h/cc.

loriab:
this is targeted for v1.5?

hokru:
Target is 1.5 yes. A lot needs to done and tested and I am not pursuing it full time.

PR #2079, opened 2021-01-08
Move DFGradient from dfocc to libmints

JonathonMisiewicz:
## Description
This PR moves the `dfocc` technology for density-fitted gradients out of `dfocc` and into the `libmints` module for use by other modules. Obligatory @jturney and @loriab pings.

**Q:** Can other modules use this code?
**A:** I can personally promise that `dct` will be using this soon, and Forte can use this for its DF-CASSCF gradients.  If you want to use this code, just send me a message! However, the other modules in Psi that currently support DF gradients shouldn't use this. The range-separated gradients mean that SCF can't use this, and DFT gradients are varied enough anyways that we shouldn't try to generalize them. Rob's DFMP2 implementation separates out the Coulomb and Exchange contributions from the derivative. The Coulomb terms allow a performance optimization which prevents this generalized solution from applying.

**Q:** Do we _really_ need to create a `compute_df` function separate from `compute` for gradients?
**A:** I'd say yes, for two reasons. First, in the DF case, it's not obvious which DF integrals you're using (`df_basis_scf` is probably one, but is the other `df_basis_mp`, `df_basis_dct`, `df_basis_cc`?), so you need to explicitly supply this. One can imagine ways to circumvent this, but those require other infrastructure changes. Second, the `compute` function is rather complicated due to its use of three different "algorithms". Combining the two functions would make them more complicated. Simplifying the `compute` function, while an important task and one I will probably need to do before I can fully put my graduate projects behind me, is outside of the scope of this PR. Do you _really_ want to review something even larger?

**Q:** How stable is the interface?
**A:** Changes aren't _necessary_, and I plan to keep this stable at least until after the `dfocc` resync. However, I may make some interface changes if I find they're convenient for non-`dfocc` modules that use this. The big change I'm considering is taking the `dfmp2` approach of including back-transforming the 3-Center density intermediate in the main loop. In that case, the calling code doesn't even need to back-transform the density. If this change works, we can eliminate some code duplication in `dfmp2`, for an added bonus. In this case, all `dfocc` would need to do is stash intermediates it already computes in a place where my code can find them and then delete its back-transformation code.

**Q:** What's this about adding a new block algorithm?
**A:** For large systems, you may not be able to assume that the three-center intermediate (naux*nmo^2) fits in memory. To remedy this, `dfmp2` uses a block algorithm, where it reads only "blocks" of this intermediate at a time. The old DF gradient code was based on Rob's `dfmp2` but in simplifying from Rob's code, had a block algorithm that used unnecessarily many blocks and incurred no memory savings compared to an in-core algorithm. The new code has a `dfmp2`-style out-of-core algorithm.

## Todos
- [x] Migrate DF gradients
- [x] Add block algorithm for the 3-Center term 

## Checklist
- [x] `ctest -L quick` and `ctest -L df[ocm]` both pass 

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
And, the multi-block case still gives the correct answer, so I'm marking this as ready for merge.

The main things I want a second opinion on are the code infrastructure changes and the details of the multi-block algorithm, _just_ in case.

CDSherrill:
@JonathonMisiewicz this sounds very nice!  At an even higher level, what does it do, take a one- and two-PDM (in what format?) and then compute all the derivative integrals and contract them to get the gradient?

JonathonMisiewicz:
@CDSherrill Here's what the caller needs:
* Store the alpha response density `Da_` on the wavefunction
* Store the beta response density `Db_` on the wavefunction
* Store the Lagrangian `Lagrangian_` on the wavefunction
* Store the "metric density" in `PSIF_AO_TPDM` for all relevant DF basis sets (lower triangular format)
* Store the "3-Center density" in `PSIF_AO_TPDM` for all relevant DF basis sets (lower triangular format)

The caller is not responsible for constructing any derivative integrals. The caller is not responsible for constructing an explicit TPDM (all derivative integrals for DF are two or three index, so an explicit TPDM is never needed). The caller is responsible for constructing the densities, as those are method dependent. _In the current implementation_, the caller is responsible for backtransforming all densities involving primary-basis orbitals to the AO basis. For the first three quantities, this is part of the wavefunction contract and requires one line of code. For the 3-Center density, I'm playing with the idea of moving the responsibility of back-transforming to the AO basis to this central code. Rob's code takes that approach.

...Now that I write this up, this would be good to mention in the code itself. I'll add it as a comment before this gets merged in.

jturney (approved):
This looks great!

With the documentation that you intend to add then I think this will be good to go.

JonathonMisiewicz:
Documentation added. I've confirmed per Lori request that standard suite and MP2 pytests pass. Let me know if anything else is necessary.

loriab (approved):
lgtm!

psi4/src/psi4/libmints/deriv.cc
```diff
@@ -385,17 +385,147 @@ Deriv::Deriv(const std::shared_ptr<Wavefunction> &wave, char needed_irreps, bool
...
+            gradient_terms.push_back(kv.second);
+        }
+        auto result = mints->three_idx_grad(aux_datum.first, "3-Center " + aux_datum.second + " Density" , "3-Center " + aux_datum.second);
+        gradient_terms.push_back(result);
+    }
+    _default_psio_lib_->close(PSIF_AO_TPDM, 1); // 1 = keep contents of PSIF_AO_TPDM
+
+    for (auto gradient: gradient_terms) {
+        gradient->symmetrize_gradient(molecule_);
+        gradient->print_atom_vector();
```
dgasmith, line 511: This seems a bit verbose?

psi4/src/psi4/libmints/deriv.cc
```diff
@@ -618,13 +746,13 @@ SharedMatrix Deriv::compute(DerivCalcType deriv_calc_type) {
...
+    opdm_contr->symmetrize_gradient(molecule_);
+    x_contr->symmetrize_gradient(molecule_);
     tpdm_contr_->symmetrize_gradient(molecule_);
 
     enuc->print_atom_vector();
-    opdm_contr_->print_atom_vector();
-    x_contr_->print_atom_vector();
+    opdm_contr->print_atom_vector();
+    x_contr->print_atom_vector();
     tpdm_contr_->print_atom_vector();
```
dgasmith, line 756: Same here.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2397,6 +2397,172 @@ std::map<std::string, SharedMatrix> MintsHelper::metric_grad(std::map<std::strin
...
+        int Pstart = Pstarts[block];
+        int Pstop = Pstarts[block + 1];
+        int NP = Pstop - Pstart;
+
+        int pstart = auxiliary->shell(Pstart).function_index();
+        int pstop = (Pstop == auxiliary->nshell() ? naux : auxiliary->shell(Pstop).function_index());
+        int np = pstop - pstart;
+
+        // Read values from disk. We assume that only the "lower triangle" of (P|mn) is stored.
+        auto temp = std::vector<double>(static_cast<size_t>(np) * ntri);
```
dgasmith, line 2461: Should this be allocated once before the loop and then reset each iteration?

dgasmith (approved):
Awesome! A few minor issues and I wouldn’t mind seeing an output, but looks great!

psi4/src/psi4/libmints/deriv.cc
JonathonMisiewicz: New behavior: For conventional and DF integrals, at print level 1, cdsalc components are not printed, and neither are gradient components. At print level 2, gradient components are printed, _but_ in the DF case, all TEI terms are folded together. At print level 3, cdsalc components are printed, and in the DF case, TEI terms are not folded together.

JonathonMisiewicz:
See the newly generated `output.ref` for user-facing interface at default print level. I've manually checked print levels 2 and 3.

JonathonMisiewicz:
This should be ready to merge. Are we specifically waiting on @dgasmith to say his minor issues are resolved, or is this good to go?

dgasmith:
LGTM, if I approve its a "minor revisions, but I don't need to see it again" :). 

Thanks for the tweaks, the output looks great!

PR #2085, opened 2021-01-16
EOM CC3 bugfix / code cleanup

psi-rking:
## Description
Cleanup EOM CC3 code a bit.

## Todos
- [x] Catchup the code with psifiles, etc., so that EOM_DEBUG toggle can be used
- [x] Change restart behavior in EOM CC3 computations to avoid odd bug.

## Checklist
- [x] No additional tests should be needed.
- [x] Tests pass.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -2024,9 +2024,13 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_int("VECS_PER_ROOT", 12);
         /*- Vectors stored in CC3 computations -*/
         options.add_int("VECS_CC3", 10);
-        /*- Do collapse with last vector? -*/
+        /*- When collapsing Davidson subspace, whether to also include the
+        previous approximate solution (for each root)? This doubles the
+        number of resulting vectors but generally improves convergence. -*/
         options.add_bool("COLLAPSE_WITH_LAST", true);
-        /*- Do collapse with last vector in CC3 iterations? -*/
+        /*- Has the same effect as "COLLAPSE_WITH_LAST" but only in 
+        CC3 computations and after the initial solution of EOM CCSD.
+        May help efficiency, but hazardous when solving for higher roots. -*/
         options.add_bool("COLLAPSE_WITH_LAST_CC3", false);
         /*- Complex tolerance applied in CCEOM computations -*/
         options.add_double("COMPLEX_TOLERANCE", 1E-12);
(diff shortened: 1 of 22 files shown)


psi4/src/read_options.cc
```diff
@@ -2026,6 +2026,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_int("VECS_CC3", 10);
         /*- Do collapse with last vector? -*/
         options.add_bool("COLLAPSE_WITH_LAST", true);
+        /*- Do collapse with last vector in CC3 iterations? -*/
+        options.add_bool("COLLAPSE_WITH_LAST_CC3", false);
```
loriab, line 2030: Would it be useful to describe the method scope of plain `COLLAPSE_WITH_LAST` above or note whether CC3 needs both or just the new one?

loriab (approved):
after typo fix, lgtm!

susilehtola:
Btw according to Henrik Koch's recent paper [doi:10.1021/acs.jctc.0c00686](https://pubs.acs.org/doi/10.1021/acs.jctc.0c00686) the CC3 implementation in Psi4 is slower by an order or two of magnitude than the most efficient implementations in other codes...

psi4/src/read_options.cc
psi-rking: Yes, better to document better - while it is in my mind again to remember and record!  I just added some verbosity.

psi-rking:
That is sad but not too surprising, since I started that EOM CCSD code over 20 years ago, and finished the EOM CC3 code maybe 15 years ago.  Thank you for the reference.  Would be interesting to dig into it and see if the difference is more basic algorithm or implementation.  But to be honest, I can't really see investing my time into that at present.  And, it is also nice to be able to open-shells.

psi-rking:
@loriab  OK, should be good to go now.  Although the 'bash' test started failing for some odd reason, after that keyword fix.

loriab:
ah, you can fix Linux CI by adding `=4` to https://github.com/psi4/psi4/blob/master/.azure-pipelines/azure-pipelines-linux.yml#L85 . I forgot that regular `conda install psi4` has constrained libxc, but CI doesn't. Is that what you meant by "bash test"?

psi-rking:
Yeah, I haven't looked at Azure much ever, but I noticed the error had
something to do with libxc.

jturney (approved):

andysim (approved):
LGTM!

PR #2088, opened 2021-01-20
Create and Python-expose functions for the DF deriv integrals

JonathonMisiewicz:
## Description
Exactly what it says in the title. New `libmints` functions mean a @jturney ping, and new functions messing with integrals mean an @andysim ping.

Also, a positive LoC PR. Sad.

## Todos

- [x] Metric derivatives and derivative of the 3-center integrals are now available to Python

## Checklist
- [x] I've tested that P4N DF-OMP2 gradients using these new functions match `dfocc` gradients to 6 decimal places, but that isn't suitable for the test suite. I'm open to suggestions on tests.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/mintshelper.cc
+++ b/psi4/src/psi4/libmints/mintshelper.cc
@@ -3360,6 +3360,199 @@ std::vector<SharedMatrix> MintsHelper::ao_tei_deriv1(int atom, double omega,
     return grad;
 }
 
+/* 1st and 2nd derivatives of metric in AO basis  */
+
+std::vector<SharedMatrix> MintsHelper::ao_metric_deriv1(int atom, const std::string& aux_name) {
+    std::array<std::string, 3> cartcomp{ {"X", "Y", "Z"} };
+
+    auto aux = get_basisset(aux_name);
+    auto factory = std::make_shared<IntegralFactory>(aux, BasisSet::zero_ao_basis_set(), aux, BasisSet::zero_ao_basis_set());
+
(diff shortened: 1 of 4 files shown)


jturney (approved):

loriab (approved):

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -3360,6 +3360,199 @@ std::vector<SharedMatrix> MintsHelper::ao_tei_deriv1(int atom, double omega,
...
+            int Pcenter = aux->shell(P).ncenter();
+            int Qcenter = aux->shell(Q).ncenter();
+
+            size_t delta;
+
+            delta = 0L;
+
+            if (Pcenter != atom && Qcenter != atom) continue;
+
+            if (Pcenter == atom && Qcenter == atom) continue;
```
andysim, line 3401: A complete brainfart from me, but what's the reason for this check?  I can see that same-center terms with odd angular momentum would integrate to zero, but why the general case?

andysim (approved):
LGTM, after clarification of the confusion on my part.

psi4/src/psi4/libmints/mintshelper.cc
JonathonMisiewicz: The displacement doesn't cause the two orbitals to move relative to each other, because they're on the same atom. The integration is still over all space, so only the relative coordinates matter.

PR #2092, opened 2021-01-25
Deprecate `Matrix::SaveType::Full`

JonathonMisiewicz:
## Description
While working on some more DF technology, I discovered a bug when trying to load using `Matrix::SaveType::Full`. The code assumed the matrix was square. One of the things this PR does is have the code work when that assumption fails.

But the bigger problem is that `Matrix::SaveType::Full` is Psi3 era tech that shouldn't exist anymore. That SaveType reads/writes a matrix _where all elements of the wrong symmetry are replaced with zeros_. The better solution is to use `Matrix::SaveType::SubBlocks`, where those zero's are neither written nor read. Accordingly, I'm deprecating the Full save in favor of SubBlocks save.

## Todos

- [x] `Matrix::SaveType::Full` is deprecated
- [x] Fixed a correctness bug with loading from `Matrix::SaveType::Full`

## Checklist
- [x] quicktests, standard stuite, and dfmp2 all pass
- [ ] I didn't bother adding a new test that I actually fixed the bug, since I accessed the buggy code through a method that's going to be deprecated anyways. Any objections?

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -481,23 +481,22 @@ void Matrix::set(const double *const *const sq) {
         throw PSIEXCEPTION("Matrix::set called on a non-totally symmetric matrix.");
     }
 
-    int h, i, j, ii, jj;
-    int offset;
-
     if (sq == nullptr) {
         throw PSIEXCEPTION("Matrix::set: Set call with a nullptr double** matrix");
     }
-    offset = 0;
-    for (h = 0; h < nirrep_; ++h) {
-        for (i = 0; i < rowspi_[h]; ++i) {
-            ii = i + offset;
-            for (j = 0; j <= i; ++j) {
-                jj = j + offset;
+
+    int row_offset = 0;
+    int col_offset = 0;
+    for (int h = 0; h < nirrep_; ++h) {
+        for (int i = 0; i < rowspi_[h]; ++i) {
+            int ii = i + row_offset;
+            for (int j = 0; j < colspi_[h]; ++j) {
+                int jj = j + col_offset;
                 matrix_[h][i][j] = sq[ii][jj];
-                matrix_[h][j][i] = sq[jj][ii];
             }
         }
-        offset += rowspi_[h];
+        row_offset += rowspi_[h];
+        col_offset += colspi_[h];
     }
 }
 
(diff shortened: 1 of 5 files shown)


loriab (approved):

loriab:
It looks like gcc5.5 has a grievance with the deprecated syntax https://dev.azure.com/psi4/psi4/_build/results?buildId=2687&view=logs&j=258bfeeb-871e-502d-292b-61cd4e6d17f6&t=7c2e2bd2-871d-5948-1d43-9089e6bce09a&l=253 . Looks like it came about in gcc6 https://stackoverflow.com/a/33096368 . I can confirm gcc 7.3 is ok with it.

What do we want to do? Give up the deprecated enum value or bump the gcc requirement?

JonathonMisiewicz:
I'm inclined to force an update, given that GCC5 hasn't been supported for years, but as I'm not familiar with cluster management, my opinion isn't very strong.

loriab:
fwiw, I don't anticipate GT having problems if GCC min boosted. alternately, can do alternate defs of the enum with preprocessor variables.

PeterKraus:
I think Debian Jessie, which is still in Extended LTS period, ships with GCC 4.9. Stretch (which is in LTS) ships with 6.3. Similarly, it looks like everything older than (and including) RHEL7 came with GCC 4.8, but for RHEL6 and RHEL7 their respective DTS packages come with GCC 6.2 or newer.

I vote to deprecate GCC 5 on principle...

jturney (approved):

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -3037,7 +3036,7 @@ void Matrix::load(psi::PSIO *const psio, size_t fileno, SaveType st) {
             psio->read_entry(fileno, name_.c_str(), (char *)fullblock[0], sizeof(double) * sizer * sizec);
 
         set(fullblock);
-        linalg::detail::free(fullblock);
```
robertodr, line 3040: Why the change?

robertodr (approved):

psi4/src/psi4/libmints/matrix.cc
JonathonMisiewicz: LGTM was reporting a mismatch error. I didn't dig into the innards of `linalg::detail::free`, but `freeblock` seems to be the standard way to clear memory allocated with `to_block_matrix`.

loriab:
if you rebase now, Azure should be healed

JonathonMisiewicz:
LGTM seems to be having problems. Not sure if that means we should delay merging this in until we understand why or not.

I need this PR in before I can continue on a bugfix PR, so I'd prefer to have this in ASAP.

loriab:
I'm expecting C++ lgtm to fail for the foreseeable future b/c the additional analysis of libxc is exceeding the 4h time limit. Not sure why it's running out of disk space <3h, but probably not the fault of this PR. will merge.

PR #2093, opened 2021-01-26
DLPNO-MP2

zachglick:
## Description
Adds the DLPNO-MP2 method to Psi4. Callable as `energy('dlpno-mp2')`.

DLPNO-MP2 (or domain-based local pair natural orbital MP2) is the method described in the following paper:

 [Pinski, Peter, et al. "Sparse maps—A systematic infrastructure for reduced-scaling electronic structure methods. I. An efficient and simple linear scaling local MP2 method that uses an intermediate basis of pair natural orbitals." The Journal of chemical physics 143.3 (2015): 034108](https://doi.org/10.1063/1.4926879)
 
DLPNO-MP2 recovers about 99.9% of the DF-MP2 correlation energy, and scales much better in both time and memory. The following graph compares this implementation with Psi4's DF-MP2 code on linear alkanes (18 core i9-10980XE, 200 GB, cc-pVTZ basis):

alkanes

## Todos
- [x] Implement DLPNO-MP2
- [x] Add tests
- [x] Add docs

## Questions
- [x] Are there any other ways that this calculation should be routed? In particular, should `DLPNO` be an option for the `MP2_TYPE` keyword? Should DLPNO be an option for `QC_MODULE`?
- [x] Related to the previous question, suppose a user wants to run a SCS DLPNO-MP2 calculation. How should that be parsed? Not a big deal if that isn't supported.
- [x] How best to handle DLPNO options? There are many thresholds associated with this method, but the error and cost of the method are primarily controlled by just two of them: `T_CUT_DO` and `T_CUT_PNO`. Should the other options be user-facing at all? For now, I've marked them expert. Opinions from anyone with experience running local correlation methods are welcome here.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (commented):
yay for new functionality!

Your mp2_type/qc_module question is an incisive one. Generally, a new approach will show up in either the method name (`energy(methodname-mp2)`) or in the algorithm option (`mp2_type=dlpno`), not both. So algorithm options (disk_df vs. mem_df) are definitely mp2_type. broad categories that give different answers (conv, df, cd) have usually been formuated as mp2_type. methods that give different answers and are known by name (e.g., fno) have been forulated by method name. I can see dlpno going either way. But the most straightforward for now would be to continue following fno, and use energy("dlpno-mp2") and not define a qc_module or mp2_type for dlpno at present.

For conv options, it'd be good to tie those options (if not set explicitly) into the main e_conv, d/r_conv options.

In fnocc, AED allows setting some conv crit by percentage, energy, and cutoff. relevant here?

hokru:
Fantastic!

We want to hook this up to double-hybrid DFT eventually, then a `mp2_type` would come in handy I think.
Needs probably a separate PR.

Users familiar with ORCA will expect the Loose/Normal/TightPNO quick settings.
image

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1498 @@
...
+    ribasis_ = get_basisset("DF_BASIS_MP2");
+}
+
+/* Utility function for easily performing making C_DGESV calls
+ *
+ * C_DGESV solves AX=B for X, given symmetric NxN matrix A and NxM matrix B
+ * B is expected in fortran layout, which complicates the call when (M > 1)
+ * The workaround used here is switching layouts of B before and after the call
+ */
+void C_DGESV_wrapper(SharedMatrix A, SharedMatrix B) {
```
jturney, line 95: This isn't already offered in Matrix?

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1498 @@
...
+
+    SharedMatrix flat = std::make_shared<Matrix>("flattened matrix list", 1, total_size);
+    //outfile->Printf("made it!\n");
+    size_t ind = 0;
+
+    for(SharedMatrix mat : mat_list) {
+        //outfile->Printf("%d %d\n", mat->rowspi(0), mat->colspi(0));
+        for(size_t r = 0; r < mat->rowspi(0); r++) {
+            for(size_t c = 0; c < mat->colspi(0); c++, ind++) {
+                flat->set(0, ind, mat->get(r,c));
```
jturney, line 144: You can directly access the memory pointer to the `flat` and `mat` data and replace these two inner for loops with a `memcpy` call.  That would be more efficient than these for loops.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1498 @@
...
+/* This function is a complement to flatten_mats(). A flattened Matrix is
+ * copied into a list of Matrix objects.
+ */
+void copy_flat_mats(SharedMatrix result, std::vector<SharedMatrix> &mat_list) {
+
+    size_t ind = 0;
+    for(SharedMatrix mat : mat_list) {
+        for(size_t r = 0; r < mat->rowspi(0); r++) {
+            for(size_t c = 0; c < mat->colspi(0); c++, ind++) {
+                mat->set(r, c, result->get(0, ind));
```
jturney, line 163: Same here.  The two inner for loops can be replaced with a memcpy.

jturney (commented):
This is awesome!

There are some places that can be BLAS-ified.  There was one place where you commented on it and a few others.

psi4/src/psi4/dlpno/mp2.cc
zachglick: Good point, thanks!

psi4/src/psi4/dlpno/mp2.cc
zachglick: I didn't see anything like this in the Matrix class. `C_DGESV` isn't used much anywhere.

psi4/src/psi4/dlpno/wrapper.cc
```diff
@@ -0,0 +1,65 @@
...
+namespace psi {
+namespace dlpno {
+
+SharedWavefunction dlpno(SharedWavefunction ref_wfn, Options& options) {
+
+    std::shared_ptr<Wavefunction> dlpno;
+    if (options.get_str("REFERENCE") == "RHF") {
+        dlpno = std::make_shared<DLPNOMP2>(ref_wfn, options);
+    } else {
+        throw PSIEXCEPTION("DLPNO-MP2 requires closed-shell reference: %s");
```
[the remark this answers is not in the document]
zachglick: Oops, good catch

zachglick:
> Fantastic!
> 
> We want to hook this up to double-hybrid DFT eventually, then a `mp2_type` would come in handy I think.
> Needs probably a separate PR.
> 
> Users familiar with ORCA will expect the Loose/Normal/TightPNO quick settings.
> image

Both good ideas! I added a `PNO_CONVERGENCE` keyword that can be set to `LOOSE`, `NORMAL` (default), or `TIGHT`, exactly analogous to Orca. Setting the individual thresholds/cutoffs is still allowed (and overrides `PNO_CONVERGENCE`), but those keywords are marked expert. Connecting this to double-hybrid DFT sounds pretty like a pretty easy follow up.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1498 @@
...
+ * The workaround used here is switching layouts of B before and after the call
+ */
+void C_DGESV_wrapper(SharedMatrix A, SharedMatrix B) {
+
+    int N = B->rowspi(0);
+    int M = B->colspi(0);
+    if(N == 0 || M == 0) return;
+
+    // create a copy of B in fortran ordering
+    double* B_fortran = new double[N * M];
```
susilehtola, line 122: You're leaking memory, since you're not deleting the array. Use `unique_ptr` or switch to `std::vector`.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1498 @@
...
+
+    X = X->get_block({zero, X->rowspi()}, {zero, dim_final});
+    n = n->get_block({zero, dim_final});
+
+    auto S_orth = linalg::triplet(X, S, X, true, false, false);
+
+    for(size_t i = 0; i < nmo_final; ++i) {
+        X->scale_column(0, i, pow(S_orth->get(i,i), -0.5));
+    }
+
```
susilehtola, line 236: This looks redundant; `orthog.cc` has code that gives you orthogonal orbitals.

zachglick:
> yay for new functionality!
> 
> Your mp2_type/qc_module question is an incisive one. Generally, a new approach will show up in either the method name (`energy(methodname-mp2)`) or in the algorithm option (`mp2_type=dlpno`), not both. So algorithm options (disk_df vs. mem_df) are definitely mp2_type. broad categories that give different answers (conv, df, cd) have usually been formuated as mp2_type. methods that give different answers and are known by name (e.g., fno) have been forulated by method name. I can see dlpno going either way. But the most straightforward for now would be to continue following fno, and use energy("dlpno-mp2") and not define a qc_module or mp2_type for dlpno at present.
> 
> For conv options, it'd be good to tie those options (if not set explicitly) into the main e_conv, d/r_conv options.
> 
> In fnocc, AED allows setting some conv crit by percentage, energy, and cutoff. relevant here?

Good to know the difference between module and type, I'll leave as `dlpno-mp2` for now.

Per Holger's suggestion, I've added a single `PNO_CONVERGENCE` option that controls the many thresholds and cutoffs specific to this method.

DLPNO's `T_CUT_PNO` keyword is analogous to fnocc's `OCC_TOLERANCE`, both control the truncation of natural orbitals based on occupation number. Keywords to truncate by energy or percentage could also be added, I'll have to see if that's offers any improvement.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1498 @@
...
+        // compute values of each basis function at each point in this block
+        point_funcs[thread]->compute_functions(block);
+
+        // the values we just computed (max_points x max_functions)
+        SharedMatrix point_values = point_funcs[thread]->basis_values()["PHI"]; 
+         
+        std::vector<int> bf_map = block->functions_local_to_global();
+
+        // resize point_values buffer to size of this block
+        SharedMatrix point_values_trim = std::make_shared<Matrix>("blah", npoints_block, nbf_block); // points x bf_block
```
susilehtola: Is name of the matrix descriptive?

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1504 @@
...
+    }
+
+    for(size_t i = 0; i < naocc; i++) {
+        for(size_t j = 0; j < naocc; j++) {
+            DOI_ij->set(i,j,sqrt(DOI_ij->get(i,j)));
+        }
+        for(size_t u = 0; u < nbf; u++) {
+            DOI_iu->set(i,u,sqrt(DOI_iu->get(i,u)));
+        }
+    }
```
susilehtola, line 343: I'm wondering whether you should just add a `sqrt` function to the `Matrix` class and simplify this as e.g.
```
DOI_ij->sqrt();
DOI_iu->sqrt();
```

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1504 @@
...
+                int centerV = basisset_->function_to_center(u);
+                double p_vv = P_i->get(v,v);
+
+                // off-diag pops (p_uv) split between u and v prop to diag pops
+                double p_uv = P_i->get(u,v);
+                mkn_pop[centerU] += p_uv * ((p_uu) / (p_uu + p_vv));
+                mkn_pop[centerV] += p_uv * ((p_vv) / (p_uu + p_vv));
+
+            }
+        }
```
susilehtola, line 511: I'm wondering whether anyone has done this with better methods, e.g. IAO. Mulliken doesn't even have a basis set limit...

susilehtola: The other part of the algorithms should work just the same, regardless of how the partial charge is computed.

hokru: Turbomole's OSV-PNO uses IAOs. Maybe Molpro's variant as well.

ORCA had initially some problems with finding different LMO solutions, e.g. for augmented basis sets, and worked on getting a really robust localizer (augmented hessian algorithm). 

So looking into what LMOs PSI4 should use is something worth spending time on. Long term, not this PR.

susilehtola: IAOs are a simple solution; they're almost a drop-in replacement to Mulliken. One can also come up with other mathematically well-defined partial charges, as we discuss in https://pubs.acs.org/doi/10.1021/ct401016x

doc/sphinxman/source/dlpnomp2.rst
```diff
@@ -0,0 +1,135 @@
...
+TODO
+
+Recommendations
+---------------
+
+Some practical notes on running the code:
+
+* DLPNO-MP2 should be used for large calculations that cannot reasonable be performed
+  with DF-MP2 (i.e. > X basis functions).
+
```
hokru, line 117: Could you please add something along the lines of below (no need to take my exact phrase):

```
* The PNO accuracy can be controlled with the |dlpno__PNO_CONVERGENCE| keyword according to recommendation by Liakos et al. [Liakos:2015:1525|_. For non-covalent interactions ``TIGHT`` is highly recommended. 
```
then http://dx.doi.org/10.1021/ct501129s would need to be added to `bibliography.rst`

psi4/src/psi4/dlpno/mp2.cc
zachglick: That's a cool paper Susi, thanks. IAO charges would be an improvement over Mulliken charges; I know Molpro's related PNO-LMP2 method uses IAO charges for domain construction. There's some IAO/IBO code buried in the `fisapt` module. Moving that code to the globally accessible `Localizer` class would be a worthwhile project.

Thanks for the info about the augmented hessian Holger. I did notice problems with the localized orbitals when using a lot of diffuse functions, so that might be another good extension.

doc/sphinxman/source/dlpnomp2.rst
zachglick: Sure, thanks for the paper

mkrompiec:
@zachglick Nice job! Are you still working on this? How difficult would it be to wire it up with the Python interface, to expose PNOs and integrals in Python?

JonathonMisiewicz:
This PR is absolutely not getting merged in until _after_ the Psi4 1.4 release, and I would much prefer that exposing anything non-essential to the Python layer be relegated to a separate PR. This one will need a very careful review as it is.

zachglick:
Hi @mkrompiec, thanks! I'm still working on local correlation methods in Psi4. This pull request is effectively done, and is only waiting on the official Psi4 1.4 release to be merged. I expect that to happen relatively soon, but I'm not sure of the exact timeline. 

Once this pull request is merged in, it would be straightforward to expose PNOs and integrals to the Python layer. By the way, if you're interested in regular (non-DLPNO) two-electron integrals, those are easily accessible through the `MintsHelper` class, [as is done here.](https://github.com/psi4/psi4numpy/blob/1233d1af2d168f91065e4f7dbb1a96e02b95b6e1/Tutorials/01_Psi4NumPy-Basics/1e_mints-helper.ipynb)

mkrompiec:
Hi @zachglick, that’s great. I know how to get the regular integrals out, it’s the integrals with (DL)PNOs I’m after.

JonathonMisiewicz:
I can start reviewing this once you fix the merge conflicts.

zachglick:
> I can start reviewing this once you fix the merge conflicts.

Merge conflicts have been fixed, have at it.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+    std::vector<SharedVector> lmo_pao_e(naocc);
+
+    for (size_t i = 0; i < naocc; ++i) {
+        R_i.push_back(Vector3(lmo_lmo_dipx->get(i, i), lmo_lmo_dipy->get(i, i), lmo_lmo_dipz->get(i, i)));
+    }
+
+    for (size_t i = 0; i < naocc; ++i) {
+        std::vector<int> pao_inds;
+        for (size_t u = 0; u < nbf; u++) {
+            if (fabs(DOI_iu->get(i, u)) > options_.get_double("T_CUT_DO_PRE")) {
```
jturney, line 325: I'd cache the value of `T_CUT_DO_PRE` outside the naocc for loop otherwise you're doing a lookup naocc * nbf times.

hokru:
Could you add `dlpno-mp2` also to the CBS driver as new method? https://github.com/psi4/psi4/blob/master/psi4/driver/driver_cbs.py#L743

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+    int nshell = basisset_->nshell();
+    int naux = ribasis_->nbf();
+    int nshellri = ribasis_->nshell();
+    int naocc = nalpha_ - basisset_->n_frozen_core(options_.get_str("FREEZE_CORE"), molecule_);
+
+    SharedMatrix C_vir = reference_wavefunction_->Ca_subset("AO", "VIR");
+    SharedVector e_vir = reference_wavefunction_->epsilon_a_subset("AO", "VIR");
+
+    SharedMatrix C_occ = reference_wavefunction_->Ca_subset("AO", "OCC");
+    SharedVector e_occ = reference_wavefunction_->epsilon_a_subset("AO", "OCC");
```
JonathonMisiewicz, line 1096: Of these four variables, the only one you use is `C_occ`. Remove unused variables.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+
+    // Localize active occupied orbitals
+    auto localizer = std::make_shared<BoysLocalizer>(
```
JonathonMisiewicz, line 1099: Superfluous pointer use is a pet peeve of mine. Why is this not just a `BoysLocalizer`?

JonathonMisiewicz, line 1099: Why do we insist on a Boys localizer as opposed to a PM localizer? I understand that the original paper used Boys, but ORCA supports both and lets the user choose, if they want.

JonathonMisiewicz (commented):
My strategy for reviewing this is going to be alternating between your code and the papers until I get a feel for how this method is supposed to work. Based on Slack comments, I assume you want comments as I have them. If you want me to wait until the end, let me know, and I'll withhold my comments until I've made a complete pass.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+    // locality determined via mulliken charges
+
+    lmo_to_ribfs_.resize(naocc);
+    lmo_to_riatoms_.resize(naocc);
+
+    for (size_t i = 0; i < naocc; ++i) {
+        // atomic mulliken populations for this orbital
+        std::vector<double> mkn_pop(natom, 0.0);
+
+        SharedMatrix P_i = reference_wavefunction_->S()->clone();
```
JonathonMisiewicz, line 429:
```suggestion
        auto P_i = reference_wavefunction_->S()->clone();
```

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+        for (size_t u = 0; u < nbf; ++u) {
+            if (fabs(DOI_iu->get(i, u)) > T_CUT_DO) {
+                lmo_to_paos_temp.push_back(u);
+            }
+        }
+
+        // if any PAO on an atom is in the list, we take all of the PAOs on that atom
+        lmo_to_paos_[i] = contract_lists(lmo_to_paos_temp, atom_to_bf_);
+
+        // equivalent to previous map
```
JonathonMisiewicz, line 481: What does equivalent mean here? It can't mean duplicate.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+
+    for (size_t thread = 0; thread < nthread; thread++) {
+        eris[thread] = std::shared_ptr<TwoBodyAOInt>(factory->eri());
+    }
+
+    outfile->Printf("\n  ==> Transforming 3-Index Integrals to LMO/PAO basis <==\n");
+
+    print_integral_sparsity();
+
+    SharedMatrix SC_lmo =
```
JonathonMisiewicz, line 644:
```suggestion
    auto SC_lmo =
```

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+        }      // M loop
+
+        SharedMatrix C_pao_slice = submatrix_rows(C_pao, riatom_to_bfs2[centerQ]);  // TODO: PAO slices
+
+        //// Here we'll refit the coefficients of C_lmo_slice to minimize residual from unscreened orbitals
+        //// This lets us get away with agressive coefficient screening
+        //// Boughton and Pulay 1992 JCC, Equation 3
+
+        // Solve for C_lmo_slice such that S[local,local] @ C_lmo_slice ~= S[local,all] @ C_lmo
+        SharedMatrix C_lmo_slice =
```
JonathonMisiewicz, line 726:
```suggestion
        auto C_lmo_slice =
```

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+            submatrix_rows_and_cols(SC_lmo, riatom_to_bfs1[centerQ], riatom_to_lmos_ext[centerQ]);
+        SharedMatrix S_aa =
```
JonathonMisiewicz, line 728:
```suggestion
        auto S_aa =
```

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1413 @@
...
+        }
+
+    }  // Q loop
+
+    timer_off("(mn|K) -> (ia|K)");
+
+    timer_on("(K|L)");
+
+    // Compute the full metric, don't invert
+    std::shared_ptr<FittingMetric> metric = std::make_shared<FittingMetric>(ribasis_, true);
```
JonathonMisiewicz, line 744: Why do we need a pointer to a `FittingMetric` at all?

JonathonMisiewicz (commented):
Next batch of comments.

Question for Zach: can you point me to exactly which paper defines how PNO's are computed? The LPNO-CEPA paper gives multiple strategies to compute them, and it isn't clear which is used here.

psi4/src/psi4/dlpno/sparse.cc
```diff
@@ -0,0 +1,219 @@
...
+
+#include <tuple>
+#include <algorithm>
+
+namespace psi {
+
+/* Args: sorted lists l1 and l2
+ * Returns: sorted union of l1 and l2
+ */
+std::vector<int> merge_lists(const std::vector<int> &l1, const std::vector<int> &l2) {
```
JonathonMisiewicz, line 39: Would it make sense to use `std::set<int>` instead of `std::vector<int>`? Then this function would just be `std::set_union`, but I don't know if this would create problems elsewhere.

psi4/src/psi4/dlpno/sparse.cc
```diff
@@ -0,0 +1,219 @@
...
+
+    // TODO: runtime is proportional to A_to_y size (system size, O(N))
+    // could maybe reduce to &y size (domain size, O(1)), probably doesn't matter
+    std::vector<int> yA;
+
+    for(int a = 0, y_ind = 0; a < A_to_y.size(); ++a) {
+
+        bool is_a = false;
+        for(auto y_val : A_to_y[a]) {
+            if (y_ind < y.size() && y[y_ind] == y_val) {
```
JonathonMisiewicz, line 83: Is it not more efficient to use `std::find(y.begin(), y.end(), y_val)`?

JonathonMisiewicz (commented):
I'm not convinced `contract_lists` is correct.

psi4/src/psi4/dlpno/sparse.cc
```diff
@@ -0,0 +1,219 @@
...
+    for(int r_new = 0; r_new < row_inds.size(); r_new++) {
+        int r_old = row_inds[r_new];
+        for(int c = 0; c < mat->colspi(0); c++) {
+            mat_new->set(r_new, c, mat->get(r_old, c));
+        }
+    }
+    return mat_new;
+}
+
+SharedMatrix submatrix_cols(SharedMatrix mat, const std::vector<int> &col_inds) {
```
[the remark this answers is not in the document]
jturney: Can't use `memcpy`. Your best bet is a `C_DCOPY` providing it the proper strides in case BLAS provides an optimized way of doing it.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+    double e_mp2_total = e_scf + e_mp2_corr;
+
+    set_scalar_variable("MP2 CORRELATION ENERGY", e_mp2_corr);
+    set_scalar_variable("CURRENT CORRELATION ENERGY", e_mp2_corr);
+    set_scalar_variable("MP2 TOTAL ENERGY", e_mp2_total);
+    set_scalar_variable("CURRENT ENERGY", e_mp2_total);
+
+    set_scalar_variable("MP2 SINGLES ENERGY", 0.0);
+    set_scalar_variable("MP2 DOUBLES ENERGY", e_mp2_corr);
+    set_scalar_variable("MP2 SAME-SPIN CORRELATION ENERGY", 0.0);      // TODO
```
JonathonMisiewicz, line 1194: Lori, should these variables even be set?

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+        if (i > j) {
+            S_pno_ij_kj[ji] = S_pno_ij_ik[ij];
+            S_pno_ij_ik[ji] = S_pno_ij_kj[ij];
+        }
+    }
+}
+
+void DLPNOMP2::lmp2_iterations() {
+    int n_lmo_pairs = ij_to_i_j.size();
+    int naocc = nalpha_ - basisset_->n_frozen_core(options_.get_str("FREEZE_CORE"), molecule_);
```
JonathonMisiewicz, line 967: Why not get this information from `nfrzc()`? This comment applies everywhere you call `BasisSet.n_frozen_core`.

psi4/src/psi4/dlpno/mp2.h
```diff
@@ -0,0 +1,180 @@
...
+    SharedMatrix e_linear; ///< upper bound to approximate pair energies
+
+    /// LMO/PAO three-index integrals
+    std::vector<SharedMatrix> qia;
+
+    /// pair natural orbitals (PNOs)
+    std::vector<SharedMatrix> K_iajb;  ///< exchange operators (i.e. (ia|jb) integrals)
+    std::vector<SharedMatrix> T_iajb;  ///< amplitudes
+    std::vector<SharedMatrix> Tt_iajb; ///< antisymmetrized amplitudes
+    std::vector<SharedMatrix> R_iajb;  ///< LMP2 residuals
```
JonathonMisiewicz, line 78: While I'm open to second opinions on this, I don't understand why `R_iajb` needs to be a member variable. It makes more sense to me to create the variable during `lmp2_iterations` and pass it by reference to the norm-measuring function. There's no need for this variable to exist outside of `lmp2_iterations`.

By contrast, I can get why exchange integrals or amplitudes would be wavefunction variables.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+    outfile->Printf("    E_CONVERGENCE = %.2e\n", options_.get_double("E_CONVERGENCE"));
+    outfile->Printf("    R_CONVERGENCE = %.2e\n\n", options_.get_double("R_CONVERGENCE"));
+    outfile->Printf("                     Corr. Energy    Delta E     Max R\n");
+
+    R_iajb.resize(n_lmo_pairs);
+
+    int iteration = 0;
+    double e_curr = 0.0, e_prev = 0.0, r_curr = 0.0;
+    bool e_converged = false, r_converged = false;
+    auto diis = std::make_shared<DIISManager>(options_.get_int("DIIS_MAX_VECS"), "LMP2 DIIS", DIISManager::LargestError,
```
JonathonMisiewicz, line 979: Why do we need a pointer here?

psi4/src/psi4/dlpno/mp2.h
```diff
@@ -0,0 +1,180 @@
...
+    SparseMap atom_to_ribf_; ///< which aux BFs are on a given atom?
+    SparseMap atom_to_shell_; ///< which orbital basis shells are on a given atom?
+    SparseMap atom_to_rishell_; ///< which aux basis shells are on a given atom?
+
+    // LMO domains
+    SparseMap lmo_to_ribfs_; ///< which aux BFs are needed for density-fitting a LMO?
+    SparseMap lmo_to_riatoms_; ///< aux BFs on which atoms are needed for density-fitting a LMO?
+    SparseMap lmo_to_paos_; ///< which PAOs span the virtual space of a LMO?
+    SparseMap lmo_to_paoatoms_; ///< PAOs on which atoms span the virtual space of a LMO?
+    std::vector<std::vector<int>> i_j_to_ij; ///< LMO indices (i, j) to LMO pair index (ij)
```
JonathonMisiewicz, line 104: Please include a comment about how you are indexing pairs of _significant_ pairs, and that an insignificant (i, j) hashes to -1.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+                                            T_iajb[ij]->get(a, b));
+                }
+            }
+
+            for (int k = 0; k < naocc; ++k) {
+                int kj = i_j_to_ij[k][j];
+                int ik = i_j_to_ij[i][k];
+
+                if (kj != -1 && i != k && fabs(F_lmo->get(i, k)) > options_.get_double("F_CUT") && n_pno[kj] > 0) {
+                    SharedMatrix temp =
```
JonathonMisiewicz, line 1010:
```suggestion
                    auto temp =
```

psi4/src/psi4/dlpno/mp2.h
```diff
@@ -0,0 +1,180 @@
...
+    void df_ints();
+
+    /// form pair exch operators (EQ 15) and SC amplitudes (EQ 18); transform to PNO basis
+    void pno_transform();
+
+    /// compute PNO/PNO overlap matrices
+    void pno_overlaps();
+
+    /// compute MP2 correlation energy w/ current amplitudes (EQ 14)
+    double eval_amplitudes();
```
JonathonMisiewicz, line 156: Rename this function. When I read this, I thought the amplitudes are the _output_ of your evaluation, not the _input_. Might I suggest `compute_energy_from_amplitudes` or `compute_iteration_energy`?

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+            diis->set_vector_size(1, DIISEntry::Matrix, T_iajb_flat.get());
+        }
+        diis->add_entry(R_iajb_flat, T_iajb_flat);
+        diis->extrapolate(T_iajb_flat);
+        copy_flat_mats(T_iajb_flat, T_iajb);
+
+#pragma omp parallel for schedule(static, 1)
+        for (int ij = 0; ij < n_lmo_pairs; ++ij) {
+            Tt_iajb[ij]->zero();
+            Tt_iajb[ij]->add(T_iajb[ij]);
```
JonathonMisiewicz, line 1061: Why not just `copy`? Less code, and probably faster code, to boot.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+            Tt_iajb[ij]->scale(2.0);
+            Tt_iajb[ij]->subtract(T_iajb[ij]->transpose());
+        }
+
+        outfile->Printf("  @LMP2 iter %3d: %16.12f %10.3e %10.3e\n", iteration, e_curr, e_curr - e_prev, r_curr);
+
+        iteration++;
```
JonathonMisiewicz, line 1068: Can we have a max iterations check? By all accounts, iterative MP2 should converge very fast. But just in case something goes terribly wrong, we should terminate the computation.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+            for (int a = 0; a < n_pno[ij]; ++a) {
+                for (int b = 0; b < n_pno[ij]; ++b) {
+                    T_iajb[ij]->add(a, b, -R_iajb[ij]->get(a, b) / ((e_pno[ij]->get(a) + e_pno[ij]->get(b)) -
+                                                                    (F_lmo->get(i, i) + F_lmo->get(j, j))));
+                }
+            }
+        }
+
+        // DIIS extrapolation
+        SharedMatrix T_iajb_flat = flatten_mats(T_iajb);
```
JonathonMisiewicz, line 1048: I know this is picky, but I'd prefer you flatten to a Vector rather than a Matrix. This object is inherently 1D.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1412 @@
...
+                int ik = i_j_to_ij[i][k];
+
+                if (kj != -1 && i != k && fabs(F_lmo->get(i, k)) > options_.get_double("F_CUT") && n_pno[kj] > 0) {
+                    SharedMatrix temp =
+                        linalg::triplet(S_pno_ij_kj[ij][k], T_iajb[kj], S_pno_ij_kj[ij][k], false, false, true);
+                    temp->scale(-1.0 * F_lmo->get(i, k));
+                    R_iajb[ij]->add(temp);
+                }
+                if (ik != -1 && j != k && fabs(F_lmo->get(k, j)) > options_.get_double("F_CUT") && n_pno[ik] > 0) {
+                    SharedMatrix temp =
```
JonathonMisiewicz, line 1016:
```suggestion
                    auto temp =
```

JonathonMisiewicz (commented):
I finished the `lmp2_iterations` function, and I have a much better feel for how the code works now.

Zach, do you want me to keep going, or do you want to respond to what I've already flooded you with? I'd rather wait for you to catch up, to avoid having many comment chains at once.

zachglick:
> I finished the `lmp2_iterations` function, and I have a much better feel for how the code works now.
> 
> Zach, do you want me to keep going, or do you want to respond to what I've already flooded you with? I'd rather wait for you to catch up, to avoid having many comment chains at once.

Thanks for the super thorough review. You've given me plenty to work through if you want to turn your attention to other things for a bit. I can ping you once I've responded to everything.

psi4/src/psi4/dlpno/mp2.cc
zachglick: I added a `DLPNO_LOCAL_ORBITALS` option that defaults to `BOYS` but may also be `PIPEK_MEZEY`. The localization scheme shouldn't matter much according to Neese and coworkers (https://doi.org/10.1063/1.3086717) who say "The default localization method is the Foster–Boys procedure that we find to be more robust than the Pipek–Mezey method that is also available. However, the results of the calculations depend very little on the actual localization method used"

psi4/src/psi4/dlpno/sparse.cc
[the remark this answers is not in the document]
zachglick: I switched to `C_DCOPY` here. Thankfully, constructing these submatrices is far from the most expensive part of the DLPNO procedure, so there's no big change in performance with or without.

psi4/src/psi4/dlpno/sparse.cc
zachglick: Using `std::set<int>` would also work here, though you'd have to cast to/from `std::vector<int>` or else rework the rest of the sparse map code to use `std::set<int>`. Besides the handy `set_union` function, I don't see any benefits to switching

psi4/src/psi4/dlpno/mp2.h
zachglick: Good point, the residuals should be effectively zero by the end of the iterations, and therefore aren't very useful to keep around. I removed them from the wavefunction.

psi4/src/psi4/dlpno/mp2.cc
zachglick: Sure, I added a `DLPNO_MAXITER` variable and iteration check

psi4/src/psi4/dlpno/mp2.cc
zachglick: I would also have liked to use a `Vector` here, unfortunately the DIIS machinery expects a `Matrix`: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libdiis/diismanager.h

zachglick:
> Could you add `dlpno-mp2` also to the CBS driver as new method? https://github.com/psi4/psi4/blob/master/psi4/driver/driver_cbs.py#L743

I added DLPNO-MP2 to the CBS driver, as well as a CBS DLPNO-MP2 test

psi4/src/psi4/dlpno/mp2.cc
JonathonMisiewicz: Are you sure? If you use [the variadic functions](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libdiis/diismanager.h#L70-L73), you can accept an arbitrary number of arguments and even different DIISEntry type. Example usage [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dct/dct_oo_UHF.cc#L69-L71]). The [DIISEntry::Vector exists](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libdiis/diisentry.h#L55), so vectors should work.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1424 @@
...
+    int nshell = basisset_->nshell();
+    int naux = ribasis_->nbf();
+    int nshellri = ribasis_->nshell();
+    int naocc = nalpha_ - nfrzc();
+
+    auto C_occ = reference_wavefunction_->Ca_subset("AO", "OCC");
+
+    // Localize active occupied orbitals
+    //    auto localizer = BoysLocalizer(basisset_, reference_wavefunction_->Ca_subset("AO", "ACTIVE_OCC"));
+    if (options_.get_str("DLPNO_LOCAL_ORBITALS") == "BOYS") {
```
JonathonMisiewicz, line 1100: Is there a reason why you can't declare a localizer outside of the if statements, define the localizer in the if statements, and call the methods outside the if statements using polymorphism? I know that C++ is picky about types, but I would expect that would still run.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1424 @@
...
+#ifdef _OPENMP
+    nthread = omp_get_max_threads();
+#endif
+    std::vector<std::shared_ptr<BasisFunctions>> point_funcs(nthread);
+    std::vector<SharedMatrix> DOI_ij_temps(nthread);
+    std::vector<SharedMatrix> DOI_iu_temps(nthread);
+    for (size_t thread = 0; thread < nthread; thread++) {
+        point_funcs[thread] = std::make_shared<BasisFunctions>(basisset_, grid.max_points(), nbf);
+        DOI_ij_temps[thread] = std::make_shared<Matrix>("(i,j) Differential Overlap Integrals", naocc, naocc);
+        DOI_iu_temps[thread] = std::make_shared<Matrix>("(i,u) Differential Overlap Integrals", naocc, nbf);
```
JonathonMisiewicz, line 223: These look to me like more completely unnecessary pointers. Are these needed for parallelization, for some reason?

JonathonMisiewicz (commented):
Alright, a few last issues from me. I'll need to look at the `sparse` function one more time, once you've responded to my comments.

There are still questions for Lori about the appropriateness of the QC variables, especially when some of them are marked zero until they're added properly.

psi4/src/psi4/dlpno/sparse.cc
```diff
@@ -0,0 +1,219 @@
...
+    return l12;
+
+}
+
+/* Args: sorted list of y, sparse map from A to y (assume sorted)
+ * Returns: sorted list yA
+ *
+ * For all a, every value in A_to_y[a] is included in yA if at least one is present in y
+ */
+std::vector<int> contract_lists(const std::vector<int> &y, const std::vector<std::vector<int>> &A_to_y) {
```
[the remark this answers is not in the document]
zachglick: I see the source of your confusion. I left out an important detail in the description of `contract_lists`, which is that the `A_to_y` SparseMap is has a very specific form: Every possible value of `y` appears exactly once in this map. (In your counterexample, the value `0` is missing, so the example doesn't apply). Also, the map is sorted such that all elements of `A_to_y[a]` are less than `A_to_y[a+1]`, and all elements within `A_to_y[a]` are sorted.

It's perhaps easier to talk about this map in un-abstract terminology. This map `A_to_y` will always be a map of atoms to basis functions. Every basis function is centered on exactly one atom, and psi4 indexes the basis functions in increasing order, following the atom indices.

psi4/src/psi4/dlpno/sparse.cc
zachglick: When `contract_lists` is called, the list `y` may be very long, but each list `A_to_y[a]` will be small, so it's more efficient to search through `A_to_y[a]` as currently written.

Also in the bigger picture, this part of the code is so fast compared to computing integrals and PNOs, that it doesn't really matter either way.

psi4/src/psi4/dlpno/mp2.cc
zachglick: Good to know, those variadic functions are super neat. I changed the code to use `DIISEntry::Vector`

psi4/src/psi4/dlpno/mp2.cc
zachglick: We need `nthread` of these items so that each thread has one of its own. I changed the `DOI_ij_temps` and `DOI_iu_temps` to use regular `Matrix` objects. make a `BasisFunctions` object without a shared pointer causes a strange compilation error, so I left it as is

psi4/src/psi4/dlpno/mp2.cc
zachglick: You might be more familiar with polymorphism in C++. I tried something like this:
```
    Localizer localizer;
    if (options_.get_str("DLPNO_LOCAL_ORBITALS") == "BOYS") {
        localizer = BoysLocalizer(basisset_, reference_wavefunction_->Ca_subset("AO", "ACTIVE_OCC"));
    } else if (options_.get_str("DLPNO_LOCAL_ORBITALS") == "PIPEK_MEZEY") {
        localizer = PMLocalizer(basisset_, reference_wavefunction_->Ca_subset("AO", "ACTIVE_OCC"));
    }
...
```
and I get the following error:
```
error: object of abstract class type "psi::Localizer" is not allowed:
            function "psi::Localizer::print_header" is a pure virtual function
            function "psi::Localizer::localize" is a pure virtual function
```

Unless you know of some pointer magic to get around declaring an object with an abstract class, it looks like the suggested change might require some modification of the `Localizer` class.

JonathonMisiewicz: That's a good argument for modifying the Localizer class, but that is beyond the scope of this PR.

JonathonMisiewicz (approved):
Two final comments, but otherwise, LGTM.

jturney (approved):
Nice work!

psi4/src/psi4/dlpno/mp2.cc
loriab: follow this pattern, I think: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/dfmp2/mp2.cc#L736-L750 . If dlpno doesn't decompose readily into SS & OS, then don't set those variables.

the block you have here will get them set on the wfn. then in proc.py add this block https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L3932-L3934 so that global P::e has the same info.

loriab (commented):
Looks great!

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -0,0 +1,1425 @@
...
+    outfile->Printf("\n");
+    outfile->Printf("    Auxiliary BFs per Local MO:\n");
+    outfile->Printf("      Average = %4d AUX BFs (%d atoms)\n", total_bfs / naocc, total_atoms / naocc);
+    outfile->Printf("      Min     = %4d AUX BFs (%d atoms)\n", min_bfs, min_atoms);
+    outfile->Printf("      Max     = %4d AUX BFs (%d atoms)\n", max_bfs, max_atoms);
+}
+
+void DLPNOMP2::print_pao_domains() {
+    size_t total_atoms = 0, min_atoms = lmo_to_paoatoms_[0].size(), max_atoms = 0;
+    for (auto atom_list : lmo_to_paoatoms_) {
```
andysim, line 1258: There are a number of `auto` loops like this where the loop variable is read-only.  IMHO it's better to declare them like `const auto &atom_list` when that's the case a) to clearly convey the read-only nature and b) to give the compiler a chance to avoid copies if it's a large object.  I don't think it's necessary to change any of the loops in here, but just wanted to weigh in with that opinion for future reference.

andysim (approved):
I can't remember the details about how the various opposite-spin and same-spin quantities should be reported, so I'll leave it to the other reviewers to comment on that.  This is a really important feature and you've done an outstanding job implementing, documenting, and testing it.  Kudos!

JonathonMisiewicz:
I've spent enough time reviewing `occ` and `dfocc` that I can probably help with OS and SS as needed.

We have three approving reviews at this point, but given the scope of the PR, we should let some time pass before merging this in. If any others want to give this PR a pass before it gets merged in, please say so now.

Zach, when you've responded to Lori and Andy's comments and the PR looks ready to merge, please say so explicitly.

doc/sphinxman/source/dlpnomp2.rst
```diff
@@ -0,0 +1,231 @@
...
+Recommendations
+---------------
+
+Some practical notes on running the code:
+
+* DLPNO-MP2 is not a drop-in replacement for DF-MP2. Instead, it should be used for
+  large calculations that cannot reasonably be performed with DF-MP2. The crossover
+  point between DF-MP2 and DLPNO-MP2 depends on details of both the calculation and
+  the hardware, but can be as low as 2,000 basis functions.
```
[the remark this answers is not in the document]
zachglick: There are no restrictions on `scf_type`. DLPNO-MP2 is run density-fitted regardless of the value of `mp2_type`. Would we want to throw an error (or print a warning) if a user tries to run DLPNO-MP2 with `mp2_type conv`?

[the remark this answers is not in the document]
loriab: yes, please, throw an error if `mp2_type` `has_changed` and equals `CONV`. better to thwart the user than surprise him. :-)

psi4/src/psi4/dlpno/mp2.cc
zachglick: This is a very good point, thanks. Since some of those lists may be very long, it was definitely worth making your suggested change.

zachglick:
Thanks all for the thorough reviews. I've added the OS/SS decomposition of the DLPNO-MP2 correlation energy and verified that SCS-DLPNO-MP2 works correctly.

As far as I know, this PR is ready to be merged.

JonathonMisiewicz:
Lori seems to have approved, and nobody else has put a hold on this PR in the week since I put out the last call, so I don't think there are objections from core devs.

Fix the merge conflict, and I can merge it in once tests pass.

loriab (approved):
lgtm, great work!

JonathonMisiewicz:
@zachglick This PR broke the docs. Docs auto-build says
```
/home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/dlpnomp2.rst:42: WARNING: undefined label: apdx:dlpnomp2

/home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/dlpnomp2.rst:42: WARNING: undefined label: apdx:dlpnomp2_psivar
```

Please submit a new PR to fix this.

loriab:
Check the formatting of the dlpno section start in read_options.cc to check it matches others, even in ways that shouldn't matter. It'll be https://github.com/psi4/psi4/blob/master/doc/sphinxman/document_options_c.pl and document_psivars that are messing up. Let me know if it proves obstinate.

PR #2095, opened 2021-01-26
Fixes get_amplitudes method when requesting frozen-core CC calculations

edeustua:
## Description
This PR introduces a fix to the `get_amplitudes` method used to export T1 and T2 amplitudes to python, which allows the method to work properly when frozen-core coupled-cluster calculations are requested.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/amps.cc
+++ b/psi4/src/psi4/cc/amps.cc
@@ -50,15 +50,15 @@ std::map<std::string, SharedMatrix> CCEnergyWavefunction::get_amplitudes() {
     std::vector<DPDMOSpace> spaces;
     if (dpd_list[0] == nullptr) {
         if (ref == "RHF") {
-            Dimension occpi_ = nalphapi_;
+            Dimension occpi_ = nalphapi_ - frzcpi_;
             Dimension virtpi_ = nmopi_ - nalphapi_;
             cachelist = cacheprep_rhf(options_.get_int("CACHELEVEL"), cachefiles.data());
             spaces = {DPDMOSpace{'o', "ijkl", occpi_}, DPDMOSpace{'v', "abcd", virtpi_}};
         } else if (ref == "ROHF") {
-            Dimension doccpi_ = nalphapi_;
-            Dimension soccpi_ = nbetapi_ - nalphapi_;
+            Dimension doccpi_ = nbetapi_ - frzcpi_;
+            Dimension soccpi_ = nalphapi_ - nbetapi_;
             Dimension occpi_ = doccpi_ + soccpi_;
-            Dimension virtpi_ = (nmopi_ - doccpi_) + soccpi_;
+            Dimension virtpi_ = nmopi_ - (occpi_ + frzcpi_);
             cachelist = cacheprep_rhf(options_.get_int("CACHELEVEL"), cachefiles.data());
             spaces = {DPDMOSpace{'o', "ijkl", occpi_}, DPDMOSpace{'v', "abcd", virtpi_}};
         } else /*UHF*/{
--- a/psi4/src/psi4/cc/amps.cc
+++ b/psi4/src/psi4/cc/amps.cc
@@ -32,19 +32,22 @@
 #include "psi4/libmints/dimension.h"
 #include "psi4/libmints/matrix.h"
 #include "psi4/libpsi4util/process.h"
+#include "psi4/libmints/molecule.h"
 #include <memory>
-namespace psi {
-namespace ccenergy {
+namespace psi::ccenergy {
+
 std::map<std::string, SharedMatrix> CCEnergyWavefunction::get_amplitudes() {
     // re-init dpd
     int** cachelist;
     auto cachefiles = std::vector<int>(PSIO_MAXUNIT);
+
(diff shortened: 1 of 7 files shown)


psi4/src/psi4/cc/amps.cc
```diff
@@ -59,8 +62,8 @@ std::map<std::string, SharedMatrix> CCEnergyWavefunction::get_amplitudes() {
             cachelist = cacheprep_rhf(options_.get_int("CACHELEVEL"), cachefiles.data());
             spaces = {DPDMOSpace{'o', "ijkl", occpi_}, DPDMOSpace{'v', "abcd", virtpi_}};
         } else /*UHF*/{
-            Dimension aoccpi_ = nalphapi_;
-            Dimension boccpi_ = nbetapi_;
+            Dimension aoccpi_ = nalphapi_ - frzcpi_;
+            Dimension boccpi_ = nbetapi_ - frzcpi_;
```
loriab, line 66: good catch! do RHF and ROHF need modification?

edeustua: I would say yes! Will push that.

loriab:
`namespace psi { namespace inner` --> `namespace psi::inner` offends gcc5 but apparently is ok for gcc 6 https://stackoverflow.com/a/39690460 . same question for core as #2092 of syntax vs compatibility.

psi4/src/psi4/cc/amps.cc
```diff
@@ -32,35 +32,38 @@
 #include "psi4/libmints/dimension.h"
 #include "psi4/libmints/matrix.h"
 #include "psi4/libpsi4util/process.h"
+#include "psi4/libmints/molecule.h"
 #include <memory>
-namespace psi {
-namespace ccenergy {
+namespace psi::ccenergy {
```
jturney, line 37: This is technically a C++17 feature.  We only compile Psi4 with C++14 and strict compilers may complain or error.  Please revert to `namespace psi { namespace ccenergy {`

For others: Are we doing this elsewhere in the code?

psi4/src/psi4/cc/amps.cc
jturney: Yes, very good catch!

psi4/src/psi4/libmints/transform.cc
```diff
@@ -80,12 +80,13 @@ void SphericalTransform::init() {
 
     // Go through and grab the values.
     int pureindex = 0;
-    int cartindex = 0;
+    //int cartindex = 0; this is not used
```
jturney, line 83: Nice. Go ahead and delete the lines not used. Thank you of the documentation you added, too!

edeustua: Can I also delete commented out code? It is confusing.

loriab: sure, thanks

psi4/src/psi4/cc/amps.cc
```diff
@@ -32,35 +32,39 @@
...
-            Dimension occpi_ = nalphapi_;
+            Dimension occpi_ = nalphapi_ - frzcpi_;
             Dimension virtpi_ = nmopi_ - nalphapi_;
             cachelist = cacheprep_rhf(options_.get_int("CACHELEVEL"), cachefiles.data());
             spaces = {DPDMOSpace{'o', "ijkl", occpi_}, DPDMOSpace{'v', "abcd", virtpi_}};
         } else if (ref == "ROHF") {
-            Dimension doccpi_ = nalphapi_;
-            Dimension soccpi_ = nbetapi_ - nalphapi_;
+            Dimension doccpi_ = nbetapi_ - frzcpi_;
+            Dimension soccpi_ = nalphapi_ - nbetapi_;
```
susilehtola, line 60:
```suggestion
            Dimension soccpi_ = nalphapi_ - nbetapi_ - frzcpi_;
```

Otherwise `soccpi_` contains the frozen core electrons, which is wrong.

edeustua: I believe this would contain frozen electrons only if the multiplicity is large enough. `nalpha_` and `nbeta_` both count for the frozen electrons, so they cancel out.

edeustua: Maybe an if statement to check that `soccpi_` is not larger than `nalphapi_ - `frzcpi_`?

susilehtola: Ah, nevermind.

Does Psi4 enforce `M >= 1`? If not, `nbetapi_` can be larger than `nalphapi_`

edeustua: I'm not completely sure about that. I assumed that was the case from looking at the HF code, but maybe not. Will ask around.

edeustua: Justin confirms that all unpaired electrons are of alpha spin.

loriab (approved):
lgtm! thanks for investigating and finding a a fix

jturney (approved):

dgasmith (approved):
Thanks, looks great!

PR #2099, opened 2021-02-01
test module plugin on windows

loriab:
## Description
test module plugin on windows

## Status
- [ ] Ready for review
- [ ] Ready for merge


loriab:
plugins on windows could be pursued, but they're not in demand. presently failing on OpenMP detection in plugin cmake configuration.

PR #2100, opened 2021-02-02
Fix DF-DCT

JonathonMisiewicz:
## Description
The primary purpose of this PR is to correct some errors in density-fitted DCT. In brief, none of the nuclear gradient, orbital residual, or amplitude residual could be derived by differentiating the density-fitted DCT energy functional. All deviations were in different ways. This PR changes the energy functional to have a simpler density-fitted energy functional and correct all of its derivatives (orbital, amplitude, and nuclear geometry).

_Unfortunately_, this is a gargantuan PR, and I owe all reviewers some beer. Simplifying some code requires adding new code that best belonged in other modules, there was a lot of new technology needed to get DCT in working order, DCT needed to be hooked up to properties in order to get tests properly working, and I did a lot of code reorganizing to make finding bugs a touch easier for me.

## Todos
- [x] DF-DCT energies, properties, density matrices, and gradients are all now correct
- [x] DCT supports one-electron properties for methods that support gradients
- [x] Adds a new libmints SaveType: ThreeIndexLowerTriangle.
- [x] Adds utility functions to `libdpd` and `libmints` with slightly different arguments/return types.
- [x] Eliminates many unneeded pointers from `dct`.
- [x] clang format DCT
- [x] DCT variables are now written to the wavefunction rather than globals
- [x] Eliminated the special procedure to get a DCT Molden file.
- [ ] There's still some cleanup to do, but it can be a new PR. This one is already very big. (I'm looking at you, incomplete use of the new SaveType, and explicit construction of the VVVV block of the TPDM.)

## Questions
- [ ] Any deprecation warnings necessary? This is an unusual case, to put it lightly.

## Checklist
- [x] Dipole and gradient finite difference tests created and _passing_.
- [x] `dct` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):
Quite the refactoring, thank you!

psi4/src/psi4/dct/dct.h
```diff
@@ -320,9 +338,9 @@ class DCTSolver : public Wavefunction {
     /// The maximum number of IDPs possible for cumulant updates
     int dim_cumulant_;
     /// The lookup array that determines which compound indices belong to orbital IDPs and which don't
-    int *lookup_orbitals_;
+    int* lookup_orbitals_;
```
jturney, line 341: Any reason these int* couldn't be a std::vector?

JonathonMisiewicz: Good call! `std::vector<bool>` is a more sensible type, since the `int` is only ever 0 or 1.

psi4/src/psi4/libdpd/dpd.h
```diff
@@ -344,6 +344,7 @@ class PSI_API DPD {
     int trace42_13(dpdbuf4 *A, dpdfile2 *B, int transb, double alpha, double beta);
 
     int file2_init(dpdfile2 *File, int filenum, int irrep, int pnum, int qnum, const char *label);
+    int file2_init(dpdfile2 *File, int filenum, int irrep, int pnum, int qnum, const std::string& label);
```
jturney, line 347: There shouldn't be the need for both of these `file2_init` since a `char*` can be converted to a `std::string`.

jturney (approved):
Looks good. Took a while to go through. 👍

psi4/driver/procrouting/proc.py
```diff
@@ -1696,18 +1699,48 @@ def run_dct_gradient(name, **kwargs):
...
+        grad = derivobj.compute_df('DF_BASIS_SCF', 'DF_BASIS_DCT')
 
     dct_wfn.set_gradient(grad)
 
     optstash.restore()
     return dct_wfn
 
 
+def run_dct_property(name, **kwargs):
+    """ FUnction encoding sequence of PSI module aclls for
```
hokru, line 1718:
```suggestion
    """ Function encoding sequence of PSI module calls for
```

JonathonMisiewicz:
The last commit also adds `dct` to the properties docstring.

jturney (approved):

psi4/src/psi4/dct/dct.h
```diff
@@ -320,9 +338,9 @@ class DCTSolver : public Wavefunction {
     /// The maximum number of IDPs possible for cumulant updates
     int dim_cumulant_;
     /// The lookup array that determines which compound indices belong to orbital IDPs and which don't
-    int *lookup_orbitals_;
+    std::vector<bool> lookup_orbitals_;
```
andysim, line 341: Be very careful with this data structure - under the hood each element is a single bit, which means it's nicely optimized for size.  However you can get crazy race conditions as a result - if 2 threads are trying to assign different elements within the same byte, trouble can happen.  If you're not writing to this array  in parallel anywhere all is well, but it's worth being aware of.

psi4/src/psi4/dct/dct_compute.cc
```diff
@@ -93,22 +69,24 @@ double DCTSolver::compute_energy() {
         total_energy = compute_energy_UHF();
     }
 
+    // If we have a variational method, the density is free.
+    if (options_.get_bool("OPDM") == true || (orbital_optimized_ && options_.get_str("THREE_PARTICLE") == "NONE")) {
```
andysim, line 73: Just out of interest, would computing the OPDM mess up the 3 particle energy correction?  Probably a stupid question - I'm not familiar with that correction.  Or is this the case that it would be something akin to a CCSD(T) energy that's computed, but the CCSD density is really what's provided?

JonathonMisiewicz: "it would be something akin to a CCSD(T) energy that's computed, but the CCSD density is really what's provided" is exactly right. If you remember the DCFT-06 gradient paper you were third author on, it's exactly the same scenario. The orbitals and amplitudes minimize an energy functional that isn't the one you care about, so computing the OPDM requires nasty response terms.

psi4/src/psi4/dct/dct.h
JonathonMisiewicz: Thanks for the warning! The code to parallelize assignments to this structure are specifically commented out in the existing code, so there are no parallelization bugs in the actually used code. The code in question is "suboptimal" and would need to be revisited to make it useful, anyways.

"suboptimal" = does second-order convergence for DCFT-06 only, which is prone to convergence problems

andysim (approved):
Your reputation for negative LoC bug fixes is well earned!  Thanks for all the effort on this one - it was a beast to look over, so must have been a lot of work to put it together.  LGTM.

PR #2101, opened 2021-02-02
contrive liblink.so

loriab:
## Description
Solely so others can see what a separate SHARED psi4 lib apart from the pb11 psi4 MODULE would look like.

### DIFF
--- a/psi4/src/link.cc
+++ b/psi4/src/link.cc
@@ -1,0 +1,1100 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2021 The Psi4 Developers.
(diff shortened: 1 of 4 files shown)


psi4/src/link.cc
```diff
@@ -0,0 +1,1100 @@
...
+void opt_clean();
+}  // namespace opt
+// Forward declare /src/bin/ methods
+namespace psi {
+
+// Declare some globals
+char* psi_file_prefix;
+std::string outfile_name;
+std::string restart_id;
+std::shared_ptr<PsiOutStream> outfile;
```
jturney, line 152: Can you try:

`PSI_API std::shared_ptr<PsiOutStream> outfile;`

loriab:
closed in favor of #2104

PR #2103, opened 2021-02-03
try shared/module toggle

loriab:
## Description
This is a non-optimal way out of https://github.com/psi4/psi4/pull/2048#issuecomment-771213188 . Non-optimal b/c libpython will be linked in to enable plugins. All indications are it'll work for v2rdm. Should but so far doesn't work for Forte on Azure.

## Todos

- [ ] Feature1
- [ ] Feature2

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -149,6 +149,7 @@ jobs:
         -DENABLE_Libint1t=OFF \
         -DENABLE_snsmp2=OFF \
         -DENABLE_v2rdm_casscf=OFF \
+        -DENABLE_PLUGINS=ON \
         -DENABLE_PLUGIN_TESTING=ON \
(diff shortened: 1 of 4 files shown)


lcyyork:
Hi Lori, thank you for the PR and I really should buy you a beer.

It would be useful to show the linking information, which I do not find under "Build Psi4". I think we should see `Linking CXX shared library core.cpython*.so` if it works as intended. However, I see `Linking CXX shared module core.cpython*.so` [here](https://dev.azure.com/fevange/Forte/_build/results?buildId=64&view=logs&j=969f314f-c02d-5da0-0600-6f21de6a62ef&t=c92288d5-3ca0-581f-2955-a086c2d0fb85&l=3375).

loriab:
good idea to add the `readelf -d`. I'll add it to the next commit on #2104

loriab:
closed in favor of #2104

PR #2106, opened 2021-02-08
Deprecate `legacy_wavefunction` getter and setter

JonathonMisiewicz:
## Description
I was investigating the "memory not being released" problem that @fevangelista brought up at the developer meeting. For the case of a simple HF energy, the wavefunction object indeed persisted after the energy call. Investigation showed its lifetime was controlled by the fact that we had the legacy wavefunction stored, and garbage collection didn't occur until the legacy wavefunction changed.

Seeing as we don't actually use the legacy wavefunction machinery for anything in Psi, except part of a plugin interface where we tell people to prefer explicit wavefunction passing, this PR deprecates the commands for removal in 1.5, so we can collect garbage faster.

This PR doesn't explain Francesco's reports of _increasing_ memory consumption for FINDIF, but it at least is a start.

## Todos
- [x] Deprecates `legacy_wavefunction` and `set_legacy_wavefunction`

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -1021,7 +1021,11 @@ void py_psi_set_n_threads(size_t nthread, bool quiet) {
 
 int py_psi_get_n_threads() { return Process::environment.get_n_threads(); }
 
+PSI_DEPRECATED("Using core.legacy_wavefunction rather than setting return_wfn=True for a computation is deprecated, "
+        "and in 1.5, it will stop working.")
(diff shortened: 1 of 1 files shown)


loriab (approved):
good idea, thanks for investigating.

looks like legacy gradient and legacy frequencies are already warned in python_helpers.py . legacy molecule I need to look into more -- if its uses aren't optking only, it may need to stick around. but yes, the plugin warning has been around for years, so legacy wfn is ready to go.

PeterKraus (approved):

hokru (approved):

PR #2111, opened 2021-02-11
Simplify EFP routines

maxscheurer:
## Description
This PR refactors EFP routines such that they use the mints functions I implemented for PE a while ago.
Less code and a significant speed-up for the electric field integral evaluations.

## Todos

- [x] Move `efp.py` to `solvent` folder
- [x] Refactor functions

## Questions
- [x] Question1

## Checklist
- [x] Tests added for any new features (already present)

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/external/upstream/libxc/CMakeLists.txt
+++ b/external/upstream/libxc/CMakeLists.txt
@@ -25,6 +25,7 @@ else()
                    -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
                    -DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}
                    -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                   -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT}
                    -DNAMESPACE_INSTALL_INCLUDEDIR=/
                    -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
                    # OpenMP irrelevant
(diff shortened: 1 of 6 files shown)


robertodr (approved):

loriab (approved):
lgtm, thank you!

PeterKraus (approved):

external/upstream/libxc/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ else()
                    -DCMAKE_INSTALL_LIBDIR=${CMAKE_INSTALL_LIBDIR}
                    -DCMAKE_INSTALL_BINDIR=${CMAKE_INSTALL_BINDIR}
                    -DCMAKE_INSTALL_INCLUDEDIR=${CMAKE_INSTALL_INCLUDEDIR}
+                   -DCMAKE_OSX_SYSROOT=${CMAKE_OSX_SYSROOT}
```
hokru, line 28: Why the libxc cmake change? (asking because it is unrelated to the PR topic)

maxscheurer: Because previously `OSX_SYSROOT` did not get picked up be the Psi4 ExternalProject & friends, so instead of using the one I defined via `cmake`, it picked up the one from the system. Don't know if there could be any further problems from this change @loriab, @susilehtola?

loriab: There shouldn't be any problems sending `OSX_SYSROOT` through. The need might not have gotten picked up on before b/c internal builds are less common on Macs.

JonathonMisiewicz: FWIW, my experience with Mac builds has been that they haven't worked out-of-the-box for the last couple months. While initial signs pointed to exactly the SDK problem that Max's fix suggests, I've only been able to tame it by passing in system clang rather than conda clang.

PR #2112, opened 2021-02-12
qcschema export bug fix

zachglick:
## Description
Fixes a bug that occurs when one runs `psi4 --qcschema` and requests that the wavefunction is returned. 

Before the orbitals / fock matrix / etc. are returned, all AO-indexed tensors are sorted according to CCA ordering: `Ca` and `Cb` are dimension [AO x MO], so their rows are sorted. `Fa`, `Fb`, `Da`, and `Db` are all dimension [AO x AO], so they are sorted by both row and column. `epsilon_a` and `epsilon_b` are not AO-indexed, so they shouldn't be sorted at all.

As an example, the following HF calculation has fewer MOs than AOs because of linear dependencies, and Psi4 will throw an error when `epsilon_a` is treated as if it were of length AO:
```
import psi4

ret = psi4.schema_wrapper.run_qcschema({
    "molecule": {
        "geometry": [
            0.0, 0.0, -0.1294769411935893,
            0.0, -1.494187339479985, 1.0274465079245698,
            0.0, 1.494187339479985, 1.0274465079245698
        ],  
        "symbols": ["O", "H", "H"],
    },  
    "driver": "energy",
    "model": {
        "method": "hf",
        "basis": "d-aug-cc-pvqz" # larger basis -> more lin deps
    },  
    "keywords": {
        "scf_type": "df",
        "s_tolerance" : 1e-4, # loosen lin dep cutoff
    },  
    "protocols": {
        "wavefunction" : "orbitals_and_eigenvalues",
    }   
    
})
print(ret.error)
```

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -306,13 +306,6 @@ def _convert_wavefunction(wfn, context=None):
                 arr = arr[ao_map[:, None]]
         return arr
 
-    def re1d(mat):
-        arr = np.array(mat)
-        if reorder:
-            arr = arr[ao_map]
-
-        return arr
-
     # get occupations in orbital-energy ordering
     def sort_occs(noccpi, epsilon):
         occs = []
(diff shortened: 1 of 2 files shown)


robertodr (approved):

hokru (approved):

dgasmith (approved):
This makes sense, but I would highly recommend an explicit test that rebuilds various quantities at this point. 

Thanks for finding and fixing!

zachglick:
> This makes sense, but I would highly recommend an explicit test that rebuilds various quantities at this point.
> 
> Thanks for finding and fixing!

That's a good idea, I updated the test

PeterKraus (approved):
Thanks for the extended tests!

PR #2126, opened 2021-03-03
Add polar, rot, and roa to can_do_properties_

bgpeyton:
## Description
~adds electric dipole polarizability, optical rotation, and electric dipole / magnetic dipole tensors to the `can_do_properties_` list so that they are correctly passed into `ret.extras` during a `run_qcshema` call. @loriab told you I'd ping you soon.~

both `omega` and `gauge` are picked up properly from Atomic Input keywords as well. 

UPDATE: `_can_do_properties` was really acting as more of a "default properties" list, so the logic in `run_qcschema` had to be changed instead. 

## Todos
- [x] `_can_do_properties` -> `_default_properties`
- [x] remove can-do filter from `run_qcschema` (kept default properties)
- [x] `run_qcschema` checks for `properties` list in `json['kwargs']` now instead of `json['model']`
- [x] new test  

## Status
- [x] Ready for review
- [x] Ready for merge


PeterKraus:
The [qcschema test](https://github.com/psi4/psi4/blob/master/tests/pytests/test_psi4_qcschema.py) should probably be extended to check for this functionality.

loriab:
If you haven't rebased recently, I wouldn't worry about the Windows error immediately, but Peter's right about tests. The test_psi4_qcschema.py is awkward to work with, so `tests/json/` is a good place to add tests. There's a lot of `run_json` in there to preserve backwards compatibility, but use `run_qcschema` instead.

bgpeyton:
yes, I don't think there's anything testing things which end up in `extras` on the return, so I'll make a quick test. 

while I can accept the windows ctest fail, the failed `test_psi4_qcschema->test_psi4_scfproperty` for linux is a bit concerning. not really sure why it's trying to calculate SCF rotations when it's not in `function_kwargs`. it works fine when putting the properties list directly into the `model` of an `AtomicInput`, so it might be an issue in `qcng.util.model_wrapper` when translating the json data? I'll dig into it

bgpeyton:
all kinds of weirdness going on, let's see if this checks all of the boxes. 

`test_psi4_scfproperty` was failing because the `properties` list was being overwritten, because `run_qcschema` was looking for properties in `model`. normally the calculation would have failed, but `_can_do_properties` was acting as a default list of properties (since it found none) and the test was only looking for a couple of things that happened to be in that list. 

to fix it, `run_qcschema` now correctly checks the kwargs for the property list. `_can_do_properties` was renamed to `_default_properties`, and the function no longer filters the properties through it (if the property is not available, psi4 should fail gracefully). once that was fixed, the "GRID_FIELD" and "GRID_ESP" properties needed to be removed from that test since they failed (could not find some `grid.dat`, I'll let someone else chime in on that one). 

also, added an additional test to hit the qc and wfn response vars (floats and arrays), both present in the `extras` return. I think this is the only test which checks the `extras` dictionary.

bgpeyton:
pinging @loriab for review, not sure who else is intimately familiar with the schema harness (besides DGAS, not sure if he's reviewing lately?)

tests/json/schema-1-response/input.py
```diff
@@ -0,0 +1,124 @@
...
+   'CC2 QUADRUPOLE XZ': 0.0,
+   'CC2 QUADRUPOLE YY': -4.3142280005261755,
+   'CC2 QUADRUPOLE YZ': 0.0,
+   'CC2 QUADRUPOLE ZZ': -5.9972667929848065,
+   'CC2 SPECIFIC ROTATION (MVG) @ 355NM': -0.0,
+   'CC2 SPECIFIC ROTATION (MVG) @ 439NM': -0.0,
+   'CC2 SPECIFIC ROTATION (VEL) @ 355NM': -0.0,
+   'CC2 SPECIFIC ROTATION (VEL) @ 439NM': -0.0,
+   'CC2 DIPOLE': np.array([0, 0, 0.99032208]),
+   'CC2 QUADRUPOLE': np.array([[-5.49672082e+00, -1.36707065e-16,  0.00000000e+00],
```
PeterKraus, line 73: Is this how `yapf` formats those?

psi4/driver/schema_wrapper.py
```diff
@@ -545,14 +545,8 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
...
-        if "properties" in json_data["model"]:
-            kwargs["properties"] = [x.lower() for x in json_data["model"]["properties"]]
-
-            extra = set(kwargs["properties"]) - can_do_properties_
-            if len(extra):
-                raise KeyError("Did not understand property key %s." % kwargs["properties"])
-        else:
-            kwargs["properties"] = list(can_do_properties_)
+        if not "properties" in kwargs:
+            kwargs["properties"] = list(default_properties_)
```
PeterKraus, line 549: What happens if the asked-for property cannot be computed for whatever reason (wrong keyword as well as incompatible method and property)? Do we abort before any calculations are done, or only when we try to compute the properties?

tests/pytests/test_psi4_qcschema.py
```diff
@@ -115,14 +115,14 @@ def test_psi4_scfproperty():
...
...eometry": [0.0, 0.0, -0.12548920214507722, 0.0, -1.6762075450210006, 0.7470892156790888, 0.0, 1.6762075450210006, 0.7470892156790888], "name": "CH2", "molecular_charge": 0.0, "molecular_multiplicity": 3, "masses": [12.0, 1.00782503223, 1.00782503223], "real": [true, true, true], "atom_labels": ["", "", ""], "atomic_numbers": [6, 1, 1], "mass_numbers": [12, 1, 1], "fragments": [[0, 1, 2]], "fragment_charges": [0.0], "fragment_multiplicities": [3], "fix_com": false, "fix_orientation": false, "provenance": {"creator": "QCElemental", "version": "v0.17.0+7.gf55d5ac.dirty", "routine": "qcelemental.molparse.from_string"}}, "driver": "properties", "model": {"method": "scf", "basis": "6-31G**"}, "keywords": {"docc": [2, 0, 0, 1], "e_convergence": 1e-08, "reference": "UHF", "scf_type": "PK", "socc": [1, 0, 1, 0], "function_kwargs": {"properties": ["DIPOLE", "QUADRUPOLE", "MULLIKEN_CHARGES", "LOWDIN_CHARGES", "WIBERG_LOWDIN_INDICES", "MAYER_INDICES", "MAYER_INDICES", "MO_EXTENTS", "ESP_AT_NUCLEI", "MULTIPOLE(5)", "NO_OCCUPATIONS"]}}, "protocols": {}, "extras": {"wfn_qcvars_only": true}, "provenance": {"creator": "Psi4", "version": "1.4a2.dev1087", "routine": "psi4.driver.p4util.procutil"}}"""
```
PeterKraus, line 118: Uh, why don't we just save the json into a separate file and load that, instead of parsing this long string-as-json?

PeterKraus (changes requested):
I am a little confused, as previously `can_do_properties_` also validated the requested properties, and now it trusts the asked-for properties are sensible.

psi4/driver/schema_wrapper.py
bgpeyton: good question, it sort of depends on the method. for cc props,`proc.py` will print `"The following properties are not currently supported: ['PROPERTY']"`. any valid properties in the list are still computed. for scf, it prints `"SCF Property: Feature 'PROPERTY' is not recognized."` and suggests alternatives if possible, but no properties are computed, even the valid ones in the list. CI goes through the `proc_utils.oeprop_validator()`, algebraic diagrammatic construction does something similar to scf, and I don't think DFMP2 checks until somewhere later down the line.  

so as far as the Schema (and whatever is using it, like QCEng) is concerned, I guess the answer is "only when we try to compute the properties," then it's up to `proc.py` to have its own sanity checks depending on the module. I think it generally does a good job, so I think another layer of validating the method/prop combo in the schema wrapper isn't necessary.

tests/pytests/test_psi4_qcschema.py
bgpeyton: not gonna lie, I have no idea. I'm happy to do it if others think it's fine!

tests/json/schema-1-response/input.py
bgpeyton: ah, I forgot about `yapf`. formatted and pushed the change, thanks!

psi4/driver/schema_wrapper.py
PeterKraus: OK, I am convinced that `schema_wrapper` may not be the best place to validate those inputs. 

I do think we need to make sure to "not crash" if it turns out a requested property (via qcschema or other spec) is not supported or is a typo. I want to avoid a situation where one requests a property that may not exist for a particular method, and then **nothing** gets returned instead of at least the stuff that has been computed already.

tests/pytests/test_psi4_qcschema.py
PeterKraus: I think the precedent of using external files for comparison in (py)tests is there - e.g. the `test_fchk_writer.py` looks into its own folder for fchk files. So I'd vote for doing it that way

hokru: There is a trick where you make a folder with the same name as the test and then use the `datadir` fixture at the top of `test_fchk_writer.py`.

bgpeyton:
okay, tests finally passing. thanks to @PeterKraus for the help! if we agree that this isn't the place to ensure soft-fail for unsupported property kwargs, then I think this is ready!

JonathonMisiewicz:
The big picture purpose of this PR is not clear to me. I understand that bullet point 4 is test coverage. Are the other three bullet points changing an interface, adding functionality, or pure code reorganization?

bgpeyton:
@JonathonMisiewicz I would classify this as changing an interface. the wrapper will now push more properties into the `extra` slot of the return for a Schema run, rather than just throwing that information away (since the full wfn is not retrievable by schema). this adds flexibility to the interface w/ e.g. QCEngine and anything else using the wrapper, and also circumvents requiring the MolSSI to re-standardize the QCSchema for esoteric properties before we can return them.

psi4/driver/schema_wrapper.py
```diff
@@ -541,14 +541,8 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
 
     # Handle special properties case
     if json_data["driver"] == "properties":
-        if "properties" in json_data["model"]:
-            kwargs["properties"] = [x.lower() for x in json_data["model"]["properties"]]
```
JonathonMisiewicz, line 545: Do we still need a case change?

JonathonMisiewicz (commented):
Okay, thanks. I have one comment on the actual code before I approve this.

As a general comment, Psi doesn't have the active reviewer base for the assumption "all the reviewers are familiar with this code" to hold. That's why it's always important for the description to supply enough context for the big picture purpose of the code to be clear, even to non-experts. That includes explicitly stating the problems/deficiencies in the current code that the PR solves.

psi4/driver/schema_wrapper.py
bgpeyton: we shouldn't as far as I'm concerned -- the driver passes the kwargs along and the method runs as per usual, including input sanitization. (plus if there's a case issue, it would probably be better to catch it at the driver level so that it's fixed for any kind of input, not just JSON)

JonathonMisiewicz (approved):

PeterKraus (approved):
Thanks for getting the tests sorted.

tests/json/schema-1-response/input.py
```diff
@@ -0,0 +1,87 @@
...
+    filename = request.module.__file__
+    test_dir, _ = os.path.splitext(filename)
+
+    if os.path.isdir(test_dir):
+        dir_util.copy_tree(test_dir, str(tmpdir))
+
+    return tmpdir
+
+
+def test_response(datadir):
```
loriab, line 28: is this test definitely running? it's using pytest fixtures and the `def test_*` collection pattern, but since it's in `tests/json/` rather than `tests/pytests/`, it's being executed by ctest and since everything's behind a `def`, I'm concerned it's a free pass. Maybe pytest is more powerful than I know, so just checking.

tests/json/schema-1-response/input.py
```diff
@@ -0,0 +1,87 @@
...
+        'CC2 DIPOLE', 'CC2 QUADRUPOLE', 'CC2 DIPOLE POLARIZABILITY TENSOR @ 355NM',
+        'CC2 DIPOLE POLARIZABILITY TENSOR @ 439NM', 'CC2 OPTICAL ROTATION TENSOR (MVG) @ 355NM',
+        'CC2 OPTICAL ROTATION TENSOR (MVG) @ 439NM', 'CC2 OPTICAL ROTATION TENSOR (VEL) @ 0NM',
+        'CC2 OPTICAL ROTATION TENSOR (VEL) @ 355NM', 'CC2 OPTICAL ROTATION TENSOR (VEL) @ 439NM',
+        'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 0 @ 355NM',
+        'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 0 @ 439NM',
+        'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ 355NM',
+        'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ 439NM',
+        'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ 355NM',
+        'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ 439NM'
```
loriab, line 77: Is this a good candidate for converting several scalars to a single array? would need to extend some lists/tech in python_helpers.py

tests/json/schema-1-response/input.py
bgpeyton: good catch. CI doesn't run any of these json tests besides energy and gradient. but if you do force test these, the function inside `schema-1-response/input.py` does not get run. I can can pull it out of the function def, axe the fixtures, and put the reference data into the input directly instead of loading it from file. would that be fine?

tests/json/schema-1-response/input.py
bgpeyton: not familiar with the current tech, but probably. extension of the helper functions might be better suited for a separate PR, though.

tests/json/schema-1-response/input.py
loriab: that would be fine. or you can move the file and ref file to tests/pytests/. either way

bgpeyton: done. to make sure it was _actually_ testing, I tried putting a "bad" value in and ran the test, and it failed as expected.

tests/json/schema-1-response/input.py
```diff
@@ -0,0 +1,326 @@
...
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 0 @ 439NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ 355NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ 439NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ 355NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ 439NM'
+]
+for a in arrays:
+    expected_response[a] = np.asarray(expected_response[a])
+
+json_ret = psi4.json_wrapper.run_qcschema(json_data).dict()
```
loriab, line 321:
```suggestion
json_ret = psi4.schema_wrapper.run_qcschema(json_data).dict()
```

tests/json/schema-1-response/input.py
```diff
@@ -0,0 +1,326 @@
...
+    'CC2 OPTICAL ROTATION TENSOR (VEL) @ 355NM', 'CC2 OPTICAL ROTATION TENSOR (VEL) @ 439NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 0 @ 355NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 0 @ 439NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ 355NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 1 @ 439NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ 355NM',
+    'CC2 QUADRUPOLE POLARIZABILITY TENSOR COMPONENT 2 @ 439NM'
+]
+for a in arrays:
+    expected_response[a] = np.asarray(expected_response[a])
```
loriab, line 319:
```suggestion
expected_response = {k: (np.asarray(v) if isinstance(v, List) else v) for k, v in expected_response.items()}
```

tests/pytests/test_psi4_qcschema.py
```diff
@@ -115,16 +148,20 @@ def test_psi4_scfproperty():
...
+    ref_file = datadir.join(f"jatin8.ref")
+    with open(ref_file) as f:
+        jatin = json.load(f)
 
-    atres = psi4.schema_wrapper.run_qcschema(json.loads(jatin))
+    atres = psi4.schema_wrapper.run_qcschema(jatin)
```
loriab, line 164: thanks, it's certainly tidier this way.

psi4/driver/schema_wrapper.py
```diff
@@ -541,14 +541,8 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
...
-
-            extra = set(kwargs["properties"]) - can_do_properties_
-            if len(extra):
-                raise KeyError("Did not understand property key %s." % kwargs["properties"])
-        else:
-            kwargs["properties"] = list(can_do_properties_)
+        if not "properties" in kwargs:
+            kwargs["properties"] = list(default_properties_)
```
loriab, line 545: I guess we won't catch properties like GRID_ESP that can't be returned, but as this is mostly for CC, perhaps that's ok.

tests/json/schema-1-response/input.py
bgpeyton: wow, that's really nifty

psi4/driver/schema_wrapper.py
bgpeyton: yes, there are some edge cases that will probably slip through. maybe we could collect them and patch them through in another way, if anyone is interested in the future.

loriab (approved):
thanks! sorry it's taken me so long to grasp the problem/sol'n in this PR.

tests/json/schema-1-response/input.py
bgpeyton: `List` needs to be `list`, oops. now test is passing again 👍

PR #2127, opened 2021-03-05
Add routine to automate computation of MBIS volume ratios

jeffschriber:
## Description
New MBIS functionality enables computation of atom-in-molecule (AIM) volumes (<r^3>). More often in force field development, a volume ratio between the AIM volume and the free atom is needed (basically describes the difference in atomic volume between a free atom and an atom in a molecule).

This PR automates the computation of MBIS volume ratios by computing free-atom densities at the same level of theory as the molecule. The free-atom computations are on the ground state of a given atom, with the multiplicity taken from the SAD code. These free-atom densities are then run through the MBIS routine, and volumes are computed along with the desired volume ratios.

## Todos

- [X] Automate computation of unique free-atom densities
- [X] Compute volume ratio in MBIS routine

## Questions
- [ ] Running the MBIS code now causes many UHF computations to run, do we need all of the printing here?

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


susilehtola:
> This PR automates the computation of MBIS volume ratios by computing free-atom densities at the same level of theory as the molecule. The free-atom computations are on the ground state of a given atom, with the multiplicity taken from the SAD code.

This also means that the density you'll get is likely symmetry-broken; does your code handle that? The multiplicities may also not be correct.

psi4/driver/p4util/prop_util.py
```diff
@@ -0,0 +1,116 @@
...
+
+        geom = f"""
+0 {int(1+reference_S[a_z])} 
+{a_sym} 0.0 0.0 0.0
+symmetry c1
+"""
+    
+        # make sure we do UHF/UKS if we're not a singlet
+        if reference_S[a_z] != 0:
+            psi4.set_options({"reference":"UHF"})
```
zachglick, line 89: Will this line make all subsequent calculations UHF? An else condition where you explicitly set the reference to RHF might be necessary. It also might be good to reset the reference at the end of this method, like you do with the active molecule

jeffschriber:
> > This PR automates the computation of MBIS volume ratios by computing free-atom densities at the same level of theory as the molecule. The free-atom computations are on the ground state of a given atom, with the multiplicity taken from the SAD code.
> 
> This also means that the density you'll get is likely symmetry-broken; does your code handle that? The multiplicities may also not be correct.

Good points. The code does run UHF/UKS as needed, and the MBIS routine can handle these cases as well. I have a few reference values for free atom volumes for elements H, Li, C, N, O, F, Ne, Si, S, and Cl, and we match these values. So I feel good about getting those ground states correct, but in general they many not be accurate, particularly for transition metals I'd imagine. I am inclined to trust the source of the ground state occupancy values from SAD, though.

psi4/driver/p4util/prop_util.py
jeffschriber: Good catch, definitely should un-set this after the atomic calculations

zachglick:
To cut down on the printing, you could use `core.be_quiet()` and `core.reopen_outfile()` depending on the print level

susilehtola:
> Good points. The code does run UHF/UKS as needed, and the MBIS routine can handle these cases as well. I have a few reference values for free atom volumes for elements H, Li, C, N, O, F, Ne, Si, S, and Cl, and we match these values. So I feel good about getting those ground states correct, but in general they many not be accurate, particularly for transition metals I'd imagine. I am inclined to trust the source of the ground state occupancy values from SAD, though.

Note that SAD hasn't been using those values in a long time; instead it runs spin-restricted calculations. Also, SAD isn't symmetry aware at the moment, which is a problem especially for transition metals and lanthanides and actinides where e.g. the s and d orbitals are very close in energy and the energy differences between different configurations are just eV. The proper way to do it is to impose symmetry, so that the occupations can be controlled. This is the way it is in PySCF; the ground states are from fully numerical calculations I did in https://journals.aps.org/pra/abstract/10.1103/PhysRevA.101.012516

JonathonMisiewicz (commented):
The way I read the conversation thus far, there's some doubt as to the correct "reference number of unpaired electrons", which is distinct from the number of unpaired electrons in the atomic ground state. It isn't clear to me how this is defined.

How serious is this doubt?

jeffschriber:
The reference ground-state multiplicities, particularly for heavier elements, may not be correct. Also there is the larger question of whether to use the broken-symmetry ground-state densities, as is done here, or to use the spin-averaged ones as is done in the SAD code. In the original MBIS paper ([https://pubs.acs.org/doi/pdf/10.1021/acs.jctc.6b00456](url), sec 5), they explore both options, and find that the broken-symmetry, ground-state densities give more accurate dispersion coefficients (computed with the Tkatchenko-Scheffler method, which relies on these volume ratios) with the MBIS method. Given that these volume ratios are used for force field development, I wouldn't be too concerned with potential inaccuracies in the multiplicities of heavier elements.

That said, I think the option to use either the broken-symmetry densities or the spin-averaged ones should be made possible. I think it would likely be cleanest for me to get the densities directly from SAD (since it can handle both cases already) and pass them to the MBIS code, rather than calling `energy` and `oeprop` as I've currently done.

psi4/driver/p4util/prop_util.py
```diff
@@ -0,0 +1,124 @@
...
+                    1, 0,                                                                         1, 2, 3, 2, 1, 0,
+                    1, 0,                                           1, 2, 3, 6, 5, 4, 3, 2, 1, 0, 1, 2, 3, 2, 1, 0,
+                    1, 0,                                           1, 2, 5, 6, 5, 4, 3, 0, 1, 0, 1, 2, 3, 2, 1, 0,
+                    1, 0, 1, 0, 3, 4, 5, 6, 7, 8, 5, 4, 3, 2, 1, 0, 1, 2, 3, 4, 5, 4, 3, 2, 1, 0, 1, 2, 3, 2, 1, 0 ]
+
+
+    # the parent molecule and reference type
+    mol = wfn.molecule()
+    basis = mol.basis_on_atom(0)
+    user_ref = core.get_global_option('REFERENCE')
```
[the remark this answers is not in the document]
JonathonMisiewicz: This line (and the `core.set_global_option` later) shouldn't be needed, now that you're using optstash. Am I missing something?

psi4/driver/p4util/prop_util.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Get unique atoms by input symbol,
+    # Be to handle different basis sets
+    atom_ids = {}
+    for atom in range(mol.natom()):
+        symbol = mol.symbol(atom)
+        Z = int(mol.Z(atom))
+        basis = mol.basis_on_atom(atom)
+    
+        for a, info in atom_ids.items():
```
JonathonMisiewicz, line 75: If I understand this correctly, you only want unique atoms (as defined by symbol and basis), and this `for` loop is supposed to filter out duplicates. This `for` loop doesn't accomplish that, because it only `continue`s the inner loop, not the outer loop.

But in this case, why are you using a `dict` structure for `atom_ids` at all? You don't do anything with the labels. I'd replace your `dict` of `dict`s with a `set` of `tuple` or `collections.namedtuple`, as a clearer data structure. Even a `list` of `dict`s would be cleaner. The advantage of a `set` is that it filters duplicates for you.

psi4/driver/p4util/util.py
```diff
@@ -61,6 +62,16 @@ def oeprop(wfn, *args, **kwargs):
         oe.set_title(kwargs['title'])
     for prop in args:
         oe.add(prop)
+            
+        # If we're doing MBIS, we want the free-atom volumes
+        # in order to compute volume ratios,
+        # but only if we're calling oeprop as the whole molecule
+        free_atom = False
+        if 'free_atom' in kwargs:
+            free_atom = kwargs['free_atom']
```
JonathonMisiewicz, line 71:
```suggestion
        free_atom = kwargs.get('free_atom', False)
```

psi4/driver/p4util/prop_util.py
jeffschriber: Using `set` is a good suggestion, thanks, and a definite blunder on my part for not filtering duplicates on the most recent commit.

JonathonMisiewicz (approved):
Two last minor things:
- `prop_util.py` could use a YAPF pass
- The currently empty `free_atom_volumes` docstring should at least say "Computes MBIS atomic volumes" 

But after that, LGTM.

JonathonMisiewicz:
If this is still a 1.4 target (which I think is completely reasonable), fix up the merge conflicts, and I'll put out the all-call on reviews.

jeffschriber:
Definitely. The only roadblock was in handling all of the printing. I used `be_quiet()` and `reopen_outfile()`, which fail during testing in the windows build. Interestingly, these functions aren't used in the code elsewhere, so they seem not to be windows-safe at all. I think a workaround shouldn't be too hard, but suggestions welcome. The printing of each atomic calculation should never be that excessive (since its only unique elements/basis), so simple solution is to just print all of the atomic outputs

psi4/driver/p4util/prop_util.py
```diff
@@ -0,0 +1,128 @@
...
+                                          elem=[a_sym],
+                                          molecular_charge=0,
+                                          molecular_multiplicity=int(1 + reference_S[a_z]))
+        a_mol.update_geometry()
+        psi4.molutil.activate(a_mol)
+
+        method = theory + "/" + basis
+
+        # Get the atomic wfn
+        at_e, at_wfn = psi4.energy(method, return_wfn=True)
```
loriab, line 110: after my latest driver organization work, I think I'm going to advocate that driver-e/g/h-calling fns like this live a top-level and not in p4util. but fine for now.

psi4/driver/p4util/prop_util.py
```diff
@@ -0,0 +1,128 @@
...
+    psi4.core.print_out(f"  Running {len(unq_atoms)} free-atom UHF computations")
+
+    optstash = optproc.OptionsState(['REFERENCE'])
+    for a_sym, a_z, basis in unq_atoms:
+
+        # make sure we do UHF/UKS if we're not a singlet
+        if reference_S[a_z] != 0:
+            psi4.core.set_global_option("REFERENCE", "UHF")
+        else:
+            psi4.core.set_global_option("REFERENCE", "RHF")
```
loriab, line 97:
```suggestion
    optstash = optproc.OptionsState(["SCF", 'REFERENCE'])
    for a_sym, a_z, basis in unq_atoms:

        # make sure we do UHF/UKS if we're not a singlet
        if reference_S[a_z] != 0:
            psi4.core.set_option("SCF", "REFERENCE", "UHF")
        else:
            psi4.core.set_option("SCF", "REFERENCE", "RHF")
```
I think this is a little better if someone had set HF ref in local scope. But I can see arguments for your way, too, for catching local correlated.

loriab (approved):
All suggested changes optional, so see what you want. If any changes made, a rebase may be needed to get in some MDI corrections.

lgtm! thank you, and sorry for the delay.

jturney (approved):

psi4/driver/p4util/prop_util.py
```diff
@@ -0,0 +1,128 @@
...
+    psi4.core.print_out(f"  Running {len(unq_atoms)} free-atom UHF computations")
+
+    optstash = optproc.OptionsState(["SCF", 'REFERENCE'])
+    for a_sym, a_z, basis in unq_atoms:
+
+        # make sure we do UHF/UKS if we're not a singlet
+        if reference_S[a_z] != 0:
+            psi4.core.set_option("SCF", "REFERENCE", "UHF")
+        else:
+            psi4.core.set_option("SCF", "REFERENCE", "RHF")
```
loriab, line 97:
```suggestion
            psi4.core.set_local_option("SCF", "REFERENCE", "UHF")
        else:
            psi4.core.set_local_option("SCF", "REFERENCE", "RHF")
```

PR #2133, opened 2021-03-19
Add autoaux basis for dzvp

leeping:
## Description
This adds a new DF basis for dzvp generated using the autoaux program (courtesy Susi Lehtola).  This is a short-term solution for the current problem that there is no working DF basis for dzvp, as it is an all-electron basis for elements going beyond Kr.

## Todos
Still need to add codes in the right place to make this the default DF basis for dzvp.

## Status
- [x] Ready for review
- [ ] Ready for merge


leeping:
I need to add another basis file - closing for now, will reopen later.

Issue #2114, opened 2021-02-15
SCF Convergence and Integral Screening


zachglick:
The following SCF calculation fails to converge when run with psi4/master, but it completed okay with psi4/v1.3.2. This is due to the default integral screening threshold (`INTS_TOLERANCE`) changing from 0.0 to 1e-12 for the SCF module. The calculation converges fine when `INTS_TOLERANCE` is reset to 0.0:

```
memory 64 GB

molecule mol {
  0 1 
  O -1.43902644 2.20001916 0.52666667
  C -0.23262529 0.21202727 0.63286928
  N -1.24873102 -2.11271381 0.47828968
  C -3.7777515 -2.85877768 0.08730535
  C -4.46145441 -5.34849185 0.74152853
  C -2.56341349 -7.19739989 1.87800982
  C -6.94682221 -6.16901094 0.3643392
  C -8.75094374 -4.55140537 -0.66556154
  C -8.08273658 -2.09627319 -1.33509151
  C -5.61324248 -1.25402226 -0.96753978
  C 2.6282311 0.28062433 1.05654588
  C 3.98467651 2.83553405 0.63381414
  N 6.54374363 2.70155247 -0.40515728
  C 8.61979675 2.25141971 0.96961847
  O 8.56480572 1.89615119 3.26884825
  H -0.04232987 -3.54814977 0.82014114
  H -3.42815216 -8.99585225 2.39447197
  H -1.03934937 -7.57742382 0.54424112
  H -1.71851694 -6.40333698 3.5817869
  H -7.49559868 -8.07309898 0.86738429
  H -10.66693706 -5.20506164 -0.95166608
  H -9.47697652 -0.84262888 -2.14880758
  H -5.16216486 0.65044373 -1.5435283
  H 3.24390387 -0.55898099 2.83213254
  H 3.51375676 -1.09698602 -0.19445282
  H 3.04888413 3.57366108 -1.05446718
  H 3.38827894 4.41440023 1.82490852
  H 6.75331425 3.01713673 -2.28241121
  H 10.34700643 2.24329388 -0.20333453
  symmetry c1
  no_com
  no_reorient
  units bohr
}

# set ints_tolerance 0.0 # uncomment to fix

energy('pbe0/aug-cc-pv_tpd_z')
```
~~[screen_zero.out.txt](https://github.com/psi4/psi4/files/5984125/screen_zero.out.txt)~~
[...]

PR #2139, opened 2021-03-22
Disable libint2 Primitive Screening

zachglick:
## Description

Disables libint2's primitive screening via the `engine_precision` parameter. This integral screening provided a minor speedup, but in rare cases led to disastrously wrong SCF behavior as noted in #2114 and [here](https://github.com/edeprince3/hilbert/issues/8). This pull request closes those issues.

Note that this primitive screening is independent of the shell-pair level Schwarz and CSAM screening performed in Psi4. The previous workaround to this problem, which was to set the `ints_tolerance` keyword to 0.0, is no longer necessary.

## Todos

- [x] Change precision

## Checklist
- [x] all ctests are unaffected by this change

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/eri.cc
+++ b/psi4/src/psi4/libmints/eri.cc
@@ -178,8 +178,6 @@ Libint2ERI::Libint2ERI(const IntegralFactory *integral, double screening_thresho
     int max_nprim = std::max(std::max(basis1()->max_nprimitive(), basis2()->max_nprimitive()),
                              std::max(basis3()->max_nprimitive(), basis4()->max_nprimitive()));
     const auto max_precision = 0.;  // equivalent in accuracy and timings is `std::numeric_limits<double>::epsilon() * 1e-30;`
-    const auto basis_precision = std::numeric_limits<double>::epsilon() * std::pow((double)max_nprim, -4.0) * 1.e-2;
-    const auto engine_precision = screening_threshold == 0.0 ? max_precision : basis_precision;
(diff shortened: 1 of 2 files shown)


andysim (approved):
I completely agree with this change.  Thanks for figuring out all of the problems related to the primitive screening; I think a lot of the confusion comes from my initial implementation so sorry for the wasted time!  LGTM

andysim:
Could you quickly grep for any tests that specify the `ints_tolerance` keyword, and check to see whether they still need it, please?

loriab:
Thank you, Zach. I'm sorry you hit the results of my loose precision choice, but thanks for investigating. I'll run a conda-build to avoid surprises.

JonathonMisiewicz:
Will this close #2114 ?

zachglick:
No worries, this was easy compared to the actual libint2 interface that you all had to write and test. I've updated the post to clarify that this PR closes #2114.

susilehtola:
Looks good

JonathonMisiewicz:
Did you check if the `ints_tolerance  1e-14` in `ao-casscf-sp` is necessary?

loriab (approved):
four full tests are clean, lgtm!

zachglick:
I don't think `ints_tolerance` is set in the `ao-casscf-sp` test. grep might've picked up the reference output file, which apparently was generated using tighter settings than the actual test.

PR #2155, opened 2021-04-14
Density-based Integral Screening and Incremental Fock Build

andyj10224:
## Description
Implements Density matrix-based integral screening algorithms for Direct SCF, as well as Incremental Fock Build, the process of building a Fock matrix using the difference in the density matrix between SCF iterations.

Reference Paper:
https://onlinelibrary.wiley.com/doi/abs/10.1002/jcc.540100111

Updated Version of PR #2062

## Todos

- [x] Provide Massive Speedups to DirectJK Builds for HF and DFT
- [x] Accuracy of Calculations remain uncompromised

## Questions
- [x] Is my current method of handling the density matrix for UHF and ROHF ideal?

## Checklist
- [x] Passing cc-pVDZ and cc-pVTZ tests for HF and DFT (e_convergence 1.0e-6)
- [x] Passing aug-cc-pVDZ, aug-cc-pVTZ tests for HF and DFT

## Status
- [x] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
Marked as a 1.5 target.

Please fix any merge conflicts. I also recommend that "adding new database files" be moved to a separate PR.

andyj10224:
Some graphs from running B3LYP on the S22 dimer set (Not S22x7, as it would take too long)


DFT Eint Diff S22 IFB
DFT Eint Diff S22 DS

DFT Eabs Difference S22 IFB
DFT Eabs Difference S22 DS

Calculation Wall Times (running the whole database, 8 cores):

Reference: 7186.40 seconds
Density Screening: 4126.07 seconds
Increfock: 3635.78 seconds

andyj10224:
> Marked as a 1.5 target.
> 
> Please fix any merge conflicts. I also recommend that "adding new database files" be moved to a separate PR.

Will do, thank you!

hokru:
I am a bit confused what the graphs show, could you explain them a bit more?

JonathonMisiewicz:
Also, how have you verified correctness? While I'm not familiar with these methods, "can induce an energy difference of over 2 _hartrees_ even at conservative cutoffs" does not seem safe.

andyj10224:
> I am a bit confused what the graphs show, could you explain them a bit more?

Yes, those graphs represent the distribution of differences expected from using density screening (or density screening + incremental Fock build) compared to not using density screening at all, on the total B3LYP energy of dimers, as well as the interaction energies in the S22 set (not S22x7). Results suggest that there is no harm in using density screening.

hokru:
What is the number in the top left corner (1e4, 1e-7) and what is plotted on the x-axis?

andyj10224:
> Also, how have you verified correctness? While I'm not familiar with these methods, "can induce an energy difference of over 2 _hartrees_ even at conservative cutoffs" does not seem safe.

I have extensively verified the correctness of this approach, across many different molecules, basis sets, etc. I have never observed an error of "2 Hartrees". The worst I have ever observed is on the order of 1.0e-5 Hartrees, better than density fitting. Here is a link to my "benzene case study".

https://docs.google.com/spreadsheets/d/1HDMZ5PV6GhnK4i68Y1t86VhauTwqMP_X2kII21GCsiY/edit?usp=sharing

I empirically discovered that the best screening threshold given an energy convergence of 1.0e-n is 1.0e-(n+3) for non-augmented basis sets and 1.0e-(n+5) for augmented basis sets.

For example, for an e_convergence of 1.0e-8 in cc-pVTZ, the ideal threshold is 1.0e-11, while it is 1.0e-13 in aug-cc-pVTZ.

andyj10224:
> What is the number in the top left corner (1e4, 1e-7) and what is plotted on the x-axis?

The numbers on the top left corner reference the order of magnitude (e.g. 1e-7 hartrees, or 1.0e-4 kcal/mol).

The values on the x-axis are violin plot represent distribution frequency (It's a violin plot).

JonathonMisiewicz:
Including the _order of magnitude_ in a corner of the plot is highly non-standard and leads to the confusion I just went through. Choose units that are of the proper order of magnitude, and include any needed decimals on the axis directly. According to your axes, you have 2 hartree error, which is utterly unacceptable. 0.2 microhartree error is much more reasonable.

What you're describing confirms that the error is small. That is different from confirming correctness. Are you able to compare the numbers to some other implementation and show agreement? Is there some rare property that the exact scheme has, which you can numerically reproduce?

andyj10224:
> Including the _order of magnitude_ in a corner of the plot is highly non-standard and leads to the confusion I just went through. Choose units that are of the proper order of magnitude, and include any needed decimals on the axis directly. According to your axes, you have 2 hartree error, which is utterly unacceptable. 0.2 microhartree error is much more reasonable.
> 
> What you're describing confirms that the error is small. That is different from confirming correctness. Are you able to compare the numbers to some other implementation and show agreement? Is there some rare property that the exact scheme has, which you can numerically reproduce?

Ah, the energy convergence criteria itself is 1.0e-6, and the errors are well below that number, so I strongly believe that we are safe.

It may not be a good idea to compare to another implementation since every implementation has different tricks thrown in. Rather, if the energy difference is less than the e_convergence criteria, we are definitely safe. Though we could try to reproduce something like MBIS charges to check if we are safe.

hokru:
Now I also get it :)

Incremental Fock builds can show numerical error accumulation. Adding a reset frequency is sensible, e.g. after 20 incremental builds do a full Fock build from scratch. The actual number should be a (expert) user option.

The incremental Fock build can be compared against a standard build. They have to agree.
Comparing density screening to other programs can probably only be done for HF calculations and even then I would expect that one can only see 'similar errors' and no exact number agreement.

andyj10224:
> Now I also get it :)
> 
> Incremental Fock builds can show numerical error accumulation. Adding a reset frequency is sensible, e.g. after 20 incremental builds do a full Fock build from scratch. The actual number should be a (expert) user option.
> 
> The incremental Fock build can be compared against a standard build. They have to agree.
> Comparing density screening to other programs can probably only be done for HF calculations and even then I would expect that one can only see 'similar errors' and no exact number agreement.

This is already done, as I turn off incremental Fock build after the density drops below 1.0e-5, to reduce the differences, though I could make this a user option. :)

JonathonMisiewicz:
Wait, what? Then how is Incremental Fock build set to energy tolerance on the order of 1e-9 giving you errors on the order of 1e-7?

andyj10224:
> Wait, what? Then how is Incremental Fock build set to energy tolerance on the order of 1e-9 giving you errors on the order of 1e-7?

Oh no, I mean the SCREENING threshold is set to 1.0e-9, the energy tolerance is 1.0e-6.

JonathonMisiewicz:
Let me make sure I have this right: you're plotting energy errors on the order of 1e-7, but you only converged the energy to 1e-6? If so, then your plots are pure noise. If not, then what am I getting wrong?

andyj10224:
> Let me make sure I have this right: you're plotting energy errors on the order of 1e-7, but you only converged the energy to 1e-6? If so, then your plots are pure noise. If not, then what am I getting wrong?

Yes sir that is right :)

JonathonMisiewicz:
You should not have made those graphs. All you know for sure is that the error is less than 1 micro hartree. You cannot tell the difference between 1.0 e-7 and 1.5 e-7, but people read graphs as if you can tell the difference between your data points.

This is also why you can't just present graphs. You need to be absolutely sure that your labels are clear, and that somebody who isn't you will be able to figure out what the graph means. If you need to add a sentence to explain them, do so.

I request additional benchmarks so we can get more precise estimates on how much error these techniques introduce. Can you increase energy convergence to `1e-10`?

Because this PR isn't coming in until 1.5 anyways, I'm going to turn my attention to other things for a while.

andyj10224:
> You should not have made those graphs. All you know for sure is that the error is less than 1 micro hartree. You cannot tell the difference between 1.0 e-7 and 1.5 e-7, but people read graphs as if you can tell the difference between your data points.
> 
> This is also why you can't just present graphs. You need to be absolutely sure that your labels are clear, and that somebody who isn't you will be able to figure out what the graph means. If you need to add a sentence to explain them, do so.
> 
> I request additional benchmarks so we can get more precise estimates on how much error these techniques introduce. Can you increase energy convergence to `1e-10`?
> 
> Because this PR isn't coming in until 1.5 anyways, I'm going to turn my attention to other things for a while.

Ah, I am learning a lot. Good teachers like you will prepare me well for grad school :)

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -452,6 +452,10 @@ def scf_iterate(self, e_conv=None, d_conv=None):
             self.Da().print_out()
             self.Db().print_out()
 
+        # Set the current energy and density convergence values
+        core.set_variable("SCF DELTA E", Ediff)
```
JonathonMisiewicz, line 456: Why add these two variables? They didn't exist in the code before this PR, and they aren't used elsewhere in this PR.

andyj10224: Ah, the DELTA D is used to determine when we can turn incremental fock build off, when we get close to convergence. Since I added DELTA D, I thought that it would make sense to add DELTA E as well. Personally, I do not like having to go to python to get delta D, and if you have any workarounds for this, I would love to hear it. :)

psi4/src/psi4/libfock/jk.cc
```diff
@@ -540,6 +547,86 @@ void JK::AO2USO() {
     delete[] temp;
 }
 void JK::initialize() { preiterations(); }
+void JK::ifb_setup() {
+
+    if (iteration_ == 0 || D_ao_prev_.size() != D_ao_.size()) {
+
+        iteration_ = 0;
```
JonathonMisiewicz, line 554: I am _strongly_ opposed to the JK object needing to know the iteration number. I'll have a more concrete counter-recommendation when I can read the original paper.

JonathonMisiewicz (changes requested):
I feel like I need a copy of the original Ahlrichs paper before I can review this PR - I don't understand the "big picture."

While I'm waiting for the library to get that to me, here are some initial comments.

psi4/src/read_options.cc
```diff
@@ -1439,6 +1439,14 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
             to make the wavefunction stable. !expert -*/
         options.add_int("MAX_ATTEMPTS", 1);
+        /*- Do Perform Incremental Fock Build? -*/
+        options.add_bool("IFB", true);
+        /*- Do Perform Density Screening for Direct SCF? [Haser:1989:104] -*/
+        options.add_bool("SCF_DENSITY_SCREENING", true);
+        /*- Threshold for SCF Density Screening -*/
+        options.add_double("DENSITY_SCREENING_TOLERANCE", 1.0e-13);
+        /*- Density Threshold to stop Incremental Fock Build -*/
+        options.add_double("IFB_D_CONVERGENCE", 1.0e-5);
```
susilehtola, line 1449: This could just be `INCFOCK_CONVERGENCE`

susilehtola:
Incremental Fock builds have the tendency to have numerical creep. As small changes incrementally accumulate in the density matrix, the incrementally formed Fock matrix may no longer be correct; this is especially noticeable in large systems. Typically this is solved by doing full Fock builds every `N` steps. How is this handled in your implementation?

psi4/src/psi4/libmints/twobody.h
```diff
@@ -144,6 +145,12 @@ class PSI_API TwoBodyAOInt {
     /// Significant shell pairs, indexes by shell
     std::vector<std::vector<int>> function_to_function_;
     std::function<bool(int, int, int, int)> sieve_impl_;
+    /// Do Density Screening?
+    bool density_screening_;
+    /// Density Screening Tolerance
+    double density_screening_threshold_;
+    /// Max Density per Shell Pair
+    std::vector<std::vector<double>> max_dens_shell_pair_;
```
JonathonMisiewicz, line 153: Move this line of code to be near `function_pair_values_` and its brethren: they're all giving lookup information used by the various screening tests. Similar to those functions, this vector of vectors should be flatten into a simple vector.

JonathonMisiewicz (changes requested):
Now that I have a copy of the paper, I've been able to review the density screening part of the PR. That half of the PR doesn't follow the standards of similar code, which is the big problem here, and fixing it requires a restructure. I request that you refactor the density screening part of this before I look at the incremental Fock build part. Trying to do both at once is too likely to cause confusion.

As for your handling of UHF and ROHF, I don't have expertise there. If you haven't already, do a quick search for density screening for open-shell references, and check the papers that cite Ahlrichs for any titles that specifically mention open-shells or new developments in Direct SCF.

andyj10224:
> Now that I have a copy of the paper, I've been able to review the density screening part of the PR. That half of the PR doesn't follow the standards of similar code, which is the big problem here, and fixing it requires a restructure. I request that you refactor the density screening part of this before I look at the incremental Fock build part. Trying to do both at once is too likely to cause confusion.
> 
> As for your handling of UHF and ROHF, I don't have expertise there. If you haven't already, do a quick search for density screening for open-shell references, and check the papers that cite Ahlrichs for any titles that specifically mention open-shells or new developments in Direct SCF.

Hi Jonathon, I just implemented the refactoring. To request density screening, the user would just set `screening density` in the options

psi4/driver/procrouting/scf_proc/scf_iterator.py
loriab: Second removing `SCF DELTA E`. If we can't find another way to pass the rms density other than by global qcvar, I think it'd be better to `core.del_variable("SCF RMS D")` when it's no longer needed (after checking `has_variable`, I think).

tests/CMakeLists.txt
```diff
@@ -41,13 +41,13 @@ foreach(test_name adc1 adc2 casscf-fzc-sp casscf-semi casscf-sa-sp ao-casscf-sp
                   cc29 cc3 cc30 cc31 cc32 cc33 cc34 cc35 cc36 cc37 cc38 cc39
                   cc4 cc40 cc41 cc42 cc43 cc44 cc45 cc46 cc47 cc48 cc49 cc4a
                   cc50 cc51 cc52 cc53 cc54 cc55 cc5a cc6 cc7 cc8 cc8a cc8b cc8c
-                  cc9 cc9a cdomp2-1 cdomp2-2 cepa1
-                  cepa2 cepa3 cepa-module ci-multi cisd-h2o+-0 cisd-h2o+-1
+                  cc9 cc9a cdomp2-1 cdomp2-2 cepa1 cepa2 cepa3 cepa-module 
+		  ci-multi cisd-h2o+-0 cisd-h2o+-1
```
JonathonMisiewicz, line 45: Formatting.

psi4/src/read_options.cc
JonathonMisiewicz: Susi has a good point: "IFB" is too opaque. Please rename the option.

psi4/src/read_options.cc
```diff
@@ -1439,6 +1439,14 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
```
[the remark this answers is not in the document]
JonathonMisiewicz: Incremental Fock build test cases still needed.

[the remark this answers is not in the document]
JonathonMisiewicz: User-facing documentation still needed.

psi4/src/psi4/libmints/twobody.h
```diff
@@ -198,6 +204,8 @@ class PSI_API TwoBodyAOInt {
     /*
      * Sieve information
      */
+    /// Update the Max Density Per Shell Pair given an updated Density Matrix (Haser 1989)
```
JonathonMisiewicz, line 207:
```suggestion
    /// Update max_dens_shell_pair_ given an updated density matrix (Haser 1989)
```

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,61 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
 
 TwoBodyAOInt::~TwoBodyAOInt() {}
 
+// Haser 1989, Equation 7 
+void TwoBodyAOInt::update_density(const std::vector<SharedMatrix>& D) {
+    
+    timer_on("Update Density");
+#pragma omp parallel for
+    for (int M = 0; M < nshell_; M++) {
```
JonathonMisiewicz, line 124: Why are these loops independent? This code counts (M, N) and (N, M). I believe it would be more efficient to only count one.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,61 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
...
+bool TwoBodyAOInt::shell_significant_density(int M, int N, int R, int S) {
+
+    // THR in Equation 9
+    double density_threshold = screening_threshold_squared_;
+
+    // Equation 13
+    double Q_MN_sq = shell_pair_values_[N * nshell_ + M];
+    double Q_RS_sq = shell_pair_values_[S * nshell_ + R];
+
+    std::initializer_list<double> block_densities {max_dens_shell_pair_[M * nshell_ + N], max_dens_shell_pair_[R * nshell_ + S],
```
JonathonMisiewicz, line 160: This isn't what I meant when I asked for an initializer list. What I recommend was like [example](https://en.cppreference.com/w/cpp/algorithm/max) `std::max({1, 10, 50, 200})` is the example I had in mind. They don't need to use an initializer list class.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,61 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
...
+        0.25 * max_dens_shell_pair_[N * nshell_ + R], 0.25 * max_dens_shell_pair_[N * nshell_ + S]};
+
+    // Equation 6
+    double max_density = std::max(block_densities);
+    // Squared to account for the fact that Q_MN is given as its square
+    max_density = max_density * max_density;
+
+    // Equations 6, 9, and 14
+    if (Q_MN_sq * Q_RS_sq * max_density > density_threshold) return true;
+    return false;
```
JonathonMisiewicz, line 171:
```suggestion
    return Q_MN_sq * Q_RS_sq * max_density > density_threshold;
```

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -112,6 +116,61 @@ TwoBodyAOInt::TwoBodyAOInt(const TwoBodyAOInt &rhs) : TwoBodyAOInt(rhs.integral_
...
+    double Q_RS_sq = shell_pair_values_[S * nshell_ + R];
+
+    std::initializer_list<double> block_densities {max_dens_shell_pair_[M * nshell_ + N], max_dens_shell_pair_[R * nshell_ + S], 
+        0.25 * max_dens_shell_pair_[M * nshell_ + R], 0.25 * max_dens_shell_pair_[M * nshell_ + S],
+        0.25 * max_dens_shell_pair_[N * nshell_ + R], 0.25 * max_dens_shell_pair_[N * nshell_ + S]};
+
+    // Equation 6
+    double max_density = std::max(block_densities);
+    // Squared to account for the fact that Q_MN is given as its square
+    max_density = max_density * max_density;
```
JonathonMisiewicz, line 167: Reading the paper, this is a major problem in your implementation. Haser's equation is that:
ε > Q_MN * Q_RS * ΔD.

You do not have his equation 13 correct. What you call Q is what Haser calls Q squared. To "correct" for this, you square ΔD, but then you would need to square ε as well. So when the user inputs a screening threshold, they are effectively getting the square root of that threshold. That's bad.

The simpler thing to do is just define Q as Haser does.

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -76,10 +76,14 @@ TwoBodyAOInt::TwoBodyAOInt(const IntegralFactory *intsfactory, int deriv)
...
     else if (screentype == "CSAM")
         screening_type_ = ScreeningType::CSAM;
+    else if (screentype == "DENSITY")
+        screening_type_ = ScreeningType::Density;
     else
         throw PSIEXCEPTION("Unknown screening type " + screentype + " in TwoBodyAOInt()");
     
     if (screening_threshold_ == 0.0) screening_type_ = ScreeningType::None;
+
+    max_dens_shell_pair_.resize(bs1_->nshell() * bs1_->nshell(), 1.0);
```
JonathonMisiewicz, line 86: Why is this necessary?

JonathonMisiewicz (changes requested):
I suspect the density screening algorithm is implemented incorrectly. I don't want to take focus away from this point, so I'm going to refrain from delving into the IFB just yet.

psi4/src/psi4/libmints/twobody.cc
andyj10224: Ah, I set the density threshold to the square of the screening threshold, so my code was still valid, let me change it to make it more clear :) Schwarz Screening also does this, to avoid taking expensive square roots.

JonathonMisiewicz: I'm going to request a second opinion from somebody more knowledgeable about screening procedures (@andysim?). What bothers me is that `screening_threshold_` is determined by `INTS_TOLERANCE`, and the documentation on `INTS_TOLERANCE` mentions nothing about this square/square-root behavior. At the very least, this is undocumented behavior.

psi4/src/psi4/libmints/twobody.cc
andyj10224: Moved this to the `create_sieve_pair_info()` function, where its other brethren are set up too.

andyj10224:
> I suspect the density screening algorithm is implemented incorrectly. I don't want to take focus away from this point, so I'm going to refrain from delving into the IFB just yet.

Hi Jonathan, I have added test cases for density screening and incfock in the pytests/test_erisieve.py file. I have yet to user-facing documentation, but feel free to review the code I have currently.

andyj10224:
I just updated the docs too. One thing I am confused about, why is the `screening` keyword only in the sapt module?

hokru:
> One thing I am confused about, why is the `screening` keyword only in the sapt module? I think we should refactor it as a global module.

That's odd! 
Since we have `INTS_TOLERANCE` for every module maybe we should do the same with `screening`?

JonathonMisiewicz:
> I just updated the docs too. One thing I am confused about, why is the `screening` keyword only in the sapt module?

Very good question. @zachglick [added the option](https://github.com/psi4/psi4/commit/e1977af9ae69b77ca518c3169efb23018b51ffb0), so perhaps he can shed some light on this one.

loriab:
`SCREENING` should probably be a top-o-the-read_options-file keyword so that it distributes identically to the modules. keywords read by libraries are iffy to place since prepare_options_for_module(mod) but mod != lib. But I may be missing a trick as to why it was placed in sapt.

zachglick:
I can't remember or come up with a good reason for placing `SCREENING` in sapt. I agree that moving the option to the top of read_options makes a lot of sense.

psi4/src/psi4/libfock/jk.h
```diff
@@ -303,6 +314,21 @@ class PSI_API JK {
     /// wK matrices: wK_mn = (ml|w|ns) C_li^left C_si^right
     std::vector<SharedMatrix> wK_ao_;
 
+    /// D, J, K, wK Matrices from previous iteration, used in Incremental Fock Builds
+    std::vector<SharedMatrix> D_ao_prev_;
```
JonathonMisiewicz, line 318: Can you elaborate on why this needs to be a vector? For alpha and beta?

JonathonMisiewicz (changes requested):
Additional "miscellanea" items:

* I don't like the fact that a user who inputs X as their integral tolerance is actually getting SQRT(X). I want a second opinion from somebody more knowledgeable about the screening code. If nothing else, this needs to be stated very clearly in the documentation for the integral tolerance option.
* The "screening" keyword should be moved to top-level and have absolutely nothing to do with sapt. Check anywhere this keyword is used in the docs, and make sure the sapt prefix is fone.
* We need user-facing documentation on what an incremental Fock build is.
* I don't like that we have to add so many variables that are just for IFB, but I don't see a way around it. If other core developers have a recommendation here?
* I still want to remove knowledge of `iteration_` from JK. I need an answer to the question I asked about the `*_ao_prev_` variables before I can recommend how to do that.

psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -272,51 +272,62 @@ void DirectJK::compute_JK() {
...
         }
     }
 
     if (do_J_ || do_K_) {
         std::vector<std::shared_ptr<TwoBodyAOInt>> ints;
         ints.push_back(std::shared_ptr<TwoBodyAOInt>(factory->eri()));
+        if (density_screening_) ints[0]->update_density(D_ref);
         for (int thread = 1; thread < df_ints_num_threads_; thread++) {
             ints.push_back(std::shared_ptr<TwoBodyAOInt>(ints[0]->clone()));
+            if (density_screening_) ints[thread]->update_density(D_ref);
```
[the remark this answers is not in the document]
andyj10224: You are right, I added the density copying to the copy constructor for a TwoBodyAOInt

CMakeLists.txt
```diff
@@ -305,7 +305,7 @@ ExternalProject_Add(psi4-core
    BUILD_ALWAYS 1)
 
 add_subdirectory(external/downstream)
-add_subdirectory(doc)
+# add_subdirectory(doc)
```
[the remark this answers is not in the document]
andyj10224: Fixed it, some of the Sherrill machines didn't compile with this line, so I removed it temporarily

psi4/src/export_mints.cc
```diff
@@ -1231,18 +1232,23 @@ void export_mints(py::module& m) {
 
     typedef bool (TwoBodyAOInt::*compute_shell_significant)(int, int, int, int);
     typedef size_t (TwoBodyAOInt::*compute_shell_ints)(int, int, int, int);
+    typedef void (TwoBodyAOInt::*update_densities)(const std::vector<SharedMatrix>&);
```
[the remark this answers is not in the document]
andyj10224: It is not, it was there because I was trying to debug. Fixed.

psi4/src/psi4/libfock/jk.h
```diff
@@ -254,6 +254,17 @@ class PSI_API JK {
     /// Do wK matrices? Defaults to false
     bool do_wK_;
 
+    /// Perform Density Screening for ERIs?
+    bool density_screening_;
+    /// Perform Incremental Fock Build for J and K Matrices?
+    bool ifb_;
+    /// Density Matrix convergence criteria
+    double d_conv_;
```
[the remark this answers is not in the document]
andyj10224: Not used, removed.

psi4/src/psi4/libfock/jk.h
```diff
@@ -254,6 +254,17 @@ class PSI_API JK {
...
+    /// The density matrix convergence value at which to stop performing IFB
+    double ifb_d_conv_;
```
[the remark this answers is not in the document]
andyj10224: Let's wait until we get other opinions on INCFOCK before we make any changes :)

psi4/src/psi4/libfock/jk.h
andyj10224: Ah, yes, D_ao_ is also a vector of matrix pointers, so D_ao_prev_ would just be an analog to it (indeed for alpha and beta in UHF and ROHF).

andyj10224:
I made the requested changes, I am wondering which .rst file I need to put the incfock documentation in. Waiting for expert opinions on incfock :)

loriab:
> I am wondering which .rst file I need to put the incfock documentation in.

I think https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/scf.rst would be good.

JonathonMisiewicz:
> I made the requested changes, I am wondering which .rst file I need to put the incfock documentation in. Waiting for expert opinions on incfock :)

Waiting for more feedback sounds good to me. Libfock is one of the pieces of the code I don't know as well.

andyj10224:
Hi all, 

I just updated the density screening code with the UHF/ROHF cases. I also added a test in the test_erisieve.py pytest file that ensures the same number of shell quartets are being screened out with RHF and UHF.

JonathonMisiewicz:
@andyj10224 @andysim What's the status of this PR? The last I heard was that on the 24th, Andy S. said, "i’m about to have a call with Andy so i’ll ask him some of the questions i have regarding density screening because i think it should be possible to do CSAM and density screening at the same time"

Are we still looking for another reviewer before re-considering my outstanding comments? Did something come of that conversation that is causing more development of this PR?

andyj10224:
> @andyj10224 @andysim What's the status of this PR? The last I heard was that on the 24th, Andy S. said, "i’m about to have a call with Andy so i’ll ask him some of the questions i have regarding density screening because i think it should be possible to do CSAM and density screening at the same time"
> 
> Are we still looking for another reviewer before re-considering my outstanding comments? Did something come of that conversation that is causing more development of this PR?

Ah, @andysim was interested in doing density screening in addition to CSAM, not as its own separate screening. Other than that I think we are good to go on the PR. Do you have any comments for incremental Fock build? @andysim should I go ahead and make the change to do density screening in addition to things like CSAM?

JonathonMisiewicz:
Unless adding it will force you to rewrite a large amount of code, or not adding it will make your results incorrect, I'm going to request you leave that to a separate PR. It's difficult enough finding people to review this, and adding more code will only make it worse.

psi4/src/psi4/libfock/jk.cc
JonathonMisiewicz: I'm still opposed to having an `iteration_` member variable, but as I look over the code, I don't understand it. Why are you treating D, J, K, and wK on the same footing in the `if` case, but treating D alone in the `else` case?

JonathonMisiewicz (changes requested):
The PR is coming along, but I suspect that I have been very confused about what this PR is for. If these changes were meant for the DirectJK algorithm only, yes, the title says that, but your description didn't, and the files you chose to edit led me to believe otherwise.

andyj10224:
@JonathonMisiewicz To reference the treating the D separately, we are calculating the change in the density matrix, or delta D if we are not in the first iteration, the delta J, K, and wKs are added to the J and K in the post-iteration step. That way, incremental Fock build is available for all JK types. Basically knowledge of the iteration is needed so we know when to set up the delta D matrices. (It is also needed for my future CFMM and linK PRs, where we do NOT want to run linK or CFMM on the SAD guess)

andyj10224:
One question, to all core developers, IMHO, we should change the default value for `df_scf_guess` from true to false, since optimized direct algorithms would be faster than density fitting for large systems. What do y'all think?

susilehtola:
> One question, to all core developers, IMHO, we should change the default value for `df_scf_guess` from true to false, since optimized direct algorithms would be faster than density fitting for large systems. What do y'all think?

But that's only true for small basis sets.

If you run small to moderate sized molecules with triple-zeta basis sets, DF is very fast.

JonathonMisiewicz:
> I made all of the requested changes, barring the `iteration_` variable, which I personally believe is necessary. If you have a workaround, please let me know :).This PR should be _really_ close to being finalized.

At risk of being grumpy, please don't say you made all the requested changes unless you actually have. There were many of my comments that still haven't been addressed.

susilehtola:
I would also consider turning off the incremental build once a preset level of `D_CONVERENCE` is achieved. The point about full rebuilds every `N` iterations was just that if, say, 40 SCF iterations are taken until the incremental SCF converges, it may have accumulated so much numerical creep that the actual orbital gradient is still significant.

andyj10224:
> Thanks. The question of "when do we _use_ an incremental build?" needs to be settled (which it isn't yet - see Susi's comment) before I can approve this.

Hi Jonathon, I just implemented Susi's suggestion about the `d_convergence`. I also updated the necessary tests (`density-screen-1` and `density-screen-2`).

andyj10224:
Besides the `Inc. Fock Build` printing issue, this PR is good to go, as I have just compiled and cleaned up the sphinxman docs.

Density Screening Screenshot

jturney (approved):
Super nice! Thanks for the feature!

doc/sphinxman/source/bibliography.rst
```diff
@@ -558,3 +558,8 @@ Bibliography
    E. Caldeweyher,  S. Ehlert,  A. Hansen, H. Neugebauer, S. Spicher, C. Bannwarth, and S. Grimme,
    *J. Chem. Phys.* **150**, 154122 (2019).
    https://doi.org/10.1063/1.5090222
+
+.. [Haser:1989:104]
+   M. Haser and R. Ahlrichs,
+   *J. Comp. Chem.* **10(1)**, 104 (1989).
+   doi: 10.1002/jcc.540100111
```
andysim, line 565:
```suggestion
   doi: 10.1002/jcc.540100111
   
.. [Thompson:2017:144101]
   T. H. Thompson and C. Ochsenfeld
   *J. Chem. Phys.* **147**, 144101 (2017).
   doi: 10.1063/1.4994190
```

doc/sphinxman/source/scf.rst
```diff
@@ -705,7 +705,29 @@ post SCF algorithms require a specific implementation.
...
+the |globals__screening| keyword to your desired option. For Schwarz screening, set it
+to ``SCHWARZ``, for CSAM, ``CSAM``, and for density matrix-based screening, ``DENSITY``.
+
+SCHWARZ
+    Uses the Cauchy-Schwarz inequality to calculate an upper bounded value of a shell quartet,
+
+.. math:: (PQ|RS) <= \sqrt{(PQ|PQ)(RS|RS)}
+
+CSAM
+    An extension of the Schwarz estimate that also screens over the long range 1/r operator.
```
andysim, line 718:
```suggestion
    An extension of the Schwarz estimate that also screens over the long range 1/r operator, described in [Thompson:2017:144101]_.
```

psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -271,52 +363,83 @@ void DirectJK::compute_JK() {
...
+    std::vector<SharedMatrix>& J_ref = (do_incfock_iter_ ? delta_J_ao_ : J_ao_);
+    std::vector<SharedMatrix>& K_ref = (do_incfock_iter_ ? delta_K_ao_ : K_ao_);
+    std::vector<SharedMatrix>& wK_ref = (do_incfock_iter_ ? delta_wK_ao_ : wK_ao_);
 
     if (do_wK_) {
         std::vector<std::shared_ptr<TwoBodyAOInt>> ints;
         for (int thread = 0; thread < df_ints_num_threads_; thread++) {
             ints.push_back(std::shared_ptr<TwoBodyAOInt>(factory->erf_eri(omega_)));
+            if (density_screening_) ints[thread]->update_density(D_ref);
         }
```
andysim, line 391: If this shows up in the timings for parallel runs, we can change it to building a single integral object, then cloning it for all threads with an ID greater than 0.  It's correct as-is, so I don't suggest we change it unless we see L2 object construction times becoming significant.

andysim (approved):
This is really outstanding work - kudos for getting it all figured out, and implemented in some of the most difficult parts of the code in terms of control flow.  My concern before was that the list of max densities is an (N^2) quantity and could get large for larger systems, while storing it only for those shell pairs that survive the Schwarz/CSAM screening would not.  This is not really a valid concern because we have overlap matrices and other O(N^2) quantities in core anyway, so the implementation is great as-is.  I added a suggestion or two to add the  CSAM reference because I can never remember which paper it comes from.  LGTM!

JonathonMisiewicz:
Per Lori request, we have a hold on merging this in until David Poole can look this over.

psi4/src/psi4/libfock/DirectJK.cc
andyj10224: iirc, update density is a REALLY cheap step.

andyj10224: `Update Density                      :      6.883u      0.050s      0.234w      9 calls`

Cheapness confirmed, this is on an 18 core machine.

psi4/src/psi4/libfock/jk.h
```diff
@@ -706,6 +706,34 @@ class PSI_API DirectJK : public JK {
...
 
+    /// Options object
+    Options& options_;
+
+    // Perform Density matrix-based integral screening?
+    bool density_screening_;
+
+    // => Incremental Fock build variables <= //
+    /// Perform Incremental Fock Build for J and K Matrices?
+    bool incfock_;
```
davpoolechem, line 717: Would it be better to change the name of this variable to "do_incfock_"? It seems that the do_ at the beginning of the variable name is something of a convention for variables like this (e.g., do_csam_, do_J_). Also, it would probably be wise to put in the comment what this variable defaults to.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -425,6 +428,9 @@ def scf_iterate(self, e_conv=None, d_conv=None):
                 if self.frac_performed_:
                     status.append("FRAC")
 
+                if incfock_performed:
```
davpoolechem, line 431: I notice that this if statement uses an incfock_performed variable determined via parameters in jk, while the if statements directly above this one use variables contained within the HF class (MOM_performed_ and frac_performed_). For consistency's sake, it might be better to move incfock_performed_ to the HF class and keep this if statement as "if self.infock_performed_:"

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -436,6 +442,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
         core.timer_off("HF: Form D")
 
         self.set_variable("SCF ITERATION ENERGY", SCFE)
+        core.set_variable("SCF D NORM", Dnorm)
```
davpoolechem, line 445: It's interesting to me that the "SCF D NORM" variable is part of core rather than self. Is there any reason for this, and would it be better for SCF D NORM to be part of self here instead of core?

psi4/src/read_options.cc
```diff
@@ -1439,6 +1439,11 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- When using |scf__stability_analysis| ``FOLLOW``, maximum number of orbital optimization attempts
             to make the wavefunction stable. !expert -*/
         options.add_int("MAX_ATTEMPTS", 1);
+        /*- Do Perform Incremental Fock Build? -*/
+        options.add_bool("INCFOCK", false);
+        /*- Frequency with which to compute the full Fock matrix if using |scf__incfock| . 
+        N means rebuild every N SCF iterations to avoid accumulating error from the incremental procedure. -*/
+        options.add_int("INCFOCK_FULL_FOCK_EVERY", 5);
```
davpoolechem, line 1446: Just as a heads up, I was able to run the IncFock code with values of INCFOCK_FULL_FOCK_EVERY below 1. Running negative values for this option seems to perform incfock for all iterations, while setting this option to 0 leads to a floating-point exception on my system. It may be wise to add some sort of option checking to make sure that the input value for this variable makes sense.

davpoolechem (commented):
I just finished looking at the code. I left a couple of comments to be reviewed before I approve these changes.

psi4/driver/procrouting/scf_proc/scf_iterator.py
loriab: Good catch -- it alarmed me, too, at one point. iirc, the answer was that directJK that reads this doesn't have access to the wfn object, so globals is the easiest way to communicate.

psi4/driver/procrouting/scf_proc/scf_iterator.py
JonathonMisiewicz: Absolutely not.

1) Class variables are a necessary evil at best. (Good luck trying to figure out where the variable was last modified.) Local variables are preferred.
2) `incfock_performed_` isn't something that the HF object should know about. It doesn't mean anything unless you're using a direct algorithm.

davpoolechem: All right, that makes sense. I had mainly suggested it for consistency's sake; but if it is better this way, then there is no real need to change it.

psi4/driver/procrouting/scf_proc/scf_iterator.py
davpoolechem: Ahhhhhhhh, I see. That would make sense!

psi4/src/psi4/libfock/jk.h
andyj10224: Good idea, but at this point, changing the associated variable would be too much of a "shotgun surgery"

psi4/src/read_options.cc
andyj10224: Just fixed

andyj10224:
I addressed all of David's comments. Ready to merge.

JonathonMisiewicz:
I'd rather have @davpoolechem officially sign off on this first. It'll be an hour before the test suite check finishes, anyway.

davpoolechem (approved):
Andy and I talked a bit about the comments that I made, and he has sufficiently addressed all of them. I give my seal of approval to this code!

PR #2158, opened 2021-04-19
only run GHA at psi4 org, not forks

loriab:
## Description
Fix problem @JonathonMisiewicz hit where the GHA docs runs on master branch of forks. Someday there ought to be a more concise way to filter https://github.com/actions/runner/issues/662.

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz (approved):

hokru (approved):

andysim (approved):
It's strange that these must be specified for every single block.  Oh well, LGTM!

PR #2159, opened 2021-04-19
Restores debug symbols in Debug/RelWithDebInfo

jturney:
## Description
Debug symbols where being mistakenly striped from the core python module.

## Todos

- [x] Debug symbols are back
- [x] Ability to break on line numbers is back.

## Checklist
- [x] Works for @JonathonMisiewicz and myself.  Doesn't affect any test cases.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/CMakeLists.txt
+++ b/psi4/CMakeLists.txt
@@ -13,6 +13,11 @@ if(${Fortran_ENABLED})
     enable_language(Fortran)
 endif()
 
+if (NOT MSVC)
+  set(CMAKE_C_FLAGS_DEBUG "-g -O0")
+  set(CMAKE_CXX_FLAGS_DEBUG "-g -O0")
+endif()
+
 list(APPEND CMAKE_MODULE_PATH ${PSI4_ROOT}/cmake)
 
 include(psi4OptionsTools)
--- a/psi4/src/CMakeLists.txt
+++ b/psi4/src/CMakeLists.txt
@@ -20,7 +20,10 @@ target_link_libraries(core
     pybind11::windows_extras
 )
 pybind11_extension(core)
-pybind11_strip(core)
+if(NOT MSVC AND NOT ${CMAKE_BUILD_TYPE} MATCHES Debug|RelWithDebInfo)
+    # Strip unnecessary sections of the binary on Linux/macOS
+    pybind11_strip(core)
+endif()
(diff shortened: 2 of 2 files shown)


JonathonMisiewicz (approved):
This approach is [good enough for Pybind](https://github.com/pybind/pybind11/blob/114be7f4ade0ad798cd4c7f5d65ebe4ba8bd892d/tools/pybind11NewTools.cmake#L249-L252).

andysim:
This is great!  Could we also get `-O0` into the flags for Debug builds?  I think it adds `-g` currently, but allows the compiler to use its default optimization (usually `-O2`), which often trips up debuggers.  We can also just defer that for a different time.

JonathonMisiewicz:
> I think it adds `-g` currently, but allows the compiler to use its default optimization (usually `-O2`), which often trips up debuggers.

I can confirm this. I've been passing the `-O0` flag manually when trying to figure this bug out.

loriab (approved):

jturney:
Let me check real quick on the `-O0` flag.

loriab:
#2158 failed all azure - Win with a mdi failure at test time and Lin with Py detection failures at Psi configure time. So far this one is ok. It's beyond the configure, but this one is using cmake 3.19. 2158 was using cmake 3.20 which has changes to Py detection. I'm keeping an eye on both PRs, but it's hardly caused by the PR changes themselves.

jturney:
Ok, I added `-O0` to the default flags for `CMAKE_C*_FLAGS_DEBUG` when not using MSVC.

Should be good to go.

psi4/CMakeLists.txt
```diff
@@ -13,6 +13,11 @@ if(${Fortran_ENABLED})
     enable_language(Fortran)
 endif()
 
+if (NOT MSVC)
+  set(CMAKE_C_FLAGS_DEBUG "-g -O0")
+  set(CMAKE_CXX_FLAGS_DEBUG "-g -O0")
```
loriab, line 18: is this debug only or all runs?

andysim (approved):
Awesome - thanks for adding the debug flag.  LGTM!

loriab:
Taylor is going to do a MDI PR to fix the Azure.

PR #2152, opened 2021-04-12
Misc. DCT Cleanup, `libmints` Function Variants

JonathonMisiewicz:
## Description
This is the first of what will likely be quite a few cleanup PRs in `dct`, in preparation for some features I'd like for 1.5.

This PR creates variants of existing constructors in `libmints`. We're no longer forced to use `SharedMatrix` in `set_block`, and we can now pass a single `Slice` object to `get_block` or `set_block` to use it for both rows and columns.

This PR can be moved to 1.5 if needed. I can also opt for fewer but longer PRs if needed.

## Checklist
- [x] `dct` test suite passes

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2165, opened 2021-04-23
Bugfix and Test Shortening

JonathonMisiewicz:
## Description
Lori brought to my attention that my previous PR accidentally introduced a bug causing tests to fail. (It was an obvious indexing bug.) That is now fixed.

In the interests of making test-running less onerous, I've also changed the test cases so that I can run all `dct` tests faster. We're checking things less tightly and using smaller basis sets.

## Todos
- [x] Fix bug introduced in #2152 
- [x] Faster `dct` tests

## Checklist
- [x] `dct` tests pass, and in under two minutes

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):
thanks for the quick fix!

andysim (approved):
Yay, faster tests!  LGTM

jturney (approved):

PR #2166, opened 2021-04-24
reorganize driver imports and API

loriab:
## Description
This produces a docs build with no warnings or errors, either overt (`make html -W`), external links (`make linkcheck`), or internal links (`make html -n`). requires a qcel docs update. hopefully this will provide a clean docs base moving forward.

This PR is the controversial part of the docs technical refresh. I think there was some technical debt left over from the 2016 inversion in the module structure of the driver. What I think was happening was that the `psi4/driver/driver.py` was the nexus, and any wrappers or utilites that it happened to import became the API. The folder psi4/driver didn't really function as a module, and any `psi4.driver` was referring to driver.py, not psi4/driver/. Then for psithon, a few extras like frac got added in in the import preamble. Only problem with the structure (besides ppl who know python finding it irregular) is that it made some functions in the driver impossible to access (i.e., run `help()` on) and link to in the docs.

https://towardsdatascience.com/whats-init-for-me-d70a312da583 outlines a few strategies for structuring python packages. There's the mostly-toplevel strategy of numpy (https://github.com/numpy/numpy/blob/main/numpy/__init__.py#L150-L156 and https://github.com/numpy/numpy/blob/main/numpy/core/__init__.py#L72-L90) and the subpackage strategy of scipy (https://github.com/scipy/scipy/blob/master/scipy/__init__.py and https://github.com/scipy/scipy/blob/master/scipy/optimize/__init__.py#L400-L423). I don't have a clear vision on how to do the driver right for psi4, but i've followed the numpy approach of both building up the top-level deliberately with `__all__` and building out the namespaced structure (`from . import x`), and it's doing what the docs linking needs, and I think it's more rational and provides much better coverage for the API docs. Also allows the only difference btwn psithon and psiapi to be whether `core.*` are at top-level.

The only tests that have needed change are "cookbook" types that were outlining driver code in input files. P4n changes are fewer than I feared, https://github.com/psi4/psi4numpy/pull/112. The driver imports are a non-trivial change and could use discussion and maybe relegation to 1.5. But they, or something like them, are necessary for a clean docs build w/o hacks.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
Per discussion on conference call, moved to 1.5. Trying to get this in will delay 1.4, and it's not the wisest to do an API reorganization after we already have a release candidate out.

loriab:
Between #2597 and #3033 , much of this PR has been incorporated. What hasn't is fixing the mushiness of `psi4.py_func` vs. `psi4.driver.py_func`, but that can be a fresh start if it causes trouble again.

PR #2182, opened 2021-05-14
Remove relax_tau

JonathonMisiewicz:
## Description
Another tiny one, in case anybody objects to my lack of deprecation.

The old keyword `relax_tau` in `dct` is eliminated. This was marked expert, was rarely a good idea to use, and wasn't consistently applied throughout the code. This keyword was implemented as a way to explore alternate convergence strategies that would have been done in P4N, nowadays.  Because this is fairly obscure code to boot, I'm eliminating it now with no formal deprecation strategy.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):

hokru (approved):

andysim (approved):
Yeah, this one's a pretty arcane feature that doesn't need a deprecation in my option.  LGTM!

Issue #2197, opened 2021-06-23
X2C analytic gradients


susilehtola:
Psi4 1.3.2 happily runs analytic derivatives when X2C is enabled, even though as confirmed by @fevangelista on Slack the gradient is not correct since it is missing the terms arising from X2C.

The code should throw an error for analytic derivatives when X2C is employed.

JonathonMisiewicz:
Closed by #2200, but the moment somebody adds a `compute_gradient` call that actually works, this will re-open.

PR #2200, opened 2021-06-24
misc, incl. x2c grad warning

loriab:
## Description
A few items I ran across. addresses #2197 

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/driver.py
+++ b/psi4/driver/driver.py
@@ -95,6 +95,10 @@ def _find_derivative_type(ptype, method_name, user_dertype):
         core.print_out('\nPCM analytic gradients are not implemented yet, re-routing to finite differences.\n')
         dertype = 0
 
+    if core.get_global_option("RELATIVISTIC") in ["X2C", "DKH"]:
+        core.print_out("\nRelativistic analytic gradients are not implemented yet, re-routing to finite differences.\n")
+        dertype = 0
+
     # Summary validation
     if (dertype == 2) and (method_name in procedures['hessian']):
         pass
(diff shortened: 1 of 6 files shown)


JonathonMisiewicz:
What happens if a user runs a computation, toggles whether X2C is on, and then calls the wavefunction's `compute_gradient` method?

loriab:
> What happens if a user runs a computation, toggles whether X2C is on, and then calls the wavefunction's compute_gradient method?

Apparently even the non-rel returns a `psi4.core.RHF`, and since that hasn't had it's `compute_gradient` implemented, the parent `Wavefunction`'s `Analytic gradients are not available for this wavefunction.` is thrown.

Not what I was expecting (or probably what someone using a C++ API was expecting), but it does shut down a way to get the wrong answer by evading the Py APi.

JonathonMisiewicz:
Because I'm actively trying to break this: `x2c` can be applied for correlated methods as well. It's great that we've caught it in `run_scf_gradient`, but what stops the x2c gradients for other electronic structure methods?

loriab:
`compute_gradient` after a `energy("ccsd")` and `gradient("scf")` also throw the wfn.h error, so no worries there.

py-side `gradient("ccsd")` is what this PR misses catching, as I see it.

loriab:
How about diverting to findif, like PCM? This'll need changing for DDD, but at least we'll notice.

(Note that PR not in self-consistent state at present. test needs replacing.)

JonathonMisiewicz:
Great idea! I still have some concerns about potential ways to abuse `compute_gradient`, but my test cases won't even run for reasons far beyond the scope of this PR. Issue list is about to grow, but I'll be happy with the PR once the tests are updated.

Thanks for this, Lori.

loriab:
This is RTG, imo.

JonathonMisiewicz (approved):

jturney (approved):

PeterKraus (approved):

PR #2210, opened 2021-06-26
[Small, Important Change] Changed phi_ao code to support puream basis sets

andyj10224:
## Description
Currently, the phi_ao function returns the value of the basis functions at a point. However, this ONLY works for cartesian and not spherical basis sets. I updated the code to make this change. This is important to the development of seminumerical methods, like chain-of-spheres exchange.

## Todos

- [x] Code is tested in my developmental chain-of-spheres exchange branch
- [x] Made an explicit test testing the validity of my bug fix

## Questions
- [x] Will this impact existing functions that use the phi_ao call?

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1171,7 +1171,13 @@ void export_mints(py::module& m) {
         .def("max_function_per_shell", &BasisSet::max_function_per_shell,
              "The max number of basis functions in a shell")
         .def("max_nprimitive", &BasisSet::max_nprimitive, "The max number of primitives in a shell")
-        .def_static("construct_from_pydict", &construct_basisset_from_pydict, "docstring");
+        .def_static("construct_from_pydict", &construct_basisset_from_pydict, "docstring")
+        .def("compute_phi", [](BasisSet& basis, double x, double y, double z) {
+            auto phi_ao = new std::vector<double>(basis.nbf());
+            auto capsule = py::capsule(phi_ao, [](void *phi_ao) { delete reinterpret_cast<std::vector<double>*>(phi_ao); });
+            basis.compute_phi(phi_ao->data(), x, y, z);
+            return py::array(phi_ao->size(), phi_ao->data(), capsule);
+        }, "Calculate the value of all basis functions at a given point x, y, and z");
 
     py::class_<OneBodyAOInt, std::shared_ptr<OneBodyAOInt>> pyOneBodyAOInt(
         m, "OneBodyAOInt", "Basis class for all one-electron integrals");
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -1240,7 +1240,12 @@ void BasisSet::move_atom(int atom, const Vector3 &trans) {
     xyz_[offset + 2] += trans[2];
 }
 
-void BasisSet::compute_phi(double *phi_ao, double x, double y, double z) {
+void BasisSet::compute_phi(double* phi_ao, double x, double y, double z) {
+    if (puream_) compute_phi_pure(phi_ao, x, y, z);
+    else compute_phi_cart(phi_ao, x, y, z);
+}
+
+void BasisSet::compute_phi_cart(double *phi_ao, double x, double y, double z) {
     zero_arr(phi_ao, nao());
 
     int ao = 0;
@@ -1268,4 +1273,46 @@ void BasisSet::compute_phi(double *phi_ao, double x, double y, double z) {
 
         ao += INT_NCART(am);
     }  // nshell
+}
+
+void BasisSet::compute_phi_pure(double *phi_ao, double x, double y, double z) {
+    zero_arr(phi_ao, nbf());
+
+    int ao = 0;
+    for (int ns = 0; ns < nshell(); ns++) {
+        const GaussianShell &shell = shells_[ns];
+        int am = shell.am();
+        int nprim = shell.nprimitive();
+        const double *a = shell.exps();
+        const double *c = shell.coefs();
+
+        auto s_transform = SphericalTransform(am);
+
+        const double *xyz = shell.center();
+        double dx = x - xyz[0];
+        double dy = y - xyz[1];
+        double dz = z - xyz[2];
+        double rr = dx * dx + dy * dy + dz * dz;
+
+        double cexpr = 0.0;
+        for (int np = 0; np < nprim; np++) cexpr += c[np] * exp(-a[np] * rr);
+
+        std::vector<double> cart_buffer(INT_NCART(am), 0.0);
+
+        for (int l = 0; l < INT_NCART(am); l++) {
+            Vector3 &components = exp_ao[am][l];
+            cart_buffer[l] += pow(dx, (double)components[0]) * pow(dy, (double)components[1]) *
+                              pow(dz, (double)components[2]) * cexpr;
+        }
+
+        for (int i = 0; i < s_transform.n(); i++) {
+            int lcart = s_transform.cartindex(i);
+            int lpure = s_transform.pureindex(i);
+            double coef = s_transform.coef(i);
+
+            phi_ao[ao + lpure] += coef * cart_buffer[lcart];
+        }
+
+        ao += INT_NPURE(am);
+    } // nshell
 }
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -1261,13 +1261,13 @@ void BasisSet::compute_phi(double* phi_ao, double x, double y, double z) {
         for (int np = 0; np < nprim; np++) cexpr += c[np] * exp(-a[np] * rr);
 
         if (puream_) {
-            auto s_transform = SphericalTransform(am);
+            const auto s_transform = SphericalTransform(am);
             std::vector<double> cart_buffer(INT_NCART(am), 0.0);
 
             for (int l = 0; l < INT_NCART(am); l++) {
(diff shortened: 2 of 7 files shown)


JonathonMisiewicz:
Thanks for the diagnosis and the fix.

Why not combine these two functions into a single function with an `if` clause to determine whether the `cart_buffer` transform is necessary?

I'm also going to request you include a test case in _this_ PR.

JonathonMisiewicz:
Following up on an issue from the conference call:
The fixed `phi_ao` functions are used to form the core Hamiltonian whenever an embedding potential is read from a file. While I haven't run anything, I would expect the bug fixed by this PR to cause correctness errors in SCF energies. Definitely agree with getting this fix in for 1.4.

andyj10224:
> Following up on an issue from the conference call:
> The fixed `phi_ao` functions are used to form the core Hamiltonian whenever an embedding potential is read from a file. While I haven't run anything, I would expect the bug fixed by this PR to cause correctness errors in SCF energies. Definitely agree with getting this fix in for 1.4.

Hi Jonathon,

As requested, I cleaned up the code by removing the helper functions, as well as added a new test case (C test phi-ao).
I think having phi_ao as a python function will be useful to Psi developers and users, as it would be immensely helpful for the development of grid-based approaches.

Thank you,
Andy

tests/phi-ao/input.dat
```diff
@@ -0,0 +1,64 @@
+#! Test of the new phi_ao code (for puream and non-puream basis functions)
```
JonathonMisiewicz, line 1:
```suggestion
#! Test computing values of basis functions (puream and non-puream) at points
```

psi4/src/export_mints.cc
```diff
@@ -1171,7 +1171,13 @@ void export_mints(py::module& m) {
         .def("max_function_per_shell", &BasisSet::max_function_per_shell,
              "The max number of basis functions in a shell")
         .def("max_nprimitive", &BasisSet::max_nprimitive, "The max number of primitives in a shell")
-        .def_static("construct_from_pydict", &construct_basisset_from_pydict, "docstring");
+        .def_static("construct_from_pydict", &construct_basisset_from_pydict, "docstring")
+        .def("compute_phi", [](BasisSet& basis, double x, double y, double z) {
+            auto phi_ao = new std::vector<double>(basis.nbf());
+            auto capsule = py::capsule(phi_ao, [](void *phi_ao) { delete reinterpret_cast<std::vector<double>*>(phi_ao); });
+            basis.compute_phi(phi_ao->data(), x, y, z);
+            return py::array(phi_ao->size(), phi_ao->data(), capsule);
```
JonathonMisiewicz, line 1179: Could you give me a reference for what you're doing here? I can't find one in the Pybind docs and I've never come across code like this before me. I'm guessing this is about [ensuring the data is deleted when appropriate](https://stackoverflow.com/questions/44659924/returning-numpy-arrays-via-pybind11/44682603#44682603)? If so, this is probably going to be very useful.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -1260,12 +1260,32 @@ void BasisSet::compute_phi(double *phi_ao, double x, double y, double z) {
         double cexpr = 0;
         for (int np = 0; np < nprim; np++) cexpr += c[np] * exp(-a[np] * rr);
 
-        for (int l = 0; l < INT_NCART(am); l++) {
-            Vector3 &components = exp_ao[am][l];
-            phi_ao[ao + l] += pow(dx, (double)components[0]) * pow(dy, (double)components[1]) *
-                              pow(dz, (double)components[2]) * cexpr;
+        if (puream_) {
+            auto s_transform = SphericalTransform(am);
```
JonathonMisiewicz, line 1264: Does this code compile if you add `const`? I don't believe we've changing this object.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -1260,12 +1260,32 @@ void BasisSet::compute_phi(double *phi_ao, double x, double y, double z) {
...
+            std::vector<double> cart_buffer(INT_NCART(am), 0.0);
+
+            for (int l = 0; l < INT_NCART(am); l++) {
+                Vector3 &components = exp_ao[am][l];
```
JonathonMisiewicz, line 1268: `const auto`

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -1260,12 +1260,32 @@ void BasisSet::compute_phi(double *phi_ao, double x, double y, double z) {
...
+                cart_buffer[l] += pow(dx, (double)components[0]) * pow(dy, (double)components[1]) *
```
JonathonMisiewicz, line 1269: I suggest explicit `static_cast<double>`.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -1260,12 +1260,32 @@ void BasisSet::compute_phi(double *phi_ao, double x, double y, double z) {
...
+                int lcart = s_transform.cartindex(ind);
+                int lpure = s_transform.pureindex(ind);
+                double coef = s_transform.coef(ind);
+
+                phi_ao[ao + lpure] += coef * cart_buffer[lcart];
+            }
+
+        } else {
+            for (int l = 0; l < INT_NCART(am); l++) {
+                Vector3 &components = exp_ao[am][l];
```
JonathonMisiewicz, line 1283: Same as previous.

tests/phi-ao/input.dat
```diff
@@ -0,0 +1,64 @@
...
+        y = np.asarray(block.y())
+        z = np.asarray(block.z())
+
+        for point in range(npoints):
+            phi_ao_points[running_points + point] = basis.compute_phi(x[point], y[point], z[point])
+            weights[running_points + point] = w[point]
+
+        running_points += npoints
+
+    # Creative manipulations for speedy calculations
```
JonathonMisiewicz, line 54:
```suggestion
```

tests/phi-ao/input.dat
```diff
@@ -0,0 +1,64 @@
...
+    basis = psi4.core.BasisSet.build(mol, "ORBITAL", psi4.core.get_global_option("BASIS"), puream=puream)
+    grid = psi4.core.DFTGrid.build(mol, basis)
+
+    # Run an SCF calculation to get Electron Density
+    e, wfn = energy('scf', return_wfn=True)
+
+    # Density Matrix
+    D = np.asarray(wfn.Da())
+    
+    # Grid information
```
JonathonMisiewicz, line 31: Move grid setup outside of the for loop. No need to do this twice.

tests/phi-ao/input.dat
```diff
@@ -0,0 +1,64 @@
...
+  0 1
+  O
+  H 1 1.0
+  H 1 1.0 2 104.5
+  symmetry c1
+  no_reorient
+  no_com
+""")
+
+TEST_NELECTRON = {}
```
JonathonMisiewicz, line 16: We don't need this. Just get the value in each for loop and compare it. It doesn't need to persist after that.

JonathonMisiewicz (commented):
I think the risk of `compute_phi` changing is minimal, so I'm not opposed to exposing it. "Useful for grid-based approaches" is too vague for me to be _for_ it, but I don't need to be.

andyj10224:
> I think the risk of `compute_phi` changing is minimal, so I'm not opposed to exposing it. "Useful for grid-based approaches" is too vague for me to be _for_ it, but I don't need to be.

An example of "grid-based approaches" would include a psi4numpy implementation of seminumerical methods like chain-of-spheres exchange.

psi4/src/export_mints.cc
andyj10224: Yes, I got this code from this source

https://github.com/pybind/pybind11/issues/1042#issuecomment-325941022

andyj10224:
> I think the risk of `compute_phi` changing is minimal, so I'm not opposed to exposing it. "Useful for grid-based approaches" is too vague for me to be _for_ it, but I don't need to be.

Hi Jonathon, I just made all of your suggested changes. Everything compiled and ran properly :)

JonathonMisiewicz (approved):

andysim (approved):
Great test case!  It's a good example of how to use these kinds of objects. LGTM

loriab (commented):
looks good, thanks for the nice fix and test! a couple suggestions to consider

tests/phi-ao/input.dat
```diff
@@ -0,0 +1,64 @@
...
+
+    phi_prime = np.einsum('uv,pu->pv', D, phi_ao_points)
+    TEST_NELECTRON = 2.0 * np.einsum('p,pv,pv->', weights, phi_prime, phi_ao_points)
+
+    if puream:
+        testname = 'SPHERICAL BASIS PHI_AO TEST'
+    else:
+        testname = 'CARTESIAN BASIS PHI_AO TEST'
+    
+    compare_values(TEST_NELECTRON, REF_NELECTRON, 4, testname)
```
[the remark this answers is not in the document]
andyj10224: Solved

psi4/src/export_mints.cc
```diff
@@ -1171,7 +1171,13 @@ void export_mints(py::module& m) {
...
+        }, "Calculate the value of all basis functions at a given point x, y, and z");
```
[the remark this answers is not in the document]
andyj10224: Discussed this with Lori in Slack, but adding the extra arguments led to a compile error.

loriab (approved):

PR #2221, opened 2021-06-28
Revert behaviour of bsse_type='cp'

### LINKED ISSUE #1691: bsse_type='cp' does unnecessary monomer-basis calculations
Psi4 1.4a2.dev66,      Git: Rev {master} 6e67529

see topic. Only recently noticed this. It used to do only dimer-basis calculations.

```
molecule ne2 {
    0 1
    Ne 0 0 0
    --
    0 1
    Ne 0 0 2
}
set scf_type pk
energy('scf/3-21G', bsse_type="cp")
```

output:
```   ===> N-Body Interaction Abacus <===
        BSSE Treatment:                     cp
        Number of 1-body computations:     4
        Number of 2-body computations:     1
```

PeterKraus:
## Description
This PR reverts the behaviour of standard `bsse_type='cp'` calculations to calculate all fragments (monomers) in the full (dimer) basis only. The PR also fixes printing interaction energies disguised as total energies in those tables.

When a calculation is requested with `bsse_type='cp', return_total_data=True`, an `bsse_type=['cp', 'nocp']` is automatically performed and the total energies are calculated & returned. With `bsse_type=['cp','nocp']` (and `return_total_data` unspecified or `False`), the total energies are calculated, but the interaction energies are returned.

The PR fixes #1691 .

## Checklist
- [x] Tests updated
- [x] `ctest -L nbody` passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -192,7 +192,7 @@ def nbody_gufunc(func: Union[str, Callable], method_string: str, **kwargs):
         ``'on'`` for gradients and Hessians. Note that the calculation of total
         counterpoise corrected energies implies the calculation of the energies of
         monomers in the monomer basis, hence specifying ``return_total_data = True``
-        may carry out more computations than ``return_Total_data = False``.
+        may carry out more computations than ``return_total_data = False``.
 
     :type levels: dict
     :param levels: ``{1: 'ccsd(t)', 2: 'mp2', 'supersystem': 'scf'}`` || ``{1: 2, 2: 'ccsd(t)', 3: 'mp2'}`` || etc
@@ -776,7 +776,7 @@ def assemble_nbody_components(metadata, component_results):
     results = {}
     results['nbody'] = nbody_dict
     for b in ['cp', 'nocp', 'vmfc']:
-        if b in metadata['bsse_type_list'] and monomer_energies != 0.0:
+        if monomer_energies != 0.0:
             results['%s_energy_body_dict' % b] = eval('%s_energy_body_dict' % b)
             results['%s_energy_body_dict' % b] = {str(i) + b: j for i, j in results['%s_energy_body_dict' % b].items()}
         else:
--- a/psi4/driver/driver_nbody.py
+++ b/psi4/driver/driver_nbody.py
@@ -112,6 +112,10 @@ def _print_nbody_energy(energy_body_dict, header, embedding=False):
     core.print_out("""\n   ==> N-Body: %s energies <==\n\n""" % header)
     core.print_out("""   n-Body     Total Energy [Eh]       I.E. [kcal/mol]      Delta [kcal/mol]\n""")
     previous_e = energy_body_dict[1]
+    if previous_e == 0.0:
+        tot_e = False
+    else:
+        tot_e = True
     nbody_range = list(energy_body_dict)
     nbody_range.sort()
     for n in nbody_range:
@@ -119,8 +123,12 @@ def _print_nbody_energy(energy_body_dict, header, embedding=False):
         delta_e_kcal = delta_e * constants.hartree2kcalmol
         int_e_kcal = (
             energy_body_dict[n] - energy_body_dict[1]) * constants.hartree2kcalmol if not embedding else np.nan
-        core.print_out("""     %4s  %20.12f  %20.12f  %20.12f\n""" % (n, energy_body_dict[n], int_e_kcal,
-                                                                      delta_e_kcal))
+        if tot_e:
+            core.print_out("""     %4s  %20.12f  %20.12f  %20.12f\n""" % (n, energy_body_dict[n], int_e_kcal,
+                                                                       delta_e_kcal))
+        else:
+            core.print_out("""     %4s  %20s  %20.12f  %20.12f\n""" % (n, " ", int_e_kcal,
+                                                                       delta_e_kcal))
         previous_e = energy_body_dict[n]
     core.print_out("\n")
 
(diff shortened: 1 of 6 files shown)


JonathonMisiewicz:
At this point, I'd favor some documentation about what the various BSSE types do. I don't think this is anywhere in the docs.

PeterKraus:
> At this point, I'd favor some documentation about what the various BSSE types do. I don't think this is anywhere in the docs.

I have added CP and NoCP definition into the docs, and expanded the examples, I'm not sure what VMFC is.

loriab:
For any newcomers, discussion is located at the issue, not here.

psi4/driver/driver_nbody.py
```diff
@@ -759,8 +776,11 @@ def assemble_nbody_components(metadata, component_results):
     results = {}
     results['nbody'] = nbody_dict
     for b in ['cp', 'nocp', 'vmfc']:
-        results['%s_energy_body_dict' % b] = eval('%s_energy_body_dict' % b)
-        results['%s_energy_body_dict' % b] = {str(i) + b: j for i, j in results['%s_energy_body_dict' % b].items()}
+        if b in metadata['bsse_type_list'] and monomer_energies != 0.0:
```
loriab, line 779: any chance this is missing the opportunity to calculate a bsse type that has data available but wasn't explicitly requested by user?

tests/nbody-cp-gradient/input.dat
```diff
@@ -166,7 +166,7 @@ for result in [  #TEST
 
 # compute nbody again with return_total_data=False
 # * note CURRENT ENERGY moved from first to second testing block
-g, wfn = gradient('SCF/STO-3G', molecule=water_trimer, bsse_type='cp', max_nbody=2,
+g, wfn = gradient('SCF/STO-3G', molecule=water_trimer, bsse_type=['cp','nocp'], max_nbody=2,
```
loriab, line 169: is the change in the test still needed now that `return_total_data` defaults to True for grad?

loriab (commented):
Thanks for the fix and the great table https://github.com/psi4/psi4/issues/1691#issuecomment-870398772 . lgtm after a couple questions

tests/nbody-cp-gradient/input.dat
PeterKraus: Yes, because we're checking with `return_total_data=False` stated explicitly.

psi4/driver/driver_nbody.py
PeterKraus: Okay, this one is a little tricky and might require a lot more work. Ideally, the driver would be clever enough to derive all BSSE treatments possible from the set of calculated energies, without filling blanks with zeroes. Probably too much work right now before DDD (or 1.4 release).

It should only print tables of whatever is in `'bsse_type_list'`, i.e. `return_total_data=True, bsse_type='cp'` would not print no-CP results, but would calculate them and return them on `wfn`, whereas `bsse_type=['cp','nocp']` printing and calculating both (with the order of the array changing what gets returned).

Before this change, the `2CP` field would return interaction energy for `energy(..., bsse_type='cp')` but a total energy for `energy(..., bsse_type=['cp', 'nocp'])`, which is what I want to avoid.

JonathonMisiewicz:
Sorry to keep making you write, but is the "big picture" view of nbody documented _anywhere_? I can't follow what `energy_body_dict` means, or why the value of one of its entries is coupled to whether the total energy is available.

I fear this is a case where there is a _lot_ of "developer was only coding to be comprehensible to themselves" code debt.

psi4/driver/driver_nbody.py
loriab: Ah, got it. I think the compute-all-available would be nice down the road (and I wouldn't mind if it printed all-available, too), but no immediate need as long as it isn't a feature loss. Yes, the `2CP` you describe sounds important to avoid.

loriab:
> Sorry to keep making you write, but is the "big picture" view of nbody documented anywhere? I can't follow what energy_body_dict means, or why the value of one of its entries is coupled to whether the total energy is available.

> I fear this is a case where there is a lot of "developer was only coding to be comprehensible to themselves" code debt.

I confess to being lost in nbody, too. And some of the psivars set aren't as descriptive as I'd like (e.g., `2CP`. However, any changes to nbody make DDD especially painful to rebase. And I think a good time to rationalize and document the pieces are when we take the nbody communication dictionary DDD uses and formulate/propose a proper schema. I guess I'm ok leaving it inscrutable (and without binding contracts :-) until then.

JonathonMisiewicz:
I can agree to that, but reviewing code I don't understand doesn't sit well with me, so I'll bow out of being one of the reviewers unless there is a desperate need.

Good luck with DDD!

PeterKraus:
To be fair, I don't understand big parts of nbody either - the whole factorial thing is a bit strange. Which is why if we'd like return an object containing all possible nbody x bsse_type results from the data calculated by the combination of bsse_type and return_total_data, it'd take me a lot of work to get there - and is probably not worth it before DDD is in.

The problem with the link between `energy_body_dict` and the returned data is that it's copied to the results directly, and if the calculation cannot compute total energies, it only returns an intermediate. For most cases that's sort of OK (since the intermediate is 0), but in the case of counterpoise without calculating monomers in monomer basis, the intermediate is the interaction energy:

```
2CP = Etot,cp = E(AB)_AB - E(A)_AB - E(B)_AB + E(A)_A + E(B)_B
2CP = Etot,cp = Eint,cp                      + monomer_energies
```

loriab (approved):
Thanks for the fix and for the what-happens table at the issue!

jturney (approved):

andysim (approved):
LGTM!

Issue #2012, opened 2020-10-02
frozen core problem for system with Na and I


hokru:
The following input, given to me via email, gives a 'bad' dMP2 correction for SAPT:
```
      delta HF,r (2)              -18.53850185 [mEh]     -11.63308554 [kcal/mol]     -48.67282991 [kJ/mol]
      delta MP2,r (2)             141.03839795 [mEh]      88.50293088 [kcal/mol]     370.29626281 [kJ/mol]
```

```
molecule {
0 1
 F                  0.00000000    0.00000000    2.96216000
 I                  0.00000000    0.00000000    0.96704400
--
0 1
 C                  0.00000000    0.00000000   -2.52999200
 N                  0.00000000    0.00000000   -1.35165800
 Na                 0.00000000    0.00000000   -4.84283500
units angstrom
no_reorient
symmetry c1
}
memory 20 GiB
set {
    basis def2-TZVP
    freeze_core true
}
energy('sapt2+dMP2')
energy('mp2',bsse_type=['nocp'])
```

The issue is reproducible for MP2/CCSD post-SCF methods. It is not directly `DF` nor `ghost` atom related. (besides a possibly small error due to https://github.com/psi4/psi4/issues/1968). And not a module specific issue.

Using different ECPs (LANL2DZ) or an all electron basis (x2c-SVPall, no rel. effects.) with global `FC=on` seems to give 'better' results in the sense that MP2 interaction energy is negative.

One, and perhaps only, issue is `frozen_core`.
Turning `FC = off` only for the supermolecular MP2 part gives OK results. For the NaCN fragment we are freezing 10 electrons, while e.g. ORCA is freezing only 3. 
At least for this case, freezing the Ne-shell by default for Na is not a good choice.

Tagging @PeterKraus.

`freeze_core -2` makes the SAPT part crash because sth goes wrong for the `IF` fragment (fragment A):
```
    FOCC A     = 18446744073709551611
    FOCC B     =         1
```

CDSherrill:
I think freezing 1s, 2s, and 2p for Na is what we decided on for the
default behavior, but I agree that it is quite iffy and one should really
be freezing only 1s.  Didn't we have a "FREEZE_CORE MIN" option a while
back that understood these kinds of rules?  That makes more sense to me
than this new (to me) "FREEZE_CORE [0,-1,-2, ...]" etc.... I don't see how
that's supposed to be a useful set of options, because the behavior needs
to be different for Na than it is for the other elements.  (The manual
[...]

PR #2222, opened 2021-06-28
Previous shell core freezing and ECPs

PeterKraus:
This PR fixes a part of #2012, where requesting to freeze n-th fixed previous shell (using `set freeze_core -n` syntax) would happily carry on, even if an ECP present on a certain atom would substitute more electrons than in the n-th shell. This means the example in #2012 using `freeze_core -2` now throws an exception.

The issue of "smart freezing" of electrons of alkali metals and a further revamp of the interface (per-atom or per-element spec.) is for a different PR.

## Questions
- [x] Do I need to add a test?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/basisset.cc
+++ b/psi4/src/psi4/libmints/basisset.cc
@@ -288,6 +288,7 @@ int BasisSet::n_frozen_core(const std::string &depth, SharedMolecule mol) {
                 int current_shell = atom_to_period(Z + ECP);
                 int delta = period_to_full_shell(std::max(current_shell - req_shell, 0));
                 // If this center has an ECP, some electrons are already frozen
+                if (delta < ECP) throw PSIEXCEPTION("ECP on atom freezes more electrons than requested by choosing a previous shell.");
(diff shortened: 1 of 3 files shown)


JonathonMisiewicz:
I'd appreciate a test for this one. [Here](https://stackoverflow.com/questions/23337471/how-to-properly-assert-that-an-exception-gets-raised-in-pytest) is a StackOverflow example of testing that an exception gets raised.

PeterKraus:
> I'd appreciate a test for this one. [Here](https://stackoverflow.com/questions/23337471/how-to-properly-assert-that-an-exception-gets-raised-in-pytest) is a StackOverflow example of testing that an exception gets raised.

Done. Although I've remained with ctest as opposed to pytest. I am not sure if there's a way to catch PSIEXCEPTION exceptions directly, or if RuntimeError is the closest we can get from within python.

loriab (approved):

JonathonMisiewicz (approved):

jturney (approved):

PR #2225, opened 2021-06-30
Implement level shifts

### LINKED ISSUE #2216: Implement level shifts
A [long-requested feature](http://forum.psicode.org/t/scf-level-shift/938): implement level shifts in Psi4.

This is an even simpler to implement than [GDM](#2214). What is needed is to edit the `form_C()` functions in RHF, UHF, and ROHF by replacing e.g.
```
    diagonalize_F(F_, C_, epsilon_);
```
with the diagonalization of an effective Fock matrix `Feff = F + shift S*Cvirt*Cvirt.T*S` where `shift` is the level shift (should be `0.5` by default), when the DIIS error is larger than the set threshold.

This would also make the SCF solver smarter, see #888.

susilehtola:
## Description

This PR implements level shifts.

Closes #2216.

## Todos

- [x] Level shifts in restricted methods
- [x] Level shifts in unrestricted methods
- [x] Level shifts in constrained unrestricted methods
- [x] Level shifts in restricted open shell methods

## Questions
- [ ] Should the level shift be turned on by default? Since Psi4 does not have EDIIS/ADIIS, it will often fail to reach the convergence radius of DIIS. Level shifts may slow down convergence somewhat, but if they're turned off at large residual gradient (e.g. 1e-2) then the slowdown should be negligible in most cases.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
Thanks for this!

Bullet points before an in-depth review:
* Why have separate `form_C` and `form_C_level_shift`? I could imagine a single `form_C` function that takes a level shift `int` that defaults to 0.
* This needs user docs on when to use level shifts, how to select a good level shift, and whether this will change the converged energy/orbitals/density-matrix/Fock-matrix
* To get the build working, please commit the `CMakeLists.txt` file for `scf-level-shift-rks`

susilehtola:
>     * Why have separate `form_C` and `form_C_level_shift`? I could imagine a single `form_C` function that takes a level shift `int` that defaults to 0.

I didn't want to change the signature of `form_C`. However, having one function is, of course, the cleanest solution. Since the level shifting is a tiny bit more work, I use plain diagonalization if the shift vanishes.

>     * This needs user docs on when to use level shifts, how to select a good level shift, and whether this will change the converged energy/orbitals/density-matrix/Fock-matrix

Added

>     * To get the build working, please commit the `CMakeLists.txt` file for `scf-level-shift-rks`

Done

psi4/src/psi4/libscf_solver/rohf.cc
```diff
@@ -426,8 +425,22 @@ void ROHF::form_F() {
     }
 }
 
-void ROHF::form_C() {
-    soFeff_->diagonalize(Ct_, epsilon_a_);
+void ROHF::form_C(double shift) {
```
JonathonMisiewicz, line 428: Orbital energies in ROHF aren't even well-defined, so it's not obvious to me how level-shifted ROHF would work. I think the canonical reference on this is [here](https://www.tandfonline.com/doi/abs/10.1080/00268977400102171), but Emory doesn't seem to have online access to Mol. Phys., so I can't access it right now.

JonathonMisiewicz (changes requested):
The reason for the build fail seems to be an issue with Pybind not knowing about the `form_C` default value.

psi4/src/psi4/libscf_solver/cuhf.cc
```diff
@@ -308,9 +308,26 @@ void CUHF::form_F() {
...
+        auto Cvir = Ca_subset("SO", "VIR");
+        auto SCvir = std::make_shared<Matrix>(nirrep_, S_->rowspi(), Cvir->colspi());
+        SCvir->gemm(false, false, 1.0, S_, Cvir, 0.0);
+        shifted_F->gemm(false, true, shift, SCvir, SCvir, 0.0);
+        shifted_F->add(Fa_);
+        diagonalize_F(shifted_F, Ca_, epsilon_a_);
+
+        Cvir = Cb_subset("SO", "VIR");
+        SCvir = std::make_shared<Matrix>(nirrep_, S_->rowspi(), Cvir->colspi());
+        shifted_F->gemm(false, true, shift, SCvir, SCvir, 0.0);
```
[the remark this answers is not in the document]
susilehtola: not sure what you mean

psi4/src/psi4/libscf_solver/uhf.cc
```diff
@@ -270,9 +268,27 @@ void UHF::form_F() {
...
+        auto Cvir = Ca_subset("SO", "VIR");
+        auto SCvir = std::make_shared<Matrix>(nirrep_, S_->rowspi(), Cvir->colspi());
+        SCvir->gemm(false, false, 1.0, S_, Cvir, 0.0);
+        shifted_F->gemm(false, true, shift, SCvir, SCvir, 0.0);
+        shifted_F->add(Fa_);
+        diagonalize_F(shifted_F, Ca_, epsilon_a_);
+
+        Cvir = Cb_subset("SO", "VIR");
+        SCvir = std::make_shared<Matrix>(nirrep_, S_->rowspi(), Cvir->colspi());
+        shifted_F->gemm(false, true, shift, SCvir, SCvir, 0.0);
```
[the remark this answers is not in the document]
susilehtola: ?

psi4/src/psi4/libscf_solver/cuhf.cc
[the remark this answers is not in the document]
JonathonMisiewicz: `SCvir` is going to be a matrix of all 0's, no? You need to repeat previous GEMM command (now using the beta virtual orbitals) for it to be non-zero. Same below.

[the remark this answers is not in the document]
susilehtola: Ah, you're 100% correct: copy-paste error.

psi4/src/psi4/libscf_solver/uhf.cc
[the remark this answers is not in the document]
susilehtola: Fixed

psi4/src/psi4/libscf_solver/rohf.cc
JonathonMisiewicz: Now that I've been able to get that paper: the ROHF level shifting approach they recommend (eq. 15) is to apply level shifter `alpha` to the open orbitals and level-shifter `alpha+beta` to the virtual orbitals. They don't do an exhaustive study of optimal level shifts, but they use `beta`=`alpha`.

psi4/src/psi4/libscf_solver/rhf.cc
```diff
@@ -286,8 +285,19 @@ void RHF::form_F() {
     }
 }
 
-void RHF::form_C() {
-    diagonalize_F(Fa_, Ca_, epsilon_a_);
+void RHF::form_C(double shift) {
```
JonathonMisiewicz, line 288: One detail worth double-checking: does Psi define the RHF Fock matrix as `F_a + F_b` or `(F_a + F_b) / 2`? If the former, we should double the shift factor as well.

susilehtola: The former choice would give the wrong orbital energies, so I don't think that's an issue.

psi4/src/psi4/libscf_solver/rohf.cc
susilehtola: But what are these orbitals in the code itself?

JonathonMisiewicz: I don't understand the question. Are you asking how to obtain the orbital spaces? The [`form_F` function documentation](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libscf_solver/rohf.cc#L361) establishes how the ROHF Fock matrix is arranged, and that `doccpi` and `soccpi` work as you would naively expect.

If that doesn't answer the question, please rephrase.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -372,7 +372,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
                 # need to ensure orthogonal orbitals and set epsilon
                 status.append(base_name + "conv")
                 core.timer_on("HF: Form C")
-                self.form_C()
+                self.form_C(0.0)
```
jturney, line 375: I would prefer the binding to be updated to know about the default parameter. When I see this function call I have no idea what 0.0 is supposed to mean.

[Pybind11 Default Arguments](https://pybind11.readthedocs.io/en/stable/basics.html#default-argumentsl)

psi4/src/psi4/libscf_solver/rohf.cc
susilehtola: Oh, okay. I've changed the code accordingly. I use the same level shift (`shift`) for virtual orbitals as in RHF, and thereby shift the open-shell orbitals by `shift/2`. However, I'm not sure it is working properly; e.g. the example calculation of Guest and Saunders
```
molecule {
units bohr
0 2
B
O 1 2.277
}

set basis pcseg-1
set reference rohf
#set reference uhf
set diis false
set level_shift 0.2
set level_shift_convergence 1e-3
energy('scf')
```

does not converge.

psi4/driver/procrouting/scf_proc/scf_iterator.py
susilehtola: Thanks for the tip!

psi4/src/psi4/libscf_solver/rohf.cc
susilehtola: So what I don't understand is the difference between `Ca_` and `Ct_`.

susilehtola: OK, solved; ROHF works.

psi4/src/psi4/libscf_solver/cuhf.cc
```diff
@@ -308,9 +308,26 @@ void CUHF::form_F() {
...
 
-void CUHF::form_C() {
-    diagonalize_F(Fa_, Ca_, epsilon_a_);
-    diagonalize_F(Fb_, Cb_, epsilon_b_);
+void CUHF::form_C(double shift) {
+    if (shift == 0.0) {
+        diagonalize_F(Fa_, Ca_, epsilon_a_);
+        diagonalize_F(Fb_, Cb_, epsilon_b_);
+    } else {
+        auto shifted_F = SharedMatrix(factory_->create_matrix("F"));
```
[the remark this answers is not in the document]
susilehtola: That really makes no difference

JonathonMisiewicz (approved):
LGTM, and thank you very much for coding this up!

jturney (approved):

psi4/src/read_options.cc
```diff
@@ -1400,6 +1400,10 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_int("DIIS_MAX_VECS", 10);
         /*- Do use DIIS extrapolation to accelerate convergence? -*/
         options.add_bool("DIIS", true);
+        /*- Do use a level shift? -*/
+        options.add_double("LEVEL_SHIFT", 0.0);
+        /*- DIIS error at which to stop applying the level shift -*/
+        options.add_double("LEVEL_SHIFT_CONVERGENCE", 1e-2);
```
loriab, line 1406: I think call it `LEVEL_SHIFT_CUTOFF` for consistency. https://psicode.org/psi4manual/master/manage_addon.html?highlight=threshold#how-to-name-keywords-in-psi4-src-read-options-cc

susilehtola: Renamed.

susilehtola: Tests pass again

loriab (approved):

Issue #1973, opened 2020-07-28
Normalization issue in Molden writer when using Cartesian basis functions


tovrstra:
Example: [molden_cartesian.zip](https://github.com/psi4/psi4/files/4988043/molden_cartesian.zip)

The ZIP file contains a trivial HF/6-31G* calculation on water with PSI4 and Gaussian. To allow for a one-on-one comparison, symmetry, reorientation etc. are disabled. The molden file written out by PSI4 results in the wrong total charge when it is loaded into Molden, which shows there is a problem:

```
                        Mulliken Charges
                        ================
 
               O  -0.8774
               H   0.4308
               H   0.4316
 
Sum of Mulliken Charges:  -0.0150
```

Repeating the calculation in Gaussian and loading the log file into Molden gives:

```
                        Mulliken Charges
                        ================
 
               O  -0.8652
               H   0.4323
               H   0.4329
 
Sum of Mulliken Charges:  -0.0000
```

I've written out the orbitals from the Gaussian log back into a Molden file with Molden, to allow for a one-on-one comparison of the orbital coefficients. This revealed that coefficients 13, 14 and 15 of each orbital in the Molden file written by PSI4 are a factor **sqrt(3)** too large compared to what they should be. These are coefficients for the `xy`, `xz` and `yz` basis functions in a d shell. Other d functions are fine. I'll check for g and f functions as well...

tovrstra:
Here is another example: [molden_cartesian2.zip](https://github.com/psi4/psi4/files/4988357/molden_cartesian2.zip)

Comparing orbital coefficients, it seems that several of them are too large compared to what they should be for the Molden format:

- The 3 last coefficients of the Cartesian d shell: `xy`, `xz`, `yz`: factor **sqrt(3)** too large.
- Coefficients 4, 5, 6, 7, 8 and 9 of the Cartesian f shell: `xyy`, `xxy`, `xxz`, `xzz`, `yzz`, `yyz`: factor **sqrt(5)** too large.
- Coefficient 10 of the Cartesian f shell (last one): `xyz`: factor **sqrt(15)** too large.
- Coefficients 4, 5, 6, 7, 8,  and 9 of the Cartesian g shell: `xxxy`, `xxxz`, `yyyx`, `yyyz`, `zzzx`, `zzzy`: factor **sqrt(7)** too large.
[...]

PR #1510, opened 2019-01-31
Molden file format: double precision orbital energies, occupation numbers and molecular orbital coefficients

tovrstra:
## Description
The Molden format allows one to store data in high precision, which is beneficial for post-processing. This PR increases the number of digits and the format of the relevant numbers to double precision.

Fixes #1486. See also #1485 

I've tested with the following example:

```
molecule {
0 1
O          1.01646        0.05097        0.09714
H          1.98435        0.06069        0.05173
H          0.73761        0.24513       -0.81034
}

set basis cc-pVTZ

E, wfn = properties('ccsd', properties=['dipole'], return_wfn=True)
Da_so = wfn.Da()
SCa = core.doublet(wfn.S(), wfn.Ca(), False, False)
Da_mo = core.triplet(SCa, Da_so, SCa, True, False, False)
molden(wfn, 'psi4.molden', density_a=Da_mo)
```

The Molden files before and after the change are attached: [psi4.molden.zip](https://github.com/psi4/psi4/files/2817866/psi4.molden.zip)

Relevant part of the output of the Molden program when reading the molden file prior to the change. Note the wrong total charge:

```
                        Mulliken Charges
                        ================
 
               O  -0.4661
               H   0.2332
               H   0.2332
 
Sum of Mulliken Charges:   0.0003
```

Same output, but after increasing the precision: (Note the correct total charge.)

```
                        Mulliken Charges
                        ================
 
               O  -0.4663
               H   0.2331
               H   0.2331
 
Sum of Mulliken Charges:   0.0000
```

(FYI: this is my last Molden-related PR.)

## Todos
Notable points (developer or user-interest) that this PR has or will accomplish.
- [x] Write out orbital energies, orbital occupation numbers and molecular orbital coefficients in double precision in a Molden file.

## Questions
None

## Checklist
- No tests added
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

Issue #1486, opened 2019-01-21
Increase precision with which the occupation numbers are written in a Molden file.


tovrstra:
This is a follow-up on #504 and #1485, but essentially a different issue. When writing natural orbitals to a Molden file, the precision of the corresponding natural orbital occupation numbers is too limited for post-processing. (It is fine for visualization, not for e.g. density analysis.)

```
molecule {
0 1
O          1.01646        0.05097        0.09714
H          1.98435        0.06069        0.05173
H          0.73761        0.24513       -0.81034
}

set basis cc-pVTZ

E, wfn = properties('ccsd', properties=['dipole'], return_wfn=True)

nmopi = wfn.nmopi()
Da_so = wfn.Da()
SCa = core.Matrix("Overlap times Ca", nmopi, nmopi)
SCa.gemm(False, False, 1.0, wfn.S(), wfn.Ca(), 0.0)
Da_mo = Matrix.triplet(SCa, Da_so, SCa, True, False, False)
molden(wfn, 'ccsd_no.molden', density_a=Da_mo)
```

When grepping the occupation numbers, one can see the limited precision:

```
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0000
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0001
 Occup=  0.0002
 Occup=  0.0002
 Occup=  0.0004
 Occup=  0.0004
 Occup=  0.0004
 Occup=  0.0004
 Occup=  0.0004
 Occup=  0.0005
 Occup=  0.0005
 Occup=  0.0007
 Occup=  0.0007
 Occup=  0.0007
 Occup=  0.0008
 Occup=  0.0009
 Occup=  0.0010
 Occup=  0.0010
 Occup=  0.0013
 Occup=  0.0013
 Occup=  0.0059
 Occup=  0.0059
 Occup=  0.0066
 Occup=  0.0066
 Occup=  0.0069
[...]

PR #2229, opened 2021-07-01
Molden Writer update

sgoodlett:
## Description
Moved old Molden writer to Python, fixed normalization issue, and added test cases

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
Happy to see this up!  Go ahead and mark this ready for review. You want to mark with a single "X" rather than "Y " or " Y" so you get pretty checkmarks.

This PR will close #1973. I was pretty heavily involved in this, so I _probably_ shouldn't review this one.

tests/PEDRA.OUT__19028
```diff
@@ -0,0 +1,247 @@
+Point group: C1
```
JonathonMisiewicz, line 1: This file shouldn't be here.

psi4/src/psi4/libmints/writer.h
```diff
@@ -81,6 +81,22 @@ class PSI_API MoldenWriter {
                std::shared_ptr<Vector> OccB, bool dovirtual);
 };
 
+class PSI_API MoldenWriter {
```
JonathonMisiewicz, line 84: I think you misunderstood me earlier. Get rid of this "new" class and move the `PSI_DEPRECATED` warnings to the functions of the _existing_ `MoldenWriter` class. This is probably why the build is failing.

jturney (approved):

loriab:
Just checking that the extra precision added https://github.com/psi4/psi4/pull/1510/files didn't get lost.

JonathonMisiewicz:
> Just checking that the extra precision added https://github.com/psi4/psi4/pull/1510/files didn't get lost.

We studied numerical precision and couldn't justify going out to the **17** decimal places reported after #1510: the Python code couldn't reproduce the C++ code out that far. My memory is hazy, but I recall some errors that were normally in the 14th decimal place but could rarely reach to the 12th decimal place, which I suspected were related to the numerical precision of S^(-1/2). This made me want to cut back the number of decimal places to something less fragile.

I [checked in](https://github.com/psi4/psi4/issues/1973) about why we had 17 decimal places in the first place, and the person who made the change simply assumed that we had that many decimals of precision. In my judgment, we do not, and it was a mistake to ever report that many decimal places.

I _also_ checked that 10 decimal places was enough to keep #1486 fixed. (That was the issue that originally prompted the increased precision in our Molden files.) While the 4 decimal places of the pre-#1486 era had the bug, the 10 decimal places of this PR did not.

While we could _maybe_ go out to 11 or 12 decimal places, trying to get 17 decimal places of precision would require an analysis of round-off error in every line of the SCF code. This seems far beyond the scope of the PR.

psi4/src/psi4/libmints/writer.cc
```diff
@@ -46,6 +46,12 @@
 using namespace psi;
 ;
 
+
+MoldenWriter::MoldenWriter(std::shared_ptr<Wavefunction> wavefunction) : wavefunction_(wavefunction) {
```
jturney, line 50: Why did you duplicate the function from line 55. The compiler is reporting an error.  This should have come up on your local compile. One of them must be deleted.

jturney (changes requested):

JonathonMisiewicz:
We might as well fix #2238 while we're at it. The problematic line of code is [right here](https://github.com/psi4/psi4/pull/2229/files#diff-21e554583cfa696eeabb7d40e85e87689e1bd9f72ff88e593fdac893c6b61ac9R356).

JonathonMisiewicz (approved):
That's much better, thank you. I'll approve this in the interests of expediting 1.4. If anybody thinks we need three _independent_ reviewers, just say so.

jturney (approved):

loriab (approved):

andysim (approved):
Excellent work!  This is exactly the kind of thing the belongs py-side.  Kudos for the extensive tests.

psi4/driver/p4util/writer.py
```diff
@@ -198,3 +198,203 @@ def write_C_matrix(C, count):
...
+        NO_Ra = core.Matrix("NO Alpha Rotation Matrix", nmopi, nmopi)
+        occupation_a = core.Vector(nmopi)
+        MO_Da.diagonalize(NO_Ra, occupation_a, core.DiagonalizeOrder.Descending)
+        Ca = core.doublet(self.Ca(), NO_Ra, False, False)
+        epsilon_a = occupation_a
+        # Betas
+        #MO_Db = core.Matrix("MO Beta Density Matrix", nmopi, nmopi)
+        #MO_Db.transform(self.Db(), self.Cb().transpose())
+        MO_Db = self.Db_subset("MO")
+        MO_Db.print_out()
```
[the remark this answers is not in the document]
JonathonMisiewicz: Good catch. This is probably a debugging statement that survived.

PR #2230, opened 2021-07-01
STO-nG basis sets from the Basis Set Exchange

susilehtola:
## Description

Includes all the STO-nG basis sets from the Basis Set Exchange.

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
The new basis sets are fine. Updates of the old basis sets are causing test failures associated with changing basis set parameters. For example, STO-3G hydrogen basis sets are changing in the ninth decimal place because we only _had_ eight decimal places previously, causing some "check numbers to the ninth decimal place" tests to fail.

While I trust the numbers from the newer EMSL, we need to make sure full tests passes before merging this in.

loriab:
How about adding the new 2g/4g/5g files and not changing the 3g/6g? At a glance the former only adds Xe and the latter mostly changes the exponent marker from E to D (which is fine -- the parser knows how to handle it). Can run the diff gbs script if need to understand 3g/6g differences.

Please add a note to basis/NOTES about addition.

Please see if aux basis sets to notate in driver/qcdb/basisfamily_other

Issue #2231, opened 2021-07-02
DCT and Linear Dependencies


JonathonMisiewicz:
The `dct` module does not seem to like linear dependencies.

Example 1:
```
molecule { 
o 
h 1 1.0 
h 1 1.0 2 104.5
}

set {
  basis 'aug-cc-pvtz'
  s_tolerance 1e-3
}

energy('dct')
```

The computation runs to completion and gives energies that are at least _plausible_. But the output file is littered with
```
        There was a problem correcting the MO phases.
        If this does not converge, try ALGORITHM=TWOSTEP
 ```
 and `stderr` receives several MKL errors.

Example 2:
```
molecule { 
o 
h 1 1.0 
h 1 1.0 2 104.5
}

set {
  basis 'aug-cc-pvtz'
  dct_type df
  s_tolerance 1e-3
}

energy('dct')
```

Using the DF code triggers a segfault.

There may well be more examples I haven't found yet as I venture into other parts of the module. Either way, I'm going to be fixing this.

PR #2232, opened 2021-07-02
Fix GEMM errors in presence of linear dependencies

JonathonMisiewicz:
## Description
This PR fixes Example 1 of #2231, some GEMM errors in the DCT module in the presence of linear dependencies. There's more debugging to do, but this is a perfectly good standalone bugfix.

I also refactored the offending function. Less code duplication, less gratuitous pointers, and eliminated an offset.

This PR is anodyne, so getting this in for 1.4 would be appreciated.

## Checklist
- [x] No tests to add? The old code completed successfully because there was a back-up algorithm when the one that was acting up errored out.
- [x] `ctest -R dct` passes

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):

jturney (approved):

andysim (approved):
Good idea to refactor the spin cases via a helper function.  LGTM!

PR #2233, opened 2021-07-02
DF-DCT with Linear Dependencies

### LINKED ISSUE #2231: DCT and Linear Dependencies

JonathonMisiewicz:
## Description
This fixes Example 2 of #2231. DF-DCT no longer segfaults in the presence of linear dependencies.

Even with this merged in, #2231 should stay open until I've had a chance to do some more thorough testing of the module. This is, once again, a stand-alone bugfix.

## Questions
- [x] Test case added! ~~It would be good for me to write a test case for everything I check as part of #2231. Should this live in ctests or pytests? I'm not clear on test best practices at this point in Psi development, since we want to mover over to pytests eventually. Even then, _right now_ is a bad time to write the test since #2232 will change the reference output.~~

## Checklist
- [x] `ctest -R dct` passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dct/dct_df_tensor.cc
+++ b/psi4/src/psi4/dct/dct_df_tensor.cc
@@ -1287,7 +1287,7 @@ void DCTSolver::build_gbarGamma_RHF() {
         std::vector<std::pair<long int, long int>> subblock;
         for (int hp = 0; hp < nirrep_; ++hp) {
             int hq = hpq ^ hp;
-            std::pair<long int, long int> subsubblock(entrance, nsopi_[hp] * nsopi_[hq]);
+            std::pair<long int, long int> subsubblock(entrance, nmopi_[hp] * nmopi_[hq]);
             subblock.push_back(subsubblock);
             entrance += subsubblock.second;
         }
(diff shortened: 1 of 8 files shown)


loriab:
> It would be good for me to write a test case for everything I check as part of DCT and Linear Dependencies #2231. Should this live in ctests or pytests? I'm not clear on test best practices at this point in Psi development, since we want to mover over to pytests eventually. Even then, right now is a bad time to write the test since Fix GEMM errors in presence of linear dependencies #2232 will change the reference output.

In the beginning I may have had (and spoken) dreams of rewriting all the ctests in psiapi for pytest, but I've since been talked down, and the plan is to allow each ctest input.dat to stay where it is and be wrapped in a pytest test (like psi4numpy). There'll have to be extra install steps to position the ctest inputs into the python module directory. Upshot is it's fine to continue choosing psithon or psiapi style for writing new tests, mostly depending on testing style convenience and whether the test would serve as an easy model for users.

jturney (approved):

hokru (approved):

loriab (approved):

PR #2246, opened 2021-07-14
Port Björn Esser's patch for flexiblas support from Fedora

susilehtola:
## Description
Fedora switched to the use of [FlexiBLAS](https://www.mpi-magdeburg.mpg.de/projects/flexiblas) in Fedora 33 as a unified interface to various BLAS/LAPACK implementations.

This PR ports the patch by Björn Esser from Fedora to Psi4 per Fedora's guidelines of contributing any patches back upstream.

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz (approved):

loriab (approved):
Also add it https://github.com/psi4/psi4/blob/master/CMakeLists.txt#L77-L80 and https://github.com/psi4/psi4/blob/master/cmake/math/ConfigMath.cmake#L14-L20, please. then lgtm!

susilehtola:
> Also add it https://github.com/psi4/psi4/blob/master/CMakeLists.txt#L77-L80 and https://github.com/psi4/psi4/blob/master/cmake/math/ConfigMath.cmake#L14-L20, please. then lgtm!

Done. I've made Flexiblas the first one to look for, since it's backend agnostic and can also be used to wrap MKL.

andysim:
Looks good to me, but I just want to check one thing (that Lori knows way more about than I) before merging it.  Do we want to check for FLEXIBLAS first?  My only concern is that we could run into the OMP vs. GOMP issue by not knowing exactly which BLAS is wrapped.  Do either of you think that's something we're likely to run into?

loriab:
Hard to say. Psi links omp libraries based on whether it finds a library with "mkl" in the name. Flexiblas could cover that up and apply its own gomp here: https://gitlab.mpi-magdeburg.mpg.de/software/flexiblas-release/-/blob/master/cmake/FindMklOpenMP.cmake#L12-27 .

I'm thinking Flexiblas will be most common in Fedora for now. Our conda builds mostly hard-code the blas libraries list to bypass the math detection cmake. I'd feel a little better if flexiblas was behind MKL in priority since MKL is our special case, but I don't know of any definite disruptions.

andysim (approved):
Thanks for the fix.  LGTM!

PR #2247, opened 2021-07-14
Move Iteration Header Printing

JonathonMisiewicz:
## Description
This PR moves the printing of the SCF pre-iteration header from `scf_iterate` (where we had to check the number of iterations to know if we needed to print) to the SCF initializer (where no checks are needed).

This can and should wait until 1.5. This is part of a larger project to remove overloading of the `iterations_` variable in HF, needed to get MOM properly working.

## Questions
- [ ] Do we need to regenerate samples for this? The printing should be exactly the same except in pathological corner cases involving SOSCF instabilities at less than two iterations.

## Checklist
- [x] quick tests pass

## Status
- [x] Ready for review
- [ ] Ready for merge


PeterKraus (approved):

loriab (approved):

andysim (approved):

JonathonMisiewicz:
And with this, the merging of 1.5 PRs has officially begun.

PR #2254, opened 2021-07-20
Make the basis set parsing a bit more accurate

### LINKED ISSUE #2253: Basis set input is not sanity checked
I just made a fool of myself on the GAMESS-US forum by thinking there was a basis set issue in a calculation, since I couldn't reproduce the energy with Psi4.

The problem is the following: if I input a wrong format, e.g.
```
H 0
S 3 1.0
     1 3.386500000000E+01 2.549380000000E-02
     2 5.094790000000E+00 1.903730000000E-01
     3 1.158790000000E+00 8.521610000000E-01
S 1 1.0
     1 3.258400000000E-01 1.000000000000E+00
S 1 1.0
     1 1.027410000000E-01 1.000000000000E+00
P 1 1.0
     1 1.500000000000E+00 1.000000000000E+00
P 1 1.0
     1 3.750000000000E-01 1.000000000000E+00
****
```
this flies through Psi4; it does not check whether the basis set format is correct.

Obviously, here it is not since there is an additional enumeration in every entry.

susilehtola:
## Description

Psi4 isn't checking that basis sets are given in a sensible format at the moment. This pull fixes the issue somewhat.

Closes #2253 

## Todos

- [x] Corrupt basis sets are detected instead of used

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz (dismissed):

loriab (approved):

loriab (changes requested):
looks like some zapa sets have and extra col that the `shell` regex `$` doesn't agree with, https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/2zapa-nr.gbs#L152

JonathonMisiewicz (commented):
If we're updating basis sets, I'll need to re-check the PR, and this belongs in 1.5.

susilehtola:
> If we're updating basis sets, I'll need to re-check the PR, and this belongs in 1.5.

They're not updated; the sets in the BSE are identical to the ones in Psi4. Only the formatting has been fixed.

JonathonMisiewicz:
That's a different story. How can you be sure that just the formatting was changed?

loriab:
I was running the basis set diff-er (with mods) on these and got the summary below (black: unchanged to 8 decimals, yellow: changed, blue: new). So changes are fairly obscure except for 2zapa carbon.
Screen Shot 2021-07-23 at 10 19 12 AM
Magnitudes of changes seen in the next printouts. I did look into one of them (0.3945 <-- 0.39945 in 5zapa Li), and it tracked with psi's being the version 0 on BSE and the PR's being the version 1. So it's probably safe to assume this is all an upgrade. I'm with Jonathon on wanting to see what's changing.
Screen Shot 2021-07-23 at 10 17 10 AM
Screen Shot 2021-07-23 at 10 16 48 AM

susilehtola:
This is what the Basis Set Exchange's comparison tool reports.

[7zapa-nr.txt](https://github.com/psi4/psi4/files/6870639/7zapa-nr.txt)
[6zapa-nr.txt](https://github.com/psi4/psi4/files/6870640/6zapa-nr.txt)
[5zapa-nr.txt](https://github.com/psi4/psi4/files/6870641/5zapa-nr.txt)
[4zapa-nr.txt](https://github.com/psi4/psi4/files/6870642/4zapa-nr.txt)
[3zapa-nr.txt](https://github.com/psi4/psi4/files/6870643/3zapa-nr.txt)
[2zapa-nr.txt](https://github.com/psi4/psi4/files/6870644/2zapa-nr.txt)

A further study shows that 2zapa-6zapa are the sets from the Supporting Information of the nZaPa paper. However, 7zapa in Psi4 is not the basis from that paper.

According to notes in the MolSSI's internal Basis Set Exchange repository, some of the contraction coefficients were fixed in the nZaPa-CV paper and the rectified contractions should also be used in the non-CV sets i.e. the nZaPa sets.

Since it looks like these are different, I'll just revert to the old Psi4 copy and fix the formatting by hand....

loriab (approved):
yay for the ignore whitespace option on GH. thanks for the format fix!

JonathonMisiewicz:
Will approve once tests pass.

JonathonMisiewicz (approved):

robertodr (approved):

PR #2257, opened 2021-07-21
Watch for Const Wfn Members in DMRG

JonathonMisiewicz:
## Description
As a preliminary PR to giving DMRG a wavefunction, this PR changes many function signatures in DMRG to either not take a wavefunction, use a const reference to a wavefunction when the wavefunction doesn't change, and use a const reference to a Matrix when the Matrix doesn't change.

This PR is 100% cleanup and should probably be saved until 1.5.

## Checklist
- [x] `chemps2` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge **squash**


loriab (approved):
nice cleanup!

I don't think it's vital to keep this in sync with https://github.com/SebWouters/CheMPS2/blob/master/integrals/psi4plugins/dmrg.cc#L167. Just fyi that that exists as largely a duplicate in case unfamiliar.

robertodr (approved):

jturney (approved):

PR #2262, opened 2021-07-29
fix scf-coverage ref after phi_ao sph correction

loriab:
## Todos

- [x] #2210 revealed that an early test case had fallen prey to `phi_ao`'s Cartesian-only usefulness
- [x] mild enhancement to `diff_gbs` utility script so it can compare out-of-order basis sets
- [x] bump a few dep from-source-build versions

## Status
- [x] Ready for review
- [x] Ready for merge


andysim (approved):

JonathonMisiewicz (approved):
Thanks to Lori and Andy for the investigation!

hokru (approved):

PR #2266, opened 2021-08-04
UHF Linear Response

### LINKED ISSUE #2265: UHF Dipole Polarizability Calculation Fails
First reported [on forums](http://forum.psicode.org/t/polarizability-error-radical-anion/2240). The following MWE crashes:
```
molecule {
O
H 1 1.0
H 1 1.0 2 104.5
}

set reference uhf

psi4.properties("scf/cc-pVDZ", properties=["DIPOLE_POLARIZABILITIES"])
```

Errors are

```
Fatal Error: UHF::onel_Hx expect incoming vector to alternate A/B
```

and

```
! Fatal Error: UHF::onel_Hx expect incoming vector to alternate A/B                !
! Error occurred in file:                                                          !
!     /Users/jonathonmisiewicz/psi4/psi4/src/psi4/libscf_solver/uhf.cc on line:    !
!     744                                                                          !
! The most recent 5 function calls were:                                           !
```

It looks like an easy case of arguments not being passed correctly.

maxscheurer:
## Description
Fixes #2265.
This is a first attempt to make the feature work in `cpscf_linear_response`, but the code is a bit ugly and needs to be cleaned up...
~~I'm also not 100% sure why/where correct pre-factors/signs need to be added 🤯 😄~~

## Todos

- [x] UHF linear response properties
- [x] removed user-provided vectors from `cpscf_linear_response`

## Questions
- [x] How should user vectors be handled?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/response/scf_response.py
+++ b/psi4/driver/procrouting/response/scf_response.py
@@ -134,6 +134,7 @@ def cpscf_linear_response(wfn, *args, **kwargs):
     # vectors will be passed to the cphf solver, vector_names stores the corresponding names
     vectors = []
     vector_names = []
+    restricted = wfn.same_a_b_orbs()
 
     # construct the list of vectors. for the keywords, fetch the appropriate tensors from MintsHelper
     for prop in complete_dict:
@@ -141,11 +142,13 @@ def cpscf_linear_response(wfn, *args, **kwargs):
             for name, vec in zip(prop['vector names'], prop['vectors']):
                 vectors.append(vec)
                 vector_names.append(name)
-
         else:
             tmp_vectors = prop['mints_function'](mints)
             for tmp in tmp_vectors:
-                tmp.scale(-2.0)  # RHF only
+                if restricted:
+                    tmp.scale(-2.0)
+                else:
+                    tmp.scale(-1.0)
                 vectors.append(tmp)
                 vector_names.append(tmp.name)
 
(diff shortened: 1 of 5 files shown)


maxscheurer:
Besides having to clean up the code, I'm not sure how user-provided vectors should be handled... My suggestion is: If user vectors are provided in AO basis with shape `(nbf, nbf)` everything is handled as usual (transformation to alpha and beta part in my newly written code). If user vectors are provided in MO basis, we must insist that the user provide alpha/beta pairs, not a single vector... Comments?

maxscheurer:
I can't quite figure out how to pass custom vectors via Psithon/PsiAPI using `psi4.properties`.
If I pass a list of vectors (e.g., dipole integrals), [this](https://github.com/maxscheurer/psi4/blob/1528bcfda6d55d32712f3e6be4b21a45abe90e55/psi4/driver/procrouting/proc.py#L2962) code block obviously fails...

Directly passing vectors to `cpscf_linear_response` obviously works somehow, but I could not find anything in the test suite that explicitly covers the "user-provided vector"-case at all... Are there any plugins relying on calling `cpscf_linear_response` directly with custom vectors as arguments, @loriab?

hokru:
Could you clarify what you mean with passing custom vectors?

maxscheurer:
Sure, `cpscf_linear_response` in principle allows one to pass a list of operators [here](https://github.com/maxscheurer/psi4/blob/1528bcfda6d55d32712f3e6be4b21a45abe90e55/psi4/driver/procrouting/response/scf_response.py#L113), so you can build your own response property.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -156,61 +159,88 @@ def cpscf_linear_response(wfn, *args, **kwargs):
...
+        nbf = c_occ.shape[0]
+        # the vectors need to be in the MO basis. if they have the shape nbf x nbf, transform.
+        for i in range(len(vectors)):
+            shape = vectors[i].shape
+            if shape == (nbf, nbf):
+                vectors[i] = core.triplet(c_occ, vectors[i], c_vir, True, False, False)
+            # verify that this vector already has the correct shape
+            elif shape != (ndocc, nvirt):
+                raise ValidationError('ERROR: "{}" has an unrecognized shape ({}, {}).'
+                                      'Must be either ({}, {}) or ({}, {})'.format(
```
JonathonMisiewicz, line 179:
```suggestion
                                      'Must be either ({}, {}) for SO basis or ({}, {}) for OCC / VIR.'.format(
```

tests/scf-response3/input.dat
```diff
@@ -0,0 +1,30 @@
...
+no_reorient
+symmetry c1
+}
+
+set reference uhf
+set scf_type pk
+
+psi4.properties("scf/cc-pVDZ", properties=["DIPOLE_POLARIZABILITIES"])
+
+alpha_ref = [
```
JonathonMisiewicz, line 18: Where does the reference come from?

tests/scf-response3/input.dat
```diff
@@ -0,0 +1,30 @@
+# UHF Dipole Polarizability Test
```
JonathonMisiewicz, line 1: Can we test that for a closed-shell molecule, the UHF and RHF polarizabilities agree?

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -156,61 +159,88 @@ def cpscf_linear_response(wfn, *args, **kwargs):
...
+        vectors_a = {k: v for k, v in zip(vector_names, vectors[::2])}
+        vectors_b = {k: v for k, v in zip(vector_names, vectors[1::2])}
+        responses_a = {k: v for k, v in zip(vector_names, responses[::2])}
+        responses_b = {k: v for k, v in zip(vector_names, responses[1::2])}
+        output = []
+        names = prop['vector names']
+        dim = len(names)
+        buf = np.zeros((dim, dim))
+        for i, i_name in enumerate(names):
+            for j, j_name in enumerate(names):
```
JonathonMisiewicz, line 238: `for i, (vector_a, response_a) in enumerate(zip(vectors_a, responses_a)):` and same with below line. The loop contents will need to change slightly.

JonathonMisiewicz (commented):
I think that requiring pairs of (Alpha, Beta) is the right interface. That's what the [`compute_products` function](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/response/scf_products.py#L488) expects.

I'm not sure how the prefactors work either. I can do some digging, but it's not a priority for me.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -156,61 +159,88 @@ def cpscf_linear_response(wfn, *args, **kwargs):
...
+        c_occ = wfn.Ca_subset("AO", "OCC")
+        c_vir = wfn.Ca_subset("AO", "VIR")
+        nbf = c_occ.shape[0]
+        # the vectors need to be in the MO basis. if they have the shape nbf x nbf, transform.
+        for i in range(len(vectors)):
+            shape = vectors[i].shape
+            if shape == (nbf, nbf):
+                vectors[i] = core.triplet(c_occ, vectors[i], c_vir, True, False, False)
+            # verify that this vector already has the correct shape
+            elif shape != (ndocc, nvirt):
```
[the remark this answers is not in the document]
maxscheurer: It will be done, once we've decided how to handle the "user-vector" case.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -156,61 +159,88 @@ def cpscf_linear_response(wfn, *args, **kwargs):
...
+        # fetch wavefunction information
+        nmo = wfn.nmo()
+        ndocc = wfn.nalpha()
+        nvirt = nmo - ndocc
+
+        c_occ = wfn.Ca_subset("AO", "OCC")
+        c_vir = wfn.Ca_subset("AO", "VIR")
+        nbf = c_occ.shape[0]
+        # the vectors need to be in the MO basis. if they have the shape nbf x nbf, transform.
+        for i in range(len(vectors)):
```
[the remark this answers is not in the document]
maxscheurer: I agree.

tests/scf-response3/input.dat
maxscheurer: The program that should not be named... 😄 I will add a comment to the input file.

maxscheurer:
> I think that requiring pairs of (Alpha, Beta) is the right interface.

Another (more clean and less error-prone)  option would be to only allow for custom operators in the AO basis...

JonathonMisiewicz:
Without a use case, I have no strong opinions on user-supplied vectors.

maxscheurer:
That's the point... I don't see the use case of the user-supplied vectors anyways 😄

JonathonMisiewicz:
Looking at the commit history, this was added by DGAS years ago as part of a commit that suffered scope creep and was never actually used for anything. Let's check to see if anybody uses it downstream, but if not, I say remove it until we have an actual use case.

maxscheurer:
I completely agree. Which downstream packages should I check for usage of the functionality?

JonathonMisiewicz:
Check anything in the list of [official add-ons](https://github.com/psi4/psi4/blob/master/tests/pytests/addons.py#L51-L71). Lori may have some others.

maxscheurer:
Official add-ons seem to be 'clean' 👍

psi4/driver/procrouting/response/scf_response.py
maxscheurer: Mhm... I'm not sure I understand this comment correctly. The loop is over Cartesian components, so to speak, and `vectors_a` and `responses_a` are both dictionaries with the same keys, so zipping them and iterating over the identical keys is probably not what we want?

JonathonMisiewicz: On rereading the code, I was confused. I think this point depends on how we want the user to input vectors, so I'll leave it for now.

psi4/driver/procrouting/response/scf_response.py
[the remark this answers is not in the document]
maxscheurer: Done.

psi4/driver/procrouting/response/scf_response.py
maxscheurer: Obsolete, code removed...

maxscheurer:
I just committed a version with the user-vector stuff removed, more transparent handling of prefactors, and less code duplication for restricted/unrestricted. The code is now a lot easier to read, I hope the reviewers will agree 😉 
I'll now add the test cases suggested by @JonathonMisiewicz 👍

tests/scf-response3/input.dat
maxscheurer: Done.

tests/scf-response3/input.dat
maxscheurer: Done.

JonathonMisiewicz (approved):
Thank you very much!

andysim (approved):
This all looks good to me; thanks for the cleanup :) Regarding user-supplied vectors, I'd be happy to just not support them at all right now, given that we can't think of a use case.  If we are going to have them, restricting to AOxAO and/or Occ x Vir MO basis quantities is a very reasonable restriction to impose.

psi4/driver/procrouting/response/scf_response.py
```diff
@@ -92,132 +92,92 @@ def cpscf_linear_response(wfn, *args, **kwargs):
...
-                for j, j_name in enumerate(names):
-                    buf[i, j] = -1.0 * vectors[i_name].vector_dot(responses[j_name])
-
-            output.append(buf)
+        names = prop['vector names']
+        dim = len(names)
+        buf = np.zeros((dim, dim))
+        for i, i_name in enumerate(names):
+            for j, j_name in enumerate(names):
+                buf[i, j] = pref * vectors[f"{i_name}_a"].vector_dot(responses[f"{j_name}_a"])
```
hokru, line 170: I don't understand why `pref` changed from -1.0 to -2.0.

maxscheurer: I removed the prefactor -2.0 from the RHS in the RHF case, which was implicitly yielding another factor of 2.0 for the solution from CPHF... now handling of the prefactors is consistent, i.e., the prefactor did not change, it is just explicitly written out in one spot now.

hokru: ah okay! Now I see it, thanks.

hokru (approved):

PR #2270, opened 2021-08-10
MOM Bugfix and Test

### LINKED ISSUE #1692: Changes in MOM from PSI4 1.1 to 1.4a2.dev78
Hello everyone. Thank you for working on this fantastic project, I've been using it a lot lately and I'm both grateful and impressed.

That said, I've been thoroughly trying to break the code by doing some... unconventional calculations.

I'm converging core holes in molecules. In the simple case of water, this leads to the following total energies, where _opt refers to the neutral GS RHF molecule and +_homo-i refers to the UHF cation, on the GS RHF geometry, with the hole in the homo-ith MO.

[psi4_11.tar.gz](https://github.com/psi4/psi4/files/3469634/psi4_11.tar.gz)

(PSI4 1.1)
water_opt -7.606667516613838e+01
water+_homo -7.566072290801020e+01
water+_homo-1 -7.558091679634278e+01
water+_homo-2 -7.541230326047109e+01
water+_homo-3 -7.423771580655676e+01
water+_homo-4 -5.571830715553878e+01

Noticeably, pulling an electron from the homo-3 and homo-4 require some crafty work, but **I have managed to converge this calculations with several basis sets and in DFT, too, using damping and other tricks, all in PSI4 1.1. Energies are qualitatively the same alway**s. All the output files are attached, in psython mode (I could not make MOM to work from python).

Now then, I wanted to check some things, so I printed the density matrices from .fchk files and plugged them to PySCF, projecting them properly and so on, just to evaluate the energy with UHF on top of those density matrices. I get the same energies for the neutral molecule and the homo, homo-1 and homo-2 cations. I get VERY different energies from the homo-3 and homo-4 states. Now thats weird, because the energies above make sense. So I assumed, okay, maybe there's something odd with the .fchk writer module. I've seen its been updated, so lets try a newer version of PSI4.

So I got the 1.4a2.dev78 and tried to run my calculations using the newest available dev release, from scratch. I get exactly the same energies, except for the homo-4  state, for which I get some very funky results:

[psi4_14.tar.gz](https://github.com/psi4/psi4/files/3469641/psi4_14.tar.gz)

(Psi4 1.4a2...)
water_opt -7.606667516615393e+01
water+_homo -7.566072290809453e+01
water+_homo-1 -7.558091679647397e+01
water+_homo-2 -7.541230326053801e+01
water+_homo-3 -7.423771580754870e+01
water+_homo-4 **9.360901611130924e+00**

And that's the story! I would try to read the .fchk files and see if I can evaluate the UHF energy on top of them with Psi4 intead of PySCF next, but this thing appeared and I guess some checks are due to know what is going on.

Thank you for your patience! I hope its nothing serious!

JonathonMisiewicz:
## Description
Closes #1692 and adds some of those examples as test cases. The heart of the PR is that MOM no longer sets the iteration number to 0, which would triggered logic resetting some wavefunction variables.

## Todos
- [x] Fixes a bug where MOM could clear all electrons

## Questions
- [x] The new tests take about 20 seconds. Do you want me to reduce the basis set size, to speed this up?

## Checklist
- [x] Tests added for newly working features
- [x] `scf` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):
lgtm, thanks! I think increasing the variety of basis sets hit is probably good, and 20 s is liveable.

psi4/src/psi4/libscf_solver/mom.cc
```diff
@@ -594,7 +593,6 @@ void HF::MOM_start() {
     outfile->Printf("\n                        Total Energy        Delta E      Density RMS\n\n");
 }
 void HF::MOM() {
-    // Go MOM go!
```
jturney, line 597: Booo for removing a fun comment.

jturney (approved):

hokru (approved):

PR #2271, opened 2021-08-10
Fix in handling of frozen core orbitals in SAPT

jeffschriber:
## Description
In very rare cases, the number of frozen core orbitals in computing the dMP2 correction in SAPT was inconsistent between the dimer and monomer calculations. I only encountered this when one monomer was a charged alkali metal. For example, when computing the dMP2 correction for Na-H20, `freeze_core true` results in one frozen orbital each for Na and H2O in the separate monomer computations, but six frozen orbitals in the dimer computation. Separately, these values make sense, but in this context they result in a meaningless dMP2 value. 

## Todos

- [X] Set # frozen core for dimer as sum of monomer # frozen core

## Checklist
- [X]  Add a warning in the manual

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/doc/sphinxman/source/sapt.rst
+++ b/doc/sphinxman/source/sapt.rst
@@ -63,6 +63,12 @@ SAPT: Symmetry-Adapted Perturbation Theory
    possible discrepancies with prior versions of the code on the order of
    0.01 kcal/mol. See https://github.com/psi4/psi4/issues/1677
 
+.. caution:: August 2021, the number of frozen core orbitals used in the dMP2 computations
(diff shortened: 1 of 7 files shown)


loriab (approved):

hokru:
Does this fix #2012 ?

jeffschriber:
This doesn't fix #2012, since in that case it seems that the number of frozen core orbitals in the dimer is sum of the number of frozen core orbitals in each monomer (at least for `freeze_core -1`, an error is thrown with `freeze_core -2`).

JonathonMisiewicz (approved):

Issue #2272, opened 2021-08-11
MBIS fails via QCEngine


jthorton:
Using psi4-1.4rc1, I was able to compute `MBIS_CHARGES` via QCEngine however after PR #2127 this now fails with the following error message. While this definitely needs fixing I think it might also be nice to have the option to turn off this feature and only calculate the ratio if passed as an scf property?
```
RuntimeError: 
Fatal Error: Wavefunction::scalar_variable: Requested variable MBIS FREE ATOM O VOLUME was not set!
Error occurred in file: /Users/github/builds/conda-builds/psi4-multiout_1628080613326/work/psi4/src/psi4/libmints/wavefunction.cc on line: 1335
```

MWE 
```python
import qcelemental as qcel
from qcelemental.models import molecule
import qcengine as qcng

mol = qcel.models.Molecule.from_data("""
     O  0.0  0.000  -0.129
     H  0.0 -1.494  1.027
     H  0.0  1.494  1.027
     """)

# build the mbis task
model = qcel.models.AtomicInput(
    molecule=mol,
    driver="energy",
    model={"method": "SCF", "basis": "sto-3g"},
    keywords={"scf_properties": ["MBIS_CHARGES"]}
)

result = qcng.compute(model, "psi4")

print(result.dict())
```
versions
psi4-1.4+9485035 
qcengine-0.19.0
qcelemental -0.21.0

loriab:
Apparently very easy to fall into endless recursion with MBIS volumes due to some necessary inputs also calculated by `oeprop()` and involving calls to `energy()`.

You're interested in MBIS CHARGES as standalone (w/o volumes and extents) to avoid the cost of all the atomic calcs since you don't need the volume?

jthorton:
Thanks for the quick response,  that makes sense!

>You're interested in MBIS CHARGES as standalone (w/o volumes and extents) to avoid the cost of all the atomic calcs since you don't need the volume?

Sort of the only bit I need is the charges and volumes but not the volume ratio so not the free atom calculations added in #2127. So basically the performance before that PR my idea would be that the ratios are only calculated if something like `MBIS RATIOS` is passed which like you say would also avoid the cost of the extra calculations.

djcole56:
[...]

PR #2273, opened 2021-08-12
split mbis oeprop

loriab:
## Description
This is a start to addressing #2272 

## Todos

- [x] `MBIS_CHARGES` and `MBIS_VOLUME_RATIOS` are now separate oeprop tasks to the user but still reusing code.
- [ ] `oeprop(wfn, "MBIS_VOLUME_RATIOS")` should be fine (indep oeprop fn), but `set scf_properties mbis_volume_ratios; energy("scf")` will still fail as #2272 reported because those are `OEProp` class instantiations called from proc.py, and the free atom volumes aren't available. The oeprop.cc code could exit gracefully when free atom volumes aren't available, but having different properties lists for the two calling routes isn't good.
- [ ] so why not add the loc from oeprop() to `OEProp`s in proc.py so that atom volumes are available? nice thought, but the fn that produces them itself calls oeprop() and energy() and descends into endless recursion. I haven't sought the logic that makes this all work together.
- [ ] add tests. probably some of the existing ones will break for only calling one mbis property but checking volrat.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/mbis-1/input.dat
+++ b/tests/mbis-1/input.dat
@@ -76,7 +76,7 @@ set {
 }
 
 e, wfn = energy('hf/cc-pvdz', return_wfn=True)
-oeprop(wfn, 'MBIS_CHARGES', "MBIS_VOLUME_RATIOS", title='H20 SCF')
+oeprop(wfn, 'MBIS_CHARGES', title='H20 SCF')
 
 #NOTE: wfn.array_variable gives you flattened atomic multipole arrays; for expanded arrays, use wfn.variable
 charges = wfn.array_variable('MBIS CHARGES')
@@ -85,7 +85,6 @@ quadrupoles = wfn.array_variable('MBIS QUADRUPOLES')
 octupoles = wfn.array_variable('MBIS OCTUPOLES')
 avols = wfn.array_variable('MBIS RADIAL MOMENTS <R^3>')
 vwidths = wfn.array_variable('MBIS VALENCE WIDTHS')
-vratios = wfn.array_variable('MBIS VOLUME RATIOS')
 
 compare_matrices(charges_ref, charges, 5, "MBIS Charges")             #TEST
 compare_matrices(dipoles_ref, dipoles, 5, "MBIS Dipoles")             #TEST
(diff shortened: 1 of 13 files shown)


JonathonMisiewicz:
The code looks anodyne, but I would appreciate information about how this fixes #2272.

From my understanding, #2272 is that when QCEngine tasks Psi with computing MBIS charges, Psi (a) would try to compute atom volumes but (b) would not have set some necessary variables. I would like to understand _why_ the call coming from QCEngine rather than a Psi input file means we get both (a) and (b) and how this PR disrupts that mechanism.

loriab:
> The code looks anodyne, but I would appreciate information about how this fixes #2272. From my understanding, #2272 is that when QCEngine tasks Psi with computing MBIS charges, Psi (a) would try to compute atom volumes but (b) would not have set some necessary variables. I would like to understand why the call coming from QCEngine rather than a Psi input file means we get both (a) and (b) and how this PR disrupts that mechanism.

* pre-MBIS-volume-capability behavior: requesting the mbis_charges property through `set scf_properties ["mbis_chargs"]; energy()` route _or_ `energy(); oeprop(..., "mbis_charges")` route called the "mbis" fns in oeprop.cc which were self-contained like all the other oeprop fns. This is the behavior that #2272 used (former route in particular) and want to use again.
* current / post-MBIS-volume-capability behavior: the "mbis" fns in oeprop.cc now compute two properties -- the original self-contained charges and the volume ratios that need atom volumes for each atom, info that's pre-computed (from energy() calcs) on orders from a line in oeprop() fn. So only the latter route works, and the former route fails for lack of atom volumes. All well and good if you want mbis volumes but a regression if you only care about mbis charges.
* this PR behavior: there are now two oeprop properties and "mbis" fns in oeprop.cc -- one for charges and one for volume ratios. so "mbis_charges" works with both routes (healing 2272), and "mbis_volume_ratios" still works only for the latter route (behavior unchanged, except you don't get volumes for free when request charges).
* future after #2299 addressed: mbis volumes will work with both routes.

Is this any clearer? It isn't the QCEngine involvement that causes trouble -- it's the breakdown in consistency btwn the two calling routes. And if you try the straightforward way of making them consistent (compute free atom volumes in scf_helper fn like the oeprop fn does), you fall into recursion.

JonathonMisiewicz (approved):

tests/mbis-1/input.dat
```diff
@@ -76,7 +76,7 @@ set {
 }
 
 e, wfn = energy('hf/cc-pvdz', return_wfn=True)
-oeprop(wfn, 'MBIS_CHARGES', title='H20 SCF')
+oeprop(wfn, 'MBIS_CHARGES', "MBIS_VOLUME_RATIOS", title='H20 SCF')
```
PeterKraus, line 79: Two things: 
- it would make sense to check that `oeprop(wfn, 'MBIS_CHARGES')` works independently from running `oeprop(wfn, 'MBIS_VOLUME_RATIOS')`; either this test, or `mbis-2` below are good candidates.
- the quotation marks are inconsistent - is this also because of `black`?

PeterKraus (changes requested):

tests/mbis-1/input.dat
loriab: good idea -- I've split `mbis-1` up now.

the quotes are just my style has changed from single to double to placate `black`. the driver codebase will get a `black` run someday, I expect, so I often convert any lines I touch for other reasons to minimize churn at format time. these tests won't likely get that treatment.

PeterKraus (approved):

tests/mbis-1/input.dat
PeterKraus: The reference test outputs are missing, but apart from that it looks OK so I've approved it.

jturney (approved):

PR #2277, opened 2021-08-15
Unify fcidump 1e and 0e integral spacings with 2e integrals

brianz98:
## Description
The fcidump.py file writes out 2e integrals by calling fcidump_helper.cc, in the `%28.20E%4d%4d%4d%4d` (cpp) format, and in fcidump.py the additional 1- and 0- electron integrals are written in the `{:29.20E} {:4d} {:4d} {:4d} {:4d}` (Python) format. The whitespace in between the Python format string causes the 1/0e integrals fields to be 1 wider than the 2e integrals, causing problems in reading the file. E.g. in Fortran the 2e integrals can be read normally by `READ(*,'(1X,G28.20E2,4I4)') Z,I,J,K,L`, but will fail to read the 1/0e integrals properly.

## Todos
- [x] Update fcidump.py
- [x] Update fcidump test outputs 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


tests/fcidump/output.ref
```diff
@@ -1,1061 +1,24 @@
...
-	Ne, C1, cc-pVDZ, UHF reference, integrals in FCIDUMP format. 2-electron energyPASSED
-	Ne, C1, cc-pVDZ, UHF reference, integrals in FCIDUMP format. SCF total energyPASSED
-	Ne, C1, cc-pVDZ, UHF reference, integrals in FCIDUMP format. MP2 correlation energyPASSED
-	Ne, C1, cc-pVDZ, UHF reference, integrals in FCIDUMP format.......PASSED
-
-    Psi4 stopped on: Monday, 02 July 2018 04:11PM
-    Psi4 wall time for execution: 0:00:01.14
-
-*** Psi4 exiting successfully. Buy a developer a beer!
+    FCIDUMP header........................................................................PASSED
```
loriab, line 1: new isn't the same file as replacement. probably the output.ref doesn't need regenerating since the code changes in the PR are only to the fcidump file.

loriab (approved):
good catch, thank you!

(for anyone else looking for the spacing changes, go to the end of the intdump.ref files.)

robertodr (approved):
Good catch! The tests read in the FCIDUMP file without relying on the format and that's why we missed this.

tests/fcidump/output.ref
brianz98: Oops missed that one, you're right, I've just reverted output.ref

PR #2280, opened 2021-08-19
Fix fractional occupation with symmetry

JonathonMisiewicz:
## Description
Closes #1673. This PR fixes an indexing error causing the wrong orbitals to be scaled during fractional occupation, when there are multiple irreps.

Now that the SCF module has its correctness errors fixed, I can get to some convergence improvements.

## Todos
- [x] Fixed a bug causing incorrect energies during fractional occupation for systems with point group symmetry

## Checklist
- [x] Tests added for any newly working features
- [x] frac tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2290, opened 2021-09-03
Fix 0 Frac Occupancy

### LINKED ISSUE #2284: `frac_traverse` broken
Discovered when running tests for a future PR. #2280 broke the `frac-traverse` test. It tried to set a fractional occupation of 0, which that test expressly disallowed.

I'll need to figure out why `frac_traverse` needs it and decide what the appropriate remedy is.

JonathonMisiewicz:
## Short Description
This PR reverts part of #2280 and restores the ability of `frac` to handle zero occupation number. At the cost of storing extra C matrices, this fixes a pre-#2280 bug in the handling of zero occupation number. Closes #2284.

## Long Description
The `frac` code is designed to allow SCF with fractional occupation numbers. Physically, this is an ensemble of Slater determinants, where the probabilities of a given orbital being occupied are uncorrelated. This is implemented by changing the C matrix from meaning "thing that contains orbitals" to "thing that produces the density", and incorporating the occupation numbers into the definition of the C matrix by multiplying by the occupation numbers. These are divided out after SCF to restore the original meaning of C.

Now, in the case of zero occupation number, multiplying by zero cannot be inverted, so the pre-#2280 code just left the orbitals. This is nonsense behavior, and #2280 banned the case of zero occupation number: the orbital should never have been included in `frac` in the first place if your use case was to describe a single ensemble. What I didn't realize was that there were legitimate use cases for this: if you had multiple ensembles and wanted to track what happens as the occupation number of one goes to zero. This is used by `frac_traverse`, so I broke the test. This slipped through the cracks because the test wasn't labeled as a frac test.

This PR allows frac to treat zero occupation number, and it corrects the previous pathological handling of the C matrix with zero occupation number by changing how we generate the restored C matrix: instead of inverting multiplication, we store the matrix where we never multiplied obitals by occupation numbers in the first place. This change keeps the symmetry fixes of #2280.

## Checklist
- [x] frac tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/frac.cc
+++ b/psi4/src/psi4/libscf_solver/frac.cc
@@ -27,12 +27,21 @@
  */
 
 /*
- * frac.cc. So close to a great Battlestar Galactica joke.
+ * frac.cc.
+ *
+ * frac handles SCF with fractional occupation numbers. Each fractionally occupied orbital is regarded as occupied
+ * for purposes of docc, socc, nalpha, nbeta, etc. However, each orbital's contribution to the density matrix is
+ * scaled by the fractional occupation number. The density matrix no longer describes a single Slater determinant
+ * but an ensemble of Slater determinants. This is useful for finite-temperature theory and removing artifacts
+ * where some orbitals but not others of a degenerate set are held occupied, artificially breaking the degeneracy.
+ *
  * The way this works:
  * 1) form_C returns 1's normalized C matrices.
  * 2) find_occupation determines the occupations of said matrix via either Aufbau or MOM selection
  * 3) find_occupation then calls this, which renormalizes the C matrices with \sqrt(val) for each frac occ
+ *    These renormalized C matrices produce the desired density matrices, but are not orthonormal.
  * 4) find_D is then computed with the renormalized C matrices, and everything is transparent until 1) on the next cycle
+ * 5) frac_renormalize sets the wfn's C matrices to be un-renormalized.
  *
  * Some executive decisions:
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/libscf_solver/frac.cc
```diff
@@ -112,15 +115,8 @@ void HF::frac() {
                     nbeta_++;
             }
 
-            // Throw if the user is insane
             if (val < 0.0)
-                throw PSIEXCEPTION(
-                    "Fractional Occupation SCF: Psi4 is not configured for positrons. Please annihilate and start "
-                    "again");
```
loriab, line 119: I liked this errormsg.

loriab (approved):
Thanks for the exposition on the PR -- it helped me considerably.

psi4/src/psi4/libscf_solver/frac.cc
```diff
@@ -112,15 +115,8 @@ void HF::frac() {
...
-            if (val == 0.0)
-                throw PSIEXCEPTION(
```
jturney, line 121: Is the `val == 0.0` check not needed anymore?

JonathonMisiewicz: Correct. As I said in the long description, "This PR allows frac to treat zero occupation number", which it was able to do (albeit incorrectly) pre-#2280.

jturney (approved):

fevangelista (approved):
Thanks for this PR! Very helpful explanations and nice streamlining of the code.

PR #2293, opened 2021-09-13
misc. qcdb compatibility

loriab:
## Description

~Ignore this for now. This is so I can get a build snapshot.~ It might well fail tests b/c needs qcel and qcng snapshots. Full tests is clean locally. Changes for the most part are RTG except for two lines noted below.

UPDATE: go ahead and review, please. I'll take out the 2 improper lines noted below before merge.

## Todos

- [x] report FD G & H more fully in core and wfn vars
- [x] add MP2 Hessian testing to stdsuite
- [x] I guess the a-ccsd(t) docs changes didn't make it into #2241 ?
- [x] some cfour interface expansions that are really only useful for internal debugging. use the regex harvesting from qcng, not here
- [x] this bit https://github.com/psi4/psi4/compare/master...loriab:sept21qcdb?expand=1#diff-3abbb4f37a59cc2e6cfbff3f4cb73b312dfd94044a19f38b00bf7ae140db13e7R540-R541 needs removing before actual merge is considered. proper way to handle would be if the select_ccsd_gradient was smart enough to say "can't do" to FC gradients, so it'd fall back to FD by ene.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/qcdb/cfour.py
```diff
@@ -54,7 +54,10 @@ def harvest_output(outtext):
 
     #for outpass in re.split(r'--invoking executable xjoda', outtext, re.MULTILINE):
     for outpass in re.split(r'JODA beginning optimization cycle', outtext, re.MULTILINE):
-        psivar, c4coord, c4grad = harvest_outfile_pass(outpass)
+        import qcengine as qcng
```
PeterKraus, line 57: `import` in a for cycle?

psi4/driver/qcdb/cfour.py
```diff
@@ -80,7 +83,7 @@ def harvest_output(outtext):
     return pass_psivar[retindx], pass_coord[retindx], pass_grad[retindx]
 
 
-def harvest_outfile_pass(outtext):
+def local_harvest_outfile_pass(outtext):
```
PeterKraus, line 86: What's this function for, now?

tests/mints5/input.dat
```diff
@@ -311,6 +311,7 @@ compare_strings("D3", ethane_gauche.get_full_point_group(), "D3 point group");
 
 
 molecule triplet_ethylene {
+  0 3
```
PeterKraus, line 314: Is there a new output reference file for this test?

tests/pytests/addons.py
```diff
@@ -68,6 +68,7 @@ def is_nvidia_gpu_present():
     "simint": psi4.addons("simint"),
     "snsmp2": which_import("snsmp2", return_bool=True),
     "v2rdm_casscf": which_import("v2rdm_casscf", return_bool=True),
+    "qcdb": False,  # capabilities of in-psi and out-of-psi qcdb not aligned
```
PeterKraus, line 71: I don't really understand what this does. Is this just to make sure that `qcdb`-requiring tests should not be run?

PeterKraus (changes requested):

psi4/driver/driver.py
```diff
@@ -1602,6 +1602,14 @@ def hessian(name, **kwargs):
         wfn.set_gradient(G0)
 
         # Explicitly set the current energy..
+        if isinstance(lowername, str) and lowername in procedures['energy']:
+            # this correctly filters out cbs fn and "hf/cc-pvtz"
```
JonathonMisiewicz, line 1606: Why is filtering out "hf/cc-pvtz" correct? I'd expect that to register as an HF TOTAL HESSIAN, HF TOTAL GRADIENT.

psi4/driver/qcdb/cfour.py
```diff
@@ -533,6 +536,10 @@ def harvest_outfile_pass(outtext):
 #        psivar['CURRENT CORRELATION ENERGY'] = psivar['%s CORRELATION ENERGY' % (mobj.group('fullCC')]
 #        psivar['CURRENT ENERGY'] = psivar['%s TOTAL ENERGY' % (mobj.group('fullCC')]
 
+    if 'CCD TOTAL ENERGY' in psivar and 'CCD CORRELATION ENERGY' in psivar:
```
JonathonMisiewicz, line 539: Please make a for loop for this string of boilerplate code.

psi4/driver/qcdb/cfour.py
JonathonMisiewicz: I don't understand your confusion. It takes the contents of a CFOUR output file as an argument and returns data structures containing variables (energies, gradients, etc.) scraped from the output file.

Admittedly, the docstring looks like somebody stopped partway through writing it.

tests/pytests/standard_suite_runner.py
```diff
@@ -59,6 +59,8 @@ def runner_asserter(inp, subject, method, basis, tnm):
     scf_type = natural_values[scf_type]
 
     atol = 1.0e-6
+    if driver == "hessian":
+        atol = 2.0e-6  # todo more elaborate
```
JonathonMisiewicz, line 63: Please elaborate on this comment to be meaningful to others.

tests/pytests/addons.py
JonathonMisiewicz: Seconded.

tests/mints5/input.dat
loriab: You'd think so, but this test is solely concerned with geometry and orientation and doesn't actually use the multiplicity. I have a test in another repo that does store HF energies for all these, so I noticed the multiplicity was missing.

tests/pytests/addons.py
loriab: right, this is signaling to pytest that any tests with `using("qcdb")` should plain not run, as I don't want to try to ensure that  qcdb imported from driver and qcdb indep module are exactly the same. In practice, it hits exactly one test, https://github.com/psi4/psi4/blob/master/tests/pytests/test_qcng_dftd3_mp2d.py#L323-L342, so not a big loss.

doc/sphinxman/source/cc.rst
```diff
@@ -117,42 +117,46 @@ continuum model (PCM) in the PTE approximation [Cammi:2009:164104]_, see
 
 .. table:: Current coupled cluster capabilities of |PSIfour|
 
-   +---------------+------------+-----------+-----------+---------------+-----------+
-   | Method        | Reference  | Energy    | Gradient  | Exc. Energies | LR Props  |
-   +===============+============+===========+===========+===============+===========+
-   | CC2           | RHF        | Y         | Y         | Y             | Y         |
-   +               +------------+-----------+-----------+---------------+-----------+
-   |               | UHF        | Y         | ---       | N             | ---       |
```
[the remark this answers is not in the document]
loriab: I don't think there is a difference. Looks like once upon a time TDC meant N := NYI and --- := not applicable. Then I shuffled them to --- := NYI (more visually distinct from Y) and n/a := not applicable. https://github.com/psi4/psi4/commit/2aa93f0b14d4fecaaca7409cb096264d386276e4 . But now all three forms are present. Agree could use cleanup.

tests/pytests/addons.py
PeterKraus: So what is the roadmap on resolving that properly, eventually?

psi4/driver/qcdb/cfour.py
PeterKraus: No, I mean, do we still need it, if we're using `qcng.programs.cfour.harvester.harvest_outfile_pass()`?

psi4/driver/driver.py
loriab: `HF TOTAL HESSIAN` is what we'd like, but without the filtering, get `HF/CC-PVTZ TOTAL HESSIAN`

psi4/driver/qcdb/cfour.py
loriab: Yep, that was silly.

psi4/driver/qcdb/cfour.py
loriab: You're right, it's dead code, as the new docstring admits. I get to delete the whole file soon, so I was avoiding deleting the fn piecemeal.

tests/pytests/addons.py
loriab: I'll probably delete the test. It's a shared test bwn qcng and psi4 that hasn't gotten run much under normal user conditions. Later `using("qcdb")` can be revived when the driver and independent flavors have less overlap, and the shared tests are more extensive.

tests/pytests/standard_suite_runner.py
loriab: done. there's an elaborate e/g/h absolute and relative tolerance criteria implemented in qcdb and qcng that hasn't been ported to psi4 yet.

psi4/driver/driver.py
JonathonMisiewicz: It should be possible to strip the basisset at this point so we can still mark it as an HF hessian. Let's file an issue once this PR is in, so we don't lose track of this.

doc/sphinxman/source/cc.rst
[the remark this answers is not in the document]
JonathonMisiewicz: File an issue, as an item for a future PR?

JonathonMisiewicz (approved):
Some items in review came up as "things to clean up in future PRs" and should be logged in issues _somewhere_, but LGTM otherwise.

psi4/driver/driver.py
loriab: I think the routing in DDD is different enough that this won't come up, but I'm adding it to the to-be-tidied list

PeterKraus (approved):
Cool, thanks for the explanation, and making #2303. LGTM.

jturney (approved):

PR #2296, opened 2021-09-21
Document `epsilon_a_subset` and `epsilon_b_subset`

JonathonMisiewicz:
## Description
Documents `epsilon_a_subset` and `epsilon_b_subset`

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_wavefunction.cc
+++ b/psi4/src/export_wavefunction.cc
@@ -155,8 +155,38 @@ void export_wavefunction(py::module& m) {
              "Returns the requested Beta Density subset.")
         .def("epsilon_a", &Wavefunction::epsilon_a, "Returns the Alpha Eigenvalues.")
         .def("epsilon_b", &Wavefunction::epsilon_b, "Returns the Beta Eigenvalues.")
-        .def("epsilon_a_subset", &Wavefunction::epsilon_a_subset, "Returns the requested Alpha Eigenvalues subset.")
-        .def("epsilon_b_subset", &Wavefunction::epsilon_b_subset, "Returns the requested Beta Eigenvalues subset.")
+        .def("epsilon_a_subset", &Wavefunction::epsilon_a_subset, R"pbdoc(
+              Returns the requested Alpha orbital energies subset.
+              Parameters
+              ----------
+              basis : str
+                  {'AO', 'SO', 'MO'}
+                  MO or SO select for Pitzer-ordering the return vector in the full computational point group symmetry.
+                  AO selects for Pitzer-ordering the return vector without point group symmetry.
+              subset : str
+                  {'ALL', 'ACTIVE', 'FROZEN', 'OCC', 'VIR', 'FROZEN_OCC', 'ACTIVE_OCC', 'ACTIVE_VIR', 'FROZEN_VIR'}
+                  Which subspace of orbital energies should be returned?
+              Returns
+              -------
+              Vector
+                  A Pitzer-ordered vector of the orbitals.
+         )pbdoc")
+        .def("epsilon_b_subset", &Wavefunction::epsilon_b_subset, R"pbdoc(
+              Returns the requested Beta orbital energies subset.
+              Parameters
+              ----------
+              basis : str
+                  {'AO', 'SO', 'MO'}
+                  MO or SO select for Pitzer-ordering the return vector in the full computational point group symmetry.
+                  AO selects for Pitzer-ordering the return vector without point group symmetry.
+              subset : str
+                  {'ALL', 'ACTIVE', 'FROZEN', 'OCC', 'VIR', 'FROZEN_OCC', 'ACTIVE_OCC', 'ACTIVE_VIR', 'FROZEN_VIR'}
+                  Which subspace of orbital energies should be returned?
+              Returns
+              -------
+              Vector
+                  A Pitzer-ordered vector of the orbitals.
+         )pbdoc")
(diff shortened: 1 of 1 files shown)


psi4/src/export_wavefunction.cc
```diff
@@ -155,8 +155,38 @@ void export_wavefunction(py::module& m) {
...
+              Parameters
+              ----------
+              basis : str
+                  {'AO', 'SO', 'MO'}
+                  MO or SO select for Pitzer-ordering the return vector in the full computational point group symmetry.
+                  AO selects for Pitzer-ordering the return vector without point group symmetry.
+              subset : str
+                  {'ALL', 'ACTIVE', 'FROZEN', 'OCC', 'VIR', 'FROZEN_OCC', 'ACTIVE_OCC', 'ACTIVE_VIR', 'FROZEN_VIR'}
+                  Which subspace of orbital energies should be returned?
+              Returns
```
loriab, line 185:
```suggestion
              Returns the requested Alpha orbital energies subset.

              Parameters
              ----------
              basis : str
                  {'AO', 'SO', 'MO'}
                  MO or SO select for Pitzer-ordering the return vector in the full computational point group symmetry.
                  AO selects for Pitzer-ordering the return vector without point group symmetry.
              subset : str
                  {'ALL', 'ACTIVE', 'FROZEN', 'OCC', 'VIR', 'FROZEN_OCC', 'ACTIVE_OCC', 'ACTIVE_VIR', 'FROZEN_VIR'}
                  Which subspace of orbital energies should be returned?

              Returns
              -------
              Vector
                  A Pitzer-ordered vector of the orbitals.
         )pbdoc")
        .def("epsilon_b_subset", &Wavefunction::epsilon_b_subset, R"pbdoc(
              Returns the requested Beta orbital energies subset.

              Parameters
              ----------
              basis : str
                  {'AO', 'SO', 'MO'}
                  MO or SO select for Pitzer-ordering the return vector in the full computational point group symmetry.
                  AO selects for Pitzer-ordering the return vector without point group symmetry.
              subset : str
                  {'ALL', 'ACTIVE', 'FROZEN', 'OCC', 'VIR', 'FROZEN_OCC', 'ACTIVE_OCC', 'ACTIVE_VIR', 'FROZEN_VIR'}
                  Which subspace of orbital energies should be returned?

              Returns
```
appease the docstring formatter :-)

loriab (approved):

robertodr (approved):

PeterKraus:
So on the topic of type "annotations" (although they're more of a "declaration" in C++): python should be annotated in function headers and not in docstrings as per @loriab's comment [here](https://github.com/psi4/psi4/pull/2298#discussion_r714910415) , but how about a case like this, in C++? Does `sphinx` pick those up from the [header](https://github.com/psi4/psi4/blob/889d03b0504e8159fc10b9469044216cb53a1f63/psi4/src/psi4/libmints/wavefunction.cc#L1226), or are they still necessary in the docstring?

loriab:
> So on the topic of type "annotations" (although they're more of a "declaration" in C++): python should be annotated in function headers and not in docstrings as per @loriab's comment here , but how about a case like this, in C++? Does sphinx pick those up from the header, or are they still necessary in the docstring?

Sphinx gets nothing from the header directly, so any doxygen markup is lost (except we do build and serve doxygen docs that I doubt anyone looks at). But Sphinx does get the types for the exported fns automatically because pybind knows them from the header files. So for the most part you can leave them out of the export docstrings, though enums are still helpful. https://psicode.org/psi4manual/master/api/psi4.core.Wavefunction.html#psi4.core.Wavefunction.alpha_orbital_space

PeterKraus:
Right, then I suppose the `: str`'s in this PR should not be there?

loriab:
> Right, then I suppose the : str's in this PR should not be there?

C-side `: str` renders ok (https://psicode.org/psi4manual/master/api/psi4.core.Wavefunction.html#psi4.core.Wavefunction.Ca_subset) and doesn't cause a sphinx warning. Py-side the `: str` in the docstring actually causes a warning. So best practice is to add the named arguments to the export fn `"basis"_a` so that `basis: str` shows up in the rendered signature. If one doesn't do the `_a`, the signature shows up as `arg1: str`, so there's still some merit to `basis : str` in the docstring.

JonathonMisiewicz:
I assume that `C-side` here means `pybind-side`?

In a case where I _do_ name the function arguments when I pybind, is there still merit in `basis: str`? How should I signify that some arguments take only specified strings, or what the significance of a particular keyword is without `basis: str`? Something like this?

```
              basis : {'AO', 'SO', 'MO'}
                  MO or SO select for Pitzer-ordering the return vector in the full computational point group symmetry.
                  AO selects for Pitzer-ordering the return vector without point group symmetry.
```

loriab:
> I assume that C-side here means pybind-side?

Yes, sphinx knows no deeper into C-side than pybind-side.

> In a case where I do name the function arguments when I pybind, is there still merit in basis: str? 

The only merit is that it'll provide a link to the python documentation for `string` :-)

> How should I signify that some arguments take only specified strings, or what the significance of a particular keyword is without basis: str?

however `alpha_orbital_space` does it renders nicely https://psicode.org/psi4manual/master/api/psi4.core.CCWavefunction.html?highlight=orbital_space#psi4.core.CCWavefunction.alpha_orbital_space

PeterKraus (approved):

PR #2298, opened 2021-09-23
Move SCF-DIIS to Python

JonathonMisiewicz:
## Description
In preparation for some PRs adding EDIIS and ADIIS to Psi, this PR moves the `compute_orbital_gradient` functions, which call DIIS, to the Python layer.

There are multiple things going on:
- For pybinding, it was convenient to convert the enums in `libdiis` to _strongly typed enums_. This means the enums have their own scope, which changes how other functions had to call them. Every change in occ, dfocc, dct, fisapt, and libdiis is _solely_ due to this. This occupies commit one.
- Many other functions had to be pybound. In particular, it was necessary to convert `diis_manager` into a property. Old code would call `HF.diis_manager()`, but now they call `HF.diis_manager_`. We're not bothering to deprecate the old way of calling it. I'm not aware if there's a way to still support the old syntax. It might work if we change the property name to `diis_manager`? I haven't tested this.
- It is not possible to pybind a variadic, so I had to pybind every `set_error_vector_size` and `set_vector_size` and `add_entry` type combination that Python might use. Sad, but necessary.

And with that done, all four of the `compute_orbital_gradient` functions could finally be moved to the Python layer. `compute_orbital_gradient` is kept as a virtual function, C++ side, to signal that Psi expects such a function to exist. My tests indicate that if the function is defined C++ side, it still works, so this won't break any SCF subclasses defined in plugins.

## Todos

- [x] _Lots_ of pybind-ing involving DIIS
- [x] Alert! `HF.diis_manager()` has changed to `HF.diis_manager_`
- [x] `HF.compute_orbital_gradient` moved to the Python layer. C-side `compute_orbital_gradient` is still supported, but not used by the core Psi.

## Questions
- [ ] Who is responsible for updating the Great DFOCC Branch with the `libdiis` API change?
- [ ] How do we feel about the new functions I Pybind-ed over? Anything amiss?

## Checklist
- [x] `ctest -L scf` passes

## Status
- [x] Ready for review
- [ ] Let's discuss the two questions before merging this in, please.

### DIFF
--- a/psi4/driver/procrouting/scf_proc/subclass_methods.py
+++ b/psi4/driver/procrouting/scf_proc/subclass_methods.py
@@ -1,0 +1,88 @@
+import math
+
+from psi4 import core
+
+# Aliases for less typing. Sadly, can't `from psi4.core.X import Y` them
+StoragePolicy = core.DIISManager.StoragePolicy
+RemovalPolicy = core.DIISManager.RemovalPolicy
+
+def _RHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
+    gradient = self.form_FDSmSDF(self.Fa(), self.Da())
+
+    if save_fock:
+        if not self.initialized_diis_manager_:
+            storage_policy = StoragePolicy.InCore if self.scf_type() == "DIRECT" else StoragePolicy.OnDisk
+            self.diis_manager_ = core.DIISManager(max_diis_vectors, "HF DIIS vector", RemovalPolicy.LargestError, storage_policy)
+            self.diis_manager_.set_error_vector_size(gradient)
+            self.diis_manager_.set_vector_size(self.Fa())
+            self.initialized_diis_manager_ = True
+        
+        self.diis_manager_.add_entry(gradient, self.Fa())
+
+    if self.options().get_bool("DIIS_RMS_ERROR"):
+        return gradient.rms()
+    else:
+        return gradient.absmax()
+
+def _UHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
+    gradient_a = self.form_FDSmSDF(self.Fa(), self.Da());
+    gradient_b = self.form_FDSmSDF(self.Fb(), self.Db());
+
+    if save_fock:
+        if not self.initialized_diis_manager_:
+            self.diis_manager_ = core.DIISManager(max_diis_vectors, "HF DIIS vector", RemovalPolicy.LargestError,
+                                                          StoragePolicy.OnDisk)
+            self.diis_manager_.set_error_vector_size(gradient_a, gradient_b)
+            self.diis_manager_.set_vector_size(self.Fa(), self.Fb())
+            self.initialized_diis_manager_ = True
+
+        self.diis_manager_.add_entry(gradient_a, gradient_b, self.Fa(), self.Fb())
+
+    if self.options().get_bool("DIIS_RMS_ERROR"):
+        return math.sqrt(0.5 * (gradient_a.rms() ** 2 + gradient_b.rms() ** 2))
+    else:
+        return max(gradient_a.absmax(), gradient_b.absmax())
+
+def _ROHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
+    # Only the inact-act, inact-vir, and act-vir rotations are non-redundant
+    dim_zero = core.Dimension(self.nirrep(), "Zero Dim")
+    noccpi = self.doccpi() + self.soccpi()
+    row_slice = core.Slice(dim_zero, noccpi)
+    col_slice = core.Slice(self.doccpi(), self.nmopi())
+    MOgradient = self.moFeff().get_block(row_slice, col_slice)
+
+    # Zero the active-active block
+    for h in range(MOgradient.nirrep()):
+        socc = self.soccpi()[h]
+        docc = self.doccpi()[h]
+
+        MOgradient.nph[h][docc:docc+socc, 0:socc] = 0
+
+    # Grab inact-act and act-vir orbs
+    # Ct is (nmo x nmo), not the (nso x nmo) you would expect
+    row_slice = core.Slice(dim_zero, self.nmopi())
+    col_slice = core.Slice(dim_zero, noccpi)
+    Cia = self.Ct().get_block(row_slice, col_slice)
+    col_slice = core.Slice(self.doccpi(), self.nmopi())
+    Cav = self.Ct().get_block(row_slice, col_slice)
+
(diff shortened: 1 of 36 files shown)


psi4/driver/procrouting/scf_proc/subclass_methods.py
```diff
@@ -0,0 +1,85 @@
+import math
+
+from psi4 import core
+
+def _RHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int):
```
PhillCli, line 5: Can you add return annotations as well?

```python
def _RHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
```

JonathonMisiewicz: Sure, I'll include that in the next set of commits.

Normally when somebody who isn't a core developer leaves comments on a PR, the PR is directly relevant to their research. Can I ask if there's some special reason for you commenting on this one?

psi4/driver/procrouting/scf_proc/subclass_methods.py
```diff
@@ -0,0 +1,85 @@
+import math
+
+from psi4 import core
+
+def _RHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
+    gradient = self.form_FDSmSDF(self.Fa(), self.Da())
+
+    if save_fock:
+        if not self.initialized_diis_manager_:
+            storage_policy = getattr(core.DIISManager.StoragePolicy, "InCore" if self.scf_type() == "DIRECT" else "OnDisk")
```
loriab, line 10:
```suggestion
            if self.scf_type() == "DIRECT":
                storage_policy = core.DIISManager.StoragePolicy.InCore 
            else:
                storage_policy = core.DIISManager.StoragePolicy.OnDisk
```
equivalent? I usually save `getattr` for when attribute only known in a variable. But that's my readability opinion. Your choice.

psi4/driver/procrouting/scf_proc/subclass_methods.py
```diff
@@ -0,0 +1,85 @@
+import math
+
+from psi4 import core
+
+def _RHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
```
loriab, line 5: general info to readers: the Sphinx docs can now process type annotations, so they're not added in vain. the Sphinx docs now also hate the old Parameters\n verbose : int style in-docstring type annotation, so those're worth moving up into the signature.

loriab (approved):
lgtm!

psi4/driver/procrouting/scf_proc/subclass_methods.py
JonathonMisiewicz: Yes, equivalent. The right thing to do for readability is probably to import `from core.DIISManager import StoragePolicy, RemovalPolicy`. Then I can have a terse ternary without using `getattr`.

loriab: that works. only other consideration is if you'd ever want to track which paths were tested with codecov.

psi4/driver/procrouting/scf_proc/subclass_methods.py
PhillCli: @JonathonMisiewicz not particularly. I'm using some of the python SCF API lately, glad to see more logic available at the python level. In general type hinting FTW!

PeterKraus: @loriab Someone ought to note that in the "developers howto", if it's not there already.

JonathonMisiewicz:
On hold until #2296 gets approved, to see if we can reach a consensus on pybind docstring format is now.

JonathonMisiewicz:
Force pushed to bring this branch up-to-date. Should be good to review again.

jturney (approved):
Looks good!

psi4/driver/procrouting/scf_proc/subclass_methods.py
```diff
@@ -0,0 +1,89 @@
...
+    else:
+        return max(gradient_a.absmax(), gradient_b.absmax())
+
+def _ROHF_orbital_gradient(self, save_fock: bool, max_diis_vectors: int) -> float:
+    # Only the inact-act, inact-vir, and act-vir rotations are non-redundant
+    dim_zero = core.Dimension(self.nirrep(), "Zero Dim")
+    noccpi = self.doccpi() + self.soccpi()
+    virpi = self.nmopi() - self.doccpi()
+    row_slice = core.Slice(dim_zero, noccpi)
+    col_slice = core.Slice(self.doccpi(), self.doccpi() + virpi)
```
PeterKraus, line 52: `self.doccpi() + virpi = self.nmopi()`

psi4/driver/procrouting/scf_proc/subclass_methods.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # Zero the active-active block
+    for h in range(MOgradient.nirrep()):
+        socc = self.soccpi()[h]
+        docc = self.doccpi()[h]
+
+        MOgradient.nph[h][docc:docc+socc, 0:socc] = 0
+
+    # Grab inact-act and act-vir orbs
+    # Ct_ is (nmo x nmo), not the (nso x nmo) you would expect
```
PeterKraus, line 63: `self.Ct()`, not `Ct_`

psi4/driver/procrouting/scf_proc/subclass_methods.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.diis_manager_.add_entry(gradient, self.soFeff())
+
+    if self.options().get_bool("DIIS_RMS_ERROR"):
+        return gradient.rms()
+    else:
+        return gradient.absmax()
+
+core.RHF.compute_orbital_gradient = _RHF_orbital_gradient
+core.UHF.compute_orbital_gradient = core.CUHF.compute_orbital_gradient = _UHF_orbital_gradient
```
PeterKraus, line 88: I'm not a terribly big fan of this triple-assignment.

PeterKraus (approved):
I only have a couple very minor style suggestions. I suggest @JonathonMisiewicz can decide whether to action them or merge straight away.

psi4/driver/procrouting/scf_proc/subclass_methods.py
JonathonMisiewicz: It looked strange to me at first as well, but I've made my peace with it. Putting these assignments on the same line makes it obvious that these two are _intended to be_ the same function. I was a little surprised that CUHF didn't need custom logic here.

PR #2302, opened 2021-09-25
Fix #2120

### LINKED ISSUE #1545: Asinine energy with DF-FNO-CCSD(T) energy
@hokru told me to post this here from this thread on the forum. 
http://forum.psicode.org/t/weird-energy-with-df-ccsd-t-on-pre-built-1-2-binary/1240

I used the pre-built binaries for 1.2. It's the python 3.5 linux version.  

I am getting an asinine energy with this method for sumanene. Here's a few of the iterations. 

0   0 1 82301976411.2597808838 82301976411.2597808838 180071.6863671333       79
  1   1 1 -11765764147947413438464.0000000000 -11765764148029716168704.0000000000 5189755552069629.0000000000     1039
  2   2 1 18577504081805034984135645462528.0000000000 18577504093570799348831936839680.0000000000 13953090784875981570048.0000000000     1042
  3   3 1 -10975754574997206597632.0000000000 -18577504092780789161197672857600.0000000000 2850375421929338594918400.0000000000     1058

### LINKED ISSUE #2120: nmopi() and Ca()->coldim() not consistent when linear dependencies present
DF-CC freaks out when linear dependencies are present, and i tracked the issue to an inconsistency between reference_wavefunction_->nmopi() and reference_wavefunction_->Ca()->coldim(). The latter has the correct value when linear dependencies are removed. 

Should nmopi() and Ca()->coldim() be consistent in this case?

Minimal input:
```
molecule { 
o 
h 1 1.0
h 1 1.0 2 104.5
}

set {
  basis 'aug-cc-pvtz'
  scf_type df
  cc_type df
  s_tolerance 1e-3
}
energy('ccsd')
```

JonathonMisiewicz:
## Description
This PR closes #2120. The user-side problem is that when there are linear dependencies in the molecule, `fnocc` reports uncontrolled energies, iteration-by-iteration. The mechanism for this is as follows:

* `run_fnocc` requests an SCF computation. `nmopi_ < nsopi_` due to linear dependencies.
* `fnocc` is incompatible with point group symmetry, so calls `c1_deep_copy` to transform the solution to a c1 solution. For concreteness, we'll assume RHF, although all HF subclasses have the same problem.
* `RHF::c1_deep_copy` calls `Wavefunction::c1_deep_copy`. The resulting wavefunction has `nmo_` equal to `nmopi_.sum()` of the previous wavefunction, as desired.
* `RHF::c1_deep_copy` calls the RHF constructor on the new wavefunction, which calls to `HF::common_init`, which initializes `nmopi_ = nsopi_`, under assumption that the subsequent SCF will call `form_Shalf`, which is _actually_ responsible for first eliminating linear dependencies and then determining `nmopi_`.
* There is no subsequent SCF, so `nmopi_ = nsopi_` at the time `fnocc` begins.
* Garbage In, Garbage Out. `fnocc` is given a garbage wavefunction and gives garbage energies.

The remedy here is to no longer have `common_init` put suspect information on the wavefunction.

Obligatory ping to @edeprince3.

## Todos

- [x] `c1_deep_copy` no longer leads to a wrong `nmopi_` in case of linear dependency. This fixes problems with linear dependencies observed in the `fnocc` module.

## Questions
- [ ] Will this fix #1545? Holger confirmed the bug there was symmetry related, and the system does have a linear dependency... It's probably worth re-investigating after this gets merged in.

## Checklist
- [x] Tests added for any newly working features
- [x] `ctest -j4` is 490/490

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -113,14 +113,12 @@ void HF::common_init() {
     H_.reset(factory_->create_matrix("One-electron Hamiltonian"));
     X_.reset(factory_->create_matrix("X"));
 
-    nmo_ = 0;
+    // nmo_ and nmopi_ not determined at present.
     nso_ = 0;
     const Dimension& dimpi = factory_->colspi();
     for (int h = 0; h < factory_->nirrep(); h++) {
         nsopi_[h] = dimpi[h];
-        nmopi_[h] = nsopi_[h];  // For now, may change in S^-1/2
         nso_ += nsopi_[h];
-        nmo_ += nmopi_[h];  // For now, may change in S^-1/2
     }
 
     density_fitted_ = false;
@@ -184,7 +182,7 @@ void HF::common_init() {
 
     // Check that we have enough basis functions
     for (int h = 0; h < nirrep_; ++h) {
-        if (doccpi_[h] + soccpi_[h] > nmopi_[h]) {
+        if (doccpi_[h] + soccpi_[h] > nsopi_[h]) {
             throw PSIEXCEPTION("Not enough basis functions to satisfy requested occupancies");
         }
     }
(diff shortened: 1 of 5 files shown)


robertodr (approved):

loriab (approved):

CDSherrill:
I'm familiar with the deep copy stuff because I wrote some of it.  Everything here seems good to me.  If there were any unexpected consequences, they should have been caught by the test suite.

CDSherrill (approved):

PR #2314, opened 2021-10-04
E(30)exch-ind in SAPT2+3 without the S^2 approximation

konpat:
## Description
Adds an initial implementation of the E(30)exch-ind SAPT correction without the single exchange (S^2) approximation, introduced in https://doi.org/10.1063/5.0035050. Original Psi4NumPy implementation by Jonathan Waldrop (@jwaldrop107), ported to C++ by Konrad Patkowski. 

The resulting numbers for a bunch of tests match the SI of the above paper and the calculation takes a tiny fraction of the total SAPT2+3 time. E(30)exch-ind,resp is approximated by scaling unrelaxed E(30)exch-ind in the same way as before and is also printed.

## Usage
The nonapproximated E(30)exch-ind and E(30)exch-ind,resp numbers will be calculated by `energy('sapt2+3')` if `do_ind30_exch_sinf` is set to `True`. 

## Technical Questions and Unresolved Issues
- [x] There's certainly room for code cleanup.
- [x] I think `libsapt_solver` is the most logical home for this code and `energy('sapt2+3')` its most logical invocation - after all, this correction forms a part of SAPT2+3 only, not the lower SAPT levels. The E(30)ind correction needs to be calculated beforehand, both for subtracting its value from the total (ind+ex-ind) calculated here, and for generating the third-order induction amplitudes that are reused here. However, unlike most of the stuff in `libsapt_solver`, the code is AO-based and the two-electron integrals enter only via generalized J/K matrices. Therefore, I had to export some additional AO quantities in `libsapt_solver/sapt.h` and I don't know if I did this in an optimal way.
- [x] Note to self: frozen core still needs to be checked.

## Design (Big-Picture) Questions - input from @CDSherrill @dgasmith @loriab and others will be highly beneficial!
- [ ] The three new(ish) beyond-S^2 corrections, E(20)exch-ind, E(20)exch-disp, and E(30)exch-ind cannot be currently calculated by a single SAPT call: the first two require `energy('sapt(dft)')` and the last one `energy('sapt2+3')`. That's quite unfortunate and we need a better workflow, possibly by making the non-S^2 E(20)exch-ind,resp and E(20)exch-disp available as part of `energy('sapt0')`.
- [ ] What should go into aggregate SAPT energy values, S^2 or non-S^2? I made the code pick the non-S^2 value of E(30)exch-ind,resp if available, but I'd be happy to revert it back to picking E(30)exch-ind,resp(S^2). At present, I don't think the non-S^2 values of E(20)exch-ind,resp and E(20)exch-disp are ever included in total SAPT numbers, maybe they should?
- [ ] Finally, the option names such as `do_ind30_exch_sinf` are somewhat unfortunate because they propagate the misconception that you include effects all the way to S^infinity. In reality, the largest power of overlap integral S you ever need is min(NelA,NelB) - after that, you run out of electrons to exchange between A and B. **Nonapproximated**, **Nonexpanded**, and **beyond-S^2** are all formally more correct names than S^infinity.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    Cr.push_back(preXiBB);
+    jk_->compute();
+
+    std::shared_ptr<Matrix> J_XiBB = J[0];
+    std::shared_ptr<Matrix> K_XiBB = K[0]->transpose();
+
+    CompleteInd30 += 4.0 * C_DDOT(nn * nn, XiAA->pointer()[0], 1, J_XiBB->pointer()[0], 1);
+    CompleteInd30 -= 2.0 * C_DDOT(nn * nn, XiAA->pointer()[0], 1, K_XiBB->pointer()[0], 1);
+    CompleteInd30 += 4.0 * C_DDOT(nn * nn, XiAB->pointer()[0], 1, J_XiBB->pointer()[0], 1);
+    CompleteInd30 -= 2.0 * C_DDOT(nn * nn, XiAB->pointer()[0], 1, K_XiBB->pointer()[0], 1);
```
JonathonMisiewicz, line 611: Please use `Matrix::vector_dot` instead of manual BLAS calls, for readability.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    std::shared_ptr<Matrix> STS_as = linalg::triplet(sAR, Tbr, sBS, false, true, false);
+    std::shared_ptr<Matrix> STS_ar = linalg::triplet(sAR, Tar, sAR, false, true, false);
+    std::shared_ptr<Matrix> STS_bs = linalg::triplet(sBS, Tbs, sBS, false, true, false);
+
+    CompleteInd30 += C_DDOT(na * nr, uAR[0], 1, omega_ar->pointer()[0], 1);
+    CompleteInd30 += C_DDOT(nb * ns, uBS[0], 1, omega_bs->pointer()[0], 1);
+    CompleteInd30 -= C_DDOT(nb * nr, STS_br->pointer()[0], 1, omega_br->pointer()[0], 1);
+    CompleteInd30 -= C_DDOT(na * ns, STS_as->pointer()[0], 1, omega_as->pointer()[0], 1);
+    CompleteInd30 -= C_DDOT(na * nr, STS_ar->pointer()[0], 1, omega_ar->pointer()[0], 1);
+    CompleteInd30 -= C_DDOT(nb * ns, STS_bs->pointer()[0], 1, omega_bs->pointer()[0], 1);
```
JonathonMisiewicz, line 560: Please use `Matrix::vector_dot` instead of manual BLAS calls, for readability.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    // => New Stuff <= //
+    // Start with T's
+    std::shared_ptr<Matrix> Sbr = linalg::triplet(CoccB_, Smat_, CvirA_, true, false, false);
+    std::shared_ptr<Matrix> Sas = linalg::triplet(CoccA_, Smat_, CvirB_, true, false, false);
+    auto Tar = std::make_shared<Matrix>("Tar", na, nr);
+    auto Tbr = std::make_shared<Matrix>("Tbr", nb, nr);
+    auto Tas = std::make_shared<Matrix>("Tas", na, ns);
+    auto Tbs = std::make_shared<Matrix>("Tbs", nb, ns);
+
+    C_DGEMM('N', 'N', na, nr, nb, 1.0, &Dp[0][na], na + nb, Sbr->pointer()[0], nr, 0.0,
```
JonathonMisiewicz, line 375: The `libmints` function gets block allows you to get a block of a Matrix. You _could_ use this to get the four blocks of the D^-1 matrix as their own matrices, and then 8 of the "New stuff" DGEMM calls become simple doublets. I expect your half-transforms and your xi-dependent terms simplify as well. There will be some additional copy overhead from this approach, but the code becomes cleaner. I won't insist on you doing this, but I do recommend you consider it.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    Ct_Kr->scale(-1);
+    Ct_Kr->add(CvirA_);
+    auto Ct_Ks = linalg::triplet(D_ia, Smat_, CvirB_, false, false, false);
+    Ct_Ks->scale(-1);
+    Ct_Ks->add(CvirB_);
+
+    // Make omega cores
+    std::shared_ptr<Matrix> AJK(J_D_ia->clone());
+    AJK->zero();
+    AJK->add(J_D_ia);
```
JonathonMisiewicz, line 462: Recurring issue: I don't understand why you clone a matrix, zero it, and then add the matrix you just cloned. Just clone it.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    CompleteInd30 -= C_DDOT(nb * ns, STS_bs->pointer()[0], 1, omega_bs->pointer()[0], 1);
+
+//All Omega-dependent contributions have been completed, now Xi-dependent (J/K) contributions
+
+    auto SCt_Na = std::make_shared<Matrix>("SCt_Na", nn, na);
+    auto SCt_Nb = std::make_shared<Matrix>("SCt_Nb", nn, nb);
+    C_DGEMM('N', 'T', nn, na, nr, 1.0, Ct_Kr->pointer()[0], nr, sAR->pointer()[0], nr, 0.0,
+            SCt_Na->pointer()[0], na);
+    C_DGEMM('N', 'T', nn, nb, ns, 1.0, Ct_Ks->pointer()[0], ns, sBS->pointer()[0], ns, 0.0,
+            SCt_Nb->pointer()[0], nb);
```
JonathonMisiewicz, line 569: I believe you can just hit these with doublet?

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    std::shared_ptr<Matrix> K_XiBB = K[0]->transpose();
+
+    CompleteInd30 += 4.0 * C_DDOT(nn * nn, XiAA->pointer()[0], 1, J_XiBB->pointer()[0], 1);
+    CompleteInd30 -= 2.0 * C_DDOT(nn * nn, XiAA->pointer()[0], 1, K_XiBB->pointer()[0], 1);
+    CompleteInd30 += 4.0 * C_DDOT(nn * nn, XiAB->pointer()[0], 1, J_XiBB->pointer()[0], 1);
+    CompleteInd30 -= 2.0 * C_DDOT(nn * nn, XiAB->pointer()[0], 1, K_XiBB->pointer()[0], 1);
+
+    e_exch_ind30_sinf_ = CompleteInd30 - e_ind30_;
+
+    Process::environment.globals["SAPT EXCH-IND30(S^INF) ENERGY"] = e_exch_ind30_sinf_;
```
JonathonMisiewicz, line 615: Q for Lori: Do we want to define the global? I know we're shifting to wavefunction variables.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    for (int b = 0; b < nb; b++) {
+        for (int s = 0; s < ns; s++) {
+            sBSp[b][s] = wABS_[b][s] / (evalsB_[b] - evalsB_[nb+s]);
+        }
+    }
+
+    std::shared_ptr<Matrix> STS_br = linalg::triplet(sBS, Tas, sAR, false, true, false);
+    std::shared_ptr<Matrix> STS_as = linalg::triplet(sAR, Tbr, sBS, false, true, false);
+    std::shared_ptr<Matrix> STS_ar = linalg::triplet(sAR, Tar, sAR, false, true, false);
+    std::shared_ptr<Matrix> STS_bs = linalg::triplet(sBS, Tbs, sBS, false, true, false);
```
JonathonMisiewicz, line 553: These are good candidates for `auto`.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +319,305 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    // Make JK's
+    jk_ = JK::build_JK(basisset_, get_basisset("DF_BASIS_SCF"), options_, false, mem_);
+    jk_->set_memory(mem_);
+    jk_->set_do_J(true);
+    jk_->set_do_K(true);
+    jk_->initialize();
+    jk_->print_header();
+
+    std::vector<SharedMatrix>& Cl = jk_->C_left();
+    std::vector<SharedMatrix>& Cr = jk_->C_right();
```
JonathonMisiewicz, line 430: `auto&` candidates. `const auto&` below.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -31,6 +31,11 @@
 #include "psi4/libpsio/psio.hpp"
 #include "psi4/libqt/qt.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
+#include "psi4/lib3index/dfhelper.h"
```
JonathonMisiewicz, line 34: Why is this header needed?

JonathonMisiewicz (commented):
Thanks for the feature. I'm not a SAPT expert, so I'll leave your big picture questions to Lori.

Overarching comment: New features always need documentation and tests. This currently has none. Do you need any guidance on what that looks like?

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: It wasn't. I copied and pasted too much.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Changed.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Changed.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Yes, I can, and now I do.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: My bad. Fixed throughout.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: I see your point, it's just that I'm quite confused about the syntax of the slice definition for `matrix->get_block`.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Done!

psi4/src/psi4/libsapt_solver/exch-ind30.cc
loriab: It's a lot simpler to switch over from P::e.globals to wfn module-by-module (or at least method-by-method). Looks like over libsapt_solver are still globals, so this new one is good as global.

konpat:
Thanks much for the review and helpful corrections!

Yes, I am fully aware that adding tests and documentation is necessary, I think I know how to go about doing this, and I will certainly add these before the PR is merge-ready. 

It would be really nice to have a test that checks all three beyond-S^2 corrections: E(20)exch-ind, E(20)exch-disp, and E(30)exch-ind. It would require two SAPT calculations at the moment though.

konpat:
Yes, @loriab, I'm looking forward to discussing the naming of SAPT variants/calls, and Psicon will be a good opportunity.

I very much like the name `do_ind30_exch_full`.

JonathonMisiewicz:
I'm marking this PR as blocked until PsiCon, pending Auburn/GATech discussions.

konpat:
This PR is ready to be unblocked and reviewed. A test has been added. Frozen-core works as expected, that is, the number is unchanged (just like regular E(30)exch-ind). The numbers agree with the original paper. @jwaldrop107 - you might want to take a look and check if everything's still working in your hands.

We had a very productive discussion with @CDSherrill @loriab @zachglick last week. We have some ideas how to make the SAPT workflows, including the stuff without the S^2 approximation, simpler and more consistent. We are also thinking of renaming all **Sinf** to **Sall**, but that's a story for a separate PR.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,264 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    }
+
+    // => Sizing <= //
+
+    int nn = basisset_->nbf();
+    int na = noccA_;
+    int nb = noccB_;
+    int nr = nvirA_;
+    int ns = nvirB_;
+    int nQ = ribasis_->nbf();
```
JonathonMisiewicz, line 338: Unused variable.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,264 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+
+    int nT = 1;
+#ifdef _OPENMP
+    nT = Process::environment.get_n_threads();
+#endif
+
+    auto uAR = std::make_shared<Matrix>("Ind30 uAR Amplitudes", na, nr);
+    auto uBS = std::make_shared<Matrix>("Ind30 uBS Amplitudes", nb, ns);
```
JonathonMisiewicz, line 346: I'd prefer `Matrix` over `shared_ptr<Matrix>`, unless there's a reason we need the pointer.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,264 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    uAR->load(psio_, PSIF_SAPT_AMPS, Matrix::SaveType::Full);
+    uBS->load(psio_, PSIF_SAPT_AMPS, Matrix::SaveType::Full);
+
+    // => Intermolecular overlap matrix and inverse <= //
+    std::shared_ptr<Matrix> Sab = linalg::triplet(CoccA_, Smat_, CoccB_, true, false, false);
```
JonathonMisiewicz, line 351:
```suggestion
    auto Sab = linalg::triplet(CoccA_, Smat_, CoccB_, true, false, false);
```

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,264 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+
+    double** Sabp = Sab->pointer();
+    auto D = std::make_shared<Matrix>("D", na + nb, na + nb);
```
JonathonMisiewicz, line 354: This can also just be a regular `Matrix`.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,264 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    nadim[0] = na;
+    nabdim[0] = na + nb;
+    auto Daa = D->get_block(Slice(zero, nadim), Slice(zero, nadim));
+    auto Dab = D->get_block(Slice(zero, nadim), Slice(nadim, nabdim));
+    auto Dbb = D->get_block(Slice(nadim, nabdim), Slice(nadim, nabdim));
+
+    // => New Stuff <= //
+    // Start with T's
+    std::shared_ptr<Matrix> Sbr = linalg::triplet(CoccB_, Smat_, CvirA_, true, false, false);
+    std::shared_ptr<Matrix> Sas = linalg::triplet(CoccA_, Smat_, CvirB_, true, false, false);
```
JonathonMisiewicz, line 377:
```suggestion
    auto Sbr = linalg::triplet(CoccB_, Smat_, CvirA_, true, false, false);
    auto Sas = linalg::triplet(CoccA_, Smat_, CvirB_, true, false, false);
```

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,264 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+
+    Cl.clear();
+    Cr.clear();
+    Cl.push_back(Cocc0AB);
+    Cr.push_back(D_Ni_a);
+    Cl.push_back(Cocc0AB);
+    Cr.push_back(D_Ni_b);
+    jk_->compute();
+
+    std::shared_ptr<Matrix> J_D_ia = J[0];
```
JonathonMisiewicz, line 432: This is another case to use `auto`.

JonathonMisiewicz (commented):
I have some very nitpicky suggestions, but this code looks pretty good to me.

Do you think the documentation needs to be updated?

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Done.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Removed.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: Then I need to reveal my ignorance and ask a basic question: how to do this so that `uAR->load()` and `uAR->vector_dot()` still work?

JonathonMisiewicz: Just change `->` to `.`.

`->` is the syntax for "this object is not a Matrix but a memory address of a Matrix, so find the Matrix at this memory address and then call this Matrix method". `.` is the syntax for "this object is a Matrix, so call this Matrix method".

konpat: OK, done, thanks!

JonathonMisiewicz (approved):
The code looks good to me - thanks for the new feature!

I'll leave it to the SAPT experts to decide if this needs documentation on the main SAPT page beyond whatever gets auto-added when a new option is introduced.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,263 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+
+    // => Sizing <= //
+
+    int nn = basisset_->nbf();
+    int na = noccA_;
+    int nb = noccB_;
+    int nr = nvirA_;
+    int ns = nvirB_;
+
+    int nT = 1;
```
jeffschriber, line 339: I don't think this threading variable is needed

jeffschriber (commented):
This is great! Code looks good to me, and I think the addition to the manual is good and necessary.

psi4/src/psi4/libsapt_solver/exch-ind30.cc
konpat: It's not. I removed it.

jeffschriber (approved):

konpat:
From my perspective, everything is in good shape now, including the documentation.

loriab (approved):
lgtm, thank you!

The issue of totaling qcvars is handled in python and so isn't covered here. But that can be addressed together with some of the other changes from the Dec sapt mini conf in a future PR.

Will merge this week unless any other comments come up!

psi4/src/psi4/libsapt_solver/exch-ind30.cc
```diff
@@ -314,5 +318,258 @@ double SAPT2p3::exch_ind30_3(double **sBS) {
...
+    double** Sabp = Sab->pointer();
+    auto D = Matrix("D", na + nb, na + nb);
+    D.identity();
+    double** Dp = D.pointer();
+    for (int a = 0; a < na; a++) {
+        for (int b = 0; b < nb; b++) {
+            Dp[a][b + na] = Dp[b + na][a] = Sabp[a][b];
+        }
+    }
+    D.power(-1.0, 1.0E-12);
```
loriab, line 356: There's also a `Matrix.general_inverse` that may be of interest: https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/matrix.h#L959-L992

PR #2319, opened 2021-10-08
Move SCF Class Variables to Function Variables

JonathonMisiewicz:
## Description
This is another minor cleanup PR that I saw was possible while working on a larger project. Three `temp` variables were previously saved to the wavefunction but only needed to be called by one or two functions. Those are better suited to be function variables than class variables, and this PR makes them so.

## Questions
- [x] Fewer class variables means less memory, but re-allocating these matrices every iteration could cause a timing hit. Is this a problem? ~~I expect the time cost to be negligible in comparison with the cost of the BLAS operations, but if anybody believes otherwise, better to say so now.~~ **Negligible in comparison to diagonalization time.**

## Checklist
- [x] `ctest -L ^scf` passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/rohf.cc
+++ b/psi4/src/psi4/libscf_solver/rohf.cc
@@ -366,7 +366,7 @@ void ROHF::form_F() {
     }
 
     // Form the orthogonalized SO basis moFeff matrix, for use in DIIS
-    soFeff_ = linalg::triplet(Ct_, moFeff_, Ct_, false, false, true);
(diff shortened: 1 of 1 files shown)


JonathonMisiewicz:
Timings test indicate that re-allocation cost is negligible. All these quantities are `n^2`, and a 5000 x 5000 matrix takes about 0.11 seconds to allocate on my laptop. That's nothing in comparison to the diagonalization time. I no longer see a barrier to merging this in.

PR #2328, opened 2021-10-19
Add saving for T-amplitudes to the QCSchema wrapper.

cgbriggs99:
## Description
The QCSchema wrapper was lacking the capability to save the t-amplitudes and the one-electron density matrix, so this was added.

## Todos
This will add the ability to save the t-amplitudes using the new "tamps" driver.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
Thanks for the suggestion. This is going to have trouble as-is because `driver` is under schema control https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/common_models.py#L57-L70 . Do you have this working locally? Could you be running `psi4 --json` rather than `psi4 --qcschema`? The former is an old but still supported route that undergoes fewer checks, iirc.

cgbriggs99:
I've made the changes requested and added a test.

loriab:
Oh dear, it looks like this PR has been contaminated by non-you commits. See if you can follow the healing advice here https://github.com/psi4/psi4/pull/1490#issuecomment-457765390 . Ping if there are any problems. Thanks!

cgbriggs99:
I think that worked. Thanks!

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,64 @@
...
+h2o_test = psi4.geometry("""
+0 1
+O 0.0 0.0 0.1294769411935893
+H 0.0 -1.494187339479985 1.0274465079245698
+H 0.0 1.494187339479985 1.0274465079245698
+""")
+e, wfn = psi4.energy("mp2/6-31g", return_wfn = True)
+tIJAB = wfn.get_amplitudes("tIJAB").to_array
+
+with open("output.json", "w") as ofile:
```
bgpeyton, line 60: I don't think you need to save the return dict to file for the test

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,64 @@
...
+    json.dump(json_ret, ofile, indent=2)
+
+psi4.compare_values(tIJAB, json_ret["extras"]["qcvars"]["tIJAB"])
```
bgpeyton, line 63: could you add a `compare_values` for the OPDM (`Da`) here as well?

bgpeyton (changes requested):

psi4/driver/schema_wrapper.py
```diff
@@ -586,6 +586,16 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
             )):
                 json_data["extras"]["qcvars"][k] = _serial_translation(v, json=json_serialization)
 
+    # Add in handling of matrix arguments which need to be obtained by a
+    # a function call.
+    if "tIJAB" in json_data["extras"]["qcvars"]:
+        json_data["extras"]["qcvars"]["tIJAB"] = wfn.get_amplitudes("tIJAB").to_array()
+    if "tIA" in json_data["extras"]["qcvars"]:
+        json_data["extras"]["qcvars"]["tIA"] = wfn.get_amplitudes("tIA").to_array()
```
loriab, line 594:
```suggestion
    if "tIJAB" in json_data["extras"]["psi4:tamps"]:
        json_data["extras"]["psi4:tamps"]["tIJAB"] = wfn.get_amplitudes("tIJAB").to_array()
    if "tIA" in json_data["extras"]["psi4:tamps"]:
        json_data["extras"]["psi4:tamps"]["tIA"] = wfn.get_amplitudes("tIA").to_array()
```
I'm worried about extending `extras.qcvars` with extra fields because the routines that write/read from there think they have full control to delete it and fill it. Better to carve out your own bit of extras. `"psi4.tamps"` is just a thought.

psi4/driver/schema_wrapper.py
```diff
@@ -586,6 +586,16 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
...
+    if "Da" in json_data["extras"]["qcvars"]:
+        json_data["extras"]["qcvars"]["Da"] = wfn.Da().to_array()
```
loriab, line 596: any way you could use the density from the wavefunction schema instead of resaving it here? or is it wrong basis and cumbersome reordering?

tests/json/schema-1-tamps/input.py
loriab: it's sometimes helpful to do just to check that nothing non-json-able has gotten added to the data structure.

JonathonMisiewicz:
Thanks for getting the git history cleaned up. I'll wait to review this until after you've responded to the existing comments.

psi4/driver/schema_wrapper.py
bgpeyton: are correlated densities properly updated in wavefunction schema? I know the `wfn` density generally only updates it from the scf if there was an `OEProp` or `properties` call, but at least that's expected. just a case of the devil you know vs the devil you don't :)

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,81 @@
...
+json_ret = psi4.json_wrapper.run_json_qcschema(json_data, True, True)
+
+h2o_test = psi4.geometry("""
+0 1
+O 0.0 0.0 0.1294769411935893
+H 0.0 -1.494187339479985 1.0274465079245698
+H 0.0 1.494187339479985 1.0274465079245698
+symmetry c1
+""")
+e, wfn = psi4.energy(f"{method}/{basis}", return_wfn = True)
```
cgbriggs99, line 61: I don't know what's going on here. For some reason, the amplitudes are not being set properly in this call to the energy. The dimensions are right, but all the values are zero. The data from the schema is being set properly and upon inspection, seems to match what is given in the output files. But direct calls are failing.

loriab: fyi, there's a report that amplitutes are/aren't available depending on if basis is set by option or set in the energy call in mtd/basis format. you might try the "set basis" route to check if that's what you're seeing.

JonathonMisiewicz: And if that is the problem, please file an issue about the amplitude availability. I haven't heard of this problem before.

loriab: that amplitude availability issue I think had to do with the mtd/basis routing through cbs to process the method string, and cbs() has clean statements btwn each part, whereas energy() doesn't. I don't remember if it's solved by ddd, but it should at least be uniform there.

JonathonMisiewicz:
_Many_ files in this PR should not have been committed. Please try again.

cgbriggs99:
Fixed

cgbriggs99:
I'm fixing the test file. Right now, I think I'm having issues with units.

cgbriggs99:
I think this is finally ready. All tests seem to be passing.

psi4/driver/schema_wrapper.py
```diff
@@ -586,6 +586,18 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
...
 
+    # Add in handling of matrix arguments which need to be obtained by a
+    # a function call.
+
+    if "psi4:arrays" in json_data["extras"] :
+        if "tIjAb" in json_data["extras"]["psi4:arrays"]:
+            json_data["extras"]["psi4:arrays"]["tIjAb"] = wfn.get_amplitudes()["tIjAb"].to_array().tolist()
+        if "tIA" in json_data["extras"]["psi4:arrays"]:
+            json_data["extras"]["psi4:arrays"]["tIA"] = wfn.get_amplitudes()["tIA"].to_array().tolist()
+        if "Da" in json_data["extras"]["psi4:arrays"]:
```
JonathonMisiewicz, line 597: I'm nervous about this because densities aren't handled sanely across all Psi modules. I'm not happy about density handling even within the `cc` megamodule and am in the slow process of refactoring. Lori, is it possible/advisable to only have this file for specific modules where we've verified the densities?

JonathonMisiewicz (commented):
I'll want @loriab and @bgpeyton feedback before approving this PR.

psi4/driver/schema_wrapper.py
bgpeyton: densities are indeed a mess, thanks a ton for chipping away at that problem. generally correlated densities (especially `cc` and non-DF `mp2`) don't get updated unless you ask for an `OEProp` or response property, and there are probably other nasty specifics elsewhere. however, seeing as how this is in `extras`, I would qualify this as an "expert" option, so I think leaving a clarifying comment here in the harness would be sufficient (unless there's another place we're documenting these entries).

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,91 @@
...
+Da = wfn.Da().to_array()
+
+with open("output.json", "w") as ofile:
+    json.dump(json_ret, ofile, indent=2)
+
+# Make sure the amplitudes compted from h2o_test are actually assigned.
+if all(map(lambda x: x == 0, tIJAB.flatten())) or \
+   all(map(lambda x: x == 0, tIA.flatten())) or \
+   all(map(lambda x: x == 0, Da.flatten())) :
+    print("Error in computing values to compare against. Could not compare.")
```
loriab, line 83: I'd raise an error if this happens. `compare(True, False, "Error in computing values to compare against")` perhaps

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,91 @@
...
+else :
+    print("Comparing T2")
+    psi4.compare_values(tIJAB, np.array(json_ret["extras"]["psi4:arrays"]["tIjAb"]))
```
loriab, line 86:
```suggestion
    psi4.compare_values(tIJAB, np.array(json_ret["extras"]["psi4:arrays"]["tIjAb"]), "comparing T2")
```
why not somthing like this?

psi4/driver/schema_wrapper.py
```diff
@@ -586,6 +586,18 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
             )):
                 json_data["extras"]["qcvars"][k] = _serial_translation(v, json=json_serialization)
 
+    # Add in handling of matrix arguments which need to be obtained by a
+    # a function call.
+
+    if "psi4:arrays" in json_data["extras"] :
```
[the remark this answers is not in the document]
bgpeyton: considering the use cases we have in mind, and the limitations already present on saving amplitudes like requiring `symmetry c1`, I think RHF-only is reasonable. would the best way to do this simply be 
`if json_data["REFERENCE"] == "RHF":`? 

note: may or may not have to check case here, I don't recall off the top of my head if the kwargs get run through `.upper()` before they get here.

loriab, line 592: do you need separate options for each array? I'm not keen on empty arrays in the input schema as a signal to save amps. what about something like below?
```
if json_data["extras"].get("psi4:save_tamps", False):
    json_data["extras"]["psi4:arrays"]["tIjAb"] = wfn.get_amplitudes()["tIjAb"].to_array().tolist()
    json_data["extras"]["psi4:arrays"]["tIA"] = wfn.get_amplitudes()["tIA"].to_array().tolist()
    json_data["extras"]["psi4:arrays"]["Da"] = wfn.Da().to_array().tolist()
```

psi4/driver/schema_wrapper.py
loriab: I think having a list a methods of verified sanity and then only saving if `json_data["model"]["method"]` in that list is cautious and good. Have this "opt-in" for a while.

loriab (commented):
first look at how this all holds together. thanks for the contribution!

psi4/driver/schema_wrapper.py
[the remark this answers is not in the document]
loriab: there's no assurance that user put in any options. here's a handle on restricted https://github.com/psi4/psi4/pull/2328/files/8f82ece41cc8e403bea358ad68a5aef2c625a53d#diff-3abbb4f37a59cc2e6cfbff3f4cb73b312dfd94044a19f38b00bf7ae140db13e7R330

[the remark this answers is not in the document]
bgpeyton: I was afraid of that. if I recall, this is _always_ present in the output schema, but not necessarily in the input I suppose. 

that link seems to just point to this PRs diff. supposed to be something else maybe, or am I missing something?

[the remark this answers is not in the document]
loriab: https://github.com/psi4/psi4/blob/master/psi4/driver/schema_wrapper.py#L330

JonathonMisiewicz (approved):

psi4/driver/schema_wrapper.py
```diff
@@ -586,6 +586,18 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
...
 
+    # Add in handling of matrix arguments which need to be obtained by a
+    # a function call.
+
+    if json_data["extras"].get("psi4:save_tamps", False):
+        if type(wfn.reference_wavefunction()) is psi4.core.RHF :
+            json_data["extras"]["psi4:arrays"] = {}
+            json_data["extras"]["psi4:arrays"]["tIjAb"] = wfn.get_amplitudes()["tIjAb"].to_array().tolist()
+            json_data["extras"]["psi4:arrays"]["tIA"] = wfn.get_amplitudes()["tIA"].to_array().tolist()
+            if json_data["model"]["method"].lower() in ["ccsd", "scf"] :
```
loriab, line 597: do you perhaps want the method filter right after the RHF filter? and won't the get_amplitudes calls have failed by this point for method=scf? maybe `"psi4:tamps"` instead of `"psi4:arrays"`? wfn.arrays() returns things gradient- and dipole-sized.

cgbriggs99: Fixed

cgbriggs99:
I fixed those few lines. Is there anything else that I should do?

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,84 @@
...
+H 0.0 1.494187339480 1.027446507906
+symmetry c1
+""")
+
+h2o_schema = psi4.core.Molecule.from_schema(json_data)
+
+psi4.compare_values(h2o_test.geometry().to_array(),
+                    h2o_schema.geometry().to_array())
+
+json_ret = psi4.json_wrapper.run_json_qcschema(json_data, True, True)
```
loriab, line 64:
```suggestion
json_ret = psi4.schema_wrapper.run_qcschema(json_data, True, True)
```
The `run_json_qcschema` fn you were calling is an internal one called by `run_json` (early effort) and `run_qcschema` (current runner), so it shouldn't be called directly. I bet you can remove the two provenance lines you added after making this change. after this, lgtm!

loriab (approved):
lgtm, thank you for the contribution!

@bgpeyton, do you want to look over it again in ~final form? And will either you or @cgbriggs99 confirm that the test runs locally? Without the "quick" marker, it won't have gotten run by CI.

bgpeyton:
(sorry for the delay, libint2 was a bit challenging to get going.) test doesn't pass locally, for a few reasons. there are a lot of comments already, so I'll list the reasons here, as well as comment on the relevant lines of code. 

1) `"schema_version"` key is deprecated in current QCSchema, this makes AtomicInput choke. remove line 12 from the test.
2) when inside `schema_wrapper.py` (which is inside the driver), you don't need to add the prefix `psi4`. `psi4.core.RHF` should be just `core.RHF` on line 594.
3) question for @loriab : `json_ret` is (once the above are fixed) of type `AtomicResult`, which is not JSON serializable. this is indeed being caught by line 72 of the test (good call). is there a simple `AtomicResult` -> `Dict` translation?

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,84 @@
...
+import psi4
+import json
+
+#Generate JSON data
+
+method = "CCSD"
+basis = "sto-3g"
+json_data = {
+    "schema_name": "qcschema_input",
+    "schema_version": 1,
```
bgpeyton, line 11: this key is deprecated, see `AtomicInput` in `qcel.models.results` to see the current verified QCSchema keywords.

psi4/driver/schema_wrapper.py
```diff
@@ -586,6 +586,18 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
             )):
                 json_data["extras"]["qcvars"][k] = _serial_translation(v, json=json_serialization)
 
+    # Add in handling of matrix arguments which need to be obtained by a
+    # a function call.
+
+    if json_data["model"]["method"].lower() in ["ccsd"] :
+        if json_data["extras"].get("psi4:save_tamps", False):
+            if type(wfn.reference_wavefunction()) is psi4.core.RHF :
```
bgpeyton, line 594: `psi4.core.RHF` -> `core.RHF` (we're already inside the `psi4` driver)

tests/json/schema-1-tamps/input.py
```diff
@@ -0,0 +1,84 @@
...
+
+json_ret = psi4.schema_wrapper.run_qcschema(json_data, True)
+psi4.set_options({"basis": basis, "scf_type": "df", "mp2_type": "df"})
+e, wfn = psi4.energy(method, return_wfn = True)
+
+tIJAB = wfn.get_amplitudes()["tIjAb"].to_array()
+tIA = wfn.get_amplitudes()["tIA"].to_array()
+Da = wfn.Da().to_array()
+
+with open("output.json", "w") as ofile:
```
bgpeyton, line 72: as it turns out, `json_ret` is of type `AtomicResult`, which is not JSON serializable. @loriab any advice?

psi4/driver/schema_wrapper.py
loriab, line 594:
```suggestion
            if type(wfn.reference_wavefunction()) is core.RHF :
```

loriab (approved):
With the suggested changes, test will likely work. No barriers to merge after that.

loriab:
The new test works locally; I think it's ready to go.

maxscheurer (approved):

PR #2346, opened 2021-11-08
Clean libmints `new`/`delete`

JonathonMisiewicz:
## Description
There's a memory leak _somewhere_ in the SCF code. This PR replaces some of the manual memory management with `new`/`delete` in `libmints` to cut down the number of places where it could be coming from. (I'd like to start from a zero-leak baseline for DIIS profiling.)

I'll probably need to use `valgrind` to track this down further...

## Checklist
- [x] Quicktests pass

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libmints/basisset.cc
```diff
@@ -96,78 +96,34 @@ BasisSet::BasisSet() {
...
-    uerd_coefficients_[0] = 1.0;
-    uoriginal_coefficients_[0] = 1.0;
+    ao_to_shell_ = std::vector<int>(1, 0);
+    function_to_shell_ = std::vector<int>(1, 0);
+    function_center_ = std::vector<int>(1, 0);
+    shell_center_ = std::vector<int>(1, 0);
+    center_to_nshell_ = std::vector<int>(1, 1);
+    center_to_shell_ = std::vector<int>(1, 0);
+    xyz_ = std::vector<double>(3, 0.0);
     shell_first_ao_[0] = 0;
```
PeterKraus, line 115: This value can also be initialised on L104, right?

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -859,13 +805,13 @@ BasisSet::BasisSet(const std::string &basistype, SharedMolecule mol,
                 shells_[shell_count] =
                     GaussianShell(shelltype, am, shell_nprim, &uoriginal_coefficients_[ustart + atom_nprim],
                                   &ucoefficients_[ustart + atom_nprim], &uerd_coefficients_[ustart + atom_nprim],
-                                  &uexponents_[ustart + atom_nprim], puream, n, xyz_ptr, bf_count);
+                                  &uexponents_[ustart + atom_nprim], puream, n, &xyz_.data()[3 * n], bf_count);
```
PeterKraus, line 808: Also this is confusing. Previously, we passed in the pointer (as opposed to a reference), which if I'm reading this right would point to 0 for n = 0, and then due to the `xyz_ptr += 3`, it works out to point in the same field as [3*n]. Correct?

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -838,9 +784,9 @@ BasisSet::BasisSet(const std::string &basistype, SharedMolecule mol,
         center_to_nshell_[n] = nshells;
         center_to_shell_[n] = shell_count;
         Vector3 xyz = molecule_->xyz(n);
-        xyz_ptr[0] = xyz[0];
-        xyz_ptr[1] = xyz[1];
-        xyz_ptr[2] = xyz[2];
+        xyz_[3 * n + 0] = xyz[0];
```
PeterKraus, line 787: I'm not sure how these two are equivalent. The bit that's puzzling me is the `xyz_ptr += 3` lines below, that were made obsolete by this change - it's clear to me what the code does now. How did it work before?

psi4/src/psi4/libmints/irrep.cc
```diff
@@ -128,13 +128,12 @@ void IrreducibleRepresentation::init(int order, int d, const char *lab, const ch
     else
         csymb = nullptr;
 
-    if (rep) {
-        delete[] rep;
-        rep = nullptr;
+    if (rep.size()) {
```
PeterKraus, line 131: Perhaps I missed this, but how can be sure that `rep` exists here?

psi4/src/psi4/libmints/molecule.cc
```diff
@@ -186,9 +184,7 @@ Molecule &Molecule::operator=(const Molecule &other) {
     // These are symmetry related variables, and are filled in by the following functions
     pg_ = std::shared_ptr<PointGroup>();
     nunique_ = 0;
-    nequiv_ = nullptr;
     equiv_ = nullptr;
```
PeterKraus, line 187: Is there a reason why equiv_ was kept as-is but nequiv_ replaced by a vector?

PeterKraus (commented):
I'm by no means a C++ expert, but the changes look sensible and the code's a lot more readable. The comments are just a couple of clarifications.

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -350,15 +306,15 @@ void BasisSet::print_summary(std::string out) const {
...
+    auto nshells = std::vector<int>(max_am_ + 1);
+    auto amtypes = std::vector<char>(max_am_ + 1);
 
     for (int A = 0; A < molecule_->natom(); A++) {
-        memset((void *)nprims, '\0', (max_am_ + 1) * sizeof(int));
-        memset((void *)nunique, '\0', (max_am_ + 1) * sizeof(int));
-        memset((void *)nshells, '\0', (max_am_ + 1) * sizeof(int));
+        memset((void *)nprims.data(), '\0', (max_am_ + 1) * sizeof(int));
+        memset((void *)nunique.data(), '\0', (max_am_ + 1) * sizeof(int));
+        memset((void *)nshells.data(), '\0', (max_am_ + 1) * sizeof(int));
```
susilehtola, line 317:
```suggestion
std::fill(nprims.begin(), nprims.end(), 0);
std::fill(nunique.begin(), nunique.end(), 0);
std::fill(nshells.begin(), nshells.end(), 0);
```

psi4/src/psi4/libmints/basisset.cc
JonathonMisiewicz: This comment and the one above are both explained by pointer syntax.
`pointer[n]` means "grab the value `n` memory locations away from where the pointer currently points" and `pointer ++ n` means "make the pointer point `n` memory locations forward". So what the old code was doing was moving the pointer ahead three every time it changed atoms, so _relative to the starting position_ it would always point at location `3 * n`.

In short - it's exactly as you said above.

psi4/src/psi4/libmints/irrep.cc
JonathonMisiewicz: C++ won't allow things that don't exist. `rep` is a length-zero vector by default.

PeterKraus (approved):
Thanks Jonathon!

jturney (approved):
Always nice to see updates to modern C++ 😄

jturney:
Yeah, using valgrind will make it easier to find the leak.

JonathonMisiewicz:
> Yeah, using valgrind will make it easier to find the leak.
> 
> At one point, we did have it leak free.

Good to know. Unfortunately, it may be a bit before I can get valgrind working - there are some issues with Francesco's cluster.

loriab (approved):

PR #2347, opened 2021-11-08
Fix SCF memory leak and `Vector::gemv`

JonathonMisiewicz:
## Description
First, this PR cleans up some manual memory management in `libscf_solver`, including a memory leak of an `nao`-by-`nso` matrix created once per call to the SCF code.

During the course of fixing that, I uncovered a correctness error in `libmints`: the `dgemv` function was using rows rather than cols for an argument. The original choice was correct in Fortran-style indexing but not C-style indexing and disagreed with every other `C_DGEMV` call in Psi. It didn't cause any issues because `Vector::dgemv` is almost never used in Psi. The one other time I found it used was for a Hermitian matrix, which of course won't have problems.

## Questions
- [x] Jet, please double-check my claim of a bug in `libmints`.

## Checklist
- [x] Quick tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -566,20 +566,22 @@ void HF::form_H() {
             // Set up AO->SO transformation matrix (u)
             MintsHelper helper(basisset_, options_, 0);
             SharedMatrix aotoso = helper.petite_list(true)->aotoso();
-            int* col_offset = new int[nirrep_];
-            col_offset[0] = 0;
-            for (int h = 1; h < nirrep_; h++) col_offset[h] = col_offset[h - 1] + aotoso->coldim(h - 1);
+            Matrix u(nao, nso);
+            int offset = 0;
 
-            double** u = block_matrix(nao, nso);
-            for (int h = 0; h < nirrep_; h++)
-                for (int j = 0; j < aotoso->coldim(h); j++)
-                    for (int i = 0; i < nao; i++) u[i][j + col_offset[h]] = aotoso->get(h, i, j);
-            delete[] col_offset;
+            for (int h = 0; h < nirrep_; h++) {
+                // These loops should be vectorized for a (small) efficiency gain.
+                for (int j = 0; j < aotoso->coldim(h); j++) {
+                    for (int i = 0; i < nao; i++) {
+                        u.set(i, j + offset, aotoso->get(h, i, j));
+                    }
+                }
+                offset += aotoso->coldim(h);
+            }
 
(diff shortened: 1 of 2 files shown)


jturney (approved):
Yes, you're absolutely right about the gemv call using the wrong value for lda.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -633,33 +635,28 @@ void HF::form_H() {
...
 
             outfile->Printf("  Perturbing H by %f %f %f V_eff.\n", dipole_field_strength_[0], dipole_field_strength_[1],
                             dipole_field_strength_[2]);
-            if (options_.get_int("PRINT") > 3) mat_print(V_eff, nso, nso, "outfile");
+            if (options_.get_int("PRINT") > 3) V_eff.print_out();
 
             if (dipole_field_type_ == dx) {
-                for (int i = 0; i < nso; i++)
-                    for (int j = 0; j < nso; j++) V_->set(i, j, V_eff[i][j]);  // ignore nuclear potential
+                V_->copy(V_eff);
```
hokru, line 655:
```suggestion
                V_->copy(V_eff); // ignore nuclear potential
```

hokru (approved):
Are the code changes covered by a test?
Didn't find 'dipole_field_type' tests on a quick look.

JonathonMisiewicz:
> Are the code changes covered by a test? Didn't find 'dipole_field_type' tests on a quick look.

Only one `if` branch is covered. The relevant keyword is `perturb_with`. While `sphere` is covered by `scf-coverge` (which passes), there is no test suite coverage on `dx` or `embpot`. While user-facing documentation lists the `perturb_with` keywords, the only documentation on what `dx` or `embpot` mean is a comment in-code that they read from external files. The specifications for the respective files `potential.dx` and `EMBPOT` are not documented. To write a test, I would either need to find somebody who knows the specification or back it out from the file parsing functions. While adding such tests is a worthy issue, I believe it's beyond the scope of this PR.

psi4/src/psi4/libscf_solver/hf.cc
JonathonMisiewicz: Is this not obvious from the code? Adding a comment felt redundant to me.

hokru: perhaps. Not insisting on it.

hokru:
> Only one `if` branch is covered

That would be sufficient for this PR in my opinion.

andysim (approved):
Nice cleanup, and good catch regarding the GEMV call.  LGTM!

PR #2349, opened 2021-11-09
Fix `qcdb` memory leak

JonathonMisiewicz:
## Description
This PR fixes a memory leak that was occurring in the Python-side `qcdb` code. The code meant to modify a _class variable_ but was actually creating and modifying an _instance variable_. Initialization happened many more times than needed, and some memory was not being cleared upon object deletion.

The only explanation I have for how this produced a memory leak is that having a class variable and an instance variable with the same name confused the Python garbage collector. I was able to confirm object creation and deletion was occurring roughly as expected.

## Checklist
- [x] Confirmed this change reduces the memory footprint for repeated SCF calls

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/qcdb/libmintsbasisset.py
+++ b/psi4/driver/qcdb/libmintsbasisset.py
@@ -192,24 +192,24 @@ class BasisSet(object):
 
 
     # <<< Methods for Construction >>>
-
-    def initialize_singletons(self):
+    @classmethod
+    def initialize_singletons(cls):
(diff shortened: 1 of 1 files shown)


loriab (approved):
yikes, I didn't know it was in my power to cause memory trouble. thanks!

jturney (approved):

hokru (approved):

PR #2350, opened 2021-11-10
Fix DIIS memory leak

JonathonMisiewicz:
## Description
This PR fixes a memory leak which I inadvertently introduced myself when refactoring `libdiis`.

What I was _trying_ to do was take heap memory and move it into the stack. You can't actually do that. So when I released my `std::vector` from the smart pointer and saved it to a class variable, the memory was _not_ cleared on object delete because it was still heap memory and not stack memory. _Mea culpa_.

## Checklist
- [x] Quick tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


jturney (approved):
Seems reasonable to use std::move like you did. LGTM

loriab (approved):

hokru (approved):

PR #2353, opened 2021-11-13
Mark `Vector::clone` const

JonathonMisiewicz:
## Description
Exactly what it says in the title. @fevangelista, we need this for my Forte fixes.

## Checklist
- [x] Code still compiles

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/vector.cc
+++ b/psi4/src/psi4/libmints/vector.cc
@@ -122,7 +122,7 @@ void Vector::init(const Dimension &v) {
     alloc();
 }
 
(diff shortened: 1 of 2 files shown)


jturney (approved):

PeterKraus (commented):
So this change makes `Vector.clone()` work with vectors that are const themselves, right?

JonathonMisiewicz:
> So this change makes `Vector.clone()` work with vectors that are const themselves, right?

Correct.

loriab (approved):

fevangelista (approved):

PR #2355, opened 2021-11-15
Vectorize libdiis

JonathonMisiewicz:
## Description
This PR uses `axpy` and `memcpy` throughout `libdiis`, for the sake of fairer comparisons when I'm doing speed profiling.

## Checklist
- [x]  quick tests pass

## Status
- [x] Ready for review

### DIFF
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -222,7 +222,6 @@ jobs:
       - script: |
           setlocal EnableDelayedExpansion
           set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
-          psi4 $(Build.SourcesDirectory)\tests\tu1-h2o-energy\input.dat
           psi4 $(Build.SourcesDirectory)\tests\castup1\input.dat
         displayName: 'Run, Run, Spot, Run Test'
         workingDirectory: $(Build.BinariesDirectory)/build
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -222,7 +222,7 @@ jobs:
       - script: |
           setlocal EnableDelayedExpansion
           set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
-          psi4 $(Build.SourcesDirectory)\tests\castup1\input.dat -o stdout
+          psi4 $(Build.SourcesDirectory)\tests\tu1-h2o-energy\input.dat -o stdout
         displayName: 'Run, Run, Spot, Run Test'
         workingDirectory: $(Build.BinariesDirectory)/build
 
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -223,6 +223,7 @@ jobs:
           setlocal EnableDelayedExpansion
           set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
           psi4 $(Build.SourcesDirectory)\tests\tu1-h2o-energy\input.dat
+          psi4 $(Build.SourcesDirectory)\tests\castup1\input.dat
         displayName: 'Run, Run, Spot, Run Test'
         workingDirectory: $(Build.BinariesDirectory)/build
 
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -222,7 +222,7 @@ jobs:
       - script: |
           setlocal EnableDelayedExpansion
           set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
-          psi4 $(Build.SourcesDirectory)\tests\castup1\input.dat
+          psi4 $(Build.SourcesDirectory)\tests\castup1\input.dat -o stdout
         displayName: 'Run, Run, Spot, Run Test'
         workingDirectory: $(Build.BinariesDirectory)/build
 
--- a/psi4/src/psi4/libdiis/diismanager.cc
+++ b/psi4/src/psi4/libdiis/diismanager.cc
@@ -247,7 +247,7 @@ bool DIISManager::add_entry(int numQuantities, ...) {
                     auto size = file2->params->rowtot[h] * file2->params->coltot[file2->my_irrep ^ h];
                     if (size) {
                         std::copy(file2->matrix[h][0], file2->matrix[h][0] + size, arrayPtr);
-                        arrayPtr += sizei;
+                        arrayPtr += size;
                     }
                 }
                 global_dpd_->file2_mat_close(file2);
--- a/psi4/src/psi4/libdiis/diismanager.cc
+++ b/psi4/src/psi4/libdiis/diismanager.cc
@@ -480,10 +480,15 @@ bool DIISManager::extrapolate(int numQuantities, ...) {
                     if (!n) matrix->zero();
                     for (int h = 0; h < matrix->nirrep(); ++h) {
                         auto size = static_cast<size_t>(matrix->rowdim(h)) * matrix->colspi(h ^ matrix->symmetry());
+                        outfile->Printf("Attempting Matrix extrapolation");
                         if (size) {
+                            outfile->Printf("Confirmed nonzero size");
                             C_DAXPY(size, coefficient, arrayPtr, 1, matrix->pointer(h)[0], 1);
+                            outfile->Printf("AXPY executed");
                             arrayPtr += static_cast<int>(size);
+                            outfile->Printf("Pointer arithmetic done");
                         }
+                        outfile->Printf("Matrix extrapolation success for this irrep");
                     }
                     break;
                 }
--- a/psi4/src/psi4/libdiis/diismanager.cc
+++ b/psi4/src/psi4/libdiis/diismanager.cc
@@ -250,7 +250,7 @@ bool DIISManager::add_entry(int numQuantities, ...) {
             {
                 matrix = va_arg(args, Matrix *);
                 for (int h = 0; h < matrix->nirrep(); ++h) {
-                    auto size = matrix->rowspi()[h] * matrix->colspi()[h ^ matrix->symmetry()];
+                    auto size = matrix->rowdim(h) * matrix->coldim(h ^ matrix->symmetry());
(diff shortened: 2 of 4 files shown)


psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -213,53 +213,57 @@ bool DIISManager::add_entry(int numQuantities, ...) {
...
                 buf4 = va_arg(args, dpdbuf4 *);
                 for (int h = 0; h < buf4->params->nirreps; ++h) {
                     global_dpd_->buf4_mat_irrep_init(buf4, h);
                     global_dpd_->buf4_mat_irrep_rd(buf4, h);
-                    for (int row = 0; row < buf4->params->rowtot[h]; ++row) {
-                        for (int col = 0; col < buf4->params->coltot[h]; ++col) {
-                            *arrayPtr++ = buf4->matrix[h][row][col];
-                        }
-                    }
+                    auto size = buf4->params->rowtot[h] * buf4->params->coltot[h ^ buf4->file.my_irrep];
```
jturney, line 229: Is the use of `^ buf4->file.my_irrep` correct? I'm not seeing it in the original code.

psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -213,53 +213,57 @@ bool DIISManager::add_entry(int numQuantities, ...) {
...
                 file2 = va_arg(args, dpdfile2 *);
                 global_dpd_->file2_mat_init(file2);
                 global_dpd_->file2_mat_rd(file2);
                 for (int h = 0; h < file2->params->nirreps; ++h) {
-                    for (int row = 0; row < file2->params->rowtot[h]; ++row) {
-                        for (int col = 0; col < file2->params->coltot[h]; ++col) {
-                            *arrayPtr++ = file2->matrix[h][row][col];
-                        }
-                    }
+                    auto size = file2->params->rowtot[h] * file2->params->coltot[file2->my_irrep ^ h];
```
jturney, line 242: Same here but with `file2->my_irrep`

psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -213,53 +213,57 @@ bool DIISManager::add_entry(int numQuantities, ...) {
...
             case DIISEntry::InputType::Matrix:
+            {
                 matrix = va_arg(args, Matrix *);
                 for (int h = 0; h < matrix->nirrep(); ++h) {
-                    for (int row = 0; row < matrix->rowspi()[h]; ++row) {
-                        for (int col = 0; col < matrix->colspi()[h]; ++col) {
-                            *arrayPtr++ = matrix->get(h, row, col);
-                        }
-                    }
+                    auto size = matrix->rowspi()[h] * matrix->colspi()[h ^ matrix->symmetry()];
```
jturney, line 253: Same.

psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -213,53 +213,57 @@ bool DIISManager::add_entry(int numQuantities, ...) {
...
             case DIISEntry::InputType::Vector:
+            {
                 vector = va_arg(args, Vector *);
-                for (int h = 0; h < vector->nirrep(); ++h) {
-                    for (int row = 0; row < vector->dimpi()[h]; ++row) {
-                        *arrayPtr++ = vector->get(h, row);
-                    }
-                }
+                auto size = vector->dimpi().sum();
+                std::copy(vector->pointer(), vector->pointer() + size, arrayPtr);
```
jturney, line 263: This copy is dependent on how the storage in Vector is done. If (for some reason) Vector's implementation changes this could fail.

psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -426,67 +430,74 @@ bool DIISManager::extrapolate(int numQuantities, ...) {
...
                     buf4 = va_arg(args, dpdbuf4 *);
                     if (!n) global_dpd_->buf4_scm(buf4, 0.0);
                     for (int h = 0; h < buf4->params->nirreps; ++h) {
                         global_dpd_->buf4_mat_irrep_init(buf4, h);
                         global_dpd_->buf4_mat_irrep_rd(buf4, h);
-                        for (int row = 0; row < buf4->params->rowtot[h]; ++row) {
-                            for (int col = 0; col < buf4->params->coltot[h]; ++col) {
-                                buf4->matrix[h][row][col] += coefficient * *arrayPtr++;
-                            }
+                        auto size = buf4->params->rowtot[h] * buf4->params->coltot[h ^ buf4->file.my_irrep];
```
jturney, line 450: Same with the `buf4->file.my_irrep` is it needed?

psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -426,67 +430,74 @@ bool DIISManager::extrapolate(int numQuantities, ...) {
...
                     file2 = va_arg(args, dpdfile2 *);
                     if (!n) global_dpd_->file2_scm(file2, 0.0);
                     global_dpd_->file2_mat_init(file2);
                     global_dpd_->file2_mat_rd(file2);
                     for (int h = 0; h < file2->params->nirreps; ++h) {
-                        for (int row = 0; row < file2->params->rowtot[h]; ++row) {
-                            for (int col = 0; col < file2->params->coltot[h]; ++col) {
-                                file2->matrix[h][row][col] += coefficient * *arrayPtr++;
-                            }
+                        auto size = file2->params->rowtot[h] * file2->params->coltot[file2->my_irrep ^ h];
```
jturney, line 467: Same.

psi4/src/psi4/libdiis/diismanager.cc
```diff
@@ -426,67 +430,74 @@ bool DIISManager::extrapolate(int numQuantities, ...) {
...
                 case DIISEntry::InputType::Matrix:
+                {
                     matrix = va_arg(args, Matrix *);
                     if (!n) matrix->zero();
                     for (int h = 0; h < matrix->nirrep(); ++h) {
-                        for (int row = 0; row < matrix->rowspi()[h]; ++row) {
-                            for (int col = 0; col < matrix->colspi()[h]; ++col) {
-                                matrix->add(h, row, col, coefficient * *arrayPtr++);
-                            }
+                        auto size = matrix->rowspi()[h] * matrix->colspi()[h ^ matrix->symmetry()];
```
jturney, line 482: And again.

psi4/src/psi4/libdiis/diismanager.cc
JonathonMisiewicz: I am 99% sure that when @andysim created the code, he wasn't thinking about non totally-symmetric tensors. I grabbed these from the `axpy` functions of the original classes.

psi4/src/psi4/libqt/qt.h
```diff
@@ -109,7 +109,7 @@ void C_DROT(size_t ntot, double* x, int incx, double* y, int incy, double costhe
 void C_DSWAP(size_t length, double* x, int incx, double* y, int inc_y);
 void C_DSCAL(size_t len, double alpha, double* vec, int inc);
 void C_DCOPY(size_t length, double* x, int inc_x, double* y, int inc_y);
-void C_DAXPY(size_t length, double a, double* x, int inc_x, double* y, int inc_y);
+void C_DAXPY(size_t length, double a, const double* x, int inc_x, double* y, int inc_y);
```
andysim, line 112: We normally omit the `const` in the BLAS/LAPACK wrappers (even though input-only args are not touched), because the underlying Fortran functions can't provide const safety.  It might be a good idea to omit it here, for consistency.

psi4/src/psi4/libdiis/diismanager.cc
JonathonMisiewicz: Change how? Store the data from a single irrep non-contiguously? Store the data from one irrep not contiguously to another irrep?

The way I see it, _some_ dependence on how storage is done in the `Vector` class is unavoidable, unless I swap out `arrayPtr` for an actual `Vector` object.

psi4/src/psi4/libqt/qt.h
JonathonMisiewicz: I can do that, but the price is losing a `const` qualifier on `arrayPtr`.

psi4/src/psi4/libdiis/diismanager.cc
andysim: I think you're right about that.  I wrote it for cumulants and Fock matrices, and likely wasn't forward-thinking enough to think about things like excited states.  The generalization to accommodate those methods is a good addition.

psi4/src/psi4/libqt/qt.h
andysim: Entirely up to you - just wanted to point out the slight inconsistency.  If it introduces the need for annoying casts everywhere in the calling code, it's better to stick with the `const` in the signature.

.azure-pipelines/azure-pipelines-windows.yml
```diff
@@ -222,7 +222,7 @@ jobs:
       - script: |
           setlocal EnableDelayedExpansion
           set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
-          psi4 $(Build.SourcesDirectory)\tests\tu1-h2o-energy\input.dat
+          psi4 $(Build.SourcesDirectory)\tests\tu1-h2o-energy\input.dat -o stdout
```
jturney, line 225: Are we wanting to keep `-o stdout`? I don't care either way.

JonathonMisiewicz: I kept that deliberately, so we get output when something goes wrong.

jturney:
For my own edification, what was needed to fix Windows?

JonathonMisiewicz:
> For my own edification, what was needed to fix Windows?

It either doesn't like adding `0` to a pointer, or doesn't like when the begin and end iterators passed to `std::copy` are the same. Either way, the fix is to check for the 0 case explicitly.

jturney (approved):

andysim (approved):
I'm sad that it's not another negative LoC PR from you, but very nice cleanup nonetheless.  We may need to switch the allocators in `Vector` and `DPDBuf4` to use aligned memory to see the full benefits of this PR, but it LGTM.

loriab (approved):

PR #2361, opened 2021-11-19
native_files field for qcschema

loriab:
## Description
In accordance with https://github.com/MolSSI/QCElemental/pull/275, return certain files in qcschema to be pruned down by user protocols. Bump qcel to 0.24.0 (and qcng to 0.23.0 once it's minted). Note that this can't handle the wfn file b/c binary. And it can't handle `timer.dat` b/c there's no fixed place it lands (I can get it to save calling schema_wrapper directly but not for QCEngine calling psi4).

## Checklist
- [x] Tests added for any new features -- tested at qcengine test_canonical_fields

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/schema_wrapper.py
+++ b/psi4/driver/schema_wrapper.py
@@ -415,8 +415,7 @@ def run_qcschema(input_data, clean=True):
             "version": __version__,
             "routine": "psi4.schema_runner.run_qcschema"
         })
-        if qcel.util.parse_version(qcel.__version__) >= qcel.util.parse_version("0.24.0"):
-            ret_data["native_files"]["input"] = json.dumps(json.loads(input_model.json()), indent=1)
+        ret_data["native_files"]["input"] = json.dumps(json.loads(input_model.json()), indent=1)
 
         exit_printing(start_time=start_time, success=True)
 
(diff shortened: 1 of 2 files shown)


psi4/driver/schema_wrapper.py
```diff
@@ -633,6 +635,14 @@ def run_json_qcschema(json_data, clean, json_serialization, keep_wfn=False):
     if keep_wfn:
         json_data["wavefunction"] = _convert_wavefunction(wfn)
 
+    files = {
+        "psi4.grad": Path(core.get_writer_file_prefix(wfn.molecule().name()) + ".grad"),
+        "psi4.hess": Path(core.get_writer_file_prefix(wfn.molecule().name()) + ".hess"),
+        # binary "psi4.180.npy": Path(core.get_writer_file_prefix(wfn.molecule().name()) + ".180.npy"),
+        "timer.dat": Path("timer.dat"),
```
JonathonMisiewicz, line 642: If "it can't handle timer.dat," why is this here? Just in case the timer is in the current file?

loriab, line 642:
```suggestion
        "timer.dat": Path("timer.dat"),  # ok for `psi4 --qcschema` but fails (no file at path) for `qcengine.run_program(..., "psi4")`
```

loriab: good catch. I wrote the PR blurb from memory w/o realizing the timer.dat line was uncommented in code. I just checked again, and it does work fine locally, so ok to leave as-is, or I can comment it if ppl prefer.

JonathonMisiewicz (approved):
I'm fine with either approach for `timer.dat`. LGTM - just remember to either include the extra comment or remove the line before this gets merged in. I'm fine with either.

andysim (approved):
LGTM!

hokru (approved):

Issue #2351, opened 2021-11-11
The calculation results of different versions (v1.3.2 vs v1.4 vs v1.4.1) of Psi4 vary greatly


blazerygy:
I used Psi4 to calculate the energy of molecules in Conformer Benchmark dataset ( https://github.com/ghutchis/conformer-benchmark/ 702 molecules, each molecule has about 10 conformations).

I tried three versions of Psi4 respectively, including V1.3.2, V1.4 and V1.4.1, and then calculated the correlation between the results of the three versions. The results are shown in the following table. The consistency is not high, especially v1.4.1, which is greatly different from the other two versions. Why are the calculation results of different versions so different?

In addition, I also calculated the correlation between Psi4 results and Orca results (from Conformer benchmark), the correlation is not high, why?

![image](https://user-images.githubusercontent.com/16797849/141259545-c20d1c6c-6fe1-44ec-a0de-1bb8980db327.png)

my code:
```Python
def run_psi4_calc(xyz_str, dtype='xyz', basis_set='WB97X-D/def2-SVP', reference='uks', max_iter=100):

    psi4.set_options({
      'reference': reference,
      'maxiter': max_iter
    })

    mol = psi4.core.Molecule.from_string(xyz_str, dtype=dtype)
    energy = psi4.energy(basis_set, molecule=mol)

    return energy



# get molecular 3d coordinates from Conformer benchmark
xyz_str = get_xyz_str()

energy = run_psi4_calc(
    xyz_str=xyz_str,
    dtype='xyz+',
    basis_set='WB97X-D/def2-SVP',
    reference='uks',
    max_iter=100
)
```

JonathonMisiewicz:
Looking at your GitHub account, I suspect that you have a background in machine learning rather than quantum chemistry, so I'll try to keep this simple.

DFT methods (like WB97X-D) rely on an expensive numerical integration procedure, and different codes use different integration grids with different schemes to remove points to reduce computational time. Chemists are usually interested in _differences_ of energies and will prune points if doing so means energy _differences_ are unchanged, even if final energies are not. Fair comparisons of DFT energies across different programs, like Psi4 and ORCA, are thus very difficult and can only be reliably accomplished by an expert.

[...]

PR #2362, opened 2021-11-22
disables wcombine option

hokru:
Reports indicate that the in-core `MemDF+wcombine` algorithm is not working correctly.
This PR disables the options. MemDF does not longer turn the option on automatically and users setting the option directly will get an exception error. A note is added to the DFT manual.

Issues: #2351 #2279 
Related: #2283 

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz (approved):

loriab (approved):

andysim (approved):
Thanks for addressing this :) Hopefully I can find some time to dig into the problem during next week's hackathon

PR #2369, opened 2021-12-01
PyDIIS

JonathonMisiewicz:
## Description
This is a preliminary PR for moving DIIS Python-side, submitted for general review over PsiCon and to see if Windows hates this.

This is going to expedite EDIIS/ADIIS, and is a prototype for moving more code to the Python layer.

## Todos
- [x] Moves DIIS to Python.

## Questions
- [x] Where should the boundary between Python and C++ be?
- [x] Can we clean up the C-side API at all?
- [x] What kind of compile-time hit do we take by the CMake changes?
- [x] Can we get more detailed estimates on time and memory costs? My benchmarks so far say there's no hit.
- [x] The new DPD functionality is somewhat hacky, but it is DPD. Are we okay with this...?

## Checklist
- [x] Quick tests pass

## Status
- [x] Ready for review
- [ ] Ready for merge


psi4/driver/procrouting/scf_proc/diis.py
```diff
@@ -0,0 +1,200 @@
...
+        self.reset_subspace()
+        if self.storage_policy == StoragePolicy.OnDisk:
+            psio = core.IO.shared_object()
+            if not psio.open_check(psif.PSIF_LIBDIIS):
+                psio.open(psif.PSIF_LIBDIIS, 1) # 1 = PSIO_OPEN_OLD
+
+    def __del__(self):
+        psio = core.IO.shared_object()
+        if psio.open_check(psif.PSIF_LIBDIIS):
+            psio.close(psif.PSIF_LIBDIIS, 1) # 1 = KEEP
```
jturney, line 55: I would prefer the file to be closed only if this object opened it.  Specifically, if line 50 (the `psio.open` call) did not execute then line 55 should not execute.

psi4/driver/procrouting/scf_proc/diis.py
```diff
@@ -0,0 +1,200 @@
...
+                    axpy(Tj, ci, Ti[j])
+            elif self.storage_policy == StoragePolicy.OnDisk:
+                for i, ci in enumerate(coeffs):
+                    psio = core.IO.shared_object()
+                    if isinstance(Tj, core.Vector):
+                        Tij = core.Vector(self.get_name("T", i, j), *self.T_template[j])
+                        Tij.load(psio, psif.PSIF_LIBDIIS)
+                    elif isinstance(Tj, (core.Matrix, core.dpdfile2, core.dpdbuf4)):
+                        Tij = core.Matrix(self.get_name("T", i, j), *self.T_template[j])
+                        Tij.load(psio, psif.PSIF_LIBDIIS, core.SaveType.SubBlocks)
```
jturney, line 187: Does this load all the data into memory? Is that safe to assume in all cases?

psi4/src/psi4/dct/dct_compute_RHF.cc
```diff
@@ -129,7 +130,8 @@ void DCTSolver::run_simult_dct_RHF() {
         "\t*---------------------------------------------------------------------------------*\n");
     auto tmp = std::make_shared<Matrix>("temp", nirrep_, nsopi_, nsopi_);
     // Set up the DIIS manager
-    DIISManager diisManager(maxdiis_, "DCT DIIS vectors");
+    py::object diis_file = py::module_::import("psi4").attr("driver").attr("scf_proc").attr("diis");
+    py::object diis_manager = diis_file.attr("DIIS")(maxdiis_, "DCT DIIS vectors");
```
jturney, line 134: Is there any way to write a a wrapper or something so that this can be hidden? It's not very friendly.

jturney: Perhaps the old C++ DIISManager can to be changed to do all of this.

JonathonMisiewicz: Francesco asked me the same thing. With two votes in favor for exploring that option, I'll give it a shot.

psi4/driver/procrouting/scf_proc/diis.py
JonathonMisiewicz: It loads all the data in memory, yes. The existing code makes even more optimistic memory assumptions: that you can load [all items at once](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libdiis/diismanager.cc#L365-L366). Removing this assumption is a feature addition, rather than a migration.

psi4/driver/procrouting/scf_proc/diis.py
JonathonMisiewicz: This is another [feature addition](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libdiis/diismanager.cc#L545-L547), but one I can do easily enough.

loriab: I wonder if a context manager could be useful here. link is one that I wrote, not one that's particularly relevant https://github.com/MolSSI/QCEngine/blob/master/qcengine/util.py#L558-L559

JonathonMisiewicz: I don't think so. Because file I/O is expensive, we want it to be closed solely on object delete. That outlives any particular function call. The current version seems best, unless I am massively misunderstanding context managers.

JonathonMisiewicz:
Code updated to use a C++ class as a wrapper to the Python interface. The result is _very _ similar to the old interface, but I was able to eliminate redundant arguments.

There's more ripping out of old `libdiis` tech that I could do (`diisentry.cc` is now unused), but I'll wait to hear that the current interface looks good before gutting any more code.

psi4/src/psi4/libscf_solver/hf.h
```diff
@@ -146,7 +148,7 @@ class HF : public Wavefunction {
     /// DIIS manager intiialized?
     bool initialized_diis_manager_;
     /// DIIS manager for all SCF wavefunctions
-    std::shared_ptr<DIISManager> diis_manager_;
+    py::object diis_manager_;
```
jturney, line 151: Why not use the new DIISManager class?

JonathonMisiewicz: Because then the Python-side needs to manipulate a C-side class which is really a wrapper to another Python-class.

I don't think there's a good solution to this: using the Python object makes C-syntax clunky, and using the C-wrapper makes Py-calls go through C++ to access Python. I don't have a firm stance on which is better.

jturney: Ah, ok, makes sense.

jturney:
Should the new 'psi4/driver/procrouting/scf_proc/diis.py' file be moved to a more common directory? New developers may think it shouldn't be used in new code if it lives in the SCF directory.

JonathonMisiewicz:
I've run some timing tests. First, with DLPNO-MP2 on 1644 basis functions, the same job in triplicate:

| Before PR | After PR |
| --- | --- |
| 2:43:56.48 | 2:44:54.51 |
| 2:44:17.27 | 2:44:03.06 |
| 2:45:08.16 | 2:44:24.98 |

And now DF-DCT with 384 basis functions, this time in duplicate:
| Before PR | After PR |
| --- | --- |
| 1:03:58.53 | 1:03:52.70|
| 1:03:49.19 | 1:03:48.30 |

I can run additional tests if requested, but the PR has negligible impact on performance time, as expected.

The PR is ready for more serious review.

jturney (approved):
I looked through this before and thought it looked good then!

JonathonMisiewicz:
After memory profiling, I've discovered that this PR _improves_ memory efficiency. For my 384 basis function DF-DCT system, memory requirements plummet from 14.0 GB to 7.8 GB. I've run some tests on a single benzene system instead of a two benzene system, and I understand the behavior:

This isn't due to shifting to Python, but shifting how we clear some large arrays. The C-DIIS code uses `clear()`, which  [doesn't necessarily de-allocate the memory.](https://www.cplusplus.com/reference/vector/vector/clear/) PyDIIS makes that not our problem. On taking the existing C-side code and forcibly de-allocating the memory, I observe a memory profile indistinguishable from the PyDIIS one.

The large change in memory is most likely because C-DIIS was keeping the allocation space for multiple T2 tensors.

psi4/src/export_dpd.cc
```diff
@@ -0,0 +1,74 @@
...
+#include "psi4/libdpd/dpd.h"
+#include "psi4/libmints/matrix.h"
+
+using namespace psi;
+namespace py = pybind11;
+using namespace pybind11::literals;
+
+void export_dpd(py::module &m) {
+    py::class_<dpdbuf4, std::shared_ptr<dpdbuf4>>(m, "dpdbuf4", "docstring")
+        .def("axpy_matrix", &dpdbuf4::axpy_matrix, "docstring")
```
zachglick, line 40: These functions are very self-explanatory, but it still wouldn't hurt to fill in the docstrings, if you felt like it

zachglick (approved):
LGTM! This PR will serve as a nice model for moving code dealing with logic to the python layer without sacrificing performance.

psi4/src/psi4/dfocc/ccsd_iterations_low.cc
```diff
@@ -63,8 +63,8 @@ void DFOCC::ccsd_iterations_low() {
         if (reference_ == "RESTRICTED") {
             ccsdDiisManager = std::shared_ptr<DIISManager>(
                 new DIISManager(cc_maxdiis_, "CCSD DIIS T Amps", DIISManager::RemovalPolicy::LargestError, DIISManager::StoragePolicy::OnDisk));
-            ccsdDiisManager->set_error_vector_size(2, DIISEntry::InputType::Matrix, T2.get(), DIISEntry::InputType::Matrix, T1.get());
-            ccsdDiisManager->set_vector_size(2, DIISEntry::InputType::Matrix, T2.get(), DIISEntry::InputType::Matrix, T1.get());
+            ccsdDiisManager->set_error_vector_size(T2.get(), T1.get());
+            ccsdDiisManager->set_vector_size(2, T2.get(), T1.get());
```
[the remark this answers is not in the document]
JonathonMisiewicz: Oops, good catch! I'm troubled this even passed tests and will look into it. I fear that quick tests have zero coverage on this.

psi4/driver/procrouting/diis.py
```diff
@@ -0,0 +1,212 @@
...
+    InCore = 1
+    OnDisk = 2
+
+def axpy(y, alpha, x):
+    if isinstance(y, (core.Matrix, core.Vector)):
+        y.axpy(alpha, x)
+    elif isinstance(y, (core.dpdbuf4, core.dpdfile2)):
+        y.axpy_matrix(x, alpha)
+    else:
+        raise Exception("Unrecognized object type for DIIS.")
```
maxscheurer, line 21: Maybe better to raise `TypeError` or `NotImplementedError`?

psi4/driver/procrouting/diis.py
```diff
@@ -0,0 +1,212 @@
...
+    if template:
+        raise Exception("Template already set.")
+
+    for arg in args:
+        if isinstance(arg, core.Vector):
+            template.append([arg.dimpi()])
+        elif isinstance(arg, (core.Matrix, core.dpdfile2, core.dpdbuf4)):
+            template.append([arg.rowdim(), arg.coldim()])
+        else:
+            raise Exception("Unrecognized object type for DIIS.")
```
maxscheurer, line 33: Same as above, I'd prefer `TypeError`.

psi4/driver/procrouting/diis.py
```diff
@@ -0,0 +1,212 @@
...
+        self.last_added = -1
+        self.cached_dot_products = dict()
+
+    def copier(self, x, new_name):
+        if isinstance(x, (core.Matrix, core.Vector)):
+            copy = x.clone()
+        elif isinstance(x, (core.dpdbuf4, core.dpdfile2)):
+            copy = core.Matrix(x)
+        else:
+            raise Exception("Unrecognized object type for DIIS.")
```
maxscheurer, line 71: As per comment above, `TypeError` would be nicer.

maxscheurer (approved):
Nice! 👍

JonathonMisiewicz:
I'll give another 48 hours for any devs to say if they want more time to review this, but otherwise, I'll merge this in on Wednesday.

psi4/src/CMakeLists.txt
```diff
@@ -2,17 +2,13 @@
...
-        Python::Module
-    )
-endif()
+include_directories(
+    ${Python_INCLUDE_DIRS}
+)
+target_link_libraries(core
+  PRIVATE
+    Python::Module
+)
```
loriab, line 11: did this come about b/c the build couldn't find Python.h at some point? using raw variables and general scope (e.g., `${Python_INCLUDE_DIRS}`) instead of targets (e.g., `Python::Module` and `core`) isn't the best practice. I've looked around https://pybind11.readthedocs.io/en/stable/compiling.html#advanced-interface-library-targets some and not fastened upon an ideal solution. `Python::Embed` is what we want to avoid iirc because that explicitly links core to libpython.so . a `target_include_directories(core SCOPE ${Python_INCLUDE_DIRS})` would be preferable if it could be here only or wherever the pybind11.h appeared only. This is rambling, so the main question is what was the error that this solved?

psi4/driver/procrouting/diis.py
```diff
@@ -0,0 +1,212 @@
...
+
+class StoragePolicy(Enum):
+    InCore = 1
+    OnDisk = 2
+
+def axpy(y, alpha, x):
+    if isinstance(y, (core.Matrix, core.Vector)):
+        y.axpy(alpha, x)
+    elif isinstance(y, (core.dpdbuf4, core.dpdfile2)):
+        y.axpy_matrix(x, alpha)
```
loriab, line 19: whoa, both `(alpha, x)` and `(x, alpha)`. typical development at different times :-)

loriab (approved):
this became quite a big code shift! much easier to read now.

psi4/driver/procrouting/diis.py
```diff
@@ -0,0 +1,212 @@
...
+            template.append([arg.dimpi()])
+        elif isinstance(arg, (core.Matrix, core.dpdfile2, core.dpdbuf4)):
+            template.append([arg.rowdim(), arg.coldim()])
+        else:
+            raise TypeError("Unrecognized object type for DIIS.")
+
+class DIIS:
+
+    def __init__(self, max_vecs: int, name: str, removal_policy = RemovalPolicy.LargestError, storage_policy = StoragePolicy.OnDisk):
+        assert isinstance(removal_policy, RemovalPolicy)
```
PhillCli, line 38: generally `assert` can be optimized away when python code is compiled to bytecode with optimization [python3 docs](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

```
The current code generator emits no code for an assert statement when optimization is requested at compile time.
```
one can replace all occurrences with `if not` + runtime exception, but it's less concise. 

I guess it's a question is how defensive psi4 is supposed to be wrt to that runtime type checking vs the readability of `if not` clauses.

JonathonMisiewicz: `assert` isn't exactly user-friendly, so I might as well replace them.

psi4/driver/procrouting/diis.py
JonathonMisiewicz: `libmints` wants `(alpha, x)` and `libdpd` wants `(x, alpha)`. But this particular function is one I created to pass `libmints` objects to `libdpd`. As the functions lived in `libdpd`, I opted to have them use `libdpd` convention, but I can go the other way.

psi4/src/psi4/libscf_solver/hf.h
```diff
@@ -35,6 +35,8 @@
 #include "psi4/libmints/vector3.h"
 #include "psi4/psi4-dec.h"
 
+#include "psi4/pybind11.h"
```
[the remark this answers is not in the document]
JonathonMisiewicz: The parts of the SCF code that need this should be Py-side anyways, but that's outside of scope. My laptop is super slow for compile time, but the hint wasn't as bad as I feared.

psi4/src/CMakeLists.txt
JonathonMisiewicz: Exactly.

```
In file included from /Users/jonathonmisiewicz/psi4/psi4/src/psi4/scfgrad/response.cc:50:
In file included from /Users/jonathonmisiewicz/psi4/psi4/src/psi4/libscf_solver/rhf.h:33:
In file included from /Users/jonathonmisiewicz/psi4/psi4/src/psi4/libscf_solver/hf.h:38:
In file included from /Users/jonathonmisiewicz/psi4/psi4/include/psi4/pybind11.h:38:
In file included from /Users/jonathonmisiewicz/miniconda3/envs/p4dev/include/pybind11/pybind11.h:44:
In file included from /Users/jonathonmisiewicz/miniconda3/envs/p4dev/include/pybind11/attr.h:13:
In file included from /Users/jonathonmisiewicz/miniconda3/envs/p4dev/include/pybind11/cast.h:13:
In file included from /Users/jonathonmisiewicz/miniconda3/envs/p4dev/include/pybind11/pytypes.h:12:
/Users/jonathonmisiewicz/miniconda3/envs/p4dev/include/pybind11/detail/common.h:128:10: fatal error: 'Python.h' file not found
#include <Python.h>
```

loriab: You could try the below in libscf_solver/CMakeLists.txt. Otherwise, merge as it, and I can play with it later.

```
 target_include_directories(scf_solver_or_whatever
      PRIVATE
        ${Python_INCLUDE_DIRS}
    )
```

JonathonMisiewicz: I played with it myself and found that I needed to include it not only in `libdiis` and `libscf_solver` (both reasonable) but anything that imports `libscf_solver/hf.h`. That would be a nightmare for downstream. I'll leave this one for you to play with later.

Issue #2299, opened 2021-09-23
allow mbis_volume_ratios from scf_properties


loriab:
The MBIS regression reported in #2272 was fixed via #2273, but there's still a few items to tackle. This part needn't be backported.

- [ ] oeprop(wfn, "MBIS_VOLUME_RATIOS") should be fine (indep oeprop fn), but set scf_properties mbis_volume_ratios; energy("scf") will still fail as MBIS fails via QCEngine #2272 reported because those are OEProp class instantiations called from proc.py, and the free atom volumes aren't available. The oeprop.cc code could exit gracefully when free atom volumes aren't available, but having different properties lists for the two calling routes isn't good.
- [ ] so why not add the loc from oeprop() to OEProps in proc.py so that atom volumes are available? nice thought, but the fn that produces them itself calls oeprop() and energy() and descends into endless recursion. I haven't sought the logic that makes this all work together.
- [ ] add tests.

JonathonMisiewicz:
These comments are too terse for me to follow. If you'll be taking this issue yourself and therefore it doesn't matter if others can understand (as I expect), that's great, but please assign this issue to yourself. Otherwise, can you provide additional context and a minimal not-working example?

loriab:
Good point -- @jeffschriber understands the issue and plans to address it sometime. The summary is that the new MBIS volume ratio functionality is only half available. The `energy(); oeprop(..., "MBIS_VOLUME_RATIOS")` route is fine, but `set scf_properties [mbis_volume_ratios]; energy()` fails b/c some necessary data (free atom volumes) isn't available. This isn't a regression. It'd just be nice to even up the capabilities, esp. as a qcschema run can only use the second form.

jeffschriber:
Exactly. I started looking into this a little while ago, but it fell off the radar a bit. I'll be able to pick it back up early next week

jthorton:
@loriab @jeffschriber Hi  is there any update on this issue as it is breaking a workflow I have where I calculate MBIS charges using wavefunctions stored in QCArchive the attached script shows how this works and results in the following error when using psi4-1.4.1.
```
Properties will be evaluated at   0.000000,   0.000000,   0.000000 [a0]
  Computing free-atom volumes
Traceback (most recent call last):
  File "charges_from_wavefunction.py", line 72, in <module>
    psi4.oeprop(hf_wfn, "MBIS_CHARGES")
[...]

PR #2370, opened 2021-12-01
Fixes to MBIS volume ratios

jeffschriber:
## Description
This PR has tho objectives:

1.  Make MBIS volume ratios callable from both `oeprop(...,'MBIS_VOLUME_RATIOS')` and `set scf_properties ['MBIS_VOLUME_RATIOS']`. So, solve #2299.   
2. A recent comment suggested that the `free_atom_volume()` function was still being called even though they only requested computation of MBIS charges. This PR also ensures that this doesn't happen anymore.

## Todos

- [X] Solve #2299 
- [X] Remove calling of free atom functions when volume ratios aren't requested.

## Checklist

## Status
- [X] Ready for review
- [X] Ready for merge

### DIFF
--- a/tests/mbis-6/input.dat
+++ b/tests/mbis-6/input.dat
@@ -61,8 +61,9 @@ set {
   max_radial_moment 4
 }
 
+set scf_properties ['MBIS_CHARGES', 'MBIS_VOLUME_RATIOS']
+
 e, wfn = energy('b3lyp/cc-pvdz', return_wfn=True)
-oeprop(wfn, 'MBIS_VOLUME_RATIOS', title='H20 DFT')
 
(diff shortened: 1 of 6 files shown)


tests/mbis-6/input.dat
```diff
@@ -61,8 +61,9 @@ set {
   max_radial_moment 4
 }
 
+set scf_properties ['MBIS_CHARGES', 'MBIS_VOLUME_RATIOS']
```
loriab, line 64: This route (set scf_properties, then energy()) should run now, right? ref.dat has error at end, so I may have missed the point.

jeffschriber: Yes, I copied the wrong output

JonathonMisiewicz (approved):

fevangelista (approved):

PR #2376, opened 2021-12-06
Tighten `fcidump` SCF convergence

JonathonMisiewicz:
## Description
`compare_fcidumps` tests to 1e-10, so let's converge to 1e-10.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/fcidump/input.dat
+++ b/tests/fcidump/input.dat
@@ -8,8 +8,8 @@ set {
   basis 6-311g
   scf_type pk
   reference rhf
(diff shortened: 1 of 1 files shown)


fevangelista:
Typically one only needs a `d_convergence` of about sqrt(e_convergence), so `d_convergence 6` should be enough. You can verify by checking the output of these tests.

JonathonMisiewicz:
Test failure healed after fixing an unrelated bug.

Issue #2023, opened 2020-10-27
XHOST detection fails


susilehtola:
Running CMake on my laptop with Fedora 32 results in
```
-- Performing Test CMAKE_CXX_FLAGS [-xHost] - Failed
-- Performing Test CMAKE_CXX_FLAGS [-march=native] - Failed
-- Performing Test CMAKE_CXX_FLAGS [/arch:AVX2] - Failed
CMake Warning at cmake/psi4OptionsTools.cmake:108 (message):
  Option unfulfilled as none of -xHost;-march=native;/arch:AVX2 valid
Call Stack (most recent call first):
  cmake/psi4OptionsTools.cmake:162 (add_C_or_CXX_flags)
  cmake/psi4OptionsTools.cmake:187 (add_CXX_flags)
  cmake/psi4OptionsTools.cmake:202 (add_flags)
  CMakeLists.txt:134 (option_with_flags)
```
CMakeError.log is
```
Performing C SOURCE FILE Test test_option failed with the following output:
Change Dir: /home/work/psi4/objdir/CMakeFiles/CMakeTmp

Run Build Command(s):/usr/bin/gmake cmTC_686f3/fast && /usr/bin/gmake  -f CMakeFiles/cmTC_686f3.dir/build.make CMakeFiles/cmTC_686f3.dir/build
gmake[1]: Entering directory '/home/work/psi4/objdir/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_686f3.dir/src.c.o
/usr/lib64/ccache/cc   -Dtest_option -xHost   -xHost -o CMakeFiles/cmTC_686f3.dir/src.c.o   -c /home/work/psi4/objdir/CMakeFiles/CMakeTmp/src.c
cc: error: language Host not recognized
cc: error: language Host not recognized
gmake[1]: *** [CMakeFiles/cmTC_686f3.dir/build.make:86: CMakeFiles/cmTC_686f3.dir/src.c.o] Error 1
gmake[1]: Leaving directory '/home/work/psi4/objdir/CMakeFiles/CMakeTmp'
gmake: *** [Makefile:141: cmTC_686f3/fast] Error 2


Source file was:
int main(void) { return 0; }
```
making it seem that the CMake files aren't even testing the flags.

If I change the order in CMakeLists.txt so that `-march=native` comes first, the output is correct:
```
-- Setting option ENABLE_XHOST: ON
[...]

PR #2377, opened 2021-12-07
Make ENABLE_XHOST flag checks dependent on compiler brand.

susilehtola:
## Description

This PR solves an issue with the use of the native optimization flags, which are different in various compilers and often get misinterpreted. It looks like the best solution is to first check the brand of the compiler, and then try whether the brand-specific flag(s) work.

Related issues https://gitlab.com/libxc/libxc/-/issues/361 and #2023

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab:
If this is the only way that works, so be it. But have you tried adding `"-tp host"` to the list? provided there isn't another compiler that implements that flag for another purpose, it theoretically should work.

susilehtola:
> If this is the only way that works, so be it. But have you tried adding `"-tp host"` to the list? provided there isn't another compiler that implements that flag for another purpose, it theoretically should work.

The point is that it's excluded since the checks are compiler vendor specific, e.g. `-tp host` is only checked for PGI.

jturney (approved):

JonathonMisiewicz (approved):

hokru (approved):
lgtm

PR #2378, opened 2021-12-07
DLPNO-MP2 Performance Improvement

zachglick:
## Description

Simple performance improvement for the DLPNO-MP2 algorithm. For DLPNO-MP2 computations on many-core machines, I see speedups around 20%

## Todos

- [x] Clone (rather than create from scratch) a `TwoBodyAOInt` for each thread
- [x] More appropriate OMP scheduling

## Checklist
- [x] DLPNO tests still pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dlpno/mp2.cc
+++ b/psi4/src/psi4/dlpno/mp2.cc
@@ -224,7 +224,7 @@ void DLPNOMP2::compute_overlap_ints() {
     }
 
     timer_on("Integration");
-#pragma omp parallel for schedule(static, 1)
+#pragma omp parallel for schedule(dynamic, 1)
     for (size_t Q = 0; Q < grid.blocks().size(); Q++) {
         size_t thread = 0;
 #ifdef _OPENMP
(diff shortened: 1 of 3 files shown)


JonathonMisiewicz (approved):
I'll trust you on dynamic being more appropriate than static here. LGTM.

susilehtola:
what about `guided`?

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -1180,7 +1181,9 @@ double DLPNOMP2::compute_energy() {
 
     print_header();
 
+    timer_on("Setup");
```
hokru, line 1184: Could this be more descriptive? Setup does the PAOs?

jturney (approved):

zachglick:
> what about `guided`?

All that matters is that we avoid `static`, which can lead to an uneven distribution of work across threads. `guided` is about as good as `dynamic` here. `dynamic` has optimal load balancing at the cost of slightly more overhead than `guided`. The work/iteration is large for these loops, making this additional overhead is negligible. 

Here's a comparison of the different schedules (wall time in seconds, large alkane on an 18 core CPU):

| Code | `static` | `guided` | `dynamic` |
|------|-----------|--------|---------|
| Numerical Integration (Loop 1)    | 36.9      | 36.9    | 36.7     |
| DF Integrals (Loop 2)    | 299.0      | 278.7    | 279.7     |
| PNO Transform (Loop 3)    | 80.9      | 80.4    | 78.4     |
| PNO Overlap (Loop 4)    | 29.8      | 29.3    | 24.4     |
| Total DLPNO-MP2    | 528.2      | 506.6    | 498.2     |

You can see that `dynamic` is possibly a hair better than `guided`, but both are better than `static`

psi4/src/psi4/dlpno/mp2.cc
zachglick: Yes, `setup()` does PAOs and also LMOs. I updated the function name to `setup_orbitals()`

loriab (approved):

Issue #2379, opened 2021-12-08
Move IBO localizer into libmints


hokru:
Consolidate https://github.com/psi4/psi4/blob/master/psi4/src/psi4/fisapt/local2.cc into https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/local.cc

The goal is to get straightforward access to the IBO localizer without the need to go into the fsapt name space.
A new subclass would need to be created for IBO.

Tentatively labeled as _new_dev_ but not very familiar with the code base.

susilehtola:
And with #2336 one can easily build the atomic overlap matrices, which are the only thing needed to implement generalized Pipek-Mezey of [J. Chem. Theory Comput. 10, 642 (2014)](https://doi.org/10.1021/ct401016x).

All variants of generalized Pipek-Mezey (Mulliken charges, IAO charges, Becke charges) should have the share the same implementation; the only difference is how the overlap matrix is defined.

PR #2385, opened 2021-12-18
Moved IBOLocalizer to Libmints from FISAPT

andyj10224:
## Description
Addresses issue #2379. IBOLocalizer is a MO localizer that exists in the FISAPT branch. Moving it to libmints cleans up the code.

## Checklist
- [ ] All fsapt tests run
- [ ] FSAPT tests passing
- [ ] FSAPT-D tests passing

## Status
- [ ] Ready for review
- [ ] Ready for merge


andyj10224:
This PR addresses #2379 

All fsapt tests pass after refactoring (except fsapt-d) Any insights? This is the error message:
`!----------------------------------------------------------------------------------!
!  Unsuccessful run. Check input, particularly geometry in [a0]. Model:            !
!     Model(method='hf-d3', basis='(auto)')                                        !
!----------------------------------------------------------------------------------!`

hokru:
Could you post the whole test output please?

And do you have the `dftd3` binary available at run time?

andyj10224:
> Could you post the whole test output please?
> 
> And do you have the `dftd3` binary available at run time?

Yep, it was a dftd3 issue. Rerunning tests now.

andyj10224:
All tests passing. PR ready for review.

susilehtola:
IBO could be combined with Pipek-Mezey; the only difference is the way the orbital overlap is computed.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -118,8 +118,8 @@ void FISAPT::localize() {
     ranges.push_back(vectors_["eps_focc"]->dimpi()[0]);
     ranges.push_back(vectors_["eps_occ"]->dimpi()[0]);
 
-    std::shared_ptr<fisapt::IBOLocalizer2> local =
```
JonathonMisiewicz, line 121: While this is tangential to the purpose of the PR, please do the trivial cleanup of using `auto` in all three code blocks where the variable is defined.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -3106,8 +3106,8 @@ void FISAPT::flocalize() {
             new Matrix("Focc", vectors_["eps_occ0A"]->dimpi()[0], vectors_["eps_occ0A"]->dimpi()[0]));
```
JonathonMisiewicz, line 3106: While you're here, please change to use `make_shared`. Do the same on 3174.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
     U_->transpose_this();
 }
 
+IBOLocalizer::IBOLocalizer(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> minao,
+                             std::shared_ptr<Matrix> C)
+    : Localizer(primary, C), minao_(minao) {
+    common_init();
+}
+IBOLocalizer::~IBOLocalizer() {}
+void IBOLocalizer::common_init() {
```
JonathonMisiewicz, line 559: `common_init` is not marked `override`. Is it meant to override the base class constructor or not?

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+    local->set_condition(options.get_double("LOCAL_IBO_CONDITION"));
+    local->set_power(options.get_double("LOCAL_IBO_POWER"));
+    local->set_use_stars(options.get_bool("LOCAL_IBO_USE_STARS"));
+    local->set_stars_completeness(options.get_double("LOCAL_IBO_STARS_COMPLETENESS"));
+
+    std::vector<int> stars;
+    for (int ind = 0; ind < options["LOCAL_IBO_STARS"].size(); ind++) {
+        stars.push_back(options["LOCAL_IBO_STARS"][ind].to_integer() - 1);
+    }
+    local->set_stars(stars);
```
JonathonMisiewicz, line 592: Why is this not in `common_init`?

The current content of `common_init` is irrelevant when you do this setting right after.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+
+    // => Overlap Integrals <= //
+
+    auto fact11 = std::make_shared<IntegralFactory>(primary_, primary_, primary_, primary_);
+    auto fact12 = std::make_shared<IntegralFactory>(primary_, minao_, primary_, minao_);
+    auto fact22 = std::make_shared<IntegralFactory>(minao_, minao_, minao_, minao_);
+
+    std::shared_ptr<OneBodyAOInt> ints11(fact11->ao_overlap());
+    std::shared_ptr<OneBodyAOInt> ints12(fact12->ao_overlap());
+    std::shared_ptr<OneBodyAOInt> ints22(fact22->ao_overlap());
```
JonathonMisiewicz, line 639: Please change to `unique`. These could be raw objects if we didn't reset.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+    std::shared_ptr<Matrix> S11_m12(S11->clone());
+    std::shared_ptr<Matrix> S22_m12(S22->clone());
+    S11_m12->copy(S11);
+    S22_m12->copy(S22);
+    S11_m12->power(-1.0 / 2.0, condition_);
+    S22_m12->power(-1.0 / 2.0, condition_);
+
+    // => Tilde C <= //
+
+    std::shared_ptr<Matrix> C = C_;
```
JonathonMisiewicz, line 689: There's a ton of `auto` to use from here on.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+    bool converged = false;
+
+    if (power != 2 && power != 4) throw PSIEXCEPTION("IAO: Invalid metric power.");
+
+    outfile->Printf("    @IBO %4s: %24s %14s\n", "Iter", "Metric", "Gradient");
+
+    for (int iter = 1; iter <= maxiter; iter++) {
+        double metric = 0.0;
+        for (int i = 0; i < nocc; i++) {
+            for (int A = 0; A < minao_inds.size(); A++) {
```
JonathonMisiewicz, line 749: Use range-based for loops.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+        metric = pow(metric, 1.0 / power);
+
+        double gradient = 0.0;
+        for (int ind = 0; ind < rot_inds.size(); ind++) {
+            int i = rot_inds[ind].first;
+            int j = rot_inds[ind].second;
+
+            double Aij = 0.0;
+            double Bij = 0.0;
+            for (int A = 0; A < minao_inds.size(); A++) {
```
JonathonMisiewicz, line 767: Range-based for loops.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+
+void IBOLocalizer::print_charges(double scale) {
+    if (!A_) build_iaos();
+
+    std::shared_ptr<Matrix> L = linalg::triplet(C_, S_, A_, true, false, false);
+
+    int nocc = L->rowspi()[0];
+    int natom = true_atoms_.size();
+
+    std::shared_ptr<Matrix> Q = orbital_charges(L);
```
JonathonMisiewicz, line 1058: There is no reason for `Q`, `N`, or `A` to be a pointer. Change these to be raw objects. That'll mean changing the signature of `orbital_charges`.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,544 @@ void PMLocalizer::localize() {
...
+        outfile->Printf("    Pi Completeness = %11.3f\n", stars_completeness_);
+        outfile->Printf("    Number of Pis   = %11zu\n", pi_orbs.size());
+        outfile->Printf("    Number of Stars = %11zu\n", stars_.size());
+        outfile->Printf("    Star Centers: ");
+        for (int ind = 0; ind < stars_.size(); ind++) {
+            outfile->Printf("%3d ", stars_[ind] + 1);
+        }
+        outfile->Printf("\n\n");
+
+        std::map<std::string, std::shared_ptr<Matrix> > ret2 =
```
JonathonMisiewicz, line 947: `auto`

JonathonMisiewicz (commented):
This is mostly moving a class over, so I don't have many comments, but there is a design decision that makes no sense to me post-move, and quite a lot of type nitpicking.

IMO, combining Pipek-Mizey and IBO localization is a refactor outside the scope of this PR.

psi4/src/psi4/libmints/local.h
```diff
@@ -154,6 +155,108 @@ class PSI_API PMLocalizer : public Localizer {
...
+    virtual ~IBOLocalizer();
+
+    /// Build IBO with defaults from Options object (including MINAO_BASIS)
+    static std::shared_ptr<IBOLocalizer> build(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> minao,
+                                                std::shared_ptr<Matrix> C, Options& options);
+
+    // => Computers <= //
+
+    /// Print out the localization algorithm and parameters
+    virtual void print_header() const;
```
zachglick, line 229: Same as the above destructor

zachglick (commented):
Thanks for taking this on Andy! It will be very useful to have IBOs outside of the FSAPT code.

I imagine a lot of psi4 users will want to export and visualize local orbitals. For this reason, we definitely want the IBOLocalizer exposed py-side as well, same as the other two localizers. See `export_mints.cc`

Something else we should think about (other devs please chime in) is how the various derived `Localizer` classes can be modified to have as similar method signatures as possible. The `BoysLocalizer` and `PMLocalizer` have a `void`-type `localize()` function and accessors `L()` and `U()`. The `IBOLocalizer` class has no accessors and instead returns a map of info from the `localize` function. I'm not sure which of these designs I prefer, but I think the three derived classes should at least be consistent.

Lastly, I noticed that there aren't really any tests or docs on the various localizers (except for class autodocs). It would be super nice if we could get one or both of those.

hokru:
> The BoysLocalizer and PMLocalizer have a void-type localize() function and accessors L() and U(). The IBOLocalizer class has no accessors and instead returns a map of info from the localize function. I'm not sure which of these designs I prefer, but I think the three derived classes should at least be consistent.

Agree. Consistency is crucial and within the PR scope. I think the accessors design is more common in psi4 but other than that I have no strong opinion on either choice.

psi4/src/psi4/libmints/local.h
```diff
@@ -154,6 +155,108 @@ class PSI_API PMLocalizer : public Localizer {
...
+    static std::shared_ptr<Matrix> reorder_orbitals(std::shared_ptr<Matrix> F, const std::vector<int>& ranges);
+    /// Orbital atomic charges (natom x nmo)
+    std::shared_ptr<Matrix> orbital_charges(std::shared_ptr<Matrix> L);
+
+   public:
+    // => Constructors <= //
+
+    IBOLocalizer(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> minao, std::shared_ptr<Matrix> C);
+
+    virtual ~IBOLocalizer();
```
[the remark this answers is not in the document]
JonathonMisiewicz: For that matter, I don't think any of the subclass destructors are necessary. They're identical to the base class destructor.

JonathonMisiewicz:
All good points from Zach.

I slightly prefer map return style; it makes it more transparent that the `Localizer` class is a glorified wrapper on the `localize` function. But my opinions are loosely held.

Incidentally, `localize_task` should be `localize` for sake of consistency.

psi4/src/psi4/libmints/local.cc
andyj10224: I am following the convention of the base Localizer class, which does the same thing (initializes instance variables in `common_init`, and then sets them with options in the `Localizer::build` method.

psi4/src/psi4/libmints/local.cc
andyj10224: I think Q (orbital_charges) should remain a `shared_ptr`, since the fsapt code needs to access the variable, and we don't wanna copy matrices.

andyj10224:
> Thanks for taking this on Andy! It will be very useful to have IBOs outside of the FSAPT code.
> 
> I imagine a lot of psi4 users will want to export and visualize local orbitals. For this reason, we definitely want the IBOLocalizer exposed py-side as well, same as the other two localizers. See `export_mints.cc`
> 
> Something else we should think about (other devs please chime in) is how the various derived `Localizer` classes can be modified to have as similar method signatures as possible. The `BoysLocalizer` and `PMLocalizer` have a `void`-type `localize()` function and accessors `L()` and `U()`. The `IBOLocalizer` class has no accessors and instead returns a map of info from the `localize` function. I'm not sure which of these designs I prefer, but I think the three derived classes should at least be consistent.
> 
> Lastly, I noticed that there aren't really any tests or docs on the various localizers (except for class autodocs). It would be super nice if we could get one or both of those.

Hey Zach and all, I just updated the code to conform to the convention of the other localizers. I split the old multi-argument `localize` function in `IBOLocalizer` into the `localize` and `update_fock` methods inherited from the Localizer base class.

psi4/src/psi4/libmints/local.h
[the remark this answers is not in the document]
susilehtola: @zachglick the virtual is important if you make a child class of this class; otherwise the child class can't override it anymore. Right?

psi4/src/psi4/libmints/local.cc
JonathonMisiewicz: I don't see how a copy would occur.

For expressions like `variable_type my_variable = my_function(my_arguments)`, C++ is smart enough not to create the return value and then copy it to the variable. It's called _return value optimization_.

psi4/src/psi4/libmints/local.cc
JonathonMisiewicz: For future reference, "following the convention" is often not a good argument if there's not a good reason for the convention.

But in this case, there is a reason that I missed the first time. Add a comment in the header file saying that the user will normally set options by passing in an `options` object through `build`, but options are initialized in `common_init` in case that doesn't happen.

JonathonMisiewicz:
I'll leave a first pass on this to Zach, but a couple notes:
* Do _not_ mark a comment as resolved unless you've actually resolved it. I see a few cases where you "resolved" some of my comments but didn't make any code changes or leave comments as to why not.
* You'll need to fix the merge conflicts with FSAPT. It's probably just the change in copyright date.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -118,13 +118,13 @@ void FISAPT::localize() {
...
     ranges.push_back(vectors_["eps_occ"]->dimpi()[0]);
 
-    std::shared_ptr<fisapt::IBOLocalizer2> local =
-        fisapt::IBOLocalizer2::build(primary_, reference_->get_basisset("MINAO"), matrices_["Cocc"], options_);
-    local->print_header();
-    std::map<std::string, std::shared_ptr<Matrix> > ret = local->localize(matrices_["Cocc"], Focc, ranges);
+    auto local = IBOLocalizer::build(primary_, reference_->get_basisset("MINAO"), matrices_["Cocc"], options_);
+    local->set_ranges(ranges);
+    local->localize();
+    local->fock_update(Focc);
```
zachglick, line 124: It appears that the `IBOLocalizer::build` call is always accompanied by exactly one call to `set_ranges` before localization and exactly one call to `fock_update` after localization. For this reason, I don't think `set_ranges` or `fock_update` should be user facing functions. They don't add any flexibility to the user, and they require the user to memorize a list of additional functional calls.

Perhaps `ranges` and `Focc` can be made arguments to `build`, `set_ranges` can be made to call `build`, and `localize` can be made to also call `fock_update`.

psi4/src/psi4/libmints/local.h
```diff
@@ -154,6 +155,106 @@ class PSI_API PMLocalizer : public Localizer {
...
+                                                std::shared_ptr<Matrix> C, Options& options);
+
+    // => Computers <= //
+
+    /// Print out the localization algorithm and parameters
+    void print_header() const override;
+
+    /// Localize function from parent class
+    void localize() override;
+    /// Fock update function for IBO (Slightly different than for regular localizers)
```
zachglick, line 237: An informative comment would be helpful here

zachglick (commented):
This is looking better Andy! I left a couple of comments. I think you might've missed part of my original review. We need to be able to export these orbitals py-side and we'll also need at least one test to demonstrate that the exporting works okay.

psi4/src/psi4/libmints/local.h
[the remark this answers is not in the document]
zachglick: That sounds right @susilehtola. I don't think we anticipate making a child class of `IBOLocalizer`, so we should be okay

psi4/src/psi4/libmints/local.cc
andyj10224: I changed N to be a raw object, but not Q, since the fisapt code expects Q to be a pointer, and not a raw object.

psi4/src/psi4/libmints/local.cc
```diff
@@ -549,4 +550,527 @@ void PMLocalizer::localize() {
...
+
+    auto D = linalg::doublet(C, C, false, true);
+    auto Dtilde = linalg::doublet(Ctilde, Ctilde, false, true);
+
+    // => A (Before Orthogonalization) <= //
+
+    auto DSDtilde = linalg::triplet(D, S11, Dtilde, false, false, false);
+    DSDtilde->scale(2.0);
+
+    auto L = linalg::doublet(S11_m12, S11_m12, false, false);  // TODO: Possibly Unstable
```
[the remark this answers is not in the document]
andyj10224: This was pre-existing

psi4/src/psi4/fisapt/fisapt.cc
andyj10224: Addressed

JonathonMisiewicz:
Ping me when this could use another review from me. From tests, it looks like there's a bug in the current code.

Issue #2374, opened 2021-12-03
Support for Yukawa kernel in range-separated integrals


susilehtola:
Now that Psi4 uses LIBINT2, one can enable support for the Yukawa kernel for range-separated functionals from Libxc, such as CAMY-B3LYP.

The code needs to be generalized so that in addition to the error function kernel, `erf_coulomb`, one can also use the Yukawa kernel, `yukawa`, in the LIBINT2 call.

susilehtola:
https://github.com/evaleev/libint/issues/228

PR #2386, opened 2021-12-19
Interfaced Yukawa Libint2 ERI Kernel

andyj10224:
## Description
Per issue #2374, this PR adds an interface to Yukawa ERIs from Libint2.

## Questions
- [x] How would I test my new code? (How are erf kernels tested?)
- [x] Do I need to add functions in MintsHelper for the Yukawa kernel?

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/eri.cc
+++ b/psi4/src/psi4/libmints/eri.cc
@@ -488,7 +488,7 @@ Libint2YukawaERI::Libint2YukawaERI(double zeta, const IntegralFactory *integral,
     } else if (!dummy1 && dummy2 && !dummy3 && dummy4) {
         braket_ = libint2::BraKet::xs_xs;
     } else {
-        throw PSIEXCEPTION("Bad BraKet type in Libint2TwoElectronInt");
+        throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI");
     }
 
     for (int der = 0; der <= deriv; ++der) {
@@ -518,7 +518,7 @@ void Libint2YukawaERI::libint2_wrapper0(const libint2::Shell &sh1, const libint2
             engines_[0].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xs, 0>(sh1, sh2, sh3, sh4, sp12, sp34);
             break;
         default:
-            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2wrapper0");
+            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2_wrapper0");
     }
 }
 
@@ -538,7 +538,7 @@ void Libint2YukawaERI::libint2_wrapper1(const libint2::Shell &sh1, const libint2
             engines_[1].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xs, 1>(sh1, sh2, sh3, sh4, sp12, sp34);
             break;
         default:
-            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2wrapper1");
+            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2_wrapper1");
     }
 }
 void Libint2YukawaERI::libint2_wrapper2(const libint2::Shell &sh1, const libint2::Shell &sh2, const libint2::Shell &sh3,
@@ -557,7 +557,7 @@ void Libint2YukawaERI::libint2_wrapper2(const libint2::Shell &sh1, const libint2
             engines_[2].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xs, 2>(sh1, sh2, sh3, sh4, sp12, sp34);
             break;
         default:
-            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2wrapper2");
+            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2_wrapper2");
     }
 }
 
--- a/psi4/src/psi4/libmints/eri.cc
+++ b/psi4/src/psi4/libmints/eri.cc
@@ -462,3 +462,103 @@ void Libint2ErfComplementERI::libint2_wrapper2(const libint2::Shell &sh1, const 
 }
 
 Libint2ErfComplementERI::~Libint2ErfComplementERI(){};
+
+//// Libint2 implementation
+Libint2YukawaERI::Libint2YukawaERI(double zeta, const IntegralFactory *integral, double screening_threshold, int deriv, 
+                                    bool use_shell_pairs, bool needs_exchange)
+    : Libint2TwoElectronInt(integral, deriv, screening_threshold, use_shell_pairs, needs_exchange) {
+    timer_on("Libint2YukawaERI::Libint2YukawaERI");
+    int max_am =
+        std::max(std::max(basis1()->max_am(), basis2()->max_am()), std::max(basis3()->max_am(), basis4()->max_am()));
+    int max_nprim = std::max(std::max(basis1()->max_nprimitive(), basis2()->max_nprimitive()),
+                             std::max(basis3()->max_nprimitive(), basis4()->max_nprimitive()));
+    const auto max_precision = 0.;  // equivalent in accuracy and timings is `std::numeric_limits<double>::epsilon() * 1e-30;`
+
+    bool dummy1 = basis1()->l2_shell(0) == libint2::Shell::unit();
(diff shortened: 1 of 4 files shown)


susilehtola:
I'm not sure if there are any Gaussian codes that have the Yukawa kernels. However, I do have the Yukawa kernels in HelFEM for atomic calculations, and for `hyb_gga_xc_camy_b3lyp` I get the following energies

| atom | energy |
|--------|-----------|
| He    |     -2.90309239 |
| N      |   -54.59195697 |
| Ne    | -128.96948058 |

If you run the calculations in Psi4 with the AHGBS-9 basis (available from the BSE) and a large quadrature grid like (250, 590), you should be able to match these to sub-microhartree accuracy.

susilehtola:
Of course, to run those tests one also needs to add the new kernels to the range-separated Fock builds and the libxc interface.

psi4/src/psi4/libmints/integral.h
```diff
@@ -499,6 +499,9 @@ class PSI_API IntegralFactory {
     /// Returns an erf complement ERI integral object (omega integral)
     virtual TwoBodyAOInt* erf_complement_eri(double omega, int deriv = 0, bool use_shell_pairs = true, bool needs_exchange = false);
 
+    /// Returns an Yukawa integral object (Slater-type geminal times Coulomb, e^(-z*r)/r)
+    virtual TwoBodyAOInt* yukawa_eri(double zeta, int deriv = 0, bool use_shell_pairs = true, bool needs_exchange = false);
```
JonathonMisiewicz, line 503: Returning a raw pointer is dangerous. Can we possibly make this a `unique_ptr`, or will that create issues elsewhere down the line?

JonathonMisiewicz:
I'll wait for reviews from @jturney and @andysim before commenting further. I don't know this part of the codebase well.

andyj10224:
> Of course, to run those tests one also needs to add the new kernels to the range-separated Fock builds and the libxc interface.

I believe that is outside the scope of this PR, is there I way I can test for numerical correctness? (Possibly expose the functions to python and call a numpy function)?

psi4/src/psi4/libmints/integral.cc
```diff
@@ -259,6 +259,13 @@ TwoBodyAOInt* IntegralFactory::erf_complement_eri(double omega, int deriv, bool
 #endif
 }
 
+TwoBodyAOInt* IntegralFactory::yukawa_eri(double zeta, int deriv, bool use_shell_pairs, bool needs_exchange) {
+    auto integral_package = Process::environment.options.get_str("INTEGRAL_PACKAGE");
```
jturney, line 263: Modify the function to only use libint2 since no other engine supports it.

psi4/src/psi4/libmints/integral.h
jturney: Leaving it as a raw pointer follows the pattern of the other integrals. Another PR could be done that makes the conversion to a unique_ptr.

jturney (changes requested):

psi4/src/psi4/libmints/integral.cc
andyj10224: I currently have the function throw an exception if the integral package is not set to Libint2. Should I just let the Yukawa ERI be created regardless of what the user option is set as?

jturney: Yes. We will always require libint2 to be installed. If another package is added that supports this integral then this ability can be added back.

andyj10224: Fixed this

jturney (approved):

JonathonMisiewicz (approved):
As far as I'm aware, the only other issue is testing, and we don't have a good way to test without massively increasing the scope of the PR. As long as the integrals aren't _exposed_ until we're sure they work, they should be fine. Anybody who would be using the Yukawa integrals assumes the risk.

psi4/src/psi4/libmints/eri.cc
```diff
@@ -462,3 +462,103 @@ void Libint2ErfComplementERI::libint2_wrapper2(const libint2::Shell &sh1, const
...
+            engines_[0].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xx, 0>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        case libint2::BraKet::xx_xs:
+            engines_[0].compute2<libint2::Operator::yukawa, libint2::BraKet::xx_xs, 0>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        case libint2::BraKet::xs_xs:
+            engines_[0].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xs, 0>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        default:
+            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2wrapper0");
```
loriab, line 521:
```suggestion
            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2_wrapper0");
```

psi4/src/psi4/libmints/eri.cc
```diff
@@ -462,3 +462,103 @@ void Libint2ErfComplementERI::libint2_wrapper2(const libint2::Shell &sh1, const
...
+            engines_[1].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xx, 1>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        case libint2::BraKet::xx_xs:
+            engines_[1].compute2<libint2::Operator::yukawa, libint2::BraKet::xx_xs, 1>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        case libint2::BraKet::xs_xs:
+            engines_[1].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xs, 1>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        default:
+            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2wrapper1");
```
loriab, line 541:
```suggestion
            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2_wrapper1");
```

psi4/src/psi4/libmints/eri.cc
```diff
@@ -462,3 +462,103 @@ void Libint2ErfComplementERI::libint2_wrapper2(const libint2::Shell &sh1, const
...
+            engines_[2].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xx, 2>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        case libint2::BraKet::xx_xs:
+            engines_[2].compute2<libint2::Operator::yukawa, libint2::BraKet::xx_xs, 2>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        case libint2::BraKet::xs_xs:
+            engines_[2].compute2<libint2::Operator::yukawa, libint2::BraKet::xs_xs, 2>(sh1, sh2, sh3, sh4, sp12, sp34);
+            break;
+        default:
+            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2wrapper2");
```
loriab, line 560:
```suggestion
            throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI::libint2_wrapper2");
```

psi4/src/psi4/libmints/eri.cc
```diff
@@ -462,3 +462,103 @@ void Libint2ErfComplementERI::libint2_wrapper2(const libint2::Shell &sh1, const
...
+    if (!dummy1 && !dummy2 && !dummy3 && !dummy4) {
+        braket_ = libint2::BraKet::xx_xx;
+    } else if (!dummy1 && dummy2 && !dummy3 && !dummy4) {
+        braket_ = libint2::BraKet::xs_xx;
+    } else if (!dummy1 && !dummy2 && !dummy3 && dummy4) {
+        braket_ = libint2::BraKet::xx_xs;
+    } else if (!dummy1 && dummy2 && !dummy3 && dummy4) {
+        braket_ = libint2::BraKet::xs_xs;
+    } else {
+        throw PSIEXCEPTION("Bad BraKet type in Libint2TwoElectronInt");
```
loriab, line 491:
```suggestion
        throw PSIEXCEPTION("Bad BraKet type in Libint2YukawaERI");
```
questionable, as preceding two-electron ints types do this, too

loriab (approved):
lgtm!

PR #2387, opened 2021-12-24
Add pb11 dependency to modules that use DIIS

andysim:
## Description
When trying to build from source, I had problems with Pybind11 headers not being found.  This is most likely due to #2369 introducing new dependencies on PB11 being introduced that the build system is unaware of.  This PR fixes the problem on my Linux/ICPC setup, but I'm open to suggestions if there's a more up to date way of telling CMake about these new dependencies.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/CMakeLists.txt
+++ b/psi4/src/psi4/cc/CMakeLists.txt
@@ -1,7 +1,5 @@
 psi4_add_module(bin cc "")
 
-target_link_libraries(cc PUBLIC pybind11::headers)
-
 # FIXME Remove once done
 list(APPEND _Clang_debug_flags
   "-Wall"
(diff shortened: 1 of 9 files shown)


JonathonMisiewicz:
I'll leave the cmake analysis to Lori, but the PR title is inaccurate. `cc` and `scfgrad` do not use PyDIIS. Do we need changes to those modules for some less direct reason?

jturney:
They may not directly use PyDIIS but they include a header file that does include pybind; therefore, a fix was needed. 

I would think that instead of adding `target_link_libraries(XXXX PUBLIC pybind11::headers)` to the modules that need it the modules should state their own dependencies. For example, `scfgrad` should have:

```
target_link_libraries(scfgrad PUBLIC scf_solver)

not

target_link_libraries(scfgrad PUBLIC pybind11::headers)
```

andysim:
Excellent call, Jet - I've updated the dependencies to properly reflect how the PB11 deps get introduced.

jturney (approved):
Looks good to me!

maxscheurer (approved):
👍🏻

PeterKraus (approved):

PR #2389, opened 2022-01-02
Fix error in MBE VMFC Hessian

alenaizan:
## Description

Looks like there is a small error in the VMFC Hessian. The `sign` variable is changed for VMFC but it is not used to multiply the Hessian. Compare this with the gradient above and the energy in line 661.

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
Can we have a correctness test on the VMFC hessian? I do not know when this code will run, what problem `_sum_cluster_ptype_data` solves, or what this change does, and an example of _what in the output file this commit changes_ would be very helpful.

I'll remind Lori that the n-body code in DDD needs to be better documented than it is right now.

alenaizan:
I have added a test case similar to the test case for the gradient. I do not know of other codes that have this capability but the test case enumerate all the necessary terms and check for the correct expansion formula for a trimer truncated at the two-body level. Without this PR, the many-body expansion terms for the Hessian are not summed up correctly if the VMFC basis set superposition error scheme is used. The `_sum_cluster_ptype_data` function is supposed to construct the total cluster gradient or Hessian from the fragment computations. Documentation is beyond the scope of this trivial PR.

JonathonMisiewicz:
> I have added a test case similar to the test case for the gradient.

I prefer a simpler test: test VMFC _analytic hessian_ against VMFC hessian by _finite difference of single point_. I don't see a need for this test to test against its own implementation of the VMFC energy formula. That's a job for a different test. Including it in this one obscures the point of the test.

> The `_sum_cluster_ptype_data` function is supposed to construct the total cluster gradient or Hessian from the fragment computations. Documentation is beyond the scope of this trivial PR.

I disagree. It is _always_ within the scope of the PR to get the documentation to such a state that the problem the PR solves is intelligible. I am not asking you to document the entire nbody module. I am asking you to document the expected behavior of the function whose behavior you are changing - that is perfectly in-scope.

`Sums gradient and hessian data from compute_list.` is a poor docstring. "data" could mean anything, and sums them to get what? "Sum gradients/hessians from n-body computations to obtain the BSSE corrected gradient/hessian" is much clearer. If that much isn't clear, then the purpose of this PR is not clear.

alenaizan:
I added the finite difference test and I documented the function.

JonathonMisiewicz (approved):
Thanks!

PeterKraus (approved):

jturney (approved):

PR #2395, opened 2022-01-06
Allow `matrix` Comparison

JonathonMisiewicz:
## Description
Provides and exposes a `matrix` method to check if two matrices are the same, within a given tolerance. This may allow for some simplifying of the test infrastructure, but I don't know the test infrastructure nearly well enough to say.

## Checklist
- [x] `test_matrix.py` passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -482,9 +482,8 @@ void export_mints(py::module& m) {
         .def("clone", &Matrix::clone, "Creates exact copy of the matrix and returns it")
         .def_property("name", py::cpp_function(&Matrix::name), py::cpp_function(&Matrix::set_name),
                       "The name of the Matrix. Used in printing.")
-
-        // def("set_name", &Matrix::set_name, "docstring").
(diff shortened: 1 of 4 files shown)


loriab (commented):
Good to have a c-side method for matrix comparison.

How prevalent is `tol` for that parameter name in libmints? If it's sparse or inconsistent, I'd advocate for `atol` for consistency with numpy (that has absolute and relative tolerance).

For testing results beyond boolean, printing the error is handy. I think one could add a kwarg `check_name` to https://github.com/psi4/psi4/blob/d9093c75c71c2b33fbe86f32b25d138675ac22eb/psi4/driver/p4util/testing.py#L79

JonathonMisiewicz:
I don't understand your comment about `_mergedapis_compare_matrices`. Assume I know nothing about the test infrastructure.

JonathonMisiewicz:
Closing PR. After discussions with Lori, while the code here is fine, it doesn't solve the problem I'm actually interested in. Others are welcome to take up this PR if they wish.

Expect a new PR shortly which does do more for the problem I care about.

PR #2402, opened 2022-01-11
Deprecate inert keyword.

### LINKED ISSUE #2401: `diis_min_vecs` Useless
The `diis_min_vecs` option currently has no effect in Psi4. We [check that is at least 1](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/scf_proc/scf_iterator.py#L402), but we [already raise an error if it isn't](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/scf_proc/scf_iterator.py#L849). I've checked, and it was this way before I started tinkering with DIIS.

This keyword accomplishes absolutely nothing besides giving the user a chance to raise errors. Do we want to _remove the keyword_, _add the intended feature_, or _deprecate and then remove the keyword_?

JonathonMisiewicz:
## Description
Closes #2401 by removing a keyword that does nothing of any use.

One less thing for me to check in an upcoming PR.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -569,6 +569,9 @@ void throw_deprecation_errors(std::string const& key, std::string const& module 
                 ". All instances of 'dcft' should be replaced with 'dct'. The method was renamed in v1.4.",
             __FILE__, __LINE__);
     }
+    if (module == "SCF" && key == "DIIS_MIN_VECS") {
+        py_psi_print_out("WARNING!\n\tRemove keyword DIIS_MIN_VECS! This keyword does nothing. Using it will raise an error in v1.7.\n");
+    }
 }
 
 Options& py_psi_get_options() { return Process::environment.options; }
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -399,7 +399,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
                 Dnorm = self.compute_orbital_gradient(add_to_diis_subspace, core.get_option('SCF', 'DIIS_MAX_VECS'))
 
-                if (add_to_diis_subspace and core.get_option('SCF', 'DIIS_MIN_VECS') - 1):
+                if add_to_diis_subspace:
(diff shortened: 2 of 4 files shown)


susilehtola (approved):
Looks good to me

psi4/src/core.cc
```diff
@@ -569,6 +569,9 @@ void throw_deprecation_errors(std::string const& key, std::string const& module
                 ". All instances of 'dcft' should be replaced with 'dct'. The method was renamed in v1.4.",
             __FILE__, __LINE__);
     }
+    if (module == "SCF" && key == "DIIS_MIN_VECS") {
+        py_psi_print_out("WARNING!\n\tRemove keyword DIIS_MIN_VECS! This keyword does nothing. Using it will raise an error in v1.7.\n");
+    }
```
loriab, line 574: I suspect that every use of diis_min_vecs (even post-scf ones that will remain legit) will hit this error during the scf part unless user specified module with `set dfocc diis_min_vecs`.

I don't have a great solution. Could tone down the warning to `WARNING!\n\tKeyword DIIS_MIN_VECS for the SCF module is deprecated (probably has not worked in a while). Directing it toward the SCF module will raise an error in v1.7.`. That's awkward enough that the read_options-only comment might be better.

JonathonMisiewicz: Incorrect.

My tests say the warning only fires if the user explicitly specifies this as an SCF option. It goes silent if it set as a global option or specific to a non-SCF module.

loriab: oh, that's great!

loriab (approved):

maxscheurer (approved):

hokru (approved):

hokru:
`test_resp_2` fails in the Linux clang build with an einsum error. (unrelated to this PR)

JonathonMisiewicz:
Since I needed to make a new commit to re-trigger tests, I slightly expanded the scope of the PR to remove some `occ` DIIS keywords that were already marked deprecated during the course of the open-heart surgery series, and should have been removed one version ago.

PR #2403, opened 2022-01-13
Fix Side Effect of Option Getting

JonathonMisiewicz:
## Description
Previously, getting an option would change the currently set module. No more.

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
Apparently, this is not a bug but a feature. Expect me to file a documentation issue shortly.

PR #2413, opened 2022-01-25
Update CMake fork of Libint2, c. 2019 -> c. 2022

loriab:
## Description

This corresponds to L2 at https://github.com/evaleev/libint/pull/233 .

## Todos

- [x] switches psi4 to use a Libint2 forked from upstream in 2022, not in 2019
- [x] changes a couple cmake knobs and changes the required components to EFV's preferred syntax (`g5` -> `eri_c4_d1_l5`)
- [x] avoided grid points discrepancies that I had hit by implementing Holger's suggestion.
- [x] I and others have reported needing `pybind11::headers` in extra places, so adding those
- [x] The defines `INT_NCART` and `INT_CARTINDEX` in psi https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/integral.h#L46-L64 are also in L2. They used to be `#define` in the current/2019 fork of L2, but now in L2 master, it’s an inline function. To avoid errors like the below, they're namespaced and inlined here now, too.

```
< namespace libint2 {
< inline int INT_NCART(int am) { return ((am + 2) * (am + 1)) >> 1; }
< }
< LIBINT_DEPRECATED("please use libint2::INT_NCART instead")
< inline int INT_NCART(int am) { return libint2::INT_NCART(am); }
---
> #define INT_NCART(am) ((((am)+2)*((am)+1))>>1)
```
With that situation (define in psi and inlined in L2), I get lots of
```
In file included from /psi/gits/hrw-pybind/psi4/src/psi4/libmints/integral.cc:28:
/psi/gits/libint2-efv/install-p01/include/libint2/cgshell_ordering.h:104:12: error: expected unqualified-id before 'int'
  104 | inline int INT_CARTINDEX(unsigned int am, int i, int j) {
      |            ^~~~~~~~~~~~~
```
I can fix the problem by renaming in psi to `INT_NCART_PSI` and `INT_CARTINDEX_PSI`. What's a cleaner solution than renaming? Should we inline? Or add to the `psi {}` namespace?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge **SQUASH**

### DIFF
--- a/external/upstream/libint2/CMakeLists.txt
+++ b/external/upstream/libint2/CMakeLists.txt
@@ -6,6 +6,7 @@
 #   This is not the default, though it may become runtime selectable in future https://github.com/evaleev/libint/issues/190#issuecomment-691485225 .
 # * In order for any DF to work, need ENABLE_ERI3 and ENABLE_ERI2 >=0
 # * In order for gradient and Hessian tests to not segfault, need ENABLE_ERI and ENABLE_ERI3 and ENABLE_ERI2 =2
+# * (new c. Jan 2022) In order for Psi4 to at all work with a libint2 build for one-electron ints, need ENABLE_ONEBODY =2
 # * The `find_package(Libint2 ... COMPONENTS ...)` commands will check the above requirements met, provided the libint2
 #   was built with cmake+cmake, not libtool+cmake.
 # * Provided you know what you're doing wrt the above options and know what subset of Psi4 calculations you're targeting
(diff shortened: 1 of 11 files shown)


andysim:
I think that renaming is perhaps the safest solution; without namespacing these kinds of macros it's possible to run into a situation where the two codes have the same macro name that returns a different result, which would cause chaos.  If you just want to get rid of the warnings (given that we know that L2 and Psi4 both have the same meaning of `INT_CART`) we could also just ensure that the L2 headers are included before Psi4 defines the macros; the `#if !defined(INT_CART)` will ensure we don't redefine the macro.

andysim:
Oh, and to clarify by "namespacing" I'm referring to adding some kind of `PSI4` to the macro name, rather than adding the define within a namespace block.  To my knowledge the preprocessor wouldn't care about whether the definition occurs withing a C++ namespace or not.

external/upstream/libint2/CMakeLists.txt
```diff
@@ -24,17 +25,16 @@
 # * If ever use a libint2 with AM>7, edit the `init_spherical_harmonics` line in libmints/integral.cc
 # * In selecting AM values, before `grep`ing lots of basis set files, consult the guide at end of this file.
 
-
 # CHOOSE! `make export`
 #find_package(Libint2 CONFIG)
 # CHOOSE! pure cmake
-find_package(Libint2 CONFIG COMPONENTS gss "e${MAX_AM_ERI}" g2 h2 eri3_e4 eri3_g3 eri3_h3 eri2_e4 eri2_g3 eri2_h3)
+find_package(Libint2 CONFIG COMPONENTS gss CXX_ho impure_sh "eri_c4_d0_l${MAX_AM_ERI}" eri_c4_d1_l2 eri_c4_d2_l2 eri_c3_d0_l4 eri_c3_d1_l3 eri_c3_d2_l3 eri_c2_d0_l4 eri_c2_d1_l3 eri_c2_d2_l3 onebody_d0_l4 onebody_d1_l3 onebody_d2_l3)
```
mbanck, line 31: :scream:

loriab: yeah, it's a little verbose, but that's a pretty minimal build AM-wise. it's going to be interesting to find a set of build conditions that all the OS QC packages can be content with.

maxscheurer (approved):

jturney (approved):
Looks good!

hokru (approved):

zachglick (approved):

Issue #2412, opened 2022-01-21
Code Cleanup Megaissue


JonathonMisiewicz:
This issue is a code cleanup wishlist. Please place a code cleanup issue here if you see it in a PR, but fixing it would be outside the scope of the PR.

- [ ] Remove `new` from `lib3index/fittingmetric.cc`. In some cases, `new` is used to initialize... a raw pointer that points to `std::shared_ptr<OneBodyAOInt>`. In 2022, this should be a `std::vector`.

susilehtola:
- [ ] Clean up `matrix.h` and `matrix.cc` to use standard C++ types instead of C-style `malloc`s

JonathonMisiewicz:
- [ ] Purge `init_array` from `libsapt_solver`. This calls a function to create a raw pointer, which later needs to be `free`d. In 2022, it's much better to use `std::vector<double>`. Feel free to break this PR into smaller ones, given how much `libsapt_solver` uses it.

PeterKraus:
- [ ] The documentation generates a lot of warnings about hardcoded links when built, see [here](https://github.com/psi4/psi4/runs/5024225771?check_suite_focus=true#step:9:5287)

JonathonMisiewicz:
- [ ] `ccenergy/rotate.cc` is the only file in all of `ccenergy` that bothers with QT -> Pitzer conversions. Refactor that file using modern libmints tech (`Dimension`, `Slice`, `Matrix`) and then purge any mention of QT vs Pitzer from `ccenergy`
- [ ] `ccenergy` has a wavefunction, so we should be able to remove most of `MOInfo` from there.

JonathonMisiewicz:
- [ ] Rename the `DFTGrid` class to `IntegrationGrid`. As written, it's confusing that non-DFT classes use it.

JonathonMisiewicz:
- [ ] Make `IntVector` use a `Dimension` internally
- [ ] Make `IntVector` accept `Dimension`, no raw pointers
- [ ] Get rid of `IntVector`'s init methods

JonathonMisiewicz:
`detci/ciwave.cc` probably should use Dimension objects in place of arrays of int. `get_block` and `set_block` may also remove a few functions.

TiborGY:
- [ ] Investigate deprecating PSI_API functions rsp() and schmidt_add() if PR #2776 is merged
- [ ] Investigate removing normalize.cc and other functions if the above is merged

TiborGY:
[...]

PR #2417, opened 2022-01-31
Cleaned fittingmetric new/delete

philipmnel:
## Description
This PR replaces `new`/`delete` in `lib3index/fittingmetric.cc` with `std::vector` per task in #2412  

## Checklist
- [x] Quick tests pass.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/lib3index/fittingmetric.cc
+++ b/psi4/src/psi4/lib3index/fittingmetric.cc
@@ -114,8 +114,8 @@ void FittingMetric::form_fitting_metric() {
 
     // == (A|B) Block == //
     IntegralFactory rifactory_J(aux_, zero, aux_, zero);
-    const auto** Jbuffer = new const double*[nthread];
-    std::shared_ptr<TwoBodyAOInt>* Jint = new std::shared_ptr<TwoBodyAOInt>[ nthread ];
+    std::vector<const double*> Jbuffer(nthread);  
+    std::vector<std::shared_ptr<TwoBodyAOInt>> Jint(nthread);
     for (int Q = 0; Q < nthread; Q++) {
         if (omega_ > 0.0) {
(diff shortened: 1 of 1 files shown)


psi4/src/psi4/lib3index/fittingmetric.cc
```diff
@@ -387,13 +380,12 @@ void FittingMetric::form_QR_inverse(double tol) {
...
         // First, find out how much workspace to provide
-        C_DORGQR(n, n, n, Qp[0], n, tau, &work_size, -1);
+        C_DORGQR(n, n, n, Qp[0], n, tau.data(), &work_size, -1);
 
         // Now, form Q
         lwork = (int)work_size;
-        work = new double[lwork];
-        C_DORGQR(n, n, n, Qp[0], n, tau, work, lwork);
-        delete[] work;
+        work.resize(lwork);
```
JonathonMisiewicz, line 387: `work` is already of size `lwork`, so this line isn't necessary.

JonathonMisiewicz:
Thanks for taking this on! There's one unnecessary line of code here, but after that, this looks good to me.

psi4/src/psi4/lib3index/fittingmetric.cc
philipmnel: `lwork` gets resized from `C_DORGQR(n, n, n, Qp[0], n, tau.data(), &work_size, -1);` if I'm not mistaken.

JonathonMisiewicz: That's right. In that case, please update the comment above the line you quoted to explicitly say it writes the optimal length into `work_size`. I would not have guessed that without checking to see what exactly that function does.

JonathonMisiewicz (approved):

jturney (approved):
Nice cleanup!

zachglick (approved):
LGTM!

PR #2418, opened 2022-02-01
Silence sphinx warnings

JonathonMisiewicz:
## Description
This is an _attempt_ to get the docs working again by resolving "Undocumented input" warnings from Sphinx. Worst case scenario, Sphinx outputs are a little shorter. Best case scenario, there was some change in Sphinx that causes crashes when this warning occurs, and we can get docs working again.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/tests/frac-sym/input.dat
+++ b/tests/frac-sym/input.dat
@@ -1,3 +1,5 @@
+#! Fractional occupation with symmetry
+
 molecule {
 0 2
 F
--- a/tests/fsapt-ext-abc/input.dat
+++ b/tests/fsapt-ext-abc/input.dat
@@ -1,3 +1,5 @@
+#! FSAPT with external charge on trimer
+
 memory 1 GB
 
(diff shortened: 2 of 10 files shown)


loriab (approved):
The "undocumented input" is a complaint from a perl script, not from Sphinx itself. But this is good cleanup in its own right.

PeterKraus (approved):
I was just gonna suggest that we should add resolving all those warnings in Sphinx into #2412. Perhaps the hard-coded hyperlinks (which this PR doesn't fix) can be still mentioned there. LGTM.

JonathonMisiewicz:
> I was just gonna suggest that we should add resolving all those warnings in Sphinx into #2412. Perhaps the hard-coded hyperlinks (which this PR doesn't fix) can be still mentioned there. LGTM.

Sounds good to me. I intended for other developers to be add items to the issue, so go right ahead.

PR #2429, opened 2022-02-08
avoid macro rename vs l2 in integral.h

loriab:
## Description

This is picked out of #2413 as independent. Depending on the version of l2, this'll either issue warnings (psi4 l2 fork where `INT_NCART` is still macro) or errors (l2 master where `INT_NCART` now inline fn) when building psi4.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/integral.h
+++ b/psi4/src/psi4/libmints/integral.h
@@ -45,28 +45,28 @@ namespace psi {
 /*! \def INT_NCART(am)
     Gives the number of cartesian functions for an angular momentum.
 */
-inline int INT_NCART(unsigned int am) {
+inline int INT_NCART(int am) {
     return (am >= 0) ? ((((am) + 2) * ((am) + 1)) >> 1) : 0;
 }
 
 /*! \def INT_PURE(am)
     Gives the number of spherical functions for an angular momentum.
 */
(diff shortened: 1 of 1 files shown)


maxscheurer (approved):

jturney (approved):

loriab:
I'm not sure what's wrong with this. It works fine locally (gcc 9) with new l2. With old l2 I get other unrelated build errors. pausing investigation.

loriab:
@maxscheurer reasoned this one out. Existing psi fork L2 still has `INT_NCART` etc. as macro, so that gets stuffed into the new `inline int` _function_ that this PR introduces. confusion ensues. Remedies:
1. close this PR and absorb the changes with #2413 when we move to L2 with namespaced and inlined INT_NCARTs
2. do some sort of suffixing of psi macros in integral.h and the couple dozen places that call them. this will allow psi to be build with a swath of L2 versions.

Unless there's support for 2., I'll close this in a couple days and proceed with 1.

JonathonMisiewicz:
We want to kill the L2 fork anyways, so let's remove the temptation to support multiple L2 and go with (1).

Thanks for your efforts, Lori and Max. I'll approve once tests pass.

PR #2434, opened 2022-02-10
Change `ccdensity` Keywords Around OPDM Relaxation

### LINKED ISSUE #2432: Can't Get CC Relaxed Density
For testing purposes, I want a CCSD OPDM including orbital relaxation contributions.

This script works:
```
molecule {    
O
H 1 1.0 
H 1 1.0 2 101.5
}

wfn = gradient('ccsd/cc-pvdz', return_wfn=True)[1]

oeprop(wfn, 'DIPOLE')
```

...and I can confirm it's correct by doing a finite difference test.

My problem is that _I have absolutely no idea how to do this without going through gradient_. If I naively go through properties, the OPDM relaxation terms get dropped because they are expensive. But if I explicitly order OPDM relaxation terms...

```
molecule {    
O
H 1 1.0 
H 1 1.0 2 101.5
}

set opdm_relax true

properties('ccsd/cc-pvdz')
```

The `cc` module rebels and informs me that it is `Turning orbital relaxation off since only onepdm is requested.` (TPDM is needed to construct the CC orbital gradient needed for the Z-vector equation.)

Who told `cc` to only compute the OPDM? [This line of driver code, which the user cannot override.](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L3124) Although [docs report](https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L1812) that this flag controls whether the onepdm is computed, internal `ccdensity` [says](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/ccdensity/Params.h#L49) the flag controls whether the 1PDM _alone_ is computed, implicitly denying the 2PDM from ever being computed.

Pinging @loriab for all things keyword standardization and @lothian in case you had a reason for [coding this](https://github.com/psi4/psi4archive/commit/575ec501ee7638cf4ff0c373f78e03ae6aac00c4).

In my view:
- Keyword `onepdm` should be renamed to something like `onepdm_only`.
- `onepdm_only` default changes to `true`.
- Requesting either `relax_opdm` or a method that inherently requires 2-electron densities (gradients for sure) trumps `onepdm_only`.
- In all other cases, the driver is no longer responsible for setting `onepdm_only`.
- Default for `relax_opdm` remains false, and therefore, 2PDM will not be computed unless necessary.

If we're agreed, I'm happy to make these changes. I'm more afraid of the current state of densities in Psi than I am of legacy code. We would need to discuss how we want to handle deprecation, and which CC properties need the 2PDM.

JonathonMisiewicz:
## Description
Closes #2432.

This is PR 1 in an ongoing series to make `ccdensity` compatible with the standard `Matrix` and `Wavefunction` machinery used elsewhere in Psi. (That means I'll have a lot of PRs for you two to review. Let me know if there's another person from VA Tech you want to take the responsibility of reviewing my PRs.)

## Todos

- [x] Adds finite difference test on CCSD dipoles
- [x] Standardizes `onepdm` vs `opdm` in `ccdensity` keywords.
- [x] Corrects incorrect description of the `onepdm`/`opdm` keyword
- [x] Creates new `ccdensity` function to convert `** double` to `Matrix`
- [x] Adds `Ca` matrix to parameters, to begin phasing out the QT ordered `** double` representation of the orbitals    

## Checklist
- [x] Passes `cc` tests
- [x] Passes finite difference test on CCSD dipoles

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/ccdensity/get_params.cc
+++ b/psi4/src/psi4/cc/ccdensity/get_params.cc
@@ -67,7 +67,7 @@ void get_params(Options& options) {
     if (options["ONEPDM_GRID_DUMP"].has_changed()) {
         outfile->Printf("\tWarning! ONEPDM_GRID_DUMP is deprecated and will be removed in 1.7. Use OPDM_GRID_DUMP instead.");
         if (not options["OPDM_GRID_DUMP"].has_changed()) {
-            params.onepdm = options.get_bool("ONEPDM_GRID_DUMP");
+            params.onepdm_grid_dump = options.get_bool("ONEPDM_GRID_DUMP");
         }
     }
 
--- a/psi4/src/psi4/cc/ccdensity/get_params.cc
+++ b/psi4/src/psi4/cc/ccdensity/get_params.cc
@@ -51,24 +51,25 @@ namespace psi {
 namespace ccdensity {
 
 void get_params(Options& options) {
-    int errcod, tol;
-    std::string junk;
-
     params.wfn = options.get_str("WFN");
-
-    // junk = options.get_str("REFERENCE");
-    // if(junk == "RHF") params.ref = 0;
-    // else if(junk == "ROHF") params.ref = 1;
-    // else if(junk == "UHF") params.ref = 2;
-    // else {
-    //  printf("Invalid value of input keyword REFERENCE: %s\n", junk.c_str());
-    //  throw PsiException("ccdensity: error", __FILE__, __LINE__);
-    //}
 
     psio_read_entry(PSIF_CC_INFO, "Reference Wavefunction", (char*)&(params.ref), sizeof(int));
 
-    params.onepdm = options.get_bool("ONEPDM");
-    params.onepdm_grid_dump = options.get_bool("ONEPDM_GRID_DUMP");
+    params.onepdm = options.get_bool("OPDM_ONLY");
+    if (options["ONEPDM"].has_changed()) {
+        outfile->Printf("\tWarning! ONEPDM is deprecated and will be removed in 1.7. Use OPDM_ONLY instead.");
+        if (not options["OPDM_ONLY"].has_changed()) {
+            params.onepdm = options.get_bool("ONEPDM");
+        }
+    }
+
+    params.onepdm_grid_dump = options.get_bool("OPDM_GRID_DUMP");
+    if (options["ONEPDM_GRID_DUMP"].has_changed()) {
+        outfile->Printf("\tWarning! ONEPDM_GRID_DUMP is deprecated and will be removed in 1.7. Use OPDM_GRID_DUMP instead.");
+        if (not options["OPDM_GRID_DUMP"].has_changed()) {
+            params.onepdm = options.get_bool("ONEPDM_GRID_DUMP");
+        }
+    }
 
     params.calc_xi = options.get_bool("XI");
     if (params.calc_xi) {
@@ -108,7 +109,7 @@ void get_params(Options& options) {
 
     /*** determine DERTYPE from input */
     params.dertype = 0;
-    junk = options.get_str("DERTYPE");
+    std::string junk = options.get_str("DERTYPE");
     if (junk == "NONE")
         params.dertype = 0;
     else if (junk == "FIRST")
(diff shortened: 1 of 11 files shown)


psi4/src/read_options.cc
```diff
@@ -1809,14 +1809,22 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_bool("XI", false);
         /*- Do use zeta?  -*/
         options.add_bool("ZETA", false);
-        /*- Do compute one-particle density matrix? -*/
+        /*- Deprecated and will be removed in 1.7. Use OPDM_ONLY. -*/
         options.add_bool("ONEPDM", false);
-        /*- Write one-particle density matrix on a grid to file opdm.dx -*/
+        /*- For internal use only! Compute the one-particle density matrix, but not the two-particle density matrix. -*/
+        options.add_bool("OPDM_ONLY", false);
```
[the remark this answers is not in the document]
JonathonMisiewicz: I could, but the way `cc` is structured, this keyword won't have the same meaning. This keyword doesn't control if the OPDM is computed: the OPDM is computed if and only if `ccdensity` is entered, and that's controlled by the driver. Rather, the keyword controls if the TPDM is computed.

If we want consistent keywords between `detci` and `cc`, we need to rename the keyword from "ONEPDM" to "TPDM" and suffer the cost that this is going to be very confusing to anybody paying attention to these keywords.

What do you want me to do?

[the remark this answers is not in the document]
loriab: OPDM_ONLY it is, then, imo. There's a limit to how far CC can be pushed to act like other methods. Good that OPDM_ONLY is an internal keyword (maybe add !expert). Thanks for the explanation.

psi4/src/read_options.cc
```diff
@@ -1809,14 +1809,22 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_bool("XI", false);
         /*- Do use zeta?  -*/
         options.add_bool("ZETA", false);
-        /*- Do compute one-particle density matrix? -*/
+        /*- Deprecated and will be removed in 1.7. Use OPDM_ONLY. -*/
         options.add_bool("ONEPDM", false);
-        /*- Write one-particle density matrix on a grid to file opdm.dx -*/
+        /*- For internal use only! Compute the one-particle density matrix, but not the two-particle density matrix. -*/
```
JonathonMisiewicz, line 1814:
```suggestion
        /*- For internal use only! Compute the one-particle density matrix, but not the two-particle density matrix. !expert -*/
```

psi4/src/read_options.cc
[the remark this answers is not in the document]
JonathonMisiewicz: Added.

lothian (approved):
Looks good.

jturney (approved):

psi4/src/psi4/cc/ccdensity/get_params.cc
```diff
@@ -51,24 +51,25 @@ namespace psi {
...
+        if (not options["OPDM_ONLY"].has_changed()) {
+            params.onepdm = options.get_bool("ONEPDM");
+        }
+    }
+
+    params.onepdm_grid_dump = options.get_bool("OPDM_GRID_DUMP");
+    if (options["ONEPDM_GRID_DUMP"].has_changed()) {
+        outfile->Printf("\tWarning! ONEPDM_GRID_DUMP is deprecated and will be removed in 1.7. Use OPDM_GRID_DUMP instead.");
+        if (not options["OPDM_GRID_DUMP"].has_changed()) {
+            params.onepdm = options.get_bool("ONEPDM_GRID_DUMP");
```
loriab, line 70: `params.onepdm_grid_dump` here?

JonathonMisiewicz, line 70:
```suggestion
            params.onepdm_grid_dump = options.get_bool("ONEPDM_GRID_DUMP");
```

JonathonMisiewicz: Good catch.

My ability to write such stupid mistakes is exactly why I write small PRs if I can get away with it.

PR #2443, opened 2022-02-16
Add LR Overlap Psivar

JonathonMisiewicz:
## Description
This PR adds a new psivariable, the overlap between the left and right CC eigenvectors. This was previously tested by `psitest.pl`, but promoting it to a psivar allows the standard `compare_values` machinery to test it. This PR integrates the new psivar into tests of `cc18`.

This PR also removes the file `overlap_LAMPS.cc`, which was unused and differs from `overlap.cc` just by a search-replace on the name of the file to which quantities were being written.

This is PR 4 in an ongoing series to make `ccdensity` compatible with the standard `Matrix` and `Wavefunction` machinery used elsewhere in Psi. **Do not merge until** we have approval from @loriab and @lothian, as we are adding a new psivar to the `cc` mega-module. The next stage will be to ensure all variables are set _on the wavefunction_, for testing purposes. This will proceed module-by-module, to minimize the chances of anything getting overlooked.

## Questions
- [x] Are there any Crawford group projects that need `overlap_LAMPS.cc`?
- [x] Please double-check that my assignments of method to ground state name are correct.

## Checklist
- [x] Tests added for any newly testable features
- [x] `cc` tests pass

## Status
- [x] Ready for review
- [ ] Ready for merge


lothian:
- No, we don't need to keep `overlap_LAMPS.cc`.
- Given that we don't have any well-defined EOM-type methods for CCSD(aT) or CCSD(T), you shouldn't need to check for these in the ground-state types.  All others look fine.

JonathonMisiewicz:
Thanks for the fast response!

Certainly, there's no EOM_CCSD_T or EOM_CCSD_aT case, but tests indicate that [the code](https://github.com/psi4/psi4/pull/2443/files#diff-2f1ed4b75aa797473dd489be513586a6cf24e9a6e22327517476beed9313c1f0R298-R313) fires even in single-state calculations, so the `CCSD_T` and `CCSD_aT` checks are necessary.

lothian:
Yes, now I see.  I agree completely.

I would probably separate out CCSD(aT) similarly to CCSD(T), since those are closely related.

JonathonMisiewicz:
Good point. I've added a _comment_ explaining why asymmetric triples are marked as having CCSD lambda equations. I had to double-check that myself when coding this.

psi4/src/psi4/cc/cclambda/cclambda.cc
```diff
@@ -296,7 +295,25 @@ double CCLambdaWavefunction::compute_energy() {
...
+                // In the CCSD_AT case, we solve the CCSD lambda equations so we can use them for a triples
+                // correction. For CCSD_AT analytic gradient theory, you would need to solve response equations
+                // including that triples correction, and only then would gs_name be CCSD_AT.
+                gs_name = "CCSD";
+            } else if (params.wfn == "CCSD_T") {
+                gs_name = "CCSD(T)";
+            } else {
+                throw PsiException("cclambda: unknown wfn", __FILE__, __LINE__);
+            }
+            reference_wavefunction_->set_scalar_variable("Left-Right " + gs_name + " Eigenvector Overlap", LR_overlap);
```
loriab, line 315: Use all caps for var names for consistency, please. Lowercase are usually for placeholders like https://psicode.org/psi4manual/master/glossary_psivariables.html#psivar-CC-ROOT-n-CORRELATION-ENERGY

Also, non-all-caps likely messes up case insensitivity of `wfn.variable(SCF total energy)` lookup.

loriab (approved):
after the caps, lgtm!

psi4/src/psi4/cc/cclambda/cclambda.cc
loriab: oh, and add an entry in glossary.rst, please. I think the docs build will complain without.

psi4/src/psi4/cc/cclambda/overlap.cc
```diff
@@ -32,6 +32,7 @@
 */
 #include <cstdio>
 #include "psi4/libdpd/dpd.h"
+#include "psi4/libmints/wavefunction.h"
```
PeterKraus, line 35: It's not obvious to me why the header is needed, given the only diff in the file is that the `overlap` function now returns a `double`.

psi4/src/psi4/cc/cclambda/cclambda.cc
```diff
@@ -296,7 +295,25 @@ double CCLambdaWavefunction::compute_energy() {
             exit_io();
             throw PsiException("cclambda: error", __FILE__, __LINE__);
         }
-        if (pL_params[i].ground) overlap(pL_params[i].irrep);
+        if (pL_params[i].ground) {
+            auto LR_overlap = overlap(pL_params[i].irrep);
+            std::string gs_name; // What theory's lambda equations did we just solve?
```
PeterKraus, line 300: `What -> Which`

psi4/src/psi4/cc/cclambda/overlap.cc
JonathonMisiewicz: Artifact of a previous draft that I forgot to remove.

maxscheurer (approved):

JonathonMisiewicz:
The Crawdad has signed off on the parts that I asked him about, so I think we're good to merge.

PR #2445, opened 2022-02-17
Loosen `adiis` BFGS Solution

JonathonMisiewicz:
## Description
During review of the ADIIS PR, it was suggested that Psi raise an error if the BFGS doesn't meet the target convergence criteria. This PR changes that, so the error will be raised _if BFGS is very far from meeting the convergence criteria_. 1e-6 would be nice, but 1e-3 is good enough for our purposes.

@loriab, these are the last of the Psi tests on my list to heal.

## Checklist
- [x] Failing tests Lori pointed out work; Can't cause tests that previously passed to fail

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/diis.py
+++ b/psi4/driver/procrouting/diis.py
@@ -257,7 +257,7 @@ class DIIS:
         result = minimize(self.adiis_energy, np.ones(len(self.stored_vectors)), method="BFGS",
                 jac = self.adiis_gradient, tol=1e-6, options={"maxiter": 200})
 
-        if not result.success:
+        if np.linalg.norm(result.jac) > 1e-3: # Even if we didn't hit the tolerance, it may be good enough.
             raise Exception("ADIIS minimization failed. File a bug, and include your entire input and output files.")
 
         return normalize_input(result.x)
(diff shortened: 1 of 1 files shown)


PeterKraus:
Have you thought about emitting a warning for cases which converge to within `1e-3` but not within `1e-6`? Also, if `1e-3` is deemed good enough to go on, why bother converging to `1e-6` in the first place?

JonathonMisiewicz:
> Have you thought about emitting a warning for cases which converge to within `1e-3` but not within `1e-6`?

In my experience, it's not worth warning about.

> Also, if `1e-3` is deemed good enough to go on, why bother converging to `1e-6` in the first place?

My reasons are primarily aesthetic. This is a very cheap optimization, so while there's no practical benefit, there is also no practical cost.

maxscheurer (approved):

loriab (approved):

PeterKraus (approved):

JonathonMisiewicz:
Because I don't know when to quit, expect another PR once #2446 comes in with a more elegant solution to this problem.

PR #2462, opened 2022-03-06
Change TDSCF Variable Names

JonathonMisiewicz:
## Description
TDSCF/ADC and EOM can't agree on how to name variables for excitation energies. Per discussion between Lori and I, this moves the TDSCF variable names to the new standard, as I see it. This PR is not complete and is submitted for feedback from the excited-state gurus, who are likely to have strong opinions: @loriab @robertodr @lothian @maxscheurer 

To demonstrate the changes, consider the example file I added. Old code would write the second transition as "TD-HF ROOT 0 -> ROOT 2 EXCITATION ENERGY - B2 SYMMETRY". The new code replaces that with three new psivars:

- TD-HF ROOT 0 (B1) -> ROOT 0 (B2) EXCITATION ENERGY
- TD-HF ROOT 0 -> ROOT 2 EXCITATION ENERGY
- TD-HF ROOT 0 -> ROOT 2 EXCITATION ENERGY - A2 SYMMETRY

We have one access call that requires the symmetries of initial and target states and the indices of both _within their irreps_. We have one access call that requires no symmetry information and just requires the index of the states _among all irreps_. The last access call is like the second, but adds on the symmetry of the _transition_. **This is a breaking change** because old code instead used the symmetry of the final state.

After coding this all up, I'm inclined to remove the last access call. The original reason for keeping it up was consistency with the current way TDSCF operates, but it's still inconsistent, per the last paragraph.

I'll add docs once we're settled about the new naming conventions. Once this PR is in, the other modules will need to update to the new style.

## Questions
- [x] Opinions on the new variable names?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/glossary_psivariables.rst
+++ b/doc/sphinxman/source/glossary_psivariables.rst
@@ -1697,7 +1697,7 @@ PSI Variables by Alpha
 
 .. psivar:: TD-fctl ROOT 0 -> ROOT m EXCITATION ENERGY
    TD-fctl ROOT 0 (h) -> ROOT m (i) EXCITATION ENERGY
-   TD-fctl ROOT 0 -> ROOT m EXCITATION ENERGY - h SYMMETRY
+   TD-fctl ROOT 0 -> ROOT m EXCITATION ENERGY - h TRANSITION
 
    The excitation energy of a given method from ground state to root m.
    DFT functional labeled if canonical.
@@ -1708,7 +1708,7 @@ PSI Variables by Alpha
 
 .. psivar:: TD-fctl ROOT m TOTAL ENERGY
    TD-fctl ROOT m (h) TOTAL ENERGY
-   TD-fctl ROOT m TOTAL ENERGY - h SYMMETRY
+   TD-fctl ROOT m TOTAL ENERGY - h TRANSITION
 
    The total energy of given method from ground state to root m in h symmetry.
    Conventions for root indexing and whether h refers to transition or root
@@ -1716,10 +1716,10 @@ PSI Variables by Alpha
 
 .. psivar:: TD-fctl ROOT 0 -> ROOT m ELECTRIC TRANSITION DIPOLE MOMENT (LEN)
    TD-fctl ROOT 0 (h) -> ROOT m (i) ELECTRIC TRANSITION DIPOLE MOMENT (LEN)
-   TD-fctl ROOT 0 -> ROOT m ELECTRIC TRANSITION DIPOLE MOMENT (LEN) - h SYMMETRY
+   TD-fctl ROOT 0 -> ROOT m ELECTRIC TRANSITION DIPOLE MOMENT (LEN) - h TRANSITION
    TD-fctl ROOT 0 -> ROOT m ELECTRIC TRANSITION DIPOLE MOMENT (VEL)
    TD-fctl ROOT 0 (h) -> ROOT m (i) ELECTRIC TRANSITION DIPOLE MOMENT (VEL)
-   TD-fctl ROOT 0 -> ROOT m ELECTRIC TRANSITION DIPOLE MOMENT (VEL) - h SYMMETRY
+   TD-fctl ROOT 0 -> ROOT m ELECTRIC TRANSITION DIPOLE MOMENT (VEL) - h TRANSITION
 
    The electric transition dipole moment in length or velocity gauge of named method
    from ground state to root m in h symmetry (if available). DFT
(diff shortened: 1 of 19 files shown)


maxscheurer:
> After coding this all up, I'm inclined to remove the last access call.

I'd definitely retain the access call with the symmetry of the transition. How about `TD-HF ROOT 0 -> ROOT 2 EXCITATION ENERGY - A2 TRANSITION` to make the distinction to the "legacy" version more clear?

Even though `adcc` does not support point group symmetry, ping @mfherbst for another opinion on excited states 😄

lothian:
I am fine with all three access calls, and I agree with Max regarding the clarification of the third one.

JonathonMisiewicz:
And, edited with Max's suggestion. Should be ready for review.

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -97,6 +97,19 @@ PSI Variables by Alpha
    The total electronic energy [Eh] and correlation energy component [Eh]
    for the averaged coupled-pair functional level of theory.
 
+.. psivar:: ADC ROOT 0 -> ROOT m EXCITATION ENERGY - h SYMMETRY
```
loriab, line 100: I think this should be `ADC ROOT 0 (h) -> ROOT m (i) EXCITATION ENERGY` even if adcc can't do symmetry, and h & i are always A.

But perhaps these additions were simply to keep the qcvars around when the multi-method blocks below split up?

robertodr (approved):

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -1682,39 +1695,92 @@ PSI Variables by Alpha
...
+   TD-fctl ROOT 0 -> ROOT m EXCITATION ENERGY - h TRANSITION
 
-   The excitation energy of given method from ground state to root m
-   in h symmetry (if available). DFT functional labeled if canonical.
+   The excitation energy of a given method from ground state to root m.
+   DFT functional labeled if canonical.
+   m in the first and third variables is the index of the root among all roots.
+   m in the second variable is the index of the root among all roots of irrep i.
+   h, i in the second variable are the irreps of the roots.
+   h in the third variable is the irrep of the transition between roots.
```
loriab, line 1707: I'm thinking for energies at least (as opposed to properties: dipole, oscillator, rotary, etc.) these entries might be easier to read if only the method was changing. For example, adc/tddft/eomcc no-sym excitation as one glossary entry. adc/tddft/eom per-sym excitation as another, and transition-sym as another. If you agree but don't want to rearrange it all now, I can do a pass at the end of the qcvars series. Glad of other opinions.

robertodr:
Good cleanup, thanks. If I recall correctly, AJ and I decided the TDSCF naming based on DALTON, with no regard for whatever conventions existed already in Psi4.

loriab (approved):

doc/sphinxman/source/glossary_psivariables.rst
JonathonMisiewicz: Exactly, the qcvars page reflects variables as they are, not variables as they'll be in a few PRs from now.

doc/sphinxman/source/glossary_psivariables.rst
JonathonMisiewicz: There are advantages either way, and I have no strong opinions on which of the two we take.

But if we shift to your suggestion, I have two requests:
* The "see X for indexing convention" line changes to "see the nth access call in `Notes on Psivars`" or whatever the appropriate sphinx-ification is.
* If we break up one variable like that, we break up _all_ variables like that.

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -97,6 +97,19 @@ PSI Variables by Alpha
    The total electronic energy [Eh] and correlation energy component [Eh]
    for the averaged coupled-pair functional level of theory.
 
+.. psivar:: ADC ROOT 0 -> ROOT m EXCITATION ENERGY - h SYMMETRY
+
+   The excitation energy of given method from ground state to root m
+   in h symmetry (if available). DFT functional labeled if canonical.
```
maxscheurer, line 103: Could you remove the "DFT functional" part of the comment? 😊

maxscheurer (approved):

JonathonMisiewicz:
@loriab After attempting to implement your suggestion, I find it looks uglier. Might I suggest that we wait until ADC and EOM are hooked up, and _then_ discuss which way of formatting these variables is best? Once we have more options, my opinion may change.

loriab:
> @loriab After attempting to implement your suggestion, I find it looks uglier. Might I suggest that we wait until ADC and EOM are hooked up, and then discuss which way of formatting these variables is best? Once we have more options, my opinion may change.

sure, sounds good.

PR #2464, opened 2022-03-09
Reduce code triplication in Molecule::atom_at_position

TiborGY:
## Description
Some code triplication may be eliminated.

## Todos

-  [X] Yay for negative LOC

## Checklist
- [X] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/molecule.cc
+++ b/psi4/src/psi4/libmints/molecule.cc
@@ -360,12 +360,7 @@ std::string Molecule::fsymbol(int atom) const { return full_atoms_[atom]->symbol
 std::string Molecule::label(int atom) const { return atoms_[atom]->label(); }
 
 int Molecule::atom_at_position1(double *coord, double tol) const {
-    Vector3 b(coord);
-    for (int i = 0; i < natom(); ++i) {
-        Vector3 a = xyz(i);
-        if (b.distance(a) < tol) return i;
-    }
-    return -1;
+    return atom_at_position2(Vector3(coord), tol);
 }
 
 int Molecule::atom_at_position2(Vector3 &b, double tol) const {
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz (approved):
FWIW, it would be better to have all these functions simply be `atom_at_position`, but that can be for another day.

JonathonMisiewicz:
> > FWIW, it would be better to have all these functions simply be `atom_at_position`, but that can be for another day.
> 
> Indeed, but I think they are used at a couple of places. My first instinct was to turn it into a function template, but I think these fns are exported to Pyside via pybind right? Not sure how that would handle a C++ template.

Templating is overcomplicated. Change the function names where they are called, C-side. In pybind, you specify which function you want to export with a typedef. For example, here is how we specify which of the three signatures for `Localizer::build` to export.
```
    typedef std::shared_ptr<Localizer> (*localizer_with_type)(const std::string&, std::shared_ptr<BasisSet>,
                                                              std::shared_ptr<Matrix>);

    py::class_<Localizer, std::shared_ptr<Localizer>>(m, "Localizer",
                                                      "Class containing orbital localization procedures")
        .def_static("build", localizer_with_type(&Localizer::build), "Build the localization scheme");
```

JonathonMisiewicz:
Making the first argument to `atom_at_position2` const should resolve the build fail. C is rightly complaining that modifying a temporary is not a sane thing to do.

maxscheurer:
One could also use `py::overload_cast` instead of typedefs 😊

JonathonMisiewicz:
We can merge this in once you say it's ready for merge.

TiborGY:
> We can merge this in once you say it's ready for merge.

This one is good to go.
I think I found a corner case in the current code that may be worth discussing, but that will be a new PR.

PR #2465, opened 2022-03-09
Fix corner case in the C++ implementation of Molecule::atom_at_position

TiborGY:
## Description

`Molecule::atom_at_position` is supposed to return the index of the atom found in a specified radius of a specified point in 3D space, or -1 if no atom is found in the radius. The current implementation assumes there is never more than one atom within the given radius, and returns the index of the first atom found to be within the tolerance.
This can hypothetically happen in two ways: the atoms are either extremely close, or the tolerance is set to a large value.

The new implementation behaves more like one would expect it to behave: under the aforementioned extreme circumstances it throws an exception and informs the user with an error message.

## Todos

- [x] Detect corner case and throw error if encountered
- [x] Update docstrings to better describe implementation

## Questions
- [x] Apparently, `qcdb.Molecule` has its own independent Python implementation of `atom_at_position`. As is, this PR means that calling `qcdb.Molecule.atom_at_position` and `psi4.core.Molecule.atom_at_position` from a Python program may return different results for the same geometry, if the aforementioned corner case happens.
Not sure what to do about this, at first glance it would be neater if both `qcdb.Molecule.atom_at_position` and `psi4.core.Molecule.atom_at_position` ended up calling the same C++ implementation, but I am not sure how feasible that would be.
Thoughts?

## Checklist
- [x] No new features
- [x] `ctest -j24` and `make pytest` both passed locally

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -924,10 +924,10 @@ void export_mints(py::module& m) {
              "Gets the multiplicity of each fragment")
         .def("atom_at_position", &Molecule::atom_at_position1,
              "Returns the index of the atom inside *tol* radius around *coord*. Returns -1 for no atoms, "
-             "returns the index of the closest atom if more than one is found.", "coord"_a, "tol"_a)
+             "throws an exception if more than one is found.", "coord"_a, "tol"_a)
         .def("atom_at_position", &Molecule::atom_at_position3,
              "Returns the index of the atom inside *tol* radius around *coord*. Returns -1 for no atoms, "
-             "returns the index of the closest atom if more than one is found.", "coord"_a, "tol"_a)
+             "throws an exception if more than one is found.", "coord"_a, "tol"_a)
         .def("print_out", &Molecule::print, "Prints the molecule in Cartesians in input units to output file")
         .def("print_out_in_bohr", &Molecule::print_in_bohr, "Prints the molecule in Cartesians in Bohr to output file")
         .def("print_out_in_angstrom", &Molecule::print_in_angstrom,
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -923,9 +923,11 @@ void export_mints(py::module& m) {
         .def("get_fragment_multiplicities", &Molecule::get_fragment_multiplicities,
              "Gets the multiplicity of each fragment")
         .def("atom_at_position", &Molecule::atom_at_position1,
-             "Tests to see if an atom is at the position *coord* with a given tolerance *tol*", "coord"_a, "tol"_a)
+             "Returns the index of the atom inside *tol* radius around *coord*. Returns -1 for no atoms, "
+             "returns the index of the closest atom if more than one is found.", "coord"_a, "tol"_a)
         .def("atom_at_position", &Molecule::atom_at_position3,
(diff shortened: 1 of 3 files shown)


JonathonMisiewicz:
It shouldn't "print a stern warning." It should raise an error. You should never ever have two atoms that are at the same point in space within symmetry tolerance.

As for qcdb, that's a @loriab question. I can never remember why we have both.

hokru:
> This can hypothetically happen in two ways: the atoms are either extremely close, or the tolerance is set to a large value.

unphysical close atoms should be acceptable as input, even if mostly of academic interest.

But I imagine in practice the error could occur if a user tries to symmetrize a molecule with a questionably large tolerance.
Then the question is if we should indeed abort or still try to symmetrize hoping to get lucky?
Potentially print the symmetrized molecule and abort after that.

TiborGY:
>  Potentially print the symmetrized molecule and abort after that.

Implementing that would require reworking the callers of `atom_at_position`, as it is a fairly low-level function that is not only called when we are trying to symmetrize. Throwing an exception in `atom_at_position` after the warning, and catching it in the symmetrization function may be feasible, other callers would not catch it and the uncaught exception would terminate execution.

> But I imagine in practice the error could occur if a user tries to symmetrize a molecule with a questionably large tolerance.
This is the primary reason why I am reworking this, although I also tend to abuse QC programs with thousands of extreme geometries.

A third option would be to raise an error in `atom_at_position` by default, but add an expert option to turn the error into a warning.

JonathonMisiewicz:
What legitimate use case is there for this being a warning instead of an error? I can't think of one.

hokru:
hmm, right. Best to keep it simple and abort then, it's a corner case after all.

TiborGY:
I am going to need to revise the docstrings now that we just error out instead of returning the closest match

zachglick:
I could see users placing two atoms at the same position if they're doing calculations with ghost atoms. However, I don't imagine `Molecule::atom_at_position` is used often, so this is very much a corner case.

TiborGY:
1. Done.
2. Good point, it could be called from other places.
3. N/A anymore
4. Sentence removed. I have taken the inspiration from the MRCC interface, which prints a warning like this. https://github.com/psi4/psi4/blob/4f7de4dca376b9c4afd2105898a98e06efcd7309/psi4/src/psi4/mrcc/mrcc.cc#L738

JonathonMisiewicz:
The MRCC interface is a different case because:
1. At the time it was coded, we had a developer who would volunteer to deal with it if requested.
2. We had a well-defined use case where we already knew that adding the feature was the correct fix.

Those are two special conditions, and there's a reason that there aren't other messages like the MRCC message in the Psi codebase!

JonathonMisiewicz (approved):

loriab (approved):
The net change is that `atom_in_position` now errors when more than one other atom is w/i radius, right? Edit the "stern warning" sentence in the PR intro to that effect, please, so we get the future release notes right.

Thanks for the change! I'll try to edit the py one to match.

TiborGY:
PR intro updated. Is a rebase going to be required before merge?

JonathonMisiewicz:
No. Rebases are only _required_ in case of merge conflict.

PR #2467, opened 2022-03-10
Remove Reference to Unimplemented Method

JonathonMisiewicz:
## Description
`LEOM_CCSD` is included in the list of EOM methods, but I can find no evidence that this method exists in Psi, other than as an option. This PR removes the option entirely.

This brings the EOM methods to create variables for down to EOMCCSD, EOMCC2, and EOMCC3.

Per Lori observation, the `LOCAL_GHOST` option is removed as unused.

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/cceom/Local.h
+++ b/psi4/src/psi4/cc/cceom/Local.h
@@ -51,7 +51,6 @@ struct Local {
     int *pairdom_len;
     int *pairdom_nrlen;
     int *weak_pairs;
-    int ghost;
     int do_singles;
     double ***V;
     double ***W;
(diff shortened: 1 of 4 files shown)


maxscheurer (approved):

loriab (approved):
option `LOCAL_GHOST` may also be associated

JonathonMisiewicz:
Oh, does the `L` stand for local? I thought it was `L` for lambda, as in lambda equations. In any event, the `LOCAL_GHOST` option isn't used either, so I'll go ahead and rip that out too.

lothian:
Yes, LEOM_CCSD refers to the local-EOM-CC code we developed for testing in Psi3.  Similarly, LOCAL_GHOST was a keyword for placing basis functions on a ghost atom to account for diffuse electronic states in the local-EOM-CC calculations. These capabilities were not transferred to Psi4, so these keywords may be safely removed.

lothian (approved):

Issue #2414, opened 2022-01-26
Implement McMurchie Davidson Integrals


andysim:
PR #2388 replaces most of the hand-coded one electron integrals with Libint2 calls, for code cleanliness and performance.  However, some of the integrals we need are not provided by that library.  We also need to replace the arbitrary order multipole moment integrals in the current implementation because they use an algorithm that is not numerically stable at high orders.  Although we could possibly use the existing Obara-Saika recursion code, arbitrary order derivatives of the potential and arbitrary order multipoles are much better handled by McMurchie Davidson code.  As a proof of concept, see the [attached Python script](https://github.com/psi4/psi4/files/7938278/integrals.py.zip) that computes these integrals through arbitrary order, as well as linear momemtum and angular momentum integrals.  All of those integrals can be computed through arbitrary orbital angular momentum and the entire code to compute them all / compare to existing Psi4 implementation is around 400 lines (with comments).  I plan to open a followup PR with this implementation, getting rid of all of the remaining Obara-Saika code afterwards, but welcome comments / opinions on this proposal.

maxscheurer:
For bookkeeping purposes, I'm going to list the integral classes that are going to need M-D:

- `AngularMomentumInt`
- `DipoleInt` (for analytic nuclear gradients)
- `MultipoleInt` (arbitrary order multipole moment integrals)
- `MultipolePotentialInt`


Integral classes that will still use OS86:

- `ThreeCenterOverlapInt`

**Untested** integral code that is not used anywhere:

- `PseudospectralInt` (`<m|erf(\omega r) / r|n>`)

zachglick:
By the way, please don't remove the `PseudospectralInt` class! Those integrals are necessary for semi-numerical exchange algorithms, which we'd like to get into the SCF code.

maxscheurer:
Good to know, I updated my comment... Would be good to have tests for these integrals though. @andysim could the pseudospectral ints be handled by M-D somehow?

zachglick:
Tests would definitely be good. I have a plugin that uses these integrals (and works), so I at least know nothing is broken.

Actually, I wonder why we aren't computing the pseudospectral integrals with libint2 now. They're closely related to standard one electron potential integrals that we use in Psi4. 

For semi-numerical exchange, I use the `PseudospectralInt` class to compute integrals of the form:
```<m|1 /(r-r_p) |n>```
where `r_p` is a DFT grid point. This is identical to the nuclear integrals here:
[...]

PR #2483, opened 2022-03-16
McMurchie-Davidson integrals for angular momentum

maxscheurer:
## Description
This PR refactors angular momentum integrals (`AngularMomentumInt`) using the McMurchie-Davidson scheme.

It is the first PR of a series to replace most of the existing OS86 code with M-D (#2414). I've chosen the angular momentum integrals because they are really simple and only require the E matrix (Hermite-to-Cartesian conversion factors).

## Todos

- [x] `AngularMomentumInt` w/ M-D scheme

## Questions
- [x] Is it worth to pre-allocate the E-matrix? This would require some boilerplate code, which could be worth packing into a small class that all future M-D integrals will inherit from?

## Checklist
- [x] Tests added for any new features (consistency check against Psi4 v1.5)

## Status
- [x] Ready for review
- [ ] Ready for merge


andysim:
Yeah, if we're going to allow C++17 we should upgrade the minimum [GCC to version 7](https://en.cppreference.com/w/cpp/compiler_support) in our CI.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,116 @@
...
+        for (int n = 0; n < am - l + 1; ++n) {
+            int m = am - l - n;
+            ret[index] = {{l, m, n, index}};
+            index++;
+        }
+    }
+    return ret;
+}
+
+inline double norm(const Point& A) { return std::sqrt(A[0] * A[0] + A[1] * A[1] + A[2] * A[2]); }
```
susilehtola, line 51: Shouldn't this be defined where `Point` is defined?

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,116 @@
...
+    Ez[0] = std::exp(-1.0 * mu * AB[2] * AB[2]);
+
+    double oo2p = 1.0 / (2.0 * p);
+    for (int i = 0; i < dim1; ++i) {
+        for (int j = 0; j < dim2; ++j) {
+            // t = 0 case
+            int idx0 = address_3d(i, j, 0, dim2, dim3);
+            if (i > 0) {
+                int idxi0 = address_3d(i - 1, j, 0, dim2, dim3);
+                int idxitp0 = address_3d(i - 1, j, 1, dim2, dim3);
```
susilehtola, line 83: This variable could be named more descriptively

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,116 @@
...
+                Ex[idx0] += PA[0] * Ex[idxi0] + Ex[idxitp0];
+                Ey[idx0] += PA[1] * Ey[idxi0] + Ey[idxitp0];
+                Ez[idx0] += PA[2] * Ez[idxi0] + Ez[idxitp0];
+            } else if (j > 0) {
+                int idxj0 = address_3d(i, j - 1, 0, dim2, dim3);
+                int idxjtp0 = address_3d(i, j - 1, 1, dim2, dim3);
```
susilehtola, line 89: same here

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,116 @@
...
+                Ex[idx0] += PB[0] * Ex[idxj0] + Ex[idxjtp0];
+                Ey[idx0] += PB[1] * Ey[idxj0] + Ey[idxjtp0];
+                Ez[idx0] += PB[2] * Ez[idxj0] + Ez[idxjtp0];
+            }
+            for (int t = 1; t < i + j + 1; ++t) {
+                int idx = address_3d(i, j, t, dim2, dim3);
+                if (i > 0) {
+                    int idxi = address_3d(i - 1, j, t, dim2, dim3);
+                    int idxit = address_3d(i - 1, j, t - 1, dim2, dim3);
+                    int idxitp = address_3d(i - 1, j, t + 1, dim2, dim3);
```
susilehtola, line 99: and here

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,116 @@
...
+                    Ex[idx] += PA[0] * Ex[idxi] + oo2p * Ex[idxit] + (t + 1) * Ex[idxitp];
+                    Ey[idx] += PA[1] * Ey[idxi] + oo2p * Ey[idxit] + (t + 1) * Ey[idxitp];
+                    Ez[idx] += PA[2] * Ez[idxi] + oo2p * Ez[idxit] + (t + 1) * Ez[idxitp];
+                } else if (j > 0) {
+                    int idxj = address_3d(i, j - 1, t, dim2, dim3);
+                    int idxjt = address_3d(i, j - 1, t - 1, dim2, dim3);
+                    int idxjtp = address_3d(i, j - 1, t + 1, dim2, dim3);
```
susilehtola, line 106: and here

psi4/src/psi4/libmints/mcmurchiedavidson.h
```diff
@@ -0,0 +1,41 @@
...
+#include <array>
+#include <vector>
+
+namespace mdintegrals {
+using Point = std::array<double, 3>;
+void fill_E_matrix(int maxam1, int maxam2, const Point& P, const Point& A, const Point& B, double a, double b,
+                   std::vector<double>& Ex, std::vector<double>& Ey, std::vector<double>& Ez);
+std::vector<std::array<int, 4>> generate_am_components_cca(int am);
+
+inline Point point_diff(const Point& A, const Point& B) { return {A[0] - B[0], A[1] - B[1], A[2] - B[2]}; }
```
susilehtola, line 39: You have `point_diff` here so you should have `point_norm` here too.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: Do you have a suggestion in mind? I always find it hard to consistently name index variables without making them too lenghty/verbose.

.azure-pipelines/azure-pipelines-linux.yml
```diff
@@ -9,13 +9,13 @@ jobs:
...
-        CXX_COMPILER: 'g++-6'
+      gcc_7:
+        CXX_COMPILER: 'g++-7'
         PYTHON_VER: '3.8'
-        C_COMPILER: 'gcc-6'
+        C_COMPILER: 'gcc-7'
         F_COMPILER: 'gfortran'
         BUILD_TYPE: 'release'
-        APT_INSTALL: 'gfortran gcc-6 g++-6'
+        APT_INSTALL: 'gfortran gcc-7 g++-7'
```
loriab, line 18: also edit here https://github.com/psi4/psi4/blob/master/cmake/custom_cxxstandard.cmake#L11 , please. and if you feel like fixing the block above it by killing off c++14 std, go for it.

maxscheurer: I've tried to update all the compiler versions in `custom_cxxstandard.cmake`, so please let me know if I got anything wrong.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: I updated all the `idx...` variable names to be more descriptive/consistent 😄

susilehtola: Well, at the least you could add a comment to say what is done in the loops, and comments on the indices explaining what they are...

maxscheurer: That's a good idea. Done 👍

jturney (approved):
Looks super nice, too.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,122 @@
...
+            index++;
+        }
+    }
+    return ret;
+}
+
+void fill_E_matrix(int maxam1, int maxam2, const Point& P, const Point& A, const Point& B, double a, double b,
+                   std::vector<double>& Ex, std::vector<double>& Ey, std::vector<double>& Ez) {
+    // computes the Hermite Gaussian expansion coefficients E_t^{ij} (eq 9.5.1)
+    // equation numbers from Molecular Electronic-Structure Theory (10.1002/9781119019572)
```
andysim, line 52: This code is correct, and it's not likely to be rate-limiting for the potential ints.  If benchmarks show that it is taking time, we can consider unrolling the loops to remove the if statements, [like we do for the OS recursions](https://github.com/psi4/psi4/blob/9f7b865b0cecf694e83e04637ab6d78b4480e43f/psi4/src/psi4/libmints/osrecur.cc#L1344-L1392).  That would remove branches from the inner loops, and I think would speed it up.  It's not important to change unless we see this routine taking a significant amount of time.  I see you already partially do this for the t loops.

psi4/src/psi4/libmints/angularmomentum.cc
```diff
@@ -26,242 +26,108 @@
...
+    const auto& comps_am2 = am_comps_[am2];
 
-    // compute intermediates
-    double AB2 = 0.0;
-    AB2 += (A[0] - B[0]) * (A[0] - B[0]);
-    AB2 += (A[1] - B[1]) * (A[1] - B[1]);
-    AB2 += (A[2] - B[2]) * (A[2] - B[2]);
+    auto A = s1.O;
+    auto B = s2.O;
+    // TODO: settable origin? (not supported previously)
```
andysim, line 72: I don't see a use case for a non-zero origin, but I think we should allow all integrals involving a third center to have that be settable.  I think that is more consistent behavior between the different engines, and it does not affect any of the code that does the computation.  Right now the Potential base class handles the setting of the origin - perhaps we should move that to the OneBody base class, and have the derived classes that don't support an origin (like kinetic, overlap) throw an error.  If we go this route, it's probably a separate PR because it will touch quite a few different objects in a very minor way.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: I've done some profiling for my standalone C++ code and filling the E-matrix was not a rate-limiting step at all, so I think it should be fine for now 👍

psi4/src/psi4/libmints/angularmomentum.cc
maxscheurer: Yes -- `set_origin` already exists in `OneBody`, but making this consistent among as many integral classes as possible is probably beyond the scope of the PR, especially since the "feature" is not needed right now.

tests/pytests/test_mcmurchie_davidson.py
```diff
@@ -0,0 +1,38 @@
...
+the McMurchie-Davidson scheme
+"""
+import numpy as np
+import pytest
+import psi4
+from pathlib import Path
+import json
+
+
+pytestmark = pytest.mark.quick
```
loriab, line 12:
```suggestion
pytestmark = [pytest.mark.psi, pytest.mark.api, pytest.mark.quick]
```
prep for a future PR

andysim (approved):
Thanks for taking the time to do this - I think it's an excellent cleanup, and it also provides some very useful machinery for other OEI types.  Kudos for the pytests - they will help us greatly as we continue to clean up the other integrals.  LGTM!

psi4/src/psi4/libmints/angularmomentum.cc
```diff
@@ -26,242 +26,108 @@
  * @END LICENSE
  */
 
-#include "psi4/libciomr/libciomr.h"
 #include "psi4/libmints/angularmomentum.h"
 #include "psi4/libmints/basisset.h"
-#include "psi4/libmints/molecule.h"
 #include "psi4/libmints/integral.h"
+#include "mcmurchiedavidson.h"
```
loriab, line 32:
```suggestion
#include "psi4/libmints/mcmurchiedavidson.h"
```
for consistency. or feel free to leave it as is, and we'll wait and see if the full addresses have actually been necessary. :-)

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -0,0 +1,122 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+#include "mcmurchiedavidson.h"
```
loriab, line 28:
```suggestion
#include "psi4/libmints/mcmurchiedavidson.h"
```
same as above

loriab (approved):
a few optional changes. lgtm!

maxscheurer:
I'll make the requested changes and squash-merge when CI passes. 👍 Thanks everyone for the reviews!

cmake/custom_cxxstandard.cmake
```diff
@@ -55,14 +55,13 @@ elseif (CMAKE_CXX_COMPILER_ID MATCHES AppleClang)
     endif()
 
 elseif (CMAKE_CXX_COMPILER_ID MATCHES Clang)
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 3.6)
-        message(FATAL_ERROR "CLANG version must be at least 3.6!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.0)
+        message(FATAL_ERROR "CLANG version must be at least 4.0!")
```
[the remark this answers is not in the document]
maxscheurer: Should I set it to 5 then? At least the structured bindings which I'm using should be covered by 4. What do you think?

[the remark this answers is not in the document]
loriab: Thanks, gtk on structured bindings as the key feature. I'm ok with leaving as 4 for now, then. I'll keep in mind that this value doesn't get exercised since we don't have a min Clang lane on Azure. Conda's running clang 12 & 13, fwiw.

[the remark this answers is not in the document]
maxscheurer: Let's go for Clang 5 then 👍

psi4/src/psi4/libmints/mcmurchiedavidson.cc
susilehtola: Much better. Even though the indices are still not explained, having the references to the equations for each line of code is sufficient information to understand the implementation.

maxscheurer: Yes, the eq numbers are helpful. I don't understand what you mean with "still not explained"? I added a comment for each index showing which element of the E matrix they point to. I think that should be explanation enough 😬

Issue #2476, opened 2022-03-14
Spin state changes during calculation


susilehtola:
Testing some calculations, I ran into very odd behavior in Psi4 1.3.2. The input
```
molecule {
0 2
Y
}

set basis seg-cc-pv5z-pp
set scf_type direct
set df_scf_guess false
set reference uhf
energy('scf')
```
leads to the output
```
  ==> Iterations <==

                        Total Energy        Delta E     RMS |[F,P]|

    Occupation by irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     2,    0,    0,    0,    0,    1,    1,    1 ]
    SOCC [     0,    0,    0,    0,    0,    1,    0,    0 ]

   @UHF iter   1:   -33.78761858648352   -3.37876e+01   4.53671e-02 DIIS
    Occupation by irrep:
             Ag   B1g   B2g   B3g    Au   B1u   B2u   B3u 
    DOCC [     1,    0,    0,    0,    0,    1,    1,    1 ]
    SOCC [     1,    1,    0,    1,    0,    0,    0,    0 ]

   @UHF iter   2:   -36.39804695906182   -2.61043e+00   2.16710e-02 DIIS
```

For some reason, the occupation update changes the spin state from a doublet (one unpaired electron) to hextet (3 unpaired electrons). Basis set is attached

[seg-cc-pv5z-pp.gbs.txt](https://github.com/psi4/psi4/files/8247541/seg-cc-pv5z-pp.gbs.txt)

susilehtola:
Reproduced with current master with `set scf_initial_accelerator none`.

susilehtola:
On further thought, I am not sure this is an error. In UHF there's no guarantee that the orbitals are spin-paired; the beta spatial orbitals are allowed to differ from the alpha ones. This naturally leads to spin contamination, but that's a feature not a bug!

JonathonMisiewicz:
You said it's a doublet in the molecule input. Psi should honor that. This is a bug.

susilehtola:
You're right. The problem is the use of `docc` and `socc`: https://github.com/psi4/psi4/blob/1ffb67bf52dcca70e357d23714c7d975f9d1356e/psi4/src/psi4/libscf_solver/hf.cc#L475

After the recent addition of the sanity check in #2488, the code crashes in the iteration where some symmetries have more beta electrons and alpha electrons.
```
Occupations
sym 0: 1 alpha 2 beta
[...]

PR #2488, opened 2022-03-20
Cleanup to find_occupation and more sanity checks

susilehtola:
## Description

Minor cleanups to `find_occupation` as well as the introduction of sanity checks for #2476 to prevent sudden changes of the spin state during the calculation.

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/hf.cc
+++ b/psi4/src/psi4/libscf_solver/hf.cc
@@ -424,46 +424,40 @@ void HF::find_occupation() {
     if (MOM_performed_) {
         MOM();
     } else {
-        // We first find the aufbau occupation.
-        // We then take our orbitals as the aufbau orbitals within the occupation.
-        std::vector<std::pair<double, int> > pairs_a;
-        std::vector<std::pair<double, int> > pairs_b;
-        for (int h = 0; h < epsilon_a_->nirrep(); ++h) {
-            for (int i = 0; i < epsilon_a_->dimpi()[h]; ++i) {
-                pairs_a.push_back(std::make_pair(epsilon_a_->get(h, i), h));
+        if (!input_docc_ && !input_socc_) {
+            // The occupations are determined by the Aufbau
+            // principle. We first collect all the orbital energies and
+            // sort them in increasing order
+            std::vector<std::pair<double, int> > pairs_a;
+            for (int h = 0; h < epsilon_a_->nirrep(); ++h) {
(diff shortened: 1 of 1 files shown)


JonathonMisiewicz:
Update your PR description. Behavior changes are not "minor cleanup", even if that behavior is as simple as raising an error where there wasn't one before.

loriab:
Between plain calcs, user-specified-occ-calcs, successive calcs, calcs involving pre-SCF-cycles, wavefunction reloading, restarts, etc., getting the SCF initialization info from places in the right priority is fraught and, I fear, not defensively tested. So it'd be useful to understand the circumstances and reasoning behind the proposed changes. I'm all for more sanity checks.

susilehtola:
> Update your PR description. Behavior changes are not "minor cleanup", even if that behavior is as simple as raising an error where there wasn't one before.

Done

jturney (approved):
More sanity checks is always a good thing.

zachglick (approved):
LGTM!

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -424,55 +424,71 @@ void HF::find_occupation() {
...
         }
 
-        int old_socc[8];
-        int old_docc[8];
+        int old_socc[nirrep_];
+        int old_docc[nirrep_];
         for (int h = 0; h < nirrep_; ++h) {
             old_socc[h] = soccpi_[h];
             old_docc[h] = doccpi_[h];
         }
```
JonathonMisiewicz, line 470: The reason the old code had `8` instead of `nirrep_` was because the old way of doing it created a statically allocated array. But`nirrep_` isn't known until runtime, so I'm worried that _even without new_, we're dynamically allocating.

`soccpi_` and `doccpi_` are Dimension objects, so why not have `old_socc` and `old_docc` by Dimension objects created by an implicit copy constructor? No chance of memory leaks and less code.

susilehtola: Sure. Using `Dimension` also resulted in other negative LOC.

JonathonMisiewicz (approved):
Thanks, Susi! Sanity checks are great.

Since Lori commented, I want to give her a chance to review before merging this in.

loriab (approved):

PR #2490, opened 2022-03-21
Ignore this one. Adds F12 integrals from libint2; removes libint1 and erd.

jturney:
## Description
WORK IN PROGRESS. PR created to run the test cases.

Adds F12 integrals from libint2 needed for F12 methods. As part of cleanup the choice of libint1 or erd integral engine is removed.

## Status
- [ ] Ready for review, only if you feel like it.
- [ ] NOT ready for merge


andysim:
Look at that deletion count!

PR #2503, opened 2022-03-25
Remove libint1 and erd integral engines.

jturney:
## Description
Removes libint1 and erd integral engines from Psi4. 

## Todos
- [x] libint1 removed
- [x] erd removed
- [ ] fix fchk writer to not use erd coefficients from the basis set object

## Status
- [ ] Ready for review
- [ ] Ready for merge


loriab:
there's more erd traces in `tests/erd/` and https://github.com/psi4/psi4/blob/cf0a0d80f5f35e9a2136253d6e476cf5a74b6145/tests/pytests/test_addons.py#L733-L738

loriab (approved):
Lgtm. I’ve run a clean full tests. I think there’s something a little wrong with the simint logic, but that has nothing to do with the topic.

JonathonMisiewicz (approved):

andysim:
Thanks, @loriab !  Sorry, I didn't get to ripping all of this out before when doing the L2 implementation.  This is going to be a very cathartic button click :)

PR #2504, opened 2022-03-25
Multipole Potential Integrals with McMurchie-Davidson

maxscheurer:
## Description
This is the third PR of the McMurchie-Davidson series, now tackling `MultipolePotentialInt` (used for PE/EFP).

I'm not happy with the performance yet, so I'll do some profiling to see what could be improved.

## Todos

- [x] implementation working
- [x] tests for arbitrary order multipole potentials (generalized existing tests)
- [ ] more equation numbers

## Questions
- [ ] Question1

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


maxscheurer:
## Small benchmark
The performance issues are resolved. :tada: As in #2496, only for small basis set angular momentum, the old code is faster than M-D, but for larger basis set/angmom, M-D outperforms the OS code.

Performed as in #2496. Here, the molecule is para-nitroaniline (typical example for EFP/PE :smile:).

### Results
| basis                | order | M-D [ms] | OS [ms] | ratio MD/OS |
|----------------------|-------|----------|---------|-------------|
| cc-pvdz (nbf = 170)  | 0     | 5.75     | 5.45    | 1.05        |
|                      | 1     | 9.55     | 7.72    | 1.24        |
|                      | 2     | 17.69    | 12.06   | 1.47        |
|                      | 3     | 30.41    | 19.61   | 1.55        |
| cc-pvtz (nbf = 384)  | 0     | 14.2     | 14.02   | 1.01        |
|                      | 1     | 27.75    | 25.62   | 1.08        |
|                      | 2     | 55.37    | 50.41   | 1.1         |
|                      | 3     | 99.74    | 89.52   | 1.1         |
| cc-pvqz (nbf = 730)  | 0     | 43.66    | 45.69   | 0.96        |
|                      | 1     | 94.84    | 101.92  | 0.93        |
|                      | 2     | 197.61   | 222.04  | 0.89        |
|                      | 3     | 366.13   | 428.15  | 0.85        |
| cc-pv5z (nbf = 1240) | 0     | 132.84   | 160.85  | 0.83        |
|                      | 1     | 323.07   | 406.33  | 0.8         |
|                      | 2     | 694.14   | 942.28  | 0.74        |
|                      | 3     | 1291.59  | 2011.75 | 0.64        |

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -160,4 +160,77 @@ void fill_M_matrix(int maxam, int maxpow, const Point& PC, double a, double b, s
...
+
+    std::vector<double> fmvals(maxam + 1);
+
+    // evaluate Boys function
+    fm_eval->eval(fmvals.data(), T, maxam);
+
+    int dim1 = maxam + 1;
+    int dim2 = dim1 * dim1 * dim1;
+    // zero out buffer
+    std::memset(&R[0], 0.0, dim1 * dim2 * sizeof(double));
```
JonathonMisiewicz, line 179: Can we use `std::fill` here? With standard compiler optimizations, that's as fast as `std::memset`, but it's cleaner.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -160,4 +160,77 @@ void fill_M_matrix(int maxam, int maxpow, const Point& PC, double a, double b, s
     }
 }
 
+void fill_R_matrix(int maxam, double p, const Point& P, const Point& C, std::vector<double>& R,
```
JonathonMisiewicz, line 163: Can you document the expected size of `R` somewhere?

JonathonMisiewicz (commented):
A couple minor comments from me. I'd be happy to approve, but you're probably better off with reviews from Lori, Andy, and Jet.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: I could change it to `std::fill(R.begin(), R.begin() + dim1 * dim2, 0.0;`, I think that should work?
We don't want to zero out the entire buffer though, that was the performance bottleneck in my previous version 😄

JonathonMisiewicz: If it doesn't affect [performance on standard optimization](https://stackoverflow.com/a/8849789), please make the change. I initially thought that `R` was length `dim1 * dim2`. Between this and the documentation comment, it should be clearer to other non-experts that is not the case.

maxscheurer: Yeah I know this post on StackOverflow... I don't know if partially zeroing out would affect the performance of `std::fill`, will have to try it out.

loriab (approved):
Nice test! I'd consider myself a gray checkmark, but lgtm!

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: Done.

psi4/src/psi4/libmints/mintshelper.h
```diff
@@ -302,7 +302,7 @@ class PSI_API MintsHelper {
     /// AO EFP Multipole Potential Integrals
     std::vector<SharedMatrix> ao_efp_multipole_potential(const std::vector<double>& origin = {0., 0., 0.},
                                                          int deriv = 0);
-    // AO EFP Multipole Potential Integrals
+    // AO Multipole Potential Integrals up to given order (in CCA lexicographic order)
     std::vector<SharedMatrix> ao_multipole_potential(const std::vector<double>& origin = {0., 0., 0.}, int max_k = 0,
```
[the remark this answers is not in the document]
maxscheurer: Done.

jturney (approved):
Looks good!

andysim (commented):
I will try to pull this and test with Intel compilers.  We know that all files that include L2's engine.h need C++14 instead of C++17 with Intel, but it's possible that the real culprit is the downstream boys.h that you are using here.  I don't think it is, but it's worth checking and an easy fix if it turns out to be the case.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -160,4 +160,78 @@ void fill_M_matrix(int maxam, int maxpow, const Point& PC, double a, double b, s
...
+                    if (t > 1) {
+                        val += (t - 1) * R[noffset + address_3d(t - 2, u, v, dim1, dim1)];  // R_{t-2,u,v}^{n+1}
+                    }
+                    val += PC[0] * R[noffset + address_3d(t - 1, u, v, dim1, dim1)];  // R_{t-1,u,v}^{n+1}
+                    R[n * dim2 + address_3d(t, u, v, dim1, dim1)] = val;              // R_{t,u,v}^{n}
+                }
+            }
+        }
+    }
+}
```
zachglick, line 235: Is this the most expensive part of the multipole potential integral calculation? If so, I wonder if there'd be any performance gains from flattening these loops.

psi4/src/export_mints.cc
```diff
@@ -1453,9 +1453,9 @@ void export_mints(py::module& m) {
         .def("ao_angular_momentum", &MintsHelper::ao_angular_momentum, "Vector AO angular momentum integrals")
         .def("so_angular_momentum", &MintsHelper::so_angular_momentum, "Vector SO angular momentum integrals")
         .def("ao_efp_multipole_potential", &MintsHelper::ao_efp_multipole_potential,
-             "Vector AO EFP multipole integrals", "origin"_a = std::vector<double>{0, 0, 0}, "deriv"_a = 0)
+             "Vector AO EFP multipole integrals", "origin"_a, "deriv"_a = 0)
         .def("ao_multipole_potential", &MintsHelper::ao_multipole_potential, "Vector AO multipole potential integrals",
-             "origin"_a = std::vector<double>{0, 0, 0}, "max_k"_a = 0, "deriv"_a = 0)
+             "order"_a, "origin"_a, "deriv"_a = 0)
         .def("electric_field", &MintsHelper::electric_field, "Vector electric field integrals",
              "origin"_a = std::vector<double>{0, 0, 0}, "deriv"_a = 0)
```
zachglick, line 1460: For consistency's sake, does is make sense to remove the default origin here too?

zachglick (approved):
This is great!

psi4/src/psi4/libmints/CMakeLists.txt
```diff
@@ -51,7 +50,6 @@ list(APPEND sources
   gaussquad.cc
   ecpint.cc
   orthog.cc
-  mcmurchiedavidson.cc
   )
 
 # l2intf is a listing of all the files that include libint2's engine.h include.  they must
```
andysim, line 55:
```suggestion
# l2intf is a listing of all the files that include libint2's boys.h include (which is included in L2's engine.h).  they must
```

psi4/src/psi4/libmints/mcmurchiedavidson.cc
```diff
@@ -160,4 +160,78 @@ void fill_M_matrix(int maxam, int maxpow, const Point& PC, double a, double b, s
...
+    // evaluate Boys function
+    fm_eval->eval(fmvals.data(), T, maxam);
+
+    int dim1 = maxam + 1;
+    int dim2 = dim1 * dim1 * dim1;
+    // R matrix buffer size needs to be at least dim1 * dim2,
+    // only zero out the required part of the buffer for performance
+    std::fill(R.begin(), R.begin() + dim1 * dim2, 0.0);
+
+    // NOTE: avoiding std::pow(-2.0 * p, n)
```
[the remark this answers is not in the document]
maxscheurer: It's actually quite slow in this case, because the compiler cannot expand the power in terms of multiplication automatically.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: What exactly do you mean by "flattening" in this case? Yes, the snippet you mention is one of the hotspots, but gathering all the contributions [here](https://github.com/psi4/psi4/blob/037ee51b2726f38f5fa79d258f1d05a223a78b7e/psi4/src/psi4/libmints/multipolepotential.cc#L148) is even more expensive.

psi4/src/export_mints.cc
maxscheurer: Done.

psi4/src/psi4/libmints/mcmurchiedavidson.cc
maxscheurer: I've tested the flattened loop version, which is unfortunately much slower than the 3-loop in this case, but it was worth a try I guess 😄

zachglick: Haha, thanks for giving it a shot! Holger is definitely right about integer division

psi4/src/psi4/libmints/multipolepotential.cc
```diff
@@ -31,201 +31,142 @@
 #include "psi4/libmints/integral.h"
 
 #include <libint2/shell.h>
+#include <libint2/boys.h>
```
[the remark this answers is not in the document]
andysim: Oh, ignore that.  I see now.  You set up the evaluator here, to pass into that routine.  All good!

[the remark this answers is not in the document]
maxscheurer: No, the Boys function evaluator is set up in the ctor of the class. Getting the singleton instance inside `fill_R_matrix` turned out to be quite slow in my tests.

andysim (approved):
Another fantastic effort!  It's great to see negative lines of code, especially with so much new functionality.  LGTM!

PR #2515, opened 2022-03-30
ExternalPotential in Bohr

loriab:
## Description
The `psi4.core.ExternalPotential` object has long stored its charge locations in the same units as its BasisSet's Molecule. Also, the usual way of initializing it is by creating a `psi4.driver.qmmm.QMMM()` object and then doing the unusual `psi4.set_global_option_python("EXTERN", qmmmobj.extern)` to a field of that object. That's the situation for "plain" external charges calcs, where there's only one potential and both charges and locations are specified.

Additionally, there's two related categories. F-SAPT can put _different_ external potentials on any of fragments A, B, C, which have been passed in as  `energy("fisapt", external_potentials={"A": extern, ...}, ...)`. Also nbody can take an array of charges (not locations) that replace fragments when the fragments are ghosted. These are passed as `energy(..., bsse_type=..., embedding_charges=[monoAchg, ...])`

Problems:
* formation of Python objects in the input file and special options setting doesn't translate to QCSchema, which is the sole means of communication for many calcs in DDD.
* the most common calc, the simple single extern has no kwarg-to-energy way to set
* the uncertainty in units of ExternalPotential can be confusing and has led to bugs. Also, in DDD, the spec order of mol and extern may not be so clear to set the units.

This is No. 2 of the DDD series, #1351.

## Todos
- [x] Replace `QMMM()` with `QMMMbohr()` and issue an update guide if the former is called.
- [x] For the common single-extern case, switch to `energy(..., external_potentials=array` instead of `QMMM()` object. This tests whether the `external_potentials` value is an array for a single extern, which gets processed and set immediately, or a dict of externs, which gets handled by the run_fsapt later on.
- [x] relaxed the array spec by which externs are initialized. instead of `q, [x, y, z]` also allow `q, x, y, z`, which makes for very easy numpy processing for units transofmation.
- [x] removed all the units handling around ExternalPotential
- [x] docs 

## Questions
- [ ] ok to use external_potentials kwarg for both simple and fsapt multi frags use cases?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -955,7 +955,7 @@ computations, |PSIfour| can perform more rudimentary QM/MM procedures via the
 
     gradient('scf', external_potentials=external_potentials)
 
-The ``external_potentials`` array has three rows for thee separate
+The ``external_potentials`` array has three rows for three separate
 particles, and it is passed to the SCF code on the last line. The
 rows are composed of the atomic charge, x coordinate, y coordinate,
 and z coordinate in that order. The atomic charge and coordinates are
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -973,10 +973,17 @@ needed to describe the full MM region.
 
     gradient('scf')
 
-   The main differences are the specification of charge locations in
-   units of the active molecule, rather than always in Bohr, and in
-   the use of the :py:class:`psi4.driver.QMMM` class, which is now
-   discouraged from being used directly.
+   The main differences are that (1) the specification of
+   charge locations in the old way used the units of the active
+   molecule, whereas the new way always uses Bohr and (2) the
+   specification of the charge and locations in the old way used the
+   :py:class:`psi4.driver.QMMM` class directly and added one charge
+   per command, whereas the new way consolidates all into an array and
+   passes it by keyword argument to the calculation.
+
+   The successor to the :py:class:`psi4.driver.QMMM` class,
+   :py:class:`psi4.driver.QMMMbohr`, is operable, but it is discouraged
+   from being used directly.
 
 To run a computation in a constant dipole field, the |scf__perturb_h|,
 |scf__perturb_with| and |scf__perturb_dipole| keywords can be used.  As an
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -947,19 +947,36 @@ computations, |PSIfour| can perform more rudimentary QM/MM procedures via the
 |scf__extern| keyword.  The following snippet, extracted from the
 :srcsample:`extern1` test case, demonstrates its use for a TIP3P external potential::
 
+    import numpy as np
+    external_potentials = [
+        [-0.834, np.array([1.649232019048,0.0,-2.356023604706]) / psi_bohr2angstroms],
+        [ 0.417, np.array([0.544757019107,0.0,-3.799961446760]) / psi_bohr2angstroms],
+        [ 0.417, np.array([0.544757019107,0.0,-0.912085762652]) / psi_bohr2angstroms]]
+
+    gradient('scf', external_potentials=external_potentials)
+
+The ``external_potentials`` array has three rows for thee separate
+particles, and it is passed to the SCF code on the last line. The
+rows are composed of the atomic charge, x coordinate, y coordinate,
+and z coordinate in that order. The atomic charge and coordinates are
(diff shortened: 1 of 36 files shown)


loriab:
looks like MDI is the first casualty of QMMM and ecosystem GHA. This PR can have a warning for QMMM, then kill it off for DDD.

psi4/driver/procrouting/proc.py
```diff
@@ -1249,12 +1249,23 @@ def scf_wavefunction_factory(name, ref_wfn, reference, **kwargs):
         # For the dimer SAPT calculation, we need to account for the external potential
         # in all of the subsystems A, B, C. So we add them all in total_external_potential
         # and set the external potential to the dimer wave function
+        from psi4.driver.qmmm import QMMMbohr
```
JonathonMisiewicz, line 1252: Import statements should be at the top of the file.

JonathonMisiewicz: I am perfectly happy to save this for #2166, after DDD.

psi4/driver/procrouting/proc.py
```diff
@@ -1249,12 +1249,23 @@ def scf_wavefunction_factory(name, ref_wfn, reference, **kwargs):
...
+
         total_external_potential = core.ExternalPotential()
 
         for frag in kwargs['external_potentials']:
             if frag.upper() in "ABC":
-                wfn.set_potential_variable(frag.upper(), kwargs['external_potentials'][frag].extern)
-                total_external_potential.appendCharges(kwargs['external_potentials'][frag].extern.getCharges())
+                chrgfield = QMMMbohr()
+                for qxyz in kwargs['external_potentials'][frag]:
```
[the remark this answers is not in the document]
loriab: True. And it'll get worse, as nbody needs a similar function in DDD. I agree on consolidation down the road, but (1) with restructuring imminent and (2) import errors when I try such a fn in utils, I prefer to defer it.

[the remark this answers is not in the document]
JonathonMisiewicz: Makes sense to me. Because I haven't been responsible for enough growth in the issue list lately, can you create a `post-DDD` issue, similar to the `post-L2` issue? I will happily accept "this needs to wait until after the rest of DDD is in" for any code cleanup suggestion I make, so long as it gets tracked on such an issue.

JonathonMisiewicz (approved):
LGTM now, many thanks for breaking DDD down.

hokru (approved):
lgtm. I think this will lead to less confusion about units.

maxscheurer (approved):

doc/sphinxman/source/scf.rst
```diff
@@ -947,19 +947,43 @@ computations, |PSIfour| can perform more rudimentary QM/MM procedures via the
...
 
+    import numpy as np
+    external_potentials = [
+        [-0.834, np.array([1.649232019048,0.0,-2.356023604706]) / psi_bohr2angstroms],
+        [ 0.417, np.array([0.544757019107,0.0,-3.799961446760]) / psi_bohr2angstroms],
+        [ 0.417, np.array([0.544757019107,0.0,-0.912085762652]) / psi_bohr2angstroms]]
+
+    gradient('scf', external_potentials=external_potentials)
+
+The ``external_potentials`` array has three rows for thee separate
```
andysim, line 958:
```suggestion
The ``external_potentials`` array has three rows for three separate
```

tests/fsapt-ext-abc2/input.dat
```diff
@@ -20,20 +20,20 @@ no_reorient
...
+Chrgfield_B[:,[1,2,3]] /= psi_bohr2angstroms
+
+Chrgfield_C = np.array([
+ 0.417,  1.1270, 1.5527, -0.1658,
+-0.834,  1.9896, 1.0738, -0.1673,
+ 0.417,  2.6619, 1.7546, -0.2910,
+ 0.417, -1.0231, 1.6243, -0.8743,
+-0.834, -0.5806, 2.0297, -0.1111,
+ 0.417, -0.9480, 1.5096,  0.6281]).reshape((-1, 4))
+Chrgfield_C[:,[1,2,3]] /= psi_bohr2angstroms
```
andysim, line 36: Excellent - great to show the user how to get bohr from Angstrom easily :)

andysim (approved):
LGTM - this will be a little easier to generalize to other input formats.  Thanks for taking care of it.  This will simplify the blurred multipole specification, methinks.

PR #2525, opened 2022-03-31
EOMCC Einstein + Rot. Strength + Tests

JonathonMisiewicz:
## Description
This PR exposes Einstein Coefficients and Rotary Strengths to `psivar`. As of this PR, cc tests 1-33, 35-45, 47, 49-53, 55-56 are ported. Four tests to go.

I'm not ultimately happy with these tests (see the six issues I filed this evening), but it's as good as we have right now.

## Questions
- [x] **Please confirm the descriptions and units of the glossary additions.**

## Checklist
- [x] Newly added tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


lothian:
How did you obtain the units?  When I look at the code (I can't find my notes on the Einstein coefficients), in SI, I get:
B <-- m^3 J^-1 s^-2
A <-- s^-1 (which is the unit reported in the tabular output that includes A)

JonathonMisiewicz:
That's right. I made three different mistakes in the units (missed the section you pointed out, didn't realize you were converting _to_ not _from_ SI at the end, and mixed up the probability-from-mean-intensity definition vs. the probability-from-energy-density definition of the Einstein B.)

Documentation updated.

tests/cc56/input.dat
```diff
@@ -0,0 +1,58 @@
...
+
+test_data = { # TEST
+"ROOT 0 (B1) -> ROOT 0 (A2)": {"OSCILLATOR STRENGTH (LEN)": 0.00407528, "EINSTEIN A (LEN)": 3.86413231e+07, "EINSTEIN B (LEN)": 1.36907729e+18}, # TEST
+"ROOT 0 (B1) -> ROOT 1 (B1)": {"OSCILLATOR STRENGTH (LEN)": 0.00317063, "EINSTEIN A (LEN)": 2.32728034e+07, "EINSTEIN B (LEN)": 1.21063018e+18}, # TEST
+"ROOT 1 (B1) -> ROOT 0 (A2)": {"OSCILLATOR STRENGTH (LEN)": 0.09372883, "EINSTEIN A (LEN)": 1.28314944e+07, "EINSTEIN B (LEN)": 2.62052705e+20}, # TEST
+}
+
+for transition, data in test_data.items(): # TEST
+    for property, value in data.items(): # TEST
+        # Change these numbers to 6 ASAP # TEST
```
loriab, line 49: the rtol? all the atol look to be 1.e-6 now.

tests/cc56/input.dat
```diff
@@ -0,0 +1,58 @@
...
+ref1B1 = -75.325751117648 # TEST
+ref0A2 = -75.260476163556 # TEST
+
+compare_values(ref1B1, core.variable("CCSD ROOT 1 (B1) TOTAL ENERGY"), 6, "Root 1 (A1) Energy") # TEST
+compare_values(ref0A2, core.variable("CCSD ROOT 0 (A2) TOTAL ENERGY"), 6, "Root 0 (A2) Energy") # TEST
+
+### Transition checks
+
+test_data = { # TEST
+"ROOT 0 (B1) -> ROOT 0 (A2)": {"OSCILLATOR STRENGTH (LEN)": 0.00407528, "EINSTEIN A (LEN)": 3.86413231e+07, "EINSTEIN B (LEN)": 1.36907729e+18}, # TEST
```
loriab, line 42: I hate to suggest it now, but what are your thoughts on `ROOT n (h)` always meaning all-root ordering and `ROOT n OF (h)` always meaning within-symmetry-ordering?

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -328,6 +328,16 @@ PSI Variables by Alpha
...
 
+.. psivar:: CCname ROOT n (h) -> ROOT m (i) EINSTEIN A (LEN)
+
+   The Einstein A coefficient, the spontaneous emission 'probability.'
+   Units are in [1/s].
+
+.. psivar:: CCname ROOT n (h) -> ROOT m (i) EINSTEIN B (LEN)
+
+   The Einstein B coefficient, the stimulated emission 'probability'
+   in terms of energy density. Units are in [m^3 / J / s^2].
```
loriab, line 339: I periodically forget this, but we now try to save qcvars in au for compatibility with QCSchema and to avoid unit conversion error. I should plan on a pass (before v1.6) to switch to au. But that'll involve tests to check pint can to the au -> SI conversion easily. For now, it's great to get these saved and in at least one set of known units :-)

loriab (approved):
excellent idea on the qcvars saver

tests/cc56/input.dat
JonathonMisiewicz: Those should be 1e-4 given the lambda convergence. I'll change in next commit.

tests/cc56/input.dat
JonathonMisiewicz: I hate to admit it now, but that _is_ cleaner.

We're keeping "ROOT n OF (h)", but do you want "ROOT n (h)" notation, to boot?

loriab: "OF" or "IN", do you think?

> do you want "ROOT n (h)" notation, to boot?

I don't know. may as well for now. up to you.

JonathonMisiewicz: You're right, "IN" is better.

I think the plan _for right now_ is to keep with the current notation, then have a cleanup PR once the series is over (we're very close), where switch all excited state variables to "IN" notation and add the new "ROOT n (h)" notation as well.

JonathonMisiewicz:
Reminder that this PR could use @lothian review now that the units are clarified.

psi4/src/psi4/cc/ccdensity/ccdensity.cc
```diff
@@ -118,7 +118,7 @@ void td_setup(const struct TD_Params& S);
 void tdensity(const struct TD_Params& S);
 void td_print();
 void oscillator_strength(std::shared_ptr<Wavefunction> wfn, struct TD_Params *S);
-void rotational_strength(MintsHelper &mints, struct TD_Params *S);
+void rotational_strength(std::shared_ptr<Wavefunction> wfn, MintsHelper &mints, struct TD_Params *S);
```
jturney, line 121: If these functions are now accepting a Wavefunction is there a need to pass in a MintsHelper, too?

JonathonMisiewicz:
I rebased the PR.

In addition to eliminating the redundant argument, I did some `const` marking.

jturney (approved):

PR #2529, opened 2022-04-01
Miscellaneous JK Improvements

zachglick:
## Description
This PR makes two small improvements to the `JK` class and SCF code. These improvements simplify the addition of new JK algorithms to Psi4, particularly an upcoming semi-numerical exchange.

The first change is transferring the responsibility of zeroing out the `J`, `K`, and `wK` matrices from the parent `JK` class to each derived `JK` class. This makes it easier to implement JK algorithms which build the fock matrix incrementally. (i.e. use the difference in density between SCF iterations to compute differences in J/K). With this change, much of the incremental fock code in the `DirectJK` class can be simplified.

The second change is the addition of an `early_screening_` member variable to the `JK`. The idea is that some future JK algorithms will increase performance by using looser screening procedures/thresholds in early SCF iterations. Screening is then tightened as the SCF approaches convergence. This logic was added to the SCF driver. The `early_screening_` variable defaults to false for all existing JK classes, so there is currently. no change in behavior.

## Todos
- [x] Derived `JK` classes are responsible for zeroing their matrices
- [x] JK screening is SCF iteration dependent

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -262,6 +262,9 @@ def scf_iterate(self, e_conv=None, d_conv=None):
     frac_enabled = _validate_frac()
     efp_enabled = hasattr(self.molecule(), 'EFP')
 
+    # does the JK algorithm use severe screening approximations for early SCF iterations?
+    early_screening = self.jk().get_early_screening()
+
     # SCF iterations!
     SCFE_old = 0.0
     Dnorm = 0.0
@@ -481,7 +484,13 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
         # Call any postiteration callbacks
         if not ((self.iteration_ == 0) and self.sad_) and _converged(Ediff, Dnorm, e_conv=e_conv, d_conv=d_conv):
-            break
+            if early_screening:
+                early_screening = False
+                self.jk().set_early_screening(False)
+                core.print_out("  Energy and wave function converged with early screening.\n")
+                core.print_out("  Continuing iterations with tighter screening.\n\n")
+            else:
+                break
         if self.iteration_ >= core.get_option('SCF', 'MAXITER'):
             raise SCFConvergenceError("""SCF iterations""", self.iteration_, self, Ediff, Dnorm)
 
--- a/psi4/src/psi4/libfock/jk.cc
+++ b/psi4/src/psi4/libfock/jk.cc
@@ -330,12 +330,6 @@ void JK::allocate_JK() {
         }
     }
 
-    // Zero out J/K for compute_JK()
-    for (size_t N = 0; N < D_.size(); ++N) {
-        if (do_J_) J_[N]->zero();
-        if (do_K_) K_[N]->zero();
-        if (do_wK_) wK_[N]->zero();
-    }
 }
 void JK::USO2AO() {
     allocate_JK();
@@ -481,11 +475,6 @@ void JK::USO2AO() {
         }
     }
 
-    for (size_t N = 0; N < D_.size(); ++N) {
-        if (do_J_) J_ao_[N]->zero();
-        if (do_K_) K_ao_[N]->zero();
-        if (do_wK_) wK_ao_[N]->zero();
-    }
 }
 void JK::AO2USO() {
     // If already C1, J/K are J_ao/K_ao, pointers are already aliased
@@ -634,11 +623,28 @@ void JK::compute() {
         C_right_.clear();
     }
 }
+
 void JK::set_wcombine(bool wcombine) {
     wcombine_ = wcombine;
     if (wcombine) {
         throw PSIEXCEPTION("To combine exchange terms, use MemDFJK\n");
     }
 }
+
+void JK::zero() {
+    if (do_J_) {
+        for(auto J : J_) J->zero();
+        for(auto J : J_ao_) J->zero();
+    }
+    if (do_K_) {
+        for(auto K : K_) K->zero();
+        for(auto K : K_ao_) K->zero();
+    }
+    if (do_wK_) {
+        for(auto wK : wK_) wK->zero();
+        for(auto wK : wK_ao_) wK->zero();
+    }
+}
+
 void JK::finalize() { postiterations(); }
 }  // namespace psi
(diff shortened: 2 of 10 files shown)


JonathonMisiewicz:
I know next-to-nothing about the CDS's group's future JK plans. Before I can approve this, I'd like some confidence that these changes really are essential preparation for work-in-progress in the CDS group that you expect to be submtited soon.

Can you give me an outline, or a draft PR, of how this responsibility transfer simplifies Incremeental Fock?

Is `early_screening_` for this "semi-numerical exchange," or something else? How close is this other PR to completion.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -481,7 +484,13 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
         # Call any postiteration callbacks
         if not ((self.iteration_ == 0) and self.sad_) and _converged(Ediff, Dnorm, e_conv=e_conv, d_conv=d_conv):
-            break
+            if early_screening:
+                early_screening = False
+                self.jk().set_early_screening(False)
```
loriab, line 489: is the idea that E_CONV and D_CONV values in the options object are going to be changing between the early_screening T and Fa stages? otherwise, I don't think I'm following the iteration logic.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -634,11 +624,28 @@ void JK::compute() {
...
+    }
+    if (do_K_) {
+        for(auto K : K_) K->zero();
+        for(auto K : K_ao_) K->zero();
+    }
+    if (do_wK_) {
+        for(auto wK : wK_) wK->zero();
+        for(auto wK : wK_ao_) wK->zero();
+    }
+}
```
loriab, line 648: this does look a lot cleaner, thanks!

jturney: Yeah, I like this better 👍

jturney (approved):
This all seem reasonable to me.

zachglick:
> I'd like some confidence that these changes really are essential preparation for work-in-progress in the CDS group that you expect to be submtited soon.

Both of these changes are necessary for semi-numerical exchange, which I plan to submit in the next week or two. Because these changes touch important parts of the SCF/JK code but shouldn't change any existing behavior, I thought it would be best to submit them separately.

> Can you give me an outline, or a draft PR, of how this responsibility transfer simplifies Incremeental Fock?

Sure, I'll do my best to explain. As you probably know, `JK` objects are used throughout the codebase via calls to `JK::compute()`. In SCF, many calls to `JK::compute()` are made interspersed with updates to the object's orbitals. `JK::compute()` does a [few things](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/jk.cc#L550): performs sanity checks, creates densities from the orbitals, zeroes out the object's `J`/`K`/`wK` matrices, and finally calls the virtual function `compute_JK()`, which is where each child class implements some algorithm to fill `J`/`K`/`wK`.

Now consider incremental Fock construction. Instead of recomputing `J`/`K`/`wK` each SCF iteration, the previous iteration's `J`/`K`/`wK` are added to a `J`/`K`/`wK` computed with the difference in densities between the current and previous iterations.

As [implemented in the `DirectJK` class](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/jk.h#L724-L734), this accomplished by storing the entire `J`/`K`/`wK` matrices of the previous iteration, computing the difference density update to `J`/`K`/`wK`, adding the two together, and storing in the `JK` objects `J`/`K`/`wK` matrices. This requires a lot of [convoluted bookkeeping](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/DirectJK.cc#L138-L217). Future JK classes that would want to do incremental Fock build would have to duplicate a lot of code, too.

Instead, a much simpler approach would be this: The call to `JK::compute()` doesn't zero out the `J`/`K`/`wK` matrices. Instead, the child JK class can choose to zero out their `J`/`K`/`wK` matrices (if performing a normal Fock build) or leave them as-is from the previous iteration and add to them (if performing an incremental Fock build). This way no previous iterations' `J`/`K`/`wK` matrices have to be saved, and the logic in `compute_JK` becomes much simpler. 

> Is `early_screening_` for this "semi-numerical exchange," or something else? How close is this other PR to completion.

Yes, this is used in semi-numerical exchange.

psi4/driver/procrouting/scf_proc/scf_iterator.py
zachglick: Not quite; `E_CONV` and `D_CONV` don't change. The idea is that JK algorithms have various parameters and thresholds that control accuracy (think `INTS_TOLERANCE`). Instead of using a tight `INTS_TOLERANCE` throughout a whole SCF calculation, you could get away with using a loose `INTS_TOLERANCE` in the earlier iterations, and then tightening the parameter in the final iterations.

`early_screening_` is a boolean attached to the `JK` class that tells the class whether or not it's allowed to get away with such "looser" cutoffs. The iteration logic works like this:  
 - If `early_screening_` is off, SCF runs to convergence as it always has.
  - If `early_screening_` is on, SCF runs to convergence, then switches `early_convergence_` to off. We then resume the SCF iterations because the switch from on to off may slightly shift the location of the SCF solution and require another iteration or two.

I've set `early_screening_` to false in all existing `JK` classes, so it doesn't currently affect any behavior. In semi-numerical exchange, the cost is proportional to the size of the integration grid. You can get a big performance boost by using a small grid until SCF convergence, and then switch to a large grid for a final iteration.

loriab: Ok, got it, I think. `INTS_TOLERANCE` or similar plays the role of hidden variable in the SCF iterations whose value is influenced by `early_screening_`. I couldn't figure how querying early_screening before iterations and resetting it after iterations was going to work. Thanks for the explanation.

JonathonMisiewicz (approved):

loriab (approved):

psi4/driver/procrouting/scf_proc/scf_iterator.py
zachglick: Exactly! I'm open to suggestions if you can think of a more immediately understandable variable name than `early_screening_`

loriab: `early_screening_` is fine by me. It was only the mechanism of `early_screening_` that confused me, and that'll be traceable once it appears in a JK.

PR #2531, opened 2022-04-04
Destroy `TaskListComputer`

JonathonMisiewicz:
## Description
This PR gets rid of the `TaskListComputer` base class, condensing it into `OEProp`, the only class that inherited from it. We gained nothing from having that class around.

## Checklist
- [x] Properties tests pass

## Status
- [x] Ready for review
- [x] Ready for merge


jturney (approved):
Seems simple enough 👍

PR #2532, opened 2022-04-05
new string cbs syntax

loriab:
## Description
These are the breaking, user-facing changes for cbs in DDD. That is, these changes aren't needed at present for the cbs back-end, but this is coercing the input front-end into the right syntax.

This is No. 6 of the DDD series, #1351.

## Todos

- [x] All `fn(cbs)` and passing to extrapolation functions are changed over to `fn("cbs")` and passing the _names_ of extrapolation functions. This is making cbs behave more like other wrappers: findif you don't call directly, you use kwarg dertype and nbody you don't call directly, you use bsse_type. Functions are trickier to keep imported and don't serialize well, so in ddd, they're registered and replaced by strings, akin to the procedures dictionary.
- [x] alias functions are replaced with dict specification instead of kwarg specification
- [x] UpdateHelpers are added so that if you feed an existing input, it prompts you to make the fn -> str changes.
- [x] Fix bug where allen_focal_point wasn't working b/c used kwargs, not dict, specification for cbs, so the higher deltas were getting lopped off.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/aliases.py
```diff
@@ -102,21 +102,25 @@ def sherrill_gold_standard(func, label, **kwargs):
...
+        'wfn_lesser': kwargs.pop('delta_wfn_lesser', 'mp2'),
+        'basis': kwargs.pop('delta_basis', 'aug-cc-pVTZ'),
+        'scheme': kwargs.pop('delta_scheme', 'xtpl_highest_1')
+    }
+
+    kwargs["cbs_metadata"] = [scf, corl, delta]
+    return driver_cbs.cbs(func, "", **kwargs)
 
 
 def allen_focal_point(func, label, **kwargs):
```
JonathonMisiewicz, line 126: The test suite has zero coverage of `allen_focal_point`. I insist on adding a regression test on energies, so we can be sure this PR doesn't change them.

tests/pywrap-cbs1/input.dat
```diff
@@ -123,5 +123,7 @@ h2.R = 0.747953788665
 #            delta_basis='cc-pV[dt]Z')
 #compare_values(-1.17505568495, e_cbs, 6, "[5a] user-def xtpl cbs()")
 
-e_cbs = energy('MP2/aug-cc-pV[T,Q]Z + D:CCSD(T)/cc-pv[dt]z', delta_scheme=myownmy_2, molecule=h2)
+
+register_xtpl_function(myownmy_2)
```
JonathonMisiewicz, line 127: We need documentation on how to use `register_xtpl_function`.

tests/cbs-xtpl-nbody/input.dat
```diff
@@ -20,26 +20,32 @@ set {
...
+    delta = {
+        'wfn': kwargs.pop('delta_wfn', 'ccsd(t)'),
+        'wfn_lesser': kwargs.pop('delta_wfn_lesser', 'mp2'),
+        'basis': kwargs.pop('delta_basis', 'cc-pVDZ'),
+        'scheme': kwargs.pop('delta_scheme', 'xtpl_highest_1')
+    }
+    kwargs["cbs_metadata"] = [scf, corl, delta]
     return driver_cbs.cbs(func, label, **kwargs)
 
+register_composite_function(unaug_focal)
```
JonathonMisiewicz, line 43: We need documentation on `register_composite_function` and the use of `cbs_metadata`.

psi4/driver/driver_nbody.py
```diff
@@ -358,6 +358,10 @@ def nbody_gufunc(func: Union[str, Callable], method_string: str, **kwargs):
     elif metadata['ptype'] == 'hessian':
         core.set_variable("CURRENT HESSIAN", nbody_results['ret_ptype'])
 
+    nbody_number = sum(len(v) for k, v in metadata["compute_dict"]["all"].items())
+    core.set_variable("NBODY NUMBER", nbody_number)
```
JonathonMisiewicz, line 362: I can't find any reference to this variable in existing Psi. Is this a new variable?

JonathonMisiewicz (changes requested):
Mostly documentation requests on my end (no surprise there), but there's a gap in the test suite that _needs_ to be fixed before I can approve this PR.

doc/sphinxman/source/psithoninput.rst
```diff
@@ -481,15 +481,15 @@ do the extrapolation).  For example,::
...
 
-  cbs('mp2', corl_basis='cc-pv[dt]z', corl_scheme=corl_xtpl_helgaker_2)
+  energy('mp2', corl_basis='cc-pv[dt]z', corl_scheme=corl_xtpl_helgaker_2)
 
   energy('mp2/cc-pv[dt]z')
 
 will compute a 2-point Helgaker extrapolation of the correlation energy
 using the cc-pVDZ and cc-pVTZ basis sets (with method MP2) and add this
 extrapolated correlation energy to the Hartree--Fock energy in the
 largest basis (cc-pVTZ). :py:func:`~psi4.cbs` can
```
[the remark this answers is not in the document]
loriab: sure it is! but no longer explicitly. any time there's a `[dt]`, cbs is active behind the scenes. (actually, in master, anytime there's a `mtd/bas`, cbs gets invoked in part. in the beginning there was `cbs("mp2")`. then in 2016, `energy(cbs)`. now `energy("cbs")`. so it's receding all the time.

psi4/driver/aliases.py
loriab: yeah, trouble is it would scarcely ever be run since it depends on mrcc. which reminds me I ought to follow Peter's recc and at least check mrcc present.

tests/cbs-xtpl-nbody/input.dat
loriab: For the most part, the updatehelpers guide the use. but never hurts to docstring the fn. `cbs_metadata` is pretty thoroughly demonstrated in https://github.com/psi4/psi4/blob/master/tests/cbs-xtpl-dict/input.dat . I'll add that hint to docs.

psi4/driver/driver_nbody.py
loriab: Yes, I've put "cbs|nbody|findif number" vars in ddd as a way to check the wrappers are planning correctly. so pre-porting those to master. need to tell the docs.

JonathonMisiewicz (approved):

PeterKraus (approved):
Everything here is clear to me and I like the "deprecation" of `cbs` as a function. It's a good idea to have the "number of steps" variables also for findif and n-body, especially as they're counting the AtomicResults (or whatever it is in DDD). LGTM.

hokru (approved):

Issue #2522, opened 2022-03-30
Fix `cclambda` Bad Initial Iteration Printout


JonathonMisiewicz:
When doing multiple `cclambda` runs, the [closing RMS from the previous run](https://github.com/psi4/psi4/blob/master/tests/cc55/output.ref#L1113) is printed as the [opening RMS for the next run](https://github.com/psi4/psi4/blob/master/tests/cc55/output.ref#L1113). That's wrong. The opening RMS should just be zero, like it is for the first run.

#943 first, yet again.

Issue #2520, opened 2022-03-30
`cclambda` Should Honor `r_convergence`


JonathonMisiewicz:
The `cc` module [forcibly stops users](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L3149-L3152) from requesting tight convergence out of `cclambda`, so as not to [break tests](https://github.com/psi4/psi4archive/commit/2b2eb6c5a710dde180b1ef7a050012069afeae8d). That's bad.

Deal with #2519 first.

lothian:
I agree that `cclambda` should honor the user's `r_convergence` choice, but the default settings we chose are not to avoid breaking tests but because tighter settings are wholly unnecessary (and often exceedingly difficult to achieve) for transition properties.  The `cc55` test case just revealed the problem on certain hardware.

bgpeyton:
agreed with @lothian -- whoever picks this up, please leave the defaults the same (or else a lot of EOM calculations will suddenly get much harder for no good reason), but the user definitely should be able to crank up the precision if they really want to. may be useful to separate the convergence criteria for the t's and lambdas, if not already done.

JonathonMisiewicz:
> the default settings we chose are not to avoid breaking tests but because tighter settings are wholly unnecessary (and often exceedingly difficult to achieve) for transition properties.

> may be useful to separate the convergence criteria for the t's and lambdas, if not already done.

The running plan is to rename `r_convergence` to `lambda_convergence` or somesuch within `cclambda`, with some logic to auto-set `lambda_convergence` is that is _not_ set but `r_convergence` _is_. `occ` and `dfocc` use a similar trick.

For context, "exceedingly difficult to achieve" means [you're trying to solve ill-conditioned equations](https://github.com/psi4/psi4/issues/2519#issuecomment-1089612589). I'm expecting to be the one who picks this up, and when the time comes, I'll do a quick convergence study to double-check the sensitivity of properties to this parameter.

lothian:
FYI, we did do a brief convergence study, which is how we came up with the 1e-4 criterion.

JonathonMisiewicz:
I figured as much. I'm just paranoid.

Issue #2519, opened 2022-03-30
Use `DIISManager` in `cc`


JonathonMisiewicz:
The `cc` module [has](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/ccenergy/diis.cc) [a](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/ccenergy/diis_RHF.cc) [habit](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/ccenergy/diis_ROHF.cc) [of](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/ccenergy/diis_UHF.cc) [rewriting](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/cclambda/diis.cc) [DIIS](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/cc/ccresponse/diis.cc) when they really should [just use `libdiis`.](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libdiis/diismanager.h)

In addition to code cleanup, I predict an important practical benefit. The central code uses Rob Parrish tricks to treat linearly dependent matrices. In my experience during the DIIS refactor, these often arise close to convergence. So by refactoring this, we can allow for effective DIIS even for tight convergence of the `cc` equations. I have [reason to believe](https://github.com/psi4/psi4archive/commit/2b2eb6c5a710dde180b1ef7a050012069afeae8d) this has a great to deal to do with anomalies in `cc55`.

Don't even think about taking this until after #943 is closed.

lothian:
Please note that `libdiis` did not exist when the DIIS code was written in the `cc` modules.    Convergence difficulties in `cclambda` are not due to DIIS difficulties but due to the fact that linear equations become poorly conditioned for excited states that are relatively close in energy.

JonathonMisiewicz:
That agrees with my knowledge of `libdiis` history: I think it was made in the late 2000's for `dct` purposes.

Noted about the poorly conditioned equations. That possibility didn't even occur to me.

PR #2533, opened 2022-04-05
Remove EOM- from prop name when irrelevant

JonathonMisiewicz:
## Description
PR #2022 labeled _ground-state_ properties with the method name... even when the method name included "EOM-". So even though `ccsd` and `eom-ccsd` compute the same ground state dipole, they weren't given the same name. This PR fixes that.

More fixes to come, but the next PR will be heftier.

Obligatory @bgpeyton ping.

## Checklist
- [x] `eom` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -3168,7 +3168,10 @@ def run_cc_property(name, **kwargs):
     if n_one > 0:
         # call oe prop for GS density
         oe = core.OEProp(ccwfn)
-        oe.set_title(name.upper())
+        # TODO: When Psi is Py 3.9+, transition to the removeprefix version.
+        title = name.upper().replace("EOM-", "")
+        #title = name.upper().removeprefix("EOM-")
+        oe.set_title(title)
         for oe_name in one:
             oe.add(oe_name.upper())
         oe.compute()
--- a/tests/cc46/input.dat
+++ b/tests/cc46/input.dat
@@ -28,21 +28,21 @@ cc_energy, wfn = properties('eom-cc2', properties=['dipole','quadrupole'],return
 
 
 gs_refs = {
-        "EOM-CC2 DIPOLE X":                 0.0000,
-        "EOM-CC2 DIPOLE Y":                 0.0000,
-        "EOM-CC2 DIPOLE Z":                 2.8972,
-        "EOM-CC2 QUADRUPOLE XX":          -11.6671,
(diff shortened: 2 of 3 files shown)


bgpeyton (approved):
ah, was this choking in `OEProp` when you pushed an EOM wfn? or was it computing properly and just giving the redundant name? regardless, good catch.

lothian (approved):

JonathonMisiewicz:
> ah, was this choking in `OEProp` when you pushed an EOM wfn? or was it computing properly and just giving the redundant name? regardless, good catch.

Computing properly and not giving the expected `CCSD DIPOLE`. This led to an error when I fixed a bug in AJ's code. That should be fixed in the next `cc` series PR.

loriab (approved):

tests/cc55/input.dat
```diff
@@ -96,11 +96,11 @@ for transition, data in test_data.items(): # TEST
     for property, value in data.items(): # TEST
         # Change these numbers to 6 ASAP # TEST
         if "EINSTEIN" not in property: # TEST
-            compare_values(value, wfn.variable(f"CC {transition} {property}"), 6, f"CC {transition} {property}") # TEST
-            compare_values(value, wfn.variable(f"CCSD {transition} {property}"), 6, f"CCSD {transition} {property}") # TEST
+            compare_values(value, wfn.variable(f"CC {transition} {property}"), 4, f"CC {transition} {property}") # TEST
```
bgpeyton, line 99: any clue what the above `# Change these numbers to 6 ASAP` is about, here? it's suspect if these oscillator strengths aren't matching to at least 1e-6.

JonathonMisiewicz: It's unjustifiable to check these quantities to 1e-6 when cc [refuses to enforce lambda convergence any tighter than 1e-4](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/proc.py#L3161-L3163). I have plans to fix this. [See](https://github.com/psi4/psi4/issues/2519) [`cclambda`'s](https://github.com/psi4/psi4/issues/2520) [sin](https://github.com/psi4/psi4/issues/25221) [list](https://github.com/psi4/psi4/issues/2522).

But that scale of refactoring is outside the scope of this review.

bgpeyton: wow. that is highly irregular, and definitely a problem. I'll be looking forward to these refactors. thanks!

PR #2536, opened 2022-04-07
forward-port ddd convergence and derivative logic

loriab:
## Description

This is No. 7 of the DDD series, #1351.

## Todos
- [x] test both old `run_json` and new `run_qcschema` modes and access patterns in the basic addons testing.
- [x] I don't remember in what order it came about, but `driver_util.py` had a function `negotiate_derivative_type` that was a simpler version (returned `analytic` and `1_0` for analytic and findif gradient, respectively) of the trio of functions in ddd (returns `(1, 1)` and `(1, 0)` for analytic and findif gradient, respectively) and was tested in `test_dertype.py` and was never run. This unused fn and its testing was removed and replaced by the ddd version.
- [x] the new `negotiate_derivative_type` function probes managed methods earlier than before, so it happens that the exact error names (not the text) expected by stdsuite change.
- [x] the new `negotiate_derivative_type` function is called early, so it now handles the methodname typo helper ("Did you mean?") that FAE added. The independent `_method_exists()` fn removed.
- [x] the former function `_set_convergence_criterion` where you passed in 5 integers for the different E/D_CONVERGENCE values has been rewritten as `negotiate_convergence_criterion` where you pass in a tuple of the case (e.g., `(2, 1)` is Hessian by findif of gradients). In actual `driver.py`, the function calls have been switched out. The values set are the same, but the arguments like `(1, 0)` aren't necessarily appropriate in this non-ddd driver code routing.
- [x] the former function `_find_derivative_type` has been replaced by three functions headed up by `negotiate_derivative_type`. For both, the goal is to take a method name, the target derivative, and any user modifications and return the appropriate analytic or findif strategy or produce a helpful error. sounds simple, but managed methods and cbs specs complicate things. In ddd, this function is called early, not piecemeal throughout the driver, so more probing has been consolidated into the new function. In actual `driver.py`, the function calls have been switched out, but it's only using half the info due to differences in non-ddd driver code routing. To mirror ddd behavior, the function is called even for energy() and properties(), mostly to take the place of `_method_exists()`.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
PIng me when merge conflicts are fixed, and I'll review it.

loriab:
merge conflicts now repaired, @JonathonMisiewicz

tests/pytests/test_dertype.py
```diff
@@ -132,9 +139,10 @@ def test_negotiate_missing_error(inp):
     (('gradient', 'Eg_man', 2)),
     (('energy', 'Eg_man', 2)),
     (('energy', 'Eg_man', 1)),
+    #(("properties", "hf", 2)),
```
JonathonMisiewicz, line 142: Why is this commented?

psi4/driver/driver_util.py
```diff
@@ -282,50 +252,102 @@ def parse_cotton_irreps(irrep: Union[str, int], point_group: str) -> int:
...
+        When input validation fails. When `user_dertype` exceeds `target_dertype`.
     MissingMethodError
         When `method` is unavailable at all. When `user_dertype` exceeds what available for `method`.
 
     """
-    egh = ['energy', 'gradient', 'hessian']
+    if target_dertype == "properties":
+        if isinstance(method, list):
+            # untested
+            analytic = [highest_analytic_derivative_available(mtd, proc) for mtd in method]
```
JonathonMisiewicz, line 299: Do you mean `highest_analytic_properties_available`?

psi4/driver/driver_util.py
```diff
@@ -282,50 +252,102 @@ def parse_cotton_irreps(irrep: Union[str, int], point_group: str) -> int:
...
+    proc: Optional[Dict] = None,
+) -> Union[Tuple[int, int], Tuple[str, str]]:
     r"""Find the best derivative level (0, 1, 2) and strategy (analytic, finite difference)
-    for `method` to achieve `ptype` within constraints of `user_dertype`.
+    for `method` to achieve `target_dertype` within constraints of `user_dertype`.
 
     Parameters
     ----------
-    ptype : {'energy', 'gradient', 'hessian'}
+    target_dertype: {"energy", "gradient", "hessian", "properties"}
```
JonathonMisiewicz, line 267: In `sort_derivative_type`, `properties` does not count as a dertype. Here it does. Is it possible/advisable to get these to agree on whether `properties` is a dertype or not?

psi4/driver/driver_util.py
loriab: quite right, now fixed. `properties("cbs")` has never been attempted, so this is just prep.

psi4/driver/driver_util.py
loriab: Not really, at least right now. The trio of functions is really suited for negotiating e/g/h; "properties" was forcibly inserted a couple weeks ago so that `properties("ccsd/cc-pvdz")` (slash invokes cbs machinery) could run. So that's why the top-level fn `negotiate_derivative_type` acknowledges properties, while the bottom-level fn `sort_derivative_type`, which doesn't get called on the properties route, doesn't deal in properties. I think the whole driver would have to be more fine-grained in how it offers what properties are available per method rather than the on/off of `procedures["properties"][method]` to justify tighter integration of properties into these probing functions. It could be nice in future, but for now I'm content with this basic working integration.

tests/pytests/test_dertype.py
loriab: I couldn't decide whether it or I was right when it didn't give me the error I expected. So it's commented as a reminder that that error pathway isn't tested. Not a great situation, but line is pretty harmless.

JonathonMisiewicz: If we can get a comment briefly explaining this? I know, you're on the odds-on favorite to probe deeper here, but just in case...

JonathonMisiewicz (approved):

tests/pytests/test_dertype.py
loriab: sure, done locally.

jturney (approved):

zachglick (approved):
LGTM! I appreciate the well-documented driver functions

PR #2540, opened 2022-04-08
more installed windows path adjustments

loriab:
## Description
believe it or not, every "fix windows conda package" commit in the past week has actually fixed the problem at hand. for this round, let's try installing databases on windows and anchoring scripts from module.

EDIT:
- [x] confirmed to actually fix windows conda-build
- [x] pytest 7 now on defaults, so let's try lifting constraints.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-  #pull_request:
+  pull_request:
 
 jobs:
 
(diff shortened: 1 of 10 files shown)


.github/workflows/ecosystem.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-  pull_request:
+  #pull_request:
```
JonathonMisiewicz, line 7: No longer running ecosystem on PRs, only on push, makes sense to me, but is this intentional?

loriab: this i'll revert. I wanted to minimize CI since it was really only the Win Azure lane failing.

JonathonMisiewicz (approved):

zachglick (approved):

maxscheurer (approved):

PR #2541, opened 2022-04-09
Renaming of excited states dipoles

JonathonMisiewicz:
## Description
More `cc` tests ported over. There's one `autotest` test left.

## Todos
- [x] Documented several CC psivars
- [x] Updated  dipole size managing functions to work with new "DIPOLE -" syntax
- [x] cc density variables renamed to new syntax
- [x] cc dipole/quadrupole variables renamed to new syntax  

## Checklist
- [x] ctests pass

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/procrouting/proc.py
```diff
@@ -3165,36 +3165,56 @@ def run_cc_property(name, **kwargs):
...
+                    if h == gs_h: i += 1
+                    root_title = title + f" ROOT {i} ({root_h_lbl})"
+                    oe.set_title(root_title)
+                    total_idx = ccwfn.total_index(i, h)
+                    set_of_names = {f"{title} ROOT {total_idx} {{}}", f"CC ROOT {total_idx} {{}}",
+                                    f"{title} ROOT {total_idx} {{}} - {trans_h_lbl} TRANSITION",
+                                         f"CC ROOT {total_idx} {{}} - {trans_h_lbl} TRANSITION",
+                                    f"{title} ROOT {i} ({root_h_lbl}) {{}}", f"CC ROOT {i} ({root_h_lbl}) {{}}"}
+                    oe.set_names(set_of_names)
+                    Da = ccwfn.variable(root_title + " Da")
```
loriab, line 3212: Are the density qcvars only for intracode communication, or are there definitely reasons to expose them to the user? There's ordering and size considerations against keeping them around.

JonathonMisiewicz: Not for intracode communication. I didn't see including them as controversial, given that we keep ground-state densities around.

loriab: I'm all for keeping them around, just not in qcvars, since those get exported and printed thoughtlessly. I'd favor a different member data on Wfn. Others thoughts?

JonathonMisiewicz:
> I'm all for keeping them [excited-state densities] around, just not in qcvars, since those get exported and printed thoughtlessly. I'd favor a different member data on Wfn. Others thoughts?

I hope you know what you're asking for. Storage of ground-state densities is lawless, let alone excited-state densities.

For excited-state densities...
* ADC won't export densities of any sort.
* TD-DFT will save left and right eigenvectors. From my (very shaky) knowledge of TD-DFT theory, those are left and right transition densities.
* EOM-CC creates but will not save its left and right transition densities. Before this PR, it would write excited state densities to the wavefunction. (In the current version of the PR, it still does)

For ground-state densities...
* For fully variational methods (including in orbitals), the density of the correlated method is unambiguous and saved to the wavefunction.
* For methods that are not fully variational, you have orbital-relaxed and orbital-unrelaxed densities. If orbital relaxation is not needed, this density may or may not get saved to the wavefunction. If it is not saved, the density is probably still the SCF density.
* Some methods have it even worse. For example, ODC-12 with perturbative lambda has _three conceivable densities_. ODC-12 density, ODC-12 density + perturbative lambda, and ODC-12 + perturbative lambda + orbital relaxtion.
* CAS-DMRG-PT2 through the `dmrg` module saves the DMRG density to the wavefunction, which is not the DMRGPT2 density.
* If the user requests a spin-scaled density, the standard densities need to be further monkeyed with to be correct in gradients. I don't know if this is actually done. I suspect it is not.

And to represent all that, the current tools we have are `Da`, `Db`, and the pool of psivars.

I'm in favor of deprecating `Da` and `Db` for anything other than internal SCF use and creating a new wavefunction member to store all the densities, _with proper labeling_. This will of course be a lot of code refactring, but that is quite literally the entire point of me doing all this `cc` PRs. And of course, somebody will need to tell `dfocc` about this API change. 🙂

loriab:
> I'm in favor of deprecating Da and Db for anything other than internal SCF use and creating a new wavefunction member to store all the densities, with proper labeling. This will of course be a lot of code refactoring, but that is quite literally the entire point of me doing all this cc PRs. And of course, somebody will need to tell dfocc about this API change. 🙂

Thanks for the enumeration of the organizational state of all the densities. The new Wfn member for finalized, labeled densities sounds good to me. Are the ES CC densities ready for that state yet, or are they still in preparation? For any that are still in preparation but still in need of saving, I guess I propose a temporary member of Wfn that is marked as having a finite lifetime. Same as Wfn.arrays, just not Wfn.arrays :-) Ben's already nervous about storing the per-atom charges that end up in qcvars, so I'm just not in favor of densities there.

JonathonMisiewicz:
The CC excited state densities are most definitely correct. If they _weren't_, EOM analytic gradients would be wrong, and `cc23` would fail. I would expect that correctness is enough to finalize these, but I defer to our resident Crawdad on this.

I don't know whether the transition densities pass some kind of finite difference test, but I am prepared to save that question for another day.

lothian:
Can you clarify what you mean about a "finite difference test" for the transition densities?

JonathonMisiewicz:
By finite-difference test, I mean "there is some property that we can compute either by finite difference of energies or by contracting appropriately defined densities against derivative integrals," e.g., geometry gradients and dipoles. By checking that both routes predict the same result, we can be much more confident that the densities are correctly implemented. (I recall a sign error in the CASPT2 gradients of another package. This error went uncaught for decades because the impact on calculations was relatively small.)

I don't know if this is an option for transition densities. I'm not sure if EOM-CC transition properties are defined by some variational criteria, some variational criteria but neglecting orbital relaxation, or something else altogether.

lothian:
OK, then we're on the same page.  Transition properties aren't clearly defined as derivatives, so I don't see how a finite-difference approach can help in this case, but we might be able to identify some limiting cases.  But I agree it can/should wait for another day.

tests/cc48/input.dat
```diff
@@ -99,4 +99,86 @@ D 2 1.00
...
+ref1A1 = -75.913588662964 # TEST
+ref2A1 = -75.859480750323 # TEST
+ref0A2 = -75.938133885345 # TEST
+ref1A2 = -75.843040508395 # TEST
+ref0B1 = -76.002650024553 # TEST
+ref1B1 = -75.886857034080 # TEST
+ref0B2 = -75.849495233717 # TEST
+ref1B2 = -75.776029952037 # TEST
+
+compare_values(ref1A1, core.variable("CCSD ROOT 1 (A1) TOTAL ENERGY"), 6, "Root 1 (A1) Energy") # TEST
```
loriab, line 128:
```suggestion
compare_values(refccsd, core.variable("CCSD ROOT 0 (A1) TOTAL ENERGY"), 6, "Root 0 (A1) Energy") # TEST
compare_values(ref1A1, core.variable("CCSD ROOT 1 (A1) TOTAL ENERGY"), 6, "Root 1 (A1) Energy") # TEST
```
maybe throw in GS test

loriab (commented):
apart from lodging the root densities outside Wfn.arrays, lgtm!

psi4/driver/procrouting/proc.py
```diff
@@ -3165,36 +3165,56 @@ def run_cc_property(name, **kwargs):
...
+                    oe.set_Da_so(Da)
+                    if core.get_global_option("REFERENCE") == "UHF":
```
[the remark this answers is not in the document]
JonathonMisiewicz: That was legacy code. I prefer `same_a_b_dens` wherever possible, so :+1:.

tests/cc46/input.dat
```diff
@@ -56,31 +56,31 @@ root_refs = {
...
-        "CC ROOT 4 QUADRUPOLE XY":    -0.0453,
+        "CC ROOT 4 DIPOLE Z":          1.7559,
+        "CC ROOT 4 QUADRUPOLE XX":   -12.1976,
+        "CC ROOT 4 QUADRUPOLE XY":    -0.9808,
         "CC ROOT 4 QUADRUPOLE XZ":     0.0000,
-        "CC ROOT 4 QUADRUPOLE YY":   -13.5795,
+        "CC ROOT 4 QUADRUPOLE YY":   -13.9980,
         "CC ROOT 4 QUADRUPOLE YZ":     0.0000,
-        "CC ROOT 4 QUADRUPOLE ZZ":   -15.3582
+        "CC ROOT 4 QUADRUPOLE ZZ":    -9.7604,
```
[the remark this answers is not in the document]
JonathonMisiewicz: Depends on how you defined scrambled. 🙃 

The root indexing _now_ means _lowest to highest energy_. The root indexing _previously_ meant the order in which `eomcc` computed them, which was _irrep first, then lowest to highest energy_. (I'd need to check the code if it was transition irrep first or target irrep first, but they're the same for this test.)

The new indexing is consistent with the previously discussed CC standard and is much more intuitive.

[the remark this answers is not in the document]
loriab: yikes. new way much better. there's already root numbering vagueness by method and by roots requested. I'm glad to see the exit of vagueness by character table.

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -735,6 +734,17 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction> {
...
+
+    /// The below members are experimental and are designed to hold densities when the
+    /// "current density" is ambiguous, e.g., non-orbital optimized methods and multi-
+    /// stage methods. ~ JPM - Apr. '22
+    /// This is public because `ccdensity` doesn't subclass wfn like it should, so we need
+    /// SOME way to let it get/set.
+    /// Vector of alpha density matrices
+    std::map<std::string, SharedMatrix> Da_map_;
+    /// Vector of beta density matrices
+    std::map<std::string, SharedMatrix> Db_map_;
```
loriab, line 746: if `Da_map_` and `Db_map_` become the long-term holders, should probably add some of the `key.upper()` features of variables and arrays, but that's for the beyond experimental stage.

loriab (approved):

lothian (approved):

andysim (approved):
I'm not really qualified to appraise this, but I don't see any obvious errors.

JonathonMisiewicz:
> I'm not really qualified to appraise this, but I don't see any obvious errors.

Alas, nobody is fully qualified to appraise this. TDC knows the `cc` code and Lori knows psivars, but densities touch so many parts of the code that we'd need expertise in every single module, which we simply don't have.

I won't be free to work on the next PR for a while, so we can leave this open to collect more comments.

hokru (approved):

PR #2359, opened 2021-11-19
Linear Exchange Matrix Build (LinK) in Direct SCF algorithm

andyj10224:
## Description
Updated version of #2187 

Implements a linear-scaling variant of building the K-matrix by taking advantage of the sparsity of ERIs combined with the Density matrix using advanced prescreening techniques.

Reference: C. Ochsenfeld, C.A. White, and M. Head-Gordon, J. Chem. Phys. 109, 1663 (1998)
(https://aip.scitation.org/doi/10.1063/1.476741)

## Todos

- [x] Faster build of K matrix based on pre-screening techniques
- [x] Verified timings and accuracy of calculations on 3D system (watercluster)

## Questions
- [x] My algorithm is slightly modified from the Oschenfeld version since we perform our DirectJK algorithm in atom-task pairs instead of shell pairs, in order to preserve the old DirectJK loop. Is that a bad idea?

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)
- [x] Updated Docs and Bibliography for new feature

## Status
- [x] Ready for review
- [x] Ready for merge

## Table of Contents
- [LinK Pseudocode](https://github.com/psi4/psi4/pull/2359#issuecomment-998996631)
- [LinK Timings Graphs](https://github.com/psi4/psi4/pull/2359#issuecomment-1003741024)
- [LinK Code Organization](https://github.com/psi4/psi4/pull/2359#issuecomment-1043550653)
- [Parallel Preprocessing Code](https://github.com/psi4/psi4/pull/2359#issuecomment-1063147739)
- [Atom vs Shell Blocking](https://github.com/psi4/psi4/pull/2359#issuecomment-992619129)

PR #2543, opened 2022-04-11
Separate J and K Build Capabilities, DirectDFJ Algorithm, and Refactor of LinK

andyj10224:
## Description
Follow-up PR to #2359. This PR implements the direct-density fitted J algorithm (https://pubs.rsc.org/en/content/articlelanding/2002/cp/b204199p) to be used with LinK to allow for performance gains. A new code infrastructure is also introduced that allows for split J/K build algorithms, through the Composite JK class (not callable by user, but a reroute by the developer).

Example of how to call my new code:
```
scf_type direct_df_linK
```

### NOTES:
- Composite JK will be extended with other Split JK algorithms in the future (CFMM @andyj10224, and Chain-of-Spheres K @zachglick)

## Todos

- [x] Introduction of Composite JK class
- [x] Implementation of Direct-DF-J algorithm
- [x] Refactor of LinK algorithm into Composite JK framework

## Questions
- [x]

## Checklist
- [x] Docs added for any new features

## Status
- [x] Ready for review
- [ ] Ready for merge

### 1.) LinK Refactor
- [Removal of LinK from DirectJK](https://github.com/psi4/psi4/pull/2543#discussion_r848777136)
- [LinK Class Header Information](https://github.com/psi4/psi4/pull/2543#discussion_r848776564)
- [New Home of LinK Algorithm](https://github.com/psi4/psi4/pull/2543#discussion_r848777810)

### 2.) DirectDFJ Algorithm
- [DirectDFJ Header File](https://github.com/psi4/psi4/pull/2543#discussion_r848778214)
- [Home of DirectDFJ Algorithm](https://github.com/psi4/psi4/pull/2543#discussion_r848778585)

### 3.) CompositeJK Structure
- [CompositeJK Header Information](https://github.com/psi4/psi4/pull/2543#discussion_r848779185)
- [SplitJK Base Class Header Information](https://github.com/psi4/psi4/pull/2543#discussion_r848779584)
- [Calling my code from jk.cc](https://github.com/psi4/psi4/pull/2543#discussion_r848780442)


JonathonMisiewicz:
I will not review this until both tests pass and somebody from GATech (Zach, Lori, David Poole) does a thorough review pass.

I have no strong opinion on whether this comes in for 1.6 or 1.7, but if this PR doesn't make 1.6, there needs to be a separate PR which implements the API changes. That PR _must_ come in for 1.6.

psi4/driver/procrouting/proc.py
```diff
@@ -1198,7 +1198,8 @@ def scf_wavefunction_factory(name, ref_wfn, reference, **kwargs):
 
     # Set the DF basis sets
     if (("DF" in core.get_global_option("SCF_TYPE")) or
-            (core.get_option("SCF", "DF_SCF_GUESS") and (core.get_global_option("SCF_TYPE") == "DIRECT"))):
+            (core.get_option("SCF", "DF_SCF_GUESS") and (core.get_global_option("SCF_TYPE") == "DIRECT")) or
+            (core.get_global_option("SCF_TYPE") == "COMPOSITE" and core.get_option("SCF", "J_TYPE") == "DIRECT_DF")):
```
loriab, line 1202: I think from a group meeting that arbitrary composites weren't planned -- only specific ones were to be allowed or encouraged? Since dozens of new combinations aren't foreseen, I'd recommend allowable ones be given their own JK scf_type (e.g., `DIRECT_DF_LINK`). 
* user only needs to set one option, not three, to access
* if you need further options for internal routing, that's fine, like https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L3221-L3224
* you won't need to program defensively for ways a user can mess up: e.g., j_type=direct_df, k_type=link, no scf_type so defaults to pk, so now you've got no fitting bases attached
* further downstream, when figuring out if a correlated method can be built atop the scf-composite reference or if a gradient method can be, it'll be easier if describable in one label.

Glad to discuss further, especially if this sounds all wrong for current plans.

andyj10224: Ah, that makes sense. I think we can keep the 'JBase' and 'KBase' classes, but instead of composite JK, we can make wrapper JK classes that call `Direct_DF_J` and `LinK`. I think a better solution is to keep the CompositeJK structure, but not allow the user to set `scf_type` to composite. For example, if user sets scf_type to `Direct_DF_LinK`, a Composite JK will be built with direct_df_j and linK

loriab: Sounds good. I see the benefit in the compositeJK structure in the JK code, so long as the user has simple access.

psi4/src/psi4/libfock/jk.h
```diff
@@ -1191,7 +1171,143 @@ class PSI_API MemDFJK : public JK {
     std::shared_ptr<DFHelper> dfh() { return dfh_; }
 };
 
+class JBase {
```
jturney, line 1174: `JBase` and `KBase` do the exact same thing and therefore code duplication. What about keeping just one of them and renaming it to `JorKBase` or something similar?

psi4/driver/procrouting/proc.py
andyj10224: Got rid of explicit calling of `CompositeJK`, to call Direct DF J and LinK, user sets `scf_type` to `DIRECT_DF_LINK`

psi4/src/psi4/libfock/jk.h
andyj10224: Made the change. Two classes are now merged in `SplitJKBase`

andyj10224:
Made all the necessary code changes. Have not had time to update docs yet (classes). Feel free to review the code, and I will update the docs tomorrow.

andyj10224:
Updated the docs, please proceed with review @zachglick @davpoolechem

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
+#include "composite.h"
+#include "jk.h"
```
davpoolechem, line 2: Since jk.h is included in composite.h, we shouldn't need the jk.h include here.

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
...
+
+#include "psi4/libmints/integral.h"
+#include "psi4/libmints/vector.h"
+#include "psi4/lib3index/dftensor.h"
```
davpoolechem, line 6: Also here, since dftensor.h is included in composite.h, we shouldn't need the dftensor.h include here.

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
...
+#include "psi4/libqt/qt.h"
+#include "psi4/libpsi4util/PsiOutStream.h"
+
+#include <vector>
+#include <unordered_set>
```
davpoolechem, line 11: Also here, since unordered_set is included in composite.h, we shouldn't need the unordered_set include here.

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
...
+        throw PSIEXCEPTION("K BUILD TYPE " + ktype_ + " IS NOT SUPPORTED IN COMPOSITE JK!");
+    }
+}
+
+size_t CompositeJK::memory_estimate() {
+    return 0;   // only Direct-based integral algorithms are currently implemented in Composite JK
+}
+
+void CompositeJK::preiterations() {}
+void CompositeJK::postiterations() {}
```
davpoolechem, line 66: If preiterations() and postiterations() are empty, it might be better to put "return;" in them to make it more clear that they are essentially empty functiopns doing nothing.

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
...
+        }
+    }
+
+    // Solve for gammaQ, (P|Q) * gammaQ = gammaP
+    SharedMatrix Jmet_copy = Jmet_->clone();
+
+    std::vector<int> ipiv(naux);
+    C_DGESV(naux, nmat, Jmet_copy->pointer()[0], naux, ipiv.data(), gamp, naux);
+
+#pragma omp parallel for num_threads(nthread_) schedule(dynamic)
```
davpoolechem, line 281: Two points regarding the parallelism in this algorithm:
1. The first loop, the one forming gammaP, parallelizes over auxiliary shells, while this second loop, the one forming J, parallelizes over one of the regular shell indices u. Why are they different?
2. If rearranging how the loops are set up isn't a big problem (likely depending on the answer to point #1), it might be more efficient from a parallel perspective to parallelize over the combined regular shell pair index uv. If that is too fine grained (though I suspect it won't be simply because each uv loop also would loop over the auxiliary shells), it's easy enough to simply set a batch size in each OpenMP parallel for pragma.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -179,6 +179,12 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
             return build_JK(primary, auxiliary, options, "DISK_DF");
         }
 
+    } else if (jk_type == "DIRECT_DF_LINK") {
+        CompositeJK* jk = new CompositeJK(primary, auxiliary, "DIRECT_DF", "LINK", options);
+        // Composite JK algorithms should use Incremental Fock build by default
+        if (!options["INCFOCK"].has_changed()) jk->set_incfock(true);
+
+        return std::shared_ptr<JK>(jk);
```
davpoolechem, line 187: A thing I want to bring to the attention of reviewers here, that Andy and I discussed yesterday:
1. This entire formation of the CompositeJK shared pointer could more cleanly be done with a call to std::make_shared. However, the method as implemented was chosen because other JK shared pointers (e.g., DirectJK), were built in this manner - we considered it a matter of convention to do the construction of the CompositeJK shared pointer the way we did. So, my question is, would it fall outside the scope of this PR to use std::make_shared for the construction of the JK shared pointer, and change the other JK shared pointers to do the same?

psi4/src/psi4/libfock/jk.cc
```diff
@@ -537,6 +550,88 @@ void JK::AO2USO() {
 }
 void JK::initialize() { preiterations(); }
 
+void JK::incfock_setup() {
```
davpoolechem, line 553: It might be better to have JK::incfock_setup() and JK::incfock_postiter() throw a PSIEXCEPTION since incfock is basically meaningless outside of DirectJK/CompositeJK, then implement the current definition of these functions into DirectJK and CompositeJK. On one hand, it's a bit of code duplication, so that's unfortunate. On the other hand, it clearly delineates which JK subclasses are meant to use these functions.

psi4/src/psi4/libfock/jk.h
```diff
@@ -1191,7 +1175,118 @@ class PSI_API MemDFJK : public JK {
     std::shared_ptr<DFHelper> dfh() { return dfh_; }
 };
 
+/**
+ * Class SplitJKBase
+ *
+ * A wrapper class that allows building of different J and K algorithms
+ * (such as Direct-DF J builds, and LinK builds)
+ */
+class SplitJKBase {
```
davpoolechem, line 1184: Would it be better for the class declaration of SplitJKBase to be in composite.h instead of jk.h? That way, jk.h is reserved for the JK class and its subclasses, and composite.h contains all references to split J/K algorithms. Plus, the SplitJKBase constructor is defined in composite, so moving the SplitJKBase class declaration cleans up the code a bit.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -179,6 +179,12 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
             return build_JK(primary, auxiliary, options, "DISK_DF");
         }
 
+    } else if (jk_type == "DIRECT_DF_LINK") {
+        CompositeJK* jk = new CompositeJK(primary, auxiliary, "DIRECT_DF", "LINK", options);
```
davpoolechem, line 183: Something I want to bring up to the reviewers, that Andy and I discussed yesterday:

Right now, we have the singular CompositeJK type "DIRECT_DF_LINK", with hardcoded J type and K type inputs into CompositeJK. Eventually, given the number of combinations of J and K there will be, it will make the code more extensible to parse the jk_type string using Python tools, extract the J and K types from the jk_type variable, and pass the J and K types in as arguments to the construction of J and K. So, this comes with two questions:
1. Would it be better to implement such a parsing mechanism now, or later on, when more types have been added to CompositeJK?
2. When we implement the parsing mechanism for jk_type, it will be necessary to use characters that clearly separate the J and K types in the jk_type variable; for example, a plus sign could be used, so jk_type would look like "DIRECT_DF+LINK", How do the reviewers feel about using such characters in the SCF_TYPE option?

loriab: I agree with dissecting the jk_type in Python. No problems anticipated with `+` as delimiter (it's no worse than `/` being the cue for "cbs" or "method/basis" runs). If you don't run into needing the parsing now, I think I'd wait for another case before implementing it -- that'll keep this PR shorter and provide a better test that you're switching as intended.

zachglick (commented):
Nice work Andy! As I see it, this PR does three things:

1. Adds a direct DFJ algorithm
2. Moves the LinK code out of the DirectJK class
3. Adds individual J/K classes in order to prevent future code duplication

Because these additions are fairly independent, I think it would be appropriate to break up this PR into two separate PRs. In the first PR, you do steps (1) and (2) by implementing a `DFJlinK` class which is derived from `JK`, just like all of the existing JK algorithms. In the second PR, you refactor the `DFJlinK` class so that it uses a `DFJ` and `LinK` class.

Also before any of this is done, I think it would be good to address the changes to density-screening we discussed last week.

psi4/src/psi4/libfock/jk.h
davpoolechem: That makes sense! Such a design would also keep in line with the current Psi4 JK class design philosophy, where JK implementations are kept in their own separate cc files.

andyj10224:
> Nice work Andy! As I see it, this PR does three things:
> 
> 1. Adds a direct DFJ algorithm
> 2. Moves the LinK code out of the DirectJK class
> 3. Adds individual J/K classes in order to prevent future code duplication
> 
> Because these additions are fairly independent, I think it would be appropriate to break up this PR into two separate PRs. In the first PR, you do steps (1) and (2) by implementing a `DFJlinK` class which is derived from `JK`, just like all of the existing JK algorithms. In the second PR, you refactor the `DFJlinK` class so that it uses a `DFJ` and `LinK` class.
> 
> Also before any of this is done, I think it would be good to address the changes to density-screening we discussed last week.

In the interest of minimizing the work of both me and the reviewers, I do _not_ think it would be a good idea to split this into two PRs, since it would involve adding a new JK class (`DFJLinK`), and then removing it, and then adding one again (`CompositeJK`). This would lead to a whole mess of merge conflicts, and involves me taking apart code that is 100% working, and then rewriting it. The second reason is that we already have reviews on CompositeJK, and to break this into two PRs at this point would also involve extra work for the core developers too. Comments on this are appreciated, but I feel __strongly__ about this opinion.

On the other hand, I do agree that it would be good to discuss the density screening changes first before this PR is merged.

JonathonMisiewicz:
> The second reason is that we already have reviews on CompositeJK, and to break this into two PRs at this point would also involve extra work for the core developers too.

You have it backwards. Most of the work in PR reviewing is reading the code and figuring out what you're trying to do. Seeing "oh, this is just moving code" around is not work for me. Figuring out what parts of your PR are about Composite JK, what parts are about refactoring LinK, and what parts are about Direct-DF-J is a lot of work. Zach's plan does mean more work for you, but a smoother review process.

psi4/src/psi4/libfock/composite.cc
andyj10224: The reason why we do that is to avoid race conditions.

The tensor contractions:
`gamma_P += (P|uv)*D_uv` (would need to block over P to avoid race conditions)
`J_uv += `(uv|Q)*gamma_Q` (would need to block over uv to avoid race conditions)

psi4/src/psi4/libfock/jk.cc
andyj10224: Hm, I am __strongly__ against the code duplication. There is currently __NO__ way incfock is called in non-Direct and non-Composite codes (i.e. those functions are called within another function in DirectJK and CompositeJK)

andyj10224:
> > The second reason is that we already have reviews on CompositeJK, and to break this into two PRs at this point would also involve extra work for the core developers too.
> 
> You have it backwards. Most of the work in PR reviewing is reading the code and figuring out what you're trying to do. Seeing "oh, this is just moving code" around is not work for me. Figuring out what parts of your PR are about Composite JK, what parts are about refactoring LinK, and what parts are about Direct-DF-J is a lot of work. Zach's plan does mean more work for you, but a smoother review process.

I am still not willing to break this into two PRs (due to the fact that I am in finals season and do not have much time). However, what I can do, is point to which parts of the code involve refactoring the LinK, Direct-DF-J, and CompositeJK in the PR description.

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
...
+    auto factory = std::make_shared<IntegralFactory>(primary_, primary_, primary_, primary_);
+    ints_.push_back(std::shared_ptr<TwoBodyAOInt>(factory->eri()));
+    for (int thread = 1; thread < nthread_; thread++) {
+        ints_.push_back(std::shared_ptr<TwoBodyAOInt>(ints_[0]->clone()));
+    }
+
+    timer_off("LinK: Build Ints");
+}
+
+// To follow this code, compare with figure 1 of DOI: 10.1063/1.476741
```
andyj10224, line 388: The old LinK code is now here.

psi4/src/psi4/libfock/composite.h
```diff
@@ -0,0 +1,128 @@
...
+#include "psi4/libmints/onebody.h"
+#include "psi4/libmints/potential.h"
+#include "psi4/libmints/twobody.h"
+#include "psi4/lib3index/dfhelper.h"
+
+#include <unordered_set>
+
+namespace psi {
+
+class DirectDFJ : public SplitJKBase {
```
andyj10224, line 44: Here is the DirectDFJ header file

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,796 @@
...
+    auto rifactory = std::make_shared<IntegralFactory>(auxiliary_, zero, primary_, primary_);
+    ints_.push_back(std::shared_ptr<TwoBodyAOInt>(rifactory->eri()));
+    for (int thread = 1; thread < nthread_; thread++) {
+        ints_.push_back(std::shared_ptr<TwoBodyAOInt>(ints_[0]->clone()));
+    }
+
+    timer_off("DirectDFJ: Build Ints");
+}
+
+void DirectDFJ::build_G_component(const std::vector<SharedMatrix>& D, std::vector<SharedMatrix>& J) {
```
andyj10224, line 160: Here is where the DirectDFJ algorithm is implemented

psi4/src/psi4/libfock/composite.cc
davpoolechem: Actually, scratch this one. It was dfhelper.h in composite.h, not dftensor.h. My bad!

psi4/src/psi4/libfock/jk.h
```diff
@@ -1191,7 +1175,118 @@ class PSI_API MemDFJK : public JK {
...
+
+};
+
+/**
+ * Class CompositeJK
+ *
+ * JK implementation allowing mixing and matching of different J and K algorithms
+ * (such as Direct-DF J builds, and LinK builds)
+ */
+class PSI_API CompositeJK : public JK {
```
andyj10224, line 1235: Here is the header information for `CompositeJK`.

psi4/src/psi4/libfock/jk.h
andyj10224, line 1184: Here is the header information for the base class in which split J/K algorithms would inherit from

psi4/src/psi4/libfock/jk.cc
```diff
@@ -179,6 +179,12 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
             return build_JK(primary, auxiliary, options, "DISK_DF");
         }
 
+    } else if (jk_type == "DIRECT_DF_LINK") {
```
andyj10224, line 182: This is how a user would call my new code `scf_type direct_df_linK`

andyj10224:
To aid the review process on this PR, I have made a table of contents in the PR description. Please proceed with review.

psi4/src/psi4/libfock/composite.cc
andyj10224: Yep, got a nice little compiler error :)

psi4/src/psi4/libfock/jk.cc
andyj10224: Cleaned it up :)

psi4/src/psi4/libfock/jk.h
andyj10224: Before I make that change, I would like reviews on the DirectDFJ algorithm

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,794 @@
...
+
+                std::vector<double> Dbuff(num_u * num_v, 0.0);
+                for (int u = u_start; u < u_start + num_u; u++) {
+                    int du = u - u_start;
+                    for (int v = v_start; v < v_start + num_v; v++) {
+                        int dv = v - v_start;
+                        Dbuff[du * num_v + dv] = Dp[u][v];
+                    }
+                }
+                C_DGEMV('N', num_p, num_u * num_v, prefactor, (double *)Puv, num_u * num_v, Dbuff.data(), 1, 1.0, &(gamp[i * naux + p_start]), 1);
```
davpoolechem, line 268: Does Puv need the cast to double * here? Since it defined with a const_cast to type double * when it is declared earlier in the code.

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,794 @@
...
+                            int du = u - u_start;
+                            for (int v = v_start; v < v_start + num_v; v++) {
+                                int dv = v - v_start;
+                                JTp[du * num_v + dv] += prefactor * (*Quv) * gamp[i * naux + q];
+                                Quv++;
+                            }
+                        }
+                    }
+                    */
+                    C_DGEMV('T', num_q, num_u * num_v, prefactor, (double *) Quv, num_u * num_v, &(gamp[i * naux + q_start]), 1, 1.0, JTp, 1);
```
davpoolechem, line 324: Same thing as before - does Quv need the cast to double * here?

psi4/src/psi4/libfock/composite.cc
```diff
@@ -0,0 +1,794 @@
...
+        }
+    }
+
+    // Solve for gammaQ, (P|Q) * gammaQ = gammaP
+    SharedMatrix Jmet_copy = Jmet_->clone();
+
+    std::vector<int> ipiv(naux);
+    C_DGESV(naux, nmat, Jmet_copy->pointer()[0], naux, ipiv.data(), gamp, naux);
+
+#pragma omp parallel for num_threads(nthread_) schedule(dynamic)
```
davpoolechem, line 279: Okay, second discussion at this point. I read your comment on my initial comments on parallelism, and after looking at the code, I agree - it is very important to avoid race conditions and unnecessary thread synchronization. However, my second point stands - you can probably gain parallel efficiency by doing some loop fusions. 

From a first glance, for this second loop, you could loop over the combined shell pair index UV instead of simply the shell U. This would make the granularity of the parallelism of this second loop smaller, improving load balance. Additionally, I have a hunch that doing this wouldn't decrease the granularity too much, because each UV loops over the entire set of auxiliary basis set shells. If looping over UV does decrease granularity too much, then the solution is simple - increase the OpenMP batch size. Doing a loop fusion as such would also come with the benefit of eliminating the need to construct shell_partners at all - a small benefit, but it's something.

psi4/src/psi4/libfock/jk.h
davpoolechem: All right, I have added a couple of comments regarding DirectDFJ specifically.

psi4/src/psi4/libfock/composite.cc
andyj10224: I agree, the `shell_partners` and separate loops were an artifact from debugging, and it can go away safely

psi4/src/psi4/libfock/composite.cc
andyj10224: Oops, artifact from old code :)

andyj10224:
After discussion with @davpoolechem and Dr. Sherrill, we have determined that this PR and the density screening cleanups can be reviewed and merged independently. Please proceed with review :)

andyj10224:
Now that seminumerical exchange is in, it would make more sense for me to refactor that code into this framework as well. Waiting on density screening refactor would also be helpful @davpoolechem

PR #2545, opened 2022-04-13
Fix fcidump.py handling of frozen orbitals

brianz98:
## Description
For large systems it's often sensible to freeze HF orbitals, rather than dumping a very large FCIDUMP file, only to be slowly read in again and frozen in the downstream program. However, I noticed that the `fcidump` functionality isn't handling the frozen keywords in the expected way (the MO slice goes from `frzcpi` to `active_mopi`, when it should obviously go to `frzcpi+active_mopi`). I've also added a test in `tests/fcidump/input.dat` to check this. The added test would fail on the current master.

Additionally, for some reason the patch in #2277 stopped working, and in the current master, the 1/0 e integrals are again 1 space wider than the 2e integrals. This patch also fixes it.

## Status
- [x] Ready for review
- [x] Ready for merge


hokru (approved):

zachglick (approved):
Thanks for the bug fix and the test!

loriab (approved):
thanks for finding and fixing the bug!

I see cc-pVDZ covers up to a d-shell. I don't see any particular susceptibility to ordering details in the fcidump format, but if you should know of any larger basis sets or other cases that are prone to mis-generation of fcidump, we'd be glad of correct examples. (background is that we're soon switching solid harmonic ordering from 0, +1, -1, ..., -l to -l, -l+1, ..., l)

JonathonMisiewicz:
Is there any reason not to have merged this in already?

PR #2547, opened 2022-04-14
Density Screening Refactor Part 1: test_erisieve.py Rework

davpoolechem:
## Description
This PR is the first in a series of planned PRs designed to remove density screening from the TwoBodyAOInt object and into the JK object. Having density screening available in TwoBodyAOInt runs the risk of applying density screening to algorithms where density screening doesn't make sense. Thus, it would be a good idea to move the logic of density screening to where it is more correctly applied, i.e., the JK object.

This PR solves two issues simultaneously:

1. The primary purpose of this PR is to change the test_erisieve.py tests to work with the planned future density screening refactor. One issue that moving density screening from TwoBodyAOInt to JK currently brings up, is that it causes the tests on density screening within the pytest test_erisieve.py to fail. These failures occur because test_erisieve.py performs its screening tests directly using an ERI object generated by IntegralFactory. With density screening being removed from the TwoBodyAOInt object, this method of density screening testing can no longer be done. The current PR is designed to address this issue for when the density screening refactor happens. 

The aforementioned issue is addressed by implementing a new variable to the HF wavefunction, computed_shells_per_iter_, which keeps track of the number of shell quartets computed per SCF iteration. The computed_shells_per_iter_ variable is accessible to the user via Python, and thus can be used to conduct screening tests. In this way, density screening tests can be performed without the need for an ERI object.

2. As a bonus from the changes introduced by this PR, the DirectJK algorithm no longer has a need to print computed shell quartet counts to bench.dat. Bench.dat is used exclusively by the DirectJK object to dump the number of shell quartets computed per SCF iteration somewhere. That data is now accessible to the user in a cleaner fashion - it can be accessed through Python, in a manipulatable format.

## Notes
1. Note that the changes in this PR have not been applied to the LinK portion of the DirectJK code. This is intentional, as Andy is planning on moving LinK out of DirectJK entirely, and editing the LinK code within DirectJK would interfere with that. Thus, the changes in this branch will be applied to LinK in a later update.
 
## Todos

- [ X ]  Addition of computed_shells_ member to JK object, which keeps track of number of shells computed during the JK build process.
- [ X ] Addition of computed_shells_per_iter_ member to HF wavefunction objects, which keep track of number of shells computed during each SCF iteration. This information can be accessed by the user via Python.
- [ X ] Modification of density screening tests in test_erisieve.py using the above class changes to allow the tests to run without construction of an ERI object.

## Questions
- [ ] Currently, only the density screening tests in test_erisieve.py use the new computed_shells_per_iter_ framework to test screening. Other tests in test_erisieve.py perform their tests using a generated ERI object. Should use of computed_shells_per_iter_ comparisons be applied to other tests in test_erisieve.py, as well?

## Checklist
- [ X ] Tests added for any new features
- [ X ] Docs added for any new features
- [ X ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [ X ] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
Details about how the integrals were computed should be the province of the JK object, not the HF wavefunction, so I disagree with creating this new variable as described.

Can we instead have `computed_shells_per_iter_` on the JK object and query the JK object, after the HF, for test purposes?

psi4/src/export_fock.cc
```diff
@@ -94,7 +94,8 @@ void export_fock(py::module &m) {
         .def("K", &JK::K, py::return_value_policy::reference_internal)
         .def("wK", &JK::wK, py::return_value_policy::reference_internal)
         .def("D", &JK::D, py::return_value_policy::reference_internal)
-        .def("print_header", &JK::print_header, "docstring");
+        .def("computed_shells_per_iter", &JK::computed_shells_per_iter, "Array containing the number of shells computed (not screened out) during each SCF iteration.")
+	.def("print_header", &JK::print_header, "docstring");
```
JonathonMisiewicz, line 98: Fix whitespace.

psi4/src/psi4/libfock/jk.h
```diff
@@ -344,6 +348,10 @@ class PSI_API JK {
     size_t memory_overhead() const;
     /// Zero out all J, K, and wK matrices
     void zero();
+    /**
+    * Return number of shells computed during the JK build process.
+    */
+    virtual size_t computed_shells();
```
JonathonMisiewicz, line 354: I think `num_computed_shells` would be slightly clearer. (Yes, I can figure out this is a number when I see the return type, but when `auto` is in play, I can't.)

JonathonMisiewicz (commented):
I have two minor code cleanup requests, but LGTM otherwise.

Short PRs are appreciated!

psi4/src/export_fock.cc
davpoolechem: Done!

psi4/src/psi4/libfock/jk.h
davpoolechem: I agree now that you mention it. Done!

jeffschriber (approved):
LGTM, no comments from me

tests/pytests/test_erisieve.py
```diff
@@ -174,39 +174,41 @@ def test_schwarz_vs_density_quartets():
...
+    density_energy, density_wfn = psi4.energy('hf/DZ', return_wfn=True)
+
+    # compare results to expected values
+    schwarz_computed_shells = schwarz_wfn.jk().computed_shells_per_iter()
+    density_computed_shells = density_wfn.jk().computed_shells_per_iter()
+
+    schwarz_computed_shells_expected = [20290, 20290, 20290, 20290, 20290, 20290, 20290, 20290, 20290]
+    density_computed_shells_expected = [13171, 19618, 19665, 19657, 19661, 19661, 19663, 19663, 19663]
+
+    for iter_ in range(0,9):
```
zachglick, line 209: It would be good to explicitly check that the lists of expected and computed shell quartets are the same length (the number of SCF iterations, which has been hardcoded as 9 here).

Better yet, I think there's a `compare_arrays` function that you could call instead of repeated calls to `compare_integers`.

tests/pytests/test_erisieve.py
```diff
@@ -225,49 +227,43 @@ def test_rhf_vs_uhf_screening():
...
-            screen_count_uhf += 1
+    # compare results to expected values
+    rhf_computed_shells = rhf_wfn.jk().computed_shells_per_iter()
+    uhf_computed_shells = uhf_wfn.jk().computed_shells_per_iter()
     
-    assert compare_integers(screen_count_rhf, 19440, 'RHF Density Screened Ints Count, Cutoff 1.0e-12')
-    assert compare_integers(screen_count_uhf, 19440, 'UHF Density Screened Ints Count, Cutoff 1.0e-12')
+    computed_shells_expected = [13171, 19618, 19665, 19657, 19661, 19661, 19663, 19663, 19663]
+
+    for iter_ in range(0,9):
```
zachglick, line 264: Same as above comment

psi4/src/export_fock.cc
```diff
@@ -94,6 +94,7 @@ void export_fock(py::module &m) {
         .def("K", &JK::K, py::return_value_policy::reference_internal)
         .def("wK", &JK::wK, py::return_value_policy::reference_internal)
         .def("D", &JK::D, py::return_value_policy::reference_internal)
+        .def("computed_shells_per_iter", &JK::computed_shells_per_iter, "Array containing the number of shells computed (not screened out) during each SCF iteration.")
```
zachglick, line 97:
```suggestion
        .def("computed_shells_per_iter", &JK::computed_shells_per_iter, "Array containing the number of ERI shell quartets computed (not screened out) during each compute call.")
```

psi4/src/psi4/libfock/jk.h
```diff
@@ -802,6 +816,11 @@ class PSI_API DirectJK : public JK {
     /// Common initialization
     void common_init();
 
+    /**
+    * Return number of shells computed during the JK build process.
```
zachglick, line 820:
```suggestion
    * Return number of ERI shell quartets computed during the JK build process.
```

psi4/src/psi4/libfock/jk.h
```diff
@@ -344,6 +348,10 @@ class PSI_API JK {
     size_t memory_overhead() const;
     /// Zero out all J, K, and wK matrices
     void zero();
+    /**
+    * Return number of shells computed during the JK build process.
```
zachglick, line 352:
```suggestion
    * Return number of ERI shell quartets computed during the JK build process.
```

psi4/src/psi4/libfock/jk.h
```diff
@@ -247,6 +247,10 @@ class PSI_API JK {
     std::vector<bool> input_symmetry_cast_map_;
     /// Use severe screening techniques? Useful in early SCF iterations (defaults to false)
     bool early_screening_;
+    /// Number of shells computed, i.e., not screened out
+    size_t num_computed_shells_;
+    /// Tally of shells computed per SCF iteration
```
zachglick, line 252:
```suggestion
    /// Tally of ERI shell quartets computed per SCF iteration 
```

zachglick (commented):
Looks good

tests/pytests/test_erisieve.py
```diff
@@ -174,39 +174,41 @@ def test_schwarz_vs_density_quartets():
...
+    # run density screening calculation
+    psi4.set_options({ 
+        "scf_type": "direct", 
+        "screening" : 'density', 
+        "df_scf_guess" : False,
+        "integral_package": 'libint2', 
+        "ints_tolerance" : 1e-12,
+        "save_jk": True,
+        "bench" : 1
+    })
```
loriab, line 199: Unless a `psi4.core.clear_options()` has been run in-between, options persist, so only `{"screening": "density"}` technically needed. 

No need for changes -- this is nice and clear -- just FYI.

loriab (approved):

tests/pytests/test_erisieve.py
loriab: good point. `compare_arrays` is legacy name -- you can use `compare_values` directly. It handles float-like, while `compare` handles int-like.

davpoolechem: This is indeed a very good point. I will make these changes!

psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -997,13 +1006,8 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
         }
     }
 
-    if (bench_) {
-        auto mode = std::ostream::app;
-        auto printer = PsiOutStream("bench.dat", mode);
-        size_t ntri = nshell * (nshell + 1L) / 2L;
-        size_t possible_shells = ntri * (ntri + 1L) / 2L;
-        printer.Printf("Computed %20zu Shell Quartets out of %20zu, (%11.3E ratio)\n", computed_shells,
-                        possible_shells, computed_shells / (double)possible_shells);
```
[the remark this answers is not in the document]
davpoolechem: Yes indeed!

tests/pytests/test_erisieve.py
davpoolechem: Hmm, this is good to know. Thanks for informing me!

tests/pytests/test_erisieve.py
davpoolechem: All right, this change should be made now!

davpoolechem:
So I now realize something - we may want to apply some of the benchmarking changes made in this PR to DFJCOSK, as well. It will increase the size of the PR, but the benchmarking changes in this PR currently only extend to DirectJK at the moment. Since DFJCOSK has two methods that it separately benchmarks, it will require a bit of retooling regarding some of the internals of the benchmarking framework. It should not have a significant impact on test_erisieve, however.

Thoughts?

loriab:
> So I now realize something - we may want to apply some of the benchmarking changes made in this PR to DFJCOSK, as well. It will increase the size of the PR, but the benchmarking changes in this PR currently only extend to DirectJK at the moment. Since DFJCOSK has two methods that it separately benchmarks, it will require a bit of retooling regarding some of the internals of the benchmarking framework. It should not have a significant impact on test_erisieve, however.

> Thoughts?

Unless the DFJCOSK changes would undo much of this PR, I think a follow-up PR would be best.

davpoolechem:
> > So I now realize something - we may want to apply some of the benchmarking changes made in this PR to DFJCOSK, as well. It will increase the size of the PR, but the benchmarking changes in this PR currently only extend to DirectJK at the moment. Since DFJCOSK has two methods that it separately benchmarks, it will require a bit of retooling regarding some of the internals of the benchmarking framework. It should not have a significant impact on test_erisieve, however.
> 
> > Thoughts?
> 
> Unless the DFJCOSK changes would undo much of this PR, I think a follow-up PR would be best.

DFJCOSK won't explicitly undo most of this PR, nicely enough, though it will require some changes to how the computed_shells member functions/variables are handled. Regardless, it won't lead to significant changes in test_erisieve, so a separate PR should work fine.

Thank you for your feedback!

PR #2550, opened 2022-04-20
Fix Incfock Convergence Bug

andyj10224:
## Description
Incremental Fock Build (In Direct JK algorithm), was supposed to have a feature where it stops after the density matrix has dropped below a certain cutoff. In my old code, I set that value to be the value of SCF D_CONVERGENCE, but that was a mistake, as this value is supposed to be higher than the SCF D_CONVERGENCE value. I fix that bug in this PR.

## Todos

- [x] Fix INCFOCK bug

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/DirectJK.cc
+++ b/psi4/src/psi4/libfock/DirectJK.cc
@@ -385,7 +385,7 @@ void DirectJK::compute_JK() {
         timer_on("DirectJK: INCFOCK Preprocessing");
         incfock_setup();
         int reset = options_.get_int("INCFOCK_FULL_FOCK_EVERY");
-        double dconv = options_.get_double("D_CONVERGENCE");
+        double dconv = options_.get_double("INCFOCK_CONVERGENCE");
         double Dnorm = Process::environment.globals["SCF D NORM"];
         // Do IFB on this iteration?
         do_incfock_iter_ = (Dnorm >= dconv) && !initial_iteration_ && (incfock_count_ % reset != reset - 1);
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -1456,6 +1456,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Frequency with which to compute the full Fock matrix if using |scf__incfock| . 
         N means rebuild every N SCF iterations to avoid accumulating error from the incremental procedure. -*/
         options.add_int("INCFOCK_FULL_FOCK_EVERY", 5);
+        /*- The density threshold in which to stop performing INCFOCK -*/
+        options.add_double("INCFOCK_CONVERGENCE", 1.0e-5);
 
         /*- Perform the linear scaling exchange (LinK) algorithm, as described in [Ochsenfeld:1998:1663]_.
             Only applies to Direct SCF. -*/


JonathonMisiewicz (changes requested):
Please rename `dconv` to something more expressive, so we're not wondering five years from now why `dconv` is not given by the `d_convergence` option.

CDSherrill (commented):
Andy and I discussed this today.  The fix looks good.

psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -385,7 +385,7 @@ void DirectJK::compute_JK() {
         timer_on("DirectJK: INCFOCK Preprocessing");
         incfock_setup();
         int reset = options_.get_int("INCFOCK_FULL_FOCK_EVERY");
-        double dconv = options_.get_double("D_CONVERGENCE");
+        double dconv = options_.get_double("INCFOCK_CONVERGENCE");
         double Dnorm = Process::environment.globals["SCF D NORM"];
         // Do IFB on this iteration?
         do_incfock_iter_ = (Dnorm >= dconv) && !initial_iteration_ && (incfock_count_ % reset != reset - 1);
```
CDSherrill, line 391: Actually people reading this code might misread "dconv" to mean the user-desired convergence on the density, which isn't what it means.  So, similarly to changing D_CONVERGENCE to INCFOCK_CONVERGENCE, how about renaming dconv to incfock_conv or something?

psi4/src/read_options.cc
```diff
@@ -1456,6 +1456,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Frequency with which to compute the full Fock matrix if using |scf__incfock| . 
         N means rebuild every N SCF iterations to avoid accumulating error from the incremental procedure. -*/
         options.add_int("INCFOCK_FULL_FOCK_EVERY", 5);
+        /*- The density threshold in which to stop performing INCFOCK -*/
```
zachglick, line 1459:
```suggestion
        /*- The density threshold at which to stop building the Fock matrix incrementally -*/
```

zachglick: while you're here, line 1454 has unnecessary capitalization. It should read "Do perform incremental Fock build?"

zachglick (approved):

psi4/src/psi4/libfock/DirectJK.cc
andyj10224: Fixed it

psi4/src/read_options.cc
andyj10224: Done

andyj10224:
> Please rename `dconv` to something more expressive, so we're not wondering five years from now why `dconv` is not given by the `d_convergence` option.

Just made the changes. Please proceed with review

JonathonMisiewicz (approved):

loriab (approved):

andyj10224:
Hmm, I got an ADIIS error in windows (on a pytest) for my tests. Is this to be expected?

loriab:
aediis error for windows + test_psi4_sapt may be intermittent since not seen in first commit. there's no reason incfock should be on in that test, right? in that case, no problem.

JonathonMisiewicz:
That test is DF, not direct, so this PR is innocent. I had observed some random ADIIS failures previously but thought I had fixed them. I recommend we rerun the failed test and save investigation into the ADIIS situation for a future PR. (Probably to be done by me.)

davpoolechem (approved):
Everything looks good to me!

PR #2551, opened 2022-04-21
`autotest` Shall Die, Plus Pair Energies

### LINKED ISSUE #943: migrate coupled-cluster test cases to independence
Presently, there are rather few `compare_values()` statements in the `tests/cc*/input.dat` test cases. This is because there's a very thorough [perl script](https://github.com/psi4/psi4/blob/master/tests/psitest.pl) that parses the output file and parses the alongside `reference.dat` file and compares everything in sight to a sensible number of decimal places. Works great (except when we accidentally break its arrangements and then everything trivially passes for months at a time until someone notices) in Make or CMake testing framework for checking CC. _Doesn't_ work great to check that psivar coverage is good for the cc* modules or independently of the ctest testing macros or in an installed (as opposed to source) testing env w/o reference.dat files or without perl.

I think the in-file compare_values() testing is an idea that has proven its worth. I suggest that:
1.  the key reference values from each `reference.dat` be collected into the various `cc*/input.dat` tests,
2. more psivars be defined as needed in the `psi4/src/psi4/cc*` modules, 
3. a `compare_coupledcluster()` be written to take in a wfn and a structure of reference values and apply all the checks that the perl script currently does through compare_values(), compare_arrays(), etc.

(3) is negotiable; it's just my current idea on how to move toward compare_values() without losing coverage. Primary aims are to make coupled-cluster tests suitable for installed testing (`assert compare_values(...)` works great in pytest) and increase coupled-cluster psivar usage (@amjames already did one follow-up pass) for the QCSchema project. Further note that in the near future, psivars need to be on Wfn, *not* `P::e`. Might be worth multiple ppl working to spread the pain.

This hasn't been run by @lothian, so he may have other ideas.

JonathonMisiewicz:
## Description
Closes #943: the last of the `cc` tests is expressed in terms of psivars instead of the autotest mechanism. The last test moved required adding psivars about pair energies for RHF CC computations.

I want a review from Zach for pair energies, from TDC for CC, and from Lori for psivars.

## Todos
- [x] Adds MP2 and CC pair energy psivars
- [x] Removes `autotest` capabilities 

## Questions
- [ ] If you close a "mugworthy" PR but already have a mug, what then?

## Checklist
- [x] `cc43` still passes
- [x] quick tests still pass 

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
By the way, the commit history is clean for this one. If the LoC count is intimidating, go commit-by-commit. I also recommend we merge rather than merge squash this one.

lothian (approved):

psi4/src/psi4/cc/ccenergy/get_params.cc
```diff
@@ -165,7 +165,6 @@ void CCEnergyWavefunction::get_params(Options &options) {
 
     params_.print_mp2_amps = options.get_bool("MP2_AMPS_PRINT");
     params_.print_pair_energies = options.get_bool("PAIR_ENERGIES_PRINT");
-    params_.spinadapt_energies = options.get_bool("SPINADAPT_ENERGIES");
```
zachglick, line 168: This looks like the only place in the code where `SPINADAPT_ENERGIES` is used, so you can also remove it from `read_options.cc`

zachglick (commented):
The pair energy stuff is all very sensible, LGTM!

tests/runtest.py
```diff
@@ -32,35 +32,34 @@
...
 
 # extract run condition from arguments
 python_exec = sys.argv[0]
 infile = sys.argv[1]
 logfile = sys.argv[2]
-psiautotest = sys.argv[3]
-top_srcdir = sys.argv[4]
-sowreap = sys.argv[5]
+top_srcdir = sys.argv[3]
+sowreap = sys.argv[4]
```
loriab, line 44: sowreap is also dead, btw. but don't mess up the commit history just on that account.

tests/CMakeLists.txt
```diff
@@ -121,9 +121,6 @@ if(ENABLE_pasture)
 endif()
 
 message(STATUS "Adding test cases: Psi4")
-if(NOT PERL_FOUND)
-   message(WARNING "Perl was not found. Some Coupled Cluster test cases will not be thoroughly checked.")
-endif()
```
loriab, line 126: You could also kill off the first line in this file: `find_package(Perl QUIET)`. But it's not doing any harm.

loriab (approved):
splendid! qcvars lgtm

tests/runtest.py
JonathonMisiewicz: I'd rather not add that to this PR. There are additional questions about what to do with the tests currently tagged `sowreap` but not in the CMake. Sounds like a worthy topic for an issue, though. 🙂

JonathonMisiewicz:
One kicked down to future work, but all others applied.

zachglick (approved):

PR #2552, opened 2022-04-21
ddd modifications to findif

### LINKED ISSUE #1683: FINDIF Doesn't Intercept Single Atom Hessian
Transferred from the [forums](http://forum.psicode.org/t/basisset-not-set-from-freq-on-an-atom-at-ump2-level/1454).

The gradient and hessian of a single atom is undefined within the Born-Oppenheimer approximation, so libmints rightfully complains when you attempt to compute an analytic gradient of a single atom. The finite difference code is not that smart. It will naively get a single point for a gradient, which is zero. But for hessians, it currently crashes. For good measure, here's a reproducible example:
```
molecule { 
0 2
F 0.0 0.0 0.0 
}
set basis 6-31G*
set reference uhf 
frequency('b3lyp')
```

It would be pointless to fix this in the current driver, because the revamped driver is a 1.4 target, according to Lori. The DDD team should get this fixed in the DDD branch. It may be caught already. When this issue is fixed there, we can close the issue with a caveat that the fix won't be live until DDD goes in.

loriab:
## Description
The first eight DDD extracts PRs reduced the delta from
99 changed files, and 5,314 additions and 2,946 deletions. c. dgasmith:recursive to
59 changed files, and 3,408 additions and 1,885 deletions. c. loriab:recursive62
so here's No. 9 of the DDD series, #1351.

## Todos
- [x] This is the non-class parts of changes to finite difference. Mostly
 * using pieces from Mol, rather than mol, to reduce dependence on psi objects
 * use more numpy, less Matrix
 * consolidate printing so can go to logging
 * start to pass finite difference options (points, steps, projection) as kwargs rather than having findif consult global options
- [x] added in the dipder from dipole but it's not used
- closes #1683
- [x] along the way, psimrcc-fd-freq1 failed. as far as I can judge now, this is an existing bug. part of the old output is below. psi is evaulating the gradient, finding it over the cutoff, and so recommending not projecting rotations. then, because of https://github.com/psi4/psi4/blob/master/psi4/driver/driver.py#L1613, that setting of FD_PROJECT gets tossed, and findif runs projecting rotations, and the test passes. I'm calling this a bug and setting fd_project in the input to force the array dimensions to match and the test to pass. glad to hear other interpretations.

```

  Based on options and gradient (rms=3.60E-02), recommend projecting translations and not projecting rotations.
hessian() will perform frequency computation by finite difference of analytic energies.

         ----------------------------------------------------------
                                   FINDIF
                     R. A. King and Jonathon Misiewicz
         ---------------------------------------------------------

  Using finite-differences of gradients to determine vibrational frequencies and
  normal modes. Resulting frequencies are only valid at stationary points.
    Generating geometries for use with 5-point formula.
    Displacement size will be 1.00e-02.
    Number of atoms is 2.
    Number of irreps is 8.
    Number of SALCs is 1.
    Translations projected? 1. Rotations projected? 1.
    Index of SALCs per irrep:
```
- [x] ADDED: moved the write_hessian fn and collected write_gradient into a matching fn.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver_findif.py
+++ b/psi4/driver/driver_findif.py
@@ -1078,6 +1078,23 @@ def prep_findif(mol, irrep, mode, gradient=None):
         return fdargs
 
 
+def hessian_write(wfn: core.Wavefunction):
+    if core.get_option('FINDIF', 'HESSIAN_WRITE'):
+        filename = core.get_writer_file_prefix(wfn.molecule().name()) + ".hess"
+        with open(filename, 'wb') as handle:
+            qcdb.hessparse.to_string(np.asarray(wfn.hessian()), handle, dtype='psi4')
+
+
+def gradient_write(wfn: core.Wavefunction):
+    if core.get_option('FINDIF', 'GRADIENT_WRITE'):
+        filename = core.get_writer_file_prefix(wfn.molecule().name()) + ".grad"
+        qcdb.gradparse.to_string(np.asarray(wfn.gradient()),
+                                 filename,
+                                 dtype='GRD',
+                                 mol=wfn.molecule(),
+                                 energy=wfn.energy())
+
+
 def _rms(arr: Union[core.Matrix, np.ndarray]) -> float:
     """Compute root-mean-square of array, be it Psi4 or NumPy array."""
     if isinstance(arr, np.ndarray):
--- a/psi4/driver/driver_findif.py
+++ b/psi4/driver/driver_findif.py
@@ -48,9 +48,8 @@ logger = logging.getLogger(__name__)
 # h is the index of an irrep.
 
 
-def _displace_cart(mass: np.ndarray, geom: np.ndarray, salc_list: "psi4.core.CdSalcList", i_m: Iterator[Tuple], step_size: float) -> Tuple[np.ndarray, str]:
+def _displace_cart(mass: np.ndarray, geom: np.ndarray, salc_list: core.CdSalcList, i_m: Iterator[Tuple], step_size: float) -> Tuple[np.ndarray, str]:
     """Displace a geometry along the specified displacement SALCs.
-    Called within :py:func:`_geom_generator` so from each displacement mode: [gradient|hessian]_from_[gradients|energies]_geometries.
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


psi4/driver/driver_findif.py
```diff
@@ -26,119 +26,154 @@
...
 # CONVENTIONS:
 # n_ at the start of a variable name is short for "number of."
 # _pi at the end of a variable name is short for "per irrep."
 # h is the index of an irrep.
 
-array_format = {"precision": 10}
-
 
-def _displace_cart(mol, geom, salc_list, i_m, step_size):
+def _displace_cart(mass: np.ndarray, geom: np.ndarray, salc_list: "psi4.core.CdSalcList", i_m: Iterator[Tuple], step_size: float) -> Tuple[np.ndarray, str]:
```
JonathonMisiewicz, line 51: Are the quotes around `psi4.core.CdSalcList` needed?

psi4/driver/driver_findif.py
```diff
@@ -26,119 +26,154 @@
...
     """Displace a geometry along the specified displacement SALCs.
+    Called within :py:func:`_geom_generator` so from each displacement mode: [gradient|hessian]_from_[gradients|energies]_geometries.
```
JonathonMisiewicz, line 53: I don't understand this comment. Do you mean "Called by :py:func:`_geom_generator`, i.e., `gradient|hessian]_from_[gradients|energies]_geometries.` Called once for each displacement."?

psi4/driver/driver_findif.py
```diff
@@ -390,41 +438,71 @@ def append_geoms(indices, steps):
...
+
+    Returns
+    -------
+    findifrec : dict
+        Dictionary of finite difference data, specified in _geom_generator docstring.
+
+"""
+
+
+gradient_from_energies_geometries = partial(_geom_generator, freq_irrep_only=-1, mode="1_0")
```
JonathonMisiewicz, line 471: Nice use of partial!

psi4/driver/driver_findif.py
```diff
@@ -850,63 +1021,70 @@ def init_string(data):
...
-
-    Notes
-    -----
-    Only symmetric displacements are necessary, so user specification of
-    symmetry is disabled.
-    """
-    return _geom_generator(molecule, -1, "1_0")
+########
 
+# This function is shifting `_energy_is_invariant` from driver to findif but isn't in final DDD form.
```
JonathonMisiewicz, line 1026: The meaning of this string is lost on me. If this is a temporary function, I'll drop the matter.

psi4/driver/driver_findif.py
loriab: quotes are used when you want to specify the type but actually importing it and having the object available would cause trouble. here, unquoted `core.CdSalcList` is probably safe. good catch.

I figure I'm going to be adjusting the signature and docstrings formatting someday when I can look at the built docs.

psi4/driver/driver_findif.py
loriab: I was trying to trace the flow through the fns. Have since found a better way so will remove.

psi4/driver/driver_findif.py
loriab: yes, it's presently ugly lines of code that are in the right place and will look more reasonable soon. suspended review requested.

JonathonMisiewicz (approved):
I'll go ahead and approve this now. Please follow-up on the unresolved items, especially whether we can close 1683.

zachglick (approved):
LGTM!

PR #2557, opened 2022-04-26
update deprecation warnings text

loriab:
## Description
This is No. 10 of the DDD series, #1351.

## Todos
- mostly a few little differences with ddd that will reduce the changed line overhead
  - [x] add logging printing, minimize imports
  - [x] `mol = core.Molecule.from_schema(molschemus, nonphysical=True)` allows freq-masses test to run in ddd. nonphysical flag allows user-set atomic masses outside the known isotope range. the schema runner shouldn't be the block to this.
  - [x] in tests add some alternate ref values. sometimes these are conventional, which can be handy reusing a test in a non-df program
- [x] adding deprecation warnings is good, but we don't often actually delete the fn promptly at the version we say it will stop working at. this is harmless, in my opinion, as it gives people longer to adapt, and developers can clear away the old fns when they actually become inconvenient. to better reflect this state of affairs, the standard message "Using blah is deprecated, and in version it will stop working" has been changed to "Using blah is deprecated, and as soon as version it will stop working". this also removes the temptation to keep incrementing <version>, which gives a misleading impression about how long ago users were warned. have backtracked on run_json accordingly from 1.7 to 1.5.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.azure-pipelines/azure-pipelines-windows.yml
+++ b/.azure-pipelines/azure-pipelines-windows.yml
@@ -222,7 +222,6 @@ jobs:
 
       # Spot Test
       - script: |
-          setlocal EnableDelayedExpansion
           set PATH=$(Build.BinariesDirectory)\install\bin;%PATH%
           psi4 $(Build.SourcesDirectory)\tests\tu1-h2o-energy\input.dat -o stdout
         displayName: 'Run, Run, Spot, Run Test'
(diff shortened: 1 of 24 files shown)


JonathonMisiewicz (approved):
Not much controversial in this one.

PeterKraus (approved):

maxscheurer (approved):

PR #2473, opened 2022-03-13
Refactor `PseudospectralInts` using Libint2

maxscheurer:
## Description
This PR refactors the existing `PseudospectralInts` (untested, unused, but maybe needed at some point, see #2414)
with some refactoring: The "old" `PseudospectralInts` did compute `<p| 1/R |q>` *or* `<p| erf(omega*R)/R |q>` with `R = |r -r_c|` for some given point/origin `r_c`. 

- The class `PseudospectralInts` is removed. To compute the "normal" Coulomb potential at a given origin, one can use either `MultipolePotentialInt` (with `max_k=0`) *OR* `ElectrostaticInt`.
- To compute the erf-attenuated Coulomb potential, I created a new class named `PotentialErfInt` to better reflect the actual "property integral" being computed. For testing purposes, I also added the erfc-attenuated Coulomb potential (`PotentialErfComplementInt`), maybe they're also useful at some point. Both use L2 under the hood.
- The point `r_c` is canonically set with `set_origin`.
- I've added the new integrals to `MintsHelper` and the Python API.
- Tests are added. (`erf(infinity*R)/R = 1/R`, `erfc(0*R)/R = 1/R`, and `1/R - erf(R)/R - erfc(R)/R = 0`).

This is a preparation to get rid of most of the remaining OS86 code as discussed in #2414.
Ping @loriab @zachglick @andysim 

## Questions
- [ ] Question1

## Checklist
- [x] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2567, opened 2022-05-02
Semi-Numerical Exchange

zachglick:
## Description
This PR implements a semi-numerical algorithm for computing the exchange (K) matrix as described in [this paper](https://doi.org/10.1016/j.chemphys.2008.10.036) and [this paper](https://doi.org/10.1063/1.3646921) as "Chain of Spheres" exchange or COSK. The implementation is added to a new `JK` class, `DFJCOSK`, which combines COSK with a simple direct density-fitted J algorithm.

COSK is easily parallelized, completely in-core, and has good scaling--between O(N^1) and O(N^2)--making it ideal for HF/DFT calculations on large systems, particularly calculations for which the `DF` algorithm must store integrals on disk. The low scaling comes at the cost of a large pre-factor, so it shouldn't be used on small systems.

In my benchmarking, `DFJCOSK` is almost always faster than `DIRECT` (even with density screening and incremental fock construction). The crossover between `DFJCOSK` and `DF` is dependent on the system, basis, CPU, memory, etc. but I've observed it as early as <3000 basis functions. I've attached a performance comparison of the `DFJCOSK`, `DIRECT`, and `DF` algorithms on a set of linear alkanes in which it's clear that `DFJCOSK` outperforms the other two algorithms for large systems.

alkanes

An encouraging fact about these results is that this implementation is far from optimal. The cost of the COSK algorithm is dominated by the evaluation of one-electron integrals on a quadrature grid. Therefore, there are a number of improvements to be made, in order of increasing difficulty:

- Do a better job at screening small integrals before they're computed (like [these](https://aip.scitation.org/doi/pdf/10.1063/1.5048491))
- Use better grids (such as [here](https://aip.scitation.org/doi/full/10.1063/1.5049435))
- Compute individual integrals faster. This could be done by reusing intermediates in the recursion for integrals involving the same shell-pair and different grid points.

## Todos

- [x] Implement semi-numerical exchange in the `DFCOSK` class
- [x] Add tests & docs

## Questions
- [x] What do we want the `SCF_TYPE` keyword for this algorithm to be? It's currently `COSK`. It could also be `DFJCOSK`. It might make more sense to call the algorithm `SNK` (for semi-numerical exchange) over `COSK`, as the latter is more general, and will still be accurate even if we change the screening procedure. (The former refers to a very specific screening procedure).
- [x] For anyone who regularly uses COSK: do you have a preference for how the tightness of the integration grids are specified? Currently, the user gives the number of spherical/angular points. We could also predefine a set of allowed spherical/angular combos. That might be easier for users (they don't have to know about lebedev numbers, for example), but it sacrifices flexibility.
- [x] This is less a question and more of a note: the `DFTGrid` class is in serious need of a refactor so that it does not read from global options. This PR kicks the can down the road by adding another constructor to the class, which takes `double`-type options. I'd be happy to hear any thoughts on what to do with this class, though that probably deserves its own issue.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


susilehtola:
What is the x axis in the figure? N?

susilehtola:
In Turbomole, seminumerical exchange is called "senex". I would prefer reusing existing acronyms over adopting new ones like the above "SNK".

susilehtola:
As to the grid tightness, IMHO quadrature grids should be inputable either in standard grid form or as a combination of radial points and angular points.

zachglick:
> What is the x axis in the figure? N?

Yes, the x-axis is `N`, the number of carbons in the alkane chain

psi4/src/psi4/libfock/cubature.cc
```diff
@@ -4309,18 +4318,33 @@ void DFTGrid::buildGridFromOptions(std::map<std::string, int> int_opts_map,
...
+            full_float_options[key] = float_opts_map[key];
+        } else {
+            full_float_options[key] = options_.get_double(key);
+        }
+        // prfloatf("%s : %d\n", key.c_str(), full_float_options[key]);
+    }
+
     // TODO: Move both the double and bool parts from a global to a grid specific option.
     //  possibly via an approach listed here:
     //  https://stackoverflow.com/questions/24702235/c-stdmap-holding-any-type-of-value/24702596
```
hokru, line 4338: std::any from C++17 following this discussion would be a suggestion how to generalize the option map.

hokru:
I would keep the "COSX" for recognition.

This PR implements 2 grids instead of 3? That Turbomole team showed it worked well, just asking.

I am in general for a simple way of setting quadrature grids, psi4-specific named grids. That is a bigger project though.
Here it would be enough to have an efficient default and in the manual a suggestion how to make a tight/accurate one.
After we gain experience with the psi4 COSX we can make a couple useful defaults.

Couple of things about grids later.

zachglick:
> I would keep the "COSX" for recognition.

That's a good point; "COSX" is a fairly well-known algorithm. For consistency throughout the psi4's `libfock` library, I think it would be good to call it "COSK" over "COSX"; all reference to the exchange matrix use the variable "K".

> This PR implements 2 grids instead of 3? That Turbomole team showed it worked well, just asking.

Yes, the SCF is first converged on a small grid (specified by `COSK_RADIAL_POINTS` and `COSK_SPHERICAL_POINTS`). Afterwards, a single iteration is performed on a larger grid (specified by `COSK_RADIAL_POINTS_FINAL` and `COSK_SPHERICAL_POINTS_FINAL`). A good future mini-optimization would be to add a third medium grid, which would be used to converge the SCF after the small grid but before the final grid evaluation.

> I am in general for a simple way of setting quadrature grids, psi4-specific named grids. That is a bigger project though. Here it would be enough to have an efficient default and in the manual a suggestion how to make a tight/accurate one. After we gain experience with the psi4 COSX we can make a couple useful defaults.

I agree, it would be nice to have named quadrature grids. For now, I'll update the manual with some recommended spherical/radial point values.

hokru:
Internally it's fine to use COSK, however user-facing it is better to use COSX, imo.

Ideally the COSK grids can use all the options the normal DFT grid uses. At the very least pruning.
However it is a bit annoying that one has to duplicate/triplicate the normal DFT grid options for this. Not sure how this can be solved best.

zachglick:
> Internally it's fine to use COSK, however user-facing it is better to use COSX, imo.

That's a good idea, done.

> Ideally the COSK grids can use all the options the normal DFT grid uses. At the very least pruning. However it is a bit annoying that one has to duplicate/triplicate the normal DFT grid options for this. Not sure how this can be solved best.

Yes, the fact that the COSX grids are affected by keywords such as `DFT_PRUNING_SCHEME`, `DFT_RADIAL_SCHEME`, `DFT_NUCLEAR_SCHEME`, etc. is not ideal. I don't think we should duplicate/triplicate these keywords (resulting in independent but parallel keywords like `COSX_RADIAL_SCHEME`, `COSX_RADIAL_SCHEME_FINAL`, etc.) because I can't imagine any users realistically wanting to manually set those values. I think it makes more sense to hardwire some of the grid settings when building the grids for COSX. For example, we probably always want do `ROBUST` pruning on those grids.

psi4/src/read_options.cc
```diff
@@ -178,7 +178,7 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     /*- What algorithm to use for the SCF computation. See Table :ref:`SCF
     Convergence & Algorithm <table:conv_scf>` for default algorithm for
     different calculation types. -*/
-    options.add_str("SCF_TYPE", "PK", "DIRECT DF MEM_DF DISK_DF PK OUT_OF_CORE CD GTFOCK");
+    options.add_str("SCF_TYPE", "PK", "DIRECT DF MEM_DF DISK_DF PK OUT_OF_CORE CD GTFOCK COSX");
```
loriab, line 181: should this be `DFJ_COSX` for future composite compatibility? or does `COSX` imply the DFJ widely enough?

psi4/src/psi4/libmints/electrostatic.cc
```diff
@@ -62,6 +63,15 @@ void ElectrostaticInt::compute(SharedMatrix& result, const Vector3& C) {
     OneBodyAOInt::compute(result);
 }
 
+void ElectrostaticInt::set_origin(const Vector3& _origin) {
+    origin_ = _origin;
+    Zxyz_vector pcs;
+    // l2 includes the electron charge, legacy psi4 code does not, so
+    // we adopt this behavior here with -1.0
+    pcs.push_back({-1.0, {origin_[0], origin_[1], origin_[2]}});
+    engine0_->set_params(pcs);
```
loriab, line 72: good catch!

loriab (commented):
lgtm!

psi4/src/psi4/libmints/electrostatic.cc
zachglick: Credit goes to @maxscheurer for that, I borrowed it from him [here](https://github.com/psi4/psi4/pull/2473/files#diff-afe4b5648da445aad249f05fa847be3a9e29f5bc0051d2ae3a2e42531c11152dR56-R65)

psi4/src/read_options.cc
zachglick: The latter, `DFJ` is the only sensible J algorithm to be combined with `COSX`, so I left it off the `SCF_TYPE` keyword in order to be more user-friendly. The `JK` class is given the full name `DFJCOSK` because it isn't user-facing

tests/pytests/test_dfjcosk.py
```diff
@@ -0,0 +1,85 @@
...
+pytestmark = [pytest.mark.psi, pytest.mark.api]
+
+
+h2o = psi4.geometry("""
+0 1
+O    0.000000000000     0.000000000000    -0.124038860300
+H    0.000000000000    -1.431430901356     0.984293362719
+H    0.000000000000     1.431430901356     0.984293362719
+units au
+""")
```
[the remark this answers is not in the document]
zachglick: I'm glad I've never come across a bug like that; I can't imagine how you'd ever discover it. Changes made.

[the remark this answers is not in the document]
loriab: Thanks for the changes. Discovery is easy -- test works in file, test fails in suite. Figuring out what was wrong as a new pytest user was memorable.

psi4/src/read_options.cc
```diff
@@ -1544,6 +1544,30 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Bump function max radius -*/
         options.add_double("DF_BUMP_R1", 0.0);
 
+        /*- SUBSECTION COSX Algorithm -*/
+
+        /*- Number of spherical points in initial COSX grid. -*/
+        options.add_int("COSX_SPHERICAL_POINTS", 50);
+        /*- Number of radial points in initial COSX grid. -*/
+        options.add_int("COSX_RADIAL_POINTS", 25);
```
davpoolechem, line 1552: It may be worth considering renaming the initial grid variables to COSX_SPHERICAL_POINTS_INITIAL and COSX_RADIAL_POINTS_INITIAL, as this clarifies their role in the calculation and makes the initial grid keywords match more closely with the final grid keywords.

davpoolechem, line 1552: It may be worth considering renaming the initial grid variables to COSX_SPHERICAL_POINTS_INITIAL and COSX_RADIAL_POINTS_INITIAL, as this clarifies their role in the calculation and makes the initial grid keywords match more closely with the final grid keywords.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1014 @@
...
+        for(size_t jki = 0; jki < njk; jki++) {
+
+            auto JTp = JT[jki][rank]->pointer();
+            auto Hp = H[jki]->pointer();
+
+            for (int p = pstart, index = 0; p < pstart + np; p++) {
+                for (int m = mstart; m < mstart + nm; m++) {
+                    for (int n = nstart; n < nstart + nn; n++, index++) {
+                        JTp[m][n] += buffer[index] * Hp[p];
+                        if (N != M) JTp[n][m] += buffer[index] * Hp[p];
```
davpoolechem, line 575: Same question as before - is this thread-safe?

davpoolechem, line 575: Same question as before - is this thread-safe?

psi4/src/read_options.cc
zachglick: Yeah, fair enough

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1014 @@
...
+        for (size_t shell_ind = 0; shell_ind < shell_map.size(); shell_ind++) {
+            shell_map_inv[shell_map[shell_ind]] = shell_ind;
+        }
+
+        // => Process Density Matrix <= //
+
+        // significant cols of D for this grid block
+        std::vector<SharedMatrix> D_block(njk);
+        for(size_t jki = 0; jki < njk; jki++) {
+            D_block[jki] = std::make_shared<Matrix>(nbf_block_all, nbf_block);
```
davpoolechem, line 754: Correct me if I am wrong, but it seems that this formulation of D_block results in the COSX implementation using O(N^3) memory - D_block is a vector of size njk=D.size()=nbf*nbf containing pointers to shared matrices each of size (nbf_block_all=nbf)*nbf_block. Possibly something to defer to a future PR, but would it be feasible to change the algo so it doesn't use O(N^3) memory?

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: Yes, each thread has its own J buffer, `JT`. That might not be ideal from a memory standpoint, but it makes the parallelism very simple.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1014 @@
...
+        for(size_t jki = 0; jki < njk; jki++) {
+
+            auto GTp = GT[jki][rank]->pointer();
+            auto Dp = D[jki]->pointer();
+
+            for (int p = pstart, index = 0; p < pstart + np; p++) {
+                for (int m = mstart; m < mstart + nm; m++) {
+                    for (int n = nstart; n < nstart + nn; n++, index++) {
+                        GTp[p] += buffer[index] * Dp[m][n];
+                        if (N != M) GTp[p] += buffer[index] * Dp[n][m];
```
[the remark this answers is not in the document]
davpoolechem: Ahhhhhhhh, I see. I missed the fact that each GT is thread-local. I assumed it was similar to Andy's implementation, where each thread was writing to a thread-shared gamma. My bad!

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: Question answered above. Consider this resolved!

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: `njk` is not `nbf ^ 2`, it's the number of unique `J` and `K` objects to be formed (1 in the RHF case, 2 in the UHF case, etc.)

Also `nbf_block` is constant in size. Thus, the total memory usage is only O(N), not O(N^3)

davpoolechem: Ah! You're completely right. D isn't just a single density matrix - it's the vector of individual density matrices. It would be D[i] that is nbf*nbf. And yeah, nbf_block is constant, and nbf_block_all is O(N), so you're right. Once again, my mistake!

Maybe I'm not in the right state of mind to review this code right now. I do apologize.

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: And for whatever it's worth, I've seen (and used) a lot of thread-private Fock buffers in the past, so I am not at all opposed to thread-private J buffers as used here.

davpoolechem:
I looked over some parts of the code, and basically everything looks good so far! I shall continue reviewing tomorrow.

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: All good! You're asking the right types of questions

davpoolechem (approved):
All right, I finished my review of the code, and everything looks good to me!

As a side comment, the benchmarks look pretty great, but it would be interesting to see how the benchmark results look with a system spread out across 3D space instead of a linear system.

zachglick:
> As a side comment, the benchmarks look pretty great, but it would be interesting to see how the benchmark results look with a system spread out three-dimensionally instead of a linear system.

Thanks, I have run additional tests on three-dimensional systems not shown here. On the [human insulin protein](https://www.rcsb.org/structure/2HIU) (405 heavy atoms), a HF/cc-pVDZ calculation with `COSX` is a bit more than twice as fast as the corresponding `DF` calculation.

susilehtola:
IIRC speedups of seminumerical methods increase with larger basis sets, so you may want to benchmark those as well...

zachglick:
> IIRC speedups of seminumerical methods increase with larger basis sets, so you may want to benchmark those as well...

You’re totally right. I have DZ timings, which agree with your statement. I need to run QZ eventually.

Side note: the integral bound used in this PR is looser for higher angular momentum. I expect future improvements to this integral bound to make the performance of higher angular momentum basis sets even better.

susilehtola:
I didn't examine the code in detail; does it use aliasing? I think Neese and coworkers use it a lot, the idea is to make the integrand smoother so that the quadrature errors are smaller and one can get away with smaller grids.

zachglick:
> I didn't examine the code in detail; does it use aliasing? I think Neese and coworkers use it a lot, the idea is to make the integrand smoother so that the quadrature errors are smaller and one can get away with smaller grids.

Yes, this PR uses aliasing (also called "overlap fitting"). This feature is enabled with the `COSX_OVERLAP_FITTING` keyword. The second reference added to the bibliography covers the theory behind this idea, and the code references equations from this paper.

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,33 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
 basis set defined for all atoms in the system, or set |scf__df_scf_guess|
 to false, which disables this acceleration entirely.
 
+COSX Exchange
+~~~~~~~~~~~~~
+
+The semi-numerical COSX algorithm (described in [Neese:2009:98]_) is so-called
+because the two-electron ERIs are evaluated analytically over one electron 
+coordinate and numerically over the other electron coordinate. The numerical
```
susilehtola, line 759:
```suggestion
The semi-numerical COSX algorithm (described in [Neese:2009:98]_) evaluates two-electron ERIs
analytically over one electron coordinate and numerically over the other electron coordinate. The numerical
```

susilehtola: One could also mention the connection to the pseudospectral method by Friesner here?

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,33 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+integration is performed on standard DFT quadrature grids, which are described
+in :ref:`DFT`. Both the accuracy of the COSX algorithm and also the computational
+cost are directly determined by the size of the integration grid, so selection
+of the grid is important. This COSX implementation uses two separate grids.
+The SCF algorithm is first converged on a smaller grid, followed by a final SCF
+iteration on a larger grid. The size of the initial grid is controlled by the
+keywords |scf__cosx_radial_points_initial| and |scf__cosx_spherical_points_initial|. The final
+grid  is controlled by |scf__cosx_radial_points_final| and
+|scf__cosx_spherical_points_final|.
+Defaults have been chosen for both grids that balance cost and accuracy.
```
susilehtola, line 769:
```suggestion
The defaults for both grids aim to balance cost and accuracy.
```

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,33 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+Screening thresholds over integrals, densities, and basis extents are set
+with the |scf__cosx_ints_tolerance|, |scf__cosx_density_tolerance|, and
+|scf__cosx_basis_tolerance| keywords, respectively. The |scf__cosx_incfock|
```
susilehtola, line 772: I'm not sure `cosx_basis_tolerance` is a descriptive keyword?

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1014 @@
...
+    timer_on("ERI1");
+#pragma omp parallel for schedule(guided) num_threads(nthreads_) reduction(+ : computed_triplets1)
+    for (size_t MNP = 0; MNP < nshelltriplet; MNP++) {
+
+        size_t MN = MNP % nshellpair;
+        size_t P = MNP / nshellpair;
+        auto bra = eri_computers_[0]->shell_pairs()[MN];
+        size_t M = bra.first;
+        size_t N = bra.second;
+        if(Dshellp[M][N] * Dshellp[M][N] * J_metric_shell_diag[P] * eri_computers_[0]->shell_pair_value(M,N) < thresh2) {
```
susilehtola, line 452: I would move the rank declaration to the top and use `rank` as index here as well.

doc/sphinxman/source/scf.rst
zachglick: Added a sentence about Friesner

doc/sphinxman/source/scf.rst
zachglick: That's fair. I added some more information about the screening keywords

psi4/src/read_options.cc
```diff
@@ -1544,6 +1544,30 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        options.add_int("COSX_RADIAL_POINTS_FINAL", 35);
+        /*- Screening criteria for integrals and intermediates in COSX -*/
+        options.add_double("COSX_INTS_TOLERANCE", 1.0E-11);
+        /*- Screening criteria for shell-pair densities in COSX !expert -*/
+        options.add_double("COSX_DENSITY_TOLERANCE", 1.0E-10);
+        /*- Screening criteria for basis function values on COSX grids !expert -*/
+        options.add_double("COSX_BASIS_TOLERANCE", 1.0E-10);
+        /*- Pruning scheme for COSX grids !expert -*/
+        options.add_str("COSX_PRUNING_SCHEME", "ROBUST", 
+                        "ROBUST TREUTLER NONE FLAT P_GAUSSIAN D_GAUSSIAN P_SLATER D_SLATER LOG_GAUSSIAN LOG_SLATER NONE");
```
susilehtola, line 1565: This looks like duplication of pruning schemes. They should be collected under one variable in `read_options` and used for all instances of grid pruning choices.

zachglick: The duplication here is intentional. 

One alternative would be to determine the COSX pruning scheme from the value of the existing `DFT_PRUNING_SCHEME` keyword. I don't think that would be a good idea because (1) it's unintuitive to expect a keyword named `DFT_PRUNING_SCHEME` to affect a COSX calculation--DFT is independent of COSX--and (2) users may very likely want different DFT and COSX grids.

In the future, we could address point (1) by collecting the three keywords `DFT_PRUNING_SCHEME`, `COSX_PRUNING_SCHEME`, and `MBIS_PRUNING_SCHEME` into a `GRID_PRUNING_SCHEME` keyword as you said, but that leaves point (2) unaddressed. Also, that change would be well outside of the scope of this PR.

psi4/src/read_options.cc
```diff
@@ -1544,6 +1544,30 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        /*- SUBSECTION COSX Algorithm -*/
+
+        /*- Number of spherical points in initial COSX grid. -*/
+        options.add_int("COSX_SPHERICAL_POINTS_INITIAL", 50);
+        /*- Number of radial points in initial COSX grid. -*/
+        options.add_int("COSX_RADIAL_POINTS_INITIAL", 25);
+        /*- Number of spherical points in final COSX grid. -*/
+        options.add_int("COSX_SPHERICAL_POINTS_FINAL", 110);
+        /*- Number of radial points in final COSX grid. -*/
+        options.add_int("COSX_RADIAL_POINTS_FINAL", 35);
```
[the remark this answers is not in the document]
zachglick: These values are pretty similar to the defaults given in the COSX papers. We can't do an apples-to-apples comparison with ORCA, because in ORCA, the number of radial points is a function of the element's row.

I've ran calculations on sulfur-containing systems just fine, and gotten reasonable results.

psi4/src/read_options.cc
susilehtola: That is not what I meant.

When one adds a new pruning scheme, one needs to remember to add it in every single instance of the options for DFT, MBIS, and COSX.

My suggestion was to collect the possible grid choices into a variable, e.g. `std::string grid_schemes = "ROBUST TREUTLER NONE FLAT P_GAUSSIAN D_GAUSSIAN P_SLATER D_SLATER LOG_GAUSSIAN LOG_SLATER NONE";` and assigning this variable to the three above mentioned parameters. (I am assuming that the code that parses the grid options is the same for the three algorithms.)

zachglick: Oh, I see. Yeah, that would make it cleaner to add a new grid option. @loriab would a `grid_schemes` string disrupt the autodocs? If not, I'll make the change

loriab: I regret that this sensible suggestion will very likely flummox the autodocs. maybe add it as a todo to read_options.cc .

psi4/src/read_options.cc
```diff
@@ -1720,7 +1744,7 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("CPHF_MEM_SAFETY_FACTOR", 0.75);
         /*- SCF Type
          -*/
-        options.add_str("SCF_TYPE", "DIRECT", "DIRECT DF PK OUT_OF_CORE PS INDEPENDENT GTFOCK");
+        options.add_str("SCF_TYPE", "DIRECT", "DIRECT DF PK OUT_OF_CORE PS INDEPENDENT GTFOCK COSX");
```
JonathonMisiewicz, line 1747: This is probably a @loriab question, but why do we have a CPHF method at all, rather than `properties('scf')`? If we removed that, this option duplication becomes unnecessary.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+#include <algorithm>
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+using namespace psi;
+
+namespace psi {
+
+SharedMatrix compute_numeric_overlap(std::shared_ptr<DFTGrid> grid, std::shared_ptr<BasisSet> primary) {
```
JonathonMisiewicz, line 53: Why are we bothering with `shared_ptr`? Thanks to return value optimization, there's no performance hit for returning a `Matrix` rather than a pointer to a Matrix, and we never need to do memory management operations on these objects.

JonathonMisiewicz, line 53: Both incoming arguments should be `const &`.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+    // DOI 10.1063/1.3646921, EQ. 9
+
+    int nbf = primary->nbf();
+    BasisFunctions bf_computer(primary, grid->max_points(), grid->max_functions());
+    auto S_num = std::make_shared<Matrix>("Numerical Overlap", nbf, nbf);
+
+    // This loop could be parallelized over blocks of grid points. However, the cost of the loop is
+    // so small (< 10 seconds for a 200 heavy atom system), parallelism isn't necessary
+
+    for (size_t bi = 0; bi < grid->blocks().size(); bi++) {
```
JonathonMisiewicz, line 64: If we're not parallelizing, then use a range-based for loop. Don't forget to use `const &` where applicable.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+            for (size_t k = 0; k < nbf_block; k++) {
+                X_blockp[p][k] = point_values->get(p, k) * std::sqrt(w[p]);
+            }
+        }
+
+        // significant basis functions at these grid points
+        std::vector<int> bf_map = block->functions_local_to_global();
+
+        auto S_num_block = linalg::doublet(X_block, X_block, true, false);
+        auto S_num_blockp = S_num_block.pointer();
```
JonathonMisiewicz, line 89: For my own edification, is it any faster to grab a pointer object and access by index rather than call `S_num_blockp->get`? If it is, why aren't we doing the same trick with `S_num[mu][nu] +=`?

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+        Matrix X_block("phi_g,u", npoints_block, nbf_block);  // points x nbf_block
+        auto X_blockp = X_block.pointer();
+        for (size_t p = 0; p < npoints_block; p++) {
+            for (size_t k = 0; k < nbf_block; k++) {
+                X_blockp[p][k] = point_values->get(p, k) * std::sqrt(w[p]);
+            }
+        }
+
+        // significant basis functions at these grid points
+        std::vector<int> bf_map = block->functions_local_to_global();
```
JonathonMisiewicz, line 86: You should be able to do `const auto&` here. `libfock` does this all the time. Cleaner code, and probably eliminates a copy.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+
+        // grid points in this block
+        auto block = grid->blocks()[bi];
+        int npoints_block = block->npoints();
+        int nbf_block = block->local_nbf();
+        auto w = block->w();
+
+        // compute basis functions at these grid points
+        bf_computer.compute_functions(block);
+        auto point_values = bf_computer.basis_values()["PHI"];
```
JonathonMisiewicz, line 74: For objects as complex as a matrix, mark `const` when you can. I think this is also `&`-able.

JonathonMisiewicz (commented):
Since we're going to be tight for reviews, I gave this a relatively quick first pass. It looks like `DFJCOSK` could use a pass for use of `const` and `&`. I'll re-review once you've done that.

When the PR is in, please put up an issue for the efficiency improvements you mentioned in your first post, so we don't lose track of them when you graduate.

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,43 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
 basis set defined for all atoms in the system, or set |scf__df_scf_guess|
 to false, which disables this acceleration entirely.
 
+COSX Exchange
+~~~~~~~~~~~~~
+
+The semi-numerical COSX algorithm (described in [Neese:2009:98]_) evaluates
+two-electron ERIs analytically over one electron coordinate and numerically
+over the other electron coordinate. COSX is related to the earlier pseudospectral
+methods of Friesner. In COSX, numerical integration is performed on standard
```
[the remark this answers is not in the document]
susilehtola: Why?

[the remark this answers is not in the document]
JonathonMisiewicz: Such historical details are inappropriate for a user guide.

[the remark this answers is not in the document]
susilehtola: I disagree. It's one sentence.

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,43 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+DFT quadrature grids, which are described in :ref:`DFT`.
+Both the accuracy of the COSX algorithm and also the computational
+cost are directly determined by the size of the integration grid, so selection
+of the grid is important. This COSX implementation uses two separate grids.
+The SCF algorithm is first converged on a smaller grid, followed by a final SCF
```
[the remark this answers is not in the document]
zachglick: That is correct. Would it be more clear if I added an additional sentence?:

"This COSX implementation uses two separate grids. The SCF algorithm is first converged on a smaller grid, followed by a final SCF iteration on a larger grid. This results in numerical errors comparable to performing the entire SCF on the expensive larger grid at a computational cost much closer to the smaller grid."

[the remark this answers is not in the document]
loriab: I like the additional sentence.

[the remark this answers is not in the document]
JonathonMisiewicz: Yes - explicitly saying the purpose of the two grid approach is much cleaner.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -147,6 +147,17 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
 
         return std::shared_ptr<JK>(jk);
 
+    } else if (jk_type == "COSX") {
+        DFJCOSK * jk = new DFJCOSK(primary, auxiliary, options);
```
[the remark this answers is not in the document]
zachglick: I think your suggested one-liner is not exactly equivalent to the current code. The current code returns and object of type `JK` (not `DFJCOSK`):
```
        DFJCOSK * jk = new DFJCOSK(primary, auxiliary, options);
        return std::shared_ptr<JK>(jk);
```
We could call `std::make_shared<DFJCOSK>` and then cast the object to `JK`, but that doesn't seem definitively cleaner to me, and I'd worry about doing the cast correctly.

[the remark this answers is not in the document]
JonathonMisiewicz: I assumed that C++ would do the cast for us upon object return, because the signature _says_ the it returns `std::shared_ptr<JK>`. If that isn't true, then disregard the comment.

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: This is a `SharedMatrix`, not a `Matrix`. I was under the impression that `const auto &` saves you from expensive copies, which is only an issue when working with `Matrix` objects, not shared pointers. Is there still a reason to `const auto &`?

JonathonMisiewicz: Oops, my mistake.

This is one of the reasons I hate Psi's overuse of `shared_ptr`. In that case, the benefits to casting `const` and `&` are _there_ but both minor. The `const` means that we can't have this pointer point to a different object, but we can still change the object. The `&` means we create a new pointer. Using it is a touch cheaper, but it's negligible on the scale of the entire algorithm.

To mark the matrix itself as `const`, which was what I was hoping for, you'd need `std::shared_ptr<const Matrix>`, which would ruin the nice `auto`.

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: Good catch, thanks

psi4/src/psi4/libfock/jk.cc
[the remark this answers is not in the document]
zachglick: You're right, the `jk` object is casted just fine on return. Change made.

psi4/src/psi4/libfock/DFJCOSK.cc
hokru: Any function call adds an overhead, iirc, but the compiler can inline the function to eliminate that.

doc/sphinxman/source/scf.rst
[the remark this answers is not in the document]
susilehtola: Users who read the documentation, meaning people who want to learn about the method. Upon another read, I agree that the organization can be improved; I'll add a suggestion

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,45 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
 basis set defined for all atoms in the system, or set |scf__df_scf_guess|
 to false, which disables this acceleration entirely.
 
+COSX Exchange
+~~~~~~~~~~~~~
+
+The semi-numerical COSX algorithm (described in [Neese:2009:98]_) evaluates
+two-electron ERIs analytically over one electron coordinate and numerically
+over the other electron coordinate. COSX is related to the earlier pseudospectral
+methods of Friesner. In COSX, numerical integration is performed on standard
```
susilehtola, line 760:
```suggestion
The semi-numerical COSX algorithm described in [Neese:2009:98]_ evaluates
two-electron ERIs analytically over one electron coordinate and numerically
over the other electron coordinate, and belongs to the family of pseudospectral 
methods originally suggested by Friesner. In COSX, numerical integration is performed on standard
```

doc/sphinxman/source/scf.rst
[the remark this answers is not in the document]
susilehtola: See below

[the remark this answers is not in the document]
loriab: I generally favor clues that connect code keywords and the literature. Even if it's a distraction at initial get-things-working read, it's potentially reassuring at the what-exactly-have-I-run-and-why stage.

psi4/src/read_options.cc
loriab: I'm not sure. The CPHF block feeds libfock but so do other blocks. There is `properties("scf")`, but mostly ppl run oeprop directly instead.

loriab (approved):
lgtm, thanks for the nice new feature!

maxscheurer (approved):
Thanks for the nice feature! LGTM!

doc/sphinxman/source/scf.rst
```diff
@@ -745,6 +751,45 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
 basis set defined for all atoms in the system, or set |scf__df_scf_guess|
 to false, which disables this acceleration entirely.
 
+COSX Exchange
+~~~~~~~~~~~~~
+
+The semi-numerical COSX algorithm described in [Neese:2009:98]_ evaluates
+two-electron ERIs analytically over one electron coordinate and numerically
+over the other electron coordinate, and belongs to the family of pseudospectral 
+methods originally suggested by Friesner. In COSX, numerical integration is performed on standard
```
JonathonMisiewicz, line 760: I still don't agree with mentioning pseudospectral methods and Friesner. Non-experts have no clue who Friesner is and would be better served by reading the Neese paper. Experts already know this. For this reason, in neither case are we helping people who want to learn about the method or people in the "what-exactly-have-I-run-and-why stage."

psi4/src/psi4/libfock/jk.h
```diff
@@ -1191,6 +1192,95 @@ class PSI_API MemDFJK : public JK {
...
+    // => Semi-Numerical Stuff <= //
+
+    /// Small DFTGrid for initial SCF iterations
+    std::shared_ptr<DFTGrid> grid_init_;
+    /// Large DFTGrid for the final SCF iteration
+    std::shared_ptr<DFTGrid> grid_final_;
+    /// Overlap fitting metric for grid_initial_
+    SharedMatrix Q_init_;
+    /// Overlap fitting metric for grid_final_
+    SharedMatrix Q_final_;
```
JonathonMisiewicz, line 1228: None of these truly need to be a pointer. The one thing you need to do is change the definition of `grid` and `Q` to be ternaries, and then you can change the signature of `compute_numeric_overlap` to remove an annoying pointer.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+#include <algorithm>
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
+using namespace psi;
+
+namespace psi {
+
+Matrix compute_numeric_overlap(const std::shared_ptr<DFTGrid> &grid, const std::shared_ptr<BasisSet> &primary) {
```
JonathonMisiewicz, line 53: This function does no memory management of `primary`. Take `const BasisSet&`, not `const std::shared_ptr<BasisSet> &`. I make comments on `shared_ptr<DFTGrid>` elsewhere.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+
+    }
+
+    S_num.hermitivitize();
+
+    return S_num;
+
+}
+
+SharedMatrix compute_esp_bound(std::shared_ptr<BasisSet> primary) {
```
JonathonMisiewicz, line 107: More unnecessary pointers! Return `Matrix` and take in `const BasisSet&`.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+    timer_on("Coulomb Metric");
+    FittingMetric J_metric_obj(auxiliary_, true);
+    J_metric_obj.form_fitting_metric();
+    J_metric_ = J_metric_obj.get_metric();
+    timer_off("Coulomb Metric");
+
+    // pre-construct per-thread TwoBodyAOInt objects for computing 3-index ERIs
+    timer_on("ERI Computers");
+    eri_computers_.resize(nthreads_);
+    std::shared_ptr<BasisSet> zero = BasisSet::zero_ao_basis_set();
```
JonathonMisiewicz, line 180:
```suggestion
    auto zero = BasisSet::zero_ao_basis_set();
```

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -0,0 +1,1012 @@
...
+        // map back from global shell index to local shell index
+        std::map<size_t, size_t> shell_map_inv;
+        for (size_t shell_ind = 0; shell_ind < shell_map.size(); shell_ind++) {
+            shell_map_inv[shell_map[shell_ind]] = shell_ind;
+        }
+
+        // => Process Density Matrix <= //
+
+        // significant cols of D for this grid block
+        std::vector<SharedMatrix> D_block(njk);
```
JonathonMisiewicz, line 750: Why is this `std::vector<SharedMatrix>` rather than `std::vector<Matrix>`?

JonathonMisiewicz (commented):
I gave this one more quick pass, in light of how quickly SETCA approaches. Again, I'm nitpicking whether pointers are appropriate, and that one line int he documentation.

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: It wasn't obvious (to me at least) how to make this a `std::vector<Matrix>` without introducing extra copies.

For one, I had trouble initializing the vector with the `Matrix` constructor included. Any other way, and I think you'd have to initialize a `Matrix` and then separately push_back / copy it into the `std::vector`.

The other challenge is that the `linalg::doublet` function is later called with these matrices. This function has two signatures: either call it with two `SharedMatrix` objects and get a `SharedMatrix` back, or call it with two `Matrix` objects and get a `Matrix` back. The `F_block` variable that's created from the `doublet` call clearly needs to be a vector of `SharedMatrix`, so therefore `D_block` should also contain `SharedMatrix` objects.

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: I can certainly change the return type to `Matrix`, that makes sense.

The basis set is already stored as a `std::shared_ptr<BasisSet> primary_` on the JK class. To me, it seems cleaner to simply pass the already existing `shared_ptr`, rather than dereference it and pass it as a `const BasisSet&`. What's the reasoning for the latter?

psi4/src/psi4/libfock/jk.h
zachglick: Were we to make `grid_init_`, `grid_final_`, `Q_init_`, and `Q_final_` objects instead of pointers, how would you initialize those objects without incurring a copy?

psi4/src/psi4/libfock/DFJCOSK.cc
JonathonMisiewicz: I prefer references to pointers when at all possible. As written, the function _only_ works for a `shared_ptr`. If I need to compute the bound with a `BasisSet`, `unique_ptr<BasisSet`, or `* BasisSet`, I can't use this function without some kind of extra work. While it's true that you need some extra work to de-reference it, it's as simple as passing in `*varname` rather than `varname`.

The other advantage is that we can now enforce `const` on the incoming `BasisSet` easily.

psi4/src/psi4/libfock/jk.h
JonathonMisiewicz: If you mean "initialize `grid_init_`", you initialize it when you construct a new object. The use of pointers isn't saving you anything.
If you mean "initialize `grid`", then `const& grid = early_screening_ ? grid_init_ : grid_final_` should work.

psi4/src/psi4/libfock/DFJCOSK.cc
JonathonMisiewicz: > For one, I had trouble initializing the vector with the `Matrix` constructor included. Any other way, and I think you'd have to initialize a `Matrix` and then separately push_back / copy it into the `std::vector`.

What's wrong with `D_block[jki] = Matrix(nbf_block_all, nbf_block);`?

> The `F_block` variable that's created from the `doublet` call clearly needs to be a vector of `SharedMatrix`

This isn't clear to me at all. Why can't it be `std::vector<Matrix>`?

psi4/src/psi4/libfock/jk.h
zachglick: I'm talking about the former.

When `grid_init_` is a `shared_ptr<DFTGrid>`, I can write:
```
grid_init_ = std::make_shared<DFTGrid>(...);
```

But when `grid_init_` is a `DFTGrid`, I have to write:
```
grid_init_ = DFTGrid(...);
```
which from my understanding first creates a grid and then assigns it to `grid_init_` (requiring a copy). RVO doesn't apply here, because `grid_init_` is already defined; RVO only works for initialization.

psi4/src/psi4/libfock/DFJCOSK.cc
zachglick: I was under the impression that the assignment:
```
D_block[jki] = Matrix(nbf_block_all, nbf_block);
```
should be avoided because it results in a redundant copy, same with the initialization of the elements of `F_block`

psi4/src/psi4/libfock/jk.h
JonathonMisiewicz: [Copy-assignment, not copy](https://stackoverflow.com/questions/37975574/what-happens-when-a-new-object-is-created-using-another-existing-object), but your point stands. (And I have learned something new today.) _Technically_ it's more efficient to use a `unique_ptr` than a `shared_ptr`, but I can't imagine that's significant for this application.

psi4/src/psi4/libfock/DFJCOSK.cc
JonathonMisiewicz: You're right, per earlier comment. You _could_ use `emplace_back` instead of assigning the size of `D_block` in advance, but the extra resizes means that performance is [probably slightly better](https://stackoverflow.com/questions/65936284/performance-of-stdvectoremplace-back-vs-assignment-for-pod-struct) with the current version.

Oh, C++, why do you make things that should be simple so very technical.

JonathonMisiewicz (approved):
Thanks for putting up with some of my pendatry during the review process, and thanks for the code. This should be a great addition!

I'll leave merging to you.

PR #2574, opened 2022-05-11
Update diis.py

JonathonMisiewicz:
## Description
Typo fix from revision cycles.

@behnle: For future, please file an issue if you spot a problem like this again. It's harder for us to miss. 🙂 

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):

behnle:
Agreed, i just thought that this is not a big issue. Regular users will never see this as whenever this exception is triggered, the program is broken, anyway :-)

zachglick (approved):

PR #2575, opened 2022-05-11
Refactor Recursive Driver Into Distributed Driver, Again

loriab:
## Description
Change the Psi4 driver from being recursive to being based on compute classes that have planning, computing and querying functions. Then, integrate the new driver with the QCArchive project for parallelizing nbody, CBS and findif.

Don't panic -- 55k of the lines added are an output.ref.

See #1351 for history and bite-sized extractions. What's left here is the gory core. Because code is moving around, there's no good way to review it. The plain code file is often easier to read than the diff. Below are the 6 key files.
* `driver.py` — mostly deletions. notice that the main e/g/h functions instead of having several layers when you enter and "bounce" to other fns now have a single plan/compute/proceed-to-analytic decision step.
* `task_planner.py` — new file. all that complexity of figuring out which of cbs/findif/nbody are active (including simple things like `energy("mp2/cc-pvdz")` parsing) has landed here. at least it's no longer repeated in several functions. the main goal is to take user info from where user is accustomed to supplying it (options or kwargs) and use it to initialize a tree of "computers" whose leaves are all AtomicComputers (analytic single-points). you can see some examples of the planner's work in `tests/pytests/test_task_planner.py`
* `task_base.py` — new file. fairly simple. how to store the pieces of a single job in a qcschema AtomicInput and how to call qcengine to run psi4 (in api, not cli, mode) to turn that into an AtomicResult. all jobs that aren't analytic single-points pass through this file.
* `driver_findif.py` — moderately modified, mostly additions. this is the least-changed wrapper. a pydantic class has been added to link the planning and assembly steps instead of a function and code that used to live in driver.py. the main idea for wrappers is that everything should be generic python except for translation from qcschema to psi4.wfn at the end (finidf does poorly in this b/c salcs call libmints). at the top of the file is a docstring outlining the procedure. all the wrappers have a mix of class and free functions for historical reasons. this can certainly be refactored once rebasing isn't an issue.
* `driver_cbs.py` — big modifications, mostly additions, but changes are mostly moving code around and renaming big dicts (`need` isn't too informative) and trying to narrow the points of data transfer. again, there's an outline at the top of the file.
* `driver_nbody.py` — big modifications, might be net deletion of code lines. Lots of renaming of variables. Some bugs fixed -- I can only hope more than were added. Docstrings and code outline should be current. I didn't rework multilevel (separate file) except to make gradient run and allow different order of keys in user `levels` kwarg. I wouldn't bother with the diff. 

## Todos

- [x] `msgpack-python` added as req'd dependency. this keeps numpy arrays serialized when communicating in schema
- [x] any fns mirroring `sherrill_gold_standard` or `allen_focal_point` need to be reformatted and registered with `register_composite_function`
- [x] extrapolation functions for composite need to be registered with register_xtpl_function
- [x] CBS and xtpl functions need to be called as strings, not objects. So `energy('cbs', scf_wfn='scf', scf_basis='cc-pV[DTQ]Z', scf_scheme='scf_xtpl_helgaker_3')` not `energy(cbs, scf_wfn='scf', scf_basis='cc-pV[DTQ]Z', scf_scheme=scf_xtpl_helgaker_3)`.
- [x] no longer always evaluating gradient before Hessian to test safety of projecting rotations. Instead assuming unsafe. Can pass ref_gradient array to test on or `set findif fd_project` T/F explicitly to control.
- [x] `json_ret = psi4.schema_wrapper.run_qcschema(json_input)` not `json_ret = psi4.json_wrapper.run_json(json_input)`. json_ret is now an object, so `.` access, not dictionary.  The old schema runner uses a schema circa pre-2018 and we've warned it's deprecated for about a year, so I went ahead and updated all the tests to the new/current run_qcschema.
- [x] Pre-ddd, CBS wrapper issues a `clean()` btwn calcs. Plain string modelchem calcs like `energy('hf/cc-pvdz')` were getting caught and also being cleaned, meaning their behavior was slightly different than `set basis cc-pvdz \n energy('hf')`. With ddd, this is no longer happening, so for occasional string modelchem calcs (that is, `energy|gradient|hessian("mtd/bas")` only), you may need to add a clean in the input. A few test cases now sport extra cleans.

Signatures of this problem are PSIO errors and, in an obscure case, CC amplitudes being all zeros when called as mtd/bas. An example is shown below, collected by modifying steps 6 & 7 of `cbs-xtpl-energy-conv` test:

|                                     | master (pre-v1.6)                           | ddd (v1.6)                                  |
|-------------------------------------|---------------------------------------------|---------------------------------------------|
| set basis bas; energy(mtd); clean() | ok                                          | ok                                          |
| set basis bas; energy(mtd)          | `PSIO_ERROR: (Incorrect block end address)` | `PSIO_ERROR: (Incorrect block end address)` |
| energy(mtd/bas); clean()            | ok                                          | ok                                          |
| energy(mtd/bas)                     | ok                                          | `PSIO_ERROR: (Incorrect block end address)` |

- [x] Output file in continuous mode now looks much as it normally does in the master branch, except there's a few more headers printed and the individual cbs/findif/nbody `AtomicResult` outputs aren't added in real time. Running through QCFractal, the output file is expected to break down, but this hasn't been tried.
- [x] A Logging file has been introduced. All its contents should be considered experimental and subject to change.
- [x] Dipole derivatives through finite difference of dipoles added, so IR intensities available
- [x] The dummy wavefunctions returned from wrappers always had some questionable properties (e.g., single method for mixed method, largeset HF basis for cbs wavefunction) can be even less realistic (e.g., always def2-svp basis for cbs).
- [x] n-body qcvars renamed (grace period applies)
```
    "CP-CORRECTED {nb}-BODY INTERACTION ENERGY": ("CP-CORRECTED INTERACTION ENERGY THROUGH {nb}-BODY", 1.7),
    "NOCP-CORRECTED {nb}-BODY INTERACTION ENERGY": ("NOCP-CORRECTED INTERACTION ENERGY THROUGH {nb}-BODY", 1.7),
    "VMFC-CORRECTED {nb}-BODY INTERACTION ENERGY": ("VMFC-CORRECTED INTERACTION ENERGY THROUGH {nb}-BODY", 1.7),
    "COUNTERPOISE CORRECTED TOTAL ENERGY": ("CP-CORRECTED TOTAL ENERGY", 1.7),
    "COUNTERPOISE CORRECTED INTERACTION ENERGY": ("CP-CORRECTED INTERACTION ENERGY", 1.7),
    "NON-COUNTERPOISE CORRECTED TOTAL ENERGY": ("NOCP-CORRECTED TOTAL ENERGY", 1.7),
    "NON-COUNTERPOISE CORRECTED INTERACTION ENERGY": ("NOCP-CORRECTED INTERACTION ENERGY", 1.7),
    "VALIRON-MAYER FUNCTION COUTERPOISE TOTAL ENERGY": ("VALIRON-MAYER FUNCTION COUNTERPOISE TOTAL ENERGY", 1.7),  # note misspelling
    "VALIRON-MAYER FUNCTION COUTERPOISE INTERACTION ENERGY": ("VMFC-CORRECTED INTERACTION ENERGY", 1.7),
``` 
- [x] added managed mode for scf gradient. it's not wholly effective at consolidating code, though, since only triggers for `gradient("hf")`, not `gradient(correlated)`. digits to compare has to be relaxed, but this has been seen with other qc programs, too.
- [x] Distributed driver should not be advertised for use with QCFractal at this time. It hasn't been tested in a while, and mem, cores, etc. setting may need tweaking in a non-continuous mode. Use it experimentally and report back.
- [x] n-body wrapper can no longer do embedding with internally calculated Mulliken charges. Charges must now be provided with  `embedding_charges` kwarg.
- [x] stdsuite: consolidated the caller fn to remove duplicate code. added HF Hessian testing, through DF can't be uncommented until refs released in qcng.
- [x] the addition of ccsd iterations qcvar to dfocc could maybe be reverted

## Questions
- [x] Any critical eye or independent tests that you can bring to the manybody wrapper much appreciated, @jeffschriber.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


zachglick (commented):
Thanks for all of your hard work Lori. I still need to go through `task_planner.py` and `driver_nbody.py` more carefully.

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1293 @@
...
+
+logger = logging.getLogger(__name__)
 
-### Math helper functions
+SubTaskComputers = Union["AtomicComputer", "CompositeComputer", "FiniteDifferenceComputer"]
+FragBasIndex = Tuple[Tuple[int], Tuple[int]]
 
 
-def nCr(n, r):
+def nCr(n: int, r: int) -> int:
```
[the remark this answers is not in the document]
loriab: aha, because math.comb was new in 3.8, and this was written c. 3.5. thanks, I'll switch it over.

psi4/driver/task_base.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # EITHER ...
+        # from psi4.driver import schema_wrapper
+        # self.result = schema_wrapper.run_qcschema(self.plan())
+        # ... OR ...
+        self.result = qcng.compute(
+            self.plan(),
+            "psi4",
+            raise_error=True,
+            # local_options below suitable for continuous mode
```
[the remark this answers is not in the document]
loriab: vs. distributed in qcfractal. I'll add some elaboration.

psi4/driver/task_base.py
```diff
@@ -0,0 +1,203 @@
...
+            local_options={
+                "memory": core.get_memory() / (2 ** 30),
```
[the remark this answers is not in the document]
loriab: good idea

psi4/driver/aliases.py
```diff
@@ -85,7 +82,7 @@ def fake_file11(wfn, filename='fake_file11.dat', **kwargs):
                 '', gradient.get(at, 0), gradient.get(at, 1), gradient.get(at, 2)))
 
 
-def sherrill_gold_standard(func, label, **kwargs):
+def sherrill_gold_standard(**kwargs):
     r"""Function to call the quantum chemical method known as 'Gold Standard'
```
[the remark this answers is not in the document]
loriab: true. docstrings have always been awkward with the wrappers since there isn't usually a centralized function. I may have a strategy, but it's not immediate. papered over the trouble for now, I think.

psi4/driver/task_planner.py
```diff
@@ -0,0 +1,256 @@
...
+        A simple (AtomicComputer) or Layered (CompositeComputer, FiniteDifferenceComputer, ManyBodyComputer) task object.
+        Layered objects contain many and multiple types of computers in a graph.
+
+    """
+
+    # Only pull the changed options
+    keywords = p4util.prepare_options_for_set_options()
+
+    keywords["function_kwargs"] = {}
+    if "external_potentials" in kwargs and "bsse_type" not in kwargs:
```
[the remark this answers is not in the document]
loriab: I think it's legacy from before external_potentials and embedding_charges got their uses (nbody charges only (Z) user-facing, nbody full Zxyz internal, full Zxyz normal calc, SAPT dict-o-Zxyz) sorted out. Everything now works fine w/o the `and "bsse_type" note in kwargs` clause.

psi4/driver/procrouting/proc.py
```diff
@@ -2585,10 +2634,11 @@ def run_scf_hessian(name, **kwargs):
     H = core.scfhess(ref_wfn)
     ref_wfn.set_hessian(H)
 
-    # Clearly, add some logic when the reach of this fn expands
-    ref_wfn.set_variable("HF TOTAL HESSIAN", H)  # P::e SCF
     ref_wfn.set_variable("SCF TOTAL HESSIAN", H)  # P::e SCF
-    core.set_variable("SCF TOTAL HESSIAN", H)  # P::e SCF
+    if ref_wfn.functional().needs_xc():
+        ref_wfn.set_variable("DFT TOTAL HESSIAN", H)  # no DFT hess at present. overwritten later for DH -- TODO when DH gradients  # P::e SCF
```
JonathonMisiewicz, line 2639: I thought we had LDA hessians.

psi4/driver/procrouting/proc.py
```diff
@@ -62,6 +63,31 @@
...
+    """Function selecting the algorithm for an SCF gradient call
+    and directing to specified or best-performance default modules.
+
+    """
+    reference = core.get_option('SCF', 'REFERENCE')
+    mtd_type = core.get_global_option('SCF_TYPE')
+    module = core.get_global_option('QC_MODULE')
+    # Considering only scf
+
+    if mtd_type == 'CD':
```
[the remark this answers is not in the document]
JonathonMisiewicz: Add a comment about this.

psi4/driver/driver_findif.py
```diff
@@ -25,20 +25,129 @@
 #
 # @END LICENSE
 #
+"""Plan, run, and assemble QC tasks to obtain derivatives by finite difference of lesser derivatives.
```
JonathonMisiewicz, line 28: What is this style of docstring? I'm not familiar with it and find it difficult to read. What is trying to accomplish?

JonathonMisiewicz (commented):
First round comments from me. I've avoided the big functions so far.

psi4/driver/aliases.py
```diff
@@ -85,11 +82,11 @@ def fake_file11(wfn, filename='fake_file11.dat', **kwargs):
                 '', gradient.get(at, 0), gradient.get(at, 1), gradient.get(at, 2)))
 
 
-def sherrill_gold_standard(func, label, **kwargs):
+def sherrill_gold_standard(**kwargs):
```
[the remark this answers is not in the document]
loriab: sure, its the dict spec for cbs

psi4/src/psi4/dfocc/ccsd_iterations.cc
```diff
@@ -27,6 +27,7 @@
  */
 
 #include "psi4/libqt/qt.h"
+#include "psi4/libpsi4util/process.h"
```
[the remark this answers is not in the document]
loriab: those were from a more hopeful era. I've tried a full tests, and they're not needed. will require squash to not complicate dfocc sync.

samples/python/cc-amps/input.py
```diff
@@ -1,127 +0,0 @@
-#! API access to CCSD amplitudes
```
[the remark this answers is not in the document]
loriab: I don't recall when/why the sample was deleted, but it'll regenerate when samples/ gets updated.

tests/pytests/test_nbody_multi_level_qcschema.py
```diff
@@ -28,10 +28,6 @@ def base_schema():
     _base_schema = {
         'schema_name': 'qcschema_input',
         'schema_version': 1,
-        "extras": {
```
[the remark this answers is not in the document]
loriab: the extras-less schema in ddd is what it's supposed to be. when I pulled the test out to master, the extras had to be added. those keywords get passed automatically in task_base.py in ddd.

tests/pytests/test_qcschema.py
```diff
@@ -52,7 +52,7 @@ def test_qcschema_energy(result_data_fixture):
     assert compare_integers(True, 'beer' in ret.stdout, "Stdout Beer")
 
     # Check Array data
-    assert compare_integers(True, isinstance(ret.extras["qcvars"]["MAYER_INDICES"], np.ndarray),
+    assert compare_integers(True, isinstance(ret.extras["qcvars"]["MAYER INDICES"], np.ndarray),
```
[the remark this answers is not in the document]
loriab: We changed it last year https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L629 . With underscore it conforms to the oeprop task but without underscore it conforms to the qcvars-use-spaces rule.

tests/pytests/test_standard_suite.py
```diff
@@ -2502,13 +2140,38 @@ def test_olccd_energy_module(inp, dertype, basis, subjects, clsd_open_pmols, req
     ],
 )
 def test_olccd_energy_default(inp, dertype, basis, subjects, clsd_open_pmols, request):
-    method = "olccd"
+    runner_asserter(*_processor(inp, dertype, basis, subjects, clsd_open_pmols, request, "olccd", "energy"))
+
+
+def _processor(inp, dertype, basis, subjects, clsd_open_pmols, request, method, driver):
```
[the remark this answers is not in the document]
loriab: Yes, it's the same cleanup I've applied to qcdb an qcng standard suite. Avoids repeating the same 15 lines for each test fn. It does make energy tests a little more complicated since errors need to be in a dict to behave like grad/hess, but that's a one-time cost.

tests/pytests/standard_suite_runner.py
```diff
@@ -60,7 +60,7 @@ def runner_asserter(inp, subject, method, basis, tnm):
 
     atol = 1.0e-6
     if driver == "hessian":
-        atol = 2.0e-6  # todo implement more elaborate e/g/h atol like at qcdb & qcng: https://github.com/qcdb/qcdb/blob/master/qcdb/tests/standard_suite_runner.py#L144-L152
+        atol = 1.0e-5  # todo implement more elaborate e/g/h atol like at qcdb & qcng: https://github.com/qcdb/qcdb/blob/master/qcdb/tests/standard_suite_runner.py#L144-L152
```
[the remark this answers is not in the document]
loriab: Nothing. I added DF Hessians. I've found Hessians generally need a both abs and rel tolerance https://github.com/qcdb/qcdb/blob/master/qcdb/tests/standard_suite_runner.py#L149-L151, but I didn't want to rework the file here yet.

psi4/driver/procrouting/proc.py
loriab: I remember a comment along those lines, but it didn't agree with my memory. We don't test any dft hess. In any case, the code edit will work better with DFT.

psi4/driver/driver_findif.py
loriab: There's no particular style to the module-level docstring. It's trying to convey what the wrapper does in a start-to-finish manner that was lost with the class conversion. All bullets at an indentation level happen in the function name of that indentation level. An indent means a called fn that could be either in-class or free.

psi4/driver/procrouting/proc.py
JonathonMisiewicz: But we do test the DFT frequencies _from_ an analytic hessian in `dft-freq-analytic`.

The more I look at this like, the more confused I am. What _is_ DH?

psi4/driver/driver_findif.py
JonathonMisiewicz: So you're saying this is supposed to show a linear path through the class?

loriab: yes, from initializing the class (happens in task_planner.py) through to returning the wfn (happens when the driver,py fns call get_psi_results().

psi4/driver/driver_findif.py
```diff
@@ -25,20 +25,129 @@
...
+FiniteDifferenceComputer.get_psi_results()
+------------------------------------------
+
+    Computer.get_results()
+    ----------------------
+
+        Computer._prepare_results()
+        ---------------------------
+        * get_results() for each job in task list
+        * arrange atres data into e/d/g/h fields as available on each of reference and displacements entries
```
JonathonMisiewicz, line 90: Is `d` dipoles?

JonathonMisiewicz, line 90: Unclear what `atres` is.

psi4/driver/driver_findif.py
```diff
@@ -25,20 +25,129 @@
...
+            ---------------------------------
+            assemble_dipder_from_dipoles()
+            ------------------------------
+
+            * form DD, G, H from lower derivative points
+
+        * place as many of DD, G, H as available onto reference entry
+
+    * pull qcvars off reference job
+    * from reference job set add'l mol, DD, G, H as available
```
JonathonMisiewicz, line 115:
```suggestion
    * from reference job, set add'l mol, DD, G, H as available
```

psi4/driver/driver_findif.py
```diff
@@ -1017,65 +1126,371 @@ def init_string(data):
...
+        return driver
+
+    @validator('molecule')
+    def set_molecule(cls, mol):
+        mol.update_geometry()
+        mol.fix_com(True)
+        mol.fix_orientation(True)
+        return mol
+
+    def __init__(self, **data):
```
JonathonMisiewicz, line 1154: A comment on `data` would be good. It looks like this is a massive list of keyword arguments. Some keyword arguments control findif operation, while others control the individual electronic structure singlepoints.

psi4/driver/procrouting/proc.py
loriab: I had forgotten that. comment edited to `# overwritten later for DH -- TODO when DH Hessians # P::e SCF` to mirror the gradient line above.

psi4/driver/driver_findif.py
```diff
@@ -1017,65 +1126,371 @@ def init_string(data):
...
+        passalong = {k: v for k, v in data.items() if k not in packet}
+        passalong.pop('ptype', None)
+
+        self.task_list["reference"] = self.computer(**packet, **passalong)
+
+        parent_group = self.molecule.point_group()
+        for label, displacement in self.findifrec["displacements"].items():
+            clone = self.molecule.clone()
+            clone.reinterpret_coordentry(False)
+            #clone.fix_orientation(True)
```
JonathonMisiewicz, line 1240: Why do we need one but not the other?

psi4/driver/driver_findif.py
```diff
@@ -1017,65 +1126,371 @@ def init_string(data):
...
+            H0 = assemble_hessian_from_gradients(self.findifrec, self.metameta['irrep'])
+            self.findifrec["reference"][self.driver.name] = H0
+
+        elif self.metameta['mode'] == '2_0':
+            try:
+                G0 = assemble_gradient_from_energies(self.findifrec)
+            except KeyError:
+                core.print_out("Unable to construct reference gradient from Hessian displacements.")
+                # TODO: this happens properly when a subset irrep doesn't
+                #  have the right displacements for grad. For both this case
```
JonathonMisiewicz, line 1354: "subset irrep doesn't have the right displacements for grad" is opaque. I think you mean "when requested symmetry block of displacements don't have the totally symmetric gradient displacements"?

psi4/driver/driver_findif.py
```diff
@@ -1017,65 +1126,371 @@ def init_string(data):
...
+    sbasis = "def2-svp" if (findif_model.model.basis == "(auto)") else findif_model.model.basis
+    basis = core.BasisSet.build(mol, "ORBITAL", sbasis, quiet=True)
+    wfn = core.Wavefunction(mol, basis)
+    if hasattr(findif_model.provenance, "module"):
+        wfn.set_module(findif_model.provenance.module)
+
+    # setting CURRENT E/G/H on wfn below catches Wfn.energy_, gradient_, hessian_
+    # setting CURRENT E/G/H on core below is authoritative P::e record
+    for qcv, val in findif_model.extras["qcvars"].items():
+        for obj in [core, wfn]:
```
JonathonMisiewicz, line 1490: Again, set is more appropriate than list.

JonathonMisiewicz (commented):
findif comments.

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
 #
 # @END LICENSE
 #
+"""Plan, run, and assemble QC tasks to obtain composite method, basis, & options treatments.
+
+===
+CBS
```
JonathonMisiewicz, line 31: Repeat my request for additional user guidance, e.g., "CBS Flow", that I made on FINDIF.

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+===
+
+cbs_text_parser()
```
JonathonMisiewicz, line 34: This may or may not fire depending on whether the use passed a string or kwargs, right?

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+* transform user string into cbs kwargs inc'l basic cbs_metadata
+
+
+----------------------------
+CompositeComputer.__init__()
+----------------------------
+
+    _process_cbs_kwargs()
+    ---------------------
+    * transform user kwargs into trial cbs_metadata format (aka dict spec)
```
JonathonMisiewicz, line 50: But these may not be user kwargs, if they were generated by `cbs_text_parser`?

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+    _build_cbs_compute()
+    --------------------
+
+        _expand_scheme_orders()
+        -----------------------
+        * form f_fields dict of entries for each zeta in a scheme (single NEED; entries related by nonlinear fn)
+
+        _contract_bracketed_basis()
+        ---------------------------
+        * form basis abbr. string from basis seq
```
JonathonMisiewicz, line 76: What's a basis sequence? Do you mean `cc-pV[345]Z` from `cc-pVTZ`, `cc-pVQZ`, `cc-pV5Z`?

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+
+    * form d_fields list of stages or stage halves from NEEDs (GRAND_NEED; items related linearly to form final val)
```
JonathonMisiewicz, line 78: What's a stage?

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+    * form list of entries (entry:= mtd-bas-opt) mentioned in GRAND_NEED (MODELCHEM; redundant, naive)
```
JonathonMisiewicz, line 79: What's the significance of the parenthetical?

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+CompositeComputer.get_psi_results()
+-----------------------------------
+
+    Computer.get_results()
+    ----------------------
+
+        Computer._prepare_results()
+        ---------------------------
+        * get_results() for each job in task list
+        * arrange atres data into e/d/g/h fields in compute_list and copy them into cbs tables
```
JonathonMisiewicz, line 106: Still not sure what `atres` is. Same with `TROVE` below. And also like before, is D down lower `dipole`?

JonathonMisiewicz (commented):
"Big picture docs pass" for CBS. I love the idea, but it's still not clear to me, as written.

psi4/driver/driver_findif.py
```diff
@@ -25,20 +25,129 @@
...
+    ----------------------------------
+
+        _geom_generator()
+        -----------------
+
+            _initialize_findif()
+            --------------------
+            * initialize CdSalcs, partition them per irrep, apply user irreps
+
+        * start the governing dict findifrec with parameters, size, mol
```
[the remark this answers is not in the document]
loriab: I'm not finding the difference here

psi4/driver/driver_findif.py
loriab: `atres` is a common variable name I choose for `qcel.models.AtomicResult` like `atin` for `AtomicInput`. I've written it out, now.

psi4/driver/driver_findif.py
loriab: I've added some notes, but these classes are still a bit of a free-for-all. All three pydantic classes need two passes: one to formalize the needed fields and add docstrings. this has happened for the manybody class. second is to forbid extra fields (see Config of base BaseComputer to see how) and find out what's getting attached and either formalize them or otherwise handle better. After those passes (or with them) a formal qcschema model for the procedures should be formed, then the Computer becomes fairly light around the Input/Result (roughly akin to AtomicComputer vs AtomicInput/AtomicResult). The pydantic objects here are fairly primitive, as they were written before the qcel ones got started, much less got developed, so these have some catching-up to do.

psi4/driver/driver_findif.py
loriab: It's usually fix_com, fix_orientation that go together, right? those two get set upon molecule validation in the class init. I probably kept the comment because line was active in the original src, and I wasn't sure if there were circs where it was needed again.

psi4/driver/driver_findif.py
loriab: Either works, but I like the ordered aspect of list (esp. for debug printing) and tend to reserve set for when uniqueness definitely wanted.

psi4/driver/driver_cbs.py
loriab: pretty much. `cbs_text_parser` is called from task_planner and fires whenever `"/"` is in the method string. it transforms into a set of cbs kwargs that may (if one of the kwargs is a cbs dict spec) trigger a subsequent composite wrapper invocation or may not (it might have been a "b3lyp/6-31G").

later on, called from the composite `__init__` is `_process_cbs_kwargs()`. that fires trivially when the user passed a cbs dict spec. or it does real work when the user passed cbs kwargs.

I've added some clarifying bullets.

psi4/driver/driver_cbs.py
loriab: yes, user specified but not necessarily user kwargs, if processed through `cbs_text_parser`

psi4/driver/driver_cbs.py
```diff
@@ -25,34 +25,152 @@
...
+        * check and supply defaults for cbs_metadata format (various calls to above two fns)
+
+* BaseComputer.__init__()
+
+    _build_cbs_compute()
+    --------------------
+
+        _expand_scheme_orders()
+        -----------------------
+        * form f_fields dict of entries for each zeta in a scheme (single NEED; entries related by nonlinear fn)
```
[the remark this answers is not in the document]
loriab: the latter. I've added your text

psi4/driver/driver_cbs.py
loriab: yes, exactly

psi4/driver/driver_cbs.py
loriab: stages are the usual decomposition of a composite method -- scf part, correlated part, deltas. I've added a definition.

psi4/driver/driver_cbs.py
loriab: it's defining a term used in further comments. I've expanded to ` (entry:= mtd-bas-opt specification)`

psi4/driver/driver_cbs.py
loriab: atres, d, dg now expanded or defined. `TROVE` is also defined somewhere in the file. It's the bundle of all the calcs you've got, whether you needed each piece or not (think ccsd, when the delta was btwn ccsd(t) and mp2).

psi4/driver/driver_findif.py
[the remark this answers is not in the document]
JonathonMisiewicz: There isn't one - artifact of editing my review.

psi4/driver/task_base.py
```diff
@@ -0,0 +1,204 @@
...
+                return self.result
+
+            self.result = result[0]
+            return self.result
+
+
+def _drink_filter(stdout: str) -> str:
+    """Don't mess up the widespread ``grep beer`` test of Psi4 doneness by printing multiple drinks per outfile."""
+
+    stdout = stdout.replace("\n*** Psi4 exiting successfully. Buy a developer a beer!", "")
```
JonathonMisiewicz, line 202: Why does this not create a problem by removing the very last one?

psi4/driver/task_planner.py
```diff
@@ -0,0 +1,257 @@
...
+from psi4.driver.driver_nbody import ManyBodyComputer
+from psi4.driver.driver_cbs import CompositeComputer, composite_procedures, cbs_text_parser
+from psi4.driver.driver_util import negotiate_derivative_type, negotiate_convergence_criterion
+
+logger = logging.getLogger(__name__)
+
+TaskComputers = Union[AtomicComputer, CompositeComputer, FiniteDifferenceComputer, ManyBodyComputer]
+
+
+def _expand_cbs_methods(method: str, basis: str, driver: DriverEnum, **kwargs) -> Tuple[str, str, Dict]:
```
JonathonMisiewicz, line 49: docstring to say what the function does, please. I can guess based on the name.

psi4/driver/task_planner.py
```diff
@@ -0,0 +1,257 @@
...
+        "cbs"
+    molecule
+        A Psi4 base molecule to use
+    kwargs
+        User keyword arguments, often used to configure task computers.
+
+    Returns
+    -------
+    computer
+        A simple (AtomicComputer) or Layered (CompositeComputer, FiniteDifferenceComputer, ManyBodyComputer) task object.
```
JonathonMisiewicz, line 94: Is "Layered" the name of an abstract base class? Otherwise, I'm not sure why this is capitalized.

psi4/driver/task_planner.py
```diff
@@ -0,0 +1,257 @@
...
+    # Check for CBS
+    elif method == "cbs":
+        kwargs.update(cbsmeta)
+        # This CompositeComputer is discarded after being used for dermode. Could have used directly for analytic except for excess printing with FD
+        simplekwargs = copy.deepcopy(kwargs)
+        simplekwargs['verbose'] = 0
+        dummyplan = CompositeComputer(**packet, **simplekwargs)
+
+        methods = [sr.method for sr in dummyplan.task_list]
+        # TODO: pass more info, so fn can use for managed_methods -- ref, qc_module, fc/ae, conv/df
```
JonathonMisiewicz, line 224: Is not having this done costing some of our current abilities?

JonathonMisiewicz (commented):
Task computer and bases. At this point, I'm largely relying on the test suite passing and checking for no glaring problems. Saving up my attention for CBS and nbody.

psi4/driver/driver_cbs.py
JonathonMisiewicz: `atres` not defined or expanded here. You use this elsewhere in this file, so defining is probably best. Thanks much for the others, though.

psi4/driver/driver_cbs.py
```diff
@@ -871,392 +996,59 @@ def cbs(func, label, **kwargs):
...
+##  Aliases  ##
+complete_basis_set = cbs
 
+######### OUTSOURCE
+######### PARSE / PREPARE
+######### PREPARE / COMPUTE
 
-######### COMPUTE / ASSEMBLE
-######### ASSEMBLE / REPORT
+#    #psioh = core.IOManager.shared_object()
```
JonathonMisiewicz, line 1031: What is this commented out section doing here?

psi4/driver/driver_cbs.py
```diff
@@ -1497,11 +1228,577 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
+        dups = -1
+        for indx_job, job in enumerate(JOBS):
+            #print('')
+            #print(job['f_wfn'], mc['f_wfn'], job['f_wfn'] == mc['f_wfn'])
+            #print(job['f_basis'], mc['f_basis'], job['f_basis'] == mc['f_basis'])
+            #print(job['f_options'], mc['f_options'], job['f_options'] == mc['f_options'],
+            #      job['f_options'] is not False)
+            if ((job['f_wfn'] == mc['f_wfn']) and (job['f_basis'] == mc['f_basis'])
+                    and (job['f_options'] == mc['f_options'])):  # and (job['f_options'] is not False)):
+                # TODO: I don't understand why I need this change
```
JonathonMisiewicz, line 1307: What change, exactly?

psi4/driver/driver_cbs.py
```diff
@@ -1497,11 +1228,577 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
+
+    # Apply chemical reasoning to choose the minimum computations to run
+    JOBS = MODELCHEM[:]
+    listfmt = """   {:>12} / {:24} for  {}{}\n"""
+
+    #     Remove duplicate modelchem portion listings
+    for mc in MODELCHEM:
+        dups = -1
+        for indx_job, job in enumerate(JOBS):
+            #print('')
```
JonathonMisiewicz, line 1300: Debug lines that should have been removed?

psi4/driver/driver_cbs.py
```diff
@@ -1497,11 +1228,577 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
+    #     Remove chemically subsumed modelchem portion listings
+    if ptype == 'energy':
+        for mc in MODELCHEM:
+            for wfn in VARH[mc['f_wfn']]:
+                for indx_job, job in enumerate(JOBS):
+                    if ((VARH[mc['f_wfn']][wfn] == VARH[job['f_wfn']][job['f_wfn']])
+                            and (mc['f_basis'] == job['f_basis'])
+                            and not (mc['f_wfn'] == job['f_wfn'])  # and (mc['f_options'] is False)):
+                            and (mc['f_options'] == job['f_options'])):
+                        # TODO: I don't understand why I need this change
```
JonathonMisiewicz, line 1326: Which change?

psi4/driver/driver_cbs.py
```diff
@@ -1497,11 +1228,577 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
+        with p4util.hold_options_state():
+            for t in reversed(self.task_list):
+                t.compute(client=client)
+
+    def _prepare_results(self, client: Optional["FractalClient"] = None):
+        results_list = [x.get_results(client=client) for x in self.task_list]
+
+        modules = [getattr(v.provenance, "module", None) for v in results_list]
+        if self.driver != "energy" and len(set(modules)) == 2 and modules.count("scf") == len(modules) / 2:
+            # signature of "MP2 GRAD" - "HF GRAD" implementation detail
```
JonathonMisiewicz, line 1617: Can you elaborate?

JonathonMisiewicz (commented):
CBS comments... It's certainly easier to read now.

psi4/driver/driver_nbody_multilevel.py
```diff
@@ -0,0 +1,146 @@
...
+    # Initialize with zeros
+    energy_result, gradient_result, hessian_result = 0, None, None
+    energy_body_contribution = {b: {} for b in self.bsse_type}
+    energy_body_dict = {b: {} for b in self.bsse_type}
+    if ptype in ['gradient', 'hessian']:
+        gradient_result = np.zeros((natoms, 3))
+    if ptype == 'hessian':
+        hessian_result = np.zeros((natoms * 3, natoms * 3))
+
+    # Get numerical label for supersystem tasks
```
JonathonMisiewicz, line 53: How do you determine an appropriate label?

psi4/driver/driver_nbody_multilevel.py
```diff
@@ -0,0 +1,146 @@
...
+    sup_level = 0
+    levels = []
+    for n,i in enumerate(self.nbodies_per_mc_level):
+        if 'supersystem' not in i:
+            levels.append(int(n+1))
+        else:
+            sup_level = n+1
+
+    nbody_list = self.nbodies_per_mc_level
+    quiet=True
```
JonathonMisiewicz, line 63:
```suggestion
    quiet = True
```

JonathonMisiewicz (commented):
Started on n-body, but I can't continue. I'm not familiar with what a "level" is. Do you have a reference on that? Maybe I don't know where to look, but I can't find n-body explained in the docs.

psi4/driver/task_planner.py
loriab: Added 

```def _expand_cbs_methods(method: str, basis: str, driver: DriverEnum, **kwargs) -> Tuple[str, str, Dict]:
    """Sort out the user input method string into recognized fields. Handles cases like:

    * (i) `"mp2"` --> passes through;
    * (ii) `"mp2/cc-pvdz"` --> broken into method and basis fields;
    * (iii) `"mp2/cc-pv[d,t]z"` --> processed into method="cbs" and CBSMetadata spec;
    * (iv) `method="cbs", cbsmeta=CBSMetadata` --> passes through.

    """
```

psi4/driver/task_planner.py
loriab: nope, ordinary adjective. I'll fix the case.

psi4/driver/task_base.py
loriab: Because it's only removing drinks from the task outputs. There's still a drink when the governing psi4 session completes:

```
  //>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>//
  //      FiniteDifference Results     //
  //<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<//

  Computing second-derivative from gradients using projected,
  symmetry-adapted, cartesian coordinates.

  9 gradients passed in, including the reference geometry.
  Generating complete list of displacements from unique ones.

    Operation 3 takes plus displacements of irrep A2 to minus ones.
    Operation 2 takes plus displacements of irrep B1 to minus ones.
    Operation 2 takes plus displacements of irrep B2 to minus ones.

-------------------------------------------------------------
    SCF hessian by gradient...............................................................PASSED

    Psi4 stopped on: Tuesday, 17 May 2022 11:15AM
    Psi4 wall time for execution: 0:00:11.25

*** Psi4 exiting successfully. Buy a developer a beer!
```

loriab:
> Started on n-body, but I can't continue. I'm not familiar with what a "level" is. Do you have a reference on that? Maybe I don't know where to look, but I can't find n-body explained in the docs.

haha, the many and varied uses of unqualified "level" in nbody were just what I was vexed about a couple weeks ago. First I should say that multilevel.py isn't the place to start and also that I only touched it up around the contact points with main nbody. It needs further harmonization and probably incorporation into the class.

there's two main "levels" in nbody:
* n-body level -- so if you have a trimer, there's 1-body, 2-body, 3-body components, and you might not even bother with the highest ones (to save computational cost)
* modelchem level -- like composite applied to many-body expansion, so you do CC on low-bodies and a cheaper mtd/bas on higher-bodies

psi4/driver/driver_nbody_multilevel.py
```diff
@@ -0,0 +1,146 @@
...
+# You should have received a copy of the GNU Lesser General Public License along
+# with Psi4; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+#
+# @END LICENSE
+#
+
+import numpy as np
+
+def prepare_results(self, client=None):
```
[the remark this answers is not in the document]
loriab: whenever you see "client", it's a QCFractal matter. client enables parallel and distributed. I've fixed it up to `client: Optional["FractalClient"] = None` like others, but they're pretty much for future, not advertised, use.

`self` is a stretch here. the fn isn't in the class but when written, passing the class (and overwriting bits of it!) was the easiest way to transmit the info. class vs. free functions in all the wrappers need a re-think, but this one especially.

psi4/driver/driver_cbs.py
loriab: details that didn't make the transition. now notated as such.

psi4/driver/driver_nbody_multilevel.py
loriab: I added that it's an index label. I think this can be handled cleaner upon revision.

psi4/driver/driver_cbs.py
loriab: added "avoid having post-scf single-method gradients/Hessians show up as "(mixed)" module just because an outright HF call in the jobs list"

psi4/driver/driver_cbs.py
loriab: the commented bit on 1306. I've moved the todo aside with explanation. same in next loop

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+task_planner()
+--------------
+* computer gets modified from task_planner outside this file!
+* user levels kwarg processed into modelchem treatment levels and the n-body levels they cover. fields nbodies_per_mc_level set and max_nbody reset
+* for each modelchem treatment level, call build_tasks() below via one of four routes, depending on simple MB or layered MB(FD), MB(CBS), or MB(FD(CBS))
+
+    ------------------------------
+    ManyBodyComputer.build_tasks()
+    ------------------------------
+    * if supersystem requested, request (frag, bas) indices for full nbody range of nocp treatment from build_nbody_compute_list()
```
JonathonMisiewicz, line 53: I had to look up what "supersystem" and "nocp" meant. Also, what are frag and bas indices? I'm guessing that you index the fragments, but what is this about bases?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+* BaseComputer.__init__()
+
+task_planner()
+--------------
+* computer gets modified from task_planner outside this file!
+* user levels kwarg processed into modelchem treatment levels and the n-body levels they cover. fields nbodies_per_mc_level set and max_nbody reset
+* for each modelchem treatment level, call build_tasks() below via one of four routes, depending on simple MB or layered MB(FD), MB(CBS), or MB(FD(CBS))
+
+    ------------------------------
+    ManyBodyComputer.build_tasks()
```
JonathonMisiewicz, line 51: If this is build _tasks_, why does it say a supersystem _or_ otherwise? Some tasks are supersystem, some are otherwise. Have I missed something?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+    ------------------------------
+    * if supersystem requested, request (frag, bas) indices for full nbody range of nocp treatment from build_nbody_compute_list()
+    * otherwise, request (frag, bas) indices for specified nbody range covering specified bsse treatments from build_nbody_compute_list()
+
+        build_nbody_compute_list()
+        --------------------------
+        * initializes dicts for each of nocp, cp, vmfc (2 for this one) with keys requested n-body levels and values empty sets
+        * use combinatorics formulas to fill each key with (frag, bas) indices (what fragments are active and what fragments have basis functions)
```
JonathonMisiewicz, line 59: Are the bas indices boolean?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+
+    ------------------------------
+    ManyBodyComputer.build_tasks()
+    ------------------------------
+    * if supersystem requested, request (frag, bas) indices for full nbody range of nocp treatment from build_nbody_compute_list()
+    * otherwise, request (frag, bas) indices for specified nbody range covering specified bsse treatments from build_nbody_compute_list()
+
+        build_nbody_compute_list()
+        --------------------------
+        * initializes dicts for each of nocp, cp, vmfc (2 for this one) with keys requested n-body levels and values empty sets
```
JonathonMisiewicz, line 58: What if a user only wants a subset of these?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+        * use combinatorics formulas to fill each key with (frag, bas) indices (what fragments are active and what fragments have basis functions)
+          needed to compute the requested bsse treatments at the requested n-body levels.
+        * merge by n-body level the sets of indices for each bsse treatment into an "all" dict. return this and all the per-bsse dicts.
+
+    * merge all the requested indices, prepend a modelchem treatment index to form `mc_(frag, bas)`
```
JonathonMisiewicz, line 63: What does "merge indices" mean?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+            * re-call build_nbody_compute_list to get the cp/nocp/vmfc lists again
+
+                build_nbody_compute_list()
+                --------------------------
+
+            * slice up the supersystem mol into fragment atom ranges to help g/h arrays build piecemeal
+            * prepare empty {bsse_type}_by_level and {bsse_type}_body_dict structs. the former have different contents for vmfc
+            * for cp and nocp, resort the build_nbody_compute_list returns into per-body lists suitable for summing
+            * note that nb loops often run over more than active nbodies_per_mc_level item due to 1-body for subtraction and multilevel complications
+            * for each possibly active n-body level and each active bsse_type, call _sum_cluster_ptype_data to build by_level structs
```
JonathonMisiewicz, line 104: What is a by_level struct?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+
+                _sum_cluster_ptype_data()
+                -------------------------
+                * build up ene, grad, Hess in per-fragment pieces based on list of (frag, bas) subjobs active for that bsse treatment
```
JonathonMisiewicz, line 108: "build up pieces" is vague.

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+
+            * compute special case of monomers in monomer basis
+            * for each of cp/nocp/vmfc, apply appropriate formula to build each n-body level of cummulative total energy into body_dict
```
JonathonMisiewicz, line 111:
```suggestion
            * for each of cp/nocp/vmfc, apply appropriate formula to build each n-body level of cumulative total energy into body_dict
```

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+            * for driver=energy, set several qcvars and call:
+
+                _print_nbody_energy()
+                ---------------------
+                * prints and logs formatted energy output. called separately for cp, nocp, vmfc
+
+            * collect qcvars and summed levels into a return dictionary with some extra aliases for target bsse_type and target driver
+
+        * merge all the assemble_nbody_components return dictionaries
+        * in struct["intermediates"], store dict of `"N-BODY (?)@(?) TOTAL ENERGY" = return_energy` for all in task_list or results kwarg
```
JonathonMisiewicz, line 121: What are the question marks here?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+        * in struct["intermediates_{ptype}"], store dict of `task_list key = return_{ptype}` for all in task_list or results kwarg
```
JonathonMisiewicz, line 122: What kind of object is `return_{ptype}`? Are these E/G/H?

JonathonMisiewicz (commented):
nbody comments on overview. This was... difficult reading.

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1290 @@
...
+    Computer.get_results()
+    ----------------------
+
+        Computer._prepare_results()
+        ---------------------------
+        * if multiple modelchems (multilevel):
+
+            multilevel.prepare_results()
+            ----------------------------
+            * separates out each modelchem calcs and calls _prepare_results on each
```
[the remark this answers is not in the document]
loriab: after all the tasks are run, you've got oodles of jobs with indices like `1_((1, 2), (1, 2, 3))` and `2_((3,), (2, 3))` in form mc_(frag, bas) where mc is a modelchem index, frag are the indices of real fragments, and set(bas - frag) are the indices of ghost fragments. this line is describing a step in a multilevel (modelchem level) calc where it groups all the 1_ jobs together, all the 2_ ones together, etc., then calls the ordinary (non-multilevel) _prepare_results on each.

psi4/driver/driver_nbody.py
loriab: I'm not quite following, but I can think of two potential points for confusion:
* (1a) there's "supersystem", the whole input molecule ignoring fragmentation
* (1b) there's "supersystem", the user kwarg that signifies treating all-the-rest n-body levels with a certain modelchem
* (2a) `build_tasks` is called in a loop, so each call is dumping more tasks into the task set

psi4/driver/driver_nbody.py
loriab: added definitions like above

psi4/driver/driver_nbody.py
loriab: no, they're user-friendly 1-indexed ints, so dimer cp is `((1, 2), (1, 2))` minus `((1,), (1, 2))` minus `((2,), (1, 2))`.

psi4/driver/driver_nbody.py
loriab: you have lists of needed (frag, bas) for all the requested bsse_types, so form set union

psi4/driver/driver_nbody.py
loriab: yes, "ptype" was predecessor to "driver" and is still used for e/g/h in more local circumstances (e.g., driver=gradient often deals with ptype=energy and ptype=gradient)

psi4/driver/driver_nbody.py
loriab: placeholders. anything return-like is better described in the nbody_return table in the prepare_results() docstring

psi4/driver/driver_nbody.py
loriab: I have learned a spelling word for the day

psi4/driver/driver_nbody.py
loriab: if user doesn't request that bsse treatment, the dict remains empty. I think the next "use combinatorics" bullet covers this.

psi4/driver/driver_nbody.py
loriab: I didn't name them, and the contents vary a bit between vmfc and other and return_total_data=T/F, so it's probably better they remain vague. think of them as intermediates between dozens of individual job results and the actual nbody energies (cumm. IE, cumm. total E, contrib to E).

tests/cbs-xtpl-energy-conv/input.dat
```diff
@@ -34,7 +34,7 @@ compare_values(nucenergy_ref, h2o.nuclear_repulsion_energy(), 9, "Nuclear repuls
 
 scf_dz = energy('SCF/cc-pVDZ')
 compare_values(scf_dz_ref, scf_dz, 6, "[1] SCF/DZ Energy Check")  #TEST
-#clean()
+clean()
```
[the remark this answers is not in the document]
loriab: yes, that's a manifestation of the below. there might have been an issue on it, too.

Pre-ddd, CBS wrapper issues a clean() btwn calcs. Plain string modelchem calcs like `energy('hf/cc-pvdz')` were getting caught and also being cleaned, meaning their behavior was slightly different than `set basis cc-pvdz \n energy('hf')`. With ddd, this is no longer happening, so for occasional string modelchem calcs, you may need to add a clean in the input. A few test cases now sport extra cleans.

psi4/driver/p4util/python_helpers.py
```diff
@@ -118,11 +119,11 @@ def _resolve_target(key, target):
 
 
 @staticmethod
-def _core_wavefunction_build(mol, basis=None):
+def _core_wavefunction_build(mol, basis=None, *, quiet: bool = False):
```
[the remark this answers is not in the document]
loriab: exactly, you can use `/` and `*` to demand positional-only and keyword-only access. It makes fns more extensible to new arguments in case some user has been doing all positional lists and looking like a dgemm call.

tests/cbs-xtpl-energy-conv/input.dat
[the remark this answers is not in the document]
zachglick: Okay, thanks. What "different behavior" is observed when the `clean()` isn't present in the input file? I don't normally `clean()` between calculations, and I expect most users don't either. If that is no longer the case, I think it would impact a lot of people.

psi4/driver/p4util/python_helpers.py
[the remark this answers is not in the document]
zachglick: Neat, thanks!

tests/cbs-xtpl-energy-conv/input.dat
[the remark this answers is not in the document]
loriab: good point -- that bullet is vague. I've added an illustrative table to the PR frontmatter. fortunately the most common case will show up as PSIO error, a situation where one reflexively cleans. Only `mtd/bas` should need a change.

psi4/driver/driver_nbody_multilevel.py
```diff
@@ -0,0 +1,148 @@
...
+        if ptype == 'hessian':
+            hessian_result -= hessian1
+
+    energy_body_dict = {str(k) + b: v for b in energy_body_dict for k, v in energy_body_dict[b].items()}
+
+    nbody_results = {
+        "ret_energy": energy_result,
+        "ret_ptype": locals()[ptype + '_result'],
+        "energy_body_dict": energy_body_dict,
+    }
```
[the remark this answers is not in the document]
loriab: I definitely agree on returning more of this info from multilevel in a subsequent pass. I think multilevel needs better incorporation with the class, so `self` isn't getting passed and mutated so freely.

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1291 @@
...
+                                       ptype_size = (1,)/(nat, 3)/(3 * nat, 3 * nat)
+                                        e/g/h := energy or gradient or Hessian
+                                        rtd := return_total_data
+
+        .. |em| unicode:: U+02003 .. em space
+
+        .. _`table:nbody_return`:
+
+        +---------------------------------------------------------------+----------------------+--------------------------------------------------------------------+--------------------------------------------------------------------------------------------------------------------+
+        | item                                                          | size                 | present / zeroed                                                   | contents / interpretation                                                                                          |
```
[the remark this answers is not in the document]
loriab: thanks. I needed it myself to find out what this mysterious `1` qcvar was holding. :-)

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1291 @@
...
+"""Plan, run, and assemble QC tasks to obtain many-body expansion and basis-set superposition error treatments.
+
+=============
+ManyBody Flow
+=============
+Bullet points are major actions
+Lines of dashes denote function calls
+e/d/dd=dg/g/h := energy, dipole, dipole derivative = dipole gradient, gradient, Hessian
+`mc_(frag, bas)` := a modelchem index, mc; indices of real fragments, frag; set(bas - frag) are indices of ghost fragments. see "intermediates_energy" in big table below for example.
+
```
JonathonMisiewicz, line 37: The concept of multichem vs. n-body levels should be explained here. Include an example of how they differ.

psi4/driver/driver_nbody.py
JonathonMisiewicz: I see! Better to say this more explicitly.

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1291 @@
...
+task_planner.py::task_planner()
+-------------------------------
+* computer gets modified from task_planner outside this file!
+* modelchem (method and basis) treatment levels for each n-body level determined from user levels kwarg. fields nbodies_per_mc_level set and max_nbody reset
+* for each modelchem treatment level, call build_tasks() below via one of four routes, depending on simple MB or layered MB(FD), MB(CBS), or MB(FD(CBS))
+
+    ------------------------------
+    ManyBodyComputer.build_tasks()
+    ------------------------------
+    * if supersystem requested, request (frag, bas) indices for full nbody range of nocp treatment from build_nbody_compute_list()
```
JonathonMisiewicz, line 54: Is it possible for `supersystem` to be _part_ of the mc level, but not _all_ of it? If so, then it's ambiguous whether this means the mc level _includes_ or _is_ the supersystem.

psi4/driver/driver_nbody.py
JonathonMisiewicz: Does this mean "compute ene/grad/Hess" for each fragment using the given n-body formula?

psi4/driver/driver_nbody.py
```diff
@@ -25,830 +25,1291 @@
...
+        Hint to shape of array data to sum.
+    ptype_dict
+        Dictionary containing computed energy, gradient, or Hessian obtained from each subsystem computation
+    compute_list
+        A list of (frag, bas) tuples notating all the required computations.
+    fragment_slice_dict
+        Dictionary containing slices that index the gradient or Hessian matrix for each of the 1-indexed fragments.
+        For He--HOOH--Me cluster, `{1: slice(0, 1, None), 2: slice(1, 5, None), 3: slice(5, 10, None)}`.
+    fragment_size_dict
+        Dictionary containing the number of atoms of each 1-indexed fragment.
```
JonathonMisiewicz, line 202: What does "1-indexed" mean?

psi4/driver/driver_nbody_multilevel.py
```diff
@@ -0,0 +1,148 @@
...
+# @END LICENSE
+#
+
+from typing import Optional
+
+import numpy as np
+
+def prepare_results(self, client: Optional["FractalClient"] = None):
+    """
+    Use different levels of theory for different expansion levels
```
JonathonMisiewicz, line 35: I think this docstring needs an overhaul. Return type isn't Pydantic, and the docstring is... vague.

psi4/driver/driver_nbody.py
loriab: as opposed to 0-indexed. so a trimer has frags 1, 2, 3, just as one would use in extract_subsets.

psi4/driver/driver_nbody.py
loriab: no, supersystem here is the keyword sense, so it just means any n-body higher than explicitly mentioned.

psi4/driver/driver_nbody.py
loriab: can't really, as `build_tasks` is a general fn inherited from BaseComputer

psi4/driver/driver_nbody.py
JonathonMisiewicz: Can you change the description of `nbodies_per_mc_level` to elaborate on the significance of this option?

JonathonMisiewicz (approved):

jeffschriber (approved):

psi4/driver/driver_cbs.py
```diff
@@ -1307,46 +1092,13 @@ def _contract_scheme_orders(needdict, datakey='f_energy'):
...
-        cbs_method_kwargs = ['scf_wfn', 'corl_wfn', 'delta_wfn']
-        cbs_method_kwargs += [f'delta{x}_wfn' for x in range(2, 6)]
-        for method in cbs_method_kwargs:
-            if method in kwargs:
-                cbs_methods.append(kwargs[method])
-    return cbs_methods
-
-
-def _parse_cbs_gufunc_string(method_name):
+def _parse_cbs_gufunc_string(method_name: str):
```
PeterKraus, line 1095: I'd argue this function and `cbs_text_parser` might belong to `task_base.py`, as that's where processing of the method name is done. Not insisting, just food for thought.

psi4/driver/driver_nbody_multilevel.py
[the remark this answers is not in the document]
zachglick: +1 for more information returned in the `energy_body_dict`. It'd also be nice to have HF/post-HF breakdowns of each energy term in this dictionary. I may run a CCSD(T) trimer and want to know how much of the two and three body energies are HF, MP2, CCSD, CCSD(T), etc.

psi4/driver/driver_cbs.py
```diff
@@ -1497,11 +1226,574 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
+    trove: List[Dict[str, Any]] = []
+
+    # Minimal (enlightened) list of jobs to run to satisfy full CBS. Keys are _f_fields. Formerly JOBS.
+    compute_list: List[Dict[str, Any]] = []
+
+    # One-to-One list of AtomicComputer-s corresponding to `compute_list`.
+    task_list: List[AtomicComputer] = []
+
+    # One-to-One list of QCSchema corresponding to `task_list`.
+    results_list: List[Any] = []
```
PeterKraus, line 1528: The issue of the default `[]` and `{}` being a class variable as opposed to an instance variable applies here too. Should be `Field(default_factory=list/dict)`

psi4/driver/driver_cbs.py
```diff
@@ -1497,11 +1226,574 @@ def _cbs_gufunc(func, total_method_name, **kwargs):
...
+                t.compute(client=client)
+
+    def _prepare_results(self, client: Optional["FractalClient"] = None):
+        results_list = [x.get_results(client=client) for x in self.task_list]
+
+        modules = [getattr(v.provenance, "module", None) for v in results_list]
+        if self.driver != "energy" and len(set(modules)) == 2 and modules.count("scf") == len(modules) / 2:
+            # signature of "MP2 GRAD" - "HF GRAD" implementation detail
+            # * avoid having post-scf single-method gradients/Hessians show up as "(mixed)" module just because an outright HF call in the jobs list
+            modules = set(modules) - set(["scf"])
```
PeterKraus, line 1613: Why not just `{"scf"}`?

psi4/driver/driver_cbs.py
```diff
@@ -429,15 +554,15 @@ def _get_default_xtpl(nbasis: int, xtpl_type: str) -> Callable:
         raise ValidationError(f"Stage treatment must be 'corl' or 'scf', not '{xtpl_type}'")
 
 
-def _validate_cbs_inputs(cbs_metadata, molecule):
+def _validate_cbs_inputs(cbs_metadata: CBSMetadata, molecule: Union["qcdb.Molecule", core.Molecule]) -> CBSMetadata:
```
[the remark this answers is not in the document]
loriab: It's a string to avoid importing. It might be that way to avoid circular imports. I have a shelved pass to clean up imports and `import *` in the driver, and I hope to clean this up then. afaik, mypy and sphinx treat the string like the class.

psi4/driver/task_base.py
```diff
@@ -0,0 +1,204 @@
...
+
+            return
+
+        logger.info(f'<<< JSON launch ... {self.molecule.schoenflies_symbol()} {self.molecule.nuclear_repulsion_energy()}')
+        gof = core.get_output_file()
+
+        # EITHER ...
+        # from psi4.driver import schema_wrapper
+        # self.result = schema_wrapper.run_qcschema(self.plan())
+        # ... OR ...
```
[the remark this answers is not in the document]
loriab: I think it's handy for a very small audience (maybe only me) :-)

psi4/driver/driver.py
```diff
@@ -46,26 +48,27 @@
...
 from psi4.driver import driver_nbody
 from psi4.driver import driver_findif
+from psi4.driver import task_planner
 from psi4.driver import p4util
 from psi4.driver import qcdb
 from psi4.driver import pp, nppp, nppp10
-from psi4.driver import qmmm
-from psi4.driver.procrouting import *
 from psi4.driver.p4util.exceptions import *
+from psi4.driver.procrouting import *
```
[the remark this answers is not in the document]
loriab: Yes, I have  a pass planned. It's alarming that we've got things like `psi4.driver.re` (regular expressions module) in the psi api.

psi4/driver/driver_cbs.py
loriab: Will do, when I get to `pydantic.Field`ifying this class.

Issue #2487, opened 2022-03-17
Fix `detci` `print_lvl`


JonathonMisiewicz:
`detci` has a variable called `print_lvl_` that is currently set at 0 and remains at 0, no matter what the user sets the `print` keyword to. As a result, it's not possible [for a user to see the S^2 value if they want](http://forum.psicode.org/t/computing-s-2-on-an-fci-calculation/2439/2).

A developer should have `print_lvl_` take the value of the standard `print` keyword.

PR #2589, opened 2022-05-26
Fix `detci` printing

philipmnel:
## Description
Fixes issues #2487 where the `print` keyword was not being properly set in part of `detci` module. S^2 value now prints out if `calc_s_squared` is turned on.

## Checklist
- [x] CI tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/detci/params.cc
+++ b/psi4/src/psi4/detci/params.cc
@@ -74,7 +74,7 @@ void CIWavefunction::get_parameters(Options &options) {
     Parameters_->wfn = options.get_str("WFN");
 
     // CDS-TODO: Check these
-    print_ = 1;
+    Parameters_->print_ = 1;
 
     Parameters_->ex_lvl = options.get_int("EX_LEVEL");
     Parameters_->cc_ex_lvl = options.get_int("CC_EX_LEVEL");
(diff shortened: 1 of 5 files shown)


psi4/src/psi4/detci/params.cc
```diff
@@ -74,7 +74,7 @@ void CIWavefunction::get_parameters(Options &options) {
     Parameters_->wfn = options.get_str("WFN");
 
     // CDS-TODO: Check these
-    print_ = 1;
+    Parameters_->print_ = 1;
```
JonathonMisiewicz, line 77: Why is this needed? I would _expect_ that `CIWavefunction` is a wavefunction subclass and thus inherits the wavefunction's `print_` field. It doesn't need additional "parameters".

JonathonMisiewicz:
Thanks for the PR! I want to make sure we're fixing #2487 the right way before merging this in.

psi4/src/psi4/detci/params.cc
philipmnel: I added `print_` to the parameters struct so that it can be passed to `CIVect`, where the un-set `print_lvl` used to live. `CIVect` gets passed the parameters struct, but not the wavefunction.

JonathonMisiewicz (approved):

loriab (approved):
thanks for the fix!

PR #2605, opened 2022-06-12
Tdscf excitations symmetry

lazaroid:
## Description
Adds contributing excitations to tdscf output for non-C1 symmetry.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/response/scf_response.py
+++ b/psi4/driver/procrouting/response/scf_response.py
@@ -486,8 +486,8 @@ def _analyze_tdscf_excitations(tdscf_results, wfn, tda, coeff_cutoff,
             nvir = (nvir, )
         for h in range(nirrep):
             h_vir = x.irrep_trans_index ^ h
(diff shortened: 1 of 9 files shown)


lazaroid:
Gah! I didn't know about the pytest test cases (the standard tdscf-n set ran OK, somewhat surprisingly). A standard copy-paste and then missing one part that needed changing! I've corrected it.

My version of pytest is too old to run the tests (apparently) but extracting the various bits out of the pytest file and running that as a standard input file worked OK.

lazaroid:
I think that should have done the trick. (I ran the tdscf tests and then overwrote the respective output.ref files with the generated output.dat files.) I think the only differences are numerical errors during SCF and then formatting of the output. All final results should be unchanged.

JonathonMisiewicz (approved):
Many thanks!

loriab (approved):
lgtm, thanks!

jturney (approved):

Issue #2594, opened 2022-06-04
SCF Count Failure


JonathonMisiewicz:
When I run the following input file on [master](https://github.com/psi4/psi4/commit/9240cbeb9343f76b39e3ec34daa844ed04cd8b66):

```
memory 2 gb

molecule mol {
-1 2
C 0.00000000 0.00000000 0.00000000
C 1.21064700 0.71690500 0.00000000
C 1.20587000 2.10633100 0.00000000
C 0.00000000 2.80649500 0.00000000
C -1.20587000 2.10633100 0.00000000
C -1.21064700 0.71690500 0.00000000
H -2.14648500 0.17133400 0.00000000
H -2.14658900 2.64534700 0.00000000
H 0.00000000 3.89046700 0.00000000
H 2.14658900 2.64534700 0.00000000
H 2.14648500 0.17133400 0.00000000
C 0.00000000 -1.42321300 0.00000000
C 0.00000000 -2.63387100 0.00000000
C 0.00000000 -4.05708400 0.00000000
C -1.21064700 -4.77398900 0.00000000
C -1.20587000 -6.16341500 0.00000000
C 0.00000000 -6.86357900 0.00000000
C 1.20587000 -6.16341500 0.00000000
C 1.21064700 -4.77398900 0.00000000
H 2.14648500 -4.22841800 0.00000000
H 2.14658900 -6.70243100 0.00000000
H 0.00000000 -7.94755100 0.00000000
H -2.14658900 -6.70243100 0.00000000
H -2.14648500 -4.22841800 0.00000000

}

set scf_type df
set basis aug-cc-pVTZ
set reference uks

mol.update_geometry()
mol.symmetrize(1e-3)

energy('b3lyp')
```

I observe
```
  ==> Pre-Iterations <==

  SCF Guess: Superposition of Atomic Densities via on-the-fly atomic UHF (no occupation information).

   -------------------------
    Irrep   Nso     Nmo    
   -------------------------
     Ag       160     160 
     B1g      130     130 
     B2g       63      63 
     B3g       84      84 
     Au        63      63 
     B1u       84      84 
     B2u      160     160 
     B3u      130     130 
   -------------------------
    Total     874     874
   -------------------------

  ==> Iterations <==

                           Total Energy        Delta E     RMS |[F,P]|

   @DF-UKS iter SAD:  -538.61970612101391   -5.38620e+02   0.00000e+00 

Traceback (most recent call last):
[...]

PR #2619, opened 2022-06-21
Compute, Don't Store, DOCC/SOCC

JonathonMisiewicz:
## Description
#2476 and #2594 indicate a major problem in Psi's SCF code: the simplistic formula `nalphapi = doccpi + soccpi` and `nbetapi = doccpi` is incapable of describing references where there are more beta than alpha orbitals of a given irrep. This occurs in MOM (where we instead use ugly hacks) and in UHF (where we crash). Furthermore, storing all of these irrep quantities is redundant.

This PR remedies the situation by only storing `nalphapi` and `nbetapi`. `doccpi` and `soccpi` can be computed from these when needed. The Py-side API is unchanged `doccpi(), soccpi()`, while the C-side API changes from `doccpi_, soccpi_` to `doccpi(), soccpi()`. While we are doing some extra work to compute docc and socc every time they're needed, the computational cost is negligible in comparison to Fock diagonalizations, integral transforms, and tensor contractions. 

To prevent scope creep, this PR aims solely to change the wavefunction. **This is a major undertaking and should not be done lightly.** A subsequent PR will fix the linked issues (if not fixed by this PR) and add them as test cases once the fix is confirmed.

Obligatory @susilehtola ping.

## Todos
- [x] DOCC and SOCC are now computed rather than stored
- [x] MOM simplified
- [x] Lots of docc/socc replaced with alphapi/betapi where more appropriate
- [x] More auto
- [x] More dimension, fewer raw arrays 

## Checklist
- [x] Passes ctest and pytest (all, except addons)

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/external/downstream/v2rdm_casscf/CMakeLists.txt
+++ b/external/downstream/v2rdm_casscf/CMakeLists.txt
@@ -16,6 +16,7 @@ if(${ENABLE_v2rdm_casscf})
 
         ExternalProject_Add(v2rdm_casscf_external
             DEPENDS psi4-core
+            # Re-change source URL once V2RDM is compatible with Vector/IntVector change and docc/socc change.
             URL https://github.com/loriab/v2rdm_casscf/archive/v2rdm8.tar.gz
             UPDATE_COMMAND ""
             CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
--- a/external/downstream/v2rdm_casscf/CMakeLists.txt
+++ b/external/downstream/v2rdm_casscf/CMakeLists.txt
@@ -16,7 +16,7 @@ if(${ENABLE_v2rdm_casscf})
 
         ExternalProject_Add(v2rdm_casscf_external
             DEPENDS psi4-core
-            URL https://github.com/loriab/v2rdm_casscf/archive/929d7fc.tar.gz
+            URL https://github.com/loriab/v2rdm_casscf/archive/v2rdm8.tar.gz
             UPDATE_COMMAND ""
             CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
                        -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/psi4/src/psi4/adc/adc.cc
+++ b/psi4/src/psi4/adc/adc.cc
@@ -66,7 +66,7 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
     if (!psio_) {
         throw PSIEXCEPTION("The wavefunction passed in lacks a PSIO object, crashing ADC. See GitHub issue #1851.");
     }
-    if (nopen_) throw PSIEXCEPTION("Openshell calculation has not been implemented yet!");
+    if (nopen_) throw PSIEXCEPTION("Open-shell calculation has not been implemented yet!");
 
     aoccount = 0, boccount = 0, avircount = 0, bvircount = 0;
     for (int h = 0; h < nirrep_; h++) {
--- a/psi4/src/psi4/adc/adc.cc
+++ b/psi4/src/psi4/adc/adc.cc
@@ -47,8 +47,8 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
     std::vector<std::string> irreps_ = molecule_->irrep_labels();
     aoccpi_ = nalphapi_ - frzcpi_;
     boccpi_ = nbetapi_ - frzcpi_;
-    avirpi_ = nmopi_ - nalphapi_ - frzvpi_;
-    bvirpi_ = nmopi_ - nbetapi_ - frzvpi_;
+    avirpi_ = nmopi_ - frzvpi_ - nalphapi_;
+    bvirpi_ = nmopi_ - frzvpi_ - nbetapi_;
 
     int naocc = aoccpi_.sum();
     int nbocc = boccpi_.sum();
--- a/psi4/src/psi4/adc/adc.cc
+++ b/psi4/src/psi4/adc/adc.cc
@@ -45,31 +45,24 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
     module_ = "adc";
 
     std::vector<std::string> irreps_ = molecule_->irrep_labels();
-    aoccpi_ = new int[nirrep_];
-    boccpi_ = new int[nirrep_];
-    avirpi_ = new int[nirrep_];
-    bvirpi_ = new int[nirrep_];
+    aoccpi_ = nalphapi_ - frzcpi_;
+    boccpi_ = nbetapi_ - frzcpi_;
+    avirpi_ = nmopi_ - nalphapi_ - frzvpi_;
+    bvirpi_ = nmopi_ - nbetapi_ - frzvpi_;
 
-    int naocc = 0, nbocc = 0, navir = 0, nbvir = 0;
+    int naocc = aoccpi_.sum();
+    int nbocc = boccpi_.sum();
+    int navir = avirpi_.sum();
+    int nbvir = bvirpi_.sum();
+
     int aoccount = 0, boccount = 0, avircount = 0, bvircount = 0;
-    for (int h = 0; h < nirrep_; h++) {
-        aoccpi_[h] = doccpi_[h] + soccpi_[h] - frzcpi_[h];
-        boccpi_[h] = doccpi_[h] - frzcpi_[h];
-        avirpi_[h] = nmopi_[h] - doccpi_[h] - soccpi_[h] - frzvpi_[h];
-        bvirpi_[h] = nmopi_[h] - doccpi_[h] - frzvpi_[h];
 
-        naocc += aoccpi_[h];
-        nbocc += boccpi_[h];
-        navir += avirpi_[h];
-        nbvir += bvirpi_[h];
-    }
     aocce_ = new double[naocc];
     bocce_ = new double[nbocc];
     avire_ = new double[navir];
     bvire_ = new double[nbvir];
 
-    nopen_ = 0;
-    for (int h = 0; h < nirrep_; h++) nopen_ += soccpi_[h];
+    nopen_ = soccpi().sum();
     if (!psio_) {
         throw PSIEXCEPTION("The wavefunction passed in lacks a PSIO object, crashing ADC. See GitHub issue #1851.");
     }
@@ -77,21 +70,23 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
 
     aoccount = 0, boccount = 0, avircount = 0, bvircount = 0;
     for (int h = 0; h < nirrep_; h++) {
-        for (int a = frzcpi_[h]; a < doccpi_[h] + soccpi_[h]; a++) aocce_[aoccount++] = epsilon_a_->get(h, a);
+        for (int a = frzcpi_[h]; a < nalphapi_[h]; a++) aocce_[aoccount++] = epsilon_a_->get(h, a);
 
-        for (int b = frzcpi_[h]; b < doccpi_[h]; b++) bocce_[boccount++] = epsilon_b_->get(h, b);
+        for (int b = frzcpi_[h]; b < nbetapi_[h]; b++) bocce_[boccount++] = epsilon_b_->get(h, b);
 
-        for (int a = doccpi_[h] + soccpi_[h]; a < nmopi_[h] - frzvpi_[h]; a++)
+        for (int a = nalphapi_[h]; a < nmopi_[h] - frzvpi_[h]; a++)
             avire_[avircount++] = epsilon_a_->get(h, a);
 
-        for (int b = doccpi_[h]; b < nmopi_[h] - frzvpi_[h]; b++) bvire_[bvircount++] = epsilon_b_->get(h, b);
+        for (int b = nbetapi_[h]; b < nmopi_[h] - frzvpi_[h]; b++) bvire_[bvircount++] = epsilon_b_->get(h, b);
     }
 
+    const auto docc = doccpi();
+    const auto socc = soccpi();
     outfile->Printf("\n\n\tIrrep  Core  Docc  Socc  aOcc  aVir  bOcc  bVir  FVir\n");
     outfile->Printf("\t*****************************************************\n");
     for (int h = 0; h < nirrep_; h++) {
         outfile->Printf("\t %3s   %3d   %3d   %3d   %3d   %3d   %3d   %3d   %3d\n", irreps_[h].c_str(), frzcpi_[h],
-                        doccpi_[h], soccpi_[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
+                        docc[h] + socc[h], soccpi()[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
     }
     outfile->Printf("\t*****************************************************\n\n");
 
--- a/psi4/src/psi4/adc/adc.cc
+++ b/psi4/src/psi4/adc/adc.cc
@@ -86,7 +86,7 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
     outfile->Printf("\t*****************************************************\n");
     for (int h = 0; h < nirrep_; h++) {
         outfile->Printf("\t %3s   %3d   %3d   %3d   %3d   %3d   %3d   %3d   %3d\n", irreps_[h].c_str(), frzcpi_[h],
-                        docc[h] + socc[h], soccpi()[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
+                        docc[h], socc[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
     }
     outfile->Printf("\t*****************************************************\n\n");
 
--- a/psi4/src/psi4/libmints/wavefunction.cc
+++ b/psi4/src/psi4/libmints/wavefunction.cc
@@ -744,7 +744,7 @@ const Dimension Wavefunction::doccpi(bool warn_on_beta_socc) const {
     for (int h = 0; h < nalphapi_.n(); h++) {
         docc_vec.push_back(std::min(nalphapi_[h], nbetapi_[h]));
         if (warn_on_beta_socc && nbetapi_[h] > nalphapi_[h]) {
-            outfile->Printf("Warning! Irrep has more beta than alpha electrons in symmetry %d orbitals.\n", d);
+            outfile->Printf("Warning! Irrep has more beta than alpha electrons in symmetry %d orbitals.\n", h);
         }
     }
     return docc_vec;
--- a/psi4/src/psi4/libmints/wavefunction.cc
+++ b/psi4/src/psi4/libmints/wavefunction.cc
@@ -145,8 +145,6 @@ Wavefunction::Wavefunction(std::shared_ptr<Molecule> molecule, std::shared_ptr<B
     frequencies_ = vectors["frequencies"];
 
     // set dimensions
-    doccpi_ = dimensions["doccpi"];
-    soccpi_ = dimensions["soccpi"];
     frzcpi_ = dimensions["frzcpi"];
     frzvpi_ = dimensions["frzvpi"];
     nalphapi_ = dimensions["nalphapi"];
@@ -213,8 +211,6 @@ void Wavefunction::shallow_copy(const Wavefunction *other) {
     energy_ = other->energy_;
     efzc_ = other->efzc_;
 
-    doccpi_ = other->doccpi_;
-    soccpi_ = other->soccpi_;
     frzcpi_ = other->frzcpi_;
     frzvpi_ = other->frzvpi_;
     nalphapi_ = other->nalphapi_;
@@ -302,8 +298,6 @@ void Wavefunction::deep_copy(const Wavefunction *other) {
     efzc_ = other->efzc_;
     variables_ = other->variables_;
 
-    doccpi_ = other->doccpi_;
-    soccpi_ = other->soccpi_;
     frzcpi_ = other->frzcpi_;
     frzvpi_ = other->frzvpi_;
     nalphapi_ = other->nalphapi_;
@@ -390,10 +384,6 @@ std::shared_ptr<Wavefunction> Wavefunction::c1_deep_copy(std::shared_ptr<BasisSe
     wfn->efzc_ = efzc_;
 
     // collapse all the Dimension objects down to one element
-    wfn->doccpi_.init(1, doccpi_.name());
-    wfn->doccpi_[0] = doccpi_.sum();
-    wfn->soccpi_.init(1, soccpi_.name());
-    wfn->soccpi_[0] = soccpi_.sum();
     wfn->frzcpi_.init(1, frzcpi_.name());
     wfn->frzcpi_[0] = frzcpi_.sum();
     wfn->frzvpi_.init(1, frzvpi_.name());
@@ -501,8 +491,6 @@ void Wavefunction::common_init() {
     nmopi_ = Dimension(nirrep_, "MOs per irrep");
     nalphapi_ = Dimension(nirrep_, "Alpha electrons per irrep");
     nbetapi_ = Dimension(nirrep_, "Beta electrons per irrep");
-    doccpi_ = Dimension(nirrep_, "Doubly occupied orbitals per irrep");
-    soccpi_ = Dimension(nirrep_, "Singly occupied orbitals per irrep");
     frzcpi_ = Dimension(nirrep_, "Frozen core orbitals per irrep");
     frzvpi_ = Dimension(nirrep_, "Frozen virtual orbitals per irrep");
 
@@ -513,8 +501,6 @@ void Wavefunction::common_init() {
     nmo_ = basisset_->nbf();
     for (int k = 0; k < nirrep_; k++) {
         nmopi_[k] = 0;
-        doccpi_[k] = 0;
-        soccpi_[k] = 0;
         nalphapi_[k] = 0;
         nbetapi_[k] = 0;
     }
@@ -784,18 +770,18 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
 }
 
 void Wavefunction::force_doccpi(const Dimension &doccpi) {
+    auto socc = soccpi();
     for (int h = 0; h < nirrep_; h++) {
-        if ((soccpi_[h] + doccpi[h]) > nmopi_[h]) {
+        if ((socc[h] + doccpi[h]) > nmopi_[h]) {
             throw PSIEXCEPTION(
                 "Wavefunction::force_doccpi: Number of doubly and singly occupied orbitals in an irrep cannot exceed "
                 "the total number of molecular orbitals.");
         }
-        doccpi_[h] = doccpi[h];
-        nalphapi_[h] = doccpi_[h] + soccpi_[h];
-        nbetapi_[h] = doccpi_[h];
+        nalphapi_[h] = doccpi[h] + socc[h];
+        nbetapi_[h] = doccpi[h];
     }
-    nalpha_ = doccpi_.sum() + soccpi_.sum();
-    nbeta_ = doccpi_.sum();
+    nalpha_ = nalphapi_.sum();
+    nbeta_ = nbetapi_.sum();
 }
 
 void Wavefunction::force_soccpi(const Dimension &soccpi) {
@@ -804,16 +790,16 @@ void Wavefunction::force_soccpi(const Dimension &soccpi) {
             "Wavefunction::force_soccpi: Cannot set soccpi since alpha and beta densities must be the same for this "
             "Wavefunction.");
     }
+    auto docc = doccpi();
     for (int h = 0; h < nirrep_; h++) {
-        if ((soccpi[h] + doccpi_[h]) > nmopi_[h]) {
+        if ((soccpi[h] + doccpi()[h]) > nmopi_[h]) {
             throw PSIEXCEPTION(
                 "Wavefunction::force_soccpi: Number of doubly and singly occupied orbitals in an irrep cannot exceed "
                 "the total number of molecular orbitals.");
         }
-        soccpi_[h] = soccpi[h];
-        nalphapi_[h] = doccpi_[h] + soccpi_[h];
+        nalphapi_[h] = docc[h] + soccpi[h];
     }
-    nalpha_ = doccpi_.sum() + soccpi_.sum();
+    nalpha_ = nalphapi_.sum();
 }
 
 void Wavefunction::set_frzvpi(const Dimension &frzvpi) {
--- a/psi4/src/psi4/libmints/wavefunction.cc
+++ b/psi4/src/psi4/libmints/wavefunction.cc
@@ -739,22 +739,22 @@ void Wavefunction::initialize_singletons() {
     done = true;
 }
 
-const Dimension Wavefunction::doccpi(bool assume_socc_alpha) const {
+const Dimension Wavefunction::doccpi(bool warn_on_beta_socc) const {
     std::vector<int> docc_vec;
     for (int h = 0; h < nalphapi_.n(); h++) {
         docc_vec.push_back(std::min(nalphapi_[h], nbetapi_[h]));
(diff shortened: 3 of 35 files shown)


JonathonMisiewicz:
Eco failure is, quite predictably, because it breaks `v2rdm`.

JonathonMisiewicz:
After discussion with Lori, work on getting this PR through will resume after MQM. We also have a plan to deal with plugin incompatibilities.

Feel free to review earlier if you wish.

loriab (approved):
quite an undertaking, lgtm!

jturney (approved):

psi4/src/psi4/adc/adc.cc
```diff
@@ -45,53 +45,48 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
...
 
     std::vector<std::string> irreps_ = molecule_->irrep_labels();
-    aoccpi_ = new int[nirrep_];
-    boccpi_ = new int[nirrep_];
-    avirpi_ = new int[nirrep_];
-    bvirpi_ = new int[nirrep_];
+    aoccpi_ = nalphapi_ - frzcpi_;
+    boccpi_ = nbetapi_ - frzcpi_;
+    avirpi_ = nmopi_ - nalphapi_ - frzvpi_;
+    bvirpi_ = nmopi_ - nbetapi_ - frzvpi_;
```
susilehtola, line 51: Aesthetic change
```suggestion
    avirpi_ = nmopi_ - frzvpi_ - nalphapi_;
    bvirpi_ = nmopi_ - frzvpi_ - nbetapi_;
```

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -378,10 +374,22 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction> {
...
+        return docc_vec;
+    }
+    /// Returns the SOCC per irrep array. Not recommended for unrestricted code.
+    const Dimension soccpi() const {
+        std::vector<int> socc_vec;
+        for (int h = 0; h < nalphapi_.n(); h++) {
+            socc_vec.push_back(std::abs(nalphapi_[h] - nbetapi_[h]));
+        }
+        return socc_vec;
+    }
```
susilehtola, line 392: Should this code issue a warning when problematic cases are encountered? In that case, results of any code using socc and docc are incorrect...

JonathonMisiewicz: Problematic meaning `nbeta > nalpha` for at least one irrep, or something else?

psi4/src/psi4/adc/adc.cc
JonathonMisiewicz: Because there's something wrong with the test suite, I'll make the change only if I need to issue a new commit for some other reason. I agree aesthetically, but it's not worth a delay to the PR getting in. The buggy ADC code is not long for this world anyways. 🙂

psi4/src/psi4/libmints/wavefunction.h
susilehtola: > Problematic meaning `nbeta > nalpha` for at least one irrep, or something else?

Exactly

JonathonMisiewicz: @loriab Sound like a good candidate for the log file?

loriab: It'll be the first warning for the log file, but sounds appropriate. Could also repeat in out file if you want.

JonathonMisiewicz:
@susilehtola There was a request among core devs for you to sign off on this PR before we get a third review.

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -753,6 +739,28 @@ void Wavefunction::initialize_singletons() {
...
+    return docc_vec;
+}
+/// Returns the SOCC per irrep array. Not recommended for unrestricted code.
+const Dimension Wavefunction::soccpi(bool assume_socc_alpha) const {
+    std::vector<int> socc_vec;
+    for (int h = 0; h < nalphapi_.n(); h++) {
+        socc_vec.push_back(std::abs(nalphapi_[h] - nbetapi_[h]));
+        if (assume_socc_alpha && nbetapi_[h] > nalphapi_[h]) {
+            outfile->Printf("Warning! Irrep has more beta than alpha electrons.\n");
+        }
```
susilehtola, line 759: `assume_socc_alpha` is not properly handled here

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -804,16 +812,16 @@ void Wavefunction::force_soccpi(const Dimension &soccpi) {
...
     for (int h = 0; h < nirrep_; h++) {
-        if ((soccpi[h] + doccpi_[h]) > nmopi_[h]) {
+        if ((soccpi[h] + doccpi()[h]) > nmopi_[h]) {
             throw PSIEXCEPTION(
                 "Wavefunction::force_soccpi: Number of doubly and singly occupied orbitals in an irrep cannot exceed "
                 "the total number of molecular orbitals.");
         }
-        soccpi_[h] = soccpi[h];
-        nalphapi_[h] = doccpi_[h] + soccpi_[h];
+        nalphapi_[h] = docc[h] + soccpi[h];
```
susilehtola, line 822: What's the idea with `docc`, `doccpi()`, and `doccpi_`? Wh yso many variables that assumedly are the same?

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: How? I don't see an issue. If you assume that all SOCC are alpha _and_ observe that there are more beta than alpha for a given irrep, you have a problem. That is exactly the problem I am trying to catch.

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: I can change l. 817 to use `docc` instead of `doccpi()`. The point is to avoid recompute.

psi4/src/psi4/libmints/wavefunction.cc
susilehtola: My issue is that it's not clear what the code should be doing for `assume_socc_alpha = false`.

JonathonMisiewicz: That we are not assuming all SOCC are alpha does not imply that we are assuming all SOCC are not alpha. If you can find a better argument name, I can change it. `warn_on_beta_socc`?

JonathonMisiewicz: @susilehtola

psi4/src/psi4/libmints/wavefunction.cc
susilehtola: What is the difference between `docc`, `doccpi()` and `doccpi_`?

susilehtola: Shouldn't `soccpi_` and `doccpi_` be removed?

external/downstream/v2rdm_casscf/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if(${ENABLE_v2rdm_casscf})
 
         ExternalProject_Add(v2rdm_casscf_external
             DEPENDS psi4-core
-            URL https://github.com/loriab/v2rdm_casscf/archive/929d7fc.tar.gz
+            URL https://github.com/loriab/v2rdm_casscf/archive/v2rdm8.tar.gz
```
susilehtola, line 19: What's the reason for this change? It does not appear to be related to docc/socc.

Put it in a separate pull.

psi4/src/psi4/adc/adc.cc
```diff
@@ -45,53 +45,48 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
...
     avire_ = new double[navir];
     bvire_ = new double[nbvir];
 
-    nopen_ = 0;
-    for (int h = 0; h < nirrep_; h++) nopen_ += soccpi_[h];
+    nopen_ = soccpi().sum();
     if (!psio_) {
         throw PSIEXCEPTION("The wavefunction passed in lacks a PSIO object, crashing ADC. See GitHub issue #1851.");
     }
     if (nopen_) throw PSIEXCEPTION("Openshell calculation has not been implemented yet!");
```
susilehtola, line 69:
```suggestion
     if (nopen_) throw PSIEXCEPTION("Open-shell calculation has not been implemented yet!");
```

external/downstream/v2rdm_casscf/CMakeLists.txt
JonathonMisiewicz: No, this is a change that I've discussed with Lori. This is an interim change for v2rdm tests not to fail.

Please ask for clarification before making demands.

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: I'm legitimately confused. `soccpi_` and `doccpi_` _are_ removed.

Like I said previously, the point of `docc` is to avoid recompute of `doccpi()` every iteration of the for loop.

psi4/src/psi4/adc/adc.cc
JonathonMisiewicz: Once again, I'll bother with this if I need to make other commits, but not otherwise. As ADC is not going to survive 1.7, this change isn't worth going through another test cycle.

psi4/src/psi4/libmints/wavefunction.cc
susilehtola: `bool sanity_check`?

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -784,18 +792,18 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
...
-        doccpi_[h] = doccpi[h];
-        nalphapi_[h] = doccpi_[h] + soccpi_[h];
-        nbetapi_[h] = doccpi_[h];
+        nalphapi_[h] = doccpi[h] + socc[h];
+        nbetapi_[h] = doccpi[h];
     }
-    nalpha_ = doccpi_.sum() + soccpi_.sum();
-    nbeta_ = doccpi_.sum();
+    nalpha_ = nalphapi_.sum();
+    nbeta_ = nbetapi_.sum();
```
susilehtola, line 806: Shouldn't there be a spin state check here?

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -784,18 +792,18 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
 }
 
 void Wavefunction::force_doccpi(const Dimension &doccpi) {
```
susilehtola, line 794:
```suggestion
void Wavefunction::force_doccpi(const Dimension & input_docc) {
```

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -784,18 +792,18 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
...
+    auto socc = soccpi();
     for (int h = 0; h < nirrep_; h++) {
-        if ((soccpi_[h] + doccpi[h]) > nmopi_[h]) {
+        if ((socc[h] + doccpi[h]) > nmopi_[h]) {
```
susilehtola, line 797:
```suggestion
        if ((socc[h] + input_docc[h]) > nmopi_[h]) {
```

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -784,18 +792,18 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
...
             throw PSIEXCEPTION(
                 "Wavefunction::force_doccpi: Number of doubly and singly occupied orbitals in an irrep cannot exceed "
                 "the total number of molecular orbitals.");
         }
-        doccpi_[h] = doccpi[h];
-        nalphapi_[h] = doccpi_[h] + soccpi_[h];
-        nbetapi_[h] = doccpi_[h];
+        nalphapi_[h] = doccpi[h] + socc[h];
+        nbetapi_[h] = doccpi[h];
```
susilehtola, line 803:
```suggestion
        nalphapi_[h] = input_docc[h] + socc[h];
        nbetapi_[h] = input_docc[h];
```

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -784,18 +792,18 @@ void Wavefunction::set_reference_wavefunction(const std::shared_ptr<Wavefunction
...
     }
-    nalpha_ = doccpi_.sum() + soccpi_.sum();
-    nbeta_ = doccpi_.sum();
+    nalpha_ = nalphapi_.sum();
+    nbeta_ = nbetapi_.sum();
 }
 
 void Wavefunction::force_soccpi(const Dimension &soccpi) {
```
susilehtola, line 809:
```suggestion
void Wavefunction::force_soccpi(const Dimension & input_socc) {
```

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -804,16 +812,16 @@ void Wavefunction::force_soccpi(const Dimension &soccpi) {
             "Wavefunction::force_soccpi: Cannot set soccpi since alpha and beta densities must be the same for this "
             "Wavefunction.");
     }
+    auto docc = doccpi();
     for (int h = 0; h < nirrep_; h++) {
-        if ((soccpi[h] + doccpi_[h]) > nmopi_[h]) {
+        if ((soccpi[h] + doccpi()[h]) > nmopi_[h]) {
```
susilehtola, line 817:
```suggestion
        if ((docc[h] + input_socc[h]) > nmopi_[h]) {
```

psi4/src/psi4/libmints/wavefunction.cc
susilehtola:
```suggestion
        nalphapi_[h] = docc[h] + input_socc[h];
```

psi4/src/psi4/libqt/qt.h
```diff
@@ -59,10 +59,10 @@ void normalize(double** A, int rows, int cols);
 double invert_matrix(double** a, double** y, int N, std::string out_fname);
 void solve_2x2_pep(double** H, double S, double* evals, double** evecs);
 PSI_API
-void reorder_qt(int* docc_in, int* socc_in, int* frozen_docc_in, int* frozen_uocc_in, int* order, int* orbs_per_irrep,
+void reorder_qt(const int* docc_in, const int* socc_in, int* frozen_docc_in, int* frozen_uocc_in, int* order, int* orbs_per_irrep,
                 int nirreps);
 PSI_API
-void reorder_qt_uhf(int* docc, int* socc, int* frozen_docc, int* frozen_uocc, int* order_alpha, int* order_beta,
+void reorder_qt_uhf(const int* docc, const int* socc, int* frozen_docc, int* frozen_uocc, int* order_alpha, int* order_beta,
```
susilehtola, line 65: These changes are unrelated to the present pull

psi4/src/psi4/libqt/reorder_qt.cc
```diff
@@ -193,7 +193,7 @@ PSI_API void reorder_qt(int *docc_in, int *socc_in, int *frozen_docc_in, int *fr
 **
 ** \ingroup QT
 */
-PSI_API void reorder_qt_uhf(int *docc, int *socc, int *frozen_docc, int *frozen_uocc, int *order_alpha, int *order_beta,
+PSI_API void reorder_qt_uhf(const int *docc, const int *socc, int *frozen_docc, int *frozen_uocc, int *order_alpha, int *order_beta,
```
susilehtola, line 196: Same here

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: The name `sanity_check` is only appropriate if having more beta than alpha is _always_ inappropriate when this function is called. Lori and I talked among ourselves and thought it was best not to make that assumption. If that assumption _were_ correct, we should raise an error rather than warn, anyways.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -208,15 +199,22 @@ void HF::common_init() {
         }
     }
 
+    // Check that we have enough basis functions
+    for (int h = 0; h < nirrep_; ++h) {
+        if (std::max(nalphapi_[h], nbetapi_[h]) > nsopi_[h]) {
+            throw PSIEXCEPTION("Not enough basis functions to satisfy requested occupancies");
```
susilehtola, line 205:
```
             throw PSIEXCEPTION("Not enough linearly independent basis functions to satisfy requested occupancies");
```

psi4/src/psi4/libqt/qt.h
JonathonMisiewicz: Tests failed without this change - it's related.

Test failure was that the `const` return value of `doccpi()` or `soccpi()` was trying to be passed into this function.

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -1227,18 +1194,20 @@ void HF::check_phases() {
...
-    for (int h = 0; h < nirrep_ - 1; ++h) outfile->Printf(" %4d,", doccpi_[h]);
-    outfile->Printf(" %4d ]\n", doccpi_[nirrep_ - 1]);
+    for (int h = 0; h < nirrep_ - 1; ++h) outfile->Printf(" %4d,", docc[h]);
+    outfile->Printf(" %4d ]\n", docc[nirrep_ - 1]);
     if (reference != "RHF" && reference != "RKS") {
         outfile->Printf("    SOCC [ ");
-        for (int h = 0; h < nirrep_ - 1; ++h) outfile->Printf(" %4d,", soccpi_[h]);
-        outfile->Printf(" %4d ]\n", soccpi_[nirrep_ - 1]);
+        for (int h = 0; h < nirrep_ - 1; ++h) outfile->Printf(" %4d,", socc[h]);
+        outfile->Printf(" %4d ]\n", socc[nirrep_ - 1]);
```
susilehtola, line 1210: The code is not printing out AOCC and BOCC

psi4/src/psi4/libqt/reorder_qt.cc
JonathonMisiewicz: See discussion [here](https://github.com/psi4/psi4/pull/2619#discussion_r938033910).

psi4/src/psi4/libscf_solver/hf.cc
JonathonMisiewicz: Correct. The _old behavior_ was to print out DOCC and SOCC. I didn't want to change it to print out AOCC and BOCC - too much for one PR.

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: You can add Dimension objects directly rather than doing it by irrep for further simplifications, but agreed that the function can and should be simplified now.

external/downstream/v2rdm_casscf/CMakeLists.txt
susilehtola: > No, this is a change that I've discussed with Lori. This is an interim change for v2rdm tests not to fail.
> 
> Please ask for clarification before making demands.

See above: "What's the reason for this change? It does not appear to be related to docc/socc."

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: I don't know. The `force` functions are [marked](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/wavefunction.h#L402-L416) as intended for expert use only because careless use can produce inconsistencies. Nothing in Psi uses them. I can't tell whether the potential to cause a spin state violation is a bug or a feature. Maybe a warning is in order if the number of unpaired electrons changes?

susilehtola: OK, in that case we can forget about it

external/downstream/v2rdm_casscf/CMakeLists.txt
JonathonMisiewicz: Current V2RDM uses `doccpi_` and `soccpi_`. This changes the copy of V2RDM we use for test purposes to a branch of Lori's where those have been replaced with the new function calls.

loriab: Because the wfn interface changed, the v2rdm plugin needed changes. there were some that arose from this PR and other from another one. last two commits of https://github.com/loriab/v2rdm_casscf/commits/v2rdm8 . the change above is loosening from a particular commit that was the branch head to the branch itself.

psi4/src/psi4/libmints/wavefunction.cc
susilehtola: Actually... I think the `force` functions may break with these changes, since `socc` and `docc` are recomputed. One would have to override `socc` and `docc` simultaneously lest the occupations be changed in the other code.

susilehtola: I would remove the force functions altogether, since it sounds like they aren't even used..

psi4/src/psi4/libmints/wavefunction.cc
susilehtola: I was legitimately confused because the array names were inconsistent. The point of these functions is that an input array is passed in and then combined with the computed arrays. I have proposed above to use a more consistent naming scheme: `socc` and `docc` for the precomputed arrays and `input_socc` and `input_docc` for the overriding inputs.

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: @loriab Fine with you? These functions are dangerous. Forte doesn't use them. V2RDM doesn't use them? Psi doesn't use them.

loriab: I need to look further. I think there was a Shannon Houck change that allowed resetting docc/socc for detci, and CDS agreed it was a legit use. Would have to see if this is that.

psi4/src/psi4/libscf_solver/hf.cc
susilehtola: Does the code already print out the number of alpha and beta electrons? I don't think it does.

Since the code changes the way the calculation is run, the printout should be changed as well. Just add printouts for AOCC and BOCC?

JonathonMisiewicz: AOCC and BOCC are printed in the "excited MOM" case. I'm happy to change to print alpha and beta always, print docc and socc if well-defined. Again, fine with @loriab ?

loriab: more printing? sure!

psi4/src/psi4/libmints/wavefunction.cc
loriab: aha, https://github.com/psi4/psi4/pull/876 . the ability added in that PR should be preserved. it likely enables things in the fockCI plugin

JonathonMisiewicz: I'll give that a look. Do we have FockCI test coverage?

JonathonMisiewicz: * The express purpose of this forcing is to [change spin state](https://github.com/shannonhouck/psi4fockci/blob/267e17103c77dcd42cf1a66a5b805db3bcd5ed9e/psi4fockci/spinflip.py#L131-L142). So the lack of a spin state check is definitely a feature and not a bug.
* This use cases forces both `doccpi` and `soccpi`. It makes sense to me to have the user force both simultaneously.

susilehtola: Then those functions definitely need to be merged into `force_docc_socc`; otherwise there is no guarantee that the code will actually use the input arrays.

loriab: > I'll give that a look. Do we have FockCI test coverage?

possibly not, 😳 . She was working on two plugins, old and new. It was near a psi release, and I got very confused and packaged the wrong one. She added https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/plugin_psi4fockci.rst though, so the internal build and that input might work.

loriab: a combo docc_and_socc can work: https://github.com/shannonhouck/psi4fockci/commit/9a3916598030cdd501c7370ca2040bd7659e018b#diff-e33326fe2aaf9f7d7d385252ae30fa0c5a6d47751f890592c8ddc82bd8b861e4R83

JonathonMisiewicz: > > I'll give that a look. Do we have FockCI test coverage?
> 
> possibly not, 😳 . She was working on two plugins, old and new. It was near a psi release, and I got very confused and packaged the wrong one. She added https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/plugin_psi4fockci.rst though, so the internal build and that input might work.

What does "internal build" mean?

loriab: toggle `ENABLE_psi4fockci=ON` in `<objdir>/CMakeCache.txt` and rebuild. hopefully cmake will pull a copy and install it into `stage/` so it's ready to run.

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: Renaming to `warn_on_beta_socc`, given aforementioned objections to `sanity_check`. Happy to re-discuss if there's a new alternative.

JonathonMisiewicz:
FYI, I'm in the process of revising the PR and am marking comment as resolved as I fix them locally.

psi4/src/psi4/libscf_solver/hf.cc
JonathonMisiewicz: Proposed rewording is misleading. This exception is thrown due to a comparison against `nsopi`, which is the number of basis functions _regardless of_ linear dependence checking. (Linear dependence checking hasn't even occurred at the time this method is called.)

I suspect but have not confirmed that a system with enough basis functions, but not enough linearly independent basis functions, would escape this check and crash Psi. If true, we should file an issue.

psi4/src/psi4/libscf_solver/hf.cc
JonathonMisiewicz: Changed proposal to "print alpha and beta always, print docc and socc **always**." A warning will fire if those might be misleading, and AOCC and BOCC should clear confusion.

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: @loriab It doesn't look like FockCI has any tests registered with Psi. Am I mistaken?

JonathonMisiewicz:
All comments except FockCI testing addressed.

psi4/src/psi4/libscf_solver/hf.cc
susilehtola: OK thanks for the clarification; I assumed that the number of symmetry functions was done *after* the linear dependence check.

psi4/src/psi4/libmints/wavefunction.h
```diff
@@ -375,9 +375,9 @@ class PSI_API Wavefunction : public std::enable_shared_from_this<Wavefunction> {
     static void initialize_singletons();
 
     /// Returns the DOCC per irrep array. Not recommended for unrestricted code.
-    const Dimension doccpi(bool assume_socc_alpha = true) const;
+    const Dimension doccpi(bool warn_on_beta_socc = true) const;
     /// Returns the SOCC per irrep array. Not recommended for unrestricted code.
```
susilehtola, line 379: Shouldn't the argument be documented here? e.g. `warn_on_beta_socc: trigger warning on singly occupied beta orbitals that break assumptions made in the old-fashioned DOCC/SOCC handling`

psi4/src/psi4/libmints/wavefunction.cc
loriab: you could try https://github.com/psi4/psi4/blob/master/tests/pytests/test_addons.py#L1236 . because of my repo confusion, though, it may not be what you're after.

JonathonMisiewicz:
@susilehtola Comments added.

Per discussion with @loriab , I'll leave a comment on the FockCI repo letting them know about this API-breaking change. It sounds like that repo has its own issues that need to be dealt with, beyond the scope of this PR.

psi4/src/psi4/adc/adc.cc
```diff
@@ -45,53 +45,48 @@ ADCWfn::ADCWfn(SharedWavefunction ref_wfn, Options& options) : Wavefunction(opti
...
     }
 
+    const auto docc = doccpi();
+    const auto socc = soccpi();
     outfile->Printf("\n\n\tIrrep  Core  Docc  Socc  aOcc  aVir  bOcc  bVir  FVir\n");
     outfile->Printf("\t*****************************************************\n");
     for (int h = 0; h < nirrep_; h++) {
         outfile->Printf("\t %3s   %3d   %3d   %3d   %3d   %3d   %3d   %3d   %3d\n", irreps_[h].c_str(), frzcpi_[h],
-                        doccpi_[h], soccpi_[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
+                        docc[h] + socc[h], soccpi()[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
```
susilehtola, line 89: You're supposed to print docc and socc so
```suggestion
                        docc[h], socc[h], aoccpi_[h], avirpi_[h], boccpi_[h], bvirpi_[h], frzvpi_[h]);
```

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -753,6 +739,28 @@ void Wavefunction::initialize_singletons() {
     done = true;
 }
 
+const Dimension Wavefunction::doccpi(bool warn_on_beta_socc) const {
+    std::vector<int> docc_vec;
+    for (int h = 0; h < nalphapi_.n(); h++) {
+        docc_vec.push_back(std::min(nalphapi_[h], nbetapi_[h]));
+        if (warn_on_beta_socc && nbetapi_[h] > nalphapi_[h]) {
+            outfile->Printf("Warning! Irrep has more beta than alpha electrons.\n");
```
susilehtola, line 747: .. in symmetry `h`

psi4/src/psi4/libmints/wavefunction.cc
```diff
@@ -753,6 +739,28 @@ void Wavefunction::initialize_singletons() {
...
+    }
+    return docc_vec;
+}
+/// Returns the SOCC per irrep array. Not recommended for unrestricted code.
+const Dimension Wavefunction::soccpi(bool warn_on_beta_socc) const {
+    std::vector<int> socc_vec;
+    for (int h = 0; h < nalphapi_.n(); h++) {
+        socc_vec.push_back(std::abs(nalphapi_[h] - nbetapi_[h]));
+        if (warn_on_beta_socc && nbetapi_[h] > nalphapi_[h]) {
+            outfile->Printf("Warning! Irrep has more beta than alpha electrons.\n");
```
susilehtola, line 758: same here

susilehtola:
@JonathonMisiewicz as far as I can see, this should be OK now except the one wrong printout above.

susilehtola (approved):

psi4/src/psi4/libmints/wavefunction.cc
JonathonMisiewicz: Changed to "in symmetry `h` orbitals" instead, but done in principle.

zachglick (approved):
LGTM, thanks for working on this. Any reason for not adding a test or two?

JonathonMisiewicz:
> LGTM, thanks for working on this. Any reason for not adding a test or two?

No tests came to mind as worth adding. If you have suggestions, I'll probably take them.

zachglick:
No suggestions, merge when ready.

JonathonMisiewicz:
[PRd to FockCI](https://github.com/shannonhouck/psi4fockci/pull/7), so I can wash my hands of the `force_.occpi` changes.

susilehtola:
> > LGTM, thanks for working on this. Any reason for not adding a test or two?
> 
> No tests came to mind as worth adding. If you have suggestions, I'll probably take them.

Any one of the various failing SCF calculations where occupations break old invalid assumptions?

JonathonMisiewicz:
> A subsequent PR will fix the linked issues (if not fixed by this PR) and add them as test cases once the fix is confirmed.

On my to-do list, I just wasn't confident those would already be fixed by this PR. Thanks for the reminder.

PR #2628, opened 2022-07-08
Refactor `Vector` and `IntVector`

JonathonMisiewicz:
## Description
**What?** This PR creates the `IrrepedVector` class template. The `Vector` and `IntVector` classes now inherit from this, adding on methods that only make sense for that particular template instance.

**Why?** During a refactor of the MOM code, I had to use `Vector` and `IntVector` features heavily and found this synchronization of the classes necessary to keep my sanity.

**Who cares?** Obligatory pings to @hokru (who may want to create `FloatVector` for faster `dfocc` operations) and @lothian (who may want to create `ComplexVector` for magnetic spectroscopies). No reviews required.

**How do I review this?** The changes to `export_mints.cc`, `vector.h`, `vector.cc`, `intvector.cc`, and `test_vector.py` are the heart of this PR. The rest are either compatibility changes or general code cleanup. Be warned that `iota` and the `sort` functions are needed for my upcoming MOM changes and are currently unused, and I've changed some functions to not use pointers.

## Todos
- [x] Created `IrrepedVector<T>`
- [x] `Vector` and `IntVector` inherit from `IrrepedVector<T>`
- [x] `Vector` and `IntVector` now have identical code logic
- [x] `iota` and `sort` methods are added
- [x]   `IntVector` now has more functionality
- [x] Added bounds checking to several `IrrepedVector<T>` methods.
- [x] Increased test coverage of `Vector`
- [x] Added test coverage of `IntVector`
- [x] Silenced compiler warning about missing override
- [x] Removed several pointers from parameter types
- [x] Removed several pointers from return types
- [x] Changed convention from "irreped" to "irrepped"

## Checklist
- [x] Full ctest passed (before some trivial edits)
- [x] `test_vector.py` passed 

## Status
- [x] Ready for review
- [x] Ready for merge **SQUASH**; Jet approval **required** before merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -347,9 +347,9 @@ void export_mints(py::module& m) {
         .def("begin", &Slice::begin, "Get the first element of this slice")
         .def("end", &Slice::end, "Get the past-the-end element of this slice");
 
-    py::class_<IrrepedVector<double>, std::shared_ptr<IrrepedVector<double>>>(m, "ProtoVector");
+    py::class_<IrreppedVector<double>, std::shared_ptr<IrreppedVector<double>>>(m, "ProtoVector");
 
-    py::class_<Vector, std::shared_ptr<Vector>, IrrepedVector<double>>(m, "Vector", "Class for creating and manipulating vectors",
+    py::class_<Vector, std::shared_ptr<Vector>, IrreppedVector<double>>(m, "Vector", "Class for creating and manipulating vectors",
                                                 py::dynamic_attr())
         .def(py::init<int>())
         .def(py::init<const Dimension&>())
@@ -429,13 +429,13 @@ void export_mints(py::module& m) {
             },
             py::return_value_policy::reference_internal);
 
-    py::class_<IrrepedVector<int>, std::shared_ptr<IrrepedVector<int>>>(m, "ProtoIntVector");
+    py::class_<IrreppedVector<int>, std::shared_ptr<IrreppedVector<int>>>(m, "ProtoIntVector");
 
     typedef int (IntVector::*int_vector_get_1)(int) const;
     typedef int (IntVector::*int_vector_get_2)(int, int) const;
     typedef void (IntVector::*int_vector_set_1)(int, int);
     typedef void (IntVector::*int_vector_set_2)(int, int, int);
-    py::class_<IntVector, std::shared_ptr<IntVector>, IrrepedVector<int>>(m, "IntVector", "Class handling vectors with integer values")
+    py::class_<IntVector, std::shared_ptr<IntVector>, IrreppedVector<int>>(m, "IntVector", "Class handling vectors with integer values")
         .def(py::init<int>())
         .def(py::init<const Dimension&>())
         .def(py::init<const std::string&, int>())
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -314,7 +314,6 @@ void export_mints(py::module& m) {
     typedef void (Vector::*vector_setitem_2)(int, int, double);
     typedef double (Vector::*vector_getitem_1)(int) const;
     typedef double (Vector::*vector_getitem_2)(int, int) const;
-    typedef void (Vector::*vector_one)(const Vector& other);
     typedef double (Vector::*vector_one_double)(const Vector& other);
     typedef void (Vector::*vector_two)(double scale, const Vector& other);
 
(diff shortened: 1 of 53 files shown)


psi4/src/psi4/libfock/solver.cc
```diff
@@ -253,9 +253,9 @@ void CGRSolver::guess() {
 }
 void CGRSolver::residual() {
     for (size_t N = 0; N < b_.size(); ++N) {
-        r_[N]->copy(Ap_[N].get());
+        r_[N]->copy(*Ap_[N]);
```
loriab, line 256: There's a lot of removing .get()s and adding dereferencing *, which isn't the direction of code changes I'm used to seeing. Looks like a switch from the stored (not managed) pointer to the value pointed at by the shared_ptr, without changing the fn signature. Is there a term for this procedure I can look up or a quick explanation?

loriab (commented):
nice code cleanup!

psi4/src/psi4/libfock/solver.cc
JonathonMisiewicz: The quick explanation is that the function signature actually did change.

Old code had one version of `copy` where the argument was passed by _pointer_ and one where it was passed by _reference_. `copy` doesn't need to manage memory or accept a `null` value, so the new code exclusively has the pass by _reference_.

Old code used pass-by-pointer (even though there was a perfectly good pass-by-reference), new code uses pass-by-reference.

tests/pytests/test_vector.py
```diff
@@ -51,8 +51,124 @@ def test_constructors_w_symmetry(name, dim):
...
+    vec.set(2, 2, 7)
+    assert vec.get(0) == 5
+    assert vec.get(2, 2) == 7
+
+def test_int_vs_float():
+    dim = Dimension(1)
+    with pytest.raises(TypeError):
+        IntVector.set(0, 0.1)
+    with pytest.raises(TypeError):
+        IntVector.add(0, 0.1)
```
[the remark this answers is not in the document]
JonathonMisiewicz: Yikes, definitely not. Will fix on next commit, thanks.

psi4/src/export_mints.cc
```diff
@@ -348,30 +347,35 @@ void export_mints(py::module& m) {
...
         .def(
             "clone",
             [](Vector& vec) {
-                std::shared_ptr<Vector> result = std::move(vec.clone());
-                return result;
+                return std::make_shared<Vector>(std::move(vec.clone()));
             },
             "Clone the vector")
         .def("zero", &Vector::zero, "Zeros the vector")
-        .def("print_out", &Vector::print_out, "Prints the vector to the output file")
```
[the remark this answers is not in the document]
JonathonMisiewicz: `vector.print_out()` is literally the same thing as `vector.print()` in the old code. I don't understand why we need `print_out`, since it doesn't _also_ offer the optional arguments that `print` does. Do we want `print_out` for consistency across different `libmints` objects?

tests/pytests/test_vector.py
```diff
@@ -173,3 +173,7 @@ def test_bounds(tested_class):
     with pytest.raises(RuntimeError):
         vec.set(0, 1000, 0)
 
+    vec.set(1, 1000)
+    assert 1000 == vec.get(1)
+    vec.add(1, 1000)
+    assert 2000 == vec.get(1)
```
loriab, line 179: huh, surprising indeed

loriab (approved):
needs v2rdm commit pin, then lgtm!

psi4/src/psi4/libmints/vector.h
```diff
@@ -40,312 +46,326 @@ class Matrix;
...
+    }
 
-    /// Sets a single element value
-    void set(int m, double val) { vector_[0][m] = val; }
+};
 
-    /// Returns a single element value
-    double get(int h, int m) const { return vector_[h][m]; }
+/*! \ingroup MINTS */
+class PSI_API Vector final : public IrrepedVector<double> {
```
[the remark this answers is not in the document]
JonathonMisiewicz: I'll change to `Irrepped`. Just wanted to have more commits to make in case I needed to restart build again.

jturney (approved):

JonathonMisiewicz:
Latest commit fulfills my IOU to @loriab and @susilehtola about moving V2RDM source used for build purposes to a commit. The newly linked version works with the modifications to the Vector classes created by this PR.

hokru (approved):
lgtm, though this isn't my strong suit. The syntax change might have people stumble for a bit.

psi4/src/psi4/dlpno/mp2.cc
```diff
@@ -849,23 +849,23 @@ void DLPNOMP2::pno_transform() {
 
         // Diagonalization of pair density gives PNOs (in basis of the LMO's virtual domain) and PNO occ numbers
         auto X_pno_ij = std::make_shared<Matrix>("eigenvectors", nvir_ij, nvir_ij);
-        SharedVector pno_occ = std::make_shared<Vector>("eigenvalues", nvir_ij);
+        auto pno_occ = Vector("eigenvalues", nvir_ij);
```
zachglick, line 852:
```suggestion
        Vector pno_occ("eigenvalues", nvir_ij);
```

PR #2637, opened 2022-07-20
Partial cleanup of sq_rsp.cc and rsp.cc

TiborGY:
## Description
PR 1 / ? of global diagonalizer cleanup (#2642)
This is a subset of #2642, and the rest of #2642 will be broken up into new PRs after this is merged.

`sq_rsp.cc` contained an entire commented out diagonalizer function, and some comments related to that, from 2010. (?)
Deleted.

rsp.cc still used its own implementation of a diagonalizer, now it has been rewritten to use the LAPACK routine instead.
This has allowed the deletion of tred2.cc and tqli.cc.

## Todos

- [x] Remove unused code from sq_rsp.cc
- [x] Rewrite rsp.cc

## Questions
- [ ] The **only** user of rsp(...) is fine with ascending order results: https://github.com/psi4/psi4/blob/ad2063594a88fd9f60d5a122398bf2f6ae2102a9/psi4/src/psi4/detci/diag_h.cc#L245
Based on this, I have not reimplemented the old function's support for reordering the results (`matz > 1`), so this will be a change in behaviour and rsp(...) is marked as `PSI_API`. Is this objectionable? It now fails with an error message if someone tries to ask for the eigenvalues in descending order.
- [x] It is not documented if the packed lower-triangular matrix on the input is supposed to be row-major-packed or column-major-packed. Based on the call site above, I guessed it is row-major packed. Is this correct? LAPACK expects column-major-packed, so I am transforming the array to be just so.
- [ ] I _think_ I got everything right, but I am not entirely sure if the DETCI test suite is thorough enough to find out if I did not. Codecov suggests the call site is not covered (https://app.codecov.io/gh/psi4/psi4/blob/master/psi4/src/psi4/detci/diag_h.cc), but the last coverage measurement seems to have been a year ago.

## Checklist
- [x] No new features
- [x] All or relevant fraction of full tests run, see [test.log](https://github.com/psi4/psi4/files/9162565/test.log)

## Status
- [x] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
How does this PR differ from #2642?

JonathonMisiewicz (approved):

hokru:
Somewhere in the cleanup process we should get rid of the `rsp` name from eispack. 

> It is not documented if the packed lower-triangular matrix on the input is supposed to be row-major-packed or column-major-packed. Based on the call site above, I guessed it is row-major packed. Is this correct? LAPACK expects column-major-packed, so I am transforming the array to be just so

I would assume this is the right approach.

hokru (approved):

TiborGY:
This PR has largely become obsolete, sq_rsp and rsp has been deprecated, and should end up deleted **after** 1.7 is released.

PR #2644, opened 2022-07-25
smother CI linux ecosys assembly error

loriab:
## Description

GXX 7.5 from conda is throwing bizarre assembly-level errors during libmints compile like `open CFI”` and `Error: .size expression for  does not evaluate to a constant`. This compiler is working fine locally and the conda env hasn't changed since it was working fine on GHA. Searching suggests things like running of of memory for compile or binutils patches. conda has moved from 7.5 to 11.2 pinning, so that's probably a good next step.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -127,9 +127,9 @@ jobs:
         EOF
         if [[ "${{ runner.os }}" == "Linux" ]]; then
           # <lang>-compiler is a metapackage on c-f but not on defaults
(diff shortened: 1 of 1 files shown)


zachglick (approved):

JonathonMisiewicz (approved):

JonathonMisiewicz:
For future reference, the commented out test is causing Ecosystem build fails. We don't understand why, but it's holding up the rest of the development pipeline, so we're commenting it out temporarily. This should be fixed by the next full release.

hokru (approved):

PR #2650, opened 2022-07-28
DFHelper Readability 1

JonathonMisiewicz:
## Description
This is the first PR of what will probably be a series to improve the readability of `dfhelper.cc`. Because this code is overwhelming, I'm breaking the PR into pieces. Rapid approval cycle appreciated, since there will likely be a lot of these.

## Checklist
- [x] quicktests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/lib3index/dfhelper.cc
+++ b/psi4/src/psi4/lib3index/dfhelper.cc
@@ -264,7 +264,7 @@ void DFHelper::prepare_sparsity() {
     std::vector<double> shell_max_vals(pshells_ * pshells_, 0.0);
     std::vector<double> fun_max_vals(nbf_ * nbf_, 0.0);
     schwarz_shell_mask_.resize(pshells_ * pshells_);
-    schwarz_fun_count_.resize(nbf_ * nbf_);
+    schwarz_fun_index_.resize(nbf_ * nbf_);
     symm_ignored_columns_.resize(nbf_);
     symm_big_skips_.resize(nbf_ + 1);
     symm_small_skips_.resize(nbf_);
@@ -327,22 +327,18 @@ void DFHelper::prepare_sparsity() {
     // => Prepare screening/indexing data <=
     double tolerance = cutoff_ * cutoff_ / max_val;
 
-    // WARNING: It's not clear why the below is commented out.
-    //#pragma omp parallel for simd num_threads(nthreads_) schedule(static)
     // ==> Is this shell pair significant? <==
     for (size_t i = 0; i < pshells_ * pshells_; i++) schwarz_shell_mask_[i] = (shell_max_vals[i] >= tolerance);
     
-    // WARNING: It's not clear why the below is commented out.
-    //#pragma omp parallel for private(count) num_threads(nthreads_)
     // ==> Is this basis function pair significant? Also start storing non-symmetric indexing. <==
     for (size_t i = 0, count = 0; i < nbf_; i++) {
         count = 0;
         for (size_t j = 0; j < nbf_; j++) {
             if (fun_max_vals[i * nbf_ + j] >= tolerance) {
                 count++;
-                schwarz_fun_count_[i * nbf_ + j] = count;
+                schwarz_fun_index_[i * nbf_ + j] = count;
             } else
-                schwarz_fun_count_[i * nbf_ + j] = 0;
+                schwarz_fun_index_[i * nbf_ + j] = 0;
         }
         small_skips_[i] = count;
     }
@@ -363,7 +359,7 @@ void DFHelper::prepare_sparsity() {
         size_t size = 0;
         size_t skip = 0;
         for (size_t j = 0; j < nbf_; j++) {
-            if (schwarz_fun_count_[i * nbf_ + j]) {
+            if (schwarz_fun_index_[i * nbf_ + j]) {
                 (j >= i ? size++ : skip++);
             }
         }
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -356,16 +354,12 @@ void DFHelper::prepare_sparsity() {
     }
     small_skips_[nbf_] = coltots;
 
-    // build indexing skips for sparse, symmetric pQq integrals
-    // symm_big_skips: outer indexing jumps. for the p index
-    // symm_ignored_columns: number of columns that should be ignored, for a given p index,
-    //                              due to triangular symmetry
-    // symm_small_skips: size of q for each p index, technically small_skips[p] - symm_ignored_columns[p]
+    // ==> Symmetric indexing. <==
```
loriab, line 357: `p`, `q`, `Q` are widespread in notation. really want to remove them, even with fuller docs below?

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -324,28 +324,26 @@ void DFHelper::prepare_sparsity() {
...
+    // => Prepare screening/indexing data <=
     double tolerance = cutoff_ * cutoff_ / max_val;
 
-    //#pragma omp parallel for simd num_threads(nthreads_) schedule(static)
-    for (size_t i = 0; i < pshells_ * pshells_; i++) schwarz_shell_mask_[i] = (shell_max_vals[i] < tolerance ? 0 : 1);
+    // ==> Is this shell pair significant? <==
+    for (size_t i = 0; i < pshells_ * pshells_; i++) schwarz_shell_mask_[i] = (shell_max_vals[i] >= tolerance);
 
-    //#pragma omp parallel for private(count) num_threads(nthreads_)
+    // ==> Is this basis function pair significant? Also start storing non-symmetric indexing. <==
```
loriab, line 333: why remove a starting-point for future parallelization work?

JonathonMisiewicz: I assumed the line didn't work, if it was commented out. Do you know/suspect otherwise?

psi4/src/psi4/lib3index/dfhelper.cc
JonathonMisiewicz: I can see that as a justification for using `p`, `q` and `Q` in the fuller docs in the header, but I'm less clear why that means these comments should stay. My concern is that the header should be the single source of truth for comments about what a variable means.

psi4/src/psi4/lib3index/dfhelper.cc
loriab: I expect you're right that it doesn't work for some reason. But Matt was proficient in threading, so if he added it, it might be a good starting-point.

psi4/src/psi4/lib3index/dfhelper.cc
loriab: I definitely support using pqQ in header docs. I suspect some folks won't go to the header and would instead grasp appreciatively at existing comment docs, but I appreciate the single-source argument, too.

JonathonMisiewicz: I moved the p/q information into the comments in the header, but I want to stay firm on "variable explanations are in the header."

loriab (approved):

psi4/src/psi4/lib3index/dfhelper.cc
JonathonMisiewicz: I would rather have the line removed. Do either of you care enough to make this a barrier to merging this in? I have no interest in mediating a disagreement between the two of you.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -260,22 +260,21 @@ void DFHelper::prepare_sparsity() {
     if (sparsity_prepared_) return;
     timer_on("DFH: sparsity prep");
 
-    // prep info vectors
+    // => Initialize vectors <=
     std::vector<double> shell_max_vals(pshells_ * pshells_, 0.0);
     std::vector<double> fun_max_vals(nbf_ * nbf_, 0.0);
     schwarz_shell_mask_.resize(pshells_ * pshells_);
-    schwarz_fun_mask_.resize(nbf_ * nbf_);
+    schwarz_fun_count_.resize(nbf_ * nbf_);
```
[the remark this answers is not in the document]
JonathonMisiewicz: `mask` means something else in the context of this class. `index` is fine.

psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -260,22 +260,21 @@ void DFHelper::prepare_sparsity() {
...
     symm_ignored_columns_.resize(nbf_);
     symm_big_skips_.resize(nbf_ + 1);
     symm_small_skips_.resize(nbf_);
     small_skips_.resize(nbf_ + 1);
     big_skips_.resize(nbf_ + 1);
 
-    // prepare eri buffers
+    // => Populate a vector of TwoBodyAOInt to make ERIs, one per thread. <=
```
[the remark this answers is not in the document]
JonathonMisiewicz: There is no conflict between "one per thread" and "at most two". `dfhelper` assumes (for some reason) there are at most two threads. That's why the next line you mention saves that into the `nthreads_` variable.

psi4/src/psi4/lib3index/dfhelper.cc
loriab: Zach judged the parallelism loop on its merits, which I didn't. Removal fine by me, then. :-)

psi4/src/psi4/lib3index/dfhelper.cc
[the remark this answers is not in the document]
zachglick: Yeah, what I was getting at is that `nthreads_` is the true number of threads psi4 has to work with. `nthreads` is the (perhaps confusingly named) number of threads used here to do setup/screening, which is capped at 2.

zachglick (approved):

psi4/src/psi4/lib3index/dfhelper.cc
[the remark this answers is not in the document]
JonathonMisiewicz: Ah, that makes much more sense. Renamed the variable to `screen_threads` and clarified the comments.

maxscheurer (approved):

PR #2633, opened 2022-07-19
add REMP methods (take 2)

loriab:
## Description
Not for review. This in contents is much a duplicate of #2354, except that was (18 months ago master + SB commits + UB commits + SB commits, all with some master commits sprinkled in) and this is (current master + UB commits + 1 LAB commit with occ/dfocc testing + SB commits). This will have more testing added and be broken up into further PRs. Just notification of progress, @behnle. Thanks for the very nice contribution.

## Status
- [ ] Ready for review
- [ ] Ready for merge

PR #2653, opened 2022-08-01
conventional REMP2 and OREMP2

loriab:
## Description
REMP methods added to `occ` and attendant testing and routing.

This is PR 1/? in the mega-dfocc-remp series.

#### PR Background (not of general interest)
There was https://github.com/loriab/psi4/tree/dfocc2 by @bozkaya and @yavuzalagoz that added to `dfocc` (1) non-oo uhf ccsd-level E & G, (2) oo ccd-level E & G, and (3) fno everything. Then, there was #2354 that added remp2 E & G to `occ` and `dfocc` and made extensive improvements to `dfocc` in the matter of combined DIIS (vital for convergence) and int overflows. All based on pre-DIIS overhaul #2369 . To untangle this, I started from master, cherry-picked the dfocc2 commits, cherry-picked the #2354 commits, modernized the DIIS calls, then added the neglected stdsuite testing and attendant qcvar fixups for existing occ/dfocc methods and new methods remp2 and categories (1) and (2) from dfocc2 branch. That mega-changeset lives at #2633 for now. This PR breaks off the `occ` portion for review.

## Todos
- [x] @behnle added REMP2 and OREMP2 hybrid perturbation theories (https://doi.org/10.1063/1.5086168) with conventional integrals added to `occ` module, as originally proposed in #2354
- [ ] note that this uses a specialty QCEngine for reference values. https://github.com/MolSSI/QCEngine/pull/375 It'll build by itself. I'll need to mint a QCEngine patch release before building conda packages with this PR in master.
- [x] docs-wise, I picked off a bit and included it in this PR. on the whole, let's defer docs edits until the dfocc PR
- [x] reworked the run_occ* driver functions a bit so that method defs are localized and there's less risk of missing an internal setting when adding new methods.
- [x] added stdsuite testing for new methods remp2 energy and oremp2 energy and gradient. added stdsuite testing  for existing E & G for omp2, omp2.5, omp3, oremp2, olccd and filled in some gaps for ccsd, ccsd(t), a-ccsd(t). all conventional ints. fixed up some mis-set byproducts uncovered by stdsuite testing: OO ROHF was printing and setting wrong plain MP2 energies,  OMP3 & OMP2.5 wasn't setting right `Wfn.energy_`, stop setting ROHF MP3 and MP2.5 out of caution.
- [x] tightened the min rms_mograd_convergence from 6.0 to 6.5 so that stdsuite (3 mol/basis sets) could reliably compute energies, gradients, and findif gradients to 1e-6.

## Questions
- [x] sometimes the non-OO value (e.g., MP2, LCCD) is available as an early byproduct of the OO calc (e.g., OMP2, OLCCD). this wasn't the case with REMP, and some QCVariables that stored a purported non-OO REMP had to be removed. @behnle, please confirm that this sounds right.
- @behnle, sorry for the long delay and scrambling of your PR. please feel free to comment and PR to this to modify.
- @bozkaya, this adds pretty cleanly onto your `occ` code, but do look it over if you want.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


behnle:
Excellent, thanks for  your tremendous effort.
Given that my test cases were not modified and still seem to work, i do not have any complaints.

> * sometimes the non-OO value (e.g., MP2, LCCD) is available as an early byproduct of the OO calc (e.g., OMP2, OLCCD). this wasn't the case with REMP, and some QCVariables that stored a purported non-OO REMP had to be removed.

MP2 should also be available from the guess. But it is of course better to not print/store something than to print something wrong. I might be mistaken, but given that `occ` does coupled DIIS for amplitudes and orbitals, the canonical LCCD enery should not be available from an OLCCD calculation (lccd is never iterated on canonical orbitals). The same holds for REMP2. So yes, if there are variables pretending to be canonical results from an orbital-optimized calculation, these should probably be removed.
No problem, it was mostly my fault to put way too many changes into a single pull request based on an ancient master branch.

JonathonMisiewicz:
> > * sometimes the non-OO value (e.g., MP2, LCCD) is available as an early byproduct of the OO calc (e.g., OMP2, OLCCD). this wasn't the case with REMP, and some QCVariables that stored a purported non-OO REMP had to be removed.
> 
> MP2 should also be available from the guess. But it is of course better to not print/store something than to print something wrong. I might be mistaken, but given that `occ` does coupled DIIS for amplitudes and orbitals, the canonical LCCD enery should not be available from an OLCCD calculation (lccd is never iterated on canonical orbitals). The same holds for REMP2. So yes, if there are variables pretending to be canonical results from an orbital-optimized calculation, these should probably be removed. No problem, it was mostly my fault to put way too many changes into a single pull request based on an ancient master branch.

This is completely correct. The lone OLCCD algorithm in `occ` does not compute LCCD with the input orbitals.

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -1290,18 +1308,44 @@ PSI Variables by Alpha
    The total electronic energy [Eh] and correlation energy component [Eh]
    for the orbital-optimized MP2 level of theory.
 
+.. psivar:: OMP2 REFERENCE CORRECTION ENERGY
+
+   The additional correction to the SCF reference energy [Eh]
```
JonathonMisiewicz, line 1313: It's unclear to me what an "additional correction to the SCF reference energy" is, but I would have guessed "a correlation energy."

Please change **all** descriptions of variables named like "X REFERENCE CORRECTION ENERGY" to something along the lines of:
`The difference between the single-determinant energy [Eh] of the final and initial orbitals for the X level of theory.`

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -1290,18 +1308,44 @@ PSI Variables by Alpha
...
+.. psivar:: OMP3 REFERENCE CORRECTION ENERGY
+
+   The additional correction to the SCF reference energy [Eh]
+   for the orbital-optimized MP3 level of theory.
+
+.. psivar:: OREMP2 TOTAL ENERGY
+   OREMP2 CORRELATION ENERGY
+
+   The total electronic energy [Eh] and correlation energy component [Eh]
+   for the orbital-optimized retain the excitation perturbation level of theory.
```
JonathonMisiewicz, line 1342: On first read, "retain" looks like it functions as a verb instead of part of a name. Possible changes:
1. retain-the-excitation-perturbation
2. REMP (retain the excitation perturbation)
3. Retain the Excitation Perturbation
4. "retain the excitation perturbation"

@behnle, opinions?

This comment applies to **all** REMP variable descriptions.

loriab:
>>> sometimes the non-OO value (e.g., MP2, LCCD) is available as an early byproduct of the OO calc (e.g., OMP2, OLCCD). this wasn't the case with REMP, and some QCVariables that stored a purported non-OO REMP had to be removed.

>> MP2 should also be available from the guess. But it is of course better to not print/store something than to print something wrong. I might be mistaken, but given that occ does coupled DIIS for amplitudes and orbitals, the canonical LCCD enery should not be available from an OLCCD calculation (lccd is never iterated on canonical orbitals). The same holds for REMP2. So yes, if there are variables pretending to be canonical results from an orbital-optimized calculation, these should probably be removed. No problem, it was mostly my fault to put way too many changes into a single pull request based on an ancient master branch.

> This is completely correct. The lone OLCCD algorithm in occ does not compute LCCD with the input orbitals.


Great, thanks. `oremp2` is in good shape, then, and I'll add negative assertions for `olccd`. Here's the summary of what gets checked (details are at QCEngine): https://github.com/psi4/psi4/pull/2653/files#diff-e2cf14f98c8e885f5abc7385ca737bfeba9f2f62caead630129e6d7cd9678e71R193-R203

doc/sphinxman/source/occ.rst
```diff
@@ -196,7 +196,18 @@ yields
 
 This final equation corresponds to the usual Newton-Raphson step.
 
-Publications resulting from the use of the orbital-optimized code should cite the following publications: 
+* OREMP
+
+The REMP hybrid perturbation theory is a constrined mixture of the |MollerPlesset| perturbation theory and the
```
JonathonMisiewicz, line 201:
```suggestion
The REMP hybrid perturbation theory is a constrained mixture of the |MollerPlesset| perturbation theory and the
```

doc/sphinxman/source/occ.rst
```diff
@@ -196,7 +196,18 @@ yields
...
+Retaining the Excitation degree perturbation theory([Fink:2006:RE]_, [Behnle:2019:REMP]_).
+The mixing ratio is determined by the parameter :math':`A`:
+
+.. math::
+   \widehat{H}^{(0)}_\text{REMP} = (1-A)\widehat{H}^{(0)}_\text{RE} + A\widehat{H}^{(0)}_\text{MP}
+
+Technically, the second order of RE corresponds to LCCD for RHF and UHF references. REMP2 and its orbital-optimized variant OREMP2
+are thus straightforward to implement in a (O)LCCD program by appropriate scaling of residuum vector contributions and density matrices.
```
JonathonMisiewicz, line 209:
```suggestion
are thus straightforward to implement in a (O)LCCD program by appropriate scaling of residual vector contributions and density matrices.
```

doc/sphinxman/source/occ.rst
```diff
@@ -370,6 +389,10 @@ preference to the default module, issue ``set qc_module occ``.
 
 Starting in v1.4, MP2.5 and MP3 default to the density-fit algorithm. Set |globals__mp_type| to ``CONV`` to get previous behavior.
 
+Publications resulting from the use of the non-OO CC codes should cite the following publications:
```
JonathonMisiewicz, line 392:
```suggestion
Publications resulting from the use of the REMP codes should cite the following publications:
```

I'm very confused why this said "non-OO CC" originally. Am I missing something?

doc/sphinxman/source/occ.rst
```diff
@@ -425,4 +448,10 @@ Starting in v1.4, MP2.5 and MP3 default to the density-fit algorithm. Set |globa
     +                         +--------------------------------------------------------------+----------------------+----------------------+---------------------------+
     |                         | Cholesky-Decomposed Lambda-CCSD(T)                           | RHF                  | ---                  | |globals__cc_type| CD     |
     +-------------------------+--------------------------------------------------------------+----------------------+----------------------+---------------------------+
+    | remp2                   | 2nd order REMP hybrid perturbation theory                    | RHF/UHF              | ---                  | |globals__cc_type| CONV   |
```
JonathonMisiewicz, line 451: _Technically_ we don't have density-fitted yet. If we're on track to get that in by 1.7, I leave it up to you whether you want this part of the docs to be accurate in the meantime.

psi4/driver/procrouting/proc.py
```diff
@@ -2205,120 +2284,51 @@ def run_occ(name, **kwargs):
...
-
-    elif name == 'omp2.5':
-        core.set_local_option('OCC', 'WFN_TYPE', 'OMP2.5')
-        core.set_local_option('OCC', 'ORB_OPT', 'TRUE')
-        core.set_local_option('OCC', 'SPIN_SCALE_TYPE', 'NONE')
-    elif name == 'custom-scs-omp2.5':
-        core.set_local_option('OCC', 'WFN_TYPE', 'OMP2.5')
-        core.set_local_option('OCC', 'ORB_OPT', 'TRUE')
-        core.set_local_option('OCC', 'SPIN_SCALE_TYPE', 'CUSTOM')
+    director = {
```
JonathonMisiewicz, line 2287: 👍

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -41,7 +41,7 @@ namespace occwave {
 void OCCWave::cepa_iterations() {
     outfile->Printf("\n  \n");
     outfile->Printf(" ============================================================================== \n");
-    outfile->Printf(" ================ Performing LCCD iterations... =============================== \n");
+    outfile->Printf(" ================ Performing LCCD iterations with REMP modification... =============================== \n");
```
JonathonMisiewicz, line 44: Do we always want this to run? What if it is a genuine CEPA computation?

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -55,38 +55,42 @@ void OCCWave::cepa_iterations() {
                  // DIIS
     DIISManager t2_diis;
     if (nooA + nooB != 1) {
-        psio_->open(PSIF_OCC_DPD, PSIO_OPEN_OLD);
```
JonathonMisiewicz, line 58: What's the rationale for moving the `psio` calls inside the if-clauses? This should always run.

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -55,38 +55,42 @@ void OCCWave::cepa_iterations() {
...
+            psio_->close(PSIF_OCC_DPD, 1);
         }
-        psio_->close(PSIF_OCC_DPD, 1);
     }
 
     // head of loop
     do {
         itr_occ++;
         timer_on("T2");
+//        t2_amps();
```
JonathonMisiewicz, line 93:
```suggestion
```

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -135,6 +139,108 @@ void OCCWave::cepa_iterations() {
...
+
+        outfile->Printf(" %3d     %12.10f    %12.10f  %12.2e %12.2e \n", itr_occ, Ecorr, Ecepa, DE, rms_t2);
+
+        if (itr_occ >= cc_maxiter) {
+            conver = 0;  // means iterations were NOT converged
+            break;
+        }
+
+        if (rms_t2 >= DIVERGE) {
+            throw PSIEXCEPTION("LCCD iterations are diverging");
```
JonathonMisiewicz, line 219:
```suggestion
            throw PSIEXCEPTION("REMP iterations are diverging");
```

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -135,6 +139,108 @@ void OCCWave::cepa_iterations() {
...
+            global_dpd_->buf4_close(&Tab);
+            psio_->close(PSIF_OCC_DPD, 1);
+        }
+    }
+
+    // head of loop
+    do {
+        itr_occ++;
+        timer_on("T2");
+//        t2_amps();
```
JonathonMisiewicz, line 194:
```suggestion
```

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -135,6 +139,108 @@ void OCCWave::cepa_iterations() {
...
+        t2_amps_remp(); // <- the only actual modification compared to regular CEPA(0)/D
+        timer_off("T2");
+        timer_on("CEPA Energy");
+        cepa_energy();
+        cepa_diis(t2_diis);    // <- CEPA diis can be reused without modifications
+        cepa_chemist();
```
JonathonMisiewicz, line 200: This is a good rationale for renaming functions, so that things labeled `cepa` are actually `cepa` only. I'd be thrilled if we could condense this and the CEPA code into a single function that just passes a flag around.

psi4/src/psi4/occ/coord_grad.cc
```diff
@@ -1796,16 +1796,34 @@ void OCCWave::effective_gfock() {
...
     // Compute oeprop
     auto oe = std::make_shared<OEProp>(shared_from_this());
+    oe->set_Da_mo(Da_);
+    if (reference_ == "UNRESTRICTED") oe->set_Db_mo(Db_);
     oe->add("DIPOLE");
     oe->add("QUADRUPOLE");
     oe->add("MULLIKEN_CHARGES");
     oe->add("NO_OCCUPATIONS");
-    oe->set_title("DF-OMP2");
+    oe->set_title(wfn_type_);
```
JonathonMisiewicz, line 1823: This `set_title` change is good. The rest of these changes are unnecessary: if the density is not explicitly set, `OEProp` will extract it from the wavefunction, which is where we want it. The extra code in `dfocc` is a problem.

If reverting these changes gives you any test failures, let me know, because that indicates a bug in `occ`.

psi4/src/psi4/occ/manager.cc
```diff
@@ -66,7 +66,8 @@ void OCCWave::omp2_manager() {
     if (ip_poles == "TRUE") omp2_ip_poles();
     if (ep_ip_poles == "TRUE") ep2_ip();
 
-    mp2_postprocessing();
+    // ordinary ROHF-MP2 not available in course of ROHF-OO
```
JonathonMisiewicz, line 69: :beer:

psi4/src/psi4/occ/manager.cc
```diff
@@ -693,6 +699,210 @@ void OCCWave::cepa_manager() {
...
+    // EcepaL = Ecepa;
+
+    std::map<std::string, double> spin_scale_energies = {{"CUSTOM", variables_["CUSTOM SCS-REMP2 TOTAL ENERGY"]},
+                                                         {"NONE", Ecepa}};
+    variables_["CURRENT ENERGY"] = spin_scale_energies[spin_scale_type_];
+    variables_["CURRENT CORRELATION ENERGY"] = variables_["CURRENT ENERGY"] - variables_["CURRENT REFERENCE ENERGY"];
+    energy_ = variables_["CURRENT ENERGY"];
+
+    // Compute Analytic Gradients
+    // NO ANALYTIC GRADIENTS YET
```
JonathonMisiewicz, line 866: So we can leave a comment for future work, what's missing to compute analytic gradients? I'd imagine the big issue is the lambda equations and the assembly of the response density matrices from those. The orbital Z-vector terms would need to be coded, but I'd think those are straightforward. We already have the general machinery to compute those, once we have the orbital gradient. I'd expect that to be straightforward once you have the response density matrices.

JonathonMisiewicz (commented):
Overall, I'm happy with this. Huge thanks to both @loriab and @behnle.

doc/sphinxman/source/glossary_psivariables.rst
behnle: Oh, originally, it was "retaining", so "RE-PT" is the "Retaining the Excitation degree Perturbation Theory" (an alternative and stringent way to derive LCCD as PT, but with the option to go to higher orders, https://doi.org/10.1016/j.cplett.2006.07.081).
REMP is the hybrid between RE-PT and MP-PT.

doc/sphinxman/source/occ.rst
behnle: Me bad, thanks for spotting.

doc/sphinxman/source/occ.rst
loriab: yes, there's a list of all the other methods and citations that I didn't pick into this PR. Probably best to consider docs as a whole later on.

psi4/src/psi4/occ/manager.cc
behnle: This is completely correct, but given that the canonical methods have only mediocre performance (we assessed that recently), i am not sure whether it is worth the effort at all.

JonathonMisiewicz: I completely agree about efficiency. Still, I'd prefer a comment explaining this in the code, so that if somebody is looking over this five years from now, they don't have to ask, "Is this code commented because it was mostly coded but just needed some debugging, or for some other reason?"

doc/sphinxman/source/glossary_psivariables.rst
loriab: ah, thanks for the clarification. I'm going with `retaining-the-excitation-degree |MollerPlesset| hybrid perturbation theory level`. that, plus `REMP` in the qcvar name should provide enough clues if read carefully to distinguish MPPT, REPT, REMP.

loriab:
@behnle, were there any changes from the review that you wanted to PR into this PR? Otherwise, I'll force push my accumulated responses.

behnle:
As far as i can see, this is fine as is.

psi4/src/psi4/occ/cepa_iterations.cc
loriab: afai see, `cepa_iterations` gets run for LCCD and `remp_iterations` for REMP2, so I'm removing the "with REMP modification"

psi4/src/psi4/occ/cepa_iterations.cc
loriab: This may have been the case I consulted you on where former restricted=ondisk and unrestricted=incore got consolidated to both ondisk. So this reverts.

psi4/src/psi4/occ/cepa_iterations.cc
loriab: for the future ...

JonathonMisiewicz (approved):
🍺

jturney (approved):

psi4/src/psi4/occ/cepa_iterations.cc
```diff
@@ -55,32 +55,35 @@ void OCCWave::cepa_iterations() {
...
         else if (reference_ == "UNRESTRICTED") {
             dpdbuf4 Taa, Tbb, Tab;
+            psio_->open(PSIF_OCC_DPD, PSIO_OPEN_OLD);
             global_dpd_->buf4_init(&Taa, PSIF_OCC_DPD, 0, ID("[O,O]"), ID("[V,V]"), ID("[O,O]"), ID("[V,V]"), 0,
                                    "T2 <OO|VV>");
             global_dpd_->buf4_init(&Tbb, PSIF_OCC_DPD, 0, ID("[o,o]"), ID("[v,v]"), ID("[o,o]"), ID("[v,v]"), 0,
                                    "T2 <oo|vv>");
             global_dpd_->buf4_init(&Tab, PSIF_OCC_DPD, 0, ID("[O,o]"), ID("[V,v]"), ID("[O,o]"), ID("[V,v]"), 0,
                                    "T2 <Oo|Vv>");
+            t2_diis = DIISManager(maxdiis_, "CEPA DIIS T2 Amps", DIISManager::RemovalPolicy::LargestError, DIISManager::StoragePolicy::InCore);
```
hokru, line 79: Is there a reason for the different the `StoragePolicy`'s for RHF and UHF?

hokru (approved):

psi4/src/psi4/occ/cepa_iterations.cc
behnle: No... Should probably be both ``OnDisk``. I also propose that the initialization of t2_diis is taken out of the
```
if (reference_ ==  "RESTRICTED") {
...
}
else if (reference_ == "UNRESTRICTED") {
...
}
```
block. It seems redundant to me.

tests/pytests/standard_suite_runner.py
```diff
@@ -107,6 +109,8 @@ def runner_asserter(inp, subject, method, basis, tnm):
         # _recorder(qcprog, qc_module_in, driver, method, reference, fcae, scf_type, corl_type, "error", "nyi: " + reason)
         return
 
+    psi4.set_output_file("asdf")
```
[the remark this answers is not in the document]
loriab: Right now it's a hack solution to a mystery. Each job is supposed to produce a `.out` and a `.log` file. The log file shows up, but the out file doesn't even though I cancel deletions in conftest.py. If I change the base to `asdf`, both files show up, so I can debug. So, intentional but temporary.

psi4/src/psi4/occ/cepa_iterations.cc
loriab: I think the different policies is what Ugur had. But makes sense to unify them. I'll do it in PR 3, though, which will go up today.

JonathonMisiewicz: For context, during my previous changes in the `occ` module, I unified them as `behnle` suggested. This was an unintended behavior change. Lori is changing back to how Ugur had it _before_ my interventions.

JonathonMisiewicz:
Three cheers for three approvals. I'm deferring merge to Lori on a PR like this.

behnle:
Thank you all very mch guys, if i ever meet you, i definitely owe you a :beer: .

loriab:
> Thank you all very mch guys, if i ever meet you, i definitely owe you a 🍺 .

You're very welcome. Thanks for (1) the new methods, (2) the additional motivation to address the technical debt in occ/dfocc, and (3) your fixes to dfocc convergence to make (2) possible :-) . I owe you a return 🍺 and certainly a Psi mug. We'll have to coordinate addresses/conferences on the latter.

Issue #2555, opened 2022-04-22
clean up tests directory


loriab:
after #2549, the list from `python tester.py` looks like the below. Most of these are easily healed by figuring out if they're intended to be a test or not. Nots should be removed to an attic or deleted. Also related is #2234 . don't rely on the below -- rerun tester.py as needed.

@psi-rking, do you prefer attic or deletion for the un-run optimizer tests?

Complaints
----------
- [ ]   1. cc5: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]   2. cookbook/manual-sow-reap: missing cmake directory registration. `vi cookbook/CMakeLists.txt`
- [ ]   3. cookbook/manual-sow-reap: missing CMakeLists. `vi cookbook/manual-sow-reap/CMakeLists.txt`
- [ ]   4. dfmp2-freq1: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]   5. dfmp2-freq2: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]   6. dfomp2p5-1: missing ctest registration. `vi dfomp2p5-1/CMakeLists.txt`
- [ ]   7. dfomp2p5-2: missing ctest registration. `vi dfomp2p5-2/CMakeLists.txt`
- [ ]   8. dfomp2p5-grad1: missing ctest registration. `vi dfomp2p5-grad1/CMakeLists.txt`
- [ ]   9. dfomp2p5-grad2: missing ctest registration. `vi dfomp2p5-grad2/CMakeLists.txt`
- [ ]  10. dft-dldf: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]  11. dft-dsd: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]  12. dft-pbe0-2: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]  13. explicit-am-basis: mismatched directory (explicit-am-basis) and ctest registration name (explicit_am_basis). `vi explicit-am-basis/CMakeLists.txt`
- [ ]  14. fsapt-diff1: missing cmake directory registration. `vi CMakeLists.txt`
- [ ]  15. fsapt-diff1: missing CMakeLists. `vi fsapt-diff1/CMakeLists.txt`
- [ ]  16. gcp/hf3c-gradient: mismatched marks ctest () and pytest (gradient). `vi gcp/hf3c-gradient/CMakeLists.txt gcp/hf3c-gradient/test_input.py`
[...]

PR #2655, opened 2022-08-02
Test Directory Cleanup

philipmnel:
## Description
Cleanup of tests directory following #2555. Heals various typos preventing some tests from getting pytest registration as well as some discrepancies between ctest and pytest labels. Deletes many unused/broken tests.

## Todos
- [x] All unused optimizer tests deleted following discussion in #2555.
- [x] All unused sowreap tests deleted.
- [x] `cc5`, `scf11-freq-from-energies`,`dft-pbe0-2`, `dft-dsd`, `dft-dldf` are all unused duplicates of other tests. Deleted.
- [x] `min-input` Broken + unused. Deleted.
- [x] `fsapt-diff1` Unused + not a test. Deleted. 
- [x] `dfmp2-freq1`, `dfmp2-freq2`, `props4` were unused but are passing, quick, and useful. These are saved and added to `CMakeLists.txt`.
- [x] `tester.py` no longer erroneously complains about ecppint tests having mismatched labels.

## Checklist
- [x] All tests that were missing registrations are passing.

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
For every duplicate test, I want to know where the duplicate is.

philipmnel:
`cc5` is a near duplicate of `cc5a`. The latter uses a smaller basis set as `cc5` is very slow.
`scf11-freq-from-energies` is a near duplicate of `fd-freq-energy-large` with a slightly different geometry. The former does not pass without adjusting findif stepsize settings.
`dft-pbe0-2` is duplicated verbatim in `dft-custom-dhdf`
`dft-dsd` is duplicated in `dft-custom-dhdf` with corrected reference values. Former does not pass.
`dft-dldf` is duplicated in `dft-custom-mgga` with corrected reference values. Former does not pass.

JonathonMisiewicz:
Okay. Please rename `cc5a` to `cc5` and update `tests/cc_index`.

tests/cc5/input.dat
```diff
@@ -1,41 +0,0 @@
-#! RHF CCSD(T) aug-cc-pvtz frozen-core energy of C4NH4 Anion
-
-memory 12 gb
-
-refnuc   =  135.0921284884196    #TEST
-refscf   = -208.2159793502       #TEST
-refccsd  =   -0.884280512675734  #TEST
-ref_t    =   -0.048772959402886  #TEST
-reftotal = -209.149032822278599  #TEST
```
loriab, line 9: copy this bit commented into the `cc5a` file, with a comment `aug-cc-pvtz`, so it's not lost if someone wants to run it again by hand.

tests/cookbook/manual-sow-reap/input.dat
```diff
@@ -1,124 +0,0 @@
-#! Frequency by finite difference of energies.
-#! Automated version of this in pywrap-freq-e-sowreap. !nosample
```
loriab, line 2: was this test not working right? it was the `energy(..., mode="sow")` that were definitely worth getting rid of. this might still be a useful model if it works.

tests/fsapt-diff1/input.dat
```diff
@@ -1,71 +0,0 @@
-#! This test case shows an example of running and analyzing a difference
-#! F-SAPT0/jun-cc-pvdz procedure for phenol dimer from the S22 database.
-
```
loriab, line 3: in what way isn't this demo working as a plain test case? it seems like it has some good guidance that I wouldn't like to lose if it doesn't live somewhere else, too.

loriab (commented):
looking good, thanks!

tests/cookbook/manual-sow-reap/input.dat
philipmnel: Currently not working. It seems `driver_findif.hessian_from_energies_geometries()` is generating displaced geometries as numpy arrays which `manual-sow-reap` is trying to JSON dump. Potentially an easy patch if this is worth keeping.

loriab: @psi-rking, is it ok with you to drop the manual-sow-reap example file? I could maybe make it work, but it'd be awkward.

psi-rking: @loriab, I think these are now irrelevant.  Their replacements are in psi4/samples/cookbook/manual-...
One of these (manual-fd-hess-grad) I just modified last week usefully - so I hope those are staying?

loriab: Thanks. Yes, the manual-fd ones are staying.

JonathonMisiewicz (approved):

tests/fsapt-diff1/input.dat
philipmnel: `fsapt-diff1/input.dat` is just a copy of `fsapt1/input.dat` (though looks like needs an updated reference value, cf. https://github.com/psi4/psi4/blame/master/tests/fsapt1/input.dat#L53). I'll add this back since it looks like the reference output (`fsaptD-ref`) is useful, but it might be nice to actually add a test for this data at some point so we know it is actually reproducible (it does seem to be currently).

loriab: neat, the nre was all it needed! I went from having 100 files left to review to having one loc change :-)

agree fsapt-diff1 could use some attention, but that can be future pass. thanks for patching it up.

loriab (approved):
Thanks very much, including improving the tester script. Nice negative LOC!

tests/tester.py
```diff
@@ -136,11 +138,16 @@ def filter_marks(marks, nested):
                 testinputpy = fp.read()
 
             pymarks = []
-            mobj = re.search(r'^@ctest_labeler\("' + r"(?P<pymarks>([a-z0-9-;]+))" + r'\"\)', testinputpy, re.MULTILINE)
+            pyplugins = []
+            mobj = re.search(r'^@uusing\("' + r"(?P<pyplugins>([a-z0-9-_;]+))" + r'\"\)', testinputpy, re.MULTILINE)
+            if mobj:
+                pyplugins = mobj.group("pyplugins").split(";")
```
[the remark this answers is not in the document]
philipmnel: Thanks for catching, will fix. (I think the only case where this bit is currently does anything is the ecpint tests since other plugin tests are nested in folders, which is already dealt with)

hokru (approved):
lgtm. Merge conflicts need to be addressed before PR can continue.

philipmnel:
Merge conflicts have been fixed.

loriab:
> @loriab Is there a hold on merging this?

No hold; merge away.

PR #2537, opened 2022-04-08
change solid harmonic ordering from gaussian to standard

loriab:
## Description
Kudos to @andysim for pointing out in January the exact 13 lines of code to change for the vast preponderance of the gaussian to standard ordering switch.

As far as timeline, there's no particular reason this needs to be in for v1.6. But if we _could_ avoid hosting forever one fewer set of L2 packages for a single release, that'd be kind of nice. But fixing the output formats (i.e., fchk, qcsk, molden, fcidump) is likely to take some careful work.

If anyone would like to tackle a part of this, go for it.

## Todos

- [ ] current test status (pretty impressive, imo)
  - after g2g commit
    - ctest: `99% tests passed, 3 tests failed out of 487`
    - pytest: `12 failed, 2971 passed, 270 skipped, 269 xfailed`
  - after first solidharmonics.cc commit
    - ctest: ~`90% tests passed, 50 tests failed out of 487`~
    - pytest: ~`43 failed, 2911 passed, 208 skipped, 269 xfailed`~
- [ ] only Linux in CI b/c only it has an appropriate L2 conda package built

- categories to examine
  - [x] tdscf
  - [x] cubeprop
  - [ ] fchk
  - [ ] qcsk atomicresult.wavefunction
  - [x] mbis
  - [ ] molden
  - [ ] fcidump
  - [ ] nbo

- fails pytest
  - [x]  tdscf
  - [x] dft_block_schemes
  - [ ] mcmurchie_davidson_consistency_angmom (fails because ref data are deposited in Gaussian ordering)
  - [ ] test_rhf_fchk[df-rhf-dft)]
  - [ ] test_rhf_fchk[conv-rhf-dct]
  - [ ] test_rhf_fchk[df-rhf-omp2]
  - [ ] test_uhf_fchk[df-uhf-mp2]
  - [ ] test_rhf_fchk[conv-rhf-ccsd] 
  - [ ] test_uhf_fchk[df-uhf-dft]
  - [ ] test_uhf_fchk[conv-uhf-ccsd]
  - [ ] test_rhf_fchk[df-rhf-mp2]
  - [ ] test_uhf_fchk[df-uhf]
  - [ ] test_rhf_fchk[df-rhf)]
  - [ ] test_rhf_fchk[conv-rhf-cc2]

- fails ctest
  - [x]	103 - cubeprop (Failed)
  - [x]	104 - cubeprop-frontier (Failed)
  - [x]	123 - density-screen-2 (Timeout)
[...]

PR #2665, opened 2022-08-08
CFMM Part 1/3: Add Libfmm and Multipole Translation and Rotation Functions

andyj10224:
## Description
This is the First PR in a series of three that will eventually get White and Head-Gordon's CFMM method into Psi4. CFMM is a method that builds J-like matrices in O(N) time, such as the J matrix in Hartree-Fock, intermediates in density-fitted HF, and J-like intermediates in SAPT.

References:
https://www.sciencedirect.com/science/article/pii/0009261494011281
Helgaker Chapter 9.13-9.14

## Todos

- [x] Add libfmm
- [x] Adds multipole translation and rotation functionalities to Psi4

## Questions
- [ ] Is this a good way to break up the CFMM PR? (1. libfmm, 2. fmm tree, 3. CFMM JK class)

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
Immediate questions:
@davpoolechem - Is JK stable enough for this PR, or do you need a hold on this while refactoring?
@jturney - Does this sound more like a new library of its own, or something that belongs in `libmints`?
@andyj10224 - Can you comment on what the difference is between `libffm` and `fmm tree`?

andyj10224:
> Immediate questions: @davpoolechem - Is JK stable enough for this PR, or do you need a hold on this while refactoring? @jturney - Does this sound more like a new library of its own, or something that belongs in `libmints`? @andyj10224 - Can you comment on what the difference is between `libffm` and `fmm tree`?

CFMM requires two parts: Multipole operations, as well as an octree structure.
FMMTree implements the octree structure needed for CFMM, and there is a lot of code there, so I am splitting this into three parts to aid the review process and to avoid too large a PR

JonathonMisiewicz:
Great. I'll proceed with reviewing this once Jet comments on whether this should be part of `libmints` or not.

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,217 @@
...
+        return 0;
+    }
+    int small = std::min(n, n-r);
+    int nCr = 1;
+    for (int t = 0; t < small; t++) {
+        nCr *= n;
+        nCr /= (t+1);
+        n -= 1;
+    }
+    return nCr;
```
loriab, line 68: psi4 is c++17. if this is indeed builtin, can it be used for this instead of hand-written helper fn? https://stackoverflow.com/a/67196949

jturney:
How does the solid harmonics code different from the solid harmonics code found in libmints? 
https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/solidharmonics.cc

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,217 @@
...
+
+// Some more helper functions
+static inline int ncart(int l) { return (l+1)*(l+2)/2; }
+
+// Some more useful Helper Functions
+static int choose(int n, int r) {
+    if (r < 0 || r > n) {
+        return 0;
+    }
+    int small = std::min(n, n-r);
```
zachglick, line 61: should this be `min(r, n-r)`?

susilehtola:
Per the last Psi4 call, I'd like to challenge you to think about how much of libcfmm could be factored out of Psi4. Many other codes might be interested in this functionality, too, if it can be made accessible through a simple API.

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,217 @@
...
+        return 2*(-m);
+    } else {
+        // 1c, 2c, 3c, ...
+        return 2*m-1;
+    }
+}
+
+extern double factorial(int n);
+
+class PSI_API MultipoleRotationFactory {
```
zachglick, line 84: Could we get a brief explanation of the purpose of this class?

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,217 @@
...
+      // Maximal Angular Momentum
+      int lmax_;
+
+      // Cached Rotation Matrices in a vector of Matrices
+      std::vector<SharedMatrix> D_cache_;
+
+      double U(int l, int m, int M);
+      double V(int l, int m, int M);
+      double W(int l, int m, int M);
+      double P(int i, int l, int mu, int M);
```
zachglick, line 102: Could we get either a very brief explanation of what these functions do, or a reference to the paper/book in which they're defined (ideally w/ equation numbers)?

zachglick (approved):
Thanks for the clear, well-documented code Andy! Just a few minor comments

andyj10224:
> Per the last Psi4 call, I'd like to challenge you to think about how much of libcfmm could be factored out of Psi4. Many other codes might be interested in this functionality, too, if it can be made accessible through a simple API.

I like your idea, but I think that is something to think about in the future because the rest of `libfmm` is currently entangled with Psi functionalities (like `IntegralFactory`).

andyj10224:
> How does the solid harmonics code different from the solid harmonics code found in libmints? https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/solidharmonics.cc

My code here uses the Helgakler multipole normalization procedure, and it introduces the multipole rotation and translation operations needed for CFMM. Does solidharmonics.cc in libmints use the Helgaker convention?

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Oops, my code still works since n-r is always less than r, but good catch

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: I will try to replace this with that call, and make sure it won't break my existing CFMM code.

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Done

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Done

andyj10224:
> > How does the solid harmonics code different from the solid harmonics code found in libmints? https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/solidharmonics.cc
> 
> My code here uses the Helgaker multipole normalization convention, and it introduces the multipole rotation and translation operations needed for CFMM. Does solidharmonics.cc in libmints use the Helgaker convention? If y'all use the same convention, then I can scrap my `HarmonicCoefficients` class in favor of solidharmonics.cc from libmints.

Iirc, @andysim  said that the old libmints solidharmonics.cc code is outdated anyways, and the Helgaker recursions are better anyways. Feel free to comment Andy.

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Done! Did not break my code!

jturney:
Fair point if the solidharmonics.cc code is outdated. I was curious if you were re-inventing something we already have.

@JonathonMisiewicz I'll hold off commenting on whether a separate library is needed until I see the other PRs.

andysim:
> Iirc, @andysim said that the old libmints solidharmonics.cc code is outdated anyways, and the Helgaker recursions are better. Feel free to comment Andy.

Both codes use regular solid harmonics with the exact same normalization (assuming that your Helgaker recursions match that code that I sent to you a while ago, Andy).  The code in libmints is stuff that was taken from MPQC, but becomes painfully slow around L=9 or so.  We had to reduce the default maximum L a while ago because it was dramatically slowing the Psi4 startup time.  So I think that, if it turns out not to be a huge task, replacing the libmints code with smarter recursions is probably not a bad idea.  Failing that, the Helgaker version is not really introducing any redundancy because the libmints version won't be able to support high enough a.m. efficiently.

andyj10224:
> > Iirc, @andysim said that the old libmints solidharmonics.cc code is outdated anyways, and the Helgaker recursions are better. Feel free to comment Andy.
> 
> Both codes use regular solid harmonics with the exact same normalization (assuming that your Helgaker recursions match that code that I sent to you a while ago, Andy). The code in libmints is stuff that was taken from MPQC, but becomes painfully slow around L=9 or so. We had to reduce the default maximum L a while ago because it was dramatically slowing the Psi4 startup time. So I think that, if it turns out not to be a huge task, replacing the libmints code with smarter recursions is probably not a bad idea. Failing that, the Helgaker version is not really introducing any redundancy because the libmints version won't be able to support high enough a.m. efficiently.

Since the Helgaker code is more efficient, should I do that replacement in this PR or the next?

andyj10224:
@jturney @JonathonMisiewicz @andysim It appears that the current solidharmonics.cc is not used ANYWHERE in the Psi4 code. I am actually in favor of moving all my fmm stuff to libmints. Here is the plan:

libfmm/multipoles_helper.h -> libmints/solidharmonics.h
libfmm/multipoles_helper.cc -> libmints/solidharmonics.cc

Thoughts?

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,230 @@
...
+#include <memory>
+#include <tuple>
+#include <vector>
+#include <unordered_map>
+
+namespace psi {
+
+enum SolidHarmonicsType {Regular, Irregular};
+
+// Copied stuff from libmints/solidharmonics.cc
```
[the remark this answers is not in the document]
andyj10224: Ah yes, the reason they are needed here is that those functions are needed in `fmm_tree.cc` later and we do not want to import the whole `multipoles_helper.cc` in `fmm_tree.cc`.

loriab:
> It appears that the current solidharmonics.cc is not used ANYWHERE in the Psi4 code.

Somehow solidharmonics.cc is getting used, because changing it changes the integral ordering (link is to the `gss` to `sss` PR: https://github.com/psi4/psi4/pull/2537/files#diff-76f4c1f378b6a48ede6c4f938378b54719f968680edca78453d9751e87cebbf9)

JonathonMisiewicz:
> > It appears that the current solidharmonics.cc is not used ANYWHERE in the Psi4 code.
> 
> Somehow solidharmonics.cc is getting used, because changing it changes the integral ordering (link is to the `gss` to `sss` PR: https://github.com/psi4/psi4/pull/2537/files#diff-76f4c1f378b6a48ede6c4f938378b54719f968680edca78453d9751e87cebbf9)
> 
> EDIT: not necessarily a block on your plan, but that's not isolated code

`transform.cc` is probably the file responsible for that. Again, not necessarily a block, just wanted to point that out before my attention moved over to something else.

andyj10224:
> > It appears that the current solidharmonics.cc is not used ANYWHERE in the Psi4 code.
> 
> Somehow solidharmonics.cc is getting used, because changing it changes the integral ordering (link is to the `gss` to `sss` PR: https://github.com/psi4/psi4/pull/2537/files#diff-76f4c1f378b6a48ede6c4f938378b54719f968680edca78453d9751e87cebbf9)
> 
> EDIT: not necessarily a block on your plan, but that's not isolated code

Looks like I did a naive `grep -r "solidharmonics.cc"` instead of its functions. My bad. In that case, refactoring solidharmonics.cc with my new code can serve as an effective test for my functions, but it does make it harder :)

JonathonMisiewicz:
My vote is to either end or temporarily pause this PR and open a new PR just for swapping out the solid harmonics technology. Of course, @jturney's judgment overrides mine.

andyj10224:
> My vote is to either end or temporarily pause this PR and open a new PR just for swapping out the solid harmonics technology. Of course, @jturney's judgment overrides mine.

I do think we can essentially move the methods in the `HarmonicCoefficients` class to `libmints/solidharmonics.cc`.  But I agree that would be 2 PRs. In that case, having a separate `libfmm` would be useful, to avoid the clutter already present in `libmints`. Of course, waiting for @jturney's judgement.

psi4/src/psi4/libfmm/multipoles_helper.h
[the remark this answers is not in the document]
jturney: Then they should exist in fmm_tree.cc since neither the helper header or implementation file needs them.

andyj10224:
Made stale by #3493

PR #2666, opened 2022-08-08
Remove legacy check

JonathonMisiewicz:
## Description
Remove a now-meaningless validation step in `detci`.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):
An allocation in psi3 that used this has gone, and detcas that used this hasn't left psi3, so I venture removing this check is safe.

jturney (approved):

hokru (approved):

Issue #2631, opened 2022-07-12
MBIS_VOLUME_RATIOS with num_frozen_docc set


tallakahath:
I have situations where I want to manually specify the number of frozen core orbitals using `num_frozen_docc`, do some work, then calculate some MBIS properties including the volume ratios. The latter causes psi4 to automatically calculate the free-atom volumes in the same method as the last-run calculation (in my case, `wb97m-d3bj`) and inherits all the settings... including `num_frozen_docc`. I can insert a `set num_frozen_docc 0; set freeze_core true` right before the `oeprop(wfn, 'MULLIKEN_CHARGES')` call and make things go away, but what worries me is just _how_ things fail.

Well, sometimes things are in fact _perfectly fine_. Other times, psi4 **segfaults**. The behavior depends on _just how large_ `num_frozen_docc` is -- I think if it ends up exceeding some other quality dependent on the calculation, things go belly-up. Here's the relevant line from a coredump.

```
#0  0x00002b4f362ef958 in psi::scf::HF::compute_fcpi (this=0x223c06c0) at [snip]/psi4/psi4/src/psi4/libscf_solver/hf.cc:830
830             for (int i = 0; i < nfzc; ++i) frzcpi_[pairs[i].second]++;
```

And here's a minimum working example of the issue that MBIS_VOLUME_RATIOS runs into, as a standalone calc:

```
set num_frozen_docc 20
set basis def2-tzvppd
set reference uks
molecule mol {
0 2
H 0 0 0
units angstrom
}
E, wf = energy('wb97m-d3bj',return_wfn=True,)
```

If you flip `num_frozen_docc` to 18, _it works_.  (19 works for me on some machines, not on others -- eek!). In all cases I'm invoking psi4 the same way: `psi4 -n1 -i run.in --memory 500MB`. 

I'm building off of 1.6.X, using icc/2020.2-108-02c7; I don't _think_ I have anything "special" set in my compile environment besides `buildtype=RelWithDebInfo`. I didn't mess with any optimization flags, for sure.

JonathonMisiewicz:
Thanks for the detailed report. I can reproduce the segfault, though for the benefit of other developers, I'll point out that this can be reproduced with `scf` instead of `wb97m-d3bj`. Sophisticated DFT functionality is not the issue here. This looks like a missing option validation at runtime.

I'll add that my advisor is _also_ unhappy with the way Psi4 assumes options for precisely situations like these, but I have a few other things to patch before I touch  options passing.

[...]

PR #2667, opened 2022-08-08
Add customizable FREEZE_CORE policy option

tallakahath:
## Description
This patch adds an option to FREEZE_CORE called "policy", which enables frozen core settings to be looked up from a list specified in the global variable FREEZE_CORE_POLICY. This is more flexible than NUM_FROZEN_DOCC for situations like SAPT where multiple molecules are run in the same command and may require different individual numbers of frozen cores, with settings more customizable than TRUE/FALSE/1/0/-1/-2

This patch addresses some (but not all) issues raised in #2631 by allowing for more flexible policies to be set appropriate to multi-part calculations. 

## Todos
- [x] Add POLICY as option to FREEZE_CORE
- [x] Add global variable FREEZE_CORE_POLICY to hold custom frozen-core policy

## Checklist
- [x] A functionality test for this flag has been added to `tests/dfmp2-ecp/input.dat`
- [x] `ctest -L quick` runs successfully, which includes the above listed test
- [ ] `ctest ` still in-flight but given the scope of this patch I don't expect any issues

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -121,12 +121,12 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
 
     /*- Specifies a custom frozen-core policy on a per-element basis. Input
     should be a list of integers representing the number of orbitals to freeze
-    for each atomic number. For example, to specify that elements H-Be should
-    have 0 frozen orbitals, B-Mg should have 1, and Al should have 2, you would
-    provide the input ``[0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 2]``. Please
-    make sure to fill in the list up to the highest atomic number included in
-    any calculations. This option is only used if |globals__freeze_core| is set
-    to ``POLICY``. -*/
+    for each atomic number MINUS one (so H is 0, He is 1, etc). For example, to
+    specify that elements H-Be should have 0 frozen orbitals, B-Mg should have
+    1, and Al should have 2, you would provide the input ``[0, 0, 0, 0, 1, 1,
+    1, 1, 1, 1, 1, 1, 2]``. Please make sure to fill in the list up to the
+    highest atomic number included in any calculations. This option is only
+    used if |globals__freeze_core| is set to ``POLICY``. -*/
     options.add("FREEZE_CORE_POLICY", new ArrayType());
 
     options.add("NUM_GPUS", 1);
--- a/psi4/src/read_options.cc
+++ b/psi4/src/read_options.cc
@@ -114,8 +114,20 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     number of frozen orbitals can be attained by using the keywords
     |globals__num_frozen_docc| (gives the total number of orbitals to freeze,
     program picks the lowest-energy orbitals) or |globals__frozen_docc| (gives
-    the number of orbitals to freeze per irreducible representation) -*/
-    options.add_str("FREEZE_CORE", "FALSE", "FALSE TRUE 1 0 -1 -2 -3");
+    the number of orbitals to freeze per irreducible representation) or by
+    the option ``POLICY`` in combination with appropriate inputs to
+    |globals__freeze_core_policy| -*/
+    options.add_str("FREEZE_CORE", "FALSE", "FALSE TRUE 1 0 -1 -2 -3 POLICY");
+
+    /*- Specifies a custom frozen-core policy on a per-element basis. Input
+    should be a list of integers representing the number of orbitals to freeze
+    for each atomic number. For example, to specify that elements H-Be should
+    have 0 frozen orbitals, B-Mg should have 1, and Al should have 2, you would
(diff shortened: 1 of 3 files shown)


jturney:
This is a little confusing to me.

It looks like basisset.cc line 240 accounts for Z = 0 but then the example array in the documentation at read_options.cc line 126 seems to have the Z = 0 in it. Or the example array could be mistyped (H-Be is just four atoms while five zeros are given). Then the example input file has the Sb atom (Z = 51) but `core_policy[50]` is used. I suspect users will be confused by this.

I'm not saying whether or not including Z=0 in the array is correct but documentation should be very clear.

tallakahath:
> Or the example array could be mistyped (H-Be is just four atoms while five zeros are given). 

Good catch! It was in fact a typo on my part, I've fixed it + added a little more clarification in the documentation for `FREEZE_CORE_POLICY`.

jturney:
Just one small addition to make and then it looks good to me.

tallakahath:
Updated!

Tbh, for large elements the use starts to get real clunky, but I don't understand the codebase well enough to instead provide a dict vs an array. Someone savvier than me should probably eventually update that (so that a user doesn't need to set 0's for a bunch of elements they Just Don't Care About), but I didn't see a mapping type available in the relevant context (just an int vector) so I went with what was there.

jturney:
Ah, yeah, for large elements it will get clunky. We _do_ have `MapType` defined [here](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/liboptions/liboptions.h#L335). But looking through the code, it looks like we don't use it anywhere and therefore may or may not work.  I think for now your solution is probably fine.

jturney (approved):

psi4/src/read_options.cc
```diff
@@ -114,8 +114,20 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
     |globals__num_frozen_docc| (gives the total number of orbitals to freeze,
     program picks the lowest-energy orbitals) or |globals__frozen_docc| (gives
-    the number of orbitals to freeze per irreducible representation) -*/
-    options.add_str("FREEZE_CORE", "FALSE", "FALSE TRUE 1 0 -1 -2 -3");
+    the number of orbitals to freeze per irreducible representation) or by
+    the option ``POLICY`` in combination with appropriate inputs to
+    |globals__freeze_core_policy| -*/
+    options.add_str("FREEZE_CORE", "FALSE", "FALSE TRUE 1 0 -1 -2 -3 POLICY");
+
+    /*- Specifies a custom frozen-core policy on a per-element basis. Input
```
JonathonMisiewicz, line 122: Please specify in the docstring that this is experimental and subject to change. The PR looks good to me, otherwise.

tallakahath: Done!

JonathonMisiewicz (approved):

JonathonMisiewicz:
I've authorized the test suite to run. This is our way of confirming that nothing is obviously broken. If everything passes (as I expect it should), there's nothing on your end yet to do. If it fails, give it a quick look and flag Lori or I if you need assistance in identifying the issue.

tallakahath:
@JonathonMisiewicz looks like most things worked except one of the linux builds failed for an issue that I don't think is related to my patch:

```CMake Error at /usr/local/share/cmake-3.24/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python (missing: Python_NumPy_INCLUDE_DIRS Interpreter
  NumPy) (found suitable version "3.8.10", minimum required is "3.8")
```

Not sure how to proceed here.

loriab:
I've restarted it, which has worked before. If it's the same problem as I've seen on Windows, it has to do with the latest NumPy. Haven't investigated why its sporadic on Linux yet though.

tests/dfmp2-ecp/input.dat
```diff
@@ -84,6 +84,17 @@ E, wfn = energy("mp2/def2-svp", return_wfn=True, molecule=sbh3)
...
+set freeze_core policy
+# Our largest element is Z=51, but freeze_core_policy is 0-indexed (H=0, He=1, etc)
+core_policy = [0 for _ in range(51)]
+# So we need to set Z-1 here to set Sb, which is 50
+core_policy[50] = 5
+psi4.set_options({'freeze_core_policy': core_policy})
+E, wfn = energy("mp2/def2-svp", return_wfn=True, molecule=sbh3)
+
+compare_values(wfn.nfrzc(), 5, 1, "Number of frozen orbitals with freeze_core_policy = {H: 0, Sb: 2}")
+compare_values(sum(wfn.doccpi()), 13, 1, "Number of occupied orbitals with freeze_core_policy = {H: 0, Sb: 2}")
```
loriab, line 96: Are the `Sb: 2` in the comment out-of-date?

tallakahath: Good catch! Yes -- fixed

PR #2663, opened 2022-08-08
dfocc tensor syntax update

loriab:
## Description
`new SharedSomething` -> `make_shared` in DFOCC.

**REVIEW ONLY SECOND COMMIT** (First commit is #2653)

This is PR No. 2 in the mega-dfocc-remp series.

## Todos
- [x] This involves no substantive changes, only syntax. So no review or cursory review fine. It will clear 10k lines of clutter from further PRs.

## Status
- [x] Ready for review
- [x] Ready for merge

Issue #1764, opened 2019-12-02
Segfault for large DF-MP3 calculation


devinamatthews:
A DF-MP3 calculation of C36H38/cc-pVDZ (36ene) fails with a segfault in the OCC module. See attached output file.
[output.txt](https://github.com/psi4/psi4/files/3913671/output.txt)

hokru:
possibly related to https://github.com/psi4/psi4/issues/1679. Although it should fit in the `int` variable raised in that issue. (i think)

Is there any other output? `stderr` captured by slurm?
What is the last content of `timer.dat`?

JonathonMisiewicz:
Technically, the problem is in the DFOCC module, not the OCC module.

Thoughts, @bozkaya? The part that grabs my attention is the line `Memory requirement for CC contractions: -49826.57 MB`. Somehow, I doubt that negative memory is accurate.

devinamatthews:
@hokru I can't find a `timer.dat` file in the submission or scratch directory (although I see it for other jobs that completed normally). All of the output is in the file attached above. I tried again with 700G memory and the symptoms were the same.

Contents of the scratch directory:
```
psi.22173.276
psi.22173.277
psi.22173.35
psi.22173.97
stdout.default.22173.180.npy
```

hokru:
Thanks. Odd having no `timer.dat` at all. Hard to test these large calculations ourselves.

@JonathonMisiewicz I've seen these overflowing numbers before without crashes. But could still point in the right direction. (mentioned here https://github.com/psi4/psi4/issues/898 )

devinamatthews:
Since I need these numbers for a paper I may be motivated enough to build a debug version and dig in deeper. Or, is there a prebuilt debug version somewhere?

loriab:
No, there's no prebuilt debug. You'll probably want to build from master, not 1.3.x. I don't see that it could cause a problem, but you may as well use a -jkfit for the scf, not -ri.

devinamatthews:
I had used -RI for the SCF because I am comparing against my dumb DF implementation in CFOUR that only uses a single fitting basis; I can try with -JKFIT.

JonathonMisiewicz:
[...]

PR #2669, opened 2022-08-09
dfocc: coupled DIIS, long int, and testing

### LINKED ISSUE #2215: Coupled DIIS for `dfocc`
The `dfocc` module has had problems converging MOs tightly. My experience with orbital optimization tells me the problem can be cured by coupling the DIIS iterations for orbital steps and amplitude steps, as is done in the `dct` module and in the `occ` module. There should be an option (I'd suggest making this the default option) to couple the DIIS steps, so examples like the following converge:

```
molecule {
1 2
units bohr
O 0.000000000000 0.000000000000 -0.143225816552
H 0.000000000000 1.638036840407 1.136548822547
H 0.000000000000 -1.638036840407 1.136548822547
}

set reference uhf 
set basis sto-3g
set MAX_MOGRAD_CONVERGENCE 8
set e_convergence 10

energy('omp2')
```

loriab:
## Description
This is the make-dfocc-work-reliably and bring-dfocc-testing-to-normal-level step of mega-dfocc-remp. All the former by @behnle.

~**REVIEW COMMIT 3 ONLY** I'll get the first two merged and out of the way shortly.~

This is PR No. 3 in the mega-dfocc-remp series.

## Todos
- [x] no new methods
- [x] coupled DIIS ported from OCC! fixes #2215. now dfocc can be converged tightly enough that 5-pt findif gradient matches analytic gradient
- [x] potential integer overflows fixed which may lead to wrong memory demand estimates for large molecules
- [x] stdsuite testing added for plain RHF DF/CD a-ccsd(t) energies and DF/CD energies and DF gradients for OO methods omp2, omp2.5, omp3, & olccd. attendantly, qcvar handling in the managers brought up to modern standards, with most saving on wfn, and freebie methods saved and checked
- [x] update run_dfocc in driver to dict/director syntax (works b/c `cc_lambda_` now set in dfocc.cc) and uniform P::e/wfn handling of qcvars
- [x] tightened convergence so that default conditions match values to 1e-6
  - [x] tightened base rms_mograd_convergence from 4 to 6.5 for oo methods (matches occ, which had to be tightened from 6 to 6.5)
  - [x] tightened r_convergence to same on the basis of e_conv for non-oo methods
  - [x] added an extra check that difference in orbitals at convergence not too big a step

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
Please request review from me once both #2663 is in and all merge conflicts are resolved.

psi4/src/psi4/dfocc/dfocc.cc
```diff
@@ -107,14 +107,27 @@ void DFOCC::common_init() {
...
+    }
+    else if (orb_opt_ == "TRUE" && cc_lambda_ == "FALSE") {
+        cc_lambda_ = "TRUE";
+        outfile->Printf("\tWarning: cc_lambda option is changed to TRUE\n");
+    }
+    else if (wfn_type_ == "DF-CCSD(AT)" && cc_lambda_ == "FALSE") {
+        cc_lambda_ = "TRUE";
+        outfile->Printf("\tWarning: cc_lambda option is changed to TRUE\n");
+    }
+
```
JonathonMisiewicz, line 122:
```suggestion
    if ((dertype == "FIRST" || orb_opt_ == "TRUE" || wfn_type_ == "DF-CCSD(AT)") && cc_lambda_ == "FALSE") {
        cc_lambda_ = "TRUE";
        outfile->Printf("\tWarning: cc_lambda option is changed to TRUE\n");
    }

```

psi4/src/psi4/dfocc/t2_2nd_gen.cc
```diff
@@ -85,6 +85,13 @@ void DFOCC::t2_2nd_gen() {
         // Denom
         Tnew->apply_denom_chem(nfrzc, noccA, FockA);
 
+        // DIIS
+        RT2 = SharedTensor2d(new Tensor2d("RT2_2 (IA|JB)", naoccA, navirA, naoccA, navirA));
```
JonathonMisiewicz, line 89: Given that we just had a PR to move to `make_shared`, could we get rid of the `new` introduced by this PR?

psi4/src/psi4/dfocc/dfocc.h
```diff
@@ -585,10 +588,11 @@ class DFOCC : public Wavefunction {
 
     // DIIS
     std::shared_ptr<DIISManager> ccsdDiisManager;
-    std::shared_ptr<DIISManager> ccsdDiisManagerAA;
-    std::shared_ptr<DIISManager> ccsdDiisManagerBB;
-    std::shared_ptr<DIISManager> ccsdDiisManagerAB;
+//    std::shared_ptr<DIISManager> ccsdDiisManagerAA; // What for?? DIISManager can handle arbitrary numbers of quantities
+//    std::shared_ptr<DIISManager> ccsdDiisManagerBB; // so there is no need for three managers except one wants to extrapolate
+//    std::shared_ptr<DIISManager> ccsdDiisManagerAB; // aa, ab and bb sepatarely... These  are used nowhere, anyway.
```
JonathonMisiewicz, line 593:
```suggestion
```

JonathonMisiewicz (commented):
Mostly, this is great. But using `new` is so old...

loriab:
All fixed up after Jonathon round 1 review.

JonathonMisiewicz (approved):

susilehtola:
`long int`? Why not just `size_t` which is **quite literally** the size type.

loriab:
> long int? Why not just size_t which is quite literally the size type.

No reason, I guess. Some modules used `long int` before we went on a `size_t` fest a few years ago. I can switch them (and update the printouts from M*B to G*B) if folks concur. It would be shorter.

behnle:
> `long int`? Why not just `size_t` which is **quite literally** the size type.

Sure, `size_t` would work equally well. Having a Fortran background, i chose `long long int` as it is guaranteed to have 64 bits on any platform. Feel free to change.

andysim:
Just be super careful if there are any subtractions; the result of subtracting two unsigned quantities is itself unsigned and, if negative, it'll wrap around and give garbage.  Therefore things like `if (A - B > tol)` should be `if (A > B + tol)`.  Probably not relevant for this case, but keep in mind that OpenMP loop iteration variables must be signed.

hokru:
Ideally, I suppose, the variables should become `size_t` from the start. As already touched upon here https://github.com/psi4/psi4/issues/1764#issuecomment-562739051

hokru (approved):

loriab:
After wading into `size_t` guidance, the dangers of mixed signed/unsigned, and the deeper troubles of #1764 that need changes to the tensor classes, not just the dfocc.h `int`s, and not just the memory estimates, I think the `long long int` serves nicely for now.

zachglick (approved):
I'm not familiar with dfocc, but LGTM

PR #2682, opened 2022-08-13
Refactor Incremental Fock Builds in Chain-of-Spheres Exchange

andyj10224:
## Description
Through tests of the `COSX` code, we have discovered numerical instabilities in larger systems, leading to slow convergence for later iterations. We discovered that the version of incremental Fock currently in `COSX` performs `incfock` every iteration, and does not include stability features included in the Incremental Fock PR from #2155 (i.e. Shutting it off every few iterations (resets) as well as shutting it off completely past a certain convergence of the density matrix. This PR refactors the inc. Fock framework in the `DFJCOSK` class with the inc. Fock framework from #2155.

## Todos

- [x] Replace old inc. fock. code from `DFJCOSK` PR with the one from #2155
- [x] Move incfock logic from DirectJK to JK
- [x] Cleans up incfock logic by removing extraneous variables like `initial_iteration_`

## Questions
- [x] What is the best way to handle JK classes that should never use `incfock` (i.e. `MemDFJK`, `CDJK`, etc.) What I currently do is to just set incfock to false in those cases no matter what. Is there a better way?
- [x] Should DirectJK use `incfock` by default or not?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

Example of incfock use in with refactor (Hexane/jun-cc-pVDZ):
Iterations in which inc. Fock is performed has `INCFOCK` appended to the end.
```
==> Iterations <==

   @RHF iter SAD:  -234.79902648013046   -2.34799e+02   0.00000e+00
   @RHF iter   1:  -235.11334205670366   -3.14316e-01   3.59965e-03 DIIS/ADIIS
   @RHF iter   2:  -235.33614986245553   -2.22808e-01   1.58174e-03 DIIS/ADIIS/INCFOCK
   @RHF iter   3:  -235.38008939725930   -4.39395e-02   2.83807e-04 DIIS/ADIIS/INCFOCK
   @RHF iter   4:  -235.38291465805983   -2.82526e-03   4.64337e-05 DIIS/INCFOCK
   @RHF iter   5:  -235.38297663987922   -6.19818e-05   1.50376e-05 DIIS/INCFOCK
   @RHF iter   6:  -235.38298211805841   -5.47818e-06   2.18669e-06 DIIS
   @RHF iter   7:  -235.38298224217732   -1.24119e-07   6.25234e-07 DIIS
  Energy and wave function converged with early screening.
  Performing final iteration with tighter screening.

   @RHF iter   8:  -235.38489215328417   -1.90991e-03   8.02480e-05 DIIS
  Energy and wave function converged.
```


susilehtola:
Yup, creeping numerical noise. What was the behavior before this change?

andyj10224:
> Yup, creeping numerical noise. What was the behavior before this change?

Before the change, the SCF took 40 iterations on a larger system. After the change, it only takes 10.

zachglick:
Yeah, for larger / more diffuse basis functions, it's often hard to converge w/ incremental fock. I've already mentioned this in the [docs](https://psicode.org/psi4manual/master/scf.html#cosx-exchange).

In general, incremental fock gets you only a very small speedup with COSX. This is because the last SCF iteration (performed on a large grid) is much more expensive than the other iterations (performed on a small grid), and the last iteration can't be done with a difference density matrix. For this reason, I generally set the `COSX_INCFOCK` option to `FALSE` when I use the JK algorithm, and I had meant to make this the default behavior.

andyj10224:
> Yeah, for larger / more diffuse basis functions, it's often hard to converge w/ incremental fock. I've already mentioned this in the [docs](https://psicode.org/psi4manual/master/scf.html#cosx-exchange).
> 
> In general, incremental fock gets you only a very small speedup with COSX. This is because the last SCF iteration (performed on a large grid) is much more expensive than the other iterations (performed on a small grid), and the last iteration can't be done with a difference density matrix. For this reason, I generally set the `COSX_INCFOCK` option to `FALSE` when I use the JK algorithm, and I had meant to make this the default behavior.

In his timings of `COSX`, @davpoolechem found discrepancies between timings (on larger systems) for doing incfock every iteration (current policy in COSX) and not doing incfock every iteration (this PR), so I believe that `incfock` being ON vs OFF would lead to significant speedups, at least for larger systems.

JonathonMisiewicz:
I'm slightly confused. So COSX and Direct have different "versions" of incremental Fock build, and this PR has them use the same incremental Fock code?

andyj10224:
> I'm slightly confused. So COSX and Direct have different "versions" of incremental Fock build, and this PR has them use the same incremental Fock code?

Yes this is correct

davpoolechem:
> > I'm slightly confused. So COSX and Direct have different "versions" of incremental Fock build, and this PR has them use the same incremental Fock code?
> 
> Yes this is correct

Yea, what Andy said is correct. I've been doing some benchmarking of different J build methods with the COSX method. When doing benchmarks on the DFJCOSK code currently present in Psi4, it often to take a notably higher number of iterations then expected to converge many of the calculations, especially when diffuse functions are in the basis set. After modifying the COSX code to use the "Direct" version of Incfock, the number of SCF iterations needs to converge goes down noticeably. It was in these timing benchmarks that we found the different Incfock implementation in COSX, and this PR is meant to remedy that.

JonathonMisiewicz:
I can start a review once @andyj10224 fixes merge conflicts.

andyj10224:
> I can start a review once @andyj10224 fixes merge conflicts.

Done

psi4/src/psi4/libfock/jk.cc
```diff
@@ -145,6 +155,9 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
         if (options["DF_INTS_NUM_THREADS"].has_changed())
             jk->set_df_ints_num_threads(options.get_int("DF_INTS_NUM_THREADS"));
 
+        // QUESTION: Should Direct JK use incfock by default or not?
+        // if (!options["INCFOCK"].has_changed()) jk->set_incfock(true);
```
davpoolechem, line 159: I noticed your question here, and I think there are two points to be made about this:

1. For a lot of more simplistic IncFock implementations, IncFock acts as a double-edged sword - while basic IncFock can greatly speed up SCF iterations close to convergence in conjunction with density screening, it can also notably increase the number of SCF iterations needed to converge in the first place. However, the IncFock implementation in Psi4 has some nice bells and whistles - particularly, using the full matrices every couple of iterations and, more importantly, disabling IncFock entirely when a certain convergence point has been reached. These bells and whistles make Psi4's IncFock much more stable and reliable; thus, IncFock isn't nearly the same double-edged sword that it often can be.
2. IncFock is especially powerful in conjunction with Density screening, since Density screening becomes much more prevelant in later SCF iterations when the difference density matrix, which approaches the zero matrix as the SCF procedure reaches convergence, is used. Thus, in conjunction with the first point, it could make the most sense to tie the default status of IncFock to the type of screening used - true if density screening is used, and false otherwise. This obviously raises the question of what to do with JK algos that have density screening hard-baked into them, such as LinK. In those cases, if we are taking the approach outlined in this comment, it would likely make the most sense and be most consistent if we enabled IncFock by default for those methods.

Those are my thoughts on the matter.

doc/sphinxman/source/scf.rst
```diff
@@ -782,7 +782,7 @@ a cutoff for the value of basis functions at grid points. This keyword is
 used to determine the radial extent of the each basis shell, and it is the
 COSX analogue to |scf__dft_basis_tolerance|.
 
-The |scf__cosx_incfock| keyword (defaults to ``false``) increases performance
+The |scf__incfock| keyword (defaults to ``true``) increases performance
```
davpoolechem, line 785: Now that scf_cosx_incfock is being removed as a keyword, I wonder how much sense it makes to keep a full paragraph of discussion surrounding scf_incfock in the COSX section. Maybe a small comment about scf_incfock can be included in the previous paragraph instead?

Also, assuming we do keep this paragraph of discussion, we would need to remove the final sentence of it entirely, as this PR switches IncFock to be true by default for COSX.

davpoolechem:
Just finished a passthrough, and it looks pretty good! I have a few very small comments, although more addressing side things than the actual code itself.

doc/sphinxman/source/scf.rst
JonathonMisiewicz: Agreed. There should be a paragraph of `incfock` discussion in general, mentioning that this is only allowed for algorithms X, Y, Z. Unless specific algorithms have special `incfock` options, no need to discuss in general.

psi4/src/export_fock.cc
```diff
@@ -95,7 +95,9 @@ void export_fock(py::module &m) {
         .def("wK", &JK::wK, py::return_value_policy::reference_internal)
         .def("D", &JK::D, py::return_value_policy::reference_internal)
         .def("computed_shells_per_iter", &JK::computed_shells_per_iter, "Array containing the number of ERI shell quartets computed (not screened out) during each compute call.")
-        .def("print_header", &JK::print_header, "docstring");
+        .def("print_header", &JK::print_header, "docstring")
+        .def("incfock_last_iter", &JK::incfock_last_iter, "Was the last Fock build incremental?")
+        .def("reset_incfock", &JK::reset_incfock, "Reset the incfock iteration number (0 means no incfock)");
```
JonathonMisiewicz, line 100: According to `jk.h`, `incfock_count_` means "The number of times INCFOCK has been performed (includes resets)". If so, why are we ever resetting this? Does 0 mean no incfock has ever been performed, or that we should not perform incfock in future?

JonathonMisiewicz (commented):
The PR seems good to me, but I want clarification on what `incfock_count_` represents before I approve.

psi4/src/export_fock.cc
andyj10224: An `incfock_count_` of 0 means that this is the first SCF iteration, so no incfock should be performed. In the context of `COSX`, the reason why this variable is reset to 0 is to avoid incremental Fock builds with the last iteration (where the COSK algorithm is performed on a larger grid). Incfock is NEVER performed at iteration 0.

JonathonMisiewicz: Okay. Then I think it would be clearer if instead of setting `incfock_iter_` to 0 (changing the meaning of the keyword), you have some dedicated flag to disable incfock on specific iterations.

psi4/src/psi4/libfock/jk.cc
andyj10224: For now, I will enable `incfock` for DirectJK only if there is density screening, and false otherwise. For DFJCOSK, `incfock` will be `true` by default.

psi4/src/export_fock.cc
andyj10224: Okay I can do that

andyj10224: Done, I added a new keyword called `disable_incfock_`

doc/sphinxman/source/scf.rst
andyj10224: In the `incfock` section, I explained which algorithms would have `incfock` on by default

JonathonMisiewicz:
Requesting review from Zach to make sure the default handling is sensible, given #2689.

zachglick (commented):
Good work Andy. I like the intent of this PR, which is to consolidate incremental Fock code between different derived `JK` classes. I think that if the incremental Fock feature is to be promoted to the parent `JK` class, it is important that the implementation be as efficient and maintainable as possible. In that respect, I think there are some general improvements to be made:

The way you've structured incremental Fock keeps eight lists of matrices attached to each `JK` object: copies of the previous `J`, `K`, `wK`, and `D` matrices, and differences between the last two iterations `J`, `K`, `wK`, `D`. In the interest of efficiency, I think it would be best to not store these matrices. Incremental Fock as implemented in the `DFJCOSK` class is a little better about this. That code adds the difference in `J`/`K`/`wK` to the appropriate buffer, which is not zeroed out between SCF iterations. This requires keeping only one list of matrices, the previous iteration `D`, which in my opinion is cleaner. (I have some thoughts on getting around storing the previous `D` matrix too.)

Speaking more broadly about the design of the `JK` class, one valuable aspect of the `JK` class is that it is stateless and decoupled from the SCF procedure. `JK` classes have a well-defined [interface](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/jk.h#L175-L209). The variables like `incfock_count_` that you've proposed adding to the `JK` class break this statelessness, and my opinion that kind of logic should be in the SCF driver, not the `JK` object.

Let me know if any of these thoughts are unclear.

andyj10224:
> Good work Andy. I like the intent of this PR, which is to consolidate incremental Fock code between different derived `JK` classes. I think that if the incremental Fock feature is to be promoted to the parent `JK` class, it is important that the implementation be as efficient and maintainable as possible. In that respect, I think there are some general improvements to be made:
> 
> The way you've structured incremental Fock keeps eight lists of matrices attached to each `JK` object: copies of the previous `J`, `K`, `wK`, and `D` matrices, and differences between the last two iterations `J`, `K`, `wK`, `D`. In the interest of efficiency, I think it would be best to not store these matrices. Incremental Fock as implemented in the `DFJCOSK` class is a little better about this. That code adds the difference in `J`/`K`/`wK` to the appropriate buffer, which is not zeroed out between SCF iterations. This requires keeping only one list of matrices, the previous iteration `D`, which in my opinion is cleaner. (I have some thoughts on getting around storing the previous `D` matrix too.)
> 
> Speaking more broadly about the design of the `JK` class, one valuable aspect of the `JK` class is that it is stateless and decoupled from the SCF procedure. `JK` classes have a well-defined [interface](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/jk.h#L175-L209). The variables like `incfock_count_` that you've proposed adding to the `JK` class break this statelessness, and my opinion that kind of logic should be in the SCF driver, not the `JK` object.
> 
> Let me know if any of these thoughts are unclear.

I agree with your points. However, making that change would require a refactor of `DirectJK` as well, as well as an overhaul of the previous `INCFOCK` code that exists in the Psi4 code. For the sake of PR brevity, I suggest punting those changes down to a future PR. The scope of this PR is to fix the numerical instabilities of `incfock` in COSX. Per @davpoolechem, using `incfock` on COSX indeed leads to significant time savings on larger systems, even if it is always OFF by default for the last iteration.

JonathonMisiewicz:
> I agree with your points. However, making that change would require a refactor of `DirectJK` (which zero out the J/K matrices at the beginning of every iteration) as well, as well as an overhaul of the previous `INCFOCK` code that exists in the Psi4 code. For the sake of PR brevity, I suggest punting those changes down to a future PR.

Is this a future PR you'll commit to doing, or a future PR you'll leave to somebody else?

andyj10224:
> > I agree with your points. However, making that change would require a refactor of `DirectJK` (which zero out the J/K matrices at the beginning of every iteration) as well, as well as an overhaul of the previous `INCFOCK` code that exists in the Psi4 code. For the sake of PR brevity, I suggest punting those changes down to a future PR.
> 
> Is this a future PR you'll commit to doing, or a future PR you'll leave to somebody else?

I will commit to doing it. I will likely open it up within the next few days.

JonathonMisiewicz:
I can accept that. Good with you, @zachglick?

zachglick:
Yeah, it's perfectly fine to split things up into multiple PRs. One sensible order I can come up with is:

1. Edit the incremental Fock in the derived `DirectJK` class so that it is efficient and maintainable (no SCF iteration counts associated with the object, minimal number of matrices cached on the object, etc.)
2. Move this improved incremental Fock to the parent `JK` class.

In a previous comment, I suggested doing away with the cached `J`/`K`/`wK`/`dJ`/`dK`/`dwK` matrices by changing the behavior of the `JK` object so that the buffers aren't cleared between SCF iterations. If you find that difficult to do, I think an acceptable compromise would be to retain those cached matrices, but initialize them in the SCF driver (rather than make them a property of the JK object).

JonathonMisiewicz (dismissed):

davpoolechem:
I wanted to comment on this PR to bring attention to it. I am working on the pilot CompositeJK implementation combining DFJCOSK and DFJLinK; and I believe it may be necessary to add this PR to Psi4 before the pilot CompositeJK subclass can be added, since this PR adds a unified Incfock framework for DFJCOSK and DFJLinK.

Technically, I could bundle the stuff in this PR with the second CompositeJK PR, but considering the previous discussion on this PR, I doubt that is something that we would want to do.

JonathonMisiewicz:
Agreed that bundling this PR into the next JK one is a bad idea.

@zachglick, thoughts? I strongly preferred your plan before, but it wasn't something I was going to insist on.

davpoolechem:
> @zachglick, thoughts? I strongly preferred your plan before, but it wasn't something I was going to insist on.

I'll drop a comment on this matter for for an extra perspective. The original plan is quite fine, and many good points were made regarding improvements to efficiency within the Incfock routine. However, the original plan proposed completes the unified Incfock formalism over two PRs, with the standardization of the Incfock process really happening in Step 2. The CompositeJK pilot implementation (i.e., the unification of DFJCOSK and DFJLinK + CompositeJK front end) really only needs Step 2 to proceed, while Step 1 of the proposed Incfock work could likely be done independently from CompositeJK. Basically, my thought is that, by switching Steps 1 and 2 of the original proposed plan for Incfock, that would allow the second CompositeJK PR to be put into Psi4 more quickly and allow parallel development of CompositeJK and the JK Incfock formalism.

I am overall fine with either ordering for Incfock PR steps, but I figured I would at least bring this up.

davpoolechem:
> > @zachglick, thoughts? I strongly preferred your plan before, but it wasn't something I was going to insist on.
> 
> I'll drop a comment on this matter for for an extra perspective. The original plan is quite fine, and many good points were made regarding improvements to efficiency within the Incfock routine. However, the original plan proposed completes the unified Incfock formalism over two PRs, and the impression I get is that the standardization of the Incfock process is really happening in Step 2. The CompositeJK pilot implementation (i.e., the unification of DFJCOSK and DFJLinK + CompositeJK front end) really only needs Step 2 to proceed, then, while Step 1 of the proposed Incfock work (i.e., the improvement of the Incfock process itself) could likely be done independently from CompositeJK. Basically, my thought is that, by switching Steps 1 and 2 of the original proposed plan for Incfock, that would allow the second CompositeJK PR to be put into Psi4 more quickly and allow parallel development of CompositeJK and the JK Incfock formalism.
> 
> I am overall fine with either ordering for Incfock PR steps, but I figured I would at least bring this up.

Thinking about this a little more, one may be able to do the Incfock development parallel to CompositeJK in the ordering suggested in Zach's original plan. Essentially, one would modify DFJLinK to use DFJCOSK's Incfock scheme, which gives the two a unified Incfock among themselves and would allow for continuation of CompositeJK. Then, the same Incfock would be added to DirectJK. And finally, the Incfock could be added to the JK class. This adds an extra PR to the list (DFJLinK Incfock modification), but both maintains Zach's original PR ordering and allows for parallel development of CompositeJK with Incfock modifications.

JonathonMisiewicz:
It would take me at least an hour to figure out what the state of JK is, with all these PRs. I concede that your work may have changed Zach's original reasoning. The only strong opinion I have is that I'd appreciate if somebody could explain all the moving parts. (Probably not on this PR - the discussion is already convoluted enough.)

davpoolechem:
> It would take me at least an hour and probably longer to figure out what the state of JK is, with all these PRs. I concede that your work may have changed Zach's original reasoning. The only strong opinion I have is that I'd appreciate if somebody could explain all the moving parts. (Probably not on this PR - the discussion is already convoluted enough.)

Yeah, there's about a gazillion moving parts in the JK class right now admittedly. Maybe this is something we could discuss in more detail at the Emory meeting or even at Psicon. I'd be happy to type something up for you either here, or send you something on Slack on the dev channel or whatnot.

davpoolechem:
Taking another look at this PR, the IncFock used here, uses a large number of matrix variables - separate matrices for the previous, current, and difference matrices for all of D, J, K, and wK. This particular IncFock formalism was replaced in the PR line of https://github.com/psi4/psi4/pull/2792, https://github.com/psi4/psi4/pull/2808, and https://github.com/psi4/psi4/pull/2816. So before this PR gets merged in, the IncFock formalism moved into the JK class in this PR needs to be updated to match.

PR #2678, opened 2022-08-12
Remove commented out code from mcscf

TiborGY:
## Description
mcscf::SCF::energy(...) has a block of commented out code for TCSCF references.
Removing it also allows the removal of unused `#include`s and an `extern FILE* outfile` declaration.
This is another shard of the https://github.com/psi4/psi4/pull/2642 mega-PR that can be merged independently.

## Todos
- [X] Unused code is removed from `psi4/src/psi4/mcscf/scf_energy.cc`

## Status
- [X] Ready for review
- [X] Ready for merge

PR #2642, opened 2022-07-23
Global diagonalizer cleanup

TiborGY:
## Description
This veritable scattercannon of a PR needs to be reworked into multiple PRs. **Not for merging.**

## Todos

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist
- [x] No new features
- [x] Almost the entire test suite passes, sans MRCC due to unrelated issues. Some runtime-loaded plugins are not covered.
```
========================================================================================================== short test summary info ==========================================================================================================
> FAILED ../tests/mrcc/ccsd_t_/test_input.py::test_mrcc_ccsd_t_ - AssertionError: Using the `local_options` keyword argument is deprecated in favor of using `task_config`, in version 0.30.0 it will stop working.
> FAILED ../tests/pytests/test_composite.py::test_allen_focal_point - NameError: name 'psi4' is not defined
> FAILED ../tests/mrcc/optfreq/test_input.py::test_mrcc_optfreq - AssertionError: Using the `local_options` keyword argument is deprecated in favor of using `task_config`, in version 0.30.0 it will stop working.
> =========================================================================== 3 failed, 3726 passed, 173 skipped, 219 xfailed, 7068 warnings in 2238.06s (0:37:18) ============================================================================

```
## Status
- [ ] Ready for review
- [ ] Ready for merge

PR #2686, opened 2022-08-15
Replace sq_rsp(...) with a new DSYEV wrapper, part 1

TiborGY:
## Description
`sq_rsp(...)` is an ancient diagonalizer function from Psi3 with many issues. While its innards could be improved, the function signature is horrid. It should be deprecated and replaced with something better.
Stemming from the awful interface, users of `sq_rsp(...)` never check if the diagonalization failed, because they _cannot_. This is now rectified by adding failure checks wherever the new DSYEV wrappers are called.
Checks against non-square matrices are also added where there is both a row and column count at the call site.

Some of the call sites would be dfocc territory, this PR does not include them to avoid conflicts with the dfocc overhaul.
This is another shard of the https://github.com/psi4/psi4/pull/2642 mega-PR that can be merged after PR #2678 and #2738 are in.

## Todos

- [x] Implement two new wrapper functions around DSYEV to replace `sq_rsp(...)` with. The new wrappers have a much cleaner interface, and no longer require the allocation of an eigenvector array if the caller only needs eigenvalues. They also no longer swallow the return value of DSYEV, in fact their return value is marked as `[[nodiscard]]`. In short, checking for diagonalization failure went from impossible to mandatory.
Please note that unlike `sq_rsp(...)`, the new functions do not take a "tolarence" value, but this is actually a null change as that parameter has been set-but-unused in `sq_rsp(...)` since ~forever ago when `sq_rsp(...)` switched to calling DSYEV internally.
- [x] Replace `sq_rsp(...)` calls in `mcscf` (`mcscf::SCF::energy` and `mcscf::MatrixBase::diagonalize`) with new wrapper. Add checks for diagonalization failure and guard against non-square matrices.
- [x] Replace `sq_rsp(...)` calls in `detci` (`detci/h0block.cc`, `detci/sem.cc` and `detci/sem_test.cc`) with new wrapper. Add checks for diagonalization failure.
- [x] Replace `sq_rsp(...)` call in `libmints/matrix.cc` (`Matrix::diagonalize`) with new wrapper. Add checks for diagonalization failure. Add a sanity checks for non-square matrices and illegal values of "diagonalization order".
- [x] Replace `sq_rsp(...)` calls in `ccenergy` (`ccenergy/d1diag.cc`, `ccenergy/d2diag.cc` and `ccenergy/new_d1diag.cc`) with new wrapper. Add checks for diagonalization failure. This allows the removal of the eigenvector array, and the code that allocates/deallocates it.
- [x] Replace `sq_rsp(...)` calls in the `libqt` Davidson solver with new wrapper. Add checks for diagonalization failure.
- [x] Replace `sq_rsp(...)` call in the RHF and ROHF stability checks with new wrapper. Add checks for diagonalization failure.
- [x] Replace `sq_rsp(...)` calls in `libsapt_solver/sapt2.cc` with new wrapper. Add checks for diagonalization failure.

## Questions
- [x] Is `libciomr` a good place for the new wrappers to live in? The one they are replacing is there, so it seemed like as good of a place as any.

## Checklist
- [x] No new features
- [x] 503/503 ctests are passing

## Status
- [x] Ready for review
- [x] Ready for merge


psi-rking:
As an old-timer, I would say yes, libciomr is the place.  At least as long as there is still a libciomr, that's where I would look.  This is nice improvement!

ccdensity has a sq_rsp too.  did you miss that one?

TiborGY:
> As an old-timer, I would say yes, libciomr is the place. At least as long as there is still a libciomr, that's where I would look. This is nice improvement!
> 
> ccdensity has a sq_rsp too. did you miss that one?

I ran a search and none of the remaining hits were from ccdensity.

psi-rking:
in cc/ccdensity/ael.cc
This code determines approximate excitation level.  It used to work.
Apparently no longer called.  I'm not sure why.
(I'm not on the very latest psi4, so it's possible it's been excised.

TiborGY:
Indeed it has, by yours truly. See #2672 for the removal. I do not know when the AEL code became defunct, but I removed it because it was never called. The code is preserved in the psi4attic repo.

psi4/src/psi4/libciomr/dsyev_ascending.cc
```diff
@@ -0,0 +1,78 @@
...
+** \param e_vecs = (optional) matrix of eigenvectors (2D row major array, one column for each eigvector)
+**
+** \ingroup CIOMR
+*/
+[[nodiscard]] int DSYEV_ascending(const int N, const double* const* const array, double* e_vals,
+                                  double* const* const e_vecs /* = nullptr*/) {
+    // We need to make a copy of the matrix before diagonalization, because LAPACK overwrites it.
+    // LAPACK also needs the mtx to be flattened to a 1D array, so a copy is inevitable.
+    // The new 1D array will correspond to a column-major array, suitable for LAPACK.
+    double* tmp_matrix = init_array(N * N);
```
JonathonMisiewicz, line 54: Use `std::vector<double> tmp_matrix(N * N)` instead. Any function signature expecting a `double *` will receive `tmp_matrix.data()` instead. This frees us from needing to manage memory manually.

TiborGY: Good idea!

loriab (approved):
Nice cleanup and consolidation, thank you!

TiborGY:
503/503 ctest tests are passing
[ctestlog.txt](https://github.com/psi4/psi4/files/10106557/ctestlog.txt)
If that is judged to be sufficient coverage, this PR is ready for merge.

loriab:
I've just run full ctests and pytests with this branch, and it's clean, so lgtm!

JonathonMisiewicz (approved):

PR #2693, opened 2022-08-18
Remove unused and pointer-argumented overloads of Matrix::diagonalize

TiborGY:
## Description
Some overloads of `Matrix::diagonalize` in `libmints` are never called anywhere, they are deleted.
On the suggestion of @JonathonMisiewicz I have only kept the overloads that do not take pointer arguments and modified the call sites where required.
This is another shard of the https://github.com/psi4/psi4/pull/2642 mega-PR that can be merged independently.

## Todos

- [x] Unused `Matrix::diagonalize` overloads are gone
- [x] `Matrix::diagonalize` no longer takes pointer arguments

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/matrix.h
+++ b/psi4/src/psi4/libmints/matrix.h
@@ -863,16 +863,8 @@ class PSI_API Matrix : public std::enable_shared_from_this<Matrix> {
 
     /// @{
     /// Diagonalizes this, eigvectors and eigvalues must be created by caller.  Only for symmetric matrices.
-    void diagonalize(Matrix* eigvectors, Vector* eigvalues, diagonalize_order nMatz = ascending);
+    void diagonalize(Matrix& eigvectors, Vector& eigvalues, diagonalize_order nMatz = ascending);
     void diagonalize(SharedMatrix& eigvectors, std::shared_ptr<Vector>& eigvalues, diagonalize_order nMatz = ascending);
-    void diagonalize(SharedMatrix& eigvectors, Vector& eigvalues, diagonalize_order nMatz = ascending);
-    /// @}
-
-    /// @{
-    /// Diagonalizes this, applying supplied metric, eigvectors and eigvalues must be created by caller.  Only for
-    /// symmetric matrices.
-    void diagonalize(SharedMatrix& metric, SharedMatrix& eigvectors, std::shared_ptr<Vector>& eigvalues,
-                     diagonalize_order nMatz = ascending);
     /// @}
 
     /// @{
@@ -1096,8 +1088,6 @@ class PSI_API Matrix : public std::enable_shared_from_this<Matrix> {
 
     /// General matrix multiply, saves result to this
     void gemm(bool transa, bool transb, double alpha, const Matrix& a, const Matrix& b, double beta);
-    /// Diagonalize a symmetric matrix. Eigvectors and eigvalues must be created by caller.
-    void diagonalize(Matrix& eigvectors, Vector& eigvalues, int nMatz = 1);
 
     /// @{
     /// Retrieves the i'th irrep
--- a/psi4/src/psi4/dlpno/mp2.cc
+++ b/psi4/src/psi4/dlpno/mp2.cc
@@ -850,7 +850,7 @@ void DLPNOMP2::pno_transform() {
         // Diagonalization of pair density gives PNOs (in basis of the LMO's virtual domain) and PNO occ numbers
         auto X_pno_ij = std::make_shared<Matrix>("eigenvectors", nvir_ij, nvir_ij);
         Vector pno_occ("eigenvalues", nvir_ij);
-        D_ij->diagonalize(X_pno_ij, pno_occ, descending);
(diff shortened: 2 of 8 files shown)


psi4/src/psi4/libmints/matrix.h
```diff
@@ -863,16 +863,8 @@ class PSI_API Matrix : public std::enable_shared_from_this<Matrix> {
 
     /// @{
     /// Diagonalizes this, eigvectors and eigvalues must be created by caller.  Only for symmetric matrices.
-    void diagonalize(Matrix* eigvectors, Vector* eigvalues, diagonalize_order nMatz = ascending);
+    void diagonalize(Matrix& eigvectors, Vector& eigvalues, diagonalize_order nMatz = ascending);
     void diagonalize(SharedMatrix& eigvectors, std::shared_ptr<Vector>& eigvalues, diagonalize_order nMatz = ascending);
-    void diagonalize(SharedMatrix& eigvectors, Vector& eigvalues, diagonalize_order nMatz = ascending);
```
loriab, line 868: pausing while we reconsider doing away with the convenience functions

TiborGY: I guess having a `(SharedMatrix&, Vector&)` overload would not be a big deal, and it _would_ remove the need for most of the call site modifications in this PR. I am OK with both options.

TiborGY: Gentle poke, has a decision been reached? #2686 is mostly done, but stalled until this PR is resolved.

JonathonMisiewicz: I can't recall any discussion on the issue.

loriab: This PR was part of my motivation for bringing up convenience functions in the conf call. Result is that they're going, so this PR is good as-is.

psi4/src/psi4/libmints/matrix.h
```diff
@@ -863,16 +863,8 @@ class PSI_API Matrix : public std::enable_shared_from_this<Matrix> {
...
-    /// @}
-
-    /// @{
-    /// Diagonalizes this, applying supplied metric, eigvectors and eigvalues must be created by caller.  Only for
-    /// symmetric matrices.
-    void diagonalize(SharedMatrix& metric, SharedMatrix& eigvectors, std::shared_ptr<Vector>& eigvalues,
```
loriab, line 874: Am I reading this and the sq_rsp PR right that the diagonalize fn with `metric` parameter are going away?

psi4/src/psi4/libmints/matrix.h
JonathonMisiewicz: I forgot about that. Glad I said "I can't recall" then!

psi4/src/psi4/libmints/matrix.h
TiborGY: Yes? This PR is removing the "diagonalize with metric" fn from `Matrix` as it is never called by anything, and the later PR is not adding it back either, for the same reason.

loriab: Ok, I'm not enthusiastic about removing the option. But 10y is a good span of opportunity for a use to be found. :-)

loriab (approved):

JonathonMisiewicz (approved):

JonathonMisiewicz:
I'm having second thoughts about this PR now. Should we deprecate the old constructors first? This PR will require changes in Forte as well.

TiborGY:
> I'm having second thoughts about this PR now. Should we deprecate the old constructors first? This PR will require changes in Forte as well.

Hmm, indeed while the functions are not `PSI_API`, the entire `Matrix` class _is_. Ouch. This is something I have completely missed, deprecation before removal is something I agree with when it comes to APIs.

The question is, should "deprecate before removal" principle be applied to all member functions/ctors of a 'PSI_API` class? I would assume yes.

loriab:
> I'm having second thoughts about this PR now. Should we deprecate the old constructors first? This PR will require changes in Forte as well.

options as I see them. I do think we should try to get the changes into one psi4 release cycle.

* (n/c) leave convenience fns as-is
* (deprecate) leave convenience fns operable but add deprecation message
* (upgradehelper) leave convenience fn header in place but have it print a message with necessary changes
* (remove) remove convenience fn header and body

JonathonMisiewicz:
I vote deprecate, _just in case_ we change our mind on this decision.

loriab:
> I vote deprecate, just in case we change our mind on this decision.

fine by me

TiborGY:
Agreed on the deprecation.

TiborGY:
> I do think we should try to get the changes into one psi4 release cycle.

Regarding the deprecation message/schedule: do you mean that they should be deprecated now and then removed before 1.7 RC1?

TiborGY:
I propose that this PR could be re-tagged as Psi4 1.8 and kept open after #2738 is merged, as a reminder to delete the functions.

PR #2695, opened 2022-08-19
Density Screening Refactor Part 2: Implementation of shell_significant()

davpoolechem:
## Description
This PR is the second in a series of planned PRs designed to remove density screening from the TwoBodyAOInt object and into the JK object, with PR #2547 being the first such PR. Having density screening available in TwoBodyAOInt runs the risk of applying density screening to algorithms where density screening doesn't make sense. Thus, it would be a good idea to move the logic of density screening to where it is more correctly applied, i.e., the JK object.

The primary purposes of this PR are twofold:
1. First, this PR introduce the shell_significant() framework to the JK class. The shell_significant() framework starts with a shell_significant() virtual function that exists in the base JK class. The shell_significant() function can then be redefined specifically for different JK derived classes as needed. The existence of shell_significant() provides a unified framework for performing screening for any JK method, and it also provides the method by which density screening can be added directly to the relevant JK classes.
2. Second, this PR uses the shell_significant() framework to reimplement screening for certain JK methods. Most significantly, as implied in the first point, the biggest change in this regard was the removal of shell_significant_density() from TwoBodyAOInt into the domain and its reimplementation into the DirectJK shell_significant() definition. With this change, density screening is now the domain of individual JK classes rather than the TwoBodyAOInt class, which was one of the primary goals of this overall refactor PR set in the first place. Additionally, another method that was subject to slight changes from the shell_significant() framework was the Yoshimine PKJK algorithm, in which its test for shell significance was implemented using the shell_significant() framework.

## Todos

- [ ] Implementation of shell_significant() framework in JK class to represent shell significance testing.
- [ ] Removal of shell_significant_density() from TwoBodyAOInt.
- [ ] Reimplementation of density screening in DirectJK via shell_significant framework.
- [ ] Reimplementation of density screening in other JK methods via shell_significant framework.
- [ ] Alteration of other shell quartet screening implementations in other JK algorithms via JK framework.

## Questions
- [ ] The shell_significant framework has not yet been added to DFJCOSK. Should that be done this PR, or added in a later PR? 

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


psi4/src/psi4/libmints/twobody.cc
```diff
@@ -77,7 +77,7 @@ TwoBodyAOInt::TwoBodyAOInt(const IntegralFactory *intsfactory, int deriv)
     else if (screentype == "CSAM")
         screening_type_ = ScreeningType::CSAM;
     else if (screentype == "DENSITY")
```
andyj10224, line 79: Would it make sense to diable setting `screening density` altogether?

psi4/src/psi4/libmints/twobody.h
```diff
@@ -226,6 +224,9 @@ class PSI_API TwoBodyAOInt {
     /// Return the maximum density matrix element per shell pair. Maximum is over density matrices, if multiple set
     double shell_pair_max_density(int M, int N) const;
 
+    /// Return the maximum density matrix element per shell pair for density matrix i
+    double shell_pair_max_density(int i, int M, int N) const;
```
andyj10224, line 228:
```suggestion
    inline double shell_pair_max_density(int i, int M, int N) const; 
```

psi4/src/psi4/libmints/twobody.cc
```diff
@@ -166,38 +166,9 @@ double TwoBodyAOInt::shell_pair_max_density(int M, int N) const {
...
-    }
-
-    // Square of Cauchy-Schwarz Q_MN terms (Eq. 13)
-    double mn_mn = shell_pair_values_[N * nshell_ + M];
-    double rs_rs = shell_pair_values_[S * nshell_ + R];
-
-    // The density screened ERI bound (Eq. 6)
-    return (mn_mn * rs_rs * max_density * max_density >= screening_threshold_squared_);
-}
+double TwoBodyAOInt::shell_pair_max_density(int i, int M, int N) const  {
```
andyj10224, line 169: Should we move `shell_pair_max_density` to JK in this PR or the next? I don't think this belongs in `TwoBodyAOInt`

andyj10224 (commented):
Looks good, I left some comments. Namely, my biggest concern is about removing the `screening density` keyword and moving `update_density` away from `TwoBodyAOInt`. I think it would make more sense to get it all done in one PR, but that is my personal opinion.

davpoolechem:
> Looks good, I left some comments. Namely, my biggest concern is about removing the `screening density` keyword and moving `update_density` away from `TwoBodyAOInt`. I think it would make more sense to get it all done in one PR, but that is my personal opinion.

Thank you very much for the comments! I do appreciate it. As for your points, my thought was to implement both of these changes in future PRs, maybe as a single future PR, maybe as their own separate ones. I think there's a good argument to be made for making a separate option for density screening in this PR, and I'd be quite willing to do that. But, I do believe that removing update_density should be its own PR, as that PR would consist of removing ALL references to the density matrix from TwoBodyAOInt.

psi4/src/psi4/libmints/twobody.cc
davpoolechem: It would be, yes. I have this planned in the future, although it could quite possibly be done in this PR as well.

psi4/src/psi4/libmints/twobody.cc
davpoolechem: My plan was to move this whenever update_density gets moved, which I'm planning on being the next PR.

psi4/src/psi4/libmints/twobody.cc
davpoolechem: The reason this is still in here is to account for cases where density screening is used in the HF step and another form of screening is used through TwoBodyAOInt for a later step in the calculation. Looking through the code, it seems that an example of this is in the DirectJKGrad class, which uses TwoBodyAOInt screening functions in its tests for shell block significance.

This will certainly be removed once density screening is separated out as a keyword, though.

psi4/src/psi4/libfock/PKmanagers.cc
```diff
@@ -1111,7 +1118,7 @@ void PKMgrYoshimine::compute_integrals(bool wK) {
             for (size_t j = 0; j <= i; ++j) {
                 int RR = sh_pairs[j].first;
                 int SS = sh_pairs[j].second;
-                if (tb[thread]->shell_significant(PP, QQ, RR, SS)) {
+                if (shell_significant(PP, QQ, RR, SS)) {
```
loriab, line 1121: so why does this no longer get applied per-thread? (read as naive, not insightful, question)

davpoolechem: The shell_significant function in PKMgrYoshimine is defined to do its shell significance test using the sieve_ variable in the base PKManager class instead of through the TwoBodyAOInt object. Theoretically, shell_significant in PKMgrYoshimine could be redefined to use tb[thread]'s shell_significant, though it would be somewhat clumsy since it would require passing { tb[thread] } as the optional argument for int.

Talking about it here though makes me wonder whether using sieve_ is the best approach for the shell_significant definition, however.

loriab:
`undefined symbol: _ZNK3psi12TwoBodyAOInt22shell_pair_max_densityEiii` in case you hadn't clicked through to it yet

davpoolechem:
> `undefined symbol: _ZNK3psi12TwoBodyAOInt22shell_pair_max_densityEiii` in case you hadn't clicked through to it yet

I did indeed notice that! I think I know the problem, so I will apply the fix in just a little bit.

davpoolechem:
All right, CI issue should be fixed now!

psi4/src/psi4/libfock/PKmanagers.cc
JonathonMisiewicz: The ERISieve class is deprecated in favor of having the `TwoBodyAOInt` object do the sieving. This was originally decided as part of the L2 discussion and has been reaffirmed on Slack just a week or two ago. I was in the process of getting rid of ERISieve when this PR came up.

We'll need to agree about the future of "what class is responsible for sieving" before this PR can be merged in.

JonathonMisiewicz (changes requested):
Blocking this PR until we can resolve the question of "is `TwoBodyAOInt` responsible for sieving, or is `ERISieve`"? Happy to talk about this after ACS Chicago.

JonathonMisiewicz:
FTR, David and I have agreed to pause this PR until after #2682 and #2665 come in.

davpoolechem:
> Blocking this PR until we can resolve the question of "is `TwoBodyAOInt` responsible for sieving, or is `ERISieve`"? Happy to talk about this after ACS Chicago.

To add an update to this, some of us (Jonathan, I, and others) discussed this issue at PsiCon 2022. We came to the conclusion that TwoBodyAOInt would be responsible for sieving, and that ERISieve will be removed from Psi4. As use of ERISieve is seemingly localized to the PKJK algorithm, I will be the one responsible for its removal.

Until then, this PR will likely be further delayed.

andyj10224:
@davpoolechem Since #3098 has been opened, should we close this PR

davpoolechem:
> @davpoolechem Since #3098 has been opened, should we close this PR?

You are correct. We should. And so I will.

PR #2696, opened 2022-08-21
Remove unused function ccdensity::resort_tei() and its dependencies

TiborGY:
## Description
`ccdensity::resort_tei()` appears to be completely unused, it is never called anywhere, it is not `PSI_API` and it is commented out from CMake as well as at the declaration site. Deleted.
There are also a bunch of functions which were only called by `resort_tei()`, or by functions only called from `resort_tei()`. They _were_ enabled in CMake. Deleted.

## Todos

- [x] Unused functions/files are removed from `ccdensity`: `classify`, `distribute`, `file_build`, `idx_error`, `idx_permute` and `resort_tei`

## Status
- [x] Ready for review
- [x] Ready for merge


lothian (commented):
Actually, this code was (is?) part of an old effort to add frozen-core capabilities to the CC gradient code.  I'd prefer to leave it in place for now until I can develop a plan to resurrect the work.

TiborGY:
> Actually, this code was (is?) part of an old effort to add frozen-core capabilities to the CC gradient code. I'd prefer to leave it in place for now until I can develop a plan to resurrect the work.

Fair enough. I have reverted the deletions, commented out the currently unused files in CMake instead, and added a comment above the declaration site pointing to this PR.
Just in case someone else gets the idea of removing it.

hokru (approved):

psi4/src/psi4/cc/ccdensity/ccdensity.cc
```diff
@@ -70,6 +70,8 @@ void onepdm(const struct RHO_Params&);
 void sortone(const struct RHO_Params&);
 void twopdm();
 void energy(const struct RHO_Params&);
+// Note: resort_tei() is to be used by a frozen-core CC gradient implementation in the future.
```
JonathonMisiewicz, line 73:
```suggestion
// Note: resort_tei() may be used by a frozen-core CC gradient implementation in the future.
```

JonathonMisiewicz:
Frozen core gradients are one of the items on my to-do list. (`libtrans` is of course the problem.) `resort_tei` may not be needed. I'll make a note to reach out to TDC when that happens. Maybe at PsiCon. 🙂

JonathonMisiewicz (approved):

loriab (approved):

PR #2699, opened 2022-08-24
Remove confusing dash/minus character from stability analysis text

TiborGY:
## Description
Currently the RHF stability analysis output looks something like this:
```
    Lowest singlet (RHF->RHF) stability eigenvalues:-
       A 0.070763         A 0.085714         A 0.095974         A 0.097368
```
Note the '-' character after the colon. What does that mean? I think it means nothing at all, but it is in a context where a stray '-' can have a lot of meaning.

This PR removes the spurious and potentially confusing '-' from the code and reference outputs.

## Todos
- [X] Do not print a spurious '-' character in RHF stability analysis

## Questions
- [ ] It must be just a typo, right?

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libscf_solver/rhf.cc
+++ b/psi4/src/psi4/libscf_solver/rhf.cc
@@ -980,9 +980,9 @@ bool RHF::stability_analysis() {
             free(evals);
         }
 
-        outfile->Printf("    Lowest singlet (RHF->RHF) stability eigenvalues:-\n");
+        outfile->Printf("    Lowest singlet (RHF->RHF) stability eigenvalues:\n");
         print_stability_analysis(singlet_eval_sym);
-        outfile->Printf("    Lowest triplet (RHF->UHF) stability eigenvalues:-\n");
+        outfile->Printf("    Lowest triplet (RHF->UHF) stability eigenvalues:\n");
         print_stability_analysis(triplet_eval_sym);
         psio_->close(PSIF_LIBTRANS_DPD, 1);
     }
--- a/tests/scf-coverage/output.ref
+++ b/tests/scf-coverage/output.ref
@@ -242,7 +242,7 @@ compare_matrices(ref_nos, scf_wfn.Ca(), 8, "Water Anion UHF NOS")
 	First half integral transformation complete.
 	Starting second half-transformation.
 	Two-electron integral transformation complete.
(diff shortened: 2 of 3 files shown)


loriab:
Looks like they've been around as long as the stability code, https://github.com/psi4/psi4/commit/088cef8a33c4ecad6b59d6c7807e5d0d631ed436 .  I agree they look confusing. Did you get all three files for the different references? Anyone recall if the dash was informative?

andysim:
It's a [colon dash](https://en.wikipedia.org/wiki/Compound_point) (well, it has a more amusing name but, to keep things family friendly, I'll allow you to look that up in the Wiki article).  I was taught that it should be used for lists when I was young, and it wasn't until I moved to the US that I learned that it's *very* outdated.

TiborGY:
> Looks like they've been around as long as the stability code, [088cef8](https://github.com/psi4/psi4/commit/088cef8a33c4ecad6b59d6c7807e5d0d631ed436) . I agree they look confusing. Did you get all three files for the different references? Anyone recall if the dash was informative?

I did a global search&replace on the repo, so unless there is an extra space or something, all instances have been replaced. The dash is not present for ROHF: https://github.com/psi4/psi4/blob/0b26beda13a4ad2569c810684039d74672babdc9/psi4/src/psi4/libscf_solver/rohf.cc#L1279
The UHF code had been rewritten and does not have a line like this.

loriab:
> It's a [colon dash](https://en.wikipedia.org/wiki/Compound_point) (well, it has a more amusing name but, to keep things family friendly, I'll allow you to look that up in the Wiki article). I was taught that it should be used for lists when I was young, and it wasn't until I moved to the US that I learned that it's very outdated.

Oh yeah, that wiki article is familiar,— maybe we've had this conversation at the uhf file, the rohf file, and now the rhf file. 🤦

TiborGY:
Now Azure CI is having some issues...`numpy.core.multiarray failed to import`

loriab:
> Now Azure CI is having some issues...numpy.core.multiarray failed to import

#2702 helps, so you can apply that here for now, though I don't want it as a long-lasting pin.

JonathonMisiewicz (approved):

hokru (approved):

loriab (approved):

Issue #2261, opened 2021-07-29
PSIO_WT_TOCLEN


JonathonMisiewicz:
The following input file from [forums](http://forum.psicode.org/t/error-in-psio-wt-toclen/2234/5) gives a PSIO error `Error in PSIO_WT_TOCLEN()!` in up-to-date master. Reported said they could also reproduce with CCSD, but I haven't tried this myself. Setting `cachelevel 0` was a workaround, but not one that should be necessary.

```memory 5 gb

molecule = {
0 2
Cl
O 1 R1
O 2 R2 1 A
R1=2.03230554
R2=1.20810395
A=115.36876732

}

set {
reference rohf
basis aug-cc-pwcvqz
print_MOs true
print 2
scf_type pk
guess sad
freeze_core -2
}

energy(‘bccd’)
```

TiborGY:
Here is an input that also reproduces it, without taking nearly as much time or disk space:
```
molecule {
-1 1
H        2.3076409400      -0.1392813200      -1.4531401200
 H        2.1805389500       0.4742872200       0.1462030400
 H        3.7756900400      -0.4179406200      -0.4177223000
 H        1.1661712500      -2.7201024200      -0.5546608900
 C        2.7098942000       0.0313565000      -0.5413217000
 O        2.0712525100      -2.8274350400      -0.8968065600
 I       -0.5922072300       0.3756625500       0.1823060400
}

basis jorge{
spherical
****
H     0
S    3   1.00
     12.98511484             0.0310449
      1.95908134             0.2173262
      0.44403649             0.7516287
S    1   1.00
      0.11282695             1.0000000
S    1   1.00
      0.02690395             1.0000000
P    1   1.00
      0.80305095             1.0000000
P    1   1.00
      0.14991095             1.0000000
****
C     0
S    5   1.00
   4949.46032594             0.0019796
    758.82582124             0.0148496
    172.62336223             0.0762545
     47.77742113             0.2872122
     15.15140603             0.7104557
S    2   1.00
      5.39400345             0.7530892
      2.11083628             0.2771972
S    1   1.00
      0.52460137             1.0000000
S    1   1.00
      0.15964728             1.0000000
S    1   1.00
      0.04532328             1.0000000
P    4   1.00
     18.97219923             0.0179023
      4.18721595             0.1114128
      1.21388239             0.3744715
      0.38698643             0.6513515
P    1   1.00
      0.12406489             1.0000000
P    1   1.00
      0.03695489             1.0000000
[...]

PR #2700, opened 2022-08-24
Rework PSIO toclen read/write

TiborGY:
## Description
PSIO toclen errors are extremely cryptic, making them even more difficult to debug,
I have recently ran into another input that produces the error in #2261 so decided to try taking a look.
I have not debugged the actual issue, but made some changes to the function raising the error to hopefully make the error messages more informative.

The `read`/`write`/`lseek` syscalls deposit an OS-defined error code in `errno`, which can be decoded to an OS-provided, human-readable (possibly non-English) error message. This should at least tell the person seeing it vaguely what is wrong, ie. is the disk full, does the file not exist, etc.

The toclen functions have been renovated in general, `const` added where appropriate, common code factored out, etc. Error conditions now properly use `psio_error(...)` and pass the OS-provided message to it.

## Todos

- [x] Error messages from `PSIO::rd_toclen` and `PSIO::wt_toclen` are now more informative
- [x] The new function `decode_errno(int)` is now available for use after system calls to decode `errno`

## Status
- [x] Ready for review
- [x] Ready for merge


hokru:
All relevant error information should end up in the error box, not just `stderr`

TiborGY:
> All relevant error information should end up in the error box, not just `stderr`

OK. Would the inverse be acceptable? ie. nothing going to `stderr`

loriab:
>> All relevant error information should end up in the error box, not just stderr

> OK. Would the inverse be acceptable? ie. nothing going to stderr, only to the error box

Haven't investigated, but my first thought is that error box only would work

TiborGY:
I have reworked this PR to a general renovation of `toclen.cc`. All errors are now handled by calling `psio_error()`, OS-provided messages are decoded by a new fn that may be useful elsewhere later and the toclen fns now first check if the stream is ought to be open and error out early with a more specific error message.

JonathonMisiewicz (approved):

JonathonMisiewicz:
Incidentally, an independent PR of mine just ran into this particular error message. I'm eager to get this in.

loriab (approved):
Thanks for tackling psio!

psi4/src/psi4/libpsio/toclen.cc
```diff
@@ -64,53 +66,67 @@ size_t PSIO::toclen(size_t unit) {
...
+    if(!open_check(unit)) psio_error(unit, PSIO_ERROR_UNOPENED);
+    // Seek to the beginning
+    rewind_toclen(unit);
+    // Read the value
+    size_t len;
+    const auto stream = psio_unit[unit].vol[0].stream;
+    const auto errcod = SYSTEM_READ(stream, (char *)&len, sizeof(size_t));
+    const auto sys_errno = errno;
+    if (errcod != sizeof(size_t)) {
+#ifdef DEBUG
```
lothian, line 100: Is this intended to occur only within the `#ifdef`?  The corresponding code in `wt_toclen()` isn't similarly blocked.

TiborGY: Good point, it is something I was on the fence about.

The `read(...)` system call returns either the number of bytes read, or `-1` if an error occurs. The old code checks `if(bytes_read_expected==bytes_read_actual)` and returns 0 if that is false. This is justified by the original code as the following:

> Note that we do not exit if the read request of the toclen from the file fails. This is because the request may be to an new file for which the toclen has not yet been written. (We allow the user open files with status PSIO_OPEN_OLD even if they don't exist, because sometimes you can't know this in advance.)

So after reading that I chose to preserve the original behavior unless the `#ifdef` is active, but now I am not so sure, as the original behavior fails to guard against actual read errors.

Given how the code in the `#ifdef` block should only trigger on actual read errors it might be best to not only remove the `#ifdef`, but make it an error exit like the other failures.

lothian: Yes, I'm pretty sure I wrote the original version of that comment when I wrote `libpsio` so long ago.  If the current version with the removal of the `#ifdef` runs correctly, then I'm happy to LGTM it.

lothian (approved):

PR #2703, opened 2022-08-24
Chip away at MOInfo (1/???)

JonathonMisiewicz:
## Description
Replace an old-style call to `moinfo_` in `cc` with something more modern: using Matrix technology to bypass the loop over the number of irreps.

## Checklist
- [x] `cc` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/ccenergy/Fmi.cc
+++ b/psi4/src/psi4/cc/ccenergy/Fmi.cc
@@ -34,8 +34,8 @@
 #include <cstdlib>
 #include "psi4/libdpd/dpd.h"
 #include "Params.h"
-#include "MOInfo.h"
 #include "psi4/cc/ccwave.h"
+#include "psi4/libmints/matrix.h"
 
 namespace psi {
 namespace ccenergy {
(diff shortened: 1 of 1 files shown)


lothian (commented):
Are you also zeroing out the diagonal for FMI in the UHF case?  I looks like only the Fmi matrix is cleared there.

JonathonMisiewicz:
> Are you also zeroing out the diagonal for FMI in the UHF case? I looks like only the Fmi matrix is cleared there.

Yes, on lines 75-77. Look for the little patch of green _between_ the red.

lothian:
> > Are you also zeroing out the diagonal for FMI in the UHF case? I looks like only the Fmi matrix is cleared there.
> 
> Yes, on lines 75-77. Look for the little patch of green _between_ the red.

OK, now I see what you're doing.  My only suggestion is that you modify the comment under RHF to read:

// This is because the amplitude update in RHF uses the full residual while the ROHF and UHF updates separate out the diagonal contributions from the Fock matrix [cf. Eqs. (1) and (2) of Stanton et al., J. Chem. Phys. 94, 4334-4345 (1991)].

lothian (approved):

JonathonMisiewicz:
Mostly adopted : I changed a few words to specify that these are true of Psi's implementations, not necessarily all implementations.

Thanks for the review! I will always accept specific journal references.

zachglick (approved):

hokru (approved):

PR #2722, opened 2022-09-19
Allow non-symmetric matrix <-> dpdfile2

JonathonMisiewicz:
## Description
Allow non-symmetric matrix <-> dpdfile2.

The hard part of this PR was figuring out how to initialize a dpd object successfully.

## Dev notes & details

- [x] `Matrix::write_to_dpdfile2` now works on non-totally symmetric matrices
- [x] Several long-deprecated Matrix methods have been removed. They now live in the `linalg` namespace. 

## Checklist
- [x] New test passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -2818,7 +2818,7 @@ void Matrix::write_to_dpdfile2(dpdfile2 *outFile) {
 
     if (outFile->my_irrep != symmetry_) {
         std::stringstream msg;
-        msg << "Symmetry mismatch. Matrix has symmetry " << outFile->my_irrep << " where dpdfile has "
+        msg << "Symmetry mismatch. Matrix has symmetry " << outFile->my_irrep << " whereas dpdfile has "
             << symmetry_ << " symmetry.";
         throw SanityCheckError(msg.str().c_str(), __FILE__, __LINE__);
     }
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -2816,9 +2816,11 @@ void Matrix::write_to_dpdfile2(dpdfile2 *outFile) {
         throw SanityCheckError(msg.str().c_str(), __FILE__, __LINE__);
     }
 
-    if (outFile->my_irrep != 0) {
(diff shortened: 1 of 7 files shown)


psi4/src/psi4/libmints/matrix.cc
```diff
@@ -2816,9 +2816,11 @@ void Matrix::write_to_dpdfile2(dpdfile2 *outFile) {
         throw SanityCheckError(msg.str().c_str(), __FILE__, __LINE__);
     }
 
-    if (outFile->my_irrep != 0) {
-        throw FeatureNotImplemented("libmints Matrix class", "Matrices whose irrep is not the symmetric one", __FILE__,
-                                    __LINE__);
+    if (outFile->my_irrep != symmetry_) {
+        std::stringstream msg;
+        msg << "Symmetry mismatch. Matrix has symmetry " << outFile->my_irrep << " where dpdfile has "
```
loriab, line 2821: perhaps "where" --> "whereas" or "!="

tests/matrix2/test_input.py
```diff
@@ -0,0 +1,6 @@
+from addons import *
+
+@ctest_labeler("quick;misc")
+def test_matrix1():
```
loriab, line 4:
```suggestion
def test_matrix2():
```

tests/matrix2/input.dat
```diff
@@ -0,0 +1,3 @@
+#! An example of using BLAS and LAPACK calls directly from the Psi input file, demonstrating
+#
+assert psi4.core.test_matrix_dpd_interface()
```
loriab, line 3: Have you checked this syntax with a forced fail? I don't think ctest works properly with `assert`, which is a pytest pattern. Try `compare(True, psi4.core.test_matrix..., "matrix dpd")` next.

JonathonMisiewicz: Yes, `ctest` works completely fine as-is. I prefer the current version.

This file works because the Psi input file is parsed to Python. Running the input file raises a Python error which is registered as a crash. This is all standard behavior.

loriab (approved):
lgtm after the test_input.py suggestion. (feel free to edit that on a later PR if this one doesn't need another CI run.)

maxscheurer (approved):
LGTM

psi-rking (approved):
LGTM

PR #2724, opened 2022-09-24
Remove unused psio_* functions

TiborGY:
## Description

There are some IO functions declared in psio.**h** instead of psio.**hpp**, which manipulate the state of the global PSIO object. Some of these functions are never called, a few other functions are declared but never defined.
This PR removes them.

- [x] No user facing or API changes

## Dev notes & details

- [x] Unused global-scope psio_* functions are removed

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dct/dct_compute.cc
+++ b/psi4/src/psi4/dct/dct_compute.cc
@@ -49,8 +49,8 @@ double DCTSolver::compute_energy() {
 
     if (options_.get_str("DCT_GUESS") == "DCT") {
         // We're reusing DCT files from a previous computation. Let's check that one exists.
-        if (!((same_a_b_orbs_ == true && _default_psio_lib_->tocentry_exists(PSIF_DCT_DPD, "Amplitude SF <OO|VV>")) ||
-              (same_a_b_orbs_ == false && _default_psio_lib_->tocentry_exists(PSIF_DCT_DPD, "Amplitude <Oo|Vv>")))) {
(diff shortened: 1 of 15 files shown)


JonathonMisiewicz (dismissed):
Given how difficult PSIO is to use for non-experts (which is anybody who is going to be using it nowadays), I'd rather keep thin but descriptive wrappers.

loriab:
btw, don't exert much effort on anything only used by optking as hopefully that gets deleted by end of year.

TiborGY:
> Given how difficult PSIO is to use for non-experts (which is anybody who is going to be using it nowadays), I'd rather keep thin but descriptive wrappers.

Very well. The two functions in question are now preserved.

JonathonMisiewicz (approved):

loriab (approved):
Looks reasonable to me, and will help define more tightly the needed I/O interface. Would be good to get 3rd review from an early developer.

lothian:
Just to confirm: these are almost all just removal of the C-based wrapper functions, correct?

TiborGY:
> Just to confirm: these are almost all just removal of the C-based wrapper functions, correct?

Yes.

As far as I can tell different modules seem to be using PSIO in slightly different way. The newer modules tend to create their own PSIO object and then call its member functions, but older modules seem to be more reliant on global state. To satisfy this, there is a "global PSIO object", and some wrapper functions have been written that usually do the same thing as the corresponding PSIO member functions, except they are regular functions and they manipulate the "global PSIO object".

This PR removes the unused ones.

lothian:
Yes, I believe the global PSIO object was to created to minimize the changes needed to older modules that worked with the pre-OOP version of the library.

lothian (approved):

PR #2727, opened 2022-09-26
Optking

AlexHeide:
## Description

Remove c++ optking. Add new python optking driver.

- [x] RN 1 The fixed_* optimization keywords have been changed to ranged_* options
- [x] RN 2 output will be changed. Check output.dat for simple convergence / step info. output.log for detailed info
- [x] RN 3 IRC convergence behavior different for minima and substep. 

## Dev notes & details

- [x] New optimizer. Most of driver is very similar
- [x] Hessian Updating and optimization logic is in optking as much as possible
- [x] Restarting optimizations now supported. Optking can write entire state to disc (json)
- [x] Driver attempts to symmetrize geometries
- [x] New keywords added. Some removed / updated. 
- [x] old optking removed from CMakeLists.txt from `/psi4/src/psi4`
- [x] optking py_funcs removed from core.cc
- [x] All tests have been verified for equal or better convergence. (except opt-irc-2)

## Questions
- [ ] A document describing differences between the optimzers may be necessary
- [ ] Unclear if I have fully removed old optking and building with new optking is fully working.

## Checklist
- [x] test15 tests restart
- [x] Can remove old deprecated set_gradient
- [x] ctest -L opt run

## Status
- [ ] Ready for review
- [ ] Ready for merge

For documentation on the new optimizer please see https://optking.readthedocs.io/en/latest/


AlexHeide:
> Just pointing out that this PR doesn't remove C-side optking code yet. Is that coming?

Just an oversight. deleted but not comitted.

susilehtola:
> > Just pointing out that this PR doesn't remove C-side optking code yet. Is that coming?
> 
> Just an oversight. deleted but not comitted.

The Optking is dead. Long live the Pyoptking.

AlexHeide:
The test that is currently failing is a test in `gcp/pbeh3c/`. This is due to an optking side issue where the `CustomHelper` class being used by optking is not accepting a `psi4.core.Molecule` the type checking was looking for `qcdb.Molecule`. As a backup optking defaulted to psi4's active molecule.

This is the call.
`E = optimize('pbeh3c/def2-msvp', molecule=unopethene)`

optking takes the molecule here in the driver and uses a default fallback instead: 
` opt_object = optking.opt_helper.CustomHelper(molecule, params=optimizer_params)`

The type check will get changed in optking to include `core.Molecule`. An alternative question this raises for me is whether the active_molecule should get updated at some point in the optimization. Updating the active molecule patches the issue but is that desired?

psi-rking:
> The test that is currently failing is a test in `gcp/pbeh3c/`. This is due to an optking side issue where the `CustomHelper` class being used by optking is not accepting a `psi4.core.Molecule` the type checking was looking for `qcdb.Molecule`. As a backup optking defaulted to psi4's active molecule.
> 
> This is the call. `E = optimize('pbeh3c/def2-msvp', molecule=unopethene)`
> 
> optking takes the molecule here in the driver and uses a default fallback instead: ` opt_object = optking.opt_helper.CustomHelper(molecule, params=optimizer_params)`
> 
> The type check will get changed in optking to include `core.Molecule`. An alternative question this raises for me is whether the active_molecule should get updated at some point in the optimization. Updating the active molecule patches the issue but is that desired?

My vote would be "no"; it's too much like changing the input.  However, I do think that some users would expect the final active molecule to be updated for them.  And workflows inside the python input may seem more intuitive that way.

psi4/src/core.cc
```diff
@@ -1371,7 +1371,7 @@ PYBIND11_MODULE(core, core) {
     core.def("occ", py_psi_occ, "ref_wfn"_a, "Runs the orbital optimized CC codes.");
     core.def("dfocc", py_psi_dfocc, "ref_wfn"_a, "Runs the density-fitted orbital optimized CC codes.");
     core.def("adc", py_psi_adc, "ref_wfn"_a, "Runs the ADC propagator code, for excited states.");
-    core.def("opt_clean", py_psi_opt_clean, "Cleans up the optimizer's scratch files.");
+    // core.def("opt_clean", py_psi_opt_clean, "Cleans up the optimizer's scratch files.");
```
loriab, line 1374: go ahead and delete the lines entirely, I think.

tests/opt15/test_input.py
```diff
@@ -0,0 +1,6 @@
+from addons import *
+
+@ctest_labeler("opt;cart;noc1")
+def test_opt14():
```
loriab, line 4:
```suggestion
def test_opt15():
```

loriab:
>> An alternative question this raises for me is whether the active_molecule should get updated at some point in the optimization. Updating the active molecule patches the issue but is that desired?

> My vote would be "no"; it's too much like changing the input. However, I do think that some users would expect the final active molecule to be updated for them. And workflows inside the python input may seem more intuitive that way.

I may not be following this right. I'd say the state of the psi4 active mol _during_ an optimization is arbitrary -- whatever works for you. In cpp-optking, I think the communication was through legacymolecule anyways. The molecule optking is acting on _should_ be updated by the time control returns to the user at the end of the opt. I thought this was already happening through https://github.com/psi4/psi4/pull/2727/files#diff-acf663ccea13592c4c656cf89c7b62e6f5bd3b2e8b4f12ba354129bd39d096f8R1296-R1297 . That's consistent with cpp-optking, and I think that must be happening b/c many of the tests check NRE before and after opt.

.azure-pipelines/azure-pipelines-windows.yml
```diff
@@ -138,6 +138,7 @@ jobs:
                         psi4/label/dev::libint2=*=h2e52968_4 ^
                         conda-forge::qcelemental=0.25.0 ^
                         conda-forge::qcengine=0.23.0 ^
+                        conda-forge::optking=0.2.0 ^
```
loriab, line 141: if you only put a version here b/c qcel/qcng had versions, then there's no need for it. easier to debug and update optking upstream if it's floating.

psi4/src/core.cc
```diff
@@ -156,10 +156,10 @@ void py_psi_plugin_close_all();
...
 extern std::map<std::string, plugin_info> plugins;
 
-namespace opt {
-psi::PsiReturnType optking(psi::Options&);
-void opt_clean();
-}  // namespace opt
+// namespace opt {
+// psi::PsiReturnType optking(psi::Options&);
+// void opt_clean();
+// }  // namespace opt
```
loriab, line 162: you can also delete legacy_molecule/wavefunction/gradient . first and last are cpp-optking-only. they're a little spread out -- defined in `libpsi4util/process.h`. not strictly necessary, so up to you.

psi-rking:
> > > An alternative question this raises for me is whether the active_molecule should get updated at some point in the optimization. Updating the active molecule patches the issue but is that desired?
> 
> > My vote would be "no"; it's too much like changing the input. However, I do think that some users would expect the final active molecule to be updated for them. And workflows inside the python input may seem more intuitive that way.
> 
> I may not be following this right. I'd say the state of the psi4 active mol _during_ an optimization is arbitrary -- whatever works for you. In cpp-optking, I think the communication was through legacymolecule anyways. The molecule optking is acting on _should_ be updated by the time control returns to the user at the end of the opt. I thought this was already happening through https://github.com/psi4/psi4/pull/2727/files#diff-acf663ccea13592c4c656cf89c7b62e6f5bd3b2e8b4f12ba354129bd39d096f8R1296-R1297 . That's consistent with cpp-optking, and I think that must be happening b/c many of the tests check NRE before and after opt.

OK, you changed my hasty mind on that.  What concerns me is that the user may, in some instances, not realize that the default active molecule is the one that optking is acting on and changing.  But the upside convenience wins, I agree.

psi4/src/read_options.cc
```diff
@@ -2633,8 +2631,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("IRC_STEP_SIZE", 0.2);
         /*- IRC mapping direction -*/
         options.add_str("IRC_DIRECTION", "FORWARD", "FORWARD BACKWARD");
-        /*- Decide when to stop IRC calculations -*/
-        options.add_str("IRC_STOP", "STOP", "ASK STOP GO");
+        /*- Number of IRC points to collect before stopping -*/
+        options.add_int("IRC_POINTS", 20);
```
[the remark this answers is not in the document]
psi-rking: Yes, there's an energy-minimum check at each converged point; it'll stop if it finds one.

loriab:
> OK, you changed my hasty mind on that. What concerns me is that the user may, in some instances, not realize that the default active molecule is the one that optking is acting on and changing. But the upside convenience wins, I agree.

If changing the mol over the course of an opt was new behavior, I think it'd be very much up for debate. I know FAE is not keen on background/globals changes. And they are weird in a Jupyter notebook. But for now, I'm for minimal surprise upon cpp-optking -> py-optking.

psi4/src/read_options.cc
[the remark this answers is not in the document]
loriab: ok, gtk. I originally asked to see if it might fit as a "MAXITER" keyword. But I now don't think that's right anyhow. With a SCF or OPT maxiter, if it converges in 2, you're pleased. Here, if it converges in 2, you have only a meager IRC.

.azure-pipelines/azure-pipelines-windows.yml
AlexHeide: That is exactly why I did :)

JonathonMisiewicz:
Please ping me for review once Lori approves. She knows this part of the code better than I do.

psi4/src/read_options.cc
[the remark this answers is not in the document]
AlexHeide: To clarify, at least in my mind, its most helpful when dealing with paths like a dissociation channel where no well defined minimum may exist to keep optking from doing unnecessary, unending work. It can also function as a distance metric i,.e. do 30 points or 3 Angstroms along the reaction path. If the irc doesn't end before 20 points, the user can just restart anyhow.

psi4/src/core.cc
AlexHeide: Now that optking isn't using `PsiReturnType.Failure` and `EndLoop` should those be deprecated and/or removed? A quick search indicates success is used by fnocc I didn't see balk used anywhere. I'm happy to perform as much or as little code cleanup with this PR as needed.

loriab: I think kill off `Balk` and `EndLoop`. `Success` and `Failure` are reasonable for general use.

AlexHeide:
In cases where reference values are not matched perfectly (but the test should pass) is it better to loosen the comparison or update the reference value. I assume updating the reference value is the way to go but I want to have some record of asking before I start slightly changing reference values.

Explanation:
I've expanded the number of tests I'm running since I started cleaning up core.cc. In the opt specific tests the convergence is usually tight enough that the nuclear repulsion energies match the reference values just fine for both optimizers. There are some tests like cc1-3 that are failing due to being just above threshold. `atol = 0.001` the difference is ~ 0.0017. These tests are using the default `qchem` convergence criteria ~ `3e-4` max_force. Both optimizers finish well below the criteria and geometries match to 1e-4 Angstroms and  0.001 degrees. Geometries are the same. There are around 5 or 6 tests failing like this.

AlexHeide:
Would y'all like this PR to include updated output.ref files for the tests or would a separate test updating PR be better? I don't see anything in the" adding tests documentation" about reference log files. I can include those as well if desired. It will just greatly increase the number of lines changed in this PR.

JonathonMisiewicz:
> Would y'all like this PR to include updated output.ref files for the tests or would a separate test updating PR be better? I don't see anything in the" adding tests documentation" about reference log files. I can include those as well if desired. It will just greatly increase the number of lines changed in this PR.

Please update `.ref` files. Probably not worth adding `.log` files, but that's a @loriab question.

> There are some tests like cc1-3 that are failing due to being just above threshold. `atol = 0.001` the difference is `~ 0.0017`. These tests are using the default `qchem` convergence criteria ~ `3e-4` max_force. Both optimizers finish well below the criteria and geometries match to 1e-4 Angstroms and  0.001 degrees. 

Could you elaborate on why tests are failing at all, and what numbers are differing? Are these Cartesian coordinates?

loriab:
> In cases where reference values are not matched perfectly (but the test should pass) is it better to loosen the comparison or update the reference value. I assume updating the reference value is the way to go but I want to have some record of asking before I start slightly changing reference values.

> Secondarily, should the reference values be updated in general at some point so that users don't stumble across an instance where the value is slightly off and wonder why?

General guidance to for ref values to be from a tightly converged/optimized calc, then loosen the comparison check to accommodate the default/existing conv crit (https://psicode.org/psi4manual/master/add_tests.html#test-contents). Tests checking opt status at a certain cycle exempt of course. That's the principle, but do feel free to change as you see fit --- the reference values (agreed, preferred thing to change if the ref is the culprit) or the comparison crit (if it's the optimizer behavior that's the instigator).

> Explanation:
I've expanded the number of tests I'm running since I started cleaning up core.cc. In the opt specific tests the convergence is usually tight enough that the nuclear repulsion energies match the reference values just fine for both optimizers. There are some tests like cc1-3 that are failing due to being just above threshold. atol = 0.001 the difference is ~ 0.0017. These tests are using the default qchem convergence criteria ~ 3e-4 max_force. Both optimizers finish well below the criteria and geometries match to 1e-4 Angstroms and 0.001 degrees. Geometries are the same. There are around 5 or 6 tests failing like this.

Thanks for the explanation. I'd view reference NRE values as less venerable. For one thing, only those that caused trouble were even updated when physical constants changed, iirc.

> Would y'all like this PR to include updated output.ref files for the tests or would a separate test updating PR be better? I don't see anything in the" adding tests documentation" about reference log files. I can include those as well if desired. It will just greatly increase the number of lines changed in this PR.

Separate, please. Like updating `samples/`, better to keep the not-for-visual-inspection changes aside. Update: I see Jonathon thinks differently. Including is ok with me, now that GH allows files to be folded up rather than scrolled through. Is `.log` where all the optking detailed output goes now? I guess we ought to start collecting them. `output.log`, perhaps. But this can also be deferred to a grand regenerate-the-refs script and PR.

JonathonMisiewicz:
I'm perfectly happy to leave `.ref` to a separate PR, so long as @AlexHeide agrees to take responsibility for getting that in by PsiCon at the latest. 🙂

AlexHeide:
> Could you elaborate on why tests are failing at all, and what numbers are differing? Are these Cartesian coordinates?

I poorly worded by explanation. I was attempting to say that the geometries are virtually the same but not numerically, they are both well converged. I assumed the discrepancy was just a numeric difference and there have been various tweaks as well over the last few years to the algorithm.

There is 1 algorithmic difference I'd be concerned about, from looking at the two outputs. New optking doesn't consider the most recent step in the hessian updating procedure - cpp-optking did. I've found three pieces of logic that explicitly or implicitly prevent updating with the most recent step in all or specific cases. I'd have to ask @psi-rking if this is a bug or was changed due to some stability concern.

>  Is .log where all the optking detailed output goes now? I guess we ought to start collecting them. output.log, perhaps. But this can also be deferred to a grand regenerate-the-refs script and PR.

Yes optking's detailed output goes to `.log` but it isn't strictly speaking necessary for the user to see the detailed logs I would say. However, if the test is failing it might be nice to have a more detailed optimization record to compare against if the optimizer is at fault.

JonathonMisiewicz:
Just updating the reference is fine with me, in that case.

psi-rking:
> There is 1 algorithmic difference I'd be concerned about, from looking at the two outputs. New optking doesn't consider the most recent step in the hessian updating procedure - cpp-optking did. I've found three pieces of logic that explicitly or implicitly prevent updating with the most recent step in all or specific cases. I'd have to ask @psi-rking if this is a bug or was changed due to some stability concern.

Interesting.  This does not ring a bell for me.  I take the question to be "given a hessian and the current forces (beyond the first step) should you update the hessian with those forces before using them to calculate the step?"  Can you point to the code?  IDK, probably best answer is whatever works better in practice and doesn't cause problems.

AlexHeide:
> I may not be following this right. I'd say the state of the psi4 active mol during an optimization is arbitrary -- whatever works for you. In cpp-optking, I think the communication was through legacymolecule anyways. The molecule optking is acting on should be updated by the time control returns to the user at the end of the opt. I thought this was already happening through https://github.com/psi4/psi4/pull/2727/files#diff-acf663ccea13592c4c656cf89c7b62e6f5bd3b2e8b4f12ba354129bd39d096f8R1296-R1297 . That's consistent with cpp-optking, and I think that must be happening b/c many of the tests check NRE before and after opt.

I want to make sure that the active molecule behavior is as expected. Whatever molecule the driver uses will be updated. This will be either the active molecule OR the passed molecule. If the molecule is passed the active molecule is not updated in any way. All the asserts pass here.

```python
import math

molecule h2o {
    pubchem:water
}

molecule h2o2 {
    pubchem:hydrogen peroxide
}

# quick comparison. h2o2 is active molecule
h2o2_nre = h2o2.nuclear_repulsion_energy()
active_nre = core.get_active_molecule().nuclear_repulsion_energy()

assert math.isclose(h2o2_nre, active_nre)

# optimize h2o2 (active molecule) expect repulsion energy to match
E = optimize("scf/sto-3g")
h2o2_opt_nre = h2o2.nuclear_repulsion_energy()
active_nre = core.get_active_molecule().nuclear_repulsion_energy()

assert math.isclose(h2o2_opt_nre, active_nre)

# optimize h2o. nuclear repulsion does not match. active molecule is still h2o2
# Currently (next commit will fix) h2o2 would be optimized by this call. (optking side issue)
E = optimize("scf/sto-3g", molecule=h2o)
h2o_opt_nre = h2o.nuclear_repulsion_energy()
active_nre = core.get_active_molecule().nuclear_repulsion_energy()

assert math.isclose(h2o2_opt_nre, active_nre)
```
The whole wrong molecule being optimized thing is an optking side issue that is fixed on optking/master.

loriab (commented):
Thanks for all the cleanup! I've now looked over everything but the actual meat, driver.py . That's next.

psi4/src/read_options.cc
```diff
@@ -2662,18 +2660,34 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+            analogous to the previous FIXED_BEND -*/
+        options.add_str("RANGED_BEND", "");
+        /*- Specify range for the dihedral angles between atoms to be constrained to (eq. value specified)
+            analogous to the previous FIXED_DIHEDRAL -*/
+        options.add_str("RANGED_DIHEDRAL", "");
+        /*- Specify range for the out-of-plane angles between atoms to be constrained to (eq. value specified)
+            analogous to the old FIXED_ keyword-*/
+        options.add_str("RANGED_OOFP", "");
+
+        options.add_str("EXT_FORCE_DISTANCE", "");
```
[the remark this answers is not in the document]
psi-rking: There is in optking/optparams.py.   So what is the strategy of documenting the keywords?  Will we do it redundantly?  Is there a machine-readable list somewhere yet?  This is a catch-up question for me.  I can help with this if needed.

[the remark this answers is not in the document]
AlexHeide: I think we'll need to do this redundantly. Keyword updates should be easy to get reviews for and changes in. As a workaround we can pass extra options into `optimize(..., **kwargs)` I think this is what geometric does by default. We could use the same key `optimizer_keywords` for the sake of consistency. 

https://github.com/psi4/psi4/blob/8c6e5bfa5c501567d7ec1ee3d4c20b26f814e90c/psi4/driver/driver.py#L843

This is how we're currently getting options so we can merge the two. Probably would be best to overwrite with whatever options are explicitly passed in.
https://github.com/psi4/psi4/blob/cdb25070b490b12c8dc0134408bf5a3ba0241fa2/psi4/driver/driver.py#L1186-L1188

> Is there a description for EXT_FORCE_DISTANCE? 

Yep it must have gotten missed or accidentally removed when I was formatting.

psi4/src/psi4/detci/ints.cc
```diff
@@ -379,7 +379,6 @@ void CIWavefunction::setup_mcscf_ints_ao() {
     std::string scf_type = options_.get_str("SCF_TYPE");
     if (scf_type == "GTFOCK") {
 #ifdef HAVE_JK_FACTORY
```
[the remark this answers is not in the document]
AlexHeide: That makes sense. I did a search for whether `get_legacy_molecule()` was ever called in detci and the other modules and didn't find anything. I just assumed the code had been missed in refactoring.

psi4/src/read_options.cc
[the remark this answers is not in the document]
loriab: I agree on redundant options for this PR. There is a way of injecting options, like for plugins, but that means C++ code, which you don't want in Py optking. Also, getting those options injected early enough for the Options object to recognize them can be iffy. Non-redundant opts should certainly be part of a future redesign.

Docs, on the other hand, are readily linked. So if you want to overhaul or de-duplicate the docs in a future PR, I don't see any problem.

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +258,219 @@ For bends, the corresponding keyword is "frozen_bend".
...
+   MTdimer = """{
+      "Natoms per frag": [12, 16],
+      "A Frag": 1,
+      "A Ref Atoms": [[1, 3, 4, 6, 8], [8], [11]],
+      "A Label": "methylthiophene",
+      "B Frag": 2,
+      "B Ref Atoms": [[13, 14, 15, 16, 17, 18], [13], [15]],
+      "B Label": "tyrosine",
+      "Frozen": ["theta_A", "theta_B", "tau", "phi_A", "phi_B"],
+   }"""
```
loriab, line 465: It could be good to have a pytest test, maybe in test_addons.py, showing the passing through of optking opts in psiapi where you shouldn't have to resort to stringifying the dict.

loriab (commented):
This is looking good to me, incl. driver.py. There's a few code suggestions, and I think you're still working on tweaking some test refs, then I think its good to merge for broader trial.

psi-rking (approved):

psi4/driver/driver.py
```diff
@@ -1196,22 +1187,49 @@ def optimize(name, **kwargs):
     molecule = kwargs.pop('molecule', core.get_active_molecule())
 
     # If we are freezing cartesian, do not orient or COM
-    if core.get_local_option("OPTKING", "FROZEN_CARTESIAN"):
+    if any([core.get_option('OPTKING', 'FROZEN_CARTESIAN'), core.get_option('OPTKING', 'EXT_FORCE_CARTESIAN')]):
+        if molecule.has_zmatrix():
+            raise ValidationError("Job includes cartesian coordinate constraints. This cannot be fully "
+                                  "obeyed due to zmatrix in input. Please convert your zmatrix to cartesian "
+                                  "coordinates if cartesian constraints are needed ")
```
loriab, line 1194: glad this is trapped

loriab (commented):
a few minor cleanups but lgtm!

psi4/driver/driver.py
psi-rking: I presume the issue here is that the user  does not know what the Cartesian geometry will be, and thus the user cannot have a clear conception of what the Cartesian constraints mean.  As far as optking is concerned, I used to just ignore the Z-matrix.  @AlexHeide , is this still what we are doing?  I agree the trap is sound regardless.

AlexHeide: Correct, the zmatrix still does not affect optking. This check was added because psi4 does not actually obey fix_com and fix_orientation when using zmatrices. By this I mean, psi4 doesn't keep the same orientation over the course of several geometry updates. I think psi4 converts our cartesians back to the zmatrix and then optking requests new cartesian coordinates forcing a zmatrix to Cartesian conversion which ignores the fix keywords.

The only thing the user is missing out on here is freezing atoms in cartesian space when using internal coordinates. Any other Cartesian constraints (like ranges), I agree, the user couldn't really hope to specify.

psi4/extras.py
```diff
@@ -276,5 +276,8 @@ def set_output_file(ofile: str, append: bool = False, *, loglevel: int = 20, exe
 
     if execute:
         core.set_output_file(str(out), append)
-        logger.addHandler(f_handler)
+        print(logger.handlers)
```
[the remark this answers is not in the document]
AlexHeide: Debug. Good catch. As an explanation, psi4 was duplicating log output when running multiple tests in a single input file. Wasn't super clear when just psi4 was logging but became very noticeable with the amount that optking logs.

doc/sphinxman/source/optking.rst
```diff
@@ -36,12 +36,12 @@
...
-.. codeauthor:: Rollin A. King
-.. sectionauthor:: Rollin A. King and Lori A. Burns
+.. codeauthor:: Rollin A. King and Alexander G. Heide
+.. sectionauthor:: Rollin A. King, Alexander G. Heide, and Lori A. Burns
 
-*Module:* :ref:`Keywords <apdx:optking>`, :ref:`PSI Variables <apdx:optking_psivar>`, :source:`OPTKING <psi4/src/psi4/optking>`
+*Module:* :ref:`Keywords <apdx:optking>`, `OPTKING <https://github.com/psi-rking/optking>`_
 
-|PSIfour| carries out molecular optimizations using a module called
+|PSIfour| carries out molecular optimizations using a python module called
```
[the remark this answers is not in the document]
AlexHeide: Should optking be in the preceding section, under

```
The following are also required for |PSIfour|, but if not detected, the
build system will automatically download and build.
```
I could remove the download from source from cmake.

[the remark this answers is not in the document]
loriab: I agree it's ambiguous what section in build_planning.rst optking should go in. On one hand, optking isn't _used_ until runtime. On the other hand, since we're requiring it, it will be built at buildtime if missing. I think just choose a place for the bullet that seems reasonable. Definitely do leave the downloading of src from cmake as-is.

external/upstream/optking/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ else()
 
     ExternalProject_Add(optking_external
         BUILD_ALWAYS 1
-        URL https://github.com/AlexHeide/optking/tarball/hessian
+        URL https://github.com/psi-rking/optking/tarball/master
```
loriab, line 22: you can replace `master` with a tag name or the short hash for the known stable commit. Otherwise there's a risk of users building a cutting edge version.

loriab (approved):
lgtm, thank you!

JonathonMisiewicz:
I'll review after merge conflicts are fixed. Sorry @AlexHeide, I probably caused them.

AlexHeide:
Np. it looks like you removed legacy freq and wfn. I had removed legacy molcule, grad, and wfn. Legacy should be fully purged! :tada: Building and testing.

AlexHeide:
After rebase, `cookbook-manual-fd-hess-grad` and `cookbook-manual-fd-hess-energy` fail due to `set_frequencies` missing. Is there something I should replace those with? I don't see an alternative recommended in the deprecation notice.

JonathonMisiewicz:
> After rebase, `cookbook-manual-fd-hess-grad` and `cookbook-manual-fd-hess-energy` fail due to `set_frequencies` missing. Is there something I should replace those with? I don't see an alternative recommended in the deprecation notice.

No. `wfn.frequencies()` will return the frequencies automagically thanks to `wfn.frequency_analysis = vibinfo`

@loriab can probably confirm.

AlexHeide:
Can I get a test re-trigger? I'm not sure why mac failed, but I'm hoping it was a fluke.

JonathonMisiewicz:
Not a fluke, but not your fault either. Ask again when [Lori's patch](https://github.com/psi4/psi4/pull/2815) is merged in.

loriab:
please rebase to fix CI

doc/sphinxman/source/optking.rst
```diff
@@ -328,57 +498,19 @@ For difficult cases, the following suggestions are made.
...
-  corresponds to an optimization that will add additional forces to move the O-H bonds
-  to 1.70 au. ::
-
-     F 1 3
-     R      1     2     1.70
-     R      1     3     1.70
-     B      2     1     3
+.. warning:: In some cases, such as the coordinate issues described above, optking will reset to maintain
+  a consistent history. If an error occurs in Psi4 due to |optking__geom_maxiter| being exceeded but
+  the final step report indicates that optking has not taken |optking__geom_maxiter| steps such a
```
JonathonMisiewicz, line 511:
```suggestion
  the final step report indicates that optking has not taken |optking__geom_maxiter| steps, such a 
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +258,219 @@ For bends, the corresponding keyword is "frozen_bend".
...
+   my_string = f"1 2 3 4 {lower} {upper}"
+   set optking ranged_dihedral = $my_string
+   E = optimize('scf')
+   PES.append((upper, E))
+
+   print("\n\tcc-pVDZ SCF energy as a function of phi\n")
+   for point in PES:
+     print("\t%5.1f%20.10f" % (point[0], point[1]))
+
+* To scan the potential energy surface without the |optking__ranged_dihedral| keyword a zmatrix
```
JonathonMisiewicz, line 303:
```suggestion
* To scan the potential energy surface without the |optking__ranged_dihedral| keyword, a zmatrix
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +258,219 @@ For bends, the corresponding keyword is "frozen_bend".
...
    for point in PES:
-     print "\t%5.1f%20.10f" % (point[0], point[1])
+     print("\t%5.1f%20.10f" % (point[0], point[1]))
+
+Multi-Fragment Optimizations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _DimerSection:
+
+Control over the intermolecular coordinates are controlled through |optking__interfrag_mode|,
```
JonathonMisiewicz, line 344:
```suggestion
The intermolecular coordinates are controlled through |optking__interfrag_mode|,
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +258,219 @@ For bends, the corresponding keyword is "frozen_bend".
...
+|optking__frag_ref_atoms|, and |optking__interfrag_coords|.
+
+* To specify the reference points to use for coordinates |optking__frag_ref_atoms| is
+  specified. Each list corresponds to a fragment. A list of indices denotes a linear combination
```
JonathonMisiewicz, line 348:
```suggestion
* Specify the reference points to use for coordinates via |optking__frag_ref_atoms|. A list of indices denotes a linear combination
```

JonathonMisiewicz (commented):
Alex, please re-read the optking doc page for Psi.

psi4/driver/driver.py
```diff
@@ -1220,88 +1238,79 @@ def optimize(name, **kwargs):
...
+        opt_object.gX = G.np
+
+        if core.get_option('OPTKING', 'CART_HESS_READ') and (n == 1):
+            opt_object.params.cart_hess_read = True
+            opt_object.params.hessian_file = f"{core.get_writer_file_prefix(molecule.name())}.hess"
+                # compute Hessian as requested; frequency wipes out gradient so stash it
+        elif 'hessian' in opt_calcs:
+            # compute hessian as requested.
+
+            # TODO test optimize dertype=0
```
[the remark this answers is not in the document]
AlexHeide: It is. `cc2` tests optimize with dertype=0. psi4 + optking + distributed driver test hasn't been explicitly tested, but optking wouldn't ("shouldn't") change anything about `psi4.gradient` or `psi4.hessian` running  through qcarchive

doc/sphinxman/source/optking.rst
```diff
@@ -277,30 +276,203 @@ how the geometry of the molecule was input (whether Z-matrix or cartesian, etc.)
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _DimerSection:
+
+The intermolecular coordinates are controlled through |optking__interfrag_mode|,
+|optking__frag_ref_atoms|, and |optking__interfrag_coords|.
+
+* Specify the reference points to use for coordinates via |optking__frag_ref_atoms|.
+  Each list corresponds to a fragment. A list of indices denotes a linear combination
+  of the atoms. In this case the first reference point for the second dimer is the center of the
```
JonathonMisiewicz, line 351:
```suggestion
  of the atoms. In this case, the first reference point for the second dimer is the center of the
```

doc/sphinxman/source/optking.rst
JonathonMisiewicz: Still need to adopt.

doc/sphinxman/source/optking.rst
```diff
@@ -277,30 +276,203 @@ how the geometry of the molecule was input (whether Z-matrix or cartesian, etc.)
...
+
+Multi-Fragment Optimizations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _DimerSection:
+
+The intermolecular coordinates are controlled through |optking__interfrag_mode|,
+|optking__frag_ref_atoms|, and |optking__interfrag_coords|.
+
+* Specify the reference points to use for coordinates via |optking__frag_ref_atoms|.
```
JonathonMisiewicz, line 349: I don't understand this section at all. What does "reference points to use for coordinates" mean? Am I expected to be able to follow this, or are you assuming the user has already ready some optking-specific documentation?

AlexHeide: No optking's documentation on this section is actually less detailed. I'll try to clarify that section and add a note that most users should just set `interfrag_mode` to `multi` and let optking handle coordinates.

AlexHeide:
@psi-rking Can you give the optking docs another read through? Anything from your experiences with forum questions and such that we should add? If you have any thoughts on further clarifying the inter-fragment section that would also be super helpful. I'm printing the html now to make another editing pass.

psi-rking:
Do we have a figure in there yet illustrating the inter-fragment
definitions?  I have no time to work on it today, but could tomorrow
(Saturday).

loriab:
`/home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/optking.rst:346: ERROR: Undefined substitution referenced: "optking_frag_mode".`

docs error above. probably needs double underscore after optking.

> Do we have a figure in there yet illustrating the inter-fragment
definitions?  I have no time to work on it today, but could tomorrow
(Saturday).

interfrag figure would be great. but if we get to the point where everything else is ready, I think that could be a separate PR.

psi-rking:
Yes, I agree that details instructions and more demonstrations could be
added in separate PR, and not a reason to stall Alex's integration efforts.

JonathonMisiewicz:
Let me be more explicit about why I find the "multi-fragment optimizations" section confusing.

* The first sentence talks about "the metric for connecting atoms" without explaining what "connecting atoms" means. Even worse, the first sentence is not obviously about dimers.
* It isn't clear to me what a "reference point" signifies. While I can tell whether something is an acceptable reference point, what are these used for? It looks like these are atoms used to define intermolecular internal coordinates.
* I don't know what it means to talk about a linear combination of atoms, or how to interpret `[[3], [1], [2]], [[1, 2, 3, 4, 5, 6], [2], [6]]` in your first example.

The third point is crucial - I can't follow your examples. I won't insist on a figure, but I do insist on examples I can understand.

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
-     print "\t%5.1f%20.10f" % (point[0], point[1])
+     print("\t%5.1f%20.10f" % (point[0], point[1]))
+
+Multi-Fragment Optimizations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. _DimerSection:
+
+In previous versions of optking, the metric for connecting atoms was increased until all atoms,
+were connected. This is the current behavior for |optking__frag_mode| `singleo`.
```
psi-rking, line 351:
```suggestion
were connected. This is the current behavior for |optking__frag_mode| `single`.
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
+Setting |optking__frag_mode| to `multi` will now add a special
+set of intermolecular coordinates between fragments - internally referred to as DimerFrag
+coordinates. For a set of molecular fragments, a set of reference points are chosen on each
```
psi-rking, line 354:
```suggestion
coordinates. For each pair of molecular fragments, a set of up to 3 reference points are chosen on each
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
+fragment. The reference points may be an atom or a linear combination of atomic positions.
```
psi-rking, line 355:
```suggestion
fragment.  Each reference point will be either an atom or a linear combination of positions of atoms within that fragment.
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
+Stretches, bends, and dihedral angles between two of the fragments will be created using
+these reference points. Up to six reference points will be created (depending upon the number
+of atoms in each fragment). See :ref:`Dimer coordinate table <table:DimerFrag>` for how
+reference points are created. For a set of three dimers A, B, and C, sets of coordinates are
+created between each pair: AB, AC, and BC. Each DimerFrag will likely use different reference
```
psi-rking, line 360:
```suggestion
created between each pair: AB, AC, and BC. Each DimerFrag may use different reference
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
+.. _DimerSection:
+
+In previous versions of optking, the metric for connecting atoms was increased until all atoms,
+were connected. This is the current behavior for |optking__frag_mode| `singleo`.
+Setting |optking__frag_mode| to `multi` will now add a special
+set of intermolecular coordinates between fragments - internally referred to as DimerFrag
+coordinates. For a set of molecular fragments, a set of reference points are chosen on each
+fragment. The reference points may be an atom or a linear combination of atomic positions.
+Stretches, bends, and dihedral angles between two of the fragments will be created using
+these reference points. Up to six reference points will be created (depending upon the number
```
psi-rking, line 357: Suggest deleting the sentence that begins with "Up to six..."

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
+of atoms in each fragment). See :ref:`Dimer coordinate table <table:DimerFrag>` for how
+reference points are created. For a set of three dimers A, B, and C, sets of coordinates are
+created between each pair: AB, AC, and BC. Each DimerFrag will likely use different reference
+points. Creation of the intermolecular coordinates can be controlled through
+|optking__interfrag_mode|, |optking__frag_ref_atoms|, and |optking__interfrag_coords|. 
+
+.. note:: Manually creating the interfragment coordinates is meant for power users.
```
psi-rking, line 364:
```suggestion
.. note:: Manual specification of the interfragment coordinates is supported for power users, and provides complete control of fragments' relative orientation.
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +260,279 @@ For bends, the corresponding keyword is "frozen_bend".
...
+       H   -1.3768100495       -2.9846751653        5.1327625515
+       H    4.9209603634       -1.7288723155        2.1638694922
+       H    2.1923374156       -0.9964630692        5.1155773223
+       nocom
+       units au
+   }
+   
+   set {
+       basis 6-31+G 
+       frag_mode MULTI
```
psi-rking, line 441:
```suggestion
       frag_mode MULTI
       The line below specifies the reference points that will be used to construct the interfragment coordinates between the two fragments (called A and B).  The format is the following:
       [[A-1], [A-2], [A-3]], [[B-1], [B-2], [B-3]]
 In terms of atoms within each fragment, the line below chooses, for water:

H3 of water for the first reference point, O1 of water for the second reference point, H2 of water for the third reference point

and for benzene:

the mean of the positions of all the C atoms, C2, one of the Carbon atoms, C6, another one of the carbon atoms.
```

psi-rking: @AlexHeide , I am concerned that this example may now not be correct.  In order to support trimers, etc., I changed the atoms numbering to the total atom number - not the number within the fragment.  See e.g., test_dimers_ne2.py for the simplest example that also uses frag_ref_atoms.

psi-rking (approved):

tests/cfour/mints5-grad/input.dat
```diff
@@ -103,29 +103,26 @@ except psi4.ConvergenceError:
...
-}
-try:
-    opt('c4-scf')
-except psi4.ConvergenceError:
-    pass
-print(variable('nuclear repulsion energy'))  #TEST
-compare_values(nre3_ref[1], variable('nuclear repulsion energy'), 3, '[2]  NRE: zmat, external values, w/dummy')  #TEST
-clean()
-opt_clean()
+# molecule NH3  {
```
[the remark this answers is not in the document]
AlexHeide: Yep we have tests elsewhere for NH3. Only a small amount of the tests in mints5 are active to begin with. For this one we would either need to just copy new optkings nuclear repulsion energy into the test, which is kinda cheating, or remove it for now. At some point, we should come back and try to make these tests 1. more reasonable and 2. repeatable. Unfortunately, when I started working on 1 cfour refused to converge the scf. So I went with removal for any that were failing.

doc/sphinxman/source/optking.rst
AlexHeide: Yep. Tests got updated but my examples didn't. Good catch.

loriab:
looks like typos in options snagging up the docs:

```
/home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/optking.rst:359: ERROR: Undefined substitution referenced: "optking_frag_ref_atoms".
/home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/optking.rst:359: ERROR: Undefined substitution referenced: "otking__frag_ref_atoms".
```

doc/sphinxman/source/optking.rst
```diff
@@ -262,45 +262,306 @@ For bends, the corresponding keyword is "frozen_bend".
...
+
+.. _DimerSection:
+
+In previous versions of optking, the metric for connecting atoms was increased until all atoms,
+were connected. This is the current behavior for |optking__frag_mode| `single`.
+Setting |optking__frag_mode| to `multi` will now add a special
+set of intermolecular coordinates between fragments - internally referred to as DimerFrag
+coordinates (see `here <DimerIntro_>` for the brief description). 
+For each pair of molecular fragments, a set of up to 3 reference points
+are chosen on each fragment. Each reference points will be either an atom or a linear combination
```
JonathonMisiewicz, line 365:
```suggestion
are chosen on each fragment. Each reference point will be either an atom or a linear combination
```

Issue #2226, opened 2021-06-30
gemm() does not check matrix sizes


susilehtola:
The `gemm()` function in libmints does not check that the matrices have the proper size.

For instance, `C.dgemm(false, false, 1.0, A, B, 0.0)` should check that 

- the row dimensions of A and C are the same
- the column dimensions of B and C are the same
- the column dimension of A is the same as the row dimension of B

TiborGY:
As far as I can tell at a glance, PR #2729 has fixed this issue.

susilehtola:
I think so too.

PR #2729, opened 2022-09-27
Checked matrix size compatability in gemm() Issue #2226

aquaticseatard:
## Description

Addressing Issue #2226, changing the gemm() method in Matrix.cc

- [x] None

## Dev notes & details

- [x] Calculates row and column totals from rowspi_ and colspi_
- [x] Throws a Psi exception error if the dimensions don't match up.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


susilehtola:
The size check should use the same code as the multiply itself
```
    for (int Ha = 0; Ha < nirrep_; ++Ha) {
        int Hb = Ha ^ (transa ? 0 : a->symmetry()) ^ (transb ? b->symmetry() : 0);
        int Hc = Ha ^ (transa ? a->symmetry() : 0);
        int m = rowspi_[Hc];
        int n = colspi_[Hc ^ symmetry_];
        int k = nlink[Ha ^ symlink];
        int lda = a->colspi_[Ha ^ a->symmetry_];
        int ldb = b->colspi_[Hb ^ b->symmetry_];
        int ldc = colspi_[Hc ^ symmetry_];
        if (m && n && k) {
            C_DGEMM(ta, tb, m, n, k, alpha, &(a->matrix_[Ha][0][0]), lda, &(b->matrix_[Hb][0][0]), ldb, beta,
                    &(matrix_[Hc][0][0]), ldc);
        }
    }
```

aquaticseatard:
> The size check should use the same code as the multiply itself
> 
> ```
>     for (int Ha = 0; Ha < nirrep_; ++Ha) {
>         int Hb = Ha ^ (transa ? 0 : a->symmetry()) ^ (transb ? b->symmetry() : 0);
>         int Hc = Ha ^ (transa ? a->symmetry() : 0);
>         int m = rowspi_[Hc];
>         int n = colspi_[Hc ^ symmetry_];
>         int k = nlink[Ha ^ symlink];
>         int lda = a->colspi_[Ha ^ a->symmetry_];
>         int ldb = b->colspi_[Hb ^ b->symmetry_];
>         int ldc = colspi_[Hc ^ symmetry_];
>         if (m && n && k) {
>             C_DGEMM(ta, tb, m, n, k, alpha, &(a->matrix_[Ha][0][0]), lda, &(b->matrix_[Hb][0][0]), ldb, beta,
>                     &(matrix_[Hc][0][0]), ldc);
>         }
>     }
> ```

Sorry, could you clarify what you mean? Should I incorporate my changes into the multiply or make the style similar?

susilehtola:
It would probably be easiest to incorporate the size check into the loop over symmetries itself... the point being that the check is currently inconsistent with what is actually attempted in the multiplication.

I have no idea what the symmetry code actually does, so someone else should explain that :sweat_smile:

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -1387,7 +1387,23 @@ void Matrix::gemm(const char &transa, const char &transb, const int &m, const in
...
-
+    // Check matrix sizes
+    int tRows;
+    int tCols;
+    int tRowsa;
+    int tColsa;
+    int tRowsb;
+    int tColsb;
+    for (int h = 0; h < nirrep; ++h) {
+        tRows += rowspi_[h];
```
TiborGY, line 1398: You should never add to an uninitialized variable in C++.

TiborGY (changes requested):

JonathonMisiewicz:
I don't know what your background is, so I'm going to write this explanation at a level accessible for a first-year graduate student.

Many matrices in electronic structure theory are indexed by two orbitals. Each orbital will change in a different way under symmetry operations of the molecule (like rotation, reflection), and we call each classification an irreducible representation (irrep). Now, our matrices further have the property that A_mn = 0 unless the symmetries of m and n are somehow related. Our matrix then only needs to store "blocks" where for all orbitals m of irreducible representation h, the only non-zero elements of A are A_mn where n is an orbital of irreducible representation i.

Because we use this block structure, it isn't enough to make sure that the total dimensions of the matrix agree (what your code does). We need the dimensions of each symmetry block to agree on the three points mentioned in the original issue. This validation should be performed for each irreducible representation.

For reference, the code that Susi linked does the following:
```
for each irreducible representation of the first index of A, Ha:
	Hb is the irreducible representation of the first index of the B block that can multiply with the A block
    Hc is the irreducible representation of the first index of the new matrix (if we're multiplying by A, this is just Ha, otherwise this is the irreducible representation of the second index for the block with first index Ha)
	m is the number of orbitals with the irreducible representation for the first index in this block of the new matrix
	n is the number of orbitals with the irreducible representation for the second index in this block of the new matrix
	k is the number of orbitals with the irreducible representation for the second index in this block of A, which is also the number of orbitals with the irreducible representation for the first index in this block of B
	lda is the number of columns of matrix A
	ldb is the number of columns of matrix B
	ldc is the number of columns of matrix C
	do the actual matrix multiplication
```

BLAS savants will notice that I simplified the explanation of `lda`, `ldb`, and `ldc` because we can assume contiguous memory.

psi4/src/psi4/libmints/matrix.cc
JonathonMisiewicz: Also true.

`tRows = rowspi_.sum()` would do the same computation without a for loop (and without adding to uninitialized variables), but as discussed in my previous comment, that isn't the data validation we need.

aquaticseatard:
> I don't know what your background is, so I'm going to write this explanation at a level accessible for a first-year graduate student, though it will still be dense.
> 
> Many matrices in electronic structure theory are indexed by two orbitals. Each orbital will change in a different way under symmetry operations of the molecule (like rotation, reflection), and we call each classification an irreducible representation (irrep). Now, our matrices further have the property that A_mn = 0 unless the symmetries of m and n are somehow related. Our matrix then only needs to store "blocks" where for all orbitals m of irreducible representation h, the only non-zero elements of A are A_mn where n is an orbital of irreducible representation i.
> 
> Because we use this block structure, it isn't enough to make sure that the total dimensions of the matrix agree (what your code does). We need the dimensions of each symmetry block to agree on the three points mentioned in the original issue. This validation should be performed for each irreducible representation.
> 
> For reference, the code that Susi linked does the following:
> 
> ```
> for each irreducible representation of the first index of A, Ha:
> 	Hb is the irreducible representation of the first index of the B block that can multiply with the A block
>     Hc is the irreducible representation of the first index of the new matrix (if we're multiplying by A, this is just Ha, otherwise this is the irreducible representation of the second index for the block with first index Ha)
> 	m is the number of orbitals with the irreducible representation for the first index in this block of the new matrix
> 	n is the number of orbitals with the irreducible representation for the second index in this block of the new matrix
> 	k is the number of orbitals with the irreducible representation for the second index in this block of A, which is also the number of orbitals with the irreducible representation for the first index in this block of B
> 	lda is the number of columns of matrix A
> 	ldb is the number of columns of matrix B
> 	ldc is the number of columns of matrix C
> 	do the actual matrix multiplication
> ```
> 
> BLAS savants will notice that I simplified the explanation of `lda`, `ldb`, and `ldc` because we can assume contiguous memory.

Thanks for the explanation. I edited the program to check for each symmetry block. I'm not entirely sure how the symmetry_ variable works (and the xor involvement), but I mimicked how it was used in other parts of the method.

JonathonMisiewicz:
Thanks for this!

Mimicking is the right thing to do here. We encode irreducible representations into bitstring so that the nonzero blocks of our matrix are exactly those with the property `row_irrep ^ col_irrep ^ symmetry_ = 0`. As a consequence, `A ^ B = C`

The check itself looks correct to me, and this is a huge improvement. For the sake of debug information, I'm tempted to ask for calls to `rowspi().print()` and `colspi().print` on all three matrices involved before the error message appears. I'd also like if the error message included the symmetries of all three matrices and the particular Ha that created the problem. Susi, what do you think?

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -1410,7 +1410,9 @@ void Matrix::gemm(bool transa, bool transb, double alpha, const Matrix *const a,
         int lda = a->colspi_[Ha ^ a->symmetry_];
         int ldb = b->colspi_[Hb ^ b->symmetry_];
         int ldc = colspi_[Hc ^ symmetry_];
-
+        if(m != (transa ? a->colspi_[Ha ^ a->symmetry_] : a->rowspi_[Ha]) || n != (transb ? b->rowspi_[Hb] : b->colspi_[Hb ^ b->symmetry_])
+            || k != (transb ? b->colspi_[Hb ^ b->symmetry_] : b->rowspi_[Hb]))
```
susilehtola, line 1414: There is way too much going on on this line with all the ternary operators.

susilehtola:
@JonathonMisiewicz please tag me next time ;)

The check should be human readable. Please declare new variables for all of the results of the ternary operations. If I were to write a size check for a matrix multiply of `C(m,n) = A(m,k) B(k,n)`, I would write the check something along the lines of `if((m_in != m_out) || (n_in != n_out) || (k1 != k2))`.

And yes, in case of an error, one should print out the dimensions of all three matrices.

aquaticseatard:
Let me know if the print statements should be revised in any way.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -1387,7 +1387,7 @@ void Matrix::gemm(const char &transa, const char &transb, const int &m, const in
 void Matrix::gemm(bool transa, bool transb, double alpha, const Matrix *const a, const Matrix *const b, double beta) {
     if (nirrep_ != a->nirrep_ || nirrep_ != b->nirrep_)
         throw PSIEXCEPTION("Matrix::gemm error: Number of irreps do not equal.");
-
+
```
susilehtola, line 1390: You need to run the code through `clang-format`

susilehtola (changes requested):
The pull has been corrupted with extraneous changes.

loriab:
> The pull has been corrupted with extraneous changes.

You're welcome to heal the git history by whatever procedure seems best to you. But if you don't know what that might be, you could do a fresh clone from master locally, set up origin and upstream, update it to psi4/psi4:master (`git pull --rebase upstream master`), check out a new branch (`git checkout -b matsizecheck`), edit just your one function in `matrix.cc`, commit that, then *force push* to the branch associated with this PR. So assuming `git remote -v` has `origin` pointing to GH:aquaticseatard/psi4 and `upstream` pointing to GH:psi4/psi4, then you'd `git push origin matsizecheck:master --force`. That will return this PR to one commit of just your changes. (Note that if you did the previous command without the `--force`, git would complain about history and suggest you pull. Pulling here is fatal and the cause of the history contamination. You *want* to overwrite history. It's fine as it's only your own, not a shared, branch.)

susilehtola (approved):
I don't know the symmetry code well enough to determine if the checks are correct. However, if the tests pass, a good assumption is that the check is okay. At least it is checking for errors that have not been caught so far...

JonathonMisiewicz (approved):
I _do_ know the symmetry code well enough to determine the checks are correct.

Please mark "ready for review" and "ready for merge."

loriab (approved):
lgtm, thanks for the contribution

hokru (approved):

PR #2736, opened 2022-10-05
Ensure that initialize_lebedev is called --> Clean up spaghetti in cubature

### LINKED ISSUE #2735: lebedev_mapping_ access is thread-unsafe (and also never initialized?)
I've been battling intermittent (read: ~1/1000 runs) hangs of DFT runs that I think I've traced back to some rather mystifying bits of code in cubature.cc... (h/t for @rmcgibbo for lots of help here)

The problems begin here:
https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/cubature.cc#L5106

(Don't mind the mismatch between line numbers I quote and line numbers in my gdb outputs -- there's some edits in a local version of psi4 we're using but I pinkie-promise they're not at all related to the code in-question! ...but to be very sure, I'll check that the same things happen with a truly vanilla build when I get home and have access to my "normal"/vanilla psi4 build.)

Which is where my runs always hang (on the [] access, specifically) when they do. If I pop open gdb, attach, and check what lebedev_mapping_ looks like, I see something somewhat disturbing:

```
#4  psi::SphericalGrid::build (scheme="", npoints=1852534357, points=0x1b2) at /tmp/garden-install.1c2be761ae20/psi4-1.6.1-desres02-c7/psi4/src/psi4/libfock/cubature.cc:5142
5142    /tmp/garden-install.1c2be761ae20/psi4-1.6.1-desres02-c7/psi4/src/psi4/libfock/cubature.cc: No such file or directory.
(gdb) print lebedev_mapping_
$5 = std::map with 6 elements = {
  [26] = 0,
  [26] = 0,
  [26] = 0,
  [26] = 0,
  [26] = 0,
  [26] = 0
}
```

Which is obviously not correct in two ways:
1. There shouldn't be multiple entries for 26
2. The entry for 26 should be 3

The threads are all deadlocking on the `[]` access, probably because multiple threads all tried to access an _unset_ entry of 26, wrote 0 to it in the process, and ran into each other. From what I can find `[]` isn't thread-safe (because of situations like this), and `.at()` should be used instead. Since `lebedev_mapping_` is a static member of `SphericalGrid`, all threads within their nice shiny `new SphericalGrid()`s inside `SphericalGrid::build` are still accessing the same global map, and mayhem is happening in the process.

...but the more disturbing problem, IMO, is that `lebedev_mapping_` appears uninitialized up until this point. I suspect in an ideal codebase all entries in `lebedev_mapping_` that would ever be accessed would already exist, so the difference between `[]` and `.at()` is just the slight performance hit in the latter of checking that the entry exists.

There _is_ a function that's supposed to take care of this... https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/cubature.cc#L5125

But from what I can tell, it never gets called. In the attached minimal example I set a breakpoint at `void SphericalGrid::initialize_lebedev()` and it's _never hit_. I also set a breakpoint at `s->order_ = lebedev_mapping_[npoints];` (the line where things hang on an Unlucky Run), and that line is definitely hit (so my breakpoints are firing correctly).

I'm not sure what the consequences of this all are, as I seem to be getting sensible energies (though I suppose I haven't compared to a known good literature benchmark calculated without psi4).

I don't think I'm the first person to run into concerns in this part of the code, though, given this comment:
https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/cubature.cc#L3812

Here's a minimal file to just check that `initialize_lebedev` is never called:
```
set freeze_core 1
set {
basis STO-3G
guess sad
fail_on_maxiter false
dft_nuclear_scheme SBECKE
dft_radial_scheme EM
dft_radial_points 99
dft_spherical_points 590
dft_vv10_radial_points 50
dft_vv10_spherical_points 194
dft_pruning_scheme robust
dft_vv10_postscf False
wcombine False
}
molecule mol {
0 1
C 0.9314972462818318 9.553873448534286 1.4618215727211896
C 0.407816021141855 8.469506309378442 2.18786320317928
C 0.8158282265296265 7.159078273321417 1.778073368593668
C 1.6705486753474763 6.941146237307343 0.7209312466052733
C 2.2362488662446047 8.008509869086277 0.003971742160492786
C 1.911850027633602 9.341079539288122 0.4146901215657727
H 0.614495362639488 10.566900793575911 1.4754165926364617
H -0.36704844148631577 8.614110059906363 3.04421928826377
H 0.4567894628980789 6.157197048912012 2.369426098541459
H 1.8364925212621395 5.872552432013023 0.3910028702339573
H 2.9471948988553818 7.873466587177986 -0.782748863606265
H 2.4659886000145983 10.188822366489614 -0.03294075189628659
units angstrom
no_reorient
no_com
symmetry c1
}
energy('pbe')
```

And below is the full file that is an example of one I ran into the hang with (note that if I run it again, it usually runs just fine... I have only noticed this horrible hang behavior because I am running a _very large_ number of calcs on a daily basis, and eventually get unlucky!). Note that when things hang, they hang on the very first step of the first SCF run and _only_ there. I think that is because if you aren't unlucky and `lebedev_mapping_` gets successfully populated, there's no more hangs that can happen.

(Sorry the below is LONG, I wanted to reproduce exactly all settings that went into a run, in case they are important. It is hard to figure out the "minimal example" as the crashes are rare _enough_ to make it very hard to reproduce them "in captivity").

```
np.set_printoptions(threshold=sys.maxsize)
set DF_INTS_NUM_THREADS = 16
set freeze_core 1
set {
basis def2-tzvppd
guess sad
fail_on_maxiter false
dft_nuclear_scheme SBECKE
dft_radial_scheme EM
dft_radial_points 99
dft_spherical_points 590
dft_vv10_radial_points 50
dft_vv10_spherical_points 194
dft_pruning_scheme robust
dft_vv10_postscf False
}
set {
wcombine False
}
molecule mol {
0 1
@C 0.265781511352271 0.841876264095443 -0.6914658815200431
@C 1.4413385866197737 0.3288357318169406 -1.1836757404873801
@C 2.2289711018973604 -0.6237867174726249 -0.5257029978620363
@C 1.8767103570771173 -0.9962343566137659 0.6986750981164255
@C 0.6545489119459532 -0.6147226134119103 1.2098027282027453
@C -0.21957387152459434 0.2564142719883635 0.5001192004375371
@H -0.3723596175396408 1.4255337233361631 -1.1634648379610923
@H 1.9177612974037643 0.7339585876082476 -2.2134537293183647
@H 3.30111923095952 -0.8660596018835396 -0.9695609764149236
@H 2.5278636235464713 -1.4596675993689474 1.3422341203604324
@H 0.5072122755030857 -0.940559021032073 2.215163414817783
@H -0.998224655511569 0.6769295855057103 1.0319031933299063
@C 1.317102898709397 -4.2559640682725135 -5.346326015162741
@C 2.6128913245771566 -4.000991995558086 -5.963528222785021
@C 2.6860602549649695 -4.080680715147077 -7.357260511037285
@C 1.61182676831893 -4.323741601719452 -8.13762758477727
@C 0.36704188574136254 -4.4498345447883 -7.491026236218776
@C 0.18935870670297217 -4.482238426772551 -6.1101157069039
@H 1.4643016900247157 -4.417830669371555 -4.268504690440918
@H 3.5167101959887903 -4.155258555420228 -5.441815579461341
@H 3.7606934564339944 -3.859729234284167 -7.848947020120773
@H 1.6613567438629124 -4.4495016057397585 -9.191513591480888
@H -0.43042396762295976 -4.715135428367288 -8.131538403210945
@H -0.7857432342736794 -4.7178574202550685 -5.387218656011778
@C -1.626234859047421 4.512413614610493 -9.823046494766203
@C -0.9826706183621148 3.6939244652828753 -8.821257560942346
@C -0.6986319218032849 2.3111056533569103 -9.261042373944903
@C -1.177145703486778 1.829413070447374 -10.546452404078376
@C -1.7690317253158327 2.758224993404836 -11.422010911826131
@C -1.898144381103452 4.0463911069083895 -11.094379169774317
@H -1.7012274142018224 5.627199302988766 -9.77335737394089
@H -0.7354458407999775 3.808094895046909 -7.760202384000629
@H -0.3592431605443381 1.650666094497526 -8.464737898970688
@H -0.8600381404496702 0.8579427307272174 -10.923006258691348
@H -2.0234174460096264 2.2620312875347652 -12.390674641213696
@H -2.3635026438425073 4.6343554516716985 -11.896791070451545
C 0.9314972462818318 9.553873448534286 1.4618215727211896
C 0.407816021141855 8.469506309378442 2.18786320317928
C 0.8158282265296265 7.159078273321417 1.778073368593668
C 1.6705486753474763 6.941146237307343 0.7209312466052733
C 2.2362488662446047 8.008509869086277 0.003971742160492786
C 1.911850027633602 9.341079539288122 0.4146901215657727
H 0.614495362639488 10.566900793575911 1.4754165926364617
H -0.36704844148631577 8.614110059906363 3.04421928826377
H 0.4567894628980789 6.157197048912012 2.369426098541459
H 1.8364925212621395 5.872552432013023 0.3910028702339573
H 2.9471948988553818 7.873466587177986 -0.782748863606265
H 2.4659886000145983 10.188822366489614 -0.03294075189628659
@C -2.883378976118354 3.679902163454515 6.605611804401257
@C -1.4418573642560812 3.600920790635927 6.670521485926086
@C -0.6393020528317843 4.660543897500103 7.199305306584455
@C -1.3266187948685466 5.747101703223244 7.774753886547491
@C -2.724567555535229 5.633863788466681 7.845926916061886
@C -3.515096369644318 4.64011668323281 7.307695334883003
@H -3.522185022496425 2.9755923057599087 6.166222160695254
@H -0.7769743608626567 2.795784357233139 6.267882984917893
@H 0.4237337787110078 4.948400919867275 7.2125183897413745
@H -0.9232262610709561 6.536591753828505 8.324128824246781
@H -3.2319912639558943 6.425323678439093 8.347870777526017
@H -4.670840490698138 4.668965154614611 7.387704241614836
C 2.2081658074790274 -5.3020164768913265 6.6510953531518195
C 3.4850970832998747 -4.830430366645136 6.603095343785539
C 3.6960958670600284 -3.5170046939420203 7.135768650558369
C 2.606915315437915 -2.6915518986780596 7.483946158304935
C 1.3596547084775734 -3.2600808377424353 7.521569539699344
C 1.1711131766388743 -4.555253723319432 7.114249143031859
H 2.116450827322973 -6.447024106625211 6.2912440007746735
H 4.147926021324595 -5.527468935787702 6.298140312113919
H 4.682959113326085 -3.1084423755103856 7.144261556531443
H 2.853922074403954 -1.7713832406320649 7.868036685742114
H 0.3668345387515738 -2.7610843910048732 7.8237465009707545
H 0.1081678166328347 -4.80913158701848 7.0546213578338275
@C -1.7405771438695425 -7.80886257185175 0.5404281767388233
@C -2.6188602604225517 -6.691098033004703 0.5011602693432181
@C -3.714140810909062 -6.576587282189308 1.3810925093268906
@C -3.942255953512009 -7.605544961402358 2.270588825183478
@C -3.03532531969506 -8.58700653050214 2.445310345258381
@C -1.9513792884695083 -8.764665860879806 1.6060394926445083
@H -0.9796992846786388 -7.840283014272942 -0.17744029414230778
@H -2.495309384578162 -5.759810624845935 -0.07447588606303623
@H -4.340889311265438 -5.63947864160487 1.3227958668441364
@H -4.551193864181943 -7.442030783029586 3.213032536935569
@H -3.1972325817701313 -9.21254035719057 3.319783691154892
@H -1.0777792942495945 -9.375803514705414 1.8259033760084107
units angstrom
no_reorient
no_com
symmetry c1
}
set basis def2-tzvppd
data = dict()
mol.set_full_geometry(Matrix.from_list([[0.5022542654449073, 1.590915569963503, -1.306681140595813],
[2.7237351815665702, 0.6214094731249062, -2.2368229698847557],
[4.212144922291332, -1.1787860562025736, -0.9934346888539755],
[3.546468590124386, -1.8826100899427796, 1.3203045855360824],
[1.2369181787496408, -1.1616573819627147, 2.286195821128258],
[-0.4149344813052484, 0.4845527485033083, 0.9450883184937658],
[-0.7036576970203194, 2.6938683185233923, -2.198629899378565],
[3.624043624500105, 1.3869807173967377, -4.182821337943252],
[6.238211251248736, -1.6366154551623187, -1.8322047065489515],
[4.77696992890693, -2.7583719957972828, 2.5364548826091995],
[0.9584922877490646, -1.7773989537967476, 4.186052175296396],
[-1.8863712097657719, 1.279211522262235, 1.9500144225202807],
[2.488963756511477, -8.042606485283144, -10.103091941619423],
[4.937648996861485, -7.5607790984247085, -11.269435077539946],
[5.07591823612626, -7.711368953670002, -13.903207393384701],
[3.045911152463411, -8.17068746090052, -15.377887439428918],
[0.6936086403173598, -8.408968589548671, -14.155987978839677],
[0.35783609498195545, -8.470203051873602, -11.546445275822556],
[2.767129157973372, -8.348490030084381, -8.06630482661384],
[6.64561913009783, -7.852300646752344, -10.28354106590362],
[7.106680671332505, -7.2938311680086105, -14.832360234725817],
[3.13950924120095, -8.408339425930736, -17.36944335867427],
[-0.8133834162821838, -8.910314600134187, -15.366380553945444],
[-1.484839517054919, -8.915458419315543, -10.180367833336788],
[-3.0731384979190186, 8.52722589264645, -18.562867584573294],
[-1.8569783394210484, 6.980505564439329, -16.669760864965355],
[-1.3202229941291759, 4.367356729918863, -17.500833715309945],
[-2.224482988369945, 3.457089671956247, -19.929906630109166],
[-3.3429854666211285, 5.212289827632862, -21.584472415838455],
[-3.5869730252828074, 7.6465709851781725, -20.96533815362649],
[-3.2148538885467297, 10.633865531333795, -18.468968754840013],
[-1.3897912186070827, 7.1962564082241745, -14.664657177429172],
[-0.6788711855737655, 3.1193068418059653, -15.996036345794762],
[-1.6252365421823092, 1.621276791687995, -20.641490286499835],
[-3.823704808747945, 4.274619618775313, -23.414981571239565],
[-4.466372691691263, 8.75766256782587, -22.48167688504687],
[1.7602746813157428, 18.05420424706374, 2.7624424155127625],
[0.7706605891926988, 16.005047335515233, 4.134462252155306],
[1.541691912880146, 13.528697241336232, 3.360071696132801],
[3.1568794742631003, 13.116865379487553, 1.3623626107690086],
[4.225897903707189, 15.133890318935643, 0.007505504920960816],
[3.6128729435857183, 17.65208203760002, 0.7836507563470771],
[1.161227940241227, 19.968548485949185, 2.7881332798114595],
[-0.6936210288798643, 16.278308820606714, 5.752740718121715],
[0.8632069814922737, 11.635416117797726, 4.477566398783909],
[3.4704678951089125, 11.097515749009613, 0.7388883386847691],
[5.5693911947308194, 14.878695501158344, -1.479180976577893],
[4.660043080476915, 19.254083805126786, -0.06224899942322865],
[-5.448796578367559, 6.954007254346888, 12.48279719591343],
[-2.7247155292187717, 6.804754090772794, 12.605458716832036],
[-1.2081057907631072, 8.807151558071153, 13.604715317009585],
[-2.5069461940826483, 10.860448229462223, 14.692155531944536],
[-5.148686488002696, 10.646459583648602, 14.82665306518665],
[-6.642569440294071, 8.76854971761692, 13.809542785134276],
[-6.6559650527771, 5.623054516429932, 11.652471107312186],
[-1.468268747886573, 5.28326673868353, 11.844582222696696],
[0.8007407915165865, 9.351122493395383, 13.62968442543807],
[-1.7446447844863573, 12.352368203223108, 15.730323703929539],
[-6.107578326059717, 12.142102014322907, 15.775189493290956],
[-8.826609299232123, 8.823065427642629, 13.96073770638817],
[4.172828613898477, -10.019359049577828, 12.568748646228057],
[6.585879005168848, -9.128190457034952, 12.478041774546314],
[6.984608919104673, -6.646175650573697, 13.484648438245728],
[4.926355976270059, -5.086295938718023, 14.142608570641505],
[2.5693750230805064, -6.160659927473747, 14.21370645735928],
[2.213083164787955, -8.608181965255541, 13.443982462683802],
[3.999512419878047, -12.183109880381776, 11.88872814465905],
[7.838444165512118, -10.445402451015424, 11.901760284360375],
[8.849510177006636, -5.874104763895764, 13.500697704537037],
[5.393131101647324, -3.34742918654667, 14.868434474560834],
[0.6932168112938935, -5.217693305978344, 14.784738155333894],
[0.20440754893479765, -9.087941596751813, 13.331302279241509],
[-3.2892141006967823, -14.756611605640634, 1.0212612440672273],
[-4.948928650864744, -12.644342755401091, 0.9470556536023668],
[-7.018708920913597, -12.427948798034729, 2.6098865953999857],
[-7.4497840653131195, -14.37239700557793, 4.290791021232555],
[-5.735933553365823, -16.22709057302199, 4.620966842252425],
[-3.687572420474477, -16.562818050920185, 3.034974786431045],
[-1.8513633325344034, -14.81598763653126, -0.335313559402],
[-4.715451333061206, -10.884464610668445, -0.14073902754797188],
[-8.203091935607064, -10.657070118309775, 2.4997219071223564],
[-8.600509943381129, -14.063399990960185, 6.07175152431976],
[-6.0418939362757165, -17.409178187152243, 6.273481969281969],
[-2.0367076889241837, -17.717700841096935, 3.450457310685484]]))
mol.update_geometry()
e, wf = energy('wb97m-v',return_wfn=True,)
```

tallakahath:
Closes #2735 

Previously, `initialize_lebedev` was never called, and in fact was getting optimized out of the module completely upon compilation. When `lebedev_mappping_[]` is then accessed across multiple OpenMP threads, the std::map is empty, and a deadlock can happen where two threads try to access-write (since [key] fills if key is not found), and the slightly slower thread ends up in a Bad State where it thinks there is a value but ends up infinitely looping on the lookup (the program will hang on `[]`).

This only happens once every several thousand runs, and only when running with a high degree of parallelism in a system with many atoms. I cannot induce it in captivity, but I have observed it in the wild.

Anyway, `[]` accesses on std::map aren't thread-safe if you aren't super-duper sure the map is fully filled for all keys you'd ever look up, which *should* be the case if `initialize_lebedev` was ever called anywhere. But it wasn't, and that was Bad.

Now it's called exactly once (thanks, c++11's `call_once`! I do see that this isn't used anywhere else in the code, but I do see mutex is imported in several files, so I don't think I'm adding any new deps here).

The hangs should be gone, though I'll have to churn through another several thousand runs to likely be sure (as, again, it is a very rare kind of hang). This will take me a few days to confirm, but given all debugging efforts point to this being the problem, I'm like 99% confident this will do the trick.

That said, as far as I can tell, besides one print function the resulting order_ that's assigned to is never *used*. Maybe a candidate to be axed in the future?

## Description
Actually invokes initialize_lebedev before accessing lebedev_mapping_ to ensure the mapping has values, and prevents a deadlock when running in parallel

Prevents a nasty, rare hang

## Dev notes & details
See the main PR body

## Questions
- [x] What does `order_` actually do in SphericalGrid? It never appears to be used anywhere except one print function that also appears unused.

## Checklist
- [x] No tests needed -- no observable changes and it's hard to write a test for a very rare race condition :)

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/cubature.cc
+++ b/psi4/src/psi4/libfock/cubature.cc
@@ -3730,13 +3730,6 @@ void MolecularGrid::buildGridFromOptions(MolecularGridOptions const &opt) {
 
     // RMP: Like, I want to keep this info, yo?
     orientation_ = std_orientation.orientation();
-    radial_grids_.clear();
-    spherical_grids_.clear();
-
-    if (opt.namedGrid == -1) {
-        radial_grids_.resize(molecule_->natom());
-        spherical_grids_.resize(molecule_->natom());
-    }
 
 #ifdef USING_BrianQC
     std::vector<std::vector<double>> atomRotations(molecule_->natom());
@@ -3772,11 +3765,6 @@ void MolecularGrid::buildGridFromOptions(MolecularGridOptions const &opt) {
             RadialGridMgr::makeRadialGrid(opt.nradpts, RadialGridMgr::MuraKnowlesHack(opt.radscheme, Z), r.data(),
                                           wr.data(), alpha);
 
-            // RMP: Want this stuff too
-            radial_grids_[A] = RadialGrid::build("Unknown", opt.nradpts, r.data(), wr.data(), alpha, Z);
-            std::vector<std::shared_ptr<SphericalGrid>> spheres;
-            spherical_grids_[A] = spheres;
-
             int currentBlockIndex = -1;
             for (int i = 0; i < opt.nradpts; i++) {
                 int numAngPts = 0;
@@ -3809,8 +3797,6 @@ void MolecularGrid::buildGridFromOptions(MolecularGridOptions const &opt) {
                 }
 #endif
 
-                // RMP: And this stuff! This whole thing is completely and utterly FUBAR.
-                spherical_grids_[A].push_back(SphericalGrid::build("Unknown", numAngPts, anggrid));
                 for (int j = 0; j < numAngPts; j++) {
                     MassPoint mp = {r[i] * anggrid[j].x, r[i] * anggrid[j].y, r[i] * anggrid[j].z,
                                     wr[i] * anggrid[j].w};
@@ -3943,11 +3929,6 @@ void MolecularGrid::buildGridFromOptions(MolecularGridOptions const &opt, const 
 
     // RMP: Like, I want to keep this info, yo?
     orientation_ = std_orientation.orientation();
-    radial_grids_.clear();
-    spherical_grids_.clear();
-
-    radial_grids_.resize(molecule_->natom());
-    spherical_grids_.resize(molecule_->natom());
 
 // Iterate over atoms
 #pragma omp parallel for schedule(static)
@@ -3964,17 +3945,9 @@ void MolecularGrid::buildGridFromOptions(MolecularGridOptions const &opt, const 
             wr[i] = ws[A][i];
         }
 
-        // RMP: Want this stuff too
-        radial_grids_[A] = RadialGrid::build("Unknown", rs[A].size(), r.data(), wr.data(), alpha, Z);
-        std::vector<std::shared_ptr<SphericalGrid>> spheres;
-        spherical_grids_[A] = spheres;
-
         for (size_t i = 0; i < rs[A].size(); i++) {
             int numAngPts = LebedevGridMgr::findNPointsByOrder(Ls[A][i]);
             const MassPoint *anggrid = LebedevGridMgr::findGridByNPoints(numAngPts);
-
-            // RMP: And this stuff! This whole thing is completely and utterly FUBAR.
-            spherical_grids_[A].push_back(SphericalGrid::build("Unknown", numAngPts, anggrid));
 
             for (int j = 0; j < numAngPts; j++) {
                 MassPoint mp = {r[i] * anggrid[j].x, r[i] * anggrid[j].y, r[i] * anggrid[j].z, wr[i] * anggrid[j].w};
@@ -4560,23 +4533,6 @@ void MolecularGrid::print(std::string out, int /*print*/) const {
     Process::environment.globals["XC GRID SPHERICAL POINTS"] = options_.nangpts;
     Process::environment.globals["XC GRID RADIAL POINTS"] = options_.nradpts;
 }
-void MolecularGrid::print_details(std::string out, int /*print*/) const {
-    std::shared_ptr<psi::PsiOutStream> printer = (out == "outfile" ? outfile : std::make_shared<PsiOutStream>(out));
-    printer->Printf("   > Grid Details <\n\n");
-    for (size_t A = 0; A < radial_grids_.size(); A++) {
-        printer->Printf("    Atom: %4zu, Nrad = %6d, Alpha = %11.3E:\n", A, radial_grids_[A]->npoints(),
-                        radial_grids_[A]->alpha());
-        for (size_t R = 0; R < spherical_grids_[A].size(); R++) {
-            double Rval = radial_grids_[A]->r()[R];
-            double Wval = radial_grids_[A]->w()[R];
-            int Nsphere = spherical_grids_[A][R]->npoints();
-            int Lsphere = spherical_grids_[A][R]->order();
-            printer->Printf("    Node: %4zu, R = %11.3E, WR = %11.3E, Nsphere = %6d, Lsphere = %6d\n", R, Rval, Wval,
-                            Nsphere, Lsphere);
-        }
-    }
-    printer->Printf("\n");
-}
 
 GridBlocker::GridBlocker(const int npoints_ref, double const *x_ref, double const *y_ref, double const *z_ref,
                          double const *w_ref, const int max_points, const int min_points, const double max_radius,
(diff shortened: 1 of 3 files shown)


susilehtola:
`lebedev_mapping_` is a member of `SphericalGrid`, so `initialize_lebedev()` should be called in the constructor of SphericalGrid. No need to add mutexes etc.

More worryingly, I also see another `lebedev_mapping_` in cubature.cc
https://github.com/psi4/psi4/blob/ac8f87a1dd3fdda2aabc3318713d6e5ce00e2c70/psi4/src/psi4/libfock/cubature.cc#L5065

loriab:
Thanks for the thorough report and fix. The seemingly extra declaration @susilehtola pointed out confused me, too, though perhaps https://stackoverflow.com/a/17392441 is the answer. The map looks to be from the boost-to-avoid-c++11-standard era, so it could be improved (https://github.com/psi4/psi4archive/blame/1.0.x/src/lib/libfock/cubature.h#L302).

tallakahath:
The "other" lebedev_mapping_: That's just the declaration for it.

The problem is that `SphericalGrid::build` is a _static_ method so the
constructor isn't called when that happens. Someone might yank out the
`new` there and still statically access `lebedev_mapping_` and we are back
where we started. There's no guarantee of construction, so I stapled it
into the one place it's actually used.

(And, again, I don't understand why it's used at all as it never seems to
show up downstream)

Even if it was in the constructor though (which would fire on the _new_) it
would still need the mutex to ensure it's initialized once as
lebedev_mapping_ is also (purposefully) static. No sense doing the rebuild
of the map on every single object instantiation.

susilehtola:
The initialization code is pretty simple, however. One could just replace the `initialize_lebedev` function by an initializer list in https://github.com/psi4/psi4/blob/ac8f87a1dd3fdda2aabc3318713d6e5ce00e2c70/psi4/src/psi4/libfock/cubature.cc#L5065
This would also avoid the need for the mutexes. 

The code is hacky also elsewhere; `lebedev_error` duplicates the same information. For what it's worth, I will be developing a general C++ Becke quadrature library that would eliminate these parts of the code.

tallakahath:
I will admit I considered that, then decided I didn't want to reformat the map into the initializer because it was long and I was lazy.

My bigger concern is still: what was the consequence of `order_` being incorrectly 0 for so long, and who actually uses this property up-stream? The git commit where this was added says `Hack DFT grids to retain indexing data for ISA` but the only files touched were:

```
src/lib/libfock/cubature.cc.REMOVED.git-id
src/lib/libfock/cubature.h
src/lib/libfock/gridblocker.h
```

And I can't find anywhere that seems to use `order_` or the `order()` method of `SphericalGrid`. 

Unfortunately, trying to look further back in the history of `cubature.cc` results in several instances of `psi4/src/psi4/libfock/cubature.cc.REMOVED.git-id` which (I think?) means things were stripped from the repo at some point, and the resulting commit stuff is all assigned to the wrong person (e.g., nearly `all` of `cubature.cc` is assigned to one commit 7e4ecf968e from dgasmith).

`print_details` is the only place the `spherical_grids_` are ever apparently used, and I can comment out the block where they print nicely. _That_ function is only ever used in `v.cc` and only if `print_ > 2`. I don't think this is used much, and is creating unnecessary complexity, and doesn't feature in any tests I can find.

I'm running through tests now to see if just... removing this entirely breaks anything. Short of `print_details` being accessible through some print settings, it doesn't appear that any of this is even exposed on the python side anyway?

susilehtola:
> I'm running through tests now to see if just... removing this entirely breaks anything. Short of `print_details` being accessible through some print settings, it doesn't appear that any of this is even exposed on the python side anyway?

Haha yes spaghetti code!

loriab:
Anyone wanting to look back in psi4 history beyond 2016 (when history was rewritten to shrink the codebase by 90% by getting rid of boost tarballs) would need to use the psi4/psi4archive repo, https://github.com/psi4/psi4archive/blame/1.0.x/src/lib/libfock/cubature.h#L302 .

Yes, acknowledged spaghetti :-(

JonathonMisiewicz:
> My bigger concern is still: what was the consequence of `order_` being incorrectly 0 for so long, and who actually uses this property up-stream? The git commit where this was added says `Hack DFT grids to retain indexing data for ISA` but the only files touched were:
> 
> ```
> src/lib/libfock/cubature.cc.REMOVED.git-id
> src/lib/libfock/cubature.h
> src/lib/libfock/gridblocker.h
> ```
> 
> And I can't find anywhere that seems to use `order_` or the `order()` method of `SphericalGrid`.

The original coder probably added it _anticipating_ it would be useful, but it seems to have never been used. I'm happy to see unused code burn.

tallakahath:
```
100% tests passed, 0 tests failed out of 520
```

> The original coder probably added it _anticipating_ it would be useful, but it seems to have never been used. I'm happy to see unused code burn.

Burn, baby, burn. Patch incoming...

tallakahath:
`RadialGrid` also looks  ripe for burning -- smoketests passed, running full tests now.

tallakahath:
`100% tests passed, 0 tests failed out of 520`

Shipping it.

psi4/src/psi4/libfock/cubature.cc
```diff
@@ -4560,23 +4533,6 @@ void MolecularGrid::print(std::string out, int /*print*/) const {
...
-            double Rval = radial_grids_[A]->r()[R];
-            double Wval = radial_grids_[A]->w()[R];
-            int Nsphere = spherical_grids_[A][R]->npoints();
-            int Lsphere = spherical_grids_[A][R]->order();
-            printer->Printf("    Node: %4zu, R = %11.3E, WR = %11.3E, Nsphere = %6d, Lsphere = %6d\n", R, Rval, Wval,
-                            Nsphere, Lsphere);
-        }
-    }
-    printer->Printf("\n");
-}
```
susilehtola, line 4579: It looks like you are removing printout of useful information. This code should be ported to use the actual data used in the code.

tallakahath: The data in SphericalGrid was never correct to begin with so I'd argue it was never very important (otherwise someone would've reported the obvious printout of 0s by now). I would put good money on nobody ever actually using this print statement in anger (it showed up in v.cc, but again, SphericalGrid has been broken for >5 years and nobody reported the weird print out).

JonathonMisiewicz (dismissed):

tallakahath:
I *think* I have a version that maintains the nice printing functionality (and shoves the necessary info into a struct, and actually uses the nicer `LebedevGridMgr::findOrderByNPoints` function and other `LebedevGridMgr` functions) but I can't actually test if this builds until I get home due to #reasons. I'll shove it upstream on my local branch and see if it breaks in the test builder.

tallakahath:
(Note that I expect this will probably fail to build or run but I cannot debug with an incremental build until late tonight at best, probably tomorrow)

tallakahath:
Clearly I can't get this working right now and I won't be able to touch this from a useful computer until tomorrow. Sorry.

tallakahath:
I now have a version that compiles at home, though I'm not getting the narrowing warning-as-error that the auto-builder is. Still I've patched for it. This *should work* now...

psi4/src/psi4/libfock/cubature.cc
```diff
@@ -4952,250 +4949,4 @@ void OctreeGridBlocker::block() {
...
-        double temp = sin(tau / (npoints + 1.0) * M_PI);
-        double w =
-            M_PI / (npoints + 1.0) * temp * temp * alpha * 2.0 / ((1.0 + x) * (1.0 + x) * sqrt(1.0 - x * x)) * r * r;
-        grid->r_[tau - 1] = r;
-        grid->w_[tau - 1] = w;
-    }
-
-    return std::shared_ptr<RadialGrid>(grid);
-}
-std::shared_ptr<RadialGrid> RadialGrid::build_treutler(int npoints, double alpha, int Z) {
```
hokru, line 5022: This is the only Treutler version with correct Eta values. It does give slightly different results compared to the currently used "Treutler" where Eta is always 1. Though I had no time to try to confirm if this one is fully correct.

tallakahath: Even if it's correct, it was never actually used anywhere. The only use of the `RadialGrid` built was to provide data in `print_details` -- it's not accessed by anything or anyone outside of that function, at least as far as all of the psi4 tests can determine. In addition, the only calls to RadialGrid::build specifically were of the type `RadialGrid::build("Unknown, ...` which never invoked `build_becke` or `build_treutler`.

hokru: Yes, i tested it in a local build. So many parts of cubature.cc are messy unfortunately. Could you move the Eta table to line 2045 as a comment and add a note that the eta mapping is missing, please?
When I added the table long time ago I wasn't aware these routines were never called.

tallakahath: Done

psi4/src/psi4/libfock/cubature.cc
tallakahath: Printing has been restored by shoving the desired data into some simple structs

psi4/src/psi4/libfock/cubature.h
```diff
@@ -56,11 +54,31 @@ class Options;
...
+};
+
+struct RadialGrid {
+    int npoints_;
+    /// Alpha scale (for user reference)
+    double alpha_;
+    /// Nodes (including alpha)
+    double* r_;
+    /// Weights (including alpha and r^2)
+    double* w_;
```
susilehtola, line 77: Please do not introduce raw pointers, when their use can easily be avoided. You can just duplicate the data into `std::vector<double>`.

psi4/src/psi4/libfock/cubature.cc
```diff
@@ -4564,15 +4579,17 @@ void MolecularGrid::print_details(std::string out, int /*print*/) const {
...
-        printer->Printf("    Atom: %4zu, Nrad = %6d, Alpha = %11.3E:\n", A, radial_grids_[A]->npoints(),
-                        radial_grids_[A]->alpha());
-        for (size_t R = 0; R < spherical_grids_[A].size(); R++) {
-            double Rval = radial_grids_[A]->r()[R];
-            double Wval = radial_grids_[A]->w()[R];
-            int Nsphere = spherical_grids_[A][R]->npoints();
-            int Lsphere = spherical_grids_[A][R]->order();
-            printer->Printf("    Node: %4zu, R = %11.3E, WR = %11.3E, Nsphere = %6d, Lsphere = %6d\n", R, Rval, Wval,
-                            Nsphere, Lsphere);
+        printer->Printf("    Atom: %4zu, Nrad = %6d, Alpha = %11.3E:\n", A,
```
JonathonMisiewicz, line 4582: Just to be clear on this point: `radial_grids_` exists now as a convenience struct for the sake of this printout, but the data being printed actually existed and is used elsewhere. Is that right, or am I missing something?

tallakahath: Yes, `radial_grids_` is a convenience struct. I checked, and it looks like, e.g., `r` and `wr` *do* get used downstream, or at least, used in the construction of the `MassPoint`s that get `push_back`'d into the `atomic_grids_` of `MolecularGrid`. I suppose I haven't checked what happens, long-term, with those, but hopefully something...

psi4/src/psi4/libfock/cubature.h
tallakahath: Fixed (I think? My C++ is very rusty, if you haven't noticed...).

psi4/src/psi4/libfock/cubature.cc
JonathonMisiewicz: Okay, in that case, does it make more sense to not print this at all but rely on the user to call `MassPoint.print()` or `MassPoint.print_details()` when necessary?

tallakahath: I looked into this and my conclusion was that grid _pruning_ also acts on the `MassPoint`s, so recovering things like the initial lebedev grid or number of grid points isn't possible. The `MassPoint`s don't really carry around sufficient information to reconstruct the state printed in `print_details`. But I'm not very familiar with the codebase OR what's happening under-the-hood here, so another set of eyes to check my assertion that this Cannot Be Done would be appreciated.

However, if ya'll are OK ditching `print_details` all together, then all these problems go away. The replacement structs were added b/c of @susilehtola 's requests to not break the existing print information.

JonathonMisiewicz: Any information that isn't actually used except in printing, I'm okay removing.

tallakahath:
I've made a new PR #2743 to JUST fix the hang bug, and then we can keep bikeshedding here on what we actually want to do about `RadialGrid` and `SphericalGrid` but have a fix in the codebase that actually stops the nasty hang that took me down this rabbit hole in the first place.

- @susilehtola this initializes with an initializer list like you requested
- @hokru this loses NO information over the prior state-of-the-world like you suggested was desirable (if I was reading correctly)
- @JonathonMisiewicz I guess you won't like it because spaghetti didn't go away but it's short and sweet, at least.

Obviously this patch would undo all of that patch but I think the scope of this discussion exceeded the original PR statement.

JonathonMisiewicz:
This is turning into a spaghetti PR, so breakup appreciated.

JonathonMisiewicz:
Thanks for getting the most important part of the PR cleared away. As far as I can see, the next steps are:
1. Fix the merge conflicts (annoying rebase)
2. Determine what information in `print_details` is actually used. I want to retain print information that is actually used, but I can do away with the rest. Once we have that, we can assess if keeping a modified `RadialGrid` is the best way to keep the useful print information. (I think we've established that `r` and `wr` are used, but `order` is not?)

This part of the code is not pretty, and I commend you for dealing with it at all.

tallakahath:
> 1. Fix the merge conflicts (annoying rebase)

I'm not sure how much rebasing you want to do, but I think things are back to the state-of-the-world that we were discussing prior to #2743 

> 2. Determine what information in `print_details` is actually used. I want to retain print information that is actually used, but I can do away with the rest. Once we have that, we can assess if keeping a modified `RadialGrid` is the best way to keep the useful print information. (I think we've established that `r` and `wr` are used, but `order` is not?)

"Used" is a funny word here. No tests when this print is removed, so it's not _so_ important that someone wrote a test to preserve it. Furthermore, the `order_` was _always_ wrong. The other elements weren't, though, but you have to set a pretty high print level to actually see anything (`v.cc` requires `print_ > 2`). 

If by "used" you mean that the piece of information represented in the print is used downstream: `r` and `wr` are the primary candidates here. That said, they aren't used as-is -- they're dotted through other elements found by a lookup in `LebedevGridMgr` and `RadialGridMgr` to make `MassPoints` (see https://github.com/psi4/psi4/blob/9db9100564eaef5d02c952817a73c397db9d657b/psi4/src/psi4/libfock/cubature.cc#L3814-L3822)

I'm not sure if we care about diagnostic print info on the intermediaries (the radii of shells around atoms we care about, and the number of angular points we end up picking for each shell). Maybe we do? If so, a struct seems like the most reasonable solution. I suppose, all-in-all, it's not so much data to carry around. Appending that data onto every `MassPoint` however feels like overkill (and that's the only other place to put it). 

But we definitely should clean up/remove the rest of the old `RadialGrid` and `SphericalGrid`, as these seem to only cause confusion about where changes should be made. `RadialGridMgr` and `LebedevGridMgr` are the classes of Actual Importance.

JonathonMisiewicz:
1 sounds good.

On 2., yes, agreed that high-level, having both `Grid` and `GridMgr` is confusing, and that's the issue to deal with. **If** the`Grid` classes are only used for print purposes, then while the print functionality belongs somewhere, it's not there. Keeping this old code around doesn't make that task easier doesn't make print setup in the _correct_ place any easier. We should then just be rid of the classes entirely, printing included.

Are the Grid class used for something other than printing?

tallakahath:
> Are the `Grid` classes used for something other than printing?

`Grid` classes are *only* used for printing. Hence why I moved things to skinnier structs, so that it's clear to future devs that "nothing goes here but storage" (and maybe even with a big comment about "THIS CONTENT IS NOT REFERENCED EXCEPT BY A PRINT". There's no reasonable way I can see to migrate this into the `MassPoints`.

I've been quiet b/c I am hoping the other primary devs will chime in here -- I'm not sure how to move forwards given conflicting desires (and given that the immediate urgency of "this causes program hangs" has been fixed).

hokru:
This `print_details` is more a debug printing for devs than for users. The quadrature points themselves can be printed with the `BlockOPoints` objects if needed. The `r` and `wr` are intermediates.

I would be in favour of removing the classes at the cost of removing the detailed printing.
It's better to re-think how to print this information instead of trying to fit it into these classes.

PR #2741, opened 2022-10-09
Coalesce PSIO system call macros

TiborGY:
## Description

Definitions of various IO system call macros are currently replicated and/or scattered across files. These macros were added to paper over differences between Windows and Linux/Mac.
This PR coalesces all of them into `psio.h`.

None

## Dev notes & details

- [x] `SYSTEM_WRITE`, `SYSTEM_READ`, `SYSTEM_LSEEK`, `SYSTEM_OPEN`, `SYSTEM_CLOSE`, `SYSTEM_UNLINK`, `PSIO_OPEN_OLD_FLAGS`, `PSIO_OPEN_NEW_FLAGS `and `PERMISSION_MODE `are now only defined in `psio.h`, which is already included by pretty much all PSIO-related files anyways.
- [x] Same goes for the Windows specific `io.h` include, and the non-Windows `unistd.h`.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libpsio/toclen.cc
+++ b/psi4/src/psi4/libpsio/toclen.cc
@@ -34,11 +34,9 @@
 #include <cstdio>
 #ifdef _MSC_VER
 #include <io.h>
-#define SYSTEM_READ ::_read
 #define SYSTEM_WRITE ::_write
 #else
 #include <unistd.h>
-#define SYSTEM_READ ::read
 #define SYSTEM_WRITE ::write
 #endif
 #include <cstdlib>
--- a/psi4/src/psi4/libpsio/toclen.cc
+++ b/psi4/src/psi4/libpsio/toclen.cc
@@ -32,13 +32,6 @@
  */
 
 #include <cstdio>
-#ifdef _MSC_VER
-#include <io.h>
-#define SYSTEM_WRITE ::_write
-#else
-#include <unistd.h>
-#define SYSTEM_WRITE ::write
-#endif
 #include <cstdlib>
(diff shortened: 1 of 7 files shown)


lothian (approved):

JonathonMisiewicz (approved):

psi4/src/psi4/libpsio/toclen.cc
```diff
@@ -32,17 +32,6 @@
  */
 
 #include <cstdio>
-#ifdef _MSC_VER
-#include <io.h>
-#define SYSTEM_LSEEK ::_lseek
```
loriab, line 37: Only possible trouble I see is here with `_lseek` vs `_lseeki64`. Docs say: `_lseek returns the offset, in bytes, of the new position from the beginning of the file. _lseeki64 returns the offset in a 64-bit integer.` Harmless?

TiborGY: Eagle-eyed review! This is something I have not noticed until now.
`volseek.cc` used `_lseeki64` on Windows, while `toclen.cc` used `_lseek`.
Re-drafting this until I have had time to ponder the docs and the impact on the code.

TiborGY: In `volseek.cc` the return value is never stored, only compared with -1, so 32/64 bit should not matter.
In `toclen.cc`, the return value is stored in an `auto` variable, which is then only compared with -1.
So I think it is harmless.

loriab (approved):
nice cleanup, thank you!

PR #2743, opened 2022-10-10
Initialize lebedev_mapping_ and make use thread-safe -- and do ONLY THAT

tallakahath:
`initialize_lebedev` wasn't ever actually called. Since we're using c++11 elsewhere, we can just use an initializer_list to build `lebedev_mapping_` and skip all this `initialize_lebedev` nonsense. Also changes the technically-thread-unsafe `[]` to a `.at()` so this can't be re-broken later.

Now, it turns out that there's a lot more ado about `SpherialGrid` and `RadialGrid` and cubature.cc in general, but, we're discussing that over in #2736 .

This PR's purpose is JUST to fix the nasty bug that could cause parallel code to hang (if multiple threads tried to touch `lebedev_mapping_` at once, and write a value because `[]` access fills-when-missing, one or more thread could get stuck forever in a tree-that-has-become-a-loop).

See #2736 for more details on who/what/where/when/why.

### DIFF
--- a/psi4/src/psi4/libfock/cubature.cc
+++ b/psi4/src/psi4/libfock/cubature.cc
@@ -5062,7 +5062,13 @@ std::shared_ptr<RadialGrid> RadialGrid::build_treutler(int npoints, double alpha
 }
 
 /// Grid npoints to order map
-std::map<int, int> SphericalGrid::lebedev_mapping_;
+std::map<int, int> SphericalGrid::lebedev_mapping_ = {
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz (approved):

susilehtola (approved):
Great, awesome!

loriab (approved):
great, thanks!

zachglick (approved):
Thanks for finding and fixing this very tricky bug!

PR #2756, opened 2022-10-20
Improve PSIO error messages for READ, WRITE and LSEEK

TiborGY:
## Description

This PR builds upon the functions added in #2711 and #2700 to provide more informative error messages if a `SYSTEM_READ`, `SYSTEM_WRITE` or `SYSTEM_LSEEK` fails, including the OS-provided error message, which may have clues for the cause of the error (disk is full, no permission, etc).

- [x] Error messages related to some IO errors are now more specific and detailed.

## Dev notes & details

- [x] New function: `psio_compose_err_msg`. This helps composing PSIO error messages without too much code duplication. It can optionally take an errno value and tack on the error message provided by the OS.
- [x] `psio_volseek` has been reworked to handle any errors internally by calling `psio_error`, instead of returning -1 on error. Appropriate error message generation and some `const` qualifiers were added.
- [x] `PSIO::rw` has been adapted to the changes to `psio_volseek` and error message generation was added after each read/write/seek.
- [x] Functions in `toclen.cc` are now also using `psio_compose_err_msg` to generate messages.

## Checklist
- [x] Tests run by the CI are passing.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libpsio/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/CMakeLists.txt
@@ -2,6 +2,7 @@ list(APPEND sources
   aio_handler.cc
   change_namespace.cc
   close.cc
+  compose_err_msg.cc
   done.cc
   decode_errno.cc
   error.cc
--- a/psi4/src/psi4/libpsio/compose_err_msg.cc
+++ b/psi4/src/psi4/libpsio/compose_err_msg.cc
@@ -1,0 +1,54 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2022 The Psi4 Developers.
+ *
+ * The copyrights for code used from other parties are included in
+ * the corresponding files.
+ *
+ * This file is part of Psi4.
+ *
+ * Psi4 is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, version 3.
+ *
+ * Psi4 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+/*!
+ ** \file
+ ** \ingroup PSIO
+ */
+#include <cstring>
+#include <optional>
+#include "psi4/libpsio/psio.h"
+
+namespace psi {
+/// @brief Composes an error message explaining that an IO system call has failed, with the error message from the OS
+/// (based on errno), some context from the caller of this function and the unit number. Callers should save the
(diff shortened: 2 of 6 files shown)


lothian (approved):

loriab (approved):

JonathonMisiewicz:
Can you provide a before-and-after example of what error messages look like?

TiborGY:
> Can you provide a before-and-after example of what error messages look like?

Hmm, yes I reckon if I ran something with a ~50 MB ramdisk as the scratch drive, that should be a reliable way to trigger a write failure. My immediate-term schedule just got a lot busier, so I might not get around to doing this for a week or so.

TiborGY:
Update: testing is currently made very difficult by an issue with exception messages not propagating out of the threaded neighborhood of https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libpsio/aio_handler.cc

TiborGY:
OK, I have finally managed to get a pair of builds (current master and this PR's branch) that do not entirely swallow the exception messages for write errors in the PK integral calculation. For me, it took building Psi4 entirely from source, with nothing from conda at all.

stdout/stderr before this PR:

```
PSIO_ERROR: unit = 34, errval = 12
terminate called after throwing an instance of 'psi::PsiException'
  what():
Fatal Error: PSIO_ERROR: 12 (error writing to file)

Error occurred in file: /home/gytibor/psi4/psi4/src/psi4/libpsio/error.cc on line: 131

psi::PSIO::rw(unsigned long, char*, psi::psio_address, unsigned long, int)
psi::PSIO::write(unsigned long, char const*, char*, unsigned long, psi::psio_address, psi::psio_address*)

Aborted (core dumped)
```
stdout/stderr after this PR:
```
PSIO_ERROR: unit = 34, errval = 12
terminate called after throwing an instance of 'psi::PsiException'
  what():
Fatal Error: WRITE failed. Only some of the bytes were written!
Error writing the first partial page, unit 34.
PSIO_ERROR: 12 (error writing to file)

Error occurred in file: /home/gytibor/psi4/psi4/src/psi4/libpsio/error.cc on line: 131

psi::PSIO::rw(unsigned long, char*, psi::psio_address, unsigned long, int)
psi::PSIO::write(unsigned long, char const*, char*, unsigned long, psi::psio_address, psi::psio_address*)

Aborted (core dumped)
```
The reason why there is no "disk is full" error message, is because `SYSTEM_WRITE` only reports a failure and sets errno if the disk is already full when it is called. As long as it can write at least one out of _however many_  bytes it does not set errno.

I both cases there is no error message at all in the output file. The current theory is that the `python` process gets aborted by a SIGABRT and cannot write the message to the output file. I think chasing that down is beyond the scope of this PR.

hokru:
It does not necessarily always cause an abort of the python process.
My test fails during libmints writing out integrals at the end of an SCF and gives a normal error message. (`cc11` test with cc-pVQZ basis set in a 50MiB RAMDISK)

```
[...]

! Fatal Error: PSIO_ERROR: 12 (error writing to file)                              !
! Error occurred in file: /Users/kruse/qc/psi4/psi4/src/psi4/libpsio/error.cc on   !
!     line: 131                                                                    !
! The most recent 5 function calls were:                                           !

    Psi4 stopped on: Sunday, 06 November 2022 04:44PM
[...]
```

I would suggest to merge and see how it behaves for our users.

hokru (approved):

JonathonMisiewicz:
@TiborGY, are you still happy to merge this in? If so, no objections to getting this PR in.

TiborGY:
Yes. It is less of an improvement than what I hoped it would be, but an improvement nevertheless.

PR #2758, opened 2022-10-22
Modify Instability Following Step

JonathonMisiewicz:
## Description
The old instability following procedure Trotterized a matrix exponential. Now we compute it (more or less) exactly.

Obligatory @susilehtola ping.

- [x] Modified the orbital instability following procedure. This should be a minor change for most applications: contact developers if you notice a problem.

## Dev notes & details

- [x] Change the algorithm we use to approximate a matrix exponential when instability following from a Trotter approximation to that of the build-in `expm` function.
- [x] Added documentation for the rationale for the orbital rotation step.

## Questions
- [ ] **Warning!** None of our tests cover instability following. I've run a check that the old and new codes mostly agree for the particular instability I'm trying to track. Is this an issue?

## Status
- [x] Ready for review
- [ ] Ready for merge after discussion

### DIFF
--- a/psi4/src/psi4/libscf_solver/stability.cc
+++ b/psi4/src/psi4/libscf_solver/stability.cc
@@ -253,31 +253,31 @@ SharedMatrix UStab::analyze() {
 }
 
 void UStab::rotate_orbs(double step_scale) {
+    // The energy is a function of orbital rotation angle, with period 2 pi. At theta = 0, we're at a local maximum.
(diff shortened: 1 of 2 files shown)


susilehtola (approved):
Not sure how expm works, but assuming it's correct this is surely an improvement since the orbitals should stay better orthonormal.

hokru (approved):

PR #2759, opened 2022-10-25
Fix crash on CBS finite-difference with zero dipole moment

philipmnel:
## Description
I recently discovered a bug where CBS finite-difference Hessians will crash on systems with zero dipole moment. This is due to `CompositeComputer` not loading a dipole with entries all exactly zero into qcvars. This fix checks that CBS results are not None rather than have any non-zero entries before loading into qcvars. I have verified that this fixes the crash and that e.g. non-existing Hessians are stiil not loaded into result qcvars.

- [x] Fixed a bug causing CBS extrapolated finite-difference Hessians to crash on systems with zero dipole moment.

## Dev notes & details

- [x] `CompositeComputer` loads variables with zero'd entries in qcvars.
- [x] `FiniteDifferenceComputer` checks for dipoles when doing finite-difference by gradients.

## Checklist
- [x] cbs tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/driver_cbs.py
+++ b/psi4/driver/driver_cbs.py
@@ -1702,24 +1702,24 @@ class CompositeComputer(BaseComputer):
             qcvars[f"CBS {self.cbsrec[dc]['d_stage'].upper()} TOTAL ENERGY"] = self.cbsrec[dc]["d_energy"] - self.cbsrec[dc + 1]["d_energy"]
 
         G0 = assembled_results["gradient"]
-        if np.count_nonzero(G0):
+        if G0 is not None:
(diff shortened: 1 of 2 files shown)


JonathonMisiewicz (approved):

loriab:
I bet there's a change in pint. pass/fail seems correlated with pint 0.17/0.20

loriab (approved):
thanks, those `count_nonzero`s look like leftovers from when all arrays were filled, whether available or not.

zachglick (approved):
Good catch!

PR #2762, opened 2022-10-26
CompositeJK Part 1: DFJLinK

davpoolechem:
## Description
We have seen a very large number of improvements to the JK code as of late. Many of these improvements involve the implementation (partial or full) of separate J or K construction algorithms, with examples including LinK in https://github.com/psi4/psi4/pull/2359, COSX and Direct DF-J in https://github.com/psi4/psi4/pull/2567, and the first step to CFMM in https://github.com/psi4/psi4/pull/2665. That said, such algorithms pose a problem for Psi4 - Psi4 doesn't have a real framework for supporting such algorithms properly. Psi4's JK class hierarchy is specifically designed at the moment for coupled JK build algorithms. Separate J and K algorithms, then, are either hard-locked to each other (as in DFJCOSK) or bolted onto the side of already-existing JK classes (as in LinK with DirectJK). Such a layout leaves no flexibility in combining separate J and K algorithms, a problem which will only get worse as new separate J and K algorithms like CFMM are added. Additionally, the scattering of separate J and K build algorithms throughout the JK class in Psi4 in an non-standardized fashion is unfortunate from a code design perspective.

The goal of the CompositeJK project is, ultimately, to create a new JK class (_CompositeJK_) that offers a proper framework for combining separate J and K algorithms for use in JK construction. Such a class will provide a home for any new separate J and K algorithm that gets added to Psi4 in the future (such as the in-progress CFMM). Additionally, the CompositeJK class will be designed in such as a way to enable arbitrary combinations of any available separate J and K build algorithm through a single JK subclass. This arbitrary mixing and matching is planned to be accesible in a user-friendly way.

DFJLinK is the first step to this process. DFJLinK, itself, is a new JK subclass that combines the integral-direct density fitted J build algorithm from DFJCOSK with the LinK algorithm previously available in DirectJK. LinK was previously available for use in Psi4 via DirectJK, but was always disrecommended due to its combination with a normal DirectJK implementation for J construction. In conjunction with Direct DF-J to form the Coulomb matrix, LinK performs much better. Some benchmarks comparing the Direct DF-J + LinK algorithm combination to other JK options in Psi4, for water clusters of varying sizes at the cc-pVDZ basis set, are attached:

Timing Plots of Water Clusters at cc-pVDZ

Small, non-diffuse basis sets are the specialty of Direct DF-J + LinK, and it reflects here, with DFJLinK outperforming the other integral-direct JK options for most of the water clusters, and being the fastest JK option tested overall for large water clusters. For specific comparisons to Direct DF-J + COSX, the closest current analogue in Psi4, DFJCOSK performs better for larger basis sets or more complex systems, while DFJLinK performs better for smaller basis sets and less complex systems from my testing. It may be worth noting that the DFJLinK in the listed benchmarks used a different implementation of Direct DF-J compared to the one in this PR, but I don't expect the results would change much because of that.

The full of CompositeJK is planned to be done in 3 PRs. This PR is the first one. The second PR in the CompositeJK process is the combining of the DFJLinK and DFJCOSK subclasses into the initial "CompositeJK" subclass and the implementation of the planned CompositeJK User API/user-facing end. The final CompositeJK PR will be the refactoring of the CompositeJK backend to truly support arbitrary mixing and matching of J/K build algorithms.

- [x] The DO_LINK option has been removed.
- [x] The SCF_TYPE option "LINK" has been added, enabling use of the integral-direct density-fitted J + Linear Exchange K algorithm combination for JK computations.

## Dev notes & details
- [x] The new JK subclass "DFJLinK" has been added to Psi4. DFJLinK combines the integral-direct density fitted formation of J with the linear-scaling formation of the exact K through the LinK algorithm. DFJLinK supports both density screening and incremental Fock construction, using machinery similar to the DirectJK class to do so.
- [x] The computation of LinK, and all references thereof, have been removed from DirectJK. This includes, e.g., removal of the build_linK() function from DirectJK entirely.
- [x] All relevant tests, samples, and documentation (particularly those pertaining to LinK) have been updated to reflect the two previous changes.

## Questions
- [ ] How do we feel about having the integral-direct DF-J implementation in two separate JK subclasses? The plan is to remove this in the second CompositeJK PR, when DFJLinK and DFJCOSK are combined, but is the dual implementation of DF-J something people would take issue with?
- [ ] Do we want to the update the definition of num_computed_shells() for DFJLinK? For now, I have this set to the default JK definition for DFJLinK, and the issue is that, since there are shell triplet counts from DFJ and shell quartet counts from LinK, there would have to be a slightly more expansive change to how num_computed_shells() works in the JK class hierarchy to support DFJLinK. I'm thinking updating this will be a problem for the second CompositeJK PR, but I'm curious to see what others have to say.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


davpoolechem:
Looking at the CI failure logs, it seems that the issue is with the package "Pint":

`ImportError: cannot import name 'quantity' from 'pint' (C:\Miniconda\envs\baseenv\lib\site-packages\pint\__init__.py)

This particular error seems to commonly show up for the failed CI tests.

psi4/src/psi4/libfock/DFJLinK.cc
```diff
@@ -0,0 +1,932 @@
...
+        outfile->Printf("    J tasked:           %11s\n", (do_J_ ? "Yes" : "No"));
+        outfile->Printf("    K tasked:           %11s\n", (do_K_ ? "Yes" : "No"));
+        outfile->Printf("    wK tasked:          %11s\n", (do_wK_ ? "Yes" : "No"));
+        if (do_wK_) outfile->Printf("    Omega:              %11.3E\n", omega_);
+        outfile->Printf("    Integrals threads:  %11d\n", nthreads_);
+        outfile->Printf("    Memory [MiB]:       %11ld\n", (memory_ *8L) / (1024L * 1024L));
+        outfile->Printf("    Incremental Fock :  %11s\n", (incfock_ ? "Yes" : "No"));
+        outfile->Printf("    Screening Type:   %11s\n", screen_type.c_str());
+        outfile->Printf("    J Screening Cutoff: %11.0E\n", cutoff_);
+        outfile->Printf("    K Screening Cutoff: %11.0E\n", linK_ints_cutoff_);
```
loriab, line 156: I put some comments in one of the output.ref's. They belong here. :-)

psi4/src/psi4/libfock/DFJLinK.cc
```diff
@@ -0,0 +1,932 @@
...
+    if (density_screening_) {
+        for (auto eri_computer : eri_computers_["4-Center"]) {
+            eri_computer->update_density(D_ref);
+	}
+    }
+
+    // => Perform matrix calculations <= //
+    
+    // Direct DF-J
+    if (do_J_) {
```
loriab, line 266: if you haven't already, it might be a useful sanity check to `set_do_J(false)` and same for K. Just a run -- no need for a test, imo.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -438,9 +436,6 @@ def scf_iterate(self, e_conv=None, d_conv=None):
                 if incfock_performed:
                     status.append("INCFOCK")
                 
-                if linK_performed:
-                    status.append("LINK")
```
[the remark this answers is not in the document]
davpoolechem: For sure! It's nice having a unified API for determining which SCF algorithm to use.

loriab:
This PR LGTM. I'm a little iffy about naming a SCF_TYPE based only on its exchange algorithm, but (1) it's in line with COSX and (2) that's an issue that I think we're kicking ahead to the actual J & K separation. There's big code changes, but from my gross comparisons, they're code moving around (collecting J from COSK and K from Direct) in new JK wrappings.

>  How do we feel about having the integral-direct DF-J implementation in two separate JK subclasses? The plan is to remove this in the second CompositeJK PR, when DFJLinK and DFJCOSK are combined, but is the dual implementation of DF-J something people would take issue with?

No problem by me for temporary dual implementation. If there's any chance this could be forgotten, please file a issue just as a reminder.

tests/linK-1/input.dat
```diff
@@ -1,6 +1,6 @@
 #! RHF Linear Exchange Algorithm test for water
 
-ref_energy = -76.04125669409474 # TEST (Compare to CSAM Screening, without LinK)
+ref_energy = -76.0412801462671837 # TEST, compared to previously-devleoped DDFJ+LinK impl
```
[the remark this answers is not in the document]
davpoolechem: That is correct!

psi4/src/psi4/libfock/jk.h
```diff
@@ -789,28 +783,6 @@ class PSI_API DirectJK : public JK {
...
-                  std::vector<SharedMatrix>& K);
-
-    /**
-     * @brief The standard J and K matrix builds for this integral class
-     * 
-     * @param ints A list of TwoBodyAOInt objects (one per thread) to optimize parallel efficiency
-     * @param D The list of AO density matrices to contract to form J and K (1 for RHF, 2 for UHF/ROHF)
-     * @param J The list of AO J matrices to build (Same size as D, 0 if no matrices are to be built)
-     * @param K The list of AO K matrices to build (Same size as D, 0 if no matrices are to be built)
-     */
```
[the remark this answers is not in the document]
davpoolechem: Ahhhhhh, nice catch. Yeah, I must have removed this accidentally while getting rid of build_linK. I will readd it!

tests/linK-2/output.ref
```diff
@@ -212,21 +211,26 @@ Scratch directory: /scratch/jiang/
...
+  ==> DFJLinK: Density-Fitted J and Linear Exchange K <==
 
+    J tasked:                   Yes
+    K tasked:                   Yes
+    wK tasked:                   No
+    Integrals threads:            1
+    Memory [MiB]:                 0
+    Incremental Fock :          Yes
+    Screening Type:       DENSITY
+    Screening Cutoff:       1E-12
```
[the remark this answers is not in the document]
davpoolechem: This is a good idea as well, cleaning up the output. I will do this, too!

davpoolechem:
> This PR LGTM. I'm a little iffy about naming a SCF_TYPE based only on its exchange algorithm, but (1) it's in line with COSX

Thanks! Yeah, this precedence is the big reason I named the SCF_TYPE for DFJLinK the way I did.

> and (2) that's an issue that I think we're kicking ahead to the actual J & K separation. 

This is correct. I plan on adding a new SCF_TYPE naming scheme for CompositeJK in the next CompositeJK PR.

> There's big code changes, but from my gross comparisons, they're code moving around (collecting J from COSK and K from Direct) in new JK wrappings.

This is also correct. It's basically a refactoring of the code to combine the DFCOSK DFJ with the old DirectJK LinK into a new, separate JK class.

> No problem by me for temporary dual implementation. If there's any chance this could be forgotten, please file a issue just as a reminder.

Awesome, thanks! This dual implementation is something I plan on removing in the next CompositeJK PR, as well.

psi4/src/psi4/libfock/DFJLinK.cc
davpoolechem: Good point! I will do this.

psi4/src/psi4/libfock/DFJLinK.cc
davpoolechem: So I think I may have missed the commit where the comments in the output.ref were added.

tests/linK-2/output.ref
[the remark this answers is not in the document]
davpoolechem: Should be done!

psi4/src/psi4/libfock/DFJLinK.cc
davpoolechem: Interesting... adding set_do_J(false) to the constructor causes a segfault, but adding set_do_K(false) to the constructor does not. I shall explore this further.

psi4/src/psi4/libfock/DFJLinK.cc
loriab: The output.refs look good to me now, so whatever happened, I think it's resolved. Thanks for the fix.

davpoolechem: All right, excellent; thanks!

psi4/src/psi4/libfock/DFJLinK.cc
loriab: Agreed on defer -- sounds like a universal weakness and probably not one likely to be hit. Thanks for investigating.

JonathonMisiewicz:
Conceptual questions before I start a proper review: is it possible to have a J and K algorithm where implementing the two _in a single algorithm_ is more efficient than implementing both in a mix-and-match fashion? If so, how are going to take advantage of the efficiency gains from combining them?

davpoolechem:
> Conceptual questions before I start a proper review: is it possible to have a J and K algorithm where implementing the two _in a single algorithm_ is more efficient than implementing both in a mix-and-match fashion? If so, how are going to take advantage of the efficiency gains from combining them?

Assuming I'm understanding this question correctly, implementing both J and K together in a single algorithm (as done in DirectJK) is generally more efficient than mix-and-match separate J and K algorithms for smaller systems, but less so for larger systems. This is because, generally, algorithms designed to build J and K separately exploit properties specific to those matrices (e.g. the exponential decay of the Density matrix with distance between basis function pairs in Exchange contributions) to lower algorithmic scaling, with the trade-off of a higher prefactor due to recomputation of integrals across separate J and K algorithms. 

If we're talking about, say, trying to compute parts of one matrix within the algorithm meant to compute the other (say, partially building J within the LinK algorithm and building the rest of J in a separate algorithm), this may be possible, and I believe Ochsenfeld discusses this possibility shortly in his LinK paper. But, I'm not so sure of the benefits of doing so, if there are any benefits at all compared to something like DirectJK. Additionally, one of the big benefits of CompositeJK (effective J or K construction in cases where only one such matrix is needed, e.g., DFT with GGA functionals) would be lost.

JonathonMisiewicz:
Wording this another way: what JK features is CompositeJK replacing, and what is it leaving be? Are there J-K combinations that we're leaving be because integrating them with CompositeJK would make things worse?

davpoolechem:
> Wording this another way: what JK features is CompositeJK replacing, and what is it leaving be? Are there J-K combinations that we're leaving be because integrating them with CompositeJK would make things worse?

Ahhhhhhhh, thanks for the clarification! I was worried I may have misinterpreted your original question.

What CompositeJK will be replacing specifically that is already in Psi4 is the DFJCOSK class and the implementation of LinK in DirectJK. Other expansions that will be made to CompositeJK, such as CFMM, are not yet in Psi4 proper. CompositeJK will not touch other JK algorithms present in Psi4, such as DirectJK or the conventional DFJK algorithms.

The original plan for CompositeJK, from many months ago, was to completely redo the JK hierarchy so that any J algorithm in Psi4 could be combined with any K algorithm. However, I was quickly convinced that this was a bad idea, because certain combinations of J and K from the entire Psi4 option list would not perform well. For example, we can see this even now, where the current LinK implementation is always run with DirectJK's formuation of J and thus displays poor performance. The current idea for CompositeJK, then, is that it will he a hub for the implementation of separate J and K build algorithms designed specifically for the construction of those matrices; e.g., Direct DF-J, LinK, COSX, CFMM.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -71,7 +71,12 @@ JK::~JK() {}
 std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary,
                                  Options& options, std::string jk_type) {
 
-    if (options.get_str("SCREENING") == "DENSITY" && !(jk_type == "DIRECT" || options.get_bool("DF_SCF_GUESS"))) {
+    bool do_density_screen = options.get_str("SCREENING") == "DENSITY";
+    bool do_df_scf_guess = options.get_bool("DF_SCF_GUESS");
+    
+    bool can_do_density_screen = (jk_type == "DIRECT" || jk_type == "LINK");
+
+    if (do_density_screen && !(can_do_density_screen || do_df_scf_guess)) {
```
JonathonMisiewicz, line 79: Why is it important to check `DF_SCF_GUESS`? It's unclear to me which keywords apply to which stages of the computation.

JonathonMisiewicz (commented):
Is there anything in this PR more sophisticated than moving code around? If not, I'm inclined to give this a very cursory review.

davpoolechem:
> Is there anything in this PR more sophisticated than moving code around? If not, I'm inclined to give this a very cursory review.

Nearly all of this PR is moving stuff around. The most notable changes to the actual algorithms themselves are small and are basically to smooth out running the two algorithms with different coding styles together (e.g., implementation of eri_computers_ as an unordered map, small changes to incfock workings for DFJLinK).

psi4/src/psi4/libfock/jk.cc
davpoolechem: This part is basically a cleanup of a code part that was previously present in Psi4, but rather messy. The basic idea is to exit out if trying to use density screening in conjunction with an algorithm that doesn't actually support it. The DF_SCF_GUESS option opens up the possibility of combining an integral-direct algorithm that supports density screening, with DFJK, which doesn't. Thus, a check for DF_SCF_GUESS would be required here.

JonathonMisiewicz (approved):
I'm happy to punt the listed questions to a future PR in this series. I'll let @loriab merge in.

Issue #2760, opened 2022-10-26
Random results when using multithreading and PK as SCF algorithm for single-electron systems


OUnke:
It seems there is a weird bug in Psi4 version 1.6.1 when using multithreading and PK as SCF algorithm, which results in essentially random results. I've attached a minimal example python script that runs the same calculation 10 times and prints the results to console.

minimal_example.py:
```
import psi4
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('--num_threads', type=int, default=1)
parser.add_argument('--scf_type', type=str, default='DF')
parser.add_argument('--element', type=int, default=1)
parser.add_argument('--multiplicity', type=int, default=2)
args = parser.parse_args()

psi4.core.be_quiet()  # Silence output for clarity.
psi4.set_num_threads(args.num_threads, quiet=True)
psi4.set_options({
  'REFERENCE': 'UHF',
  'SCF_TYPE': args.scf_type,
})

for i in range(10):
  energy = psi4.energy(
    'pbe0/sto-3g',
    molecule=psi4.core.Molecule.from_arrays(
        geom=[0, 0, 0],
        elez=[args.element],
        molecular_charge=0,
        molecular_multiplicity=args.multiplicity))
  print(energy)
```

To reproduce the behaviour, run
`python minimal_example.py --num_threads 6 --scf_type PK --element 1 --multiplicity 2`

For me, the output of this is (the exact numbers change on re-runs):
> -0.41204465858903644
-0.18510932345625386
0.18101301722463536
-0.17603191005094254
-0.1034126028084521
-0.04894812237658419
-0.2577286306987443
-0.3757350049677912
-0.26680604410405556
-0.26680604410405556

Interestingly, the issue is not present when using PK and a single thread,
`python test.py --num_threads 1 --scf_type PK --element 1 --multiplicity 2`

nor when using multiple threads, but DIRECT instead of PK
`python test.py --num_threads 6 --scf_type DIRECT --element 1 --multiplicity 2`

In both of these cases, I always get the same output:
> -0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424
-0.46650913902090424

Also, the issue is not present when using PK in multi-threaded environments, but using a different input, e.g.
`python test.py --num_threads 6 --scf_type PK --element 6 --multiplicity 3`
>-37.303469175322675
[...]

PR #2763, opened 2022-10-31
PK Bugfix 

davpoolechem:
## Description
This PR fixes the bug in the PK code brought up in https://github.com/psi4/psi4/issues/2760. It was observed that, for sufficiently small systems and small basis sets run in parallel, the calculation would converge to random energies. After some investigation, it was found that the buffer size assigned per thread in the PK algorithm could potentially be set to zero:

```    
    size_t buffer_size = pk_size() / nthreads(); // DP: can be set to 0 in edge cases
    size_t lastbuf = pk_size() % nthreads();

    for (size_t i = 0; i < nthreads(); ++i) {
        // DEBUG        outfile->Printf("start is %lu\n",start);
        SharedPKWrkr buf = std::make_shared<PKWrkrInCore>(primary(), sieve(), buffer_size, lastbuf, J_ints_.get(),
                                                          K_ints_.get(), wK_ints_.get(), nthreads());
        fill_buffer(buf);
        set_ntasks(nthreads());
    }
```

This leads to very unusual behavior occurring within the PK algorithm and is the root cause of the issue brought up regarding the PKJK algorithm. To address this issue, the buffer size per thread is always set to have at least a value of 1. When a minimum buffer size is enforced, the issue presented in https://github.com/psi4/psi4/issues/2760 disappears. Additionally, with these changes, the PK option produces the same answer as the DIRECT algorithm for the system in question (the H atom):

```
(p4dev) dpoole34@ds6:~$ cat minimal_example.py 
import psi4
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('--num_threads', type=int, default=1)
parser.add_argument('--scf_type', type=str, default='DF')
parser.add_argument('--element', type=int, default=1)
parser.add_argument('--multiplicity', type=int, default=2)
args = parser.parse_args()

psi4.core.be_quiet()  # Silence output for clarity.
psi4.set_num_threads(args.num_threads, quiet=True)
psi4.set_options({
  'REFERENCE': 'UHF',
  'SCF_TYPE': args.scf_type,
})

for i in range(10):
  energy = psi4.energy(
    'pbe0/sto-3g',
    molecule=psi4.core.Molecule.from_arrays(
        geom=[0, 0, 0],
        elez=[args.element],
        molecular_charge=0,
        molecular_multiplicity=args.multiplicity))
  print(energy)
(p4dev) dpoole34@ds6:~$ python3 minimal_example.py --num_threads 6 --scf_type PK --element 1 --multiplicity 2
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
(p4dev) dpoole34@ds6:~$ python3 minimal_example.py --num_threads 6 --scf_type DIRECT --element 1 --multiplicity 2
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
-0.466509139020904
```
## Dev notes & details
- [x] Fix bug present within PK calculations executed on small systems with small basis sets, wherein random energies would result from the calculation.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/PKmanagers.cc
+++ b/psi4/src/psi4/libfock/PKmanagers.cc
@@ -1472,6 +1472,7 @@ void PKMgrInCore::allocate_buffers() {
 
     // TODO We probably need more buffer for more tasks for better load balancing
     size_t buffer_size = pk_size() / nthreads();
+    if (buffer_size == 0) buffer_size = 1;
     size_t lastbuf = pk_size() % nthreads();
 
     for (size_t i = 0; i < nthreads(); ++i) {
--- a/psi4/src/psi4/libfock/PKmanagers.cc
+++ b/psi4/src/psi4/libfock/PKmanagers.cc
@@ -176,6 +176,11 @@ PKManager::PKManager(std::shared_ptr<BasisSet> primary, size_t memory, Options& 
     nthreads_ = 1;
 #ifdef _OPENMP
     nthreads_ = Process::environment.get_n_threads();
+    if (nthreads_ > pk_size_) {
+	outfile->Printf("  WARNING! More threads than unique shell quartets to compute!\n");
+	outfile->Printf("  Decreasing thread count to %d.\n", pk_size_);	
+	nthreads_ = pk_size_;
+    }
 #endif
 }
 
@@ -1472,7 +1477,6 @@ void PKMgrInCore::allocate_buffers() {
 
     // TODO We probably need more buffer for more tasks for better load balancing
     size_t buffer_size = pk_size() / nthreads();
-    if (buffer_size == 0) buffer_size = 1;
     size_t lastbuf = pk_size() % nthreads();
 
     for (size_t i = 0; i < nthreads(); ++i) {
--- a/conda/win/meta.yaml
+++ b/conda/win/meta.yaml
@@ -15,7 +15,7 @@ requirements:
     - numpy
     - pytest>=7.0.1
     - python={{ PY_VER }}
-    - qcelemental=0.25.0
+    - qcelemental=0.25.1
     - qcengine=0.24.1
     - msgpack-python
     - gau2grid


JonathonMisiewicz (approved):

hokru:
So the edge case is that there are less `pk_pairs` than `nthreads`, right? 
How about setting `nthreads = pk_pairs` instead?

davpoolechem:
> So the edge case is that there are less pk_pairs than nthreads, right?

Technically `pk_size` instead of `pk_pairs`, but otherwise yeah. The former describes the number of symmetry-unique shell quartets, while the latter describes the number of symmetry-unique shell pairs.

> How about setting nthreads = pk_pairs instead?

One would set `nthreads` to `pk_size`, per above, but I'm overall neutral on whether this change is better or not. I implemented your suggestion in a private repo, and it works just as fine as my change for fixing the bug. The changes this PR introduces are so small size-wise, and apply to such small calculations, that it's really a matter of code design than anything. On one hand, your suggestion is more elegant than my implementation, which is basically a brute-force fixing of the problem; and your suggestion is a lot cleaner for adding a warning to the user that "Hey, you're using too many threads!". On the other hand, I'm somewhat averse to forcibly lowering the thread count from what the user asked for, from a principle standpoint. Overall though, I'm fine with either approach being used for fixing this PK bug, and I'll be happy to use whatever approach people think is better.

psi4/src/psi4/libfock/PKmanagers.cc
```diff
@@ -1472,6 +1472,7 @@ void PKMgrInCore::allocate_buffers() {
 
     // TODO We probably need more buffer for more tasks for better load balancing
     size_t buffer_size = pk_size() / nthreads();
+    if (buffer_size == 0) buffer_size = 1;
```
hokru, line 1475: I think also `lastbuf = 0` is needed

hokru:
As long as the threads are only reduced during the PK steps it would only have benefits. No threads that do no work and allocate memory (Even if it doesn't matter in practice)

The automatic selection probably ensures the in-core algorithm for these small calculations but could you test `pk_algo = YOSHIMINE/REORDER` and `PK_NO_INCORE True` as well? We probably have no test coverage.
Limiting `nthreads` in the PKManager https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/PKmanagers.cc#L154 might solve headaches in the other algorithms. I saw a `ntask < nthreads` check in Yoshimine.

If that exceeds time you planned to invest for the PK bug fix that is also fine.

davpoolechem:
> As long as the threads are only reduced during the PK steps it would only have benefits. No threads that do no work and allocate memory (Even if it doesn't matter in practice).

This is all true. Any opposition I would have to reducing thread counts is from a principle standpoint - essentially having the code forcibly doing something in opposition to what the user wanted it to do. That said, this is a minor qualm on my part, and everything you said here is correct regardless.

> The automatic selection probably ensures the in-core algorithm for these small calculations but could you test pk_algo = YOSHIMINE/REORDER and PK_NO_INCORE True as well? We probably have no test coverage.
Limiting nthreads in the PKManager https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/PKmanagers.cc#L154 might solve headaches in the other algorithms. I saw a ntask < nthreads check in Yoshimine.

Sure! I tested the Yoshimine and Reorder algorithms (with PK_NO_INCORE set to True) with the minimal_example.py file provided in issue https://github.com/psi4/psi4/issues/2760; and as it turns out, using PK_ALGO=REORDER leads to a segfault when using my implementation. Your suggestion of changing nthreads directly fixes that REORDER segfault issue as well. In light of that, I will be happy to suggest your suggested methodology for fixing this PR!

loriab (approved):

loriab:
FYI, I'm going to make a couple more little changes to this branch to kill two bugs with minimal CI runs.

davpoolechem:
> FYI, I'm going to make a couple more little changes to this branch to kill two bugs with minimal CI runs.

Awesome; thank you very much!

JonathonMisiewicz (approved):

hokru (approved):

PR #2388, opened 2021-12-30
Libint2 one electron integrals

andysim:
## Description
Adds libint2 one electron integrals.  A few integrals still use the hand-written code for now, but they will be addressed in subsequent PRs, as linked below.
## Todos

- [x] Add libint2 one electron integrals.
- [x] Convert all double shell loops to use Onebody object's shell pair list instead.
- [x] Convert multipole potential ints to use L2 engines or McMurchie-Davidson recursions (#2414)
- [x] Convert X2C ints to use L2 engines.
- [x] Convert AngularMomentum ints to use L2 engines. (#2414)
- [x] Delete unused hand written code for recursions and integrals. (#2414)
- [x] Document changes
- [x] Build Windows Libint2 Conda package

## New Features for Release Notes
- Screening of one electron integrals to improve efficiency of PCM, EFP, and embedding methods
- Parallelized PCM integral computations, which are typically rate-limiting for implicit solvent SCF

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2767, opened 2022-11-07
Implement interface with pyddx

mfherbst:
## Description
This PR strives to implement an interface of psi4 to the [ddX library](https://github.com/ACoM-Computational-Mathematics/ddX), which implements solvation models (COSMO, PCM, linearised Poisson-Boltzmann) following a domain-decomposition approach.

At its current stage I open the PR to get some feedback from devs about the suggested changes and structure and to finalise the upstream python interface of ddX. Note that this PR Is currently deliberately done on top of an outdated master, since any commit after #2388 introduces segfaults (details see below), which so far I have not yet been able to trace down. Any help on that would be much appreciated.

- Implementation of PCM and COSMO solvation models based on the ddx library.

## Dev notes & details
- Introduction of a NumIntHelper class to compute some integrals numerically using a DFT grid
- Introduction of ddx solvation model and new ddx options

## Reproducer for the mysterious segfault
As part of rebasing against the current master I encountered a really strange segfault. I managed to produce a minimal example, which has really nothing to do with ddx and only adds a python interface to a simple numerical electrostatic integral. See here for a [trimmed-down diff](https://github.com/psi4/psi4/compare/master...mfherbst:psi4:segfault). On my machine checking out this `segfault` branch with `0_configure.sh`, building and running the `mytests/runtests.sh` script gives a segfault inside the numerical integration in the `PCMPotentialInt` class. Note that the code I added *is not even called*, the call to `PCMPotentialInt` happens from the pcm code *which I have not modified*.

Now, commenting out
```c++
PrintIntegralsFunctor printer;
potential_integrals_->compute(printer);
```
the segfault disappears. I'm getting the weird feeling I'm doing something really stupid here and I just missed it.

## Questions
- [x] Thoughts on the NumIntHelper?
- [x] Is D -> D_cart needed or not ([this stuff](https://github.com/psi4/psi4/pull/2767/commits/108b6bb1a53cdc428933475dea991e08a3330d98#diff-48947c7a095c933d6843e0564043f051b0e54dff111fff222bf25cd968f6f9daR1769-R1780))
- [x] Any idea on the segfault when rebasing against master?

## Checklist
- [x] Find and fix segfaults when using psi4 master
- [x] Remove debug restriction to one thread in numerical integrals
- [x] Documentation
- [x] Bring forward more options to change numerical grid details for DDX solvation models
- [x] ~~Merge to the largest extent possible PCM and DDX solvation options~~ (I don't think this can be done easily. Leaving it for now)

## Status
- [x] Ready for review
- [x] Ready for merge

Tagging @hokru, @loriab and @andysim who were involved in discussions about this some point (but a long time ago ... sorry for taking so long to tackle this :smile:).

### DIFF
--- a/doc/sphinxman/source/ddx.rst
+++ b/doc/sphinxman/source/ddx.rst
@@ -54,8 +54,8 @@ Interface to ddx by A. Mikhalev, A. Jha, M. Nottoli and M. F. Herbst
 by A. Mikhalev *et. al.*. The library provides a linear-scaling implementation
 of standard continuum solvation models using a domain-decomposition ansatz
 [Cances:2013:054111]_ [Stamm:2016:054101]_.
-Currently the conductor-like screening model (COSMO) [Barone:1998:1995]_
-and the polarisable continuum model (PCM) [Tomasi:2005:2999]_
+Currently the conductor-like screening model (COSMO) [Klamt:1993:799]_ [Lipparini:2014:184108]_
+and the polarisable continuum model (PCM) [Tomasi:2005:2999]_ [Nottoli:2019:6061]_
 are supported.
 No additional licence or configuration is required to use ddx with Psi4.
 
@@ -98,8 +98,9 @@ PCMSolver is based on a boundary-element discretisation [Cances:1998:309]_,
 while ddx is based on a domain decomposition approach
 [Cances:2013:054111]_ making it linear scaling.
 For more details about PCMSolver see the :ref:`section on PCMsolver <sec:pcmsolver>`.
-For a concise introduction to the `theory behind ddx <https://ddsolvation.github.io/ddX/md_docs_theory.html>`
-or further `literature references <https://ddsolvation.github.io/ddX/label_references.html>`
+For a concise introduction to the
+`theory behind ddx <https://ddsolvation.github.io/ddX/md_docs_theory.html>`_
+or further `literature references <https://ddsolvation.github.io/ddX/label_references.html>`_
 see the ddx documentation.
 
 The usage of ddx-based solvation models is enabled
@@ -194,12 +195,12 @@ of the implict description of the solvation.
 
 |ddx__dft_radial_points| and |ddx__dft_spherical_points| influence
 the accuracy of the numerical grid used to obtain the representation
-of the electric field of the solute density
-(for which a standard DFT integration grid is used).
-Unlike the case for DFT grids a finer grid than the employed default is
-rarely needed. In fact often even a coarser grid than the default
-can be used without trading too much accuracy
-(e.g. 50 radial and 230 spherical points).
+of the electric potential / field of the solute density,
+since a standard DFT integration grid is used to obtain these quantities.
+In contrast to the integration of DFT quantities much lower accuracy
+is required, such that for this step considerably smaller grids are employed.
+If extremly high accuracy reference solutions are required, the DDX
+DFT integration parameters might need to be increased, but this is rarely needed.
 
 |ddx__lmax| and |ddx__n_lebedev| determine the accuracy of the discretisation
 on the boundary of the spheres around each atom. The defaults are usually good.
--- a/doc/sphinxman/source/ddx.rst
+++ b/doc/sphinxman/source/ddx.rst
@@ -98,6 +98,9 @@ PCMSolver is based on a boundary-element discretisation [Cances:1998:309]_,
 while ddx is based on a domain decomposition approach
 [Cances:2013:054111]_ making it linear scaling.
 For more details about PCMSolver see the :ref:`section on PCMsolver <sec:pcmsolver>`.
+For a concise introduction to the `theory behind ddx <https://ddsolvation.github.io/ddX/md_docs_theory.html>`
+or further `literature references <https://ddsolvation.github.io/ddX/label_references.html>`
+see the ddx documentation.
 
 The usage of ddx-based solvation models is enabled
 by specifying |globals__ddx| ``true`` in your input file.
--- a/doc/sphinxman/source/ddx.rst
+++ b/doc/sphinxman/source/ddx.rst
@@ -169,8 +169,8 @@ The full list understood by ddx can be obtained using ::
 The **cavity** in ddx is defined as a union of spheres around each atom.
 Usually the spehere radii for each atom are determined using a standard
 set of tabulated radii per atomic species, determined by the |ddx__radii_set| parameter.
-Currently ``bondi`` and ``uff`` are supported for |ddx__radii_set|
-with ``uff`` selected by default.
+Currently ``bondi`` [Bondi:1964:441]_ and ``uff`` [Rappe:1992:114]_
+are supported for |ddx__radii_set| with ``uff`` selected by default.
 These radius values are conventionally scaled by an additional factor before use,
 conventionally 1.1 for ``uff`` and 1.2 for ``bondi``. Customisation of the scaling
 is possible using the |ddx__radii_scaling| parameter.
--- a/doc/sphinxman/source/ddx.rst
+++ b/doc/sphinxman/source/ddx.rst
@@ -102,10 +102,10 @@ For more details about PCMSolver see the :ref:`section on PCMsolver <sec:pcmsolv
 The usage of ddx-based solvation models is enabled
 by specifying |globals__ddx| ``true`` in your input file.
 The solvation model itself is selected using the |ddx__model| parameter.
-
-|Psifour| understands a number of other options to configure
-numerical details in the library (discretisation, iterative solver, cavity setup),
-see the next section.
+Additionally the definition of the solvent and solute cavity is required
+and further parameters allow to influence details of discretisation,
+numerical integration and iterative solvers,
+see the next sections for details.
 
 .. note:: At present dd-based solvation models
           can only be used for energy calculations with SCF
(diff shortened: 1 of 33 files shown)


TiborGY:
FYI rebasing onto https://github.com/TiborGY/psi4/tree/toc_dbg should allow you to get a debug build, with line numbers in GDB working. Might help with debugging the segfault.

maxscheurer:
I think I'm going to have a look at the mysterious "segfault" since I've been doing quite some libmints/libint2-related debugging for #2414 etc. It should not be too hard to track down (hopefully) 😄

mfherbst:
Thanks @maxscheurer :+1:. Let me know if you need something beyond my reproducer.

maxscheurer:
Unfortunately I cannot reproduce the segfault... Are you building in a conda environment w/ libint2 from the psi4 channel?

maxscheurer:
> No, I'm builing everything from source locally. I can try that though if you think that's better.

Ok, I don't know if self-build L2 could be the reason @loriab.
If you could create an environment `conda create -n psi4_dev psi4-dev -c psi4/label/dev -c defaults --override-channels`
and try building in that environment with what `psi4-path-advisor --psi4-compile` gives you (plus the enable PCM solver flag), that'd be worth a shot I think.

mfherbst:
Thanks for that tip. I tried that and I get the same segfault.

maxscheurer:
I've been able to reproduce the segfault on Linux and fix it, see #2770.

mfherbst:
Good catch! Thanks!

maxscheurer:
> Is D -> D_cart needed or not?

I think the manual conversion was needed before L2, now it is not required anymore. Let me double-check the commit history though 😁

maxscheurer:
I can confirm that the `D_cart` conversion stuff is not needed, it was removed in the L2 OEI PR [here](https://github.com/psi4/psi4/pull/2388/commits/91731f89e9b36b725726db1fac728392f0e91084#diff-94c1f3db6d9f3a05612ed101c8e726a172a7ea40f4be040178c63c4f38438022L217-L227), for example, and is still the case in the current [PCM implementation](https://github.com/psi4/psi4/blob/6384381a24a446c0f37d96f54fa69b203c5a7893/psi4/src/psi4/libpsipcm/psipcm.cc#L217).

mfherbst:
@hokru  @loriab @maxscheurer This is ready for a first review ;). In particular I'd like to hear your thoughts on the numerical integral class and what should be changed.

mfherbst:
Also I'm not sure how to deal with the ambivalence between DDX and PCM.

PCMsolver can do PCM and COSMO, but DDX can provide domain-decomposition variants of these plus in the future linearised Poisson-Boltzmann (LPB), which is in some sense an extension to PCM. So looking ahead it feels weird to have a flag PCM to decide whether LPB is run. Similar DDX and PCMsolver will not give the same values for the solvation energy, but are sort of doing the same thing (continuum solvation models). So how shall this be dealt with?

maxscheurer (commented):
This looks great, thanks a lot for the contribution! I will go through the interface code again in detail, but it's very clean und easy to follow.

Some things that came to mind:
- If the `elec_only` functionality is working, one could add LR-DDX for TDDFT. I can look up where the line of code needs to be added. If we want to add this, you could probably generate reference data with Gaussian, if it's not too much to ask.
- Will it be possible to add nuclear gradients in a follow-up PR? I think this would be a much appreciated feature by a lot of Psi users.
- As for the numerical integration, I leave the comments to @hokru et al ☺️
- What about performance in general? Have you ran some small benchmark comparing to the existing PCM implementation? Just curious 🧐

mfherbst:
Hi @maxscheurer. Let me briefly react to your comments

> If the elec_only functionality is working, one could add LR-DDX for TDDFT. I can look up where the line of code needs to be added. If we want to add this, you could probably generate reference data with Gaussian, if it's not too much to ask.

Nice idea, but it is not yet been tested to the point where I would be confident in it. I have this planned as a follow-up to this one ... I'll add a todo for now.

> Will it be possible to add nuclear gradients in a follow-up PR? I think this would be a much appreciated feature by a lot of Psi users.

Yes absolutely. That takes a bit of work (as more quantities are needed on the psi4-side), but ddx has them and naturally it would make sense to carry that forward to psi4.

> What about performance in general? Did you run some small benchmark comparing to the existing PCM implementation? 

I did not benchmark things rigorously, but e.g. on nitro-aniline in an STO-3G basis the timings were noticably different. In that setup the main load of the SCF is on the pcm. Here the SCF needs around 150s for PCMsolver and around 40s for ddx. Please take this with a grain of salt as I have done zero testing in how this scales or translates to realistic setups.

maxscheurer (approved):

psi4/src/read_options.cc
```diff
@@ -302,6 +304,80 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+
+        /*- Tolerance to which DDX linear systems are solved -*/
+        options.add_double("SOLVATION_CONVERGENCE", 1e-8);
+
+        /*- Number of spherical points used to compute the integrals for DDX calculations
+            (A :ref:`Lebedev Points <table:lebedevorder>` number) -*/
+        options.add_int("DFT_SPHERICAL_POINTS", 110);
+
+        /*- Number of radial points used to compute the integrals for DDX calculations -*/
+        options.add_int("DFT_RADIAL_POINTS", 35);
```
loriab, line 351: `ddx__dft_radial_points` is the logical option name the way the options class was written. But, a flatter options namespace has become preferred over the years, so feel free to do `ddx__ddx_radial_points` like https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L280-L283 so that users don't necessarily have to set the module.

tests/pytests/test_ddx.py
```diff
@@ -0,0 +1,250 @@
...
+    psi4.set_options({
+        "ddx": True,
+        "basis": "3-21g",
+        "guess": "core",
+        "scf_type": "direct",
+        #
+        "ddx__model": "cosmo",
+        "ddx__radii_set": "uff",
+        "ddx__lmax": 3,
+        "ddx__n_lebedev": 302,
```
loriab, line 189: isn't the lebedev number 29 for sph 302? should both options be available?

doc/sphinxman/source/ddx.rst
```diff
@@ -0,0 +1,272 @@
...
+These parameters can be altered to balance the cost and accuracy
+of the implict description of the solvation.
+
+|ddx__dft_radial_points| and |ddx__dft_spherical_points| influence
+the accuracy of the numerical grid used to obtain the representation
+of the electric potential / field of the solute density,
+since a standard DFT integration grid is used to obtain these quantities.
+In contrast to the integration of DFT quantities much lower accuracy
+is required, such that for this step considerably smaller grids are employed.
+If extremly high accuracy reference solutions are required, the DDX
```
loriab, line 202:
```suggestion
If extremely high accuracy reference solutions are required, the DDX
```

loriab (approved):
Thanks for the much-wanted contribution and nice documentation and testing!

loriab:
> Also I'm not sure how to deal with the ambivalence between DDX and PCM in terms of the user-facing flags to enable the models and the Psi variables to store results / energy terms.

> PCMsolver can do PCM and COSMO, but DDX can provide domain-decomposition variants of these plus in the future linearised Poisson-Boltzmann (LPB), which is in some sense an extension to PCM. So looking ahead it feels weird to have a flag PCM to decide whether LPB is run. Similar DDX and PCMsolver will not give the same values for the solvation energy, but are still sort of doing the same thing (continuum solvation models). Any thoughts?

Psi likes to have a uniform interface when one can get the same value from different engines, but this has really only worked out for empirical dispersion. So the other principle is that it's fine for options to be a passthrough to the upstream project. Then PCM, PE, and DDX are effectually engine/upstream specifiers rather than the method flags that "PCM" suggests.

loriab:
Looks like there's a trivial merge conflict. Lmk if you prefer (1) I resolve with the GH interface, (2) I rebase and force-push, or (3) you want to handle it. Thanks!

tests/pytests/test_ddx.py
mfherbst: I see. Indeed that might be confusing. What you refer to is the Lebedev order, but what is meant here (and in the appropriate DDX option) is the number of actual points used to discretise the unit sphere, which for Lebedev order 29 is 302. I've used this flag for consistency with the pyddx interface.

tests/pytests/test_ddx.py
```diff
@@ -0,0 +1,250 @@
...
+        "ddx__dft_radial_points": 75,
+        "ddx__shift": 0.0,
+    })
+    for key in inp["ddx"].keys():
+        psi4.set_options({"ddx__" + key: inp["ddx"][key]})
+    scf_e, wfn = psi4.energy('SCF', return_wfn=True, molecule=mol)
+    ddx_e = wfn.scalar_variable("dd solvation energy")
+
+    assert compare_values(inp["solvation"], ddx_e, 4, "DDX solvation energy versus Gaussian")
+    assert compare_values(inp["ref"], scf_e, 4, "Total SCF energy with DDX versus Gaussian")
```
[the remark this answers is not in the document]
mfherbst: Might be tricky, I'll ask a colleague, who did the Gaussian calculations. It's not just about the SCF tightness, but also about making sure the same DFT grid is used to get the electrostatic potential. I doubt Gaussian and Psi4 can both be tricked into using the same grid and pruning schemes etc.

psi4/src/read_options.cc
mfherbst: Ok. I've put it into global.

maxscheurer:
Let's postpone the tighter conv checking and merge this to get it into the upcoming 1.7 release ☺️ Are you ok with this @mfherbst @loriab?

psi4/src/read_options.cc
```diff
@@ -286,10 +286,83 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        Advanced parameter, which usually does not need to be modified. Valid
+        values are within the range [0, 1]. !expert -*/
+    options.add_double("DDX_ETA", 0.1);
+
+    /*- Shift for characteristic function of sphere overlap.
+        Advanced parameter, which usually does not need to be modified. Valid values
+        are within the range [-1, 1] with -100 denoting an automatic selection of the
+        best shift. !expert -*/
+    options.add_double("DDX_SHIFT", -100.0);
+
```
loriab, line 358: Sorry, I was unclear. Please model on PCM, so move all the DDX options (except the governing boolean below) back to a `if (name == "DDX" || options.read_globals()) {` section of their own. But they can keep the prefix; that is, the `FMM` -> `DDX_FMM` change is good. Thanks

tests/pytests/test_ddx.py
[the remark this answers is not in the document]
loriab: Oh right, their default pruned grid isn't published, iirc. So the only way might be to use very tight unpruned grids in both programs. So this existing looser check under everyday conditions is probably the better. Thanks.

mfherbst:
> Let's postpone the tighter conv checking and merge this to get it into the upcoming 1.7 release

Fine by me.

psi4/src/read_options.cc
mfherbst: ok. I'll make the adaption.

loriab:
>> Let's postpone the tighter conv checking and merge this to get it into the upcoming 1.7 release

> Fine by me.

Me, too. Only TODO is the options semi-reversion. Sorry for the extra work.

PR #2768, opened 2022-11-07
Expand Testing of COSX and LINK SCF Types

davpoolechem:
## Description
The goal of this PR is to expand test coverage of the DFJCOSK and DFJLinK JK subclasses within Psi4 by including their testing within the scf5 test. The scf5 test in Psi4 tests a variety of combinations of SCF_TYPE and SCF reference for singlet and triplet O2. As it were, the COSX and LinK SCF_TYPES were missing from the list of methods analyzed within this test. This PR simply adds COSX and LinK to the list of methods tested in scf5, expanding testing of these two methods beyond what was previously present.

NOTE TO REVIEWERS: This PR is a general JK maintenance/cleanup/improvement PR and is independent of the developments occurring regarding CompositeJK and its direct implementation into Psi4. 

- [x] N/A

## Dev notes & details

- [x] Adds COSX and LINK to the list of SCF_TYPEs tested in the scf5 test option.

## Questions
- [ ] Is the methodology by which COSX and LinK tested acceptable? Unfortunately, the reference energies used in scf5 are either canonical (i.e., non-DF) or DF, neither of which fully describe the COSX and LINK methods. The approach I took is to compare each COSX and LINK energy to the corresponding canonical energy with an atol of 1E-4. Any feedback on this particular approach to testing COSX and LINK would be appreciated.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/samples/scf5/input.dat
+++ b/samples/scf5/input.dat
@@ -47,11 +47,11 @@ E = energy('scf')
 set scf_type mem_df
 E = energy('scf')
 
-set scf_type cosx 
+set scf_type cosx
 E = energy('scf')
 
 set scf_type link
-set screening density 
+set screening density
 E = energy('scf')
 
 set scf reference uhf
@@ -72,11 +72,11 @@ E = energy('scf')
 set scf_type mem_df
 E = energy('scf')
 
-set scf_type cosx 
+set scf_type cosx
 E = energy('scf')
 
 set scf_type link
-set screening density 
+set screening density
 E = energy('scf')
 
 set scf reference cuhf
@@ -97,11 +97,11 @@ E = energy('scf')
 set scf_type mem_df
 E = energy('scf')
 
-set scf_type cosx 
+set scf_type cosx
 E = energy('scf')
 
 set scf_type link
-set screening density 
+set screening density
 E = energy('scf')
 
 activate(triplet_o2)
@@ -130,11 +130,11 @@ E = energy('scf')
 set scf_type mem_df
 E = energy('scf')
 
-set scf_type cosx 
+set scf_type cosx
 E = energy('scf')
 
 set scf_type link
-set screening density 
+set screening density
 E = energy('scf')
 
 clean()
@@ -160,11 +160,11 @@ E = energy('scf')
 set scf_type mem_df
 E = energy('scf')
 
-set scf_type cosx 
+set scf_type cosx
 E = energy('scf')
 
 set scf_type link
-set screening density 
+set screening density
 E = energy('scf')
 
 clean()
@@ -190,9 +190,9 @@ E = energy('scf')
 set scf_type mem_df
 E = energy('scf')
 
-set scf_type cosx 
+set scf_type cosx
 E = energy('scf')
 
 set scf_type link
-set screening density 
+set screening density
 E = energy('scf')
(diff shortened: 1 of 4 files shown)


TiborGY:
Would testing COSX results against the implementation in Orca 5 make sense? Of course the grid dependence makes such things tricky.

davpoolechem:
> Would testing COSX results against the implementation in Orca 5 make sense? Of course the grid dependence makes such things tricky.

Practically speaking, this should be doable, but one would need to ensure that as many factors between the two implementations as possible are standardized and made the same. Some of these factors (e.g., screening tolerances) shouldn't be that hard, but others (like grids, as you mentioned) would be quite a bit harder and could be quite a thorn in our side for comparing the two implementations. 

This question opens up the discussion to another approach I was thinking for this PR - using separate reference energies for COSX and LINK in the scf5 test, rather than comparing against canonical or DF energies. I am rather neutral on that matter, and am definitely open to doing so; but there are some concerns that should be considered with the separate reference energies approach. These are concerns that might not matter as much now; but the scf5 test will undoubtedly be expanded as CompositeJK and new separate J and K algorithms are added to Psi4, and these concerns might pop up then:

1. Using separate reference energies for each CompositeJK JK build combination, for each O2 multiplicity and SCF reference, could quickly make scf5 quite unwieldy and messy. 
2. Against what do we compare reference energies? In some cases, it won't be too big of a problem (as with here, where we can compare against ORCA), but I'm willing to bet that there are certain combinations of J and K algorithms that CompositeJK will eventually be able to do, that might not actually be present in other codes to compare against. 

Of course, these concerns are more specific to CompositeJK testing, so maybe they're an issue that can punted down the road for when scf5 gets updated again.

loriab:
> Is the methodology by which COSX and LinK tested acceptable? Unfortunately, the reference energies used in scf5 are either canonical (i.e., non-DF) or DF, neither of which fully describe the COSX and LINK methods. The approach I took is to compare each COSX and LINK energy to the corresponding canonical energy with an atol of 1E-4. Any feedback on this particular approach to testing COSX and LINK would be appreciated.

I think it'd be most useful to continue scf5 as a regression test. That is, record new refs for the in-between-conv-and-df algorithms collected at tight convergences, then check them against usual 6 decimal places. Then we'll know in future if code changes affect the algorithms.

How are you generating the `samples/` files? If it's any other way than autogenerated via the docs build, please don't go to the trouble. We can catch the changes en masse before release.

JonathonMisiewicz:
I strongly agree with Lori about having this as a regression test rather than an accuracy test. Put the new, approximate energies with all the others. That should be a good indicator to others working with the SCF system of what accuracy to expect.

davpoolechem:
I did a couple more things here:

1) At the suggestion of the reviewers (assuming I understtod it correctly), I added individual reference energies for the current composite methods (which were acquired at tight tolerances - e-10 for D and E convergence, and e-14 for ERI screening). The COSX and LinK methods now test against these corresponding reference energies.
2) With 1) in mind, I made more sweeping changes as well. First, the reference energies are now kept in a dictionary rather than spread out across different variables. This improves readability and should facilitate the addition of new methods, Composite or otherwise, to the testing suite in Psi4. Additionally, composite methods are tested in a loop per test case. This will facilitate the testing of new Composite methods added to Psi4, as will happen in the future.

JonathonMisiewicz (approved):

loriab (approved):
lgtm, thanks for the expanded testing!

PR #2769, opened 2022-11-08
KE/PE/Virial

JonathonMisiewicz:
## Description
This is take one at adding the kinetic energy, potential energy, and virial ratio as psivars, for the purposes of `cc` testing. I'm not confident on how this handles PCM, EFP, PE, and point charges, so putting it up very early for feedback.

- [x] Kinetic energy, potential energy, and virial energy are added as Psi variables for fully QM HF and CC computations. The variable names are subject to change.

## Dev notes & details

- [x] Standardize `E_h` in docs

## Checklist
- [ ] Do we want tests for the new psivars now, or in the next PR?

## Status
- [x] Ready for review
- [x] Ready for merge - need Lori re-review.

### DIFF
--- a/doc/sphinxman/source/glossary_psivariables.rst
+++ b/doc/sphinxman/source/glossary_psivariables.rst
@@ -415,7 +415,11 @@ PSI Variables by Alpha
 
 .. psivar:: CC CORRELATION VIRIAL RATIO
 
-   The correlation correction to the virial ratio, as computed by a coupled cluster method. Only defined for a fully quantum mechanical computation, i.e., not QM/MM.
+   The correlation virial ratio, as defined in DOI: 10.1063/1.1535440 for basis set completeness analysis. Computed using coupled cluster.
+
+.. psivar:: CC VIRIAL RATIO
+
+   The virial ratio, as computed by a coupled cluster method. Only defined for a fully quantum mechanical computation, i.e., not QM/MM.
 
 .. psivar:: CC T1 DIAGNOSTIC
    CC D1 DIAGNOSTIC
--- a/doc/sphinxman/source/glossary_psivariables.rst
+++ b/doc/sphinxman/source/glossary_psivariables.rst
@@ -46,23 +46,23 @@ PSI Variables by Alpha
 
 .. psivar:: [T] CORRECTION ENERGY
 
-   The coupled-cluster bracket perturbative triples correction [Eh].
+   The coupled-cluster bracket perturbative triples correction [E_h].
 
 .. psivar:: (T) CORRECTION ENERGY
 
-   The coupled-cluster perturbative triples correction [Eh].
+   The coupled-cluster perturbative triples correction [E_h].
 
 .. psivar:: (AT) CORRECTION ENERGY
    A-(T) CORRECTION ENERGY
 
-   The coupled-cluster asymmetric perturbative triples correction [Eh].
+   The coupled-cluster asymmetric perturbative triples correction [E_h].
 
 .. psivar:: AAA (T) CORRECTION ENERGY
    AAB (T) CORRECTION ENERGY
    ABB (T) CORRECTION ENERGY
    BBB (T) CORRECTION ENERGY
 
-   Spin components of the UHF-based coupled-cluster perturbative triples correction [Eh].
+   Spin components of the UHF-based coupled-cluster perturbative triples correction [E_h].
 
 .. psivar:: ACPF DIPOLE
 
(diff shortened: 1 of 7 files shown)


JonathonMisiewicz:
Better idea: in the case of EFP/PE/DDX/etc., we just don't define a virial ratio. If somebody _really_ needs that, they can do the work of figuring out how the virial theorem interacts with these non-quantum terms themselves. Thoughts?

psi4/src/psi4/cc/ccdensity/kinetic.cc
```diff
@@ -117,6 +117,10 @@ void kinetic(std::shared_ptr<Wavefunction> wfn) {
     outfile->Printf("\t-V/T (corr)            = %20.15f\n", -vcorr / tcorr);
     outfile->Printf("\t-V/T (total)           = %20.15f\n", -vtot / ttot);
 
+    wfn.set_scalar_variable("CC DELTA KINETIC ENERGY", tcorr);
+    wfn.set_scalar_variable("CC DELTA POTENTIAL ENERGY", vcorr);
+    wfn.set_scalar_variable("CC DELTA VIRIAL RATIO", -vtot / ttot + vref / tref);
```
loriab, line 122: What is the "delta" between? Looking at the `tcorr` etc. to be stored I might have named them `CC CORRELATION KINETIC ENERGY`, `CC CORRELATION POTENTIAL ENERGY`, `CC CORRELATION VIRIAL RATIO`.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -739,6 +740,10 @@ def scf_print_energies(self):
         #self.set_variable(self.functional().name() + ' FUNCTIONAL TOTAL ENERGY', hf_energy + exc + evv10)
         self.set_variable("DFT TOTAL ENERGY", dft_energy)  # overwritten later for DH  # P::e SCF
     else:
+        auto potential = total_energy - ke
```
loriab, line 743: You've been in C++ too long, `auto`. :-)

loriab (commented):
> Better idea: in the case of EFP/PE/DDX/etc., we just don't define a virial ratio.

That sounds reasonable to me. I haven't looked for an ideal "universal" place in the code yet, but may as well `del_variable` the virial qcvars at the end if EFP etc. detected.

Note that the vars will need to have a glossary.rst entry to appease the docs.

psi4/src/psi4/cc/ccdensity/kinetic.cc
JonathonMisiewicz: My one concern is that CC CORRELATION RATIO might _look like it means_ "correlation potential" / "correlation kinetic." Using "delta" instead doesn't have that ambiguity. If you aren't worried about that, as long as the psivars are defined clearly in the glossary, I'm happy to change "DELTA" to "CORRELATION."

loriab: I agree CORRELATION is ambiguous. In my initial reading I would call DELTA vague. I'm not sure which is worse. For consistency I'd probably choose CORRELATION accompanied by clear glossary definitions, but I'm not adamant. There's also CORRECTION that's often used for parts that aren't a full tot - ref, if that seems much better to you.

JonathonMisiewicz: After reviewing Psi usage, "CORRECTION" either appears in contexts that are too dissimilar or that should clearly be "CORRELATION." So "CORRECTION" it is.

loriab: Where are you seeing CORRECTIONs that should be CORRELATIONs? My intent was `CORL = TOT – REF`, and if a piece didn't meet that, it became a CORRECTION. (Even pieces like `(T) = CCSD(T) – CCSD` are CORRECTION, though that's certainly the perturbative correction to the _correlation_ energy.)

JonathonMisiewicz: With that clarification, I retract my previous statement: Psi usage is correct, and I'll move my keyword to "CORRELATION."

psi4/src/psi4/cc/ccdensity/kinetic.cc
```diff
@@ -117,6 +117,10 @@ void kinetic(std::shared_ptr<Wavefunction> wfn) {
     outfile->Printf("\t-V/T (corr)            = %20.15f\n", -vcorr / tcorr);
     outfile->Printf("\t-V/T (total)           = %20.15f\n", -vtot / ttot);
 
+    wfn->set_scalar_variable("CC CORRELATION KINETIC ENERGY", tcorr);
+    wfn->set_scalar_variable("CC CORRELATION POTENTIAL ENERGY", vcorr);
+    wfn->set_scalar_variable("CC CORRELATION VIRIAL RATIO", -vtot / ttot + vref / tref);
```
lothian, line 122: The correlation energy virial ratio is -vcorr/tcorr, which isn't the same as what you've got here, unless I'm missing something due to lack of sleep.

JonathonMisiewicz: There's a definition question.

@loriab and I prefer defining the correlation virial ratio so that `reference + correlation = total`. That property is lost with your definition, but your definition is the one adopted by the current `cc` output. I can adopt either, so long as you and Lori agree on it. I'm not familiar enough with virial theorem literature to say whether either candidate definition is standard.

lothian: I discussed this in [a paper in JCP](https://aip.scitation.org/doi/abs/10.1063/1.1535440) a while back, which is where I used these values as a measure of basis-set completeness for the correlated wave function only.  Like the total and Hartree-Fock energies, the ratio of the correlation contributions to the potential and kinetic energies also converges separately to -2, but it's not clear that what you're computing will do that.  Won't it converge to 0?  What values are you getting in your tests?

JonathonMisiewicz: Oh, I had no idea that you could use the correlation virial ratio for basis set completeness analysis. I've changed the variables and edited the documentation accordingly.

JonathonMisiewicz:
@loriab, please re-review. The last commit changed variable definitions.

lothian (approved):

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -405,6 +405,22 @@ PSI Variables by Alpha
...
+
+   The correlation correction to the kinetic energy [E_h], as computed by a coupled cluster method.
+
+.. psivar:: CC CORRELATION POTENTIAL ENERGY
+
+   The correlation correction to the potential energy [E_h], as computed by a coupled cluster method.
+
+.. psivar:: CC CORRELATION VIRIAL RATIO
+
+   The correlation virial ratio, as defined in DOI: 10.1063/1.1535440 for basis set completeness analysis. Computed using coupled cluster.
```
loriab, line 418:
```suggestion
   The correlation virial ratio, as defined in https://doi/org/10.1063/1.1535440 for basis set completeness analysis. Computed using coupled cluster.
```
This way DOIs get processed into links. Feel free to postpone to future PR.

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -405,6 +405,22 @@ PSI Variables by Alpha
...
+
+.. psivar:: CC VIRIAL RATIO
+
+   The virial ratio, as computed by a coupled cluster method. Only defined for a fully quantum mechanical computation, i.e., not QM/MM.
```
loriab, line 422:
```suggestion
   The virial ratio, as computed by a coupled cluster method. Only defined for a fully quantum mechanical computation, i.e., not QM/MM or solvated.
```

loriab (approved):
Sure, new vars well defined, so works for me.

PR #2771, opened 2022-11-10
Fix some incorrect printf format specifiers

TiborGY:
## Description

Printf format specifiers are required to match the type of the variable being printed. Using an incorrect format specifier may be UB. This PR resolves instances where the format specifier was incorrect. At least some of these have been triggering GCC warnings, so Psi4 now compiles with a couple fewer warnings.

- [x] Likely no user-visible changes

## Dev notes & details

- [x] `cctransort/memcheck`: Replace 14 instances of `%1lu` being used for printing an `int`, with `%d`. Make argument `const`.
- [x] `detci/olsengraph.cc`: fix one instance of `%3d` being used for printing a `size_t` integer
- [x] `dlpno/mp2.cc`: fix 12 instances of %d being used for `size_t` integers
- [x] `fnocc/lowmemory_triples.cc`: fix `%i` being used to print a `long`
- [x] `libfock/DirectJK.cc`: fix %3d being used for `size_t` integer
- [x] `libfock/solver.cc`: fix %d being used for `size_t` integer
- [x] `libmints/cdsalclist.cc`: fix 5 instances of %d being used for `size_t` integers
- [x] `libmints/extern.cc`: fix 3 instances of %d being used for `size_t` integers. Make argument `const &`.

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cctransort/memcheck.cc
+++ b/psi4/src/psi4/cctransort/memcheck.cc
@@ -32,7 +32,7 @@
 namespace psi {
 namespace cctransort {
 
-void memcheck(int reference) {
+void memcheck(const int reference) {
     size_t irrep_size, size;
     dpdbuf4 Z;
 
@@ -44,7 +44,7 @@ void memcheck(int reference) {
         for (int h = 0; h < Z.params->nirreps; h++) {
             irrep_size = (size_t)Z.params->rowtot[h] * Z.params->coltot[h];
             size += irrep_size;
-            outfile->Printf("\tSize of irrep %1lu of <ab|cd> integrals: %10.3lf (MW) / %10.3lf (MB)\n", h,
+            outfile->Printf("\tSize of irrep %d of <ab|cd> integrals: %10.3lf (MW) / %10.3lf (MB)\n", h,
                             irrep_size / 1e6, (irrep_size / 1e6) * sizeof(double));
         }
         global_dpd_->buf4_close(&Z);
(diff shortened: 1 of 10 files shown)


psi4/src/psi4/cctransort/memcheck.cc
```diff
@@ -44,7 +44,7 @@ void memcheck(int reference) {
         for (int h = 0; h < Z.params->nirreps; h++) {
             irrep_size = (size_t)Z.params->rowtot[h] * Z.params->coltot[h];
             size += irrep_size;
-            outfile->Printf("\tSize of irrep %1lu of <ab|cd> integrals: %10.3lf (MW) / %10.3lf (MB)\n", h,
+            outfile->Printf("\tSize of irrep %d of <ab|cd> integrals: %10.3lf (MW) / %10.3lf (MB)\n", h,
```
loriab, line 47: I had a habit of using `%zu` for `size_t`, and https://stackoverflow.com/a/2524675 concurs. Is there an advantage to `%d` here? I see you're using some `%zu`s below.

TiborGY: As far as I could gather, `%zu` is not an "universal automatic-size integer format", but "the format that works for size_t integers". And in this case, the type of `h` is just `int`, so unless `size_t` happens to be `int`, `%zu` would not be appropriate here, I think.

loriab: Ack, I was looking at `irrep_size`, not `h`, for the var-to-be-formatted. Sorry, you're right.

loriab (approved):
lgtm, thanks for the cleanup!

JonathonMisiewicz (approved):

PR #2775, opened 2022-11-14
change raw pointer return in `libmints` to `std::unique_ptr`

### LINKED ISSUE #2493: `libmints` Raw to Unique Pointer
Several methods in `libmints` export raw pointers. (See the file `libmints/integral.h` and [discussion](https://github.com/psi4/psi4/pull/2386#discussion_r832737185).) Those should be changed to `unique_ptr`.

Please do not start until #2386 is in. Check in here or on Slack before starting this issue, to confirm that you're not going to cause a problem for other ongoing PRs.

hmacdope:
Fixes #2493 

## Description

For memory safety, the integrals in libmints should be returned as unique_ptrs rather than raw pointers

## Dev notes & details

- [x] All integrals in libmints now return a unique_ptr rather than a raw pointer
- [x] Call sites refactored to match
## Questions
- [x] Are my refactors to the call sites correct? Many are just an immediate release of the unique_pointer, with the assumption that the memory management / pointer deletion occurs elsewhere
- [x] Should iterators eg `CartesianIter` also return unique_ptrs?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1304,7 +1304,7 @@ void export_mints(py::module& m) {
         .def("shell_significant", compute_shell_significant(&TwoBodyAOInt::shell_significant),
              "Determines if the P,Q,R,S shell combination is significant");
 
-    py::class_<Libint2ERI, std::shared_ptr<Libint2ERI>>(m, "ERI", pyTwoBodyAOInt,
+    py::class_<Libint2ERI, std::unique_ptr<Libint2ERI>>(m, "ERI", pyTwoBodyAOInt,
                                                         "Computes normal two electron repulsion integrals");
 #ifdef ENABLE_Libint1t
     py::class_<F12, std::shared_ptr<F12>>(m, "F12", pyTwoBodyAOInt, "Computes F12 electron repulsion integrals");
--- a/psi4/src/psi4/fisapt/fisapt.cc
+++ b/psi4/src/psi4/fisapt/fisapt.cc
@@ -525,7 +525,7 @@ void FISAPT::nuclear() {
 
     auto Vfact = std::make_shared<IntegralFactory>(primary_);
     std::shared_ptr<PotentialInt> Vint;
-    Vint = std::shared_ptr<PotentialInt>(static_cast<PotentialInt*>(Vfact->ao_potential()));
+    Vint = std::shared_ptr<PotentialInt>(static_cast<PotentialInt*>(Vfact->ao_potential().release()));
 
     // > Molecular Centers < //
 
@@ -3356,7 +3356,7 @@ void FISAPT::felst() {
 
     // => Nuclear Part (PITA) <= //
     auto Vfact2 = std::make_shared<IntegralFactory>(primary_);
-    std::shared_ptr<PotentialInt> Vint2(static_cast<PotentialInt*>(Vfact2->ao_potential()));
+    std::shared_ptr<PotentialInt> Vint2(static_cast<PotentialInt*>(Vfact2->ao_potential().release()));
     auto Vtemp2 = std::make_shared<Matrix>("Vtemp2", nn, nn);
 
 
@@ -3698,7 +3698,7 @@ void FISAPT::find() {
     // => Nuclear Part (PITA) <= //
 
     auto Vfact2 = std::make_shared<IntegralFactory>(primary_);
-    std::shared_ptr<PotentialInt> Vint2(static_cast<PotentialInt*>(Vfact2->ao_potential()));
+    std::shared_ptr<PotentialInt> Vint2(static_cast<PotentialInt*>(Vfact2->ao_potential().release()));
     auto Vtemp2 = std::make_shared<Matrix>("Vtemp2", nn, nn);
 
     double* ZAp = vectors_["ZA"]->pointer();
--- a/psi4/src/psi4/libmints/mintshelper.cc
+++ b/psi4/src/psi4/libmints/mintshelper.cc
@@ -1399,7 +1399,7 @@ void MintsHelper::add_dipole_perturbation(SharedMatrix potential_mat) {
 
     OperatorSymmetry msymm(1, molecule_, integral_, factory_);
     std::vector<SharedMatrix> dipoles = msymm.create_matrices("Dipole");
-    OneBodySOInt *so_dipole = integral_->so_dipole();
+    std::unique_ptr<OneBodySOInt> so_dipole = integral_->so_dipole();
     so_dipole->compute(dipoles);
 
     if (lambda[0] != 0.0) {
--- a/psi4/src/psi4/libmints/mintshelper.cc
+++ b/psi4/src/psi4/libmints/mintshelper.cc
@@ -1719,7 +1719,7 @@ std::vector<SharedMatrix> MintsHelper::electric_field(const std::vector<double> 
 SharedMatrix MintsHelper::induction_operator(SharedMatrix coords, SharedMatrix moments) {
     SharedMatrix mat = std::make_shared<Matrix>("Induction operator", basisset_->nbf(), basisset_->nbf());
     ContractOverDipolesFunctor dipfun(moments, mat);
-    auto field_integrals_ = static_cast<ElectricFieldInt *>(integral_->electric_field().release());
+    auto field_integrals_ = std::unique_ptr<ElectricFieldInt>(static_cast<ElectricFieldInt *>(integral_->electric_field().release()));
     field_integrals_->compute_with_functor(dipfun, coords);
     mat->scale(-1.0);
 
@@ -1727,7 +1727,7 @@ SharedMatrix MintsHelper::induction_operator(SharedMatrix coords, SharedMatrix m
 }
 
 SharedMatrix MintsHelper::electric_field_value(SharedMatrix coords, SharedMatrix D) {
-    auto field_integrals_ = static_cast<ElectricFieldInt *>(integral_->electric_field().release());
+    auto field_integrals_ = std::unique_ptr<ElectricFieldInt>(static_cast<ElectricFieldInt *>(integral_->electric_field().release()));
 
     SharedMatrix efields = std::make_shared<Matrix>("efields", coords->nrow(), 3);
     auto fieldfun = ContractOverDensityFieldFunctor(efields, D);
--- a/psi4/src/psi4/libmints/mintshelper.cc
+++ b/psi4/src/psi4/libmints/mintshelper.cc
@@ -1719,7 +1719,7 @@ std::vector<SharedMatrix> MintsHelper::electric_field(const std::vector<double> 
 SharedMatrix MintsHelper::induction_operator(SharedMatrix coords, SharedMatrix moments) {
     SharedMatrix mat = std::make_shared<Matrix>("Induction operator", basisset_->nbf(), basisset_->nbf());
     ContractOverDipolesFunctor dipfun(moments, mat);
-    auto field_integrals_ = static_cast<ElectricFieldInt *>(integral_->electric_field());
+    auto field_integrals_ = static_cast<ElectricFieldInt *>(integral_->electric_field().release());
     field_integrals_->compute_with_functor(dipfun, coords);
     mat->scale(-1.0);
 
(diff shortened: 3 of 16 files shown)


psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1719,15 +1719,15 @@ std::vector<SharedMatrix> MintsHelper::electric_field(const std::vector<double>
 SharedMatrix MintsHelper::induction_operator(SharedMatrix coords, SharedMatrix moments) {
     SharedMatrix mat = std::make_shared<Matrix>("Induction operator", basisset_->nbf(), basisset_->nbf());
     ContractOverDipolesFunctor dipfun(moments, mat);
-    auto field_integrals_ = static_cast<ElectricFieldInt *>(integral_->electric_field());
+    auto field_integrals_ = static_cast<ElectricFieldInt *>(integral_->electric_field().release());
```
hmacdope, line 1722: Can this be `get()` rather than `release()`? I wasn't sure where the memory management occurs otherwise eg in `compute_with_functor`

psi4/src/psi4/libmints/orbitalspace.cc
```diff
@@ -148,9 +148,9 @@ SharedMatrix OrbitalSpace::overlap(const std::shared_ptr<BasisSet> &basis1, cons
     SharedMatrix Smat =
         std::make_shared<Matrix>("Overlap between space1 and space2", sobasis1.dimension(), sobasis2.dimension());
 
-    OneBodySOInt *S = mix_ints.so_overlap();
+    std::unique_ptr<OneBodySOInt> S = mix_ints.so_overlap();
     S->compute(Smat);
-    delete S;
+    S.reset();
```
hmacdope, line 153: This is a bit redundant but I kept them for clarity with original code

psi4/src/psi4/libmints/orbitalspace.cc
```diff
@@ -133,9 +133,9 @@ SharedMatrix OrbitalSpace::overlap(const OrbitalSpace &space1, const OrbitalSpac
     SharedMatrix Smat =
         std::make_shared<Matrix>("Overlap between space1 and space2", p1.SO_basisdim(), p2.SO_basisdim());
 
-    OneBodySOInt *S = mix_ints.so_overlap();
+    std::unique_ptr<OneBodySOInt> S = mix_ints.so_overlap();
     S->compute(Smat);
-    delete S;
+    S.reset();
```
hmacdope, line 138: This is a bit redundant but I kept them for clarity with original code

hmacdope (commented):
Few questions @JonathonMisiewicz :)

loriab:
[tests/test_erisieve.py](https://github.com/psi4/psi4/blob/master/tests/pytests/test_erisieve.py) is unhappy with the changes. You can run it locally via `<objdir># pytest -v ../tests/pytests/test_eriseive.py` after executing the results of `<objdir># stage/bin/psi4 --psiapi`.

psi4/src/psi4/libmints/mintshelper.cc
JonathonMisiewicz: Can this just be `auto field_integrals = integral_->electric_field()`? Looking at the code, there's no reason to bother with the static cast. The same applies below.

hmacdope: Seems like the static cast is required to resolve `compute_with_functor`. I have released, cast and then put in another unique_ptr, its a bit ugly but i suppose it works. Let me know if you want it done differently.

hmacdope:
> [tests/test_erisieve.py](https://github.com/psi4/psi4/blob/master/tests/pytests/test_erisieve.py) is unhappy with the changes. You can run it locally via `<objdir># pytest -v ../tests/pytests/test_eriseive.py` after executing the results of `<objdir># stage/bin/psi4 --psiapi`.

Thats strange, I couldn't reproduce locally ☹️

psi4/src/psi4/libmints/mintshelper.cc
JonathonMisiewicz: @jturney Why does `integral_->electric_field()` return a `OneBodyAoInt` rather than an `ElectricFieldInt`? I would assume it should return an object of the child class and leave it to the caller to decide when to upcast to a member of the parent class.

hmacdope:
@JonathonMisiewicz @loriab  I'm also a bit confused as to why none of the callsites for the `eri` integrals needed to be changes and everything compiled no sweat. I feel I am probably missing something?

loriab:
hmm, do `which psi4 python` and `python -c "import psi4;print(psi4.__file__, psi4.__version__)"` to see if anything looks suspicious (or post them here if you're unsure).

hmacdope:
> python -c "import psi4;print(psi4.__file__, psi4.__version__)"

```
/Users/hugomacdermott/Desktop/dev/psi4_build/bin/psi4
/Users/hugomacdermott/miniconda3/envs/flow/bin/python
/Users/hugomacdermott/Desktop/dev/psi4_build/lib/psi4/__init__.py 1.7a1.dev109
```

But this should not affect the cmake compilation in my objdir?

hmacdope:
@loriab to add to this the CI seems to **build** fine. Some of the tests no dice.

loriab:
If `/Users/hugomacdermott/Desktop/dev/psi4_build/` is your `CMAKE_INSTALL_PREFIX`, then this is fine, but you can make development a little easier by skilling the `install` step. In your `objdir/` is a `stage/` dir with a complete install that gets updated with every `make`. It's this installation that gets run with `ctest`. I'd seek that, then run the results of `objdir/stage/bin/psi4 --psiapi` and check that the `which` & `python -c ` commands returns paths in `objdir/stage/`. This way there's a quick turnaround between edit/make/test.

I don't see anything necessarily wrong with what you posted. The above is just a setup that can reduce confusion if you forget the install step.

hmacdope:
Ah yep thanks for the tips makes sense . It is my CMAKE_INSTALL_PREFIX. 

I'm still confused though as to how it's even compiling

JonathonMisiewicz:
> @JonathonMisiewicz @loriab I'm also a bit confused as to why none of the callsites for the `eri` integrals needed to be changes and everything compiled no sweat. I feel I am probably missing something?

A quick grep tells me that the raw pointers returned by `eri` were converted to `shared_ptr` almost immediately. That's the reason why, I'm sure.

hmacdope:
@JonathonMisiewicz @loriab I think I understand now, I thought you had to use an explicit `std::move` but that is only for an **`lvalue`**, you can do 

```c++
std::shared_ptr<TwoBodyAOInt>(factory->erf_eri(omega_)
```

where `erf_eri` returns a `std::unique_ptr` because its a  **temporary `rvalue`**

My bad sorry for the misunderstanding. 

Regardless I will track down the failing test.

psi4/src/export_mints.cc
```diff
@@ -1304,7 +1304,7 @@ void export_mints(py::module& m) {
         .def("shell_significant", compute_shell_significant(&TwoBodyAOInt::shell_significant),
              "Determines if the P,Q,R,S shell combination is significant");
 
-    py::class_<Libint2ERI, std::shared_ptr<Libint2ERI>>(m, "ERI", pyTwoBodyAOInt,
+    py::class_<Libint2ERI, std::unique_ptr<Libint2ERI>>(m, "ERI", pyTwoBodyAOInt,
```
hmacdope, line 1307: @loriab  @JonathonMisiewicz  After many hours staring at valgrind I figured out the issue was that the PyBind11 holder type for the `eri` needed to be a unique ptr to avoid  default deletion of the `ERI` prior to calling the `screen_impl_` lambda function. Do we anticipate this being an issue with the other `2BodyAOInts`? Are they well covered by tests?

JonathonMisiewicz: Can you go into more detail about what the original problem was and why shared-vs-unique matters? It isn't clear to me what default deletion of `ERI` means, and I can't see which test was failing.

psi4/src/psi4/libmints/mintshelper.cc
JonathonMisiewicz: While this should be a separate PR, `integral_->electric_field()` should return an `ElectricFieldInt` rather than a `OneBodyAOInt`. This will clean up `MintsHelper::induction_operator`. I'm not sure your version of the code handles memory correctly, but it can't be worse than the old way of doing it.

hmacdope: I can raise an issue.

psi4/src/export_mints.cc
hmacdope: Originally the tests in `test_erisieve.py` were failing due to a segfault when calling any of the methods associated with the `ERI` TwoBodyAOInt object, called through the `_screen_impl` lambda.

 This is because the `unique_ptr` holding the C++ TwoBodyAOInt goes is allowed to expire because the `sharedptr` that holds the python class object has no reference counts.  

Making it a `unique_ptr` seems to fix the reference counting on the python object but I'm not exactly sure why.

hmacdope: Perhaps it is because the lambda doesn't extend the lifetime, but I'm fairly sure the issue is with the PyBind11 layer. Is anyone here in particular a Pybind expert?

psi4/src/psi4/libmints/mintshelper.cc
hmacdope: raised #2793

hmacdope:
How do we want to move forward here @JonathonMisiewicz @loriab etc? :)

JonathonMisiewicz:
Apologies for the delay; you decided to get this PR moving again when Lori and I were both extremely busy with PsiCon and 1.8.

The one outstanding issue here seems to be the Pybind holder conversion. I'll look that over later today.

psi4/src/psi4/libmints/x2cint.cc
```diff
@@ -144,7 +144,7 @@ void X2CInt::compute_integrals() {
     if ((lambda_[0] != 0.0) or (lambda_[1] != 0) or (lambda_[2] != 0)) {
         OperatorSymmetry msymm(1, molecule_, integral_, soFactory_);
         std::vector<SharedMatrix> dipoles = msymm.create_matrices("Dipole");
-        OneBodySOInt *so_dipole = integral_->so_dipole();
+        std::unique_ptr<OneBodySOInt> so_dipole = integral_->so_dipole();
```
JonathonMisiewicz, line 147: Can't we use `auto` here?

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -525,7 +525,7 @@ void FISAPT::nuclear() {
 
     auto Vfact = std::make_shared<IntegralFactory>(primary_);
     std::shared_ptr<PotentialInt> Vint;
-    Vint = std::shared_ptr<PotentialInt>(static_cast<PotentialInt*>(Vfact->ao_potential()));
+    Vint = std::shared_ptr<PotentialInt>(static_cast<PotentialInt*>(Vfact->ao_potential().release()));
```
JonathonMisiewicz, line 528: Does `Vint = Vfact->ao_potential()` not suffice here (and many other places throughout this PR)?

JonathonMisiewicz (commented):
Now that the holidays are over (and I remembered about this PR, whoops), time to start getting this in.

psi4/src/psi4/fisapt/fisapt.cc
hmacdope: Not without changing `std::shared_ptr<PotentialInt> Vint;` on L527, Im not libmints expert but does this have downstream effects or are we all good to cast to PotentialInt?

hmacdope: Lots of cases like this in this PR so probably good to decide what the strategy is.

JonathonMisiewicz: To be more explicit: this line of code is both downcasting the underlying object and changing the type of the holding smart pointer to a shared pointer. I missed that first point in my original comment.

The underlying `ao_potential` method should be changed so that this downcasting isn't necessary, but that's beyond scope for this PR. This can stay as-is, begrudgingly, knowing that the removal of this ugliness is deferred to a follow-up PR. I'll do it if need be.

hmacdope: I see. This is in the same vein as #2793 and related PR #2795.  Perhaps should make a more general issue. Happy to help out fixing.

JonathonMisiewicz (approved):
Good enough for now. The integral factory methods need to stop upcasting their return values in a different PR, but otherwise, LGTM.

jturney (approved):

JonathonMisiewicz:
Is there a reason not to merge this now?

loriab (approved):
lgtm, thanks for the modernization.

JonathonMisiewicz:
> Is there a reason not to merge this now?

To be clear, this is directed @hmacdope, since you don't have "Ready to merge" checked.

PR #2788, opened 2022-11-21
Begin moving OPDM management out of libtrans

JonathonMisiewicz:
At long last, some gradient refactoring. Supersedes [my previous attempt](https://github.com/psi4/psi4/pull/1745). Only the OPDM and Lagrangian are a 1.7 target.

- [x] Orbital-optimized `occ` densities now available on the wavefunction
- [x] DC-06 densities now symmetrized

## Dev notes & details

- [x] Conventional gradients now check whether the density and Lagrangian are set on the wavefunction. If so, it uses those rather than reading MO-basis quantities from disk. `dct` and `occ` use this new system. `cc` will use it in the next PR. The old system will be removed for 1.8. Old system vs new system is toggled by the `reset_oneel` flag.

## Questions
- [ ] How to create a warning that plugin developers should use new-style gradients?

## Checklist
- [x] All `dct`, `omp`, `mp`, `olccd`, lccd`, `cepa`, and `cc` tests pass

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libtrans/integraltransform.h
+++ b/psi4/src/psi4/libtrans/integraltransform.h
@@ -155,7 +155,9 @@ class PSI_API IntegralTransform {
     void transform_tei_first_half(const std::shared_ptr<MOSpace> s1, const std::shared_ptr<MOSpace> s2);
     void transform_tei_second_half(const std::shared_ptr<MOSpace> s1, const std::shared_ptr<MOSpace> s2,
                                    const std::shared_ptr<MOSpace> s3, const std::shared_ptr<MOSpace> s4);
-    void backtransform_density();
+    // WARNING! reset_oneel is set to true for backwards compatibility. Soon, this option will be removed
+    // and only the reset_oneel = false logic will be available.
+    void backtransform_density(bool reset_oneel = true);
     void backtransform_tpdm_restricted();
     void backtransform_tpdm_unrestricted();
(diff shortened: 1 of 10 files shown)


psi4/src/psi4/libtrans/integraltransform.h
```diff
@@ -155,7 +155,9 @@ class PSI_API IntegralTransform {
     void transform_tei_first_half(const std::shared_ptr<MOSpace> s1, const std::shared_ptr<MOSpace> s2);
     void transform_tei_second_half(const std::shared_ptr<MOSpace> s1, const std::shared_ptr<MOSpace> s2,
                                    const std::shared_ptr<MOSpace> s3, const std::shared_ptr<MOSpace> s4);
-    void backtransform_density();
+    // WARNING! reset_oneel is set to true for backwards compatibility. Soon, this option will be removed
+    // and only the reset_oneel = false logic will be available.
+    void backtransform_density(bool reset_oneel = true);
```
loriab, line 160: By changing the signature, you've already required plugins+thisPsi to recompile, so you could add a message to the plugin outfile like we used to do for wfn: https://github.com/psi4/psi4/blob/1.4.x/psi4/src/export_plugins.cc#L111-L112 .

loriab (approved):

maxscheurer (approved):

PR #2795, opened 2022-11-24
Refactor `libmints::IntegralFactory.electric_field()` to return ElectricFieldInt

### LINKED ISSUE #2793: Make `electric_field()` return an `ElectricFieldInt` rather than a `OneBodyAOInt``
While this should be a separate PR, `integral_->electric_field()` should return an `ElectricFieldInt` rather than a `OneBodyAOInt`. This will clean up `MintsHelper::induction_operator`. 

_Originally posted by @JonathonMisiewicz in https://github.com/psi4/psi4/pull/2775#discussion_r1030527587_

hmacdope:
## Description

Refactor `electric_field()` to return ElectricFieldInt rather than OneBodyAOInt

Fixes #2793 

- [x] ElectricFieldInt returned now by `IntegralFactory.electric_field()`
- [x] Equivalent changes in `libmintshelper` made to stop immediate `static cast`

## Dev notes & details

- [x] ElectricFieldInt returned now by `IntegralFactory.electric_field()`
- [x] Equivalent changes in `libmintshelper`

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/integral.cc
+++ b/psi4/src/psi4/libmints/integral.cc
@@ -197,7 +197,7 @@ OneBodySOInt* IntegralFactory::so_traceless_quadrupole() {
     return new OneBodySOInt(ao_int, this);
 }
 
-OneBodyAOInt* IntegralFactory::electric_field(int deriv) {
+ElectricFieldInt* IntegralFactory::electric_field(int deriv) {
     return new ElectricFieldInt(spherical_transforms_, bs1_, bs2_, deriv);
 }
 
--- a/psi4/src/psi4/libmints/integral.h
+++ b/psi4/src/psi4/libmints/integral.h
@@ -38,6 +38,7 @@ PRAGMA_WARNING_POP
 
 #include "onebody.h"
 #include "twobody.h"
+#include "electricfield.h"
 
 
 namespace psi {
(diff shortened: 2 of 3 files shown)


maxscheurer:
Isn't the purpose of the integral factory that one does not return the ptr to the specific integral class? Do we want to change the other integral types in the factory, too?

hmacdope:
> Isn't the purpose of the integral factory that one does not return the ptr to the specific integral class? Do we want to change the other integral types in the factory, too?

I was just following a suggestion by @JonathonMisiewicz documented in #2793. Happy to close if required :)

JonathonMisiewicz:
> Isn't the purpose of the integral factory that one does not return the ptr to the specific integral class? Do we want to change the other integral types in the factory, too?

The point of the integral factory is that we don't need to pass the same arguments, over and over. Returning a pointer to the base class rather than the derived class gets us nothing. C++ will cast to the base class when necessary.

While I support changing the other integral types as well, I won't insist on it for this PR.

hmacdope:
If so shall we push forward for this small change to avoid conflicts with #2775 that would occur if we changed everything?

JonathonMisiewicz:
Fine by me.

hmacdope:
@JonathonMisiewicz have we had any additional thoughts here? #2775 is probably a blocker

JonathonMisiewicz:
No. I agree that getting #2775 in first is best.

PR #2797, opened 2022-11-25
Do not use or suggest the undocumented -H option for CMake

TiborGY:
## Description

The `-H<path>` option for CMake is technically not part of the CMake public interface, it is not documented and was never meant for public use. Since CMake 3.13 there is a proper alternative, `-S<path>`, so Psi4 should use that and stop suggesting users to use `-H<path>`.

- [x] Psi4 no longer suggests users to use the undocumented `-H<path>` CMake option when building from source.

## Dev notes & details

- [x] Replace `-H<path>`  with `-S<path>` in CMake invokations.

## Questions
- [ ] I have no idea where to make a PR for fixing this in psi4-path-advisor.

## Checklist
- [x] No new features
- [x] CI tests are passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -117,7 +117,7 @@ jobs:
       #   b/c their c++ symbols don't mix with the different Travis compilers. for other
       #   reasons, pkgs are being compiled less statically, sad for CI.
       # * can't enable trivial plugins b/c no psi4 for them to detect at start (e.g., snsmp2)
-      cmake -Bbuild -H. \
+      cmake -Bbuild -S. \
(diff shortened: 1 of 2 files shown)


maxscheurer:
`psi4-path-advisor.py` is [here](https://github.com/psi4/psi4meta/blob/master/conda-recipes/psi4-dev/src/psi4-path-advisor.py) in psi4meta.

loriab (approved):
Good call, thanks. I got the main one https://github.com/psi4/psi4/blob/master/CMakeLists.txt#L30 but forgot these. I'll catch the psi4-path-advisor one if you haven't already gotten there.

maxscheurer (approved):

PR #2808, opened 2022-11-29
IncFock Standardization Part 2: DirectJK

davpoolechem:
## Description
This PR is a continuation of the IncFock standardization effort initially started in https://github.com/psi4/psi4/pull/2682, and restarted in https://github.com/psi4/psi4/pull/2792. It is the second PR listed in the PR submission scheme discussed in the comments in https://github.com/psi4/psi4/pull/2682.

For all intents and purposes, the PR has basically the same changes as https://github.com/psi4/psi4/pull/2792, except applied to DirectJK instead of DFJLinK. The same benefits to DFJLinK seen in https://github.com/psi4/psi4/pull/2792 are applied to DirectJK here, and many of the same points of the benefits of that PR apply here, as well.

A couple of notes to be had for reviewers, some of which arise from discussions in https://github.com/psi4/psi4/pull/2792:
1. Unlike https://github.com/psi4/psi4/pull/2792, this PR is independent of the CompositeJK effort, since this PR is localized within DirectJK.
2. This PR makes slightly more expansive changes to the DirectJK infrastructure to support the new IncFock formalism, but it is still not a massive number of changes; and most of these changes are similar to those observed in DFJLinK in https://github.com/psi4/psi4/pull/2792.
3. As with https://github.com/psi4/psi4/pull/2792, this PR does not remove the bells and whistles related to IncFock seen in DirectJK (e.g., recomputing the full Fock matrix every so often, disabling IncFock past a given convergence threshold). This PR incorporates the DFJCOSK IncFock formalism as much as possible into DirectJK, while maintaining the IncFock bells and whistles seen in DirectJK.
4. It should be noted that there are no directJK-x tests akin to the linK-x tests, and thus, no IncFock efficiency tests were added in this PR, as was done for DFJLinK in https://github.com/psi4/psi4/pull/2792. Technically, IncFock efficiency tests could be added to scf5, but I feel that goes against the spirit of the scf5 test.
5. I looked at stability tests, as well. STABILITY_ANALYSIS=CHECK works for all of RHF, ROHF, and UHF references.

N/A

## Dev notes & details
- [X] Switches DirectJK to using the incremental Fock build implementation used in DFJCOSK. This change standardizes the incremental Fock implementation between DirectJK and DFJCOSK and improves the memory usage of DirectJK in the process.
- [X] Changes DirectJK machinery to support new incremental Fock formalism.

## Questions
- [x] Do we want to add extra tests for DirectJK to allow for Incremental Fock efficiency tests?

## Checklist

## Status
- [X] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/DirectJK.cc
+++ b/psi4/src/psi4/libfock/DirectJK.cc
@@ -161,7 +161,7 @@ void DirectJK::incfock_postiter() {
 }
 
 void DirectJK::compute_JK() {
-
+   
 #ifdef USING_BrianQC
     if (brianEnable) {
         // zero out J, K, and wK matrices
@@ -396,6 +396,22 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
     if (!build_J && !build_K) return;
     
     timer_on("build_JK_matrices()");
+
+    // => Zeroing... <= //
+    
+    // Ideally, this wouldnt be here at all
+    // It would be better covered in incfock_setup()
+    // But removing this causes a couple of tests to fail for some reason
+    
+    if (!do_incfock_iter_) {
+        for (auto& Jmat : J) {
+            Jmat->zero();
+        }
+    
+        for (auto& Kmat : K) {
+            Kmat->zero();
+        }
+    }
 
     // => Sizing <= //
 
--- a/psi4/src/psi4/libfock/DirectJK.cc
+++ b/psi4/src/psi4/libfock/DirectJK.cc
@@ -135,12 +135,12 @@ void DirectJK::incfock_setup() {
         size_t njk = D_ao_.size();
 
         // If there is no previous pseudo-density, this iteration is normal
-        if(initial_iteration_ || D_prev_.size() != njk) {
-            initial_iteration_ = true;
+        if (initial_iteration_ || D_prev_.size() != njk) {
+	        initial_iteration_ = true;
 
             D_ref_ = D_ao_;
             zero();
-        } else { // Otherwise, the iteraction is incremental
+        } else { // Otherwise, the iteration is incremental
             for (size_t jki = 0; jki < njk; jki++) {
                 D_ref_[jki] = D_ao_[jki]->clone();
                 D_ref_[jki]->subtract(D_prev_[jki]);
--- a/psi4/src/psi4/libfock/DirectJK.cc
+++ b/psi4/src/psi4/libfock/DirectJK.cc
@@ -162,12 +162,12 @@ void DirectJK::incfock_postiter() {
 
 void DirectJK::compute_JK() {
 
-    // zero out J, K, and wK matrices
-    zero();
-
 #ifdef USING_BrianQC
     if (brianEnable) {
-        brianBool computeCoulomb = (do_J_ ? BRIAN_TRUE : BRIAN_FALSE);
+        // zero out J, K, and wK matrices
+        zero();
+        
+	brianBool computeCoulomb = (do_J_ ? BRIAN_TRUE : BRIAN_FALSE);
         brianBool computeExchange = ((do_K_ || do_wK_) ? BRIAN_TRUE : BRIAN_FALSE);
 
         if (do_wK_ and not brianEnableDFT) {
@@ -397,15 +397,6 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
     
     timer_on("build_JK_matrices()");
 
-    // => Zeroing <= //
-    for (auto& Jmat : J) {
-        Jmat->zero();
-    }
-    
-    for (auto& Kmat : K) {
-        Kmat->zero();
-    }
-    
     // => Sizing <= //
 
     int nshell = primary_->nshell();
@@ -718,6 +709,17 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
         if (!touched) continue;
 
         // => Stripe out <= //
+        if (build_J) {
+	    for (auto& JTmat : JT[thread]) {
+                JTmat->scale(2.0);
+	    }
+        }
+        
+        if (build_K) {
+	    for (auto& KTmat : KT[thread]) {
+                KTmat->scale(2.0);
+	    }
+        }
 
         // if (thread == 0) timer_on("JK: Atomic");
         for (size_t ind = 0; ind < D.size(); ind++) {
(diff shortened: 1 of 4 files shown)


psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -131,93 +131,42 @@ void DirectJK::preiterations() {
...
 void DirectJK::compute_JK() {
-
-    // zero out J, K, and wK matrices
-    zero();
-
+   
 #ifdef USING_BrianQC
     if (brianEnable) {
+        // zero out J, K, and wK matrices
+        zero();
```
loriab, line 168: just checking that editing brianqc code intentional.

loriab (approved):
Thanks for looking at stability. Probably later we can add a test for each of the stabilities that parameterizes all the scf methods and incfock combinations.

davpoolechem:
> Thanks for looking at stability. Probably later we can add a test for each of the stabilities that parameterizes all the scf methods and incfock combinations.

You're welcome! This would probably be a good idea, especially since, from my testing, some issues pop up with stability tests UHF references depending on the SCF_TYPE method used currently.

psi4/src/psi4/libfock/DirectJK.cc
davpoolechem: Indeed it is!

The original DirectJK code has a call to zero() at the very beginning of compute_JK(). Unfortunately, this zero() call breaks the IncFock code when using the new IncFock formalism, so essentially, the call to zero() was moved down further in DirectJK. _However_, the call to zero() also occurs before the BrianQC code. So, I assumed that just flat-out getting rid of the call to zero() at the beginning would break the BrianQC code. The addition of zero() to BrianQC, then, is essentially making sure that zero() is still called at the beginning when BrianQC is used.

Thinking about it further, though, it's possible this same effect can be had by just replacing the original call to zero() with something like

`if (!incfock_) zero();`

Then, the zero() call is only done for non-IncFock runs without changing the BrianQC code at all. Might be safer to do it that way, since I am not sure if we have immediate access to BrianQC to test changes to BrianQC code.

loriab: We don't have immediate BrianQC access, so I think the consolidated `zero()` you propose would be good.

JonathonMisiewicz (approved):
This all looks good to me. Let us know when this is ready to merge.

davpoolechem:
> This all looks good to me. Let us know when this is ready to merge.

Will do! I think I want to run a quick final test set with the DFJLinK PR merged in, to make sure nothing broke. Then, it should be good to go!

psi4/src/psi4/libfock/DirectJK.cc
davpoolechem: Will do, thanks! My only concern is some edge case where someone uses BrianQC while requesting IncFock, then the initial zeroing doesn't happen, and then BrianQC possibly breaks.

davpoolechem:
So the CI error here is... interesting.

`FAILED: src/psi4/libdpd/CMakeFiles/dpd.dir/buf4_mat_irrep_wrt.cc.obj 
C:\PROGRA~1\LLVM\bin\clang-cl.exe  /nologo -TP -DUSING_LAPACK_MKL -D_CRT_NONSTDC_NO_DEPRECATE -D_CRT_NONSTDC_NO_WARNINGS -D_CRT_SECURE_NO_WARNINGS -D_USE_MATH_DEFINES -ID:\a\1\s\psi4\include -ID:\a\1\s\psi4\src -imsvc C:\tools\miniconda3\include -imsvc C:\tools\miniconda3\Library\include -imsvc C:\tools\miniconda3\Library\include\eigen3 /arch:AVX /O2 /Ob2 /DNDEBUG -MD /EHsc -Xclang -fopenmp -std:c++17 /showIncludes /Fosrc\psi4\libdpd\CMakeFiles\dpd.dir\buf4_mat_irrep_wrt.cc.obj /Fdsrc\psi4\libdpd\CMakeFiles\dpd.dir\dpd.pdb -c -- D:\a\1\s\psi4\src\psi4\libdpd\buf4_mat_irrep_wrt.cc
In file included from D:\a\1\s\psi4\src\psi4\libdpd\buf4_mat_irrep_wrt.cc:33:
In file included from C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.34.31933\include\cstdio:9:
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.34.31933\include\yvals_core.h(784,1): error: static_assert failed "Error in C++ Standard Library usage."
_EMIT_STL_ERROR(STL1000, "Unexpected compiler version, expected Clang 14.0.0 or newer.");
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.34.31933\include\yvals_core.h(456,5): note: expanded from macro '_EMIT_STL_ERROR'
    static_assert(false, "Error in C++ Standard Library usage.")
    ^             ~~~~~
1 error generated.`

Two overall errors on the Windows Psi4 build in this vein.

loriab:
I've gotten that once today, too. Then I've gotten distracted by other matters. Even if a restart fixes it, another CI failure will crop up later on, so please rebase, and we'll see if this is a consistent issue.

loriab:
I got the static_assert error again now. The file it's triggering on hasn't had recent changes, so I don't know why it's popping up now, unless it's an intermittant/brownout strategy to push upgrades. Anyways, switching the llvm version https://github.com/psi4/psi4/pull/2791/files#diff-9f5f8e4d87a7d1fbb0e8f30d1c680c5f42f6af10c04aa81d342c7dfa5af9a99fR10 at least gets the build started.

davpoolechem:
> I got the static_assert error again now. The file it's triggering on hasn't had recent changes, so I don't know why it's popping up now, unless it's an intermittant/brownout strategy to push upgrades. Anyways, switching the llvm version https://github.com/psi4/psi4/pull/2791/files#diff-9f5f8e4d87a7d1fbb0e8f30d1c680c5f42f6af10c04aa81d342c7dfa5af9a99fR10 at least gets the build started.

Hmmmm, I see. Thank you for looking into this! I just rebased and pushed again, so we will see if that helps.

loriab:
The LLVM bump was a trial in a branch, not master, so you'll have to edit, not just rebase. Sorry I was unclear.

PR #2816, opened 2022-12-01
IncFock Standardization Part 3: DFJCOSK

davpoolechem:
## Description
... and another one!

This PR is a continuation of the IncFock standardization effort initially started in https://github.com/psi4/psi4/pull/2682, and continued in https://github.com/psi4/psi4/pull/2792 and https://github.com/psi4/psi4/pull/2808.

This PR continues to standardize the Incremental Fock process across the current integral-direct JK algorithms present in Psi4. However, this PR stands somewhat in contrast to https://github.com/psi4/psi4/pull/2682 and https://github.com/psi4/psi4/pull/2792. In those PRs, their IncFock schemes were adapted to use that of DFJCOSK, without removing the bells and whistles of their IncFock implementations (e.g., recomputing the full Fock matrix every so often, disabling IncFock after a given convergence threshold). In contrast, DFJCOSK is the template IncFock upon which the two previous PRs were based; however, DFJCOSK does not have the IncFock bells and whistles that DirectJK and DFJLinK had. Unmitigated, the incremental Fock procedure can actually significantly increase the number of SCF iterations needed to converged; and the IncFock bells and whistles in DirectJK and DFJLinK notably mitigate the effect of IncFock on SCF convergence counts. Without these bells and whistles, DFJCOSK runs the significant risk of excessive SCF iterations needed to converge when IncFock is enabled, an issue I have run into in my own calculations.

This PR seeks to rectify the aforementioned issue by implementing the IncFock bells and whistles of DirectJK and DFJLinK into DFJCOSK. With these extra features, DFJCOSK can now recompute the full Fock matrix every n iterations, or disable IncFock past a given convergence threshold, at the will of the end user. These features can notably reduce the amount of SCF iterations needed to converge.

This PR is also the continued effort to standardize IncFock among all integral-direct SCF algorithms in Psi4. With this PR, DirectJK, DFJLinK, and DFJCOSK will have effectively the exact same incremental Fock schemes, each featuring the lower memory usage of DFJCOSK's former IncFock scheme, and the bells and whistles of DirectJK and DFJLinK's former IncFock schemes. The final step to the IncFock standardization process, then, is the movement of the IncFock scheme of these algorithms into the base JK class itself - the final step of the PR scheme discussed in the comments in https://github.com/psi4/psi4/pull/2682.

It is worth noting for the reviewers that, since this PR more closely matches the IncFock of DFJCOSK to that of DFJLinK, it should be considered relevant to the CompositeJK PR chain as well, as it smooths the combination of DFJLinK and DFJCOSK into a single class as planned in the next CompositeJK step.

- [X] DFJCOSK now uses the "INCFOCK" keyword to control usage of the incremental Fock procedure, rather than "COSX_INCFOCK", standardizing the incremental Fock keywords among the integral-direct SCF algorithms.
- [X] DFJCOSK can now recompute the Fock matrix every n iterations during a calculation using incremental Fock. The "INCFOCK_FULL_FOCK_EVERY" keyword controls the number of SCF iterations between full Fock matrix recomputations for DFJCOSK.
- [X] DFJCOSK can now disable the incremental Fock process when a specific SCF convergence threshold. The "INCFOCK_CONVERGENCE" keyword controls the SCF convergence threshold point at which the incremental Fock build process is disabled.

## Dev notes & details
- [X] Improves the DFJCOSK incremental Fock procedure by allowing it to recompute the full Fock matrix every n iterations.
- [X] Improves the DFJCOSK incremental Fock procedure by allowing it to disable the incremental Fock procedure past a certain SCF convergence threshold.
- [X] Improves the DFJCOSK incremental Fock prodecure by computing the full Fock matrix during the first SCF iteration of the calculation, even with INCFOCK enabled.
- [X] Adds incremental Fock build efficiency tests to test_dfjcosk.py. 
- [X] Updates documentation to reflect new changes.

## Questions
N/A

## Checklist

## Status
- [X] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/DFJCOSK.cc
+++ b/psi4/src/psi4/libfock/DFJCOSK.cc
@@ -318,8 +318,10 @@ void DFJCOSK::incfock_setup() {
 	size_t njk = D_ao_.size();
 
         // If there is no previous pseudo-density, this iteration is normal
-        if(D_prev_.size() != njk) {
-            D_ref_ = D_ao_;
+        if(initial_iteration_ || D_prev_.size() != njk) {
+            initial_iteration_ = true;
+	    
+	    D_ref_ = D_ao_;
             zero();
         } else { // Otherwise, the iteraction is incremental
             for (size_t jki = 0; jki < njk; jki++) {
@@ -349,10 +351,14 @@ void DFJCOSK::compute_JK() {
     if (incfock_) {
         timer_on("DFJCOSK: INCFOCK Preprocessing");
 
-        // always do incfock on this iteration
-	// TODO: Adds bells and whistles from DFJLinK
-        do_incfock_iter_ = incfock_; 
-        
+        int reset = options_.get_int("INCFOCK_FULL_FOCK_EVERY");
+        double incfock_conv = options_.get_double("INCFOCK_CONVERGENCE");
+        double Dnorm = Process::environment.globals["SCF D NORM"];
+        // Do IFB on this iteration?
+        do_incfock_iter_ = (Dnorm >= incfock_conv) && !initial_iteration_ && (incfock_count_ % reset != reset - 1);
+
+        if (!initial_iteration_ && (Dnorm >= incfock_conv)) incfock_count_ += 1;
+	
 	incfock_setup();
         
 	timer_off("DFJCOSK: INCFOCK Preprocessing");
@@ -364,23 +370,30 @@ void DFJCOSK::compute_JK() {
     // Direct DF-J
     if (do_J_) {
         timer_on("DFJ");
-        build_J(D_ref_, J_ao_);
-        timer_off("DFJ");
+   
+   	build_J(D_ref_, J_ao_);
+   
+   	timer_off("DFJ");
     }
     
     // COSX
     if (do_K_) {
         timer_on("COSK");
-        build_K(D_ref_, K_ao_);
-        timer_off("COSK");
+   
+   	build_K(D_ref_, K_ao_);
+   
+   	timer_off("COSK");
     }
 
     // => Finalize Incremental Fock if required <= //
+
     if (incfock_) {
         timer_on("DFJCOSK: INCFOCK Postprocessing");
         incfock_postiter();
         timer_off("DFJCOSK: INCFOCK Postprocessing");
     }
+    
+    if (initial_iteration_) initial_iteration_ = false;
 }
 
 void DFJCOSK::postiterations() {}
--- a/psi4/src/psi4/libfock/DFJCOSK.cc
+++ b/psi4/src/psi4/libfock/DFJCOSK.cc
@@ -165,6 +165,9 @@ void DFJCOSK::common_init() {
     // and a large grid for the final iterations
     early_screening_ = true;
 
+    // incremental Fock setup
+    incfock_ = options_.get_bool("COSX_INCFOCK");
+    
     // => Direct Density-Fitted Coulomb Setup <= //
 
     // pre-compute coulomb fitting metric
@@ -298,7 +301,7 @@ void DFJCOSK::print_header() const {
         if (do_wK_) outfile->Printf("    Omega:              %11.3E\n", omega_);
         outfile->Printf("    Integrals threads:  %11d\n", nthreads_);
         outfile->Printf("    Memory [MiB]:       %11ld\n", (memory_ *8L) / (1024L * 1024L));
-        outfile->Printf("    Incremental Fock :  %11s\n", (options_.get_bool("COSX_INCFOCK") ? "Yes" : "No"));
+        outfile->Printf("    Incremental Fock :  %11s\n", (incfock_ ? "Yes" : "No"));
         outfile->Printf("    J Screening Type:   %11s\n", screen_type.c_str());
         outfile->Printf("    J Screening Cutoff: %11.0E\n", cutoff_);
         outfile->Printf("    K Screening Cutoff: %11.0E\n", options_.get_double("COSX_INTS_TOLERANCE"));
@@ -310,55 +313,76 @@ void DFJCOSK::print_header() const {
 
 void DFJCOSK::preiterations() {}
 
-void DFJCOSK::compute_JK() {
-
-    int njk = D_ao_.size();
-
-    // range-separated semi-numerical exchange needs https://github.com/psi4/psi4/pull/2473
-    if (do_wK_) throw PSIEXCEPTION("COSK does not support wK integrals yet!");
-
-    // D_eff, the effective pseudo-density matrix is either:
-    //   (1) the regular density: D_eff == D_lr = C_lo x C*ro
-    //   (2) the difference density: D_eff == dD_lr = (C_lo x C_ro)_{iter} - (C_lo x C_ro)_{iter - 1}
-    //
-    std::vector<SharedMatrix> D_eff(njk);
-
-    if(options_.get_bool("COSX_INCFOCK")) {
+void DFJCOSK::incfock_setup() {
+    if (do_incfock_iter_) {
+	size_t njk = D_ao_.size();
 
         // If there is no previous pseudo-density, this iteration is normal
         if(D_prev_.size() != njk) {
-            D_eff = D_ao_;
+            D_ref_ = D_ao_;
             zero();
         } else { // Otherwise, the iteraction is incremental
             for (size_t jki = 0; jki < njk; jki++) {
-                D_eff[jki] = D_ao_[jki]->clone();
-                D_eff[jki]->subtract(D_prev_[jki]);
+                D_ref_[jki] = D_ao_[jki]->clone();
+                D_ref_[jki]->subtract(D_prev_[jki]);
             }
         }
+    } else {
+        D_ref_ = D_ao_;
+        zero();
+    }
+}
 
+void DFJCOSK::incfock_postiter() {
+    if (do_incfock_iter_) {
         // Save a copy of the density for the next iteration
         D_prev_.clear();
         for(auto const &Di : D_ao_) {
             D_prev_.push_back(Di->clone());
         }
+    }
+}
 
+void DFJCOSK::compute_JK() {
+    // range-separated semi-numerical exchange needs https://github.com/psi4/psi4/pull/2473
+    if (do_wK_) throw PSIEXCEPTION("COSK does not support wK integrals yet!");
+
+    // explicit setup of Incfock for this SCF iteration
+    if (incfock_) {
+        timer_on("DFJCOSK: INCFOCK Preprocessing");
+
+        // always do incfock on this iteration
+	// TODO: Adds bells and whistles from DFJLinK
+        do_incfock_iter_ = incfock_; 
+        
+	incfock_setup();
+        
+	timer_off("DFJCOSK: INCFOCK Preprocessing");
     } else {
-        D_eff = D_ao_;
+	D_ref_ = D_ao_;
         zero();
     }
-
+ 
+    // Direct DF-J
     if (do_J_) {
         timer_on("DFJ");
-        build_J(D_eff, J_ao_);
+        build_J(D_ref_, J_ao_);
         timer_off("DFJ");
     }
     
+    // COSX
     if (do_K_) {
         timer_on("COSK");
-        build_K(D_eff, K_ao_);
+        build_K(D_ref_, K_ao_);
         timer_off("COSK");
     }
 
+    // => Finalize Incremental Fock if required <= //
+    if (incfock_) {
+        timer_on("DFJCOSK: INCFOCK Postprocessing");
+        incfock_postiter();
+        timer_off("DFJCOSK: INCFOCK Postprocessing");
+    }
 }
 
 void DFJCOSK::postiterations() {}
--- a/psi4/src/psi4/libfock/DFJCOSK.cc
+++ b/psi4/src/psi4/libfock/DFJCOSK.cc
@@ -315,7 +315,7 @@ void DFJCOSK::preiterations() {}
 
 void DFJCOSK::incfock_setup() {
     if (do_incfock_iter_) {
-	    size_t njk = D_ao_.size();
+	    auto njk = D_ao_.size();
 
         // If there is no previous pseudo-density, this iteration is normal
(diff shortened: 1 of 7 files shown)


davpoolechem:
So the CI issue on this one seems... interesting, as well.

`2022-12-01T16:23:23.8944651Z Err:1 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libjsoncpp1 amd64 1.7.4-3.1ubuntu2
2022-12-01T16:23:23.8945670Z   503  Service Unavailable [IP: 52.147.219.192 80]
2022-12-01T16:23:28.9307925Z Get:2 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libllvm6.0 amd64 1:6.0.1-14 [15.2 MB]
2022-12-01T16:29:39.8525876Z Get:3 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libclang-common-6.0-dev amd64 1:6.0.1-14 [3015 kB]
2022-12-01T16:30:07.1867680Z Get:4 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libclang1-6.0 amd64 1:6.0.1-14 [7472 kB]
2022-12-01T16:31:01.2307594Z Get:5 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 clang-6.0 amd64 1:6.0.1-14 [9831 kB]
2022-12-01T16:32:37.6678852Z Get:6 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libomp5-10 amd64 1:10.0.0-4ubuntu1 [300 kB]
2022-12-01T16:32:38.5410720Z Get:7 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libomp-10-dev amd64 1:10.0.0-4ubuntu1 [47.7 kB]
2022-12-01T16:32:38.5769178Z Get:8 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 llvm-6.0-runtime amd64 1:6.0.1-14 [207 kB]
2022-12-01T16:32:38.6670727Z Get:9 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 llvm-6.0 amd64 1:6.0.1-14 [4889 kB]
2022-12-01T16:33:14.3718902Z Get:10 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 llvm-6.0-dev amd64 1:6.0.1-14 [24.0 MB]
2022-12-01T16:35:33.4054716Z Get:11 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libomp-dev amd64 1:10.0-50~exp1 [2824 B]
2022-12-01T16:35:33.4069243Z Fetched 65.0 MB in 12min 10s (89.1 kB/s)
2022-12-01T16:35:33.4339396Z E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/libj/libjsoncpp/libjsoncpp1_1.7.4-3.1ubuntu2_amd64.deb  503  Service Unavailable [IP: 52.147.219.192 80]
2022-12-01T16:35:33.4341544Z E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
2022-12-01T16:35:33.4444758Z ##[error]Bash exited with code '100'.
2022-12-01T16:35:33.4494014Z ##[section]Finishing: Apt-Get Packages
2022-12-01T16:35:33.6248939Z ##[section]Starting: Checkout psi4/psi4@refs/pull/2816/merge to s
2022-12-01T16:35:33.6261343Z `

Seems to be an issue with loading and installing packages with apt-get? Maybe this is a one-time thing that can be fixed with a CI restart.

tests/pytests/test_dfjcosk.py
```diff
@@ -86,3 +85,60 @@ def test_dfjcosk(inp, mols):
...
+    # compute DFJCOSK energy+wfn with Incfock 
+    psi4.set_options({"incfock" : True})
+    energy_dfjcosk_inc, wfn_dfjcosk_inc = psi4.energy(inp["method"], molecule=molecule, bsse_type=inp["bsse_type"], return_wfn=True)
+
+    # how do energies compare?
+    assert compare_values(energy_dfjcosk_noinc, energy_dfjcosk_inc, atol=1e-6)
+    
+    # how do SCF iteration counts compare?
+    if hasattr(wfn_dfjcosk_noinc, "iteration_") and hasattr(wfn_dfjcosk_inc, "iteration_"):
+        assert compare_values(wfn_dfjcosk_noinc.iteration_, wfn_dfjcosk_inc.iteration_, 3)
```
loriab, line 144: `niter_noinc = int(wfn_dfjcosk_noinc.variable("SCF ITERATIONS"))` can give you the count, too

I'm not sure what you're doing in the assert. why would the the niter be the same to 1.e-3? You could do `assert compare(True, (niter_noinc - niter_inc) >= 3, "reduced by at least 3 iterations by incfock")` if that's what you had in mind.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -310,55 +313,87 @@ void DFJCOSK::print_header() const {
...
+   
+   	timer_off("COSK");
     }
 
+    // => Finalize Incremental Fock if required <= //
+
+    if (incfock_) {
+        timer_on("DFJCOSK: INCFOCK Postprocessing");
+        incfock_postiter();
+        timer_off("DFJCOSK: INCFOCK Postprocessing");
```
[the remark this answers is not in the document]
davpoolechem: Yeah, this particular label is this long in DirectJK and DFJLinK, as well. Wouldn't be hard to shorten them if you want, however.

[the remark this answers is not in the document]
loriab: As long as they're consistent, it's fine for now, thanks. Something to consider and change all at once if it ever hurts readability.

tests/pytests/test_dfjcosk.py
davpoolechem: What I was trying to do with the test was to make sure that the SCF iterations for the IncFock run was not significantly higher than that of the non-IncFock run. IncFock won't generally reduce the number of SCF iterations significantly, but it CAN greatly increase them. This test is designed to ensure that the IncFock run doesn't require significantly more iterations (defined as more than 3 in this test); and this assert is the point at which the comparison is made. 

Thank's for the clarification on the compare functions!

psi4/src/psi4/libfock/DFJCOSK.cc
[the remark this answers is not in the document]
davpoolechem: Will do! It's a good target to clean up, admittedly.

psi4/src/read_options.cc
```diff
@@ -1592,8 +1592,6 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
                         "ROBUST TREUTLER NONE FLAT P_GAUSSIAN D_GAUSSIAN P_SLATER D_SLATER LOG_GAUSSIAN LOG_SLATER NONE");
         /*- Do reduce numerical COSX errors with overlap fitting? !expert -*/
         options.add_bool("COSX_OVERLAP_FITTING", true);
-        /*- Do allow for improved COSX screening performance by constructing the Fock matrix incrementally? !expert -*/
-        options.add_bool("COSX_INCFOCK", false);
```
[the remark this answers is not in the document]
davpoolechem: I'll double-check this with testing; but I'm pretty sure that, since incfock_ and related variables are not used in any cc file for JK except the integral-direct ones (DFJLinK, DirectJK, DFJCOSK), it should just be ignored.

tests/pytests/test_dfjcosk.py
loriab: Ah, I see now, thanks. do you want an `abs(niter_noinc - niter_inc)` then?

davpoolechem: You're welcome! Abs could be something we do, because it would catch cases where IncFock gives significantly lower SCF iterations than no-IncFock. It may be worth catching this in case it indicates some other error.

psi4/src/read_options.cc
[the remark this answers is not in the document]
davpoolechem: Just checked this! Adding `set incfock true` to the beginning of the scf5 test results in no errors; all tests still run and pass fine! So it does seem that other JK methods simply ignore the INCFOCK keyword.

tests/pytests/test_dfjcosk.py
davpoolechem: All right, I think this file has been largely refined based on the discussion here!

psi4/src/psi4/libfock/jk.h
```diff
@@ -1196,9 +1196,24 @@ class PSI_API DFJCOSK : public JK {
...
+
     /// Previous iteration pseudo-density matrix
     std::vector<SharedMatrix> D_prev_;
+    
+    // D_ref_, the effective pseudo-density matrix is either:
+    //   (1) the regular density: D_eff == D_lr = C_lo x C*ro
+    //   (2) the difference density: D_eff == dD_lr = (C_lo x C_ro)_{iter} - (C_lo x C_ro)_{iter - 1}
+    std::vector<SharedMatrix> D_ref_;
 
+    // Is the JK currently on a guess iteration
```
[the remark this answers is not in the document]
davpoolechem: I dug into this a little bit. For DF_SCF_GUESS=True runs, it shows as true on the first CONV direct SCF iteration and false otherwise. For runs using a BASIS_GUESS, it shows up as true for the first iteration of each SCF cycle present (i.e., once for both the smaller and the larger basis set SCFs) and false otherwise. This is all from my testing. 

Further looking into it, this is a feature of the old DirectJK/DFJLinK IncFock scheme that does two things. First, that disables incremental Fock construction on the first SCF iteration of a given cycle. Second, it is set to true when the size of the previous D matrix vector is not the same length as the current D matrix vector. The second component is described in a comment in the original IncFock scheme in DirectJK as being "used to handle stability analysis case". I couldn't tell you how or why that particular component works, but I carried it over to make sure nothing broke.

Regardless, you're right; the comment could use a better description. I'm thinking something like `// Is the JK currently on the first SCF iteration of this SCF cycle?`

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -310,55 +313,85 @@ void DFJCOSK::print_header() const {
...
+   	timer_off("DFJ");
     }
     
+    // COSX
     if (do_K_) {
         timer_on("COSK");
-        build_K(D_eff, K_ao_);
-        timer_off("COSK");
+   	    build_K(D_ref_, K_ao_);
+   	    timer_off("COSK");
```
loriab, line 383: I wonder if your editor isn't counting spaces right or not converting tabs to 4 spaces. No harm or need to edit further, as we'll catch it in a clang-format someday.

psi4/src/psi4/libfock/jk.h
```diff
@@ -755,7 +755,7 @@ class PSI_API DirectJK : public JK {
     /// Pseudo-density matrix to be used this iteration
     std::vector<SharedMatrix> D_ref_;
 
-    // Is the JK currently on a guess iteration
+    // Is the JK currently on the first SCF iteration of this SCF cycle?
```
loriab, line 758: Thanks for investigating. I think this is clearer.

loriab (approved):
lgtm, thanks!

susilehtola:
This reminds me, there *is* a technique to avoid recomputing every `N` iterations. Namely, I think Ahlrichs(?) described a procedure where one keeps track of the change of the density matrix from the reference one by setting `D_0=0` and `D_i = D_{i-1} + max{|Duv|}` where `Duv` is the difference density matrix at iteration `i` and `D_i` tracks the changes. Instead of doing the reset every `N` iterations, one does a full rebuild when `D_i >= eps` because by then sufficient numerical error may have creeped into the calculation.

davpoolechem:
> This reminds me, there _is_ a technique to avoid recomputing every `N` iterations. Namely, I think Ahlrichs(?) described a procedure where one keeps track of the change of the density matrix from the reference one by setting `D_0=0` and `D_i = D_{i-1} + max{|Duv|}` where `Duv` is the difference density matrix at iteration `i` and `D_i` tracks the changes. Instead of doing the reset every `N` iterations, one does a full rebuild when `D_i >= eps` because by then sufficient numerical error may have creeped into the calculation.

Thank you for telling me about this! This is an intriguing idea overall. Essentially, instead of hard-setting a number of iterations to fully recompute the Fock matrix, one does it dynamically based on how much the density matrix has changed over time.

While I don't think this PR is the right place to implement such a scheme (with 1.7 is right around the corner; and since such a scheme would break standardization of IncFock between the different integral-direct JK builds Psi4 has currently, assuming we implement it only in DFJCOSK for this PR), it could be worth investigating in the future. The potential for removing a human-set parameter (INCFOCK_FULL_FOCK_EVERY in this case) is always appealing.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -310,55 +313,85 @@ void DFJCOSK::print_header() const {
...
+    // range-separated semi-numerical exchange needs https://github.com/psi4/psi4/pull/2473
+    if (do_wK_) throw PSIEXCEPTION("COSK does not support wK integrals yet!");
+
+    // explicit setup of Incfock for this SCF iteration
+    if (incfock_) {
+        timer_on("DFJCOSK: INCFOCK Preprocessing");
 
+        int reset = options_.get_int("INCFOCK_FULL_FOCK_EVERY");
+        double incfock_conv = options_.get_double("INCFOCK_CONVERGENCE");
+        double Dnorm = Process::environment.globals["SCF D NORM"];
```
JonathonMisiewicz, line 356:
```suggestion
        auto reset = options_.get_int("INCFOCK_FULL_FOCK_EVERY");
        auto incfock_conv = options_.get_double("INCFOCK_CONVERGENCE");
        auto Dnorm = Process::environment.globals["SCF D NORM"];
```

JonathonMisiewicz (approved):
Some nitpicky changes, but LGTM otherwise.

Three cheers for standardization!

davpoolechem:
All right, I just did a final rebase. Given no issues with CI, this should be ready to merge!

PR #2818, opened 2022-12-02
`cc` Lagrangian and OPDM to Matrix; UHF Bugfix

JonathonMisiewicz:
## Description
`cc` now uses `Matrix` rather than `block_matrix` for the OPDM and Lagrangian. During the course of this transition, I discovered an indexing bug in the conversion of the UHF densities to Matrix. This dates [back to 2016](https://github.com/psi4/psi4/commit/d4bbb98291afe4b500476d0d46c87ded3821bed6). As-is seen from the test values that had to change, this can lead to dramatic errors in properties.

- [x] Fixed a bug where UHF-CCSD properties were incorrect when some orbitals were singly occupied

## Dev notes & details

- [x] `cc` now stores its Lagrangian and OPDM as `Matrix` objects rather than block_matrices rather than write them to disk
- [x] `cc` now uses the wfn based interface, rather than the disk interface, to gradients 
- [x] Adds a finite difference correctness test to  UHF-CCSD one-electron properties
- [x] Much heavier use of slice technology 

## Questions
- [x] Lori, does the test look okay to you?
- [x] Should I do another check on the dipole that changed the most in our test? That change is _massive_.

## Checklist
- [x] `cc` tests pass
- [x] Edited test with incorrect reference values 

## Status
- [x] Ready for review
- [ ] Ready for merge


psi4/src/psi4/cc/ccdensity/sortI_ROHF.cc
```diff
@@ -53,145 +54,84 @@ namespace ccdensity {
...
+    temp2 = temp.get_block(aocc_in_occ, avir_in_vir);
+    O->set_block(aocc_in_full, avir_in_full, temp2);
+    temp2 = temp2->transpose();
+    O->set_block(avir_in_full, aocc_in_full, temp2);
     global_dpd_->file2_close(&D);
 
     /* Sort beta components */
+    // vir is stored in DPD as UOCC then SOCC.
+    // The standard order used by Psi is SOCC then UOCC.
+    // This inconsistency is why we need to treat uocc and socc separately.
```
loriab, line 90: valuable explanation, thanks

tests/pytests/test_dipoles.py
```diff
@@ -45,3 +45,31 @@ def test_dipole(inp):
     analytic_dipole = wfn.variable(inp['varname'] + " DIPOLE")
 
     assert compare_values(findif_dipole, analytic_dipole, 5, "findif vs. analytic dipole")
+
+@ pytest.mark.slow
+@pytest.mark.parametrize("inp", [
+    pytest.param({'name': 'ccsd', 'options': {'opdm_relax': 'true'}, 'varname': 'CCSD'}, id='ccsd')
```
loriab, line 51: good test. why not merge it with `test_dipole` since all but the reference repeated?

ah, molecule. can do `mol = h2o_doublet if "reference" not in inp["options"] else h2o_singlet`

tests/cc56/input.dat
```diff
@@ -39,9 +39,9 @@ compare_values(ref0A2, core.variable("CCSD ROOT 0 (IN A2) TOTAL ENERGY"), 6, "Ro
 ### Dipole check
 
 test_data = {                   # TEST
-(0, "B1", 0, "A1"):  1.0109909, # TEST
-(1, "B1", 1, "A1"):  0.4953101, # TEST
-(0, "A2", 2, "B2"): -0.0535458, # TEST
+(0, "B1", 0, "A1"):  0.9924065, # TEST
+(1, "B1", 1, "A1"): -0.1336723, # TEST
```
loriab, line 43: this second one is a jump -- check through cfour sometime?

loriab (approved):
wow, congratulations on the indexing slog. I can see the value in slices :-)

psi4/src/psi4/cc/ccdensity/sortI_RHF.cc
```diff
@@ -60,96 +61,36 @@ namespace ccdensity {
...
     global_dpd_->file2_init(&D, PSIF_CC_OEI, 0, 1, 1, "I'AB");
-    global_dpd_->file2_mat_init(&D);
-    global_dpd_->file2_mat_rd(&D);
-    for (h = 0; h < nirreps; h++) {
-        for (a = 0; a < virtpi[h]; a++) {
-            A = qt_vir[vir_off[h] + a];
-            for (b = 0; b < virtpi[h]; b++) {
-                B = qt_vir[vir_off[h] + b];
-
-                O[A][B] += 2.0 * D.matrix[h][a][b];
```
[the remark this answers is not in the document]
JonathonMisiewicz: The normalization convention for the Lagrangian that `cc` wrote to **disk** was too large (relative to the **wfn** convention) by a factor of 2. The 2.0 scale on those lines is accounted for by changing the normalization convention from disk to wfn.

tests/cc56/input.dat
JonathonMisiewicz: Sounds good. I don't think Emory has CFOUR, but if we do, I'll be happy to run this before the PR is merged on. I'll check on Monday.

tests/pytests/test_dipoles.py
JonathonMisiewicz: So, define both `h2o_doublet` and `h2o_singlet` inside the test?

psi4/src/psi4/cc/ccdensity/sortI_RHF.cc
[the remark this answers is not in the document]
loriab: aha, another factor of 2 that's a fine convention until it collides with another convention :-)

tests/pytests/test_dipoles.py
loriab: yes, and then pass `properties(..., molecule=mol)`

loriab: actually, `mol = h2o_singlet if inp["options"].get("reference", "rhf") == "rhf" else h2o_doublet` would be more robust

maxscheurer (approved):

lothian (approved):
It would be good to talk about the reasons for the orbital ordering in ROHF cases (perhaps at PsiCon), but I think the PR looks good.  We have CFOUR access here if that would help with output comparisons.

tests/cc56/input.dat
loriab: I can make cfour generate all three new transition dipoles, so I think this is good to go whenever Jonathon wants to merge.

Orbital ordering whyfors at psicon sounds interesting :-)

JonathonMisiewicz:
I'm happy to talk orbital ordering at PsiCon. One of the points I'll be making in my talk is the importance of keeping this sort of thing documented.

I'll be pushing one more commit momentarily, when build and dipole tests pass again. After that, it'll be good to merge.

PR #2684, opened 2022-08-14
Remove unused diagonalizers and linear solvers from dfoccwave::Tensor2d and dfoccwave::Array2d

TiborGY:
## Description
Much like `occ` (PR #2679), `dfocc` also has these functions that are never called anywhere C++-side, and not `PSI_API`.
This is another shard of the https://github.com/psi4/psi4/pull/2642 mega-PR that can be merged independently.

## Todos
- [x] Unused code is removed from dfoccwave::Array2d
- [x] Unused code is removed from dfoccwave::Tensor2d

## Status
- [x] Ready for review
- [x] Ready for merge

PR #1077, opened 2018-07-05
Port libqt to Windows

raimis:
## Description
This is part of Psi4 porting to Windows (#933).

## Todos
Notable points that this PR has either accomplished or will accomplish.
* **Developer Interest**
  - [x]  Replace `bzero` with `memset`
  - [x] Match function signatures
  - [x] Move a function to avoid a forward-declaration
  - [x] Fix PSI_API usage
  - [x] Add missing header
  - [x] `and` --> `&&`
  - [x] Disable system/user timer
  - [x] Move `<sys/timers.h>` porting to a separate header (after #1078 merged)
 
## Checklist
- [x] ~~Tests added for any new features~~
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2822, opened 2022-12-04
Add deprecation message to schmidt_add in libqt

TiborGY:
## Description

After the removal of `sem_test` in #2776, the only remaining internal user of `libqt/schmidt_add.cc` is `libqt/david.cc`. That too is destined for removal, if its only user in `dfocc` is removed by #2684. So it looks like `schmidt_add` could be removed in the future, when all of its callers are gone.

To keep the promise of not randomly breaking API without fair warning, this PR deprecates the function, setting up its eventual removal. Note that this `schmidt_add` is not the same function as the `schmidt_add` in `cceom` or `detci`, even though the name is the same.

- [x] `PSI_API` function `int schmidt_add(double** A, int rows, int cols, double* v)` is deprecated and 1.7 will be the last release with it present.

## Dev notes & details

- [x] Deprecation message is added to `schmidt_add` in `libqt`

## Questions
- [ ] This function was tagged as `PSI_API` in #1077 by @raimis, do we know why it was necessary back then?

## Checklist
- [x] No new features
- [x] CI tests are passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libqt/blas_intfc.cc
+++ b/psi4/src/psi4/libqt/blas_intfc.cc
@@ -211,7 +211,7 @@ void PSI_API C_DROT(size_t length, double *x, int inc_x, double *y, int inc_y, d
  * \ingroup QT
  */
 
-double PSI_API C_DDOT(size_t length, const double* const x, int inc_x, const double* const y, int inc_y) {
+double PSI_API C_DDOT(size_t length, const double *const x, int inc_x, const double *const y, int inc_y) {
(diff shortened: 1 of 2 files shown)


loriab:
Good call on adding the deprecation warning. Unfortunately, I have no memory of why that `PSI_API` was added; fortunately, this is reversible.

JonathonMisiewicz (approved):

loriab (approved):

PR #2824, opened 2022-12-05
Comment `compute_V`; Parallelize UKS `compute_V`

JonathonMisiewicz:

- [x] Fixed a bug where UKS was not properly parallelized

## Dev notes & details

- [x] Adds a lot of comments to `compute_V` in preparation for fixing `compute_Vx` for GGAs
- [x] Fixes _a_ bug in `compute_Vx` for GGAs. There's more to find, but that's beyond-the-scope 
- [x] Things to review: Are the comments comprehensible? Is anything obvious missing, needed to understand `compute_V` up to GGA in either polarization?
- [x] Things to review: I added [this parallelization line](https://github.com/JonathonMisiewicz/psi4/blob/75709af51d7fd68f926d2e28ac2d3ae9a4cea9c9/psi4/src/psi4/libfock/v.cc#L2490), which is the counterpart of [this line](https://github.com/JonathonMisiewicz/psi4/blob/75709af51d7fd68f926d2e28ac2d3ae9a4cea9c9/psi4/src/psi4/libfock/v.cc#L1978) in RKS. Anything obviously wrong with the parallelization?
- [x] Notice to @davpoolechem: per the above, any UKS speed benchmarks need to be redone with this PR. 
- [x] 1.7 preferred but not demanded

## Checklist
- [x] quick tests passed

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/v.cc
+++ b/psi4/src/psi4/libfock/v.cc
@@ -1254,12 +1254,14 @@ void RV::initialize() {
 void RV::finalize() { VBase::finalize(); }
 void RV::print_header() const { VBase::print_header(); }
 void RV::compute_V(std::vector<SharedMatrix> ret) {
+    // => Validate object <=
     timer_on("RV: Form V");
     
     if ((D_AO_.size() != 1) || (ret.size() != 1)) {
         throw PSIEXCEPTION("V: RKS should have only one D/V Matrix");
     }
     
+    // => Special BrianQC Logic <=
 #ifdef USING_BrianQC
     if (brianEnable and brianEnableDFT) {
         double DFTEnergy;
@@ -1288,6 +1290,7 @@ void RV::compute_V(std::vector<SharedMatrix> ret) {
     }
 #endif
 
+    // => Initialize variables, esp. pointers and matrices <=
     // Thread info
     int rank = 0;
 
@@ -1320,10 +1323,11 @@ void RV::compute_V(std::vector<SharedMatrix> ret) {
 
 // VV10 kernel data if requested
 
+    // => Compute V <=
 // Traverse the blocks of points
 #pragma omp parallel for private(rank) schedule(guided) num_threads(num_threads_)
     for (size_t Q = 0; Q < grid_->blocks().size(); Q++) {
-// Get thread info
+        // ==> Define block/thread-specific variables <==
 #ifdef _OPENMP
         rank = omp_get_thread_num();
 #endif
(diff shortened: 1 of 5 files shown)


psi4/src/psi4/libfock/v.cc
```diff
@@ -2886,9 +2907,9 @@ void UV::compute_Vx(std::vector<SharedMatrix> Dx, std::vector<SharedMatrix> ret)
             rho_bk_y = R_rho_bk_y[rank]->pointer();
             rho_bk_z = R_rho_bk_z[rank]->pointer();
             gamma_bbk = R_gamma_bk[rank]->pointer();
-            rho_bx = pworker->point_value("RHO_AX")->pointer();
-            rho_by = pworker->point_value("RHO_AY")->pointer();
-            rho_bz = pworker->point_value("RHO_AZ")->pointer();
+            rho_bx = pworker->point_value("RHO_BX")->pointer();
+            rho_by = pworker->point_value("RHO_BY")->pointer();
+            rho_bz = pworker->point_value("RHO_BZ")->pointer();
```
susilehtola, line 2912: uhm this looks like a pretty important bug fix...?

JonathonMisiewicz: Thanks, I forgot about this one and will edit the main post accordingly.

`UV::compute_Vx` wasn't correct for GGAs. The two parts of Psi that might use this function, TDDFT and stability analysis, both refuse to run for UKS GGAs, which is why Psi hasn't been producing wrong results. The lines in question fix an obvious bug in that function, but the function _still_ isn't correct. When I enable UTDDFT with GGAs, the old behavior was positive semidefinite violations. The new behavior is convergence failure. This suggests another bug.

Once this PR is in, I'll do a similar commenting of `compute_Vx` up to GGAs. The process of doing that will presumably uncover the bug, and I can re-enable TDDFT and stability analysis for unrestricted GGAs.

There are no other lines in this PR that will change numerical values.

davpoolechem:
Thank you for letting me know!

Regarding parallelization, nothing seems especially amiss. I can see `schedule(guided)` being preferred over `schedule(dynamic)` here.

The comments look largely good to me, as well!

psi4/src/psi4/libfock/v.h
```diff
@@ -170,8 +172,10 @@ class RV : public VBase {
     void initialize() override;
     void finalize() override;
 
+    // compute_V assuming same orbitals for different spin. Computes V_alpha, not spin-summed V.
     void compute_V(std::vector<SharedMatrix> ret) override;
-    void compute_Vx(std::vector<SharedMatrix> Dx, std::vector<SharedMatrix> ret) override;
+    /// compute_Vx assuming same orbitals for different spin. Vectors are length one.
```
hokru, line 177: I think it can take multiple `Dx` as long as `Dx` and `ret` are the same size. Or i misunderstand what you mean with "vectors are length one".

Similar for the UV class.

psi4/src/psi4/libfock/v.cc
hokru: oh wow!

hokru (approved):
lgtm

psi4/src/psi4/libfock/v.h
JonathonMisiewicz: Edited comment, thanks. The crux is that these arguments come in (alpha, beta) pairs for UKS, but not for RKS.

loriab (approved):
code documentation looks much improved to me

JonathonMisiewicz:
Thanks, Holger and Lori! Let's merge this in for 1.7. Users might appreciate more parallelizable DFT.

PR #2833, opened 2022-12-12
CompositeJK Part 2: Pilot Implementation

davpoolechem:
## Description
This is it. It's here.

This PR is the initial implementation of the CompositeJK framework, the culmination of many, MANY of the JK-related PRs that are either in progress, or have been added to Psi4 already.

Many methods have been added to Psi4's JK class which enable utilization of algorithms that build J or K separately to improve performance. The ones in Psi4 as of now are the JK subclasses DFJLinK and DFJCOSK, which, between the two, contain integral-direct density-fitted J construction, the Linear Exchange method (LinK), and the Chain-of-Spheres Exchange method (COSX). Further ones, such as the Continuous Fast Multipole Method (CFMM), are planned for the future, and even more (e.g., J-Engine, Local DF methods) exist beyond the work done in Psi4 for the past year. The problem is, Psi4's JK class does not have a framework that truly supports the existence of these separate J and K algorithms. DFJLinK and DFJCOSK work for now, given the low number of separate J and K algorithms current in Psi4. But, as more separate J and K algorithms are added to Psi4, the number of JK subclasses will skyrocket, assuming every combination of separate J and separate K algorithm is stored in its own JK subclass. This will lead to a nasty problem with code duplication, not even discussing the pollution of the JK hierarchy that would result. Thus, for continued research into separate J and K construction algorithms, a proper framework for supporting such algorithms is paramount.

CompositeJK is that framework. CompositeJK is a single JK subclass which enables the combination and execution of any separate J and K build algorithm available in Psi4. CompositeJK can enable this without duplication of separate build algorithms across different subclasses. Additionally, CompositeJK is designed to enable simple expansion of itself, enabling the easy addition of new separate J and K builds to Psi4 with minimal programming issues.

The current PR is the first implementation of the CompositeJK framework into Psi4, and it has two components:
- Front end. The front end changes control of how composite combinations of separate J and K build algorithms can be called by the user. The most important aspect is that, for CompositeJK, any combinations of its sub-algorithms can be called by setting SCF_TYPE to `J_alg+K_alg`, where _J_alg_ is the separate J build algorithm of choice, and _K_alg_ is the separate K algorithm of choice. As of now, the two CompositeJK options available are `DIRECTDFJ+LINK` representing the old DFJLinK class, and `DIRECTDFJ+COSX` representing the old DFJCOSK class. The key here, is that the "+" delimiter in SCF_TYPE serves as a signal to Psi4 that a CompositeJK algorithm is being used. 
- Back end. Basically all of the back-end changes can be classified into two categories - merging DFJLinK and DFJCOSK into the initial CompositeJK implementation, and modifications of functions where necessary to support the above front-end changes. 

- [X] The "COSX" keyword for "SCF_TYPE" has been replaced with "DIRECTDFJ+COSX".
- [X] The "LINK" keyword for "SCF_TYPE" has been replaced with "DIRECTDFJ+LINK".

## Dev notes & details
- [X] Refactors the DFJLinK and DFJCOSK classes into the combined CompositeJK JK subclass. CompositeJK can perform the computations previously performed by both DFJLinK and DFJCOSK within a single JK subclass. These changes are largely localized to the commits f0c4f6c8e31696f6690e101d71bef1863bcf324c - ca66c91976cb5e8c42c8292fcf90c6d28b885fd0, with bfdff974dbfbb9513c85a460b00495598e5f240e also counting. 
- [X] Implements a new scheme for "SCF_TYPE" names associated with CompositeJK algorithm combinations. This scheme uses the "+" delimiter in conjunction with the names of the separate J and K algorithms present in CompositeJK to use a specific combination of separate J and K build algorithms. 
- [X] Changes functionalities in DFJLinK and DFJCOSK, as well of the JK base class itself, as necessary to support the standardized "SCF_TYPE" naming scheme and the general flexibility and ease of expansion that CompositeJK is aiming for.
- [X] Updates relevant tests and documentation.

## To-do
- [x] Update CompositeJK methods with comments on paper references.
- [x] Update CompositeJK::num_computed_shells() function.

## Questions
- [ ] Hopefully this PR isn't too large? While the lines changed looks pretty large, a lot of it comes from the removal of the DFJLinK and DFJCOSK classes and their reimplementation into CompositeJK, as well as the updating of associated reference outputs.

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge


davpoolechem:
To ease the review process, I am going to go through the diff and mark out what should be prioritized by reviewers, and what should not be.

doc/sphinxman/source/scf.rst
```diff
@@ -648,7 +648,9 @@ ERI Algorithms
 The key difficulty in the SCF procedure is treatment of the four-index ERI
 contributions to the Fock Matrix. A number of algorithms are available in
 |PSIfour| for these terms. The algorithm is selected by the |globals__scf_type|
-keyword, which may be one of the following
+keyword. Some such algorithms consist of a single algorithm applied to
```
davpoolechem, line 651: Changes in doc/* files are documentation updates regarding CompositeJK - worth looking at.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -39,6 +39,7 @@
 #include "psi4/liboptions/liboptions.h"
```
davpoolechem, line 39: The diff interpreted DFJCOSK as being moved to CompositeJK.cc. A lot of this is renaming DFJCOSK functions to CompositeJK functions, but I will point things out where they appear.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -144,186 +145,273 @@ Matrix compute_esp_bound(const BasisSet &primary) {
...
+    density_screening_ = options_.get_str("SCREENING") == "DENSITY";
+    set_cutoff(options_.get_double("INTS_TOLERANCE"));
     early_screening_ = true;
 
-    // incremental Fock setup
-    incfock_ = options_.get_bool("INCFOCK");
+    // pre-construct per-thread TwoBodyAOInt objects for computing 3- and 4-index ERIs
+    timer_on("CompositeJK: ERI Computers");
     
-    // => Direct Density-Fitted Coulomb Setup <= //
```
davpoolechem, line 171: This block is just moved to later on in the code.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -144,186 +145,273 @@ Matrix compute_esp_bound(const BasisSet &primary) {
...
-    eri_computers_[0] = std::shared_ptr<TwoBodyAOInt>(rifactory.eri());
+    eri_computers_["3-Center"][0] = std::shared_ptr<TwoBodyAOInt>(rifactory.eri());
+
+    // create each threads' ERI computers
     for(int rank = 1; rank < nthreads_; rank++) {
-        eri_computers_[rank] = std::shared_ptr<TwoBodyAOInt>(eri_computers_.front()->clone());
+        eri_computers_["4-Center"][rank] = std::shared_ptr<TwoBodyAOInt>(eri_computers_["4-Center"].front()->clone());
+        eri_computers_["3-Center"][rank] = std::shared_ptr<TwoBodyAOInt>(eri_computers_["3-Center"].front()->clone());
     }
-    timer_off("ERI Computers");
```
davpoolechem, line 189: This block is just moved to later on in the code.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -144,186 +145,273 @@ Matrix compute_esp_bound(const BasisSet &primary) {
...
-    timer_on("Overlap Metric Solve");
 
-    int nbf = primary_->nbf();
-    std::vector<int> ipiv(nbf);
+    timer_off("CompositeJK: ERI Computers");
 
-    // solve: Q_init_ = S_an @ S_num_init_^{-1}
-    Q_init_ = S_an->clone();
-    C_DGESV(nbf, nbf, S_num_init.pointer()[0], nbf, ipiv.data(), Q_init_->pointer()[0], nbf);
+    // => Set up separate J algorithm <= //
```
davpoolechem, line 211: Here is where the aforementioned moved blocks of code are moved to.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -144,186 +145,273 @@ Matrix compute_esp_bound(const BasisSet &primary) {
...
+    return JK::num_computed_shells();
 }
 
-size_t DFJCOSK::memory_estimate() {
+size_t CompositeJK::memory_estimate() {
     return 0;  // Memory is O(N^2), which psi4 counts as effectively 0
 }
 
-void DFJCOSK::print_header() const {
+void CompositeJK::print_header() const {
```
davpoolechem, line 347: The header printing scheme was changed a little bit for CompositeJK - this is worth looking at.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -335,21 +423,22 @@ void DFJCOSK::incfock_setup() {
...
+void CompositeJK::incfock_postiter() {
     // Save a copy of the density for the next iteration
     D_prev_.clear();
     for(auto const &Di : D_ao_) {
         D_prev_.push_back(Di->clone());
     }
 }
 
-void DFJCOSK::compute_JK() {
+void CompositeJK::compute_JK() {
```
davpoolechem, line 434: Some changes here to accomodate the new front end.

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -358,45 +447,65 @@ void DFJCOSK::compute_JK() {
...
-    
+
     if (initial_iteration_) initial_iteration_ = false;
 }
 
-void DFJCOSK::postiterations() {}
+void CompositeJK::postiterations() {}
 
-void DFJCOSK::build_J(std::vector<std::shared_ptr<Matrix>>& D, std::vector<std::shared_ptr<Matrix>>& J) {
+void CompositeJK::build_DirectDFJ(std::vector<std::shared_ptr<Matrix>>& D, std::vector<std::shared_ptr<Matrix>>& J) {
```
davpoolechem, line 508: The changes here really only fully standardize Direct DF-J between LinK and COSX, because DFJCOSK and DFJLinK had _slight_ differences in implementation before this. Nothing here that's too important to look at unless you want to, since all of the changes here were in DFJLinK already.

psi4/src/psi4/libfock/jk.cc
```diff
@@ -71,10 +71,12 @@ JK::~JK() {}
 std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary,
```
davpoolechem, line 71: Changes in this function are related to the CompositeJK front end - probably worth a look.

psi4/src/psi4/libfock/jk.h
```diff
@@ -1185,127 +1185,28 @@ class PSI_API MemDFJK : public JK {
 };
 
 /**
- * Class DFJCOSK
+ * Class CompositeJK 
  *
- * JK implementation using a direct density-fitted coulomb algorithm
```
davpoolechem, line 1190: This block is just the removal of DFJCOSK.

psi4/src/psi4/libfock/jk.h
```diff
@@ -1185,127 +1185,28 @@ class PSI_API MemDFJK : public JK {
...
+ * J: Direct DF-J
+ * K: COSX, LinK
  *
- * JK implementation using the direct density-fitted coulomb algorithm of Weigend
- * and the Linear Exchange algorithm of Ochsenfeld
- * TODO: Combine DFJLink with DFJCOSX to start CompositeJK
+ * TODO: Implement SplitJK companion framework for truly arbitrary mixing and matching
  */
-class PSI_API DFJLinK : public JK {
+class PSI_API CompositeJK : public JK {
```
davpoolechem, line 1198: The diff interpreted the CompositeJK implementation as a modification of DFJLinK.

The new CompositeJK class is essentially a union of the DFJLinK and DFJCOSK classes, with small changes to support the new front end.

davpoolechem:
All right, I went through the diff and tried to do some clarifications on the changes made. For other clarifications or more questions, please let me know.

davpoolechem:
Also, here's what I get for the CI error:

`Error: The process 'C:\Miniconda\condabin\conda.bat' failed with exit code 1`

This seems unrelated to the changes in this PR?

psi4/src/psi4/libfock/CompositeJK.cc
```diff
@@ -144,186 +145,273 @@ Matrix compute_esp_bound(const BasisSet &primary) {
...
-        if (do_wK_) outfile->Printf("    Omega:              %11.3E\n", omega_);
-        outfile->Printf("    Integrals threads:  %11d\n", nthreads_);
-        outfile->Printf("    Memory [MiB]:       %11ld\n", (memory_ *8L) / (1024L * 1024L));
-        outfile->Printf("    Incremental Fock :  %11s\n", (incfock_ ? "Yes" : "No"));
-        outfile->Printf("    J Screening Type:   %11s\n", screen_type.c_str());
-        outfile->Printf("    J Screening Cutoff: %11.0E\n", cutoff_);
+        outfile->Printf("  ==> CompositeJK: Mix-and-Match J+K Algorithm Combos <==\n\n");
+
+        outfile->Printf("    J tasked:          %11s\n", (do_J_ ? "Yes" : "No"));
+        outfile->Printf("    K tasked:          %11s\n", (do_K_ ? "Yes" : "No"));
```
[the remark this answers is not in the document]
davpoolechem: This is probably a good idea, and wouldn't be hard to implement. I will do that!

psi4/src/psi4/libfock/jk.cc
```diff
@@ -71,10 +71,12 @@ JK::~JK() {}
...
                                  Options& options, std::string jk_type) {
 
+    bool is_composite = jk_type.find("+") != std::string::npos; // does SCF_TYPE contain +?
+
     bool do_density_screen = options.get_str("SCREENING") == "DENSITY";
     bool do_df_scf_guess = options.get_bool("DF_SCF_GUESS");
     
-    bool can_do_density_screen = (jk_type == "DIRECT" || jk_type == "LINK");
+    bool can_do_density_screen = (jk_type == "DIRECT" || jk_type == "DIRECTDFJ+LINK");
```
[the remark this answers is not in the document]
davpoolechem: This is indeed intended behavior! Although it honestly might be better changed.

In the current version of the code, DFJCOSX fails the subsequent check using the can_do_density_screen variable, which happens when can_do_density_screen is set to false. This can be seen by looking at the original code before alteration here, where only the jk_types DIRECT and LINK result in can_do_density_screen=true, and not a jk_type of COSX. Practically, this means that SCF_TYPE=COSX throws an exception when SCREENING is set to DENSITY.  I believe the reason the reason things were programmed this way, is because SCREENING=DENSITY does density screening from within TwoBodyAOInt, while DFJCOSX has its own density screening scheme implemented within itself.

The intentional filter on both J+K here is a continuation of the current behavior, where using the COSX algorithm in conjunction with SCREENING=DENSITY throws an exception.

samples/matrix1/input.dat
```diff
@@ -48,7 +48,7 @@ info = psi4.DSYEV(0, 'V','U', n, A, n, D, W, 4*n)
 A.name = "Eigenvectors"
 
 # Make first element of eigenvector positive if desired
-for i in range(1,n):
+for i in range(0,n):
```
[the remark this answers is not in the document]
davpoolechem: I do remember you mentioning this previously, but thanks again! I did run the docs build as part of modifying the documentation regarding composite methods, but it's entirely possible I did a hand edit of this one, as well. My bad!

psi4/src/psi4/libfock/CompositeJK.cc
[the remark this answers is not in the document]
davpoolechem: Done and done!

davpoolechem:
CI error with the recent updates: 

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100     9  100     9    0     0     46      0 --:--:-- --:--:-- --:--:--    49
tar: Error opening archive: Unrecognized archive format
Error: Process completed with exit code 1.
```

Seems to be the CI acting goofy again.

davpoolechem:
Further CI update - it seems the ddd-deriv test fails with this PR... but only on Windows.

loriab:
I think the ddd-deriv is a fluke -- we were seeing random Windows fails in Dec. After #2843 is merged, you can rebase to fix the ecos and give azure win another chance.

doc/sphinxman/source/scf.rst
```diff
@@ -648,7 +648,9 @@ ERI Algorithms
...
+the consutruction of both the Coulomb and Exchange constructions
+to the Fock Matrix, such as follows:
```
loriab, line 653:
```suggestion
keyword. Most consist of a single algorithm applied to
the construction of both the Coulomb and Exchange parts
of the Fock Matrix:
```

doc/sphinxman/source/scf.rst
```diff
@@ -685,19 +687,45 @@ CD
...
+|PSIfour| also features the capability to use arbitrary combinations 
+of specialized algorithms that construct either the Coulomb
+or the Exchange matrix separately. These algorithms can be mixed and
+matched arbitrarily to optimize performance based on the calculation
+in question. In general, such combinations of algorithms display lower
+scaling then their combined counterparts, and thus perform better 
+with larger systems. Such combinations of algorithms can be utilized 
+by setting the |globals__scf_type| keyword to ``J_alg+K_alg``, 
+where *J_alg* is the name of the separate Coulomb construction algoritm to use, 
+and *K_alg* is the name of the separate Exchange construction algorithm to use.
```
loriab, line 700:
```suggestion
or the Exchange matrix separately. In general, such combinations of algorithms display lower
scaling than their combined counterparts and thus perform better 
for larger systems. Such combinations can be accessed 
by setting |globals__scf_type| to ``J_alg+K_alg``, 
where *J_alg* and *K_alg* are the names of the separate Coulomb
and Exchange construction algorithms, respectively.
```

doc/sphinxman/source/scf.rst
```diff
@@ -769,6 +797,56 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+mixed and matched with each other freely. The algorithms availble for use in a composite fashion are as follows:
+
+Separate J Construction: DIRECTDFJ 
+
+Separate K Construction: COSX, LINK
+
+Two composite algorithms can be combined with each other by setting the |globals__scf_type| keyword to ``J_alg+K_alg``, where *J_alg* is an available
+algorithm for the separate construction of J, and *K_alg* is an available algorithm for the separate construction of K. For example, if one wanted
+to run an integral-direct density-fitted algorithm for Coulomb construction (i.e., DIRECTDFJ) in conjunction with the Linear Exchange algorithm
+for Exchange construction (i.e., LINK), one would use the |globals__scf_type| keyword ``DIRECTDFJ+LINK``.
```
loriab, line 827: Isn't this (sec:compositejk) pretty repetitive from the new part further up?

psi4/driver/procrouting/proc.py
```diff
@@ -1446,7 +1446,8 @@ def scf_wavefunction_factory(name, ref_wfn, reference, **kwargs):
         wfn._disp_functor = _disp_functor
 
     # Set the DF basis sets
-    df_needed = core.get_global_option("SCF_TYPE") in ["DF", "MEM_DF", "DISK_DF", "COSX", "LINK"]
+    df_needed = core.get_global_option("SCF_TYPE") in ["DF", "MEM_DF", "DISK_DF" ]
+    df_needed |= "DIRECTDFJ" in core.get_global_option("SCF_TYPE")
```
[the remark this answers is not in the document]
davpoolechem: Yeah, we ended up with DF-DirJ in the paper. It'd probably be better to match the paper in that regard, but I will double-check and see if there is any reason in the code not to switch.

doc/sphinxman/source/scf.rst
davpoolechem: Hmm... I suppose one could come to that conclusion. There are a couple of details mentioned in `sec:compositejk` that aren't mentioned in the CompositeJK summary (i.e. the "new part further up"), such as the exploitation of different properties in J and K for composite algorithms, the existence of redundant ERI computation in composite algorithms, and composite algorithms performing worse for small systems (the latter is implied in the summary, but not explicitly stated). How important these details are, and whether they could simply be bundled into the summary, is the question.

psi4/driver/procrouting/proc.py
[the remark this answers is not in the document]
davpoolechem: All right, there's no real reason _not_ to switch to DF-DirJ, so I went with that. I only switched the user-facing stuff, though; not internal function names. Would it be good to switch the internal function names, as well?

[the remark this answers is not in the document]
davpoolechem: That's true. We can probably leave it as is, then.

doc/sphinxman/source/scf.rst
loriab: I vote to bundle the redundant ERI, worse for large, diff J, K properties, etc. details into the summary. That is, have one big section or a little intro section and a big section, rather than the two big sections there are now.

loriab (approved):

davpoolechem:
So before this gets merged in officially, I would like to regenerate the reference output files, especially with the DirectDFJ->DFDirJ change.

jturney (approved):

PR #2834, opened 2022-12-13
Fix screening in Vx

JonathonMisiewicz:
@susilehtola 

- [x] The behavior of the `DFT__RHO_V2_CUTOFF` keyword has been slightly modified for UKS.
- [x] The `DFT__RHO_V2_CUTOFF` default has been changed to the LibXC default.

## Dev notes & details

- [x] RKS and closed-shell UKS now screen out the same points.
- [x] Hermiticity of the UKS energy hessian (should) be restored.
- [x] Test references changed to use tighter density thresholds. (Still derived from Psi.) 

## Questions
- [x] Do we want to tighten the `DFT__RHO_V2_CUTOFF` default?

## Checklist
- [x] `tdscf` ctests pass
- [x] `test_tdscf_excitations` pytest passes
- [x] SVWN analytic hessian pytests pass 

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
The test failure is because of a 0.1 cm^-1 change in hessian frequencies due to the keyword change. I either need to adjust the screening value in the test, adjust the reference, or adjust the default screening value. Opinions?

loriab:
I don't think you can adjust the reference b/c it's analytic vs fd. https://github.com/psi4/psi4/blob/f8a5f63abeb916750e00fdad7c3bb55a5dfa481f/tests/dft-freq-analytic/input.dat#L18-L19

maybe adjust the screening in the test for now. have you run stdsuite? there could be more of these.

JonathonMisiewicz:
`svwn_hessian_module` shows 15 failures, now that I check that... Can I just halve the default value? That will make RKS behavior as it was before. UKS will stay change, but I think that's unavoidable.

loriab:
> svwn_hessian_module shows 15 failures, now that I check that... Can I just halve the default value? That will make RKS behavior as it was before. UKS will stay change, but I think that's unavoidable.

Unless you think new (w/PR) analytic value is better than before (in which case I need to re-collect refs), I agree on halving the default. That's equivalent to tightening up occ/dfocc to work under standard options settings. All stdsuite is run under standard conv and options (besides points=5; otherwise nothing would match).

psi4/src/psi4/libfock/v.cc
```diff
@@ -2988,14 +2991,12 @@ void UV::compute_Vx(std::vector<SharedMatrix> Dx, std::vector<SharedMatrix> ret)
                 std::fill(Tap[P], Tap[P] + nlocal, 0.0);
                 std::fill(Tbp[P], Tbp[P] + nlocal, 0.0);
 
-                if (rho_a[P] > v2_rho_cutoff_) {
+                if (rho_a[P] > v2_rho_cutoff_ || rho_b[P] > v2_rho_cutoff_) {
```
susilehtola, line 2994:
```suggestion
                if (rho_a[P] + rho_b[P] > v2_rho_cutoff_) {
```

psi4/src/psi4/libfock/v.cc
```diff
@@ -3031,24 +3032,20 @@ void UV::compute_Vx(std::vector<SharedMatrix> Dx, std::vector<SharedMatrix> ret)
 
                 // This one is a doozy
                 for (int P = 0; P < npoints; P++) {
+                    if ((rho_a[P] < v2_rho_cutoff_) || (rho_b[P] < v2_rho_cutoff_)) continue;
```
susilehtola, line 3035:
```suggestion
                    if (rho_a[P] + rho_b[P] < v2_rho_cutoff_) continue;
```

psi4/src/psi4/libfock/v.cc
```diff
@@ -1235,7 +1235,10 @@ void SAP::compute_V(std::vector<SharedMatrix> ret) {
 }
 
 RV::RV(std::shared_ptr<SuperFunctional> functional, std::shared_ptr<BasisSet> primary, Options& options)
-    : VBase(functional, primary, options) {}
+    : VBase(functional, primary, options) {
+    // For convenience, turn a spindensity cutoff into a spatial density cutoff.
+    v2_rho_cutoff_ *= 2;
```
susilehtola, line 1240:
```suggestion
```

psi4/src/read_options.cc
```diff
@@ -1700,8 +1700,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("DFT_OMEGA_C", 0.0);
         /*- The DFT Correlation hybrid parameter -*/
         options.add_double("DFT_ALPHA_C", 0.0);
-        /*- Minima rho cutoff for the second derivative -*/
-        options.add_double("DFT_V2_RHO_CUTOFF", 1.e-6);
+        /*- Minima spindensity cutoff for the second derivative -*/
+        options.add_double("DFT_V2_RHO_CUTOFF", 5.e-7);
```
JonathonMisiewicz, line 1704:
```suggestion
        /*- Minima spin-summed density cutoff for the second derivative -*/
        options.add_double("DFT_V2_RHO_CUTOFF", 1e-6);
```

JonathonMisiewicz:
Quick tests on a 510 basis function system show that timing savings between old and new defaults are utterly negligible. Changed, per @susilehtola's recommendation.

susilehtola:
Does it have to be its own keyword, could one not just use the same one as for the SCF part?

JonathonMisiewicz:
Good question. Again, it makes sense for me for these two to be the same, but I'm not fond of removing functionality. I'll take other opinions.

JonathonMisiewicz:
We could, perhaps, give it as a _candidate_ for deprecation and deprecate it two releases from now, if nobody objects?

loriab:
Could let `DFT_RHO_V2` default from `DFT_DENSITY_TOL` by checking `has_changed`. That would keep them consistent but keep the extra knob if needed.

JonathonMisiewicz:
Adopted Lori's suggestion. _Hopefully_ the PR will be fine, now.

psi4/src/psi4/libfock/v.cc
```diff
@@ -2988,14 +2992,12 @@ void UV::compute_Vx(std::vector<SharedMatrix> Dx, std::vector<SharedMatrix> ret)
                 std::fill(Tap[P], Tap[P] + nlocal, 0.0);
                 std::fill(Tbp[P], Tbp[P] + nlocal, 0.0);
```
susilehtola, line 2993: Wouldn't a single `std::fill` be better than calling the function consecutively in a loop? If you did that, you could also use a consistent style for the screening i.e. `if (rho_a[P] + rho_b[P] < v2_rho_cutoff_) continue;`

psi4/src/read_options.cc
```diff
@@ -1700,8 +1700,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_double("DFT_OMEGA_C", 0.0);
         /*- The DFT Correlation hybrid parameter -*/
         options.add_double("DFT_ALPHA_C", 0.0);
-        /*- Minima rho cutoff for the second derivative -*/
-        options.add_double("DFT_V2_RHO_CUTOFF", 1.e-6);
+        /*- Minima spin-summed density cutoff for the second derivative. Defaults to the density tolerance. -*/
+        options.add_double("DFT_V2_RHO_CUTOFF", -1.0);
```
susilehtola, line 1704: Maybe `0.0` would be a better default? Also, one should note here that Libxc does internal screening.. did the changing the density tolerance change the Libxc threshold as well?

susilehtola (changes requested):
Other than the comments above, looks good

psi4/src/read_options.cc
JonathonMisiewicz: I don't understand : what changing the density tolerance? Your use of past tense makes me think you're referring to some change in this PR, but no part of this PR matches that description.

psi4/src/psi4/libfock/v.cc
JonathonMisiewicz: Probably.

`Tap` and `Tbp` are pointers to rows of a `Matrix` object, so the question is "are the columns guaranteed to be of length `nlocal`"? If so, then we're zeroing contiguous memory in a loop, and this way of doing it is silly. If not, then we need to check what the contiguous memory is and benchmark adding the overhead of zeroing additional elements versus removing the overhead of a bigger loop.

While this is a good issue to add, it's beyond the PR scope.

psi4/src/read_options.cc
JonathonMisiewicz: > If one changes the density tolerance, does the internal Libxc screening threshold also get changed?

[Yes.](https://github.com/psi4/psi4/blob/33eb8fcd3163f1b77dc3edb63b87cc43d9cf0d5f/psi4/driver/procrouting/dft/superfunctionals.py#L81-L84) That is the one job of the "DFT_DENSITY_TOLERANCE" keyword.

psi4/src/psi4/libfock/v.cc
JonathonMisiewicz: Added #2846.

jturney (approved):

psi4/src/read_options.cc
JonathonMisiewicz: And because I forgot to bring this up - the "DFT_DENSITY_TOLERANCE" defaults to -1, which is why this keyword is defaulting to -1.

loriab (approved):
lgtm!

For future ref, the svwn ref values in qcng standard_suite_ref file were collected mid-2021 in the capabilities overhaul and came solely from psi4. So it's not surprising that tweaking cutoffs needed new ref values in ref_local. These will be moved to qcng soon so that ref_local can be cleared.

Issue #2234, opened 2021-07-03
Broken tests


susilehtola:
While running the tests for #2230, I encountered the following failures
- large-atoms is trying to run analytical ECP gradietns
- min-atoms fails in `KeyError: 'psi4.core.variable: Requested variable CURRENT ENERGY was not set!\n'`
- pywrap-freq-e-sowreap fails in `NameError: name 'get_frequencies' is not defined`
- pywrap-freq-g-sowreap fails in `NameError: name 'get_frequencies' is not defined`

loriab:
How did you come to be running those as none are in the test suite? `large-atoms` is waiting on ECP tech. `min-input` was essentially replaced by the stdsuite. The latter two can't run outside of ctest anyways, and I don't think we want to encourage the sowreap capability.

susilehtola:
Oh I just grepped the ones that use STO-3G... I didn't expect to find broken tests in the tests directory

JonathonMisiewicz:
Are we clear to remove those four tests, then?

loriab:
We should keep `large-atoms`. It's the start of a good test iirc once psi catches up. Agree on ditching the other three. There's probably more dead tests in the suite like the `optking/` folder and these: https://github.com/psi4/psi4/blob/master/tests/CMakeLists.txt#L33-L35. Can do a thorough cleanout when switch to pytest.

loriab:
except for `large-atoms`, #2655  should fix up this issue

PR #2836, opened 2022-12-16
Add large atoms test

philipmnel:
## Description
The `large-atoms` test was previously sitting around unused, waiting for improved ECPs (see discussion in #2234). This PR renovates that test, checking SCF and DF-MP2 energies and gradients on period 5/6 atoms. 

~~TODO (future PR): after MP2 ECP gradients are validated+enabled, should be added to this test~~ 01/25 update: DF-MP2 gradients now enabled and tested

## Dev notes & details

- [x] Add `test_large_atoms` pytest
- [x] Enable DF-MP2 gradients 

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
If we plan to test many methods in this way, shouldn't this be a pytest?

tests/CMakeLists.txt
```diff
@@ -64,7 +64,7 @@ foreach(test_name aediis-1
                   fd-freq-energy fd-freq-energy-large fd-freq-gradient
                   fd-freq-gradient-large fd-gradient freq-isotope1 freq-isotope2 fnocc1 fnocc2
                   fnocc3 fnocc4 fnocc5 fnocc6 fnocc7 frac frac-ip-fitting frac-sym frac-traverse ghosts gibbs
-                  lccd-grad1 lccd-grad2 matrix1 matrix2
+                  large-atoms lccd-grad1 lccd-grad2 matrix1 matrix2
```
loriab, line 67: I think I agree with Jonathon that as a fresh test, this is easier to read as a pytest. Particularly since now libecpint is an optional add-on (ecp capability was built-in when large-atoms was first drafted), this test would need to be moved to the `ENABLE_ecpint` section of this file for the extra logic for cmake to decide when it's runable. shifting to pytest is probably preferred.

tests/large-atoms/input.dat
```diff
@@ -1,124 +1,56 @@
-#! Sample with post-Argon atoms !nosample
+#! Test SCF and DF-MP2 on post-Ar atoms
 
 molecule {
 Ba
-Ba 1 2.0
-units bohr
+Ba 1 4.0
```
loriab, line 5: Might the approximate unit conversion have gone the wrong way?

tests/large-atoms/input.dat
```diff
@@ -1,124 +1,56 @@
...
-  scf_type pk
-}
+ref_energy =  -50.11011922912507 #TEST
 
-thisenergy = energy('mp2')
+analytic_grad, wfn = gradient('scf', return_wfn=True)
+this_energy = wfn.energy()
+findif_grad = gradient('scf', dertype=0)
+assert compare_values(findif_grad, analytic_grad, 6, "Analytic vs. FinDiff Gradient (SCF)") #TEST
+assert compare_values(ref_energy, this_energy, 6, "Calculated vs. Reference Energy (SCF)") #TEST
```
loriab, line 19: ping me if the pytest conversion is entirely foreign. roughly, you'll want to parameterize each of the gradient jobs with a different mol, method, options (basis, etc.), and ref value.

mol is trickiest, and you can use https://github.com/psi4/psi4/blob/master/tests/pytests/test_weird_basis.py#L3-L228 as a model, ignoring all the specialty basis set stuff. 

thanks for working on this!

tests/CMakeLists.txt
philipmnel: Alright, great. I'll switch this over to pytest then.

tests/large-atoms/input.dat
philipmnel: Ah, no. At 2 bohr, the gradient is huge (for Ba-Ba, ~15 Hartree/bohr) which makes getting converged finite differences really tough. I assumed these inputs were boilerplate and made the geometries more sensible (unless having overlapping atoms here was intentional?).

loriab: yes, they're boilerplate, so sure, whatever's sensible. using Bohr can be a little easier to match between programs.

JonathonMisiewicz (approved):

loriab (approved):

PR #2838, opened 2022-12-17
Remove deprecated rsp diagonalizer

TiborGY:
## Description

This PR removes the `rsp` diagonalizer which was deprecated in #2798, and its supporting functions `tqli` and `tred2`, which are not used anywhere else.
These can be preserved in psi4attic after this PR is merged.

- [x] Deprecated `PSI_API` function `void rsp(int nm, int n, int nv, double *array, double *evals, int matz, double **evecs, double toler)` is removed.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):
thanks for the cleanup!

JonathonMisiewicz (approved):

PR #2848, opened 2023-01-11
DFJK Expert I/O Control Part 1: MemDFJK

davpoolechem:
## Description
This PR is the start of a two-part mini-project regarding the two DFJK algorithms present in Psi4 - MemDFJK and DiskDFJK. This first PR handles some changes to MemDFJK, while the next PR in this project will handle the same changes to DiskDFJK.

So, what are the changes? MemDFJK and DiskDFJK hold a rather interesting property. Each of them is actually a combination of two sub-algorithms - one sub-algorithm which is optimized for storing ERIs in memory, and the other sub-algorithm which is optimized for storing ERIs on disk. This, in total, leads to a potential combination of 4 algorithms that can be utilized when one sets SCF_TYPE to DF. Now, one can set SCF_TYPE to either MEM_DF to DISK_DF to use MemDFJK or DIskDFJK, respectively. However, one cannot explicitly use a keyword to control whether MemDFJK or DiskDFJK uses its corresponding in-core or out-of-core sub-algorithm. Which sub-algorithm is used, is entirely dictated by the amount of memory allocated to Psi4 compared to the amount of memory needed to store ERIs.

This PR project seeks to remedy the aforementioned issue by implementing exactly such a keyword for both MemDFJK and DiskDFJK, to control whether they use their in-core or out-of-core sub-algorithms. The new keyword is FORCE_MEM, which can take three options. FORCE_MEM=AUTO (the default) maintains the previous behavior, in which a sub-algorithm is selected based on user-supplied memory. FORCE_MEM=NO_INCORE forces the DF algorithm to use its out-of-core sub-algorithm, even if enough memory is given to run the calculation in-core. Finally, FORCE_MEM=FORCE_INCORE forces the DF algorithm to run its in-core sub-algorithm, and causes the calculation to throw an exception if not enough memory is given to Psi4 to utilize the in-core sub-algorithm.

So why, exactly, do we want this? In truth, this is mostly a debug and expert option. The entire reason I started this PR project is actually because the lack of control over in-core vs. out-of-core DF algorithms was a bit of a thorn in my side in my JK benchmarks. That said, it still works well as a debug option, wherein you can enforce looking at either the in-core or out-of-core algorithm for either MemDFJK or DiskDFJK. Also, FORCE_MEM=FORCE_INCORE has a neat practical application in that you can lock the calculation out of doing on-disk calculations, if one desires to not do so for time's sake.

This first PR implements the FORCE_MEM keyword and applies it to the MemDFJK algorithm. The second half of this PR project will apply the FORCE_MEM keyword to DIskDFJK.

- [X] The "FORCE_MEM" expert option has been added to Psi4, enabling more fine-grained control over the behavior of the SCF_TYPE="DF" algorithms.

## Dev notes & details
- [X] Adds a new keyword "FORCE_MEM" to Psi4. The FORCE_MEM keyword allows the user to specify, for MemDFJK (in this PR) and DiskDFJK (the next PR), whether the algorithm uses its in-core (FORCE_MEM=FORCE_INCORE) or out-of-core (FORCE_MEM=NO_INCORE) sub-algorithm, or whether the sub-algorithm is selected automatically based on allocated memory (FORCE_MEM=AUTO, the default).
- [X] Implements functionality of the FORCE_MEM keyword to the MemDFJK class (i.e., the MEM_DF SCF_TYPE).

## Questions
- [x] Would "FORCE_DF_MEM" be a better name for the new keyword?

## Checklist

## Status
- [X] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -727,9 +727,9 @@ can manually switch between the in-memory and on-disk options for MEM_DF using
 the |scf__force_mem| option. Using ``SCF_SUBTYPE = AUTO``, which allows |PSIfour| 
 to automatically select the in-memory or on-disk option for MEM_DF, is the default 
 and recommend option. However, the in-memory option for MEM_DF can be forced by using
-``SCF_SUBTYPE = FORCE_INCORE``, and the on-disk option for MEM_DF can be forced by using
-``SCF_SUBTYPE = NO_INCORE``. Note that an exception will be thrown if 
-``SCF_SUBTYPE = FORCE_INCORE`` is used with MEM_DF without allocating sufficient memory to 
+``SCF_SUBTYPE = INCORE``, and the on-disk option for MEM_DF can be forced by using
+``SCF_SUBTYPE = OUT_OF_CORE``. Note that an exception will be thrown if 
+``SCF_SUBTYPE = INCORE`` is used with MEM_DF without allocating sufficient memory to 
 |PSIfour|.
 
 For some of these algorithms, Schwarz and/or density sieving can be used to
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -703,7 +703,7 @@ LINK
     Exchange term. See :ref:`sec:scflink` for more information.
 
 In some cases the above algorithms have multiple implementations that return
-the same result, but are optimal under different molecules sizes and hardware/no_
+the same result, but are optimal under different molecules sizes and hardware
 configurations. Psi4 will automatically detect the correct algorithm to run and
 only expert users should manually select the below implementations. The DF
 algorithm has the following two implementations
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -723,12 +723,12 @@ be selected in all cases so that the correct implementation can be selected by
 |PSIfours| internal routines. Expert users can manually switch between MEM_DF and
 DISK_DF; however, they may find documented exceptions during use as several
 post SCF algorithms require a specific implementation. Additionally, expert users 
-can manually switch between the in-memory and on-disk options for MEM_DF using 
-the |scf__scf_subtype| option. Using ``SCF_SUBTYPE = AUTO``, which allows |PSIfour| 
-to automatically select the in-memory or on-disk option for MEM_DF, is the default 
-and recommend option. However, the in-memory option for MEM_DF can be forced by using
-``SCF_SUBTYPE = INCORE``, and the on-disk option for MEM_DF can be forced by using
-``SCF_SUBTYPE = OUT_OF_CORE``. Note that an exception will be thrown if 
+can manually switch between the in-memory and on-disk options *within* MEM_DF using 
+the |scf__scf_subtype| option. Using ``SCF_SUBTYPE = AUTO``, where |PSIfour| 
+automatically selects the in-memory or on-disk option for MEM_DF based on memory and molecule, is the default 
+and recommend option. However, the in-memory or on-disk algorithms for MEM_DF can be forced by using
+``SCF_SUBTYPE = INCORE`` or ``SCF_SUBTYPE = OUT_OF_CORE``, respectively.
+Note that an exception will be thrown if 
 ``SCF_SUBTYPE = INCORE`` is used with MEM_DF without allocating sufficient memory to 
 |PSIfour|.
 
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -724,7 +724,7 @@ be selected in all cases so that the correct implementation can be selected by
 DISK_DF; however, they may find documented exceptions during use as several
 post SCF algorithms require a specific implementation. Additionally, expert users 
 can manually switch between the in-memory and on-disk options for MEM_DF using 
-the |globals__force_mem| option. Using ``FORCE_MEM = AUTO``, which allows |PSIfour| 
+the |scf__force_mem| option. Using ``FORCE_MEM = AUTO``, which allows |PSIfour| 
 to automatically select the in-memory or on-disk option for MEM_DF, is the default 
 and recommend option. However, the in-memory option for MEM_DF can be forced by using
 ``FORCE_MEM = FORCE_INCORE``, and the on-disk option for MEM_DF can be forced by using
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -703,7 +703,7 @@ LINK
     Exchange term. See :ref:`sec:scflink` for more information.
 
 In some cases the above algorithms have multiple implementations that return
-the same result, but are optimal under different molecules sizes and hardware
+the same result, but are optimal under different molecules sizes and hardware/no_
 configurations. Psi4 will automatically detect the correct algorithm to run and
 only expert users should manually select the below implementations. The DF
 algorithm has the following two implementations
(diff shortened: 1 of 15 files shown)


psi4/src/read_options.cc
```diff
@@ -1408,6 +1408,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
             orbitals before switching to the use of exact integrals in
             a |globals__scf_type| ``DIRECT`` calculation -*/
         options.add_bool("DF_SCF_GUESS", true);
+        /*- Forcibly disable or enable in- or out-of-core algorithms for MemDF/DiskDF. For debug purposes. !expert -*/
+        options.add_str("FORCE_MEM", "AUTO", "AUTO FORCE_INCORE NO_INCORE");
```
loriab, line 1412: I guess what troubles me about this PR is that it's adding a new option `FORCE_MEM` that pertains to one SCF_TYPE, but there's a `PK_NO_INCORE` already that does a partial job of the same but for a different single SCF_TYPE.

I wonder if there should be a single `INCORE_TYPE` option with values `AUTO`, `INCORE`, `OUTOFCORE` (or `MEM`/`DISK` or `FORCE_INCORE`/`NO_INCORE` or any combination thereof) and have it work for `PK` (will need more logic b/c mem-only not selectable at present), `MEM_DF`, `DISK_DF` (in future PR), and maybe `DF` (auto-selection of MEM_DF/DISK_DF)? What do folks think?

davpoolechem: Just as a note, this PR applies FORCE_MEM to both SCF_TYPE=MEM_DF and SCF_TYPE=DF (the latter when MemDFJK is automatically selected, of course). The implementation of FORCE_MEM in MemDFJK/DFHelper is independent of the SCF_TYPE used to utilize it.

I do see what you mean about overlapping keywords, though. This PR set is currently planned to be a two-part project, where FORCE_MEM will apply to both the MemDFJK and DiskDFJK algorithms by the end. Based on what you mentioned, it may be worth combining the PK_NO_INCORE keyword with the FORCE_MEM option so that all of MEM_DF, DISK_DF, DF, and PK use it. 

Such a plan should be doable across separate PRs, of which this PR is one. That said, this all does bring up the question of PR ordering. This PR introduces the FORCE_MEM option and implements it into MemDFJK, but maybe it is better to split those two and have the introduction of FORCE_MEM replace PK_NO_INCORE for the initial PR?

loriab: what do you think of the below?
```
/*- For certain |globals__scf_type| algorithms that have internal sub-algorithms
    depending on available memory or other hardware constraints, allow the best
    sub-algorithm for the molecule and conditions (``AUTO`` ; usual mode) or
    forcibly select a sub-algorithm (usually only for debugging or profiling).
    Presently, ``SCF_TYPE=MEM_DF`` can have ``INCORE`` or ``OUT_OF_CORE`` selected.
    (This also applies for ``SCF_TYPE=DF`` when ``MEM_DF`` active.) !expert -*/
options.add_str("SCF_SUBTYPE", "AUTO", "AUTO INCORE OUT_OF_CORE");
```

davpoolechem: All right, after further discussion of the idea at the Psi4 dev meeting, I understand the scope of this suggestion better; and I will implement it!

davpoolechem: Done and done!

psi4/src/psi4/lib3index/dfhelper.h
```diff
@@ -52,6 +52,8 @@ class TwoBodyAOInt;
 class PSI_API DFHelper {
    public:
     DFHelper(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> aux);
+    DFHelper(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> aux,
+        Options& options);
```
loriab, line 56: There's no easy way around adding Options to DFHelper?

davpoolechem: Theoretically, one could simply pass the setting of the "SCF_SUBTYPE" keyword itself through the constructor instead of the entire options object, as "SCF_SUBTYPE" is the only option DFHelper really uses right now. I went with this way, though, because I consider it more elegant (more in line with the JK constructors), and it helps future-proof in the case that options are needed in DFHelper in the future.

davpoolechem: Looking into it some more, it may be possible to access the "SCF_SUBTYPE" variable through Process::environment.globals instead of passing an options object. I'll be happy to switch to this methodology if that is preferred!

davpoolechem: So it seems Process::environment can be used to get "SCF_SUBTYPE", instead of passing an options object! If this is the preferred methodology, I'll be happy to use that way instead.

davpoolechem: After further discussion with Lori, I have reimplemented options access in DFHelper using Process::environment!

psi4/src/psi4/libfock/MemDFJK.cc
```diff
@@ -53,8 +53,9 @@ using namespace psi;
 
 namespace psi {
 
-MemDFJK::MemDFJK(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary)
-    : JK(primary), auxiliary_(auxiliary) {
+MemDFJK::MemDFJK(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary,
+    Options& options) : JK(primary), auxiliary_(auxiliary), options_(options) {
```
loriab, line 57: Options remains in the MemDFJK constructor?

davpoolechem: This actually opens up a potential route of discussion! 

Yes, options remains in the MemDFJK constructor. Of the JK derived classes, all of them contain an options_ member variable, _except_ for GTFockJK, MemDFJK, and DiskDFJK (and CDJK by extension). I originally added options to the MemDFJK constructor to define its own options_ member variable, which was then passed to DFHelper for this PR. However, even with the removal of explicit options-passing to DFHelper, having an options_ member variable for MemDFJK is still reasonable from a standardization perspective. in my opinion.

What are your thoughts on the matter?

loriab: ah, thanks for the clarification! yes, I agree with adding `options_` for standardization. it'll be easier to deal with these classes if they have consistent member data.

davpoolechem: Yes, I would agree! In that case, I will keep this as it is, then.

loriab (approved):
lgtm, thanks for the new knob!

davpoolechem:
> To be clear, `AO_core_` controls which subalgorithm is used? If so, that should be mentioned in a comment.

That's correct! `AO_core_ = true` means the in-core subalgorithm is used, while `AO_core_ = false` means the out-of-core subalgorithm is used. I added comments to clarify.

doc/sphinxman/source/scf.rst
```diff
@@ -722,7 +722,15 @@ algorithm is chosen. It is therefore highly recommended that the keyword "DF"
 be selected in all cases so that the correct implementation can be selected by
 |PSIfours| internal routines. Expert users can manually switch between MEM_DF and
 DISK_DF; however, they may find documented exceptions during use as several
-post SCF algorithms require a specific implementation.
+post SCF algorithms require a specific implementation. Additionally, expert users 
+can manually switch between the in-memory and on-disk options *within* MEM_DF using 
+the |scf__scf_subtype| option. Using ``SCF_SUBTYPE = AUTO``, where |PSIfour| 
+automatically selects the in-memory or on-disk option for MEM_DF based on memory and molecule, is the default 
+and recommend option. However, the in-memory or on-disk algorithms for MEM_DF can be forced by using
```
JonathonMisiewicz, line 729:
```suggestion
and recommended option. However, the in-memory or on-disk algorithms for MEM_DF can be forced by using
```

JonathonMisiewicz (approved):
Please fix verb tense.

PR #2885, opened 2023-02-25
Triplet TDDFT

### LINKED ISSUE #2841: TDDFT Not Triplet Spin-Integrated
One of the [limitations](https://psicode.org/psi4manual/master/tdscf.html#known-limitations) of Psi's TDSCF implementation is that it raises an error upon requesting TDDFT triplets from a restricted reference.

To understand the problem, it's easiest to work with a toy system. I'm choosing the following:
```
molecule {
O
H 1 1.0 
H 1 1.0 2 104.5
}

mtd = "hf/sto-6g"

set tdscf_tda true
set reference uks 
set tdscf_states [0, 0, 2, 0]

energy("td-" + mtd)

set reference rks 
set tdscf_states [0, 0, 1, 0]
energy("td-" + mtd)

set tdscf_triplets only
energy("td-" + mtd)
```

Within the TDA approximation, we need to find the roots of a Hermitian matrix. This matrix has the structure J + V + X. J consists of _Coulomb integrals_, V consists of _the exchange correlation second derivative elements_, and X is everything else - it's not relevant for this problem. Because of the spin-properties of the Hamiltonian (and approximate DFT functionals), the matrix becomes block-diagonal upon transforming the basis states (i-a) α and (i-a) β into their normalized plus and minus combinations. The plus eigenvectors correspond to singlets, and the negative eigenvectors to triplets. Matrix elements take the form 1/2 * [(α|α) ± (α|β) ± (β|α) + (β|β)], where parentheses denote a spin-dependent matrix element. Upon exploiting spin-restriction, this reduces to [(α|α) ± (α|β)].

Now, J matrix elements are spinfree, i.e., the Coulomb integral (pq|pq) doesn't care whether the bra and the ket have the same spin-index or not, so long as p and q have the same spinindex. Therefore, the plus spin combination reduces to 2 * (pq|pq), and the minus spin combination reduces to 0.

The V matrix elements are not spinfree. Whether you differentiate with respect to an alpha excitation twice or an alpha followed by a beta produces different integrals. The formula [(α|α) ± (α|β)] does not simplify.

Unfortunately, Psi combines the V terms into the J terms. Psi currently computes the + case of the V terms with no mechanism to request the - terms. _This_ is why we can't do triplets. This should be fixed, once the code is better organized and commented so it's clear exactly which terms need the freedom to change sign.

JonathonMisiewicz:
## Description
This PR has TDDFT triplets for LDA and GGA functionals. (Psi does not currently support any TDDFT for meta functionals.) This requires lots of moving parts, so this PR serves as a reference for how they all fit together. For ease of reviewing, I'll have smaller PRs that pull off independent pieces for analysis.

While I'm waiting for reviews, I'll update comments to show exactly how I know these spin-integration formulae are correct, for the benefit of future debuggers.

Closes #2841.

## Status

There are four parts that I can split into separate PRs for reviewer convenience. Then I can bring in this PR.

- [x] #2886
- [x] #2887
- [x] #2888
- [x] #2889

- [x] Triplet TDDFT excitations from RKS are now supported.

## Theoretical Analysis

Why were DFT triplets harder than HF triplets? To understand this, we need to understand both the origin of the triplet matrix and the spin properties of the relevant matrix elements.

1. Starting from the UKS RPA/TDA matrices when Ca = Cb, we can do a similarity transformation to obtain the singlet and triplet RKS matrices. The new basis elements for the singlet block all take the form (i->a α + i->a β) / sqrt(2), while the new basis elements for the triplet block take form (i->a α - i->a β) / sqrt(2).
2. The electron potential is spin-free. Because the coulomb J and exchange K terms are expectation values of this, the associated integrals are spin free, assuming spin does not integrate to zero. For J, spin only integrates to zeroes if there is a spin mismatch in either the bra or the ket. Because we only consider Sz preserving excitations, no spin mismatches are possible. For K, a bra orbital needs to have the same spin as a ket orbital. Because both bra orbitals have the same spin, and both ket orbitals have the same spin, this requires that _all_ orbitals have the same spin. Meanwhile, the DFT V terms are second derivatives of the DFT energy with respect to orbital rotation generators. These are not spin-free.
3. Now let's combine the two above facts. After performing the spin-integration in the triplet case, you end up with [(α|α) - (α|β) - (β|α) + (β|β)]. Upon exploiting spin-restriction, this reduces to [(α|α) - (α|β)]. For J, the second term is equal to the first, so the two cancel. For K, the second term is zero, so you have the first integral. For V, the two terms are neither equal nor zero. The V term cannot be neglected, even though it's normally added to the J term, which here is zero.

## Dev notes & details

- [x] V is still bundled with J, but the RSCF products now mark that they may need to get a J-like term. This is no longer equivalent to being singlet or not.
- [x] Several methods have been modified to have a singlet flag, necessary to pass to compure_Vx whether to compute the singlet or triplet term.
- [x] HF classes no longer have a `potential_` attribute. Individual classes may need to access signatures of the specific subclass they have. Instead, subclasses now have a specific subclass for their `potential_` attribute if applicable. An abstract method has been added to the HF base class to get the potential when the subclass doesn't change the method signature.
- [x] RV::compute_Vx_full now exists alongside RV::compute_Vx. The former needs to exist so we can have a flag to control the spin-integration. The latter needs to exist to not break polymorphism when we don't need that flag.
- [x] A new function has been created to make a UKS version of an RKS functional.
- [x] If a triplet is requested, `compute_functional` will build a UKS functional, compute for that, and cannibalize the pieces to get the properly triplet spin-integrated quantity.
- [x] Ability to do `DAXPBY` added.
- [x] Updated a bad test value.

## Checklist
- [x] `test_tdscf_excitations.py` passes. All 70 of the tests.

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfunctional/superfunctional.cc
+++ b/psi4/src/psi4/libfunctional/superfunctional.cc
@@ -662,7 +662,66 @@ void SuperFunctional::allocate() {
     }
 }
 std::map<std::string, SharedVector>& SuperFunctional::compute_functional(
-    const std::map<std::string, SharedVector>& vals, int npoints) {
+    const std::map<std::string, SharedVector>& vals, int npoints, bool singlet) {
+    if (!singlet && is_unpolarized() and deriv_ == 2) {
+        // We want triplet spin integration for the second derivatives.
+
+        // First, make a UKS version of this functional.
+        auto UKS = build_polarized();
+
+        // Now compute the UKS version of this functional. Naturally, the input values will need to be UKS-ified first.
+        std::map<std::string, SharedVector> UKS_vals;
+        if (true) {
+            UKS_vals["RHO_A"] = std::make_shared<Vector>(std::move((vals.at("RHO_A")->clone())));
+            UKS_vals["RHO_A"]->scale(0.5); // Un-spinsum
+            UKS_vals["RHO_B"] = UKS_vals["RHO_A"];
+        }
+        if (vals.count("RHO_AX")) {
+            UKS_vals["RHO_AX"] = std::make_shared<Vector>(std::move(vals.at("RHO_AX")->clone()));
+            UKS_vals["RHO_AX"]->scale(0.5); // Un-spinsum
+            UKS_vals["RHO_BX"] = UKS_vals["RHO_AX"];
+            UKS_vals["RHO_AY"] = std::make_shared<Vector>(std::move(vals.at("RHO_AY")->clone()));
+            UKS_vals["RHO_AY"]->scale(0.5); // Un-spinsum
+            UKS_vals["RHO_BY"] = UKS_vals["RHO_AY"];
+            UKS_vals["RHO_AZ"] = std::make_shared<Vector>(std::move(vals.at("RHO_AZ")->clone()));
+            UKS_vals["RHO_AZ"]->scale(0.5); // Un-spinsum
+            UKS_vals["RHO_BZ"] = UKS_vals["RHO_AZ"];
+            UKS_vals["GAMMA_AA"] = std::make_shared<Vector>(std::move(vals.at("GAMMA_AA")->clone()));
+            UKS_vals["GAMMA_AA"]->scale(0.25); // Un-spinsum
+            UKS_vals["GAMMA_AB"] = UKS_vals["GAMMA_AA"];
+            UKS_vals["GAMMA_BB"] = UKS_vals["GAMMA_AA"];
+        }
+
+        auto temp = UKS->compute_functional(UKS_vals, npoints);
+        values_ = std::move(UKS->compute_functional(UKS_vals, npoints));
+
+        // Now we take the magic triplet combinations.
+        if (true) {
+            values_.erase("V_RHO_B_RHO_B");
+            values_.at("V_RHO_A_RHO_A")->axpby(-0.5, 0.5, *values_.at("V_RHO_A_RHO_B"));
+            values_.erase("V_RHO_A_RHO_B");
+        }
+        if (vals.count("RHO_AX")) {
+            values_.erase("V_GAMMA_BB");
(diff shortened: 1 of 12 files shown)


JonathonMisiewicz:
Ready for review! I'll want this open for a bit before merging it, to give time for comments.

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -662,7 +662,66 @@ void SuperFunctional::allocate() {
...
+    const std::map<std::string, SharedVector>& vals, int npoints, bool singlet) {
+    if (!singlet && is_unpolarized() and deriv_ == 2) {
+        // We want triplet spin integration for the second derivatives.
+
+        // First, make a UKS version of this functional.
+        auto UKS = build_polarized();
+
+        // Now compute the UKS version of this functional. Naturally, the input values will need to be UKS-ified first.
+        std::map<std::string, SharedVector> UKS_vals;
+        if (true) {
```
jturney, line 674: Is there a reason for couple of if(true)s that I see?

JonathonMisiewicz: This convention traces back to [the LibXC interface](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfunctional/LibXCfunctional.cc#L407) but is fundamentally so the LDA block of code "looks like" the GGA and eventually meta blocks. I can also check that "RHO_A" is in the dictionary, but that should never ever be false, so checking it as if it _could_ be false would be misleading.

jturney: Oh okay, how odd. That's fine. I didn't know the history of it.

jturney (approved):

loriab (approved):

PR #2891, opened 2023-03-03
Parallelize compute_esp_over_grid

q-posev:
## Description

There was a race condition in the parallel calculation of ESP over grid in memory. This was fixed in PR #1900.
This PR brings back the OpenMP parallelization, the race condition is avoided by allowing each thread to hold its own copy of the `ElectrostaticInt` object. Thanks to @JonathonMisiewicz for the initial hint.

- [ ] Significant acceleration of the calculation of ESP over grid in memory

## Dev notes & details
- [ ] Created a vector of thread-specific `<ElectrostaticInt>` and `<Matrix>` objects for computing the ESP at a given grid point
- [ ] Added `#pragma omp parallel for schedule(dynamic)` to parallelize the outer loop over grid points

## Questions
- [ ] While working on this part of the code I noticed that `cubeprop` uses a scheme different from the one used by `oeprop` to compute ESP on a grid. I am not sure this is documented. Should it be?
- [ ] Initially my goal was to parallelize the computation of `compute_esp_over_grid` function which reads the grid from `grid.dat` file. The way it is designed now, the grid is read iteratively, so one grid point read->one ESP point computed->one ESP point written to the output `grid_esp.dat` file. This is not parallelizible. However, if one would read grid points in batches - this can be parallelized. What do you think? 

## Checklist
- [ ] Tests added for any new features => test already in place

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/oeprop.cc
+++ b/psi4/src/psi4/libmints/oeprop.cc
@@ -26,6 +26,10 @@
  * @END LICENSE
  */
 
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
 #include "psi4/psifiles.h"
 #include "psi4/psi4-dec.h"
 #include "psi4/physconst.h"
@@ -1042,12 +1046,17 @@ void OEProp::compute_esp_over_grid() { epc_.compute_esp_over_grid(true); }
 void ESPPropCalc::compute_esp_over_grid(bool print_output) {
     auto mol = basisset_->molecule();
 
+    // TODO: move this into openMP section ?
     std::shared_ptr<ElectrostaticInt> epot(dynamic_cast<ElectrostaticInt*>(integral_->electrostatic().release()));
 
     if (print_output) {
-        outfile->Printf("\n Electrostatic potential computed on the grid and written to grid_esp.dat\n");
+        outfile->Printf("\n Electrostatic potential to be computed on the grid and written to grid_esp.dat\n");
     }
+        
+    // TODO: remove this
+    outfile->Printf("\n DEBUG PRINT Q-POSEV  \n");
 
+    // TODO: move this into openMP section ?
     SharedMatrix Dtot = wfn_->matrix_subset_helper(Da_so_, Ca_so_, "AO", "D");
     if (same_dens_) {
         Dtot->scale(2.0);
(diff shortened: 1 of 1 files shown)


psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1045,9 +1049,9 @@ void ESPPropCalc::compute_esp_over_grid(bool print_output) {
     std::shared_ptr<ElectrostaticInt> epot(dynamic_cast<ElectrostaticInt*>(integral_->electrostatic().release()));
 
     if (print_output) {
-        outfile->Printf("\n Electrostatic potential computed on the grid and written to grid_esp.dat\n");
+        outfile->Printf("\n Electrostatic potential to be computed on the grid and written to grid_esp.dat\n");
     }
-
+
```
JonathonMisiewicz, line 1054:
```suggestion

```

JonathonMisiewicz (commented):
One minor comment.

loriab (approved):
Thanks for the parallelization!

q-posev:
Thanks for reviewing @loriab @JonathonMisiewicz !

psi4/src/psi4/libmints/oeprop.cc
q-posev: @JonathonMisiewicz I accidentally marked this as resolved, please let me know if additional changes are needed for approval of this PR.

JonathonMisiewicz (approved):

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -1107,22 +1109,47 @@ SharedVector ESPPropCalc::compute_esp_over_grid_in_memory(SharedMatrix input_gri
...
+    std::vector<std::shared_ptr<Matrix>> VtempT;
+    std::vector<std::shared_ptr<ElectrostaticInt>> VintT;
+
+    for (int thread = 0; thread < nthreads; thread++) {
+        VtempT.push_back(std::make_shared<Matrix>("ints", nbf, nbf));
+        VintT.push_back(std::shared_ptr<ElectrostaticInt>(static_cast<ElectrostaticInt*>(integral_->electrostatic().release())));
+    }
+
+#ifdef _OPENMP
+#pragma omp parallel for schedule(dynamic)
```
susilehtola, line 1126: `schedule(guided)` would be better

PR #2887, opened 2023-02-25
Add DAXPBY to Psi and to the Vector class.

JonathonMisiewicz:
## Description
Adds `DAXPBY` to `libqt` so I can add `axpby` to `Vector`. Needed for #2885.

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2823, opened 2022-12-04
Remove long-deprecated LAPACK functions

TiborGY:
## Description

Psi4 has wrappers for seven LAPACK functions which have been marked as deprecated by the LAPACK devs for a long time. These functions have also been marked as deprecated within Psi4. Luckily they are not marked as `PSI_API`, nor are they used by Psi4. This PR removes them.

Note that the default git diff algorithm is getting confused by the repetitive text in `lapack_intfc.cc` and incorrectly indicates that some lines were added. The correct diff can be obtained with `--diff-algorithm=minimal` :
```
$ git diff --stat --diff-algorithm=minimal master remove_deprecated_lapack
 psi4/src/export_blas_lapack.cc            |    7 -
 psi4/src/psi4/libmints/psimath.cc         |   50 --
 psi4/src/psi4/libmints/psimath.h          |   27 -
 psi4/src/psi4/libqt/CMakeLists.txt        |   19 -
 psi4/src/psi4/libqt/lapack_intfc.cc       | 1021 -----------------------------
 psi4/src/psi4/libqt/lapack_intfc_mangle.h |   35 -
 psi4/src/psi4/libqt/qt.h                  |   25 -
 7 files changed, 1184 deletions(-)
```

## User API & Changelog headlines

- [x] Psi4 now requires the LAPACK library to provide `DGGSVD3` and `DGGSVP3`.

## Dev notes & details

- [x] Fortran, C and C++ style wrappers for `DGEGS`, `DGEGV,` `DGELSX,` `DGEQPF,` `DGGSVD,` `DGGSVP` and `DTZRQF` are removed.
- [x] CMake checks for the availability of  `DGGSVD3` and `DGGSVP3` are removed.

## Checklist
- [x] No new features
- [x] CI tests are passing

## Status
- [x] Ready for review
- [x] Ready for merge

PR #2895, opened 2023-03-13
daxpby workaround

loriab:
## Dev notes & details

- [x] Some setups were using a BLAS that doesn't support the extension AXPBY, so #2887 caused runtime can't-find-symbol errors. Most notably, this occurred for Mac users linking to Accelerate. This bypasses the trouble by rerouting to dscal+daxpy
- [x] clarifies who's getting scaled in the export docs.
- Note that the CI error was due to ongoing changes in QCArchive `next` branch. I've pinned the dep back a release so that psi4's interface to it can be fixed later.

## Questions
- [ ] Can someone confirm this works with the Accelerate setup? Below is a quick test.
```
import psi4
import numpy as np


vecX = np.array([1., 1., 1.])
pvecX = psi4.core.Vector.from_array(vecX)
print(pvecX.np)  # 1

vecY = np.array([5., 5., 5.])
pvecY = psi4.core.Vector.from_array(vecY)
print(pvecY.np)  # 5

pvecX2 = pvecX.clone()
pvecX2.axpy(2.0, pvecY)
print(pvecX2.np)  # 11

pvecX2 = pvecX.clone()
pvecX2.axpby(2.0, 1.0, pvecY)
print(pvecX2.np)  # 11

pvecX2 = pvecX.clone()
pvecX2.axpby(2.0, 3.0, pvecY)
print(pvecX2.np)  # 13
```

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libqt/CMakeLists.txt
+++ b/psi4/src/psi4/libqt/CMakeLists.txt
@@ -31,6 +31,10 @@ list(APPEND CMAKE_REQUIRED_LIBRARIES tgt::blas)
 check_function_exists(daxpby _has_daxpby)
 cmake_pop_check_state()
 
+if(NOT _has_daxpby)
+  message(WARNING "${Yellow}Your BLAS library does not seem to be providing the DAXPBY subroutine. If you are seeing this message, and you are not building with Apple Accelerate, you may want to re-check that the correct BLAS/LAPACK libraries are being used by the build system.${ColourReset}")
+endif()
+
 psi4_add_module(lib qt sources)
 target_compile_definitions(qt
   PRIVATE
--- a/psi4/src/psi4/libqt/CMakeLists.txt
+++ b/psi4/src/psi4/libqt/CMakeLists.txt
@@ -21,8 +21,20 @@ list(APPEND sources
   timer.cc
   )
 
+# cmake_symbol_exists() or check_cxx_source_runs() are preferable to check_function_exists()
+#   according to CMake docs. But both require an explicit header, either in the call or in the code,
+#   and that's tough with multiple BLAS/LAPACK backends. So, we'll see if this is robust enough.
+include(CMakePushCheckState)
+include(CheckFunctionExists)
+cmake_push_check_state()
+list(APPEND CMAKE_REQUIRED_LIBRARIES tgt::blas)
+check_function_exists(daxpby _has_daxpby)
+cmake_pop_check_state()
+
 psi4_add_module(lib qt sources)
 target_compile_definitions(qt
   PRIVATE
+    $<$<BOOL:${_has_daxpby}>:BLAS_HAS_DAXPBY>
     FC_SYMBOL=${FC_SYMBOL}
   )
+unset(_has_daxpby)
--- a/psi4/src/psi4/libqt/CMakeLists.txt
+++ b/psi4/src/psi4/libqt/CMakeLists.txt
@@ -27,18 +27,28 @@ list(APPEND sources
 include(CMakePushCheckState)
 include(CheckFunctionExists)
 cmake_push_check_state()
-list(APPEND CMAKE_REQUIRED_LIBRARIES tgt::blas)
+list(APPEND CMAKE_REQUIRED_LIBRARIES tgt::lapack)
 check_function_exists(daxpby _has_daxpby)
-cmake_pop_check_state()
-
 if(NOT _has_daxpby)
-  message(WARNING "${Yellow}Your BLAS library does not seem to be providing the DAXPBY subroutine. If you are seeing this message, and you are not building with Apple Accelerate, you may want to re-check that the correct BLAS/LAPACK libraries are being used by the build system.${ColourReset}")
+  message(WARNING "${Yellow}Your BLAS/LAPACK library does not seem to be providing the DAXPBY subroutine. DAXPBY calls will be re-routed through DSCAL and DAXPY. If you are seeing this message, and you are not building with Apple Accelerate, you may want to re-check that the correct BLAS/LAPACK libraries are being used by the build system. Redo CMake configuration for changes to register.${ColourReset}")
 endif()
+
+check_function_exists(dggsvd3 _has_dggsvd3)
+check_function_exists(dggsvp3 _has_dggsvp3)
+if(NOT (_has_dggsvd3 AND _has_dggsvp3))
+  message(WARNING "${Yellow}Your BLAS/LAPACK library does not seem to be providing the DGGSVD3 and DGGSVP3 subroutines. No re-routing is available. If you are seeing this message, and you are not building with Apple Accelerate, you may want to re-check that the correct BLAS/LAPACK libraries are being used by the build system. Redo CMake configuration for changes to register.${ColourReset}")
+endif()
+cmake_pop_check_state()
 
 psi4_add_module(lib qt sources)
 target_compile_definitions(qt
   PRIVATE
     $<$<BOOL:${_has_daxpby}>:BLAS_HAS_DAXPBY>
+    $<$<BOOL:${_has_dggsvd3}>:BLAS_HAS_DGGSVD3>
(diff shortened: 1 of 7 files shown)


psi4/src/psi4/libqt/CMakeLists.txt
```diff
@@ -42,9 +52,11 @@ endif()
 psi4_add_module(lib qt sources)
 target_compile_definitions(qt
   PRIVATE
+    $<$<BOOL:${_has_daxpby}>:BLAS_HAS_DAXPBY>
```
TiborGY, line 55: I do not think this will work as well as you would hope for. When working on #2823 I have initially tried to write a fancy build-system check that would have errored out early and told the user if they lack DGGSVD3/DGGSVP3, but when I tried it I realized that detection is broken, so I abandoned the entire concept of giving an error message early.

It has been a while since I was working on it, but I _think_ that the problem was that for some reason CMake was always failing to find DGGSVD3/DGGSVP3, even though it was available, so the new non-deprecated functions were not getting used, ever.

I have abandoned the idea of completely debugging CMake detection of LAPACK extensions, but I seem to recall that one of the issues was that `check_fortran_function_exists` does not (and probably cannot) get called if `Fortran_ENABLED == false`. So having these extension availability checks work reliably may come at the cost of always requiring a Fortran compiler, or something like that.

TiborGY (changes requested):

psi4/src/psi4/libqt/CMakeLists.txt
loriab: Seeking this simple check also gave me fits this weekend. I'm pretty sure the PR can detect both found and missing w/o Fortran for at least the Linux+MKL case. MKL has symbols for 3 of the 4 FC_MANGLE variations, and I don't hear much about people having to change it, so I'm hoping that =2 is fairly settled. I'd be content if this check works only for MKL or Accelerate and lets one avoid the Fortran compiler.

I'd much prefer `check_cxx_source_runs`, and I have a snippet, but (1) it's calling `cblas_daxpby` and (2) it has `mkl.h` explicitly which won't do for multiple backends. So I share your reservations, but I'd like to see an Accelerate test to see if it's any use.

TiborGY: I actually would not mind at all if having a Fortran compiler would become an essential build-time dependency. In fact, requiring a modern (F2003 or newer) compiler would actually open the way for completely eradicating having to deal with FC_MANGLE at all. (if someone finds the time to write the required shims)

I would strongly prefer if Psi4 retained at least some compatibility with BLAS/LAPACK implementations beyond MKL/Accelerate, like OpenBLAS which does implement extensions like daxpby.
It is disappointing that Apple's budget does not seem to cover updating their LAPACK library with contemporary extensions.

loriab:
> I would strongly prefer if Psi4 retained at least some compatibility with BLAS/LAPACK implementations beyond MKL/Accelerate, like OpenBLAS which does implement extensions like daxpby.

Oh, yes, I agree that Psi4 should continue to/strive to work with a range of implementations. Right now it's not bad at detection. But non-MKL doesn't get tried out too often. I'll see how OpenBLAS fares again soon when I try c-f psi4. There, the recc is to compile against netlib, then let user choose the backend at runtime. Alternately, one can hard-code MKL, etc. It's good to see that OpenBLAS with OpenMP is now an option (https://github.com/conda-forge/openblas-feedstock/blob/main/recipe/meta.yaml#L20). I think it was pthreads only a few years ago.

My comment on "works only for MKL or Accelerate" was meant for only the daxpby check, so we're not blocking the Accelerate folks.

> It is disappointing that Apple's budget does not seem to cover updating their LAPACK library with contemporary extensions.

Hear, hear! And fix bugs.

psi4/src/psi4/libqt/CMakeLists.txt
```diff
@@ -24,6 +24,16 @@ list(APPEND sources
...
+# cmake_symbol_exists() or check_cxx_source_runs() are preferable to check_function_exists()
+#   according to CMake docs. But both require an explicit header, either in the call or in the code,
+#   and that's tough with multiple BLAS/LAPACK backends. So, we'll see if this is robust enough.
+include(CMakePushCheckState)
+include(CheckFunctionExists)
+cmake_push_check_state()
+list(APPEND CMAKE_REQUIRED_LIBRARIES tgt::blas)
+check_function_exists(daxpby _has_daxpby)
+cmake_pop_check_state()
+
```
TiborGY, line 36: Maybe adding a warning when DAXPBY is not detected would be a good idea? If detection break for whatever reason in the future, this would give people a chance to notice it eventually.
```suggestion
if(NOT _has_daxpby)
  message(WARNING "${Yellow}Your LAPACK library does not seem to be providing the DAXPBY subroutine. If you are seeing this message, and you are not building with Apple Accelerate, you may want to re-check that the correct BLAS/LAPACK library is being used by the build system.${ColourReset}")
endif()

```

loriab: good idea, will do.

> I actually would not mind at all if having a Fortran compiler would become an essential build-time dependency.

This probably isn't going to happen because it's unnecessarily burdensome (e.g., Mac gfortran was/is poorly supported for a while, and Win fortran compilers are half-a-dozen and none does it all and isn't proprietary, at least to my acquaintance). It's easier to get started if req'd deps are minimal (numpy, qcengine, optking, libint, libxc, gau2grid). Also, there's enough devs around who remember the novelty of a C-based QC program that they're not going to want required Fortran.

Nevertheless, I consider Fortran not a required dep but certainly a commonplace one so that addons can be in any language.

TiborGY: > > I actually would not mind at all if having a Fortran compiler would become an essential build-time dependency.
> 
> This probably isn't going to happen because it's unnecessarily burdensome (e.g., Mac gfortran was/is poorly supported for a while, and Win fortran compilers are half-a-dozen and none does it all and isn't proprietary, at least to my acquaintance). It's easier to get started if req'd deps are minimal (numpy, qcengine, optking, libint, libxc, gau2grid). Also, there's enough devs around who remember the novelty of a C-based QC program that they're not going to want required Fortran.

Yeah, I also prefer programs that keep the required deps few and easy, and I am not a huge fan of Fortran either. Still, in this case, Psi4 is paying the price in the C/Fortran interface complexity, and brittleness. The minimal-diff change that could cure that is probably to use CBLAS/LAPACKE, instead of directly calling the Fortran subroutines, and this would not require a Fortran compiler.
But I think we are getting a bit too off-topic for this PR.

JonathonMisiewicz (approved):
Rubber stamp.

jturney (approved):

fevangelista:
@loriab: if the goal is to re-enable compilation on apple Silicon with Accelerate, then we also need to take care of DGGSVD3/DGGSVP3, which are not shipped with the library. These functions are not used or exported to the python side by psi4.

loriab:
> if the goal is to re-enable compilation on apple Silicon with Accelerate, then we also need to take care of DGGSVD3/DGGSVP3, which are not shipped with the library. These functions are not used or exported to the python side by psi4.

Yes, that was the intent -- to reenable Accelerate. Thanks for posting to get this on record.

As far as I recall and observe, the lapack plan for psi has been for libqt to provide a C interface for all lapack, whether used internally or not, so that plugins can depend on psi for all blas/lapack. Python exports are as-needed. The latter function generically (not Matrix) could be handled by numpy nowadays. Whether we want to continue to serve the role of clapack is probably a question for a psi conf call.

I agree, checking for those fns and conditionally excluding them at build-time is what I should do with this PR. What I don't like about it is that it's not going to work with the c-f scheme (https://conda-forge.org/docs/maintainer/knowledge_base.html#how-it-works) where one builds against a standards-compliant generic lapack, then, perhaps, switchs to Accelerate at runtime. I haven't run across a substitution I can define for DGGSVD3/DGGSVP3 -- has anyone else? Apple's recalcitrant.

PR #2899, opened 2023-03-15
New link assignments in ISAPT

konpat:
## Description

A preliminary implementation of new functionalities in intramolecular SAPT proposed by Du Luu and Konrad Patkowski in [https://doi.org/10.1021/acs.jpca.2c06465](https://doi.org/10.1021/acs.jpca.2c06465). The code is coauthored by @dluu12.

## Dev notes & details

- The new available options for `FISAPT_LINK_ASSIGNMENT` are `SAO0, SAO1, SAO2, SIAO0, SIAO1, SIAO2`. The `SIAO1` variant is the recommended one that solves some serious issues of the default `FISAPT_LINK_ASSIGNMENT=C` variant and, after some more testing, will hopefully take over as the production ISAPT scheme.
- A couple minor options for the new variant (orthogonalization of link orbitals, making cube files) have been added and documented in the manual.
- After quite a bit of effort, FISAPT (that is, the F-SAPT partitioning of ISAPT) seems to be working fine.
- A suitable fragment in the manual and two tests (isapt-siao1 and fisapt-siao1) have been added.

## Questions
- The code does what it is supposed to, but certainly needs extensive general cleanup.
- Not tested with external potentials.
- Presently, there are two functions that produce different cube files: `FISAPT::raw_plot()` and `FISAPT::do_cubes()`. They have different capabilities but should probably be merged into one.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


tests/fisapt-siao1/test_input.py
```diff
@@ -0,0 +1,6 @@
+from addons import *
+
+@ctest_labeler("sapt")
+def test_fisapt_siao1():
+    ctest_runner(__file__)
```
loriab, line 5:
```suggestion
from pathlib import Path
import psi4
from addons import *

@ctest_labeler("sapt")
def test_fisapt_siao1():
    fsaptpy_installed = (Path(psi4.core.get_datadir()) / "fsapt" / "fsapt.py").resolve()

    ctest_runner(__file__, [
        fsaptpy_installed,
    ])
```

loriab (commented):
Wow, that's quite the involved method. I found a few comments, but lgtm! Main question is: are unsuitable complications like extern-in-system, charged molecule?, etc. stopped early? Thanks for the contribution!

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -159,6 +161,21 @@ void FISAPT::partition() {
         }
     }
 
+    // => Lookup table for atoms (1-A, 2-B, 3-C) <= //
+    vectors_["FRAG"] = std::make_shared<Vector>("FRAG", nA);
```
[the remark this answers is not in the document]
konpat: `nA` is the total number of atoms in the molecule, not in a fragment:
`int nA = mol->natom();`

konpat:
> Wow, that's quite the involved method. I found a few comments, but lgtm! Main question is: are unsuitable complications like extern-in-system, charged molecule?, etc. stopped early? Thanks for the contribution!

Thanks for the feedback @loriab! I don't think that either a charged molecule or a presence of external potential mean that new ISAPT is a no-go (just the printed dipole moments will be origin-dependent, which is no big deal). We just haven't tested such cases yet. Let me think of a good ionic system to try out.

loriab:
> Thanks for the feedback @loriab! I don't think that either a charged molecule or a presence of external potential mean that new ISAPT is a no-go (just the printed dipole moments will be origin-dependent, which is no big deal). We just haven't tested such cases yet. Let me think of a good ionic system to try out.

I don't have any particular concerns -- I just noticed your paper was on neutrals and the PR intro mentioned "Not tested with external potentials.".

loriab (approved):
Thanks for the great new SAPT flexibility!

psi4/share/psi4/fsapt/fsapt.py
```diff
@@ -836,6 +837,39 @@ def compute_fsapt(dirname, links5050, completeness = 0.85):
...
 
+    # For I-SAPT/SAOn and I-SAPT/SIAOn, we need to add one extra orbital weight for the reassigned link orbital. It belongs
+    # to the atom of A/B directly connected to the linker C. The user needs to supply a file link_siao.dat that has two lines
+    #   A (atomnumber)
+    #   B (atomnumber)
+    # to specify the numbers of atoms which are connected to C. We will now check if this file exists and read it.
+    if os.path.exists("%s/link_siao.dat" % dirname):
+        (fragsiao,keyssiao) = read_fragments("%s/link_siao.dat" % dirname)
+        if ("A" not in keyssiao) or ("B" not in keyssiao):
+            print("\n Invalid syntax of the link_siao.dat file\n")
```
davpoolechem, line 848: Would it be cleaner to raise an exception here, instead of a print out? 

This is more of a formality than anything, since I think the code will throw on one of the next two lines anyway (depending on which of "A" or "B" is missing) if this line is printed.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -308,44 +331,53 @@ void FISAPT::partition() {
...
     std::vector<int> orbsA;
     std::vector<int> orbsB;
     std::vector<int> orbsC;
+    std::vector<int> orbsL;
 
     // => Assign Links <= //
 
     std::string link_assignment = options_.get_str("FISAPT_LINK_ASSIGNMENT");
-    if (!(link_assignment == "AB" || link_assignment == "C")) {
+    if (!(link_assignment == "AB" || link_assignment == "C" || link_assignment == "SAO0" || link_assignment == "SAO1" || link_assignment == "SAO2" || link_assignment == "SIAO0" || link_assignment == "SIAO1" || link_assignment == "SIAO2" )) {
```
davpoolechem, line 348: As a possible thought for cleanup, the entire set of `link_assignment == SAOx/SIAOx` comparisons could probably be collapsed cleanly using regex. Something like:

```
bool is_sao = std::regex_match(link_assignment, std::regex("S[I]?AO[0-2]"));

if (!(link_assignment == "AB" || link_assignment == "C" || is_sao))
 ```
 
Might need an `#include <regex>` in the header list if it is not already included via another header file.

Not necessary by any means, but maybe something worth considering.

File: psi4/src/psi4/fisapt/fisapt.cc [hunk 5]
```diff
@@ -562,7 +628,7 @@ void FISAPT::nuclear() {
 
     matrices_["VC"] = std::make_shared<Matrix>("VC", nm, nm);
     Vint->compute(matrices_["VC"]);
-
+
```
davpoolechem, line 631: Seems there's a bit of extra whitespace to get rid of here.

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -1441,11 +2461,11 @@ void FISAPT::exch() {
     for (int k = 0; k < Exch10_2M_terms.size(); k++) {
         Exch10_2M += Exch10_2M_terms[k];
     }
-    // for (int k = 0; k < Exch10_2M_terms.size(); k++) {
-    //    outfile->Printf("    Exch10(S^2) (%1d)     = %18.12lf [Eh]\n",k+1,Exch10_2M_terms[k]);
-    //}
+    for (int k = 0; k < Exch10_2M_terms.size(); k++) {
```
davpoolechem, line 2464: Just to be clear, the things here were uncommented and left uncommented intentionally?

psi4/src/psi4/fisapt/fisapt.cc
```diff
@@ -1580,12 +2600,457 @@ void FISAPT::exch() {
...
+        auto Dri_ss = linalg::triplet(matrices_["AlloccA"], Dii_ss, matrices_["AlloccA"], false, false, true);
+        Dri_ss->add(linalg::triplet(matrices_["AlloccB"], Dji_ss, matrices_["AlloccA"], false, false, true));
+        auto Dri_os = linalg::triplet(matrices_["AlloccA"], Dii_os, matrices_["AlloccA"], false, false, true);
+        Dri_os->add(linalg::triplet(matrices_["AlloccB"], Dji_os, matrices_["AlloccA"], false, false, true));
+    
+        double Enuc = 0.0;
+        double** Enuc2p = matrices_["E NUC"]->pointer();
+        Enuc += 2.0 * Enuc2p[0][1];  // A - B
+
+//The full first-order energy (elst+exch) is given by Eq. (21)
```
davpoolechem, line 2921: Another small cleanup thing, could be adjusting comments like these so that they are aligned with they code they are commenting on.

davpoolechem:
I decided to take a look through, as well. I focused moreso on cleanup aspects, but I left a few comments here and there.

psi4/src/psi4/fisapt/fisapt.cc
konpat: I opted against shifting all comments far to the right, but I'll be happy to do so if that's the generally preferred style. However, I was able to remove some unnecessary indentation so this should now be less of an issue.

psi4/src/psi4/fisapt/fisapt.cc
konpat: Yup, that was debug printing that is not needed anymore. Good catch!

psi4/src/psi4/fisapt/fisapt.cc
konpat: Fixed.

psi4/src/psi4/fisapt/fisapt.cc
konpat: I know I'm old :) so the current code looks more readable to me, but if the youngsters want this change, I'll be happy to make it...

psi4/share/psi4/fsapt/fsapt.py
konpat: Good idea, fixed.

psi4/src/psi4/fisapt/fisapt.cc
loriab: Someday a formatter will have its way with the comments. Then we can decide if it's worth resisting. :-) Thanks for the indentation change.

Issue #2890, opened 2023-03-02
COSX Grid Issues with 146 Spherical Grid Points + ROBUST Grid Pruning Scheme


davpoolechem:
Hi, all! I mentioned this issue yesterday at the Psi4 Dev meeting, and I wanted to post the issue here, with more collected information.

The crux of the issue, as I have determined, is that the COSX code, when using the ROBUST grid pruning scheme in conjunction with 146 spherical points on either of its grids (initial or final), simply breaks.

Consider the following input:
```
molecule mol {
    0 1
    O
    H 1 0.96
    H 1 0.96 2 104.5
    symmetry c1
    no_reorient
    no_com
}

set {
    scf_type cosx
    df_scf_guess false
    basis aug-cc-pVDZ
    e_convergence 1.0e-10
    incfock true
    incfock_full_fock_every 4
    ints_tolerance 1.0e-12
    cosx_spherical_points_final 146
    cosx_radial_points_final 50
    cosx_pruning_scheme robust
}

energy = energy('scf')
```

A simple expansion of the final COSX grid compared to default. This calculation breaks upon entering the final SCF iteration (where the tight grid is used) with the following error:
```
!----------------------------------------------------------------------------------!
!                                                                                  !
!  ADIIS minimization failed. File a bug, and include your entire input and output !
!     files.                                                                       !
!                                                                                  !
!----------------------------------------------------------------------------------!
```

The following input, an expansion of both the initial and final grids, also fails:
```
molecule mol {
    0 1
    O
    H 1 0.96
    H 1 0.96 2 104.5
    symmetry c1
    no_reorient
    no_com
}

set {
    scf_type cosx
    df_scf_guess false
    basis aug-cc-pVDZ
    e_convergence 1.0e-10
    incfock true
    incfock_full_fock_every 4
    ints_tolerance 1.0e-12
    cosx_spherical_points_initial 146 
    cosx_radial_point_initial 50 
    cosx_spherical_points_final 302
    cosx_radial_points_final 75 
    cosx_pruning_scheme robust
}

energy = energy('scf')
```

This one fails in the SAD-guess iteration, with the following error:
```
!----------------------------------------------------------------------------------!
!                                                                                  !
! Fatal Error: Matrix::diagonalize: DSYEV failed!                                  !
! Error occurred in file:                                                          !
!     /theoryfs2/ds/dpoole34/Documents/Codes/psi4/psi4/src/psi4/libmints/matrix.cc !
!     on line: 1715                                                                !
[...]

PR #2906, opened 2023-03-23
Better Error Message for Issue #2890

davpoolechem:
## Description
This PR provides a more useful error message when the issue described in https://github.com/psi4/psi4/issues/2890 is encountered. In short, the COSX K build method breaks with certain grid configurations - specifically, grid configurations which result in negative grid weights. How to best handle that issue - removing such grid configurations from Psi4 entirely, handling negative grid weights differently within COSX itself - is still under discussion, and this PR is _not_ meant to be a final fix to the aforementioned issue.

Rather, this PR is meant to provide a more informative error message in the case that an end user _does_ encounter this error. Currently, when this error is encountered, it shows up as either a DGESV error (if the initial COSX grid has negative weights) or an ADIIS minimization error (if the final COSX grid has negative weights). These error outputs don't help the end user figure out what is wrong at all. So, what this PR does, is add sanity checks to ensure that the COSX grids don't have negative weights, and throw an exception if they do. The sanity check exceptions explain to the user what the error is and how they can fix it (i.e., changing the grid pruning scheme and/or number of spherical points).

Again, I want to emphasize that this is _not_ intended to be a full fix to https://github.com/psi4/psi4/issues/2890. Rather, it is an intermediate step that is designed to better help end users avoid the problem if they run into it themselves.

N/A

## Dev notes & details
- [X] Add sanity checks to COSX to ensure that neither of the COSX grids have negative grid weights. An exception is thrown if either of the grids do.

## Questions
- [x] Probably better discussed on https://github.com/psi4/psi4/issues/2890, but how do we want to officially solve this problem? There has been some discussion on simply removing grid configurations which lead to negative weights, as well as discussions on reformulating the X matrix (the matrix central to the observed error) to be able to handle positive grid weights. 

## Checklist

## Status
- [X] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/DFJCOSK.cc
+++ b/psi4/src/psi4/libfock/DFJCOSK.cc
@@ -95,7 +95,10 @@ Matrix compute_numeric_overlap(const DFTGrid &grid, const std::shared_ptr<BasisS
                 S_nump[mu][nu] += S_num_blockp[mu_local][nu_local];
             }
         }
-
+        outfile->Printf("S_num_block: \n");
+        outfile->Printf("----------- \n");
+	S_num_block.print_out();
+	outfile->Printf("\n");
     }
 
     S_num.hermitivitize();
--- a/psi4/src/psi4/libfock/DFJCOSK.cc
+++ b/psi4/src/psi4/libfock/DFJCOSK.cc
@@ -247,7 +247,7 @@ void DFJCOSK::common_init() {
         double* w = init_block->w();
         for (int ipoint = 0; ipoint < init_block->npoints(); ++ipoint) {
             if (w[ipoint] < 0.0) {
-	      throw PSIEXCEPTION("The initial COSX grid configuration contains negative weights! This will crash COSX, as per https://github.com/psi4/psi4/issues/2890. A proper fix is incoming, but for now, adjust either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_INITIAL to remove negative weights.");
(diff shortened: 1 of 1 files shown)


loriab (approved):

JonathonMisiewicz:
Please update the error message to say that the grid is unusable with COSX, rather than "better fix" incoming.

JonathonMisiewicz (changes requested):
See above.

loriab:
Maybe language like "The definition of the current grid includes negative weights. As these are not suitable for the COSX implementation, please choose another grid through ...".

davpoolechem:
> Please update the error message to say that the grid is unusable with COSX, rather than "better fix" incoming.

> See above.

> Maybe language like "The definition of the current grid includes negative weights. As these are not suitable for the COSX implementation, please choose another grid through ...".

Done and done! Wording changed to match these suggestions.

Also, for those not at the Psi4 dev meeting today, it was agreed upon that this PR would indeed be the final fix to the COSX grid issue.

JonathonMisiewicz (approved):
I'd mark `w` as `const auto` for good measure, but I'm fine with the PR being merged as-is.

davpoolechem:
> I'd mark `w` as `const auto` for good measure, but I'm fine with the PR being merged as-is.

Yeah, I can do that! Give a quick sec, and we can probably merge after that.

davpoolechem:
Done! We should be good on this front.

loriab:
@mfherbst , there's a couple of the ddx tests failing now https://github.com/psi4/psi4/actions/runs/4703123664/jobs/8341938828?pr=2906#step:23:739, and I suspect it's because of the v0.4.1 released yesterday since it was working fine with v0.3.0 after your #2918 . Does this sound right? If so, we can pin to v0.3 unless you have a PR in the works.

PR #2907, opened 2023-03-23
Rework factorial and n-choose-k functions in libqt

TiborGY:
## Description

This PR reworks and simplifies `libqt/probabil.cc`. Excessive conditionals are removed, sanity checks now throw instead of returning zero, and the `double` kludge is supplanted by `uint64_t`

## Dev notes & details

- [x] `factorial` and `combinations` now return `uint64_t` instead of `double`, and take `const uint64_t` instead of `int`
- [x] `factorial` now throws if the result would overflow
- [x] `combinations` now throws if computing n-choose-k is impossible

## Checklist
- [x] No new features
- [x] CI tests are passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libqt/probabil.cc
+++ b/psi4/src/psi4/libqt/probabil.cc
@@ -32,73 +32,26 @@
   \ingroup QT
 */
 
+#include "qt.h"
+#include "psi4/libpsi4util/exception.h"
 namespace psi {
 
-/*!
-** factorial(): Returns n!
-**
-** Parameters:
-**    \param n  = number to take factorial of
-**
-** Returns:
-**    n factorial, as a double word (since n! can get very large).
-** \ingroup QT
-*/
-double factorial(int n) {
-    if (n == 0 || n == 1) return (1.0);
-    if (n < 0)
-        return (0.0);
-    else {
-        return ((double)n * factorial(n - 1));
-    }
+/// @brief Returns n!
+/// @param n : number to take factorial of
+/// @return n factorial, as a 64-bit int (since n! can get very large)
+/// \ingroup QT
+uint64_t factorial(const uint64_t n) {
+    if (n <= 1) return 1;
+    return (n * factorial(n - 1));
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/libqt/probabil.cc
```diff
@@ -32,73 +32,27 @@
...
-        return ((double)n * factorial(n - 1));
-    }
+/// @brief Returns n!
+/// @param n : number to take factorial of
+/// @return n factorial, as a 64-bit int (since n! can get very large)
+/// \ingroup QT
+uint64_t factorial(const uint64_t n) {
+    if (n <= 1) return 1;
+    if (n > 20) throw PSIEXCEPTION("Cannot compute n! if n > 20, as the result would overflow in a 64-bit integer.");
+    return (n * factorial(n - 1));
```
davpoolechem, line 46: If we're already fixing up this function anyway, would it be worth it to replace the recursive factorial algorithm with an iterative one? One could even tabulate the factorial results in an array and pull from that, since the input is limited to [0, 20) anyway.

Of course, it's not necessary to do this, but I'll ask regardless.

davpoolechem (commented):
Everything you have changed up looks good! I added one possible other thing that could be done, but it's probably not necessary.

psi4/src/psi4/libqt/probabil.cc
TiborGY: All good suggestions, but I think `factorial` is only ever called by `combinations`, which in turn is only ever called once in the entire codebase. So I decided to not spend any more time on them.

davpoolechem: That's reasonable! If it's called that little, then it's probably not worth the extra effort. Maybe if it ever does find more common use within the Psi4 codebase, it will be worth reconsidering a different algorithm then.

davpoolechem (approved):

PR #2908, opened 2023-03-23
fix test_dft_blocking 

loriab:
## Dev notes & details

- [x] one test failed in full tests. I can't check it easily, but I think this is the fix after #2885 
- [ ] Feature2

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz (approved):
This is correct.

I think it would be better practice to pybind over that `singlet` defaults to `True` for `compute_functional`, so none of our users get bitten by this.

susilehtola (approved):

PR #2909, opened 2023-03-24
Add deprecation notices to Matrix::swap_rows and Matrix::swap_columns

TiborGY:
## Description

`Matrix::swap_rows` and `Matrix::swap_columns` do no seem to have any internal users, but since `Matrix` as a whole is `PSI_API`, it would be rude to suddenly remove them. This PR adds deprecation notices.

The motivation for removal is that these two functions are the only internal users of the DSWAP BLAS interface.

- [x] `PSI_API` functions `Matrix::swap_rows` and `Matrix::swap_columns` are now deprecated due to a lack of users. Unless someone speaks up, 1.8 will be the last release to have them.

## Dev notes & details

- [x] Deprecation notices added to `matrix.h`

## Checklist
- [x] No new features
- [ ] CI tests are passing

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/matrix.h
+++ b/psi4/src/psi4/libmints/matrix.h
@@ -1009,8 +1009,14 @@ class PSI_API Matrix : public std::enable_shared_from_this<Matrix> {
     ///
     void sort_cols(const IntVector& sortvec);
     /// Swap rows i and j
+    PSI_DEPRECATED(
+        "Matrix::swap_rows and Matrix::swap_columns are being deprecated due to a lack of users. Unless someone speaks "
+        "up, 1.8 will be the last release to have them.")
     void swap_rows(int h, int i, int j);
     /// Swap cols i and j
+    PSI_DEPRECATED(
+        "Matrix::swap_rows and Matrix::swap_columns are being deprecated due to a lack of users. Unless someone speaks "
+        "up, 1.8 will be the last release to have them.")
     void swap_columns(int h, int i, int j);
 
     /*! Average off-diagonal elements */


loriab:
Is DSWAP deprecated upstream or is there another reason to prune it? I think part of the plan with the libqt exports was to provide a full blas/lapack interface, even if not everything got used by psi internally. That was before `np.linalg` and Eigen were around/prominent, so perhaps it's not so vital, but it's nice for downstream to know if psi is exporting the full set or as-needed.

fwiw, I've no strong opinion on `Matrix::swap_rows` and `Matrix::swap_columns`.

TiborGY:
> Is DSWAP deprecated upstream or is there another reason to prune it? I think part of the plan with the libqt exports was to provide a full blas/lapack interface, even if not everything got used by psi internally. That was before `np.linalg` and Eigen were around/prominent, so perhaps it's not so vital, but it's nice for downstream to know if psi is exporting the full set or as-needed.
> 
> fwiw, I've no strong opinion on `Matrix::swap_rows` and `Matrix::swap_columns`.

There is no upstream deprecation for this one, so comments and discussion is welcome.

I was not aware that the intent was to expose the full set of functions to downstream. Some of the Level 1 and 2 BLAS functions that Psi4 has interfaces for are not exported to Python. So I assumed that Psi4 only provides interfaces for so many of them, because it was much easier to auto-generate all of them, than try to figure out what is needed.

The bigger picture is that by providing the entire Fortran-->C++-->Python interface to downstream, Psi4 is taking on a maintenance burden, with the really problematic part being the Fortran-->C++ part. (eg. the Fortran character argument ABI violation horror story-we are relying on UB and the mercy of compiler developers)
Currently, the sheer number of BLAS/LAPACK functions makes any overhaul of the interface a daunting prospect.

I would wager that if all BLAS/LAPACK functions which have no internal or external users were removed, then the task of transitioning to a less fragile solution for getting linear algebra into Psi4 (regardless of how that will be achieved), will be much easier done.

JonathonMisiewicz:
I don't like the idea of removing these functions. Some of us in Francesco's group use `swap_columns` to move orbitals in the C matrix.

TiborGY:
V

> I don't like the idea of removing these functions. Some of us in Francesco's group use `swap_columns` to move orbitals in the C matrix.

Thanks, I have no intention of causing downstream pain by axing it if it is actually being used.

loriab:
> There is no upstream deprecation for this one, so comments and discussion is welcome.

gtk, thanks

> I was not aware that the intent was to expose the full set of functions to downstream. Some of the Level 1 and 2 BLAS functions that Psi4 has interfaces for are not exported to Python. So I assumed that Psi4 only provides interfaces for so many of them, because it was much easier to auto-generate all of them, than try to figure out what is needed.

Quite right, I think it was for downstream C++ plugins that the full export set was intended. Python exports (relatively newfangled c. 2016) have been as-needed.

> The bigger picture is that by providing the entire Fortran-->C++-->Python interface to downstream, Psi4 is taking on a maintenance burden, with the really problematic part being the Fortran-->C++ part. (eg. the Fortran character argument ABI violation horror story-we are relying on UB and the mercy of compiler developers)
Currently, the sheer number of BLAS/LAPACK functions makes any overhaul of the interface a daunting prospect.

Agree on the daunting. In the same category as the libqt exports are the CMake LAPACK detection and testing with other LAPACK distributions. Right now, I'm hardwiring the psi4 conda-forge recipe to MKL, but it'd be nice in future to relax that to normal usage (compile against a generic interface, then allow runtime selection of distribution) if we could do so with assurance (last time OpenBLAS was tried, we had threading oversubscription for SAPT iirc).

> I would wager that if all BLAS/LAPACK functions which have no internal or external users were removed, then the task of transitioning to a less fragile solution for getting linear algebra into Psi4 (regardless of how that will be achieved), will be much easier done.

PR #2921, opened 2023-04-17
pin pyddx

loriab:
## Description
pyddx is in rapid development, so I think psi4 needs a PR to catch up to v0.4

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -229,7 +229,7 @@ jobs:
           - geometric
           - openfermion>=1.0
           - openfermionpsi4
-          - pyddx
+          - pyddx=0.3.0
           - pymdi
           - qcengine
           - qcfractal


davpoolechem (commented):
Thanks for this PR! DDX issues seem to have been the root for quite a few CI failures lately.

A quick thought on my end, since we're enforcing pyddx 0.3.0 here for now. In the ddx/CMakeLists.txt file in the external/upstream/ directory, there is a line for finding the pyddx Python module:

`find_python_module(pyddx ATLEAST 0.3.0 QUIET)`

Would it be necessary, or even desirable, to temporarily switch this to

`find_python_module(pyddx EXACT 0.3.0 QUIET)`

to simply not allow pyddx v0.4 to be used with Psi4 at the moment?

mfherbst:
I'm a bit surprised this is needed as we did not introduce python-side changes in 0.4. Maybe there is a bug we introduced in 0.4? Can you point me to what is failing?

On top I should add we follow semantic versioning, so all 0.3.x should be compatible. I don't know how conda / cmake        deal with this, but in this case one should target the range 0.3.0 to less than 0.4.0

loriab:
> I'm a bit surprised this is needed as we did not introduce python-side changes in 0.4. Maybe there is a bug we introduced in 0.4? Can you point me to what is failing?

Sure, here's one: https://github.com/psi4/psi4/actions/runs/4735120869/jobs/8404940333#step:23:737

> On top I should add we follow semantic versioning, so all 0.3.x should be compatible. I don't know how conda / cmake deal with this, but in this case one should target the range 0.3.0 to less than 0.4.0

gtk, I can add `0.3.*` if the pin needs to persist.

davpoolechem:
So what's the current status of this PR? pyddx seems to be pinned to 0.3.0 from work in other PRs, although there seems to be discussion of pinning pyddx to 0.3.* instead.

loriab:
I've filed https://github.com/ddsolvation/ddX/issues/142 over at the parent project. Further investigation makes me uncertain whether our test reference may just have a built-in bug, in which case I hate to pin to a worse version.

PR #2924, opened 2023-04-20
DFJK Expert I/O Control Part 2: DiskDFJK

davpoolechem:
## Description

This PR is part two of a mini-project regarding the two DFJK algorithms present in Psi4. The first PR (https://github.com/psi4/psi4/pull/2848) applied to MemDFJK, and the second PR (this one) applies to DiskDFJK. The goal of this PR project is to add an expert keyword to allow user control over which subalgorithm is used by either MemDFJK or DiskDFJK. See, MemDFJK and DiskDFJK, despite their name, each have separate "subalgorithms" optimized to be run either in-core or out-of-core. While MemDFJK and DiskDFJK can be independently selected with `SCF_TYPE`, their subalgorithms cannot be; subalgorithm selection was previously controlled exclusively through memory allocated to Psi4. The mini-project this PR is part of, is meant to rectify that issue.

In the previous PR of this project, the `SCF_SUBTYPE` keyword was introduced, an expert option allowing for forced execution of a given subalgorithm. The default option, `AUTO`, has the code select the subalgorithm automatically, as before. `INCORE` forces MemDFJK to use the in-core subalgorithm, and throws an exception if not enough memory is allocated to Psi4 to do so. `OUT_OF_CORE` forces MemDFJK to use the out-of-core subalgorithm, even if enough memory is allocated to Psi4 to use the in-core subalgorithm.

This current PR takes the `SCF_SUBTYPE` keyword, previously applied to MemDFJK, and applies it to the DiskDFJK algorithm, with the same effects.

## User API & Changelog headlines
- N/A

## Dev notes & details
- [X] Implements functionality of the SCF_SUBTYPE keyword to the DiskDFJK class (i.e., the DISK_DF SCF_TYPE).

## Questions
- N/A

## Checklist
- [X] Tests added for any new features
- [X] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [X] Ready for review
- [x] Ready for merge

PR #2926, opened 2023-04-21
DFJK Expert I/O Control Bonus: PKJK

davpoolechem:
## Description
This PR is a companion PR to https://github.com/psi4/psi4/pull/2848 and https://github.com/psi4/psi4/pull/2924. Those two PRs introduced the `SCF_SUBTYPE` keyword to Psi4, allowing for specifications of subalgorithms in SCF methods where appropriate, and applied the `SCF_SUBTYPE` keyword to MemDFJK and DiskDFJK, respectively.

As it turns out, the DFJK algorithms aren't the only JK algorithms where this new keyword may be useful.

The PKJK algorithm contains a number of subalgorithms, as well - an in-core algorithm (labeled InCore in the code) and two out-of-core algorithms (labeled Reorder and Yoshimine in the code). PKJK has some amount of control over which subalgorithm is utilized already - the `PK_ALGO` keyword specifies whether the Reorder or Yoshimine out-of-core algorithm is to be used, if an out-of-core algorithm is used; while the `PK_NO_INCORE` keyword disables the InCore PK algorithm and forces one of the out-of-core algorithms to be used. If the functionality of `PK_NO_INCORE` seems familiar, that's because it is - it is effectively the _exact_ same thing as `SCF_SUBTYPE=OUT_OF_CORE` for MemDFJK (in https://github.com/psi4/psi4/pull/2848) and DiskDFJK (in https://github.com/psi4/psi4/pull/2924).

The goal of this PR is to clean that up; `PK_NO_INCORE` is replaced by `SCF_SUBTYPE` for PKJK. `SCF_SUBTYPE=OUT_OF_CORE` has the exact same effect that `PK_NO_INCORE=TRUE` had previously. `SCF_SUBTYPE=AUTO` has PKJK select in-core or out-of-core by default, as per usual. Finally, `SCF_SUBTYPE=INCORE` allows PKJK to force-execute its in-core algorithm, a new functionality for PKJK as far as I am aware. As usual with `SCF_SUBTYPE`, setting it to `INCORE` without allocating sufficient memory to Psi4 will throw an exception.

- [X] The `PK_NO_INCORE` keyword has been removed.
- [X] The `SCF_SUBTYPE` keyword has been added for PKJK. `SCF_SUBTYPE=AUTO` has PKJK select the subalgorithm by default. `SCF_SUBTYPE=OUT_OF_CORE` forces PKJK to use one of the two out-of-core algorithms (equivalent to `PK_NO_INCORE=TRUE` previously). `SCF_SUBTYPE=INCORE` forces PKJK to use its in-core algorithm, and throws an exception if insufficient memory is allocated.

## Dev notes & details
- [X] The `PK_NO_INCORE` keyword for PKJK has been replaced with the new `SCF_SUBTYPE` keyword.

## Questions
-N/A

## Checklist
- [X] Tests added for any

 new features
## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -215,7 +215,7 @@ def scf_initialize(self):
         core.print_out("\n  ==> Pre-Iterations <==\n\n")
 
         # force SCF_SUBTYPE to AUTO during SCF guess
-        optstash = p4util.OptionsState(["SCF_SUBTYPE"])
+        optstash = p4util.OptionsState(["SCF", "SCF_SUBTYPE"])
         core.set_local_option("SCF", "SCF_SUBTYPE", "AUTO")
 
         core.timer_on("HF: Guess")
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -214,16 +214,15 @@ def scf_initialize(self):
 
         core.print_out("\n  ==> Pre-Iterations <==\n\n")
 
-        # disable scf_subtype for SAD guess
-        scf_subtype_cache = core.get_local_option("SCF", "SCF_SUBTYPE")
+        # force SCF_SUBTYPE to AUTO during SCF guess
+        optstash = p4util.OptionsState(["SCF_SUBTYPE"])
         core.set_local_option("SCF", "SCF_SUBTYPE", "AUTO")
 
         core.timer_on("HF: Guess")
         self.guess()
         core.timer_off("HF: Guess")
 
-        # re-enable scf_subtype after SAD guess is done
-        core.set_local_option("SCF", "SCF_SUBTYPE", scf_subtype_cache)
+        optstash.restore()
 
         # Print out initial docc/socc/etc data
         if self.get_print():                    
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -214,9 +214,17 @@ def scf_initialize(self):
 
         core.print_out("\n  ==> Pre-Iterations <==\n\n")
 
+        # disable scf_subtype for SAD guess
+        scf_subtype_cache = core.get_local_option("SCF", "SCF_SUBTYPE")
+        core.set_local_option("SCF", "SCF_SUBTYPE", "AUTO")
+
         core.timer_on("HF: Guess")
         self.guess()
         core.timer_off("HF: Guess")
+
+        # re-enable scf_subtype after SAD guess is done
+        core.set_local_option("SCF", "SCF_SUBTYPE", scf_subtype_cache)
+
         # Print out initial docc/socc/etc data
         if self.get_print():                    
             lack_occupancy = core.get_local_option('SCF', 'GUESS') in ['SAD']
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -556,6 +556,9 @@ void throw_deprecation_errors(std::string const& key, std::string const& module 
     if (module == "SCF" && key == "PK_NO_INCORE") {
         py_psi_print_out("WARNING!\n\tRemove keyword PK_NO_INCORE! PK_NO_INCORE has been replaced by the SCF_SUBTYPE=NO_INCORE option. Using PK_NO_INCORE will raise an error in v1.8.\n");
     }
+    if (module == "SCF" && key == "PK_ALGO") {
+        py_psi_print_out("WARNING!\n\tRemove keyword PK_ALGO! PK_ALGO has been replaced by the SCF_SUBTYPE=YOSHIMINE_OUT_OF_CORE and REORDER_OUT_OF_CORE options. Using PK_ALGO will raise an error in v1.8.\n");
+    }
 }
 
 Options& py_psi_get_options() { return Process::environment.options; }
--- a/psi4/src/core.cc
+++ b/psi4/src/core.cc
@@ -553,6 +553,9 @@ void throw_deprecation_errors(std::string const& key, std::string const& module 
     if (module == "SCF" && key == "DIIS_MIN_VECS") {
         py_psi_print_out("WARNING!\n\tRemove keyword DIIS_MIN_VECS! This keyword does nothing. Using it will raise an error in v1.7.\n");
     }
+    if (module == "SCF" && key == "PK_NO_INCORE") {
+        py_psi_print_out("WARNING!\n\tRemove keyword PK_NO_INCORE! PK_NO_INCORE has been replaced by the SCF_SUBTYPE=NO_INCORE option. Using PK_NO_INCORE will raise an error in v1.8.\n");
+    }
 }
 
 Options& py_psi_get_options() { return Process::environment.options; }
--- a/psi4/src/psi4/libfock/PKmanagers.cc
+++ b/psi4/src/psi4/libfock/PKmanagers.cc
@@ -111,17 +111,19 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
     bool do_reord = false;
     bool do_yosh = false;
     bool do_incore = false;
-    if (options["PK_ALGO"].has_changed() && subalgo != "INCORE") {
-        if (algo == "REORDER") {
-            do_reord = true;
-        } else if (algo == "YOSHIMINE") {
-            do_yosh = true;
-        }
-    } else {
-        if (algo_factor * memory > pk_size) {
-            do_reord = true;
+    if (subalgo != "INCORE") {
+        if (options["PK_ALGO"].has_changed()) { 
+            if (algo == "REORDER") {
+                do_reord = true;
+            } else if (algo == "YOSHIMINE") {
+                do_yosh = true;
+            }
         } else {
-            do_yosh = true;
+            if (algo_factor * memory > pk_size) {
+                do_reord = true;
+            } else {
+                do_yosh = true;
+            }
         }
     }
     
--- a/psi4/src/psi4/libfock/PKmanagers.cc
+++ b/psi4/src/psi4/libfock/PKmanagers.cc
@@ -88,7 +88,7 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
                                                       size_t memory, Options& options, bool dowK, double omega_in) {
     std::string algo = options.get_str("PK_ALGO");
     std::string subalgo = options.get_str("SCF_SUBTYPE");
-    
+
     // We introduce another safety factor in the memory, otherwise
     // we are apparently prone to being killed by the OS.
     // TODO: Check for memory leaks ? Trace memory usage ?
@@ -106,13 +106,13 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
     if (dowK) {
         ncorebuf = 3;
     }
-  
-    // determine which sub-algorithm to use  
+
+    // determine which sub-algorithm to use
     bool do_reord = false;
     bool do_yosh = false;
     bool do_incore = false;
     if (subalgo != "INCORE") {
-        if (options["PK_ALGO"].has_changed()) { 
+        if (options["PK_ALGO"].has_changed()) {
             if (algo == "REORDER") {
                 do_reord = true;
             } else if (algo == "YOSHIMINE") {
@@ -126,11 +126,11 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
             }
         }
     }
-    
+
     // throw exception if INCORE requested without enough memory...
     if (ncorebuf * pk_size > memory && subalgo == "INCORE") {
         throw PSIEXCEPTION("SCF_SUBTYPE=INCORE was specified, but there is not enough memory to do in-core! Increase the amount of memory allocated to Psi4 or allow for out-of-core to be used.\n");
-    
+
     // ..or do INCORE if OUT_OF_CORE not explicitly requested and enough memory 
     } else if (ncorebuf * pk_size < memory && subalgo != "OUT_OF_CORE")  {
 	do_incore = true;
--- a/psi4/src/psi4/libfock/PKmanagers.cc
+++ b/psi4/src/psi4/libfock/PKmanagers.cc
@@ -131,7 +131,7 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
     if (ncorebuf * pk_size > memory && subalgo == "INCORE") {
         throw PSIEXCEPTION("SCF_SUBTYPE=INCORE was specified, but there is not enough memory to do in-core! Increase the amount of memory allocated to Psi4 or allow for out-of-core to be used.\n");
 
-    // ..or do INCORE if OUT_OF_CORE not explicitly requested and enough memory 
+    // ..or do INCORE if OUT_OF_CORE not explicitly requested and enough memory
     } else if (ncorebuf * pk_size < memory && subalgo != "OUT_OF_CORE")  {
 	do_incore = true;
     }
--- a/psi4/src/psi4/libfock/PKmanagers.cc
+++ b/psi4/src/psi4/libfock/PKmanagers.cc
@@ -87,8 +87,9 @@ void ijklBasisIterator::next() {
 std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio, std::shared_ptr<BasisSet> primary,
                                                       size_t memory, Options& options, bool dowK, double omega_in) {
     std::string algo = options.get_str("PK_ALGO");
-    bool noincore = options.get_bool("PK_NO_INCORE");
-
+    std::string subalgo = options.get_str("SCF_SUBTYPE");
+    bool noincore = subalgo == "OUT_OF_CORE";
+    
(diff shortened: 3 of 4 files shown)


loriab:
Nice idea to apply SCF_SUBTYPE to PK, too. Only thing that concerns me is that now selecting the algorithm (which most ppl won't) is a three-keyword affair: `scf_type=pk, scf_subtype=out_of_core, pk_algo=yoshimine`. What do you/others think of dropping `pk_algo` and having several scf_subtype options applicable to PK: `AUTO` (of course), `INCORE`, `YOSHIMINE_OUT_OF_CORE`, `REORDER_OUT_OF_CORE`, `OUT_OF_CORE`?

davpoolechem:
> Nice idea to apply SCF_SUBTYPE to PK, too. Only thing that concerns me is that now selecting the algorithm (which most ppl won't) is a three-keyword affair: `scf_type=pk, scf_subtype=out_of_core, pk_algo=yoshimine`. What do you/others think of dropping `pk_algo` and having several scf_subtype options applicable to PK: `AUTO` (of course), `INCORE`, `YOSHIMINE_OUT_OF_CORE`, `REORDER_OUT_OF_CORE`, `OUT_OF_CORE`?

Thank you! I had thought about doing this very thing, and I am fine with it if everyone else is. My initial hesitation in doing so, is that `YOSHIMINE_OUT_OF_CORE` and `REORDER_OUT_OF_CORE` are completely inapplicable to the DF algorithms; whereas with the current arrangement, all of the options in `PK_ALGO` and `SCF_SUBTYPE` can be used with their respective applicable algorithms.

Practically speaking, it's not a huge problem, since MemDFJK and DiskDFJK already throw exceptions for incompatible `SCF_SUBTYPE` options (as `YOSHIMINE_OUT_OF_CORE` and `REORDER_OUT_OF_CORE` would be). It just feels a bit... not ideal to me, I guess? If this option is considered better than the three-keyword approach currently used, I'll be happy to implement it, however!

loriab:
The some-subtype-values-not-applicable-to-some-scftype-values doesn't bother me too much -- after all, direct, cd, etc. can't take any subtype value besides auto. So long as the read_options docstring is clear and the code rejects non-meaningful subtype choices, I don't think users are led astray.

All the same, I'm not hearty on my suggestion because it seems a little indistinct between algorithm choice (usually scftype) and memory transition choice (usually subtype). Maybe, if designed from scratch, there'd be `YOSHIMINE_PK`, `REORDERED_PK`, and `PK` as `scftype`s and then PK `subtype`s `OUT_OF_CORE` and `INCORE` would fit the DF pattern nicely. That might well be considered too much reengineering at this point in the release cycle.

So minimum change, unless others chime in, is leave as-is and add the `PK_NO_INCORE` to the deprecation messages here https://github.com/psi4/psi4/blob/master/psi4/src/core.cc#L553

davpoolechem:
> The some-subtype-values-not-applicable-to-some-scftype-values doesn't bother me too much -- after all, direct, cd, etc. can't take any subtype value besides auto. So long as the read_options docstring is clear and the code rejects non-meaningful subtype choices, I don't think users are led astray.

That's all fair, to be honest, although I agree moreso with your second paragraph below, and what you said about indistinctness between algorithm choice and memory transition choice is applicable for sure.
 
> All the same, I'm not hearty on my suggestion because it seems a little indistinct between algorithm choice (usually scftype) and memory transition choice (usually subtype). Maybe, if designed from scratch, there'd be `YOSHIMINE_PK`, `REORDERED_PK`, and `PK` as `scftype`s and then PK `subtype`s `OUT_OF_CORE` and `INCORE` would fit the DF pattern nicely. That might well be considered too much reengineering at this point in the release cycle.

This is true, and a point I hadn't considered, either. I wouldn't mind taking such a project on, but I would probably save it for after v1.8.

> So minimum change, unless others chime in, is leave as-is and add the `PK_NO_INCORE` to the deprecation messages here https://github.com/psi4/psi4/blob/master/psi4/src/core.cc#L553

Will do!

davpoolechem:
Deprecation message added! I'm less familiar with deprecating things in Psi4, so let me know if anything else should be done on that front.

loriab (approved):

psi4/src/read_options.cc
```diff
@@ -1403,8 +1403,6 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_int("PK_MAX_BUCKETS", 500);
         /*- Select the PK algorithm to use. For debug purposes, selection will be automated later. !expert -*/
         options.add_str("PK_ALGO", "REORDER", "REORDER YOSHIMINE");
```
JonathonMisiewicz, line 1405: After this PR, does this keyword name make sense?

loriab: I think so -- it's still the only way to choose btwn reorder and yoshimine when scf_type=PK. Am I missing your point?

JonathonMisiewicz: If PK is the algorithm and the subalgorithm is in core or out-of-core, is this a sub-sub-algorithm?

davpoolechem: > EDIT: docstring might be edited to mention if you want PK+INCORE, use subtype.

This is a good idea! I'd be happy to put that in.

davpoolechem: > If PK is the algorithm and the subalgorithm is in core or out-of-core, is this a sub-sub-algorithm?

One could frame it as such, yes - `YOSHIMINE` and `REORDER` are effectively sub-sub-algorithms to the out-of-core PK subalgorithm.

loriab: > is this a sub-sub-algorithm?

Yeah, that's what I didn't like about it. But the most logical fix I could devise (see above) is expanding scf_type, so this is the least intrusive for now.

JonathonMisiewicz: What about collapsing `PK_ALGO` into `SCF_SUBTYPE`? So for PK, allowed options are `INCORE`, `OOC_REORDER`, and `OOC_YOSHIMINE`?

davpoolechem: If there's enough support for folding `PK_ALGO` into `SCF_SUBTYPE`, however, I wouldn't mind doing that for this PR. I would also be happy to tackle Lori's proposal, but I'd rather that be for another PR.

JonathonMisiewicz: I'm still for folding `PK_ALGO` into `SCF_SUBTYPE`, thanks for considering.

Agreed that Lori's proposal is beyond-scope.

davpoolechem:
Hi all! I was working on folding the `PK_ALGO` keywords into `SCF_SUBTYPE`, and I ran into an issue that may be worth discussing. 

By default, using `OOC_YOSHIMINE` and `OOC_REORDER` as `SCF_SUBTYPE`s the way we were discussing causes the calculation to throw an exception. The issue is because, in the SAD guess, DFJK is used by default; and `SCF_SUBTYPE` applies to the SAD guess DFJK (at least MemDFJK currently, and soon DiskDFJK) as well. Since `OOC_YOSHIMINE` and `OOC_REORDER` would be used as the keyword for the SAD DFJK, and since DFJK doesn't support those keywords explicitly for `SCF_SUBTYPE`, the calculation throws an exception. Note that setting `SAD_SCF_TYPE=DIRECT` eliminates this issue entirely.

There are probably a few ways to resolve this, if we choose to explicitly continue on this path:

1. As mentioned above, simply force `SAD_SCF_TYPE` to another option (like `DIRECT`) for PK calculations. I'll admit I'm not a big fan of this approach since DF is more performant, although it may not matter in the grand scheme of things since it's just the SAD guess. 
2. Force `OOC_YOSHIMINE` and `OOC_REORDER` to work with the DFJK algorithms. The logical way to do this would be to have those keywords act the exact same as `SCF_SUBTYPE=`OUT_OF_CORE` for DFJK. I'm a bit iffy on this one, as well, since it may imply that the three out-of-core keywords have different impacts on DFJK when they really don't. Documentation can clear that up, but still.

What does everyone think?

davpoolechem:
> If it's not awful to implement, I'd make the SAD guess impervious to SCF_SUBTYPE. The user setting SCF_SUBTYPE is thinking about the main iterations, not the guess, and if this is some highly technical timing, the guess type might be reset anyways.

Yeah, I'd agree with this. I can look into it and see if it's a quick fix or not.

> Also, I think it should be `REORDER_OUT_OF_CORE` not `REORDER_OOC` for consistency with the other values.

Yeah, I'll do that, as well!

davpoolechem:
All right, I disabled `SCF_SUBTYPE` for SAD guess runs! I went with a pretty basic solution - just adjust the Python driver and force-set `SCF_SUBTYPE=AUTO` for the guess part of the run. Let me know if this is an acceptable solution.

loriab:
> All right, I disabled SCF_SUBTYPE for SAD guess runs! I went with a pretty basic solution - just adjust the Python driver and force-set SCF_SUBTYPE=AUTO for the guess part of the run. Let me know if this is an acceptable solution.

Ah, turning it off py-side is simple and great, but there's a better way. The minor trouble with your approach is that it's restoring the value but not the status of whether the user has set it. That latter piece of info is often used to decide whether an option is default or not or whether, once control has returned to the user, to write the option explicitly in subsequent calcs. The usual remedy that you'll see scattered in proc.py are optstash=OptionsState([list o options]) ... do work ... optstash.restore(). So I'd suggest using that or the content manager like https://github.com/psi4/psi4/pull/2926/files#diff-2aec42d854b50db94cde6a3f9d3c665eed11e8f1f083040807d9e4dd4696895aL67-L68 `["SCF", "SCF_SUBTYPE"]` for "local" setting.

davpoolechem:
> > All right, I disabled SCF_SUBTYPE for SAD guess runs! I went with a pretty basic solution - just adjust the Python driver and force-set SCF_SUBTYPE=AUTO for the guess part of the run. Let me know if this is an acceptable solution.
> 
> Ah, turning it off py-side is simple and great, but there's a better way. The minor trouble with your approach is that it's restoring the value but not the status of whether the user has set it. That latter piece of info is often used to decide whether an option is default or not or whether, once control has returned to the user, to write the option explicitly in subsequent calcs. The usual remedy that you'll see scattered in proc.py are optstash=OptionsState([list o options]) ... do work ... optstash.restore(). So I'd suggest using that or the content manager like https://github.com/psi4/psi4/pull/2926/files#diff-2aec42d854b50db94cde6a3f9d3c665eed11e8f1f083040807d9e4dd4696895aL67-L68 `["SCF", "SCF_SUBTYPE"]` for "local" setting.

Done and done! This is all great knowledge; thank you!

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -214,9 +214,16 @@ def scf_initialize(self):
 
         core.print_out("\n  ==> Pre-Iterations <==\n\n")
 
+        # force SCF_SUBTYPE to AUTO during SCF guess
+        optstash = p4util.OptionsState(["SCF_SUBTYPE"])
+        core.set_local_option("SCF", "SCF_SUBTYPE", "AUTO")
```
loriab, line 219: either the first line needs to be `["SCF", "SCF_SUBTYPE"]` or the second needs to be `set_global_option`

davpoolechem: Ah, I see. Addressed!

psi4/src/read_options.cc
```diff
@@ -1419,9 +1415,10 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
             depending on available memory or other hardware constraints, allow the best
             sub-algorithm for the molecule and conditions (``AUTO`` ; usual mode) or
             forcibly select a sub-algorithm (usually only for debugging or profiling).
-            Presently, ``SCF_TYPE=MEM_DF`` can have ``INCORE`` or ``OUT_OF_CORE`` selected.
-            (This also applies for ``SCF_TYPE=DF`` when ``MEM_DF`` active.) !expert -*/
-	options.add_str("SCF_SUBTYPE", "AUTO", "AUTO INCORE OUT_OF_CORE");
+            Presently, ``SCF_TYPE=MEM_DF`` (including ``SCF_TYPE=DF`` using the in-core subalgorithm)
+	    can have ``INCORE`` and ``OUT_OF_CORE`` selected; and ``SCF_TYPE=PK``  can have ``INCORE``,
+	    ``OUT_OF_CORE``, ``OOC_YOSHIMINE``, and ``OOC_REORDER`` selected. !expert -*/
+	options.add_str("SCF_SUBTYPE", "AUTO", "AUTO INCORE OUT_OF_CORE OOC_YOSHIMINE OOC_REORDER");
```
loriab, line 1421:
```suggestion
	    ``OUT_OF_CORE``, ``YOSHIMINE_OUT_OF_CORE``, and ``REORDER_OUT_OF_CORE`` selected. !expert -*/
	options.add_str("SCF_SUBTYPE", "AUTO", "AUTO INCORE OUT_OF_CORE YOSHIMINE_OUT_OF_CORE REORDER_OUT_OF_CORE");
```
I put the algo first in analogy to `MEM_DF` vs `DF`

psi4/src/core.cc
```diff
@@ -553,6 +553,9 @@ void throw_deprecation_errors(std::string const& key, std::string const& module
     if (module == "SCF" && key == "DIIS_MIN_VECS") {
         py_psi_print_out("WARNING!\n\tRemove keyword DIIS_MIN_VECS! This keyword does nothing. Using it will raise an error in v1.7.\n");
     }
+    if (module == "SCF" && key == "PK_NO_INCORE") {
+        py_psi_print_out("WARNING!\n\tRemove keyword PK_NO_INCORE! PK_NO_INCORE has been replaced by the SCF_SUBTYPE=NO_INCORE option. Using PK_NO_INCORE will raise an error in v1.8.\n");
+    }
```
loriab, line 558:
```suggestion
    if (module == "SCF" && key == "PK_NO_INCORE") {
        py_psi_print_out("WARNING!\n\tRemove keyword PK_NO_INCORE! PK_NO_INCORE has been replaced by the SCF_SUBTYPE=NO_INCORE option. Using PK_NO_INCORE will raise an error in v1.8.\n");
    }
    if (module == "SCF" && key == "PK_ALGO") {
        py_psi_print_out("WARNING!\n\tRemove keyword PK_ALGO! PK_ALGO has been replaced by the SCF_SUBTYPE=YOSHIMINE_OUT_OF_CORE and REORDER_OUT_OF_CORE options. Using PK_ALGO will raise an error in v1.8.\n");
    }
```

psi4/src/psi4/libfock/PKmanagers.cc
```diff
@@ -107,24 +110,48 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
...
+	} else if (algo_factor * memory > pk_size) {
+            do_reord = true;
+	} else {
+            do_yosh = true;
+	}
+
+    // throw an exception on an invalid SCF_SUBTYPE
+    } else {
+        throw PSIEXCEPTION("Invalid SCF_SUBTYPE option! The valid choices of SCF_SUBTYPE for SCF_TYPE=PK are AUTO, INCORE, OUT_OF_CORE, OOC_YOSHIMINE, and OOC_REORDER.");
+    }
```
loriab, line 154: `OOC_YOSHIMINE` and `OOC_REORDER` need switching out in this file, too.

psi4/src/psi4/libfock/PKmanagers.cc
```diff
@@ -107,24 +110,48 @@ std::shared_ptr<PKManager> PKManager::build_PKManager(std::shared_ptr<PSIO> psio
...
+            throw PSIEXCEPTION("SCF_SUBTYPE=INCORE was specified, but there is not enough memory to do in-core! Increase the amount of memory allocated to Psi4 or allow for out-of-core to be used.\n");
+        } else {
+            do_incore = true;
+	}
+
+    // ...or just let psi4 pick any subalgorithm
+    } else if (subalgo == "AUTO") {
+        if (ncorebuf * pk_size < memory) {
+            do_incore = true;
+	} else if (algo_factor * memory > pk_size) {
```
loriab, line 145: logic lgtm, but the formatting/spacing is deceptive. :-)

loriab (changes requested):

psi4/src/psi4/libfock/PKmanagers.cc
davpoolechem: Oh right, forgot to switch that out. Will do!

davpoolechem: Done and done!

psi4/src/psi4/libfock/PKmanagers.cc
loriab: `do_record` on line 120 should be a tab inwards. `} else if` on 146 should be aligned with 144. 148 and 149 are two tabs apart. It's very common in your PRs, so you might want to look at your editor's whitespace settings.

davpoolechem: Ah, I see. Yeah, that's something I've been meaning to do, and just haven't gotten around to. I'll do that now and recommit with proper spacing.

davpoolechem: All right, it should be better now, I think

loriab (approved):
thanks! much easier to read.

jturney (approved):

davpoolechem:
So now that this PR and https://github.com/psi4/psi4/pull/2924 have been approved for adding to the merge queue, my guess is that there will be a merge conflict between the two, at the `SCF_SUBTYPE` docstring in scf_subtype.cc.

davpoolechem:
I suppose if anyone wants to check the merged `SCF_SUBTYPE` docstring, they can; else, I'll add this to the merge queue by the time I finish up today.

PR #2929, opened 2023-04-24
Fix for QCFractal next beta10

loriab:
- [x] fix bug where `optimize(..., engine="geometric", molecule=mol)` would not run
- [x] undo some hack fixes and make distributed driver runs through QCFractal "next" branch more reliable by making array qcvars shaped numpy arrays. Works with QCFractal >=v0.50beta10. Show examples of using `get_psi_results()` to prepare the results of distributed driver runs in "normal" psi4 way.

## Dev notes & details
- [x] In the early days (before ddd entered psi4), all array-like pieces of AtomicResults came back as flat lists, so we had to "plump" them into e.g., (nat, 3) for gradient. Then, by the time of v1.6 and QCFractal master v15.8, the AtomicResults preserved all that, so we stopped using the plumper. Now in qcfractal next, they're back to lists again _and_ qcvars have been mingled with other properties, so both separation and plumping are required.
- [x] promote `get_psi_results` as a drop-in replacement after running distributed. It sets up a dummy wfn and qcvariables, just like an internal run, so one doesn't have to deal with picking results off the AtomicResult QCSchema.
- [x] Added labels to tests (mostly findif) so that one can catch all ddd-using tests with `pytest -m "findif and cbs and nbody"`
- [x] set up a mechanism where can parameterize both pytests and ctests with a snowflake with minimum clutter.

## Questions
- [ ] This won't work with current qcf release. I'll ask @bennybp to mint a beta10, then this should work.

## Checklist

## Status
- [x] Ready for review (for review, the 4th commit has the material changes. 3rd are test changes. 1st & 2nd are labeling and syntax.)
- [ ] Ready for merge


psi4/driver/task_base.py
```diff
@@ -35,6 +35,7 @@
 import abc
 import copy
 import logging
+import pprint
```
maxscheurer, line 38: Is `pprint` used or just a debug leftover?

maxscheurer (approved):

jturney (approved):

PR #2931, opened 2023-04-24
COSX Reformulation for Negative Grid Weights

davpoolechem:
## Description
This PR is a follow-up to https://github.com/psi4/psi4/pull/2906, and what can be considered an official solution to the issue discussed in https://github.com/psi4/psi4/issues/2890. The current issue is that COSX does not work with certain grid configurations - specifically, it does not work with grids that have negative grid weights, due to the use of an intermediate matrix in COSX that uses the square root of negative grid weights. https://github.com/psi4/psi4/pull/2906 "fixed" this issue by having COSX throw an exception when grids with negative weights were encountered. This PR provides a COSX reformulation that allows COSX to work with negative grid weights.

The reformulation does two primary things:
- The $X$ matrix (Eq. 4 in [Neese 2009](https://doi.org/10.1016/j.chemphys.2008.10.036)) is redefined as $X_{\kappa g} = \sqrt{|w_{g}|} \kappa(r_{g})$. In words, $X_{\kappa g}$ now uses the square root of the _magnitude_ of the weights, instead of the square root of the raw weights.
- To correct for the above, the computation of the $G$ matrix (Eq. 7 in [Neese 2009](https://doi.org/10.1016/j.chemphys.2008.10.036)) is performed as $G_{\nu g} = \sum_{\tau} \text{sign}(w_{g}) A_{\nu \tau} (r_{g}) F_{\tau g}$ . In words, the sign of the corresponding grid weights are included in the formation of $G_{\nu g}$, when $A_{\nu \tau}$ and $F_{\tau g}$ are contracted.

One other thing comes out as a consequence of this:
- For overlap fitting, the numerical overlap matrix $S_{N}$ (Eq. 13 in [Izsák 2011](https://doi.org/10.1063/1.3646921)) is computed as $S_{N} = X * X_{\text{alt}}^{T}$, where $X_{\kappa g} = \sqrt{|w_{g}|} \kappa(r_{g})$ (the same as the first bullet point in the reformulation above), and $X_{\text{alt},\kappa g} = \text{sign}(w_{g}) \sqrt{|w_{g}|} \kappa(r_{g})$. Essentially, $S_{N}$ now uses two variants of the $X$ matrix in its formulation, one of which folds the grid weight sign into itself.

- [X] COSX can now be used with a wider variety of grids, as a bug preventing COSX to be used with specific grid configurations has been fixed.
## Dev notes & details
- [X] The COSX implementation has been reformulated to enable calculations with grids containing negative grid weights.

## Questions
- [x] Is the current way of handling computation of the numerical overlap matrix acceptable? Currently, two X matrices are used to form the numerical overlap matrix. One goal of folding the grid weight sign into the formation of G was to prevent the use of multiple X matrices. However, I don't think that can be done here without removing the call to `linalg::doublet` in the `compute_numeric_overlap` function. One could probably do a slightly-modified, manually-implemented matrix multiply with the grid weight folded in, and construct $S_{N}$ with a single $X$ matrix that way. But that comes at the cost of not utilizing BLAS. What does everyone consider preferable?
- [ ] What, exactly, is the best way to test this? I would like to add a test to `test_dfjcosk.py` for a new grid weight option with negative weights. Unfortunately, it seems that other codes that have COSX just don't have grid options that lead to negative grid weights. For now, I've been comparing against Psi4 COSX calculations using similar grid sizes and ensuring the calculations have similar energies, but I certainly wouldn't consider that robust.

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/CompositeJK.cc
+++ b/psi4/src/psi4/libfock/CompositeJK.cc
@@ -258,7 +258,6 @@ void CompositeJK::common_init() {
             linK_ints_cutoff_ = options_.get_double("LINK_INTS_TOLERANCE");
         } else {
             linK_ints_cutoff_ = cutoff_;
-
         }
     
     // Chain-of-Spheres Exchange (COSX)
@@ -329,6 +328,7 @@ void CompositeJK::common_init() {
 	            }
             }
 	        if (warning_printed_init) break;
+        }
 
         auto warning_printed_final = false;
         for (const auto &final_block : grid_final_->blocks()) {
@@ -392,7 +392,6 @@ void CompositeJK::common_init() {
         ;
     } else {
         throw PSIEXCEPTION("Invalid Composite K algorithm selected!");
-
     }
 }
 
--- a/psi4/src/psi4/libfock/CompositeJK.cc
+++ b/psi4/src/psi4/libfock/CompositeJK.cc
@@ -249,6 +249,7 @@ void CompositeJK::common_init() {
         throw PSIEXCEPTION("Invalid Composite J algorithm selected!");
     }
 
+<<<<<<< HEAD:psi4/src/psi4/libfock/CompositeJK.cc
     // => Set up separate K algorithm <= //
 
     // Linear Exchange (LinK)
@@ -258,12 +259,14 @@ void CompositeJK::common_init() {
             linK_ints_cutoff_ = options_.get_double("LINK_INTS_TOLERANCE");
         } else {
             linK_ints_cutoff_ = cutoff_;
+
         }
     
     // Chain-of-Spheres Exchange (COSX)
     } else if (k_type_ == "COSX") {
         timer_on("CompositeJK: COSX Grid Construction");
 
+<<<<<<< HEAD:psi4/src/psi4/libfock/CompositeJK.cc
         // TODO: specify bool "DFT_REMOVE_DISTANT_POINTS" in the DFTGrid constructors
 
         // Create a small DFTGrid for the initial SCF iterations
@@ -322,7 +325,7 @@ void CompositeJK::common_init() {
             const auto w = init_block->w();
             for (int ipoint = 0; ipoint < init_block->npoints(); ++ipoint) {
                 if (w[ipoint] < 0.0) {
-	                outfile->Printf("  WARNING: The definition of the current initial grid includes negative weights, which the original COSX formulation does not support! If this is of concern, please choose another initial grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_INITIAL.\n");
+	                outfile->Printf("  WARNING: The definition of the current initial grid includes negative weights, which the original COSX formulation does not support!\n    If this is of concern, please choose another initial grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_INITIAL.\n\n");
 	                warning_printed_init = true;
 		            break;
 	            }
@@ -335,7 +338,7 @@ void CompositeJK::common_init() {
             const auto w = final_block->w();
             for (int ipoint = 0; ipoint < final_block->npoints(); ++ipoint) {
                 if (w[ipoint] < 0.0) {
-	                outfile->Printf("  WARNING: The definition of the current final grid includes negative weights, which the original COSX formulation does not support! If this is of concern, please choose another final grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_FINAL.\n");
+	                outfile->Printf("  WARNING: The definition of the current final grid includes negative weights, which the original COSX formulation does not support!\n    If this is of concern, please choose another final grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_FINAL.\n\n");
                     warning_printed_final = true;
 	                break;
 	            }
@@ -1596,7 +1599,7 @@ void CompositeJK::build_COSK(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
 
                     // contract A_nu_tau with F_tau to get contribution to G_nu
                     // symmetry permitting, also contract A_nu_tau with F_nu to get contribution to G_tau
-                    // we fold sign(w) into the formation of G to correct for the modified definition of X 
+                    // we fold sign(w) into the formation of G to correct for the modified definition of X
                     for(size_t jki = 0; jki < njk; jki++) {
                         auto F_blockp = F_block[jki]->pointer();
                         auto G_blockp = G_block[jki]->pointer();
--- a/psi4/src/psi4/libfock/CompositeJK.cc
+++ b/psi4/src/psi4/libfock/CompositeJK.cc
@@ -1448,14 +1448,14 @@ void CompositeJK::build_COSK(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
 
         // resize the buffer of basis function values
         auto X_block_nosign = std::make_shared<Matrix>(npoints_block, nbf_block);  // points x nbf_block
-        auto X_block_sign = std::make_shared<Matrix>(npoints_block, nbf_block);  // points x nbf_block
+        //auto X_block_sign = std::make_shared<Matrix>(npoints_block, nbf_block);  // points x nbf_block
 
         auto X_block_nosignp = X_block_nosign->pointer();
-	auto X_block_signp = X_block_sign->pointer();
+	//auto X_block_signp = X_block_sign->pointer();
         for (size_t p = 0; p < npoints_block; p++) {
             for (size_t k = 0; k < nbf_block; k++) {
                 X_block_nosignp[p][k] = point_values->get(p, k) * std::sqrt(std::fabs(w[p]));
-                X_block_signp[p][k] = sign(w[p])*X_block_nosignp[p][k];
+                //X_block_signp[p][k] = sign(w[p])*X_block_nosignp[p][k];
             }
         }
 
@@ -1464,11 +1464,11 @@ void CompositeJK::build_COSK(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
         auto X_block_bfmaxp = X_block_bfmax.pointer();
         for (size_t p = 0; p < npoints_block; p++) {
             for (size_t k = 0; k < nbf_block; k++) {
-                X_block_bfmaxp[p] = std::max(X_block_bfmaxp[p], std::abs(X_block_signp[p][k]));
+                X_block_bfmaxp[p] = std::max(X_block_bfmaxp[p], std::abs(X_block_nosignp[p][k]));
             }
         }
 
-        double X_block_max = X_block_sign->absmax();
+        double X_block_max = X_block_nosign->absmax();
 
         // => F Matrix <= //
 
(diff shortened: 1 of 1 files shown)


psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -73,19 +80,28 @@ Matrix compute_numeric_overlap(const DFTGrid &grid, const std::shared_ptr<BasisS
         bf_computer.compute_functions(block);
         auto point_values = bf_computer.basis_values()["PHI"];
 
+        // lambda for returning sign of double
+        auto sign = [ ](double val) {
```
susilehtola, line 84: I do not like this: the return type should be obvious from the function definition.

It looks like the function returns an `int`.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -939,6 +969,12 @@ void DFJCOSK::build_K(std::vector<std::shared_ptr<Matrix>>& D, std::vector<std::
 
         const auto & int_buff = int_computers[rank]->buffers()[0];
 
+        // lambda for returning sign of double
+        // needed for formation of G
+        auto sign = [ ](double val) {
```
susilehtola, line 974: same comment here

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -239,26 +255,35 @@ void DFJCOSK::common_init() {
     };
     grid_final_ = std::make_shared<DFTGrid>(primary_->molecule(), primary_, grid_final_int_options, grid_final_str_options, grid_final_float_options, options_);
 
-    // Sanity-check of grids to ensure no negative grid weights
-    // COSX crashes when grids with negative weights are used,
+    // Print out warning if grid with negative grid weights is used
+    // Original Nesse COSX formulation does not support negative grid weights
```
susilehtola, line 259: Neese

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -239,26 +255,35 @@ void DFJCOSK::common_init() {
...
-    // See https://github.com/psi4/psi4/issues/2890
+    // the Psi4 COSX implementation is slightly modified to work with negative grid weights
+    // See https://github.com/psi4/psi4/issues/2890 for discussion
+    auto warning_printed_init = false;
     for (const auto &init_block : grid_init_->blocks()) {
         const auto w = init_block->w();
         for (int ipoint = 0; ipoint < init_block->npoints(); ++ipoint) {
             if (w[ipoint] < 0.0) {
-	      throw PSIEXCEPTION("The definition of the current initial grid includes negative weights. As these are not suitable for the COSX implementation, please choose another initial grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_INITIAL.");
+	        outfile->Printf("  WARNING: The definition of the current initial grid includes negative weights, which the original COSX formulation does not support!\n    If this is of concern, please choose another initial grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_INITIAL.\n\n");
```
susilehtola, line 268: I think this should not be a WARNING but INFO (does this exist in Psi4?)

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -239,26 +255,35 @@ void DFJCOSK::common_init() {
...
+	if (warning_printed_init) break;
     }
 
+    auto warning_printed_final = false;
     for (const auto &final_block : grid_final_->blocks()) {
         const auto w = final_block->w();
         for (int ipoint = 0; ipoint < final_block->npoints(); ++ipoint) {
             if (w[ipoint] < 0.0) {
-	      throw PSIEXCEPTION("The definition of the current final grid includes negative weights. As these are not suitable for the COSX implementation, please choose another final grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_FINAL.");
+	        outfile->Printf("  WARNING: The definition of the current final grid includes negative weights, which the original COSX formulation does not support!\n    If this is of concern, please choose another final grid through adjusting either COSX_PRUNING_SCHEME or COSX_SPHERICAL_POINTS_FINAL.\n\n");
```
susilehtola, line 281: same here

susilehtola (changes requested):
Looks pretty good, please fix the issues above

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -239,26 +255,35 @@ void DFJCOSK::common_init() {
     };
     grid_final_ = std::make_shared<DFTGrid>(primary_->molecule(), primary_, grid_final_int_options, grid_final_str_options, grid_final_float_options, options_);
 
-    // Sanity-check of grids to ensure no negative grid weights
-    // COSX crashes when grids with negative weights are used,
+    // Print out warning if grid with negative grid weights is used
```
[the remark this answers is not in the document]
davpoolechem: Yeah, it's probably not necessary, honestly. I'll get rid of it.

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: Geh, nice catch. Thank you!

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: I'm happy to switch to INFO, but Psi4 seems to trend towards using WARNING instead of INFO for printout like this, at least from my observation.

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: Yeah, on second glance, you're right, it does return an `int`. We can definitely clear up the return type; and likely, we want it to return a `double`, to boot.

These goals could be achieved pretty easily using one of two methods:
1. Place a `static_cast<double>` around the returned value in the lambda function.
2. Simply replace the current lambda function implementation with one using a ternary operator.

I will fix this up with idea #1 to start, as the entire current implementation avoids calling the equivalent of an `if` statement for each function call; but if people prefer using a ternary operator for clarity, I'll be happy to do that, as well.

susilehtola: Yes a ternary would be preferable

davpoolechem: All right, I'll use a ternary instead, then!

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: Added! I'll leave the conversation open, though, in case others want to comment on INFO vs. WARNING.

psi4/src/psi4/libfock/DFJCOSK.cc
davpoolechem: Done and done!

davpoolechem:
> Looks pretty good, please fix the issues above

Thanks! I think everything you suggested should be covered now.

psi4/src/psi4/libfock/DFJCOSK.cc
loriab: > I think this should not be a WARNING but INFO (does this exist in Psi4?)

There's no proper logging from c-side. :-( Only `PSIEXCEPTION` that stops execution and `Printf("WARNING/INFO/ETC` that goes to the output file. Admittedly not great. I don't think the situation has been seriously examined for improvement since pybind11 (and having python on the outside of psi4) was young.

> in case others want to comment on INFO vs. WARNING.

I agree it's probably INFO at this point (rounding the grid up to down to a pure-positive one would be WARNING), but there's not standardization.

davpoolechem: > > I think this should not be a WARNING but INFO (does this exist in Psi4?)
> 
> There's no proper logging from c-side. :-( Only `PSIEXCEPTION` that stops execution and `Printf("WARNING/INFO/ETC` that goes to the output file. Admittedly not great. I don't think the situation has been seriously examined for improvement since pybind11 (and having python on the outside of psi4) was young.
> 
> > in case others want to comment on INFO vs. WARNING.
> 
> I agree it's probably INFO at this point (rounding the grid up to down to a pure-positive one would be WARNING), but there's not standardization.

All good to know! Maybe such logging is something that could be more seriously looked at in the future. It wouldn't be a bad functionality to have.

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -852,16 +876,21 @@ void DFJCOSK::build_K(std::vector<std::shared_ptr<Matrix>>& D, std::vector<std::
...
         auto point_values = bf_computers[rank]->basis_values()["PHI"];
 
         // resize the buffer of basis function values
         auto X_block = std::make_shared<Matrix>(npoints_block, nbf_block);  // points x nbf_block
+
         auto X_blockp = X_block->pointer();
         for (size_t p = 0; p < npoints_block; p++) {
             for (size_t k = 0; k < nbf_block; k++) {
-                X_blockp[p][k] = point_values->get(p, k) * std::sqrt(w[p]);
+                X_blockp[p][k] = point_values->get(p, k) * std::sqrt(std::fabs(w[p]));
```
susilehtola, line 893:
```suggestion
                X_blockp[p][k] = point_values->get(p, k) * std::sqrt(std::abs(w[p]));
```

psi4/src/psi4/libfock/DFJCOSK.cc
```diff
@@ -73,19 +80,28 @@ Matrix compute_numeric_overlap(const DFTGrid &grid, const std::shared_ptr<BasisS
...
-        auto X_blockp = X_block.pointer();
+        Matrix X_block_nosign("phi_g,u", npoints_block, nbf_block);  // points x nbf_block
+        Matrix X_block_sign("phi_g,u", npoints_block, nbf_block);  // points x nbf_block
+
+        auto X_block_nosignp = X_block_nosign.pointer();
+        auto X_block_signp = X_block_sign.pointer();
         for (size_t p = 0; p < npoints_block; p++) {
             for (size_t k = 0; k < nbf_block; k++) {
-                X_blockp[p][k] = point_values->get(p, k) * std::sqrt(w[p]);
+                X_block_nosignp[p][k] = point_values->get(p, k) * std::sqrt(std::fabs(w[p]));
```
susilehtola, line 96:
```suggestion
                X_block_nosignp[p][k] = point_values->get(p, k) * std::sqrt(std::abs(w[p]));
```

susilehtola (approved):

susilehtola:
I'm good with the way of computing the overlap matrix. One does want to use BLAS and this just means that you need an extra copy of the basis function value matrix.

davpoolechem:
> I'm good with the way of computing the overlap matrix. One does want to use BLAS and this just means that you need an extra copy of the basis function value matrix.

Yeah, I was thinking it would be preferable to use BLAS, as well.

davpoolechem:
So with https://github.com/psi4/psi4/pull/2955 (SplitJK) also being a thing, I think it would be preferable to get that one in first, and then rebase this one to adjust for the changes.

jturney (approved):

PR #2932, opened 2023-04-26
set_output_file enhancements

### LINKED ISSUE #2893: Add a way to produce a full output file when using PsiAPI
I'd like to be able to use Psi4 in a Jupyter notebook, but part of my workflow involves feeding the output into `cclib` for conversion to CJSON. Unfortunately, the truncated output from PsiAPI can't be parsed by the `cclib` parsers. Is there (or can there be) a way to toggle a "full" output using `psi4.set_output_file()`?

loriab:

- [x] If you call `psi4.set_output_file()` (note that this is the usual fancy one, not the low-level `psi4.core.set_output_file()`), that file will have a psi4 header so suitable for cclib parsing. closes #2893
- [x] Sometimes you don't want psi4 managing the logging. Added a `inherit_loglevel` to exe and set_output_file so that psi4 doesn't tamper with loglevel. @bennybp request

## Dev notes & details

- [x] we always had `set_output_file` and `print_header` separate, but I'm guessing that that was originally b/c former in C++ and latter in Py. Now that we have the former in python to set up logging, too, it's reasonable to combine them, I think. Note that in exe, we do need to set up threads & mem first so that header stats are as correct as possible.
- [x] I set it up so that the header prints whenever a new file is opened. seemed reasonable. had to rewire core.cc calls to take notice of "stdout". Any problem with that?
- [x] `-o stdout` now gets a logging file `stdout.log` bug? feature?
- [x] a plain `import psi4; psi4.energy()` is still going to print to screen w/o header. `import psi4; set_output_file("asdf"); psi4.energy()` newly has a header. DDD calcs still have repetitive header, but it's no worse that before.

## Questions
- [x] @AlexHeide, does removing those two `root` lines in `psi4.__init__.py` mess up your optking logging?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/__init__.py
+++ b/psi4/__init__.py
@@ -102,6 +102,4 @@ if "@ENABLE_libefp@".upper() in ["1", "ON", "YES", "TRUE", "Y"]:  # pylibefp
 
 # Create a custom logger
 import logging
-root = logging.getLogger()  # create root
-root.setLevel("ERROR")
 logger = logging.getLogger(__name__)  # create initial psi4 from root to be configured later in extras
--- a/psi4/extras.py
+++ b/psi4/extras.py
@@ -30,12 +30,13 @@ import atexit
 import datetime
 import itertools
 import os
-from typing import List, Union
+from typing import List, Optional, Union
 from pathlib import Path
 
 from qcelemental.util import which, which_import
 
 from . import core
+from .header import print_header as _print_header
 
 # Numpy place holder for files and cleanup
 numpy_files = []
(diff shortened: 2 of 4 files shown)


AlexHeide:
I don't see any issues cropping up. I should make some documentation updates though to inform users how `optking` inherits logging from `psi4`. Just that users should make sure to import psi4 before optking for optking to inherit logging from psi4.

loriab:
> I don't see any issues cropping up. I should make some documentation updates though to inform users how optking inherits logging from psi4. Just that users should make sure to import psi4 before optking for optking to inherit logging from psi4.

Great, thanks for checking, @AlexHeide !

jturney (approved):

davpoolechem (approved):

PR #2933, opened 2023-04-28
Begone, ERISieve!

davpoolechem:
## Description
This PR is something of an independent project, although it is also a precursor of the continuation of https://github.com/psi4/psi4/pull/2695, as agreed upon by discussion there.

This PR completely eliminates the `ERISieve` class, and replaces all of its old functionalities with that of `TwoBodyAOInt`. `ERISieve` is a class in libmints that does many of the same things as `TwoBodyAOInt`. The amount of features that `ERISieve` has, that `TwoBodyAOInt` doesn't, is minimal. It's really just absolute redundancy. Currently, `ERISieve` _only_ finds use within the PKJK class and its subalgorithms.

As mentioned, this PR eliminates `ERISieve` and replaces it with `TwoBodyAOInt` in all of the former's previous use cases. Any previously-utilized functionality that was present in `ERISieve` and not in `TwoBodyAOInt` (a couple functions determining basis function significance) were added to `TwoBodyAOInt`. Furthermore, any adjustments needed to support `TwoBodyAOInt` within the PKJK code were made.

With all of this, `ERISieve` was also ripped out of the code entirely. Maybe it could go into the attic.

- [X] The core.ERISieve class has been removed, along with its associated functions core.ERISieve.build and core.ERISieve.shell_significant. The removed functionalities, specifically shell_significant, can be accessed via the construction of a TwoBodyAOInt object.

## Dev notes & details
- [X] The `ERISieve` class in libmints has been removed from Psi4 and replaced with `TwoBodyAOInt`.
- [X] Functionalities in `ERISieve` that were not in `TwoBodyAOInt` and were required in the code, were added to `TwoBodyAOInt`.
- [X] The PKJK files have been adjusted as necessary to support the use of `TwoBodyAOInt`.

## TODO
- [ ] Remove ERISieve from all other external plugins (currently focusing on v2rdm_casscf).

## Questions
- [ ] I guess `ERISieve` can go into the attic now?

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge


psi4/src/psi4/libmints/twobody.h
```diff
@@ -215,6 +215,14 @@ class PSI_API TwoBodyAOInt {
      inline double shell_ceiling2(int M, int N, int R, int S) {
         return shell_pair_values_[N * nshell_ + M] * shell_pair_values_[R * nshell_ + S];
     }
+    /// Is the function pair (mn| ever significant according to sieve (no restriction on mn order)
+    inline bool function_pair_significant(int m, int n) {
```
TiborGY, line 219:
```suggestion
    inline bool function_pair_significant(const int m, const int n) {
```

psi4/src/psi4/libmints/twobody.h
```diff
@@ -215,6 +215,14 @@ class PSI_API TwoBodyAOInt {
...
+        return function_pair_values_[m * nbf_ + n] * max_integral_ >= screening_threshold_squared_;
+    }
+    /// Is the integral (mn|rs) significant according to sieve? (no restriction on mnrs order)
+    inline bool function_significant(int m, int n, int r, int s) {
```
TiborGY, line 223:
```suggestion
    inline bool function_significant(const int m, const int n, const int r, const int s) {
```

TiborGY (commented):
Just a formality in this case, but marking arguments that are not modified in the function `const` is good practice.

davpoolechem:
Ah, so it seems I made a mistake in my assumptions somewhere. The failed CI tells me this:

`/home/runner/work/psi4/psi4/objdir/external/downstream/v2rdm_casscf/v2rdm_casscf_external-prefix/src/v2rdm_casscf_external/threeindexintegrals.cc:33:10: fatal error: psi4/libmints/sieve.h: No such file or directory
 #include <psi4/libmints/sieve.h>`
 
 It seems that PKJK is, in fact, _not_ the only place that uses ERISieve. This v2rdm_casscf plugin does, as well.

loriab:
v2rdm strikes again! :-)

fwiw, the original author (DePrince) isn't dedicated to preserving the plugin, but it's handy for catching things that may affect downstream users. Right now we're pulling from https://github.com/loriab/v2rdm_casscf/tree/v2rdm8, so if you want to PR a switchout of ERISeive to there for post-1.8, that'd be fine. In any case, it might be worth getting in a deprecation warning now for v1.8 on ERISeive.

davpoolechem:
> v2rdm strikes again! :-)
> 
> fwiw, the original author (DePrince) isn't dedicated to preserving the plugin, but it's handy for catching things that may affect downstream users. Right now we're pulling from https://github.com/loriab/v2rdm_casscf/tree/v2rdm8, so if you want to PR a switchout of ERISeive to there for post-1.8, that'd be fine. In any case, it might be worth getting in a deprecation warning now for v1.8 on ERISeive.

Thank you for the heads-up on this! I'll work on getting rid of ERISieve there, as well, and deprecate ERISieve here for 1.8.

davpoolechem:
I just put up https://github.com/psi4/psi4/pull/2935 to deprecate `ERISieve`. That one should go in before this one.

davpoolechem:
I officially opened up a PR for v2rdm_casscf, at https://github.com/loriab/v2rdm_casscf/pull/5, which removes ERISieve from v2rdm_casscf and replaces it with TwoBodyAOInt. Once that is added in, we can progress with this PR, as that is the current bottleneck preventing this PR from passing CI testing. Ideally, there's no fun surprises after that.

davpoolechem:
All right, the necessary v2rdm_casscf changes have been made and incorporated into this PR.

psi4/driver/p4util/python_helpers.py
```diff
@@ -1554,43 +1553,3 @@ def _core_triplet(A, B, C, transA, transB, transC):
...
-    -------
-    >>> sieve = psi4.core.ERISieve.build(bas, cutoff, csam)
-    """
-
-    warnings.warn(
-        "`ERISieve` is deprecated in favor of `TwoBodyAOInt`, and will be removed as soon as Psi4 v1.9 is released.\n",
-        category=FutureWarning,
-        stacklevel=2)
-
-    return core.ERISieve(orbital_basis, cutoff, do_csam)
```
loriab, line 1593: you could raise an UpgradeHelper build-48s/src/lib/libint/library-prefix/src/library-build/src/configuration.cc saying how to replace this instead of deleting completely. that is, instead of warning and continuing, it stops with advice. up to you.

loriab: helps if I send a link to the code rather than a path on my local machine :-)

https://github.com/psi4/psi4/blob/819c588cb7cae7d86637b7727fcd6f5899473b3f/psi4/driver/p4util/python_helpers.py#L824

davpoolechem: I think this is a pretty good idea! I will give it a look.

davpoolechem: So I do have a version of the code that raises the `UpgradeHelper` for calling `core.ERISieve.build` instead of deleting `ERISieve` completely, at https://github.com/davpoolechem/psi4/tree/dpoole34/erisieve-upgrade-helper. That is probably better from an API maintenance approach than simply deleting `ERISieve`, but we would probably want to close down this PR and push the other branch instead if we wanted the `UpgradeHelper` approach.

davpoolechem: I'm thinking a new PR would be easier, so I will likely do that.

davpoolechem:
Closing this one up in favor of https://github.com/psi4/psi4/pull/2974.

PR #2934, opened 2023-04-28
Fix fsapt + external potential - nocom/noreorient

loriab:
## Description

@carolinesargent identified a bug where running FSAPT with an external potential w/o having frozen the orientation with no_com + no_reorient would run but give the wrong answer. :-(

External potentials has long been one of those cases where we required the user to freeze the orientation at molecule creation time so that the potential could be set in the same frame. This couldn't be fixed driver-side because as soon as the `core.Mol` builds w/o freeze directives, it loses the original Cartesian coordinates. (The clone, set_nocom, set_noreorient calls in the driver allow _regular_ sapt to forego user setting by preventing the dimer, monoA, monoB from having different frames.)

Happily, in the intervening period, @maxscheurer ran into exactly this problem for polarizable embedding potentials and solved it by tacking a copy of the original Cartesians onto the molecule. So we're applying this to FSAPT also. 

I've been getting some segfaults that I think are a quirk of my directory, hence the cc31 testing.

- [x] Fixes bug where FSAPT with an external potential and without no_com/no_reorient set would return wrong answer.

## Dev notes & details

- [x] Let's use this route to accommodate aux info in the frame of the Cartesianmol

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/mdi_engine.py
+++ b/psi4/driver/mdi_engine.py
@@ -81,7 +81,10 @@ class MDIEngine():
 
         # Molecule all MDI operations are performed on
         input_molecule = kwargs.pop('molecule', psi4.core.get_active_molecule())
+        _ini_cart = getattr(input_molecule, "_initial_cartesian", None)
         self.molecule = input_molecule.clone()
+        if _ini_cart:
+            self.molecule._initial_cartesian = _ini_cart
         psi4.core.set_active_molecule(self.molecule)
 
         # Most recent SCF energy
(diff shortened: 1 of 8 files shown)


loriab:
@taylor-a-barnes, just a heads-up to this small change to MDIEngine. It'll be in v1.8. I don't anticipate any implications for MDI.

jturney (approved):

davpoolechem (approved):

PR #2935, opened 2023-04-28
Deprecate ERISieve

davpoolechem:
## Description
This PR is a companion to https://github.com/psi4/psi4/pull/2933. The goal of https://github.com/psi4/psi4/pull/2933 is to completely eliminate the `ERISieve` class, for reasons explained there. However, it turns out that eliminating `ERISieve` requires fiddling with the v2rdm_casscf plugin, as well, since it also uses `ERISieve`.

This PR is meant to be a deprecation of the `ERISieve` class that will show up in v1.8 in the meantime, in case `ERISieve` isn't fully removed by then.

- [X] Deprecates the `ERISieve` class.

## Dev notes & details
- [X] Fully deprecates the `ERISieve` class C++-side by tagging it with the deprecated attribute.
- [X] Deprecates the `ERISieve` class Python-side by creating helper functions for the Python-facing `ERISieve` functions that warn about deprecation.

## Questions
- [x] Is `PSI_DEPRECATED` the preferred mechanism by which to perform this deprecation?
- [X] Assuming the answer to the first question is "yes", are there any other functions in `ERISieve` that should be tagged? I chose specifically the user-facing functions.
- [x] I _loathe_ having to use a global variable for the Python-side helper functions. But, unbelievably, it is what I considered the best one I could think of. I would be very happy to hear out alternative suggestions.
- [ ] Using the `PSI_API` and `PSI_DEPRECATED` macros together to tag the `ERISieve` class cause the compiler to complain. On further exploration, replacing `PSI_DEPRECATED` with its textual definition seems to work fine. Is this an acceptable solution?

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -1502,8 +1502,8 @@ def _core_wavefunction_frequencies(self):
     """Returns the results of a frequency analysis.
 
     Parameters
-    ----------
-    self
+    ------G----
+    selfG
         Wavefunction instance.
 
     Returns
@@ -1590,4 +1590,17 @@ def _core_erisieve_build(
 
     return core.ERISieve(orbital_basis, cutoff, do_csam)
 
+def _core_erisieve_shell_significant(self, M, N, R, S):
+    """
+    Determine if a given shell quartet (MN | RS) is sigificant or not. Return True if it is, and False if it is not.
+    """
+
+    warnings.warn(
+        "`ERISieve` is deprecated in favor of `TwoBodyAOInt`, and will be removed as soon as Psi4 v1.9 is released.\n",
+        category=FutureWarning,
+        stacklevel=2)
+
+    return core.ERISieve.shell_significant(M, N, R, S)
+
 core.ERISieve.build = _core_erisieve_build
+core.ERISieve.shell_significant = _core_erisieve_shell_significant
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -1588,7 +1588,12 @@ def _core_erisieve_build(
         category=FutureWarning,
         stacklevel=2)
 
-    return core.ERISieve(orbital_basis, cutoff, do_csam)
+    factory = core.IntegralFactory(orbital_basis)
+    
+    global erisieve_twobody 
+    erisieve_twobody = factory.eri(0)
+
+    return core.ERISieve(orbital_basis, cutoff, do_csam) 
 
 def _core_erisieve_shell_significant(self, M, N, R, S):
     """
@@ -1599,8 +1604,8 @@ def _core_erisieve_shell_significant(self, M, N, R, S):
         "`ERISieve` is deprecated in favor of `TwoBodyAOInt`, and will be removed as soon as Psi4 v1.9 is released.\n",
         category=FutureWarning,
         stacklevel=2)
-
-    return core.ERISieve.shell_significant(M, N, R, S)
+   
+    return erisieve_twobody.shell_significant(M, N, R, S)
 
 core.ERISieve.build = _core_erisieve_build
 core.ERISieve.shell_significant = _core_erisieve_shell_significant
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -37,7 +37,7 @@ Also, many Python extensions to core classes:
  - JK (constructor)
  - VBase (grid)
  - OEProp (avail prop)
-
+ - ERISieve (constructor, shell_significant)
 """
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


TiborGY:
It looks like the entire `ERISieve` class is marked as `PSI_API`, so AFAIK that is giving downstream the permission to rely on both the object and all of its public members being a stable API, not just a few functions.

loriab:
Yes, it would have to be `PSI_API` for v2rdm to have used it at all. Those are the only fns to escape pb11's settings to hide all symbols. I think it's still ok to replace medium term, but I agree with @TiborGY that should blanket deprecate. And yes, `PSI_DEPRECATED` is the way to go. It complains at build-time.

davpoolechem:
All right, blanket deprecated ERISieve!

... but there's a catch. It seems the compiler complains when trying to use `PSI_API` and `PSI_DEPRECATE("message")` together on the same class. I essentially had to hand-insert the associated macros to get the code to compile.

loriab:
Ok, two different levels here:
* v2rdm (and potentially other plugins or codes that link to psi4 (not that I know any in the latter case)) are using `ERISeive` c-side. They can do so b/c it's `PSI_API`d so the symbols are exposed in `psi4/core.*so`. These should be deprecated by `PSI_DEPRECATED`. If the plugin owner is compiling his plugin but never psi4 itself, he might still miss the notice, but at least we're giving such a fair chance.
* anyone could be using `ERISeive` py-side if it's exported by pb11. (And this has nothing to do with whether it's `PSI_API`d.) I think the way to deprecate those is to wrap the fns like https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L1480-L1495

Whenever dealing with deprecations, I think it's a good idea to use the "_as soon as_ v1.x" phrasing. We want to get the urgency across, but we almost never clear things out ASAP so that phrasing keeps the message accurate.

For semi-completeness, I bring up a couple other deprecation/removal items:
* When you've got to make a breaking change py-side or you want to provide some guidance even after removal, there's `UpgradeHelper`s https://github.com/psi4/psi4/blob/master/psi4/driver/driver_cbs.py#L1062 that stop the calc but provide advice for fixing the input.
* When you want to remove an option, there's a pattern in core.cc
* When you want to remove or change the meaning of a QCVariable, there's a list in python_helpers.py

davpoolechem:
> Ok, two different levels here:

First off, thank you very much for sharing some details on all of this!
 
> * v2rdm (and potentially other plugins or codes that link to psi4 (not that I know any in the latter case)) are using `ERISeive` c-side. They can do so b/c it's `PSI_API`d so the symbols are exposed in `psi4/core.*so`. These should be deprecated by `PSI_DEPRECATED`. If the plugin owner is compiling his plugin but never psi4 itself, he might still miss the notice, but at least we're giving such a fair chance.

That make sense. I recently blanket-deprecated the `ERISieve` class with (the equivalent of) `PSI_DEPRECATED`. The issue I saw, was that using both `PSI_API` and `PSI_DEPRECATED` for the same class caused the compiler to complain, although using their textual equivalents seem to work fine. Hopefully, that settles the deprecation C-side.

> * anyone could be using `ERISeive` py-side if it's exported by pb11. (And this has nothing to do with whether it's `PSI_API`d.) I think the way to deprecate those is to wrap the fns like https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L1480-L1495

Also very good to know! I think I will switch this to a draft PR in the meantime, since it seems some more needs to be done on my end. 

> 
> Whenever dealing with deprecations, I think it's a good idea to use the "_as soon as_ v1.x" phrasing. We want to get the urgency across, but we almost never clear things out ASAP so that phrasing keeps the message accurate.
> 
> For semi-completeness, I bring up a couple other deprecation/removal items:
> 
> * When you've got to make a breaking change py-side or you want to provide some guidance even after removal, there's `UpgradeHelper`s https://github.com/psi4/psi4/blob/master/psi4/driver/driver_cbs.py#L1062 that stop the calc but provide advice for fixing the input.
> * When you want to remove an option, there's a pattern in core.cc
> * When you want to remove or change the meaning of a QCVariable, there's a list in python_helpers.py

TiborGY:
> The issue I saw, was that using both `PSI_API` and `PSI_DEPRECATED` for the same class caused the compiler to complain, although using their textual equivalents seem to work fine. Hopefully, that settles the deprecation C-side.

It is strange that this is happening to you, I had no such trouble when deprecating `PSI_API` functions, see https://github.com/psi4/psi4/pull/2798/commits/e11f826e4b5fe3fc96c2a213bab0a0c4f1ad051c

davpoolechem:
> > The issue I saw, was that using both `PSI_API` and `PSI_DEPRECATED` for the same class caused the compiler to complain, although using their textual equivalents seem to work fine. Hopefully, that settles the deprecation C-side.
> 
> It is strange that this is happening to you, I had no such trouble when deprecating `PSI_API` functions, see [e11f826](https://github.com/psi4/psi4/commit/e11f826e4b5fe3fc96c2a213bab0a0c4f1ad051c)

Deprecating a PSI_API _function_ works fine, which I verified myself. Deprecating a PSI_API _class_ is where I find issues.

Something like:

``class PSI_API PSI_DEPRECATED("msg") ERISieve {``

causes the compiler to complain. That said, replacing the macros with their associated definitions, something like

``class __attribute__((visibility("default"))) __attribute__((deprecated("msg"))) ERISieve { ``

works just fine.

davpoolechem:
Okay, I opened this up for official review again, since I think I covered all of the bases mentioned by @loriab. I am not confident in every design decision here, so don't be afraid to suggest alternatives if you have them.

psi4/driver/p4util/python_helpers.py
```diff
@@ -1553,3 +1553,62 @@ def _core_triplet(A, B, C, transA, transB, transC):
...
+
+    warnings.warn(
+        "`ERISieve` is deprecated in favor of `TwoBodyAOInt`, and will be removed as soon as Psi4 v1.9 is released.\n",
+        category=FutureWarning,
+        stacklevel=2)
+
+    return erisieve_twobody.shell_significant(M, N, R, S)
+
+
+core.ERISieve.build = _core_erisieve_build
```
loriab, line 1613: Probably only the constructor is needed since it must happen before shell_significant can be run, right? But both is fine. You've built one to confirm the warning and hookups?

Btw, I suspect the mac ecosys is caught in the qcf hanging that I introduced yesterday but fixed as of #2946 merge. I'll cancel it, but that shouldn't count against the PR.

loriab: Ok, I think I'm for removing `shell_significant`, then. (Thanks for the thoroughness in including it, though.) And thanks for the testing. After that, lgtm!

davpoolechem: > Ok, I think I'm for removing `shell_significant`, then. (Thanks for the thoroughness in including it, though.) And thanks for the testing. After that, lgtm!

Yeah, I was very unhappy with needing a global variable to make it work, so I'm for removing it, as well. I will do so. Thanks!

loriab: Oh, and either rebase or kill off the mac ecosys, please, to avert wasting CI.

loriab (approved):

JonathonMisiewicz (approved):
I'm not confident in all the design decisions either, but this is good enough, and I have nothing better to propose. Good riddance.

PR #2938, opened 2023-05-01
temp pin adcc

loriab:
## Description
At the moment, psi4 + adcc 0.15.14 is passing but 0.15.16 is not. @maxscheurer suspects a bugfix in adcc that the psi4 ref data needs to be updated to reflect. Pinning this for now so other PRs don't need to be held up.

## Status
- [x] Ready for review
- [ ] Ready for merge


maxscheurer (approved):

davpoolechem (approved):

maxscheurer:
Should we close this PR now that we have #2944?

PR #2947, opened 2023-05-01
add third-party-licenses

loriab:
## Description

c-f required something like this for PCMSolver, so I went ahead with it for Psi4 also. None present a problem with overall LGPL-3.0-only as far as I can see. Some of the folks we're still in contact with and could ask to dual-license, if there was a problem.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/THIRD-PARTY-LICENSES
+++ b/THIRD-PARTY-LICENSES
@@ -1,0 +1,437 @@
+This project incorporates material from the project(s) listed below
+(collectively, "Third Party Code"). Psi4 is not the original author
+of the Third Party Code. The original copyright notice and license under
+which Psi4 received such Third Party Code are set out below. This
+Third Party Code is licensed to you under their original license terms
+set forth below.
+
+
+<< 1 >> psi4/src/psi4/libmints/corrtab.h & corrtab.cc
+        psi4/src/psi4/libmints/petitelist.cc
+        psi4/src/psi4/libmints/maketab.cc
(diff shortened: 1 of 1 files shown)


jturney (approved):

davpoolechem (approved):
I think it looks good! 

As for compatibility, the third-party licenses I saw were BSD, MIT, and GPL2+3. While GPL2 is incompatible with LGPL3 as per [FSF](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility), the phrasing of the GPL2-licensed codes here allow for substitution of later GPL versions which _are_ compatible with LGPL3. So I think we are good on the compatibility front.

davpoolechem:
> > As for compatibility, the third-party licenses I saw were BSD, MIT, and GPL2+3. While GPL2 is incompatible with LGPL3 as per [FSF](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility), the phrasing of the GPL2-licensed codes here allow for substitution of later GPL versions which are compatible with LGPL3. So I think we are good on the compatibility front.
> 
> For GNU, I only saw "Lesser" and "Library" licences, both of which are LGPL. Having GPL code itself _would_ be a problem. Did you or I misread?

Ah, if "Library" is a reference to LGPL, then we should be good. I took "Library" as referring to the regular GPL, and "Lesser" as referring to LGPL. The more you know!

loriab:
https://opensource.org/license/lgpl-license-html/

yeah "General" is the key word for regular GPL

PR #2948, opened 2023-05-02
modernize paths and bugfixes

loriab:
## Description

~Note: this isn't going to pass until after #2946~

## Dev notes & details
- [x] We had the odd situation where the install directories (GNUInstallDirs) were pretty flexible. But it wasn't quite flexible enough for the usual Windows install layout (see https://github.com/psi4/psi4/blob/master/conda/win/meta.yaml#L32-L41). And if you got those all in the right place, the paths computation in `run_psi4.py` (-> `bin/psi4`) and `psi4/__init__.py` wasn't quite as flexible as GNUInstallDirs allowed. So I modernized the lot with pathlib and hopefully systemitized it a bit, too.
- [x] There's a weird case on linux c-f where we're in a git repo, but it's not the psi git repo. Thus versioning is much confused. The versioner can now escape from that case and version as if tarball.

## Checklist
- [x] works for lin, mac, win conda-builds

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/cmake/FindPythonModule.cmake
+++ b/cmake/FindPythonModule.cmake
@@ -6,6 +6,7 @@
 # * Remove ``if(NOT PY_${module})`` so runs each time, also remove module caps
 # * Module name, not path in fphsa
 # * Added version handling via parse_version call
+# * Switched PYTHON -> Python
 
 #.rst:
 #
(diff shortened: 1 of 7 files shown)


JonathonMisiewicz (approved):
Rubber-stamped.

PR #2952, opened 2023-05-05
Fix tests to reflect pyddx 0.4

### LINKED ISSUE #2940: update to pyddx 0.4
Reminder to finish pinning to 0.3 or resolve tests to allow 0.4 before v1.8. Just making an issue so it's tracked. @mfherbst

mfherbst:
Fixes #2940. Closes #2921

@mnottoli  Could you quickly give this a second pair of eyes to check what I'm doing [here when I'm lowering the tolerance](https://github.com/psi4/psi4/blob/6173fba2dc820bba05918f3dfa599550704de4f5/tests/pytests/test_ddx.py#L112) is reasonable --- I'm thinking this is the escaped charge problem for this small system. Maybe we should just remove the test alltogether?

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -230,7 +230,7 @@ jobs:
           - geometric
           - openfermion>=1.0
           - openfermionpsi4
-          - pyddx=0.3.0
+          - pyddx>=0.4
           - pymdi
           - qcengine
           - qcfractal
--- a/external/upstream/ddx/CMakeLists.txt
+++ b/external/upstream/ddx/CMakeLists.txt
@@ -1,7 +1,7 @@
 if(${ENABLE_ddx})
     if(NOT (${CMAKE_DISABLE_FIND_PACKAGE_ddx}))
         include(FindPythonModule)
-        find_python_module(pyddx ATLEAST 0.3.0 QUIET)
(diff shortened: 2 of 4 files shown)


maxscheurer (approved):

loriab (approved):
Thanks!

We'll hold off on merging until you hear back from your colleague and give the ok.

mnottoli:
Since ddLPB has the electric field as RHS, it may be that the error in the numerical derivatives (numerical Fock) is higher than in ddPCM/ddCOSMO. To be sure, you could try to lower the step size, you should observe a decreasing error in the analytical-numerical Fock comparison (at least until it hits the finite numerical precision), if it is not like that, there may be a problem.

loriab:
This is the variable to vary? https://github.com/ddsolvation/ddX/blob/main/tests/matrix_derivatives.f90#L129

I'm having problems getting an editable pyddx build working with psi, so I'm inclined to merge this as-is. This PR against c-f 0.4.1 is clean against psi4 current master.

mnottoli:
Sorry for the late reply.
I meant this: https://github.com/psi4/psi4/blob/6173fba2dc820bba05918f3dfa599550704de4f5/tests/pytests/test_ddx.py#LL76C48-L76C51

In any case the pull request looked good for me.

PR #2953, opened 2023-05-06
Renovate Dimension object

TiborGY:
## Description

This PR aims to improve the Dimension object in various ways, by adding machine-readable (for eg. VSCode) docstrings, adding const to arguments, replacing `int` with `size_t` to match `std::vector::operator[](std::size_t)` and cautiously deprecating member functions that deal with pointers.

Not only are pointers more error-prone, they are currently involved in some strange interaction between `Dimension` and `Matrix` objects, resulting in awkward roundabout initialization/construction that I hope to untangle before Psi4 1.11? (1.11, or maybe 1.12?).
Please let me know if you feel the three deprecations in this PR would be too disruptive or otherwise undesirable, this PR is very much a "request for comments".

- [x] API change: The custom assignment operator for `Dimension` objects (`Dimension& operator=(const int*)`) is being deprecated. Unless someone speaks up, 1.10 may be the last release to have it.
- [x] API change: Cast-to-pointer operators for `Dimension` objects (`operator int*()` and `operator const int*() const`) are being deprecated. Unless someone speaks up, 1.10 may be the last release to have them.
- [x] Minor API change: Several constructors and member functions of `Dimension` are now using `size_t` instead of `int` for indexing:
−`Dimension::Dimension(int, const std::string&)` is now `Dimension::Dimension(size_t, const std::string&)`
−`void Dimension::init(int, const std::string&)` is now `void Dimension::init(size_t, const std::string&)`
−`int Dimension::n() const` is now `size_t Dimension::n() const`
−`int& Dimension::operator[](int)` is now `int& Dimension::operator[](size_t)`
−`const int& Dimension::operator[](int) const` is now `const int& Dimension::operator[](size_t) const`
−`const int& Dimension::get(int) const` is now `const int& Dimension::get(size_t) const`
−`void Dimension::set(int, int)` is now `void Dimension::set(size_t, int)`

## Dev notes & details

- [x] Machine-readable docstrings have been added to `dimension.h` to improve suggestions offered by IDEs like VSCode
- [x] `size_t` is now used instead of `int` when dealing with array indexing. Python bindings have been updated to reflect the change in constructor arguments.
- [x] Local variables have been made `const` where possible
- [x] Deprecation notices have been added to `Dimension& operator=(int*)`, `operator int*()` and `operator const int*() const`

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/dimension.cc
+++ b/psi4/src/psi4/libmints/dimension.cc
@@ -61,8 +61,7 @@ void Dimension::print() const {
 }
 
 Dimension& Dimension::operator=(const int* other) {
-    std::copy(other, other + n(), blocks_);
-
+    std::copy(other, other + n(), blocks_.begin());
     return *this;
 }
 
--- a/psi4/src/psi4/libmints/dimension.cc
+++ b/psi4/src/psi4/libmints/dimension.cc
@@ -61,14 +61,14 @@ void Dimension::print() const {
 }
 
 Dimension& Dimension::operator=(const int* other) {
-    for (int i = 0, maxi = n(); i < maxi; ++i) blocks_[i] = other[i];
+    for (size_t i = 0, maxi = n(); i < maxi; ++i) blocks_[i] = other[i];
 
     return *this;
 }
 
 Dimension& Dimension::operator+=(const Dimension& b) {
     if (n() == b.n()) {
-        for (int i = 0, maxi = n(); i < maxi; ++i) blocks_[i] += b.blocks_[i];
+        for (size_t i = 0, maxi = n(); i < maxi; ++i) blocks_[i] += b.blocks_[i];
     } else {
         std::string msg = "Dimension operator+=: adding operators of different size (" + std::to_string(n()) + " and " +
                           std::to_string(b.n()) + ")";
@@ -80,7 +80,7 @@ Dimension& Dimension::operator+=(const Dimension& b) {
 
 Dimension& Dimension::operator-=(const Dimension& b) {
     if (n() == b.n()) {
-        for (int i = 0, maxi = n(); i < maxi; ++i) blocks_[i] -= b.blocks_[i];
+        for (size_t i = 0, maxi = n(); i < maxi; ++i) blocks_[i] -= b.blocks_[i];
     } else {
         std::string msg = "Dimension operator-=: subtracting operators of different size (" + std::to_string(n()) +
                           " and " + std::to_string(b.n()) + ")";
@@ -96,7 +96,7 @@ PSI_API bool operator!=(const Dimension& a, const Dimension& b) { return !operat
 PSI_API Dimension operator+(const Dimension& a, const Dimension& b) {
     Dimension result = a;
     if (a.n() == b.n()) {
-        for (int i = 0, maxi = a.n(); i < maxi; ++i) result[i] += b[i];
+        for (size_t i = 0, maxi = a.n(); i < maxi; ++i) result[i] += b[i];
     } else {
         std::string msg = "Dimension operator+: adding operators of different size (" + std::to_string(a.n()) +
                           " and " + std::to_string(b.n()) + ")";
@@ -109,7 +109,7 @@ PSI_API Dimension operator+(const Dimension& a, const Dimension& b) {
 PSI_API Dimension operator-(const Dimension& a, const Dimension& b) {
     Dimension result = a;
     if (a.n() == b.n()) {
-        for (int i = 0, maxi = a.n(); i < maxi; ++i) result[i] -= b[i];
+        for (size_t i = 0, maxi = a.n(); i < maxi; ++i) result[i] -= b[i];
     } else {
         std::string msg = "Dimension operator-: subtracting operators of different size (" + std::to_string(a.n()) +
                           " and " + std::to_string(b.n()) + ")";
@@ -140,7 +140,7 @@ bool Slice::validate_slice() {
     }
 
     // Check that begin[h] >= 0 and end[h] >= begin[h]
-    for (int h = 0, max_h = begin_.n(); h < max_h; h++) {
+    for (size_t h = 0, max_h = begin_.n(); h < max_h; h++) {
         if (begin_[h] < 0) {
             valid = false;
             msg = "Invalid Slice: element " + std::to_string(h) + " of begin Dimension object is less than zero (" +
--- a/psi4/src/psi4/libmints/dimension.cc
+++ b/psi4/src/psi4/libmints/dimension.cc
@@ -40,11 +40,11 @@ namespace psi {
 
 Dimension::Dimension() : name_("(empty)") {}
 
-Dimension::Dimension(int n, const std::string& name) : name_(name), blocks_(n, 0) {}
+Dimension::Dimension(const size_t n, const std::string& name) : name_(name), blocks_(n, 0) {}
 
 Dimension::Dimension(const std::vector<int>& v) : blocks_(v) {}
 
-void Dimension::init(int n, const std::string& name) {
+void Dimension::init(const size_t n, const std::string& name) {
     name_ = name;
     blocks_.assign(n, 0);
 }
--- a/psi4/src/psi4/libmints/dimension.cc
+++ b/psi4/src/psi4/libmints/dimension.cc
@@ -61,7 +61,7 @@ void Dimension::print() const {
 }
 
 Dimension& Dimension::operator=(const int* other) {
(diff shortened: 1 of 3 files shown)


fevangelista:
I had to write some code recently using the Dimension object and I was thinking that it would be useful to support initialization from a list, for example:
```
Dimension orbs({3,0,1,1})
```
This is particularly useful when working with Slices to avoid doing this:
```
        auto start_dim = psi::Dimension(1);
        start_dim[0] = start;
```
I think it might already be supported on the python side. Another useful thing would be to have direct access to the underlying std::vector object, or const iterators for it. That way, one could use Dimensions in loops more naturally.

psi4/src/psi4/libmints/dimension.h
```diff
@@ -37,59 +37,88 @@
...
+    Dimension(const size_t n, const std::string& name = "");
+
+    /// @brief Constructs a Dimension object from an std::vector<int> object, leaving the name a zero-length string.
+    /// @param other : object to copy the block numbers from
     Dimension(const std::vector<int>& other);
 
-    /// Assignment operator, this one can be very dangerous
+    /// @brief Assignment operator, this one can be very dangerous
+    PSI_DEPRECATED(
+        "The assignment operator for psi::Dimension is being deprecated. Unless someone speaks up, 1.10 may be the "
```
loriab, line 63:
```suggestion
        "The assignment operator for psi::Dimension is being deprecated. Unless someone speaks up, v1.9 may be the "
```
1.9 instead? that would mean the fns could be removed before the May 2024 v1.10 release.

loriab (commented):
Looks reasonable to me, though I'm not a big user of `Dimension` and will defer to others. This would be a good time to get the deprecation warnings in, so the fns can be removed by 1.10. Are there any Python bindings that hit the deprecated fns? I think we could inject runtime depr warnings in python_helpers.py. The downside of the c-side warnings is that one only sees them when compiling.

TiborGY:
> Are there any Python bindings that hit the deprecated fns?

I don't think so? They are not used in `py::class_<Dimension>` in `export_mints.cc`, if there are other places to look for Python bindings let me know and I will check.

> This would be a good time to get the deprecation warnings in, so the fns can be removed by 1.10

Perhaps, but I delayed it because

1. I was not sure that this would get thorough reviews before 1.9 is released. This is something that could disrupt people's codes/plugins that rely on it. I do not know how widely they are used, thus I want to give people time to stumble into the deprecation warning and yell before it is too late.
2. These functions are not exactly unused, even in Psi4 itself. It will not be a trivial task to rip them out, and I doubt that I will have enough free time to complete it before May 2024.

loriab (approved):

psi4/src/psi4/libmints/dimension.cc
```diff
@@ -37,25 +37,20 @@
 #include "psi4/libpsi4util/exception.h"
 
 namespace psi {
-
 Dimension::Dimension() : name_("(empty)") {}
-
-Dimension::Dimension(int n, const std::string& name) : name_(name), blocks_(n, 0) {}
-
+Dimension::Dimension(const size_t n, const std::string& name/* = ""*/) : name_(name), blocks_(n, 0) {}
```
fevangelista, line 41: Here `const size_t` seems superfluous since n is copied and used only once. I'm in favor of using `const` as much as possible, especially to explicitly express that a variable should not be changed. But here we copy n and it's only used once.

psi4/src/psi4/libmints/dimension.cc
```diff
@@ -37,25 +37,20 @@
...
 Dimension::Dimension(const std::vector<int>& v) : blocks_(v) {}
 
-void Dimension::init(int n, const std::string& name) {
+void Dimension::init(const size_t n, const std::string& name/* = ""*/) {
```
fevangelista, line 44: Same point here about const size_t

psi4/src/psi4/libmints/dimension.cc
```diff
@@ -37,25 +37,20 @@
...
 }
 
 int Dimension::sum() const { return std::accumulate(blocks_.begin(), blocks_.end(), 0); }
-
 int Dimension::max() const { return *(std::max_element(blocks_.begin(), blocks_.end())); }
 
 void Dimension::zero() { std::fill(blocks_.begin(), blocks_.end(), 0); }
-
-void Dimension::fill(int v) { std::fill(blocks_.begin(), blocks_.end(), v); }
+void Dimension::fill(const int v) { std::fill(blocks_.begin(), blocks_.end(), v); }
```
fevangelista, line 53: Same here, remove `const`

psi4/src/psi4/libmints/dimension.cc
```diff
@@ -66,14 +61,14 @@ void Dimension::print() const {
 }
 
 Dimension& Dimension::operator=(const int* other) {
-    for (int i = 0, maxi = n(); i < maxi; ++i) blocks_[i] = other[i];
+    for (size_t i = 0, maxi = n(); i < maxi; ++i) blocks_[i] = other[i];
```
fevangelista, line 64: This shoud work:
```suggestion
    std::copy(other, other + n(), blocks_);
```

psi4/src/psi4/libmints/dimension.h
```diff
@@ -37,59 +37,88 @@
...
-    void fill(int v);
+    /// @brief Fill all elements in blocks_ with given value
+    void fill(const int v);
 
     void print() const;
 
     // Only used for python
-    const int& get(int i) const { return blocks_[i]; }
-    void set(int i, int val) { blocks_[i] = val; }
+    const int& get(const size_t i) const { return blocks_[i]; }
```
fevangelista, line 120: Remove `const` in `get` and `set`

psi4/src/psi4/libmints/dimension.h
```diff
@@ -37,59 +37,88 @@
...
     void set_name(const std::string& name) { name_ = name; }
 
-    /// Blocks access
-    int& operator[](int i) { return blocks_[i]; }
-    const int& operator[](int i) const { return blocks_[i]; }
+    /// @brief Access a block number at a partcular index. Not bounds-checked.
+    int& operator[](const size_t i) { return blocks_[i]; }
+
+    /// @brief Access a block number at a partcular index. Not bounds-checked.
+    const int& operator[](const size_t i) const { return blocks_[i]; }
```
fevangelista, line 88: remove `const`

psi4/src/psi4/libmints/dimension.h
```diff
@@ -37,59 +37,88 @@
...
 
-    /// Set the name of the dimension
+    /// @brief Set the name of the Dimension object
     void set_name(const std::string& name) { name_ = name; }
 
-    /// Blocks access
-    int& operator[](int i) { return blocks_[i]; }
-    const int& operator[](int i) const { return blocks_[i]; }
+    /// @brief Access a block number at a partcular index. Not bounds-checked.
+    int& operator[](const size_t i) { return blocks_[i]; }
```
fevangelista, line 85: remove `const`

fevangelista (changes requested):
Thanks for the changes. I requested some changes, some of which should be propagated to both the .h and .cc files.

psi4/src/psi4/libmints/dimension.h
TiborGY: Done

psi4/src/psi4/libmints/dimension.h
TiborGY: Done

psi4/src/psi4/libmints/dimension.cc
TiborGY: Done

psi4/src/psi4/libmints/dimension.cc
TiborGY: Done

psi4/src/psi4/libmints/dimension.cc
TiborGY: I have a habit of doing "const by default", but it is indeed making thing more verbose than necessary here.
Change done.

psi4/src/psi4/libmints/dimension.cc
TiborGY: Good idea, but as-is this leads to a build fail, for reasons I do not fully understand. My guess is that there is no suitable overload for handling `std::copy(T*, T*, std::vector<T>)`.
Getting the iterator of the destination with `blocks_.begin()` appears to make it work and pass the tests run by the CI.

fevangelista (approved):

PR #2968, opened 2023-05-24
Update gdma to v2.3; convert to runtime

loriab:

- [x] gdma is bumped from upstream v2.2.06 to v2.3.3 (can now use h-functions)
- [x] gdma is now selectable at runtime, not just compile-time. use `conda install pygdma -c conda-forge` (presently in staged-recipes https://github.com/conda-forge/staged-recipes/pull/22882 )

## Dev notes & details
- [x] A perhaps obscure point is that psi hasn't been using exactly upstream GDMA, which is a Fortran executable. Instead, Andy Simmonett did lots of I/O conversion and turned it into a library and bound some functions to C so psi could extract results w/o parsing. That's great, but it's meant we've been fixed at upstream v2.2.06.
- [x] I was startled to realize gdma is GPL. For background, roughly speaking, GPL probably does adhere to psi4 if libgdma statically linked, might adhere if dynamically linked (FSF thinks yes; everyone else thinks no; result is that any GPL makes ppl nervous), doesn't adhere if dlopen'ed or runtime optional. Conda-package-wise, we've always dynamically linked to libgdma, and scouring my memory, there was a plan to have an alternate gdma- and chemps2-free conda build of psi4 to remove doubt; that never happened.
- [x] To get past these problems and get a c-f package going, I
  - [x] Started with current upstream https://gitlab.com/anthonyjs/gdma/-/tree/master and applied Andy's and Holger's gdma->libgdma changes
  - [x] Updated the cmake buildsys with the usual checklist https://github.com/psi4/gdma/issues/6
  - [x] Added back the gdma executable so c-f package wouldn't lose functionality from upstream
  - [x] pybind11-'d the C functions Andy created
  - [x] psi4-side, moved the gdma_interface c-side function over to the driver, so it's pure runtime
  - [x] got the conda package going https://github.com/conda-forge/staged-recipes/pull/22882 for L, M, W
- [x] need to add comment notation so the driver qcvars are associated with gdma 
- [x] UNRELATED: fix mrcc addon so that can pass it FNO orbitals again.

## Checklist
- [x] ~Tests added for any new features~

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz (commented):
Do we have a test on the distributed multipoles?

loriab:
> Do we have a test on the distributed multipoles?

Just one: https://github.com/psi4/psi4/blob/master/tests/gdma/gdma1/input.dat that's mirrored at https://github.com/psi4/psi4/blob/master/tests/pytests/test_addons.py#L14

psi4/driver/driver.py
```diff
@@ -1802,7 +1802,57 @@ def gdma(wfn, datafile=""):
...
+    gof = core.get_output_file()
+    gdma.run_gdma(gof, commands)
+    core.set_output_file(gof, True)
+    core.reopen_outfile()
+
+    nsites = gdma.get_nsites()
+    maxorder = 0
+    for site in range(1, nsites + 1):
+        site_order = gdma.get_order(site)
+        maxorder = max(maxorder, site_order)
```
JonathonMisiewicz, line 1828:
```suggestion
    maxorder = max(gdma.get_order(site) for site in range(1, nsites + 1))
```

JonathonMisiewicz (approved):

psi4/driver/driver.py
loriab: good iea -- I was translating directly from the c++

jturney (approved):

PR #2973, opened 2023-05-30
Adds user option for ScreeningType::None

EricaCMitchell:
## Description
Add ScreeningType::None as a user option and sets ScreeningType::Schwarz as the default. This should allow use of mixed basis set types to be used in integrals as well.

N/A

## Dev notes & details
- [X] Breaks out of setup_sieve() in twobody.cc before exception can be thrown
- [X] Adds "NONE" to list of options in read_options.cc

## Checklist
- [X] All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/twobody.cc
+++ b/psi4/src/psi4/libmints/twobody.cc
@@ -78,10 +78,12 @@ TwoBodyAOInt::TwoBodyAOInt(const IntegralFactory *intsfactory, int deriv)
         screening_type_ = ScreeningType::CSAM;
     else if (screentype == "DENSITY")
         screening_type_ = ScreeningType::Density;
+    else if (screentype == "NONE")
+        screening_type_ = ScreeningType::None;
     else
         throw PSIEXCEPTION("Unknown screening type " + screentype + " in TwoBodyAOInt()");
     
-    if (screening_threshold_ == 0.0) screening_type_ = ScreeningType::None;
+    if (screening_threshold_ == 0.0) screening_type_ = ScreeningType::Schwarz;
 
 }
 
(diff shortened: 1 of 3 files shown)


jturney (approved):
Looks good!

davpoolechem:
So I cross-checked this with the CompositeJK methods, and; as I somewhat suspected might happen, the CompositeJK methods do not work with `screening none`.  `DFDIRJ+LINK` segfaults, while `DFDIRJ+COSX` gives the wrong answer, from my testing. 

The composite methods partially implement the screening machinery within themselves instead of relying entirely on the `TwoBodyAOInt::shell_significant()` function for sieving. Some functions, such as `TwoBodyAOInt::shell_pair_values` (in DF-DirJ), and `TwoBodyAOInt::shell_pair_significant` (in LinK) are used to assist with this. However, these functions rely on values that (specifically, `shell_pair_values_`) are initialized during the call to `TwoBodyAOInt::create_sieve_pair_info`, which is never called when `screening_type` is set to `none` due to the `return` upon registering `ScreeningType::None` for the sieve during `TwoBodyAOInt::setup_sieve`.

Given this wasn't caught by CI, it may be reasonable to add a new test for different JK methods at different screening types, but that wouldn't be for this PR.

loriab (approved):
lgtm, thanks!

I agree with DP that a parametrize on the test_erisieve going through the scf_types could be a good precaution -- we can schedule that as a group programming segment.

davpoolechem:
Talking to @loriab some, I'll be happy to let this PR be merged in as-is, and fix up the issues I brought up earlier in other PRs.

PR #2974, opened 2023-05-31
Begone, ERISieve (Kind of)!

davpoolechem:
## Description
This PR is a reimplementation of https://github.com/psi4/psi4/pull/2933; but instead of entirely deleting the `ERISieve` class, it makes `ERISieve` inaccessible for use through the following:

- Using `core.ERISieve.build` now throws an `UpgradeHelper` exception when called, as compared to the previous behavior of giving a warning.
- Calling the `ERISieve` constructor C++-side now throws an exception, with the same message as the UpgradeHelper would give.

Aside from that, this PR does the same things as https://github.com/psi4/psi4/pull/2933. `ERISieve` is replaced with `TwoBodyAOInt` in all use cases, with adjustments as necessary. Additionally, the new version of v2rdm_casscf is used, which also uses `TwoBodyAOInt` instead of `ERISieve`.

- [X] Calling `core.ERISieve.build` Py-side, or constructing an `ERISieve` object C++-side, now throws an exception. The removed functionalities, specifically `shell_significant`, can be accessed via the construction of a TwoBodyAOInt object.

## Dev notes & details
- [X] The `ERISieve` class in libmints has been made inaccessible, now throwing exceptions when constructed (with an UpgradeHelper Py-side, and a PSIEXCEPTION C++-side).
- [X] Functionalities in `ERISieve` that were not in `TwoBodyAOInt` and were required in the code, were added to `TwoBodyAOInt`.
- [X] The PKJK files have been adjusted as necessary to support the use of `TwoBodyAOInt`.

## TODO
- [X] Remove ERISieve from all other external plugins (currently focusing on v2rdm_casscf).

## Questions
- N/A

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -1565,7 +1565,7 @@ def _core_erisieve_build(
     Previously constructed a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
     ERI screening and an optional input to enable CSAM screening (over Schwarz screening).
 
-    However, the ERISieve class has been removed from Psi4 as of v1.9. So, the function now throws with an UpgradeHelper
+    However, as the ERISieve class was removed from Psi4 in v1.9, the function now throws with an UpgradeHelper
     exception, and lets the user know to use TwoBodyAOInt instead.
 
     Parameters
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -1562,7 +1562,7 @@ def _core_erisieve_build(
         do_csam: bool = False
     ) -> core.ERISieve:
     """
-    Previously constructed a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
+    This function previously constructed a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
     ERI screening and an optional input to enable CSAM screening (over Schwarz screening).
 
     However, as the ERISieve class was removed from Psi4 in v1.9, the function now throws with an UpgradeHelper
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -1562,8 +1562,11 @@ def _core_erisieve_build(
         do_csam: bool = False
     ) -> core.ERISieve:
     """
-    Constructs a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
+    Previously constructed a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
     ERI screening and an optional input to enable CSAM screening (over Schwarz screening).
+
+    However, the ERISieve class has been removed from Psi4 as of v1.9. So, the function now throws with an UpgradeHelper
+    exception, and lets the user know to use TwoBodyAOInt instead. 
 
     Parameters
     ----------
@@ -1574,7 +1577,7 @@ def _core_erisieve_build(
     do_csam
         Use CSAM screening? If True, CSAM screening is used; else, Schwarz screening is used. By default,
         Schwarz screening is utilized.
-
+    
     Returns
     -------
     ERISieve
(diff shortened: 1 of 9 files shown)


loriab (approved):

psi4/driver/p4util/python_helpers.py
```diff
@@ -1562,9 +1562,12 @@ def _core_erisieve_build(
         do_csam: bool = False
     ) -> core.ERISieve:
     """
-    Constructs a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
+    Previously constructed a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
```
susilehtola, line 1565:
```suggestion
    This function previously constructed a Psi4 ERISieve object from an input basis set, with an optional cutoff threshold for
```

psi4/driver/p4util/python_helpers.py
```diff
@@ -1562,9 +1562,12 @@ def _core_erisieve_build(
...
     ERI screening and an optional input to enable CSAM screening (over Schwarz screening).
 
+    However, the ERISieve class has been removed from Psi4 as of v1.9. So, the function now throws with an UpgradeHelper
```
susilehtola, line 1568:
```suggestion
    However, as the ERISieve class was removed from Psi4 as of v1.9, the function now throws with an UpgradeHelper
```

psi4/src/psi4/libfock/PKmanagers.h
```diff
@@ -110,7 +110,7 @@ class PKManager {
     double omega_;
 
     /// Sieving object for the integrals
-    std::shared_ptr<ERISieve> sieve_;
+    std::shared_ptr<TwoBodyAOInt> sieve_;
```
susilehtola, line 113: This doesn't make sense to me. This is not a sieve!

psi4/src/psi4/libfock/PKmanagers.h
```diff
@@ -153,7 +153,7 @@ class PKManager {
     double cutoff() const { return cutoff_; }
     int nthreads() const { return nthreads_; }
     int nbf() const { return nbf_; }
-    std::shared_ptr<ERISieve> sieve() const { return sieve_; }
+    std::shared_ptr<TwoBodyAOInt> sieve() const { return sieve_; }
```
susilehtola, line 156: This should be called something like `integral_evaluator`, no?

davpoolechem: Yeah, that would definitely be sensible, especially because similar objects in other JK classes take names like `eri` or `eri_computer`. I'll be happy to make that change here, too, as a general PKJK cleanup.

psi4/src/psi4/libfock/PKmanagers.h
davpoolechem: Yeah, it would be described more accurately as an integral evaluator. As per your other comment, I will fix that up.

davpoolechem: All right, done and done!

psi4/src/psi4/libfock/PKmanagers.h
davpoolechem: Done!

psi4/src/psi4/libfock/PK_workers.cc
```diff
@@ -40,10 +39,10 @@ namespace psi {
 
 namespace pk {
 
-AOShellSieveIterator::AOShellSieveIterator(std::shared_ptr<BasisSet> prim, SharedSieve sieve_input)
-    : shell_pairs_(sieve_input->shell_pairs()) {
+AOShellSieveIterator::AOShellSieveIterator(std::shared_ptr<BasisSet> prim, SharedSieve eri_input)
```
susilehtola, line 42: Is `SharedSieve` accurate, should it be named e.g. `ERIEvaluator`?

susilehtola (approved):

psi4/src/psi4/libfock/PK_workers.cc
davpoolechem: Ah, nice catch! Given it is a `shared_ptr` to a `TwoBodyAOInt` object, something like `SharedERI` or `SharedInt` would probably be the best name for it.

davpoolechem:
In line with the other comments on this PR, I would presume it would be a good cleanup to also rename the `AOShellSieveIterator`/`AOFctSieveIterator` classes? I'm thinking something like `AOShellQuartetIterator` and `AOFctQuartetIterator`.

PR #2978, opened 2023-06-05
Add Comprehensive JK Build+Screening Testing

davpoolechem:
## Description
This PR is designed to enable testing of a wide variety of combinations of `SCF_TYPE`, `SCF_SUBTYPE`, and `SCREENING` keywords available in Psi4. Of the currently-available JK tests, scf5 covers a wide variety of build methods, but only at a single screening type per method (density or csam, depending on the method). Meanwhile, `test_erisieve.py`, after the updates introduced in https://github.com/psi4/psi4/pull/2973, tests a wide variety of screening types, but with limited testing in conjunction with different JK builds (the only tests that don't use the Python interface of `TwoBodyAOInt` directly, use `SCF_TYPE=DIRECT` or `DF` as the JK method for screening comparisons). This leaves a lot of untested JK build+screening combos, which may be potentially broken and uncaught by the CI as a result. As a matter of fact, such cases actually do exist in the code currently (e.g., CompositeJK methods + no screening).

This PR adds a new pytest module, `test_comprehensive_jk_screening.py`. It is effectively an expanded version of the scf5 test module, testing one of the scf5 systems (singlet oxygen) with the same basis set, but also including different screening methods and algorithmic subtypes available in Psi4. Screening is assumed to have an insignificant impact on energy within the tolerance used, so all screening types for a given method use the same reference energy. Some combinations of method and screening type throw an exception by design; this is accounted for in the test by testing that such combinations do indeed throw an exception as expected. Other combinations of method and algorithm are broken at the moment and error out; these are simply skipped for now. They are all logged in the same spot, and can and will be addressed in future PRs.

- [X] N/A

## Dev notes & details
- [X] Adds a new pytest module to Psi4, `test_comprehensive_jk_screening.py` , to test different combinations of JK build algorithms and ERI screening methods.

## Questions
- [x] Would the test in `test_comprehensive_jk_screening.py` be better placed in `test_erisieve.py`? I placed the test in the former because I considered it large enough to warrant not having the quick pytest mark, but I'm ambivalent about where the test goes between those two test modules.

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/tests/pytests/test_comprehensive_jk_screening.py
+++ b/tests/pytests/test_comprehensive_jk_screening.py
@@ -63,8 +63,8 @@ def test_comprehensive_jk_screening(scf_type, scf_subtype, screening):
     should_error_out = should_error_out or (scf_type in Eref["Singlet"]["Composite"].keys() and screening == "NONE")
     #== .. DFDIRJ+LINK with SCREENING=SCHWARZ or CSAM... ==#
     should_error_out = should_error_out or (scf_type == "DFDIRJ+LINK" and screening in [ "SCHWARZ", "CSAM" ])
-    #== .. and DISK_DF, DIRECT, or Yoshimine PK with SCREENING=NONE ==# 
-    should_error_out = should_error_out or (scf_type == "PK" and scf_subtype == "YOSHIMINE_OUT_OF_CORE" and screening == "NONE")
+    #== .. and DISK_DF, DIRECT, or PK with SCREENING=NONE ==#
+    should_error_out = should_error_out or (scf_type == "PK" and screening == "NONE")
     should_error_out = should_error_out or (scf_type == "DISK_DF" and screening == "NONE")
     should_error_out = should_error_out or (scf_type == "DIRECT" and screening == "NONE")
   
--- a/tests/pytests/test_comprehensive_jk_screening.py
+++ b/tests/pytests/test_comprehensive_jk_screening.py
@@ -5,7 +5,10 @@ from utils import compare, compare_integers, compare_values
 
 pytestmark = [pytest.mark.psi, pytest.mark.api] 
 
-def test_comprehensive_jk_screening():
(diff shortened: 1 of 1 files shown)


tests/pytests/test_comprehensive_jk_screening.py
```diff
@@ -0,0 +1,121 @@
...
+        "print": 2,
+    })
+    
+    #== define various scf params to combine test ==#
+    scf_params = {
+        "SCF_TYPE": [ "PK", "DIRECT", "OUT_OF_CORE", "DISK_DF", "MEM_DF", "DFDIRJ+LINK", "DFDIRJ+COSX" ],
+        
+        "SCF_SUBTYPE": [ "AUTO", "INCORE", "OUT_OF_CORE", "YOSHIMINE_OUT_OF_CORE", "REORDER_OUT_OF_CORE" ],
+        
+        "SCREENING": [ "SCHWARZ", "DENSITY", "CSAM", "NONE" ],
```
loriab, line 49: nice test! did you consider doing this as three nested `pytest.parametrize`? it'd parallelize better, and the output might be clearer as 140 tests instead of 1.

tests/pytests/test_comprehensive_jk_screening.py
```diff
@@ -0,0 +1,102 @@
...
+
+    #== define reference energies ==#
+    Eref = {  
+        "Nuclear"       :   30.7884922572,     #TEST
+        "Singlet": {
+            "Canonical" : -149.58723684929720, #TEST
+            "DF"        : -149.58715054487624, #TEST
+            "Composite": {
+              "DFDIRJ+COSX"    : -149.58722317236171, #TEST
+              "DFDIRJ+LINK"    : -149.58726772171027  #TEST
```
loriab, line 23: fwiw, the `#TEST` don't do anything in `tests/pytests/`

loriab (approved):
lgtm, thanks!

tests/pytests/test_comprehensive_jk_screening.py
davpoolechem: Ah, good to know. I will probably remove it, then.

jturney (approved):

PR #2981, opened 2023-06-09
More SCF Convergence Control for COSX Final Grid

davpoolechem:
## Description
Currently, COSX in Psi4 uses a two-grid scheme to complete the SCF - the SCF is first converged on a smaller grid, and then a single iteration is performed on a larger grid, a scheme proposed by Neese. This scheme works quite well for energies; but since the SCF isn't converged on the larger grid, it may not be as ideal if wavefunction properties (e.g., gradients) are desired.

This PR fixes that by adding a new keyword, `COSX_MAXITER_FINAL`, an integer that determines the maximum number of SCF iterations to run on the final COSX grid. By default, this keyword is set to 1, mimicking the current behavior of the code. A more detailed description of behavior can be broken down as follows:

- If `COSX_MAXITER_FINAL` < 0; the SCF will attempt to fully converge on the final COSX grid (within the constraints provided by `MAXITER`, of course).
- If `COSX_MAXITER_FINAL` == 0; the SCF will not run on the final grid at all, effectively turning the COSX algorithm into a single-grid implementation.
- If `COSX_MAXITER_FINAL` > 0; the SCF will run on the final grid for a maximum number of iterations specified by the keyword. 
    - If the SCF converges on the final grid before the iteration count specified by `COSX_MAXITER_FINAL`, the SCF will exit upon convergence.
    - If the iteration count specified by `COSX_MAXITER_FINAL` is reached before the SCF is converged on the final grid, the SCF will exit, treated as a success.
    - If the `MAXITER` SCF iteration limit is hit before the `COSX_MAXITER_FINAL` limit is reached, the calculation will fail.

Also of note, this PR changes CompositeJK so that early_screening is disabled for LinK-based composite methods. Practically, this means that LinK-based methods do not run the extra post-early-screening SCF iteration that is present in COSX-based methods.

- [X] Adds a new keyword, `COSX_MAXITER_FINAL`, for controlling the maximum number of SCF iterations to run on the final COSX grid.

## Dev notes & details
- [X] Adds a new keyword, `COSX_MAXITER_FINAL`, for controlling the maximum number of SCF iterations to run on the final COSX grid. The keyword can be used to run a preset number of SCF iterations or fully converge the SCF on the final grid, or to skip the final grid entirely.
- [X] Changes the early screening feature so that it is only turned on for COSX-based composite methods.

## Questions
- [x] N/A

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -270,6 +270,8 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
     # does the JK algorithm use severe screening approximations for early SCF iterations?
     early_screening = self.jk().get_early_screening()
+    # does the JK algorithm fully converge after early screening is disabled? 
+    early_screening_full_scf = core.get_option('SCF', 'COSX_FULL_SCF') 
 
     # has early_screening changed from True to False?
     early_screening_disabled = False
@@ -497,7 +499,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
             continue
 
         # this is the first iteration after early screening was turned off
-        if early_screening_disabled:
+        if early_screening_disabled and not early_screening_full_scf:
             break
 
         # Call any postiteration callbacks
@@ -518,7 +520,10 @@ def scf_iterate(self, e_conv=None, d_conv=None):
                     self.jk().clear_D_prev()
 
                 core.print_out("  Energy and wave function converged with early screening.\n")
-                core.print_out("  Performing final iteration with tighter screening.\n\n")
+                if early_screening_full_scf:
+                    core.print_out("  Continuing SCF iterations with tighter screening.\n\n")
+                else:
+                    core.print_out("  Performing final iteration with tighter screening.\n\n")
             else:
                 break
 
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -270,7 +270,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
     # does the JK algorithm use severe screening approximations for early SCF iterations?
     early_screening = self.jk().get_early_screening()
-    # does the JK algorithm fully converge after early screening is disabled? 
+    # does the SCF fully converge after early screening is disabled? 
     early_screening_full_scf = core.get_option('SCF', 'COSX_FULL_SCF') 
 
     # has early_screening changed from True to False?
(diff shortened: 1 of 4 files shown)


psi4/src/read_options.cc
```diff
@@ -1681,6 +1681,8 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_int("COSX_RADIAL_POINTS_FINAL", 35);
         /*- Screening criteria for integrals and intermediates in COSX -*/
         options.add_double("COSX_INTS_TOLERANCE", 1.0E-11);
+        /*- Maximum number of SCF iterations to run on on larger (i.e., final) COSX grid -*/
```
loriab, line 1684: Maybe be more expansive about the >1, ==1, ==0, <0 values here since this is what gets slurped for the docs.

psi4/driver/procrouting/scf_proc/scf_iterator.py
```diff
@@ -270,13 +270,16 @@ def scf_iterate(self, e_conv=None, d_conv=None):
 
     # does the JK algorithm use severe screening approximations for early SCF iterations?
     early_screening = self.jk().get_early_screening()
+    # maximum number of scf iterations to run after early screening is disabled
+    scf_maxiter_post_screening = core.get_option('SCF', 'COSX_MAXITER_FINAL')
```
loriab, line 274: Is there any practical difference between cosx_maxiter_final == -1 and -10? If the total total limit is still controlled by SCF_MAXITER, then no? If that's the case, maybe only allow -1.

loriab (commented):
nice additional flexibility

psi4/src/read_options.cc
davpoolechem: Yeah, that's fair, since COSX_MAXITER_FINAL is a bit more complex than it appears at a first glance. I will do so!

davpoolechem: Also done! Hopefully, the new description in the code makes sense.

psi4/src/read_options.cc
```diff
@@ -1681,6 +1681,11 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         options.add_int("COSX_RADIAL_POINTS_FINAL", 35);
         /*- Screening criteria for integrals and intermediates in COSX -*/
         options.add_double("COSX_INTS_TOLERANCE", 1.0E-11);
+        /*- Controls SCF iteration behavior for the larger (i.e., final) COSX grid.
+        -1 fully converges the SCF on the final grid if possible, ending early if MAXITER total SCF iterations are performed.
+        0 disables the final COSX grid entirely.
+        n runs up to n iterations on the final COSX grid, ending early if SCF convergence is reached or MAXITER total SCF iterations are performed. -*/
```
loriab, line 1687:
```suggestion
        n runs up to n iterations on the final COSX grid, ending early if SCF convergence is reached (success) or if |scf__maxiter| total SCF iterations are reached (failure). -*/
```

loriab (approved):
See if you like the proposed changes.

davpoolechem:
> See if you like the proposed changes.

I do! I accepted all of them. Thank you for the suggestions!

PR #2982, opened 2023-06-09
CABS, rework build_cabs_space

EricaCMitchell:
## Description
Provides users the ability to form a basisset composed of two combined basissets, the two can be combined simply or through use of the complementary auxiliary basis set (CABS) approach of Valeev et al.

## Dev notes & details
- [X ] Uncomments `pyconstruct_combined` in qcdb with minor changes to fit updated python
- [X ] Changes input to `build_ri_space` in OrbitalSpace to require a prebuilt combined `BasisSet` object
- [X ] Changes `build_cabs_space` in OrbitalSpace to use a full SVD
- [X ] Adds pytest `test_orbitalspace.py` to check orthonormality between orbital basis set and CABS
- [X ] Adds updated F12 basissets from the Basis Set Exchange
- [X ] Use of the CABS created works with my F12 plugin giving correct energies

## Questions
- [ ] Shoud I generalize `_pybuild_basis` aka `BasisSet.build` to work with a list of keys, targets, roles, and other? Or create a new function to more easily access the building of a combined basisset?

## Checklist
- [X ] `test_orbitalspace.py`
- [X ] Passes quick tests [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [X ] Ready for review
- [ ] Ready for merge


susilehtola:
The fix to one-electron integrals should be its own pull request. It has nothing to do with the other changes, which are pretty extensive by themselves to review.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,86 @@ def build(molecule, shells):
...
+        for at in range(len(keys)):
+            bas = BasisSet.pyconstruct(mol, keys[at], targets[at], fitroles[at], others[at])
+            name += targets[at] + " + "
+            keywords += keys[at] + " + "
+            blends += bas.name.upper() + " + "
+            sets.append(bas)
+
+        name = name[:-3].strip()
+        keywords = keywords[:-3].strip()
+        blends = blends[:-3].strip()
```
loriab, line 554:
```suggestion
        sets = [BasisSet.pyconstruct(mol, keys[at], targets[at], fitroles[at], others[at]) for at in range(len(keys))]
        name = " + ".join(targets)
        keywords = " + ".join(keys)
        blends = " + ".join([bas.name.upper() for bas in sets])
```
something like this could also work.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,86 @@ def build(molecule, shells):
...
+            for label, basis_map in atom_basis_shell.items():
+                if label not in combined_atom_basis_shell:
+                    combined_atom_basis_shell[label] = collections.OrderedDict()
+                    combined_atom_basis_shell[label][name] = []
+                for basis, shells in basis_map.items():
+                    combined_atom_basis_shell[label][name].extend(shells)
+
+        # sort the shells by angular momentum
+        for label, basis_map in combined_atom_basis_shell.items():
+            combined_atom_basis_shell[label][name] = sorted(combined_atom_basis_shell[label][name],\
```
loriab, line 570:
```suggestion
            combined_atom_basis_shell[label][name] = sorted(combined_atom_basis_shell[label][name],
```
as long as there's an open parenthesis or list or similar, python doesn't need the line continuation

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,86 @@ def build(molecule, shells):
...
+                    key=lambda shell: shell.l)
+
+        # Molecule and parser prepped, call the constructor
+        mol.set_basis_all_atoms(name, "CABS")
```
loriab, line 574: This isn't allowing `basis { ... } ` style input, but that's fine for now.

loriab (commented):
Thanks for all the basis set work!

loriab (approved):

JonathonMisiewicz:
@konpat, do you think this PR would benefit from you looking it over? I know you have some F12 work in mind.

konpat:
Yup, @JonathonMisiewicz, I do have this PR on my radar and it's overall a super useful contribution! Unfortunately, I'm very short on time for the next couple weeks and it might still be a while until I properly test it out.

tests/pytests/test_orbitalspace.py
```diff
@@ -0,0 +1,48 @@
...
+    bs1 = s1.basisset()
+    bs2 = s2.basisset()
+    C1 = np.array(s1.C())
+    C2 = np.array(s2.C())
+    mints = psi4.core.MintsHelper(bs1)
+    S_ao = np.array(mints.ao_overlap(bs1, bs2))
+    S_mo = np.linalg.multi_dot([C1.T, S_ao, C2])
+
+    if o1 != o2:
+        np.testing.assert_allclose(np.dot(S_mo.T, S_mo), np.zeros((C2.shape[1], C2.shape[1])), rtol=1e-05, atol=1e-07)
```
konpat, line 46: Does this test really need to verify that S<sup>T</sup>S is identity or zero? I think we just need to check the matrix S itself for orthonormality (or orthogonality between different spaces):
```    
    if o1 != o2:
        np.testing.assert_allclose(S_mo, np.zeros((C1.shape[1], C2.shape[1])), rtol=1e-05, atol=1e-07)
    else:
        np.testing.assert_allclose(S_mo, np.eye(C1.shape[1]), rtol=1e-05, atol=1e-07)
```

konpat (approved):
Great contribution! It took some work, but I made my development F12 code to work with it, and the results agree with my old CABS code (which I'm happy to ditch).

Sorry it took so long.

tests/pytests/test_orbitalspace.py
JonathonMisiewicz: Another (very picky) issue: shouldn't we prefer to check if `o1` and `o2` are approximately equal, to within some numerical threshold?

psi4/src/psi4/libmints/orbitalspace.cc
```diff
@@ -206,79 +206,54 @@ OrbitalSpace orthogonal_compliment(const OrbitalSpace &space1, const OrbitalSpac
...
-        }
+    // Half-transform to RIBS
+    auto C12 = std::make_shared<Matrix>("C12", space1.C()->colspi(), space2.C()->colspi());
+    C12->gemm(false, false, 1.0, O12, space2.C(), 0.0);
+
+    // SVD of MO overlap matrix
+    std::tuple<SharedMatrix, SharedVector, SharedMatrix> svd_a_tmps = C12->svd_a_temps();
+    SharedMatrix U = std::get<0>(svd_a_tmps);
+    SharedVector S = std::get<1>(svd_a_tmps);
+    SharedMatrix Vt = std::get<2>(svd_a_tmps);
```
JonathonMisiewicz, line 220: Just at a glance, is there a reason not to use `std::tie` here?

I can do a more thorough review later.

EricaCMitchell: You're right, I've updated to an even shorter form within C++17 standard.

tests/pytests/test_orbitalspace.py
EricaCMitchell: @konpat You're right! I've fixed the test to just look at the S_mo matrix.

@JonathonMisiewicz The two spaces are going to be unequal just by construction with the CABS space consisting of the basis functions of the OBS. What I'm checking for here is that we are fulfilling the condition that S12 C = 0, as described in [the paper by Valeev](https://doi.org/10.1016/j.cplett.2004.07.061). I guess I do check more cases than is necessary in the test though.

jturney (approved):
I've been meaning to approve this for a while. Looks good!

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,76 @@ def build(molecule, shells):
...
-#TRIAL#            return basisset
-#TRIAL#        else:
-#TRIAL#            bsdict = {}
-#TRIAL#            bsdict['message'] = text
-#TRIAL#            bsdict['name'] = basisset.name
-#TRIAL#            bsdict['puream'] = int(basisset.has_puream())
-#TRIAL#            bsdict['shell_map'] = basisset.export_for_libmints("CABS")
-#TRIAL#            return bsdict
+    @staticmethod
+    def pyconstruct_combined(mol, keys, targets, fitroles, others):
```
JonathonMisiewicz, line 524: Please add a docstring to make clear what this function is supposed to do.

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,76 @@ def build(molecule, shells):
...
+        # load in the basis sets
+        sets = [BasisSet.pyconstruct(mol, keys[at], targets[at], fitroles[at], others[at]) for at in range(len(keys))]
+        name = " + ".join(targets)
+        keywords = " + ".join(keys)
+        blends = " + ".join([bas.name.upper() for bas in sets])
+
+        # work our way through the sets merging them
+        combined_atom_basis_shell = collections.OrderedDict()
+        for at in range(len(sets)):
+            atom_basis_shell = sets[at].atom_basis_shell
```
JonathonMisiewicz, line 549:
```suggestion
        for basis in sets:
            atom_basis_shell = basis.atom_basis_shell
```

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,76 @@ def build(molecule, shells):
...
+
+            for label, basis_map in atom_basis_shell.items():
+                if label not in combined_atom_basis_shell:
+                    combined_atom_basis_shell[label] = collections.OrderedDict()
+                    combined_atom_basis_shell[label][name] = []
```
JonathonMisiewicz, line 554:
```suggestion
                    combined_atom_basis_shell[label] = collections.OrderedDict(name=[])
```

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,76 @@ def build(molecule, shells):
...
+                for basis, shells in basis_map.items():
+                    combined_atom_basis_shell[label][name].extend(shells)
```
JonathonMisiewicz, line 556:
```suggestion
                for shells in basis_map.values():
                    combined_atom_basis_shell[label][name].extend(shells)
```

psi4/driver/qcdb/libmintsbasisset.py
```diff
@@ -520,76 +520,76 @@ def build(molecule, shells):
...
+        mol.update_geometry()
+
+        # load in the basis sets
+        sets = [BasisSet.pyconstruct(mol, keys[at], targets[at], fitroles[at], others[at]) for at in range(len(keys))]
+        name = " + ".join(targets)
+        keywords = " + ".join(keys)
+        blends = " + ".join([bas.name.upper() for bas in sets])
+
+        # work our way through the sets merging them
+        combined_atom_basis_shell = collections.OrderedDict()
```
JonathonMisiewicz, line 547: Please document what this represents, for non-specialists.

psi4/src/export_mints.cc
```diff
@@ -1617,20 +1617,18 @@ void export_mints(py::module& m) {
         .def("integral", &OrbitalSpace::integral, "The integral factory used to create C")
         .def("dim", &OrbitalSpace::dim, "MO dimensions")
         .def("print_out", &OrbitalSpace::print, "Print information about the orbital space to the output file")
-        .def_static("build_cabs_space", &OrbitalSpace::build_cabs_space,
+	.def_static("build_cabs_space", &OrbitalSpace::build_cabs_space,
                     "Given two spaces, it projects out one space from the other and returns the new spaces \
                     The first argument (orb_space) is the space to project out. The returned space will be orthogonal to this \
                     The second argument (ri_space) is the space that is being projected on. The returned space = this space - orb_space \
```
JonathonMisiewicz, line 1623:
```suggestion
                    The second argument (ri_space) is the space that is being projected on. The returned space = ri_space - orb_space \
```

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -1791,6 +1795,10 @@ void Matrix::svd(SharedMatrix &U, SharedVector &S, SharedMatrix &V) {
 }
 
 void Matrix::svd_a(SharedMatrix &U, SharedVector &S, SharedMatrix &V) {
+#ifdef _OPENMP
```
JonathonMisiewicz, line 1798: What are these lines of code doing?

tests/pytests/test_orbitalspace.py
```diff
@@ -0,0 +1,48 @@
...
+    s2 = spaces[o2]
+    bs1 = s1.basisset()
+    bs2 = s2.basisset()
+    C1 = np.array(s1.C())
+    C2 = np.array(s2.C())
+    mints = psi4.core.MintsHelper(bs1)
+    S_ao = np.array(mints.ao_overlap(bs1, bs2))
+    S_mo = np.linalg.multi_dot([C1.T, S_ao, C2])
+
+    if o1 != o2:
```
JonathonMisiewicz, line 45: What if `o1` and `o2` are _ever so slightly_ off? I understand that in practice, we won't encounter that, but please either don't check this case at all, or have a more robuster check to see if those two matrices are similar.

JonathonMisiewicz (commented):
A few nitpicky additional comments.

tests/pytests/test_orbitalspace.py
EricaCMitchell: The function spaces returns the orbital basis set (OBS) and the complementary auxiliary basis set (CABS). By definition, the CABS is the orthogonal complement to the OBS so the overlap of the MOs between the two should be rigorously zero.

EricaCMitchell: `@pytest.mark.parametrize("o1,o2", [(0, 0), (0, 1), (1, 0), (1, 1)])`

The test looks at the orthonormality of the OBS with itself (`o1 = 0` and `o2 =0`), the OBS with the CABS (`o1 = 0` and `o2 = 1`, or vice versa), and the CABS with itself (`o1 = 1` and `o2 =1`). So `o1 != o2` is sometimes false.

JonathonMisiewicz: Okay, that makes much more sense. I'm used to different parameter values being entirely different computations, not different checks on a single computation. Can you please add a docstring explaining what this test does, so that doesn't bite anybody else? Example: "Test that the CABS and OBS spaces are orthonormal." Renaming "o1" and "o2" to "idx1" and "idx2" would also make a lot of difference.

JonathonMisiewicz:
I'll happily approve after these final comments. Unless anybody else objects, I can merge this in last. Thanks again for your patience!

JonathonMisiewicz:
> Okay, that makes much more sense. I'm used to different parameter values being entirely different computations, not different checks on a single computation. Can you please add a docstring explaining what this test does, so that doesn't bite anybody else? Example: "Test that the CABS and OBS spaces are orthonormal." Renaming "o1" and "o2" to "idx1" and "idx2" would also make a lot of difference.

Just want to make sure this one didn't get lost. 😅

JonathonMisiewicz (approved):
Devs have had ample time to review this one if interested, so I'll add this to merge queue.

Thanks for the code!

JonathonMisiewicz:
The reason for docs build failure is unrelated to this PR. I've asked Lori to take a look.

loriab:
Almost certainly because Ben took down some old qcel/qcng docs in preparation for the qcfractal release. I'll hunt down the replacement. Probably this can still join the merge queue b/c only Azure is required to pass.

PR #2986, opened 2023-06-14
Fix #2985: Symmetry OEINTS

### LINKED ISSUE #2985: One-electron integrals - sometimes square, sometimes triangle
Since moving to LibInt2, the functions that access one-electron integrals, such as `ao_overlap()`, `ao_kinetic()`, ... behave differently when given no parameters (so the integrals are computed within the orbital basis set) and when given two basis set parameters (which is e.g. needed to access integrals in an auxiliary basis). The former gives an appropriate symmetric matrix with all integrals filled in. The latter produces a matrix with only the lower triangle filled with integrals and the upper triangle containing zeros. I think it would be better if the integral functions always filled in the other triangle (by symmetry) and returned a symmetric matrix.

Here's the Python code that demonstrates the issue in case of the AO overlap integrals:
```
import sys
import numpy as np
import psi4

psi4.set_memory('16 GB')
np.set_printoptions(threshold=sys.maxsize)

mol = psi4.geometry("""
0 1
O -0.0669991400 0.0000000000 1.4943547400
H 0.8157342700 0.0000000000 1.8658663900
H 0.0688551000 0.0000000000 0.5391427700

units angstrom
symmetry c1
""")

psi4.set_options({'basis': 'aug-cc-pvdz',
                  'df_basis_mp2':'aug-cc-pvdz-ri',
                })

conv = psi4.core.BasisSet.build(mol,'BASIS', psi4.core.get_global_option('BASIS'))
aux = psi4.core.BasisSet.build(mol,'DF_BASIS_MP2',"", "RIFIT", psi4.core.get_global_option('DF_BASIS_MP2'))

wfn = psi4.core.Wavefunction.build(mol, psi4.core.get_global_option('BASIS'))
mints = psi4.core.MintsHelper(wfn.basisset())

Soo = np.asarray(mints.ao_overlap())
Sxx = np.asarray(mints.ao_overlap(aux,aux))

print("Orbital overlap matrix")
print(Soo)
print("Auxiliary overlap matrix")
print(Sxx)
```

EricaCMitchell:
## Description
Quick fix to symmetry in one-electron integral calls in MintsHelper. Closes #2985 

## Dev notes & details
- [X] Changes the bool to `one_body_ao_computer` to check if the provided basis sets are the same in `mintshelper.cc`

## Questions
- [X] N/A

## Checklist
- [X] Adds a test for the one-electron overlap integrals in `test_mints.py`

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/mintshelper.cc
+++ b/psi4/src/psi4/libmints/mintshelper.cc
@@ -618,7 +618,7 @@ SharedMatrix MintsHelper::ao_overlap(std::shared_ptr<BasisSet> bs1, std::shared_
         ints_vec.push_back(std::shared_ptr<OneBodyAOInt>(factory.ao_overlap()));
     }
     auto overlap_mat = std::make_shared<Matrix>(PSIF_AO_S, bs1->nbf(), bs2->nbf());
-    one_body_ao_computer(ints_vec, overlap_mat, false);
+    one_body_ao_computer(ints_vec, overlap_mat, bs1 == bs2);
     return overlap_mat;
 }
 
@@ -650,7 +650,7 @@ SharedMatrix MintsHelper::ao_kinetic(std::shared_ptr<BasisSet> bs1, std::shared_
         ints_vec.push_back(std::shared_ptr<OneBodyAOInt>(factory.ao_kinetic()));
     }
     auto kinetic_mat = std::make_shared<Matrix>("AO-basis Kinetic Ints", bs1->nbf(), bs2->nbf());
-    one_body_ao_computer(ints_vec, kinetic_mat, false);
+    one_body_ao_computer(ints_vec, kinetic_mat, bs1 == bs2);
     return kinetic_mat;
 }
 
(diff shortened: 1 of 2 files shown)


susilehtola:
Does it work if the basis sets are different?

EricaCMitchell:
It works for both cases, basis sets are the same and basis sets are different.

EricaCMitchell:
When the basis sets are the same only the lower triangular of the matrix is computed. In MintsHelper, the `one_body_ao_computer` was switched to loop through `OneBodyAOInt::shellpairs` since Psi4 1.6
```
const auto &shell_pairs = ints[0]->shellpairs();
    size_t n_pairs = shell_pairs.size();
```

The `OneBodyAOInt::shellpairs` are computed in that class which checks for equivalent basis sets
```
std::vector<std::pair<int, int>> build_shell_pair_list_no_spdata(std::shared_ptr<BasisSet> bs1,
                                                                 std::shared_ptr<BasisSet> bs2, double threshold) {
    const auto nsh1 = bs1->nshell();
    const auto nsh2 = bs2->nshell();
    const auto bs1_equiv_bs2 = (bs1 == bs2);
```

Since OneBodyAOInt does a separate check for symmetry, the input symmetry to `one_body_ao_computer` was irrelevant due to the shellpairs computed
```
        if (symm) {
...
                    outp[nu][mu] = outp[mu][nu] = ints_buff[index++];
                }
            }
        } else {
...
                    outp[mu][nu] = ints_buff[index++];
                }
```

konpat (approved):
LGTM. I can confirm that this PR fixed the symmetry issue #2985 for me.

susilehtola (approved):

loriab (approved):

PR #2994, opened 2023-06-27
Refactor `JK_deriv2`

JonathonMisiewicz:
## Description
The `JK_deriv2` function exists to compute the overlap-derivative-times-TEI part of the right side of the CPHF equation. It's a J-term and a K-like term. with overlap derivative integrals instead of a density.[1] The previous algorithm to do this first computed the alpha spin part in one function call, and then the beta spin part in another function call. This approach was redundant. The first function call had all the intermediates _necessary_ to compute the beta part, but didn't use them. As a result, the function re-computed JK.

This PR refactors `JK_deriv2` so it computes both spin cases in a single function call.

[1] = There's also a Vx term, but its implementation was both unused and buggy. Now it's just unused. Trying to use it in the _old_ way would be even more redundant.

- [x] The UHF hessian algorithm has been slightly adjusted, which should lead to slightly faster computations. Please report any errors.

## Dev notes & details

- [x] Refactored `JK_deriv2` to compute both spin cases in a single function call.
- [x] Makes UKS LDA hessians much less ugly.
- [x] Makes `JK_deriv2` comply with `compute_Vx`'s expected function signature  

## Checklist
- [x] `ctest -R scf-hess` passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/scfgrad/response.cc
+++ b/psi4/src/psi4/scfgrad/response.cc
@@ -2509,8 +2509,7 @@ std::shared_ptr<Matrix> USCFDeriv::hessian_response()
     jk->initialize();
 
     // Jpi/Kpi
-    JK_deriv2(jk,mem, Ca, Ca_occ, Cb, Cb_occ, nso, naocc, nbocc, navir, true);
-    JK_deriv2(jk,mem, Cb, Cb_occ, Ca, Ca_occ, nso, nbocc, naocc, nbvir, false);
+    JK_deriv2(jk,mem, Ca, Ca_occ, Cb, Cb_occ, nso, naocc, nbocc, navir);
 
     VXC_deriv(Ca, Ca_occ, nso, naocc, navir, true);
     VXC_deriv(Cb, Cb_occ, nso, nbocc, nbvir, false);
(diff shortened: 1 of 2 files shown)


jturney (approved):
Yeah looks cleaner

psi4/src/psi4/scfgrad/response.cc
```diff
@@ -4399,52 +4407,64 @@ void USCFDeriv::JK_deriv2(std::shared_ptr<JK> jk, int mem,
...
+                J[2*a+1]->subtract(Vx[2*a+1]);
+            }
 
-            C_DGEMM('N','N',nso,n1occ,nso,1.0,J[2*a+1]->pointer()[0],nso,C1op[0],n1occ,0.0,Tp[0],n1occ);
-            C_DGEMM('T','N',nmo,n1occ,nso,-1.0,C1p[0],nmo,Tp[0],n1occ,1.0,Up[0],n1occ);
+            // TODO: We can simplify the code by converting these to transform calls, but that means allocating
+            //       temporary intermediates in a loop. In the (likely) event the allocation time cost is
+            //       negligible in comparison to the J/K build, simplify away.
+            C_DGEMM('N','N',nso,naocc,nso,1.0,J[2*a]->pointer()[0],nso,Caop[0],naocc,0.0,Tap[0],naocc);
+            C_DGEMM('T','N',nmo,naocc,nso,1.0,Cap[0],nmo,Tap[0],naocc,0.0,Uap[0],naocc);
```
loriab, line 4458: Change of constants from -1,1 to 1,0 deliberate?

loriab (approved):
Looks reasonable (and a lot more readable) to me. I'm curious if there was an intended generalization by the c1/c2 setup.

psi4/src/psi4/scfgrad/response.cc
JonathonMisiewicz: Yes.

In the old code, we backtransform four times and then add the pieces together. In the new code, we add the four pieces together and backtransform once. The job that those constants used to do is now fulfilled by `J[2*a]->add(J[2*a+1]); J[2*a]->scale(-1);` a few lines above.

JonathonMisiewicz:
> Looks reasonable (and a lot more readable) to me. I'm curious if there was an intended generalization by the c1/c2 setup.

Yes, and getting rid of that intended generalization is _precisely_ the point of this PR.

The intended generalization of the c1/c2 setup was that you could pass in alpha and then beta to compute the alpha block, or you could pass in beta and then alpha to compute the beta block. We're now computing both at once, so we don't need that. Having c1/c2 now is just confusing. Instead, we make explicit that alpha comes before beta.

PR #2995, opened 2023-06-27
Update oeprop.rst

CDSherrill:
Clarify that the Hirshfeld volume ratios are Hirshfeld volume ratios, not "free atom volumes" (which are pre-tabulated)

## Status
- [x] Ready for review
- [x] Ready for merge


maxscheurer (approved):

loriab (approved):
If this passes the docs build, all well and good. But I suspect it's going to fail because the table spacing and delimiters got out of alignment.

Issue #2970, opened 2023-05-27
v1.7 PSIO_ERROR


njzjz:
I use psi4 v1.7 to perform gradient calculations under the WB97M-D3BJ/DEF2-TZVPPD level. Most structures work fine, but I get the following error in some structures without a clear error message.

```
  ==> Integral Setup <==

PSIO_ERROR: unit = 97, errval = 12
forrtl: error (76): Abort trap signal
Image              PC                Routine            Line        Source
libgdma.so         00007F6F7C68B10F  for__signal_handl     Unknown  Unknown
libpthread-2.28.s  00007F6F9C906B20  Unknown               Unknown  Unknown
libc-2.28.so       00007F6F9BDDC37F  gsignal               Unknown  Unknown
libc-2.28.so       00007F6F9BDC6DB5  abort                 Unknown  Unknown
libstdc++.so.6.0.  00007F6F8C6B535A  Unknown               Unknown  Unknown
libstdc++.so.6.0.  00007F6F8C6B53C5  Unknown               Unknown  Unknown
libstdc++.so.6.0.  00007F6F8C6B5658  Unknown               Unknown  Unknown
core.cpython-310-  00007F6F88B5A129  Unknown               Unknown  Unknown
core.cpython-310-  00007F6F88B69438  _ZN3psi4PSIO2rwEm     Unknown  Unknown
core.cpython-310-  00007F6F88B6B963  _ZN3psi4PSIO5writ     Unknown  Unknown
core.cpython-310-  00007F6F88B566E9  Unknown               Unknown  Unknown
libstdc++.so.6.0.  00007F6F8C6DFBF4  Unknown               Unknown  Unknown
libpthread-2.28.s  00007F6F9C8FC14A  Unknown               Unknown  Unknown
libc-2.28.so       00007F6F9BEA1DC3  clone                 Unknown  Unknown
/var/spool/slurm/slurmd/job83930/slurm_script: line 27: 2245656 Aborted                 (core dumped)
```

I used conda to install psi4 in a new environment.
```
psi4                      1.7+6ce35a5     py310h769327c_1    psi4
```

Below is the full log, which includes the geometry:

<details>

```

  Memory set to   7.451 GiB by Python driver.
  Threads set to 4 by Python driver.

Scratch directory: /scratch/njzjz/83930/
gradient() will perform analytic gradient computation.

*** tstart() called on aerosol-020
*** at Fri May 26 20:10:38 2023

   => Loading Basis Set <=

[...]

PR #2997, opened 2023-07-02
Print full PSIO error message to cerr as well

TiborGY:
## Description

For some poorly understood confluence of reasons, for some users error messages can _sometimes_ get completely lost when Psi4 exits with a `PSIEXCEPTION`, especially when a PSIO error happens.
This can be observed in recent bug reports, see:

- https://github.com/psi4/psi4/issues/2970
- https://forum.psicode.org/t/sapt0-calculation-crashes-pointer-error/2839

In both of these cases the users were only seeing the error message that has been written to `cerr`, but not the string that `PSIEXCEPTION` should have printed.

IIRC when I tried to research this quite a while ago, I have found this may be a bug in some versions of libstdc++. This PR adds a simple workaround: the exception message is also printed to `cerr` in `psio_error`.

The only side effect, is that some users, who are currently not affected by the lost message problem, may start getting the same error message twice when a PSIO error happens. Spammy error exits are not great, but PSIO errors are not supposed to happen too often, and when they do loosing the error message is more annoying than having it be duplicated.

- [x] PSIO error messages are now also printed to `cerr`, fixing an issue where (under some circumstances) large parts of the error message would never reach the user, 

## Dev notes & details

- [x] The same error message that gets passed to `PSIEXCEPTION` is now printed to `cerr` before throwing in `psio_error`

## Checklist
- [x] There is no easy way to add tests for "are error messages getting delivered or not?"

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libpsio/error.cc
+++ b/psi4/src/psi4/libpsio/error.cc
@@ -131,6 +131,7 @@ void psio_error(size_t unit, size_t errval, std::string prev_msg /* = ""*/) {
                         " If you're a developer, get yourself some coffee.\n";
             break;
     }
+    std::cerr << prev_msg << std::endl;
     throw PSIEXCEPTION(prev_msg);
 }
 


loriab (approved):
I completely agree better two error messages than zero. Thanks for the fix.

Hopefully #1657 will fix the CI.

loriab:
I'm going to update the branch and then add to merge queue, so heads-up that local != remote.

PR #3012, opened 2023-07-12
Cleanup remove_distant_points function.

susilehtola:
## Description

I was looking through the source code to understand what the `DFT_REMOVE_DISTANT_POINTS` keyword does, and realized I could not understand what the `remove_distant_points` function was doing. I also suspect that that code was wrong.

This PR cleans up that function so that it now does what it should.

However, it is also evident that the logic of this function is not optimal. Instead of checking if any nucleus is within the maximal spatial extent of the basis set, one should check if any nucleus is within the extent of that nucleus' basis functions. This is just a drive-by commit, as I don't think it's worthwhile to spend time on Psi4's quadrature code, since [IntegratorXX](https://github.com/wavefunction91/IntegratorXX/) and [GauXC](https://github.com/wavefunction91/GauXC) offer a reusable solution that could be used also in Psi4 with significant negative LOCs. 

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/cubature.cc
+++ b/psi4/src/psi4/libfock/cubature.cc
@@ -4273,7 +4273,8 @@ DFTGrid::DFTGrid(std::shared_ptr<Molecule> molecule, std::shared_ptr<BasisSet> p
     buildGridFromOptions(int_opts_map, str_opts_map, float_opts_map);
 }
 DFTGrid::DFTGrid(std::shared_ptr<Molecule> molecule, std::shared_ptr<BasisSet> primary,
-                 std::map<std::string, int> int_opts_map, std::map<std::string, std::string> str_opts_map, Options &options)
+                 std::map<std::string, int> int_opts_map, std::map<std::string, std::string> str_opts_map,
+                 Options &options)
     : MolecularGrid(molecule), primary_(primary), options_(options) {
     std::map<std::string, double> float_opts_map;
     buildGridFromOptions(int_opts_map, str_opts_map, float_opts_map);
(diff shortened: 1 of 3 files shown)


psi4/src/psi4/libfock/cubature.cc
```diff
@@ -4480,43 +4481,59 @@ void MolecularGrid::block(int max_points, int min_points, double max_radius) {
...
-    int offset = 0;
-    int point_index = 0;
-    std::vector<std::vector<MassPoint>> temp_grids(molecule_->natom());
-    for (int atom = 0; atom < atomic_grids_.size(); ++atom) {
-        for (int Q = 0; Q < atomic_grids_[atom].size(); ++Q) {
-            auto P = atomic_grids_[atom][Q];
-            Vector3 v = molecule_->xyz(atom);
-            double R = (P.x - v[0]) * (P.x - v[0]) + (P.y - v[1]) * (P.y - v[1]) + (P.z - v[2]) * (P.z - v[2]);
-            for (int A = 0; A < molecule_->natom(); A++) {
-                if (Q == A) continue;
```
susilehtola, line 4496: Specifically, this looks odd to me: why is the code comparing the atomic grid index with an atom index?

loriab: here's the last edit of that block: https://github.com/psi4/psi4/commit/6141a691070c2c22aade1dd9e420679f522dfe53#diff-b6a47943fd1f813dc381d6a9e5de7b37bf671d4997497cbaa911a29a359ab66fR4473 . @hokru might be able to comment.

PR #3013, opened 2023-07-12
qcarchive tag, protocol passing

loriab:
## Description

GT will trial this before it's ready for merging, but it's good to get eyes on it and to not lose track of the changes.

- [x] Users of QCFractal `next` can now customize protocols, tags, priorities, and owner_groups

## Dev notes & details

- [x] Allow psi4 to turn all the knobs qcportal allows: https://github.com/MolSSI/QCFractal/blob/next/qcportal/qcportal/client.py#L903-L915

## Questions
- [x] @bennybp look ok? Ben has looked this over and approved

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


tests/ddd-deriv/input.dat
```diff
@@ -17,31 +17,33 @@ molecule h2o {
 set basis cc-pVDZ
 ans = -76.0266327341067125
 
-energy('scf')
+largs = {"protocols": {"stdout": False}, "tag": "test1"}
```
JonathonMisiewicz, line 20: Are the `protocols` and `tag` kwarg documented anywhere? I'm not sure what they do.

loriab: Docs aren't great. `protocols` is here https://molssi.github.io/QCElemental/api/qcelemental.models.AtomicInput.html#qcelemental.models.AtomicInput.protocols , but it looks like I need to add that class to the autodoc list for expanded documentation. Basically, protocols doesn't change what gets computed, but it's a way for the user to change the layout/contents of what gets returned -- don't want miles of text, turn off `stdout` or do want every stage of the optimization, switch from `first_and_last` to `all`.

`tag` is in the qcf domain, so I'm not so familiar with finding its docstring. Generally have to isolate the `next` branch and look it up https://github.com/MolSSI/QCFractal/blob/next/qcportal/qcportal/client.py#L903 . What it means is that you want to pass a string "tag" to the distributed job. Then in the qcf config, if you've labeled a series of compute nodes (with say, local scratch or high memory) with that same tag, qcf knows to send these jobs there. Useful when you've got heterogeneous projects or types of calcs running on a cluster and you want to direct them as you could with slurm.

JonathonMisiewicz (approved):
This is as good as it's going to be. I'm still unhappy with the docs, but that's upstream's problem.

jturney (approved):

PR #3036, opened 2023-08-25
Implement fitted SAP guess

susilehtola:
## Description

This is an implementation of the fitted version of the SAP guess, described in [J. Chem. Phys. 152, 144105 (2020)](https://doi.org/10.1063/5.0004046).

The repulsive potential on every atom is fit in terms of error function potentials, which means that the guess can be evaluated with few two-electron integrals; namely

$$ {\bf F}^\text{SAP}  = {\bf H} + {\bf V}^\text{SAP} $$

$$ V_{\mu \nu}^\text{SAP} =  \sum_{A} (A|\mu \nu) $$

I have no idea how the integrals work in Psi4. I followed [a Psi4Numpy example](https://github.com/psi4/psi4numpy/blob/master/Moller-Plesset/DF-MP2_NAF.py) @loriab hinted me at.

The appears to compile; the only problem being the final assembly, i.e. the sum over the 3-center integrals. If anyone can tell me what I should be doing, help would be appreciated...

## User API & Changelog headlines

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Implement scheme and get something that matches ERKALE
- [ ] Use optimal storage

## Questions
- [ ] Question1

## Checklist
- [ ] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #3037, opened 2023-08-28
Modularize Libint2 constructor

susilehtola:
## Description

To implement the SAP guess (#3036), I need to toy with the contraction coefficients. Currently, there is duplicated data in BasisSet, as the Libint2 shells are built alongside the basis set itself. The idea of this PR is to modularize the Libint2 shell data creation so that the data can be refreshed based on the data in the BasisSet.

The PR also fixes an issue in that the `n_prim_per_shell_` member of BasisSet was never populated.

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
Let me make sure I understand this:

The old behavior was that l2_shells_ was being populated at object construct time. This makes perfect sense if various fields used in constructing the libint2::shell were const, but they are not. Instead, we're migrating the libint2 construction to its own function so it can be called if these fields ever get changed, as #3036 does.

Is that all correct?

JonathonMisiewicz (approved):
Looks good to me. I ping @jturney on libmints changes, by default.

jturney (approved):
Seems good to me!

PR #3060, opened 2023-10-10
Fixes to SCF_TYPE + SCF_SUBTYPE + SCREENING Combinations

davpoolechem:
## Description
This PR changes the code to fix the majority of combinations of `SCF_TYPE`/`SCF_SUBTYPE`/`SCREENING` keywords which were previously broken.

It was discovered that there were a lot of combinations of `SCF_TYPE`, `SCF_SUBTYPE`, and `SCREENING` keywords (e.g., `SCF_TYPE = {CompositeJK}` with `SCREENING = NONE`) that would break upon use, either with an exception or with a hard error such as a segfault. PR https://github.com/psi4/psi4/pull/2978 added testing to detect these broken combinations. This PR actually fixes the broken combinations detected through the added test.

The most notable change as a result of fixing these issues is the handling of how certain variables within the `TwoBodyAOInt` class are initialized. To enable universal support of JK builds with `SCREENING = NONE`, an option has been added to manually call the `TwoBodyAOInt::create_sieve_pair_info()` function via the new wrapper function `TwoBodyAOInt::initialize_sieve()`. Multiple JK builds rely on values of variables initialized via `create_sieve_pair_info()` , and `initialize_sieve()` is now used to initialize the required variables if `SCREENING = NONE` is set. In this way, `SCREENING = NONE` can exist with the benefits it provides, while still working with all JK builds. Also, changes are made to ensure that no screening actually occurs in JK builds when `SCREENING` is set to `NONE`.

In line with the above, some tests have been expanded. `test_comprehensive_jk_screening.py` has been changed to account for the `SCF_TYPE`/`SCF_SUBTYPE`/`SCREENING` combinations that were fixed. `test_erisieve.py` has been updated to more thoroughly test `SCREENING=NONE`.

It is worth noting that, unfortunately, `TwoBodyAOInt::initialize_sieve()` must be defined per integral engine, and so the fixes of this PR only work with Libint2. The issues this PR tries to fix, still persist if Simint is used instead.

- [X] N/A

## Dev notes & details
- [X] Adds a new function to `TwoBodyAOInt`, `initialize_sieve`, which allows for manual initialization of screening variables via `create_sieve_pair_info()` needed for JK builds.
- [X] Uses the above function to fix combinations of `SCF_TYPE`, `SCF_SUBTYPE`, and `SCREENING` which were broken.
- [X] Updates the `test_comprehensive_jk_screening.py` and `test_erisieve.py` test to account for the previously-described fixes.

## Questions
-[] Should we keep the `initialize_sieve` name? I named it as such, because it is basically a light wrapper to manually call `create_sieve_pair_info`. But the name of "sieve" might be objectionable.

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/lib3index/dfhelper.cc
+++ b/psi4/src/psi4/lib3index/dfhelper.cc
@@ -316,7 +316,7 @@ void DFHelper::prepare_sparsity() {
     auto rifactory = std::make_shared<IntegralFactory>(primary_, primary_, primary_, primary_);
     std::vector<std::shared_ptr<TwoBodyAOInt>> eri(screen_threads);
     eri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
-    if (!(eri.front()->initialized())) eri.front()->initialize_sieve();
+    if (!(eri.front()->sieve_initialized())) eri.front()->initialize_sieve();
 #pragma omp parallel num_threads(screen_threads) if (nbf_ > 1000)
     {
         int rank = 0;
@@ -425,7 +425,7 @@ void DFHelper::prepare_AO() {
     auto rifactory = std::make_shared<IntegralFactory>(aux_, zero, primary_, primary_);
     std::vector<std::shared_ptr<TwoBodyAOInt>> eri(nthreads_);
     eri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
-    if (!(eri.front()->initialized())) eri.front()->initialize_sieve();
+    if (!(eri.front()->sieve_initialized())) eri.front()->initialize_sieve();
     for(int rank = 1; rank < nthreads_; rank++) {
         eri[rank] = std::shared_ptr<TwoBodyAOInt>(eri.front()->clone());
     }
@@ -497,7 +497,7 @@ void DFHelper::prepare_AO_wK() {
     auto rifactory = std::make_shared<IntegralFactory>(aux_, zero, primary_, primary_);
     std::vector<std::shared_ptr<TwoBodyAOInt>> eri(nthreads_);
     eri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
-    if (!(eri.front()->initialized())) eri.front()->initialize_sieve();
+    if (!(eri.front()->sieve_initialized())) eri.front()->initialize_sieve();
 #pragma omp parallel num_threads(nthreads_)
     {
         int rank = 0;
@@ -517,7 +517,7 @@ void DFHelper::prepare_AO_core() {
     auto rifactory = std::make_shared<IntegralFactory>(aux_, zero, primary_, primary_);
     std::vector<std::shared_ptr<TwoBodyAOInt>> eri(nthreads_);
     eri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
-    if (!(eri.front()->initialized())) eri.front()->initialize_sieve();
+    if (!(eri.front()->sieve_initialized())) eri.front()->initialize_sieve();
 #pragma omp parallel num_threads(nthreads_)
     {
         int rank = 0;
@@ -594,10 +594,10 @@ void DFHelper::prepare_AO_wK_core() {
     std::vector<std::shared_ptr<TwoBodyAOInt>> weri(nthreads_);
 
     eri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
-    if (!(eri.front()->initialized())) eri.front()->initialize_sieve();
+    if (!(eri.front()->sieve_initialized())) eri.front()->initialize_sieve();
 
     weri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->erf_eri(omega_));
-    if (!(weri.front()->initialized())) weri.front()->initialize_sieve();
+    if (!(weri.front()->sieve_initialized())) weri.front()->initialize_sieve();
 #pragma omp parallel num_threads(nthreads_)
     {
         int rank = 0;
@@ -1885,7 +1885,7 @@ void DFHelper::transform() {
     auto rifactory = std::make_shared<IntegralFactory>(aux_, zero, primary_, primary_);
     std::vector<std::shared_ptr<TwoBodyAOInt>> eri(nthread);
     eri[0] = std::shared_ptr<TwoBodyAOInt>(rifactory->eri());
-    if (!(eri.front()->initialized())) eri.front()->initialize_sieve();
+    if (!(eri.front()->sieve_initialized())) eri.front()->initialize_sieve();
 #pragma omp parallel num_threads(nthreads_)
     {
         int rank = 0;
--- a/psi4/src/psi4/libfock/jk.cc
+++ b/psi4/src/psi4/libfock/jk.cc
@@ -185,6 +185,8 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
     /// handle composite methods
     } else if (is_composite) {
         auto jk = std::make_shared<CompositeJK>(primary, auxiliary, options);
+        // INTS_TOLERANCE handling in CompositeJK::common_init() to better account for behavior when SCREENING=NONE
+
         if (options["SCREENING"].has_changed()) jk->set_csam(options.get_str("SCREENING") == "CSAM");
         if (options["PRINT"].has_changed()) jk->set_print(options.get_int("PRINT"));
         if (options["DEBUG"].has_changed()) jk->set_debug(options.get_int("DEBUG"));
--- a/psi4/src/psi4/libmints/twobody.h
+++ b/psi4/src/psi4/libmints/twobody.h
@@ -116,7 +116,7 @@ class PSI_API TwoBodyAOInt {
      */
     typedef std::vector<std::pair<int, int>> PairList;
     /// Is sieve initialized?
-    bool initialized_;
+    bool sieve_initialized_;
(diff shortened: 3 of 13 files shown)


psi4/src/psi4/lib3index/dfhelper.cc
```diff
@@ -316,7 +316,10 @@ void DFHelper::prepare_sparsity() {
 #ifdef _OPENMP
         rank = omp_get_thread_num();
 #endif
-        if (rank) eri[rank] = std::shared_ptr<TwoBodyAOInt>(eri.front()->clone());
+        if (rank) {
+            eri[rank] = std::shared_ptr<TwoBodyAOInt>(eri.front()->clone());
+            if (!eri[rank]->initialized()) eri[rank]->initialize_sieve();
```
susilehtola, line 321: Does `clone` clone the sieve as well? Should one instead first just make sure `eri.front()` has been `initialized()`?

psi4/src/psi4/libfock/jk.cc
```diff
@@ -158,13 +158,12 @@ std::shared_ptr<JK> JK::build_JK(std::shared_ptr<BasisSet> primary, std::shared_
...
             jk->set_df_ints_num_threads(options.get_int("DF_INTS_NUM_THREADS"));
 
-        return std::shared_ptr<JK>(jk);
+        return jk;
 
     /// handle composite methods
     } else if (is_composite) {
         auto jk = std::make_shared<CompositeJK>(primary, auxiliary, options);
 
-        if (options["INTS_TOLERANCE"].has_changed()) jk->set_cutoff(options.get_double("INTS_TOLERANCE"));
```
susilehtola, line 167: why remove this?

psi4/src/psi4/lib3index/dfhelper.cc
davpoolechem: It should, but I will double check. If it does, your suggestion would be a better implementation than what I did here.

davpoolechem: Yeah, it does copy the sieve as well. So I have changed all initialization instances where a clone is involved, to what you suggested!

psi4/src/psi4/libfock/LinK.cc
```diff
@@ -68,7 +68,7 @@ LinK::LinK(std::shared_ptr<BasisSet> primary, Options& options) : SplitJK(primar
     lr_symmetric_ = true;
 
     // set up LinK integral tolerance
-    if (options["LINK_INTS_TOLERANCE"].has_changed()) {
+    if (options["LINK_INTS_TOLERANCE"].has_changed() && options.get_str("SCREENING") != "NONE") {
         linK_ints_cutoff_ = options.get_double("LINK_INTS_TOLERANCE");
```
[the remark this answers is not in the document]
davpoolechem: It defaults to 1e-12 if any `SCREENING` other than `NONE` is selected; and 0.0 if `SCREENING = NONE`.

psi4/src/psi4/libfock/jk.cc
davpoolechem: This has been moved into `CompositeJK::common_init()` to better account for behavior when `SCREENING=NONE`.

loriab (commented):
I agree with all your PR headmatter text, but (1) I'm not quite following how the screening=None case is different in the code. If screening!=None, `setup_sieve` is calling `c_s_p_i()` (create_sieve_pair_info no args) which in turn calls `c_s_p_i(bs)` or throws. If screening==None, `setup_sieve` isn't calling any `c_s_p_i` in the parent class, but the frontmatter text suggests `c_s_p_i()` is getting called in the `initialize_sieve()` fn in the derived L2 class for the screening=None. If that's right, is the difference in the timing of the calls and whether some cutoff or threshold member data has been set? Or more likely I'm entirely missing something.

(2) The creation of a `c_s_p_i()` signature whose purpose is to perform various logic and then to call the existing `c_s_p_i(bs)` signature seems a little weird to me. Then there's `initialize_sieve()` whose whole purpose is to call `c_s_p_i()` but whose names aren't related? I've diagrammed the old (left) and new (right) setups below as I see them. Sorry for the scratch paper.


IMG_2604

samples/linK-1/input.dat
```diff
@@ -12,15 +12,15 @@ molecule mol {
 }
 
 set {
-    scf_type dfdirj+link
+    scf_type dfdirj+cosx
```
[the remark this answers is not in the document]
davpoolechem: Yeah, this is a result of my usual testing on `samples/` files, and not anything meaningful. This is discardable.

davpoolechem:
Thank you for the review, @loriab! I will address both of your primary comments above individually.

davpoolechem:
> I agree with all your PR headmatter text, but (1) I'm not quite following how the screening=None case is different in the code. If screening!=None, setup_sieve is calling c_s_p_i() (create_sieve_pair_info no args) which in turn calls c_s_p_i(bs) or throws. If screening==None, setup_sieve isn't calling any c_s_p_i in the parent class, but the frontmatter text suggests c_s_p_i() is getting called in the initialize_sieve() fn in the derived L2 class for the screening=None. If that's right, is the difference in the timing of the calls and whether some cutoff or threshold member data has been set? Or more likely I'm entirely missing something.

The big difference is that the code associated with `c_s_p_i()` is called _at all_ when `ScreeningType::None` is used in the current PR, where it is _not_ in the current iteration of Psi4. 

In both versions of the code, you are right in that the logic associated with `c_s_p_i()` isn't called within `setup_sieve()` in either this PR or current Psi4 when `ScreeningType::None` is used. However, because `c_s_p_i()` is only ever called within `setup_sieve()` currently, this means that the logic associated with `c_s_p_i()` is _never_ called when `ScreeningType::None` is used.  This leaves a lot of values that are uninitialized because they are initialized by the `c_s_p_i()` logic; and these values are commonly needed in the various JK objects. This leads to a lot of breakage. 

The `initialize_sieve()` function fixes this problem by adding a second point at which the logic for `c_s_p_i()` can be called. Now, the logic for `c_s_p_i()` can actually be executed at all with `ScreeningType::None`, given that `initialize_sieve()` is also called.

davpoolechem:
>(2) The creation of a c_s_p_i() signature whose purpose is to perform various logic and then to call the existing c_s_p_i(bs) signature seems a little weird to me.` 

`c_s_p_i()` encapsulates the logic that was previously performed in `setup_sieve`, after the initialization of the `sieve_impl_` variable, into its own function. This makes it easier to call the logic both from `initialize_sieve()` (when `ScreeningType::None` is used) and `setup_sieve()` (when `ScreeningType::None` is not used). Perhaps this could be refactored in a more elegant way, however.

>Then there's initialize_sieve() whose whole purpose is to call c_s_p_i() but whose names aren't related? I've diagrammed the old (left) and new (right) setups below as I see them. Sorry for the scratch paper.`

No worries! Tehnically, `Libint2TwoElectronInt::initialize_sieve()` does things aside from calling `c_s_p_i()` (basically everything in `Libint2TwoElectronInt::common_init()` after `setup_sieve()` is called as well), but again, I'm open to better ways to implement this.

andyj10224 (approved):
LGTM, after all previous comments by @loriab are addressed

davpoolechem:
So there is something that has long bothered me about this PR, and I want to bring it up because I consider it rather important. And that is the interaction between the `SCREENING=NONE` keyword option and the CompositeJK methods.

Currently, the composite Psi4 methods are in this weird limbo where integral screening occurs both within the `TwoBodyAOInt` object, and within the composite method itself. This is, presumably, undesirable. And here, it opens the question of how to best handle how composite methods behave when `SCREENING=NONE` is set. There are two options:

1. `SCREENING=NONE` disables integral screening only within `TwoBodyAOInt`, while composite method-side screening can be disabled by setting the corresponding `INTS_TOLERANCE` value to 0.0. This is very unintuitive, and something I could easily see throwing someone off in the future (because `SCREENING=NONE` could technically still do integral screening in composite methods). But this at least allows for composite methods to be run without integral screening via `TwoBodyAOInt` easily.
2. `SCREENING=NONE` disables integral screening both within `TwoBodyAOInt` and within the composite methods. This is undoubtedly the most intuitive behavior, but doesn't address the aforementioned issue where integral screening occurs both in `TwoBodyAOInt` and composite methods.

What do you all think?

psi4/src/psi4/libmints/twobody.h
```diff
@@ -203,6 +206,8 @@ class PSI_API TwoBodyAOInt {
     /*
      * Sieve information
      */
+    /// Is sieve initialized?
+    bool initialized() { return initialized_; };
```
jturney, line 210: This is unintuitive that it is referring to the sieve. If I see initialized() I would guess it refers to the object not a subcomponent of the object.

psi4/src/psi4/libmints/twobody.h
```diff
@@ -316,6 +321,9 @@ class PSI_API TwoBodyAOInt {
 
     /// Results go back to buffer_
     void pure_transform(int, int, int, int, int nchunk, bool copy_to_source = true);
+
+    /// Manually set up sieve if desired
+    virtual void initialize_sieve();
```
jturney, line 326: There's already a `setup_sieve` how is this function different and how is the user supposed to know which to use?

jturney (approved):

loriab (approved):
Nice fix to get more combinations working and to not have screening=None taking different routes through the code. I think I understand it well enough :-) I've rebased and added a commit 2973f1870ac8c6b2a3978229c084fa7439b4d9da that incorporates some renames and other review suggestions. And, inevitably, fix CI.

PR #3064, opened 2023-10-26
make `psi4 --version` faster

loriab:

- [x] When running `psi4 --version`, short circuit the `import psi4` command that can be slow on networked drives. Instead, just read the string out of the file.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/run_psi4.py
+++ b/psi4/run_psi4.py
@@ -33,6 +33,7 @@ import atexit
 import datetime
 import json
 import os
+import re
 import sys
 import warnings
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


psi4/run_psi4.py
```diff
@@ -212,10 +219,6 @@ def custom_formatwarning(msg, *args, **kwargs):
 sys.path.insert(1, lib_dir)
 import psi4  # isort:skip
```
JonathonMisiewicz, line 220: If we want to short-circuit the `import psi4`, shouldn't we remove the `import psi4`?

loriab: that's later, after the `args["version"]:` call has already `sys.exit()`ed.

JonathonMisiewicz (approved):

zachglick (approved):

PR #3080, opened 2023-11-17
Turning off FDDS disp when SAPT_DFT_FUNCTIONAL = HF

carolinesargent:
## Description
The purpose of this PR is to turn off FDDS dispersion when SAPT_DFT_FUNCTIONAL = HF in SAPT(DFT). 

SAPT0 uses a HF description of monomers, whereas SAPT(DFT) replaces this description with KS-DFT.  When SAPT(DFT)’s functional is set to HF, the results should be equivalent to those of SAPT0. Currently, SAPT(DFT) / SAPT_DFT_FUNCTIONAL HF computes coupled dispersion with FDDS. This step is costly and unnecessary. 

- [x] The dispersion energy for SAPT(DFT) with HF will now match SAPT0 and no longer calculate or print Disp2,r or Est. Exch-Disp2,r
- [x] No change necessary in input files

## Dev notes & details

- [x] Places SAPT(DFT)'s FDDS dispersion into an if statement so that FDDS disp is only computed when the functional is not HF
- [x] Carries information regarding the functional into the printing code so that the correct dispersion terms are printed out

## Questions
- [ ] Question1

## Checklist
- [x] sapt-dft1, sapt-dft2, and sapt-compare all pass; no need for the creation of additional tests

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/sapt/sapt_proc.py
+++ b/psi4/driver/procrouting/sapt/sapt_proc.py
@@ -266,7 +266,7 @@ def run_sapt_dft(name, **kwargs):
 
     # Call SAPT(DFT)
     sapt_jk = wfn_B.jk()
-    sapt_dft(dimer_wfn, wfn_A, wfn_B, sapt_jk=sapt_jk, data=data, print_header=False, delta_hf=delta_hf)
+    sapt_dft(dimer_wfn, wfn_A, wfn_B, do_dft=do_dft, sapt_jk=sapt_jk, data=data, print_header=False, delta_hf=delta_hf)
 
     # Copy data back into globals
     for k, v in data.items():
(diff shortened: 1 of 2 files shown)


psi4/driver/procrouting/sapt/sapt_util.py
```diff
@@ -122,8 +122,7 @@ def print_sapt_hf_summary(data, name, short=False, delta_hf=False):
         return ret
 
 
-def print_sapt_dft_summary(data, name, short=False):
-
+def print_sapt_dft_summary(data, name, do_dft=True, short=False):
```
loriab, line 125:
```suggestion
def print_sapt_dft_summary(data, name, *, do_dft=True, short=False):
```
If there are any other changes to be made, I'd add a star here so that callers have to name their arguments and don't risk mixing up do_dft and short.

psi4/driver/procrouting/sapt/sapt_proc.py
```diff
@@ -266,7 +266,7 @@ def run_sapt_dft(name, **kwargs):
 
     # Call SAPT(DFT)
     sapt_jk = wfn_B.jk()
-    sapt_dft(dimer_wfn, wfn_A, wfn_B, sapt_jk=sapt_jk, data=data, print_header=False, delta_hf=delta_hf)
+    sapt_dft(dimer_wfn, wfn_A, wfn_B, do_dft=do_dft, sapt_jk=sapt_jk, data=data, print_header=False, delta_hf=delta_hf)
```
loriab, line 269: aha, making use of Yi's already established `do_dft` to identify Hartree–Fock — good strategy.

loriab (approved):
Thanks for the shortcut!

jturney (approved):

PR #3087, opened 2023-11-27
Created variable and method for storing and retrieving the gradients …

johnppederson:
…on the external point charges of the ExternalPotential.

## Description
Provides a way to get the gradient of the potential between a Wavefunction object and an ExternalPotential object on the external point charges. These gradients are collected and stored in a protected SharedMatrix object of the ExternalPotnetial during the ExternalPotential.computePotentialGradients() routine, and they are accessible through a ExternalPotential.gradient() method which is bound to a corresponding method in the Python API.

- [ ] Given a Wavefunction object with an ExternalPotential for which a gradient call has been made, the corresponding gradient on the embedded point charges represented by the ExternalPotential can be retrieved by calling gradient() on the ExternalField

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -1640,6 +1640,7 @@ void export_mints(py::module& m) {
              "Append a vector of charge tuples to a current ExternalPotential")
         .def("addBasis", &ExternalPotential::addBasis, "Add a basis of S auxiliary functions iwth Df coefficients",
              "basis"_a, "coefs"_a)
+        .def("gradient", &ExternalPotential::gradient, "Get the gradient on the embedded charges")
         .def("clear", &ExternalPotential::clear, "Reset the field to zero (eliminates all entries)")
         .def("computePotentialMatrix", &ExternalPotential::computePotentialMatrix,
              "Compute the external potential matrix in the given basis set", "basis"_a)
--- a/psi4/src/psi4/libmints/extern.cc
+++ b/psi4/src/psi4/libmints/extern.cc
@@ -61,6 +61,8 @@ void ExternalPotential::addBasis(std::shared_ptr<BasisSet> basis, SharedVector c
     bases_.push_back(std::make_pair(basis, coefs));
 }
 
+SharedMatrix ExternalPotential::gradient() { return gradient_; }
+
 void ExternalPotential::print(const std::string& out) const {
     std::shared_ptr<psi::PsiOutStream> printer = (out == "outfile" ? outfile : std::make_shared<PsiOutStream>(out));
     printer->Printf("   => External Potential Field: %s <= \n\n", name_.c_str());
@@ -205,7 +207,9 @@ SharedMatrix ExternalPotential::computePotentialGradients(std::shared_ptr<BasisS
     int natom = mol->natom();
     int nextc = charges_.size();
     auto grad = std::make_shared<Matrix>("External Potential Gradient", natom, 3);
+    auto extg = std::make_shared<Matrix>("Gradient on External Charges", nextc, 3);
     double **Gp = grad->pointer();
+    double **EGp = extg->pointer();
 
     std::vector<std::pair<double, std::array<double, 3>>> Zxyz;
     for (size_t i=0; i< charges_.size(); ++i) {
@@ -231,6 +235,9 @@ SharedMatrix ExternalPotential::computePotentialGradients(std::shared_ptr<BasisS
             Gp[cen][0] += charge * x / (r * r2);
             Gp[cen][1] += charge * y / (r * r2);
             Gp[cen][2] += charge * z / (r * r2);
+            EGp[ext][0] -= charge * x / (r * r2);
+            EGp[ext][1] -= charge * y / (r * r2);
+            EGp[ext][2] -= charge * z / (r * r2);
         }
     }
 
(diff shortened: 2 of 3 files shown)


psi4/src/psi4/libmints/extern.h
```diff
@@ -61,6 +61,8 @@ class PSI_API ExternalPotential {
     std::vector<std::tuple<double, double, double, double> > charges_;
     /// Auxiliary basis sets (with accompanying molecules and coefs) of diffuse charges
     std::vector<std::pair<std::shared_ptr<BasisSet>, SharedVector> > bases_;
+    /// Gradient, if available, as number of charges x 3 SharedMatrix
+    SharedMatrix gradient_;
```
loriab, line 65: I think member data named `gradient_` that *isn't* natom x 3 and `gradient_ != grad` in the .cc file is going to be confusing. I'd rename it `gradient_on_charges_` or something.

davpoolechem: I'm going to agree with Lori on this one. Building on her idea, maybe it's also worth naming the other gradient `grad` to `gradient_on_atoms` or something likewise, to clearly mark what each gradient in this part of the code refers to.

loriab: Note that the `grad` isn't member data, just a local var, so it's name doesn't have to be so formal. I haven't checked but probably it gets stored on wfn in an array of natom x 3 grads. It's only the ncharge x 3 that's stored (newly) on the ExPot object. With the negative sign, should it be labeled a force?

loriab:
Don't worry about the Eco/Linux error -- it's https://github.com/psi4/psi4/pull/3088#issuecomment-1831177563

psi4/src/export_mints.cc
```diff
@@ -1640,6 +1640,7 @@ void export_mints(py::module& m) {
              "Append a vector of charge tuples to a current ExternalPotential")
         .def("addBasis", &ExternalPotential::addBasis, "Add a basis of S auxiliary functions iwth Df coefficients",
              "basis"_a, "coefs"_a)
+        .def("gradient", &ExternalPotential::gradient, "Get the gradient on the embedded charges")
```
loriab, line 1643:
```suggestion
        .def("gradient_on_charges", &ExternalPotential::gradient_on_charges, "Get the gradient on the embedded charges")
```

loriab: Any further thought on `gradient_on_charges` vs. `force_on_charges`?

psi4/src/psi4/libmints/extern.cc
```diff
@@ -61,6 +61,8 @@ void ExternalPotential::addBasis(std::shared_ptr<BasisSet> basis, SharedVector c
     bases_.push_back(std::make_pair(basis, coefs));
 }
 
+SharedMatrix ExternalPotential::gradient() { return gradient_on_charges_; }
```
loriab, line 64:
```suggestion
SharedMatrix ExternalPotential::gradient_on_charges() { return gradient_on_charges_; }
```

psi4/src/psi4/libmints/extern.h
```diff
@@ -101,6 +103,9 @@ class PSI_API ExternalPotential {
     // Compute the interaction of this potential with an external potential
     double computeExternExternInteraction(std::shared_ptr<ExternalPotential> other_extern);
 
+    /// Returns the gradient on the external potential point charges from the wfn-extern interaction
+    SharedMatrix gradient();
```
loriab, line 107:
```suggestion
    SharedMatrix gradient_on_charges();
```

loriab (commented):
Thanks for the name change. It'd be good to have a test case, too. Do you have one in mind or do you want to add to tests/extern*/input.dat. I can advise on the next steps after.

davpoolechem (approved):

psi4/src/psi4/libmints/extern.h
johnppederson: Let's stick with the gradient terminology over force, since the negative sign in the nuclear contribution comes from reusing the displacement used for the nuclear contribution on the `gradient_on_atoms` (it's a case of r_ij = -r_ji).

johnppederson:
To test of this feature, I have been comparing the `gradient_on_charges` with the `gradient_on_atoms`, since they should sum up to zero along each coordinate. It's not currently possible to get the `gradient_on_atoms` directly from the ExternalPotential, so I have been testing this locally by subtracting the gradient of the `wfn` (SCF in the presence of an ExternalPotential) with the ExternalPotential set to `None` from the gradient of the `wfn` with the ExternalPotential. I can add my local test to the test suite or I can write it into an existing test; which would make more sense?

loriab:
> To test this feature, I have been comparing the gradient_on_charges with the gradient_on_atoms, since they should sum up to zero along each coordinate. It's not currently possible to get the gradient_on_atoms directly from the ExternalPotential, so I have been testing this locally by subtracting the gradient of the wfn (SCF in the presence of an ExternalPotential) with the ExternalPotential set to None from the gradient of the wfn with the ExternalPotential. I can add my local test to the test suite or I can write it into an existing test; which would make more sense?

I see what you mean -- you'd need `SCFDeriv::gradients_` to be exported to get at the `gradient_on_atoms` array. If that'd be generally useful, we could do it. But if it's just for testing, your procedure sounds good. Either way on new test or adding that extra computation and check to an existing extern* test. If you have a different molecule (we're stuck on water and N2), it might be nice to have a new test for variety. If you do do a new test, the procedure is to make a new dir in `tests/`, copy over the CMakeLists.txt, test_input.py, and input.dat from an existing extern test. Edit the name and the input.dat contents, and register the test in `tests/CMakeLists.txt`.

johnppederson:
Exporting `SCFDeriv::gradients_` might be a useful feature to decompose the gradient (and force) contributions on the QM atoms, but for the test on `gradient_on_charges`, I have opted to use existing machinery in Psi4 in a new test directory, `extern4`.

loriab (approved):
lgtm, thanks for the contribution!

PR #3089, opened 2023-11-30
Disable cmake policy

fevangelista:
## Description
Disable cmake policy CMP0144 due to limited cmake support.

- [x] Remove cmake policy CMP0144 from the main CMakeLists.txt file

## Questions
- [x] Is this policy critical? It appears not.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,6 @@
 cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 cmake_policy(SET CMP0094 NEW)
 cmake_policy(SET CMP0075 NEW)
-cmake_policy(SET CMP0144 NEW)
 if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
     # Avoid warning about DOWNLOAD_EXTRACT_TIMESTAMP in CMake 3.24:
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -2,7 +2,6 @@
 cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 cmake_policy(SET CMP0094 NEW)
 cmake_policy(SET CMP0075 NEW)
-cmake_policy(SET CMP0144 NEW)
```
loriab, line 5:
```
if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.27.0")
    cmake_policy(SET CMP0144 NEW)
endif()
```
Sorry, I saw your message this morning and never got to it. This should both silence the warning for 3.27 and not raise errors for <3.27

loriab: GH wouldn't let me do a code suggestion on a deleted line, so I added this fix to #3090

loriab (approved):
From Fedora in #3092, maybe there isn't a safe guard, so let's live with warnings but allow src to configure.

davpoolechem (approved):

PR #3102, opened 2023-12-09
Print final MBIS shell populations and widths

kzinovjev:
## Description
For some applications it may be helpful to know the converged populations and widths of all shells in MBIS, so here goes the patch. As for the initial values, the info is only printed when debug>=1

N/A

## Dev notes & details
- [x] Print converged MBIS shell populations and widths when debug>=1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


davpoolechem (approved):

jthorton:
Hi @kzinovjev great job adding this. 

@loriab It would be great to have these available via the qcengine interface so we could store them in QCArchive when MBIS charges are requested on a calculation, can you adviseif any other changes would be needed to be made to make this possible?

jthorton:
Hi @kzinovjev just checking over this again and noticed that the [populations](https://github.com/psi4/psi4/blob/e105cc2346fba36ccacb77bf8944a3dcf6c097f2/psi4/src/psi4/libmints/oeprop.cc#L2099-L2104) and [widths](https://github.com/psi4/psi4/blob/e105cc2346fba36ccacb77bf8944a3dcf6c097f2/psi4/src/psi4/libmints/oeprop.cc#L2182-L2188) might already be printed can you check they match what you are printing?

kzinovjev:
> Hi @kzinovjev just checking over this again and noticed that the [populations](https://github.com/psi4/psi4/blob/e105cc2346fba36ccacb77bf8944a3dcf6c097f2/psi4/src/psi4/libmints/oeprop.cc#L2099-L2104) and [widths](https://github.com/psi4/psi4/blob/e105cc2346fba36ccacb77bf8944a3dcf6c097f2/psi4/src/psi4/libmints/oeprop.cc#L2182-L2188) might already be printed can you check they match what you are printing?

Hi @jthorton, currently only the atomic charges (so, the sum of populations of all the shells) and the widths of the last shell for each atom are printed. In contrast, with this patch the full result of MBIS density fitting is printed, so populations and widths of all the shells. This is helpful if one wants to split the atomic charges into core/valence shells, as in [MEDFF](http://dx.doi.org/10.1021/acs.jctc.6b00969), [IPML](https://doi.org/10.1063/1.5009502) or [EMLE](https://doi.org/10.1021/acs.jctc.2c00914) methods.

psi4/src/psi4/libmints/oeprop.cc
```diff
@@ -2027,6 +2027,17 @@ std::tuple<SharedMatrix, SharedMatrix, SharedMatrix, SharedMatrix> PopulationAna
 
     if (!is_converged) throw ConvergenceError<int>("MBIS", max_iter, conv, delta_rho_max_0, __FILE__, __LINE__);
 
+    // Final population and width
+    if (print_output && debug >= 1) {
+        for (int atom = 0; atom < num_atoms; atom++) {
+            for (int m = 0; m < mA[atom]; m++) {
+                outfile->Printf("  FINAL ATOM %d, SHELL %d, POP %8.5f, WIDTH %8.5f\n", atom + 1, m + 1, Nai[atom][m],
```
andyj10224, line 2034: Would it be helpful to also print out the angular momentum of the shell (L)?

andyj10224 (approved):
LGTM

psi4/src/psi4/libmints/oeprop.cc
kzinovjev: Hi @andyj10224, in MBIS the shells are just Slater functions, so they are spherically symmetric (L=0).

kzinovjev:
Was solved in a better way in https://github.com/psi4/psi4/pull/3226, so closing this one

PR #3104, opened 2023-12-10
Touch up block_matrix in libciomr

TiborGY:
## Description

`block_matrix` should not let its callers discard the arrays it allocates. This PR adds the `[[nodiscard]]` attribute.
Also, if an allocation fails, `block_matrix` now throws instead of calling `exit(1)` and tells the user how many bytes it has tried and failed to allocate.

- [x] PSI_API function `double **block_matrix(size_t n, size_t m, bool mlock = false)` now has the `[[nodiscard]]` attribute, as discarding its return value is a guaranteed memory leak (unless n or m is zero, in which case no allocation happens).

## Dev notes & details

- [x] `[[nodiscard]]` attribute added to declaration and definition
- [x] `exit(PSI_RETURN_FAILURE)` replaced with `throw PSIEXCEPTION`
- [x] VScode can now parse the docstring of the function

## Checklist
- [x] No new features
- [x] All tests run by the CI are passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libciomr/block_matrix.cc
+++ b/psi4/src/psi4/libciomr/block_matrix.cc
@@ -84,14 +84,16 @@ PSI_API [[nodiscard]] double **block_matrix(size_t n, size_t m, bool memlock) {
     if (A == nullptr) {
         outfile->Printf("block_matrix: trouble allocating memory \n");
         outfile->Printf("n = %ld\n", n);
-        exit(PSI_RETURN_FAILURE);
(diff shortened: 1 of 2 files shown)


fevangelista (approved):

davpoolechem (approved):

TiborGY:
Had to swap the order of `[[nodiscard]]` and `PSI_API`, GCC is lenient about this but Clang is not.
GHA Eco (Linux and MacOS) CI tests are still broken for some reason unrelated to this PR.
Otherwise this should now be ready for merge, when convenient.

PR #3110, opened 2023-12-15
[Draft] MP2-F12

EricaCMitchell:
## Description
New feature to compute the explicitly correlated MP2 energy. Only computes the specific case of MP2-F12/3C(FIX) which has been shown to be the most robust of all versions of MP2-F12. Includes conventional and density-fitted versions of the code and also allows choice between a in-core and disk implementation. Relies on the tensor library Einsums.

Equations come from:
[Werner, Adler, and Manby](https://pubs.aip.org/aip/jcp/article/126/16/164102/905556/General-orbital-invariant-MP2-F12-theory)
[Kong, Bischoff, and Valeev](https://pubs.acs.org/doi/10.1021/cr200204r)
[Manby](https://pubs.aip.org/aip/jcp/article/119/9/4607/817158/Density-fitting-in-second-order-linear-r12-Moller)

### Example Input
```
molecule {
0 1
C    0.000000000    0.000000000    0.000000000
H    1.185992116    1.185992116    1.185992116
H    1.185992116   -1.185992116   -1.185992116
H   -1.185992116    1.185992116   -1.185992116
H   -1.185992116   -1.185992116    1.185992116

units bohr
symmetry c1
}

set {
  BASIS cc-pvtz-f12
  FREEZE_CORE true
  E_CONVERGENCE 1.e-10
}

set mp2-f12 {
  CABS_BASIS cc-pvtz-f12-optri
  DF_BASIS_F12 aug-cc-pvtz-ri
  F12_TYPE df
  CABS_SINGLES true
}

energy('mp2-f12')
```

### Timings
Timings and maxvmem are an average over 50 runs.
Orbital basis set (OBS) is cc-pVTZ-F12 (VTZ-F12) and CABS is cc-pVTZ-F12-OPTRI. For DF, auxiliary basis set (AUX) is aug-cc-pVTZ-RI.

Timings and Max RAM Usage for MP2-F12/3C(FIX):
VTZ-F12 | NOBS | NCABS | Total (s) | Total (min) | maxvmem (GB)
-- | -- | -- | -- | -- | --
CH4 | 125 | 239 | 176.48 | 2.94 | 27.554
NH3 | 107 | 198 | 96.64 | 1.61 | 15.060
H2O | 89 | 157 | 53.02 | 0.88 | 7.828
HF | 71 | 116 | 17.96 | 0.30 | 4.081

Timings and Max RAM Usage for DF-MP2-F12/3C(FIX):
VTZ-F12 | NOBS | NCABS | NAUX | Total (s) | maxvmem (GB)
-- | -- | -- | -- | -- | --
CH4 | 125 | 239 | 290 | 6.24 | 3.219
NH3 | 107 | 198 | 244 | 4.91 | 2.763
H2O | 89 | 157 | 198 | 2.13 | 2.282
HF | 71 | 116 | 152 | 1.18 | 2.128

### VTune Analysis for MP2-F12/3C(FIX)
Memory consumption is most egregious in the `form_teints` where the allocation of the AO ERI is quite large with the largest AO being (NOBS, NOBS, NRI, NRI) e.g. CH4 would be (125, 125, 364, 364)
image

CPU Time all goes back to the `form_teints` function and specifically the `two_body_ao_computer`
image

- [X] MP2-F12 single-point energy

## Dev notes & details

- [X] Computes in-core MP2-F12/3C(FIX) energy
- [X] Computes disk MP2-F12/3C(FIX) energy
- [X] Computes in-core DF-MP2-F12/3C(FIX) energy
- [X] Computes disk DF-MP2-F12/3C(FIX) energy

## Questions
- [ ] I am unsure if I have done the disk implementation correctly.
- [ ] The max RAM usage for the conventional is quite large. I could use some suggestions on how to get this down.
- [ ] The conventional MP2-F12/3C(FIX) is not as usable as I would like with the high max RAM usage and slow integral computation. DF-MP2-F12/3C(FIX) is recommended over the conventional.
- [ ] This version of DF-MP2-F12/3C(FIX) uses a more robust scheme than ORCA and MPQC for the density-fitting. 

## Checklist
- [ ] Documentation

## Status
- [ ] Ready for review
- [ ] Ready for merge


loriab:
Thanks for the PR, Erica! I pushed some lines to the Azure CI so that einsums is enabled and your code has a chance of running :-) . It won't always be this ugly -- ultimately Einsums will be req'd. There's also a blas dependency detail (mkl=2022 vs. 2023) I need to work out to get rid of that openblas pkg.

konpat:
This might be something dumb on my part, but I cannot get this build to run on our Linux system:
```
>>> import psi4
  File "<stdin>", line 1, in <module>
  File "/mmfs1/home/kjp0013/psi4_erika/psi4/objdir_p4d19/stage/lib/psi4/__init__.py", line 71, in <module>
    from . import core
ImportError: /mmfs1/home/kjp0013/psi4_erika/psi4/objdir_p4d19/stage/lib/psi4/core.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN3psi6mp2f126mp2f12ESt10shared_ptrINS_12WavefunctionEERNS_7OptionsE
```

I installed einsums from conda-forge. What else do I need to do?

loriab:
@konpat you may need to `-D ENABLE_Einsums=ON`. You can toggle the value in objdir/CMakeCache.txt, then rebuild.

konpat:
@loriab this did the trick, thank you!

konpat:
Fantastic job on this PR, @EricaCMitchell! It took me a while (I apologize), but I finally translated @mkodrycka 's dispersion-F12 code to an MP2-F12 one and, after some tweaking, was able to reproduce your DF-MP2-F12 correlation energy exactly. 

I learned quite a bit in the process: initially, our implementations (both based on the formulas from the same Werner-Adler-Manby paper) gave minimally different results, and I found out that our programmed expressions differ by several terms that vanish in the GBC approximation. This approximation is pretty good but neither one of us makes it explicitly in the implementation (no elements of the Fock matrix are zeroed). I think this is completely OK.

One avenue to possibly speed up the code is fully exploiting the fact that our F12 amplitudes are diagonal and we don't need to compute off-diagonal elements of some matrices. For example, out of the entire B matrix, we only use terms of the form  `B(i,j,i,j)` and `B(i,j,j,i)`. I know computing just the diagonal elements is easier said than done, but I think there is room for speedup there.

Finally, I know this was not directly a part of this PR, but do you happen to know the source of the 6-Gaussian fit of the Slater correlation factor? Here's what Psi4 uses for `GEM_BETA == 1.0`:

https://github.com/psi4/psi4/blob/f20a7c61ca0f4939885aa28f96d7d88058a71816/psi4/src/psi4/libmints/mintshelper.cc#L930-L933

and this is what Molpro 2012.1 prints out (I don't have a newer version):
` Alpha:                 0.19532     0.81920     2.85917     9.50073    35.69989   197.79328`
` Coeff:                 0.27070     0.30552     0.18297     0.10986     0.06810     0.04224`

This discrepancy, if not removed, does lead to small differences in the final results.

EricaCMitchell:
Thank you @konpat for trying out my code! :)

I will look into speeding up the F12 intermediate matrices by computing only the diagonal elements. I am also looking into speeding up the integral overhead by taking advantage of permutational symmetry of the integrals, but that affects the `TwoBodyAOInt` class as well.

The 6-Gaussian fit comes from [Tew and Klopper](https://pubs.aip.org/aip/jcp/article/123/7/074101/931738). These are close to the coefficients and exponents they reported and are the ones given by MPQC4 and ORCA, which use a linear solve similar to Molpro.

loriab (commented):
Thanks for all your work on this!

I was looking for why windows was failing on CI, and I think I have a fix. Noticed a couple other things along the way.

EricaCMitchell:
Thanks @loriab for looking into the CI failure!
I'll include all these edits in my next commits and I am also working on making some ctests as well.

I haven't had a lot of time to to focus on computing just the diagonal elements of the F12 intermediates, so that's on the backburner.

EricaCMitchell:
CI failures are caused by Einsums v0.5. Once v0.6 goes up @jturney , hopefully, the CI checks pass.

EricaCMitchell:
CI failing due to mamba v2 compatibility issues with [conda-incubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda/issues/366). Fix looks to be on the way, but this will effect all Psi4 CI jobs.

EricaCMitchell:
@jturney What are the plans for the integration of Einsums into Psi4? Given that the base of this PR is Einsums, it would be good to know where it is heading, especially since other groups may be using Psi4/Einsums for their plug-ins.

Also @loriab the speed of this PR is hindered by the integrals, until we can figure out how Psi4 is going to accommodate Einsums, it seems a bit much to write a whole separate JK builder just for the F12 right now.

jturney:
Einsums is currently going to be made an optional dependency; and therefore, MP2-F12 will be made optional. There are some condo compiler issues we're not ready to tackle just yet.

EricaCMitchell:
CI/CD passes for all the computations but fails for the Docs with an error of 
`/home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/autodir_psivariables/module__f12.rst:18: WARNING: 'psivar' reference target not found: F12 OPPOSITE-SPIN CORRELATION ENERGY`

Haven't had the time to create a new JK Integral Engine for this module.

Overall, the [ntegrals.cc](https://github.com/psi4/psi4/pull/3110/files#diff-625d6de7772adf9fd1f968564d02553d31949787252909545bf09820f0f1110d) is the sloppiest code in this PR. If @jturney could review and give some suggestions for code improvement, especially if there is a better conversion from Psi4 matrices to Einsums tensors, that would be very helpful.

psi4/src/psi4/f12/f12_intermediates.cc
```diff
@@ -0,0 +1,875 @@
...
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+#include "mp2.h"
+
+#include "einsums.hpp"
```
jturney, line 31: This becomes:

```
#include <Einsums/Tensor/Tensor.hpp>
#include <Einsums/TensorAlgebra.hpp>
```

psi4/src/psi4/f12/f12_intermediates.cc
```diff
@@ -0,0 +1,875 @@
...
+
+namespace psi {
+namespace f12 {
+
+void MP2F12::form_fock(einsums::Tensor<double, 2> *f, einsums::Tensor<double, 2> *k) {
+    using namespace einsums;
+    using namespace tensor_algebra;
+    using namespace tensor_algebra::index;
```
jturney, line 39:
```
using namespace einsums;
using namespace einsums::tensor_algebra;
using namespace einsums::index;
```

psi4/src/psi4/f12/f12_intermediates.cc
```diff
@@ -0,0 +1,875 @@
...
+    form_oeints(f);
+
+    Tensor Id = create_identity_tensor("I", nocc_, nocc_);
+    {
+        outfile->Printf("     Forming J\n");
+        auto J = std::make_unique<Tensor<double, 4>>("Coulomb", nri_, nocc_, nri_, nocc_);
+        form_teints("J", J.get(), {'O', 'O', 'o', 'o', 'O', 'C', 'o', 'o', 'C', 'C', 'o', 'o'});
+
+        Tensor<double, 4> J_sorted{"pqiI", nri_, nri_, nocc_, nocc_};
+        sort(Indices{p, q, i, I}, &J_sorted, Indices{p, i, q, I}, J);
```
jturney, line 50: `sort` becomes `permute`

EricaCMitchell:
@loriab @cgbriggs99  Thank you for updating this to the latest Einsums and fixing CI issues!

- If I remember correctly, including the CABS with the HF energy better matches with ORCA. This is a good rearrange.
- Should not mix DF and CONV so obeying MP2_TYPE is preferred.
- F12_SUBTYPE is good for consistency with other Psi4 modules.
- DF_BASIS_F12 does set the auxiliary basis set for all three modules ([SCF](https://github.com/EricaCMitchell/psi4/blob/e2cc1fdb6a32b9813cd36ec4446b4fddc85838d8/psi4/driver/procrouting/proc.py#L4417), [MP2](https://github.com/EricaCMitchell/psi4/blob/e2cc1fdb6a32b9813cd36ec4446b4fddc85838d8/psi4/driver/procrouting/proc.py#L4435), and [F12](https://github.com/EricaCMitchell/psi4/blob/e2cc1fdb6a32b9813cd36ec4446b4fddc85838d8/psi4/src/psi4/f12/mp2.cc#L79)). The pointer stored in `DFBS_` is then called during [DF-integral evaluation](https://github.com/EricaCMitchell/psi4/blob/8702bdf45c0b09c0bd73f036d8b3ff9cd567b596/psi4/src/psi4/f12/integrals.cc#L242). The SCF with RI is unusual but ensures that we are running DF-MP2-F12 with a density-fitted Hamiltonian. Essentially, if we are running DF, all parts of the computation should be density-fitted to ensure we have a proper reference.
- There is the F12 CABS CORRECTION ENERGY and the F12 correlation energy can be thought of as a correction to the conventional MP2 energy.

EricaCMitchell:
Ah, I see what you're saying now. Option (c)! We want all computed energies to come from the same Hamiltonian.

doc/sphinxman/source/mp2f12.rst
```diff
@@ -0,0 +1,282 @@
...
+   H 1 1.0 2 104.5
+   symmetry c1
+   }
+
+   set {
+     basis         cc-pvtz-f12
+     cabs_basis    cc-pvtz-f12-optri
+     df_basis_f12  aug-cc-pvtz-ri
+     freeze_core   True
+     f12_type      df
```
loriab:
```suggestion
     mp2_type      df
```

doc/sphinxman/source/mp2f12.rst
```diff
@@ -0,0 +1,282 @@
...
+     f12_beta      1.0
+     cabs_singles  True
+   }
+
+   energy('mp2-f12')
+
+The ``energy('mp2-f12')`` call to :py:func:`~psi4.driver.energy` executes the predefined MP-F12 procedure, 
+first calling the SCF module with a default RHF reference. The two-electron integrals may be solved either with 
+or without DF, as specified with ``F12_TYPE``. The general outline of an MP2-F12 computation is as follows:
```
loriab:
```suggestion
or without DF, as specified with |f12__mp2_type|. The general outline of an MP2-F12 computation is as follows:
```

doc/sphinxman/source/mp2f12.rst
```diff
@@ -0,0 +1,282 @@
...
+MP2-F12 is a complicated theory but the implemented module should be simple to use. All keywords are 
+appropriately documented in the Appendix :ref:`apdx:f12` and care should be taken to read through 
+them before changing defaults. Some basic recommendations for running the module are included below:
+
+* MP2-F12 runs much faster with density-fitted integrals. Unless high accuracy is needed, DF requires much 
+  less computational resources and still provides quality results. If DF is chosen, the HF and MP2 modules 
+  are also set to run DF. Choose a DF basis set with at least triple-zeta quality as 
+  |PSIfours| :ref:`DF-SCF <sec:scf>` module performs poorly with lower quality DF basis sets. 
+  This module utilizes robust density-fitting [Dunlap:2000:2113]_ so energies may differ from other programs.
+  The default ``F12_TYPE`` is ``DF``.
```
loriab:
```suggestion
  The default |f12__mp2_type| is ``DF``.
```

doc/sphinxman/source/mp2f12.rst
```diff
@@ -0,0 +1,282 @@
...
+  that the exponents 0.9, 1.0, and 1.0 :math:`a_0^{-1}` are recommended for cc-pVXZ-F12 where X = D, T, and Q, 
+  respectively, while those for aug-cc-pVXZ where X = D, T, and Q are 1.1, 1.2, and 1.4 :math:`a_0^{-1}`. 
+  The default value is set to 1.0 :math:`a_0^{-1}` to allow flexibility of use.
+
+* MP2-F12 likes threads. The module was developed with threading in mind and the formation of the 
+  integrals and the tensor contractions rely on the threads for efficiency.
+
+* MP2-F12 is memory greedy. The core algorithm has been developed to compute the integrals on-the-fly but 
+  the size of some integrals cannot be avoided. Care should be taken to choose whether the core or disk 
+  algorithm is chosen, ``F12_TYPE``.
```
loriab:
```suggestion
  algorithm is chosen, |f12__f12_subtype| .
```

psi4/src/read_options.cc
```diff
@@ -3254,6 +3254,28 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        options.add_str("F12_SUBTYPE", "INCORE", "INCORE DISK");
+        /*- Whether to read-in stored integrals from previous computation -*/
+        options.add_bool("F12_READ_INTS", false);
+        /*- Set contracted Gaussian-type geminal beta value -*/
+        options.add_double("F12_BETA", 1.0);
+        /*- Choose a basis for Complementary Auxiliary Basis Set -*/
+        options.add_str("CABS_BASIS", "");
+        /*- Whether to compute the CABS Singles Correction -*/
+        options.add_bool("CABS_SINGLES", true);
+        /*- Choose a density-fitting basis for integrals -*/
```
loriab:
```suggestion
        /*- Choose a density-fitting basis for integrals. For |f12__mp2_type| =DF, this is applied to non-F12
        SCF and MP2 parts, too, regardless of |scf__df_basis_scf| or |dfmp2__df_basis_mp2| .  -*/
```

loriab (approved):
Integration lgtm! In future I ought to fill out stdsuite testing more (and come to terms with non-matching HF) and see if there's some default aux setting that can be done for DF_BASIS_F12. Thanks for a great new feature!

jturney (approved):

psi4/src/psi4/f12/f12_intermediates.cc
```diff
@@ -0,0 +1,877 @@
...
+
+    {
+        Tensor<double, 4> F_oooc{"<oo|F|oC>", nact_, nact_, nocc_, ncabs_};
+        form_teints("F", &F_oooc, {'o', 'o', 'o', 'C'});
+
+        Tensor<double, 4> tmp{"Temp", nact_, nact_, nact_, nact_};
+        {
+            einsum(Indices{i, j, k, l}, &tmp, Indices{i, j, m, q}, F_oooc, Indices{k, l, m, q}, F_oooc);
+            permute(1.0, Indices{i, j, k, l}, &(*X), -1.0, Indices{i, j, k, l}, tmp);
+            permute(1.0, Indices{i, j, k, l}, &(*X), -1.0, Indices{j, i, l, k}, tmp);
```
TiborGY, line 125: @EricaCMitchell I see a lot of this `&(*foo)` pattern in the F12 code, which seems redundant at first glance. I mean unless Einsums is doing something special (lazy evaluation?) with operator overloading, dereferencing a pointer then taking the address of the result just yields the original pointer, no?

EricaCMitchell: I see what you mean, probably a lapse in my C++ when I first wrote this. It would probably be good to go back through and ensure that the MP2F12 is up-to-date with all the changes [Einums](https://github.com/Einsums/Einsums) has gone through since this module was developed.

Likely won't get to this until during the holiday break.

PR #3111, opened 2023-12-16
Implement fixed reference variant of DirectJK.

susilehtola:
## Description

This is the variant of DirectJK I mentioned at the end of @davpoolechem's talk at PsiCon last Saturday. The idea is simple: instead of updating the reference every iteration, which is subject to creeping numerical noise, use a fixed reference density and JK matrix for all iterations.

I also simplified some of the DirectJK code, and added missing documentation.

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Should full formation still happen every N iterations? I don't know if this truly is less susceptible to numerical noise accumulation...

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -837,7 +837,7 @@ of integrals that need to be calculated.
 The typical implementation of incremental Fock matrix formation
 updates the reference at every iteration. This makes the screening
 better and better as the wave function converges, because the density
-changes less and less. But, since small changes can accumulate to a
+changes between SCF iterations go to zero as convergence is approached. But, since small changes can accumulate to a
 non-negligible outcome over many iterations, this procedure is
 susceptible to creeping numerical error. For this reason, a fresh Fock
 matrix needs to be rebuilt every ``N`` iterations
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -862,7 +862,7 @@ It has the following options
 
   |scf__incfock|: Toggles incremental formation of the Fock matrix. Defaults to ``true``.
   |scf__incfock_fixed_reference|: Chooses the algorithm: use a fixed reference Fock matrix instead of updating it every iteration? Defaults to ``false``.
-  |scf__incfock_full_every|: The number of iterations between full rebuilds of the Fock matrix. Defaults to ``5``.
+  |scf__incfock_full_fock_every|: The number of iterations between full rebuilds of the Fock matrix. Defaults to ``5``.
   |scf__incfock_convergence|: The convergence threshold to turn off incremental formation of the Fock matrix. Defaults to ``1e-5``.
 
 .. _`sec:scfddfj`:
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -831,7 +831,7 @@ compute the full matrix at every iteration, only the change from a
 previously computed Fock matrix.  This behavior is toggled by the
 |scf__incfock| keyword. Because the elements in the difference density
 matrix are much smaller than the elements in the density matrix, this
-makes the screening of the integrals much tighter, reducing the number
+makes the screening of the integrals much tighter when density-based screening is used, reducing the number
 of integrals that need to be calculated.
 
 The typical implementation of incremental Fock matrix formation
--- a/doc/sphinxman/source/scf.rst
+++ b/doc/sphinxman/source/scf.rst
@@ -684,7 +684,7 @@ DIRECT
     up to 1500 basis functions, uses zero disk (if DF pre-iterations are
     turned off), and can obtain significant
     speedups with negligible error loss if |scf__ints_tolerance|
-    is set to 1.0E-8 or so.
+    is set to 1.0E-8 or so. See the :ref:`sec:directscf` section for more information.
 DF [:ref:`Default <table:conv_scf>`]
     A density-fitted algorithm designed for computations with thousands of
     basis functions. This algorithm is highly optimized, and is threaded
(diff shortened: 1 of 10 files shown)


tests/density-screen-3/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+include(TestingMacros)
+
+add_regression_test(density-screen-3 "psi;dft;scf;medlong")
+set_tests_properties(density-screen-3 PROPERTIES COST 500)
```
loriab, line 4: Lest this test never gets run, add the name of the test to the big list in `tests/CMakeLists.txt`, please. And add a `test_input.py` file to this dir.

susilehtola: Thanks, done

psi4/src/psi4/libfock/DirectJK.cc
```diff
@@ -118,62 +119,84 @@ void DirectJK::print_header() const {
...
+    // Add in the reference J and K
+    if (do_incfock_iter_) {
+        size_t njk = D_ao_.size();
+        for (size_t jki = 0; jki < njk; jki++) {
+            if (do_J_) J_ao_[jki]->add(J_reference_[jki]);
+            if (do_K_) K_ao_[jki]->add(K_reference_[jki]);
+            if (do_wK_) wK_ao_[jki]->add(wK_reference_[jki]);
+        }
+
+        // If we don't do incremental formation, we can update the reference
```
davpoolechem, line 164: The main thing I'd like to point out is right here. With this implementation, the update of the reference is tied to the `INCFOCK_FULL_FOCK_EVERY` keyword, because the reference will only be updated when an IncFock iteration isn't done, i.e., when the full Fock matrix is constructed. This is, of course, great for numerical stability. But, to get consistent reference updates, it forces you to set `INCFOCK_FULL_FOCK_EVERY` to a smallish value. Psi4 uses a small default value already (5), but one goal of the IncFock changes is to avoid consistent reconstruction of the Fock matrix (for some comparisons, ORCA has a similar keyword set to 15 by default if I recall, and GAMESS never recomputes the full Fock matrix). 

Would it be good to decouple the regeneration of the reference from the formation of the full Fock matrices? With such an implementation, some references could be difference matrices themselves, which could be a  numerical stability concern. But perhaps this is okay in practice?

doc/sphinxman/source/scf.rst
```diff
@@ -816,23 +816,72 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+exact Fock matrix by recomputing integrals on the fly at every
+iteration and contracting them with the density matrix. As the
+electron density changes very little in later iterations, the direct
+SCF method often employs incremental formation for the Fock matrix:
+since the Fock matrix is linear in the density, one does not have to
+compute the full matrix at every iteration, only the change from a
+previously computed Fock matrix.  This behavior is toggled by the
+|scf__incfock| keyword. Because the elements in the difference density
+matrix are much smaller than the elements in the density matrix, this
+makes the screening of the integrals much tighter, reducing the number
```
davpoolechem, line 834:
```suggestion
makes the screening of the integrals much tighter when density-based screening is used, reducing the number
```

doc/sphinxman/source/scf.rst
```diff
@@ -816,23 +816,72 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+of integrals that need to be calculated.
+
+The typical implementation of incremental Fock matrix formation
+updates the reference at every iteration. This makes the screening
+better and better as the wave function converges, because the density
+changes less and less. But, since small changes can accumulate to a
```
davpoolechem, line 840:
```suggestion
because the difference in density between SCF iterations approaches zero as convergence is approached.
```

davpoolechem:
Thank you for this PR! I greatly appreciate it!

The cleanup is quite nice, and pretty necessary honestly. Aside from a few small doc wording suggestions, my biggest thought is regarding coupling the regeneration of the reference matrices to construction of the full Fock matrix.

doc/sphinxman/source/scf.rst
```diff
@@ -816,23 +816,72 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+non-negligible outcome over many iterations, this procedure is
+susceptible to creeping numerical error. For this reason, a fresh Fock
+matrix needs to be rebuilt every ``N`` iterations
+(|scf__incfock_full_every|, defaults to 5). In addition, due to the
```
loriab, line 844:
```suggestion
(|scf__incfock_full_fock_every|, defaults to 5). In addition, due to the
```
GHA docs build will be fixed by this

doc/sphinxman/source/scf.rst
```diff
@@ -816,23 +816,72 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+need to be computed for every incremental step. This algorithm is used
+instead of the typical algorithm discussed above, if
+|scf__incfock_fixed_reference| is ``true``.
+
+
+It has the following options
+
+  |scf__incfock|: Toggles incremental formation of the Fock matrix. Defaults to ``true``.
+  |scf__incfock_fixed_reference|: Chooses the algorithm: use a fixed reference Fock matrix instead of updating it every iteration? Defaults to ``false``.
+  |scf__incfock_full_every|: The number of iterations between full rebuilds of the Fock matrix. Defaults to ``5``.
```
loriab, line 865:
```suggestion
  |scf__incfock_full_fock_every|: The number of iterations between full rebuilds of the Fock matrix. Defaults to ``5``.
```

psi4/src/psi4/libfock/DirectJK.cc
susilehtola: I'm not sure I follow what you say, but I'll have another look at the code with clearer eyes when I have the time.

One thing one could do with the changes is to use the fixed-reference variant in the background even when one uses traditional incremental formation: the "full" rebuild would be done with respect to the fixed reference matrix, which would make the consecutive full rebuilds much faster, as one would only need to compute the full matrix once.

doc/sphinxman/source/scf.rst
```diff
@@ -816,23 +816,72 @@ To avoid this, either set |scf__df_basis_scf| to an auxiliary
...
+instead of the typical algorithm discussed above, if
+|scf__incfock_fixed_reference| is ``true``.
+
+
+It has the following options
+
+  |scf__incfock|: Toggles incremental formation of the Fock matrix. Defaults to ``true``.
+  |scf__incfock_fixed_reference|: Chooses the algorithm: use a fixed reference Fock matrix instead of updating it every iteration? Defaults to ``false``.
+  |scf__incfock_full_fock_every|: The number of iterations between full rebuilds of the Fock matrix. Defaults to ``5``.
+  |scf__incfock_convergence|: The convergence threshold to turn off incremental formation of the Fock matrix. Defaults to ``1e-5``.
```
loriab, line 866:
```suggestion
.. include:: autodir_options_c/scf__incfock.rst
.. include:: autodir_options_c/scf__incfock_fixed_reference.rst
.. include:: autodir_options_c/scf__incfock_full_fock_every.rst
.. include:: autodir_options_c/scf__incfock_convergence.rst
```
*If* all the description and default is in read_options.cc, you can let this include it all for you.

andyj10224 (approved):
Thanks for doing this Susi. LGTM

susilehtola:
Superseded by #3530. This branch had gone stale (~2 years, ~180 commits behind master) and its incremental-Fock infrastructure was overtaken by the IncFock standardization in #2808, which landed the same `D_ref_`/`D_prev_` machinery under different names. Rather than force a conflicting rebase of the two implementations, the genuinely new part — the fixed-reference incremental build (`INCFOCK_FIXED_REFERENCE`) — has been reimplemented cleanly on top of the current incfock structure in #3530, along with the `density-screen-3` regression test and SCF-manual documentation. Closing in favor of #3530.

PR #3112, opened 2023-12-17
Clean up #includes, macros, unnecessary array init, error exits, and other miscellaneous things in libmoinfo

TiborGY:
## Description

This PR aims to make a lot of tiny improvements to libmoinfo.
While trying to modernize a bit of libmoinfo code to eliminate some (ab)use of the Dimension functions deprecated by PR #2953 I had to try to get at least some grip on what-does-what, but libmoinfo has accumulated some clutter which made that harder than necessary.
One thing led to another, and now this PR is here to hopefully make it easier for the next person.
Docstring addition and more are the subject of a future PR.

- [x] No user-facing changes

## Dev notes & details

- [x] Removed a some unused macro definitions, inlined the single use of `#define size_det`
- [x] Removed `std::vector<size_t> ioff`, its getter function, and initialization from `MOInfoBase`. It looks like this 5 M element array was getting non-trivially initialized every time an `MOInfoBase` object (or one derived from it) was constructed. RAM usage of these objects should now be 40 MB lower, and their construction should be faster by an unknown amount.
- [x] Removed some class declarations, included the appropriate `.h` instead where required instead of them (Note: some of this was reverted after reviews, see discussion on this PR.)
- [x] Touched-up what-includes-what, along the lines of removing what is not required, but adding what is
- [x] Removed declarations for `tuning()` and `read_mo_spaces2()`, as they are never defined
- [x] Removed some getters from `MOInfo`, either unused or already defined in the base class.
- [x] Removed unused fn `MOInfoBase::get_irr_labs()`
- [x] Renamed `MOInfoBase::get_irr_labs(int i)` to `get_irr_lab`, as it only gets one label. Rename propagated to all of its call sites.
- [x] Removed unused fn `MOInfoBase::get_scf_mos(int i)`
- [x] Moved the `intvec` and `boolvec` typedefs from global namespace, to `namespace psi`. Also modernized them to the `using` syntax.
- [x] Removed an unused `extern ModelSpace* model_space` declaration which was cluttering `namespace psi`
- [x] <strike>Replaced `exit(PSI_RETURN_FAILURE)` and `exit(1)` with `throw PSIEXCEPTION`</strike> Obviated by PR #3118

## Checklist
- [x] No new features
- [x] Tests run by the CI were passing, but now they are failing for unclear reasons

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmoinfo/model_space.h
+++ b/psi4/src/psi4/libmoinfo/model_space.h
@@ -57,9 +57,6 @@ class ModelSpace {
     std::vector<int> unique_to_all;  // spin-unique  determinants
     MOInfo* moinfo_obj;
 };
-
-extern ModelSpace* model_space;
-
 }  // namespace psi
 
 #endif  // _psi_src_lib_libmoinfo_model_space_h_
--- a/psi4/src/psi4/libmoinfo/model_space.h
+++ b/psi4/src/psi4/libmoinfo/model_space.h
@@ -35,11 +35,9 @@
 */
 
 #include "slater_determinant.h"
+#include "moinfo.h"
 
 namespace psi {
-
-class MOInfo;
-
 class ModelSpace {
    public:
     ModelSpace(MOInfo* moinfo_obj_);
--- a/psi4/src/psi4/libmoinfo/model_space.h
+++ b/psi4/src/psi4/libmoinfo/model_space.h
@@ -35,9 +35,11 @@
 */
 
 #include "slater_determinant.h"
-#include "moinfo.h"
 
 namespace psi {
+
+class MOInfo;
+
 class ModelSpace {
    public:
     ModelSpace(MOInfo* moinfo_obj_);
--- a/psi4/src/psi4/libmoinfo/moinfo.h
+++ b/psi4/src/psi4/libmoinfo/moinfo.h
@@ -36,8 +36,6 @@
 
 #include "moinfo_base.h"
 
-#define size_det 2048
-
 namespace psi {
 
 enum ReferenceType { AllRefs, UniqueRefs, ClosedShellRefs, UniqueOpenShellRefs };
@@ -74,7 +72,7 @@ class MOInfo : public MOInfoBase {
         const MOInfo* moinfo;
 
        public:
-        typedef std::bitset<size_det> bitdet;
+        typedef std::bitset<2048> bitdet;
         SlaterDeterminant(const MOInfo*);
         ~SlaterDeterminant();
         void set(int n) { bits.set(n); }
--- a/psi4/src/psi4/libmoinfo/moinfo.h
+++ b/psi4/src/psi4/libmoinfo/moinfo.h
@@ -130,7 +130,6 @@ class MOInfo : public MOInfoBase {
     int get_nocc() const { return (nocc); }
     int get_nvir() const { return (nvir); }
 
-    intvec get_sopi() const { return (sopi); }
     intvec get_mopi() const { return (mopi); }
     intvec get_docc() const { return (docc); }
     intvec get_actv() const { return (actv); }
--- a/psi4/src/psi4/libmoinfo/moinfo.h
+++ b/psi4/src/psi4/libmoinfo/moinfo.h
@@ -37,9 +37,6 @@
 #include "moinfo_base.h"
 
 namespace psi {
-
-enum ReferenceType { AllRefs, UniqueRefs, ClosedShellRefs, UniqueOpenShellRefs };
-
 class MOInfo : public MOInfoBase {
     typedef std::vector<std::string> strvec;
     typedef std::vector<std::pair<int, int> > intpairvec;
@@ -68,6 +65,7 @@ class MOInfo : public MOInfoBase {
       4) Uses
         STL vector
     *********************************************************/
+    enum ReferenceType { AllRefs, UniqueRefs, ClosedShellRefs, UniqueOpenShellRefs };
     class SlaterDeterminant {
         const MOInfo* moinfo;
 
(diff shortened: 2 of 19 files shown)


psi4/src/psi4/libmoinfo/model_space.h
```diff
@@ -35,11 +35,9 @@
 */
 
 #include "slater_determinant.h"
+#include "moinfo.h"
 
 namespace psi {
-
-class MOInfo;
```
loriab, line 41: I understood it to be better (faster compile) to forward declare the class like this rather than including the header. Am I missing something here?

psi4/src/psi4/libmoinfo/moinfo.h
```diff
@@ -74,7 +72,7 @@ class MOInfo : public MOInfoBase {
         const MOInfo* moinfo;
 
        public:
-        typedef std::bitset<size_det> bitdet;
+        typedef std::bitset<2048> bitdet;
```
loriab, line 75:
```suggestion
        typedef std::bitset<2048> bitdet;  // adjust based on "size det"?
```

psi4/src/psi4/libmoinfo/moinfo_base.h
```diff
@@ -34,25 +34,14 @@
...
-typedef std::vector<int> intvec;
-typedef std::vector<bool> boolvec;
+#include <vector>
+#include "psi4/libmints/wavefunction.h"
+#include "psi4/liboptions/liboptions.h"
 
 namespace psi {
-
-class Options;
-class Wavefunction;
```
loriab, line 55: also here on forward decl

loriab (commented):
Besides the forward decl issue, this looks reasonable to me. These are FAE modules, though, so will seek input. Thanks for the modernization!

psi4/src/psi4/libmoinfo/model_space.h
TiborGY: Hmm, I have not thought of that to be honest. I made the change because I felt like it makes the code more readable, easier to see what a particular file depends on. But of course, I can revert if you feel strongly about the potential compile time impact of it.

loriab: I'd favor reverting to the forward decl. But no hurry if you want to wait until FAE chimes in.

fevangelista: I think forward declaration is better because it avoids including (expanding the content of the .h file) in all the files that also include `model_space.h`.

psi4/src/psi4/libmoinfo/moinfo_base.h
fevangelista: Agree.

fevangelista:
Thank you for making these improvements! This code was written before C++11 so it's really old style. I've left some comments here and there for things that need to be taken care.

psi4/src/psi4/libmoinfo/moinfo.h
```diff
@@ -74,7 +72,7 @@ class MOInfo : public MOInfoBase {
         const MOInfo* moinfo;
 
        public:
-        typedef std::bitset<size_det> bitdet;
+        typedef std::bitset<2048> bitdet;  // adjust based on "size det"?
```
fevangelista, line 75: Just to double check: was `size_det` used only here? It might be a good idea to make this a static constant expression in the class `SlaterDeterminant`.

psi4/src/psi4/libmoinfo/model_space.h
```diff
@@ -59,9 +57,6 @@ class ModelSpace {
     std::vector<int> unique_to_all;  // spin-unique  determinants
     MOInfo* moinfo_obj;
 };
-
-extern ModelSpace* model_space;
```
fevangelista, line 63: I agree, this should go, if possible.

psi4/src/psi4/libmoinfo/moinfo_base.cc
```diff
@@ -91,27 +90,23 @@ void MOInfoBase::compute_number_of_electrons() {
     nbel = nel - nael;
 }
 
-void MOInfoBase::compute_ioff() {
```
fevangelista, line 94: Thanks for removing this. It's probably faster to compute `ioff` directly than to read it off from memory.

psi4/src/psi4/libmoinfo/moinfo_scf.h
```diff
@@ -51,7 +48,7 @@ class MOInfoSCF : public MOInfoBase {
     void print_mo();
 };
 
-extern MOInfoSCF* moinfo_scf;
+extern MOInfoSCF* moinfo_scf;  // Unfortunate, but psi::mcscf assumes that this is declared.
```
fevangelista, line 51: Thanks for leaving the note. Ideally this would not be a global variable. But I am not sure it's worth the effort to fix this.

fevangelista (changes requested):

psi4/src/psi4/libmoinfo/model_space.h
TiborGY: Fair point, reverted.

psi4/src/psi4/libmoinfo/moinfo.h
TiborGY: Yes, this was the only hit for `size_det` in the entire repo. I have also searched for 2048 to check for things assuming this size, and found nothing. So having it "inline" like this seems like the most straightforward solution.

psi4/src/psi4/libmoinfo/moinfo_base.h
TiborGY: Reverted

psi4/src/psi4/libmoinfo/moinfo_scf.h
TiborGY: My thoughts exactly. Briefly looked into it, seemed painful to change.

loriab (approved):

fevangelista (approved):

PR #3117, opened 2024-01-04
Throw PSIEXCEPTION instead of exit(1) in DPD

TiborGY:
## Description

A lot of error exits in `libdpd` are implemented with `exit(1)` or equivalent. This makes debugging harder, and results in less informative error messages when a test fails in the CI environment.

This PR modernizes all `exit(1)` error exits in `libdpd` into `throw PSIEXCEPTION`.

- [x] Psi4 now prints more detailed error messages if an error happens in its DPD module

## Dev notes & details

- [x] All `exit(1)` and `exit(PSI_RETURN_FAILURE)` calls in `libdpd` have been replaced with a `throw PSIEXCEPTION`.

## Checklist
- [x] No new features
- [x] CI tests are failing only due to libint being in flux

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libdpd/block_matrix.cc
+++ b/psi4/src/psi4/libdpd/block_matrix.cc
@@ -50,7 +50,7 @@
 ** zero-priority entries remain), there is still insufficient memory
 ** available to satisfy the request, a nullptr pointer is returned to the
 ** caller, indicating that either an out-of-core algorithm must be
-** used, or the caller must exit().
+** used, or the caller must throw PSIEXCEPTION("").
 **
 ** TDC, 6/24/00
 */
@@ -64,6 +64,7 @@
 #include "psi4/libqt/qt.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
 #include "psi4/psi4-dec.h"
+#include "psi4/libpsi4util/exception.h"
 
 #include <cstdio>
 #include <cstdlib>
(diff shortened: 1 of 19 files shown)


susilehtola (approved):
Great change. However, I would propose eliminating the old print statements, since the same information is in the error message.

TiborGY:
> Great change. However, I would propose eliminating the old print statements, since the same information is in the error message.

I have considered doing that, but I am not entirely comfortable with it because `throw PSIEXCEPTION` is not perfectly reliable when it comes to delivering the error message to the user. For example see https://github.com/psi4/psi4/pull/2997.

Until a reliable solution is found for the error messages getting swallowed up on some platforms, I am reluctant to remove the traditional error messages. But in principle I agree completely.

TiborGY:
PS: In #2997 I added the band-aid workaround of printing to `stderr` for PSIO errors, since those are quite common due to disks getting full and whatnot, but doing that globally in `PSIEXCEPTION` itself seems inelegant.
Users who are not affected by the missing error message would get the entire stack trace, etc. twice, in addition to probably a bunch of noise from Python when it gets a C++-side exception...

susilehtola:
The cause of missing error messages is likely the improper use of `exit()`. I have prepared a companion pull request to this one to get rid of them.

TiborGY:
> The cause of missing error messages is likely the improper use of `exit()`. I have prepared a companion pull request to this one to get rid of them.

I think there are two separate issues here. One is the improper use of `exit()` scattered around the codebase, but I still think that there is a separate issue here where not even `throw PSIEXCEPTION(...)` is delivering the error message, in certain environments. I think that would still be a problem even if every instance of `exit()` were to be eradicated from the repo.

In https://github.com/psi4/psi4/issues/2970, the part of PSIO that was raising the error was already using `throw PSIEXCEPTION(...)`, it was not calling `exit()`. Unless removing all of the `exit()` calls elsewhere from the codebase magically changes how `PSIEXCEPTION` works, I cannot see how that would fix that.

loriab (approved):
lgtm, thanks for the cleanup!

PR #3118, opened 2024-01-04
Do not call exit

susilehtola:
## Description

As discussed in #2997 and #3117, no debugging information is produced in many places in Psi4. I believe this is because of improper calls to `exit()`, which is a remnant of the C history of Psi. #3117 replaces the calls to `exit()` with `throw`s in libdpd. This PR replaces the calls in the rest of Psi4.

Sorry for the verbose changes; I ran clang-format on the files since my editor's indentation doesn't match that in Psi4.

- [x] Obsoleted calls to `exit()` have been replaced by exceptions, which should better guarantee the passing of error messages to the end user.

## Dev notes & details

- [x] Spurious calls to `exit()` have been removed in favor of exceptions, which should better guarantee error messages getting through to the user.

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/dfocc/ccd_tpdm.cc
+++ b/psi4/src/psi4/dfocc/ccd_tpdm.cc
@@ -39,143 +39,530 @@ namespace dfoccwave {
 void DFOCC::ccd_tpdm() {
     timer_on("tpdm");
 
- // RHF
- if (reference_ == "RESTRICTED") {
-    SharedTensor2d T, U, Tau, G, G2, V, X, Y, Z;
-    SharedTensor2d Ts, Ta, Vs, Va, S, A;
+    // RHF
+    if (reference_ == "RESTRICTED") {
+        SharedTensor2d T, U, Tau, G, G2, V, X, Y, Z;
+        SharedTensor2d Ts, Ta, Vs, Va, S, A;
 
-    //============================
-    // OO-Block Correlation TPDM
-    //============================
+        //============================
+        // OO-Block Correlation TPDM
+        //============================
 
-    // G_ij^Q += P+(ij) V_ij^Q
-    G = std::make_shared<Tensor2d>("Correlation 3-Index TPDM (Q|IJ)", nQ, naoccA, naoccA);
-    V = std::make_shared<Tensor2d>("V (Q|IJ)", nQ, naoccA, naoccA);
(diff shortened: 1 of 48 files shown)


psi4/src/psi4/psimrcc/operation_compute.cc
```diff
@@ -205,8 +205,8 @@ void CCOperation::zero_target_block(int h) {
 void CCOperation::fail_to_compute() {
     outfile->Printf("\n\nSolve couldn't perform the operation ");
     print_operation();
-
-    exit(EXIT_FAILURE);
+    outfile->Flush();
```
JonathonMisiewicz, line 208: Why do you `flush` some times but not others? I assume (perhaps naively) that information would all get flushed before the program actually terminates, even without that line.

JonathonMisiewicz (commented):
One question before I approve

psi4/src/psi4/psimrcc/operation_compute.cc
susilehtola: I've found through experience that all information may not get flushed to output in C++ when an exception is thrown. I've added flushes wherever there's a danger of pertinent output not being printed out. Does Psi4 catch exceptions and flush the output there?

JonathonMisiewicz: Not that I've ever seen.

JonathonMisiewicz (approved):
Don't forget to resolve the merge conflict.

psi4/src/psi4/dfocc/ccd_tpdm.cc
```diff
@@ -39,534 +39,143 @@ namespace dfoccwave {
 void DFOCC::ccd_tpdm() {
     timer_on("tpdm");
 
- // RHF
```
loriab, line 42: Maybe revert all or just the non-throw changes from dfocc/ccd_tpdm.cc and dfocc/tensor.cc ? dfocc still needs some reconciliation with uncommitted changes, and all the spacing changes here are going to make that hard. sorry for the trouble. otherwise, lgtm!

susilehtola: Done. Yeah, stupidly I ran clang-format before committing the changes, and then realized what a mess it had made.

PR #3130, opened 2024-02-07
AM symbol list

loriab:
## Dev notes & details

- [x] remove P=1=12 from angmom list

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/conda/win/meta.yaml
+++ b/conda/win/meta.yaml
@@ -19,7 +19,7 @@ requirements:
     - msgpack-python
     - optking
     - gau2grid
-    - libint 2.7.3dev1
+    - libint
     - libxc-c
     - scipy
 
@@ -52,6 +52,7 @@ build:
 test:
   requires:
     - pytest-xdist
+    - pytest<8
     - dftd3-python
     - gcp-correction
   commands:
--- a/psi4/driver/p4util/util.py
+++ b/psi4/driver/p4util/util.py
@@ -359,7 +359,7 @@ def _l2_config_style_eri_llll():
     skel = {"onebody_": [], "eri_c4_": [], "eri_c3_": [], "eri_c2_": []}
     skel_re = {"onebody_": r"onebody_\w_d\d", "eri_c4_": r"eri_\w..._d\d", "eri_c3_":  r"eri_\w.._d\d", "eri_c2_": r"eri_\w._d\d"}
 
-    amstr = "SPDFGHIKLMNOPQRTUVWXYZ"
+    amstr = "SPDFGHIKLMNOQRTUVWXYZ"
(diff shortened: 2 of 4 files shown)


JonathonMisiewicz (approved):
I really want to know how you stumbled onto this.

loriab:
> I really want to know how you stumbled onto this.

:-) Yeah, so "superbigam" is now 7-7-4-12-7-5, and when I built that in Dec for 2.8.0 I only set the 12 for eri3, forgetting my own advice to make eri2 equal eri3. So rebuilding for 2.8.2, I knew I had to add symbols to the configuration.cc patch for eri2. Started searching patterns and realized I had P twice in the spdfg string. Then realized libmints (where I thought I had copied it from) had the same trouble. Then realized that libint had a mix of troubles.

loriab:
> Uhm, this is not Gaussian'94 format, which doesn't skip J.

Psi4 is "HIK" convention: https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/basissets.rst?plain=1#L95 and the change is only for AM>=12. Am I missing something?

susilehtola:
> > Uhm, this is not Gaussian'94 format, which doesn't skip J.
> 
> Psi4 is "HIK" convention: https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/basissets.rst?plain=1#L95 and the change is only for AM>=12. Am I missing something?

It's not related to this change but the skipping of `J` in general.

loriab:
> Uhm, this is not Gaussian'94 format, which doesn't skip J.

Ok, so this is just a general observation? The immediate use of the amchar string is to decrypt the cmake component strings coming from L2, so it's worth following it's conventions. https://github.com/evaleev/libint/pull/325

I think this is good to know, unless I'm not following your point.

susilehtola:
> > Uhm, this is not Gaussian'94 format, which doesn't skip J.
> 
> Ok, so this is just a general observation? The immediate use of the amchar string is to decrypt the cmake component strings coming from L2, so it's worth following it's conventions. [evaleev/libint#325](https://github.com/evaleev/libint/pull/325)

Is there a reason for them not being numeric? This would fully eliminate any ambiguities...

loriab:
> Is there a reason for them not being numeric? This would fully eliminate any ambiguities...

It was fully numeric for a while (e.g., `eri_c4_l10_d0`), but then to more concisely express puream shells and the independent paired/unpaired AMs for eri3, EFV preferred symbolic, https://github.com/evaleev/libint/pull/271#discussion_r1359984476 . So then the bit in p4util/util.py is translating back to numeric for clarity in the psi4 output file.

jturney (approved):

PR #3131, opened 2024-02-07
Avoid Disk algorithm of MemDF

loriab:
## Description

Collect SAPT changes in the progress of SAPT consolidation.

- [x] avoid stumbling into memdf with disk algorithem

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/lib3index/dfhelper.cc
+++ b/psi4/src/psi4/lib3index/dfhelper.cc
@@ -67,6 +67,12 @@ DFHelper::DFHelper(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> 
     if(Process::environment.options["SCF_SUBTYPE"].has_changed()) {
         subalgo_ = Process::environment.options.get_str("SCF_SUBTYPE");
     }
+    else {
+        // Default to in-core for MemDF b/c out-of-core is so slow we don't want it to be hit w/o explicit user SCF_SUBTYPE.
+        // By setting this, checks below will crash if memory is insufficient.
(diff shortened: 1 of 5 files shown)


davpoolechem (approved):

loriab:
@davpoolechem, tests/python/3-index-transforms/input.py isn't running correctly. Do you happen to understand how to edit it while still preserving the intent of the test?

```
Fatal Error: SCF_SUBTYPE=INCORE was specified, but there is not enough memory to do in-core! Increase the amount of memory allocated to Psi4 or allow for out-of-core to be used.
```

davpoolechem:
> @davpoolechem, tests/python/3-index-transforms/input.py isn't running correctly. Do you happen to understand how to edit it while still preserving the intent of the test?
> 
> ```
> Fatal Error: SCF_SUBTYPE=INCORE was specified, but there is not enough memory to do in-core! Increase the amount of memory allocated to Psi4 or allow for out-of-core to be used.
> ```

After some quick tinkering, I was fix the issue by adding the line

`psi4.set_options({ "scf_subtype": "auto" })`

under the definition of `mol`. Setting `SCF_SUBTYPE` explicitly like this circumvents the switch to the `INCORE` MemDF algorithm in the DFHelper constructor that is causing the exception. 

Cross-checking against the test output from another Psi4 build leads to basically no difference in the output files, specifically in MemDF header printout, so I believe the intent of the test is also preserved (i.e., there are no alterations into what configurations of MemDF are tested).

JonathonMisiewicz (approved):

PR #3136, opened 2024-02-23
openorbitaloptimizer

loriab:
## User API & Changelog headlines

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist
- [ ] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [ ] Ready for review
- [ ] Ready for merge

PR #3138, opened 2024-02-24
Orbitals should be multiplied by sqrt(occ) not occ

### LINKED ISSUE #3137: Potential bug in sad.cc: definition of pseudo-orbitals
As a result of in-person collaboration on #3136 with @loriab in Lausanne, I noticed a bug in the SAD code:
https://github.com/psi4/psi4/blob/9dbdd59dd474947d64522508476ffba1fb6b59c8/psi4/src/psi4/libscf_solver/sad.cc#L716

The pseudo-orbitals should be obtained by multiplying the orbital with **the square root of** their occupation to reproduce the correct density matrix  C occ C^T.

susilehtola:
## Description

Fixes a bug in `sad.cc`:  pseudo-orbitals should be obtained by multiplying the orbital with the square root of their occupation to reproduce the correct density matrix  C occ C^T.

Closes #3137.

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


TiborGY (commented):
Would it be possible to add a test that fails before but passes after this change? Clearly there is a testing gap, given how long this bug has been present, and how something like this can be changed without having to change any of tests.

susilehtola:
Yes, by testing the guess energy for H atom.

davpoolechem (approved):

susilehtola:
Looks like there are some unrelated test failures, but this one is ready to go. As shown by the changes in the pytest guess energies, the new code reaches a considerably lower SAD guess energy for the HF molecule: -100.02909951427 vs -99.63941801281894. Large improvements are expected especially for systems containing hydrogen atoms, since the existing code only includes 0.5 electrons on those atoms.

jturney (approved):

tests/pytests/test_scf_options.py
```diff
@@ -67,31 +67,31 @@ def test_guess_mix_for_broken_symmetry(inp):
...
-            "core": -88.99202896495355,
-            "gwh": -94.35585144644561,
-            "sad": -99.63941801281894,
-            "sadno": -99.11882081641045,
-            "huckel": -99.20374047108815,
-            "modhuckel": -99.17689049002892,
-            "sap": -99.50144270485751,
+            "core": -88.99202896495,
+            "gwh": -94.35585144645,
+            "sad": -99.34836107817,
```
loriab, line 81: Is it a concern that with the change SAD UHF & ROHF (and often sadno, *huckel) guess energies are _less_ deep?

susilehtola: No, the issue is that the test here is FH+, which has orbitals that are more compact than the neutral FH. It is logical that the proper guess for the neutral molecule yields a slightly higher energy for the cation.

Like I mention above, right now the SAD code is placing $n_i^2$ electrons on each orbital that should have a fractional occupation of $n_i$. For hydrogen this means that currently the guess has `0.5` electrons for the hydrogen atom, whereas fluorine probably has 7 electrons distributed evenly on 4 spatial orbitals [no symmetry, this is the key feature in the OpenOrbitalOptimizer guess in #3136] which means that its number of electrons is also wrong by `-0.875`.  Therefore, the current guess is spuriously good for FH+ as it corresponds to a cation with charge 1.375.

andyj10224 (approved):
LGTM! Anything holding this PR up from merging?

tests/nbody-cp-gradient/input.dat
```diff
@@ -210,14 +210,14 @@ for result in [  #TEST
...
-    compare_values(ref_2cp_grad, result, 8, 'rtd=F CP-Corrected Gradient')  #TEST
+    compare_values(ref_2cp_grad, result, 7, 'rtd=F CP-Corrected Gradient')  #TEST
 
 for result in [  #TEST
     g,  #TEST
     wfn.variable("CURRENT GRADIENT"),  #TEST
     variable("CURRENT GRADIENT"),  #TEST
 ]:  #TEST
-    compare_values(ref_2cp_grad_ie, result, 8, 'rtd=F CP-Corrected Gradient')  #TEST
+    compare_values(ref_2cp_grad_ie, result, 7, 'rtd=F CP-Corrected Gradient')  #TEST
```
loriab, line 220: I'd really rather these stayed at 8, but #3160 is going to change up this test a lot, so I'll defer recomputing the refs until then. I don't know why the snowflake version was failing (1.3e-8) while internal was passing. That's a little troublesome, too.

loriab (approved):
few comments, but let's merge this while CI is happy.

Issue #3141, opened 2024-03-01
SAPT(DFT) might forget about freezing core


konpat:
When one runs SAPT0 via the SAPT(DFT) pathway with the following options:
```
set SAPT_DFT_FUNCTIONAL HF
set SAPT_DFT_MP2_DISP_ALG FISAPT
energy('sapt(dft)')
```
the algorithm forgets to freeze the core in the exchange-dispersion term. This does not happen when `SAPT_DFT_MP2_DISP_ALG == 'FISAPT'`.

I'll submit a PR with the fix very soon.

PR #3142, opened 2024-03-01
Pass frozen-core information from SAPT(DFT) to FISAPT dispersion

konpat:
## Description
When SAPT(DFT) is used with `SAPT_DFT_FUNCTIONAL == 'HF'` to run SAPT0, two pathways for dispersion and exchange-dispersion calculations are possible, through the SAPT0-SAPT2+3 code (requested by `set SAPT_DFT_MP2_DISP_ALG SAPT`) or through the FISAPT code (`set SAPT_DFT_MP2_DISP_ALG FISAPT`). However, in the latter case, the calculation forgot to freeze the core as requested. This PR should preserve frozen-core in this specific situation.

This fixes https://github.com/psi4/psi4/issues/3141. It's quite likely that the FISAPT pathway can be faster (thanks to @carolinesargent for extensive testing), that's why we want to fix it.

## Questions
- [ ] I assumed no frozen core is ever needed for true SAPT(DFT). Is that valid? 

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/sapt/sapt_mp2_terms.py
+++ b/psi4/driver/procrouting/sapt/sapt_mp2_terms.py
@@ -271,11 +271,11 @@ def df_mp2_fisapt_dispersion(wfn, primary, auxiliary, cache, nfrozen_A, nfrozen_
 
     # If frozen core, trim the appropriate matrices and vectors. We can do it with NumPy slicing.
     if nfrozen_A > 0:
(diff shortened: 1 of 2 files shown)


loriab (approved):
lgtm, thanks!

I agree that there's no need for fc for sapt(dft).

jturney (approved):

PR #3143, opened 2024-03-07
Add creation of Eigen::Map representations of Psi4 matrices

davpoolechem:
## Description
This is something I have been working on as part of the upcoming Psi4-GauXC interface; but it seems to be of interest for other use cases, so I will be adding it here as a separate PR.

What this change does, is it adds two new member functions to the `Psi4::Matrix`  class, `eigen_map()`  and `eigen_maps()` . Both of these functions serve the same purpose - take the Psi4 matrix in question and return a representation of said matrix through Eigen's `Map` objects. The `Eigen::Map` class acts the same way as a normal `Eigen::Matrix` object, but uses an external array as its data buffer rather than an internal data buffer. The `eigen_map()` and `eigen_maps()` functions, then, return Eigen matrix-like objects that directly link with the Psi4 matrix data buffer of the Psi4 matrix for which the function was called. This provides an efficient fashion by which to utilize Psi4 matrices in contexts which require Eigen constructs, with no data deepcopying or Eigen-to-Psi4 back-conversions required.

There is a key difference between the two functions. `eigen_map()`  assumes that the matrix has a single irrep, and returns a single `Eigen::Map` object. `eigen_maps()` , on the other hand, is used for matrices with multiple irreps, and returns a `std::vector` of `Eigen::Map` objects, each `Map` in the `vector` corresponding to one irrep of the Psi4 matrix.

- [X] Adds two new functions, `eigen_map()` and `eigen_maps()`, usable in Psi4 plugins and downstream programs, that return  a formulation of the calling `Psi::Matrix` object, that is usable in contexts where `Eigen::Matrix` objects are required/desired.

## Dev notes & details

- [X] Adds two new member functions to `Psi4::Matrix`, `eigen_map()` and `eigen_maps()`, that return `Eigen::Map` representations of the Psi4 matrix object.

## Questions
- [X] N/A

## Notes
- [X] I did look into forward declaring the necessary Eigen classes, but forward declaring `Eigen::Map` turns out to be a bit ugly, as it is a class template that, to forward declare, ends up requiring forward declarations for other Eigen classes and enums (such as `Stride` and `AlignmentType`) that aren't specifically necessary for the Psi4 use case. Therefore, I have skipped forward declarations here.
- [X] I can confirm the correctness of the `eigen_map()` function, as I have implemented these PR changes into my Psi-GauXC interface, wherein `eigen_map()` replaces the original formulation I used for the equivalent result. `eigen_maps()`, however, is currently untested as there is no current use case for it yet, and should thus be considered experimental. 

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -438,7 +438,7 @@ std::vector<Eigen::Map<Eigen::MatrixXd>> Matrix::eigen_maps() {
         eigen_maps.emplace_back(get_pointer(h), rowdim(h), coldim(h));
     } 
     
-    return std::move(eigen_maps);
+    return eigen_maps;
 }
 
 void Matrix::alloc() {
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -404,6 +404,41 @@ void Matrix::copy(const Matrix &cp) { copy(&cp); }
 
 void Matrix::copy(const SharedMatrix &cp) { copy(cp.get()); }
 
+// produces an Eigen::Map object mapping to the matrix data buffer 
+// of a matrix with a single irrep 
+Eigen::Map<Eigen::MatrixXd> Matrix::eigen_map() {
+    // Sanity checks
+    // only works for C1 symmetry at the moment
+    // could be improved if symmetry is utilized in JK builds in the future
+    if (nirrep() != 1) {
+        throw PSIEXCEPTION("Psi4::Matrix must be in C1 symmetry to be transformed into Eigen::MatrixXd!");
+    }
+
+    // create Eigen matrix "map" using Psi4 matrix data array directly
+    return std::move(
+        Eigen::Map<Eigen::MatrixXd>(
+            get_pointer(), nrow(), ncol()
+        )
+    );
+}
+
+// produces Eigen::Maps object mapping to the matrix data buffer
+// of a matrix with multiple irreps 
+// NOTE: this impl for mapping Psi4 matrices to Eigen maps
+// is currently experimental, as it is unused in the code 
+// currently
+std::vector<Eigen::Map<Eigen::MatrixXd>> Matrix::eigen_maps() {
+    std::vector<Eigen::Map<Eigen::MatrixXd>> eigen_maps;
+    eigen_maps.reserve(nirrep());
+
+    for (int h = 0; h != nirrep(); ++h) {
+        eigen_maps.emplace_back(get_pointer(h), rowdim(h), coldim(h));
+    } 
+    
+    // create Eigen matrix "map" using Psi4 matrix data array directly
+    return std::move(eigen_maps);
+}
+
 void Matrix::alloc() {
     if (matrix_) release();
 
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -431,11 +431,12 @@ std::vector<Eigen::Map<Eigen::MatrixXd>> Matrix::eigen_maps() {
     std::vector<Eigen::Map<Eigen::MatrixXd>> eigen_maps;
     eigen_maps.reserve(nirrep());
 
+    // create Eigen matrix "map"s for each irrep, 
+    // using Psi4 matrix data array directly
     for (int h = 0; h != nirrep(); ++h) {
         eigen_maps.emplace_back(get_pointer(h), rowdim(h), coldim(h));
     } 
     
-    // create Eigen matrix "map" using Psi4 matrix data array directly
     return std::move(eigen_maps);
 }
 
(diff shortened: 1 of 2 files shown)


TiborGY:
> ## User API & Changelog headlines
> 
>     * [x]  N/A

The entire `psi::Matrix` class is marked as `PSI_API`, so I think adding any public member functions amounts to a change in API.

davpoolechem:
> > ## User API & Changelog headlines
> > ```
> > * [x]  N/A
> > ```
> 
> The entire `psi::Matrix` class is marked as `PSI_API`, so I think adding any public member functions amounts to a change in API.

The reason I didn't mark anything here as an API change is because neither of the new `eigen_map` functions are added to the list of exported `Psi4::Matrix` functions in `export_mints.cc`, which means that they cannot be utilized in the PsiAPI or  Psithon input modes. Which is fine, because you'd probably prefer the `Psi4::Matrix` as a NumPy array in those contexts anyway. 

The `PSI_API` flag makes the symbols for these functions visible in the Psi4 shared object library, but I'm not sure if that alone constitutes a User API change. I suppose it could count for plugin developers, as it is a new functionality within Psi4 that they can utilize in their plugins.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -404,6 +404,43 @@ void Matrix::copy(const Matrix &cp) { copy(&cp); }
...
+    std::vector<Eigen::Map<Eigen::MatrixXd>> eigen_maps;
+    eigen_maps.reserve(nirrep());
+
+    // create Eigen matrix "map"s for each irrep, 
+    // using Psi4 matrix data array directly
+    for (int h = 0; h != nirrep(); ++h) {
+        eigen_maps.emplace_back(get_pointer(h), rowdim(h), coldim(h));
+    } 
+    
+    return std::move(eigen_maps);
```
JonathonMisiewicz, line 441: Is this necessary?

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -404,6 +404,43 @@ void Matrix::copy(const Matrix &cp) { copy(&cp); }
...
+}
+
+// produces Eigen::Maps object mapping to the matrix data buffer
+// of a matrix with multiple irreps 
+// NOTE: this impl for mapping Psi4 matrices to Eigen maps
+// is currently experimental, as it is unused in the code 
+// currently
+std::vector<Eigen::Map<Eigen::MatrixXd>> Matrix::eigen_maps() {
+    std::vector<Eigen::Map<Eigen::MatrixXd>> eigen_maps;
+    eigen_maps.reserve(nirrep());
```
JonathonMisiewicz, line 433: Why not pass `nirrep()` to the constructor on the line above.

davpoolechem: `Eigen::Map<Eigen::MatrixXd>` doesn't seem have a constructor definition for the case with no input arguments, i.e., `Eigen::Map<Eigen::MatrixXd>()`. Trying to do 

```std::vector<Eigen::Map<Eigen::MatrixXd>>(nirrep())``` 

will attempt to create a series of `Eigen::Map` objects using that constructor which, because it's undefined, causes compilation to fail. 

The way I implemented it gets around that by using `emplace_back` to construct each `Eigen::Map` in place, with the proper arguments, and add it to the overall `std::vector` upon construction.

File: psi4/src/psi4/libmints/matrix.cc (hunk 1 above)
Comment (davpoolechem) [reply psi4/src/psi4/libmints/matrix.cc:441]: It's probably not "necessary", strictly speaking, but I did this on the understanding that doing a `std::move` on the output vector would be more efficient due to avoiding unnecessary object copying/duplication.

Comment (JonathonMisiewicz) [reply psi4/src/psi4/libmints/matrix.cc:441]: Wouldn't return value optimization cover this?

Comment (davpoolechem) [reply psi4/src/psi4/libmints/matrix.cc:441]: Yeah, when I double-checked my assumption online, it does seem that `std::move` can hinder RVO. So with that in mind, I'll get rid of the move.

**Review by JonathonMisiewicz — APPROVED** [review]

File: psi4/src/psi4/libmints/matrix.cc (hunk 1 above)
Comment (davpoolechem) [reply psi4/src/psi4/libmints/matrix.cc]: Done and done!

**Review by loriab — APPROVED** [review]
lgtm, thanks!

I agree that this is an API change on the plugin side, and thanks for pointing that out @TiborGY . It's less disruptive than some b/c it's an addition. And we've never achieved API stability on the c++ side; plugins builds just have to pin to a particular psi4 bld.

**Comment by davpoolechem** [comment]
All right, added a note about the relevant API change as introduced by this PR!

File: psi4/src/psi4/libmints/matrix.h [hunk 3]
```diff
@@ -36,6 +36,8 @@
 
 #include "dimension.h"
 
+#include <eigen3/Eigen/Core>
+
```
susilehtola, line 40: Isn't this a new requirement? I don't think Eigen3 has been used in Psi4 previously...

loriab: Right, I did mean to mention this. Correct that Psi4 doesn't use Eigen3 itself. But Eigen3 and boost header-only libs are needed to compile against L2 header-only, so it isn't a net new requirement.

PR #3145, opened 2024-03-07
Improve Compiler Portability Through "Pedantic" Compilation

davpoolechem:
## Description

This PR does two things:

1. Adds a new compile-time option, `FORCE_PEDANTIC`, that adds the `-pedantic-errors` flag to the compilation flag list when enabled. The `-pedantic-errors` flag has the compilation fail any time the code does not strictly follow the C++ language standard, i.e., whenever a compiler-specific language extension is used. `FORCE_PEDANTIC` is currently set to be enabled by default. 
2. Fixes any instances of the code where the above option/flag would cause a compilation failure. This only requires two changes currently - removing extra semicolons in `libdiis/diismanager.h`, and replacing use of variable-length C-style arrays in `libmints/extern.cc`.

The goal of these changes is to improve code portability across different compilers. By effectively disabling compiler-specific language extensions, the code no longer relies on specific compiler functionalities to work, and therefore should be compatible with a larger range of compilers across a wider variety of systems.

- [X] N/A

## Dev notes & details

- [X] Adds a new compile-time setting, `FORCE_PEDANTIC`, that causes compilation to fail whenever use of a compiler-specific extension is detected.
- [X] Removes extraneous semicolons in `libddis/diismanager.h` that led to compiler error with the `FORCE_PEDANTIC` option enabled.
- [X] Replaces the use of variable-length C-style arrays in `libmints/extern.cc` with `std::vector` constructs.

## Questions
- [X] Should this new option be on by default?

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge


susilehtola (approved):

loriab:
looks like it found another one 

```
[  5%] Building CXX object src/psi4/libdiis/CMakeFiles/diis.dir/diismanager.cc.o
cd /home/vsts/work/1/s/build/psi4-core-prefix/src/psi4-core-build/src/psi4/libdiis && /usr/bin/g++-10 -DUSING_LAPACK_MKL -I/home/vsts/work/1/s/psi4/include -I/home/vsts/work/1/s/psi4/src -I/usr/share/miniconda/envs/p4env/include/python3.8 -isystem /usr/share/miniconda/envs/p4env/include -march=native -pedantic-errors -O3 -DNDEBUG -std=c++17 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -fopenmp -MD -MT src/psi4/libdiis/CMakeFiles/diis.dir/diismanager.cc.o -MF CMakeFiles/diis.dir/diismanager.cc.o.d -o CMakeFiles/diis.dir/diismanager.cc.o -c /home/vsts/work/1/s/psi4/src/psi4/libdiis/diismanager.cc
In file included from /home/vsts/work/1/s/psi4/src/psi4/libdiis/diismanager.h:38,
                 from /home/vsts/work/1/s/psi4/src/psi4/libdiis/diismanager.cc:29:
/home/vsts/work/1/s/psi4/include/psi4/pybind11.h:46:52: error: extra ‘;’ [-Wpedantic]
   46 | PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>);
      |                                                    ^
make[5]: *** [src/psi4/libdiis/CMakeFiles/diis.dir/build.make:76: src/psi4/libdiis/CMakeFiles/diis.dir/diismanager.cc.o] Error 1
make[5]: Leaving directory '/home/vsts/work/1/s/build/psi4-core-prefix/src/psi4-core-build'
```

davpoolechem:
Thanks for the heads-up @loriab! Weird that it didn't show up when I compiled it, but since this one seems to be outside the `psi4/src` dir, perhaps that has something to do with it. Regardless, the issue should be fixed now.

davpoolechem:
So there's another CI issue. It'll be easy to fix, but it's really curious that these issues show up in CI, but not when I compile it myself. Regardless, I will work through all CI issues as they pop up!

loriab:
np, that's why it's nice to have a variety of compilers in CI. thanks for the fixes.

loriab (approved):
quite a lot of lonesome semicolons you found

loriab:
uh-oh, it's finding pedantic errors in integratorxx now `/include/integratorxx/generators/s2_factory.hpp:24:2: error: extra ‘;’ [-Wpedantic]` . this would be a fault of our appending to `CMAKE_CXX_FLAGS`, which wasn't so outrageous back in the day.

thinking about it, there would be a lot more of these interferences with other projects if the CI wasn't detecting so much as pre-built. which means that developers building deps from source will likely have builds fail since force_pydantic defualts to ON and if the dep projects aren't pedantic clean. so this was probably a lucky block to the merge.

I'm not sure right off if there's a quick+good solution. you could move the option to here https://github.com/psi4/psi4/blob/master/CMakeLists.txt#L222 _after_ the externals have been added. and maybe turn the default to OFF but add it (`-D`) as ON to the azure CI. I'm going to put a needs-changes block so I remember it isn't just a spurious test fail in the merge queue. lmk if you disagree with any of my hasty conclusions.

loriab (changes requested):
see https://github.com/psi4/psi4/pull/3145#issuecomment-2015930547

davpoolechem:
> uh-oh, it's finding pedantic errors in integratorxx now `/include/integratorxx/generators/s2_factory.hpp:24:2: error: extra ‘;’ [-Wpedantic]` . this would be a fault of our appending to `CMAKE_CXX_FLAGS`, which wasn't so outrageous back in the day.
> 
> thinking about it, there would be a lot more of these interferences with other projects if the CI wasn't detecting so much as pre-built. which means that developers building deps from source will likely have builds fail since force_pydantic defualts to ON and if the dep projects aren't pedantic clean. so this was probably a lucky block to the merge.
> 
> I'm not sure right off if there's a quick+good solution. you could move the option to here https://github.com/psi4/psi4/blob/master/CMakeLists.txt#L222 _after_ the externals have been added. and maybe turn the default to OFF but add it (`-D`) as ON to the azure CI. I'm going to put a needs-changes block so I remember it isn't just a spurious test fail in the merge queue. lmk if you disagree with any of my hasty conclusions.

Yeah, I feel like I ran into this issue previously with Einsums. It's part of why I was so concerned about this PR in the first place, and whether `FORCE_PEDANTIC` should be on by default. I can definitely toy around and see if I can get the pedantic flag to only apply to Psi-side code, like you suggested. Worst-case scenario, we just turn `FORCE_PEDANTIC` off by default?

davpoolechem:
All right, I moved the location of the `FORCE_PEDANTIC` definition, as per the suggestion of @loriab. This seems to work, as when I compile a local instance of this Psi4 branch with an internally-built IntegratorXX, it now compiles, whereas it did not previously.

loriab:
The Mac Eco error is probably a fluke -- I don't see any recent changes to dftd4-python. What do you think of defaulting this to OFF for the protection of plugins? I don't quite understand how they're behaving now with this. See below that it's warning but not erroring. I do think the cmake is slightly in error in that i'm not setting CXX_EXTENSIONS=OFF in the inner psi4/CMakeLists.txt


```
2024-03-25T14:45:09.8381208Z [67/68] Building Fortran object CMakeFiles/v2rdm_casscf.dir/focas_interface.F90.o
2024-03-25T14:45:09.8382362Z /home/runner/work/psi4/psi4/objdir/external/downstream/v2rdm_casscf/v2rdm_casscf_external-prefix/src/v2rdm_casscf_external/focas_interface.F90:274:53:
2024-03-25T14:45:09.8383299Z 
2024-03-25T14:45:09.8383512Z   274 |         max_dim = max(size(integrals_1,dim=1),nnz_d1,maxval(nnz_int))
2024-03-25T14:45:09.8383966Z       |                                                     1
2024-03-25T14:45:09.8384366Z Warning: GNU Extension: Different type kinds at (1)
2024-03-25T14:45:09.8385389Z /home/runner/work/psi4/psi4/objdir/external/downstream/v2rdm_casscf/v2rdm_casscf_external-prefix/src/v2rdm_casscf_external/focas_interface.F90:161:46:
2024-03-25T14:45:09.8386144Z 
2024-03-25T14:45:09.8386330Z   161 |         max_dim = max(size(integrals_1,dim=1),maxval(nnz_int))
2024-03-25T14:45:09.8386747Z       |                                              1
2024-03-25T14:45:09.8387135Z Warning: GNU Extension: Different type kinds at (1)
2024-03-25T14:45:09.8387546Z [68/68] Linking CXX shared module v2rdm_casscf.so
2024-03-25T14:45:09.8404861Z [17/19] Performing install step for 'v2rdm_casscf_external'
```

davpoolechem:
I'm fine defaulting it to `OFF`, to be honest! Probably safer that way, in regards to unforeseen errors/issues.

loriab:
> I'm fine defaulting it to `OFF`, to be honest! Probably safer that way, in regards to unforeseen errors/issues.

sounds good. I think it's worth turning on in azure, if that wasn't already in your plans.

loriab:
maybe a bkgd dependency like clang version changed such that mac isn't getting dftd4 3.6 but an oldr 3.3 hence the errors.

loriab (approved):

davpoolechem:
And here, we learn why it's important to be careful when force-pushing commits after a rebase. CI should be okay after this, though?

PR #3148, opened 2024-03-21
Fix initial cartesian units

timostrunk:
## Description

This PR fixes #3147 . Basically the psithon way of defining molecules converted _initial_cartesian on parsing.
When adding an external potential this was used as
molecule.set_geometry(clone_molecule._initial_cartesian).
set_geometry applies the conversion again.

The two python ways I found, which set initial_cartesian did not do the conversion leading to incorrect geometries, when using external_potential and molecules defined as unit=angstrom.

## User API & Changelog headlines

- [x] Fixes #3147: Parsing molecules with angstrom units and using external_potential from python was broken

## Dev notes & details

- [x] Fixes #3147: Parsing molecules from psithon and python left mol._initial_cartesian in a different unit. This aligns the units. Downstream this had the effect of breaking using molecules with angstrom units with an external_potential from python

## Questions
- [x] Question1

## Checklist
- [x] Tests added for any new features
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #3149, opened 2024-03-21
External hamiltonian

timostrunk:
## Description
It would be nice to be able to pass in an external hamiltonian to be able to test-drive some code (e.g. for the calculation of external potentials) without always needing to recompile. This PR adds this possibilty.

I am testing it by making a calculation without an external hamiltonian but with a potential (and saving the nuclear_repulsion addition, which is not mediated by the additional H term) and comparing it to a simulation without an external potential, but with the captured potential hamiltonian.

FYI: This is on top of #3148 , because otherwise the tests don't pass.
Also fyi: I did not build the docs.

- [ ] Added possibility to pass in an external hamiltonian contribution via `energy(..., external_hamiltonian = np.array(...))`.

## Dev notes & details

- [ ] Added possibility to pass in an external hamiltonian contribution via `energy(..., external_hamiltonian = np.array(...))`.
- [ ] Tested via simulation with external_potential and an equivalent external_hamiltonian

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge


jturney:
This seems like a neat ability.

PR #3153, opened 2024-04-08
Modernize relativistic potential

JonathonMisiewicz:
## Description
Paying off an old IOU. I'll get to what Francesco and Daniel Nascimento need after this PR is in.

- [x] `RelPotentialInt` now has the same charge field API as `PotentialInt` and should be faster

## Dev notes & details

- [x] `RelPotentialInt` now delegates construction of W0 integrals to Libint
- [x] `RelPotentialInt` now has the same charge field API as `PotentialInt` 

## Checklist
- [x] `ctest -L x2c` passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/rel_potential.cc
+++ b/psi4/src/psi4/libmints/rel_potential.cc
@@ -31,11 +31,9 @@
 #include "psi4/libmints/integral.h"
 #include "psi4/libmints/matrix.h"
 #include "psi4/libmints/molecule.h"
+#include "psi4/libpsi4util/PsiOutStream.h"
 
-#include <libint2/shell.h>
 #include <libint2/engine.h>
-
-#include <algorithm>
 
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 
@@ -44,7 +42,6 @@
 ;
 using namespace psi;
 
-// Initialize potential_recur_ to +1 basis set angular momentum
 RelPotentialInt::RelPotentialInt(std::vector<SphericalTransform>& st, std::shared_ptr<BasisSet> bs1,
                                  std::shared_ptr<BasisSet> bs2, int deriv)
     : OneBodyAOInt(st, bs1, bs2, deriv) {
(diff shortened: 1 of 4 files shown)


psi4/src/psi4/libmints/rel_potential.cc
```diff
@@ -31,71 +31,55 @@
...
 
 ;
 using namespace psi;
 
-// Initialize potential_recur_ to +1 basis set angular momentum
 RelPotentialInt::RelPotentialInt(std::vector<SphericalTransform>& st, std::shared_ptr<BasisSet> bs1,
                                  std::shared_ptr<BasisSet> bs2, int deriv)
     : OneBodyAOInt(st, bs1, bs2, deriv) {
     if (deriv > 0) {
         throw PSIEXCEPTION("RelPotentialInt: deriv > 0 is not supported.");
```
fevangelista, line 49: Just for curiosity: will be there be native support for derivatives with this and near-future set of changes? I am asking because one of the reasons to implement the integrals natively in libint2 was to make this possible.

psi4/src/psi4/libmints/rel_potential.cc
```diff
@@ -31,71 +31,55 @@
...
     }
 
+    if (bs1 != bs2) {
```
fevangelista, line 52: Thank you for pointing this out. I am wondering if we should throw and exception here since this is serious stuff.

fevangelista (approved):

psi4/src/psi4/libmints/rel_potential.cc
JonathonMisiewicz: It's already possible in L2. I'd just need to double-check that the L2 version Psi uses has those derivatives, and then double-check how the derivatives interface with the fact that this is a multi-component operator.

psi4/src/psi4/libmints/rel_potential.cc
JonathonMisiewicz: I just copy-pasted from `potential.cc`.

fevangelista: Thanks for checking. All looks good to me!

psi4/src/psi4/libmints/rel_potential.cc
JonathonMisiewicz: Ah. Yes, this PR _requires_ 2.8.0. How do I enforce that dependency?

loriab: * update https://github.com/psi4/psi4/blob/master/external/upstream/libint2/CMakeLists.txt#L38
* comment out COMPONENTS and OPTIONAL_COMPONENTS below it
* same here https://github.com/psi4/psi4/blob/master/psi4/CMakeLists.txt#L180-L191

JonathonMisiewicz: Does that mean commenting out the parts in all caps, or also the lines under them?

loriab: lines under them, too. upstream L2 doesn't export any components.

jturney (approved):

PR #3168, opened 2024-05-21
Throw PSIEXCEPTION instead of calling abort() in detci

TiborGY:
## Description

Detci still has a few places where error termination is effected through printf+abort(). This PR replaces the calls to abort() in with `throw PSIEXCEPTION(msg)`.

The rest of the codebase has more of this pattern, those are subject to future PRs to keep this one small.

- [x] A few error terminations in the detci module now print a more detailed error information to the console, instead of just the error message in the output file.

## Dev notes & details

- [x] Calls to `abort()` in detci are replaced with `throw PSIEXCEPTION(msg)`, where `msg` is the relevant error message taken from the context of the `abort()`

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/detci/import_vector.cc
+++ b/psi4/src/psi4/detci/import_vector.cc
@@ -82,15 +82,15 @@ void CIWavefunction::parse_import_vector(SlaterDetSet *sdset, int *ialplist, int
 
     /* now figure out how the frozen stuff is going to map */
     if (CalcInfo_->num_drc_orbs > alphastrings->ndrc || CalcInfo_->num_drc_orbs > betastrings->ndrc) {
-        outfile->Printf(
-            "(parse_import_vector): Can't drop more orbitals now"
-            " than in the imported guess!\n");
-        abort();
+        const std::string msg = "(parse_import_vector): Can't drop more orbitals now than in the imported guess!\n";
+        outfile->Printf(msg.c_str());
+        throw PSIEXCEPTION(msg);
     }
 
     if (alphastrings->ndrc != betastrings->ndrc) {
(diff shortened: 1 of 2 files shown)


psi4/src/psi4/detci/import_vector.cc
```diff
@@ -82,15 +82,15 @@ void CIWavefunction::parse_import_vector(SlaterDetSet *sdset, int *ialplist, int
 
     /* now figure out how the frozen stuff is going to map */
     if (CalcInfo_->num_drc_orbs > alphastrings->ndrc || CalcInfo_->num_drc_orbs > betastrings->ndrc) {
-        outfile->Printf(
-            "(parse_import_vector): Can't drop more orbitals now"
-            " than in the imported guess!\n");
-        abort();
+        const std::string msg = "(parse_import_vector): Can't drop more orbitals now than in the imported guess!\n";
+        outfile->Printf(msg.c_str());
```
davpoolechem, line 86: Not a big critique here, but isn't the `Printf` call here redundant, since throwing should print the message anyway?

TiborGY: Yes it _should_. Unfortunately, on some systems that does not happen reliably due to unclear reasons, possibly a `libstdc++` bug or something of that nature. So Psi4 is left with the choice of either printing the message twice or have some users getting no message at all.

I have some ideas on how this might be at least somewhat cleaned up later, but for now I just want to get rid of `abort()` calls.

davpoolechem: Hmm, I see, I see. I can accept that, then!

davpoolechem (approved):

loriab (approved):
Thanks for the clean-up!

PR #3173, opened 2024-05-23
memdf in saptdft

Awallace3:
## Description

In, SAPT(DFT), I removed setting "DF_INTS_IO" to "SAVE" because it blocked scf_type=="DF" from accessing MemDF on its own

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


davpoolechem (approved):

psi4/driver/procrouting/sapt/sapt_proc.py
```diff
@@ -107,15 +107,9 @@ def run_sapt_dft(name, **kwargs):
     core.IO.set_default_namespace('dimer')
     data = {}
 
-    if (core.get_global_option('SCF_TYPE') == 'DF'):
```
loriab, line 110: It won't let me do this as a suggestion, but I suggest all these deleted lines be restored and `== 'DF'` be replaced by `== 'DISK_DF'`

loriab (commented):
This one has troubled me b/c while it's making most SAPT(DFT) computations more efficient by allowing the scf_type=DF default to access memDF, it's making diskDF less efficient b/c not saving ints between dimer and monomers. The culprit is https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libfock/jk.cc#L242 . I've proposed a modification that will still allow saving for explicit `set scf_type disk_df`. And added a note to the SAPT document to address this later. fyi @carolinesargent that if you're running master with this in place, input files will behave differently for the KP/DGAS route.

PR #3174, opened 2024-05-28
Support all W integrals

JonathonMisiewicz:
## Description
`RelPotentialInt` and associated integral classes now expose all four W integral types instead of just the W0 integrals.

This PR seems way too easy. @brianz98, please let me know if there are issues. Obligatory @fevangelista and @dnascimento13 pings.

- [x] `RelPotentialInt` and associated integral classes now expose all four W integral types instead of just the W0 integrals.

## Dev notes & details

- [x] Changes `RelPotentialInt` to expose W0, Wx, Wy, and Wz integrals.

## Questions
- [x] Any objections to changing an existing integral type rather than creating a new one?
- [x] Any documentation edits needed?
- [x] Do any of the users need the derivative integrals?

## Checklist
- [ ] Are any tests necessary?
- [x] `ctest -L x2c` passes, in addition to standard tests

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/src/psi4/libmints/rel_potential.cc
+++ b/psi4/src/psi4/libmints/rel_potential.cc
@@ -65,6 +65,7 @@ RelPotentialInt::RelPotentialInt(std::vector<SphericalTransform>& st, std::share
             {bs1_->molecule()->x(A), bs1_->molecule()->y(A), bs1_->molecule()->z(A)}});
     }
 
+    set_chunks(4);
     engine0_ = std::make_unique<libint2::Engine>(libint2::Operator::opVop, max_nprim, max_am, 0);
     engine0_->set_params(params);
     // If you want derivatives of these integrals, just take the code from potential.cc's constructor


fevangelista (approved):
This does indeed seem too easy. I assume the X2C tests in psi4 pass or that you have tested them locally. Answers to your questions:

Q: Any objections to changing an existing integral type rather than creating a new one?
A: **No**

Q: Any documentation edits needed?
A: **I don't think so. Relativistic integrals are not documented.**

Q: Do any of the users need the derivative integrals?
A: **Count me in. With additional extra work, that would easily enable X2C derivatives of all methods in psi4.**

fevangelista (approved):

fevangelista:
And just to clarify: to get the four components of W the user would use this function, passing a std::vector of `SharedMatrix` to fill up?
```
virtual void compute(std::vector<SharedMatrix> results)
```

brianz98:
Thanks Jonathon, looks good to me!  Would be nice to have some comments alongside the code as it's not totally transparent what it's doing / why it would work.

susilehtola:
Hm, GitHub is telling me that the only change is the addition of 
```
    set_chunks(4);
```
in `psi4/src/psi4/libmints/rel_potential.cc`.

JonathonMisiewicz:
> Hm, GitHub is telling me that the only change is the addition of
> 
> ```
>     set_chunks(4);
> ```
> 
> in `psi4/src/psi4/libmints/rel_potential.cc`.

Well, yes. That's the one change that seems necessary. Just compare against [the dipole code](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/dipole.cc). L2 is **already** computing all the W0, Wx, Wy, Wz integrals because `libint2::Operator::opVop` on line 69 tells it to. There isn't a way to tell L2 to only compute W0 any more than there's a way to tell it to only compute the Y component of the dipole. The question is just "how many buffers do we need to read from?", which is what `set_chunks` is there for.

The surprise is that Psi handles a mismatch between the number of chunks and the number of matrices silently. Presumably, sending in fewer buffers than chunks means that some generated integrals are never grabbed from L2 (I haven't checked this myself). What I know for sure is that sending in a single matrix just grabs buffer 0. Sending in a vector of Matrix indeed just grabs one buffer for each chunk.

In my mind, the question is just what tests you all want for new integrals.

loriab (approved):
lgtm!

If you want a test model, there's https://github.com/psi4/psi4/blob/master/tests/mints3/input.dat, but testing ints directly is the exception rather than the rule, and it sounds like the ints will be in use downstream soon enough.

JonathonMisiewicz:
I think this is enough for now. I want to very carefully check the handling of perturbations before making this accessible enough to be tested by our normal suite.

For posterity, standard usage with the current interface is:
```
    std::shared_ptr<OneBodySOInt> wOBI(integral_->so_rel_potential());
    OperatorSymmetry msymm(OperatorSymmetry::L, molecule_, integral_, soFactory_);
    auto wMats = msymm.create_matrices("SO Relativistic Potential");
    wMats.insert(wMats.begin(), soFactory_->create_matrix("SO Relativistic Potential"));
    wOBI->compute(wMats);
```

PR #3176, opened 2024-05-30
Update liboptions.cc

JonathonMisiewicz:
## Dev notes & details

- [x] Changed unneeded accumulate to range-based-for-loop, removing passing ints by ref
- [x] Pretty sure this will get #3113 past a step of the build, and it's also good practice. Passing ints by reference is normally bad for efficiency, certainly more confusing.

## Checklist
- [x] If it compiles, we're good

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/liboptions/liboptions.cc
+++ b/psi4/src/psi4/liboptions/liboptions.cc
@@ -842,7 +842,6 @@ Data& Options::use(std::string& key) {
                 choices_joined.append(" ");
             }
             choices_joined.append(s);
-            return 0;
         };
 
         printf("\nDid you mean? %s\n\n", choices_joined.c_str());
--- a/psi4/src/psi4/liboptions/liboptions.cc
+++ b/psi4/src/psi4/liboptions/liboptions.cc
@@ -837,13 +837,13 @@ Data& Options::use(std::string& key) {
         }
 
         std::string choices_joined;
-        std::accumulate(std::begin(choices), std::end(choices), 0, [&choices_joined](int, std::string& s) {
+        for (const std::string& s : choices) {
             if (!choices_joined.empty()) {
                 choices_joined.append(" ");
             }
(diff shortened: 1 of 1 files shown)


fevangelista (approved):

psi4/src/psi4/liboptions/liboptions.cc
```diff
@@ -837,7 +837,7 @@ Data& Options::use(std::string& key) {
         }
 
         std::string choices_joined;
-        std::accumulate(std::begin(choices), std::end(choices), 0, [&choices_joined](int&, std::string& s) {
+        std::accumulate(std::begin(choices), std::end(choices), 0, [&choices_joined](int, std::string& s) {
```
fevangelista, line 840: Why not just do this, which expresses the intention more clearly? There is no need to use accumulate, since we're not accumulating anything here.
```suggestion
for (const std::string& s : choices) {
    if (!choices_joined.empty()) {
        choices_joined.append(" ");
    }
    choices_joined.append(s);
}
```

fevangelista (changes requested):
By the way, in this call to `accumulate` the `int` is not used at all. It

psi4/src/psi4/liboptions/liboptions.cc
```diff
@@ -837,13 +837,13 @@ Data& Options::use(std::string& key) {
         }
 
         std::string choices_joined;
-        std::accumulate(std::begin(choices), std::end(choices), 0, [&choices_joined](int&, std::string& s) {
+        for (const std::string& s : choices) {
             if (!choices_joined.empty()) {
                 choices_joined.append(" ");
             }
             choices_joined.append(s);
             return 0;
```
fevangelista, line 845: I think the line `return 0;` should be removed.

loriab (approved):

jturney (approved):

PR #3201, opened 2024-08-08
Add missing basis set to nbody example in docs

berquist:
## Description

The example in https://psicode.org/psi4manual/master/nbody.html is not copy-pasteable as-is, it's missing a basis set.  I picked something small and not horrible so it finishes in a reasonable amount of time.

## Status
- [x] Ready for review
- [ ] Ready for merge


loriab (approved):
whoops, thanks

JonathonMisiewicz (approved):

PR #3228, opened 2024-10-02
Enabling SCF calculations with arbitrary external potential

### LINKED ISSUE #3227: Computing SCF with an arbitrary external potential operator
My collaborators are trying to compute the SCF energy with an external one-electron potential added to the Fock matrix (this potential does NOT result from a set of point charges). We were trying to use the `wfn.push_back_external_potential()` functionality to accomplish this, but the external potential is ignored.

I tracked down the issue to [this line in `scf_iterator.py`](https://github.com/psi4/psi4/blob/master/psi4/driver/procrouting/scf_proc/scf_iterator.py#L312) which erases the provided external potential, resulting in only potentials from `PCM/DDX/PE`, added immediately afterwards, being allowed.

The problem is completely fixed (at least in our case) by commenting out the aforementioned line of code. I will submit a one-liner PR to do this.

konpat:
## Description

A one-line change to enable an SCF calculation with an arbitrary one-electron matrix `X` added to the Fock matrix via `wfn.push_back_external_potential(X)`. Fixes #3227.

## Questions
- [x] Is this the right way to accomplish this goal (computing SCF with an extra 1-electron operator added to the Fock matrix)?
- [x] Does this mess up PCM/DDX/PE in any way? (It shouldn't).

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/scf_proc/scf_iterator.py
+++ b/psi4/driver/procrouting/scf_proc/scf_iterator.py
@@ -309,7 +309,7 @@ def scf_iterate(self, e_conv=None, d_conv=None):
             self.H().add(Vefp)
 
         SCFE = 0.0
-        self.clear_external_potentials()
+#       self.clear_external_potentials()
 
         # Two-electron contribution to Fock matrix from self.jk()
         core.timer_on("HF: Form G")


susilehtola:
This is not ok.

The SCF class header `hf.h` states
```
    /// List of external potentials to add to Fock matrix and updated at every iteration
    /// e.g. PCM potential
    std::vector<SharedMatrix> external_potentials_;
```

Indeed, if you look at the SCF driver, this array gets populated at every iteration
```
        if core.get_option('SCF', 'PCM'):
            calc_type = core.PCM.CalcType.Total
            if core.get_option("PCM", "PCM_SCF_TYPE") == "SEPARATE":
                calc_type = core.PCM.CalcType.NucAndEle
            Dt = self.Da().clone()
            Dt.add(self.Db())
            upcm, Vpcm = self.get_PCM().compute_PCM_terms(Dt, calc_type)
            SCFE += upcm
            self.push_back_external_potential(Vpcm)
```
which is why it is cleared at every iteration.

The right way to do this is to add a definition of a persistent external potential, which gets summed in at every iteration.

konpat:
Closed as [#PR3243](https://github.com/psi4/psi4/pull/3243) makes this one obsolete.

PR #3239, opened 2024-11-01
Fix EMBPOT Functionality in SCF and Add EMBPOT Gradients

johnppederson:
## Description

This pull request fixes the functionality of the EMBPOT perturbation to the core Hamiltonian in SCF calculations and adds appropriate gradients for the EMBPOT potential.

The EMBPOT perturbation works by reading x, y, z, w, and v coordinates from a file (`EMBPOT`) during the call to `HF::form_H()`. The core Hamiltonian of the `HF` object is then modified to include a contribution that is calculated by performing numerical integration of the basis over the EMBPOT coordinates, weights, and potentials. This functionality assumed that the values of phi calculated in the `BasisSet::compute_phi()` routine are always in a cartesian basis and not in a spherical basis, which was true up until the fix in #2210, which was included in the 1.4 release. Accordingly, an unnecessary AO to SO transformation is applied in the `HF::form_H()` routine every time, which results in inaccurate EMBPOT matrices being added to the core Hamiltonian. In order to correct this in the current version of Psi4, I have removed the AO to SO transformation and call `BasisSet::compute_phi()` on an appropriately sized vector. I have also added numerical gradients over the EMBPOT potential using the gau2grid library.

- [x] The user may supply an arbitrary potential evaluated on a numerical quadrature grid to an SCF calculation by saving the x, y, z, w, and v values to a human-readable `EMBPOT` file. The first line of the `EMBPOT` file must have the number of points inside of the file. The user must also include `set perturb_h true` and `set perturb_with embpot` in the Psi4 input. The potential will then be evaluated and included in the core Hamiltonian construction, and energy and gradient calculations.

## Dev notes & details

- [x] Fixed the EMBPOT functionality in `HF::form_H()` by removing AO to SO transformation and supplying an appropriately sized vector to the `BasisSet::compute_phi()` call.
- [x] Added function `MintsHelper::embpot_grad()` to calculate the component of the gradient from the EMBPOT perturbation.

## Checklist
- [x] Added test `embpot1` to compare energies and gradients computed by including embedded point charges analytically, through the `external_potentials` keyword, and numerically, through the EMBPOT functionality.

## Status
- [x] Ready for review
- [ ] Ready for merge


loriab (approved):
lgtm! Thanks for the bugfix, the new deriv functionality, and the tidy validating test case.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2424,6 +2433,106 @@ std::map<std::string, SharedMatrix> MintsHelper::metric_grad(std::map<std::strin
...
+    int npoints;
+    int statusvalue = fscanf(input, "%d", &npoints);
+    double x, y, z, w, v;
+    std::vector<double> xyz(npoints * 3);
+    std::vector<double> w_vec(npoints);
+    std::vector<double> v_vec(npoints);
+
+    // Pull out data
+    for (int k = 0; k < npoints; k++) {
+        statusvalue = fscanf(input, "%lf %lf %lf %lf %lf", &x, &y, &z, &w, &v);
```
jturney, line 2455: Should check the value of `statusvalue` to ensure the read in was successful.

loriab: Perhaps something like:
```

if (statusvalue != 5) {
    throw PSIEXCEPTION("MintsHelper::embpot_grad error: EOF or wrong number of inputs read from EMBPOT, return=" + std::to_string(statusvalue)");
}
```

loriab:
Hi @johnppederson, I think this one is ready for a rebase to resolve conflicts and get CI passing again. Also, Jet has a couple comments (above). I can also handle this locally if you're busy. Sorry for the extreme delay.

Since #3291 is finally in, your #3243 could be rebased, but that's for v1.11, so at your leisure.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -2424,6 +2433,106 @@ std::map<std::string, SharedMatrix> MintsHelper::metric_grad(std::map<std::strin
...
+    int natom = molecule_->natom();
+    bool puream = basisset_->has_puream();
+
+    auto ret = std::make_shared<Matrix>("EMBPOT dervatives (i.e. Nx3)", natom, 3);
+    double **Gp = ret->pointer();
+    double **Dp = D->pointer();
+
+    FILE* input = fopen("EMBPOT", "r");
+    int npoints;
+    int statusvalue = fscanf(input, "%d", &npoints);
```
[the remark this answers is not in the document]
johnppederson: I've added checks for the `statusvalue` return on `fscanf` calls to read the EMBPOT file header and body in both `mintshelper.cc` and `hf.cc`, and I've also added a check to see if the EMBPOT file size (based on `npoints`) exceeds the allocated memory.

PR #3243, opened 2024-11-08
Alternate way to add an arbitrary potential to an SCF calculation

johnppederson:
## Description
This is an alternative to the fix proposed in #3228 for issue #3227, where an arbitrary one-electron potential provided by the user can be included in an SCF calculation.

I had a conversation about this functionality with @konpat, and it seems like this would be a generally useful feature. Following the suggestion made by @susilehtola on #3228, I have added a permanent potential data structure (shared matrix), `Vperm_`, to the `HF` class, which is set when the user provides a matrix to the `permanent_potential` keyword in the `energy()` call. The permanent potential is added to the core Hamiltonian in the `HF::form_H()` call in a similar manner to the `PERTURB_WITH` potentials (dipoles, EMBPOT, etc.), and so it persists during the SCF, along with the other parts of the core Hamiltonian. Seeing as the user is performing integrals outside of Psi4 to generate the 1e matrix, this should be considered an advanced usage feature.

- [x] Users may provide an arbitrary one-electron potential to be included in the core Hamiltonian during an SCF call as a `Matrix` or `numpy.ndarray` object with the `permanent_potential` keyword.

## Dev notes & details
- [x] Added a check to `scf_wavefunction_factory` in `proc.py` for the `permanent_potential` keyword. A validation error is thrown if the argument type is not `Matrix` or `numpy.ndarray`; otherwise, the permanent potential is sent to the wavefunction object.
- [x] Added the `Vperm_` shared matrix data structure to the `HF` class, which is added onto the core Hamiltonian in the `HF::form_H()` call if it is populated. The implementation is comparable to the `PERTURB_H` machinery, and checks are performed to ensure c1 symmetry and that the matrix is square and `nbf` by `nbf`.

## Questions
- [ ] Any gradients performed with this permanent potential will not include the Hellmann-Feynman contribution from the permanent potential, and I don't think there is any way to back this out from the potential matrix since the potential is arbitrary. We could require that the user provide the appropriate Hellmann-Feynman contribution to the gradient when they call `gradient()` with a `permanent_potential` through an additional keyword, or we could print a warning and leave it up to the user to add the Hellmann-Feynman contribution onto the gradient *post hoc*. How should we approach this?

## Checklist
- [x] Added test `perm` which compares energies computed by including embedded point charges through the `external_potentials` keyword and by providing the appropriate 1e matrix from this potential through the `permanent_potential` keyword (the 1e matrix is constructed by instantiating a separate `ExternalPotential` object, adding the point charges, and calling `ExternalPotential::computePotentialMatrix()` on the basis).

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/driver/procrouting/proc.py
+++ b/psi4/driver/procrouting/proc.py
@@ -1493,6 +1493,15 @@ def scf_wavefunction_factory(name, ref_wfn, reference, **kwargs):
         sapgau = core.BasisSet.build(wfn.molecule(), "SAPGAU_BASIS", core.get_global_option("SAPGAU_BASIS"))
         wfn.set_basisset("SAPGAU", sapgau)
 
+    if 'permanent_potential' in kwargs:
+        potential_matrix = kwargs["permanent_potential"]
+        if isinstance(potential_matrix, np.ndarray):
+            potential_matrix = core.Matrix.from_array(potential_matrix)
+        elif not isinstance(potential_matrix, core.Matrix):
+            raise ValidationError("Argument of 'permanent_potential' keyword has unrecognized type '%s', " %
+                                  type(potential_matrix) + "please provide a 'psi4.core.Matrix' or 'numpy.ndarray' object.")
+        wfn.set_permanent_potential(potential_matrix)
+
     if hasattr(core, "EXTERN") and 'external_potentials' in kwargs:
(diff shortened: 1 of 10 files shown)


johnppederson:
This has been has been superseded by #3349.

Issue #2407, opened 2022-01-14
Overhaul Threading Documentation


JonathonMisiewicz:
We recently had a user confused by the [threading documentation](https://psicode.org/psi4manual/master/external.html#threading), so that needs an overhaul as well:

- In (3), change the sample code to only set the number of threads once. That's a more common use case, and those who _need_ to change the number of threads are more likely to read the supporting details.
- Either remove (1) or relegate it to a sidebar.
- The introduction to the threading section is overly technical. The first paragraph can be simplified to "Psi4 threading uses both OpenMP and MKL, so it is important that your installation is linked against both if you want to use threading. If you have a conda install, this is already taken care of."
- Replace "certain circumstances" in (4) with a more specific statement.

Harishankar14:
Hi,

I noticed this issue and would like to take it on. I believe I can simplify the threading documentation as described while maintaining clarity for all users. If it's okay with you, could you please assign this to me?, My approach would be  to  simplify the intro, update section (3) to set threads once, move section (1) to advanced notes, and clarify section (4) with specific examples. I’ll submit changes after testing locally."

PR #3248, opened 2024-12-01
Overhaul Threading Documentation #2407

Harishankar14:
Updated Threading Documentation:

Simplified the introduction to threading to make it more accessible for new users, explaining that Psi4 uses OpenMP and MKL for threading and clarifying that conda installations handle this automatically.

Refined the sample code in section (3) to set the number of threads only once, aligning with the more common use case. The updated example emphasizes simplicity for most users while providing additional details for those who need more control.

Relegated section (1) to a sidebar for clarity, or removed it entirely, as it was deemed unnecessary for the main content.
Replaced vague language like "certain circumstances" in section (4) with more specific statements for better clarity.
These changes aim to improve readability and ensure the documentation is more user-friendly, especially for those new to threading in Psi4.


loriab:
Hi @Harishankar14,

Thanks for your help with the documentation, as the newcomer perspective is valuable. A few guidance notes before we run the CI:
* There's a lot of files generated by the docs build that shouldn't be in the repo. One can build the documentation through cmake through `make sphinxman`, but that does require a lot of extra tools installed, so feel free to keep your sphinx-build Makefile, but don't `git add` it. Only add the files you're edited, like `external.rst`. To remedy, you'll want to `git rm _build` on your existing branch, or start over with a new local branch from master and only add `external.rst`, then push to this github branch.
* It's definitely challenging to write docs that are concrete enough for newcomers, comprehensive enough for experienced users, and compact enough to have a chance of being read. :-) With settling on 4 for the number of threads in the example and changing "may be changed multiple" to "can be reset once", I think it might be too focused on an explicit example. Could you rework it to cover both the explicit and abstract viewpoints, please? (When I saw the edit, I thought "oh dear, has he tested multiple resetting, and it's broken?", so that's the reaction to avoid.)
* I agree with moving the MKL_NUM_THREADS to a sidebar or note. (The reason for keeping it is in case people search the docs for that envvar to find out if it's relevant.) And renumbering 2-4 as 1-3 is good.

This will be a good start, then we can run the docs builder CI on it. Thanks again!

Harishankar14:
@loriab, sure I get your suggestions. thanks for letting me know it, will make the changes

Harishankar14:
Summary of Updates:

Threading Documentation:

Reworked the threading example to cover both explicit and abstract viewpoints, addressing concerns about multiple thread resets.
The threading information has been made clearer to avoid confusion and ensure accuracy.
MKL_NUM_THREADS:

Moved the MKL_NUM_THREADS environment variable setup to a sidebar for better clarity and visibility.
Updated the documentation to guide users on controlling threading effectively via environment variables and other methods.
Unnecessary Files:

Removed the _build directory and the Makefile from the repository, which were previously included unintentionally.
Added the _build/ directory to .gitignore to prevent it from being added again in the future.
Miscellaneous:

Renumbered the sections as per review feedback, adjusting section numbering for better clarity.
Ensured that all changes are consistent with the documentation style and structure.
These changes have been pushed and should address the reviewer's comments. Please let me know if there are any further updates needed!

Issue #2201, opened 2021-06-24
X2C Linear Dependencies


JonathonMisiewicz:
There have been reports of X2C not being well-behaved in the presence of linear dependencies. Calling on @fevangelista or @susilehtola to give details. Quoting Francesco from Slack:

> I think it is still unclear to me if the general solution is to a) implement an algorithms that eliminates lin. dependencies consistently in both the small and large components of the modified Dirac equation, or even simpler b) define auxiliary basis functions that are linearly independent. Right now X2C solves the Dirac equation in a decontracted computational basis, so when you go up to QZ or higher basis sets you can easily run into linear dependency issues and the code does not handle them correctly.

susilehtola:
This is an inherent drawback of X2C which employs a restricted minimal balance basis. The orbital basis set of non-relativistic theory is the large component set that depicts the electronic part of the relativistic wave function, and the small component set is obtained by acting on the large component with something that looks like `sigma . p` that has the classical three-momentum operator which is essentially nabla. This approach becomes inefficient when the orbital basis is large; this is why four-component codes favor the use of separate basis sets for the small and large components.

JonathonMisiewicz:
Closed, duplicate of #868. Transferring relevant discussion there.

Issue #868, opened 2017-11-24
X2C linear dependencies?


susilehtola:
I'm trying to run some simple x2c jobs using the new Karlsruhe x2c basis sets. The SVP jobs run fine, but the larger basis sets cause problems. For instance,
```
molecule {
 units bohr
 0 1
 Ag
 Ag 1 4.60
}

set basis x2c-TZVPall-decon
set basis_relativistic x2c-TZVPall-decon
set relativistic x2c
set scf_type direct
set df_scf_guess false
set dft_radial_points 99
set dft_spherical_points 590

energy('svwn')
```
fails to converge.

For SVP, I get
```
    The 1-norm of |H_X2C - H_Dirac| is: 0.000000088631
```
but for TZVP
```
    The 1-norm of |H_X2C - H_Dirac| is: 181094.365380967443

    WARNING: The X2C and Dirac Hamiltonians have substatially different eigenvalues!
             This is probably caused by the recontraction of the basis set.
```

This must be related to linear dependencies in the basis.

pyscf appears to run the same calculations without problems. It uses a threshold for small exponents for the small exponents (https://programtalk.com/vs2/python/12572/pyscf/scf/x2c.py/ line 68); maybe one should also be used in psi4?

fevangelista:
This problem is on my todo list. The threshold for small exponents might not be a bad idea, but the code currently assumes that the small and large components have the same size. One option is to manufacture and read a `basis_relativistic` basis that excludes functions with small exponents. This basis will be used for both the small and large components, an approach that is probably more consistent with kinetic balance.

susilehtola:
Btw does the x2c code actually do the exact treatment, i.e. update the transformation matrices at (every) iteration?

PR #3249, opened 2024-12-07
Increase the cutoff of matrix power in x2c

### LINKED ISSUE #1908: Wrong X2C energy when using cartesian functions
Hi all,

I've been running x2c calculations and when I checked for a simple Fe atom I encountered a bug I guess, as the energy is plain wrong with x2c in the cc-pVDZ basis when ```puream = false```

Here is the script I use:

```
import psi4
  
psi4.geometry("""
              Fe 0. 0. 0.
              symmetry c1
              """)
psi4.set_options({'basis':"cc-pvdz-decon",
                  'basis_relativistic': "cc-pvdz-decon",
                  'reference': "rhf",
                  'relativistic': "X2C",
                  'puream':"false"})

scf_e, scf_wfn = psi4.energy('scf', return_wfn=True)
print("Hartree-Fock energy: {}".format(scf_e))
```

Just replacing ```puream = true``` fixes the issue. Any idea of the reason of this ?
Note that ```sto-3g-decon``` works like a charm with ```puream = false```.

lcyyork:
## Description
This PR tries to "fix" issue of X2C, where a huge difference is observed for `1-norm of |H_X2C - H_Dirac|` (see #1908).
Previous discussions suggest that this is caused by basis set linear dependencies, e.g., #868 and #2201.
After some tests, it seems to me the major reason for the huge 1-norm difference is the insufficient cutoff in `Matrix::power` function.
This cutoff is important because it is directly called to build the transformation matrix in symmetric orthogonalization (T = S^{-1/2}) when diagonalizing the Dirac Hamiltonian.
After increasing the cutoff to `1.0e-20` (1.0e-16 is already enough), the Fe atom calculation (#1908) appears to be reasonable, though the 1-norm difference and the converged SCF energy (-1270.9903006996810291) is still noticeably different from others (PYSCF: -1270.4887320302198, DIRAC: -1271.1221679632611, according to #1908).

```
==> X2C Options <==

    Computational Basis: CC-PVDZ-DECON
    X2C Basis: CC-PVDZ-DECON
    The X2C Hamiltonian will be computed in the X2C Basis

    The 1-norm of |H_X2C - H_Dirac| is: 0.000189942455
```

- [ ] RN 1
- [ ] RN 2

## Dev notes & details

- [ ] Feature1
- [ ] Feature2

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

closes #1908

### DIFF
--- a/psi4/src/psi4/libmints/x2cint.cc
+++ b/psi4/src/psi4/libmints/x2cint.cc
@@ -231,7 +231,7 @@ void X2CInt::diagonalize_dirac_h() {
      * Diagonalize the Dirac Hamiltonian with X2C 1e Hamiltonian
      */
 
-    SXMat->power(-0.5, 1.0e-20);
+    SXMat->power(-0.5, 1.0e-16);
     dMat->transform(SXMat);
     dMat->diagonalize(dtmpMat, E_LS_Mat);                    // diagonalize Dfock
     C_LS_Mat->gemm(false, false, 1.0, SXMat, dtmpMat, 0.0);  // C = X C'
@@ -328,7 +328,7 @@ void X2CInt::form_R() {
      */
 
     S_inv_half->copy(sMat);
-    S_inv_half->power(-0.5, 1.0e-20);
+    S_inv_half->power(-0.5, 1.0e-16);
 #if X2CDEBUG
     S_inv_half->print();
 #endif
@@ -340,7 +340,7 @@ void X2CInt::form_R() {
     //    sTmp->gemm(false, false, 1.0, clMat, S_inv_half,0.0);      // S^{-1/2} S_tilda S^{-1/2}
     // TOCHECK -> this line below should take care of the two above.
     sTmp1->transform(S_tilde, S_inv_half);
-    sTmp1->power(-0.5, 1.0e-20);
+    sTmp1->power(-0.5, 1.0e-16);
 
     /*
      * S^{-1/2} (S ^{-1/2} S_tilda S^{-1/2})^{-1/2} S^{1/2}
@@ -475,7 +475,7 @@ void X2CInt::test_h_FW_plus() {
     SharedMatrix S_inv_half(S_x2c_->clone());
     SharedMatrix H_x2c = T_x2c_->clone();
     H_x2c->add(V_x2c_);
-    S_inv_half->power(-0.5, 1.0e-20);
+    S_inv_half->power(-0.5, 1.0e-16);
     H_x2c->transform(S_inv_half);
     H_x2c->diagonalize(Evec_x2c, Eval_x2c);
 
--- a/psi4/src/psi4/libmints/x2cint.cc
+++ b/psi4/src/psi4/libmints/x2cint.cc
@@ -231,7 +231,7 @@ void X2CInt::diagonalize_dirac_h() {
      * Diagonalize the Dirac Hamiltonian with X2C 1e Hamiltonian
      */
 
-    SXMat->power(-1.0 / 2.0);
+    SXMat->power(-0.5, 1.0e-20);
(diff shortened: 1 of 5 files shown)


susilehtola:
1e-20 is smaller than floating point epsilon. I would suggest changing the default to something like 1e-14...

andyj10224:
Try the macro `PSI_ZERO` for machine precision 

(See https://github.com/psi4/psi4/blob/master/psi4/include/psi4/psi4-dec.h#L52)

loriab:
* I tried some lesser zeros, and 1.e-16 is the largest that's reasonable, though SCF is only converged to ~6th decimal. I left it at 16, but maybe it should go back to 20?


|  zero | 1-norm of H_X2C - H_Dirac | SCF ene |
| ------------- | -------------: | ----------------: |
| as-was | 1040983.949606436770 | 48.72725914 |
| `PSI_ZERO`=1.e-14 |  36662.887649390992 | -73883.88659546 |
| 1.e-16 | 0.000039556999 | -1270.99040286 |
| 1.e-18 | 0.000312855581 | -1270.99040342 |
| 1.e-20 (York had) | 0.000238637010 | -1270.99040467 |

* also, rebased and test case from issue added

loriab (approved):
Thanks for the fix!

JonathonMisiewicz:
> * I tried some lesser zeros, and 1.e-16 is the largest that's reasonable, though SCF is only converged to ~6th decimal. I left it at 16, but maybe it should go back to 20?
> 
> 
> zero 	1-norm of H_X2C - H_Dirac 	SCF ene
> as-was 	1040983.949606436770 	48.72725914
> `PSI_ZERO`=1.e-14 	36662.887649390992 	-73883.88659546
> 1.e-16 	0.000039556999 	-1270.99040286
> 1.e-18 	0.000312855581 	-1270.99040342
> 1.e-20 (York had) 	0.000238637010 	-1270.99040467
> 
>     * also, rebased and test case from issue added

What happens if you set the cutoff even lower than `1.e-20`? I don't have a good intuition for where the zero should be just yet.

Also, rather than having the magic number appear four times, let's create a temp variable `cutoff` and use that throughout.

lcyyork:
Thanks Lori. I did not check my mailbox lately. The tests and the change is fine to me. I chose 1.0e-20 because I saw this value was used in some other softwares.

loriab:
Last commit is a quick update to consolidate the var and extend the table (below) with tighter converged SCF. Haven't done any ref with OOO, and the Mac GHA lanes will probably still fail, as they differ in ~4th decimal.


|  zero | 1-norm of H_X2C - H_Dirac | SCF ene |
| ------------- | -------------: | ----------------: |
| as-was | 1040983.949606436770 | 48.72725914 |
| `PSI_ZERO`=1.e-14 |  36662.887649390992 | -73883.88659546 |
| 1.e-16 | 0.000312855581 | -1270.99040342 |
| 1.e-18 | 0.000312855581 | -1270.99040342 |
| 1.e-20 (York had) | 0.000039556999 | -1270.99040286 |
| 1.e-22 | 0.000262323157 | -1270.99040343 |
| 1.e-24 | 0.000039556999 | -1270.99040286 |

JonathonMisiewicz (approved):
g2g when tests pass

loriab:
The Mac testing lanes differ from the Linux/Win value in the 3rd to 4th decimal place. Any ideas? It's not purely BLAS b/c it's happening for Intel Mac (MKL) and both OpenBLAS and Accelerate on my Silicon Mac.

JonathonMisiewicz (changes requested):
Buying time until I can consult with others in the EFV group.

PR #3251, opened 2024-12-13
Update read_options.cc

JonathonMisiewicz:
## Description
Unmarks an option as case-sensitive.

- [x] The "summation_fields" option for polarizable embedding is no longer case-sensitive.

## Dev notes & details

- [x] Marks an option as case-insensitive that was never supposed to be case sensitive.

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):
Agree, it's lowercased in the only use. Good catch. https://github.com/psi4/psi4/blob/a25e71659d3233537d7f147a9ddc78b9a76533e0/psi4/driver/procrouting/solvent/pol_embed.py#L65

loriab:
Error in M-Intel is a known fluke. Rerunning usually heals it, but I don't think it's worth it. RTG, imo.

jturney (approved):

PR #3256, opened 2024-12-16
Saptdft reuse jk for scf

Awallace3:
## Description

Previously, SAPT(DFT) `scf_helper()` calls would re-compute integrals for SAPT0 dimer and monomer computations. Furthermore, the DFT `scf_helper()` calls would also recompute 3-index integrals by regenerating the JK-object. This PR enables developers to pass in a jk object to scf_helper to not re-compute 3-index DF integrals if desired. This change was largely drafted during a meeting between @cdsgroup with @konpat on Dec. 13th, 2024. Further tested in a 7 Feb 2025 meetup.

closes #3173

## User API & Changelog headlines

- [x] SAPT(DFT) and SAPT(HF) computations will be faster through re-using integrals on SCF's

## Dev notes & details

- [x] Added kwargs option of passing a JK-object to `scf_helper()` to prevent re-computing integrals for SCF.

## Questions
- [x] Question1

## Checklist
- [X] Passes previous sapt-dft1 tests
- [x] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge

PR #3271, opened 2025-02-06
Reuse the JK object from SCF into FISAPT

konpat:
## Description

This PR makes it possible to pass a jk object from the initial HF calculation into FISAPT to avoid re-computing 3-index DF integrals. This is inspired by [PR#3256](https://github.com/psi4/psi4/pull/3256) by @Awallace3 and originates from the same discussion with @cdsgroup, but it is orthogonal - can work with or without the other PR.  

- [x] FISAPT computations will be slightly faster through re-using integrals

## Dev notes & details

- [ ] A premade JK object is now passed to `fisapt_compute_energy()`.

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge


andyj10224 (approved):
This looks good! Thanks @konpat!

JonathonMisiewicz (approved):

psi4/driver/procrouting/sapt/fisapt_proc.py
```diff
@@ -35,7 +35,7 @@
 from .. import empirical_dispersion
 
 
-def fisapt_compute_energy(self, external_potentials=None):
+def fisapt_compute_energy(self, jk_obj, external_potentials=None):
```
loriab, line 38:
```suggestion
def fisapt_compute_energy(self, jk_obj, *, external_potentials=None):
```

loriab (approved):
lgtm, thank you!

one suggestion that hopefully won't be disruptive since it's an internal fn.

psi4/driver/procrouting/sapt/fisapt_proc.py
konpat: No problem!

PR #3281, opened 2025-04-25
shake up versions and add newly avail for windows

loriab:
## User API & Changelog headlines

- [x] enable new dftd4 (windows), einsums, cppe, integratorxx, python 3.13

## Dev notes & details

- [x] Drop testing and explicit support for python 3.8. Start testing 3.13.
- [x] Accommodate new einsums v1
  - [x] no longer need mkl/openblas variants, hooray!
  - [x] made possible by Einsums/Einsums#157 / v1.0.3 that allow the python module to be build here
  - [x] update target and syntax for einsums in psi4, including the dummy usage
- [x] advancing to cppe 0.3.2 from 0.3.1 because build matrix is better. but somehow it's a little off on the pe+ecp cross tests, hence a few explicit exceptions. (the adc test seems only for old py on mac, so hopefully it ages out)
- [x] `mkl-version` not acutally used, but the prohibition on mkl 2024.2 for windows is well-merited (tied to build mkl*24 to hold back to 2024.1)
- [x] newly enabled dftd4 for windows (in modern version that has what we need for 3c methods; there wer ~v1.0 builds available). to use this, need qcengine v0.32.0
- [x] added integratorxx since it's now available as a conda package. unfortunately gauxc is maybe using intxx paths from a src build so can't use the installed version?
- [x] force workaround of problem Jonathon hit where default static build of chemps2 won't link the chemps2 executable also built by default
- [x] update upstream src for gauxc for a problem building with latest gcc compilers that Jonathon fixed

## Not fully tackled
- [ ] h5py compatibility  needed for adcc
- [ ] openfermion stuck on np v1
- [ ] torsiondrive needs c-f release for np v2
- [ ] torsiondrive will enable qcfractal again
- [ ] still stuck at cmake 3.29 for some arch. need to enable particularly policy 117? for windows to generally allow cmake v4

## Checklist
- [ ] Tests added for any new features
- [ ] [All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [x] Ready for merge

PR #3285, opened 2025-05-05
Update aug-cc-pvtz-dk

jaclark5:
## Description
This PR updates the `aug-cc-pvtz-dk` basis set to include all supported elements in basis set exchange. Note that the `diff` shows many changes which are rearrangements between the current basis set exchange output and the older version from which the current `.gbs` file was sourced. I manually rearranged the basis set functions for the first few atoms to highlight this fact. 

Note that the existing .gbs file appears to be for `cc-pvtz-dk` in error rather than `aug-cc-pvtz-dk`, like the label. There are a few extra basis functions added for previously supported elements to accommodate the diffuse functions.

- [x] Update aug-cc-pvtz-dk to include diffuse functions for all atoms and support for Mo, Nb, Pd, Rh, Ru, Tc, Y, and Zr

## Dev notes & details
- [x] Update aug-cc-pvtz-dk.gbs file to include diffuse functions for all atoms and support for  Mo, Nb, Pd, Rh, Ru, Tc, Y, and Zr

## Checklist

## Status
- [X] Ready for review
- [ ] Ready for merge


JonathonMisiewicz:
Thanks!

This PR is going to fail tests due to unrelated build problems. When PR #3281 is merged in, please rebase onto that to heal tests.

loriab:
Thanks for the basis set update. I think you might have generated this w/o checking the "Optimize General Contractions" box off the "Advanced" link on BSE. The spec will match the existing file better that way and the basis will be smaller/run faster.

Ideally, I try to keep the cc-pvtz-dk https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/primitives/basis-cc-pvtz-dk.gbs and the diffuse only https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/primitives/diffuse-aug-cc-pvtz-dk.gbs and then run a script to add them together to update both bases simultaneously. But let's see what your diff looks like with the optimized contractions before further plans.

loriab:
Thanks, the optimized general contractions made the diff a lot cleaner. I don't see the "`cc-pvtz-dk` in error rather than `aug-cc-pvtz-dk`" anymore, do you? Is the diff only new atoms now, or are there corrections to existing atoms? Do you think it'd be hard to split into cc-pvtz-dk and diffuse files? Sorry this is a little complicated.

jaclark5:
Yeah, I can give it a go, does that mean this aug-cc-pvtz-dk file will be deleted? Otherwise it's straightforward for users to choose a basis set that is incorrect. It seems to me that it would be cleaner to have .gbs files for "cc-pvtz-dk" and "aug-cc-ppz-dk" since that's what users would expect.

loriab:
> does that mean this aug-cc-pvtz-dk file will be deleted?

correct, it'd be the two files in the `primatives/` directory (linked above) that get edited.  (Don't worry, I won't make you run the adder script -- it's in ancient perl.)

jaclark5:
@loriab How will the base and diffuse basis functions be combined in practice in psi4? I ask because I need this for use in QCPortal where only a single basis set can be defined. I'm worried that this change will effectively remove the augmented basis set from my "toolbox".

loriab:
It'll still be one basis set to call in qcportal. The two "primitives" basis files you update will get added together (this is how all the Dunning basis sets work https://github.com/psi4/psi4/blob/master/psi4/share/psi4/basis/NOTES#L817 ) so there will be a aug-cc-pvtz-dk.gbs file in the end, it'll just be autogenerated from the primitives.

jaclark5:
Ok great! How can I go about separating the base and diffuse terms, you said there was a perl script? Should I instead have this PR delete the aug-cc-pvtz-dk.gbs file?

loriab:
The perl script is for adding, not separating. Probably you'd copy the new elements from BSE cc-pvtz-dk into the primitives/basis-cc-pvtz-dk.gbs file. And copy the difference in BSE cc-pvtz-dk and aug-cc-pvtz-dk, namely the diffuse functions, into primitives/diffuse-aug-cc-pvtz-dk.gbs

jaclark5:
@loriab this PR is now up to date with the master branch and the diff is as clean as I can make it. 

What do you think of the header?

I wrote my own python module to separate the diffuse basis functions and needed to make some ordering decisions so that it would be successful. I'll link the module here when it's polished and on GitHub.

jaclark5:
> Just a notice that Lori (and many of the other Psi developers) are at a conference, so expect slow responses for the next few days.

@JonathonMisiewicz Thanks for letting me know! I'm also at a symposium so I was trying to get this in

loriab:
Screenshot 2025-05-15 at 1 59 39 PM

Thanks for generating the primitives. I've run the manipulation scripts and regenerated the full basis sets. Above is a specialized diff (shell ordering doesn't matter; lc black is unchanged; uc blue is new elements) at various points to show all's well.

I think this is pretty much rtg. lmk if you have any concerns. 

Oh, and it occurred to me at the conf. last week that I may have been answering the wrong question for you -- one can  access BSE basis sets directly by prepending `bse:` to the name, thanks to a PR by @bennybp . See, for example, the change to the ddd-deriv test below to allow it to pass w/o using the psi4 cc-pVDZ library basis set. Note, though, that BSE bases don't have tailored default for auxiliary DF bases set up, so it'll default to the universal quad-zeta one that produces slightly different answers unless you set it explicitly, like below.

https://github.com/psi4/psi4/blob/master/tests/ddd-deriv/input.dat
```
-set basis cc-pVDZ
+set basis bse:cc-pVDZ
+set df_basis_scf cc-pVDZ-JKFIT
```

loriab (approved):

jaclark5:
I don't have any concerns, thanks!

JonathonMisiewicz (approved):

JonathonMisiewicz:
Test failures are unrelated to this PR and are also showing up in my work... We'll want #3286 merged in, once it's on master.

PR #3286, opened 2025-05-05
cmake v4, test more deps w/latest hdf5 and numpy

loriab:

- [x] update psi4 and plugins for cmake v4

## Dev notes & details

- [x] fix up frontmatter versions/arch
- [x] all at once for hdf5/h5py/adcc/einsums/ambit now (since they're now all built with hdf5 1.14.6)
- [x] show ddx works with qcschema and make other ddx tests actually check answers
- [x] fix up at least psi4 itself (not necessarily all deps) for cmake v4 (and release the practical cmake 3.29 pin I've had for a while)
- [x] fix up plugins for cmake v4 (v2rdm_casscf got extra work)
- [x] updated compiled addons in psi4Config and finished enabling integratorxx checks
- [x] something about the conda env or cmake cache setup, but v2rdm build on Mac (I & S) was adamantly not detecting threads right `Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success` even though it had earlier during the psi4-core build. maybe this is a little new b/c the unix c-f L2 packages have OMP now (wasn't finding that the same way either). Reproducible on laptop. Resistant to debugging, but moving the detection higher up the file (psi4Config -- really happens in find_package(L2)) fixed it, so running with that.
- [x] we're not using modules, so tell cmake not to scan for them even if std is c++20 otherwise need add'l `clang-scan-deps` tool at buildtime
- [x] bump optking to v0.3
- [x] accommmodate some deps that haven't had cmake update in a while by forcing a later policy std on them (g2g, libxc, gauxc)
- [x] after torsiondrive work to get a np v2 compatible release, we're now solving environments entirely with numpy v2. (TD was needed with qcf which we like to test with).
- [x] openfermion suppressed in codedeps so that ppl solve for np v2 envs in general. The latest np v1 was build for py39/310/311/312 and np v2 py310/311/312/313 so we don't want openfermion as a choking the env sol'n
- [x] On Windows, when libgcc 15 is present, `import psi4` throws `ImportError: DLL load failed while importing core: The specified procedure could not be found.` It seems like a runtime problem, not a buildtime (that is, v15 ok to have around during compile). So for now, we're evading it. Thank goodness I had successive CI runs last week where one passed, then failed. Having eliminated code changes, it had to be env. I've no idea what's actually going wrong.
- [x] also includes changes found for conda packaging https://github.com/conda-forge/psi4-feedstock/pull/39

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


CMakeLists.txt
```diff
@@ -122,6 +123,7 @@ option(ENABLE_Einsums "Enables the Einsums tensor library" OFF)
 option(ENABLE_simint "Enables use of SIMINT two-electron integral library" OFF)
 option(ENABLE_gauxc "Enables use of GauXC density functional library" OFF)
 option(ENABLE_gdma "Enables Stone's GDMA multipole code (requires Fortran; can also be added at runtime)" OFF)
+option(ENABLE_IntegratorXX "Enables use of IntegratorXX density functional grids library" OFF)
```
JonathonMisiewicz, line 126: Do we actually use this anywhere?

psi4/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
-cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.19...4.0)
 cmake_policy(SET CMP0094 NEW)
+if (POLICY CMP0167)
+    # L2 not prepared for BoostConfig instead of FindBoost
+    cmake_policy(SET CMP0167 OLD)
```
JonathonMisiewicz, line 5: Is there anything that needs to be done on the EFV end?

CMakeLists.txt
loriab: There's a dummy module (like Einsums), so we check it compiles, not just detects. GauXC uses it (unfortunately it has to build its own b/c it expects different include paths than the installed project). OpenOrbitalOptimizer uses it for testing (that's irrelevant to Psi4). So the answer is "not really", but there's addons we use less.

JonathonMisiewicz (approved):

psi4/CMakeLists.txt
loriab: Hard to say, as the last time I looked at L2 deps was evaleev/libint#314 (https://github.com/loriab/libint/compare/configconfig...loriab:libint:l2deps). That mentions updating from ancient Boost (1.29) and reorganizing some of EFV's special checks on the install. From https://stackoverflow.com/a/79147222 , it sounds like requiring Boost 1.70 would be necessary and probably some `find_package(Boost CONFIG)` conversions.

It's not a big concern for me.

jturney (approved):

PR #3291, opened 2025-05-19
expanding external_potential: parsing, take 2

loriab:
## Description

replace half of #3272

- [x] `psi4.QMMMbohr` is further discouraged, and external point charges specification (always in Bohr, regardless of molecule units) now more flexibly takes in lists, numpy arrays, and psi4.core.Matrix es. (It also takes in more arrangements, but none of them will save keystrokes from the point charges case.)

## Dev notes & details

- [x]  As discussed at the 7 Feb Patkowski/Pederson/CDSgroup meeting, we want to route more user permanent potentials through the external_potential keyword (embedding charges in MBE). This is step 1 to allow point charges, diffuse charges, and (Nao, Nao) matrices to pass through, validated, with backwards compatible syntax. (diffuse and matrices are disallowed at the next step before hitting c-side.)
- [x]  Started trying to retire driver/qmmm.py, as the pb11 export of core.ExternalPotential should be sufficient. Fortunately, its direct use was already deprecated. Only MDI needed a line change. `QMMMbohr()` will raise a FutureWarning now
- [x] Filled in the test mark so `-m extern` runs all EP tests.
- [x] The `extern.cc` looks like a lot of changes, but it's really just hiding a block behind `if (charges_.size()) {...}` which is future prep.
- [x] Ubuntu image 20.04 has retired, so we're oging to have to bump testing from clang v6 (2018) to clang v11 (2020)

## Questions
- [ ] Question1

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/p4util/python_helpers.py
```diff
@@ -742,6 +743,148 @@ def anon(mol, role):
...
+    mode_keys_list = ["points", "diffuse", "matrix"]
+
+    if isinstance(external_potential, dict) and ({k.upper() for k in external_potential.keys()} <= frag_keys):
+        ep = {k.upper(): v for k, v in external_potential.items()}
+    else:
+        # assign "C" to mean whole molecule in the non-SAPT case
+        ep = {"C": external_potential}
+
+    # expand input structure: may be single points list or types list or types dict or fragments dict (for SAPT)
+    ep1 = {}
```
JonathonMisiewicz, line 847: The PR _seems_ fine, but I'm having a very hard time reading it.

Can you go through this file and use longer but more expressive variable names? Keeping track of `ep` vs `ep1` and `ep2` is a headache.

loriab: sure, done

psi4/driver/p4util/python_helpers.py
```diff
@@ -742,6 +743,148 @@ def anon(mol, role):
...
+    # expand input structure: may be single points list or types list or types dict or fragments dict (for SAPT)
+    ep_building = {}
+    for frag, frag_ep in ep_outer_structure.items():
+        if isinstance(frag_ep, dict):
+            if frag_ep.keys() <= set(mode_keys_list):
+                ep_building[frag] = frag_ep
+            else:
+                raise ValidationError(f"external_potential: primary or sec keys should be among {mode_keys_list}, not {frag_ep.keys()}. Full input: {external_potential}")
+        elif isinstance(frag_ep, list):
+            if (w := validate_charge_list(frag_ep)):
```
JonathonMisiewicz, line 855: A comment explaining the logic in this if-clause would be helpful.

JonathonMisiewicz (approved):
Yep, that change was what was needed to make the code more readable. There's one more comment I'd like you to add, but otherwise, LGTM.

psi4/driver/p4util/python_helpers.py
loriab:
```
# list might be new-fangled structure holding one or many among points, diffuse, matrix (the `else` branch), or it might be the outer list of an longstanding list-o-points (the `if` branch), in which case it'll pass the `validate_charge_list` check and needs wrapping in an outer list to match the new-fangled structure.
```

Sure, I think the above is what's needed. I'll add it to the follow-up PR, as I'd like to get this in since everything needs rebasing against it to fix CI.

jturney (approved):

PR #3292, opened 2025-05-20
Adding ZORA

NathanGillispie:
## Description
Adding the zeroth-order regular approximation (ZORA) for a scalar-relativistic hamiltonian.

Here we use the model potential proposed by van Wüllen in *Molecular density functional calculations in the regular relativistic approximation* [doi.org/10.1063/1.476576](https://doi.org/10.1063/1.476576). The exact values of the model potential come from [NWCHEM](https://www.nwchem-sw.org/). Given atomic orbital $\chi$, we evaluate the following integral on a grid:  

$$T_{ij}^\text{SR}=\int d\mathbf{r}^3 \frac{c^2}{2c^2-v_\text{eff}(\mathbf{r})}\nabla\chi_i^\dagger(\mathbf{r}) \cdot\nabla\chi_j(\mathbf{r})$$

The effective potential $v_\text{eff}(\mathbf{r})$ is the potential experienced at a point due to nuclear attraction and electron repulsion from model potentials that reproduce the nuclear cusp on each atom. The generated matrix replaces the kinetic energy for SCF etc. Because this is only evaluated once, don't cheap out on the grid! The ZORA time is usually negligible. For transition metals, the energy difference with and without ZORA is much greater than numerical error from grid integration, although the grid settings can be changed. Also, the `zora_nr_debug true` option can be set to test numerical grid error (assuming analytic kinetic integrals otherwise).

Nathan Gillispie, University of Memphis, 2025

- Set `relativistic zora` to make the core Hamiltonian scalar-relativistic! (with minimal cost)

## Dev notes & details

- Simply replaces the kinetic energy integral in `libmints` with its scalar-relativistic counterpart.
- Added options (zora, zora grid settings, zora debug flag)
- Tested with all symmetry types, single and multi-threaded configurations, and all references.
- Not tested with calculations other than `energy('scf')`
- Implemented in a way similar to X2C

## Questions
- [x] The documentation I added in `read_options.cc` will pop up online eventually, right?

## Checklist

## Status
- [x] Ready for review
- [ ] Ready for merge

### DIFF
--- a/doc/sphinxman/source/relativistic.rst
+++ b/doc/sphinxman/source/relativistic.rst
@@ -31,9 +31,9 @@
 .. index::
    relativistic
 
-================================
 .. _`sec:relativistic`:
 
+================================
 Scalar Relativistic Hamiltonians
 ================================
 


psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,263 @@
...
+#endif
+        pworkers[thread]->compute_functions(block);
+        auto phi_x = pworkers[thread]->basis_value("PHI_X");
+        auto phi_y = pworkers[thread]->basis_value("PHI_Y");
+        auto phi_z = pworkers[thread]->basis_value("PHI_Z");
+
+
+        // Preprocess kernel c²/(2c²-veff) * weight
+        double* w = block->w();
+        double kernel[npoints];
```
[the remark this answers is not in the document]
NathanGillispie: Thanks for reviewing :D I was under the assumption that that was like a 2KB stack allocation (npoints is 100-256). I don't need the pointer, it's just that I'm coming from C-land.

susilehtola:
Is the gradient correction also implemented?

JonathonMisiewicz:
> Is the gradient correction also implemented?

It isn't added, so the gradient code will need to be edited to default to FINDIF if ZORA is on... [We do something similar for the other relativistic corrections](https://github.com/psi4/psi4/pull/2200/files), but the relevant code may have migrated since the PR were those warnings were added.

NathanGillispie:
> > Is the gradient correction also implemented?
> 
> It isn't added, so the gradient code will need to be edited to default to FINDIF if ZORA is on... [We do something similar for the other relativistic corrections](https://github.com/psi4/psi4/pull/2200/files), but the relevant code may have migrated since the PR were those warnings were added.

Sorry but I'm not familiar with this part of PSI4. Could you elaborate on FINDIF? I just pushed a commit with a change to `driver.py` that forces it to use numerical gradients, like the referenced PR.

JonathonMisiewicz:
FINDIF = finite difference = numerical gradients

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,263 @@
...
+		int thread = omp_get_thread_num();
+		pworkers[thread]->compute_functions(block);
+		auto phi_x = pworkers[thread]->basis_value("PHI_X");
+		auto phi_y = pworkers[thread]->basis_value("PHI_Y");
+		auto phi_z = pworkers[thread]->basis_value("PHI_Z");
+
+
+		// Preprocess kernel c²/(2c²-veff) * weight
+		double* w = block->w();
+		double kernel[npoints];
```
JonathonMisiewicz: As I said before, this is a memory leak. This is also the cause of the build failures. Use something like `std::vector<double> kernel(npoints)` instead.

NathanGillispie: The build failure is fixed now. It appears the issue was non-standard array declaration. In moving to heap allocation, memory deletion is now explicit.

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,269 @@
...
+#endif
+
+		pworkers[thread]->compute_functions(block);
+		auto phi_x = pworkers[thread]->basis_value("PHI_X");
+		auto phi_y = pworkers[thread]->basis_value("PHI_Y");
+		auto phi_z = pworkers[thread]->basis_value("PHI_Z");
+
+		// Preprocess kernel c²/(2c²-veff) * weight
+		double* w = block->w();
+		double* kernel = new double[npoints];
```
TiborGY: New code probably should not use new/delete and raw pointers to begin with. Is there a reason why std::vector or some type of smart pointer isn't used here?

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,269 @@
...
+{
+	// Speed of light in atomic units
+	double C = pc_c_au;
+
+	int nthreads = 1;
+#ifdef _OPENMP
+    nthreads = Process::environment.get_n_threads();
+#endif
+
+#pragma omp parallel for schedule(dynamic) num_threads(nthreads)
```
TiborGY: Is the difference between per-thread work large enough to make schedule(dynamic) better than the default, which is static on most systems IIRC?

psi4/src/psi4/libmints/zora.cc
NathanGillispie: Simply because `std::vector` is slower and over-engineered for this task. Initially I used a stack allocated array which de-allocates when going out of scope, the choice was obvious due to the hard-coded limit on `npoints` making `kernel` around 2KB. To make the de-allocation explicit, I switched to `new`, which I agree is not a good option, but it was a simple replacement. I see now that the style guidelines suggest not using `new` so I will change it.

psi4/src/psi4/libmints/zora.cc
NathanGillispie: The number of points in a block can vary greatly, so theoretically the workloads could too. I couldn't tell a difference in my testing though.

TiborGY: In that case I would leave the schedule on whatever the OpenMP implementation defaults to, instead of explicitly setting either static or dynamic.

psi4/src/read_options.cc
```diff
@@ -231,17 +231,29 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+    options.add_str("RELATIVISTIC", "NO", "NO X2C ZORA");
 #endif
     /*- Auxiliary basis set for solving Dirac equation in X2C and DKH
         calculations. Defaults to decontracted orbital basis. -*/
     options.add_str("BASIS_RELATIVISTIC", "");
     /*- Order of Douglas-Kroll-Hess !expert -*/
     options.add_int("DKH_ORDER", 2);
 
+	/*- Number of radial points for the ZORA effective potential grid -*/
+    options.add_int("ZORA_RADIAL_POINTS", 140);
```
TiborGY, line 247: Shouldn't these new options be added to the docs as well?

TiborGY (approved):

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,265 @@
...
+		{"DFT_RADIAL_POINTS", options_.get_int("ZORA_RADIAL_POINTS")},
+	};
+
+	std::map<std::string, double> grid_double_options = {
+		{"DFT_BS_RADIUS_ALPHA",   1.0},
+		{"DFT_PRUNING_ALPHA",     1.0},
+		{"DFT_WEIGHTS_TOLERANCE", 1e-15},
+		{"DFT_BLOCK_MAX_RADIUS",  3.0},
+		{"DFT_BASIS_TOLERANCE",   options_.get_double("ZORA_BASIS_TOLERANCE")},
+	};
```
susilehtola: should these really be hardcoded?

psi4/src/psi4/libmints/zora.h
```diff
@@ -0,0 +1,291 @@
...
+	std::shared_ptr<std::unordered_map<int, SharedVector>> veff_;
+
+	void setup();
+    /// Effective potential
+	void compute_veff();
+    /// Compute the T integral
+	void compute_TSR(std::vector<std::shared_ptr<BasisFunctions>>, SharedMatrix&);
+
+	/// Model basis
+	static constexpr double coeffs[2919] = {0.0063990067,0.0745660339,0.3008381442,0.4565180734,0.1616787418,
```
susilehtola, line 98: shouldn't this be a 2D array?

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,265 @@
...
+#endif
+
+		pworkers[thread]->compute_functions(block);
+		auto phi_x = pworkers[thread]->basis_value("PHI_X");
+		auto phi_y = pworkers[thread]->basis_value("PHI_Y");
+		auto phi_z = pworkers[thread]->basis_value("PHI_Z");
+
+		// Preprocess kernel c²/(2c²-veff) * weight
+		double* w = block->w();
+		std::vector<double> kernel(npoints);
```
susilehtola: you could declare this out of the loop so that the memory doesn't get reallocated nblocks times.

psi4/src/read_options.cc
NathanGillispie: I thought Sphinx automatically extracted the comments in `read_options.cc`. Honestly, I've struggled to compile the documentation so I can't check.

psi4/src/psi4/libmints/zora.cc
NathanGillispie: I was just following other code here. See [libfock/COSK.cc:211](https://github.com/psi4/psi4/blob/85c5672ca665457a8c2dbbcaeb322fb66e9c9ae3/psi4/src/psi4/libfock/COSK.cc#L211).

psi4/src/psi4/libmints/zora.h
NathanGillispie: Absolutely, the issue is that each atom has a different number of gaussian primitives. I could have had a 2D array with a lot of zeros, but I chose a 1D-array with indices for the start of each atom.

susilehtola: but it could be a `std::vector` of `std::vector`s...

psi4/src/psi4/libmints/zora.cc
NathanGillispie: Because you have to have a different kernel for each thread.

psi4/src/psi4/libmints/zora.h
NathanGillispie: True, but hardcoding for 104 atoms is hard. I don't expect the basis set to be changed, or even reused, since it was created specifically for this task.

psi4/src/psi4/libmints/zora.cc
TiborGY: > Because you have to have a different kernel for each thread.

That would perhaps be better done through OpenMP's firstprivate clause

susilehtola: ... or just declare it in the `#openmp parallel` region before going into the loop.

NathanGillispie: I tried `firstprivate` and it worked well.

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,270 @@
...
+	// Speed of light in atomic units
+	double C = pc_c_au;
+
+	int nthreads = 1;
+#ifdef _OPENMP
+    nthreads = Process::environment.get_n_threads();
+#endif
+
+	int max_points = grid_->max_points(); //Set in grid_int_options
+	std::vector<double> kernel(max_points);
```
susilehtola: now there's an extra copy of `kernel`

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,270 @@
...
+            veff_->insert({index, veff_block});
+        }
+    }
+}
+
+
+//Scalar Relativistic Kinetic Energy Matrix
+void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, SharedMatrix &T_SR) {
+	// Speed of light in atomic units
+	double C = pc_c_au;
```
susilehtola: since $c$ only appears as $c^2$, you could just use a variable for $c^2$

psi4/src/psi4/libmints/zora.cc
NathanGillispie: I declare kernel after omp parallel in the newest commit.

psi4/src/psi4/libmints/zora.h
```diff
@@ -0,0 +1,309 @@
...
+	Options& options_;
+    /*! Purpose of the map: the ``block->index()`` does not index like ``grid_->blocks()``.
+     *  For multithreading to work, this is a solution.
+     */
+    std::shared_ptr<std::map<int, SharedVector>> veff_;
+
+	void setup();
+	void compute_veff();
+	void compute_debug_veff();
+	void compute_TSR(std::vector<std::shared_ptr<BasisFunctions>>, SharedMatrix&);
```
susilehtola, line 92: these could be commented

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,273 @@
...
+
+            //einsums("i,ip->p", 𝕔[i], erf(α[i]⊗ r[p]))/r[p]
+            for (int p = 0; p < npoints; p++) {
+                double dist = std::hypot(pos_a[0]-x[p], pos_a[1]-y[p], pos_a[2]-z[p]);
+                double outer = 0;
+                for (int i = 0; i < nc_a; i++) {
+                    outer += std::erf(dist * alpha_a[i]) * coef_a[i];
+                }
+                outer /= dist;
+                outer -= Z/dist;
```
susilehtola:
```suggestion
                outer -= Z;
                outer /= dist;
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,273 @@
...
+
+        double* x = block->x();
+        double* y = block->y();
+        double* z = block->z();
+
+        auto veff_block = std::make_shared<Vector>(npoints);
+
+        for (int a = 0; a < natoms; a++) {
+            int Z = molecule_->Z(a);
+            if (Z > 104) throw PSIEXCEPTION("Z too big. Max value 104");
```
susilehtola:
```suggestion
            if (Z > 104) throw PSIEXCEPTION("Too heavy atom for ZORA implementation. Tabulation only available to Z=104, Rf.\n");
```

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1450,6 +1453,31 @@ void MintsHelper::add_dipole_perturbation(SharedMatrix potential_mat) {
     }
 }
 
+void MintsHelper::compute_so_zora_ints(bool include_perturbations) {
+    outfile->Printf(" OEINTS: Using relativistic (ZORA) kinetic integrals.\n");
+
+    if (include_perturbations && options_.get_bool("PERTURB_H")) {
+		throw PSIEXCEPTION("Perturbations of the ZORA hamiltonian are not implemented.");
+    }
```
susilehtola, line 1461: It is a bit weird to me that there are two flags for the same purpose... do other parts of the code do the same?

susilehtola: I guess this makes sense if the perturbation is weak and doesn't affect the relativistic correction that much but still...

NathanGillispie: I agree, x2c does the same thing (at `mintshelper.cc:1493`) and this is just a precaution in case anyone tries.

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,271 @@
...
+    for (const auto &block : grid_->blocks()) {
+        int index = block->index();
+        int npoints = block->npoints();
+        auto veff_block = std::make_shared<Vector>(npoints);
+        veff_block->zero();
+        veff_->insert({index, veff_block});
+    }
+}
+
+void ZORA::compute_veff()
```
andyj10224, line 157: It would be good to comment a paper/equation that explains what v_eff(r) is

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,271 @@
...
+
+#pragma omp critical
+        {
+            veff_->insert({index, veff_block});
+        }
+    }
+}
+
+
+void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, SharedMatrix &T_SR) {
```
andyj10224, line 216: See comment above… good to reference a paper and equation number

psi4/src/psi4/libmints/zora.h
andyj10224: I am not personally a huge fan of putting all these coefficients here, I believe it better to put inside a psi4/include file like the one given here: https://github.com/psi4/psi4/blob/master/psi4/include/psi4/masses.h

andyj10224 (changes requested):
Thank you for adding this feature Nathan! Good work! I do have a couple of comments that need to be addressed, and I do insist on writing docs for it (as it is a new feature). 

Check out https://github.com/andyj10224/psi4/blob/dlpno_ccsdt_q/doc/sphinxman/source/dlpnoccsdt_q.rst for inspiration. 

It will also be good to add the paper you referenced to the bibliography: https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/bibliography.rst

Let me know if you have questions, or if there is anything else I can do to help :)

psi4/src/psi4/libmints/zora.cc
NathanGillispie: Thanks :D So the comment above the ZORA class declaration has everything you're asking for, but I can include a comment there too.

psi4/src/psi4/libmints/zora.cc
NathanGillispie: Yeah same as above, I already have a latex equation in the header.

NathanGillispie:
@andyj10224 To respond to your suggestion about the location of the coefficients: I also don't like all the coefficients there, but they were made to solve this specific problem, so I doubt they will be used anywhere else. Splitting this header file into two would probably just add unnecessary complexity.

psi4/src/read_options.cc
loriab: relativistic.rst might be good place to add narrative docs (show up as https://psicode.org/psi4manual/master/relativistic.html). 

yes, the read_options comments get scraped into a short documentation snippet. this can be incorporated into the narrative docs like the last two lines of https://github.com/psi4/psi4/blob/master/doc/sphinxman/source/relativistic.rst?plain=1

andyj10224:
> @andyj10224 To respond to your suggestion about the location of the coefficients: I also don't like all the coefficients there, but they were made to solve this specific problem, so I doubt they will be used anywhere else. Splitting this header file into two would probably just add unnecessary complexity.

Got it! Might be a necessary evil for now

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+#pragma omp parallel for schedule(auto) num_threads(nthreads)
+    for (const auto &block : grid_->blocks()) {
+        int npoints = block->npoints();
+        int index = block->index();
+
+        double* x = block->x();
+        double* y = block->y();
+        double* z = block->z();
+
+        auto veff_block = std::make_shared<Vector>(npoints);
```
andyj10224, line 179:
```suggestion
        auto veff_block = std::make_shared<Vector>(npoints);
        veff_block->zero();
```
I know Psi4 automatically initializes vectors and matrices to zero, but this is helpful for clarity

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+#ifdef _OPENMP
+    nthreads = Process::environment.get_n_threads();
+#endif
+
+	int max_points = grid_->max_points(); //Set in grid_int_options
+#pragma omp parallel num_threads(nthreads)
+	{
+		// Give each thread their own scratch space
+		std::vector<double> kernel(max_points);
+		auto tmp = std::make_shared<Matrix>(T_SR->ncol(), T_SR->nrow());
```
andyj10224:
```suggestion
		auto tmp = std::make_shared<Matrix>(T_SR->nrow(), T_SR->ncol());
```

Perhaps not a big deal for Hermitian matrices

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+			auto phi_y = pworkers[thread]->basis_value("PHI_Y");
+			auto phi_z = pworkers[thread]->basis_value("PHI_Z");
+
+			// Preprocess kernel c²/(2c²-veff) * weight
+			double* w = block->w();
+			for (int p = 0; p < npoints; p++) {
+				kernel[p] = C2 /(2.*C2 - veff_block->get(p)) * w[p];
+			}
+
+			tmp->zero();
```
andyj10224:
```suggestion
			tmp->zero();
			
			// Pointer for T_SR buffer for this thread (dimension: (nbf, nbf))
			double **tmp_p = tmp->pointer();
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+
+			int thread = 0;
+#ifdef _OPENMP
+			thread = omp_get_thread_num();
+#endif
+
+			pworkers[thread]->compute_functions(block);
+			auto phi_x = pworkers[thread]->basis_value("PHI_X");
+			auto phi_y = pworkers[thread]->basis_value("PHI_Y");
+			auto phi_z = pworkers[thread]->basis_value("PHI_Z");
```
andyj10224:
```suggestion
			auto phi_x = pworkers[thread]->basis_value("PHI_X");
			auto phi_y = pworkers[thread]->basis_value("PHI_Y");
			auto phi_z = pworkers[thread]->basis_value("PHI_Z");
			
			// Pointers for basis function values (dimensions: (npoints, nbf_local))
			double **phi_x_p = phi_x->pointer();
			double **phi_y_p = phi_y->pointer();
			double **phi_z_p = phi_z->pointer();
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+				int mu = bf_map[l_mu];
+				for (int l_nu = l_mu; l_nu < local_nbf; l_nu++) {
+					int nu = bf_map[l_nu];
+					for (int p = 0; p < npoints; p++) {
+						// ∇²φ(r)*kernel
+						tmp->add(mu,nu, kernel[p]*(
+							phi_x->get(p,l_mu)*phi_x->get(p,l_nu) +
+							phi_y->get(p,l_mu)*phi_y->get(p,l_nu) +
+							phi_z->get(p,l_mu)*phi_z->get(p,l_nu)
+						));
```
andyj10224:
```suggestion
						tmp_p[mu][nu] += (phi_x_p[p][l_mu] * phi_x_p[p][l_nu] + phi_y_p[p][l_mu] * phi_y_p[p][l_nu] + phi_z_p[p][l_mu] * phi_z_p[p][l_nu]) * kernel[p];
```

Normally I would not fuss about this but since this is in such a deep loop, using pointers rather than directly calling functions would make this operation a lot more efficient :)

psi4/src/psi4/libmints/zora.h
```diff
@@ -0,0 +1,314 @@
...
+	/// Creates same structure as above but all zeros. TSR->T as veff->0
+	void compute_debug_veff();
+	/// Uses veff_ to compute the Scalar Relativistic Kinetic Energy Matrix
+	void compute_TSR(std::vector<std::shared_ptr<BasisFunctions>>, SharedMatrix&);
+
+    /*! I'm so sorry.
+     *  These are the contraction coefficients and exponents of the model basis. The
+     *  last array indexes the atoms by atomic number.
+     */
+	static constexpr double coeffs[2919] = {0.0063990067,0.0745660339,0.3008381442,0.4565180734,0.1616787418,
```
andyj10224, line 98: Nothing more to address here, I just find it absolutely hilarious whenever a header file is longer than the corresponding `.cc` file 🤣

andyj10224 (approved):
LGTM! Great work and thanks for bearing with me! I do have a few minor comments regarding efficiency, but you can take it or leave it 👏

psi4/src/psi4/libmints/zora.cc
NathanGillispie: You know, I would think that too. I tested the two and on my machine the results were not statistically different to a 95% CI 😳. I bet the compiler is optimizing the function calls out. I'll just leave this as is.

andyj10224: Good to know! Thank you for running the tests!

susilehtola: it is not surprising, since `add` is defined in the header of `matrix.h` as
```
    void add(int h, int m, int n, double val) {
#ifdef PSIDEBUG
        if (m > rowspi_[h] || n > colspi_[h ^ symmetry_]) {
            outfile->Printf("out of bounds: symmetry_ = %d, h = %d, m = %d, n = %d\n", symmetry_, h, m, n);

            throw PSIEXCEPTION("What are you doing, Rob?");
        }
#endif
        matrix_[h][m][n] += val;
    }
```
so it does get inlined.

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+        int index = block->index();
+
+        double* x = block->x();
+        double* y = block->y();
+        double* z = block->z();
+
+        auto veff_block = std::make_shared<Vector>(npoints);
+
+        for (int a = 0; a < natoms; a++) {
+            int Z = molecule_->Z(a);
```
TiborGY:
```suggestion
            const int Z = molecule_->Z(a);
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,277 @@
...
+            if (Z > 104) throw PSIEXCEPTION("Too heavy atom for ZORA implementation. Tabulation only available to Z=104, Rf.\n");
```
TiborGY: Does this error message reach the user? I seem to recall that throwing from an OMP parallel region can be a bit fussy.
Perhaps it would be better to check molecule_->Z with std::any_of before entering the parallel region.

NathanGillispie: I do get the error on my computer, but the message seems to be cut off by abort. After putting it outside the OMP block it does indeed seem to give the correct error message with traceback and so on. Thanks for the suggestion.

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,283 @@
...
+
+void ZORA::setup() {
+    outfile->Printf("\n        ____  ____  ___  ___\n");
+    outfile->Printf("       /_  / / __ \\/ _ \\/ _ |\n");
+    outfile->Printf("        / /_/ /_/ / , _/ __ |\n");
+    outfile->Printf("       /___/\\____/_/|_/_/ |_|\n");
+    outfile->Printf("        by Nathan Gillispie\n");
+    outfile->Printf("      ========================\n");
+
+    timer_on("Make Grid");
```
loriab:
```suggestion
    timer_on("ZORA: Build Grid");
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,283 @@
...
+        {"DFT_BS_RADIUS_ALPHA",   1.0},
+        {"DFT_PRUNING_ALPHA",     1.0},
+        {"DFT_WEIGHTS_TOLERANCE", 1e-15},
+        {"DFT_BLOCK_MAX_RADIUS",  3.0},
+        {"DFT_BASIS_TOLERANCE",   options_.get_double("ZORA_BASIS_TOLERANCE")},
+    };
+
+    grid_ = std::make_shared<DFTGrid>(primary_->molecule(), primary_, grid_int_options, grid_str_options, grid_double_options, options_);
+
+    timer_off("Make Grid");
```
loriab:
```suggestion
    timer_off("ZORA: Build Grid");
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,283 @@
...
+
+    // Basis function computer on each thread
+    std::vector<std::shared_ptr<BasisFunctions>> pworkers;
+    for (int i = 0; i < nthreads; i++) {
+        auto p_tmp = std::make_shared<BasisFunctions>(primary_, max_points, max_funcs);
+        p_tmp->set_deriv(1);
+        pworkers.push_back(p_tmp);
+    }
+
+    timer_on("Effective Potential");
```
loriab:
```suggestion
    timer_on("ZORA: Effective Potential");
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,283 @@
...
+    veff_ = std::make_shared<std::map<int, SharedVector>>();
+    if (options_.get_bool("ZORA_NR_DEBUG")) {
+        compute_debug_veff();
+    } else {
+        compute_veff();
+    }
+    timer_off("Effective Potential");
```
loriab:
```suggestion
    timer_off("ZORA: Effective Potential");
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,283 @@
...
+
+    timer_on("Scalar Relativistic Kinetic");
```
loriab:
```suggestion
    timer_on("ZORA: Scalar Relativistic Kinetic");
```

psi4/src/psi4/libmints/zora.cc
```diff
@@ -0,0 +1,283 @@
...
+    compute_TSR(pworkers, T_SR);
+    timer_off("Scalar Relativistic Kinetic");
```
loriab:
```suggestion
    timer_off("ZORA: Scalar Relativistic Kinetic");
```

doc/sphinxman/source/relativistic.rst
```diff
@@ -31,15 +31,141 @@
...
+Keywords
+^^^^^^^^
+
+.. include:: autodir_options_c/globals__relativistic.rst
+.. include:: autodir_options_c/globals__zora_radial_points.rst
+.. include:: autodir_options_c/globals__zora_spherical_points.rst
+.. include:: autodir_options_c/globals__zora_pruning_scheme.rst
+.. include:: autodir_options_c/globals__zora_basis_tolerance.rst
+.. include:: autodir_options_c/globals__zora_nr_debug.rst
+
```
loriab, line 161: perfect -- this will inline the read_options into the docs

yet another indictment of the options handling that these zora options are in the globals section of r_o

doc/sphinxman/source/relativistic.rst
```diff
@@ -31,15 +31,141 @@
 .. index::
    relativistic
 
-.. _`sec:relativistic`:
-
-Scalar relativistic Hamiltonians
 ================================
+Scalar Relativistic Hamiltonians
```
loriab, line 37:
```suggestion
.. _`sec:relativistic`:

Scalar Relativistic Hamiltonians
```
leave the sec marker just in case someone has an html link to it

doc/sphinxman/source/relativistic.rst
NathanGillispie: Yes, I was finally able to compile the documentation to make sure.

loriab:
If you rebase, that'll (1) address the conflict above and (2) fix the docs and 3rd Azure lane issue that was fixed in an earlier merge. After that, it's rtg

loriab:
drat, let me look
```
2025-12-06T22:44:02.0057485Z reading sources... [100%] tutorial
2025-12-06T22:44:02.0057562Z reading sources... [100%] variables
2025-12-06T22:44:02.0057568Z 
2025-12-06T22:44:02.0058050Z /home/runner/work/psi4/psi4/code/objdir/doc/sphinxman/source/relativistic.rst:34: CRITICAL: Missing matching underline for section title overline.
2025-12-06T22:44:02.0058060Z 
2025-12-06T22:44:02.0058134Z ================================
2025-12-06T22:44:02.0058209Z .. _`sec:relativistic`:
2025-12-06T22:44:02.0058311Z looking for now-outdated files... none found
2025-12-06T22:44:02.0058394Z pickling environment... done
```

PR #3296, opened 2025-05-25
Add DFDIRJ+COSX Gradients

philipmnel:
## Description

This PR implements chain-of-spheres (COSX) exchange gradients (see https://doi.org/10.1063/1.3646921) and direct density-fitted Coulomb gradients, accessed with the `"SCF_TYPE": "DFDIRJ+COSX"` keyword. The COSX intergral derivatives can be computed as:

$(\mu' \nu | \rho \sigma) =  \sum_g X_{\mu g}' A_{\rho \sigma} X_{\nu g}$

$A_{\rho \sigma} = \int \frac{\rho(r) \sigma (r)}{|r-r_g|} dr$

$X_{\mu g}' = w_g^{1/2} \phi_{\mu}'(\mathbf{r}_g)$

The only extra quantities needed beyond those necessary to compute the exchange energy are the derivatives of the basis functions at each grid point, $\phi_{\mu}'(\mathbf{r}_g)$. Therefore, the computation of the exchange gradient is done alongside the computation of the energy (in `libfock/COSK.cc`) and then saved to the wavefunction. In COSX, the energy is converged on a small grid, then recomputed on a larger one. The gradients are only computed on this final grid. The Coulomb contribution to the gradient is computed in `scf_grad/dfjcosk_grad.cc`.

- [x] Enables SCF nuclear gradients for DFDIRJ+COSX.

## Dev notes & details

- [x] Enables DFDIRJ+COSX gradients.
- [x] Exchange gradients computed with energies, then saved to wavefunction (`"COSK_EXCHANGE_GRADIENT"`).
- [x] Option `COSX_DO_GRADIENT` is set in `proc.py` so that the initial energy call will know if the exchange gradients are requested.
- [x] Exchange gradients should ideally be computed once, on the final COSX grid. However, Psi4 allows the user to set an arbitrary number of SCF cycles to be computed on the final grid (`COSX_MAXITER_FINAL`, default: 1). `scf_iterator.py` will set the gradients to be computed on the final iteration. If `COSX_MAXITER_FINAL` is set to -1 (full convergence on final grid), gradients will be computed every iteration.
- [x] Test added in `tests/pytests/test_compositejk.py`. 

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/psi4/libfock/COSK.cc
```diff
@@ -344,16 +344,24 @@ void COSK::build_G_component(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
...
+    std::vector<std::vector<SharedMatrix>> KTgrad(njk, std::vector<SharedMatrix>(nthreads_));
 
     // initialize per-thread objects
     IntegralFactory factory(primary_);
     for(size_t thread = 0; thread < nthreads_; thread++) {
         int_computers[thread] = std::shared_ptr<ElectrostaticInt>(static_cast<ElectrostaticInt *>(factory.electrostatic().release()));
         bf_computers[thread] = std::make_shared<BasisFunctions>(primary_, grid->max_points(), grid->max_functions());
+        if (do_gradient_) {
+            bf_computers[thread]->set_deriv(1);
+        }
```
susilehtola, line 356:
```suggestion
bf_computers[thread]->set_deriv(do_gradient_);
```

psi4/src/psi4/libfock/COSK.cc
```diff
@@ -344,16 +344,24 @@ void COSK::build_G_component(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
 
     // per-thread K Matrix buffers (for accumulating thread contributions to K)
     std::vector<std::vector<SharedMatrix>> KT(njk, std::vector<SharedMatrix>(nthreads_));
+    std::vector<std::vector<SharedMatrix>> KTgrad(njk, std::vector<SharedMatrix>(nthreads_));
```
susilehtola, line 347: should the initialization be conditional on the gradient?

psi4/src/psi4/libfock/COSK.cc
```diff
@@ -344,16 +344,24 @@ void COSK::build_G_component(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
...
     IntegralFactory factory(primary_);
     for(size_t thread = 0; thread < nthreads_; thread++) {
         int_computers[thread] = std::shared_ptr<ElectrostaticInt>(static_cast<ElectrostaticInt *>(factory.electrostatic().release()));
         bf_computers[thread] = std::make_shared<BasisFunctions>(primary_, grid->max_points(), grid->max_functions());
+        if (do_gradient_) {
+            bf_computers[thread]->set_deriv(1);
+        }
         for(size_t jki = 0; jki < njk; jki++) {
             KT[jki][thread] = std::make_shared<Matrix>(nbf, nbf);
+            KTgrad[jki][thread] = std::make_shared<Matrix>(natom, 3);
```
susilehtola, line 359: same question here

psi4/src/psi4/libfock/COSK.cc
```diff
@@ -344,16 +344,24 @@ void COSK::build_G_component(std::vector<std::shared_ptr<Matrix>>& D, std::vecto
...
         }
     }
+    
+    // exchange gradient
+    auto Kgrad = std::make_shared<Matrix>("Exchange Gradient", natom, 3);
```
susilehtola, line 364: and here

psi4/src/psi4/libfock/COSK.cc
philipmnel: If you define a variable within an if-statement its scope will be limited to that if-statement block. So not possible to make this conditional on `do_gradient`.

psi4/src/psi4/libfock/COSK.cc
philipmnel: Moved to final if-statement block, since this is only used there.

psi4/src/psi4/libfock/COSK.cc
susilehtola: Well you could declare it but only initialize it if `do_gradient`. However, the size is `njk*nthreads_*sizeof(std::vector)` which is probably small enough to ignore, since the other way does involve more complicated code.

tests/pytests/test_compositejk.py
```diff
@@ -333,3 +333,30 @@ def test_j_algo_bp86(j_algo, k_algo, df_basis_scf, mols):
...
+                  },
+                  id="h2o (rhf)"),
+    pytest.param({"molecule" : "nh2",
+                  "options": {"reference" : "uhf"},
+                  "ref" :  [[ 0.0,  0.0,         0.01517592],
+                            [ 0.0,  0.00610143, -0.00739917],
+                            [ 0.0, -0.00610143, -0.00739917]]
+                  },
+                  id="nh2 (uhf)"),
+])
```
andyj10224, line 352: Would it be too difficult to also include an ROHF case?

andyj10224 (approved):
LGTM! @philipmnel If you can include an ROHF case, that would be awesome!

tests/pytests/test_compositejk.py
andyj10224: Also, make sure to rebase with the current Psi4 branch

PR #3297, opened 2025-05-25
Split JK gradient algorithms to separate files

philipmnel:
## Description
Splits `jk_grad.cc` into two files, `dfjk_grad.cc` and `directjk_grad.cc`. This PR only moves code. The purpose of this PR is to make the gradient code a bit more navigable, particularly considering new JK gradient algorithms added (#3296), rather than having one file with 3000+ LOC.

n/a

## Dev notes & details

- [x] Moves `DirectJKGrad` class to `directjk_grad.cc`
- [x] Moves `DFJKGrad` class to `dfjk_grad.cc`

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
1. Is there a preferred merge order between this and #3296?
2. Is this a target for 1.10 or 1.11? The final call is up to @loriab.

susilehtola:
As a side note, one should try to unify the Cholesky code with the density fitting routines. The most cost-efficient way to implement Cholesky is to do it in two steps: 1. identify the pivot functions and 2. do density fitting in the basis defined by the pivot functions.

The slight complication over normal density fitting is that the fit functions are now orbital products (possibly on 2 centers), but the equations are exactly the same as in the usual DF case.

loriab:
Thanks for the moving-code PR -- that'll make subequent review much easier. My only surprise was that the original file remained with ~100 loc. I wonder if the dfjk class should stay in that file and only the directjk_grad get split out. Whatever you and others think best. I checked with the two nearest of DLP's PRs, and they don't touch these files, so I'm ok with getting the split merged.

PR #3308, opened 2025-06-17
convey libgcc windows

loriab:
## Description

~There's been a strange error message the past few weeks on Windows at runtime that's fixed by `libgcc<14.3`. Let's test the rebuild of g2g 2.0.8 and see if this might have fixed this issue, as g2g was the oldest dep on c-f.~

- [x] conda env spec yaml files including addons are now saved in the repo

## Dev notes & details

- [x] new adcc version. averts pkg_resources/setuptools restriction
- [x] update reason openfermionpsi4 is blocked (it's fine but restricts the whole conda env to numpy v1)
- [x] now save full conda env with addons yaml files
- [x] convey some windows+conda peculiarities through codedeps rather than GHA sed's
Later
- [x] deployed v1.0.6 Einsums and new conda envs (probably in another PR too)
- [x] v2rdm stopped linking the with "no files found" for Intel Mac. no reason. I just disabled it since on its way out
- [x] the middle (3/4) Linux lane started reliably failing on two of the new GRAC tests with "SVD did not converge" even though it passed them fine three times last week. I added a nan_to_num call b/c there was at least one NaN and lots o e-308 . Austin didn't add that code, just triggered it in a new way. it should be investigated further
- [x] I got rid of the KMP stuff and added lots to detect and coney OpenMP. gauxc is using different (more modern) detection and needs help to see the conda files. If it sees the apt-get omp files, that's when multiple get linked in and triggers the runtime error

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/devtools/conda-envs/osx-64-buildrun-addons.yaml
+++ b/devtools/conda-envs/osx-64-buildrun-addons.yaml
@@ -5,17 +5,15 @@ channels:
 dependencies:
     # build
   - c-compiler
-  - clangdev=17.0.6           # req'd with cxx-compiler; Avoid omp problem with v19, v20.
   - cmake
   - cxx-compiler
-  - libgcc<14.3               # req'd with cxx-compiler; Avoid 'specified procedure could not be found' DLL error.
   - ninja
     # non-qc buildtime required
   - blas-devel                # req'd with libblas
   - eigen                     # req'd with libint
-  - intel-openmp
   - libblas=*=*mkl
   - libboost-headers          # req'd with libint
+  - llvm-openmp
   - numpy
   - pip                       # req'd with python; Package installer not strictly needed but needs to be tied to this python if used.
   - pybind11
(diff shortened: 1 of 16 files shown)


devtools/conda-envs/osx-64-buildrun-addons.yaml
```diff
@@ -0,0 +1,66 @@
...
+  #- psi4::libint              # opt'l with libint; Optionally for linux-64, an AM=7 build is available here.
+  - libint
+  - libxc-c
+  - optking
+  - qcelemental
+  - qcengine
+    # buildtime optional
+  - ambit
+  - dkh
+  - einsums>=1.0.3
```
jturney: Should this go ahead and be bumped to 1.0.6?

loriab: They're auto-generated from `conda/psi4-path-advisor.py deploy`, so I'll catch them all next run. I'm pretty sure these envs will solve to 1.0.6 in the meantime anyways.

jturney (approved):

zachglick (approved):

PR #3315, opened 2025-08-07
[DRAFT] DLPNO-CCSDT(Q)!!!

andyj10224:
## Description
This is a draft of the DLPNO-CCSDT(Q) method that will eventually be available in Psi4, described in these two works:

[DLPNO-CCSDT](https://doi.org/10.1021/acs.jctc.4c01716)
[DLPNO-CCSDT(Q)](https://doi.org/10.1063/5.0257672)

This is currently research-quality software for anybody seeking to use DLPNO-CCSDT(Q) for their research projects, a sample input file looks like:

  ```
  memory 2 GB

   molecule h2o {
   0 1
   O
   H 1 0.96
   H 1 0.96 2 104.5
   symmetry c1
   }
   
   set basis cc-pvdz
   set scf_type direct
   set freeze_core True
   set pno_convergence very_tight
   
   energy('dlpno-ccsdt(q)')
  ```

For all average users, we DO NOT recommend changing any of the default parameters for DLPNO-CCSDT(Q), which are:
```
pno_convergence very_tight 
t_cut_pno 1.0e-8
t_cut_pairs 1.0e-8
t_cut_tno 1.0e-9 # for (T)
t_cut_tno_full 1.0e-7 # for full T
t_cut_qno 3.33e-7
```
The above parameters are defaults and do not need to be manually set by the user. 

## Practical Advice

* DLPNO-CCSDT/(Q) makes the impossible possible! Platinum-standard theoretical benchmarks that would
  normally require millions or billions of years with canonical CCSDT(Q) are now possible with less
  than a couple weeks of computation time (or a day if you have access to lots of RAM/CPUs).

* For most systems (other than very sparse water clusters), asymptotic linear-scaling behavior 
  for DLPNO-CCSDT/(Q) will not be observed in the range of computations currently possible on the 
  typical lab workstation.

* DLPNO-CCSDT/(Q) computations can still take a while, after all, it is designed to accurately
  reproduce the result of one of the most expensive quantum chemistry algorithms in the world, to within
  a relative energy error of ``0.1 kcal mol^{-1}``. It probably is not (yet) possible to run a protein
  calculation with DLPNO-CCSDT/(Q).

* DLPNO-CCSDT/(Q) loves RAM and disk! It consumes computer resources like a growing adolescent. The computations
  you will be able to perform will be largely dictated by the computational resources you have access to!
  As a rule of thumb, a typical lab workstation can handle a benzene dimer computation with a polarized double
  zeta basis set!

* Always use ``VERY_TIGHT`` ``pno_convergence`` for a DLPNO-CCSDT/(Q) computation, and ``t_cut_pairs``
  is set to ``1.0e-8`` to maintain the accuracy needed for typical applications. This is done by default, and users
  who attempt to use custom parameters for DLPNO-CCSDT/(Q) do so at their own peril!

* Users are encouraged to use DLPNO-CCSDT(Q) as part of composite or focal point schemes, with 
  the T(Q) correction being computed through the difference of the DLPNO-CCSDT(Q) energy and a
  canonical DF-CCSD(T) or rank-corrected DLPNO-CCSD(T) energy (either canonical MP2,
  tighter ``t_cut_tno`` cutoff, or a CPS extrapolation). This approach has been shown to yield
  the best error cancellation!

* In DLPNO methods, it is recommended to freeze core orbitals (by setting ``freeze_core``
  to ``True``), since core excitations are known to be more sensitive to PNO truncations than
  valence truncations.

## Compilation instructions
```
 git clone https://github.com/andyj10224/psi4.git && cd psi4
 git checkout dlpno_ccsdt_q
 git remote add upstream https://github.com/psi4/psi4.git
 git fetch upstream "refs/tags/*:refs/tags/*"
 eval $(conda/psi4-path-advisor.py env --python 3.12 --name p4dev --disable docs)
 eval $(conda/psi4-path-advisor.py cmake)
 eval $(objdir_p4dev/stage/bin/psi4 --psiapi)
 psi4 --test
```


TiborGY:
Wow, I have not seen these papers yet, fantastic work! And thank you for sharing this draft.

andyj10224:
This is now stale with #3506

PR #3317, opened 2025-08-18
Pybind options

AlexHeide:
## Description

This PR moves optking's type and value validation for options from Psi4 to pydantic.

Allowed options are detected at compile time. Removes need to sync options. (REVERTED)

Some tests needed to be updated after changes to the IRC code. 

- [x] This PR accompanies the 0.4 optking release which includes additional linear coordinate handling and IRC fixes. 
- [x] Alphabetical keywords list has now moved into optking docs page.
- [x] irc_forward and irc_backward were not being written since 1.7. `irc_traj.<pid>.xyz` file will now be written for irc jobs and `opt_traj.<pid>.xyz` file can also be requested for non irc jobs.
- [x] Setting a coordinate constraint to a whitespace containing string (i.e. `set frozen_bend (" ")`) no long works. Please use `set frozen_bend ""` to erase a previously set value.
- [x] Adds 'opt' to allowed cli test groups 

## Dev notes & details

- [x] An embedded python interpreter is used to check optking's pydantic model for fields and the types 
of those fields to be added to the options object. (REVERTED)
- [x] Some Psi4 specific options are left as module options.
- [x]  Had to replace sphinx substitutions for optking with explicit links.
- [x] A sign error was hiding the fact that the current forward / backward direction doesn't always match old optking.
- [x] Addition of linear coordinates during IRC affects termination for HCN. Test updated.
- [x] Optking is pydantic v1 and v2 ready. optking.op is assigned to v1/optparams or v2/optparams as appropriate given detected pydantic version so updating Psi4's environment shouldn't break anything on optking's end. (Still true but since read_options no longer references optparams, less relevant to Psi devs)
- [x] Removes `python setup.py` from optking/Cmakelists.txt. Prepares for removal of `setup.py`. I'm delaying optkings move to pyproject.toml until after this PR so there'll be a deprecation notice for the time being.

## Checklist
- [x] Tests added for any new features (All relevant optking options explicitly tested in `test_optking.py`)

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/src/read_options.cc
```diff
@@ -30,13 +30,17 @@
...
+#include <cstdio>
 #include "psi4/physconst.h"
 #include "psi4/psi4-dec.h"
 #include "psi4/psifiles.h"
 
 #include "psi4/liboptions/liboptions.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
 
+#include "pybind11/pybind11.h"
+#include "pybind11/embed.h"
```
loriab: Alex and I have been talking about this. My initial reaction was the below. I think he's going to revise this (or open a separate PR) reverting back to static options to avoid this. And maybe we can do it better with Jonathon's options rework.

`embed.h` is necessary? this links `libpython.so` into the `psi4/core.so`, and we try pretty hard to avoid this. (got rid of it in 2016 and this year had to prevent einsums for inadvertently bringing it back.)

JonathonMisiewicz (approved):

psi4/src/read_options.cc
```diff
@@ -2775,214 +2775,238 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        options.add_bool("ADD_AUXILIARY_BONDS", true);
+        /* Skipping ALG_GEOM_MAXITER */
+        /* This factor times standard covalent distance is used to add extra stretch coordinates. */
+        options.add_double("AUXILIARY_BOND_FACTOR", 2.5);
+        /* SKIPPING BT_X KEYWORDS */
+        /* Do read Cartesian Hessian?  Only for experts - use
+        :py:attr:`~optking.v1.optparams.OptParams.full_hess_every` instead. */
+        options.add_bool("CART_HESS_READ", false);
+        /* */
+        options.add_str("CONJUGATE_GRADIENT_TYPE", "");
```
loriab:
```suggestion
        /*- :py:attr:`~optking.v1.optparams.OptParams.conjugate_gradient_type` -*/
        options.add_str("CONJUGATE_GRADIENT_TYPE", "");
```
not mandatory at all. I _think_ the above suggestion would be a way to get the normal autodoc file generated for the option w/o much code duplication. I don't have a strong opinion at the moment on whether that's worth it.

psi4/src/read_options.cc
```diff
@@ -2775,214 +2775,238 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
...
+        options.add_double("COVALENT_CONNECT", 1.3);
+        /* A string formatted as a dicitonary containing a set of coordinates. Coordinates can be
+            appended to Optking's coordinate set or used on their own - expected optking 0.5 */
         options.add_str("CUSTOM_COORDS", "");
-        /*- Do follow the initial RFO vector after the first step? -*/
-        options.add_bool("RFO_FOLLOW_ROOT", false);
-        /*- Root for RFO to follow, 0 being lowest (for a minimum) -*/
-        options.add_int("RFO_ROOT", 0);
-        /*- Starting level for dynamic optimization (0=nondynamic, higher=>more conservative) -*/
+        /* Starting level for dynamic optimization (0=nondynamic, higher=>more conservative) */
```
loriab: checking you know that `/*-` and `-*/` are special signals to autodoc option on the following line. fine to remove to consistency from rest, just checking.

tests/pytests/test_optking.py
```diff
@@ -271,3 +272,108 @@ def test_dimers_ne2_long(inp):
...
+    ('V3D_TORS_ANGLE_LIM', 0.01),
+    ('V3D_TORS_COS_TOL', 0.01),
+    ('WRITE_TRAJECTORY', True)]
+)
+def test_all_keywords(option, test_val):
+
+    import optking
+    params = optking.op.OptParams(**{option: test_val})
+    processed = params.to_dict(by_alias=True)
+    assert processed.get(option) == test_val
```
loriab, line 381: nice and reassuring. I imagine optking itself checks that bad values get sensible errors?

doc/sphinxman/source/optking.rst
```diff
@@ -41,6 +41,15 @@ Geometry Optimization
...
 *Module:* :ref:`Keywords <apdx:optking>`, `OPTKING <https://github.com/psi-rking/optking>`_
 
+.. note::
+
+   As of version `1.10`, the validation of OptKing's options has been moved from |PSIfour| into
+   OptKing. Any options can still be set directly in |PSIfour| like in previous versions; however,
+   validation of any OptKing keywords will not take place until optimize is called.
+   If an option is not available in your version of PSI4 please make sure that |PSIfour|
+   and OptKing are up-to-date. If an option is still not available, use the ``optimizer_keywords``
+   argument to :py:func:`~psi4.driver.optimize`.
```
loriab, line 51: I like the compromise. Glad psi4 isn't preventing any optking functionality now.

psi4/src/read_options.cc
AlexHeide: I'm not opposed if having optking's options in the main keyword list feels important to anyone - that seems fairly easy to maintain. Currently, there are three copies of optking's options `v1/optparams.py`, `v2/optparams.py`, and `psi4/read_options.cc` so I just wanted to avoid having to copy documentation over again.

psi4/src/read_options.cc
AlexHeide: Yes, I removed the '-' but kept the multiline comments because it seemed the easiest way to remove the keywords from docs but be able to add them back if desired.

tests/pytests/test_optking.py
AlexHeide: This is what the errors look like. The regex is fairly nasty (trying to prevent an accidental extra character being a problem). output.log does also attempt to direct users to go back and look at the examples in optking's docs for how to use the keyword.
```
!  unable to parse user provided options.                                          !
! 1 validation error for OptParams                                                 !
! RANGED_BEND                                                                      !
!   string does not match regex "^\s*(?:\W?\s*\d+\W?\s+\d+\W?\s+\d+\s*\W*s*\W?-    !
!     ?\d+\.\d+\W?\s+-?\d+\.\d+\s*\W*\s*)*$" (type=value_error.str.regex; pattern= !
!     ^\s*(?:\W?\s*\d+\W?\s+\d+\W?\s+\d+\s*\W*s*\W?-?\d+\.\d+\W?\s+-               !
!     ?\d+\.\d+\s*\W*\s*)*$)                                                       !
! RANGED_BEND:    ()                                                               !
```

```
2025-08-21:16:56:44,359 CRITICAL [psi4.optking.optwrapper:256] 
Encountered an <class 'pydantic.error_wrappers.ValidationError'> error parsing options 1 validation error for OptParams
RANGED_BEND
  string does not match regex "^\s*(?:\W?\s*\d+\W?\s+\d+\W?\s+\d+\s*\W*s*\W?-?\d+\.\d+\W?\s+-?\d+\.\d+\s*\W*\s*)*$" (type=value_error.str.regex; pattern=^\s*(?:\W?\s*\d+\W?\s+\d+\W?\s+\d+\s*\W*s*\W?-?\d+\.\d+\W?\s+-?\d+\.\d+\s*\W*\s*)*$)
2025-08-21:16:56:44,359 CRITICAL [psi4.optking.optwrapper:257] 
For valdidation errors please check the documenation to find an example of the desired
keyword being uitilized: https://optking.readthedocs.io/en/latest/options.html#options
```

AlexHeide:
c-f shows 0.4.2 as latest online, but Psi4 can't find it yet (nor can I find it with a local `conda search`. Should be good to go if we re-run this in an hour.

tests/pytests/test_optking.py
loriab: "OptParams" and "RANGED_BEND" seem like good and sufficient clues to me

psi4/src/read_options.cc
loriab: Definitely don't want to replicate the option descriptions. The suggestion would be only copying the names again. Bu tup to you.

tests/opt7/input.dat
```diff
@@ -40,7 +40,10 @@ molecule {
 }
 
 set optking {
-  ranged_distance = (" ")
+  # This doesn't work anymore. If it was truly an empty string it would pass validation but parser adds
```
JonathonMisiewicz: I think you mean "The commented-out option setting below doesn't work anymore."?

AlexHeide: Correct. `ranged_distance = (" ")` no longer works but `ranged_distance = ""` does.

psi4/src/read_options.cc
AlexHeide: Added links.

PR #3322, opened 2025-10-14
Added Hellmann-Feynman Gradients to NumIntHelper

johnppederson:
## Description

This pull request adds the ability to evaluate the Hellmann-Feynman gradient term for arbitrary potentials via numerical quadrature with the `psi4.core.NumIntHelper`.

At the 7 Feb Patkowski/Pederson/CDSgroup meeting, it was suggested that the user ought to be responsible for calculating the Hellmann-Feynman gradient component when supplying Nao x Nao matrices to the `external_potentials` keyword (#3291), since there is no way to determine how the gradient should be calculated from its potential integral matrix. The `psi4.core.NumIntHelper` can currently evaluate the potential integral matrices from an arbitrary potential. If these potential integral matrices are going to be able to be supplied to the SCF machinery, then adding support for evaluating the corresponding Hellmann-Feynman contribution to the gradient would be helpful to users.

- [x] The user may evaluate the Hellmann-Feynman gradient term associated with an arbitrary potential via numerical quadrature with `psi4.core.NumIntHelper.potential_gradient()`. The potential must be supplied as a `list` of `psi4.core.Vector`, where each vector object corresponds to a block in the `psi4.core.DFTGrid` used to instantiate the `psi4.core.NumIntHelper`. A single density matrix must also be supplied.

## Dev notes & details

- [x] Added function `NumIntHelper::potential_gradient()` to calculate the Hellmann-Feynman contribution to the gradient from an arbitrary potential via numerical quadrature. This function is primarily based on `RV::compute_gradient()` and `NumIntHelper::potential_integral()`. The function makes a call to a `gradient_integrator()` function, which is defined in the anonymous namespace of `libmints/numinthelper.cc` and is comparable to the `rks_gradient_integrator()` defined in `libfock
/dft_integrators.h`.
- [x] Added a test `numint1` to the test suite, which compares the numerical potential integral matrix (`psi4.core.NumIntHelper.potential_integral()`) and Hellmann-Feynman gradient (`psi4.core.NumIntHelper.potential_gradient()`) from a set of point charges against their analytic counterparts calculated by `psi4.core.ExternalPotential`.

## Checklist
- [x] All or relevant fraction of full tests run

## Status
- [x] Ready for review
- [ ] Ready for merge


jturney (approved):

tests/numint1/input.dat
```diff
@@ -0,0 +1,78 @@
...
+0.417,0.544757019107,0.0,-3.799961446760,
+0.417,0.544757019107,0.0,-0.912085762652]).reshape((-1, 4))
+# convert coordinates columns to bohr
+external_potentials[:,[1,2,3]] /= psi_bohr2angstroms
+
+set {
+    scf_type df
+    d_convergence 12
+    basis 6-31G*
+    print 5
```
TiborGY, line 29: Do we really need print 5 for this test to work? The reference output currently stands to commit 11.5k lines to the repo...
I would never be the one to oppose adding good reference data, but as-is this seems a bit excessive.

psi4/src/psi4/libmints/numinthelper.cc
```diff
@@ -183,4 +233,73 @@ SharedMatrix NumIntHelper::potential_integral(const std::vector<SharedVector>& g
...
+        parallel_timer_on("finish operator", rank);
+        gradient_integrator(basisset, block, grid_data[Q], worker, G_local[rank], D_local, U_local[rank]);
+        parallel_timer_off("finish operator", rank);
+    }  // grid blocks
+    // Sum up the matrix
+    auto G = std::make_shared<Matrix>("NumIntHelper Gradient", natom, 3);
+    for (auto const& val : G_local) {
+        G->add(val);
+    }
+    G->print();
```
TiborGY, line 301: Perhaps it would be a good idea to only print the entire gradient if `print > 0` ?

loriab: A lot of the Matrix `print()`s in libmints are behind debug or printlvl guards. Otoh, this array is the key point. So I'm with however way you choose on this, John. Any further changes or good to merge?

TiborGY: I guess my main gripe is that `print 5` dumps a lot of arrays into the output that are not being tested here, as far as I can tell. E.g. F, C and D gets dumped twice every SCF iteration.

loriab: Perhaps the extra info will be useful for debugging someday.

PR #3330, opened 2025-11-29
Fix redundant LibXC build caused by GauXC's ExchCXX dependency handling

TiborGY:
## Description

Currently, when GauXC is enabled at build time, the Psi4 build system builds it from source, but GauXC's build system then needs ExchCXX and the build of ExchCXX always ends up building its own copy of LibXC, even if a suitable binary package for LibXC has already been found by Psi4's CMake. This slows down any CI lanes where GauXC is enabled, and contributes to the intermittent CI failures due to timeouts on the lanes which enable GauXC.

This PR fixes the issue by passing the path to LibXC to the GauXC build system.

- [x] Building from source with GauXC enabled should be faster if GauXC is built from source but a suitable LibXC binary is found in the environment.

## Dev notes & details

- [x] Pass STAGED_INSTALL_PREFIX in CMAKE_PREFIX_PATH when building GauXC so the ExchCXX CMake can find the LibXC binaries
- [x] Bump LibXC version pin to latest (fix did not work with old version)
- [x] Accommodate GauXC fixing a typo in their API

## Checklist
- [x] No new features
- [x] All CI lanes run

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/codedeps.yaml
+++ b/codedeps.yaml
@@ -1457,7 +1457,7 @@ data:
       host: github
       account: wavefunction91
       name: GauXC
-      commit: 5c85f6b
+      commit: d9cdad3
     cmake:
       name: gauxc
       constraint: null


external/upstream/gauxc/CMakeLists.txt
```diff
@@ -91,7 +91,10 @@ if(${ENABLE_gauxc})
                          -DCMAKE_CXX_COMPILER_ARG1:STRING=${CMAKE_CXX_COMPILER_ARG1}
                          -DOpenMP_C_LIB_NAMES:STRING=${OpenMP_C_LIB_NAMES}
                          -DOpenMP_CXX_LIB_NAMES:STRING=${OpenMP_CXX_LIB_NAMES}
-                         -DBLAS_LIBRARIES:STRING=${BLAS_LIBRARIES})
+                         -DBLAS_LIBRARIES:STRING=${BLAS_LIBRARIES}
+                         # CMAKE_PREFIX_PATH propagates to GauXC's FetchContent builds (ExchCXX)
+                         # allowing ExchCXX to find Psi4's LibXC without rebuilding from source
+                         -DCMAKE_PREFIX_PATH:STRING=${STAGED_INSTALL_PREFIX};${CMAKE_PREFIX_PATH})
```
loriab, line 97: Thanks for taking the time to investigate this.

loriab (approved):
lgtm, thanks!

PR #3331, opened 2025-12-03
fix docs and plugins CI

loriab:
## Description

This fixes two CI failures. One (affecting azure lane 3) is described in a comment below. The second is a single warning in the docs build that fails the build b/c we have strict enabled. I don't think it's correctable in psi, and I do think it has to do with old (c. 2013) autodoc-pydantic mixed with new everything else. Since hopefully that package's end is nigh, I'm just pinning it. folks building docs themselves on mac will need to comment out a line in their env.yaml.

## Status
- [x] Ready for review
- [x] Ready for merge


andyj10224 (approved):
LGTM! Requesting other reviewers to move this along!

PR #3335, opened 2025-12-12
Fix potential nullptr dereference in DSYEV_descending

TiborGY:
## Description

DSYEV returns the eigenvalues and optionally eigenvectors sorted in the order of ascending eigenvalues.
DSYEV_descending swaps the order to descending without checking if the eigenvectors have been requested, resulting in a nullptr dereference if not. This PR fixes that corner case.

Thankfully, nothing in the current Psi4 codebase seems to be using this particular combination of "only eigenvalues, descending order", so the impact should be only on out-of-tree codes using Matrix::diagonalize with that particular setting.

- [x] PSI_API: fixed corner case in `Matrix::diagonalize`, no longer dereferences a nullptr if `nMatz == evals_only_descending`

## Dev notes & details

- [x] Make sure the eigenvector pointer is not nullptr before trying to swap its elements in DSYEV_descending

## Checklist
- [x] No new features
- [x] Tests run by CI are passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libciomr/dsyev_descending.cc
+++ b/psi4/src/psi4/libciomr/dsyev_descending.cc
@@ -56,7 +56,7 @@ namespace psi {
         std::swap(e_vals[i], e_vals[N - i - 1]);
     }
     // Reverse the order of columns of the row-major 2D eigenvector array
-    if (e_vecs != nullptr){
+    if (e_vecs != nullptr) {
         for (int64_t i = 0; i < N; i++) {
             for (int64_t j = 0; j < N / 2; j++) {
                 std::swap(e_vecs[i][j], e_vecs[i][N - j - 1]);


TiborGY:
> @TiborGY It seems like there is a CMakeError with Windows. Can you look into it? In the meantime, I will put #3334 into the merge queue

Seems to be completely unrelated to this PR, and to be honest the error message makes zero sense to me:

```
> CMake Error at C:/Program Files/CMake/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
>   Could NOT find Python (missing: Python_NumPy_INCLUDE_DIRS NumPy) (found
>   suitable version "3.12.9", minimum required is "3.10")
```
Found 3.12.9, but could not find Python?
This is such a "task failed successfully" type of message...
Anyhow I will try poking the CI with a trivial change

loriab:
fwiw, `Could NOT find Python (missing: Python_NumPy_INCLUDE_DIRS NumPy)` nearly always means that there was a connection or solver fluke, the conda env didn't complete, and numpy and all other deps are missing. poke or restart is the correct sol'n. occasionally, though, it does mean what it says (e.g., Mac + cmake 3.30).

PR #3337, opened 2025-12-12
Fixed several memory leaks tied to psimrcc

cgbriggs99:
## Description
I found some memory leaks in psimrcc. One of them was a cross-dependence between shared pointers that resulted in the wavefunction never being freed. Also patched a few smaller ones. There is still one issue where one set of matrices has values allocated with both new and malloc, so deleting them becomes undefined behavior no matter what I do.

- [ ] Fixed memory leaks related to psimrcc.
- [ ] Fixed a segfault in mcscf that prevented optimizing with psimrcc.

## Dev notes & details
- [ ] Fixed memory leaks related to psimrcc.
- [ ] Fixed deprecation warnings in psimrcc. I may go through some other modules as well if I have time and a lack of patience.
- [ ] Fixed a segfault when optimizing with psimrcc.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -170,6 +170,7 @@ jobs:
         fi
         if [[ "${{ runner.os }}" == "macOS" && "${{ matrix.cfg.label }}" == "Conda Clang (M-Intel)" ]]; then
           :
+          sed -E -i.bak "s;- memory_profiler;#- memory_profiler;g" env_p4build.yaml
           # sed -E -i.bak "s;;;g" env_p4build.yaml
         fi
         if [[ "${{ runner.os }}" == "macOS" && "${{ matrix.cfg.label }}" == "Conda Clang (M-Silicon)" ]]; then


psi4/src/psi4/libtrans/integraltransform_tei_2nd_half.cc
```diff
@@ -98,7 +98,10 @@ void IntegralTransform::transform_tei_second_half(const std::shared_ptr<MOSpace>
         }
     }
 
-    if (useIWL_) iwl = new IWL(psio_.get(), iwlAAIntFile_, tolerance_, 0, 0);
+    if (useIWL_) {
+        delete iwl;
```
andyj10224:
```suggestion
    if (iwl != nullptr) {
         delete iwl;
    }
```

psi4/src/psi4/psimrcc/manybody.cc
```diff
@@ -98,7 +98,7 @@ CCManyBody::~CCManyBody() {
 void CCManyBody::generate_integrals() {
     // CCSort reads the one and two electron integrals
     // and creates the Fock matrices
-    std::make_shared<CCSort>(wfn_, out_of_core_sort);
+    auto sort = std::make_shared<CCSort>(wfn_, out_of_core_sort);
```
andyj10224: Where is this variable used?

psi4/src/psi4/libtrans/integraltransform_moinfo.cc
```diff
@@ -482,10 +482,17 @@ void IntegralTransform::process_spaces() {
 
     // End by adding the AO orbital space - this is always needed
     spacesUsed_.push_back(MOSPACE_NIL);
-    spaceArray_.push_back(sopi_);
+
+    int *sopi_data = new int[sopi_.n()];
```
andyj10224, line 486: Is this freed later on?

psi4/src/psi4/psimrcc/manybody.cc
cgbriggs99: It's not, but it's needed to make sure that the shared pointer is active for the duration of the scope. Without it, the temporary variable may be allowed to be destroyed at any time.

psi4/src/psi4/libtrans/integraltransform_moinfo.cc
cgbriggs99: Here: https://github.com/psi4/psi4/blob/714461223fac38ceb4711210c7798c053ff4d53a/psi4/src/psi4/libtrans/integraltransform.cc#L329

cgbriggs99: I'm not a fan of this method, but it's the only way without a full rewrite of the classes.

psi4/src/psi4/psimrcc/manybody.cc
andyj10224: Okay, it may be helpful to add a comment here on why you are initiating the variable then :)

loriab:
```suggestion
    [[maybe_unused]] auto sort = std::make_shared<CCSort>(wfn_, out_of_core_sort);
```
perhaps this is the official way

loriab (approved):

andyj10224:
@cgbriggs99 if this is ready for merge, you can mark it in the checklist and I will add it to the queue!

cgbriggs99:
Looks like the mac test still doesn't work.

PR #3341, opened 2025-12-21
Py314 and QCSchema v2

loriab:

- [x] Python v3.14 is now useable. Some QCSchema v1 functionality may be unavailable due to pydantic constraints (i.e., Pydantic does not allow QCSchema v1 objects to be created with Py >=3.14)
- [x] `psi4.run_qcschema(..., return_version)` and `psi4 --return-version` learned new keyword `return_version=1|2` to specify what version QCSchema to return. Without this keyword, version in is version out. Old `run_json` and `psi4 --json` is removed. `run_qcschema(..., return_dict)` keyword is also available to get a dictionary rather than a model (class instance); `return_dict=True` can provide more capability for QCSchema v1 with Python v3.14. Note that `run_qcschema` is a top-level fn now accessible without longer `psi4.schema_wrapper.run_qcschema()` address.
- [x] For compatibility, now need pydantic >=2.11, qcelemental >=0.50.1, qcengine >=0.50.0, qcmanybody >=0.7.1, optking >=0.5.0, qcfractal >=?, dftd3-python >=1.3.0, dftd4-python >=4.1.0
- [x] For qcfractal integration, the deprecated `tag` and `priority` have been removed in favor of `compute_tag` and `compute_priority`.

## Dev notes & details
- [x] old `run_json` (`json_wrapper`) dropped -- hasn't been the modern mode for 6+ y . `psi4 --json` is dead
- [x] internal driver interactions with QCEngine (cbs, findif, mbe) converted to qcschema v2
- [x] note that qcschema v2 is not final -- maybe don't run masses of data if not prepared to reprocess and convert them for storage
- [x] pydantic v2 now required
- [x] `psi4.run_qcschema()` and `psi4` cmd now take a `return_version` to select qcschema version to return
- [x] above can also take in both v1 & v2 versions
- [x] note that b/c of pydantic restrictions, QCSchema v1 is not compatible with Python 3.14+ . if you deal in v1 dictionaries/json (pass it into the fn or cmd and demand it back with return_dict=T) instead of objects, we have a private workaround to keep functionality
- [x] schema tests have been duplicated or adapted to provide adequate testing
- [x] `test_qcschema_addons.py` now fully adapted
- [x] adcc test case `cn_*any_6` has root 6 alternating btwn two values on different runs, even locally. I've been excusing the tests for a long time for affected architectures. Now, excluding just the 6th root (of 6).
- [x] TODO qcfractal py314 -- this had to wait on qcel/qcng 0.50 which was only available by 24 May
  - [ ] TODO change 0.64 version logic to 0.70 after proper release
  - [ ] TODO modify M-Silicon to do qcf 0.64 and L to do qcf 0.70 . that's the way it is at the moment but it'll be different after proper qcf release
- [x] proper optking tag -- now v0.5.0, thanks Alex Heide and Rollin King
- [x] docs build switched to pydantic API v2 (autodoc-pydantic and lots of optking switches). cleaned up old nitpick exceptions, though had to add another extension to access both QCSchema v1 and v2 that can't exist in the same docs build upstream.
- [x] add docs (and maybe examples - depends on overlap with Austin PR) updating how to deal with fisapt and fragment chgmult. One can specify total and per-frag charge and multiplicity through the string input. Needn't fall back on from_arrays.
- [x]  `state_to_atomicinput` now takes `dtype` for schema_version
- [x] fixed in `sapt_proc` a molecule building step that ran afoul of new serialization in qcel 0.50.0rc4
- [x] convert internal procedures to qcschema v2 (only way for py314 to work)
  - [x] cbs
  - [x] findif
  - [x] nbody
  - [x] base single-point
- [x] convert outgoing calls to qcschema v2 (only way for py314 to work)
  - [x] calls to empirical dispersion converted through qcengine
  - [x] calls to optking
- [ ] TODO handle failed single rec in cbs/findif etc
- [x] add logic to `py_psi_print_out` to handle errors ocurring before outstream is defined. now get error msg instead of segfault (this was so useful I added it to other PRs that got merged ahead of this one)
- [x] add back optking option INTRAFRAG_HESS that had gotten lost
- [x] TODO fix tests after qcel rc5 serialization: extern1, extern2, fsapt_AtomicOutput_external_potentials, qcschema_keyword_error
- [x] TODO recheck output to json in tests/json/ files
- [x] REDO SPHINXMAN_STRICT=ON in `.github/workflows/docs-pr.yml`‎

## Feeder PRs
- https://github.com/conda-forge/einsums-feedstock/pull/40
- https://github.com/dftd3/simple-dftd3/pull/170
- https://github.com/conda-forge/dftd3-python-feedstock/pull/33
- https://github.com/conda-forge/libefp-feedstock/pull/13
- https://github.com/dftd4/dftd4/pull/341
- https://github.com/conda-forge/dftd4-python-feedstock/pull/40
- https://github.com/psi-rking/optking/pull/110
- https://github.com/conda-forge/cppe-feedstock/pull/33
- https://github.com/conda-forge/adcc-feedstock/pull/23
- https://github.com/conda-forge/pyddx-feedstock/pull/27
- https://github.com/conda-forge/gcp-correction-feedstock/pull/12
- https://github.com/MolSSI/QCElemental/compare/v0.30.2...v0.50.0
- https://github.com/MolSSI/QCEngine/compare/v0.34.2...v0.50.0
- https://github.com/MolSSI/QCManyBody/pull/43
- https://github.com/MolSSI/QCFractal/pull/1004

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


psi4/driver/procrouting/sapt/sapt_proc.py
```diff
@@ -421,18 +421,15 @@ def compute_GRAC_shift(molecule, sapt_dft_grac_convergence_tier, label):
...
         given_charge = mol_qcel_dict["molecular_charge"]
 
-        mol_qcel = qcel.models.Molecule(**mol_qcel_dict)
-        mol_given = core.Molecule.from_schema(mol_qcel.dict())
-
+        mol_given = core.Molecule.from_schema(mol_qcel_dict)
         mol_qcel_dict["molecular_charge"] += 1
-        mol_qcel = qcel.models.Molecule(**mol_qcel_dict)
-        mol_cation = core.Molecule.from_schema(mol_qcel.dict())
+        mol_cation = core.Molecule.from_schema(mol_qcel_dict)
```
loriab, line 432: Due to serialization changes for QCSchema v2 in QCElemental v0.50.0rc4, the processing & alteration of chgmult no longer works with the previous code. The replacement code is both simpler and works with v2.

psi4/driver/molutil.py
```diff
@@ -84,6 +85,7 @@ def _molecule_from_string(cls,
         fix_com=fix_com,
         fix_orientation=fix_orientation,
         fix_symmetry=fix_symmetry,
+        nonphysical=nonphysical,
```
loriab, line 88: extension from https://github.com/MolSSI/QCElemental/issues/396

tests/json/schema-1-ghost/input.py
```diff
@@ -6,7 +6,7 @@
 
 # Generate JSON data
 json_data = {
-    "schema_name": "qc_schema_input",
+    "schema_name": "qcschema_input",
     "schema_version": 1,
```
susilehtola, line 9: this crashes with py 314

tests/json/schema-1-gradient/input.py
```diff
@@ -6,7 +6,7 @@
 
 # Generate JSON data
 json_data = {
-  "schema_name": "qc_schema_input",
+  "schema_name": "qcschema_input",
   "schema_version": 1,
```
susilehtola, line 9: same here

tests/json/schema-1-ghost/input.py
loriab: Hmm, these two tests are working ok for me through ctest, pytest, and direct python run and with both py313 and py314. I'll check them on CI now.

If these were local runs for you (rather than fresh clone and env), from `objdir/` `rm -rf external/upstream/qc*` and `rm -rf stage/lib/qc*` and rebuild, as those can help clear out older installations.

JonathonMisiewicz:
Can you elaborate on the three marked todos?

loriab:
> Can you elaborate on the three marked todos?

Good question

- [x] TODO qcfractal py314 -- this had to wait on qcel/qcng 0.50 which was only available by 24 May

This is snowflake tests for Py3.14. The QCFractal PR with compatibility is https://github.com/MolSSI/QCFractal/pull/1004 posted today, and I'm testing it locally now. It wouldn't hurt to leave this outstanding and I can fix this in a separate PR once the QCFractal packages are properly published b/c it'd be helpful for QCF if I could publish pre-release c-f packages for _their_ CI testing.

- [ ] TODO handle failed single rec in cbs/findif etc

This is a longstanding weakness that I'd _like_ to solve now, but it's not a regression if I don't. Basically if, for instance, a findif gradient fails one energy job, what sort of error gets thrown or what object returned. I'm not going to be able to look at this in the next week-and-a-half

EDIT: that endorsed plugins adcc did manage to trigger an error, so I'll address that in a follow-up PR

 - [x] TODO recheck output to json in tests/json/ files

Working on it now. I wasn't sure when I wrote that in ~Nov what serialization was going to look like.  It has to do with all these lines in json/ tests. Will push.

```
    with open("output.json", "w") as ofile:                                                    #TEST
        json.dump(json_ret.json(), ofile, indent=2)       
```

psi4/driver/task_base.py
```diff
@@ -29,20 +29,24 @@
...
-from pydantic.v1 import Field, validator
+from pydantic import Field, field_validator
 
 import qcelemental as qcel
-from qcelemental.models import AtomicInput, AtomicResult, DriverEnum
-from qcelemental.models.results import AtomicResultProtocols
-qcel.models.molecule.GEOMETRY_NOISE = 13  # need more precision in geometries for high-res findif
+from qcelemental.util.importing import parse_version
+from qcelemental.models.v2 import AtomicInput, AtomicResult, DriverEnum, AtomicProtocols, ProtoModel
+from qcelemental.models import QCEL_V1V2_SHIM_CODE
```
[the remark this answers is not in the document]
loriab: this'll return

JonathonMisiewicz (approved):
This PR is too ginormous for me to look at anything closely, but not seeing any problems, I'll approve it.

Thanks for doing this.

psi4/driver/psifiles.py
```diff
@@ -209,10 +209,10 @@
...
 PSIF_DFOCC_MIABC_BBAA       =  293  # DFOCC M_iaBC
 PSIF_SAD                    =  300  # A SAD file (File for SAD related quantities
-PSIF_DLPNO_QAB_PAO          =  310  # DLPNO (Q|uv)
-PSIF_DLPNO_QIA_PNO          =  311  # DLPNO (Q_ij|m_ij a_ij)
-PSIF_DLPNO_QAB_PNO          =  312  # DLPNO (Q_ij|a_ij b_ij)
-PSIF_DLPNO_TRIPLES          =  313  # DLPNO X_{ijk}^{abc}
+PSIF_DLPNO_QAB_PAO          =  310  # A PSIO file to store all (AUX | PAO PAO) integrals in a DLPNO computation
+PSIF_DLPNO_QIA_PNO          =  311  # A PSIO file to store all (AUX | LMO PNO) integrals in a DLPNO computation
+PSIF_DLPNO_QAB_PNO          =  312  # A PSIO file to store all (AUX | PNO PNO) integrals in a DLPNO computation
+PSIF_DLPNO_TRIPLES          =  313  # A PSIO file to store all X_{ijk}^{abc} like terms in a DLPNO-CCSD(T) computation
```
susilehtola, line 215: is this change meaningful? It's not related to this PR...

susilehtola (approved):
I looked through the changes and I didn't see anything obviously wrong. This is needed and if it passes CI, it should be correct. I will sic Claude for a deeper look; maybe it finds something important.

susilehtola (commented):
Took a pass over this. Overall it reads as a faithful, well-organized migration — the bulk is mechanical pydantic v1→v2 (`.dict()`→`.model_dump()`, `.json()`→`.model_dump_json()`, `@validator`→`@field_validator`, `qcel.models.X`→`qcel.models.v2.X`), the v2 `specification` nesting, and `to_schema` dtype bumps, plus welcome cleanups (the long-dead `run_json`/`--json`, deprecated `tag`/`priority`). The `read_options.cc` diff is purely Sphinx doc-comment URLs (`optking.v1`→`v2`), no logic. The `return_version`/`return_dict` truth tables are genuinely nice. A few things I'd raise before merge:

### Worth addressing

**1. `inspect.stack()` caller-sniffing in `run_qcschema` (schema_wrapper.py).**
```python
if return_dict is True or inspect.stack()[1][3] == "_psi4_qcschema_cli":
    return_version = -12
```
Deciding whether to permit the private v1 shim by inspecting the caller's frame name is fragile (silently breaks if `_psi4_qcschema_cli` is renamed, wrapped, decorated, or reached through an intermediate frame) and `inspect.stack()` is expensive — it reads source files for every frame to get one name (`sys._getframe(1).f_code.co_name` is the cheap equivalent). Since `_psi4_qcschema_cli` is the *direct* caller and already passes `return_version=...`, it could just pass an explicit private flag (e.g. `_allow_v1_dict_shim=True`) instead. I'd prefer the explicit param over frame inspection.

**2. `except`-path `input_data.model_dump()` may mask the real error.**
```python
if not isinstance(input_data, dict):
    input_data = input_data.model_dump()
```
If `build_input_model` raises while `input_data` is a genuine pydantic-v1-API model (v1 on Py<3.14 uses `.dict()`), `.model_dump()` may not exist, so the handler raises `AttributeError` and hides the original exception. Could you confirm v1 inputs expose `.model_dump()` under the supported pydantic, or guard it?

**3. `state_to_atomicinput(dtype=1)` default is a Py3.14 footgun.** It's a public exported util with no internal callers, and `dtype=1` builds `AtomicInput_v1`, which can't instantiate on 3.14 — the exact failure this PR fixes. A user calling it without `dtype=2` on 3.14 hits the v1 wall. Could default to 2 (or a Py-aware default), or at least document that 3.14 requires `dtype=2`.

**4. codedeps pydantic note vs constraint.** Constraint is `pydantic>=2.11.0,<3` but the note says *"For Py v3.14, require Pydantic v2.12."* If 2.12 is actually required on 3.14, `>=2.11.0` under-constrains (a 3.14 env could resolve 2.11). Stale note, or should the floor be env-marker'd?

### Non-blocking / track-don't-block

- `BaseComputer.protocols` is now `Optional[Dict[str, Any]]` with the typed `AtomicProtocols` alternative commented (`# ONCE`/`# FUTURE`) — drops validation on protocols during the transition. Fine as temporary, but there are several such `# ONCE`/`# FUTURE`/`# v2:` breadcrumbs in the driver; worth a tracking issue so they don't rot.
- `--json`/`run_json` removal is reasonable (6+ yr deprecated) and is in the changelog checklist — just make sure release notes surface it since it's user-facing.
- adcc `cn_*any_6` root-6 exclusion: pragmatic and documented (oscillating root), not introduced here.
- Nit: leftover `# print(f"RETVER ...")` debug line in `schema_wrapper`.

No blocking correctness issues in the core logic from my read; #1 is the one I'd really like changed, and #2/#3 are quick confirms since they're the precise Py3.14/v1 edge cases this PR is about.

psi4/driver/psifiles.py
loriab: Yes, it keeps the python psifiles in sync with psifiles.h https://github.com/psi4/psi4/blob/master/psi4/include/psi4/psifiles.h#L283-L286 . A local docs build syncs them.

loriab:
> Took a pass over this. Overall it reads as a faithful, well-organized migration — the bulk is mechanical pydantic v1→v2 (`.dict()`→`.model_dump()`, `.json()`→`.model_dump_json()`, `@validator`→`@field_validator`, `qcel.models.X`→`qcel.models.v2.X`), the v2 `specification` nesting, and `to_schema` dtype bumps, plus welcome cleanups (the long-dead `run_json`/`--json`, deprecated `tag`/`priority`). The `read_options.cc` diff is purely Sphinx doc-comment URLs (`optking.v1`→`v2`), no logic. The `return_version`/`return_dict` truth tables are genuinely nice. A few things I'd raise before merge:
> 
> ### Worth addressing
> **1. `inspect.stack()` caller-sniffing in `run_qcschema` (schema_wrapper.py).**
> 
> ```python
> if return_dict is True or inspect.stack()[1][3] == "_psi4_qcschema_cli":
>     return_version = -12
> ```
> 
> Deciding whether to permit the private v1 shim by inspecting the caller's frame name is fragile (silently breaks if `_psi4_qcschema_cli` is renamed, wrapped, decorated, or reached through an intermediate frame) and `inspect.stack()` is expensive — it reads source files for every frame to get one name (`sys._getframe(1).f_code.co_name` is the cheap equivalent). Since `_psi4_qcschema_cli` is the _direct_ caller and already passes `return_version=...`, it could just pass an explicit private flag (e.g. `_allow_v1_dict_shim=True`) instead. I'd prefer the explicit param over frame inspection.

Good idea and implemented. Also updated to use the new `QCEL_V1V2_SHIM_CODE` imported from qcel rather than the raw `-12`.

> **2. `except`-path `input_data.model_dump()` may mask the real error.**
> 
> ```python
> if not isinstance(input_data, dict):
>     input_data = input_data.model_dump()
> ```
> 
> If `build_input_model` raises while `input_data` is a genuine pydantic-v1-API model (v1 on Py<3.14 uses `.dict()`), `.model_dump()` may not exist, so the handler raises `AttributeError` and hides the original exception. Could you confirm v1 inputs expose `.model_dump()` under the supported pydantic, or guard it?

Yes, this psi4 requires qcel >=0.50, and those QCSchema v1 models have been retrofitted with `model_dump()` https://github.com/MolSSI/QCElemental/blob/master/qcelemental/models/v1/basemodels.py#L121-L123

> **3. `state_to_atomicinput(dtype=1)` default is a Py3.14 footgun.** It's a public exported util with no internal callers, and `dtype=1` builds `AtomicInput_v1`, which can't instantiate on 3.14 — the exact failure this PR fixes. A user calling it without `dtype=2` on 3.14 hits the v1 wall. Could default to 2 (or a Py-aware default), or at least document that 3.14 requires `dtype=2`.

Yeah, right now this fn is for optking (https://github.com/psi-rking/optking/blob/master/optking/optwrapper.py#L122-L135) and since its aiming to remain compatible with pre- and post-this-PR psi4 for now, dtype needs to remain 1. Note added to docstring.

> **4. codedeps pydantic note vs constraint.** Constraint is `pydantic>=2.11.0,<3` but the note says _"For Py v3.14, require Pydantic v2.12."_ If 2.12 is actually required on 3.14, `>=2.11.0` under-constrains (a 3.14 env could resolve 2.11). Stale note, or should the floor be env-marker'd?

Fair point but codedeps (and machinery that uses it) doesn't know the python version. Moreover, the OpenForceField can't use >=2.12, so I don't want to over-constrain to catch py314. In practice, envs resolve to beyond pyd212, so I think the note is sufficient.

> ### Non-blocking / track-don't-block
> * `BaseComputer.protocols` is now `Optional[Dict[str, Any]]` with the typed `AtomicProtocols` alternative commented (`# ONCE`/`# FUTURE`) — drops validation on protocols during the transition. Fine as temporary, but there are several such `# ONCE`/`# FUTURE`/`# v2:` breadcrumbs in the driver; worth a tracking issue so they don't rot.

Part of several things to do when QCFractal 0.70 is minted. All in one file, so I think I can track.

> * `--json`/`run_json` removal is reasonable (6+ yr deprecated) and is in the changelog checklist — just make sure release notes surface it since it's user-facing.

Yep, already in RN section for frontmatter

> * adcc `cn_*any_6` root-6 exclusion: pragmatic and documented (oscillating root), not introduced here.

Glad you agree -- I'm pleased to be rid of the exception

> * Nit: leftover `# print(f"RETVER ...")` debug line in `schema_wrapper`.

killed it off

> No blocking correctness issues in the core logic from my read; #1 is the one I'd really like changed, and #2/#3 are quick confirms since they're the precise Py3.14/v1 edge cases this PR is about.

PR #3346, opened 2026-01-08
Pin the version of github-push-action used for docs, try to not force push

TiborGY:
## Description

Using ad-m/github-push-action@master in the action that pushes the updated docs is technically a software supply chain risk.
It handles sensitive data like a GH token, and could probably do bad things if the account of one of its maintainers was to be compromised. While the risk is remote, given how rarely it is normally updated, pinning its version would be good practice.

~~I don't see why the action was set to force push, but perhaps it would be worth seeing if we can do away with that.~~

## Dev notes & details

- [x] Pin ad-m/github-push-action to the recent 1.0 release

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -163,4 +163,4 @@ jobs:
         directory: ./docs
         repository: psi4/psi4docs
         github_token: ${{ secrets.psi4docs_from_psi4 }}
-
+        force: true
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -158,10 +158,9 @@ jobs:
 
     - name: Push Changes to psi4/psi4docs
       if: github.repository == 'psi4/psi4'
-      uses: ad-m/github-push-action@master
+      uses: ad-m/github-push-action@v1.0.0
       with:
         directory: ./docs
         repository: psi4/psi4docs
         github_token: ${{ secrets.psi4docs_from_psi4 }}
-        force: true
 


loriab (commented):
Great precaution to pin that version. I do suspect that the `force` is necessary to avoid dealing with conflicts since https://github.com/psi4/psi4docs/tree/master/sphinxman/master is overwritten for each push to GH:psi4/psi4 master . It's a little less serious to force-push to GH:psi4/psi4docs than if it was a gh-pages branch to GH:psi4/psi4 itself. I've just taken a recent snapshot of psi4docs in case of calamity. Do you still want to try dropping force?

TiborGY:
> Great precaution to pin that version. I do suspect that the `force` is necessary to avoid dealing with conflicts since https://github.com/psi4/psi4docs/tree/master/sphinxman/master is overwritten for each push to GH:psi4/psi4 master . It's a little less serious to force-push to GH:psi4/psi4docs than if it was a gh-pages branch to GH:psi4/psi4 itself. I've just taken a recent snapshot of psi4docs in case of calamity. Do you still want to try dropping force?

I dont immediately see why that necessitates the force push, but TBH I dont fully understand how this GHA runs, so I trust your judgement better than mine on this.

loriab (approved):

JonathonMisiewicz (approved):

PR #3353, opened 2026-02-01
Libint v2.13 (cmake+cmake)

loriab:

- [x] Require Libint >= v2.13 to consolidate code and ensure detected pre-built Libint has the integrals Psi4 needs. Linking against >= v2.8.1 is still possible if one un-does the mostly checking and high-level (configuration and citation) connections in this PR.

## Dev notes & details

- [x] EFV is releasing a new Libint2 version that's cmake+cmake in build https://github.com/evaleev/libint/pull/389 . Provided you're _detecting_ L2 (not internally building), Psi4 needs no changes to use it. To take full adv of 2.13, this PR: 
    - [x] switches over some variable names for internal build
    - [x] now that cmake components are available, we check that Psi4 has the ints it needs of different classes, deriv, and AM through `find_package(Libint2 COMPONENTS ...)`
    - [x] now that c++ -side integral code components (e.g., `eri_ffff_d2` or `eri_ssI_d0` (3-center)) are always available, add py-side `psi4.driver.p4util.libint_supports(6)` so it's easy to check what's accessible py-side (deriv and centers args, too). `libint2::supports` is the direct c++ call. Previous to v2.13, the c++ was often available b/c I patched a big string into the c-f builds but wasn't guaranteed for libtool+cmake builds
    - [x] citation printed and consolidated the output file reporting of L2 capabilities. (The removed `eri_c4_d0_l2`-style was from my old cmake+cmake versions before we returned to upstream c. v1.9.1 ~December 2023. The new-style equivalent is `eri_dddd_d0`)
- [x] fill in codedeps.yaml
- [x] fill in tarball URLs
- [x] boost AM on c-f for FAE
- [x] I was getting unrelated failures since corrected (qcengine not detecting psi4 b/c python freethreading was printing a warning; maybe only on py314) when conda solved to the freethreading variant (>=py313). since we don't know if psi4 is safe yet, I'm adding the `python-gil` constraining package
- [x] In putting #3388 together, I had bad problems with OOM again on Linux with libmints. I solved this by dialing down core target to batch size of 5 and further restraining l2intf to `one_job` at a time. After rebasing this, similar problems appeared for GCC. I think what's happened is that incidentally here I'm no longer testing an old `psi4::libint=2.8.1` high-AM build on Azure linux, and the additional complexity of the L2 2.13 header is too much for "old" (c. 2021) GCC. I've introduced `psi4_libmints_UNITY_BUILD_BATCH_SIZE` to throttle libmints, and that works for latest GCC 13.3 . The GCC 11.4/11.5 lane I can't get to complete with any unity build on. :-( So it's single-thread GNU Make with reduced testing load so it doesn't run much longer than the other lanes. For ref, `-O0` takes 45m/65m for build/compile; -O3 (default Release) is 59/28; and -O2 that I left it on is 52/28. Hopefully this won't hit too many innocents on laptops. Fortunately the conda gcc hasn't needed any mods.
- [x] borrowed the CXX from Susi's #3400 to minimize build time with -O2

## Status
- [x] Ready for review
- [x] Ready for merge


tests/fcidump/input.dat
```diff
@@ -17,6 +17,9 @@ if psi4.core.get_option("scf", "orbital_optimizer_package") == "INTERNAL":
 else:
     fci_tol = 8
 
+if psi4.core.get_option("scf", "orbital_optimizer_package") != "INTERNAL":
+    psi4.set_options({"e_convergence": 9, "d_convergence": 8e-8})
```
TiborGY, line 21: Do we know _why_ OOO is getting stuck at such levels of convergence? I think I saw tolerances loosened for OOO in other tests too. Naively, I would think that it would have as good or better convergence than the internal solvers, unless this is all a mirage stemming from e.g. the convergence statistic having a different definition or something. Or am I missing something here?

loriab: No we don't know why OOO is getting stuck, though Susi wasn't too surprised. We added a callback fn to OOO so I'm pretty confident OOO is using the same criteria as the internal.

susilehtola: One would have to look in-depth at what is going on; my guess is that the cleanup of the DIIS history may be the culprit. This can be checked by changing the variable value. OOO is also adamant about not going up in energy.

maxscheurer (approved):

tests/fcidump/input.dat
loriab: This'll be fixed in #3392 , so I'd give it a pass for now.

PR #3354, opened 2026-02-04
Fix deprecation warnings + cleanup in `libscf_solver`

JonathonMisiewicz:
## Description
This PR fixes the deprecation warnings that I found in `libscf_solver` and also does some miscellaneous code cleanup. Please pay attention to my changes in the level shift procedure.

## Dev notes & details

- [x] Fixes several deprecation warnings involving casts to pointer in `libscf`.
- [x] This PR does create a merge conflict with #3351 because the `HF::form_Fia` signature changes. The solution of this PR is preferred.

## Checklist
- [x] Quicktests pass

## Status
- [x] Ready for review
- [x] Ready for merge

PR #3355, opened 2026-02-04
GauXC Integration

JonathonMisiewicz:
## Description
This is the work-in-progress PR for allowing Psi4 to outsource Kohn-Sham integration to the GauXC program. This is both desired for Microsoft's Skala work (notifying @jhrmnn as the point-person from Microsoft) and to allow Psi4 to account for grid weights in nuclear derivative computations. The scope of the current PR is to add this support for RKS energy computations as proof-of-concept. Once that is done, adding support for UKS and gradient computations will be very straightforward.

I'll be pushing bite-sized commits here for ease of review, frequent review, and coordination with Microsoft. The older commits on history are various remove-warning PRs that are cherry-picked into history so that when I build locally, I get cleaner output. However, those should actually be merged in as separate PRs, such as #3354.

I'm fond of keeping commit history clean for future investigation, which often involves force pushes. Once the PR actually supports the target feature, I'll stop doing that. I'm happy to change if that would be more convenient on Microsoft's end.

## Current Status
- [ ] PR is currently **not** functional. It's just infrastructure.
- [ ] Next target is to set up the GauXC functional object.

- [x] Setting `GAUXC_INTEGRATE` to true dispatches Kohn-Shamm integration to GauXC. This is currently supported only for RKS energy computations.

## Dev notes & details

- [x] Adds `GAUXC_INTEGRATE`
- [x] Refactors `compute_V` into different compute algorithms for diff

## Checklist
- [ ] New tests will come when features are complete
- [x] [Quick tests pass](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)

## Status
- [x] Ready for review
- [ ] Ready for merge


psi4/src/read_options.cc
```diff
@@ -1929,6 +1929,10 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Save the UHF NOs -*/
         options.add_bool("SAVE_UHF_NOS", false);
 
+        /*- SUBSECTION GAUXCINT -*/
+        /*- Outsource all Kohn-Shame numerical integration to the GauXC program? -*/
+        options.add_bool("GAUXC_INTEGRATE", false);
```
loriab: fwiw, the pattern we've been following for alternate software to do the same task is https://github.com/psi4/psi4/blob/master/psi4/src/read_options.cc#L196-L202 . bool is fine for now but should eventually be revisited.

https://github.com/psi4/psi4/blob/master/.azure-pipelines/azure-pipelines-linux.yml#L106 is a handy way to get a non-default option broadly tested.

JonathonMisiewicz: Since we're now at _three_ packages to do the same task, this looks useful. Thanks!

psi4/src/read_options.cc
```diff
@@ -1929,6 +1929,10 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
         /*- Save the UHF NOs -*/
         options.add_bool("SAVE_UHF_NOS", false);
 
+        /*- SUBSECTION GAUXCINT -*/
+        /*- Outsource all Kohn-Shame numerical integration to the GauXC program? -*/
```
TiborGY:
```suggestion
        /*- Outsource all Kohn–Sham numerical integration to the GauXC program? -*/
```

JonathonMisiewicz:
I'll want to make sure that GauXC integration is operational before anything else, but I figure this was worth putting up for discussion now:

For an actual merge-able PR, I'm thinking I may need to rewrite the class structure... There'll be a central `IntegrationManager` which dispatches to either a `VBase` object (traditional Psi), BrianQC, or GauXC. The current union approach is confusion.

psi4/src/read_options.cc
susilehtola: GauXC is a library not a program

psi4/src/psi4/libfock/v.cc
```diff
@@ -670,6 +677,52 @@ void VBase::initialize() {
     }
 #endif
 }
+void VBase::initialize_gauxc() {
+#ifdef USING_gauxc
+    // TODO: Allow for Device execspace, depending on flags. This will add GPU support.
+    const auto gauxc_execspace = GauXC::ExecutionSpace::Host;
+    GauXC::LoadBalancerFactory lb_factory(gauxc_execspace, options_.get_str("SNLINK_LOAD_BALANCER_KERNEL"));
```
susilehtola: the keyword should probably be changed to GAUXC_LOAD_BALANCER_KERNEL

psi4/src/psi4/libfock/v.cc
```diff
@@ -670,6 +677,52 @@ void VBase::initialize() {
...
+    // generate map for radial quadrature schemes 
+    std::unordered_map<std::string, GauXC::RadialQuad> radial_scheme_map = { 
+      {"TREUTLER", GauXC::RadialQuad::TreutlerAhlrichs},
+      {"MURA", GauXC::RadialQuad::MuraKnowles},
+      {"EM", GauXC::RadialQuad::MurrayHandyLaming}
+    };
+    auto radial_scheme = options_.get_str("GAUXC_RADIAL_SCHEME");
+    auto grid_batch_size = options_.get_int("GAUXC_GRID_BATCH_SIZE");
+    auto radial_points = options_.get_int("SNLINK_RADIAL_POINTS");
+    auto spherical_points = options_.get_int("SNLINK_SPHERICAL_POINTS");
```
susilehtola: SNLINK or DFT?

psi4/src/psi4/libmints/basisset.cc
```diff
@@ -1283,3 +1283,47 @@ void BasisSet::convert_sap_contraction() {
...
+// converts a Psi4::BasisSet object to a GauXC::BasisSet object
+template <typename T>
+GauXC::BasisSet<T> BasisSet::to_gauxc_basisset(double basis_tol, bool force_cartesian) {
+    using prim_array = typename GauXC::Shell<T>::prim_array;
+    using cart_array = typename GauXC::Shell<T>::cart_array;
+
+    GauXC::BasisSet<T> gauxc_basisset(nshell());
+
+    for (size_t ishell = 0; ishell != nshell(); ++ishell) {
+        auto psi4_shell = shell(ishell);
```
susilehtola: I think one needs to also check that the number of primitives isn't greater than the length of the GauXC static arrays
```suggestion
        auto psi4_shell = shell(ishell);
        if(psi4_shell.nprimitive() > GauXC::detail::shell_nprim_max)
            throw std::runtime_error("Shell has " << psi4_shell.nprimitive() << " primitives but the shell datatype in GauXC is limited to " << GauXC::detail::shell_nprim_max << "!\nTo work around this issue, segment your basis set!\n");
```

psi4/src/psi4/libfock/v.cc
```diff
@@ -670,6 +677,52 @@ void VBase::initialize() {
     }
 #endif
 }
+void VBase::initialize_gauxc() {
```
awvwgk: Probably needs to have the [`MolecularWeightsFactory`](https://awvwgk.github.io/GauXC/api/cxx/molgrid/#_CPPv4N5GauXC23MolecularWeightsFactoryE) here as well.

JonathonMisiewicz:
All suggestions have been made _locally_, thanks, but because Psi is also massively out-of-date on code-formatting, I'm going to need to do some dedicated code formatting PRs so that the commit hooks we have don't clutter this PR to the point of unreadability when I push those fixes.

psi4/src/psi4/libfock/v.cc
```diff
@@ -182,7 +188,9 @@ void VBase::initialize() {
         // Need a functional worker per thread
         functional_workers_.push_back(functional_->build_worker());
     }
-    
+#ifdef USING_Gauxc
```
loriab:
```suggestion
#ifdef USING_gauxc
```

JonathonMisiewicz: I fixed that in a change that was lost during last night's tussle with CMake. Thanks!

awvwgk (commented):
Can Psi4 use the XCIntegrator with the Psi4 matrix class? As far as I can see the only functions expected by GauXC are:

```c++
const int64_t cols()
const int64_t rows()
value_type data()
void setZero()
```

And the possibility to construct the matrix based `matrix(ncols, nrows)` for returning the VXC component.

JonathonMisiewicz:
I'll continue on the internal refactoring over the weekend, whether the BrianQC maintainers respond to me or not.

JonathonMisiewicz:
The last commit updated BrianQC test values. Several settings were updated for tighter convergence, whether this meant tighter SCF densities, tighter geometry convergence, or tighter integration grids (especially for gradients).

JonathonMisiewicz:
Psi is currently having build issues, c.f. #3474. Work on this PR will resume when build issues are fixed. Should be soon.

awvwgk:
> [cd6dbe1](https://github.com/psi4/psi4/pull/3355/commits/cd6dbe12bae9f6fc92f156da0d1b86d52fcbadcf) Support GauXC compute_Vx. At time of writing, GauXC's implementation is not correct.

@JonathonMisiewicz Is this on the GauXC side? If so do you have a reproducer?

JonathonMisiewicz:
> > [cd6dbe1](https://github.com/psi4/psi4/pull/3355/commits/cd6dbe12bae9f6fc92f156da0d1b86d52fcbadcf) Support GauXC compute_Vx. At time of writing, GauXC's implementation is not correct.
> 
> @JonathonMisiewicz Is this on the GauXC side? If so do you have a reproducer?

The problem is definitely on GauXC's end, though I wanted to do some more probing before I sent the bug report. The symptom is that the following script passes when the functional is SVWN, but not when the functional is PBE - and it fails due to TDSCF failing to converge when using GauXC but not when using Psi4. If it was a problem in Psi's Davidson, our TD-SCF would never have converged. If it was a problem in my implementation, SVWN never would have passed.

I've determined that the problem is the <v[i], Av[j]> elements are non-Hermitian. Psi[ forces hermiticity in our Davidson algorithm](https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/solvers.py#L778), but then A_ij != <v[i], Av[j]>, which causes the Davidson algorithm to converge, even though there's a non-zero residual.

Note that this is *not* an apples-to-apples comparison. At time of writing, Psi4 neglects grid point derivatives terms. That's why I went to an absurd number of grid points for the comparison, to make those effects negligible.

[input.txt](https://github.com/user-attachments/files/31954482/input.txt)

susilehtola:
Note that grid point derivatives are included in #3458

JonathonMisiewicz:
> > [cd6dbe1](https://github.com/psi4/psi4/pull/3355/commits/cd6dbe12bae9f6fc92f156da0d1b86d52fcbadcf) Support GauXC compute_Vx. At time of writing, GauXC's implementation is not correct.
> 
> @JonathonMisiewicz Is this on the GauXC side? If so do you have a reproducer?

Reproducer included in the issue. https://github.com/wavefunction91/GauXC/issues/225

PR #3356, opened 2026-02-04
Dimension deprecation: int* --> Dimension

NathanGillispie:
## Description
Breaks up #3351

Part of a larger effort to deprecate those pesky Dimension operators. This one has a lot of changes, but they are straightforward. This is for the simplest case of replacing objects declared as type `int*` to `Dimension`, which avoids the cast-to-pointer operation.

## Dev notes & details
- Only involves changing `int *` types to `Dimension` to avoid deprecated cast-to-pointer operation.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/ccenergy/diagnostic.cc
+++ b/psi4/src/psi4/cc/ccenergy/diagnostic.cc
@@ -50,7 +50,7 @@ double CCEnergyWavefunction::diagnostic() {
     auto uoccpi = moinfo_.uoccpi;
     auto openpi = moinfo_.openpi;
 
-    int *occpi, *virtpi;
+    Dimension occpi, virtpi;
     int *occ_sym, *vir_sym;
     if (params_.ref != 2) {
         occpi = moinfo_.occpi;
--- a/psi4/src/psi4/cc/cceom/diagSS.cc
+++ b/psi4/src/psi4/cc/cceom/diagSS.cc
@@ -77,9 +77,9 @@ void diagSS(int C_irr) {
     int pf, cnt, irr_occ, irr_virt;
 
     nirreps = moinfo.nirreps;
-    auto openpi = moinfo.openpi;
-    auto occpi = moinfo.occpi;
-    auto virtpi = moinfo.virtpi;
+    const auto& openpi = moinfo.openpi;
+    const auto& occpi = moinfo.occpi;
+    const auto& virtpi = moinfo.virtpi;
 
     if (params.eom_ref == 2) { /* UHF */
         aoccpi = moinfo.aoccpi;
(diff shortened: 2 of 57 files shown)


JonathonMisiewicz (changes requested):
Please learn what `const auto&` does, and then apply it.

JonathonMisiewicz (approved):

JonathonMisiewicz (changes requested):
It looks like there are still some code failures. I strongly suspect these are cases where changing function signatures to require a `const Dimension&` is the way forward. Don't forget to run tests locally!

JonathonMisiewicz (approved):

loriab (approved):

maxscheurer (approved):

psi4/src/psi4/cc/ccenergy/diagnostic.cc
```diff
@@ -50,7 +50,7 @@ double CCEnergyWavefunction::diagnostic() {
     auto uoccpi = moinfo_.uoccpi;
     auto openpi = moinfo_.openpi;
 
-    int *occpi, *virtpi;
+    Dimension occpi, virtpi;
```
TiborGY, line 53: This now creates a local copy of moinfo_.occpi and moinfo_.virtpi instead of pointing to the original array.

JonathonMisiewicz: @NathanGillispie Please fix in another PR.

NathanGillispie: The issue here is the if statement immediately below this line. You can't declare it as Dimension& without initializing it.

JonathonMisiewicz: In this case, the fix is very easy. Move the definitions of `occpi` and `virtpi` into the if-clause where they're actually used, and get rid of the unused `occ_sym` and `vir_sym` variables. The `if (params_.ref != 2)` check can then be removed.

TiborGY: @NathanGillispie your PRs are merged, but I dont see that this has been addressed

psi4/src/psi4/cc/cceom/diagSS.cc
```diff
@@ -70,16 +70,16 @@ void diagSS(int C_irr) {
     int L, h, i, j, k, a, C_index, errcod, keep_going = 1, numCs, iter = 0;
     double norm, tval, *lambda, *lambda_old, zero = 0.0;
     double **G, *work, *evals_complex, **alpha, **evectors_left;
-    int nirreps, *openpi, *occpi, *virtpi, range;
-    int *aoccpi, *avirtpi, *boccpi, *bvirtpi;
+    int nirreps, range;
+    Dimension aoccpi, avirtpi, boccpi, bvirtpi;
```
TiborGY, line 74: @NathanGillispie same issue here as ccenergy/diagnostic.cc, there is now a local copy of these Dimension objects

PR #3358, opened 2026-02-05
Dimension deprecation: std::pair DPD constructor

NathanGillispie:
## Description
Breaks up #3351

Part of a larger effort to deprecate those pesky Dimension operators. Previously, DPD objects were often created by pushing back two `int *` pointers at a time. The first of these was often of type `Dimension` hence the cast-to-pointer warning. This PR creates a new DPD constructor (and `dpd_init` function) for type `std::vector<std::pair<Dimension, int *>>` and changes all relevant cast-to-pointer cases to use this constructor.

- API: new DPD constructor and `dpd_init` functions. 
- API: Deprecating old DPD constructor.

## Dev notes & details
- See Description

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/conf.py.in
+++ b/doc/sphinxman/source/conf.py.in
@@ -454,10 +454,10 @@ intersphinx_mapping = {'python': ('https://docs.python.org/3.12', None),
                        "numpy": ("https://numpy.org/doc/stable/", None),
                        'scipy': ('https://docs.scipy.org/doc/scipy/', None),
                        'matplotlib': ('https://matplotlib.org/stable/', None),
-                       "qcelemental": ("https://molssi.github.io/QCElemental/", None),
-                       "qcengine": ("https://molssi.github.io/QCEngine/", None),
+                       "qcelemental": ("https://molssi.github.io/QCElemental/dev/", None),
+                       "qcengine": ("https://molssi.github.io/QCEngine/dev/", None),
                        "qcfractal": ("https://molssi.github.io/QCFractal/", None),
-                       "qcmanybody": ("https://molssi.github.io/QCManyBody/", None),
+                       "qcmanybody": ("https://molssi.github.io/QCManyBody/dev/", None),
                        "optking": ("https://optking.readthedocs.io/en/latest", None),
                       }
 
--- a/doc/sphinxman/source/nitpick-exceptions
+++ b/doc/sphinxman/source/nitpick-exceptions
@@ -39,6 +39,7 @@ py:obj qcmanybody.computer.ManyBodyComputer.set_max_nbody
 py:obj qcmanybody.computer.ManyBodyComputer.set_return_total_data
 py:obj qcmanybody.computer.ManyBodyComputer.set_supersystem_ie_only
 py:class mbres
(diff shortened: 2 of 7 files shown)


psi4/src/psi4/cc/ccdensity/ccdensity.cc
```diff
@@ -162,27 +162,21 @@ PsiReturnType ccdensity(std::shared_ptr<ccenergy::CCEnergyWavefunction> ref_wfn,
     if (params.ref == 0 || params.ref == 1) { /** RHF or ROHF **/
         cachelist = cacheprep_rhf(params.cachelev, cachefiles);
 
-        std::vector<int *> spaces;
-        spaces.push_back(moinfo.occpi);
-        spaces.push_back(moinfo.occ_sym.data());
-        spaces.push_back(moinfo.virtpi);
-        spaces.push_back(moinfo.vir_sym.data());
+        std::vector<std::pair<Dimension, int *>> spaces;
+        spaces.push_back(std::make_pair(moinfo.occpi, moinfo.occ_sym.data()));
```
JonathonMisiewicz: The `push_back(std::make_pair(*, *))` can be replaced with `emplace_back(*, *)`.

NathanGillispie: Interesting I didn't know you could do that.

psi4/src/psi4/libdpd/dpd.h
```diff
@@ -330,6 +330,8 @@ class PSI_API DPD {
         dpd_file4_cache_entry *priority, int num_subspaces, std::vector<int *> &spaceArrays);
```
JonathonMisiewicz, line 331: The `std::vector<int *>` version of `DPD` and `dpd_init` should be deprecated.

psi4/src/psi4/libdpd/dpd.h
```diff
@@ -330,6 +330,8 @@ class PSI_API DPD {
...
     DPD(int dpd_num, int nirreps, long int memory, int cachetype, int *cachefiles, int **cachelist,
         dpd_file4_cache_entry *priority, int num_subspaces, std::vector<DPDMOSpace> &moSpaces);
+    DPD(int dpd_num, int nirreps, long int memory, int cachetype, int *cachefiles, int **cachelist,
+        dpd_file4_cache_entry *priority, int num_subspaces, std::vector<std::pair<Dimension, int *>> &spaceArrays);
```
JonathonMisiewicz: The variable `num_subspaces` is redundant and can be removed. `num_subspaces = spaceArrays.size()`.

JonathonMisiewicz (changes requested):

psi4/src/psi4/libdpd/dpd.h
NathanGillispie: Looks like it it can be removed immediately, do you think it's best to do so or use `PSI_DEPRECATED`?

JonathonMisiewicz: Please make the deprecation message more descriptive. `DPD construction via a vector of int* is being deprecated and will be removed as soon as 1.12.`

PR #3361, opened 2026-02-07
Dimension deprecation: reorder_qt

NathanGillispie:
## Description
Breaks up #3351

Part of a larger effort to deprecate those pesky Dimension operators. Changes function signatures of `reorder_qt` and `reorder_qt_uhf` from `int*` to `Dimension const&`.

- Minor API: `reorder_qt[_uhf]` function signatures changed.

## Dev notes & details
- See description.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/doc/sphinxman/source/nitpick-exceptions
+++ b/doc/sphinxman/source/nitpick-exceptions
@@ -39,6 +39,7 @@ py:obj qcmanybody.computer.ManyBodyComputer.set_max_nbody
 py:obj qcmanybody.computer.ManyBodyComputer.set_return_total_data
 py:obj qcmanybody.computer.ManyBodyComputer.set_supersystem_ie_only
 py:class mbres
+py:class qcmanybody.models.v1.manybody_input_pydv1.ManyBodyInput
 
 ## numpy inherited docstrings
 #py:obj dtype


psi4/src/psi4/libqt/reorder_qt.cc
```diff
@@ -66,17 +66,17 @@ namespace psi {
...
-PSI_API void reorder_qt(const int *docc_in, const int *socc_in, int *frozen_docc_in, int *frozen_uocc_in, int *order,
-                        int *orbs_per_irrep, int nirreps) {
+PSI_API void reorder_qt(Dimension const& docc_in, Dimension const& socc_in, Dimension const& frozen_docc_in,
+                        Dimension const& frozen_uocc_in, int *order, Dimension const& orbs_per_irrep) {
     int cnt = 0, irrep, point, tmpi;
     int *used, *offset;
     int *docc, *socc, *frozen_docc, *frozen_uocc;
     int *uocc;
 
+    int nirreps = docc_in.n();
```
JonathonMisiewicz: 1. It's best to use `auto` in a case like this.
2. At risk of paranoia, can we add a check that _all_ dimensions have the same number of irreps and throw an exception if this fails?

NathanGillispie: For 2. if I enforce all inputs to have the same number of irreps, a significant number (84/122) of the cc tests fail due to the check, both in the rhf and uhf version.

NathanGillispie: The only possibility is if there are more irreps in other Dimension objects that are never used. Otherwise you would have to get a segfault. Indices 0 to nirreps are accessed for all Dimension objects, yet number of irreps are not all equal.

JonathonMisiewicz: Can you do some investigation? Just use `Dimension::print` to print out all the dimensions for one of the failing test cases.

NathanGillispie: Got it, I was running the comparison incorrectly. Everything is right in the world again.

JonathonMisiewicz (approved):

PR #3365, opened 2026-02-11
Fix GauXC Path Printout

JonathonMisiewicz:
Let's make the GauXC printout a little more useful, shall we?

## Dev notes & details

- [x] GauXC path is now printed out properly when building with it

## Status
- [x] Ready for review
- [x] Ready for merge


JonathonMisiewicz:
@loriab requested this be merged into a larger PR.

PR #3320, opened 2025-09-22
WIP: Interface to OpenTrustRegion library

jonas-greiner:
## Description
This PR interfaces to the OpenTrustRegion library for second-order trust region orbital optimization for mean-field methods

## User API & Changelog headlines
- [x] Enables robust second-order orbital optimization for mean-field methods with the OpenTrustRegion library through the `otr_scf` keyword

## Dev notes & details
- [x] Adds interfaces in hf.cc, rhf.cc, rohf.cc, uhf.cc

## Questions
- [ ] The build system is likely not optimal, codedeps.yaml needs to be improved?

## Checklist
- [ ] Tests added for any new features

## Status
- [ ] Ready for review
- [ ] Ready for merge

PR #3369, opened 2026-02-28
CGHF added as an SCF reference

mnward10:
## Description
Complex generalized Hartree-Fock (CGHF) utilizes the new Einsums backend for complex orbitals, but also spin-blocks the Hamiltonian to facilitate symmetry breaking. This will also enable future relativistic methods as they are inherently complex.

- CGHF is added as an SCF reference

## Dev notes & details
- all initial guesses are available, but only core and SAD have been thoroughly tested.
- DIIS is available as the only option for convergence acceleration
- CGHF uses all of the functions inherited from the HF class to run the SCF
- cannot use the inherited variables (`Da_`, `Fa_`, etc) due to the inability of the SharedMatrix object to handle complex numbers, so Einsums BlockTensor objects are used for storing the data and performing the linear algebra routines
- preiterations is an additional Pybind meant to be called after the initial guess is formed. This is used to populate some spin-blocked matrices.
- code does not have any features that would break spin-symmetry. I'd like to get the "base" CGHF code reviewed before adding these, but they will come in upcoming PRs
- DIIS is in C++ compared to the other HF algorithms which implement it in Python.
- form_einsums_FDSmSDF_ is added as a Pybind to enable DIIS on the Python side
- J and K calculation currently handled with an explicit 4-index `for` loop, but testing is underway with Andy Jiang's new EinsumsDFJK object. This will completely replace the current for loop.
- C1 symmetry is only available due to the ao_eri mints object being used without transformation. This will also be deprecated with EinsumsDFJK. The remainder of the code is able to handle point group symmetry.
- Before Einsums 2.0, there is a bug with syev/hyev where heev and gesv are transposed incorrectly. Therefore, I had to transpose in a couple of instances. This will be deprecated once Einsums 2.0 is in Psi4
- This is currently standalone; there is no post-SCF calculations

## Checklist

## Status
- [X] Ready for review
- [X] Ready for merge


JonathonMisiewicz (changes requested):
Thank you for the PR! This is a great new feature for relativistic methods.

We want our code to be easily maintainable and straightforward for new developers, since we rely so heavily on graduate students, and understand that a lot of our new developers are new to programming. As such, I'm going to ask you to wait to merge this PR in as we make some recommendations to get this code up to our quality standards. I'm going to go slowly so as not to be overwhelming and encourage you to think of this as an opportunity to develop some more useful programming skills.

psi4/src/psi4/scfgrad/scf_grad.cc
```diff
@@ -1004,6 +1005,7 @@ SharedMatrix SCFDeriv::compute_hessian()
     if (options_.get_str("REFERENCE") == "RHF" || 
         options_.get_str("REFERENCE") == "RKS" || 
         options_.get_str("REFERENCE") == "UHF" ||
+        options_.get_str("REFERENCE") == "CGHF" ||
```
JonathonMisiewicz: Given what this if-clause does, does enabling it for CGHF make sense? Why or why not?

mnward10: Thanks for reviewing this and the encouragement! This is for a later feature, but I may be getting a bit too ahead of myself. I think I should have an exception here that the feature is not enabled yet. Does this seem right? This was a mistake on my part and doesn't make sense, as I think it would seg fault

JonathonMisiewicz: Yes, this is a later feature, and you're getting a bit ahead of yourself. The correct thing to do here is to raise the exception in the `else` block, and simply remove this line.

mnward10: Awesome, thank you. I removed that line and pushed.

psi4/src/psi4/scfgrad/scf_grad.cc
```diff
@@ -54,6 +54,7 @@
 #include "psi4/libscf_solver/hf.h"
 #include "psi4/libscf_solver/uhf.h"
 #include "psi4/libscf_solver/rhf.h"
+#include "psi4/libscf_solver/cghf.h"
```
JonathonMisiewicz: Alright, next question: what is the purpose of this include? I could ask the same thing for all your changes in `scfgrad`.

mnward10: Looking back, I had a working stability analysis in the source once upon a time, but it was in no condition to be submitted as a PR. It looks like I missed some deprecations/includes when I did my cleanup there. The new push should have removed all traces/hints of CGHF in scf_grad.

psi4/src/psi4/libscf_solver/cghf.cc
```diff
@@ -0,0 +1,632 @@
...
+ * Fills the density matrix with the precomputed SAD guess from sad.cc
+ * This seems to be more robust than SAP at the moment since SAP
+ */
+void CGHF::compute_SAD_guess(bool natorb) {
+    HF::compute_SAD_guess(natorb);
+
+    for (int h = 0; h < nirrep_; h++)
+        for (int p = 0; p < nsopi_[h]; p++)
+            for (int q = 0; q < nsopi_[h]; q++) {
+                D_->block(h)(p, q) = Da_->get(h, p, q);
```
JonathonMisiewicz, line 495: Is there really not a better way to do this? Setting element-by-element is slow compared to copying contiguous memory. I'm not familiar with einsums, but I'd expect there would be _something_ better for this.

For comparison, Psi4 Matrices have block gets and sets.

mnward10: Alright. After some trial and error with @cgbriggs99 for a little while, I believe the new push should address this. There is a way to do this with Einsums Tensors, which I took the liberty of doing in my `form_D` routine, but with double `SharedMatrix` to std::complex<double> `Tensor`, it's a bit harder. This would be a smooth copy with a block get/set like mentioned, but Einsums cannot copy due to the mismatch in types, at least without some black magic/hacky stuff.

NathanGillispie: I want to add, since I just saw the proposed changes, einsums has efficient array copying with `operator=` but the types from HF (`Da_`, `Fa_`, etc.) make that hard. One optimization may be to just parallelize the outer loop. I think the best option is to simply revert [4864d23](https://github.com/psi4/psi4/pull/3369/commits/4864d231b7f31ce63a8dccda5cb3384dc1ed95b4).

NathanGillispie: Also, my spin-orbit code fails after  [4864d23](https://github.com/psi4/psi4/pull/3369/commits/4864d231b7f31ce63a8dccda5cb3384dc1ed95b4) so I think it would be best if we had a way to test a complex Fock matrix during this process. Maybe some sort of complex constant perturbation in the AB, BA blocks when a certain option is given?

JonathonMisiewicz: I can't tell what you're trying to communicate. I think what you're trying to say is that, "The new einsums feature used by [commit] is broken. To get working code, I'm going to revert the older solution but with parallelization for speed's sake. Since the break was detected not by this PR but by a correlated method on top of this reference, the tests aren't good enough. Perhaps we can add a test where the orbitals are not pure spinorbitals?"

_Assuming that's what you mean_...
1. Reverting that is fine for now, but I want the new einsums feature working and used before this code is merged.
2. If the CGHF tests only converge to complex UHF solutions, then indeed, the tests aren't enough. Please add a test case where the solution requires breaking Sz symmetry.

NathanGillispie: Yes, the new einsums feature is what was causing a segfault in #3371 in the M-intel runner, which for some reason is not running in this PR. Connor is working right now on getting that sorted.

Originally, we had planned on this PR being as small as possible for reviewers' sake, and we were testing complex orbitals with #3371. But it's becoming clear to me now that that isn't sustainable and we need some way of adding and testing complex orbitals in this PR. I'm still not entirely sure how to do that in a simple way.

To address 2. We often do get lower energies than UHF when running open-shell molecules, which seems impossible because how does the symmetry breaking occur? I think the answer comes down to a lack of numerical precision in the eigenvalue solver. The full explanation is rather long-form and I would be glad to talk about it, but perhaps elsewhere (I have an [issue](https://github.com/pyscf/pyscf/issues/3146) on pyscf about this if you want to read more). This lack of precision causes errors that compound over the course of the SCF procedure, often giving different solutions for different computers, especially on open-shell molecules, which is a huge problem for tests that are supposed to yield the same energy on every computer.

NathanGillispie: Continuing the reliable tests idea: stability analysis resolves the issue of converging to different solutions, but it isn't implemented. Point group symmetry also helps, but there are cases where you can't obtain the lowest energy solution with it. Perhaps there is a better way to test that the SCF procedure "works" without comparing the final converged energy.

cgbriggs99: Could this also be a case of UHF wavefunctions not being true eigenfunctions of the spin operator? The spin-contamination metric might shed some light on this.

NathanGillispie: Perhaps. Theoretically, non-zero AB, BA blocks can't be introduced into the density matrix if you start with a density matrix and core hamiltonian that is zero in those blocks. There's just no way to spontaneously break the Sz symmetry in that case without some perturbation of the off-diagonal blocks. The reason I think this happens in reality is numerical noise.

mnward10: Spin contamination is pretty much the case, yeah. I agree that a metric would be useful here. I'll work today on adding a routine for evaluating the S^2 expectation value. To introduce spin-broken solutions, there are a wide variety of approaches. The simplest and easiest is HOMO/LUMO mixing which I have ready to go, and there is already a `GUESS_MIX` keyword. We can also perform orbital rotations as highlighted in this paper: https://pubs.acs.org/doi/abs/10.1021/acs.jctc.7b01016. The problem with the latter approach is that it's pretty arbitrary. I have, however, been investigating non-collinear initial guesses in my research.

The introduction of complex orbitals is a different story. In a non-relativistic calculation, the only way to do this (from what I've seen) is either through a random perturbation -- which I disagree with since it can be non-deterministic -- or by constructing a complex perturbation based off the core Hamiltonian. We can also obtain complex orbitals via orbital rotations, but like mentioned it is kind of arbitrary and not consistent.

Both HOMO/LUMO mixing and a complex perturbation are ready to be added. I was waiting on these at to not overload the reviewers, but if you feel like they are needed here, then I will be happy to add these as well. I am especially delighted to add these because they capture the true spirit of CGHF.

mnward10: I think another useful feature to add at this stage is to read in an initial guess density, either from NumPy or by reading in from a file. This seems to be the most robust way to test if complex orbitals work. I understand the dilemma of using the ZORA code as a test here, because ZORA relies on CGHF being correct

loriab: could a format like https://github.com/psi4/psi4/blob/master/psi4/driver/p4util/python_helpers.py#L355 from a "normal" calc be hacked into to provide a complex guess? Psi4 either does or could be persuaded to guess from that structure.

NathanGillispie: I like PySCF's solution, they essentially take the restricted guess diagonal and add `0.05*` that to the off-diagonal blocks. We could equivalently do `+-0.05i*` **that** to make it complex. Every GHF guess in PySCF does that except core. So it may not be necessary to read complex orbitals from a file or previous guess.

mnward10: I think it would be fruitful to use the recommendation of @loriab to test the validity of my algorithm for complex orbitals, since we will know the final energy beforehand. Once that's valid, the PySCF complex guess may be the route to go.

mnward10: Alright. I haven't gotten around to @loriab 's recommendation yet, but I have added HOMO/LUMO mixing to break the Sz spin symmetry. I have found lower energies than UHF in Li3 and odd-numbered hydrogen rings. I have added the CGHF_MIXING_DEGREE parameter (default 20 degrees) to control the amount of rotation here when doing the mixing. Here is the data. All of these are done with the cc-pVDZ basis set with SAD as the initial guess. 

The Li3 Triangle (2.7 Angstrom bond length)
| ghf_energy   | uhf_energy   | degrees |
|--------------|-------------|---------|
| -22.31332    | -22.31332   | 10      |
| -22.314882   | -           | 20      |
| -22.314882   | -           | 30      |
| -22.314882   | -           | 40      |
| -22.314882   | -           | 50      |
| -22.314882   | -           | 60      |
| -22.314882   | -           | 70      |

And then the hydrogen rings separated with equal bond lengths of 1.0 Angstroms. This is why I have the CGHF_MIXING_DEGREE parameter, since the default will not always work.

| Molecule | ghf_energy | uhf_energy | degree |
| -------- | ---------- | ---------- | ------ |
| H3       | \-1.50627  | \-1.50627  | 10     |
|          | NaN        |            | 20     |
|          | \-1.50773  |            | 30     |
| H7       | NaN        | \-3.6844   | 10     |
|          | \-3.68991  |            | 20     |
|          | \-3.68991  |            | 30     |

I have the max iterations set to 500, so the two cases that didn't converge truly failed. These are difficult systems that SAD by itself may or may not struggle with.

I'll finish by saying that, in my own research, I find that electrons need to be canted more than 10 degrees off their quantization axis, otherwise they'll kind of snap back to being collinear.

mnward10: Oh -- currently the s2_expectation_value routine is still bugged. I am working on that. It gives dubious results, so I am likely doing the math wrong somewhere.

JonathonMisiewicz: Thanks. Is there any way we can compare these GHF results to results from some other package, or some computation in the literature? What I'm fundamentally looking for is an independent correctness check.

mnward10: I think forte2 does it this way, so I'll try to compare with those results.

It is hard to compare our data with the literature since articles always use stability analysis. To my knowledge, PySCF doesn't do HOMO/LUMO mixing, instead opting to break symmetry with what was mentioned above.

I'll add the PySCF broken symmetry guess next for a side-by-side comparison

mnward10: One piece of evidence is, for the H3 case, the difference in GHF vs UHF energy exactly matches the energy from Table 1 in this article,

https://pubs.aip.org/aip/jcp/article/142/15/154109/211957/Stability-of-the-complex-generalized-Hartree-Fock

But misses the mark with H7. As expected, we would need stability analysis for that.

mnward10: I have replicated PySCF's off-diagonal perturbation and the results exactly.

This is for the O2 molecule being dissociated with a SAD guess and cc-pVDZ basis set. I do not use the density matrix from the last bond length; it is a brand new calculation at each bond length. We tend to have strange results for the `no_break_symm` case, but it has been previously established that numerical precision is an issue at non-equilibrium bond lengths. I'm not sure why this goes away with the `break_symm` guess. 

I have included a UHF triplet dissociation (with UHF stability analysis) for reference. This was taken from Psi4. O2 is a really difficult case for dissociation curves, so it's fully expected that GHF won't get a lower energy there until stability analysis is implemented. 

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/matth/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/matth/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#467886" vlink="#96607D">


  | Bond length | Psi4     break_symm | PySCF     break symm | PySCF     no break_symm | Psi4     no break symm | UHF (triplet)
-- | -- | -- | -- | -- | -- | --
0 | 0.6 | -146.1980575 | -146.1980574 | -146.134285 | -146.134285 | -146.198057
1 | 0.7 | -147.9907597 | -147.9907597 | -147.92141 | -147.92141 | -147.99076
2 | 0.8 | -148.9012481 | -148.9012481 | -148.82654 | -148.82654 | -148.901248
3 | 0.9 | -149.3489809 | -149.3489809 | -149.323265 | -149.323265 | -149.348981
4 | 1 | -149.5510414 | -149.5510414 | -149.524033 | -149.551041 | -149.551041
5 | 1.1 | -149.6235644 | -149.6235644 | -149.595748 | -149.539815 | -149.623564
6 | 1.2 | -149.6289923 | -149.6289923 | -149.600768 | -149.600768 | -149.628992
7 | 1.3 | -149.6010159 | -149.6010159 | -149.515833 | -149.515833 | -149.605212
8 | 1.4 | -149.5576521 | -149.5576521 | -149.529425 | -149.529425 | -149.579367
9 | 1.5 | -149.5082982 | -149.5082982 | -149.423492 | -149.423492 | -149.553471
10 | 1.6 | -149.4578128 | -149.4578128 | -149.373552 | -149.430136 | -149.540886
11 | 1.7 | -149.4087301 | -149.4087301 | -149.381414 | -149.40873 | -149.502413
12 | 1.8 | -149.362333 | -149.362333 | -149.27953 | -149.3354 | -149.539405
13 | 1.9 | -149.3192008 | -149.3192008 | -149.237218 | -149.237218 | -149.435536
14 | 2 | -149.2795244 | -149.2795244 | -149.253368 | -149.198376 | -149.543384
15 | 2.1 | -149.2432836 | -149.2432836 | -149.162954 | -149.162954 | -149.544844
16 | 2.2 | -149.2103418 | -149.2103418 | -149.130795 | -149.184906 | -149.545843
17 | 2.3 | -149.1805004 | -149.1805004 | -149.155389 | -149.101687 | -149.54648
18 | 2.4 | -149.1535325 | -149.1535325 | -149.075394 | -149.128718 | -149.360244
19 | 2.5 | -149.1292042 | -149.1292042 | -149.104659 | -149.051679 | -149.547097



</body>

</html>
 

Interestingly, this doesn't give a lower energy solution for the Li3 in either PySCF or Psi4. This time, it was done with a 3-21G basis set, but the same thing happens with cc-pVDZ.


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/matth/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/matth/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#467886" vlink="#96607D">


Psi4     (break_symm) | Psi4     (no_break_symm) | PySCF     (break_symm) | PySCF     (no_break_symm)
-- | -- | -- | --
-22.16023688 | -22.16023688 | -22.16023688 | -22.16023688



</body>

</html>


Nor does it give a lower energy than any of my HOMO/LUMO calculations. I'll omit the table there for redundancy reasons. I'm suspecting this is because the perturbation is too tiny, and the electrons are snapping back to being collinear.

NathanGillispie:
@JonathonMisiewicz You requested that we test for true GHF solutions, not just complex UHF solutions. Converging to these solutions often requires a combination of stability analysis and finagling of the guess. Suppose we prepared stable GHF MOs externally, ran SCF on those, and showed the energy/density matrix did not change. Would that be sufficient to show CGHF works without implementing stability analysis? If not, a few priorities might have to shift. Stability is already on the agenda but I have been under the assumption that including it in this PR would be too much to review.

susilehtola:
I would heavily suggest using the OpenTrustRegion interface (#3320 ) to obtain robust second-order convergence in the implementation. GHF is infamous for local minima and poor SCF convergence; this is where trust region methods shine. I have the intuition that OTR will converge faster with respect to wall time than first-order methods just because the optimization landscape is quite challenging.

The implementation would not be difficult; it just needs an implementation of the `hess_on_x` function that calculates the action of the orbital Hessian on a vector of orbital rotation parameters; the same function that is needed for stability analysis.

Please refer to our recent paper for further details: 
> J. Greiner, I.-M. Høyvik, S. Lehtola, and J. J. Eriksen, [A reusable library for second-order orbital optimization using the trust region method](https://doi.org/10.1021/acs.jctc.5c01576), J. Chem. Theory Comput. 22, 881 (2026). doi:10.1021/acs.jctc.5c01576 [arXiv:2509.13931](https://arxiv.org/abs/2509.13931)

JonathonMisiewicz:
> @JonathonMisiewicz You requested that we test for true GHF solutions, not just complex UHF solutions. Converging to these solutions often requires a combination of stability analysis and finagling of the guess. Suppose we prepared stable GHF MOs externally, ran SCF on those, and showed the energy/density matrix did not change. Would that be sufficient to show CGHF works without implementing stability analysis? If not, a few priorities might have to shift. Stability is already on the agenda but I have been under the assumption that including it in this PR would be too much to review.

Yes, that'd be enough for now. I agree with you that waiting for stability analysis is probably for the best.

mnward10:
> 
> Please refer to our recent paper for further details:
> 
> > J. Greiner, I.-M. Høyvik, S. Lehtola, and J. J. Eriksen, [A reusable library for second-order orbital optimization using the trust region method](https://doi.org/10.1021/acs.jctc.5c01576), J. Chem. Theory Comput. 22, 881 (2026). doi:10.1021/acs.jctc.5c01576 [arXiv:2509.13931](https://arxiv.org/abs/2509.13931)

Susi, this is a great paper and I believe I have the capacity to implement this. I'll begin prototyping in Python. Thanks!

> > @JonathonMisiewicz You requested that we test for true GHF solutions, not just complex UHF solutions. Converging to these solutions often requires a combination of stability analysis and finagling of the guess. Suppose we prepared stable GHF MOs externally, ran SCF on those, and showed the energy/density matrix did not change. Would that be sufficient to show CGHF works without implementing stability analysis? If not, a few priorities might have to shift. Stability is already on the agenda but I have been under the assumption that including it in this PR would be too much to review.
> 
> Yes, that'd be enough for now. I agree with you that waiting for stability analysis is probably for the best.

With this and Susi's comment, are we saying that I should have stability analysis implemented in the first release, or should this come at a later time?

JonathonMisiewicz:
_Under the circumstances_, I think it's better for stability analysis to come later. The two relevant circumstances are:

1. It's relatively straightforward to come up with an alternative way to verify correctness.
2. You're relatively new to coding, so simpler will probably be easier on us both. After all, I haven't even gotten to what I think is the biggest issue with this PR yet. 🙂 

If it'd be less trouble for you to write stability analysis, then go right ahead. But I'm not insisting on it or even pushing for it.

mnward10:
Alright, I have a progress update. It is actually fairly difficult to get a converged solution from PySCF inserted as a guess for my CGHF code. Previously, I thought the main limitation was complex matrices, but I also learned while doing this that the size of the matrix is important too. The GHF matrices will be 2x the size of UHF. That is, if Ca_ and Cb_ are (15, 15), then in GHF they are (30, 30) for the real and imaginary components, respectively. This caused seg faults down the line for reasons that are still pretty unclear to me. I tried to fix this by adding an additional conditional in HF::guess, but to no avail. I am leaving that in the current commit as I work on it, since I think the option of adding a custom guess is important here.

In good news, I was able to get proof that the code works in another way. I added an additional Pybind, `CGHF::read_from_wfn` which I execute after running a calculation. This is implemented in my Python code as (ugly as it may be):

```
import psi4
import numpy as np

#H3
mol = psi4.geometry("""
0 2
H 0.57735027 0.0 0.0
H -0.28867513 0.5 0.0
H -0.28867513 -0.5 0.0
symmetry c1
""")

psi4.core.set_output_file("output.dat", False)

# Need to use a CORE guess on hydrogen rings since SAD will converge to a quartet
psi4.set_options({'basis': 'sto-3g', "reference": "cghf", "scf_type": "pk", "guess": "core"})
E_scf, wfn = psi4.energy('scf', molecule=mol, return_wfn=True)

guess_F = np.load("pyscf_data/H3_fock.npy")

Ca = wfn.Ca()

wfn.guess_Ca(Ca)
wfn.guess_Cb(wfn.Cb())

wfn.Fa().copy(psi4.core.Matrix.from_array(guess_F.real))
wfn.Fb().copy(psi4.core.Matrix.from_array(guess_F.imag))

wfn.reset_diis_subspace()
wfn.read_from_wfn()
wfn.form_C()
wfn.form_D()
wfn.form_G()
wfn.compute_energy()
```

This was ran on, again, the hydrogen rings with the STO-3G basis set. I ran into a bunch of difficulty when trying to do anything bigger than this since the conversion from PySCF to Psi4 AO orderings isn't straightforward. I followed this issue: https://github.com/pyscf/pyscf/issues/2123. Thus, hydrogen rings with a STO-3G basis set should be a direct translation. Here is the data. These solutions are both complex and break Sz symmetry. The UHF calculations were ran on Psi4 with `stability_analysis: "FOLLOW"`. 


<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/matth/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/matth/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#467886" vlink="#96607D">


  | UHF | Psi4 CGHF | PySCF CGHF
-- | -- | -- | --
H3 | -1.335980055 | -1.340440343 | -1.340440343
H5 | -2.461887557 | -2.47103509 | -2.47103509
H7 | -3.52044518 | -3.532521235 | -3.532521235
H9 | -4.619586139 | -4.632769513 | -4.632769513



</body>

</html>

When the guess is supplied in this manner, the SCF converges immediately. e.g.,
```
   @CGHF iter   0:    -1.34044034292794   -1.34044e+00   1.61119e-08
   @CGHF iter   1:    -1.34044034292799   -4.37428e-14   1.26910e-08
```

Finally, I should mention that I tested a wide variety of molecules that have CGHF solutions with the PySCF guess alone, and there is never a case where it provides a lower energy solution than UHF. This is ubiquitous to both the PySCF GHF code and my code. The only time a lower energy solution is given is when stability analysis is ran, or when the HOMO/LUMO mixing routine I implemented is used. This is why I had to supply a guess from PySCF.

I'm currently working on cleaning up some changes that I made while doing this since it appears to have broken one of the test cases, but see the current commit for how I did this.

JonathonMisiewicz:
Great, make sure that test is added so we can have a correctness check, and let me know when that's done.

mnward10:
Deprecated a bunch of the stuff, so now the guess density matrix is only offered in test cases (cghf4, cghf5, cghf6). It ended up being far too difficult to get this "generalized" to any guess density matrix. All tests pass.

JonathonMisiewicz:
> All tests pass.

Clearly, they don't.

mnward10:
> > All tests pass.
> 
> Clearly, they don't.

I, like a fool, forgot to add the `density.dat` files to the commit

JonathonMisiewicz:
> > > All tests pass.
> > 
> > 
> > Clearly, they don't.
> 
> I, like a fool, forgot to add the `density.dat` files to the commit

1. I don't see any `density.dat` files in your latest commit.
2. Even that won't be enough for tests to pass. Click on the red X associated with the last commit and do some exploring. You should find some compiler error messages.

mnward10:
> > > > All tests pass.
> > > 
> > > 
> > > Clearly, they don't.
> > 
> > 
> > I, like a fool, forgot to add the `density.dat` files to the commit
> 
>     1. I don't see any `density.dat` files in your latest commit.
> 
>     2. Even that won't be enough for tests to pass. Click on the red X associated with the last commit and do some exploring. You should find some compiler error messages.

That's confusing. I'm not sure why the simple Pybind isn't being recognized only in the clang build. Let me investigate. I didn't see these because everything compiles and all tests pass on my machine.

mnward10:
Ah, I have Einsums installed on my machine, and the Pybinds require this. Building without Einsums on my machine caused the same error. The new commit adds dummy routines (exactly like the other Pybinds) in the header file to fix this, and it builds on my machine when I toggle Einsums off now.

PR #3371, opened 2026-03-03
Adding ZORA Spin-orbit Coupling 

NathanGillispie:
## Description
Adding spin-orbit coupling to Psi4. Given the atomic orbital $\chi$ and effective potential $v_\text{eff}(\mathbf{r})$ (see #3292 for explanation), the spin-orbit coupling Hamiltonian is computed as

$$H_{\mu\nu}^\text{SO}=\sigma\cdot\int d\mathbf{r}^3\frac{v_\text{eff}(\mathbf{r})}{4c^2-2v_\text{eff}(\mathbf{r})}\nabla\chi_\mu^\dagger(\mathbf{r})\times\nabla\chi_\nu(\mathbf{r})$$

This corresponds to the third term on the RHS of the [ZORA Kohn-Sham](https://psicode.org/psi4manual/master/relativistic.html#equation-zks) Hamiltonian:

$$\hat h^\text{ZKS}=\frac{\mathbf{p}^2}{2}+\mathbf{p}\left(\frac{\kappa -1}{2}\right)\mathbf{p}+\frac{\kappa^2}{4c^2}\mathbf{\sigma}\cdot\left(\nabla v^\text{KS}\times \mathbf{p}\right)+v^\text{KS}$$

> [!IMPORTANT]
> This is a branch of [complex_wfn](https://github.com/NathanGillispie/psi4/tree/complex_wfn) which is the megafeature branch for complex wavefunctions. `complex_wfn` is still the reference for the new stack starting with #3487, so don't expect any major changes. After the CGHF PR is open on that stack, I'll rebase this.

Pytests have been added that demonstrate $2p$ and $3d$ orbital splitting.

- [x] Added boolean option `SPIN_ORBIT_COUPLING` for CGHF reference.
- [x] Added pybind `ao_zora_spin_orbit` to get ZORA spin-orbit coupling Hamiltonian directly from libmints.

## Dev notes & details
- [x] Computes spin-orbit coupling Hamiltonian within zora.cc.
- [x] Caches $H^\text{SO}$ in vector form with three real matrices to be recombined inside of `cghf.cc`.

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


NathanGillispie:
This, along with #3379 should probably be moved to the v1.13 milestone. I'll rebase this onto the new CGHF stack when that is opened.

doc/sphinxman/source/scf.rst
```diff
@@ -341,6 +383,26 @@ keyword to true::
...
+In other cases, such as spin-frustrated systems or bond dissociation,
+constraining electrons to be purely spin-up or spin-down is insufficient.
+A common cause for this is near-degeneracies in which multiple states close
+in energy (but not perfectly identical) are in competition with one another,
+yet only a single solution can be obtained. Removing the constraint that
+electrons must be collinear is often enough to break these near-degeneracies
+yielding a lower energy, spin-broken solution.
+
+In even rarer cases, lower energy solutions are possible when complex 
+conjugation symmetry (:math:`\hat K`) is no longer enforced. While real orbitals
```
EricaCMitchell, line 394: For (:math:`\hat K`), can we try to match either Dyall & Faegri (eq 6.126) or Reiher & Wolf (eq 10.24) notation-wise for the complex conjugation operator?

psi4/src/psi4/libmints/zora.cc
```diff
@@ -278,6 +267,86 @@ void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, Sh
     }
 
     T_SR->copy_upper_to_lower();
+    timer_off("ZORA: Scalar Relativistic Kinetic");
+}
+
+// Compute the spin-orbit coupling Hamiltonian in the AO basis.
+void ZORA::compute_HSO(SharedMatrix& Hx, SharedMatrix& Hy, SharedMatrix& Hz) {
+    outfile->Printf("\n\n    ZORA: You called compute_HSO. Rejoice!\n");
```
EricaCMitchell: This line prints on every spin-orbit calculation. Do we need this? Seems like debug printing.

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -788,6 +788,25 @@ SharedMatrix MintsHelper::so_dkh(int dkh_order) {
...
 
+std::vector<SharedMatrix> MintsHelper::ao_zora_spin_orbit() {
+    bool include_perturbations = true;
+
+    std::string socx(PSIF_AO_SOCX);
+    std::string socy(PSIF_AO_SOCY);
+    std::string socz(PSIF_AO_SOCZ);
+
+    if (!are_ints_cached(socx, include_perturbations)) {
+        compute_so_zora_ints(include_perturbations, true);
```
EricaCMitchell, line 829: This call always writes the ZORA kinetic integrals into `cached_oe_ints_` under the key `PSIF_SO_T`. See line 1513. The function does not read the `RELATIVISTIC` option first.

As a result, a Python user who calls `mints.ao_zora_spin_orbit()` replaces the kinetic integrals of that `MintsHelper` object. A later non-relativistic energy from the same object is then wrong, and the program gives no message.

```python
mints = psi4.core.MintsHelper(basis)   # RELATIVISTIC is not set
T_nr = mints.so_kinetic()              # correct non-relativistic T, cached
soc  = mints.ao_zora_spin_orbit()      # overwrites the cached T with the ZORA T
T_2  = mints.so_kinetic()              # returns the ZORA T, not T_nr
```

Two corrections are possible:

- Give `compute_so_zora_ints` a flag that stops the write to `PSIF_SO_T`.
- Save the previous `PSIF_SO_T` entry in `ao_zora_spin_orbit()`, then put it back.

The first correction is cleaner.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
```diff
@@ -278,6 +267,86 @@ void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, Sh
...
+                    int nu = bf_map[l_nu];
+                    for (int p = 0; p < npoints; p++) {
+                        auto x = kernel[p] * (phi_y->get(p, l_mu) * phi_z->get(p, l_nu) -
+                                              phi_z->get(p, l_mu) * phi_y->get(p, l_nu));
+                        auto y = kernel[p] * (phi_z->get(p, l_mu) * phi_x->get(p, l_nu) -
+                                              phi_x->get(p, l_mu) * phi_z->get(p, l_nu));
+                        auto z = kernel[p] * (phi_x->get(p, l_mu) * phi_y->get(p, l_nu) -
+                                              phi_y->get(p, l_mu) * phi_x->get(p, l_nu));
+
+                        tmp_x->add(mu, nu, x);
```
EricaCMitchell, line 333: Each `add` call is a read-modify-write at a random position in an `nbf x nbf` matrix. The six targets do not depend on `p`, so the compiler cannot vectorize the loop.

Accumulate three local scalars over the point loop. Then do the six writes one time for each basis-function pair:

```cpp
double sx = 0.0, sy = 0.0, sz = 0.0;
for (int p = 0; p < npoints; p++) {
    sx += kernel[p] * (phi_y->get(p, l_mu) * phi_z->get(p, l_nu) - ...);
    sy += ...;
    sz += ...;
}
tmp_x->add(mu, nu, sx);  tmp_x->add(nu, mu, -sx);
tmp_y->add(mu, nu, sy);  tmp_y->add(nu, mu, -sy);
tmp_z->add(mu, nu, sz);  tmp_z->add(nu, mu, -sz);
```

This change removes `6 * npoints` scattered writes for each pair. Six writes stay. `compute_TSR` has the same pattern at line 255, but with one write for each point.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1476,6 +1511,19 @@ void MintsHelper::compute_so_zora_ints(bool include_perturbations) {
...
     // Overwrite cached integrals
     cached_oe_ints_[std::make_pair(PSIF_SO_T, include_perturbations)] = so_kinetic_zora;
+
+    if (spin_orbit) {
+        // Don't consider SO integrals until CGHF supports point-group symmetry
+        auto ao_Hx = std::make_shared<Matrix>("Spin-orbit Hx", basisset_->nbf(), basisset_->nbf());
+        auto ao_Hy = std::make_shared<Matrix>("Spin-orbit Hy", basisset_->nbf(), basisset_->nbf());
+        auto ao_Hz = std::make_shared<Matrix>("Spin-orbit Hz", basisset_->nbf(), basisset_->nbf());
+
+        zoraint.compute_HSO(ao_Hx, ao_Hy, ao_Hz);
```
EricaCMitchell, line 1551: `compute_TSR` at line 1500 and `compute_HSO` at this line each call `pworkers_[thread]->compute_functions(block)` for every block of the grid. When spin-orbit coupling is on, the program evaluates the first derivatives of all basis functions over the full grid two times.

Both kernels use the same three arrays `PHI_X`, `PHI_Y`, and `PHI_Z`. Add a combined function that fills `T_SR`, `Hx`, `Hy`, and `Hz` in one pass over the grid. Then `compute_TSR` and `compute_HSO` can stay as thin wrappers.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
```diff
@@ -278,6 +267,86 @@ void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, Sh
...
+#ifdef _OPENMP
+    nthreads = Process::environment.get_n_threads();
+#endif
+
+    int max_points = grid_->max_points();  // Set in grid_int_options
+#pragma omp parallel num_threads(nthreads)
+    {
+        // Give each thread their own scratch space
+        std::vector<double> kernel(max_points);
+        auto tmp_x = std::make_shared<Matrix>(Hx->nrow(), Hx->ncol());
```
EricaCMitchell, line 289: Each thread allocates three full `nbf x nbf` matrices. With 32 threads and 2000 basis functions, this is approximately 3 GB. `compute_TSR` allocates one such matrix for each thread.

Two alternatives are possible. Give each thread one matrix and use three OpenMP passes. Or reduce into a shared buffer with an atomic operation. If you keep the current code, state the memory cost in the header comment.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1453,16 +1472,32 @@ void MintsHelper::add_dipole_perturbation(SharedMatrix potential_mat) {
...
-void MintsHelper::compute_so_zora_ints(bool include_perturbations) {
-    outfile->Printf(" OEINTS: Using relativistic (ZORA) kinetic integrals.\n");
+void MintsHelper::compute_so_zora_ints(bool include_perturbations, bool force_spin_orbit) {
+    bool spin_orbit = options_.get_bool("SPIN_ORBIT_COUPLING");
+
+    // force_spin_orbit is used when requesting SOC integrals from python.
+    if (force_spin_orbit) {
+        spin_orbit = true;
+    } else {
+        if (spin_orbit && (options_.get_str("REFERENCE") != "CGHF"))
```
EricaCMitchell, line 1512: This guard is inside `compute_so_zora_ints`. The program gets to this function only when `RELATIVISTIC` is `ZORA`. The second guard, at `cghf.cc:214`, is inside CGHF.

As a result, this input runs to completion and ignores the option without a message:

```
set reference rhf
set spin_orbit_coupling true
```

The same is true for `reference rhf` with `relativistic x2c`. Add a check in `psi4/driver/procrouting/proc.py`, or in `HF::HF`. The check must reject `SPIN_ORBIT_COUPLING true` for every reference that is not CGHF.

Generated by: Claude Opus 5

psi4/include/psi4/psifiles.h
```diff
@@ -321,6 +321,9 @@
 #define PSIF_AO_TYY              "AO-basis T-YY Ints"        /* Traceless electric quadrupole moment integrals */
 #define PSIF_AO_TYZ              "AO-basis T-YZ Ints"        /* Traceless electric quadrupole moment integrals */
 #define PSIF_AO_TZZ              "AO-basis T-ZZ Ints"        /* Traceless electric quadrupole moment integrals */
+#define PSIF_AO_SOCX             "AO-basis spin-orbit X"
```
EricaCMitchell: These three labels are above the divider comment at line 328. The matrices behind them are anti-symmetric. `zora.cc` writes `+x` at `(mu, nu)` and `-x` at `(nu, mu)`. The labels belong under the anti-Hermitian comment.

Move the three lines below that comment. Add a trailing comment to each line, as the neighbor lines have.

Generated by: Claude Opus 5

tests/zora-soc/output.ref
```diff
@@ -0,0 +1,551 @@
...
+  Minimum eigenvalue in the overlap matrix is 3.8336369020E-03.
+  Reciprocal condition number of the overlap matrix is 1.1652196509E-03.
+    Using symmetric orthogonalization.
+
+  ==> Pre-Iterations <==
+
+  SCF Guess: Core (One-Electron) Hamiltonian.
+
+
+  CGHF: ZORA Spin-orbit coupling selected.
```
EricaCMitchell: This line has the old text `CGHF: ZORA Spin-orbit coupling selected.`. The source at `cghf.cc:209` now prints `CGHF: Spin-orbit coupling selected.`.

Commit 98d963a5b renamed the option, but the reference output kept the old text. Generate the file again after you delete the debug print at `zora.cc:275`.

Generated by: Claude Opus 5

tests/zora-soc/input.dat
```diff
@@ -0,0 +1,29 @@
...
+    Kr 0 0 0
+symmetry C1
+}
+
+set {
+    reference cghf
+    scf_type pk
+    e_convergence 1e-9
+    basis 3-21g
+    relativistic zora
```
EricaCMitchell, line 17: The test compares the energy against `-2787.5680652543642` to 6 decimal places. The test does not set `zora_radial_points` or `zora_spherical_points`. The value thus depends on the default grid.

`tests/zora/input.dat` sets both options for this reason. Set the same two options here.

Generated by: Claude Opus 5

psi4/src/read_options.cc
```diff
@@ -261,6 +261,10 @@ int read_options(const std::string &name, Options &options, bool suppress_printi
     Useful when comparing analytic and grid-based methods. !expert -*/
     options.add_bool("ZORA_NR_DEBUG", false);
 
+    /*- Add the spin-orbit coupling Hamiltonian on top of (scalar-relativistic) Hamiltonians.
+    Only available with `REFERENCE CGHF`. Currently, only `RELATIVISTIC ZORA` is supported. -*/
```
EricaCMitchell: Single backticks give an italic title reference in RST. The psi4 convention is `|globals__reference|` for an option name and double backticks for a value.

```
/*- Add the spin-orbit coupling Hamiltonian on top of the scalar-relativistic Hamiltonian.
Only available with |globals__reference| set to ``CGHF``. At this time, only
|globals__relativistic| set to ``ZORA`` is supported. -*/
```

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
```diff
@@ -278,6 +267,86 @@ void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, Sh
     }
 
     T_SR->copy_upper_to_lower();
+    timer_off("ZORA: Scalar Relativistic Kinetic");
+}
+
+// Compute the spin-orbit coupling Hamiltonian in the AO basis.
+void ZORA::compute_HSO(SharedMatrix& Hx, SharedMatrix& Hy, SharedMatrix& Hz) {
```
EricaCMitchell, line 274: The old `ZORA::compute` had an outer `timer_on("ZORA")` pair. The refactor removed it. `compute_TSR` keeps its own timer. This function has none.

Add `timer_on("ZORA: Spin-Orbit Coupling")` and the matching `timer_off`.

Generated by: Claude Opus 5

psi4/src/psi4/libscf_solver/cghf.cc
```diff
@@ -0,0 +1,804 @@
...
+        outfile->Printf("\n  CGHF: Spin-orbit coupling selected.\n");
+        std::vector<SharedMatrix> H_SO(3);
+        if (options_.get_str("RELATIVISTIC") == "ZORA") {
+            H_SO = mintshelper()->ao_zora_spin_orbit();
+        } else {
+            throw PSIEXCEPTION("RELATIVISTIC option must be in [\"ZORA\"] when SPIN_ORBIT_COUPLING TRUE");
+        }
+        std::complex<double> i(0, 1);
+
+        for (int h = 0; h < nirrep_; h++) {
```
EricaCMitchell: The loop runs over `h`, but `H_SO[k]->get(p, q)` reads irrep 0 of an AO matrix. This is correct only for C1. `common_init` throws for `nirrep_ > 1` at line 83, so the code is safe at this time. Add a comment that names this dependency.

Generated by: Claude Opus 5

psi4/src/psi4/libscf_solver/cghf.cc
```diff
@@ -0,0 +1,804 @@
...
+            int nso = nsopi_[h];
+            for (int p = 0; p < nso; p++) {
+                for (int q = 0; q < nso; q++) {
+                    // Note that I multiply by i to get H_SO Hermitian
+                    F0_->block(h)(p, q)         += H_SO[2]->get(p,q)*i; // +Hz [0,0]
```
EricaCMitchell: `H_SO[0]`, `H_SO[1]`, and `H_SO[2]` are magic indices. Give them names, for example `const auto& Hx = H_SO[0];`.

The comment `//+iHy [0,1]` at line 227 is also confusing. The value that the code adds is `+Hy`. The result is correct, but the comment says something different.

Generated by: Claude Opus 5

psi4/src/psi4/libscf_solver/cghf.cc
```diff
@@ -0,0 +1,804 @@
...
+    // If SOC, then add Hx, Hy, Hz to F0_
+    if (options_.get_bool("SPIN_ORBIT_COUPLING")) {
+        outfile->Printf("\n  CGHF: Spin-orbit coupling selected.\n");
+        std::vector<SharedMatrix> H_SO(3);
+        if (options_.get_str("RELATIVISTIC") == "ZORA") {
+            H_SO = mintshelper()->ao_zora_spin_orbit();
+        } else {
+            throw PSIEXCEPTION("RELATIVISTIC option must be in [\"ZORA\"] when SPIN_ORBIT_COUPLING TRUE");
+        }
+        std::complex<double> i(0, 1);
```
EricaCMitchell, line 275: `std::complex<double> i(0, 1)` needs `<complex>`. This file gets the header through `cghf.h` or through Einsums. Include `<complex>` directly.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -788,6 +788,25 @@ SharedMatrix MintsHelper::so_dkh(int dkh_order) {
...
+    std::string socx(PSIF_AO_SOCX);
+    std::string socy(PSIF_AO_SOCY);
+    std::string socz(PSIF_AO_SOCZ);
+
+    if (!are_ints_cached(socx, include_perturbations)) {
+        compute_so_zora_ints(include_perturbations, true);
+    }
+
+    std::vector<SharedMatrix> H_SO_zora(3);
+    H_SO_zora[0] = cached_oe_ints_[std::make_pair(socx, include_perturbations)];
```
EricaCMitchell, line 833: `operator[]` returns a null `SharedMatrix` on a miss. Use `.at(key)` so that a miss throws.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -788,6 +788,25 @@ SharedMatrix MintsHelper::so_dkh(int dkh_order) {
     return dkh;
 }
 
+std::vector<SharedMatrix> MintsHelper::ao_zora_spin_orbit() {
+    bool include_perturbations = true;
```
EricaCMitchell, line 822: `ao_zora_spin_orbit()` hard-codes `include_perturbations = true`. As a result, the call throws when `PERTURB_H` is on. The spin-orbit integrals do not depend on the perturbation.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/mintshelper.cc
```diff
@@ -1476,6 +1511,19 @@ void MintsHelper::compute_so_zora_ints(bool include_perturbations) {
 
     // Overwrite cached integrals
     cached_oe_ints_[std::make_pair(PSIF_SO_T, include_perturbations)] = so_kinetic_zora;
+
+    if (spin_orbit) {
+        // Don't consider SO integrals until CGHF supports point-group symmetry
```
EricaCMitchell, line 1546: This comment sits above code that does compute the integrals. The comment means that the matrices stay in the AO basis. Rewrite it.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.h
```diff
@@ -61,35 +61,47 @@ class BasisFunctions;
...
     /*! @{
      * Fills an empty SharedMatrix with the computed ZORA kinetic integral.
      * @param T Shared matrix object that will hold the ZORA kinetic integrals.
      */
-    void compute(SharedMatrix);
+    void compute_TSR(SharedMatrix&);
+    /*! @} */
+
+    /*! @{
+     * Fills three empty SharedMatrix objects with the real x, y, and z components
```
EricaCMitchell, line 78: `compute_HSO` adds to the three matrices. It does not fill them. Write "Adds the x, y, and z components to three SharedMatrix objects".

Two more notes for this block. `compute_TSR` and `compute_HSO` both need `setup()` first. Neither function makes sure that `grid_` and `pworkers_` exist, so a direct call gives a null dereference. Record this order in the comment, or add a guard.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
```diff
@@ -100,10 +91,15 @@ void ZORA::setup() {
     outfile->Printf("    Total number of batches: %d\n", nblocks);
 }
 
-void ZORA::compute(SharedMatrix T_SR) {
-    timer_on("ZORA");
+void ZORA::setup() {
```
EricaCMitchell, line 94: `setup()` pushes into `pworkers_` and does not clear the vector first. A second call to `setup()` doubles the vector.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
```diff
@@ -278,6 +267,86 @@ void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, Sh
...
+            auto phi_y = pworkers_[thread]->basis_value("PHI_Y");
+            auto phi_z = pworkers_[thread]->basis_value("PHI_Z");
+
+            // Preprocess kernel veff/(4c²-2veff) * weight
+            double* w = block->w();
+            for (int p = 0; p < npoints; p++) {
+                kernel[p] = veff_block->get(p) / (4. * C2 - 2. * veff_block->get(p)) * w[p];
+            }
+
+            for (int l_mu = 0; l_mu < local_nbf; l_mu++) {
```
EricaCMitchell, line 320: `auto local_nbf = bf_map.size()` is a `size_t`. The comparison `int l_mu < local_nbf` gives a `-Wsign-compare` warning. Use the same type for both. The same pattern is at line 249.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
```diff
@@ -278,6 +267,86 @@ void ZORA::compute_TSR(std::vector<std::shared_ptr<BasisFunctions>> pworkers, Sh
     }
 
     T_SR->copy_upper_to_lower();
+    timer_off("ZORA: Scalar Relativistic Kinetic");
+}
+
+// Compute the spin-orbit coupling Hamiltonian in the AO basis.
+void ZORA::compute_HSO(SharedMatrix& Hx, SharedMatrix& Hy, SharedMatrix& Hz) {
+    outfile->Printf("\n\n    ZORA: You called compute_HSO. Rejoice!\n");
+    // Speed of light in atomic units squared
```
EricaCMitchell, line 276: When `ZORA_NR_DEBUG` is true, `veff_` is zero. `H^SO` is then exactly zero. The user gets no message. Print a warning for this combination of options.

Generated by: Claude Opus 5

psi4/src/export_mints.cc
```diff
@@ -1489,6 +1489,7 @@ void export_mints(py::module& m) {
         // One-electron properties and
         .def("ao_pvp", &MintsHelper::ao_pvp, "AO pvp integrals")
         .def("ao_dkh", &MintsHelper::ao_dkh, "AO dkh integrals")
+        .def("ao_zora_spin_orbit", &MintsHelper::ao_zora_spin_orbit, "AO ZORA Spin-orbit coupling integrals")
```
EricaCMitchell, line 1646: The docstring is `"AO ZORA Spin-orbit coupling integrals"`. It does not say that the return value is three anti-symmetric AO matrices. It also does not say that the call forces the ZORA integral code to run.

Generated by: Claude Opus 5

tests/pytests/test_zora_soc_ints.py
```diff
@@ -0,0 +1,26 @@
+"""
+This file tests the ZORA spin-orbit coupling integrals from libmints
+"""
+import numpy as np
+np.set_printoptions(suppress=True, linewidth=200, precision = 6)
```
EricaCMitchell, line 5: This line runs at import and changes the numpy state for the full pytest session. Delete it.

Generated by: Claude Opus 5

tests/pytests/test_zora_soc_splitting.py
```diff
@@ -0,0 +1,81 @@
...
+This file tests the ZORA spin-orbit coupling 2p and 3d orbital energy splitting.
+"""
+import numpy as np
+import pytest
+import psi4
+from addons import uusing
+
+pytestmark = [pytest.mark.api, pytest.mark.quick]
+
+@pytest.fixture
```
EricaCMitchell, line 11: The `mo_energies` fixture has function scope. Three tests use it, so 6 SCF calculations run instead of 2. Use `@pytest.fixture(scope="module")`.

Generated by: Claude Opus 5

tests/pytests/test_zora_soc_splitting.py
```diff
@@ -0,0 +1,81 @@
...
+    e_so, wfn_so = psi4.energy('scf', return_wfn=True)
+
+    def get_2p3d_energies(wfn):
+        """Returns 2p and 3d energies of single atom."""
+        # GHF energies are split into alpha/beta due to factory limitations.
+        mo_energy_a = wfn.epsilon_a().to_array()
+        mo_energy_b = wfn.epsilon_b().to_array()
+        mo_energies = np.sort(np.append(mo_energy_a, mo_energy_b))
+
+        mo_e_2p = mo_energies[4:10]
```
EricaCMitchell, line 42: `mo_energies[4:10]` and `mo_energies[18:28]` are hard-coded. They are correct for Kr with 3-21G only. Add a comment that names the shell order.

Generated by: Claude Opus 5

tests/pytests/test_zora_soc_splitting.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return (mo_e_2p, mo_e_3d, mo_e_2p_so, mo_e_3d_so)
+
+@uusing("einsums")
+def test_2p_splitting(mo_energies):
+    """
+    Demonstrates the lack of 2p orbital splitting without SOC and shows splitting with SOC.
+    """
+    mo_e_2p, _, mo_e_2p_so, _ = mo_energies
+    assert np.allclose(mo_e_2p, mo_e_2p[0])
```
EricaCMitchell, line 57: `np.allclose` uses `rtol=1e-5` by default. For a 2p energy near -190 Ha, this permits a difference of 1.9e-3 Ha. Give an explicit `atol` and `rtol` here and in the other asserts.

Generated by: Claude Opus 5

tests/zora-soc/input.dat
```diff
@@ -0,0 +1,29 @@
...
+
+e = energy('scf')
+
+set {
+    spin_orbit_coupling true
+}
+
+e_so = energy('scf')
+
+compare_values(True, e_so < e, "ZORA spin-orbit coupling lowers energy")
```
EricaCMitchell, line 30: `compare_values(True, e_so < e, "...")` passes booleans to a float comparison. Use `compare(True, e_so < e, "...")`.

Generated by: Claude Opus 5

tests/zora-soc/input.dat
```diff
@@ -0,0 +1,29 @@
...
+compare_values(e_so, ref_e_so, 6, "ZORA spin-orbit energy")
```
EricaCMitchell, line 31: The degeneracy pattern test in `test_zora_soc_splitting.py` is a good check, because it tests the physics and not only the code path. Three gaps stay:

- **No comparison against an independent program.** The `zora` test compares against pyscf. This test compares only against its own earlier output. The 2p and 3d splittings of Kr are the correct quantities to compare against NWChem, ORCA, or pyscf. Add one such number here.
- **Single atoms only.** Both tests use one atom. The Becke grid for one atom is a special case. Add a diatomic molecule, for example HI or Br2. This tests the multi-center grid and the partition function.
- **No rotational invariance test.** `Hx`, `Hy`, and `Hz` transform as an axial vector under rotation, so the total energy must not change. A grid-based integral can break this rule. Rotate the molecule by 45 degrees and compare the energy.

Generated by: Claude Opus 5

doc/sphinxman/source/scf.rst
```diff
@@ -341,6 +383,26 @@ keyword to true::
...
+can break collinearity, they are always coplanar; the wavefunction will always
+be an eigenfunction of one of the :math:`\hat Sx`, :math:`\hat Sy`, or :math:`\hat Sz` operators. 
+To obtain a completely spin-broken solution, complex orbitals must also
+be introduced to break the final two spin-symmetries enforced: :math:`\hat K` and :math:`\hat Sn`. 
+This is modeled using complex generalized Hartree-Fock (CGHF) and is available
+as an SCF reference with limited capabilities::
```
EricaCMitchell, line 400: Should the limitations be listed here? If there is a hard restriction, such as the atomic number 104 elsewhere, should it be explicit here that the code only supports certain elements.

EricaCMitchell (changes requested):
# First-Pass Review
Two items block a merge:

- A debug print reaches the user output at `zora.cc:275`.
- `ao_zora_spin_orbit()` overwrites the cached kinetic integrals.

Two performance items are small and remove most of the cost of the new code:

- The inner grid loop in `compute_HSO` does six scattered matrix writes for each point.
- `compute_TSR` and `compute_HSO` each make one full pass over the grid.

### Personal Nits (Human-Generated)

I'd change `compute_HSO` to `compute_H_SOC` and `compute_TSR` to `compute_T_SR`. This way it is more explicit that it is computing the spin-orbit coupling Hamiltonian, and it is computing the scalar relativistic kinetic energy. For the `compute_TSR`, it took me a second to understand it was the scalar relativistic corrections and not a misspelling of an abbreviation for time-reversal symmetry.

Be careful with the use of `SO` vs. `SOC`. `SO` is used throughout Psi4 to stand for symmetry orbitals, I would use `SOC` so there are no collisions within the codebase and that it is plain that spin-orbit coupling is what's being addressed.

### Mixed tabs and spaces

`psi4/src/psi4/libmints/mintshelper.cc:1492-1496` and `1505-1510` mix tabs and spaces. `psi4/src/psi4/libmints/zora.h:90-91` has the same problem. Some of these lines come from the CGHF base branch. Run clang-format on both files.

### What I checked and found correct

This list records the checks that passed, so that a later reader does not repeat the work.

- **The kernel.** `compute_HSO` uses `veff / (4c^2 - 2 veff)`, and `compute_TSR` uses `c^2 / (2c^2 - veff)`. The difference is exactly 1/2 for every value of `veff`. The extra term is the integral of a curl over all space, which is zero. The form in the code goes to zero as `veff` goes to zero, which is the better choice.
- **No division by zero.** The van Wüllen model potential is attractive, so `veff <= 0` and `4c^2 - 2 veff >= 4c^2`.
- **The skipped diagonal.** For `l_mu == l_nu`, the cross product is zero at every point. The skip is exact.
- **The spin structure.** `cghf.cc:223-228` builds `i * (sigma_x Hx + sigma_y Hy + sigma_z Hz)`. Each of the six lines matches one Pauli term. The assembly in `test_zora_soc_ints.py:25` agrees with it.
- **One integral evaluation for each SCF.** `HF::form_H` and `ao_zora_spin_orbit()` use the same cache key and the same `MintsHelper` object. The reference output confirms that the ZORA banner prints one time for each energy call.
- **The test result.** Kr is a closed-shell atom, so the first-order spin-orbit energy is zero by symmetry. The decrease of 2.45 mHa comes from second-order orbital relaxation. The assertion `e_so < e` is thus the correct variational statement.

Generated by: Claude Opus 5

psi4/src/psi4/libmints/zora.cc
NathanGillispie: LOL yeah that's leftover.

psi4/src/psi4/libmints/mintshelper.cc
NathanGillispie: Ah that's a good catch, I'll add that.

doc/sphinxman/source/scf.rst
NathanGillispie: This is @mnward10 stuff from the old branch. That branch is outdated, but the new complexwavefunction stack will have the documentation near the end of the stack (I'm thinking layer 9). Ideally if you have comments you could put them there, but it's not open yet and I suspect won't be for a bit. If you want, I can just make documentation the next layer and put things out of order. Or I can just keep these in mind for later. Just let me know.

tests/zora-soc/input.dat
NathanGillispie: Tests against independent programs are very difficult here because ZORA uses a grid-based approach. NWchem is the closest program to test the SR ZORA against since it is also grid-based and uses identical model potential. Unfortunately, NWChem does not include a variational SOC because it does not have a generalized SCF. AFAIK, ORCA doesn't have a variational SOC either. PySCF doesn't have ZORA. However, my advisor wrote an in-house program that does do variational SOC with generalized SCF which is what I used to verify correctness. It also uses PySCF as a backend.

I agree, more tests are needed and in greater variety.

doc/sphinxman/source/scf.rst
NathanGillispie: I didn't write this, but I do think it's relevant to state at this point in the documentation what CGHF can do. scf.rst is mostly in narrative form and I think anyone reading about CGHF would want to know. I'm pretty sure I put the ZORA documentation in a separate file like `relativistic.rst` or something. I think that would be the best place to put the atom limitation. (but also who is running ZORA on Dubnium??)

NathanGillispie:
Thanks for the comments @EricaCMitchell I agreed with all of them unless otherwise stated. Like I said in one of the comments, most of CGHF is outdated. I will need to rebase this branch onto the new ComplexWavefunction stuff, and since I now have a ComplexMatrix class, I don't have to keep `H_SOC` decomposed. I would still say >80% of the comments will be applicable after I do that.

PR #3373, opened 2026-03-06
Copy of CGHF

cgbriggs99:
For some reason, I can't get it to run the runners on my end. It keeps saying that it can't find the tags when running. Hopefully, I can get it to run here.


.github/workflows/ecosystem.yml
```diff
@@ -312,7 +313,7 @@ jobs:
     - name: Test Addons with CTest (M)
       if: runner.os == 'macOS'
       working-directory: ./objdir
-      run: ctest -L quick -j2 --output-on-failure  # `-LE d2ints` needed for Windows
+      run: ctest -L cghf -j2 --output-on-failure  # `-LE d2ints` needed for Windows
```
ConstLike, line 316: Revert this line back to `ctest -L quick` to fix your MacOS problem. I verified this locally: configured the build without Einsums, `ctest -L cghf -N` shows `Total Tests: 0`, which means CTest returns non-zero `("No tests were found!!!")`.

ConstLike (commented):
Hi, I have a guess, see below.

.github/workflows/ecosystem.yml
cgbriggs99: I’m actually trying to do bug fixes for Eunsums, so I only care about the two tests that are causing errors.

cgbriggs99: I’m putting diagnostics in the cghf tests so I can find out why the dot product is raising exceptions on Mac intel only

ConstLike: I think this is your case:

In [`libs/Einsums/BLASVendor/src/dot.cpp`](https://github.com/Einsums/Einsums/blob/main/libs/Einsums/BLASVendor/src/dot.cpp), `zdotc` has two code paths (see the `extern "C"` block line 19 and the implementation around line 90):

```cpp
#ifdef EINSUMS_HAVE_MKL
// pointer ABI: result written to first argument (Fortran convention)
extern void FC_GLOBAL(zdotc, ZDOTC)(std::complex<double> *, int_t *, ...);
#else
// return value ABI: result as C++ return value
extern std::complex<double> FC_GLOBAL(zdotc, ZDOTC)(int_t *, ...);
#endif
```

Apple Accelerate on **Intel** uses the same pointer ABI as MKL, but Einsums takes the `#else` branch -> ABI mismatch -> segfault. On ARM (Apple Silicon), the return convention is compatible, so it works there.

To confirm, add this to your `do_diis()` before the `true_dot` call:

```cpp
// Test: manual dot to prove it's a BLAS ABI issue, not a data issue
std::complex<double> manual_sum{0, 0};
for (int h = 0; h < nirrep_; h++)
    for (int p = 0; p < irrep_sizes_[h]; p++)
        for (int q = 0; q < irrep_sizes_[h]; q++)
            manual_sum += std::conj(ei->block(h)(p, q)) * ej->block(h)(p, q);
printf("manual_dot: (%f, %f)\n", manual_sum.real(), manual_sum.imag());

// This will segfault on Intel Mac:
sum += einsums::linear_algebra::true_dot(*ei, *ej);
```

If manual works and `true_dot` crashes = confirmed.

cgbriggs99: Great catch. Do you have a link to this behavior so that I can make sure we fix it properly?

cgbriggs99: I figured out a general fix for the issue. When we build Einsums, it now checks to see if running the subroutine version causes segfaults and chooses an algorithm based on that.

PR #3379, opened 2026-04-03
Adding optional spin-dependence (spin-orbit coupling) to X2C-1e

NathanGillispie:
> [!IMPORTANT]
> This is a branch of the megafeature `complex_wfn` branch, which is set to track the complex wavefunction stack #3487. Eventually, this will be rebased onto the complex wavefunction work. Don't mind the diffs, there's only about 900 in this leaf.

## Description
This PR creates a new class `SOX2C1e`, separate from the existing `X2CInts` class, to compute the full spin-dependent X2C integrals. The spin-free modified Dirac equation for one electron is typically expressed as

$$h^{\text{D},\text{sf}}=\begin{bmatrix} V & T \\\ T & \frac{1}{4c^2}W-T \end{bmatrix}$$

Where $V$ and $T$ are the usual potential and kinetic energy integrals in the AO basis, and $W$ is the scalar integral $\mathbf{p}V\cdot \mathbf{p}$. Nominally, the spin-dependent equation does not change, but $W$ becomes

$$W\equiv(\sigma\cdot\mathbf{p})V(\sigma\cdot\mathbf{p})=\underbrace{\mathbf{p}V\cdot \mathbf{p}}_\text{spin-free}+\underbrace{i\sigma\cdot\mathbf{p}V\times \mathbf{p}}_\text{spin-dep.}$$

In practice, $h^{\text{D},\text{sf}}$ benefits from a size reduction due to symmetry reasons. So for the spin-dependent Dirac equation, the $T$ and $V$ integrals are twice the size you would encounter in the spin-free version. Thankfully, the spin-dependent part of $W$ is already given in Libint2, so few modifications were needed to access it.

Aside from the larger Dirac equation and additional terms to $W$, the rest of X2C is conducted as usual with the exception of returning $V_\text{X2C}$ and $T_\text{X2C}$. Instead of returning the 2-component integrals, which would have twice the expected dimension, they are decomposed into pauli form. That is, the non-scalar part of $V_\text{X2C}+T_\text{X2C}$ is returned as the vector of shared matrices labeled (`PSIF_SO_SOX`, `PSIF_SO_SOY`, `PSIF_SO_SOZ`) and only the scalar parts of $V$ and $T$ are returned as themselves. I could not mathematically prove that such a decomposition (into entirely real matrices) always exists. To check this, the Frobenius norm of the residual imaginary components is printed to the output. In my testing, this is less than 1e-8 for each decomposed matrix, and a warning is shown if it exceeds an arbitrary threshold.

- [x] Added boolean option `SPIN_ORBIT_COUPLING`
- [x] Added mintshelper pybind to access X2C spin-orbit integrals: `so_x2c_spin_orbit`

## Dev notes & details
- [x] Renamed class `X2CInt` to `SFX2C1e`
- [x] Designed to be similar to the SF X2C1e implementation, but most functions have more useful signatures and most working quantities are kept alive within the scope of `SOX2C1e::compute` rather than all as private class members. 
- [x] Merges cleanly with #3371.
- [x] Accessing integrals is done in the same way as above.
- [x] Does not assume C1 symmetry, but gives a warning to users when they use non-C1 symmetry.

## Checklist

## Status
- [ ] Ready for review


NathanGillispie:
I have no idea why github builds are failing, seems to have something to do with libint2?

JonathonMisiewicz:
I believe this bug occurred on an older version of 3369. I recommend the two of you get 3369 passing tests, and you then re-sync.

JonathonMisiewicz:
Correction: more specifically, you need [this](https://github.com/psi4/psi4/pull/3377/files#diff-8da6fd2db04fd21f34aabd66c718127b4c110589d9a18885d0cea72e79058f39R738), as had been mentioned on Slack.

NathanGillispie:
Okay, that fix worked. Now the issue seems to be outside my scope: something with `dftd3`.

PR #3387, opened 2026-04-23
fix multiple defs for psimrcc for unity builds

loriab:

- @fevangelista , I've had significant success speeding up the psi4 build (2-3x) with unity/jumbo builds, but for the psimrcc module I hit some multiple definition errors when it concatenates the files together. This PR is AI's solution. Does it look ok or make you uneasy? Tests pass, but I haven't explored farther.
- It's a one-liner to exempt psimrcc from the unity build at a reproducible but not hefty cost, so that's the easy backup if this PR needs a lot of work. See https://github.com/psi4/psi4/commit/7c61b21d8f9bd339eef0fae91f3e7435d00bedca for commands to quickly force the unity build if you want to try it out.
- I've run a full test suite with these changes under unity build, and it's clean.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/psimrcc/psimrcc_wfn.h
+++ b/psi4/src/psi4/psimrcc/psimrcc_wfn.h
@@ -32,8 +32,12 @@
 #include "psi4/libmints/wavefunction.h"
 
 #include "psi4/libqt/qt.h"
+#ifndef START_TIMER
 #define START_TIMER(a) timer_on((a));
+#endif
+#ifndef END_TIMER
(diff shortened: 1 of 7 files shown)


JonathonMisiewicz (approved):
Let's leave final review to @fevangelista.

fevangelista (approved):

fevangelista:
@loriab: This looks like a good fix. At this point in time (I am not sure what I didn't do this in the past), all of these macros could just be turned into functions. But this is a reasonable patch.

PR #3389, opened 2026-05-07
make qcmanybody optional

loriab:

- [x] QCManyBody is an optional addon, no longer a required addon like in v1.10, nor using internal code like in pre-v1.10 . Computations involving `bsse_type` need additional `conda install qcmanybody -c conda-forge` or `pip install qcmanybody` .

## Dev notes & details

- [x] We try to keep req'd deps to a minimum, and 3 py (qcel, qcng, optking) and 3 compiled (L2, Libxc, g2g) are well justified by broad use. QCManyBody was req'd on the grounds that its capability had previously been internal, and it's awkward with the distributed driver to signal for install-this. But I think this wasn't a great choice. A few tweaks here makes it optional.
- [x] the pytest label `nbody` had been the signal for these types of calcs before. now it's `qcmanybody` in keeping with other addons. but that's aliased to nbody so no behavior lost
- [x] TODO if not v1.12, correct exact version in a few places (nbody.rst, build_planning.rst, codedeps.yaml)
- [x] Windows started having a lot of trouble with § symbol, so there's changes to compensate for that
- [x] I realized there were <addon>_PYMOD that weren't defined, so I cleared those out of psi4/__init__

## Questions
- [x] ok to optional this?

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


susilehtola (approved):
There are many unrelated changes to the core but this looks fine

TiborGY:
I gave the diff of this PR to Claude Fable 5 and these are the comments it generated:

> 
> ### Main finding — codedeps `conda.cmake` not updated, so CI's ctest lanes silently lose 20 tests
> 
> Every comparable buildable-optional addon in `codedeps.yaml` (gdma, libefp, pcmsolver, bse) carries `ENABLE_<addon>: true` in its `conda.cmake` block — that's the mechanism by which the path-advisor-generated cmake cache does "the `-D ENABLE_<addon>=ON` work" for CI (ecosystem.yml's own comment). The patch flips qcmanybody to `required: false` but leaves its `conda.cmake` with only the commented-out `//CMAKE_INSIST/DISABLE` entries. Consequences:
> 
> 1.  The generated cache never sets `ENABLE_qcmanybody=ON`, so ecosystem CI's `ctest -L quick` and Azure Windows' `ctest --label-regex smoke` (which passes no addon flags at all) will no longer register the 20 gated ctests — including the smoke test `dfmp2-1` and tutorial `tu6-cp-ne2`. Pytest lanes keep coverage via runtime `uusing` detection since qcmanybody stays in every conda env, but the ctest lanes and any default source build (`cmake` with no flags) lose them.
> 2.  `psi4-path-advisor cache --insist` becomes a no-op for qcmanybody, since the INSIST check now lives inside `if(ENABLE_qcmanybody)`.
> 
> Suggested fix: add `"ENABLE_qcmanybody": true` to that entry, mirroring bse's.
> 
> ### Secondary finding — the headline feature has zero CI coverage
> 
> `test_missing_qcmanybody.py` self-skips when qcmanybody is installed, and it's installed in every CI env (all `devtools/conda-envs/*` files, Azure's conda install). So the graceful-degradation path this PR exists for is never exercised in CI. Worth adding a lane that removes qcmanybody before pytest — the disabled `if: false` "Special Config — QCManyBody Dep" steps in ecosystem.yml/docs-pr.yml are a ready-made template.
> 
> ### Nits
> 
> In the `except ImportError` branch, `AtomicSpecification = Any` clobbers the class successfully imported from **qcelemental** two lines above — harmless given the guards, but it's not a qcmanybody symbol and the reassignment is misleading. In `_normalize_nbody_text`, the third replacement `s.replace("\xa7", "§")` is a literal no-op (`"\xa7"` _is_ `"§"`). The re-flowed dft test line in `tests/CMakeLists.txt` carries trailing whitespace (the `git apply` warning). Only `nbody-multi-level/input.dat` was ASCII-fied; three other inputs (`nbody-vmfc-hessian`, `nbody-cp-gradient`, `dftd3/nbody-cp-gradient`) still contain literal `§` — the new `PYTHONIOENCODING=utf-8` in `ctest_runner` mitigates on Windows, but the fix is inconsistent. And the codedeps note "Optional for v1.12. Required for v1.10." skips v1.11, where it was also required.
> 
> ### Behavioral notes reviewers should sign off on deliberately
> 
> A default source build no longer installs qcmanybody or registers the nbody ctests; users hit a clear, actionable `ModuleNotFoundError` at runtime. And because `uusing("qcmanybody")` applies the `addon` mark (plus the `nbody` alias, a nice back-compat touch), the former core tests like `dfmp2-1` now shift into `-m "addon"` CI selections and out of any `-m "not addon"` lanes — intended, but it changes lane membership.
>

JonathonMisiewicz (approved):

psi4/driver/driver_nbody.py
```diff
@@ -118,10 +118,33 @@
...
+
+_QCMANYBODY_IMPORT_ERROR = None
+try:
+    from qcmanybody.models.v2 import ManyBodyInput, ManyBodyResult, BsseEnum
+    from qcmanybody import ManyBodyCore
+    from qcmanybody.v2 import ManyBodyComputer as ManyBodyComputerQCNG
+    from qcmanybody.utils import delabeler, labeler, translate_qcvariables, modelchem_labels
+except ImportError as err:
+    _QCMANYBODY_IMPORT_ERROR = err
+    AtomicSpecification = Any
```
loriab:
```suggestion
```

.github/workflows/ecosystem.yml
```diff
@@ -456,6 +456,26 @@ jobs:
...
+        conda list
+        PYTHONPATH=stage/lib python -c "import psi4; print(psi4.__file__)"
+
+    - name: dlopen -- Test Addons with Pytest
+      if: matrix.cfg.python-version == '3.14' && runner.os != 'Windows'
+      working-directory: ./objdir
+      run: |
+        export PYTHONPATH=stage/lib
+        pytest --cache-clear -v -rws --color yes -n auto -m qcmanybody stage/lib/psi4/tests/
+        pytest --cache-clear -v -rws --color yes -n auto stage/lib/psi4/tests/test_missing_qcmanybody.py‎
```
loriab:
```suggestion
        pytest --cache-clear -v -rws --color yes -n auto -k missing stage/lib/psi4/tests/
```

PR #3372, opened 2026-03-05
F-SAPT(DFT), SAPT(DFT)-D3/-D4, and SAPT(DFT) External Potentials

Awallace3:
## Description
This PR extends SAPT(DFT) to have an FSAPT energy decomposition, new SAPT(DFT)-D3/-D4 methods, external-potentials with MM charges (point or diffuse). It updates SAPT/FISAPT docs with new usage guidance and adds pytest coverage for the new SAPT(DFT), FSAPT(DFT), -D3/-D4, and external-potential behavior. Also, timers can be exported to a python dictionary or csv file. This PR must wait until the [QCEngine PR#495](https://github.com/MolSSI/QCEngine/pull/495) is added for the empirical dispersion damping parameters here require those definitions.

## User API & Changelog headlines

- [x] 1: FSAPT(DFT) support with ``SAPT_DFT_DO_FSAPT`` option. Another PR will regularize both this and SAPT(DFT) functional calls.
- [x] 2: Added new SAPT(DFT) energy methods: ``sapt(dft)-d3(i)``, ``sapt(dft)-d3(s)``, ``sapt(dft)-d4(i)``, ``sapt(dft)-d4(s)``, ``dft-d3(sapt)``, and ``dft-d4(sapt)``.
- [x] 3: External point/diffuse charge support in SAPT(DFT).
- [x] 4: Timer data to Python Dictionary or CSV file. 

## Dev notes & details

- [x] Feature1: Implemented FSAPT(DFT) decomposition paths (using PyEinsums or FISAPT object).
  - SAPT(DFT) cache now includes additional matrices for localization/fragment decomposition.
  - In ``sapt_dft``:
    - einsums path supports localization/partition + F-SAPT components when available;
    - Alternatively, a path builds an FISAPT object for ``felst/fexch/find/fdisp``;
    - empirical pairwise dispersion can be used with FSAPT via the ``FSAPT_EMPIRICAL_DISP`` variable. Note, this is the only way to use SAPT(DFT) with a DFT functional (other than `SAPT_DFT_FUNCTIONAL=HF`) to get FSAPT energies. FDDS dispersion is for FSAPT is not available yet with a check before running the calculation to ensure the user is not requesting this.
  - Added ``saptdft_fisapt.py`` to bridge SAPT(DFT) caches/scalars into FISAPT object internals and variable dropping.

- [x] Feature2: Added SAPT(DFT)-D3/-D4 method routing and dispersion engines.
[...]

PR #3393, opened 2026-05-12
needs_grac_ is now computed rather than cached. code modernization.

JonathonMisiewicz:
## Dev notes & details

- [x] `needs_grac_` variable is removed in favor of a `needs_grac` method
- [x] Miscellaneous code modernization - lots of range based for loops and auto 

## Checklist
- [x] `grac` test and dft smoke still passes

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfunctional/superfunctional.cc
+++ b/psi4/src/psi4/libfunctional/superfunctional.cc
@@ -223,7 +223,7 @@ void SuperFunctional::print(std::string out, int level) const {
         int nxc = 0;
         int nexch = 0;
         int ncorr = 0;
-        for (const auto& [_, xc_code, __]: mix_data) {
+        for (const auto& [_1, xc_code, _2]: mix_data) {
             if (xc_code == 0) {
                 nexch++;
             } else if (xc_code == 1) {
--- a/psi4/src/psi4/libfunctional/superfunctional.cc
+++ b/psi4/src/psi4/libfunctional/superfunctional.cc
@@ -63,7 +63,6 @@ void SuperFunctional::common_init() {
     c_os_alpha_ = 0.0;
     c_ss_alpha_ = 0.0;
 
-    needs_grac_ = false;
     grac_shift_ = 0.0;
     grac_alpha_ = 0.5;
     grac_beta_ = 40.0;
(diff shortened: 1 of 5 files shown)


psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -134,8 +133,7 @@ std::shared_ptr<SuperFunctional> SuperFunctional::build_polarized() {
         sup->vv10_c_ = vv10_c_;
         sup->vv10_beta_ = vv10_beta_;
     }
-    if (needs_grac_) {
-        sup->needs_grac_ = true;
+    if (needs_grac()) {
```
loriab, line 136: So functional.cc keeps `needs_grac_` as data, but for superfunctional, it's now a method?

psi4/src/psi4/libfunctional/superfunctional.cc
```diff
@@ -243,21 +237,21 @@ void SuperFunctional::print(std::string out, int level) const {
 
         if (nxc) {
             printer->Printf("   => Exchange-Correlation Functionals <=\n\n");
-            for (int i = 0; i < mix_data.size(); i++) {
-                if (std::get<1>(mix_data[i]) != 2) continue;
+            for (const auto& [name, xc_code, coeff]: mix_data) {
```
loriab, line 240: I especially like this cleanup

loriab (commented):
checked that there's no conflict with #3372

psi4/src/psi4/libfunctional/superfunctional.cc
JonathonMisiewicz: I don't follow. `functional.cc` knows nothing about GRAC. `superfunctional.cc` stores a `Functional` for GRAC X and another for GRAC C. `superfunctional.cc` orders those two functionals to compute their derivatives and then uses those to edit the derivatives of the "core" superfunctional. On this model, only the `superfunctional.cc` needs to know about GRAC.

If you can point me to a line where `needs_grac_` survives or where GRAC information is used in `functional.cc`, let me know, and I'll get rid of it. I don't see either.

psi4/src/psi4/libfunctional/superfunctional.cc
JonathonMisiewicz: Structured bindings are great. C++-26 makes them even better, but that's for the future.

psi4/src/psi4/libfunctional/superfunctional.cc
loriab: aha, this is a build mtd, so it's ~ copying the same needs_grac data w/i a class. somehow I saw `sup` and though of superclass. all good!

loriab (approved):

jturney (approved):

PR #3394, opened 2026-05-15
MP2-F12 Performance Improvements

EricaCMitchell:
# MP2-F12 Performance Improvements

## Description

This PR delivers performance and correctness improvements to the MP2-F12/3C(FIX) explicitly
correlated subsystem (`psi4/src/psi4/f12/`).

The F12 changes reduce unnecessary memory allocation and redundant I/O by restructuring the energy
computation so that only the quantities actually needed for each electron-pair contribution are
computed or read at any given time. Several helper functions that allocated large rank-4 temporaries
have been eliminated in favor of direct scalar extraction and range-based tensor assignment. Parallel
loop structures are also improved to avoid load imbalance from skipped iterations.

---

### `psi4/src/psi4/f12/`

- [x] **Removed `V_Tilde` and `B_Tilde` helper methods** (`mp2.h`, `mp2.cc`, both `MP2F12` and
  `DiskMP2F12`) — the energy loop now directly extracts the two scalar B̃ and Ṽ elements needed per
  (i,j) pair rather than forming full `nact×nact×nvir×nvir` temporaries. The fixed-amplitude energy
  expression is inlined.
- [x] **Serial triangular-pair loop in `form_f12_energy`** — iterates over unique (i,j) pairs
  (i ≤ j) via a serial nested loop; the scalar B̃/Ṽ extraction per pair is retained but OpenMP
  parallelization is absent because the einsums contractions inside the loop are not thread-safe.
- [x] **Eliminated `set_ERI` overloads** (`mp2.h`, `mp2.cc`, both classes) — manual
  element-by-element copies into `TensorView`/`DiskView` slices replaced with direct einsums
  range-based assignment (`(*ERI)(Range{...}, ...) = *tensor`).
- [x] **Merged `convert_C` overloads into one** — the no-`use_frzn` overload is removed; callers
  pass `/* use_frzn = */ false` explicitly. Implementation rewritten to use `std::memcpy` row-by-row
  (or a single memcpy when the full matrix is contiguous) instead of a nested loop.
- [x] **Hoisted DF metric inverse computation in `form_metric_ints`** — `J^{-1/2}` is built once
  before the orbital-space loop instead of being recomputed for every block; `FittingMetric` is
  now stack-allocated rather than heap-allocated through a `shared_ptr`.
- [x] **Pre-built shell-quartet list in `two_body_ao_computer`** — unique (M,N,P,Q) quartets are
  collected into a flat vector before the OpenMP region (bra and ket permutation symmetry applied
  at build time), replacing the nested four-level loop with `schedule(dynamic)` over the compact
  list; inner fill loops simplified to direct element indexing (`index_Q + q`) removing pointer
  arithmetic.
- [x] **Pre-built shell-triplet list in `three_index_ao_computer`** — unique (B,P,Q) triplets are
  collected analogously (P ≤ Q symmetry at build time), replacing `collapse(3) schedule(guided)`
  with `schedule(dynamic)` over the flat triplet vector; fill loops simplified to direct indexing.
- [x] **Exploited (A|op|B) symmetry in `form_oper_ints`** — only the lower triangle of the
  auxiliary-basis matrix is computed; the upper triangle is filled by copying the transposed value in
  the same pass, halving the number of AO integral evaluations.
- [x] **`form_D` loop re-indexed** — virtual indices now run 0 to `nvir_` (no repeated offset
  arithmetic) and diagonal Fock elements are cached in a local vector before the parallel region.
- [x] **`DiskMP2F12::form_D` made serial** — removed OpenMP parallelization; concurrent HDF5
  writes to the same dataset are not thread-safe.
- [x] **`DiskMP2F12::form_f12_energy` reads B, X, V once** — the three nact⁴ tensors are loaded
  from disk into in-memory tensors before the pair loop; G and D are read per (i,j) pair (nvir²
  each), and C is read only for the (i,j) and (j,i) slices actually needed.
- [x] **Disk ERI stitch-ins batched** — `DiskMP2F12::form_teints` now writes each orbital-space
  block in a single hyperslab assignment instead of `nmo1×nmo3` individual 2-D writes, reducing
  HDF5 round-trips.
- [x] **Added `timer_on`/`timer_off` around `DiskMP2F12::compute_energy`**.
- [x] **Removed empty destructors** — `~MP2F12()` and `~DiskMP2F12()` replaced with `= default`.
- [x] **`t_` made `static`** — the fixed-amplitude lookup has no instance state.
- [x] **`DiskMP2F12` constructor no longer calls `common_init()`** — it was being invoked twice
  (once from `MP2F12::MP2F12`, once from the `DiskMP2F12` constructor).
- [x] **Replaced `pow(f_iA(i,A), 2)` with `f_iA(i,A) * f_iA(i,A)`** in `form_cabs_singles`
  (both classes).
- [x] **Fixed UB: `*target++` → `target++`** in `two_body_ao_computer` — the dereference was a
  no-op but constituted undefined behavior on a raw pointer.
- [x] **Removed shadowing local variable** `std::vector<OrbitalSpace> bs_` in `common_init`.

---

## Notes

- The energy values produced by the new `form_f12_energy` are numerically identical to the original
  — the refactoring extracts exactly the same scalar elements of B̃ and Ṽ that were computed before;
  no approximations are introduced.
- The `DiskMP2F12::form_D` serialization is safe because the per-(i,j) arithmetic cost is
  O(nact² · nvir²) and negligible; the bottleneck for the disk algorithm is I/O, not CPU.
- Shell-quartet (resp. shell-triplet) precomputation in `two_body_ao_computer`
  (resp. `three_index_ao_computer`) adds a small upfront allocation but eliminates branch
  mispredictions in the parallel loop, enables `schedule(dynamic)` load balancing for shells of
  unequal size, and groups all (P,Q) integrals under the same (M,N) auxiliary block — improving
  cache locality during the immediately following MO transformation.
- `form_f12_energy` is not parallelized because the einsums `einsum` calls inside the loop mutate
  internal library state and are not safe to call concurrently from multiple threads.
- The (A|op|B) symmetry exploit in `form_oper_ints` is exact for symmetric two-center operators
  (Coulomb, F12 correlation factor).


psi4/src/psi4/f12/integrals.cc
```diff
@@ -118,13 +82,26 @@ void MP2F12::two_body_ao_computer(const std::string& int_type, einsums::Tensor<d
         ints.push_back(std::shared_ptr<TwoBodyAOInt>(ints[0]->clone()));
     }
 
+    auto& gao = *GAO;
     auto bs1_equiv_bs2 = (bs1 == bs2);
     auto bs3_equiv_bs4 = (bs3 == bs4);
 
-#pragma omp parallel for collapse(2) schedule(guided) num_threads(nthreads_)
+    // Pre-build the flat list of unique (M,N) shell pairs so that the parallel
+    // loop has no wasted iterations from triangular-skip branches.
```
JonathonMisiewicz: Could you clarify this comment? I think that the essential point is that it's more efficient to parallelize over `mn` pairs rather than `M` because the workload per `mn` is going to be more even than per `m`, so there's less danger of some threads being idle. From the comment, it is unclear _if_ parallelization is part of the motivation for pre-build is ambiguous, let alone why the parallelization is important.

psi4/src/psi4/f12/integrals.cc
```diff
@@ -521,6 +493,18 @@ void MP2F12::form_metric_ints(einsums::Tensor<double, 3>* DF_ERI, bool is_fock)
 
     const bool use_frzn = !is_fock;
 
+    // Build J^{-1/2} once; it is independent of the orbital-space block.
+    Tensor<double, 2> AB{"JinvAB", naux_, naux_};
+    {
+        auto metric = std::make_shared<FittingMetric>(DFBS_, true);
+        metric->form_full_eig_inverse(1.0e-12);
+        SharedMatrix Jm12 = metric->get_metric();
```
JonathonMisiewicz: Is it worth giving `Matrix` a `to_tensor` method if einsums is detected? This seems like a very common operation.

psi4/src/psi4/f12/integrals.cc
```diff
@@ -521,6 +493,18 @@ void MP2F12::form_metric_ints(einsums::Tensor<double, 3>* DF_ERI, bool is_fock)
 
     const bool use_frzn = !is_fock;
 
+    // Build J^{-1/2} once; it is independent of the orbital-space block.
+    Tensor<double, 2> AB{"JinvAB", naux_, naux_};
+    {
+        auto metric = std::make_shared<FittingMetric>(DFBS_, true);
```
JonathonMisiewicz: It's nitpicky, I know, but why is `metric` on-heap at all? You don't need any sophisticated memory management here.

psi4/src/psi4/f12/mp2.cc
```diff
@@ -165,48 +166,100 @@ void MP2F12::form_f12_energy(einsums::Tensor<double, 4>* V, einsums::Tensor<doub
     using namespace einsums::tensor_algebra;
     using namespace einsums::index;
 
-    auto E_f12_s = 0.0;
-    auto E_f12_t = 0.0;
-    int kd;
+    // Pre-build triangular pair list for load-balanced parallel dispatch.
+    struct Pair { size_t i, j; };
```
JonathonMisiewicz: What's wrong with the structure of `emplace_back` into `std::vector<std::pair<size_t>>` from earlier?

psi4/src/psi4/f12/mp2.cc
```diff
@@ -165,48 +166,100 @@ void MP2F12::form_f12_energy(einsums::Tensor<double, 4>* V, einsums::Tensor<doub
     using namespace einsums::tensor_algebra;
     using namespace einsums::index;
 
-    auto E_f12_s = 0.0;
-    auto E_f12_t = 0.0;
-    int kd;
+    // Pre-build triangular pair list for load-balanced parallel dispatch.
```
JonathonMisiewicz: Again, vague comment.

JonathonMisiewicz (dismissed):
Good stuff overall, but I have some nitpicks.

JonathonMisiewicz:
Also, please ping us to rerun tests when you see the problems are all linked to @susilehtola's OpenOrbitalOptimizer package. As @loriab can attest, the convergence is very sensitive, which gives it problems with the test suite.

In this case, the next commit will trigger tests to rerun anyways, so there's no point in doing it now.

EricaCMitchell:
## Timing Analysis: AO Integral Loop Restructuring

Benchmarked on H₂O, C₂H₂, and HNCO using MP2-F12/3C(FIX) (conventional and DF paths), aug-cc-pVDZ / cc-pVDZ-F12, 4 threads, 3-run averages.

Two approaches were compared against the original (baseline) code:

1. **Quartets/triplets pre-build (this PR):** Pre-builds a flat vector of all unique `(M,N,P,Q)` / `(B,P,Q)` tuples before the parallel region, then dispatches over the flat vector with `schedule(dynamic)`.
2. **Pairs-only pre-build (rejected alternative):** Pre-builds only the outer pair list (`mn_pairs` / `pq_pairs`), parallelises over pairs, with inner loops sequential — no flat vector.

### Results vs baseline

| Path | Molecule | Total wall Δ | AO kernel Δ | MO trans Δ |
|---|---|---|---|---|
| Conventional | H₂O | −1.1% | +7.6% | −9.6% |
| Conventional | C₂H₂ | −3.4% | +6.5% | −13.7% |
| Conventional | HNCO | **−7.0%** | +5.8% | **−16.4%** |
| DF | H₂O | −4.1% | +9.6% | −10.0% |
| DF | C₂H₂ | −1.1% | +8.6% | −10.6% |
| DF | HNCO | **−3.9%** | +2.6% | **−12.4%** |

The quartets/triplets approach is **1–7% faster overall** (conventional) and **1–5% faster** (DF), with gains growing with system size. The net improvement comes entirely from **MO transformation** (10–16% faster, conv; 10–12% faster, DF): the flat-vector write order keeps `gao`/`bpq` cache lines warm for the immediately following MO contraction. The AO integral kernels are 3–10% slower due to flat-vector allocation overhead, but this cost is outweighed at all tested system sizes.

The pairs-only alternative was tested and rejected: it recovers the AO kernel overhead (5–9% faster kernel) but loses the cache-warming benefit, making MO transformation 11–20% slower and producing a net regression of 1–8% vs the quartets/triplets code.

The Round 1 gain and Round 2 regression are equal in magnitude and scale identically with system size, confirming the cache-warming of `gao`/`bpq` as the driver. Conventional results have CoV ≤ 2%; all HNCO Δ% values exceed 2σ and are statistically reliable.

loriab (approved):
lgtm! thanks for the performance improvements and timings data.

hhjhhaaa:
Hi, 

I ran into the same MP2-F12 performance bottleneck in my own calculations, so I tried a different implementation inspired by local/PNO-F12 and integral-direct ideas.

The main change is pretty simple: keep the reusable three-index quantities around, build the larger pieces only when needed, and contract the F12 correction pair by pair instead of storing the full four-index intermediates. The underlying DF-MP2-F12/3C(FIX) equations are unchanged.

I compared it with this PR applied on top of current master, using the same machine and the same fixed protocol. On the two-water test, #3394 INCORE took 7.25 s vs 2.97 s at 1 thread, and 5.57 s vs 1.74 s at 4 threads. Peak memory was also quite a bit lower. The numerical differences were around 4e-11 Eh.

This is only a small fixed-protocol test, so I don't want to read too much into the exact speedups. But it looks like the pair-wise / on-demand approach may still help on top of the optimizations here.

If this seems useful, I can clean up the implementation and share the code / benchmark details.

EricaCMitchell:
> This is only a small fixed-protocol test, so I don't want to read too much into the exact speedups. But it looks like the pair-wise / on-demand approach may still help on top of the optimizations here.
> 
> If this seems useful, I can clean up the implementation and share the code / benchmark details.

@hhjhhaaa  All contributions are welcome! Just open a PR and I'll look at it. I think I understand the approach you're taking with the "pair by pair".

hhjhhaaa:
> > This is only a small fixed-protocol test, so I don't want to read too much into the exact speedups. But it looks like the pair-wise / on-demand approach may still help on top of the optimizations here.
> > If this seems useful, I can clean up the implementation and share the code / benchmark details.
> 
> @hhjhhaaa All contributions are welcome! Just open a PR and I'll look at it. I think I understand the approach you're taking with the "pair by pair".

Thanks — I opened a draft PR with the streamed implementation and the updated comparison: #3557.

PR #3400, opened 2026-05-18
cmake: forward per-config flag variants to psi4-core subbuild

susilehtola:
## Summary

The outer superbuild's `ExternalProject_Add(psi4-core ...)` forwards `CMAKE_<LANG>_FLAGS` but none of the per-configuration variants (`CMAKE_<LANG>_FLAGS_{DEBUG,RELEASE,RELWITHDEBINFO,MINSIZEREL}`).

A user override like `-DCMAKE_CXX_FLAGS_RELEASE="-DNDEBUG"` on the outer cmake lands in the outer cache but never reaches the inner `psi4-core` build, which silently falls back to CMake's auto-populated defaults. The result is silently-ignored user flags whenever they target a specific build configuration.

This PR forwards all four configuration variants for C, C++, and Fortran (12 new lines in `CMakeLists.txt`).

Without this patch, the user-set flags aren't honored, and e.g. distribution packages aren't built with the flags they're supposed to be built.

## Test plan

- [x] Configure with `-DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE="-O2 -DNDEBUG -DCANARY_FLAG"` and confirm `-DCANARY_FLAG` appears in `objdir/stage/.../compile_commands.json` for files compiled by the inner `psi4-core` subproject (previously it would only appear for outer-layer targets).
- [ ] Existing Release / Debug / RelWithDebInfo builds still succeed unchanged (when the user doesn't override these flags, the defaults CMake populates in the outer scope now flow inward identically — no behavior change).


loriab:
Timely, I was just considering this over at #3353 where I found cmake's config variants overriding my optimization level set through `CMAKE_CXX_FLAGS`.

loriab (approved):
lgtm. I included some of this in #3353 and also tested it there and locally. note that setting these replaces the cmake defaults, not appends.

jturney (approved):

PR #3405, opened 2026-05-19
cmake: fix HDF5 2.x detection (resolve imported targets to file paths)

susilehtola:
## Summary

HDF5 2.x ships its own \`HDF5Config.cmake\` (config mode) and populates \`HDF5_LIBRARIES\` with imported-target names like \`hdf5-shared\` instead of the absolute library file paths that HDF5 1.x's \`FindHDF5.cmake\` returned. Our \`external/common/hdf5/CMakeLists.txt\` stored those names into \`tgt::hdf5\`'s \`INTERFACE_LINK_LIBRARIES\` and re-exported the target. Downstream consumers (ambit, chemps2, einsums, psi4-core) imported \`tgt::hdf5\` from a scope where \`hdf5-shared\` was never an imported target, so the bare names fell through to the linker as \`-lhdf5-shared\`:

\`\`\`
/usr/bin/ld.bfd: cannot find -lhdf5-shared: No such file or directory
\`\`\`

This PR resolves any imported targets in \`HDF5_LIBRARIES\` to their actual library file paths (preferring \`IMPORTED_LOCATION_<config>\` from the first \`IMPORTED_CONFIGURATIONS\` entry, falling back to \`IMPORTED_LOCATION\`) before storing them in \`tgt::hdf5\`. Non-target entries pass through unchanged, so HDF5 1.x builds keep working.

Reported against HDF5 2.1.1 on Fedora.

## Test plan

- [x] Configure with \`ENABLE_CheMPS2=ON\` (or \`ENABLE_ambit\`/\`ENABLE_Einsums\`) against system HDF5 2.x; \`Found HDF5\` line should now print an absolute \`/usr/lib*/libhdf5*.so\` path; \`psi4-core\` link succeeds.
- [ ] Repeat against a system HDF5 1.x — output and link both unchanged.

### DIFF
--- a/external/common/hdf5/CMakeLists.txt
+++ b/external/common/hdf5/CMakeLists.txt
@@ -6,8 +6,38 @@ if(ENABLE_ambit OR ENABLE_CheMPS2 OR ENABLE_Einsums)
 
     find_package(HDF5 QUIET REQUIRED)
 
+    # HDF5 1.x's FindHDF5 sets HDF5_LIBRARIES to absolute library file paths,
+    # but HDF5 2.x ships its own HDF5Config.cmake which (in config mode) sets
+    # HDF5_LIBRARIES to imported-target names like 'hdf5-shared'. Those
+    # targets are only visible in the scope of the HDF5 config file, so once
+    # we stash them into our INTERFACE library and re-export it for
+    # downstream consumers the bare names fall through as link library
+    # names ('-lhdf5-shared'), and the link fails. Resolve any imported
+    # targets to their actual library file paths before storing.
+    set(_HDF5_RESOLVED_LIBS)
+    foreach(_lib IN LISTS HDF5_LIBRARIES)
+        if(TARGET ${_lib})
+            get_target_property(_imp_cfgs ${_lib} IMPORTED_CONFIGURATIONS)
+            set(_loc)
+            if(_imp_cfgs)
+                list(GET _imp_cfgs 0 _first_cfg)
+                get_target_property(_loc ${_lib} IMPORTED_LOCATION_${_first_cfg})
+            endif()
+            if(NOT _loc)
+                get_target_property(_loc ${_lib} IMPORTED_LOCATION)
+            endif()
+            if(_loc)
+                list(APPEND _HDF5_RESOLVED_LIBS ${_loc})
+            else()
+                list(APPEND _HDF5_RESOLVED_LIBS ${_lib})
+            endif()
+        else()
+            list(APPEND _HDF5_RESOLVED_LIBS ${_lib})
+        endif()
+    endforeach()
+
     add_library(hdf5 INTERFACE)
-    set_property(TARGET hdf5 PROPERTY INTERFACE_LINK_LIBRARIES ${HDF5_LIBRARIES})
+    set_property(TARGET hdf5 PROPERTY INTERFACE_LINK_LIBRARIES ${_HDF5_RESOLVED_LIBS})
     set_property(TARGET hdf5 PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${HDF5_INCLUDE_DIRS})
     set (${PN}_VERSION ${HDF5_VERSION})
 


loriab:
We know HDF5 v1.14 is working b/c that's what's been in CI for a while, and several of the Azure lanes are building Einsums and worked fine with this patch.

I had AI switch out HDF5 to 2.1.0 and build Einsums again. Einsums isn't presently compatible with v2, but there's a patch below in case anyone else wants to try it out. (I understand Einsums is going to drop H5CPP and do HDF5 directly, so not pursuing.) With that (and the changeset in this PR of course), it was able to build Einsums and link it into psi4 and run all the mp2f12 tests. Moreover, the i_l_l below is a full path, as the PR intends. lgtm!

### local ` stage/share/cmake/TargetHDF5/TargetHDF5/TargetHDF5Targets.cmake`
```
set_target_properties(tgt::hdf5 PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/psi/toolchainconda/envs/py313/include"
  INTERFACE_LINK_LIBRARIES "/psi/toolchainconda/envs/py313/lib/libhdf5.so.320.1.0"
```

### patch for H5CPP in current v1 Einsums
```
diff --git a/h5cpp/H5Zpipeline_basic.hpp b/h5cpp/H5Zpipeline_basic.hpp
index 1234567..89abcde 100644
--- a/h5cpp/H5Zpipeline_basic.hpp
+++ b/h5cpp/H5Zpipeline_basic.hpp
@@ -44,16 +44,28 @@ inline void h5::impl::basic_pipeline_t::read_chunk_impl( const hsize_t* offset,

        switch( tail ){ // tail = index pointing to queue holding filters
                case 0: // no filters, ( if blocking ) -> data == chunk0 otherwise directly from container
-                       H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk0);
+#if H5_VERSION_GE(2,0,0)
+                       H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk0, &length);
+#else
+                       H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk0);
+#endif
                        break;
                case 1: // single filter
-                       H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk1);
+#if H5_VERSION_GE(2,0,0)
+                       H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk1, &length);
+#else
+                       H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk1);
+#endif
                        length = filter[0](chunk0, chunk1, nbytes, flags[0], cd_size[0], cd_values[0] );
                        break;
                default: // more than one filter
                        throw std::runtime_error("filters not implemented yet...");
                        if( tail % 2 ){
-                               H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk0);
+#if H5_VERSION_GE(2,0,0)
+                               H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk0, &length);
+#else
+                               H5Dread_chunk(ds, dxpl, offset, &filter_mask, chunk0);
+#endif
                                for(int j=tail; j>0; j--){ // invariant: out == buffer holding final result
                                        tmp = in, in = out, out = tmp;
                                        length = filter[j](out,in,length, flags[j], cd_size[j], cd_values[j]);
```

loriab (approved):

external/common/hdf5/CMakeLists.txt
```diff
@@ -6,8 +6,38 @@ if(ENABLE_ambit OR ENABLE_CheMPS2 OR ENABLE_Einsums)
 
     find_package(HDF5 QUIET REQUIRED)
 
+    # HDF5 1.x's FindHDF5 sets HDF5_LIBRARIES to absolute library file paths,
```
JonathonMisiewicz, line 9: Thanks Claude for the comments. :slightly_smiling_face:

JonathonMisiewicz (approved):
Add to merge queue when you think it's ready, @susilehtola.

PR #3411, opened 2026-05-26
fix pytest markers and docs strictness in CI

loriab:
## Dev notes & details

- [x] with the release of QCEngine v0.50, pytest has new markers to complain about, so let's fix them pronto since all the CI is failing. borrowed from https://github.com/psi4/psi4/pull/3341/changes#diff-fb6a686182f16eb54af3c628f38593f347f68aba31de903983023c560288d7a1
- [x] some of the QCSchema v2 update have broken intersphinx links in docs, which causes docs build to fail w/strict so turn that off temporarily. this is also fixed properly by 3341
- [x] we're having trouble getting PRs to 1.10.x to run, so editing the azure control file here on master, too, in case that helps.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/.github/workflows/docs-pr.yml
+++ b/.github/workflows/docs-pr.yml
@@ -102,7 +102,7 @@ jobs:
           -C ${{ github.workspace }}/code/cache_p4docs@objdir.cmake \
           -D CMAKE_BUILD_TYPE=Debug \
           -D CMAKE_CXX_FLAGS="-O0" \
-          -D SPHINXMAN_STRICT=ON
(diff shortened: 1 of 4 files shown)


jturney (approved):

JonathonMisiewicz (approved):
Let's get the proper fixes in ASAP.

loriab:
> Let's get the proper fixes in ASAP.

Yes, docs fixes are ready to review at #3341

PR #3412, opened 2026-05-26
Pybind11v3.0

pot8osince98:
## Description

Upgrades psi4 to pybind11 v3.0 APIs.

- [x] Pybind11 v3.0 is required

## Dev notes & details

- [x] Updated `CMakeLists.txt` files to reflect pybind11 v3.0 requirement
- [x] Replaced `py::module` with`py::module_`
- [x] Replaced `py::class_` with `py::classh`
- [x] Migrated `py::enum_` to `py::native_enum`

## Checklist

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/codedeps.yaml
+++ b/codedeps.yaml
@@ -735,16 +735,16 @@ data:
       host: github
       account: pybind
       name: pybind11
-      commit: v2.13.1
+      commit: v3.0.0
     cmake:
       name: pybind11  # detect Python before pybind11
-      constraint: 2.12.0
+      constraint: 3.0.0
       target: pybind11::module
     conda:
       channel: conda-forge
       name: pybind11
-      constraint: null
-      constraint_note: "Numpy v2 needs >=2.12. For docs, '>=2.10.*,<2.12.0' has adv; search FixedSize."
+      constraint: ">=3.0.0"
+      constraint_note: "v3.0.0 required for py::classh and py::native_enum support."
       cmake:
         pybind11_DIR:
           unix: ${CONDA_PREFIX}/share/cmake/pybind11
--- a/codedeps.yaml
+++ b/codedeps.yaml
@@ -735,7 +735,7 @@ data:
       host: github
       account: pybind
       name: pybind11
-      commit: v3.0.0
+      commit: v3.0.3
     cmake:
       name: pybind11  # detect Python before pybind11
       constraint: 3.0.0
--- a/external/upstream/pybind11/CMakeLists.txt
+++ b/external/upstream/pybind11/CMakeLists.txt
@@ -1,13 +1,5 @@
 
-if(${Python_NumPy_VERSION} GREATER_EQUAL "2")
-        find_package(pybind11 2.12.0 CONFIG QUIET)  # edit in codedeps
-else()
-    if(${Python_VERSION_MINOR} GREATER_EQUAL "11")
-        find_package( pybind11 2.10.1 CONFIG QUIET)
-    else()
-        find_package( pybind11 2.6.2 CONFIG QUIET)
-    endif()
-endif()
+find_package(pybind11 3.0.0 CONFIG QUIET)  # edit in codedeps
 
 if(${pybind11_FOUND})
     message(STATUS "${Cyan}Found pybind11${ColourReset}: ${pybind11_INCLUDE_DIR} (found version ${pybind11_VERSION})")
@@ -25,7 +17,7 @@ else()
     include(ExternalProject)
     message(STATUS "Suitable pybind11 could not be located, ${Magenta}Building pybind11${ColourReset} instead.")
     ExternalProject_Add(pybind11_external
-        URL https://github.com/pybind/pybind11/archive/v2.13.1.tar.gz  # edit in codedeps
+        URL https://github.com/pybind/pybind11/archive/v3.0.0.tar.gz  # edit in codedeps
         UPDATE_COMMAND ""
         CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/external/upstream/pybind11/CMakeLists.txt
+++ b/external/upstream/pybind11/CMakeLists.txt
@@ -17,7 +17,7 @@ else()
     include(ExternalProject)
     message(STATUS "Suitable pybind11 could not be located, ${Magenta}Building pybind11${ColourReset} instead.")
     ExternalProject_Add(pybind11_external
-        URL https://github.com/pybind/pybind11/archive/v3.0.0.tar.gz  # edit in codedeps
+        URL https://github.com/pybind/pybind11/archive/v3.0.3.tar.gz  # edit in codedeps
         UPDATE_COMMAND ""
         CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${STAGED_INSTALL_PREFIX}
                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/psi4/include/psi4/pybind11.h
+++ b/psi4/include/psi4/pybind11.h
@@ -36,6 +36,7 @@
 #include <memory>
 
 #include <pybind11/pybind11.h>
+#include <pybind11/native_enum.h>
 #include <pybind11/stl.h>
 #include <pybind11/stl_bind.h>
 #include <pybind11/operators.h>
--- a/psi4/include/psi4/pybind11.h
+++ b/psi4/include/psi4/pybind11.h
@@ -43,6 +43,4 @@
 
 namespace py = pybind11;
 
-PYBIND11_DECLARE_HOLDER_TYPE(T, std::shared_ptr<T>)
-
 #endif  // PSI4_CORE_PYBIND11_H_H
--- a/psi4/src/export_mints.cc
+++ b/psi4/src/export_mints.cc
@@ -500,9 +500,9 @@ void export_mints(py::module_& m) {
         .finalize();
 
     py::native_enum<Molecule::FragmentType>(m, "FragmentType", "enum.IntEnum", "Fragment activation status")
-        .value("Absent", Molecule::Absent)
-        .value("Real", Molecule::Real)
-        .value("Ghost", Molecule::Ghost)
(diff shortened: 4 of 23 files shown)


external/upstream/pybind11/CMakeLists.txt
```diff
@@ -25,7 +17,7 @@ else()
     include(ExternalProject)
     message(STATUS "Suitable pybind11 could not be located, ${Magenta}Building pybind11${ColourReset} instead.")
     ExternalProject_Add(pybind11_external
-        URL https://github.com/pybind/pybind11/archive/v2.13.1.tar.gz  # edit in codedeps
+        URL https://github.com/pybind/pybind11/archive/v3.0.0.tar.gz  # edit in codedeps
```
loriab:
```suggestion
        URL https://github.com/pybind/pybind11/archive/v3.0.3.tar.gz  # edit in codedeps
```
In `find_project` one wants min useable, while in the url one wants max avail.

codedeps.yaml
```diff
@@ -733,16 +733,16 @@ data:
       host: github
       account: pybind
       name: pybind11
-      commit: v2.13.1
+      commit: v3.0.0
```
loriab:
```suggestion
      commit: v3.0.3
```

psi4/src/export_mints.cc
```diff
@@ -487,21 +487,24 @@ void export_mints(py::module& m) {
...
 
-    py::enum_<Molecule::FragmentType>(m, "FragmentType", "Fragment activation status")
-        .value("Absent", Molecule::Absent)  // Neglect completely
-        .value("Real", Molecule::Real)      // Include, as normal
-        .value("Ghost", Molecule::Ghost)    // Include, but with ghost atoms
-        .export_values();
+    py::native_enum<Molecule::FragmentType>(m, "FragmentType", "enum.IntEnum", "Fragment activation status")
+        .value("Absent", Molecule::Absent)
+        .value("Real", Molecule::Real)
+        .value("Ghost", Molecule::Ghost)
```
loriab: Could you look into whether pb11 finally defined a way to include docstrings on enum values and, if so, add the previous comments as docstrings?

pot8osince98: Pb11 added a way to include doctstrings on enum values. I have added the old comments as docstrings.

psi4/include/psi4/pybind11.h
```diff
@@ -36,13 +36,12 @@
 #include <memory>
 
 #include <pybind11/pybind11.h>
+#include <pybind11/native_enum.h>
 #include <pybind11/stl.h>
 #include <pybind11/stl_bind.h>
```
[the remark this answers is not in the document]
pot8osince98: In pybind11 v3.0, `smart_holder` was merged into the core library. `py::smart_holder` is available directly from `pybind11/pybind11.h`.

loriab (approved):

TiborGY (commented):
This is what Claude Fable 5 generated when I plugged this PR's diff into it:

> 
> 
> ~~### Blocking bug: missing `native_enum.h` in two translation units~~
> 
> ~~The patch adds `#include <pybind11/native_enum.h>` only to the shared header `psi4/include/psi4/pybind11.h`. But **`export_mints.cc` and `export_trans.cc` don't include that header** — they include pybind11 headers directly (`stl.h`, `pytypes.h`, `stl_bind.h`, `operators.h`, and `numpy.h` for mints). I walked their full transitive include graphs: neither ever reaches `psi4/pybind11.h` or `native_enum.h`. `native_enum.h` is opt‑in in v3 (not pulled in by `pybind11.h`), and a compile probe with exactly their include set confirms the result:~~
> 
> ```
> error: 'native_enum' is not a member of 'py'
> ```
> 
> ~~So these two files — which contain 11 of the 15 enum conversions — will not compile. `core.cc` and `export_pcm.cc` are fine because they include the shared header. Fix: add `#include <pybind11/native_enum.h>` (or `"psi4/pybind11.h"`) to both files.~~

> 
> ### User-visible behavior change worth a release note
> 
> `native_enum` produces real `enum.IntEnum` types, and on Python ≥3.11 `str()`/`format()` of an IntEnum member yields the integer: `str(core.SaveType.LowerTriangle)` is now `"1"`, not `"SaveType.LowerTriangle"`; `repr` changes to `<SaveType.LowerTriangle: 1>`. Members are now `isinstance(..., int)` and JSON‑serializable. I grepped the psi4 driver and tests: all in‑repo uses (`DiagonalizeOrder` in `writer.py`/SAPT, `PCM.CalcType` in `scf_iterator.py`, `SaveType` in `diis.py`) are attribute access passed straight into bound functions, so nothing in‑repo breaks — but downstream scripts that print these enums will see different output.
> 
> ### Downstream plugin/ecosystem impact — recommend two additions
> 
> pybind11 v3 bumped its cross‑extension ABI, so extensions passing objects between compiled modules must all use pybind11 3.0; older ABIs aren't supported. Every psi4 plugin must therefore be rebuilt. More importantly, I tested the case of a plugin that binds a _subclass_ of a psi4 class the old way (`py::class_<Derived, std::shared_ptr<Derived>, Base>`) against the new smart‑holder base: it compiles and imports, but passing the instance where `std::shared_ptr<Base>` is expected fails at runtime with `RuntimeError: Non-owning holder (load_as_shared_ptr)`. So plugins binding, e.g., `Wavefunction` subclasses need a code change to `py::classh`, not just a rebuild. Two suggestions: [Iscinumpy](https://iscinumpy.dev/post/pybind11-3-0-0/)
> 
> 1.  `psi4/psi4Config.cmake.in` line 130 still says `find_package(pybind11 CONFIG REQUIRED)` with no version — bump it to `3.0` so plugin builds against a stale pybind11 fail fast at configure time rather than mysteriously at runtime.
> 2.  Add a note to the release notes / `plugins.rst` about the rebuild + `classh` requirement.
> 
> Minor related nit: the non‑lock conda env yamls (`linux-64-buildrun.yaml`, `linux-64-docs.yaml`, etc.) still list bare `pybind11`. That currently resolves to 3.0.3 on conda‑forge so it works, but mirroring the `pybind11>=3.0.0` pin there (as the patch did for the Windows pipeline and docs‑cf) would be more robust.
> 
> ### Pre-existing latent bug the patch preserves (`BSVec`)
> 
> Not a regression, but since the patch touches the line: `py::bind_vector<std::vector<ShellInfo>>(m, "BSVec")` is broken **both before and after** this patch. psi4's shared setup includes `pybind11/stl.h` in the same translation units as `bind_vector` with no `PYBIND11_MAKE_OPAQUE(std::vector<ShellInfo>)`, so the generic list caster and the bound type collide — any Python‑side call on a `BSVec` instance (`len(v)`, `v.append(...)`) recurses infinitely through `list_caster::load → sequence::size → __len__ → ...` and segfaults. I reproduced this under both 2.13.6 and 3.0.3, with and without `smart_holder`. `BSVec` is referenced nowhere else in the repo, so nobody has hit it; while editing that line anyway, I'd either delete the binding or make the type opaque. The `py::smart_holder` template argument the patch adds there is harmless either way.
> 
> ### Small cosmetic notes
> 
> Removing the shared_ptr `PYBIND11_DECLARE_HOLDER_TYPE` is safe (shared_ptr support is built in), and the `py::module` → `py::module_` renames are optional in v3 (the alias still exists) but good hygiene. A handful of continuation lines left over from the mechanical `py::class_<X, std::shared_ptr<X>>` → `py::classh<X>` shortening now have drifted hanging indents (e.g., the `ERI`, `QuadrupoleInt`, `AOShellCombinationsIterator` declarations in `export_mints.cc`) — worth a clang‑format pass.
> 
> **Bottom line:** with the `native_enum.h` include added to `export_mints.cc` and `export_trans.cc`, this looks mergeable — the conversion is complete and every runtime pattern I could exercise behaves correctly. The `psi4Config.cmake.in` version floor and a plugin‑facing release note are the two things I'd most want alongside it.
>

TiborGY:
CI is probably only passing because unity builds are the default:

> 
> **CI passes because psi4's unity build masks the missing includes.**
> 
> Here's the full picture. psi4 builds `core` with `psi4_CMAKE_UNITY_BUILD` defaulting to **ON** (top-level `CMakeLists.txt` line 210, added recently as "#3388 turn on a basic unity build"), with `UNITY_BUILD_BATCH_SIZE 5` set on the `core` target, and no CI leg overrides it. A unity build concatenates sources into jumbo TUs in listed order, and the order in `psi4/src/CMakeLists.txt` happens to save the patch twice over:
> 
> -   Batch 1 is `core.cc, export_mints.cc, ...` — `core.cc` includes the patched `psi4/pybind11.h` (which now pulls in `native_enum.h`), so by the time the compiler reaches `export_mints.cc`'s code in the concatenated TU, `py::native_enum` is already declared.
> -   The batch containing `export_trans.cc` begins with `export_pcm.cc`, which also includes the shared header. Same masking. (I checked both batch sizes 5 and 8 — the masking holds either way.)
> 
> I verified this end-to-end: a standalone compile of an `export_mints.cc`-like file fails with `'native_enum' is not a member of 'py'`, while the same code concatenated CMake-unity-style after a `core.cc`-like file compiles cleanly. So green CI and my finding are both real — they're just testing different TU configurations.
> 
> So the corrected assessment: **not a blocking bug, but a genuine latent defect worth fixing in the PR anyway.** Concretely, `cmake -Dpsi4_CMAKE_UNITY_BUILD=OFF` breaks the build — and that's not an exotic configuration; psi4's own option help text advertises it ("For dev, can turn off overall or per-module"), since developers routinely disable unity for faster incremental rebuilds, better error locality, or IDE tooling. It's also fragile in dumber ways: reordering the `target_sources` list, changing the batch size, or a per-module unity opt-out would surface the error later with no obvious connection to this PR. And it violates include-what-you-use in exactly the spot pybind11's upgrade guide calls out — the header is not included automatically and must be added explicitly. [pybind11](https://pybind11.readthedocs.io/en/stable/changelog.html)
> 
> The fix is still the same two lines: add `#include <pybind11/native_enum.h>` (or `#include "psi4/pybind11.h"`) to `export_mints.cc` and `export_trans.cc`. I'd suggest it as a review comment on the draft rather than a blocker — something like "these two TUs only compile under the unity build because a batch-mate provides native_enum.h; please add the include so non-unity builds work."

PR #3414, opened 2026-05-27
Add TrexIO save/load Python interface

susilehtola:
## Summary
- Add `psi4.driver.trexio` (`save`, `load`, `load_wavefunction`) on top of the official `trexio` Python package.
- Covers geometry / nuclei / basis / AO metadata / electrons / MO coefficients / energies / occupations / spin labels, with optional AO and MO 1-electron integrals, AO ERIs (8-fold sparse), and CI determinant lists.
- Round-trip inside Psi4 is exact; AO ordering follows Psi4's native convention and is recorded in the `metadata_description` tag so consumers can re-permute if needed.

## Details
- New module `psi4/driver/trexio.py` (~590 lines). All TrexIO calls are routed through a small set of helpers, with a lazy `import trexio` so users without the package see a clear error.
- `save(wfn, path, ...)` accepts `save_ao_integrals`, `save_mo_integrals`, `save_eri`, `ci_vector`, `determinants`, `back_end`, `overwrite`, `description`.
- `load(path)` returns a dict keyed by group (`metadata`, `nucleus`, `basis`, `ao`, `electron`, `mo`, `ao_1e`, `mo_1e`, `ao_2e`, `determinant`); groups absent from the file are omitted.
- `load_wavefunction(path, reference=..., basis_name=...)` rebuilds a C1 `core.Wavefunction` using the same constructor that `Wavefunction.from_file` uses, populating Ca/Cb/Da/Db, epsilons, and the relevant dimension/int/string fields.
- AO ordering: Cartesian (`xx,xy,xz,yy,yz,zz`) and pure spherical (`m=0,+1c,+1s,+2c,+2s,...`). External consumers that expect a different layout (e.g. `m=-L…+L` for QMC=Chem) will need a permutation step; this is left to a follow-up.

## Build / packaging
- `psi4/driver/CMakeLists.txt` adds `trexio` to the cythonized-sources list so the module ships with the install.
- `pytest.ini` registers a `trexio` marker so tests skip cleanly when the package isn't installed.

## Test plan
New `tests/pytests/test_trexio.py` with 7 tests:
- [x] `test_save_roundtrip_rhf` — RHF wfn writes and reads back; coefficient magnitudes, energies, and occupations match Psi4 exactly.
- [x] `test_save_uhf` — UHF wfn produces `mo_type == "UHF"`, doubled MO block, correct spin labels.
- [x] `test_ao_one_electron_integrals` — saved S/T/V/h match `MintsHelper` to ≤1e-12.
- [x] `test_mo_one_electron_integrals` — saved MO overlap recovers identity to ≤1e-10.
- [x] `test_ao_eri_sparse` — reconstructing the full ERI tensor from the 8-fold-symmetric sparse list matches `MintsHelper.ao_eri` to ≤1e-12.
- [x] `test_overwrite_protection` — existing file raises `TrexIOError` unless `overwrite=True`.
- [x] `test_load_wavefunction` — rebuilt wavefunction has matching `nbf`, `natom`, and `epsilon_a`.

Tests probe for HDF5/TEXT TrexIO back ends at module load and skip if neither is available. All 7 pass locally against the staged install.

## Known follow-ups
- Optional canonical AO permutation for cross-tool interop.
- Reading basis exponents/coefficients back from TrexIO (so `load_wavefunction` can handle non-library bases).
- MO ERI writer (currently AO ERI only).


susilehtola:
pingg @scemama

PR #3415, opened 2026-05-27
Adds Einsums interfaces to MintsHelper and DFTensor

jturney:
Updates Psi4 to use an in-progress version of Einsums 2. Adds interfaces to MintsHelper and DFTensor to return Einsums RuntimeTensor objects.

## Status
- [ ] Ready for review
- [ ] Ready for merge

### DIFF
--- a/psi4/__init__.py
+++ b/psi4/__init__.py
@@ -93,7 +93,25 @@ atexit.register(core.finalize)
 from .driver import endorsed_plugins
 
 # Manage threads. Must be after endorsed plugins, honestly.
-core.set_num_threads(1, quiet=True)
+#
+# Importing psi4 sets the thread count for the whole PROCESS, not just for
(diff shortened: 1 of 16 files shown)


NathanGillispie:
This would be super useful for my Complex Wavefunction code. Are there any other types that you would recommend instead of TiledRuntimeTensor? I'm working with other people and I can't expect everyone to compile their own Einsums, so I was planning on temporarily converting these types to something else on my branch, at least until Einsums v2 hits conda.

NathanGillispie:
Okay I think I figured it out. The RuntimeTensor is so the python bindings work for arbitrary rank. I'll copy your changes but remove references to RuntimeTensor (not available in v1.1). I'll also make `so_*_einsums` integrals return `BlockTensor<double, 2>` because you can probably static\_cast into `TiledTensor` later if needed.

PR #3420, opened 2026-05-30
Remove unnecessary local copy of Dimension objects from cctriples/count_ijk

TiborGY:
## Description

PR #3356 introduced local mutable copies of the `moinfo.occpi`,  `moinfo.virtpi` obejcts and their open-shell counterparts into cctriples/count_ijk
This PR provides a better solution via `const&`.

- [x] No user-facing changes

## Dev notes & details

- [x] Replace declaring mutable dimension objects unconditionally and assigning them if needed, with `const&` in the smallest scope they are needed in. This restores the lack of copying before PR #3356.

## Checklist
- [x] No new features
- [x] CI tests are passing

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/cc/cctriples/count_ijk.cc
+++ b/psi4/src/psi4/cc/cctriples/count_ijk.cc
@@ -47,8 +47,6 @@ void count_ijk() {
     int Ga, Gb, Gc;
     int a, b, c;
     int A, B, C;
-    Dimension occpi, aoccpi, boccpi;
-    Dimension virtpi, avirtpi, bvirtpi;
     const int *occ_off, *aocc_off, *bocc_off;
     const int *vir_off, *avir_off, *bvir_off;
     int nijk;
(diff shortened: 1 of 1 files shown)


JonathonMisiewicz (dismissed):

JonathonMisiewicz:
I know this is nit-picking, but why not just use `auto`? We're not doing any type-conversion.

TiborGY:
I have considered it and in fact started with `auto` at first, but I think being explicit about the type is (very) slightly better here.

- The type name is not long. If this was a shared pointer to a tuple of std::chrono objects and std::filesystem paths I would certainly use auto.
- Auto makes it a bit opaque what type things are, this way it is obvious at a glance.
- Auto might increase the blast radius of (hypothetical) future changes to the `moinfo` object. With an explicit type here, if whatever is in `moinfo` isn't similar enough to a `Dimension` object for the `const&` to bind, there will be a compile error, with a hopefully informative message. Auto is more flexible, but that might not always be a good thing, eg. if some type changes from `int` to `size_t` (effectively int32_t --> uint64_t on most systems) we might not want that to spread around virally through auto. Who knows if there is some loop somewhere that occasionally does negative increments and is an underflow risk with uints...

TiborGY:
@JonathonMisiewicz When you have the time, could you take another look at this please?

psi4/src/psi4/cc/cctriples/count_ijk.cc
```diff
@@ -57,24 +55,25 @@ void count_ijk() {
...
-                            J = occ_off[Gj] + j;
-                            for (k = 0; k < occpi[Gk]; k++) {
-                                K = occ_off[Gk] + k;
-
-                                if (params.dertype == 1)
-                                    nijk++;
-                                else if (I >= J && J >= K)
-                                    nijk++;
+        {
+            const Dimension& occpi = moinfo.occpi;
```
JonathonMisiewicz: almost always auto

TiborGY: Well, I have made my arguments why I decided against auto in this particular case above, but fair enough I can change over to auto.

JonathonMisiewicz: Right - for me, the central point (which I thought I made earlier, oops) is that this _decreases_ the blast radius of future type changes. `auto` signals we do not want type casting. Implicit type casts increase the blast radius, which is part of the justification for [almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/).

TiborGY: Done.

JonathonMisiewicz (approved):

jturney (approved):
LGTM

PR #3422, opened 2026-06-02
Throw if Qmn_ is nullptr in DiskDFJK (in-core mode)

TiborGY:
## Description

DiskDFJK and its derived objects have some footguns, and currently improper usage results in segfaults instead of an informative exception. I ran into this with @andyj10224 while working on DLPNO.
This PR does not fix the underlying issue of these objects arguably being a bit unergonomic, but at least the error message will be a bit better than a segfault.

- [x] None, I think?

## Dev notes & details

- [x] `manage_JK_core()` and `compute_JK()` in DiskDFJK now throw if called when the object is in a state where calling these functions would result in a nullptr dereference. (DiskDFJK is in in-core mode, and `Qmn_` is not a valid `shared_ptr` holding a ptr to a `Matrix`)
- [x] This PR also affects classes derived from DiskDFJK that do not override both of these functions, such as CDJK (only overrides  `manage_JK_core()` )

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/src/psi4/libfock/DiskDFJK.cc
+++ b/psi4/src/psi4/libfock/DiskDFJK.cc
@@ -1661,6 +1661,9 @@ void DiskDFJK::rebuild_wK_disk() {
     // No need to close
 }
 void DiskDFJK::manage_JK_core() {
+    if (!Qmn_ || Qmn_.use_count() == 0)
+        throw PSIEXCEPTION(
+            "DiskDFJK(in-core mode) tried to manage_JK_core with a Qmn_ that does not point to a Matrix object!");
     for (int Q = 0; Q < auxiliary_->nbf(); Q += max_rows_) {
         int naux = (auxiliary_->nbf() - Q <= max_rows_ ? auxiliary_->nbf() - Q : max_rows_);
         if (do_J_) {
--- a/psi4/src/psi4/libfock/DiskDFJK.cc
+++ b/psi4/src/psi4/libfock/DiskDFJK.cc
@@ -504,6 +504,7 @@ void DiskDFJK::compute_JK() {
         }
     }
 }
+
 void DiskDFJK::postiterations() {
     Qmn_.reset();
     Qlmn_.reset();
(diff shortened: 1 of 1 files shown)


TiborGY:
CI flakiness was due to PubChem servers being overloaded, this is now ready for review.

loriab (approved):
Thanks for avoiding a segfault, especially affecting DLPNO.

TiborGY:
I am not sure if it can be triggered with the current master, but with full-Q and future DLPNO methods in development there is currently a sharp edge where only a few integral types are supported and since the JK infra is not hardened against holding it wrong, segfaults can result if the user requests an unsupported integral type. This will be smoothed over later when we get to upstreaming it, but the JK objects themselves can also be made more ergonomic.

PS. This PR closes one of those pitfalls.

psi4/src/psi4/libfock/DiskDFJK.cc
```diff
@@ -469,19 +469,23 @@ void DiskDFJK::preiterations() {
...
     zero();
-
     max_nocc_ = max_nocc();
     max_rows_ = max_rows();
 
     if (do_J_ || do_K_) {
         initialize_temps();
-        if (is_core_)
+        if (is_core_) {
+            if (!Qmn_ || Qmn_.use_count() == 0) {
```
jturney, line 480: Are these checks not equivalent?

TiborGY: Almost but not completely.

> An empty shared_ptr (where use_count() == 0) may store a non-null pointer accessible by get(), e.g. if it were created using the aliasing constructor. 


[https://en.cppreference.com/cpp/memory/shared_ptr/operator_bool](https://en.cppreference.com/cpp/memory/shared_ptr/operator_bool)

This is not an issue currently in the Psi4 codebase as far as I can tell, but it is not impossible to produce a shared_ptr that is holding a non-null ptr, but with `use_count() == 0`. Writing the check this way remains robust against that.

jturney: Gotcha. Sounds good :)

jturney (approved):

PR #3423, opened 2026-06-03
Dont predicate throw in CDJK on printlevel

TiborGY:
## Description

I noticed that currently the check that throws in CDJK if it is used with "wK tasked" (which is not supported in CDJK) is disarmed if the CDJK print level is zero. A sanity check like this should not be predicated on a print verbosity setting, this PR makes it unconditional.

- [x] Some improper usages of Cholesky decomposition methods in unimplemented ways now throw an error even if the output print verbosity is set to minimal

## Dev notes & details

- [x] Check if `do_wK_ == true` in `CDJK::print_header` and throw even if `CDJK::print_ == 0`

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge


loriab (approved):

TiborGY:
This should be good to go now.

JonathonMisiewicz (approved):

PR #3424, opened 2026-06-08
Adapt for QCFractal v0.65

loriab:
## Dev notes & details

- [x] Not sure the exact mechanism of difference between qcfractal 0.64 and 0.65 (since I can see this error with psi4 src fixed, adcc pkg fixed, and qcel/qcng fixed ~0.50), but importing adcc under certain circumstances (snowflake) fails, causing the qcengine records to return as errors and the psi4 logic to fail.
- AI suggested some better error handling but going to defer that to #3034 . For now, just want to get CI working again.

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/endorsed_plugins.py
+++ b/psi4/driver/endorsed_plugins.py
@@ -63,6 +63,10 @@ try:
     import adcc
 except ImportError:
     pass
+except Exception:
+    # Some adcc versions evaluate sys.stdout at import time and can fail in
+    # redirected/headless worker contexts (e.g., QCFractal compute workers).
+    pass
 
 try:
     import psi4fockci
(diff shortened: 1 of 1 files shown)


psi4/driver/endorsed_plugins.py
```diff
@@ -63,6 +63,10 @@
     import adcc
 except ImportError:
     pass
+except Exception:
+    # Some adcc versions evaluate sys.stdout at import time and can fail in
+    # redirected/headless worker contexts (e.g., QCFractal compute workers).
+    pass
```
loriab:
```suggestion
except AttributeError as e:
    if "'NoneType' object has no attribute 'isatty'" in str(e):
        # Some adcc versions evaluate sys.stdout at import time and can fail in
        # redirected/headless worker contexts (e.g., QCFractal compute workers).
        # (originates from adcc/timings.py default arg evaluation during import adcc)
        pass
    else:
        raise
```

jturney (approved):

JonathonMisiewicz (approved):

Issue #54, opened 2014-03-20
Case where df-hf fails in sapt


CDSherrill:
From Daniel Schofield, 2/7/14: "I have been attempting to run SAPT calculations in PSI4, but am having little success due to my inability to converge the df-hf step for one of my monomers (H2).  I think this is due to a combination of using density fitting and the size of the ghosted basis set relative to this monomer i.e. if I run in a monomer centered basis set the calculation converges and if I run direct SCF (outside of SAPT) the calculation converges."

I have the test input/output files.  Apparently the DF-HF is going non-variational on the dimer computation.

dgasmith:
Can we get an input file where this fails?

I dont think we have seen this in quite some time, there was an issue where the molecules would be re-centered so that the DF integrals and one-electron integrals didnt have the same center which caused some issues. This was fixed however.

dgasmith:
Closing this out, if someone has an input where this fails please post a new issue with the reproducible error.

PR #3427, opened 2026-06-10
OpenTrustRegion as an alternate second-order SCF optimizer

loriab:
## Description

Interfaces [OpenTrustRegion](https://github.com/eriksen-lab/opentrustregion) (OTR), a
black-box second-order orbital optimizer by J. Greiner, as an alternative to psi's
internal SOSCF code. OTR does not replace the SCF loop: first-order iterations run as
usual until the orbital gradient reaches `SOSCF_START_CONVERGENCE`, then OTR takes the
handoff and carries the calculation to convergence.

Selecting it splits the existing keyword in two. `ORBITAL_OPTIMIZER_PACKAGE`
(`INTERNAL|OOO`) drives the first-order iterations; a new
`SECOND_ORDER_ORBITAL_OPTIMIZER_PACKAGE` (`INTERNAL|OTR`) drives the second-order ones
and is consulted only when `SOSCF` is on. The two compose freely.

## Questions
- [ ] **Should the occupation-restart loop stay?** OpenTrustRegion optimizes at a fixed
      occupation, so `_run_opentrustregion` checks whether the aufbau rule still agrees with the
      occupation it converged, and restarts at the new one if not. That was written when
      OpenTrustRegion drove the whole SCF from a cold guess and could settle on the guess's
      occupation. Under the handoff the first-order iterations fix the occupation before it ever
      runs, and I could not produce a disagreement: 8 near-degenerate systems (singlet and
      stretched O2, C2, B2, Be, NO, NO+, CN), two bases, core and gwh guesses, and
      `SOSCF_START_CONVERGENCE` raised to force the handoff at iteration 3 -- no hits. The
      branch is exercised only by an injection test (which is awful-looking). Three ways to go:

      * *Keep it.* ~35 lines that recover automatically if a real case exists. Cost: the only
        coverage is injected (and I think should drop that test), and the block went through several revisions before it was right.
      * *Detect and warn.* Keep the check, drop the restart: report the disagreement and return
        the converged solution. Honest, much less code, no automatic recovery.
      * *Remove it.* Simplest, and it also retires `_aufbau_occupation`, the ~8 lines of python
        mirroring the aufbau rule in `HF::find_occupation` that are flagged below as a drift
        risk. Cost: a genuine disagreement would pass unnoticed.

- [x] New `SECOND_ORDER_ORBITAL_OPTIMIZER_PACKAGE` selects the second-order SCF optimizer,
      `INTERNAL` or `OTR`. `ORBITAL_OPTIMIZER_PACKAGE` covers only the first-order
      iterations is still `INTERNAL|OOO`.
- [x] Nine `OTR_*` keywords expose OpenTrustRegion settings with no internal counterpart
      (`OTR_SUBSYSTEM_SOLVER`, `OTR_N_RANDOM_TRIAL_VECTORS`, `OTR_JACOBI_DAVIDSON_START`,
      `OTR_LINE_SEARCH`, `OTR_START_TRUST_RADIUS`, `OTR_GLOBAL_RED_FACTOR`,
      `OTR_LOCAL_RED_FACTOR`, `OTR_SEED`, `OTR_PRINT`). Everything with a genuine
      counterpart reuses the existing keyword.
- [x] `SOSCF` now raises rather than misbehaving for a CUHF reference, meta-GGA and VV10
      functionals, fractional occupation, and incremental Fock builds. Fractional
      occupation previously returned a converged energy 1.5 Eh off; the others died in C++
      or were unreachable. None of this is specific to OTR.
- [x] A line above the iteration table names the optimizers actually in force.

## Dev notes & details
- [x] Four conditions under which second-order convergence silently misbehaved now raise.
  These are pre-existing bugs in the internal SOSCF path, surfaced by testing it harder.
- [x] OTR is engaged from the existing SOSCF trigger in `scf_iterate`; the run block is
      factored into `_run_opentrustregion`, which owns the remaining iterations and returns.
      OOO's convergence callback stops at `SOSCF_START_CONVERGENCE` when `SOSCF` is on and
      the driver falls through, so `SOSCF` is no longer an OOO incompatibility.
- [x] `conv_tol` takes the tighter of `E_CONVERGENCE`/`D_CONVERGENCE` and two more digits.
      This is to make sure convergence gets tight enough for Psi4 convcrit. Costs nothing measurable and avoids macro-iterations
      that stop short. `n_micro` is deliberately unmapped: `SOSCF_MAX_ITER` defaults to 5
      against OTR's 50.
- [x] Guards for CUHF and meta-GGA/VV10 sit in `scf_iterate` rather than beside the
      existing `check_non_symmetric_jk_density` in `proc.py`, because they need
      `self.functional()`, which does not exist at that point.
- [x] `module_roles_` is carried through shallow/deep copies and round-trips through
      `to_file`/`from_file` under a `module_role:` prefix, the file format holding a flat
      string map. 13 direct `module_ = "..."` assignments become `set_module(...)`.
- [x] Test parametrization runs the optimizer packages as a grid: a six-way for tests where
      second order is reachable, a four-way for tests that turn `SOSCF` on themselves, and
      a two-way where the method rules second order out. There are enough cases for a decent check even w/o redefaulting to OOO or OTR. The
      assertions check `module_roles()`, so a silent fallback to the internal optimizer
      fails rather than passing quietly.
- [x] One upstream OTR fix came out of this (eriksen-lab/opentrustregion#47); distinguishable
      solver error codes came from #54 and are mapped so ordinary non-convergence is subject
      to `FAIL_ON_MAXITER` while a failed solver step is not.
- [x] `Wavefunction::module_` is generalized into a map keyed by role, so a caller can ask
  which package actually optimized the orbitals rather than which one was requested.
  This is on branch `wfn_module_roles` as a standalone, no-behavior-change PR. (#3541 ; start of rework into better recording code path that I mentioned on slack)
- [ ] Requires an OpenTrustRegion carrying both the 102/202 error codes. This is why one test is presently failing. The `find_package` floor still needs setting.

## AI
- [ ] After Jonas's work (#3320), Claude Code did most of the implementation, test triage and docs under close direction:
      the handoff restructuring, keyword mapping, the four guards, the parametrization grid,
      and the upstream OTR bug reproducers. Notably it diagnosed the J/K aliasing bug behind
      ~30 early failures and the fractional-occupation wrong answer.

## Checklist

## Status
- [x] Ready for review


susilehtola:
Any progress on this PR?

loriab:
Ok, I'm happy for this to be reviewed. I don't care ultimately if it's merged from this PR or the original #3320 @jonas-greiner

PR #3426, opened 2026-06-09
libiwl: phase 1 modernization (collapse dup, overflow check, RAII)

susilehtola:
Phase 1 of the libiwl modernization tracked in #3425. An internal refactor of `psi4/src/psi4/libiwl/` with **no caller-facing API change and no on-disk format change**, plus a standalone round-trip test for the bucket I/O. The ~30 in-tree consumers that poke `struct iwlbuf` fields directly all keep compiling unchanged.

## Description
The libiwl bucket reader/writer had two byte-identical bodies for every operation (a C `iwl_buf_*` free function and a C++ `IWL::` method), C-style macros, mixed `malloc`/`new[]` ownership, and silent 32-bit→16-bit narrowing at the `Label` cast that could corrupt a bucket file without warning. This PR consolidates the duplication, fixes the latent correctness/ownership issues, and adds a round-trip test — all without touching the public field layout or the on-disk format.

## User API & Changelog headlines
- [x] `Label` overflow (orbital index exceeding the 16-bit bucket field) now throws a `PSIEXCEPTION` naming the offending indices instead of silently narrowing and corrupting the IWL file.

## Dev notes & details
- `config.h`: replace `#define IWL_INTS_PER_BUF` / `IWL_KEY_*` with `inline constexpr`; document the bucket layout.
- `iwl.hpp`: remove dead declarations `IWL::read` / `IWL::to_end` (never defined); delete copy ctor / assign; document the deferred-init pattern used by `libtrans`.
- Collapse every duplicated C-function / C++-method body into a single shared static helper per operation (one implementation site instead of two byte-identical ones). New internal `iwl_impl.h` holds the shared `check_label_fits()` / `make_printer()` helpers.
- **Overflow check at every `(int) → Label` cast** in the three write paths (throws instead of silently narrowing).
- `buf_init`: throw on `tocscan` failure instead of `Printf`'ing and leaving a half-initialized buffer the caller then `fetch()`es from.
- Unify allocators on `new[]`/`delete[]` (removes the cross-API `malloc`/`free` vs `new[]`/`delete[]` ownership UB).
- Hoist per-integral `PsiOutStream` construction in the write paths to one per-call instance, only when `printflag` is set (previously one file open per nonzero integral).
- Default-construct all `IWL` members including `keep_`; drop `MAX0`/`MIN0`/`INDEX` macros in `buf_wrt_mat.cc` for `std::max`/`std::min` and an inline `packed_index()`.

[...]

Issue #3425, opened 2026-06-09
libiwl: modernization assessment and three-phase refactor plan


susilehtola:
## Summary

`psi4/src/psi4/libiwl/` is the Psi3-derived "Integrals With Labels" library: a sparse `(p,q,r,s,value)` stream serialized as fixed-size buckets via libpsio. It is small (~1,200 LOC across 10 `.cc` + 3 headers) but 60+ source files across 14 modules depend on it, so any change has a wide blast radius. The library carries a number of correctness, safety, and code-health issues from its 1995 origin, including its own README explicitly warning that "any mistake in this area will be nearly impossible to track down later." This issue catalogs the defects and proposes a three-phase refactor that can land incrementally without breaking the on-disk format until the final phase.

The library exposes two parallel APIs over the same on-disk layout:

* **C-style**: `struct iwlbuf` + free functions `iwl_buf_init` / `iwl_buf_fetch` / `iwl_buf_put` / `iwl_buf_flush` / `iwl_buf_wrt_val` / `iwl_buf_close` / `iwl_rdone` (malloc/free).
* **C++ class**: `class IWL` with `init`/`fetch`/`put`/`flush`/`write`/`write_matrix`/`write_value`/`close`, `static read_one`/`write_one` (new/delete).

Every `buf_*.cc` file contains both implementations side-by-side, byte-identical apart from `Buf->` vs `this->`.

On-disk bucket layout (fixed `IWL_INTS_PER_BUF = 2980`):

```
[int lastbuf][int inbuf][Label labels[4*2980]][Value values[2980]]
```

with `Label = short int` and `Value = double` — ~28 kB per bucket.

### Correctness / safety

1. **`Label = short int` → orbital index hard-capped at 32767.** `config.h:34`. Every module that stores integrals through IWL silently truncates if `nso > 32767`. There is no overflow check; `(Label)p` is a plain narrowing cast in `buf_wrt.cc:75-78`, `buf_wrt_mat.cc:99-102`, `buf_wrt_val.cc:54-62`.

2. **The reader relies on a sign-bit hack** that is not documented anywhere in libiwl. `cc/ccenergy/fock_build.cc:77` reads back `std::abs((int)lblptr[idx++])` because psimrcc writes negative `p` as a sentinel. Invisible contract.

[...]

PR #3428, opened 2026-06-10
libiwl: modernization — IWLReader/IWLWriter, internals off the C API (extends #3426)

susilehtola:
## Summary

This is the **full libiwl modernization** for #3425, and it **supersedes / contains #3426** (the phase-1-only PR). This branch is stacked on top of the phase-1 work, so its diff includes everything in #3426 plus phases 2–3 of the cleanup. If you'd rather review the whole thing in one place, you can review this PR and close #3426; if you'd prefer to land it incrementally, #3426 is the first slice.

`libiwl` is the Psi3-derived "Integrals With Labels" library. Net effect of this branch: **−185 lines**, a single modern C++ API, the dead C API removed, and every consumer migrated except one documented special case.

## What's in here (commit by commit)

1. **Phase 1 — internal hardening** (`= #3426`): collapse the duplicated C-function/C++-method bodies to one implementation each; RAII (`IWL` throws on init failure instead of returning a half-open buffer; move ctor/assign); a 16-bit-`Label` overflow check that throws instead of silently truncating an orbital index; `inline constexpr` constants; remove dead `IWL::read`/`to_end` decls; one `PsiOutStream` per write call instead of one per integral.
2. **Standalone test** (`libiwl/test/`): a round-trip exerciser covering empty / 1 / `B-1` / `B` / `B+1` / `3.5·B` integrals and the overflow throw, cross-checking the low-level `IWL` accessors, `IWLReader`, and `IWLWriter` against the source data. (Disabled-by-default link via a small stubs TU; see the commit for why.)
3. **`IWLReader`** — a range-for input iterator (`for (const auto& I : reader) { I.p, I.q, I.r, I.s, I.value; }`). Labels are returned exactly as stored, so the psimrcc sign-of-`p` sentinel is preserved (callers apply `std::abs` as before).
4. **`IWLWriter`** — RAII writer; flushes the last buffer and closes on destruction, with an explicit idempotent `close()` for the paths that reopen the same psio unit. `write()` reuses the validated `write_value` path (cutoff + overflow check).
5. **Consumer fan-out** — every `iwl_buf_*` C-API call site migrated: sequential readers (`ccenergy/fock_build`, `dct/*`, `libtrans/integraltransform_sort_mo_tpdm`, `occ/{tei_sort_iabc,gfock}`, `mcscf/scf_read_so_tei`, `psimrcc/transform_presort`, `libfock/DiskJK`), helper-threaded readers (`ccenergy/{AO_contribute,BT2_AO}`, `cclambda/BL2_AO`, `fnocc/sortintegrals`+`mp2`), and the writer/dump pipelines (`cc/ccdensity/*`, `dct/dct_gradient_*`, `occ/coord_grad`, `libdpd::DPD::buf4_dump`).
6. **Delete the dead C API** — `struct iwlbuf`, `iwl_buf_*`, `iwl_rdone`, and `iwl.h` are gone (the last `iwl_rdone` one-electron reads moved to `IWL::read_one`); the ~60 `#include "iwl.h"` sites were repointed to `iwl.hpp` or dropped where stale.
7. **`IWLWriter` by reference** — non-owning, never-null `IWLWriter*` parameters are now `IWLWriter&` (the owned per-bucket arrays in `idx_permute`/`file_build` stay `std::vector<std::unique_ptr<IWLWriter>>`).

## One documented deferral

`libfock/PKmanagers.cc` is intentionally left on the `IWL` class. Its PK-supermatrix reader consumes a **multi-stream** IWL file (the Yoshimine presort writes each batch terminated by its own `lastbuf` marker, and the loop finalizes a batch on each `lastbuf` then keeps fetching). A single-pass `IWLReader` stops at the first `lastbuf`, so it is not a drop-in — this wants either a "continue past `lastbuf`" `IWLReader` mode or a small dedicated batch-stream reader, plus regression coverage, in a follow-up. Its writers are the separate `IWLAsync_PK` async path.

## Testing

Full `core` links cleanly after every commit, and the standalone IWL round-trip test passes. I could **not** run the CTest/pytest regression suites locally (this box has a Python 3.14 / pydantic-v1 / qcelemental import-time failure that aborts before any psi4 C++ runs, independent of these changes). So per-call-site runtime equivalence rests on CI here — each migration is a behavior-preserving 1:1 transform validated by the reader/writer round-trip test, but a green CI run is the real gate before merge. Flagging that explicitly.

Closes nothing automatically; see #3425 for the plan and #3426 for the phase-1 slice.

### DIFF
--- a/psi4/src/psi4/cc/ccdensity/add_core_ROHF.cc
+++ b/psi4/src/psi4/cc/ccdensity/add_core_ROHF.cc
@@ -33,6 +33,7 @@
 #include <cstdio>
 #include "psi4/libciomr/libciomr.h"
 #include "psi4/libiwl/iwl.h"
+#include "psi4/libiwl/iwl_writer.h"
 #include "MOInfo.h"
 #include "Params.h"
 #include "Frozen.h"
@@ -41,7 +42,7 @@
 namespace psi {
 namespace ccdensity {
 
-void add_core_ROHF(struct iwlbuf *OutBuf) {
+void add_core_ROHF(IWLWriter *OutBuf) {
     const auto actpi = moinfo.occpi + moinfo.virtpi - moinfo.openpi;
     int mo_offset = 0;
 
@@ -53,8 +54,8 @@ void add_core_ROHF(struct iwlbuf *OutBuf) {
                 double p_qt = moinfo.pitzer2qt[p + mo_offset];
                 double q_qt = moinfo.pitzer2qt[q + mo_offset];
                 for (int m = 0; m < moinfo.nfzc; m++) {
-                    iwl_buf_wrt_val(OutBuf, p_qt, q_qt, m, m, value, 0, "outfile", 0);
-                    iwl_buf_wrt_val(OutBuf, p_qt, m, m, q_qt, -0.5 * value, 0, "outfile", 0);
+                    OutBuf->write(p_qt, q_qt, m, m, value);
+                    OutBuf->write(p_qt, m, m, q_qt, -0.5 * value);
                 }
             }
         }
--- a/psi4/src/psi4/cc/ccdensity/add_core_ROHF.cc
+++ b/psi4/src/psi4/cc/ccdensity/add_core_ROHF.cc
@@ -32,7 +32,6 @@
 */
 #include <cstdio>
 #include "psi4/libciomr/libciomr.h"
-#include "psi4/libiwl/iwl.h"
 #include "psi4/libiwl/iwl_writer.h"
 #include "MOInfo.h"
 #include "Params.h"
@@ -42,7 +41,7 @@
 namespace psi {
 namespace ccdensity {
 
-void add_core_ROHF(IWLWriter *OutBuf) {
+void add_core_ROHF(IWLWriter &OutBuf) {
     const auto actpi = moinfo.occpi + moinfo.virtpi - moinfo.openpi;
     int mo_offset = 0;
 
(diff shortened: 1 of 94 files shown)


susilehtola:
**CI fix (force-pushed):** the ecosystem build failed because `v2rdm_casscf` includes `<psi4/libiwl/iwl.h>` and uses the `struct iwlbuf` / `iwl_buf_*` C API directly. That API is **public plugin API, not dead code**, so removing it was wrong. I've restored it: `iwl.h` is now byte-identical to master and the C functions are retained.

The modernization itself is unchanged in intent — psi4's own code is fully migrated onto the `IWL` class / `IWLReader` / `IWLWriter`, and thanks to the phase-1 dedup the retained C entry points share their implementation with the class methods (no logic duplication). The last commit was amended accordingly (title/message updated; `iwl.h` + the C function bodies kept).

Net: internal code modernized, public C API preserved for downstream plugins.

jturney:
Nice. I'm liking this clean up :)

TiborGY:
Sounds good, but this PR is huge. I think smaller and more self-contained PRs would be easier to review.

susilehtola:
Closing in favor of splitting this into smaller, independently reviewable PRs (it had grown to ~94 files). The phase-1 foundation #3426 is retained, and the rest is now a three-PR stack on top of it:

- #3439 — **IWLReader / IWLWriter API** + standalone round-trip test (no consumer changes)
- #3440 — migrate **read-side** consumers to `IWLReader`
- #3441 — migrate **write-side** consumers to `IWLWriter` + stale-`#include` cleanup

Stack order: #3426 → #3439 → #3440 → #3441. The union is byte-identical to this PR's final state, rebased onto current `master`; each piece builds independently. No content lost.

Issue #3429, opened 2026-06-10
libpsio: modernization assessment and phased cleanup plan


susilehtola:
## Summary

`psi4/src/psi4/libpsio/` is the Psi3-derived disk-I/O layer (≈4,300 LOC across **40 one-function-per-file `.cc` files** + 4 headers). It's on the critical path for essentially every Psi4 run, and it carries a lot of pre-C++11 idiom. This issue catalogs the cleanup opportunities and proposes a phased modernization, in the same spirit as the libiwl work in #3425.

A "unit" (file) is striped across up to 8 "volumes"; data is addressed by `(page, offset)`, indexed by a hand-rolled linked-list table of contents, and read/written through raw POSIX `::open/::read/::write/::lseek`.

## Antiquated aspects

1. **Dual API.** A C free-function surface (`psio_open/close/read/write/...` in `psio.h`) sits alongside `class PSIO` / `PSIOManager` (`psio.hpp`). The C functions are mostly thin delegators (`psio_read` → `_default_psio_lib_->read(...)`), so there's little duplicated *logic*, but it's a redundant parallel API. A few `psio_*` entries are genuine standalone helpers (`psio_volseek`, `psio_get_address`, `psio_error`, `psio_compose_err_msg`, `decode_errno`).
2. **C memory management throughout** — `malloc`/`free`/`strdup`/`sprintf`, `char* path`, `psio_unit = (psio_ud*)malloc(sizeof(psio_ud)*PSIO_MAXUNIT)`. `psio_ud`/`psio_vol` are C structs holding raw `int` file descriptors and `char*` paths.
3. **Hand-rolled linked-list TOC** (`psio_tocentry` with manual `next`/`last` surgery across `tocwrite/tocread/tocdel/tocclean/toclast`) — wants `std::list`/`std::vector`.
4. **Fixed arbitrary caps**: `PSIO_MAXUNIT 500` (a flat malloc'd array) and `PSIO_MAXVOL 8`, both `#define`. A `std::map<size_t, …>` removes the unit cap.
5. **Error handling**: 21 `#define PSIO_ERROR_*` codes + an ~80-line `switch` in `error.cc`, plus scattered `fprintf(stderr, …)` right before each `psio_error` call. It already throws `PSIEXCEPTION` (good); the codes should be an `enum class` + message table.
[...]

PR #3430, opened 2026-06-10
libpsio: modernize config.h constants + add round-trip test (#3429)

susilehtola:
## Description

Non-breaking internal modernization of `libpsio`, plus the behavior gate the library never had. First slice of the plan in #3429.

Two things only: `config.h`'s preprocessor macros become `inline constexpr int`, and a standalone round-trip test is added and wired into `ctest`.

An earlier revision of this PR also removed the C `psio_*` free-function API. **That has been dropped.** As the analogous libiwl PR (#3428) showed, those free functions are public API that downstream plugins call directly (e.g. `v2rdm_casscf`), so removing them breaks the ecosystem build; and unlike libiwl's range-for `IWLReader`, they are no worse than the verbose `_default_psio_lib_->...`, so there was little upside to forcing the migration. Any future C-API change needs to be coordinated with downstream plugin updates rather than done unilaterally.

The deeper internal rewrites from #3429 — `AIOHandler` single job-queue, `psio_ud`/TOC/`char*`-path/fd-RAII — remain follow-ups that want this test as a safety net first. The multi-volume striping removal is #3452, stacked on this PR.

- [ ] No user-visible change. `psio_error()`'s public signature is unchanged (error codes stay `int` rather than becoming an `enum class`), so there is no downstream impact.

## Dev notes & details
- [ ] `config.h`: `#define` → `inline constexpr int` for `PSIO_KEYLEN`, `PSIO_MAXUNIT`, `PSIO_PAGELEN`, `PSIO_MAXVOL` and the `PSIO_ERROR_*` codes. Same names, same values. Gives the constants a type, a scope, and debugger visibility.
- [ ] New `psi4/src/psi4/libpsio/test/`: a standalone round-trip exerciser. Writes then reads back four entries whose sizes deliberately straddle the branches of `PSIO::rw` — 64 B (sub-page partial), exactly `PSIO_PAGELEN` (one full page, no remainder), `3 * PSIO_PAGELEN + 137` (multiple full pages plus a trailing partial), and 8 B.
- [ ] The test links a curated subset of static archives (`psio psi4util options ciomr`) plus a small `stubs.cc` supplying the runtime globals normally provided by `core.cc`. No Python, no libdpd, no libmints/Libint closure.
- [ ] Not built on Windows. The curated link line resolves under GNU/Clang but not clang-cl/lld-link, which emits inline and implicit functions that GCC discards as unused: `psio.lib` ends up referencing `DPDMOSpace::~DPDMOSpace` (via `std::vector<DPDMOSpace>` in `dpd.h`), and `psi4util.lib` references `Molecule::Molecule(const Molecule&)` (inline `Molecule::clone`, `molecule.h:200`) and `ExternalPotential::print` (inline `py_print`, `extern.h:131`). Not a unity or PCH artifact — Linux is unity-built too and its archives carry no such references, and Azure Windows builds with PCH off. (An earlier revision of this description blamed the unity build; that was wrong.)
- [ ] Labelled `"psi;quicktests;smoketests;libpsio"`. `quicktests` alone only reached the GH macOS lane; the Azure lanes drive ctest by `-L "plug|smoke"` (`devtools/scripts/ci_run_test.py`) and `--label-regex smoke`. Verified the test is selected by all three selectors.
- [ ] Fixed the `DPD` stub to declare `file4_cache_del_filenum(std::size_t)`, matching `dpd.h:483`. It previously said `unsigned long`, which mangles differently on Windows x64 — that accounted for one of the four undefined symbols above.
- [ ] Registration lives in the superbuild (`tests/libpsio/`), not next to the executable. `psi4-core` is configured as a separate `ExternalProject` with its own binary directory and never calls `enable_testing()`, so an `add_test()` under `psi4/src/` generates no `CTestTestfile.cmake` at all — and CI invokes `ctest` from the superbuild directory, which cannot descend into the core build tree in any case. Resolving the ExternalProject's `BINARY_DIR` from `tests/` makes `ctest -L quick` actually execute it.

## Questions
- [ ] `tests/CMakeLists.txt` gains an `add_subdirectory(libpsio)` under "Add internal projects". #3426 adds `add_subdirectory(libiwl)` at the same spot, so whichever of the two lands second will need a one-line conflict resolution. Happy to introduce a shared `add_core_unit_test()` macro in `cmake/TestingMacros.cmake` instead if you would rather have one entry point for C++ unit tests.
- [ ] Is skipping the unit tests on Windows acceptable, or would you prefer the executables link against the full closure so they run everywhere?

## AI
- [ ] Claude (Opus 4.8) assisted with the mechanical `#define` → `inline constexpr` conversion, drafting the round-trip test and its stubs, and diagnosing the ctest-registration and Windows-link problems. Reviewed and submitted by me; I am happy to answer questions about any of it.

## Checklist
- [x] Tests added for any new features — this PR *is* the test
- [x] Docstring or narrative docs/sphinxman/source updated if needed — n/a, internal only
- Verified locally: `ctest -R libpsio-roundtrip` passes, and reports failure when the binary is removed. Before the registration fix, `ctest -N -R psio_roundtrip` listed no tests at all.

## Status
- [x] Ready for review

---

Thanks @TiborGY — you were right on both counts. The description was stale, and the ctest never ran.

### DIFF
--- a/psi4/src/psi4/libpsio/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/CMakeLists.txt
@@ -34,6 +34,8 @@ list(APPEND sources
   )
 psi4_add_module(lib psio sources)
 
+add_subdirectory(test)
+
 if (psi4_ENABLE_PRECOMPILE_HEADERS)
     # Clearly from the paths, libpsio is a weird place to be defining the pch. However, if one sensibly
     #   puts it in mints, CMake creates a dependency of, say, fnocc on mints, and since mints is the
--- a/psi4/src/psi4/libpsio/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/CMakeLists.txt
@@ -34,11 +34,16 @@ list(APPEND sources
   )
 psi4_add_module(lib psio sources)
 
-# The standalone round-trip test links a curated subset of static archives
-# plus a small stubs TU. That works with GNU/Clang on Linux/macOS but not with
-# Windows lld-link, where the unity+PCH build pulls in extra symbol references
-# (Molecule, DPDMOSpace, ...) the curated link line doesn't provide. Restrict
-# it to non-Windows rather than drag in the full dependency closure.
+# The standalone round-trip test links a curated subset of static archives plus a
+# small stubs TU. That resolves with GNU/Clang on Linux/macOS but not with
+# clang-cl/lld-link, which emits inline and implicit functions that GCC discards
+# as unused, so the Windows objects carry undefined references the curated link
+# line does not provide: psio.lib wants DPDMOSpace::~DPDMOSpace (instantiated by
+# std::vector<DPDMOSpace> in dpd.h, reached via close.cc), and psi4util.lib wants
+# Molecule::Molecule(const Molecule&) (from inline Molecule::clone) and
+# ExternalPotential::print (from inline py_print). Not a unity or PCH artifact --
+# Linux is unity-built too, and Azure Windows builds with PCH off. Restrict the
+# test to non-Windows rather than drag in the full libmints/Libint closure.
 if(NOT WIN32)
     add_subdirectory(test)
 endif()
--- a/psi4/src/psi4/libpsio/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/CMakeLists.txt
@@ -34,7 +34,14 @@ list(APPEND sources
   )
 psi4_add_module(lib psio sources)
 
-add_subdirectory(test)
+# The standalone round-trip test links a curated subset of static archives
+# plus a small stubs TU. That works with GNU/Clang on Linux/macOS but not with
+# Windows lld-link, where the unity+PCH build pulls in extra symbol references
+# (Molecule, DPDMOSpace, ...) the curated link line doesn't provide. Restrict
+# it to non-Windows rather than drag in the full dependency closure.
+if(NOT WIN32)
+    add_subdirectory(test)
+endif()
 
 if (psi4_ENABLE_PRECOMPILE_HEADERS)
     # Clearly from the paths, libpsio is a weird place to be defining the pch. However, if one sensibly
--- a/psi4/src/psi4/libpsio/test/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/test/CMakeLists.txt
@@ -2,13 +2,15 @@
 # liboptions/libciomr static archives plus a tiny stubs TU (test/stubs.cc) for
 # the runtime globals normally provided by core.cc, so the test is
 # self-contained (no python, no libdpd).
+#
+# Only the executable is defined here. psi4-core is configured as a separate
+# ExternalProject with its own binary dir and never calls enable_testing(), so
+# an add_test() in this scope would generate no CTestTestfile.cmake and would
+# be invisible to the `ctest` that CI runs from the superbuild directory.
+# Registration therefore lives in the superbuild, in tests/libpsio/.
 add_executable(psio_roundtrip_test.x psio_roundtrip_test.cc stubs.cc)
 target_link_libraries(psio_roundtrip_test.x PRIVATE psio psi4util options ciomr)
 target_include_directories(psio_roundtrip_test.x PRIVATE
     ${CMAKE_SOURCE_DIR}/src
     ${CMAKE_BINARY_DIR}/src
 )
-add_test(NAME psio_roundtrip
-         COMMAND psio_roundtrip_test.x
-         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-set_tests_properties(psio_roundtrip PROPERTIES LABELS "libpsio;quicktests")
--- a/psi4/src/psi4/libpsio/test/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/test/CMakeLists.txt
@@ -8,7 +8,3 @@
 # ExternalProject that never calls enable_testing()).
 add_executable(psio_roundtrip_test.x psio_roundtrip_test.cc stubs.cc)
 target_link_libraries(psio_roundtrip_test.x PRIVATE psio psi4util options ciomr)
-target_include_directories(psio_roundtrip_test.x PRIVATE
-    ${CMAKE_SOURCE_DIR}/src
-    ${CMAKE_BINARY_DIR}/src
-)
--- a/psi4/src/psi4/libpsio/test/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/test/CMakeLists.txt
@@ -3,11 +3,9 @@
 # the runtime globals normally provided by core.cc, so the test is
 # self-contained (no python, no libdpd).
 #
-# Only the executable is defined here. psi4-core is configured as a separate
-# ExternalProject with its own binary dir and never calls enable_testing(), so
-# an add_test() in this scope would generate no CTestTestfile.cmake and would
-# be invisible to the `ctest` that CI runs from the superbuild directory.
-# Registration therefore lives in the superbuild, in tests/libpsio/.
+# Only the executable is defined here. See ctest registration in `tests/libpsio/`
+# (an add_test() in this scope would not register at all: psi4-core is a separate
+# ExternalProject that never calls enable_testing()).
 add_executable(psio_roundtrip_test.x psio_roundtrip_test.cc stubs.cc)
 target_link_libraries(psio_roundtrip_test.x PRIVATE psio psi4util options ciomr)
 target_include_directories(psio_roundtrip_test.x PRIVATE
--- a/psi4/src/psi4/libpsio/test/CMakeLists.txt
+++ b/psi4/src/psi4/libpsio/test/CMakeLists.txt
@@ -1,0 +1,14 @@
+# Standalone round-trip exerciser for libpsio. Links the libpsio/libpsi4util/
+# liboptions/libciomr static archives plus a tiny stubs TU (test/stubs.cc) for
+# the runtime globals normally provided by core.cc, so the test is
+# self-contained (no python, no libdpd).
+add_executable(psio_roundtrip_test.x psio_roundtrip_test.cc stubs.cc)
+target_link_libraries(psio_roundtrip_test.x PRIVATE psio psi4util options ciomr)
+target_include_directories(psio_roundtrip_test.x PRIVATE
+    ${CMAKE_SOURCE_DIR}/src
+    ${CMAKE_BINARY_DIR}/src
+)
+add_test(NAME psio_roundtrip
+         COMMAND psio_roundtrip_test.x
+         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+set_tests_properties(psio_roundtrip PROPERTIES LABELS "libpsio;quicktests")
--- a/psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
+++ b/psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
@@ -1,0 +1,119 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2025 The Psi4 Developers.
+ *
+ * The copyrights for code used from other parties are included in
+ * the corresponding files.
+ *
+ * This file is part of Psi4.
+ *
+ * Psi4 is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, version 3.
+ *
+ * Psi4 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+// Standalone round-trip test for libpsio. Writes several TOC entries to a new
+// unit (including a multi-page entry that exercises the paging path in
+// PSIO::rw), closes and reopens the file, reads everything back, and checks the
+// table-of-contents lookups. This is the behavior gate for any internal
+// refactor of the file/TOC/paging machinery.
+
+#include <cstdio>
+#include <cstring>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "psi4/libpsio/psio.h"
+#include "psi4/libpsio/psio.hpp"
+
+namespace {
+
+using namespace psi;
+
+bool g_failed = false;
+
+void check(bool cond, const std::string &what) {
+    if (!cond) {
+        std::cerr << "  FAIL: " << what << "\n";
+        g_failed = true;
+    }
+}
+
+// A TOC entry's worth of test data of a given byte size, deterministically
+// filled so we can verify it after a round trip.
+std::vector<char> make_payload(std::size_t nbytes, unsigned seed) {
+    std::vector<char> v(nbytes);
+    for (std::size_t i = 0; i < nbytes; ++i) v[i] = static_cast<char>((i * 31 + seed) & 0xff);
+    return v;
+}
+
+}  // namespace
+
+int main() {
+    psio_init();
+    auto psio = PSIO::shared_object();
+
+    const size_t unit = 84;
+
+    // Sizes chosen to span the single-partial-page, multi-full-page, and
+    // trailing-partial-page branches of PSIO::rw (PSIO_PAGELEN = 65536).
+    struct Entry {
+        std::string key;
+        std::size_t size;
+        unsigned seed;
+    };
+    std::vector<Entry> entries = {
+        {"Alpha small", 64, 1},
+        {"Beta one page", PSIO_PAGELEN, 2},
+        {"Gamma multi-page", 3 * PSIO_PAGELEN + 137, 3},
+        {"Delta empty-ish", 8, 4},
+    };
+
+    std::vector<std::vector<char>> payloads;
+    for (auto &e : entries) payloads.push_back(make_payload(e.size, e.seed));
+
+    // --- Write pass ---
+    psio->open(unit, PSIO_OPEN_NEW);
+    for (std::size_t i = 0; i < entries.size(); ++i) {
+        psio->write_entry(unit, entries[i].key.c_str(), payloads[i].data(), entries[i].size);
+    }
+    psio->close(unit, /*keep*/ 1);
+    check(!psio->open_check(unit), "unit closed after write pass");
+
+    // --- Read pass (reopen, verify each entry + TOC lookups) ---
+    psio->open(unit, PSIO_OPEN_OLD);
+    check(psio->open_check(unit) != 0, "unit open after reopen");
+
+    for (std::size_t i = 0; i < entries.size(); ++i) {
+        check(psio->tocentry_exists(unit, entries[i].key.c_str()), "tocentry_exists: " + entries[i].key);
+        std::vector<char> got(entries[i].size, 0);
+        psio->read_entry(unit, entries[i].key.c_str(), got.data(), entries[i].size);
+        check(got == payloads[i], "payload round-trips: " + entries[i].key);
+    }
+
+    check(!psio->tocentry_exists(unit, "No Such Key"), "absent key reported absent");
+
+    psio->close(unit, /*keep*/ 0);  // erase the scratch file
+
+    if (g_failed) {
+        std::cerr << "psio_roundtrip: FAILED\n";
+        return 1;
+    }
+    std::cout << "psio_roundtrip: all checks passed (" << entries.size() << " entries)\n";
+    return 0;
+}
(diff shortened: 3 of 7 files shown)


susilehtola:
**Scope reduced (force-pushed).** I dropped the C-`psio_*`-API-removal commit. As the analogous libiwl PR (#3428) just showed, the `psio_*` free functions are **public API that downstream plugins use directly** (e.g. v2rdm_casscf), so removing them breaks the ecosystem build — and unlike libiwl's range-for `IWLReader`, the free functions are no worse than the verbose `_default_psio_lib_->...`, so there was little upside to forcing the migration.

What remains is the genuinely safe, non-breaking internal modernization:
- `config.h` macros → `inline constexpr int` (same names + values; the error codes stay `int` rather than becoming an `enum class`, so `psio_error()`'s public signature is unchanged — no downstream impact).
- the standalone round-trip test (the behavior gate libpsio never had).

The deeper internal rewrites from the plan (#3429) — `AIOHandler` single job-queue, `psio_ud`/TOC/`char*`-path/fd-RAII — remain follow-ups that want CI/regression coverage, and any future C-API change needs to be coordinated with downstream-plugin updates rather than done unilaterally.

TiborGY:
The PR description does not look like the Psi4 PR description template and talks about things that have since been backed out of this PR. Please update the description and squash/merge the information in the comment into it.

psi4/src/psi4/libpsio/test/CMakeLists.txt
TiborGY, line 1: Are you sure this ctest actually runs? IIRC this is not how ctests are usually registered in Psi4.

susilehtola: You're right, it does not run. Thanks for catching it.

Two independent reasons, either one sufficient:

1. `psi4-core` is configured as a separate `ExternalProject` with its own binary directory, and that project never calls `enable_testing()` (it lives only in `cmake/ConfigTesting.cmake`, included from the top-level `CMakeLists.txt`). So an `add_test()` in this scope generates no `CTestTestfile.cmake` at all.
2. CI invokes `ctest` from the superbuild directory — `ecosystem.yml` uses `working-directory: ./objdir`, Azure uses `workingDirectory: $(Build.BinariesDirectory)/build` — which cannot descend into the core build tree regardless.

Confirmed on my checkout: `ctest -N -R psio_roundtrip` from either directory reported `Total Tests: 0`, while `psio_roundtrip_test.x` sat built in the core build tree. Same defect in the libiwl test in #3426.

Fixed by moving registration into the superbuild, `tests/libpsio/`, resolving the ExternalProject's `BINARY_DIR`. Now:

```
$ ctest -N -R roundtrip
  Test #538: libpsio-roundtrip
$ ctest -R libpsio-roundtrip
100% tests passed, 0 tests failed out of 1
```

and, since a test that has never been observed to fail isn't worth much, with the binary moved aside:

```
$ ctest -R libpsio-roundtrip
538 - libpsio-roundtrip (Not Run)
```

Description rewritten onto the template as well.

One open question, in the PR body too: #3426 adds `add_subdirectory(libiwl)` at the same spot in `tests/CMakeLists.txt`, so whichever lands second gets a one-line conflict. If you'd rather have a single entry point for C++ unit tests, I'm happy to add an `add_core_unit_test()` macro to `cmake/TestingMacros.cmake` instead.

tests/libpsio/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+# The executables are built inside the psi4-core ExternalProject, which has its
+# own binary directory and does not call enable_testing(). Registration has to
+# happen here, in the superbuild, because that is where `enable_testing()` runs
+# (cmake/ConfigTesting.cmake) and where CI invokes `ctest`.
+ExternalProject_Get_Property(psi4-core BINARY_DIR)
+
+add_test(NAME libpsio-roundtrip
+         COMMAND ${BINARY_DIR}/src/psi4/libpsio/test/psio_roundtrip_test.x${CMAKE_EXECUTABLE_SUFFIX}
+         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+set_tests_properties(libpsio-roundtrip PROPERTIES LABELS "psi;quicktests;libpsio")
```
loriab: looks good, I think the test is much more findable (by a psi4 person) here.

loriab:
```suggestion
set_tests_properties(libpsio-roundtrip PROPERTIES LABELS "psi;smoketests;libpsio")
```
The new test isn't running at present b/c we run scarcely any tests in CI through ctest b/c they're all doubled by pytest, and it's easier to have one driver. This isn't eligible to be registered with pytest b/c it's not a `psi4 input.dat`. Smoke and plugin _are_ run through ctest, so change this here (or start a new `cxxunit` label and add that to https://github.com/psi4/psi4/blob/master/devtools/scripts/ci_run_test.py#L7

psi4/src/psi4/libpsio/CMakeLists.txt
```diff
@@ -34,6 +34,15 @@ list(APPEND sources
   )
 psi4_add_module(lib psio sources)
 
+# The standalone round-trip test links a curated subset of static archives
+# plus a small stubs TU. That works with GNU/Clang on Linux/macOS but not with
+# Windows lld-link, where the unity+PCH build pulls in extra symbol references
+# (Molecule, DPDMOSpace, ...) the curated link line doesn't provide. Restrict
+# it to non-Windows rather than drag in the full dependency closure.
```
loriab: Hmm, I was going to say we could just move the PCH definition to libqt or libciomr, but the fact that it would include `psio.hpp` might bring in the same issues.

Or could `if(NOT(WIN32 AND psi4_ENABLE_PRECOMPILE_HEADERS))` . It doesn't look like PCH is enabled for the Azure Windows lane.

psi4/src/psi4/libpsio/test/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# Standalone round-trip exerciser for libpsio. Links the libpsio/libpsi4util/
+# liboptions/libciomr static archives plus a tiny stubs TU (test/stubs.cc) for
+# the runtime globals normally provided by core.cc, so the test is
+# self-contained (no python, no libdpd).
+#
+# Only the executable is defined here. psi4-core is configured as a separate
+# ExternalProject with its own binary dir and never calls enable_testing(), so
+# an add_test() in this scope would generate no CTestTestfile.cmake and would
+# be invisible to the `ctest` that CI runs from the superbuild directory.
+# Registration therefore lives in the superbuild, in tests/libpsio/.
```
loriab:
```suggestion
# Only the executable is defined here. See ctest registration in `tests/libpsio/`
```
if you like

psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
```diff
@@ -0,0 +1,119 @@
...
+
+    psio->close(unit, /*keep*/ 0);  // erase the scratch file
+
+    if (g_failed) {
+        std::cerr << "psio_roundtrip: FAILED\n";
+        return 1;
+    }
+    std::cout << "psio_roundtrip: all checks passed (" << entries.size() << " entries)\n";
+    return 0;
+}
```
loriab, line 119: I've confirmed locally that reversing some checks here trips the ctest to FAILED.

loriab (commented):
thanks! lgtm after confirmation that the test is running in Azure

psi4/src/psi4/libpsio/test/CMakeLists.txt
susilehtola: Correction to my reply above, and to the PR body: I said the Windows link failure came from the unity build coalescing each static archive into one object. **That was wrong, and I hadn't checked it.**

Linux is unity-built too (`libpsi4util.a` is a single `unity_0_cxx.cxx.o`), and `nm -C --undefined-only` on it shows no reference to any of the symbols in question. So unity isn't the discriminator. PCH isn't either — `psi4_ENABLE_PRECOMPILE_HEADERS` defaults to `OFF` (`CMakeLists.txt:212`) and `azure-pipelines-windows.yml` never sets it, yet that lane failed identically.

lld-link's `>>> referenced by` lines, which I should have read the first time, give the actual cause: clang-cl emits inline and implicit functions that GCC discards as unused, so the Windows objects carry undefined references to out-of-line definitions the curated link line doesn't provide.

```
undefined symbol: psi::DPDMOSpace::~DPDMOSpace(void)
>>> referenced by psio.lib(unity_0_cxx.cxx.obj):(std::vector<psi::DPDMOSpace, ...>::_Assign_...)

undefined symbol: psi::Molecule::Molecule(class psi::Molecule const &)
>>> referenced by psi4util.lib(unity_0_cxx.cxx.obj):(public: class psi::Molecule __cdecl psi::Molecule::clone(void))

undefined symbol: psi::ExternalPotential::print(...)
>>> referenced by psi4util.lib(unity_0_cxx.cxx.obj):(public: void __cdecl psi::ExternalPotential::py_print(void) const)
```

`Molecule::clone()` is inline at `molecule.h:200`, `py_print()` at `extern.h:131`, and `std::vector<DPDMOSpace>` is a member of the real `DPD` in `dpd.h:327`, reached because `close.cc` includes `dpd.h`.

The fourth undefined symbol was **our own bug**, not a toolchain quirk — see the Copilot thread. Comments and commit messages corrected in the latest push.

tests/libpsio/CMakeLists.txt
susilehtola: Done — thanks, this is the bit I'd missed. Applied as `"psi;quicktests;smoketests;libpsio"` rather than swapping `quicktests` out, since 13 existing tests carry both (`cc1`, `fcidump`, `ddd-deriv`, ...) and that keeps the GH macOS `ctest -L quick` lane running it too.

Checked against the actual selectors rather than guessing:

```
ctest -N -L "plug|smoke"     -> Test #538: libpsio-roundtrip   (Azure Linux, ci_run_test.py)
ctest -N --label-regex smoke -> Test #538: libpsio-roundtrip   (Azure Windows)
ctest -N -L quick            -> Test #538: libpsio-roundtrip   (GH macOS)
```

For the record on "isn't running at present": it *was* running, but only on the one lane — GH macOS, `192/222 Test #540: libpsio-roundtrip ... Passed 0.05 sec`. The Azure lanes ran 21–25 tests and it was in none of them. So your diagnosis was right where it counts; I've now got it in all three.

No need for the `cxxunit` label unless you'd prefer one for future C++ unit tests — happy to add it to `ci_run_test.py` if so.

psi4/src/psi4/libpsio/CMakeLists.txt
susilehtola: I chased this and PCH turns out not to be the trigger — `if(NOT(WIN32 AND psi4_ENABLE_PRECOMPILE_HEADERS))` would leave the test building on Azure Windows, where it fails.

You're right that PCH isn't enabled on that lane: `psi4_ENABLE_PRECOMPILE_HEADERS` defaults to `OFF` (`CMakeLists.txt:212`) and `azure-pipelines-windows.yml` never sets it (only `azure-pipelines-linux.yml` and `ecosystem.yml` do). But that lane failed anyway, with the same four undefined symbols — Azure build 8394, `Windows VS2022`. So PCH-off is exactly the configuration that already breaks.

I'd also blamed the unity build in the comment you're replying to. That's wrong too: Linux is unity-built and `nm -C --undefined-only libpsi4util.a` shows none of these references.

The real cause is in lld-link's `>>> referenced by` lines — clang-cl emits inline and implicit functions that GCC discards as unused:

| symbol | emitted into | from |
|---|---|---|
| `DPDMOSpace::~DPDMOSpace` | `psio.lib` | `std::vector<DPDMOSpace>` internals (`dpd.h:327`, via `close.cc`) |
| `Molecule::Molecule(const&)` | `psi4util.lib` | inline `Molecule::clone()`, `molecule.h:200` |
| `ExternalPotential::print` | `psi4util.lib` | inline `py_print()`, `extern.h:131` |

So it isn't about where the PCH is defined; moving it to libqt/libciomr wouldn't help either. Narrowing the guard would need stubs for those three (on top of the `DPD` one we already have), which starts to look like reimplementing chunks of libmints. I've left it at `if(NOT WIN32)` and rewritten the comment to state the actual mechanism, so the next person doesn't re-litigate PCH.

(The fourth symbol, `DPD::file4_cache_del_filenum`, was our own stub declaring `unsigned long` where `dpd.h` says `size_t` — fixed.)

psi4/src/psi4/libpsio/test/stubs.cc
```diff
@@ -0,0 +1,66 @@
...
+
+// PSIO::close calls global_dpd_->file4_cache_del_filenum on every file close.
+// In a non-DPD context the cache is empty, so a minimal local DPD with a no-op
+// method and a non-null instance keeps the call well-defined. Linked without
+// libdpd, so no ODR clash with the real definition.
+class DPD {
+   public:
+    void file4_cache_del_filenum(unsigned long);
+};
+void DPD::file4_cache_del_filenum(unsigned long) {}
```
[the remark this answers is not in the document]
susilehtola: Correct, and this is a real bug rather than a style point. `dpd.h:483` declares `void file4_cache_del_filenum(size_t)`. The stub said `unsigned long`. They're the same type on LP64 and nowhere else.

It's already visible in the Windows CI logs — lld-link asks for exactly the 64-bit spelling, and our 32-bit stub doesn't provide it:

```
lld-link: error: undefined symbol: public: void __cdecl psi::DPD::file4_cache_del_filenum(unsigned __int64)
>>> referenced by psio.lib(unity_0_cxx.cxx.obj):(public: void __cdecl psi::PSIO::close(unsigned __int64, int))
```

So one of the four undefined symbols on that lane was self-inflicted. Fixed to `std::size_t` here and in the sibling libiwl stub (#3426), which had the same declaration.

psi4/src/psi4/libpsio/test/stubs.cc
```diff
@@ -0,0 +1,66 @@
...
+
+// libciomr's DSYEV_ascending lands in the same unity TU as helpers libpsio
+// pulls in; the standalone link references its LAPACK wrapper though the test
+// never diagonalizes. A no-op leaf keeps the test self-contained.
+void C_DSYEV(char, char, int, double *, int, double *, double *, int) {}
+
+// PSIO::close calls global_dpd_->file4_cache_del_filenum on every file close.
+// In a non-DPD context the cache is empty, so a minimal local DPD with a no-op
+// method and a non-null instance keeps the call well-defined. Linked without
+// libdpd, so no ODR clash with the real definition.
```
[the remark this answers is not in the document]
susilehtola: Agreed — the comment was wrong. `close.cc` is compiled against the real `psi::DPD` from `libdpd/dpd.h`, while the stub TU defines a *different* `psi::DPD` in the same namespace. That's an ODR violation regardless of whether libdpd is linked; it holds together only because the call is non-virtual and the mangled name matches (which is precisely why the parameter type has to be spelled `size_t` — see the sibling comment).

Comment rewritten to say so instead of claiming there's no issue.

tests/libpsio/CMakeLists.txt
loriab: thanks, this should be good coverage. No need on my part for a `cxxunit` until/unless we start accumulating more. And I'm ok with skipping Windows.

loriab (approved):
I confirm the new test is passing on Azure-L and GHA-M. lgtm, thanks!

psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
TiborGY: Thanks!

psi4/src/psi4/libpsio/CMakeLists.txt
```diff
@@ -34,6 +34,20 @@ list(APPEND sources
...
+# small stubs TU. That resolves with GNU/Clang on Linux/macOS but not with
+# clang-cl/lld-link, which emits inline and implicit functions that GCC discards
+# as unused, so the Windows objects carry undefined references the curated link
+# line does not provide: psio.lib wants DPDMOSpace::~DPDMOSpace (instantiated by
+# std::vector<DPDMOSpace> in dpd.h, reached via close.cc), and psi4util.lib wants
+# Molecule::Molecule(const Molecule&) (from inline Molecule::clone) and
+# ExternalPotential::print (from inline py_print). Not a unity or PCH artifact --
+# Linux is unity-built too, and Azure Windows builds with PCH off. Restrict the
+# test to non-Windows rather than drag in the full libmints/Libint closure.
+if(NOT WIN32)
```
TiborGY, line 47: Is WIN32 true on Windows-on-ARM systems?

psi4/src/psi4/libpsio/test/stubs.cc
```diff
@@ -0,0 +1,74 @@
...
+namespace psi {
+
+std::string restart_id;
+char *psi_file_prefix = strdup("psio_test");
+std::shared_ptr<PsiOutStream> outfile;
+
+// libciomr's DSYEV_ascending lands in the same unity TU as helpers libpsio
+// pulls in; the standalone link references its LAPACK wrapper though the test
+// never diagonalizes. A no-op leaf keeps the test self-contained.
+void C_DSYEV(char, char, int, double *, int, double *, double *, int) {}
```
TiborGY:
```suggestion
int C_DSYEV(char, char, int, double *, int, double *, double *, int) {}
```
To match the actual declaration being stubbed.

psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
```diff
@@ -0,0 +1,119 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2025 The Psi4 Developers.
```
TiborGY:
```suggestion
 * Copyright (c) 2007-2026 The Psi4 Developers.
```

psi4/src/psi4/libpsio/test/stubs.cc
```diff
@@ -0,0 +1,74 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2025 The Psi4 Developers.
```
TiborGY:
```suggestion
 * Copyright (c) 2007-2026 The Psi4 Developers.
```

psi4/src/psi4/libpsio/test/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+#
+# Only the executable is defined here. See ctest registration in `tests/libpsio/`
+# (an add_test() in this scope would not register at all: psi4-core is a separate
+# ExternalProject that never calls enable_testing()).
+add_executable(psio_roundtrip_test.x psio_roundtrip_test.cc stubs.cc)
+target_link_libraries(psio_roundtrip_test.x PRIVATE psio psi4util options ciomr)
+target_include_directories(psio_roundtrip_test.x PRIVATE
+    ${CMAKE_SOURCE_DIR}/src
+    ${CMAKE_BINARY_DIR}/src
+)
```
TiborGY:
```suggestion
```
This is probably not required.

TiborGY (changes requested):

psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
loriab: fwiw, when I do the license year bump, I usually go back a couple years with the `sed` to catch files like this.

psi4/src/psi4/libpsio/CMakeLists.txt
susilehtola: Good question — yes. CMake sets `WIN32` true for every Windows target, including ARM64 (Windows-on-ARM), so `if(NOT WIN32)` excludes the test there too. That is what we want: the same clang-cl/lld-link closure problem applies on WoA, so it should be skipped there as well. No change needed.

psi4/src/psi4/libpsio/test/stubs.cc
susilehtola: Done in 3b48752. I spelled it `int C_DSYEV(...) { return 0; }` — matching the `int` return from `qt.h`, but with an explicit `return 0;` since an empty body on a non-`void` function is UB. (The return type is not part of the mangled name, so this was cosmetic for the link, but agreed the stub should match the real declaration.)

psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
susilehtola: Bumped to `2007-2026` in 3b48752.

psi4/src/psi4/libpsio/test/stubs.cc
susilehtola: Bumped to `2007-2026` in 3b48752.

psi4/src/psi4/libpsio/test/CMakeLists.txt
susilehtola: Removed in 3b48752 — you were right, both lines are redundant. The top-level `include_directories(src)` and `include_directories(include)` (psi4/CMakeLists.txt) already propagate to this subdirectory, and nothing is generated into the binary `src` tree, so `${CMAKE_BINARY_DIR}/src` adds nothing either. I confirmed both test TUs compile with only the global include dirs.

psi4/src/psi4/libpsio/test/psio_roundtrip_test.cc
susilehtola: Bumped these two to 2026. Agreed a broader `sed` sweep going back a couple years is the right way to catch stragglers like this across the tree — happy to leave that to your usual license-bump pass rather than scattering it here.

TiborGY (approved):
LGTM

PR #3436, opened 2026-06-14
Mark the Python exports of Options::set_read_globals and Options::set_read_globals as deprecated

TiborGY:
## Description

`edit_globals_` is a private `bool` mode flag in the `Options` class. The reason it exists it that `Options` objects keep options in two separate stores:

- globals_:  a single flat `map<string,Data>` the global keyword table.
- locals_: a map<string, `map<string,Data>>` keyed by module name, so each module (SCF, CCENERGY, DFMP2, …) has its own private option table

`edit_globals_` is the flag that decides which of those two stores the option machinery reads from and writes to. It is only ever turned on transiently, when Psi4 starts up, during `clean_options()` and if a plugin is being loaded. Its public face is the getter `read_globals()` and the setter `set_read_globals(bool)`. This is all fine, but these are also currently part of the Python API, which is probably not needed and in case of the setter, slightly dangerous.

To use `set_read_globals` correctly one has to call it to set it to `true`, do whatever has to be done on the globals, and then call it again with false to restore the `Options` object to its normal state. If a Python caller sets it true and never resets it (perhaps because an exception is thrown and handled before the reset) the options object is left silently corrupted, and mayhem ensues.

There is nothing in the Psi4 tree currently that is using these Pybind11 exports, so unless someone likes having these in for debugging or prototyping stuff, IMO it would be best to deprecate and remove them. A pure-Python plugin might conceivably want to inject options into the global space the same way the C++ plugin loader does, and that would need these exports, but I am not aware of anyone doing that.

Since these are not really acute problems, I am setting the deprecation time horizon to 1.12. Hopefully that will be enough time for any downstream users of the Python API to notice the deprecation messages.

- [x] Python API: The Python API export of `Options::set_read_globals(bool)` is now deprecated due to disuse and because it is slightly hazardous. Unless someone speaks up, 1.12 will be the last release to have it.
- [x] Python API: The Python API export of `Options::read_globals()` is deprecated due to disuse and to better hide C++ implementation details. Unless someone speaks up, 1.12 will be the last release to have it.

## Dev notes & details

- [x] Add a universal deprecation helper to driver/p4util/python_helpers.py
- [x] Add deprecation notices for `Options.read_globals()` and `Options.set_read_globals(bool)`

## Checklist
- [x] No new features

## Status
- [x] Ready for review
- [x] Ready for merge

### DIFF
--- a/psi4/driver/p4util/python_helpers.py
+++ b/psi4/driver/p4util/python_helpers.py
@@ -618,6 +618,15 @@ def set_module_options(module: str, options_dict: Dict[str, Any]) -> None:
     for k, v, in options_dict.items():
         core.set_local_option(module.upper(), k.upper(), v)
 
+_original_set_read_globals = core.Options.set_read_globals  # capture first
+def _deprecated_set_read_globals(self, b):
+    warnings.warn(
+        "Options.set_read_globals is deprecated due to disuse and because it is slightly hazardous. Unless someone speaks up, 1.12 will be the last release to have it.",
+        category=FutureWarning,
+        stacklevel=2,
+    )
+    return _original_set_read_globals(self, b)
+core.Options.set_read_globals = _deprecated_set_read_globals
 
 ## OEProp helpers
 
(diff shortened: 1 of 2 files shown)


loriab (approved):
From discussion at https://github.com/psi4/psi4/pull/725/changes, these were originally exposed to integrate with pybind11 plugins, namely Forte. I don't now see it used in Forte, v2rdm_casscf, or hilbert. So it should be ok to retract, and this deprecation warning should let us know.

jturney (approved):

PR #3437, opened 2026-06-15
toward aotoso/sotoao inverses

### LINKED ISSUE #1763: PetiteList.sotoao() Incorrect
According to [the documentation](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/petitelist.h#L329-L330) of `sotoao()`, the function is supposed to return the inverse of `PetiteList.aotoso()`. This matrix is not guaranteed to be invertible. If the matrix is _not_ invertible, multiplying `sotoao()` and `aotoso()` together does not yield the identity matrix.

The problem is that `aotoso()` isn't even guaranteed to be square. The SO space will always be [the number of basis functions](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/petitelist.h#622-L639). The AO space will be [the number of basis functions or the number of atomic orbitals ](https://github.com/psi4/psi4/blob/master/psi4/src/psi4/libmints/petitelist.h#615-L620), depending on whether `include_pure_transform_` is set. In other words, you may be transforming from either Cartesian or pure angular momentum functions to symmetry-adapted pure angular momentum functions. If `d` or higher angular momentum functions are present, then choosing to transform from Cartesian functions means you have more atomic orbitals than basis functions, so the row and column dimensions don't match. The transformation isn't even square, let alone invertible, so of course `sotoao()` returns nonsense.

The following input file almost demonstrates the problem. You'll need to PyBind the `PetiteList` initializer over as well for it to run.

```
molecule mol {
0 3
symmetry c1
        C     0.0000000000   0.0000000000  -0.5928430915
        H    -0.0000000000   0.9469373770  -1.1509808737
        H     0.0000000000  -0.9469373770  -1.1509808737
}

basis = core.BasisSet.build(mol, "ORBITAL", "cc-pvdz", quiet=True)
factory = core.IntegralFactory(basis)

# np.dot(so2ao, ao2so) should always be the identity. The functions are inverses.

# We have a problem!
petite = core.PetiteList(basis, factory, True)
so2ao = petite.sotoao().nph[0]
print(so2ao)
print(so2ao.shape) # (24, 25) 
ao2so = petite.aotoso().nph[0]
print(np.dot(so2ao, ao2so))

# Works fine!
petite = core.PetiteList(basis, factory, False)
so2ao = petite.sotoao().nph[0]
print(so2ao)
print(so2ao.shape) # (24, 24) 
ao2so = petite.aotoso().nph[0]
print(np.dot(so2ao, ao2so))
```

The next question is what we do about this. I see two solutions:
1. Put a flag in sotoao that throws if `include_pure_transform_` is true. I've run this change through full tests and didn't get any errors. However, somebody can assume `sotoao()` is the transpose of `aotoso()` and get rubbish.
2. Remove the functionality of the `include_pure_transform` flag entirely. The transformations are always invertible, but we lose whatever efficiency gain motivated including the flag in the first place.

I don't know which is the best solution, or if there's a third.

Pinging @andysim, @susilehtola, @jturney, and @fevangelista, all of whom have either touched this code or expressed interest in this issue previously.

loriab:
## Description

Deprecate `PetiteList(include_pure_transform=true)` path and migrate internal users to explicit CartAO<->AO transforms (fixes #1763). This PR addresses the long-standing `PetiteList.sotoao(` correctness issue in #1763 : when `include_pure_transform=true`, AO<->SO transforms can be rectangular (`nao != nbf`), so `sotoao()` is not generally an inverse of `aotoso()`.

To remove reliance on that ambiguous/non-invertible path, this PR migrates internal users to an explicit composition:

- `cartao_to_ao_transform()` (CartAO -> AO/BF), then
- `petite_list(false)->aotoso()` (AO/BF -> SO)

closes #1763 

- [x] `PetiteList(include_pure_transform=true)` is deprecated. See code examples for replacement.

#### How this resolves #1763

Issue #1763 highlights that `sotoao()` is not guaranteed to invert `aotoso()` when `include_pure_transform=true` (rectangular transforms for pure-AM shells with l>=2). This PR resolves that at the usage level by:

- avoiding internal dependence on `petite_list(true)` in production paths,
- using explicit CartAO->AO transform + invertible `petite_list(false)` transform,
- adding regression/sentinel tests that fail if the pure-AM CartAO transform is incorrect.

So we no longer depend on the problematic assumed-inverse semantics in those routes.

#### Routes using `include_pure_transform=true` and how they were handled

1. snLinK / GauXC force-cartesian route
  - Before: local `PetiteList(..., true).sotoao()`-style behavior.
  - Now: uses shared `MintsHelper::cartao_to_ao_transform()` + `petite_list(false)` composition.
  - Preserves existing SNLINK_FORCE_CARTESIAN C1 restriction.
  - renamed member data since storing something a little different
2. HF perturbation path (embpot, sphere, dx) in hf.cc
  - Before: `helper.petite_list(true)->aotoso()` to build `u`.
  - Now: `helper.cartao_to_ao_transform()` + `helper.petite_list(false)->aotoso()` to build equivalent `u`.
  - AO→SO setup for the embpot/sphere/dx block replaces a triple element-wise loop with `C_DGEMM('T','N', nao, nsol, nbf, ...)` computing `u_h = cart_to_ao^T · aotoso_false_h` per irrep.
3. DX writer in dx_write.cc
  - Before: `helper.petite_list(true)->aotoso()`.
  - Now: same explicit shared transform composition as HF perturbation path.
4. PCM path
  - Removed stale/unused `PetiteList(..., true) setup (my_aotoso_)` from libpsipcm; no functional dependence remains there.

#### Dev Notes

- [x] Shared utility introduced
  - Added `MintsHelper::cartao_to_ao_transform()` and exposed it to Python.
  - This centralizes CartAO->AO construction and avoids multiple localized implementations.
- [x] `PetiteList(include_pure_transform=True)` now emits deprecation messaging with migration guidance. The `include_pure_transform=False` path is unaffected. (Can't use `PSI_DEPRECATED` b/c only `true` affected.)
- [x]Tests added/updated to ensure correct values. (It'd be nice if they made incorrect pure-AM handling (new fn return `I` mat) fail loudly but only the `force_cartesian_guard` actually fails.):
  - [x] `tests/pytests/test_petitelist_i1763.py` i1763-focused shape/invertibility characterization, pure-AM sphere perturbation sentinel
  - [x] nothing for pcmsolver since its independent of petitelist
  - [x] `tests/embpot1-puream/` new ctest duplicate that uses cc-pVDZ rather than 6-31G*. also added concrete ref values to both
  - [x] `tests/pytests/test_compositejk.py::test_snlink_force_cartesian_guard` (snLinK sentinel). Also refactored the parameterization of some scf tests so it's not showing >100 tests as skipped

### Known separate issue with question not addressed here — `phi_ao` buffer misalignment in the sphere path

- [ ] Claude suggested a quick fix and a nice test case comparing sphere perturbation and embpot. But we also have permission from TDC to remove the sphere perturbation (phone call to CDS 7 Feb 2025 -- see longrunning SAPT notes). Should we (a) remove it; (b) fix it; (c) fix and deprecate it?
  - [x] I've deprecated it, fixed it, changed the two test ref values, and added an AI generated trio of tests.

The `sphere` sub-path of `HF::form_H` has a pre-existing bug orthogonal to the above:

```cpp
// Before fix:
Vector phi_ao(nao);
basisset_->compute_phi(phi_ao.pointer(), x, y, z);  // zeros nbf() entries, fills with pure offsets
phi_so.gemv(true, 1.0, u, phi_ao, 0.0);             // u expects Cartesian layout — wrong for d+
```

`BasisSet::compute_phi` zeros `nbf()` elements and, for puream bases, fills them using pure-shell offsets (incrementing by `INT_NFUNC(puream_, am)` = n_pure per shell). For d shells (n_pure = 5 < n_cart = 6), the extra Cartesian slot is never written and stays zero, so `uᵀ·phi_ao` picks up wrong coefficients for every d+ shell. The `embpot` sub-path immediately above has always been correct — it calls `compute_phi(phi_so, x, y, z)` directly into the `nso`-sized buffer. The sphere path should do the same:

```cpp
// Fix (sphere path only):
basisset_->compute_phi(phi_so.pointer(), x, y, z);
```

The sentinel energy in `test_petitelist_i1763.py` was generated from PR code and will not catch this bug. A separate test (`test_sphere_phi_buf_bug.py`) demonstrates it by running both the sphere path and the embpot path on an identical integration grid for H₂O / cc-pVDZ puream — the energies must agree after the fix.

> **Reviewers:** the sphere-path fix is a one-liner and could reasonably be included here rather than in a follow-on PR. LAB -- fix now included

## Checklist

## Status
- [x] Ready for review
- [x] Ready for merge


tests/embpot1/input.dat
```diff
@@ -102,3 +102,11 @@ grad_embpot.add(grad_nuc)
...
 compare_matrices(grad_analytic, grad_embpot, 5, "Gradient with analytic point charge embedding vs. with numerical EMBPOT embedding") #TEST
 compare_values(e_analytic, e_embpot, 6, "Energy with analytic point charge embedding vs. with numerical EMBPOT embedding")  #TEST
+
+ene_ref = -76.0194112288343  #TEST
+grad_ref = ([
+        [-6.82055788e-03, -4.91941583e-07, -1.37359568e-02],
+        [ 1.97327569e-03,  1.21536585e-02,  8.95300981e-03],
+        [ 1.97320516e-03, -1.21531649e-02,  8.95261948e-03]])
+compare_values(ene_ref, e_analytic, 6, "fixed compare ene")  #TEST
+compare_values(grad_ref, grad_analytic, 5e-5, "fixed compare grad")  #TEST
```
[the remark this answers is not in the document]
loriab: It's fine -- they get converted.

psi4/src/psi4/libmints/petitelist.h
```diff
@@ -303,9 +303,10 @@ class PSI_API PetiteList {
         */
     SharedMatrix aotoso();
 
-    /** @return the SO->AO coefficient matrix (the inverse of AO->SO; for Abelian point groups it
-        is a transpose of AO->SO matrix). The columns correspond to AOs (see AO_basisdim() )
-        and rows to SOs (see SO_basisdim() ).
+    /** @return the SO->AO coefficient matrix (the inverse of AO->SO (unless
```
JonathonMisiewicz, line 306: Nested parentheses are hard to read.

psi4/src/psi4/libfock/snLinK.cc
```diff
@@ -279,15 +278,13 @@ snLinK::snLinK(std::shared_ptr<BasisSet> primary, Options& options) : SplitJK(pr
...
     // create matrix for spherical-to-cartesian matrix transformations
     if (force_cartesian && primary_->has_puream()) {
-        const auto factory = std::make_shared<IntegralFactory>(primary, primary, primary, primary);
-        PetiteList petite(primary, factory, true);
-        sph_to_cart_matrix_ = petite.sotoao(); 
+        MintsHelper helper(primary_, options_, 0);
+        cartao_to_ao_matrix_ = helper.cartao_to_ao_transform();
 
         // SNLINK_FORCE_CARTESIAN only works with C1 symmetry currently
         // TODO: Fix this!
```
JonathonMisiewicz, line 285: Is this error message still needed, after this fix?

psi4/src/psi4/libmints/petitelist.h
```diff
@@ -303,9 +303,10 @@ class PSI_API PetiteList {
...
+        include_pure_transform_==true and basis_.has_puream()==true and basis_ contains l>=2);
+        for Abelian point groups it is a transpose of AO->SO matrix). The columns correspond to AOs
+        (see AO_basisdim() ) and rows to SOs (see SO_basisdim() ).
```
loriab, line 309:
```suggestion
    /** @return the SO->AO coefficient matrix (the inverse of AO->SO; for Abelian point groups, it
        is a transpose of AO->SO matrix). The columns correspond to AOs (see AO_basisdim() )
        and rows to SOs (see SO_basisdim() ).
        Note that the contract to be the inverse of AO->SO is *not* valid when
        include_pure_transform_==true and basis_.has_puream()==true and basis_ contains l>=2.
```

psi4/src/psi4/libmints/petitelist.h
loriab: I agree -- misplaced priorities when I was writing that. Hopefully it's clearer now.

psi4/src/psi4/libfock/snLinK.cc
loriab: As far as I know, yes. I didn't investigate that aspect of the snlink hook-up.

jturney (approved):

psi4/src/psi4/libscf_solver/hf.cc
```diff
@@ -636,9 +638,7 @@ void HF::form_H() {
 
                             double jacobian = weight * r * r * sin(theta);
 
-                            basisset_->compute_phi(phi_ao.pointer(), x, y, z);
-
-                            phi_so.gemv(true, 1.0, u, phi_ao, 0.0);
+                            basisset_->compute_phi(phi_so.pointer(), x, y, z);
```
JonathonMisiewicz: For clarity's sake, can we move the declarations of the `phi_ao` and `phi_so` variables to the scopes where they're actually used? For instance, this `phi_so` really should be a `std::vector<double>(nso, 0)` created inside the if-clause.

tests/pytests/test_petitelist.py
```diff
@@ -0,0 +1,386 @@
...
+    nao=25 > nbf=24) the sphere and embpot paths must give the same energy.
+
+    With the phi_ao buffer misalignment bug present, compute_phi fills nbf=24
+    pure-function values into a nao=25 Cartesian buffer, the extra slot stays
+    zero, and u^T * phi_ao gives wrong phi_so values for the d-shell.  The
+    resulting V_eff and SCF energy deviate from the correct embpot answer by
+    an amount that far exceeds numerical integration error.
+
+    After the fix (sphere path calls compute_phi(phi_so,...) directly, like
+    embpot), the two energies agree to 10 decimal places.
```
JonathonMisiewicz, line 292: :confetti_ball:

psi4/src/psi4/libfock/snLinK.cc
JonathonMisiewicz: It most definitely is. You removed the (broken) implementation of symmetry handling... I'll have to fix that after GauXC is wrapped.

JonathonMisiewicz (commented):
Some minor clarity changes requested, but good enough for now otherwise.

psi4/src/psi4/libscf_solver/hf.cc
loriab: Done. `phi_so` was in all three branches, while `phi_ao` was only in the one. Unfortunately there's no test hitting `perturb_with dx`, but `data()` doesn't usually cause problems.

JonathonMisiewicz (approved):

Issue #2725, opened 2022-09-25
Rehabilitate or remove PSIO multivolume striping functionality


TiborGY:
Poking around the PSIO machinery, it seems to me that it is meant to be capable of doing a sort of "userland RAID0", where IO is striped over multiple scratch directories. But as far as I can tell, this is not (in any shape or form) accessible to users, so the number of volumes is always 1.

While this is a somewhat interesting feature, and I can imagine that there are circumstances where it would be useful, it comes at the cost of increasing code complexity in PSIO.

Right now, it is the worst of both worlds: noone can benefit from the feature, and PSIO code is more complicated. So IMO, this feature should be either rehabilitated and documented, or stripped out.

I am not volunteering for either, so this may very well be one of those "for the sake of posterity" issues.

lothian:
Yes, I wrote the original version of `libpsio`, and this I/O structure was an important part of its use for many years.  Indeed, it was intended to replicate some of the key performance features of an earlier I/O library (long since removed) contained within the old `libciomr` that have infrastructure for about four different types of I/O.   However, at some point in the development of Psi4, someone wrote the new, much more complicated code for building the file paths, they broke this capability.  Chesterton's fence strikes again.

JonathonMisiewicz:
The "correct" solution is probably to replace PSIO with hdf5. I'd like to talk about that at PsiCon.

TiborGY:
> The "correct" solution is probably to replace PSIO with hdf5. I'd like to talk about that at PsiCon.

Q5Cost ?
I have seen some evidence of Q5Cost support in the MRCC codebase, so that could maybe replace the currently very fragile MRCC interface.

JonathonMisiewicz:
Further discussion of replacing `libpsio` should be moved to a new issue. As for your specific proposal:

I'm not familiar with Q5Cost, and a quick Google search gives me a 14-year-old manual with so many out-of-date links that I doubt Q5Cost is actively maintained. That's a compelling argument against adopting it.

Q5Cost was a proposed standard format for tensors-with-metadata. If MRCC is the only other major user and we have no reason to believe in other users, interoperability is not a compelling reason.

TiborGY:
Closed by #3452 : the PSIO multivolume striping functionality is removed.

PR #3452, opened 2026-07-03
libpsio: remove multi-volume file striping

susilehtola:
## Summary

Removes the long-unused **multi-volume file striping** from `libpsio`. A PSIO
unit could historically be striped across up to `PSIO_MAXVOL` (8) physical
files, with consecutive 64 KiB pages distributed round-robin across the
volumes. In practice the feature was never exercised:

- The default was hard-wired to a single volume (`NVOLUME=1`) and nothing in
  the tree ever configured more.
- The per-volume path lookup (`get_volpath`) was already dead code — `open()`
  discarded its result and used `PSIOManager::get_file_path()` instead.
- No code outside `libpsio` reads `numvols`, `PSIO_MAXVOL`, or `psio_vol[]`
  (verified by an exhaustive grep of `src/`, `include/`, and `driver/`).

Each unit is now backed by a single file.

## Changes

- **`psio_ud`**: drop `numvols` and the `vol[PSIO_MAXVOL]` array → a single
  `psio_vol vol`.
- **`config.h`**: drop `PSIO_MAXVOL` and the three striping-only error codes
  (`MAXVOL`, `NOVOLPATH`, `IDENTVOLPATH`). Their numeric values (3, 4, 19) are
  left as gaps so the remaining **public** error-code values are unchanged.
- **`rw.cc`**: seek once and stream sequentially, instead of seeking every
  volume and selecting one per page.
- **`volseek.cc`**: drop the `numvols` dimension (final offset is
  `page*PAGELEN + offset`); the overflow-safe chunked `lseek` is preserved.
- **`open`/`exists`/`close`/`init`**: collapse the per-volume loops to a single
  file; drop the now-meaningless identical-volume-path check.
- Delete `get_numvols.cc` and `get_volpath.cc`; stop seeding the `NVOLUME` /
  `VOLUMEn` default keywords.

Net: **−254 lines**, 2 files removed.

## Validation

The standalone `psio_roundtrip` test (added in #3430) passes unchanged. It
deliberately spans the **single-partial-page**, **multi-full-page**
(`3*PSIO_PAGELEN + 137` bytes), and **trailing-partial-page** branches of
`PSIO::rw`, so the rewritten sequential I/O path is covered.

## Stacking

Stacked on **#3430** (libpsio config.h modernization); both edit
`libpsio/config.h`. GitHub shows the cumulative diff until #3430 merges, after
which this rebases to the incremental change. Review the last commit only, or
compare against `libpsio/modernize`.


loriab:
Historical note that multi-volume file striping did work in Psi3. I'm not surprised if that aspect of libpsio didn't get built out in the port to Psi4.

susilehtola:
Thanks — that matches what the code shows, and it's worth recording since the Psi3 provenance is the reason the machinery looked plausible.

The striping was carried over structurally but never wired up in Psi4:

- `psio_get_numvols()` reads `NVOLUME` from the file-config keywords, but no default is ever seeded above 1, and nothing in-tree sets it.
- `psio_get_volpath()` still resolves per-volume paths, but `PSIO::open()` discards the result and uses `PSIOManager::get_file_path(unit)` for every volume — the in-tree comment even says *"A bit of a hack in psio open at the moment, breaks volumes and some error checking"*.
- Consequently `vol[i]` for `i > 0` was never opened, and `psio_volseek()`'s `numvols` divisor was always 1.

So this PR removes a feature that was already inert in Psi4, rather than one that worked and is being dropped. Nothing outside libpsio reads `numvols`, `PSIO_MAXVOL` or `psio_vol[]`. The round-trip test from #3430 covers the rewritten `PSIO::rw` paths (sub-page, full-page, multi-page + trailing partial), which is where a striping-removal bug would show up.

If multi-volume striping is ever wanted again, it would want doing against `PSIOManager` rather than resurrecting the `vol[PSIO_MAXVOL]` array.

loriab (approved):
I reviewed the last commit -- lgtm and @CDSherrill has given verbal approval to the removal of the psi3-era feature.

Needs rebase now that its parent is merged and conflicts resolved.

TiborGY:
I'll review this after the rebase.

susilehtola:
Rebased onto current master. The four `#3429` test-infrastructure commits are already upstream (merged via #3430), so they dropped out; this branch is now the single striping-removal commit. Rebuilt clean and the `psio_roundtrip` standalone test passes (4/4 entries). Ready for review, @TiborGY.

psi4/src/psi4/libpsio/close.cc
```diff
@@ -64,24 +64,18 @@ void PSIO::close(size_t unit, int keep) {
         this_entry = next_entry;
     }
 
-    /* Close each volume (remove if necessary) and free the path */
-    for (i = 0; i < this_unit->numvols; i++) {
-        int errcod;
+    /* Close the file (remove if necessary) and free the path */
+    int errcod = SYSTEM_CLOSE(this_unit->vol.stream);
```
TiborGY:
```suggestion
    const int errcod = SYSTEM_CLOSE(this_unit->vol.stream);
```

psi4/src/psi4/libpsio/config.h
```diff
@@ -62,7 +61,7 @@ inline constexpr int PSIO_ERROR_KEYLEN = 15;
 inline constexpr int PSIO_ERROR_BLKSIZ = 16;
 inline constexpr int PSIO_ERROR_BLKSTART = 17;
 inline constexpr int PSIO_ERROR_BLKEND = 18;
-inline constexpr int PSIO_ERROR_IDENTVOLPATH = 19;
+// Value 19 (IDENTVOLPATH) is retired with multi-volume striping.
 inline constexpr int PSIO_ERROR_MAXUNIT = 20;
```
TiborGY, line 65: The removal of these symbols could be a C++ API change, depending on their visibility, which I am not entirely sure about. So I would mention them in the release notes as an API change.

TiborGY (approved):

TiborGY:
IDK why the Azure Windows CI lane is so flaky, but here we go again.

TiborGY:
Merging this closes #2725 where I noted the defunct multivolume functionality.

PR #3455, opened 2026-07-09
Fix missing include in `timer.cc`

NathanGillispie:
## Description

Fixes missing include that doesn't appear for unity builds. I turn off unity builds because my nvim LSP likes the compile_commands.json when each entry is for a source file. #3443 introduced `get_timer_records()` and `TimerRecord` usage into `timer.cc`. Both are declared in `qt.h` but never included.

With unity builds on, `timer.cc` is bundled with other sources that do include `qt.h`, in particular `schmidt.cc`. I suspect newer CMake versions may change which sources are bundled together which could cause this bug to randomly appear later. Heisenbug averted?

none

## AI
- AI was used to find this bug and provide definitive proof for the cause.

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libqt/timer.cc
+++ b/psi4/src/psi4/libqt/timer.cc
@@ -118,6 +118,7 @@ typedef int omp_lock_t;
 #include "psi4/libciomr/libciomr.h"
 #include "psi4/libpsi4util/PsiOutStream.h"
 #include "psi4/libpsi4util/exception.h"
+#include "psi4/libqt/qt.h"
 
 /* guess for HZ, if missing */
 #ifndef HZ


NathanGillispie:
@loriab, we had a discussion on slack where you planned out a cmake flag to include -isystem headers via `-Wsystem-headers` so deprecation warnings wouldn't be hidden sometimes. Perhaps it would be prudent to have some option to help developers not miss bugs like this? I'm thinking hijack `FORCE_PEDANTIC` to also set `psi4_CMAKE_UNITY_BUILD=OFF`?

loriab (approved):

loriab:
Good catch, thanks. I thought I had one CI lane that wasn't using unity, but I bet that was in #3353 that isn't merged yet.

> we had a discussion on slack where you planned out a cmake flag to include -isystem headers via -Wsystem-headers so deprecation warnings wouldn't be hidden sometimes. Perhaps it would be prudent to have some option to help developers not miss bugs like this? I'm thinking hijack FORCE_PEDANTIC to also set psi4_CMAKE_UNITY_BUILD=OFF?

I'm not quite recalling that discussion or where isystem/system-headers factor in. Definitely we should try to avoid missing includes. unity=off can add up to 30m to the CI time, so I want to stay minimal with that. Why tie the force_pedantic and unity options together?

NathanGillispie:
>  Why tie the force_pedantic and unity options together?

My idea was to have an option to be more verbose and help devs catch simple errors that could go under the radar.

> I'm not quite recalling that discussion or where isystem/system-headers factor in.

Essentially, some legitimate deprecation warnings were not appearing because of a quirk/bug in GCC. By turning those warnings on, you actually get notified of all deprecations. It wasn't an uncommon case either, it happened when using a deprecated cast within a make_shared constructor. For the above idea that might be a useful inclusion.

Now that I think about it though, we could also just have some sort of linter that finds missing includes. I'm currently testing clang-include-cleaner which does that and also finds unused includes.

NathanGillispie:
#3481 was merged instead

PR #3456, opened 2026-07-09
Remove a few unused includes

NathanGillispie:
## Description

This seeks to improve clean build times by removing includes unused for builds with disabled addons. These were found with the `clang-include-cleaner`, however this program notoriously reports false positives. To automate the process and identify false positives, a few things were done.
1. I used a non-unity build with no addons (hence the addition to `timer.cc` to reflect the fix in #3455) to make the TUs more isolated.
2. `clang-include-cleaner` identified lines of unused includes.
3. For each candidate line, the include was removed then the TU was recompiled.
4. If errors occured, the line was kept and marked as a false positive.
5. Commits were made after a sufficient number of diffs so that I could split up the PR or run `git bisect` if necessary.

## Stats
\[These stats are dated to June. Since then, I've restored some includes and ran another pass of the cleaner (~560 diffs), but they are still indicative.\]
Of the 1189 files analyzed, 914 (77%) had flagged unused includes. This amounts to 2068 unused include lines. As you can see from the PR diffs, there were actually **1691 unused includes** implying a false positive rate of 18%. Considering 793 files were changed, the false positive rate on a file basis is 13%.

From this reconnaissance, the top offenders were `psimrcc/transform_read_so.cc`, `occ/omp3_ip_poles.cc`, `occ/omp2_ip_poles.cc`, `libfock/DirectJK.cc`, and `scfgrad/scf_grad.cc`. The most over-included headers were
```
222  "psi4/libpsi4util/PsiOutStream.h"
132  "MOInfo.h"
128  "psi4/libciomr/libciomr.h"
125  "psi4/libqt/qt.h"
120  "Frozen.h"
110  "psi4/libpsio/psio.hpp"
 90  "defines.h"
 78  "Local.h"
```
Personally, I have seen `psi4/libpsi4util/PsiOutStream.h` unnecessarily included many times. I think developers just copy and paste and think that there's a good reason for it.

### Build times

On my computer (AMD 6900HX), I observed a -3.4% reduction in wall build times. For a non-unity build, I observed a -2.5% reduction in build times. On a different computer, I automated 10 clean builds each for master and this branch. I found a -3% reduction in unity build times with no plugins, but the results were not statistically significant at p=.303. When looking at the TU-specific changes for a non-unity build, it seems that most source files see no improvement whatsoever, and improvements are dominated by a handful of loss leaders. This is likely due to the *preprocessor* removing redundant includes using the `#ifndef` mechanism. This can happen if the same header is included both transitively and directly. 

None

## Dev notes

- [x] Aside from removing includes. The only code change was in `export_pcm.cc`. Here, includes should have been placed after `#ifdef USING_pcmsolver`. No includes were deleted in this file.
- [x] See my comment for information about removing pybind11 includes.  

## AI
- [x] Sonnet 5 medium thinking was used to write the script that automated finding unused includes.
- [x] The same model was used to commit those scripts and temporary logs which have since been reverted.

## Checklist

## Status
- [x] Ready for review


NathanGillispie:
Found the issue for `export_pcm.cc`. I suspect there might be a lot of includes that **should** be behind `#ifdef USING_*` statements... I'll recompile with all addons and see what other issues pop up.

NathanGillispie:
So it seems that the PCM issue is completely resolved. The problem is that I had removed includes that genuinely were unused for builds with pcmsolver disabled. In `export_pcm.cc`, the includes should have been after `#ifdef USING_pcmsolver`. In `export_wavefunction.cc`, pybind includes were removed that helped runtime implicit casts causing tests to fail.

The only other files where includes related to pybind11 were edited are:
- [`psi4/src/core.cc`](https://github.com/psi4/psi4/pull/3456/changes/3084579fff701d43714695915ac1bc5a7cbf2c13#diff-8f5f16f3222b618b036494ab237e899ef0f547044cec6b2065547fad51350ca7L40)
- [`psi4/src/export_mints.cc`](https://github.com/psi4/psi4/pull/3456/changes/cf7fa19e16e5f1473d5a783a69f5bb5c1dc906d4#diff-5c8ff36b76af6b7aabec09f273435b8b4c0f09f114d1b2f295036d3a41226b86L29)
- [`psi4/src/export_trans.cc`](https://github.com/psi4/psi4/pull/3456/changes/adaf900484b61c49914d69f68298ada3c2c800ea#diff-2044e3a644cb95e0308f1ae3ffb34d9e5ca10d6b126279c0f1413de1c99dd20cL31-L33)
- [`psi4/src/psi4/detci/civect.cc`](https://github.com/psi4/psi4/pull/3456/changes/f9d92252e027a6d73976e489b3e4b16894302f1c#diff-903cd55a09421b7c4903f1ce10c1276c98daa70c3e27e5d33e50377fb23565bdL52)
- [`psi4/src/psi4/mrcc/mrcc.cc`](https://github.com/psi4/psi4/pull/3456/changes/57b57d4db890eaa1121354d71632a03383919551#diff-cd98d8c191f70efa1b854b9617d3bf2b5efcba051351da3f941a5d2e2e44edacL35)

Out of an abundance of caution I could leave these.

NathanGillispie:
Updates on the pybind includes:
- `core.cc`: valid removal because it was indirectly included through `python_data_types.h`.
- `civect.cc`: valid removal because it was indirectly included through `civect.h`.
- `export_mints.cc`, `export_trans.cc`: I was able to get runtime failures due to missing includes in non-unity builds. I missed this for three reasons. First, clang-include-cleaner notoriously marks pybind includes as false positives. Second, these evaded detection because compiling the TU gives no compile-time warnings. Third, during the testing stage, I forgot to turn off unity builds which wouldn't have pulled the pybind includes in from other files. That's my bad because I should have remembered. That also explains why the CI passes, and again justifies keeping one run as non-unity.
- `mrcc.cc`: Out of an abundance of caution I'm returning the pybind includes here.

TiborGY (commented):
I love to see the negative delta-LoC and build speed improvement, but perhaps this PR could be split into several, given how many files it touches?

NathanGillispie:
Sure, that should be easy. How do you think I should split it up? I could shoot for a max number of diffs and split it based on module.

TiborGY:
I think I lean towards per-module. And in case of CC, since the subdirectories are already laid out and the are a _lot_ of changed files in CC, per-submodule.

susilehtola:
Also, make sure to test this via a non-unity build, since unity builds aren't sensitive to missing includes.

susilehtola:
IMHO since the whole PR is just deleting includes, it doesn't need much manual reviewing; I would prefer a single PR since it's easier to check that it still builds

TiborGY:
> IMHO since the whole PR is just deleting includes, it doesn't need much manual reviewing; I would prefer a single PR since it's easier to check that it still builds

That's a fair point. I was thinking maybe having multiple commits would be better for being able to revert/bisect/git-blame with a finer granularity if any issues are found later.

NathanGillispie:
Yeah I think having bisectable commits is ideal. you can have your cake and eat it too by rebase merging. That would be one PR + multiple commits. I'll work on getting the correct partition when I get back to my computer.

NathanGillispie:
Also I ran another pass of the include cleaner this time in the opposite direction and it was able to find a bunch more unused includes. So I'll see if I can milk more out of it.

loriab:
Sidenote that bisectable commits is very valuable. But because we nearly always squash merge PRs, that keeps psi4 master bisectable while giving developers free-rein on commit history w/i PRs. Haven't been following the whole conversation, though. Working on the CI problems.

NathanGillispie:
I decided to automate 10 clean builds for both this and master and it seems that the 3% improvement remains, but it's not statistically significant (p=.30). When looking at TU-specific build times, only a handful of files dominated improvements like `zero_pdm.cc` which went from 1.58s to 0.49s for a non-unity build. Everything else is probably transitively included, and the preprocessor already deletes any redundant includes by the `#ifndef something_h_` mechanism. I think that's the only explanation, because when I look at the `-ftime-report` breakdown, a majority of compile time is spent parsing the input. That step includes creating an AST for -possibly unnecessary- headers, the exact type of thing that should be improved by this PR.

Also, it's possible some sources require symbols from one header `A.h`, but transitively include them through a larger unnecessary header `B.h`. The unused include cleaner can't identify those scenarios. It's possible that I could have removed the `A.h` include and left `B.h`. For now, I think a majority of these diffs are cosmetic, and I'm not sure how I feel about that.

To actually make progress, I think I could identify cases where forward declarations could be used. Also, clang does have an "include what you use" tool which is more sophisticated than their include cleaner, and would identify those transitive include scenarios.

NathanGillispie:
Ok I repartitioned the diffs into 17 commits organized by module or cc submodule. That means it can be rebase merged without putting junk in the master history.

Each commit should be under 200 diffs.

PR #3465, opened 2026-07-18
libiwl: add standalone round-trip test

susilehtola:
A standalone characterization test for the libiwl bucket I/O, landed on its own ahead of the phase-1 modernization so the refactor has a behavior-preserving guard to build on. **No library source changes** — this is purely a new test.

Split out of #3426 at TiborGY's request (tests first, modernization respun on top).

## Description
`test/iwl_roundtrip_test.cc` writes integrals through the C++ `IWL` API and reads them back through the C-style `iwl_buf_*` API (the dominant in-tree read pattern), checking every bucket-boundary case the 1995 README warned about: empty, `1`, `B-1`, `B`, `B+1`, and `3.5*B` integrals (`B = IWL_INTS_PER_BUF`). It links the libiwl/libpsio/libpsi4util/liboptions/libciomr archives plus a tiny stubs TU (`test/stubs.cc`) supplying the runtime globals normally provided by core.cc/libdpd (`global_dpd_`, `psi_file_prefix`, a no-op `C_DSYEV`), so it needs neither Python nor libdpd.

Built inside the psi4-core ExternalProject and registered as the `libiwl-roundtrip` ctest in the superbuild (`tests/libiwl/`, where `enable_testing()` runs), labeled `psi;quicktests;smoketests;libiwl`. Skipped on Windows, where the curated link line does not resolve under clang-cl/lld-link (see `psi4/src/psi4/libiwl/CMakeLists.txt`).

- [ ] None (test-only).

## Dev notes & details
- [x] new `libiwl-roundtrip` ctest; C++ write / C-API read round-trip over the bucket-boundary sizes
- [x] `test/stubs.cc` hosts the runtime globals so the exerciser links without Python/libdpd
- [x] non-Windows only (clang-cl/lld-link cannot resolve the curated link)

## Questions
- [ ] None outstanding.

## AI
- [x] AI usage: the test and its hosting stubs.

## Checklist
- [ ] Docstring or narrative docs/sphinxman/source updated if needed (test-only)
- Test running is well covered by CI.

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
+++ b/psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
@@ -28,7 +28,15 @@
 
 // Standalone round-trip test for libiwl. Exercises every bucket-boundary case
 // the 1995 README warned about: empty, single, exactly one bucket, just over
-// one bucket, mid-bucket termination, and all-zeros sweeps.
+// one bucket, and mid-bucket termination.
+//
+// Also covers cutoff filtering. write_value() keeps an integral only when
+// |value| > cutoff, so sub-cutoff entries are *dropped*, not stored as zeros --
+// nothing is read back for them at all. That makes filtering a repacking
+// operation: survivors slide forward into earlier bucket slots, so which
+// bucket the dropped run falls in changes the packing downstream of it. The
+// positional cases below (first/second/second-to-last/last bucket fully
+// filtered, plus all-filtered and interleaved) pin that behavior down.
 
 #include <cmath>
 #include <cstdio>
@@ -73,6 +81,21 @@ std::vector<Quartet> make_test_data(std::size_t n) {
     return out;
 }
 
+// Sub-cutoff stand-ins. Exactly +0.0 is the common case in a real integrals
+// sweep, but a nonzero magnitude below the cutoff must be dropped just the
+// same, and |value| == cutoff must *also* drop because write_value() tests
+// strictly greater-than. Cycling all three keeps that boundary honest.
+double subcutoff_value(std::size_t i) {
+    switch (i % 3) {
+        case 0:
+            return 0.0;
+        case 1:
+            return (i % 6 == 1) ? 1.0e-20 : -1.0e-20;
+        default:
+            return 1.0e-14;  // == cutoff, so filtered by the strict >
+    }
+}
+
 // Drive an IWL write loop using the C++ API.
 void write_all(psi::IWL &buf, const std::vector<Quartet> &data) {
     for (const auto &q : data) {
@@ -108,9 +131,12 @@ std::vector<Quartet> read_all(int itap) {
     return out;
 }
 
+// Exact comparison, deliberately. IWL neither scales nor rounds: write_value()
+// casts to Value (= double, so a no-op) and put()/fetch() move the raw bytes
+// through libpsio. A surviving integral must therefore come back bit-identical,
+// and any tolerance here would only hide a real corruption.
 bool quartets_equal(const Quartet &a, const Quartet &b) {
-    return a.p == b.p && a.q == b.q && a.r == b.r && a.s == b.s &&
-           std::fabs(a.value - b.value) < 1.0e-12;
+    return a.p == b.p && a.q == b.q && a.r == b.r && a.s == b.s && a.value == b.value;
 }
 
 // Run one write/read round-trip for the given size and tape unit. Returns
@@ -160,6 +186,42 @@ bool round_trip_empty(int itap) {
     return true;
 }
 
+// Round-trip N integrals of which a contiguous run [lo, hi) is pushed below the
+// cutoff. Only the survivors should come back, in order and bit-exact; the
+// dropped run must leave no trace (no zero-valued placeholder, no gap).
+bool round_trip_filtered(std::size_t n, std::size_t lo, std::size_t hi, int itap, const std::string &what) {
+    auto data = make_test_data(n);
+    std::vector<Quartet> expected;
+    for (std::size_t i = 0; i < n; ++i) {
+        if (i >= lo && i < hi) {
+            data[i].value = subcutoff_value(i);
+        } else {
+            expected.push_back(data[i]);
+        }
+    }
+
+    {
+        psi::IWL buf(psi::_default_psio_lib_.get(), itap, 1.0e-14,
+                     /*oldfile*/ 0, /*readflag*/ 0);
+        write_all(buf, data);
+        buf.set_keep_flag(true);
+    }
+
+    auto got = read_all(itap);
+
+    if (got.size() != expected.size()) {
+        std::cerr << "  " << what << ": expected " << expected.size() << " survivors, read " << got.size() << "\n";
+        return false;
+    }
+    for (std::size_t i = 0; i < expected.size(); ++i) {
+        if (!quartets_equal(expected[i], got[i])) {
+            std::cerr << "  " << what << ": survivor " << i << " mismatch\n";
+            return false;
+        }
+    }
+    return true;
+}
+
 }  // namespace
 
 int main() {
@@ -182,6 +244,33 @@ int main() {
         std::cout << "[iwl_roundtrip] N = " << n << "\n";
         if (!round_trip(n, unit++)) ++failures;
     }
+
+    // Cutoff filtering, positioned bucket by bucket over a four-bucket write.
+    // Each case drops a full bucket's worth of input, so the survivors have to
+    // repack across the remaining buckets -- the first and last cases also
+    // exercise "the very first thing written is dropped" and "the file ends on
+    // a dropped run", where an off-by-one in flush/lastbuf would show up.
+    const std::size_t N4 = 4 * B;
+    struct FilterCase {
+        const char *what;
+        std::size_t lo, hi;
+    };
+    const std::vector<FilterCase> filtered = {
+        {"first bucket below cutoff", 0, B},
+        {"second bucket below cutoff", B, 2 * B},
+        {"second-to-last bucket below cutoff", 2 * B, 3 * B},
+        {"last bucket below cutoff", 3 * B, N4},
+        {"all below cutoff", 0, N4},
+    };
+    for (const auto &fc : filtered) {
+        std::cout << "[iwl_roundtrip] " << fc.what << "\n";
+        if (!round_trip_filtered(N4, fc.lo, fc.hi, unit++, fc.what)) ++failures;
+    }
+
+    // A sub-cutoff run that straddles a bucket boundary, so survivors on both
+    // sides of it repack into the same bucket.
+    std::cout << "[iwl_roundtrip] run straddling a bucket boundary\n";
+    if (!round_trip_filtered(3 * B, B - 7, B + 11, unit++, "straddling run")) ++failures;
 
     if (failures) {
         std::cerr << failures << " case(s) failed.\n";
--- a/psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
+++ b/psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
@@ -1,0 +1,192 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2026 The Psi4 Developers.
+ *
+ * The copyrights for code used from other parties are included in
+ * the corresponding files.
+ *
+ * This file is part of Psi4.
+ *
+ * Psi4 is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, version 3.
+ *
+ * Psi4 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+// Standalone round-trip test for libiwl. Exercises every bucket-boundary case
+// the 1995 README warned about: empty, single, exactly one bucket, just over
+// one bucket, mid-bucket termination, and all-zeros sweeps.
+
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "psi4/libiwl/iwl.h"
+#include "psi4/libiwl/iwl.hpp"
+#include "psi4/libpsio/psio.h"
+#include "psi4/libpsio/psio.hpp"
+#include "psi4/libpsi4util/PsiOutStream.h"
+
+// Hosting globals (`outfile`, `restart_id`, `psi_file_prefix`, `global_dpd_`)
+// live in test/stubs.cc; the link line pulls that TU in.
+
+namespace {
+
+using namespace psi;
+
+struct Quartet {
+    int p, q, r, s;
+    double value;
+};
+
+// Deterministic pseudo-random test data with bounded indices.
+std::vector<Quartet> make_test_data(std::size_t n) {
+    std::vector<Quartet> out;
+    out.reserve(n);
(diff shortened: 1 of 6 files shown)


psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,192 @@
...
+        iwl_buf_fetch(&B);
+        lastbuf = B.lastbuf;
+    }
+    iwl_buf_close(&B, /*keep*/ 0);
+    return out;
+}
+
+bool quartets_equal(const Quartet &a, const Quartet &b) {
+    return a.p == b.p && a.q == b.q && a.r == b.r && a.s == b.s &&
+           std::fabs(a.value - b.value) < 1.0e-12;
```
TiborGY: All IWL does here with these numbers is store them, write them to a file etc, right? Then input/output should match exactly, or at least within cutoff (1e-14 here), no?
```suggestion
           std::fabs(a.value - b.value) < 1.0e-14;
```

In test cases where the values written are below the cutoff it should be tested what gets read back, I am guessing it all should be exactly +0.0?

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,192 @@
...
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+// Standalone round-trip test for libiwl. Exercises every bucket-boundary case
+// the 1995 README warned about: empty, single, exactly one bucket, just over
+// one bucket, mid-bucket termination, and all-zeros sweeps.
```
TiborGY: `all-zeros sweeps`
Where does that get tested here? Also I would like a test for what happens when some of the values but not all are below cutoff, but non-zero. Maybe these:

1. The first bucket is all below cutoff
2. The second bucket is all below cutoff
3. The last bucket is all below cutoff
4. The second to last bucket is all below cutoff

TiborGY (changes requested):

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
susilehtola: Agreed, and I went one further than the suggestion: the comparison is now exact (`a.value == b.value`).

IWL neither scales nor rounds — `write_value()` casts to `Value`, which is `typedef double`, so a no-op, and `put()`/`fetch()` move the raw bytes through libpsio. A surviving integral has to come back bit-identical, so any tolerance at all could only hide corruption.

On what happens to sub-cutoff values: they aren't read back as `+0.0` — they aren't read back at all. `write_value()` only stores the integral when `std::fabs(value) > cutoff_`, so a sub-cutoff entry is dropped outright and no slot is ever allocated for it. That's now covered explicitly (see the other thread).

To confirm the exact comparison actually bites, I mutated the library to scale stored values by `(1 + 1e-15)` — comfortably inside the old 1e-12 tolerance — and 10 cases fail.

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
susilehtola: Good catch — it didn't get tested anywhere. That phrase was left over from an earlier draft and described coverage that never existed; I've dropped it and added the real thing.

Because filtering *drops* entries rather than zeroing them, survivors slide forward into earlier bucket slots, so where the sub-cutoff run sits changes the packing downstream of it. That's what the new cases pin down, over a four-bucket write:

- first bucket entirely sub-cutoff
- second bucket entirely sub-cutoff
- second-to-last bucket entirely sub-cutoff
- last bucket entirely sub-cutoff
- *all* buckets sub-cutoff (the genuine all-zeros sweep — expects zero survivors, mirroring the empty-buffer case)
- a run straddling a bucket boundary, so survivors on both sides repack into the same bucket

Each expects only the survivors back, in order and bit-exact, with no placeholder and no gap. The first and last cases also exercise "the very first thing written is dropped" and "the file ends on a dropped run", where an off-by-one in `flush`/`lastbuf` would surface.

The sub-cutoff stand-ins cycle three kinds: exactly `+0.0`, `±1e-20`, and a value *exactly equal* to the cutoff — the last because `write_value()` tests strictly greater-than, so `|value| == cutoff` must drop too, and that boundary was previously untested.

Verified by mutation rather than by assumption: relaxing the cutoff test from `>` to `>=` fails all six new cases.

susilehtola:
Both review points are addressed in 17042da, and #3466 has been rebased on top so its copy of the test (which additionally covers the `Label` overflow throw) carries the same changes.

Verification, since neither change is worth much untested:

| build | cases | result |
|---|---|---|
| this PR's test vs. current libiwl | 12 | all pass |
| #3466's test vs. modernized libiwl | 13 | all pass |
| mutation: cutoff `>` → `>=` | | 6 filtering cases fail |
| mutation: stored values × (1 + 1e-15) | | 10 cases fail under exact compare |

The two mutations are there to show the new assertions can actually fail — a green test that cannot go red isn't coverage.

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+
+using namespace psi;
+
+struct Quartet {
+    int p, q, r, s;
+    double value;
+};
+
+// Deterministic pseudo-random test data with bounded indices.
+std::vector<Quartet> make_test_data(std::size_t n) {
```
TiborGY:
```suggestion
std::vector<Quartet> make_test_data(const std::size_t n) {
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+        out.push_back({p, q, r, s, v});
+    }
+    return out;
+}
+
+// Sub-cutoff stand-ins. Exactly +0.0 is the common case in a real integrals
+// sweep, but a nonzero magnitude below the cutoff must be dropped just the
+// same, and |value| == cutoff must *also* drop because write_value() tests
+// strictly greater-than. Cycling all three keeps that boundary honest.
+double subcutoff_value(std::size_t i) {
```
TiborGY:
```suggestion
double subcutoff_value(const std::size_t i) {
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+// casts to Value (= double, so a no-op) and put()/fetch() move the raw bytes
+// through libpsio. A surviving integral must therefore come back bit-identical,
+// and any tolerance here would only hide a real corruption.
+bool quartets_equal(const Quartet &a, const Quartet &b) {
+    return a.p == b.p && a.q == b.q && a.r == b.r && a.s == b.s && a.value == b.value;
+}
+
+// Run one write/read round-trip for the given size and tape unit. Returns
+// true on success, false otherwise (with diagnostic to stderr).
+bool round_trip(std::size_t n, int itap) {
```
TiborGY:
```suggestion
bool round_trip(const std::size_t n, const int itap) {
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+    auto data = make_test_data(n);
+
+    {
+        psi::IWL buf(psi::_default_psio_lib_.get(), itap, 1.0e-14,
+                     /*oldfile*/ 0, /*readflag*/ 0);
```
TiborGY: This does not look like it has been clang-formatted. Please confirm that the choice of formatting is intentional.

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+// through libpsio. A surviving integral must therefore come back bit-identical,
+// and any tolerance here would only hide a real corruption.
+bool quartets_equal(const Quartet &a, const Quartet &b) {
+    return a.p == b.p && a.q == b.q && a.r == b.r && a.s == b.s && a.value == b.value;
+}
+
+// Run one write/read round-trip for the given size and tape unit. Returns
+// true on success, false otherwise (with diagnostic to stderr).
+bool round_trip(std::size_t n, int itap) {
+    auto data = make_test_data(n);
```
TiborGY:
```suggestion
    const auto data = make_test_data(n);
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+
+    {
+        psi::IWL buf(psi::_default_psio_lib_.get(), itap, 1.0e-14,
+                     /*oldfile*/ 0, /*readflag*/ 0);
+        write_all(buf, data);
+        buf.set_keep_flag(true);
+    }
+
+    auto got = read_all(itap);
```
TiborGY:
```suggestion
    const auto got = read_all(itap);
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+            return false;
+        }
+    }
+    return true;
+}
+
+// Empty buffer: no integrals written, but flush(lastbuf=1) must still produce
+// a readable file that read_all() returns as zero entries. The 1995 README
+// specifically calls out this case.
+bool round_trip_empty(int itap) {
```
TiborGY:
```suggestion
bool round_trip_empty(const int itap) {
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+    {
+        psi::IWL buf(psi::_default_psio_lib_.get(), itap, 1.0e-14,
+                     /*oldfile*/ 0, /*readflag*/ 0);
+        buf.flush(/*lastbuf*/ 1);
+        buf.set_keep_flag(true);
+    }
+    auto got = read_all(itap);
```
TiborGY:
```suggestion
    const auto got = read_all(itap);
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+    if (!got.empty()) {
+        std::cerr << "  empty round-trip returned " << got.size()
+                  << " entries\n";
```
TiborGY: This could probably be formatted to be 1 line

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+        return false;
+    }
+    return true;
+}
+
+// Round-trip N integrals of which a contiguous run [lo, hi) is pushed below the
+// cutoff. Only the survivors should come back, in order and bit-exact; the
+// dropped run must leave no trace (no zero-valued placeholder, no gap).
+bool round_trip_filtered(std::size_t n, std::size_t lo, std::size_t hi, int itap, const std::string &what) {
```
TiborGY:
```suggestion
bool round_trip_filtered(const std::size_t n, const std::size_t lo, const std::size_t hi, const int itap, const std::string &what) {
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+    auto data = make_test_data(n);
```
TiborGY, line 234:
```suggestion
    const auto data = make_test_data(n);
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    {
+        psi::IWL buf(psi::_default_psio_lib_.get(), itap, 1.0e-14,
+                     /*oldfile*/ 0, /*readflag*/ 0);
+        write_all(buf, data);
+        buf.set_keep_flag(true);
+    }
+
+    auto got = read_all(itap);
```
TiborGY:
```suggestion
    const  auto got = read_all(itap);
```

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
```diff
@@ -0,0 +1,281 @@
...
+    // libpsio needs init before anything opens a unit.
+    psi::psio_init();
+
+    // Use tape unit numbers in a range unlikely to clash with the few entries
+    // libpsio's filecfg might preset.
+    const int base_unit = 80;
+    int unit = base_unit;
+    int failures = 0;
+
+    const std::size_t B = static_cast<std::size_t>(IWL_INTS_PER_BUF);
```
TiborGY: Why do we need the static cast here?
Could this be constexpr instead of const?

TiborGY (changes requested):
Current tests seem to excercize C++ write API and C read API. Could it also test the C write and C++ read APIs?

susilehtola:
Thanks @TiborGY — addressed in e28efcb84:

- **C write / C++ read coverage.** Factored the loops into `write_all_cpp`/`write_all_c` and `read_all_c`/`read_all_cpp`, and every case now runs through **all four** {C, C++} write × {C, C++} read combinations — so a byte written through either interface is shown to read back bit-identically through either. Each reader deletes its scratch file, so the tape-unit range is reused per API pass.
- **const-correctness.** By-value parameters (`n`, `itap`, `lo`, `hi`, `i`) and the local `data`/`got` bindings are now `const` / `const auto`.
- **line 237 `static_cast`.** Dropped it; the buffer size is now `constexpr std::size_t B = IWL_INTS_PER_BUF;` (`IWL_INTS_PER_BUF` is a literal, so the widening to `size_t` is implicit and compile-time).
- **clang-format.** Ran clang-format over the file (picks up the two flagged spots).

Verified locally: all cases pass under all four API combinations ("All round-trip cases passed.").

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
TiborGY: It looks like this comment was...skipped?

TiborGY (approved):
LGTM assuming my one remaining comment receives a response and all the CI lanes are green.

psi4/src/psi4/libiwl/test/iwl_roundtrip_test.cc
susilehtola: Not skipped — deliberately left non-`const` here, and this is the one spot where the suggestion cannot apply. This `data` gets mutated a few lines down: the filtering loop reassigns `data[i].value = subcutoff_value(i)` for the run `[lo, hi)` to push those integrals under the cutoff before writing. The other two `data` bindings (in `round_trip`/`round_trip_empty`) are write-once and are `const auto`; this one is not, because it has to be edited in place to build the sub-cutoff run. The `got` binding just below (line 245) is `const auto` as suggested.

I could split it into a `const` original plus a separate mutated copy, but that just adds a copy to satisfy the keyword without making anything safer. Happy to do so if you prefer it for consistency.

TiborGY: OK, thanks, it's OK as-is.

TiborGY:
> ## Description
> 
> `test/iwl_roundtrip_test.cc` writes integrals through the C++ `IWL` API and reads them back through the C-style `iwl_buf_*` API ...

As the PR has evolved beyond its initial state, the description needs updating.
That, plus a rebase to heal CI, and this PR should be all set for a second reviewer and merge.

psi4/src/psi4/libiwl/test/stubs.cc
```diff
@@ -0,0 +1,87 @@
...
+
+void DPD::file4_cache_del_filenum(std::size_t) {}
+
+namespace {
+DPD stub_dpd_instance;
+}
+
+DPD *global_dpd_ = &stub_dpd_instance;
+
+// libciomr's DSYEV_ascending lands in the same unity translation unit as the
```
loriab, line 79: If this continues to be a problem, we can define a separate unity group for the pieces these unit tests need.

loriab (approved):
Really nice to be finally unit testing this!

PR #3466, opened 2026-07-18
libiwl: phase 1 modernization (collapse dup, overflow check, RAII)

susilehtola:
Phase 1 of the libiwl modernization tracked in #3425. An internal refactor of `psi4/src/psi4/libiwl/` with **no caller-facing API change and no on-disk format change**. The ~30 in-tree consumers that poke `struct iwlbuf` fields directly all keep compiling unchanged.

Stacked on #3465 (the round-trip test), which acts as the behavior-preserving guard for this refactor. This is the "modernization respun on top of the tests" half of the #3426 split TiborGY requested; **#3426 is superseded by #3465 + this PR** and will be closed.

## Description
The libiwl bucket reader/writer had two byte-identical bodies for every operation (a C `iwl_buf_*` free function and a C++ `IWL::` method), C-style macros, mixed `malloc`/`new[]` ownership, and silent 32-bit→16-bit narrowing at the `Label` cast that could corrupt a bucket file without warning. This consolidates the duplication and fixes the latent correctness/ownership issues, without touching the public field layout or the on-disk format.

- [x] `Label` overflow (orbital index exceeding the 16-bit bucket field) now throws a `PSIEXCEPTION` naming the offending indices instead of silently narrowing and corrupting the IWL file.

## Dev notes & details
- `config.h`: `#define IWL_INTS_PER_BUF` / `IWL_KEY_*` → `inline constexpr`; document the bucket layout.
- `iwl.hpp`: remove dead declarations `IWL::read` / `IWL::to_end` (never defined); delete copy ctor / assign; document the deferred-init pattern. Field comments retained.
- Collapse every duplicated C-function / C++-method body into one shared static helper per operation; new internal `iwl_impl.h` holds `check_label_fits()` / `make_printer()`.
- Overflow check at every `(int) → Label` cast in the three write paths (throw vs silent narrowing).
- `buf_init`: throw on `tocscan` failure instead of `Printf`'ing and leaving a half-initialized buffer.
- Unify allocators on `new[]`/`delete[]` (removes the cross-API ownership UB).
- Hoist per-integral `PsiOutStream` construction to one per-call instance, only when `printflag` is set.
- Default-construct all `IWL` members; drop `MAX0`/`MIN0`/`INDEX` macros in `buf_wrt_mat.cc` for `std::max`/`std::min` and an inline `packed_index()`.
- The round-trip test gains the `Label`-overflow case (verifying the new throw) and switches to the `psi::IWL_INTS_PER_BUF` constexpr.

## What this PR does *not* do
- No change to `struct iwlbuf`'s field layout; direct field pokes keep compiling.
- No on-disk format change; the v1 bucket layout is preserved bit-for-bit.
- The 16-bit orbital-index cap is unchanged; only the failure mode changes (loud throw vs silent corruption). 32-bit is phase 3.
- Iterator-based `IWLReader`/`IWLWriter` + caller migration is phase 2 (follow-up). See #3425.

## Questions
- [ ] None outstanding.

## AI
- [x] AI usage: refactor and bug-seeking (the `Label` overflow and allocator-mismatch issues).

## Checklist
- [x] Tests added for any new features (the overflow case in #3465's round-trip test)
- [ ] Docstring or narrative docs/sphinxman/source updated if needed (internal refactor)
- Test running is well covered by CI.

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libiwl/CMakeLists.txt
+++ b/psi4/src/psi4/libiwl/CMakeLists.txt
@@ -11,3 +11,19 @@ list(APPEND sources
   wrtone.cc
   )
 psi4_add_module(lib iwl sources)
+
+# test/ contains a standalone round-trip exerciser (iwl_roundtrip_test.cc).
+# It links against the libiwl/libpsio/libpsi4util/liboptions/libciomr static
+# archives plus a tiny stubs TU (test/stubs.cc) that supplies the runtime
+# globals normally provided by core.cc / libdpd. That curated link resolves with
+# GNU/Clang on Linux/macOS but not with clang-cl/lld-link, which emits inline and
+# implicit functions that GCC discards as unused, so the Windows objects carry
+# undefined references the link line does not provide: psio.lib wants
+# DPDMOSpace::~DPDMOSpace (instantiated by std::vector<DPDMOSpace> in dpd.h), and
+# psi4util.lib wants Molecule::Molecule(const Molecule&) (from inline
+# Molecule::clone) and ExternalPotential::print (from inline py_print). Not a
+# unity or PCH artifact -- Linux is unity-built too, and Azure Windows builds with
+# PCH off. Restrict the test to non-Windows rather than drag in the full closure.
+if(NOT WIN32)
+    add_subdirectory(test)
+endif()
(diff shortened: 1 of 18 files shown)


TiborGY (dismissed):
LGTM assuming my one remaining comment receives a response and all the CI lanes are green.

PR #3472, opened 2026-07-30
Support for Dimension range-based for loop

NathanGillispie:
## Description

Allows the following syntax:
```c++
for (auto& h : nmopi_) { h *= 2; }
```
Or for einsums-type quantities
```c++
std::vector<size_t> dim(nirreps_);
for (auto&& [h, d] : std::views::zip(nsopi_, dim)) {
    d = static_cast<size_t>(2*h);
}
ComplexMatrix D_{"Density", dim, dim};
```
This showed up in my ComplexWavefunction work a bit and I think it makes things a bit easier.

- [x] Added `Dimension::begin()` and `::end()`. Forwards to internal `blocks_.begin()` and `end()`. Allows for range-based for loops over Dimension objects.

## AI
None

## Checklist

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libmints/dimension.h
+++ b/psi4/src/psi4/libmints/dimension.h
@@ -116,6 +116,14 @@ class PSI_API Dimension {
 
     void print() const;
 
+    // Forward underlying vector's iterators
+    // This allows for range-based for loops:
+    //    for (auto h : nsopi_) { ... }
+    auto begin() { return blocks_.begin(); }
+    auto end() { return blocks_.end(); }
+    auto begin() const { return blocks_.begin(); }
+    auto end() const { return blocks_.end(); }
+
     // Only used for python
     const int& get(size_t i) const { return blocks_[i]; }
     void set(size_t i, int val) { blocks_[i] = val; }
--- a/psi4/src/psi4/libmints/dimension.h
+++ b/psi4/src/psi4/libmints/dimension.h
@@ -119,10 +119,10 @@ class PSI_API Dimension {
     // Forward underlying vector's iterators
     // This allows for range-based for loops:
     //    for (auto h : nsopi_) { ... }
-    auto begin() { return blocks_.begin(); }
-    auto end() { return blocks_.end(); }
-    auto begin() const { return blocks_.begin(); }
-    auto end() const { return blocks_.end(); }
+    auto begin() noexcept { return blocks_.begin(); }
+    auto end() noexcept { return blocks_.end(); }
+    auto begin() const noexcept { return blocks_.begin(); }
+    auto end() const noexcept { return blocks_.end(); }
 
     // Only used for python
     const int& get(size_t i) const { return blocks_[i]; }


susilehtola (approved):

psi4/src/psi4/libmints/dimension.h
```diff
@@ -116,6 +116,14 @@ class PSI_API Dimension {
 
     void print() const;
 
+    // Forward underlying vector's iterators
+    // This allows for range-based for loops:
+    //    for (auto h : nsopi_) { ... }
+    auto begin() { return blocks_.begin(); }
+    auto end() { return blocks_.end(); }
+    auto begin() const { return blocks_.begin(); }
+    auto end() const { return blocks_.end(); }
```
susilehtola: Wouldn't the single `const` declaration suffice?

NathanGillispie: It would not compile if someone tries to iterate over a const instance/reference of Dimension. For the first example I showed, not applicable. But it's not unreasonable for someone to use the second example within a function where someone has passed in `const Dimension& occpi`.

NathanGillispie: Oh I misunderstood your question. I think it wouldn't work to have just const because in the first example I modify the data, which is not allowed within const functions unless `blocks_` is mutable.

TiborGY: If I am reading the standard reference correctly (https://en.cppreference.com/cpp/container/vector/end) these should be both `constexpr` and `noexcept`.

NathanGillispie: yeah that appears to be the case. I assume in all cases the compiler would inline these, but it doesn't hurt to be explicit.

NathanGillispie: So after looking into it a bit more, adding constexpr should have no effect because there is no constexpr constructor for Dimension. You need a constexpr instance to call a non-static member function in a compile-time context. As far as I can tell, nobody creates Dimension objects at compile-time.

`noexcept` could be added, but the effect would be infinitesimal.

TiborGY: Thanks for investigating. Once CI is healed I am happy to merge this.

TiborGY (approved):

TiborGY:
@NathanGillispie Please rebase to heal CI.
Also, the merge queue would currently reject this due to "Tests added for any new features" not being checked. If you are not planning to add any tests that would exercise these functions, you can remove this line from the description.
And in the "User API & Changelog headlines" you should list the addition of these functions to the C++ API, since we are adding member functions to a class marked as PSI_API.

TiborGY (changes requested):
GH description update

NathanGillispie:
> you should list the addition of these functions to the C++ API, since we are adding member functions to a class marked as PSI_API.

Gotcha, I will work on that. I never knew exactly what that meant before.

TiborGY (approved):

TiborGY:
@NathanGillispie It looks like azure is being flaky again for no apparent reason...
Feel free to retrigger CI by pushing a trivial change (eg. one whitespace or something in the comment)

NathanGillispie:
@TiborGY You can use `git commit --allow-empty -m "CI"` for that, fun fact.

TiborGY:
@NathanGillispie This is getting a bit ridiculous, but one lane healed and two more broke now on azure...so retry i guess?

PR #3478, opened 2026-08-10
Thread the density-fitted and one-electron nuclear-Hessian integral loops

susilehtola:
## What

The analytic nuclear Hessian evaluated most of its integral work **serially on a single core**, even though the sibling gradient code has long been threaded. This PR threads those loops, mirroring the established `DFJKGrad::compute_gradient` pattern (per-thread integral engines + a parallel loop over auxiliary/primary shells, with per-thread accumulators reduced afterwards). No math changes — only parallelization.

Threaded regions:

1. **DF two-electron second-derivative Hessian** (`DFJKGrad::compute_hessian`, `jk_grad.cc`) — the dominant cost. All five auxiliary-shell loops (the `(A|mn)` build, the first/second-derivative `(A|mn)^x`/`(A|mn)^xy` loops, and the metric `(A|B)^x`/`(A|B)^xy` loops) now run in parallel. Loops writing aux-partitioned intermediates are race-free by construction; the deriv1 K-scratch is made per-thread; the two `compute_shell_deriv2` loops accumulate into per-thread `JHess`/`KHess` reduced into `hessians_` before the serial tail.

2. **DF orbital-response (CPHF) perturbed-Fock build** (`RSCFDeriv::hessian_response` + `USCFDeriv::JK_deriv1`, `response.cc`) — the `(A|mn)` build plus the metric and main `(A|mn)^x` contraction loops. These scatter into per-perturbation Fock matrices keyed by shell centers (not partitioned by P), so each thread scatters into a **private copy of the perturbation batch** that is reduced afterwards. The batch size `max_a` is re-budgeted so the per-thread copies fit inside the memory budget (`df_mem / ((1+nthreads)·nso²)`, capped at `3·natom`).

3. **One-electron Hessian terms** (`SCFDeriv::compute_hessian`, `scf_grad.cc`) — the Potential, Kinetic, and Overlap shell-pair loops, with per-thread one-electron engines and a per-thread accumulator. (ECP left serial — rare, differently structured.)

## Bonus bug fix

While threading (2), fixed a **pre-existing latent fault** in the multi-batch response path: the loops mark a whole atom in-core when *any* of its perturbations falls in a batch and then write all three of its Fock derivatives, so a batch boundary that split an atom wrote into an unmapped accumulator (segfault). `max_a` is now forced to a multiple of 3 (atom-aligned).

## Validation

Analytic DF Hessians (def2-SVP, `scf_type df`) computed at 1 vs 6 threads, RHF and UHF:

- Small molecules (H2O, NH2), single batch: agreement to **1e-10 … 6e-10**.
- Multi-batch reduction isolated at fixed thread count (single vs multi batch): **~7e-9** — the same magnitude as two *identical* runs of the same configuration (**~1e-8** for a water tetramer), i.e. pure floating-point summation-order noise from the threaded reduction (dynamic scheduling), not a correctness difference. This is the same run-to-run nondeterminism already present in psi4's other threaded quadrature/JK code, and is physically negligible for a Hessian.

Normal-memory runs use a single batch (`max_a` capped at `3·natom`), so results there are identical to the previous serial code up to threading FP noise; existing Hessian tests are unaffected.

### DIFF
--- a/psi4/src/psi4/scfgrad/jk_grad.cc
+++ b/psi4/src/psi4/scfgrad/jk_grad.cc
@@ -1342,8 +1342,17 @@ void DFJKGrad::compute_hessian() {
     auto Pmnfactory = std::make_shared<IntegralFactory>(auxiliary_, BasisSet::zero_ao_basis_set(), primary_, primary_);
     auto PQfactory = std::make_shared<IntegralFactory>(auxiliary_, BasisSet::zero_ao_basis_set(), auxiliary_,
                                                        BasisSet::zero_ao_basis_set());
-    std::shared_ptr<TwoBodyAOInt> Pmnint(Pmnfactory->eri(2));
-    std::shared_ptr<TwoBodyAOInt> PQint(PQfactory->eri(2));
+    // Per-thread integral engines (the deriv-2 object also serves the deriv0/1
+    // calls). Mirrors DFJKGrad::compute_gradient so the shell loops below can
+    // be threaded over the auxiliary index.
+    int nthread = df_ints_num_threads_;
+    std::vector<std::shared_ptr<TwoBodyAOInt>> Pmnints(nthread), PQints(nthread);
+    Pmnints[0] = std::shared_ptr<TwoBodyAOInt>(Pmnfactory->eri(2));
+    PQints[0] = std::shared_ptr<TwoBodyAOInt>(PQfactory->eri(2));
+    for (int t = 1; t < nthread; t++) {
+        Pmnints[t] = std::shared_ptr<TwoBodyAOInt>(Pmnints[0]->clone());
+        PQints[t] = std::shared_ptr<TwoBodyAOInt>(PQints[0]->clone());
+    }
     auto Amn = std::make_shared<Matrix>("(A|mn)", np, nso*nso);
     auto Aa_mi = std::make_shared<Matrix>("(A|mi)", np, nso*na);
     auto Aa_ij = std::make_shared<Matrix>("(A|ij)", np, na*na);
@@ -1372,7 +1381,12 @@ void DFJKGrad::compute_hessian() {
     double **Bb_mnp = Bb_mn->pointer();
     double **Db_PQp = Db_PQ->pointer();
 
+#pragma omp parallel for schedule(dynamic) num_threads(nthread)
     for (int P = 0; P < nauxshell; ++P) {
+        int thread = 0;
+#ifdef _OPENMP
+        thread = omp_get_thread_num();
+#endif
         int nP = auxiliary_->shell(P).nfunction();
         int oP = auxiliary_->shell(P).function_index();
         for (int M = 0; M < nshell; ++M) {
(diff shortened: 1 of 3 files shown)


susilehtola:
Heads up on the red Azure/"Propose Docs" checks: they are **not** caused by this PR. The Azure build fails to compile `export_fock.cc` (the `SADGuess.build_SAD` binding) under pybind11 3.x — a pre-existing incompatibility on master that this PR does not touch (it only changes three `scfgrad` files, and its own Eco build matrix is green). Same failure hits other open PRs. Fixed separately in #3480; once that merges to master, this PR’s merge build should go green.

loriab:
In case anyone (or your AI) was concerned about nested OMP directives, here's an OpenAI report:

     I found that the full PR version is best on your “believable” scale (benzene analytic DF Hessian ~150s baseline), and a small modification I tested does not materially improve it.                                                    
                                                                                                                                                   
     Method                                                                                                                                        
     - Bench input: objdir_py314_nextd/profile_df_hess_rhf_benzene.dat (RHF, DF, 6-31G*, analytic                                                  
     Hessian, C1 benzene), baseline runtime ~150s at 1 thread.                                                                                     
     - Critical detail: must use psi4 -n N; setting only OMP_NUM_THREADS left Psi4 at 1 thread.                                                    
     - BLAS pinned single-thread for consistency (MKL/OPENBLAS/BLIS_NUM_THREADS=1).                                                                
     - Rebuilt between variants by swapping only:                                                                                                  
       - psi4/src/psi4/scfgrad/scf_grad.cc (region 1)                                                                                              
       - psi4/src/psi4/scfgrad/jk_grad.cc (region 2)                                                                                               
       - psi4/src/psi4/scfgrad/response.cc (region 3)                                                                                              
     - “Before PR” commit: a260858888; “PR head” commit: 2cf42e372e.                                                                               
                                                                                                                                                   
     Three region impact (isolated)                                                                                                                
     - base000 (before PR in all 3): t1/t4/t8 = 148.9 / 94.0 / 87.8 s (speedup at 8 threads: 1.69x)                                                
     - r1_100 (only region 1 from PR): 152.3 / 96.9 / 80.1 s (8-thread: 1.90x)                                                                     
     - r2_010 (only region 2 from PR): 152.5 / 73.3 / 63.1 s (8-thread: 2.42x)                                                                     
     - r3_001 (only region 3 from PR): 147.2 / 60.6 / 47.8 s (8-thread: 3.08x) 
         
     Combinations                                                                                                                                  
     - r12_110: 142.1 / 74.8 / 62.0 s                                  
     - r13_101: 143.0 / 58.4 / 44.6 s                        
     - r23_011: 139.9 / 43.5 / 26.0 s                              
     - all111 (full PR): 151.2 / 44.8 / 26.4 s (8-thread: 5.73x)                          
                                                                                                                                                   
     Interpretation:                       
     - Region 3 gives the largest single uplift.   
     - Region 2 is the next biggest.                                                                                                               
     - Region 1 is smaller, but not harmful in full combination.                                                                                   
     - Full PR threading is dramatically better than pre-PR for multi-thread performance.                                                          
                                                                                                                                                   
     “Modify PR?” check                                                                                                                            
     - I tested a concrete tweak: removing nested inner #pragma omp parallel for sites in jk_grad.cc                                               
     inside already-threaded outer loops.                                                                                                          
     - Result vs PR head was effectively noise-level:                                                                                              
       - Modified: t1/t4/t8 = 142.6 / 44.2 / 26.5 s                                                                                                
       - PR head rerun: 143.0 / 44.3 / 26.4 s                                                                                                      
     - So this modification does not meaningfully beat current PR.                                                                                 
                                                                                                                                                   
     Correctness sanity                                                                                                                            
     - Hessian numeric differences across thread counts were tiny (~1e-8 max abs element diff),                                                    
     consistent with reduction-order FP noise.                                                                                                     
     - Energies matched for the benchmarked runs.

loriab (approved):

fevangelista (approved):

PR #3485, opened 2026-08-14
use v1 conda compilers for now

loriab:
Conda v2 compiler metapackages are designed to be easier to use _outside_ conda-build, with names like `g++`, not `x86_64-conda-linux-gnu-c++`, and fewer enironment variables, and less (no?) flag injection. However, this does throw a wrench in the path-advisor handling that was closely attended to the envvar the v1 metapackages set. I'll have to take a closer look at changes for v2, but this should get us past the immediate CI crisis.

### DIFF
--- a/.github/workflows/ecosystem.yml
+++ b/.github/workflows/ecosystem.yml
@@ -244,7 +244,7 @@ jobs:
         echo "::endgroup::"
 
     - name: Setup Conda
-      uses: conda-incubator/setup-miniconda@v4
+      uses: conda-incubator/setup-miniconda@v3
       with:
         activate-environment: p4build
         environment-file: env_p4build.yaml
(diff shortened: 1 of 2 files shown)


NathanGillispie (approved):
I was able to fix it by setting the `CONDA_TOOLCHAIN_HOST` env [directly](https://github.com/NathanGillispie/psi4/pull/32/changes/aad1d245603adacb096bfb31f7214e1985dfca96)... but that doesn't feel like a long term solution.

loriab:
> I was able to fix it by setting the CONDA_TOOLCHAIN_HOST env [directly](https://github.com/NathanGillispie/psi4/pull/32/changes/aad1d245603adacb096bfb31f7214e1985dfca96)... but that doesn't feel like a long term solution.

I agree, esp. with those compiler versions for Mac. And I'd like to keep the script working for local users. Thanks for the triplets

TiborGY (approved):
I have not reviewed exactly what the v2 compilers will do, but since we are just constraining the selection to what was previously the only option (v1), I am rubber-stamping this to get CI back on its feet.

andyj10224 (approved):
LGTM!

PR #3486, opened 2026-08-14
Refactor `DirectJK::build_JK_matrices` for clarity using range helpers

NathanGillispie:
## Description

Refactors the `DIRECT` JK build. Replaces cryptic, index-heavy task blocking with meaningful names and range helpers. No algorithmic change was made.

When typing up the ComplexDirectJK for the CGHF work, I found the existing DirectJK inscrutible. Since I will eventually open a PR for that, I decided to apply the same fixes to the real side, so both appear more similar.

*None*
## Dev notes & details

- [x] One of the biggest improvements is in the anonymous namespace helpers `partition` and `partition_with_idx`. These turn the atom-to-shell boundary array (previously `task_starts`, now `atom_to_shell`) into an interable of shell ranges with optional center index. 

For example, consider the following refactor. \[Names have been changed to appear similar\]
```diff
-for (size_t task = 0; task < ntask; task++) {
+for (auto tasks : partition(task_starts)) {
     size_t size = 0L;
-    for (int P2 = task_starts[task]; P2 < task_starts[task + 1]; P2++) {
-        size += primary_->shell(task_shells[P2]).nfunction();
+    for (size_t shell : tasks) {
+        size += primary_->shell(shell).nfunction();
```
- [x] The previous ubiquitous `task_starts[Ptask]..task_starts[Ptask+1]` loop indexing is completely replaced by a much nicer syntax using the `shells_on_center` lambda.

For example, the following code
```cpp
for (int P2 = P2start; P2 < P2start + nPtask; P2++) {
    for (int Q2 = Q2start; Q2 < Q2start + nQtask; Q2++) {
        int P = task_shells[P2];
        int Q = task_shells[Q2];
```
Can be simplified into
```cpp
for (int P : shells_on_center(Ptask)) {
    for (int Q : shells_on_center(Qtask)) {
```
- [x] A few in-function lambdas were written to encapsulate the atom-shell-basis index mapping so that contraction and stripe-out blocks can stop re-deriving the indices inline. Best of all, `basis_index_of_shell_from_atom` provides the "local orbital" offset within a center. It replaces the old `Poff2 = task_offsets[P2] - task_offsets[P2start]` mess.

If this diff doesn't convince you, I don't know what will. All the `task_starts` and `task_offsets` indexing is handled already, eliminating the need for these variables.
```diff
-        int nPtask = task_starts[Ptask + 1] - task_starts[Ptask];
-        int nQtask = task_starts[Qtask + 1] - task_starts[Qtask];
-        int nRtask = task_starts[Rtask + 1] - task_starts[Rtask];
-        int nStask = task_starts[Stask + 1] - task_starts[Stask];
 
-        int P2start = task_starts[Ptask];
-        int Q2start = task_starts[Qtask];
-        int R2start = task_starts[Rtask];
-        int S2start = task_starts[Stask];
-
-        int dPsize = task_offsets[P2start + nPtask] - task_offsets[P2start];
-        int dQsize = task_offsets[Q2start + nQtask] - task_offsets[Q2start];
-        int dRsize = task_offsets[R2start + nRtask] - task_offsets[R2start];
-        int dSsize = task_offsets[S2start + nStask] - task_offsets[S2start];
+        size_t Patom_size = nfunctions_on_center(Patom);
+        size_t Qatom_size = nfunctions_on_center(Qatom);
+        size_t Ratom_size = nfunctions_on_center(Ratom);
+        size_t Satom_size = nfunctions_on_center(Satom);
```
- [x] `nfunctions_on_center` is the per-center block size. It replaces the old `dPsize` strides.
- [x] `shells_on_center` returns a shell-index view for the given center. Ex. above.
- [x] Variable names are more meaningful. `task_starts` was renamed to `atom_to_shell` because you provide an atom index and the list returns the index of the shell starting that atom. Likewise for `task_offsets` --> `shell_to_basis`.
- [x] Global/local offsets become `*ao` and `*lo`. Task indices become `[PQRS]atom`. This makes it more clear what basis/index is being used where.

For example, in the ERI contraction, It is now obvious that `Jp` operates on the whole of `J` and `J1p` operates on an atom block only.
```diff
- Jp[p + Poff][q + Qoff] += J1p[(p + Poff2) * dQsize + q + Qoff2];
+ Jp[p + Pao][q + Qao] += J1p[(p + Plo) * Qatom_size + q + Qlo];
```
- [x] The (mostly identical) changes to the ComplexDirectJK side were verified by unit tests. The changes here were verified by running the test suite with `SCF_TYPE` default changed to `DIRECT`. The only errors were with `DF_BASIS_DCT` and once when the literal `SCF_TYPE` option was asserted, both of which were unrelated to the built JK matrices.
## AI

- [x] AI was used to explore modern C++ solutions to the indexing problems. Actual implementation was done by hand.
## Checklist
- [x] ~~Tests added for any new features~~ \[*N/A*\]
- [x] ~~Docstring or narrative docs/sphinxman/source updated.~~ \[*N/A*\]
## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libfock/DirectJK.cc
+++ b/psi4/src/psi4/libfock/DirectJK.cc
@@ -50,7 +50,6 @@
 #include <span>
 #include <sstream>
 #include <unordered_set>
-#include <utility>
 #include "psi4/libpsi4util/PsiOutStream.h"
 #ifdef _OPENMP
 #include <omp.h>
@@ -425,20 +424,20 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
     bool build_K = (!K.empty());
 
     if (!build_J && !build_K) return;
-
+    
     timer_on("build_JK_matrices()");
 
     // => Zeroing... <= //
-
+    
     // Ideally, this wouldnt be here at all
     // It would be better covered in incfock_setup()
     // But removing this causes a couple of tests to fail for some reason
-
+    
     if (!do_incfock_iter_) {
         for (auto& Jmat : J) {
             Jmat->zero();
         }
-
+    
         for (auto& Kmat : K) {
             Kmat->zero();
         }
@@ -498,10 +497,10 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
     // => Welcome to the jungle <= //
 
     {
-        int atomic_ind = -1;
+        shell_to_basis.push_back(0);
 
         size_t total_nfuncs = 0;
-        shell_to_basis.push_back(0);
+        int atomic_ind = -1;
         for (int P = 0; P < nshell; P++) {
             const auto& shell = primary_->shell(P);
 
@@ -563,7 +562,6 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
         }
     }
     size_t n_pair = significant_pairs.size();
-    size_t n_pair2 = n_pair * n_pair;
 
     // => Intermediate Buffers <= //
 
@@ -584,7 +582,7 @@ void DirectJK::build_JK_matrices(std::vector<std::shared_ptr<TwoBodyAOInt>>& int
     std::vector<std::vector<SharedMatrix>> KT;
     if (build_K) {
         for (int thread = 0; thread < nthread; thread++) {
-            std::vector<SharedMatrix> K2;
+            std::vector<SharedMatrix > K2;
             for (size_t ind = 0; ind < D.size(); ind++) {
                 // The factor of 4 or 8 comes from exploiting ERI permutational symmetry
                 K2.push_back(std::make_shared<Matrix>("KT", (lr_symmetric_ ? 4 : 8) * max_nfuncs_per_center,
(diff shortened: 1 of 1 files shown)


NathanGillispie:
Looks like theres some sort of bug in the compiler/libstdc++ version for clang_latest. Doesn't appear for other clang build, not even a warning. I wonder if it would be okay to just roll with it and hope people don't land on that specific build combination. Either way, I'll do some more investigating, but any code change would require making a struct that recreates the functionality of `std::views:iota` which is easy, but I don't like that option.

NathanGillispie:
Okay it get's really complicated, but essentially it's unique to clang < 17 (15 fails for certain) combined with GCC < 14 (12 and 13 fail for sure). Even Ubuntu 24.04 LTS ships GCC 13, so making GCC >= 14 a hard constraint probably isn't a good idea. Anyone compiling with system stdlibc++ and clang on affected Ubuntu versions would run into this issue. That scenario isn't unlikely even for conda-installed clang.

I could pinpoint where the issue is occuring in the stdlibc++ headers, but not exactly why clang fails and gcc doesn't. Anyways, the rabbithole is too deep on this so I'll just replicate the iota struct.

NathanGillispie:
I guess I C++20'd too hard. Changing the title to be more boring 😔️

susilehtola:
C++20 might be too new a standard to be fully supported on all platforms...

PR #3493, opened 2026-08-18
Add CFMM and Density-Fitted CFMM For Linear-Scaling J Builds

andyj10224:
## Description

This PR revives and modernizes Psi4's Continuous Fast Multipole Method (CFMM) implementation from the earlier `local_df2` development branch and integrates it with the current `CompositeJK` infrastructure.

Two Coulomb-matrix algorithms are introduced:

* `CFMM`, which partitions the direct four-center Coulomb build into exact near-field interactions and multipole-expanded far-field interactions.
* `DFCFMM`, which applies the CFMM machinery to the three-center contractions required for the density-fitted Coulomb build.

These Coulomb builders can be combined with the existing CompositeJK exchange algorithms, including LinK, COSX, and sn-LinK. The implementation supports conventional, density-screened, and incremental Fock builds and includes tests for restricted, unrestricted, restricted open-shell, and hybrid-DFT calculations.

The implementation follows the original CFMM and DF-CFMM literature:

* C. A. White, B. G. Johnson, P. M. W. Gill, and M. Head-Gordon, “The continuous fast multipole method,” *Chem. Phys. Lett.* **230**, 8–16 (1994). [[doi:10.1016/0009-2614(94)01128-1](https://doi.org/10.1016/0009-2614(94)01128-1)](https://doi.org/10.1016/0009-2614%2894%2901128-1)
* C. A. White, B. G. Johnson, P. M. W. Gill, and M. Head-Gordon, “Linear scaling density functional calculations via the continuous fast multipole method,” *Chem. Phys. Lett.* **253**, 268–278 (1996). [[doi:10.1016/0009-2614(96)00175-3](https://doi.org/10.1016/0009-2614(96)00175-3)](https://doi.org/10.1016/0009-2614%2896%2900175-3)
* A. Sodt, J. E. Subotnik, and M. Head-Gordon, “Linear scaling density fitting,” *J. Chem. Phys.* **125**, 194109 (2006). [[doi:10.1063/1.2370949](https://doi.org/10.1063/1.2370949)](https://doi.org/10.1063/1.2370949)
* R. Łazarski, A. M. Burow, and M. Sierka, “Density Functional Theory for Molecular and Periodic Systems Using Density Fitting and Continuous Fast Multipole Methods,” *J. Chem. Theory Comput.* **11**, 3029–3041 (2015). [[doi:10.1021/acs.jctc.5b00252](https://doi.org/10.1021/acs.jctc.5b00252)](https://doi.org/10.1021/acs.jctc.5b00252)
* F. Weigend, “A fully direct RI-HF algorithm: Implementation, optimized auxiliary basis sets, demonstration of accuracy and efficiency,” *Phys. Chem. Chem. Phys.* **4**, 4285–4291 (2002). [[doi:10.1039/B204199P](https://doi.org/10.1039/B204199P)](https://doi.org/10.1039/B204199P)

* Adds `CFMM` and `DFCFMM` Coulomb-matrix algorithms to the CompositeJK framework.
* Adds composite `SCF_TYPE` choices such as `CFMM+LINK` and `DFCFMM+LINK`, together with corresponding COSX and sn-LinK combinations.
* Adds the `CFMM_ORDER` option controlling the maximum multipole-expansion order.
* Adds the `CFMM_GRAIN` option controlling the CFMM tree depth.
* Adds the `CFMM_EXTENT_TOLERANCE` option controlling Gaussian shell-pair extents.
* Enables automatic construction of the SCF auxiliary basis for `DFCFMM`.
* Supports density-based integral screening and incremental Fock builds.

## Dev notes & details

* Adds a new `libfmm` module containing:

  * the hierarchical CFMM tree and box structures;
  * primary- and auxiliary-basis shell-pair organization;
  * real solid harmonic multipole coefficients, translations, and contractions;
  * near-field exact-integral and far-field multipole contributions.
* Adds `DirectCFMM` and `DFCFMM` implementations derived from the `SplitJK` interface.
* Extends `CompositeJK` construction and dispatch to select the correct three- or four-center ERI engines for each Coulomb algorithm.
* Extends `TwoBodyAOInt` with accessors needed for bra/ket shell-pair lists and per-density shell-pair maxima.
* Preserves CompositeJK's additive-output contract:

  * full builds clear the physical J/K matrices in `CompositeJK`;
  * incremental builds retain the previous J/K matrices and add the response to the difference density;
  * DFCFMM auxiliary intermediates are cleared and reconstructed on every build.
* Adds regression coverage for:

  * RHF water calculations with direct CFMM;
  * water-dimer interaction energies with Direct, DF-DirJ, CFMM, and DFCFMM;
  * B3LYP phenol-dimer calculations with density screening and incremental Fock builds;
  * UHF and ROHF sodium–water calculations;
  * comparisons of DFCFMM against DirectDFJ and CFMM against conventional direct builds.

## Questions

* [ ] Is the current explicit enumeration of CompositeJK combinations in `SCF_TYPE` acceptable, or would the maintainers prefer a more general parser for arbitrary `J_ALGORITHM+K_ALGORITHM` combinations? (May be outside the scope for this PR)
* [ ] Are the names and initial defaults for `CFMM_ORDER`, `CFMM_GRAIN`, and `CFMM_EXTENT_TOLERANCE` appropriate?
* [ ] Should `libfmm` remain a separate Psi4 module, or would the maintainers prefer that this functionality reside entirely within `libfock` or even `libmints`?

## AI (GPT 5.6 Sol Extra High)

* [x] AI assistance was used for source inspection, debugging, and pre-maintainer code review; analysis of density screening, incremental-Fock accumulation, and intermediate-matrix initialization; implementation of targeted fixes for null dereferences, bounds checking, benchmark bookkeeping, input validation, and thread-safe access; improvements to naming, comments, methodological references, API visibility, and general code quality; implementation and testing of the `num_computed_shells()` reporting; expansion of regression coverage; preparation of Sphinx documentation; and refinement of this PR description. Author feedback guided the final design, including preservation of the planned `ContractionType::METRIC` extension point. **The underlying CFMM and DF-CFMM implementation predates this assistance and was originally developed in 2022.** The author reviewed the resulting changes and remains responsible for their technical correctness and validation.

## Checklist

* [x] Tests added for new features.
* [x] Docstring or narrative docs/Sphinx documentation updated if needed.
* Test running is well covered by CI. For running locally, see [[All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests).

## Status

* [x] Ready for review.
* [ ] Ready for merge — use Draft/Open GitHub status to signal readiness.


andyj10224:
Just rebased after the CI fix!

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1430 @@
...
+
+static double axis_sign(double coordinate) {
+    if (std::abs(coordinate) < 1.0e-8) {
+        return 0.0;
+    } else if (coordinate < 0) {
+        return -1.0;
+    } else {
+        return 1.0;
+    }
+}
```
TiborGY:
```suggestion
```
Eliminate data-dependent branching if possible, please. See below for alternative solution.

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1430 @@
...
+
+void CFMMBox::set_regions() {
+
+    // Creates a temporary parent shared pointer
+    std::shared_ptr<CFMMBox> parent = parent_.lock();
+
+    // Parent is not a nullpointer
+    if (parent) {
+        // Near field or local far fields are from children of parents
```
TiborGY, line 272: This is a branchless equivalent, in the form of a lambda local to the only scope where it is used.
Or alternatively, `static_cast<double>((coordinate > 1.0e-8) - (coordinate < -1.0e-8))` could be pasted to all three call sites of `axis_sign`.

```suggestion
        constexpr double axis_sign = [](double coordinate) constexpr noexcept {
            return static_cast<double>((coordinate > 1.0e-8) - (coordinate < -1.0e-8));
        };
        // Near field or local far fields are from children of parents
```

TiborGY (changes requested):

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done!

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done!

andyj10224:
Just rebased with upstream! @TiborGY requesting second round of reviews

psi4/src/psi4/libfmm/multipoles_helper.cc
```diff
@@ -0,0 +1,707 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+#include "psi4/pragma.h"
```
TiborGY: Is there anything in this file that needs pragma.h?
```suggestion
```

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,215 @@
...
+
+#include <algorithm>
+#include <cmath>
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+namespace psi {
+
+enum SolidHarmonicsType { Regular, Irregular };
```
TiborGY: Is there a reason why this shouldn't be an `enum class`?
```suggestion
enum class SolidHarmonicsType { Regular, Irregular };
```

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,215 @@
...
+
+// Indexing conventions shared with libmints/solidharmonics.cc.
+static inline int npure(int l) { return 2 * l + 1; }
+static inline int icart(int a, int b, int c) { return (((((a + b + c + 1) << 1) - a) * (a + 1)) >> 1) - b - 1; }
+static inline int ipure(int, int m) { return m < 0 ? 2 * -m : (m == 0 ? 0 : 2 * m - 1); }
+
```
TiborGY: Are there three functions used anywhere in CFMM?
```suggestion
```

TiborGY (changes requested):

psi4/src/psi4/libfmm/multipoles_helper.cc
andyj10224: Good catch — nothing in this file uses `PRAGMA_WARNING_*`, `PSI_API`, or `PSI_LOCAL`, so the
include is dead. Removed. Same was true of `fmm_tree.cc`, so I dropped it there too. The two headers do still need it (`multipoles_helper.h` and `fmm_tree.h` each use `PSI_API`), so those includes stay.

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: No good reason — done. It was leaking `Regular` and `Irregular` into all of `namespace psi`, which is exactly the argument for scoping it. Updated the 13 call sites in `multipoles_helper.cc` and `fmm_tree.cc`.

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: They aren't — no call sites anywhere in `libfmm` or `libfock`. `npure` and `icart` were verbatim copies of the file-static helpers in `libmints/transform.cc`, and `ipure` duplicated `m_addr` right below it (which *is* used). Deleted all three. The rest of that block — `ncart`, `choose`, `m_addr`, `cfmm_factorial` — is still in use.

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,210 @@
...
+#include <algorithm>
+#include <cmath>
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+namespace psi {
+
+enum class SolidHarmonicsType { Regular, Irregular };
```
TiborGY, line 45:
```suggestion
enum class SolidHarmonicsType { Regular, Irregular };
//TODO: Investigate if it would be better to do the Regular/Irregular split through a template parameter
```

psi4/src/psi4/libfmm/multipoles_helper.cc
```diff
@@ -0,0 +1,705 @@
...
+    mpole_terms_.resize(lmax_+1);
+
+    for (int l = 0; l <= lmax_; l++) {
+        Rc_[l].resize(l+1);
+        Rs_[l].resize(l+1);
+        mpole_terms_[l].resize(2*l+1);
+    }
+
+    if (type_ == SolidHarmonicsType::Regular) compute_terms_regular();
+    if (type_ == SolidHarmonicsType::Irregular) compute_terms_irregular();
```
TiborGY: If SolidHarmonicsType ever gained a third type this would break.
```suggestion
    switch (type_) {
        case SolidHarmonicsType::Regular:
            compute_terms_regular();
            break;
        case SolidHarmonicsType::Irregular:
            compute_terms_irregular();
            break;
    }
    throw PSIEXCEPTION("Tried to construct HarmonicCoefficients with an invalid solid-harmonic type!");
```

psi4/src/psi4/libfmm/multipoles_helper.cc
```diff
@@ -0,0 +1,705 @@
...
+    if ((new_center - center_).norm() <= std::numeric_limits<double>::epsilon()) return copy();
+
+    switch (type_) {
+        case SolidHarmonicsType::Regular:
+            return translate_regular(new_center);
+        case SolidHarmonicsType::Irregular:
+            return translate_irregular(new_center);
+        default:
+            throw PSIEXCEPTION("RealSolidHarmonics has an invalid solid-harmonic type.");
+    }
```
TiborGY, line 522: That default case might accidentally disarm a compiler warning about not all types being handled if SolidHarmonicsType gains a third type. Since both cases return we could put that throw beyond the switch statement where it should be unreachable.
```suggestion
    switch (type_) {
        case SolidHarmonicsType::Regular:
            return translate_regular(new_center);
        case SolidHarmonicsType::Irregular:
            return translate_irregular(new_center);
    }
    throw PSIEXCEPTION("RealSolidHarmonics has an invalid solid-harmonic type.");
```

psi4/src/psi4/libfmm/multipoles_helper.cc
```diff
@@ -0,0 +1,705 @@
...
+        if (l == 0) {
+            Rc_[0][0][0] = 1.0;
+            Rs_[0][0][0] = 0.0;
+        } else {
+            // m < l-1 terms
+            for (int m = 0; m < l-1; m++) {
+                double denom = (l+m)*(l-m);
+
+                // Rc_[l-1][m] contributions to Rc_[l][m]
+                for (const std::pair<int, double>& rpair : Rc_[l-1][m]) {
```
TiborGY: Every loop like this ends up creating a copy of the current pair, to which the const& can bind to. This is because the type of `std::unordered_map<int, double>` is not `std::pair<int, double>` but `std::pair<const int, double>`.
A structured binding would probably be the best choice here
```suggestion
                for (const auto& [ind, coef] : Rc_[l-1][m]) {
```

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,210 @@
...
+        return 2 * (-m);
+    } else {
+        // 1c, 2c, 3c, ...
+        return 2 * m - 1;
+    }
+}
+
+static inline double cfmm_factorial(const double n) {
+    return std::tgamma(n + 1.0);
+}
```
TiborGY: Multiple issues here.
This takes a double argument and calls tgamma which is an expensive function. The code never seems to call this with anything other than an integer argument, so tgamma is not needed in principle.
Also tgamma is from libm, which comes from whatever the system C library is, usually some version of glibc but could be musl. The result they return can be slightly different from system to system and between versions of glibc.
Since we only need integer arguments, this could be a constexpr look up table, generated at compile time without tgamma, which isnt constexpr.
So a binary build will use the same values for factorials regardless of where it is run.

```suggestion
// Implementation limit for CFMM_ORDER. The factorial would overflow at around 85.
inline constexpr size_t max_cfmm_order = 32;

// compute_terms_regular() evaluates (l ± m)! with |m| <= l <= max_cfmm_order
inline constexpr size_t max_cfmm_factorial = 2 * max_cfmm_order;

inline constexpr auto cfmm_factorial_table = [] {
    std::array<double, max_cfmm_factorial + 1> table{};
    table[0] = 1.0;
    for (std::size_t n = 1; n < table.size(); ++n) table[n] = table[n - 1] * static_cast<double>(n);
    return table;
}();

constexpr double cfmm_factorial(const size_t n) {
    if (n > max_cfmm_factorial) throw PSIEXCEPTION("cfmm_factorial: argument outside the tabulated range of [0," + std::to_string(max_cfmm_factorial) + "]");
    return cfmm_factorial_table[n];
}
```

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1422 @@
...
+    nlevels_ = options_.get_int("CFMM_GRAIN");
+    if (nlevels_ <= 2) {
+        throw PSIEXCEPTION("CFMM_GRAIN must be at least 3.");
+    } else if (nlevels_ >= 6) {
+        throw PSIEXCEPTION("CFMM_GRAIN must not exceed 5.");
+    }
+    lmax_ = options_.get_int("CFMM_ORDER");
+    if (lmax_ < 0) {
+        throw PSIEXCEPTION("CFMM_ORDER must be nonnegative.");
+    }
```
TiborGY, line 434: We should probably also guard against absurdly high CFMM_ORDER, if for no other reason the because eventually n! in cfmm_factorial would overflow a double. 32 sounds sensible, LMK if you think otherwise.
```suggestion
    } else if (lmax_ > max_cfmm_order) {
        throw PSIEXCEPTION("The maximum CFMM_ORDER supported by the implementation is " + std::to_string(max_cfmm_order));
    }
```

TiborGY (changes requested):

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: I accept this resolution, added the TODO on top of the enum class

psi4/src/psi4/libfmm/multipoles_helper.cc
andyj10224: Agreed, though we should use `return` instead of `break` as otherwise it would always lead to the `PSIEXCEPTION`! This is cleaner and more resilient code in practice!

psi4/src/psi4/libfmm/multipoles_helper.cc
andyj10224: Accepted for similar reasons to previous comment!

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: This is actually a smart idea! I am a personal fan of dynamic programming :)

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: I think having a max multipole_order of 32 is more than enough, since Q-Chem recommends 25 for gradients https://manual.q-chem.com/latest/sect_cfmm.html, so even if some unlucky soul is forced to implement CFMM gradients in the future, the multipole order limit should still be more than enough :)

psi4/src/psi4/libfmm/multipoles_helper.cc
andyj10224: Agreed! This change has also been applied to other loops in this file!

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,231 @@
...
+    return table;
+}();
+
+constexpr double cfmm_factorial(const int n) {
+    if (n < 0 || n > max_cfmm_factorial) {
+        throw PSIEXCEPTION("cfmm_factorial: argument outside the tabulated range of [0, " +
+                           std::to_string(max_cfmm_factorial) + "].");
+    }
+    return cfmm_factorial_table[static_cast<std::size_t>(n)];
+}
```
TiborGY: Given that compute_terms_regular() is the only call site of cfmm_factorial(), perhaps all of this could be moved into multipoles_helper.cc, instead of this header? That would avoid everything that includes this (transitively) seeing these symbols and potentially getting a copy of the table.
An anonymous namespace could be used for this, like this:
```
namespace {
    constexpr int max_cfmm_factorial = 2 * max_cfmm_order;
    
    constexpr auto cfmm_factorial_table = [] {
    ...
    }();
    
    constexpr double cfmm_factorial(const int n) {
    ...
    }
}
```

TiborGY (changes requested):
For some reason the Azure CI tests are failing and it does not look like a fluke.

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,231 @@
...
+
+class PSI_API MultipoleRotationFactory {
```
TiborGY: Does `MultipoleRotationFactory` need to be part of the C++ API?

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,231 @@
...
+
+    public:
+      // Constructor
+      MultipoleRotationFactory(Vector3 source_center, Vector3 target_center, int lmax);
+
+      SharedMatrix get_D(int l);
+
+}; // End MultipoleRotationFactory
+
+class PSI_API HarmonicCoefficients {
```
TiborGY: Same here about PSI_API

psi4/src/psi4/libfmm/multipoles_helper.h
```diff
@@ -0,0 +1,231 @@
...
+
+    public:
+      // Constructor
+      HarmonicCoefficients(int lmax, SolidHarmonicsType type);
+      // Returns a reference to the terms
+      const std::unordered_map<int, double>& get_terms(int l, int mu) const { return mpole_terms_[l][mu]; }
+
+};
+
+class PSI_API RealSolidHarmonics {
```
TiborGY: Also here about PSI_API

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Moved it to `multipoles_helper.cc`, `max_cfmm_order` kept in header since it is later referenced in other files.

andyj10224:
> For some reason the Azure CI tests are failing and it does not look like a fluke.

Just rebased from upstream, so hopefully that fixes it!

TiborGY:
> > For some reason the Azure CI tests are failing and it does not look like a fluke.
> 
> Just rebased from upstream, so hopefully that fixes it!

Nope, it did not.

loriab:
Your cfmm-2 and -4 tests use nbody, but ctest doesn't know that, and it's an optional dependency. Azure tests non-addons so that's why it's getting caught there. `dfmp2-2` is a similar test to model upon.

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1426 @@
...
+
+void CFMMBox::set_regions() {
+
+    // Creates a temporary parent shared pointer
+    std::shared_ptr<CFMMBox> parent = parent_.lock();
+
+    // Parent is not a nullpointer
+    if (parent) {
+        constexpr auto axis_sign = [](double coordinate) constexpr noexcept {
+            return static_cast<double>((coordinate >= 1.0e-8) - (coordinate <= -1.0e-8));
```
TiborGY: This is a hardcoded threshold. Why 1e-8? What would be the effects of setting this to 1e-5, or to 1e-11?

TiborGY (changes requested):

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: No, we got rid of it

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Fixed!

psi4/src/psi4/libfmm/multipoles_helper.h
andyj10224: Fixed!

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done! We changed this tolerance so it can never be negative

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1421 @@
...
+    std::vector<std::vector<std::shared_ptr<ShellPair>>> child_shell_pair_primary_buffer(nchild);
+    std::vector<std::vector<std::shared_ptr<ShellPair>>> child_shell_pair_auxiliary_buffer(nchild);
+
+    // Max WS at the child's level
+    int child_level_max_ws = std::max(2, (int) std::pow(2, level_+1));
+    int diffuse_child_max_ws = child_level_max_ws;
+
+    // Fill order (ws,z,y,x) (0)000 (0)001 (0)010 (0)011 (0)100 (0)101 (0)110 (0)111
+    // (1)000 (1)001 (1)010 (1)011 (1)100 (1)101 (1)110 (1)111
+    for (std::shared_ptr<ShellPair> shell_pair : primary_shell_pairs_) {
```
TiborGY: This does a std::shared_ptr<ShellPair> copy construction every iteration, no? That is not exactly free, as there is a reference counter that must be atomic updated in the ctor/dtor. Could `const auto& shell_pair` be used instead?

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1421 @@
...
+
+        int boxind = 8 * rbit + 4 * zbit + 2 * ybit + 1 * xbit;
+        child_shell_pair_primary_buffer[boxind].push_back(shell_pair);
+
+        int child_ws = std::max(2, (int)std::ceil(2.0 * extent / length_));
+        if (child_ws > diffuse_child_max_ws) diffuse_child_max_ws = child_ws;
+    }
+
+    // auxiliary_shell_pairs_ is empty for a direct primary-basis contraction.
+    for (std::shared_ptr<ShellPair> shell_pair : auxiliary_shell_pairs_) {
```
TiborGY: same here

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1421 @@
...
+
+    // Creates a temporary parent shared pointer
+    std::shared_ptr<CFMMBox> parent = parent_.lock();
+
+    // Parent is not a nullpointer
+    if (parent) {
+        // Near field or local far fields are from children of parents
+        // and children of parent's near field
+        for (std::shared_ptr<CFMMBox> parent_nf : parent->near_field_) {
+            for (std::shared_ptr<CFMMBox> child : parent_nf->children_) {
```
TiborGY: also here

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1421 @@
...
+    }
+}
+
+void CFMMBox::compute_mpoles_from_children() {
+
+    int nmat = 0;
+    // Upward pass: translate each child's regular expansion to the parent
+    // center and accumulate it. This is the multipole-to-multipole step of the
+    // CFMM hierarchy described by White et al. (1994, 1996).
+    for (std::shared_ptr<CFMMBox> child : children_) {
```
TiborGY: and here

psi4/src/psi4/libfmm/fmm_tree.cc
```diff
@@ -0,0 +1,1421 @@
...
+        Vff_.resize(nmat);
+    }
+
+    // Create multipoles and far field vectors for each density matrix
+    for (int N = 0; N < nmat; N++) {
+        mpoles_[N] = std::make_shared<RealSolidHarmonics>(lmax_, center_, SolidHarmonicsType::Regular);
+        Vff_[N] = std::make_shared<RealSolidHarmonics>(lmax_, center_, SolidHarmonicsType::Irregular);
+    }
+
+    for (std::shared_ptr<CFMMBox> child : children_) {
```
TiborGY: as well as here

TiborGY (changes requested):
Some of the tests are still failing in Azure CI.

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Yes it can! Thanks for the suggestion

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done!

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done!

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done!

psi4/src/psi4/libfmm/fmm_tree.cc
andyj10224: Done!

psi4/src/psi4/libfock/CFMM.cc
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @END LICENSE
+ */
+
+#include "jk.h"
+#include "SplitJK.h"
+
+#include <algorithm>
+
+using namespace psi;
```
TiborGY, line 34: Do we need using namespace psi;? Usually not the best choice to pull in large namespaces like this and the code itself is already in that namespace...

psi4/src/psi4/libfock/CFMM.cc
```diff
@@ -0,0 +1,137 @@
...
+
+namespace psi {
+
```
TiborGY, line 37: What is your opinion of putting the CFMM implementation into its own namespace?
It is not done consistently through the codebase but DLPNO and some other parts of it do it, and I tend to think it's a good idea.
The same would apply to everything in libfmm of course.

```suggestion
namespace psi {
namespace CFMM {
```

psi4/src/psi4/libfmm/fmm_tree.h
```diff
@@ -0,0 +1,344 @@
...
+      CFMMTree(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary, Options& options);
+
+      // Build the J matrix of CFMMTree
+      void build_J(std::vector<std::shared_ptr<TwoBodyAOInt>>& ints,
+                    const std::vector<SharedMatrix>& D, std::vector<SharedMatrix>& J,
+                    const std::vector<double>& metric_shell_diagonal_max = {});
+      // Returns the max tree depth
+      int nlevels() { return nlevels_; }
+      // Returns the max multipole AM
+      int lmax() { return lmax_; }
```
TiborGY, line 332: Not only required for DirectCFMM to de-pointer, but a good idea anyways.
```suggestion
      // Returns the max tree depth
      int nlevels() const { return nlevels_; }
      // Returns the max multipole AM
      int lmax() const { return lmax_; }
```

psi4/src/psi4/libfock/SplitJK.h
```diff
@@ -249,6 +250,93 @@ class PSI_API LinK : public SplitJK {
     std::string name() override { return "LinK"; }
 };
 
+class PSI_API DirectCFMM : public SplitJK {
+  protected:
+   /// The CFMMTree object used to compute the CFMM integrals
+   std::shared_ptr<CFMMTree> cfmmtree_;
```
TiborGY, line 256: Does DirectCFMM need to hold a pointer to its CFMMTree? Could it just hold the actual object instead?
I don't see the anything that benefits from the shared_ptr functionality, and I dislike needless pointers.
```suggestion
   CFMMTree cfmmtree_;
```

psi4/src/psi4/libfock/SplitJK.h
```diff
@@ -249,6 +250,93 @@ class PSI_API LinK : public SplitJK {
...
+   /**
+    * print name of method
+    */
+   std::string name() override { return "CFMM"; }
+};
+
+class PSI_API DFCFMM : public DirectDFJ {
+  protected:
+   /// CFMMTree used to calculate the three-center integrals
+   std::shared_ptr<CFMMTree> df_cfmm_tree_;
```
TiborGY, line 297: Same here.

psi4/src/psi4/libfock/CFMM.cc
```diff
@@ -0,0 +1,137 @@
...
+DirectCFMM::DirectCFMM(std::shared_ptr<BasisSet> primary, Options& options) : SplitJK(primary, options) {
+    cfmmtree_ = std::make_shared<CFMMTree>(primary_, nullptr, options_);
+}
```
TiborGY, line 40: Required if DirectCFMM switches to holding its CFMMTree directly
```suggestion
DirectCFMM::DirectCFMM(std::shared_ptr<BasisSet> primary, Options& options)
    : SplitJK(primary, options), cfmmtree_(primary, nullptr, options) {}
```

psi4/src/psi4/libfock/CFMM.cc
```diff
@@ -0,0 +1,137 @@
...
+
+size_t DirectCFMM::num_computed_shells() {
+    return num_computed_shells_;
+}
+
+DFCFMM::DFCFMM(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary, 
+               Options& options) : DirectDFJ(primary, auxiliary, options) {
+
+    df_cfmm_tree_ = std::make_shared<CFMMTree>(primary_, auxiliary_, options_);
+
```
TiborGY, line 71: If DFCFMM starts holding its CFMMTree directly
```suggestion
DFCFMM::DFCFMM(std::shared_ptr<BasisSet> primary, std::shared_ptr<BasisSet> auxiliary, Options& options)
    : DirectDFJ(primary, auxiliary, options), df_cfmm_tree_(primary, auxiliary, options) {
```

TiborGY (changes requested):

PR #3494, opened 2026-08-18
Matrix constructor refresh; `Dimension::operator=` deprecations

NathanGillispie:
## Description

This PR updates the Matrix constructors to use more initializer lists. In general, use of initializer lists is preferred over what happens currently: the compiler is implicitly initializing the values first then setting them *again* in the ctor block. Initializer lists are faster, require less allocations, allow the compiler to optimize moves vs. copies, and allow for const classes to be initialized more easily. The last point doesn't benefit Matrix outside of copying one Matrix to another, but developers should still be used to seeing them and using them.

Based on [rule of five](https://en.cppreference.com/cpp/language/rule_of_three#Rule_of_five) and our circumstances, we will need to implement the move `Matrix(const Matrix&& other)` ctor and move assignment `operator=(const Matrix&& other)` operator since the compiler does not do this by default. A function returning `Matrix` by value (technically prvalue) may go from copies to moves due to compiler [RVO](https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization). For example, a `Matrix` may be converted to `SharedMatrix` _without_ any copy operations: `std::make_shared<Matrix>(Factory())`. The obvious benefit here is that moving is near instantaneous and essentially constant time.

This would also have benefits for code within matrix.cc. Consider the random example `Matrix::pivoted_cholesky`:
```C++
auto U = std::make_shared<Matrix>(rowspi_, nchol);
U->zero();
for (int h = 0; h < nirrep_; ++h)
    for (int m = 0; m < rowspi_[h]; ++m)
        for (int n = m; n < nchol[h]; ++n)
            U->set(h, m, n, get(h, n, m));
*this = *U;
```
The Matrix class uses the copy assignment operator here, but it should instead move. For correct behavior, just convert to an rvalue `*this = std::move(*U);`. This code would do nothing before because no explicit or implicit move assignment operator existed. So adding support for move semantics is essentially free, and gives you free performance. There are really only pros and const to this PR.

*None*

## Dev notes & details

- [x] Deprecates Matrix ctors using `const int *` parameters.
- [x] Adds four new constructors to replace the above deprecated ctors `(const int *, const int)` --> `(const Dimension&, const int)`. These function identically, but use Dimension instead, and most of it can be done in the initializer list because `std::vector` has a handy "fill" constructor. 
- [x] Removes the `Dimension::operator=(const int * start)` lines in these ctors and replaces them with `Dimension(std::vector<int>(start, start + nirrep_))`. Since Matrix has a implicit conversion operator from `std::vector<int>` to `Dimension`, this just works without requiring any copy operations. It is somewhat similar to the previous approach, but more modern and shifts the deprecation burden away from Dimension.
- [x] The `Matrix(string, ...)` and `Matrix(...)` constructor pairs are merged by defining `Matrix(...) : Matrix("", ...) {};`. This works because the default value for `std::string` is already `""`, thus it behaves like the same ctor with a default option.
- [x] Explicit defaults were added for `matrix_`, `nirreps_` and `symmetry_`. `int` and pointer types initialize to garbage by default. This was already being explicitly set within each constructor, so this only simplifies redundant lines of code.
- [x] The now redundant default ctor `Matrix::Matrix()` was explicitly set to `= default`, again simplifying the code.
- [x] Added move semantics.
## AI

- [x] None

## Checklist

## Status
- [x] Ready for review

How it feels opening 6 consecutive PRs with no user-facing changes:
Programmer typing deviously

### DIFF
--- a/psi4/src/psi4/libmints/matrix.cc
+++ b/psi4/src/psi4/libmints/matrix.cc
@@ -63,6 +63,7 @@
 #include <regex>
 #include <tuple>
 #include <memory>
+#include <utility>
 
 // In molecule.cc
 namespace psi {
@@ -94,16 +95,18 @@ Matrix& Matrix::operator=(const Matrix &c) {
 
 Matrix::Matrix(Matrix&& m) noexcept
     : rowspi_(std::move(m.rowspi_)), colspi_(std::move(m.colspi_)), nirrep_(m.nirrep_), matrix_(m.matrix_),
-      symmetry_(m.symmetry_), name_(m.name_), numpy_shape_(std::move(m.numpy_shape_)) {
+      symmetry_(m.symmetry_), name_(std::move(m.name_)), numpy_shape_(std::move(m.numpy_shape_)) {
     // Copy pointer to data then nullify the previous pointer to prevent double free.
     m.matrix_ = nullptr;
 }
 
 Matrix& Matrix::operator=(Matrix&& m) noexcept {
+    if (this == &m) return *this;
+
     release();
     nirrep_ = m.nirrep_;
     symmetry_ = m.symmetry_;
-    name_ = m.name_;
+    name_ = std::move(m.name_);
     rowspi_ = std::move(m.rowspi_);
     colspi_ = std::move(m.colspi_);
     numpy_shape_ = std::move(m.numpy_shape_);
@@ -345,11 +348,10 @@ SharedMatrix Matrix::matrix_3d_rotation(Vector3 axis, double phi, bool Sn) {
         R(2, 1) = R(1, 2) = 2 * wy * wz;
         Matrix tmp(nrow(), 3);
         tmp.gemm(false, true, 1.0, rotated_coord, R, 0.0);
-        rotated_coord.copy(tmp);
+        rotated_coord = std::move(tmp);
     }
 
-    SharedMatrix to_return = rotated_coord.clone();
-    return to_return;
+    return std::make_shared<Matrix>(std::move(rotated_coord));
 }
 
 void Matrix::copy_to_row(int h, int row, double const *const data) {
@@ -924,7 +926,7 @@ void Matrix::symmetrize_gradient(std::shared_ptr<Molecule> molecule) {
     // Obtain atom mapping of atom * symm op to atom
     auto atom_map = compute_atom_map(molecule);
 
-    SharedMatrix ret(clone());
+    auto ret = std::make_shared<Matrix>(name_, rowspi_, colspi_, symmetry_);
     ret->zero();
     Matrix temp = *this;
 
@@ -960,7 +962,7 @@ void Matrix::symmetrize_hessian(SharedMolecule molecule) {
 
     auto atom_map = compute_atom_map(molecule);
 
-    auto symm = std::make_shared<Matrix>(clone());
+    auto symm = std::make_shared<Matrix>(name_, rowspi_, colspi_, symmetry_);
     symm->zero();
     double **pH = pointer();
     double **pS = symm->pointer();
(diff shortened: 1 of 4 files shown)


TiborGY:
I see there is one remaining todo in the description, is that expected to be done in this PR or later?

NathanGillispie:
I think this PR is big enough so I'll just remove the todo. After a cursory look into it, I don't think I want to change the behavior of any of the constructors now out of a fear of breaking things. There's no telling how much code relies on the assumption that Matrix behaves like that.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -101,88 +93,71 @@ Matrix &Matrix::operator=(const Matrix &c) {
...
-    nirrep_ = c->nirrep_;
-    symmetry_ = c->symmetry_;
-    name_ = c->name();
-    alloc();
-    copy_from(c->matrix_);
+Matrix::Matrix(Matrix&& m) noexcept
+    : rowspi_(std::move(m.rowspi_)), colspi_(std::move(m.colspi_)), nirrep_(m.nirrep_), matrix_(m.matrix_),
+      symmetry_(m.symmetry_), name_(std::move(m.name_)), numpy_shape_(std::move(m.numpy_shape_)) {
+    // Copy pointer to data then nullify the previous pointer to prevent double free.
+    m.matrix_ = nullptr;
```
jturney, line 102: `m.matrix_` is nulled but `m.nirrep_` is left untouched, while `m.rowspi_/m.colspi_` are moved-from (empty vectors in practice). So the source now claims some number of irreps with no data and no dimensions.

The destructor survives this, `release()` opens with `if (!matrix_) return;` so nothing crashes on scope exit. But every other member function reads `nirrep_` and then indexes: `Matrix::set(double)` does `matrix_[h][0][I]` for `h < nirrep_` → null deref; `copy_from` the same; `rowspi_[h]` on an empty Dimension → out of bounds. Same issue in the move-assignment operator.

One line each fixes it — set `m.nirrep_ = 0` (and clear `rowspi_/colspi_`) so a moved-from Matrix is exactly a default-constructed one. `alloc()`'s existing `if (!nirrep_) { matrix_ = nullptr; return; }` already handles that state correctly.

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -63,32 +63,24 @@
...
-    : matrix_(nullptr), nirrep_(0), name_(name), symmetry_(symmetry) {}
+    : name_(name), symmetry_(symmetry) {}
 
-Matrix::Matrix(const Matrix &c) : rowspi_(c.rowspi_), colspi_(c.colspi_) {
-    matrix_ = nullptr;
-    nirrep_ = c.nirrep_;
-    symmetry_ = c.symmetry_;
-    name_ = c.name();
+Matrix::Matrix(const Matrix &c)
+    : rowspi_(c.rowspi_), colspi_(c.colspi_), nirrep_(c.nirrep_), symmetry_(c.symmetry_), name_(c.name_) {
```
jturney: Can you add copying the numpy_shape_ here?

psi4/src/psi4/libmints/matrix.cc
```diff
@@ -63,32 +63,24 @@
...
     alloc();
     copy_from(c.matrix_);
 }
 
-Matrix &Matrix::operator=(const Matrix &c) {
+Matrix& Matrix::operator=(const Matrix &c) {
```
jturney, line 84: Can you add copying the numpy_shape_ here, too?

psi4/src/psi4/libmints/matrix.cc
NathanGillispie: That makes sense. Even if I moved the nirreps, it doesn't set the moved-from objects's nirreps to 0. So this change prevents people from doing something silly (dereferencing nullptr) with a Matrix after moving it.

NathanGillispie: Implemented the nirrep fix.

So it appears that the [move assignment operator](https://en.cppreference.com/cpp/container/vector/operator%3D#Version_2) does leave `other` in a "valid but unspecified state afterwards". But there is no such mention in the regular move constructor of vector: [`vector(vector&& other)` version 8](https://en.cppreference.com/cpp/container/vector/vector#Version_8). I'm not sure what to take from that but that more investigation is needed to be sure.

Also, the clear option may not be best because it doesn't reduce the capacity of the source vector, thus leaving unused memory allocated. I think it might be better to create an empty vector in the initializer then swap the two. That would be a backup plan if the move constructor doesn't do what I want.

PR #3496, opened 2026-08-18
cuEST wK progress

vlita:
- Exposed usage of range-separated (rs) functionals (energies only) by configuring `cuest_df_plan_` in `cuESTJK.cc` with range separated parameters. As of opening this PR, rs functional energies computed with the cuEST backend are not currently within the acceptable error range set for other functionals in `test_cuest.py`

- Created an option to toggle between cuEST's default mixed precision emulation (intended to boost performance) and usage of full double-precision throughout. Note that while emulated energies match full precision very closely (when using cuEST), there is still some discrepancy between the values calculated by cuEST vs by Psi4. The cause of this has not yet been fully resolved. 

- Fixed bug in which non rs functional gradients were scaled by HF exchange fraction twice, inadvertently caused by configuring rs parameters in `cuESTJK.cc`


vlita:
- Exposed two parameters (`CUEST_DFK_SLICES` and `CUEST_DFK_MODULI`) through options which allow the user to tune mixed precision emulation of the density-fitted K build. See cuEST's [docs](https://docs.nvidia.com/cuda/cuest/gemm_emulation.html) for more info

vlita:
- had AI examine our cuEST interface, and it found a bug in the code that Andy and I introduced.  In our code, we do not use Psi4's wK when using cuEST—Instead, both wK & K are packed into the cuEST K. Furthermore, we request the cuEST K if and only if alpha is non-zero. Since we weren't checking beta, cuEST calculations with wB97 (which is long-range only) did not request either cuEST K or Psi4's wK. This explains the 2 hartree error we were seeing yesterday.

vlita:
- Changes added by Andy as discussed in slack

psi4/src/psi4/scfgrad/jk_grad.cc
```diff
@@ -3090,7 +3090,7 @@ void cuESTJKGrad::compute_gradient() {
         cudaMemcpy(d_C, Ca_row_major.data(), sizeof(double) * na * nao, cudaMemcpyHostToDevice);
 
         std::shared_ptr<cuESTJK> cuest_jk = std::dynamic_pointer_cast<cuESTJK>(jk_);
-        cuestDFIntPlan_t cuest_df_plan = cuest_jk->cuest_df_plan();
+        cuestDFIntPlan_t cuest_df_plan = cuest_jk->cuest_df_plan(); // The df_plan configured in cuESTJK.cc is re-used for nuclear gradient computations!
```
loriab: own-line comments are preferred for all but the shorted comments to keep diff on the code cleaner

loriab:
LAB: Below is claude's review (Opus 5 High) in case anyone else is alarmed by the `component_scale` trace here. I'm good with it now with the explanation that it's skipping SuperFunctional. That's a down-the-road goal since cuEST will already need to be refactored to work with Jonathon's gauxc rework. The missing guard claude mentions sounds important to re-include. And some variety of testing is good -- I can add that broken to #3497 and then this PR can fix it. If you make changes and none of the windows CI lanes pass, it's not your fault -- I may need to rebase xptl_v1110.

## Review: `component_scale` scope, cleanliness, and performance

I went in worried that the `component_scale` work in da56e71 (and its `accumulate_cuest_xc_component` replacement in 5599bdd) was reaching outside cuEST. **It isn't** — those two commits are strictly cuEST-gated. But there *is* a cross-backend leak in this PR, in `scf_grad.cc`, from a different commit. Details below.

TL;DR:

- [x] `component_scale` / `accumulate_cuest_xc_component` — cuEST-only, verified mechanically. No action needed.
- [x] Performance — measured, non-issue (the refactor is marginally *faster*).
- [ ] **`scf_grad.cc` α/β reset lost its runtime `USE_CUEST` guard** — wrong hybrid-DFT gradients on the CPU path in any cuEST build, *and* wrong LRC gradients. See (A).
- [ ] The α-scaling fix ships with essentially no test coverage. See (B).

---

### 1. Scope: does this affect other backends?

No. I built the preprocessor stack over the post-PR `v.cc` and intersected it with the changed line numbers:

```
added/changed lines in new v.cc:            143
lines NOT inside a USING_cuEST region:        2   ->  146: #endif
                                                     147: (blank)
```

Every functional line is inside `#ifdef USING_cuEST`, and within that, inside `if (options_.get_bool("USE_CUEST"))` — e.g. the `RV::compute_V` cuEST block spans lines 1421–2021. Compile-time **and** runtime gated. The libxc CPU path, BrianQC, and everything else are untouched. Same story for `cuESTJK.cc`; `rhf.cc` / `uhf.cc` do it correctly via a `use_cuest` local that is `false` in non-cuEST builds.

### 2. Why so widespread? Because it isn't really an LRC fix

Two things are going on.

**(a) The cuEST Vxc path bypasses `SuperFunctional`.** Rather than calling `functional_->compute_functional()`, it hand-calls `xc_lda_exc_vxc` / `xc_gga_exc_vxc` / `xc_mgga_exc_vxc` and re-implements the accumulation. `LibXCFunctional::compute_functional` applies `alpha_` via `C_DAXPY` (`libfunctional/LibXCfunctional.cc:619`); the cuEST reimplementation summed raw libxc output with an implicit weight of 1.0. Since that accumulation is duplicated across 4 methods x 3 ansätze x {x, c}, a correct fix *has* to touch ~24 sites. The breadth is a symptom of the pre-existing duplication, not of the fix.

**(b) This is not primarily a wK bug.** I built all 1677 registered functionals and inspected component alphas: 194 distinct component signatures, **28 with a component α ≠ 1, of which exactly one is range-separated**.

| functional | `is_x_lrc` | components (α) |
|---|---|---|
| `lc-wpbe0` | **True** | `GGA_X_HJS_PBE` 0.75 / `GGA_C_PBE` 1.0 |
| `pw6b95` | False | `GGA_X_MPW91` 0.72 / `MGGA_C_BC95` 1.0 |
| `mpw1lyp` | False | `GGA_X_MPW91` 0.75 / `GGA_C_LYP` 1.0 |
| `pbe50` | False | `GGA_X_PBE` 0.5 / `GGA_C_PBE` 1.0 |
| `pbeh-3c` | False | `GGA_X_PBE` 0.58 / `GGA_C_PBE` 1.0 |
| `revpbe0`, `mpw1pbe`, `ptpss`, `pwpb95`, `pbe0-2`, `pbe0-dh`, `b2plyp`, `b2gpplyp`, `core-dsd-blyp`, `dsd-*` | False | 27 non-LRC signatures total |

Magnitude of the pre-PR error (water / cc-pVDZ, correct vs. α-dropped):

```
mpw1lyp   -76.3865034  vs  -78.6539721   ->  -1423 kcal/mol
pw6b95    -76.4998290  vs  -79.0841397   ->  -1622
pbe50     -76.3444611  vs  -80.8857934   ->  -2850
pbeh-3c   -76.2783587  vs  -79.9577281   ->  -2309
lc-wpbe0  -76.3586325  vs  -78.2192983   ->  -1168
```

So this is a **pre-existing, catastrophic cuEST bug affecting ~27 global and double hybrids**, found while chasing wK. Worth saying so in the commit message / PR description — right now it reads as an LRC detail.

### 3. Performance

Non-issue. I extracted the helper and the original hand-written mGGA-UKS loop into a standalone harness (clang, `-O2`, C++20):

```
original   4.45 ns / point
refactor   4.14 ns / point      (~7% faster)
```

Clang constant-propagates `components_per_point` through the `std::array` temporary and unrolls, so the abstraction is free. For scale, the `xc_mgga_exc_vxc` call it immediately follows measures **~91 ns/point** — the accumulate loop is ~4% of that, and both are noise next to the GPU collocation and the host<->device density/potential memcpys that dominate this path.

I also checked the refactor is bit-for-bit identical to the hand-written loop, NaN/inf-skip semantics included, across all field layouts (1/2/3 components, NaN and ±inf seeded in each field independently).

### 4. Is there a cleaner way?

**Long-term, yes, and it's a big win:** route the cuEST path through `functional_->compute_functional(vals, npoints)` and read back `V_RHO_A` / `V_GAMMA_AA` / `V_TAU_A`. The post-accumulation code only consumes `full_f`, `full_f_rho`, `full_f_gamma`, `full_f_tau` — exactly what `compute_functional` returns. That deletes ~500 lines of ansatz-switch duplication and inherits α, the tau convention, the NaN policy, GRAC, and every future `SuperFunctional` feature for free.

One caveat to budget for: psi4's `V_TAU_A` is `0.5 * α * f_tau` whereas the cuEST path packs raw libxc `f_tau` into `p_Vxc_grid[5*point + 4]`, so that needs a documented factor of 2.

**For this PR:** 5599bdd is the better of the two commits and I'd take it as-is. It's semantically identical and it collapses 24 copy-pasted NaN-check-and-accumulate blocks into one place — which is precisely what let the α bug hide for so long. Optional nit: `components_per_point` is a compile-time constant at every call site, so it could be a template parameter rather than a runtime struct member.

---

## (A) `scf_grad.cc` — α/β reset lost its runtime `USE_CUEST` guard

This is the one place where the "does this affect other backends?" answer is **yes**.

81498e5d added this correctly: `#ifdef USING_cuEST` **plus** `if (options_.get_bool("USE_CUEST"))`, placed *after* JK construction. cd99e065 then hoisted it above the branch and dropped the runtime check:

```cpp
#else
    alpha = 1.0;          // <-- unconditional in any cuEST-enabled build
    beta  = 1.0;

    std::shared_ptr<JKGrad> jk;
    if (options_.get_str("SCF_TYPE").find("DF") != std::string::npos
        && options_.get_bool("USE_CUEST") && !functional_->is_x_lrc()) {
        jk = JKGrad::build_cuESTJKGrad(1, jk_, -alpha, -beta);
    } else {
        jk = JKGrad::build_JKGrad(1, mintshelper_);   // returns UNSCALED K
    }
#endif
    ...
    gradients_["Exchange"]->scale(-alpha);            // now -1.0, not -x_alpha()
```

Three consequences, all in a cuEST-enabled build:

1. Running with `USE_CUEST false` (or a non-DF `SCF_TYPE`) takes the CPU `JKGrad` path, which returns unscaled K, and then scales exchange by −1.0 instead of −0.2. **B3LYP gradients are silently wrong on the ordinary CPU path.**
2. The branch predicate excludes `is_x_lrc()`, so LRC functionals deliberately fall back to CPU `JKGrad` — and hit the same bug. **This PR enables wB97X energies but leaves its gradients broken.**
3. `build_cuESTJKGrad(1, jk_, -alpha, -beta)` now always receives `(-1.0, -1.0)`, because the reset was hoisted above the call. If the reused `df_plan` genuinely already carries α/β (as the new comment says), then those two arguments are dead and should be removed or asserted rather than silently passed as 1.0.

`rhf.cc` and `uhf.cc` already have the right pattern — a `use_cuest` local, `#ifdef` + runtime check at each site. `scf_grad.cc` is the sole outlier. Hoisting the branch predicate into a named local and conditioning the reset on it fixes all three:

```cpp
#ifdef USING_cuEST
    const bool use_cuest_grad = options_.get_str("SCF_TYPE").find("DF") != std::string::npos
                                && options_.get_bool("USE_CUEST")
                                && !functional_->is_x_lrc();
    if (use_cuest_grad) {
        // cuEST folds alpha/beta into the reused df_plan, so don't apply them again here
        alpha = 1.0;
        beta  = 1.0;
    }
#endif
```

## (B) The α-scaling fix has essentially no test coverage

Every functional exercised by `test_cuest.py` — `svwn`, `blyp`, `b3lyp`, `m06`, `b97m-v` — has **all component alphas equal to 1.0**. That is exactly why this bug survived.

`test_cuest_range_separated.py` filters on `is_x_lrc()`, which adds exactly one α ≠ 1 case (`lc-wpbe0`); the headline `wb97x` / `wb97x-d` are single-component with α = 1, so they don't exercise the scaling either.

Suggest parametrizing a few non-LRC α ≠ 1 functionals. `pw6b95`, `mpw1lyp`, and `pbe50` are cheap and between them cover GGA-x with mGGA-c, GGA/GGA, and the hybrid-mixing case.

## (C) Minor

- `dfk_slices_` / `dfk_moduli_` are `uint64_t` assigned from `options.get_int()`; a negative input wraps to a huge value handed straight to cuEST. Worth a bounds check.
- Typos in the new `scf_grad.cc` comment: "seperated" -> "separated", "functioinals" -> "functionals".

---

<details>
<summary>How this was checked</summary>

- Preprocessor-region analysis of post-PR `v.cc` against `git diff -U0` line numbers.
- All 1677 registered functionals built via `build_superfunctional_from_dictionary`, component `alpha()` / `omega()` inspected, deduplicated by component signature.
- Error magnitudes from paired SCF energies: reference functional vs. the same dictionary with every component `alpha` forced to 1.0.
- Equivalence and timing from a standalone harness reproducing the original UKS-mGGA loop and the new helper side by side.
- libxc cost from `LibXCFunctional('XC_MGGA_X_TPSS').compute_functional` over 2e5 points, single-threaded.

**Caveat:** my build has `ENABLE_cuEST=OFF`, so I could not compile or run the cuEST branch itself. (A) and (B) are from source analysis; the equivalence and perf numbers come from the extracted harness, not an in-tree build. Baseline `tests/dft1` passes on my tree.

</details>

vlita:
> 
> ```c++
> #ifdef USING_cuEST
>     const bool use_cuest_grad = options_.get_str("SCF_TYPE").find("DF") != std::string::npos
>                                 && options_.get_bool("USE_CUEST")
>                                 && !functional_->is_x_lrc();
>     if (use_cuest_grad) {
>         // cuEST folds alpha/beta into the reused df_plan, so don't apply them again here
>         alpha = 1.0;
>         beta  = 1.0;
>     }
> #endif
> ```
> 

I added the `options_.get_bool("USE_CUEST")` guard, but imo psi4 should just throw an exception if cuEST is requested with a non-df scf type or wK gradients. I can add this if desired.

loriab:
### `scf_grad.cc`: α/β reset guard doesn't match the JK-selection guard

Thanks for restoring the runtime check in 6fe95c3 — that closes the case I was most worried about. One case is still open, and it's the one this PR is meant to enable.

**Fixed:** with the guard back, a cuEST-enabled build running `USE_CUEST=false` keeps the true `x_alpha` / `x_beta`, so ordinary CPU-path hybrid gradients (B3LYP etc.) are correct again.

**Still open:** the reset is guarded on `USE_CUEST`, but the JK-object selection two lines below uses a *different* predicate:

```cpp
    if (options_.get_bool("USE_CUEST")) {          // <-- guard 1
        alpha = 1.0;
        beta  = 1.0;
    }

    std::shared_ptr<JKGrad> jk;
    if (options_.get_str("SCF_TYPE").find("DF") != std::string::npos
        && options_.get_bool("USE_CUEST")
        && !functional_->is_x_lrc()) {             // <-- guard 2, note the extra term
        jk = JKGrad::build_cuESTJKGrad(1, jk_, -alpha, -beta);
    } else {
        jk = JKGrad::build_JKGrad(1, mintshelper_);   // CPU path: returns UNSCALED K and wK
    }
```

The two disagree exactly on the range-separated case. In a cuEST build with `SCF_TYPE` = DF:

| `USE_CUEST` | `is_x_lrc()` | JKGrad used | α/β at `scale()` | |
|---|---|---|---|---|
| `false` | any | CPU | `x_alpha` / `x_beta` | correct (fixed by 6fe95c3) |
| `true` | `false` | `cuESTJKGrad` | 1.0 / 1.0 | correct — df_plan carries α/β |
| `true` | **`true`** | **CPU** | **1.0 / 1.0** | **wrong — needs `x_alpha` / `x_beta`** |

In that third row the reset fires, but `!is_x_lrc()` sends us down the **CPU** `JKGrad` branch, which returns unscaled K and wK. Then:

```cpp
    gradients_["Exchange"]->scale(-alpha);       // -1.0 instead of -x_alpha()
    gradients_["Exchange,LR"]->scale(-beta);     // -1.0 instead of -x_beta()
```

So **wB97X and friends get wrong gradients under `USE_CUEST=true`** — the exact functionals this PR adds support for. Energies are fine; only the gradient path is affected.

Separately, `build_cuESTJKGrad(1, jk_, -alpha, -beta)` now always receives `(-1.0, -1.0)`, since the reset precedes the call. If the reused `cuest_df_plan_` really does already carry α/β (as the comment says), those two arguments are dead — worth dropping them or asserting, rather than passing 1.0 and relying on the reader to work out why that's harmless.

**Suggested fix** — condition the reset on the same predicate that selects the JK object, so the two can't drift apart again:

```cpp
    const bool use_cuest_grad = options_.get_str("SCF_TYPE").find("DF") != std::string::npos
                                && options_.get_bool("USE_CUEST")
                                && !functional_->is_x_lrc();

    // cuEST folds alpha/beta into the reused cuest_df_plan_, so don't apply them again below.
    // The LRC case falls back to the CPU JKGrad, which returns unscaled K/wK and still needs them.
    if (use_cuest_grad) {
        alpha = 1.0;
        beta  = 1.0;
    }

    std::shared_ptr<JKGrad> jk;
    if (use_cuest_grad) {
        if (!jk_) {
            throw PSIEXCEPTION(
                "cuESTJKGrad requires the SCF JK object to be preserved across finalization. "
                "Set the SAVE_JK option to True before running the SCF calculation.");
        }
        jk = JKGrad::build_cuESTJKGrad(1, jk_, -alpha, -beta);
    } else {
        jk = JKGrad::build_JKGrad(1, mintshelper_);
    }
```

Nit while you're in here: the comment just above has two typos — "seperated" -> "separated", "functioinals" -> "functionals".

vlita:
Functionals currently tested by `test_cuest_range_separated.py` (and passing for energy error within `5e-4`):

|  |  |
|---|---|
| CAM-B3LYP | CAM-LDA0 |
| HJS-B97X | HJS-PBE |
| HJS-PBE-SOL | HSE03 |
| HSE06 | LC-BOP |
| LC-VV10 | LRC-WPBE |
| LRC-wPBEh | M11 |
| MN12-SX | N12-SX |
| TUNED-CAM-B3LYP | wB97 |
| wB97M-D3BJ | wB97M-V |
| wB97X | wB97X-D |
| wB97X-D3 | wB97X-D3BJ |
| wB97X-V | wB97X3c |
| wPBE | wPBE0 |

- HJS-B88 failing with max iter for Psi4 and ADIIS error for cuEST

vlita:
GPT-5.6 Sol Pro's explanation of the bug fixed by [7f83da7](https://github.com/psi4/psi4/pull/3496/commits/7f83da7ba14ebbdeda74cdf8a52ac5ec424b7158) is below. I tested the cuEST path with all functional combinations for an energy error of `5e-4` against Psi4, and it passed for everything except HJS-B88 and MP2D for non cuEST related issues (as far as I can tell).  In the commits below ([501dc5e](https://github.com/psi4/psi4/pull/3496/commits/501dc5e0ac906a2fb4a6b4db9df0dd21a656efcc) - [fac060e](https://github.com/psi4/psi4/pull/3496/commits/fac060e88c92004b21525b958144210e5c780ef9)) I also added RHF/UHF PW6B95 to  `test_cuest`, as it combines GGA exchange with meta-GGA correlation. All tests pass with existing tolerances.

# cuEST Mixed-Rung XC Components

## What Was Wrong Before

Psi4 assigns the entire `SuperFunctional` the highest rung used by any component:

```cpp
int SuperFunctional::ansatz() const {
    if (is_meta()) return 2;
    if (is_gga()) return 1;
    return 0;
}
```

For a functional containing GGA exchange and meta-GGA correlation, the overall ansatz is meta-GGA.

The previous cuEST code then used that overall ansatz to evaluate every component. Inside the meta-GGA branch, it called:

```cpp
xc_mgga_exc_vxc(...)
```

for both the meta-GGA correlation and the GGA exchange.

That is wrong for the GGA component. A GGA LibXC handle must be evaluated with:

```cpp
xc_gga_exc_vxc(...)
```

## What The Commit Changes

The commit adds:

```cpp
evaluate_cuest_xc_component(...)
```

This helper examines each component individually:

```cpp
if (component->is_meta()) {
    xc_mgga_exc_vxc(...);
} else if (component->is_gga()) {
    xc_gga_exc_vxc(...);
} else {
    xc_lda_exc_vxc(...);
}
```

The overall functional still requests density information at its highest rung:

```text
LDA functional       -> rho
GGA functional       -> rho + gradient
meta-GGA functional  -> rho + gradient + tau
```

But each component now consumes only the ingredients appropriate for its own rung.

For a GGA plus meta-GGA functional:

```text
Overall density calculation: meta-GGA ingredients
GGA component evaluation:    xc_gga_exc_vxc
meta-GGA evaluation:         xc_mgga_exc_vxc
```

That is the correct structure.

## Other Accomplishments

The commit also:

- Applies each component's `alpha()` weight to its energy and derivative fields.
- Uses the same helper in RKS potential, UKS potential, RKS XC gradient, and UKS XC gradient.
- Validates all components during initialization.
- Rejects non-LibXC semilocal components with a clear error.
- Rejects Laplacian-dependent meta-GGAs, which cuEST does not support.
- Removes the shared `LibXCFunctional*` variable from OpenMP regions, eliminating a potential thread race.
- Removes several hundred lines of duplicated LDA/GGA/meta-GGA component loops.

It does not change JK gradients, range-separated gradient support, or the alpha/beta handling in `scf_grad.cc`.

PR #3500, opened 2026-08-21
Clean up confusing is_core handling between DiskDFJK and CDJK

TiborGY:
## Description

Currently CDJK declares `virtual bool is_core()`, which hides DiskDFJK's `is_core()` instead of overriding it.
This has several negative effects, primarily that CDJK's `is_core()` is not the one getting called when it really should be.

Since CDJK does not implement its own `preiterations()` and `iaia()`, calls to these functions on a CDJK object lead to the parent implementations running (`DiskDFJK::preiterations()` and `DiskDFJK::iaia()`, respectively). But since CDJK does not override `is_core()`, when `DiskDFJK::preiterations()` and `DiskDFJK::iaia()` runs, they call `DiskDFJK::is_core()`, which might select the out-of-core algo and return false. As CDJK only supports incore, mayhem ensues.

This PR switches `is_core()` from hiding to overriding its parent counterpart.

- [x] C++ API change: `DiskDFJK::is_core()` is now marked as virtual
- [x] C++ API change: `CDJK::is_core()` now overrides `DiskDFJK::is_core()` instead of hiding it. `CDJK::is_core()` now throws if SCF_SUBTYPE is neither AUTO or INCORE.
- [x] SCF_SUBTYPE=CD with SCF_SUBTYPE=AUTO previously failed with 'Disk algorithm for CD JK not implemented.', if the a sufficiently pessimistic memory estimate was obtained, which was bad UX. It now proceeds to the decomposition and either succeeds or fails with the authoritative 'Not enough memory for CD.' check.
- [x] SCF_SUBTYPE=CD with SCF_SUBTYPE=OUT_OF_CORE previously failed with a misleading message, it now fails immediately with an explanation and suggested alternatives.

## Dev notes & details

- [x] `DiskDFJK::is_core()` is now marked as virtual so that it can be overridden by CDJK.
- [x] `virtual bool is_core() { return true; }` for CDJK is now `bool is_core() override;` with a body that throws if an unsupported SCF_SUBTYPE is requested.
- [x] The dead initial read of the cached `is_core_` flag in `DiskDFJK::is_core()` is removed (every branch overwrote it).
- [x] Comments and docstrings are added or expanded where relevant.
- [x] Some PSIEXCEPTION messages were improved
- [x] `CDJK::initialize_JK_disk()` should now be unreachable

## AI

- [x] LLMs were used to write the initial version of this patch, which was then manually applied and refined in the process. The initial version of the PR description was generated by LLMs, which was then manually rewritten.

## Checklist
- [x] No new features
- [x] Docstrings and comments in the source files were added/updated

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libfock/CDJK.cc
+++ b/psi4/src/psi4/libfock/CDJK.cc
@@ -52,7 +52,26 @@ namespace psi {
 CDJK::CDJK(std::shared_ptr<BasisSet> primary, Options& options, double cholesky_tolerance)
     : DiskDFJK(primary, primary, options), cholesky_tolerance_(cholesky_tolerance) {}
 
-void CDJK::initialize_JK_disk() { throw PSIEXCEPTION("Disk algorithm for CD JK not implemented."); }
+void CDJK::initialize_JK_disk() {
+    throw PSIEXCEPTION(
+        "CDJK::initialize_JK_disk(): internal error: the out-of-core path was selected even though CDJK::is_core() "
+        "guarantees the in-core subalgorithm. If you are seeing this, there is a bug!");
+}
+
+bool CDJK::is_core() {
+    // CD has no out-of-core algorithm, so the only meaningful subalgorithm is in-core.
+    if (subalgo_ != "AUTO" && subalgo_ != "INCORE") {
+        throw PSIEXCEPTION(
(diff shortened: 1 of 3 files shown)


loriab (approved):
yikes. thanks for the cleanup!

TiborGY:
It might be a good idea to add tests that would have failed before this PR, but that will be a separate PR.

andyj10224 (approved):
LGTM! Thank you for this patch!

PR #3501, opened 2026-08-22
Scope external potentials by fragment in many-body calculations

Awallace3:
## Summary

- allow `external_potentials` to use a dictionary keyed by 1-indexed fragment numbers in many-body calculations
- apply each fragment's potentials only when that fragment contains real atoms, not when it is present solely as a CP ghost fragment
- preserve flat-list behavior while warning that those potentials apply to every many-body component
- validate fragment dictionary keys and document the new `bsse_type` usage
- (LAB) conform to c-f and stop ecosystem testing on py310 since new package builds won't be 310-compat.

## Motivation

Previously, external point charges were passed to every monomer and supersystem calculation in a many-body expansion. This could place charges in calculations for fragments they were not intended to affect.

Fragment-scoped input such as:

```python
external_potentials={2: potentials_b}
```

now applies `potentials_b` to monomer B and supersystems containing real fragment B. A calculation containing only ghost fragment B does not receive those potentials.

## Tests

Added three water-cluster tests covering:

- a NoCP dimer result against explicitly computed monomer and dimer SCF energies
- a CP dimer numerical regression
- trimer task planning across monomers, dimers, the trimer, and real-versus-ghost fragment selection


tests/pytests/test_nbody_external_potentials.py
```diff
@@ -0,0 +1,500 @@
...
+    c_points = [[-0.3, [4.0, 5.0, 6.0]]]
+    c_diffuse = [[0.1, 7.0, 8.0, 9.0, 0.5]]
+
+    global_potential = {"A": pot_a, "B": pot_b}
+    if c_partition == "points-list":
+        global_potential["C"] = c_points
+    elif c_partition == "mode-dict":
+        global_potential["C"] = {"points": c_points, "diffuse": c_diffuse}
+
+    plan = task_planner(
```
loriab: this is probably fine, but it's a little more usual for users to call `energy(..., return_plan=True)`

psi4/driver/procrouting/proc.py
```diff
@@ -1511,134 +1511,8 @@ def scf_wavefunction_factory(name, ref_wfn, reference, **kwargs):
     return wfn
 
 
-def validate_external_potential(external_potential) -> Dict:
```
loriab, line 1514: glad to see the duplicate go

tests/pytests/test_extern.py
```diff
@@ -392,3 +392,37 @@ def test_extern_points_diffuse(ep, anskey):
 
     ene = psi4.energy("hf/6-31G*", molecule=water, external_potentials=ep)
     psi4.compare_values(ans[anskey]["energy"], ene, 6, anskey)
+
+
+def test_extern_wavefunction_uses_shared_validator(monkeypatch):
+    """Verify the wavefunction setup normalizes with :py:func:`psi4.p4util.validate_external_potential`.
+
+    A spelling taught only to the p4util validator must reach the wavefunction, so the
+    accepted-spelling grammar has a single owner rather than forked copies that can drift.
```
loriab: Is this just saying the functions were consolidated and the keys are steady?

tests/pytests/test_nbody_external_potentials.py
```diff
@@ -0,0 +1,500 @@
...
+from psi4.driver.driver_nbody import ManyBodyComputer
+from psi4.driver.p4util.exceptions import ValidationError
+from psi4.driver.task_planner import task_planner
+
+from addons import uusing
+
+
+pytestmark = [pytest.mark.psi, pytest.mark.api]
+
+
```
loriab, line 15: Does the nbody report table need "N/A"ing in the presence of potentials like embedding does? https://github.com/MolSSI/QCManyBody/blob/main/qcmanybody/tests/test_computer_het4_gradient.py#L169

loriab (commented):
It's a complicated change, and I appreciate the plans for the edge cases and failure modes. It looks reasonably correct to me, and Sol 5.6 didn't produce suggestions I consider important. I wonder if we need to have a better/any printout of what potentials are present (whole and per-frag) and what frags they're going to be activated for. It's probably not clear from the output file alone as stands. I'd feel better with a little more end-to-end testing like the qcmanybody/ runs (one linked in one of the comments) that checks each psivar is getting modified in the way it's expected. 

But on the whole, I think this is rtg after minor revisions. If there's a glaring error I think it's more likely to come out during research than from staring. Thanks for the nice feature!

tests/pytests/test_extern.py
Awallace3: This test was for TDD and should not actually be kept. As such, I am dropping it for it is confusing and quite unhelpful going forward. This is related to the dropped duplicate above.

tests/pytests/test_nbody_external_potentials.py
Awallace3: Yes, it does need "N/A"ing on the QCManyBody side if I understand correctly. Right now is populates nan

tests/pytests/test_nbody_external_potentials.py
Awallace3: I updated these to use the more common energy route instead of the task_planner

Awallace3 (commented):
seems ready for re-reviewing

Awallace3:
> It's a complicated change, and I appreciate the plans for the edge cases and failure modes. It looks reasonably correct to me, and Sol 5.6 didn't produce suggestions I consider important. I wonder if we need to have a better/any printout of what potentials are present (whole and per-frag) and what frags they're going to be activated for. It's probably not clear from the output file alone as stands. I'd feel better with a little more end-to-end testing like the qcmanybody/ runs (one linked in one of the comments) that checks each psivar is getting modified in the way it's expected.
> 
> But on the whole, I think this is rtg after minor revisions. If there's a glaring error I think it's more likely to come out during research than from staring. Thanks for the nice feature!

Great! I have added the following logging information for informing the user at the start of the computation which external potentials map to which real fragments. Also, to check qcvars explicitly, I have added `test_nbody_external_potentials_nocp_qcvars()` for checking that the external potential does produce specific values for populated QC vars (and the plain vs external_potential routes do not have the same numbers).

```
  //>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>//
  //   ManyBody Setup: N-Body Levels 2 //
  //<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<//

    Model chemistry "(auto)" (§A):          3
        Number of 1-body computations:      2 (nocp: 0, cp: 2, vmfc_compute: 0)
        Number of 2-body computations:      1 (nocp: 0, cp: 1, vmfc_compute: 0)

External potential for computation on real fragments [2]:
[[0.917, [-4.842990114323323, -1.5626145331410781, -3.15508673906439]],
 [-0.934, [-3.382420791956967, -0.7609927107218674, -2.4128023169845547]],
 [0.917, [-3.588211967019324, -0.9435402544410937, -0.5805238657406447]]]

External potential for computation on real fragments [1]:
[[0.417, [-4.842990114323323, -1.5626145331410781, -3.15508673906439]],
 [-0.434, [-3.382420791956967, -0.7609927107218674, -2.4128023169845547]],
 [0.417, [-3.588211967019324, -0.9435402544410937, -0.5805238657406447]]]

External potential for computation on real fragments [1, 2]:
[[0.417, [-4.842990114323323, -1.5626145331410781, -3.15508673906439]],
 [-0.434, [-3.382420791956967, -0.7609927107218674, -2.4128023169845547]],
 [0.417, [-3.588211967019324, -0.9435402544410937, -0.5805238657406447]],
 [0.917, [-4.842990114323323, -1.5626145331410781, -3.15508673906439]],
 [-0.934, [-3.382420791956967, -0.7609927107218674, -2.4128023169845547]],
 [0.917, [-3.588211967019324, -0.9435402544410937, -0.5805238657406447]]]
```

For handling the final outputs, a QCManyBody PR https://github.com/MolSSI/QCManyBody/pull/53 attempts to address the nans to get the following returning table result in psi4:
```
  //>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>//
  //          ManyBody Results         //
  //<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<//


   ==> N-Body: Counterpoise Corrected (CP) energies <==

              MC n-Body  Total Energy            Interaction Energy                          N-body Contribution to Interaction Energy
                         [Eh]                    [Eh]                  [kcal/mol]            [Eh]                  [kcal/mol]
              §A  1        N/A                              N/A                   N/A        0.000000000000        0.000000000000
     FULL/RTN §A  2        N/A                              N/A                   N/A        0.005150000059        3.231673826703

   MC Legend: §A: "(auto)"
```

loriab:
Thanks for the qcmb PR! It's now ready as v0.8.0. Remaining failures are b/c c-f dropped py310 support, so the new package think it can't install so those lanes are getting v0.7.x . Easiest for us to just follow suit ...

codedeps.yaml
```diff
@@ -1535,17 +1535,18 @@ data:
       host: github
       account: MolSSI
       name: QCManyBody
-      commit: v0.7.1
+      commit: v0.8.0
       githttps: true
     cmake:
       name: qcmanybody
-      constraint: "ATLEAST 0.7.1"
+      constraint: "ATLEAST 0.8.0"
```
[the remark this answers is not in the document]
loriab: not an issue in practice since conda collects most recent. it'll get copied over en masse when codedeps deploys

loriab (approved):
Thanks for the logging and per-var test.

PR #3503, opened 2026-08-22
Add GauXC as an optional backend for the exchange-correlation quadrature

wafgo:
## Description

Psi4 already builds and links GauXC when `ENABLE_gauxc=ON`, and `snLinK.cc`
drives it through `eval_exx` for the exchange matrix. But
`XCIntegrator::eval_exc_vxc` is never called anywhere in the tree — the XC
quadrature itself always runs through `RV::compute_V` / `UV::compute_V` on the
CPU. `codedeps.yaml` records this explicitly:

```
required_note: "Allow using the GauXC density functional library for LinK (not DFT)."
```

This PR closes that gap: a new option `DFT_V_ALGORITHM = INTERNAL|GAUXC`
(default `INTERNAL`, existing behaviour untouched) routes the XC quadrature
through GauXC, optionally on the GPU via `DFT_V_USE_GPU`.

Motivation, measured rather than assumed: in a profile of Psi4 1.11, the XC
quadrature accounts for **70 % of SCF wall time** for (H2O)8/cc-pVTZ/BP86 — of
which 70 % is collocation and density evaluation and only 2 % the actual libxc
evaluation. The cost sits exactly where GauXC operates.

- [ ] New option `DFT_V_ALGORITHM = INTERNAL|GAUXC` selects the backend for the
      exchange-correlation quadrature. Default `INTERNAL` — no existing
      calculation changes.
- [ ] With `DFT_V_ALGORITHM GAUXC`, `DFT_V_USE_GPU` runs that quadrature on the
      GPU. Requires a GauXC built with `-Dgauxc_ENABLE_GPU=ON`.
- [ ] Both options are no-ops unless Psi4 was built with `ENABLE_gauxc=ON`;
      requesting them otherwise raises a clear error.

## Dev notes & details

- [ ] Lifted the Psi4→GauXC conversion out of `snLinK.cc` into a shared
      `gauxc_interface.{h,cc}` (molecule, basis, CCA permutation,
      spherical→Cartesian, runtime environment). Pure refactor, no behaviour
      change; `snLinK.cc` shrinks by 139 lines.
- [ ] `GauXCRV : public RV` and `GauXCUV : public UV` override only
      `compute_V`. Gradients, VV10 and everything else keep using the existing
      code paths.
- [ ] Selection happens in the `VBase::build_V` factory. `RHF/UHF::setup_potential`
      previously bypassed that factory; this PR routes them through it.
- [ ] Grid parameters come from the existing `DFT_*` options, so both backends
      integrate on the same grid and are directly comparable. Worth noting for
      anyone benchmarking: GauXC partitions with Stratmann-Scuseria-Frisch,
      Psi4 defaults to Treutler — without matching them the partitioning
      dominates the difference by roughly a factor of 30.
- [ ] No functional whitelist. The ExchCXX functional is constructed from
      Psi4's own `SuperFunctional` decomposition — each ingredient carries its
      libxc kernel name, and ExchCXX builds kernels from exactly that. Covers
      LDA, GGA, meta-GGA, global and range-separated hybrids.
- [ ] Hard errors instead of silent divergence for VV10, GRAC, and grid schemes
      with no GauXC equivalent.
- [ ] `tests/pytests/test_gauxc_v.py`: 17 tests, skipped when GauXC is absent.
- [ ] The harness I used to produce the numbers below (build + three-way
      comparison + timings in one run) is deliberately *not* part of this PR --
      `devtools/scripts/` is CI-only by convention and this is scaffolding, not
      product. Happy to attach it to a comment or open a follow-up if reviewers
      want to re-measure on their own hardware.

**Numerics** (water, cc-pVDZ, (75,302), `DFT_NUCLEAR_SCHEME STRATMANN` on both sides)

| Functional | Class | ΔE vs. internal |
|---|---|---|
| svwn | LDA | −5.9e−08 |
| blyp / pbe | GGA | −1.1e−07 / −9.7e−08 |
| b3lyp / pbe0 | global hybrid | −8.3e−08 / −6.9e−08 |
| m06-2x | meta-GGA hybrid | 8.4e−07 |
| wb97x | range-separated | −3.6e−07 |
| svwn/blyp/b3lyp/m06-2x | UKS | < 1.1e−07 |

The residual comes from point screening and shrinks on finer grids (UKS:
4e−07 at (75,302) → 8e−08 at (99,590)).

**CPU performance** (glycine, B3LYP, 16 threads, quadrature time only)

| Basis | GauXC vs. internal |
|---|---|
| cc-pVDZ | 0.48× (slower) |
| cc-pVTZ | **2.29×** |
| cc-pVQZ | **2.02×** |

Small bases lose — the setup does not pay off there. This is one reason the
default stays `INTERNAL`.

**GPU**

Verified on a GeForce GT 1030 (sm_61, CUDA 12.6): the GPU path reproduces the
GauXC CPU results to all printed digits, for every functional above, RKS and
UKS, cc-pVDZ and cc-pVTZ.

Being straight about speed: on that card the GPU is **slower** than 32 CPU
threads. The code is FP64 throughout and NVIDIA restricts FP64 on consumer
cards to 1/32–1/64 of FP32. That card demonstrates correctness, not
performance; the latter needs datacenter hardware (FP64 1:2), which I do not
have access to. **So the GPU path is presented as verified-correct, not as
verified-fast.** If someone with a V100/A100 has cycles to spare, I am glad
to hand over the harness I used so the numbers are directly comparable.

**Tests**

Full regression on this branch: 192 passed, 117 skipped, 0 failed.
`test_gauxc_v.py`: 17/17.

## Questions

- [ ] Was leaving the XC side out a deliberate decision during the sn-LinK
      work, or simply out of scope? If deliberate, I would like to understand
      the objection before you spend review time on this.
- [ ] Is `DFT_V_ALGORITHM` the right name and place for the switch? It mirrors
      how `SCF_TYPE` already selects sn-LinK, but I have no attachment to it.
- [ ] Routing `RHF/UHF::setup_potential` through `VBase::build_V` is the one
      change that touches existing behaviour. It looked like an oversight
      rather than intent — please confirm.
- [ ] Interest in a `GAUXC_ENABLE_HIP` mapping in
      `external/upstream/gauxc/CMakeLists.txt`? Only `GAUXC_ENABLE_CUDA` is
      wired up (line 72), though GauXC supports AMD as well. Happy to add it,
      but I cannot test it.
- [ ] `-Dgauxc_ENABLE_GPU=ON` is mandatory when building Psi4, otherwise the
      interface is silently built CPU-only with only a CMake *warning*. That
      cost me a day. Would you take a patch making it a hard error?

## AI

- [ ] Substantial. An AI assistant (Claude, via GitHub Copilot CLI) did the
      prior-art survey, wrote most of the implementation, the tests, the
      verification harness and the docs, and found the spherical-d/f segfault
      described below. I directed the work, made the design decisions, ran and
      reviewed everything on my own hardware, and I am the one answering
      questions about it.
- [ ] Reviewers may want to look hardest at the segfault fix in `GauXCV.cc`:
      in the GPU path the density is transformed to Cartesian (25×25 for water
      /cc-pVDZ), after which the *spherically* sized permutation matrix (24×24)
      was still being applied. Eigen does not check dimensions in release
      builds (`NDEBUG`), so this wrote past the buffer. The fix drops the
      permutation in the Cartesian path entirely — it only reorders *within*
      solid-harmonic shells, which do not exist in the Cartesian basis. It is
      only ever active with `psi4_SHGSHELL_ORDERING=2` (Gaussian ordering).
      It went unnoticed at first because sto-3g (s/p only) and 6-31G
      (Cartesian) never trigger the mismatch.

## Checklist

## Status
- [x] Ready for review

### DIFF
--- a/doc/sphinxman/source/dft.rst
+++ b/doc/sphinxman/source/dft.rst
@@ -450,11 +450,48 @@ implemented in |PSIfour| for correctness. If you find an error in a DFT
 functional or have a request for a new functional, please let us know on our
 forum or GitHub page.
 
+.. _`sec:xc-quadrature-backend`:
+
+Exchange-Correlation Quadrature Backend
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The numerical integration of the exchange-correlation potential is selected
+with |scf__dft_v_algorithm|:
+
+``INTERNAL`` (default)
+    |PSIfour|'s own grid code. Supports every functional |PSIfour| knows,
+    including VV10 non-local correlation and GRAC asymptotic corrections.
+
+``GAUXC``
+    Delegates the quadrature to the `GauXC <https://github.com/wavefunction91/GauXC>`_
+    library. Requires |PSIfour| built with ``ENABLE_gauxc=ON``. Supports RKS
+    and UKS with LDA, GGA, meta-GGA, global hybrid and range-separated hybrid
+    functionals. GauXC additionally offers GPU execution, enabled with
+    |scf__dft_v_use_gpu|.
+
+The ``GAUXC`` backend does *not* support VV10 or GRAC and rejects those
+functionals with an explicit message. It also requires grid settings that have
+a GauXC counterpart: |scf__dft_pruning_scheme| must be ``ROBUST``, ``TREUTLER``
+or ``NONE``, |scf__dft_radial_scheme| must be ``TREUTLER``, ``MURA`` or ``EM``,
+and |scf__dft_nuclear_scheme| must be ``BECKE`` or ``STRATMANN``.
+
+Note that GauXC partitions atomic weights with the Stratmann-Scuseria-Frisch
+scheme, whereas the |PSIfour| default is Treutler. When comparing the two
+backends numerically, set ``DFT_NUCLEAR_SCHEME STRATMANN`` on both sides;
+otherwise the difference is dominated by the partitioning, not the quadrature.
+
+Both backends agree to better than 1e-6 :math:`E_h` on total energies. The
+residual difference stems from how the two libraries sieve grid points and
+weights, and shrinks as the grid is refined.
+
+On CPU, which backend is faster depends strongly on system size: for small
+basis sets the internal code wins, while from roughly cc-pVTZ upwards GauXC is
+about twice as fast (glycine, 16 threads, (75,302) grid).
+
 .. _`sec:grid-selection`:
 
 Grid Selection
 ~~~~~~~~~~~~~~
-
 |PSIfour| uses the standard Lebedev-Laikov spherical quadratures in concert with a
 number of radial quadratures and atomic partitioning schemes. 
 The default grid in |PSIfour| is a Lebedev-Treutler (75,302) grid with a Treutler
--- a/doc/sphinxman/source/dft.rst
+++ b/doc/sphinxman/source/dft.rst
@@ -488,6 +488,26 @@ On CPU, which backend is faster depends strongly on system size: for small
 basis sets the internal code wins, while from roughly cc-pVTZ upwards GauXC is
 about twice as fast (glycine, 16 threads, (75,302) grid).
 
+GPU execution
+^^^^^^^^^^^^^
+
+Setting |scf__dft_v_use_gpu| runs the quadrature on the GPU. This requires a
+GauXC installation built with device support (``-Dgauxc_ENABLE_GPU=ON``, plus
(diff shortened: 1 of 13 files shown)


JonathonMisiewicz:
Hello, thanks for your interest in allowing Psi4 to use GauXC as an integration backend.

This work is already in-progress, as part of PR #3355, and I'm spending this weekend trying to do the final GPU tests before I put this up formally for review. (Unfortunately, all the GPU computations on my cluster are taking a very long time to get out of queue.) You are welcome to give comments there. We've made some more infrastructure changes that will let it integrate with Psi4 better.

PR #3504, opened 2026-08-24
Update matrix includes and Eigen for faster builds

NathanGillispie:
## Description

From the previous work on removing unused includes, I said that a different approach was needed. I came to this conclusion because GCC shows that for most files, a majority of compilation is spent on parsing. \[This is honestly surprising to me since I would assume optimizations or conversion to IR or machine code would be longer.\] This step involves making the AST, which should be smaller when fewer headers are included. After removing nearly 2.5k unused includes I found a 3%, albeit statistically insignificant, reduction in build times. This insignificant effect is likely because the preprocessor removed redundant includes, and few real changes were made. Unlike #3456, this actually does tip the needle.

**Note:** By "compile-only", I mean the sum of all compilation steps for each TU. That does not include linking, for example. I also mean non-unity builds, which is how I benchmarked build times unless otherwise stated. "compile-only" improvements apply most directly to a serial, non-unity build. In practice, unity builds and imperfect parallelism mean that common build configurations do not see the same improvements in wall time.

> [!NOTE]
> Bottom line: Combined with other improvements outside this PR, I was able to get -37% compile-only times (1.6x faster!) with minimal API changes. That translates to a real wall time reduction of -11% for unity builds compared to master. For this PR, l measured a compile-only reduction of -15%, and a wall-time reduction of approximately -8%. 

As stated previously, the majority of build time is spent parsing. To find out where improvements could be made, I started by running some analysis.
1. I used the `compile_commands.json` file of a non-unity build to find a TUs compile command.
2. For each TU, I ran `clang++ -fsyntax-only -ftime-trace` which produces a JSON file. From that, you can extract how long parsing took for each header in that TU.
3. For each header measured, I produced a cost by summing the time spent parsing that header over all TUs. Thus, one should be able to reduce build times by reducing how long it takes to parse the header or reducing the number of times it is parsed.
4. For the highest cost headers, the next thing to do is just look at the data and figure out a plan of action. This can involve adjusting how this header is included in other files, or adjusting the header itself. This PR does both for `matrix.h`.

Some headers like `diismanager.h` didn't need `matrix.h` at all so it was just deleted. Some headers only need Matrix to be forward declared, so I could just swap to `typedefs.h`. Others just needed a refactor like `ccdensity/MOInfo.h` (different PR).

`matrix.h` was the most costly header. By looking at the timings for includes of `matrix.h`, I found that only 2% of the time spent parsing `matrix.h` was on the real body. The largest portion of time was spent parsing Eigen/Core and its friends. This PR shuffles things around for Eigen, putting it in its own header and source file to be included by the only real user of Eigen: `snLinK.cc`.

- [x] `Matrix::eigen_map[s]()` member methods have been moved to free methods `linalg::eigen_map[s](Matrix&)`.

## Dev notes & details

- [x] Creates `matrix_eigen.h` and `matrix_eigen.cc` to allow for the removal of the `<eigen3/Eigen/Core>` include from `matrix.h`.

## AI

- [x] AI wrote scripts automating the include analysis. AI was also used to automate builds, measur times, and report back on a separate computer.

## Checklist

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libmints/matrix.h
+++ b/psi4/src/psi4/libmints/matrix.h
@@ -36,7 +36,6 @@
 
 #include "dimension.h"
 
-#include <eigen3/Eigen/Core>
 #ifdef USING_OpenOrbitalOptimizer
 #ifdef USING_LAPACK_MKL
 #include <mkl.h>
@@ -291,9 +290,6 @@ class PSI_API Matrix : public std::enable_shared_from_this<Matrix> {
     void copy(const Matrix* cp);
     /** @} */
 
-    /// returns an Eigen::Map object to the underlying matrix data buffer
-    Eigen::Map<Eigen::MatrixXd> eigen_map();
-    std::vector<Eigen::Map<Eigen::MatrixXd>> eigen_maps();
     /// Returns an Armadillo matrix
     arma::mat to_armadillo_matrix(int h=0);
     /// Copies data from an Armadillo matrix
--- a/psi4/src/psi4/libmints/matrix.h
+++ b/psi4/src/psi4/libmints/matrix.h
@@ -28,6 +28,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <string>
 #include <vector>
 #include <memory>
--- a/psi4/src/psi4/libmints/matrix_eigen.h
+++ b/psi4/src/psi4/libmints/matrix_eigen.h
@@ -1,0 +1,45 @@
+/*
+ * @BEGIN LICENSE
+ *
+ * Psi4: an open-source quantum chemistry software package
+ *
+ * Copyright (c) 2007-2026 The Psi4 Developers.
+ *
+ * The copyrights for code used from other parties are included in
+ * the corresponding files.
+ *
+ * This file is part of Psi4.
+ *
+ * Psi4 is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, version 3.
+ *
+ * Psi4 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with Psi4; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * @END LICENSE
+ */
+
+#pragma once
+
+#include "psi4/libmints/matrix.h"
+
+#include <eigen3/Eigen/Core>
+
+namespace psi {
+namespace linalg {
+
+/// Map a single-irrep Psi4 Matrix onto its underlying data.
+PSI_API Eigen::Map<Eigen::MatrixXd> eigen_map(Matrix& matrix);
+
+/// Map each irrep block of a Psi4 Matrix onto its underlying data.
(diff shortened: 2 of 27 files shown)


JonathonMisiewicz (approved):
With this refactor, this PR *must* come in before my GauXC PR can proceed. I'll need to edit my own code.

NathanGillispie:
Wow thanks for the fast turnaround time.

`SplitJK.h` includes GauXC stuff which is itself included by jk.h which is included by a ton of TUs. I have a [commit](https://github.com/psi4/psi4/commit/ff2d6e6f01b8047ab0a0f28acdbee7671a33dfbc) that splits the GauXC portion from `SplitJK.h` and puts it into `snLinK.h`. Let me know if that conflicts with your stuff.

JonathonMisiewicz:
Yes, it conflicts. Just at a glance, conceptually `snLinK` has zero business owning the conversion from `psi4_to_gauxc_molecule`, and that conflicts with the fact that the new `GauXCBase` class [needs that](https://github.com/psi4/psi4/pull/3355/files#diff-1bf57182e170e0fd643d545212cbc5b43ba496e6e38217c46c0d71b31a1cdb3cR68). I will block any PR that contains that change.

Please check changes to `snLinK` from PR #3355. Any functions I moved out of there have probably been moved elsewhere and need to be accessed by code other than `snLinK`. Having `snLinK` own them is not an acceptable solution. Let me know if you need further guidance.

NathanGillispie:
Ok I'll keep that in mind. I'll just drop that commit since so much will change with GauXC anyways. I will add that it was number 3 in terms of significant changes to build times.

JonathonMisiewicz:
I'm perfectly happy to have a special header in `libmints` for GauXC-tech that will only be included where needed. That should keep build times quick while keeping the code accessible for any files that truly need it.

psi4/src/psi4/libmints/matrix_eigen.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#pragma once
+
+#include "psi4/libmints/matrix.h"
+
+#include <eigen3/Eigen/Core>
+
+namespace psi {
+namespace linalg {
```
TiborGY, line 36: I am not sure if I like the new namespace being called `linalg`. Currently it only contains eigen, not all kinds of linalg, and there is a lot of arguably linalg stuff (BLAS/LAPACK, etc.) outside of `namespace linalg`. Since Eigen wraps their symbols in the Eigen namespace anyways, do we need this new namespace? Or perhaps we could come up with a better name for it.

NathanGillispie: Yeah I don't love that option. My thinking was that linalg is usually where the details and free methods go to for Matrix. I'll look into the eigen namespaces, but if that's not ideal we should come up with one.

loriab: iirc, Roberto DiRemigio invented `linalg::` to free doublet and triplet from `Matrix`. So while the linalg namespace is much less than it says on the tin, I think it's ok style-wise. I'm not adamant either way.

susilehtola: I think we should continue with `linalg::` for that reason

TiborGY: Alright, thanks for the discussion, I am OK with `linalg::` given this.

psi4/src/psi4/libmints/matrix.h
```diff
@@ -36,7 +36,6 @@
 
 #include "dimension.h"
 
-#include <eigen3/Eigen/Core>
```
loriab, line 39: Yeah, I remember David Poole really trying to avoid including the full header, but nothing else would do. Glad to get it out of the main matrix.h

loriab (approved):
lgtm, thanks!

susilehtola:
It's failing to build; fixes coming up

PR #3505, opened 2026-08-24
Fix MP2-F12 when linear dependencies are removed from a basis

### LINKED ISSUE #3498: MP2-F12 fails, when linear dependencies removed from orbital basis
If some basis functions are removed due to linear dependencies in the MO space, the MP2-F12 calculation fails.

A minimal example is:

```python

import psi4

hcn_dimer = psi4.geometry("""
0 1
H    0.000000000  0.000000000  3.855213060
C    0.000000000  0.000000000  2.786499760
N    0.000000000  0.000000000  1.631507910
H    0.000000000  0.000000000 -0.593774920
C    0.000000000  0.000000000 -1.668098240
N    0.000000000  0.000000000 -2.825250560

symmetry c1
""")


psi4.set_options({'basis':'d-aug-cc-pvqz',
                   'df_basis_f12':'aug-cc-pvqz-ri',
                   'cabs_basis': 'aug-cc-pvqz-ri',
                   'scf_type': 'df',
                        'freeze_core': 'False',
                        'e_convergence': 1e-8,
                        'd_convergence': 1e-8,
                        'INTS_TOLERANCE': 1e-15,
    })

energy = psi4.energy('mp2-f12')

print("MP2-F12 energy:", energy)
```

With the default `S_TOLERANCE`, some vectors are removed because of linear dependencies, and the MP2-F12 code fails due to the matrix size mismatch in the formation of CABS.
```
 Complementary Auxiliary Basis Set (CABS)
    Orthogonalizing basis for space RIBS.
    39 linear dependencies will be 'removed'.
    Projecting out 'ALL' from 'RIBS' to obtain space 'CABS'
Row and column block dimensions of A
   (n = 1): 544
   (n = 1): 1376
Row and column block dimensions of B
   (n = 1): 1376
   (n = 1): 1376
Row and column block dimensions of C
   (n = 1): 542
   (n = 1): 1376
Asym: 0, Bsym: 0, Csym: 0
Mismatch in index Ha = 0
  File "/home/dominik/test.py", line 28, in <module>
    energy = psi4.energy('mp2-f12')
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dominik/miniconda3/envs/psi4_2026/lib/python3.11/site-packages/psi4/driver/driver.py", line 527, in energy
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dominik/miniconda3/envs/psi4_2026/lib/python3.11/site-packages/psi4/driver/procrouting/proc.py", line 4492, in run_mp2f12
    mp2f12_wfn.compute_energy()
RuntimeError:
Fatal Error: Matrix::gemm error: Number of rows and columns do not match.
Error occurred in file: /home/conda/feedstock_root/build_artifacts/psi4_1770332549233/work/psi4/src/psi4/libmints/matrix.cc on line: 1441

psi::OrbitalSpace::build_cabs_space(psi::OrbitalSpace const&, psi::OrbitalSpace const&, double)
```

If `S_TOLERANCE` is decreased sufficiently so that all vectors are retained, the calculation runs successfully.

JustinKyleKirkland:
Closes #3498.

Three related problems kept MP2-F12 from running once linear dependencies were removed:

* `orthogonal_complement` built the MO overlap as S12 C2, labelling its rows with the *orbital* count of space1 while actually producing one row per *basis function*. Those agree only when nothing was removed from space1; otherwise the gemm aborts with a row/column mismatch. Transform both sides, C1^T S12 C2.

* `orthogonalize` zeroed the dependent directions of S^-1/2 instead of dropping them, so the RI space kept the full basis dimension and handed its null vectors on to CABS. Switch to canonical orthogonalization, X = U_kept s_kept^-1/2, when there is something to remove; bases without linear dependencies still get the symmetric orthogonalization as before.

* MP2F12 set `nobs_` from the number of basis functions rather than the number of orbitals, which overruns the reference coefficient matrix and mis-sizes the one-electron MO integrals.

The orbital space tests now cover linear dependencies in the orbital basis, in the RI basis, and in both.

- [x] MP2-F12 no longer aborts with "Matrix::gemm error: Number of rows and columns do not match" when linear dependencies are removed from the orbital basis (#3498).
- [x] The RI space now really drops linearly dependent vectors instead of retaining them as null vectors, so CABS comes out orthonormal and with the right dimension.

## Dev notes & details

- [x] OrbitalSpace::orthogonal_complement built the MO overlap as S12 C2 but allocated it with space1.C()->colspi() rows. That product has one row per basis function of space1, not one per orbital, and the two only agree when nothing was removed from space1. It now transforms both sides, C1^T S12 C2. This is the immediate cause of the traceback in the issue.
- [x] orthogonalize, which builds the RI space, zeroed the dependent directions of S^-1/2 instead of dropping them. That leaves a rank deficient square matrix, so the space keeps claiming the full basis dimension and hands its null vectors on to CABS, which is then not orthonormal. It now uses canonical orthogonalization, X = U_kept s_kept^-1/2, when there is something to remove. Bases with no linear dependencies still get symmetric orthogonalization as before, so nothing changes for cases that already worked.
- [x] MP2F12::common_init took nobs_ from basisset()->nbf() rather than the number of orbitals. Once linear dependencies are removed that overruns the reference coefficient matrix in convert_C and mis-sizes the one-electron MO integrals in form_oeints, which does C1^T h C2 and so returns nmo by nmo.
- [x] tests/pytests/test_orbitalspace.py now runs the existing orthonormality checks over four cases: no linear dependencies, dependencies in the orbital basis, in the RI basis, and in both. The orbital basis cases truncate the SCF coefficients rather than leaning on an ill conditioned basis, so they are deterministic. A CABS dimension check is also added.

Testing done locally (macOS arm64, conda toolchain, Einsums 1.1.5):
- Water / d-aug-cc-pVTZ with S_TOLERANCE 1e-4 removes 2 of 126 orbitals and reproduces the reported gemm error on master. With this branch it runs: -76.3880318670 with nothing removed, -76.3879290140 with 2 removed, -76.3874095731 with 6 removed (S_TOLERANCE 1e-3). The issue's own example (HCN dimer, d-aug-cc-pVQZ) is too large to run here, so this is the same code path on a smaller system.
- On master the new test cases fail as expected: the orbital basis ones error in the fixture on the gemm, and ribs_lindep fails the CABS orthonormality check without any crash.
- ctest mp2f12-1 and dfmp2f12-1 pass with unchanged energies, and the quicktests label is 190/190.

## AI

- [x] Used to create local test cases in order to reproduce the issue.

## Checklist
- [ ] Docstring or narrative docs/sphinxman/source updated if needed

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libmints/orbitalspace.cc
+++ b/psi4/src/psi4/libmints/orbitalspace.cc
@@ -171,28 +171,54 @@ OrbitalSpace orthogonalize(const std::string &id, const std::string &name, const
     SharedMatrix overlap = OrbitalSpace::overlap(bs, bs);
     Dimension SODIM = overlap->rowspi();
     auto evecs = std::make_shared<Matrix>("evecs", SODIM, SODIM);
-    auto sqrtm = std::make_shared<Matrix>("evecs", SODIM, SODIM);
     auto evals = std::make_shared<Vector>("evals", SODIM);
 
-    int nlindep = 0;
     overlap->diagonalize(evecs, evals);
+
+    Dimension nlindep(SODIM.n());
     for (int h = 0; h < SODIM.n(); h++) {
         for (int i = 0; i < SODIM[h]; i++) {
-            if (std::fabs(evals->get(h, i)) > lindep_tol) {
-                sqrtm->set(h, i, i, 1.0 / sqrt(evals->get(h, i)));
-            } else {
-                sqrtm->set(h, i, i, 0.0);
-                nlindep++;
-            }
+            if (std::fabs(evals->get(h, i)) <= lindep_tol) nlindep[h]++;
         }
     }
 
-    sqrtm->back_transform(evecs);
-
-    outfile->Printf("    %d linear dependencies will be \'removed\'.\n", nlindep);
+    outfile->Printf("    %d linear dependencies will be \'removed\'.\n", nlindep.sum());
 
     auto localfactory = std::make_shared<IntegralFactory>(bs);
-    return OrbitalSpace(id, name, sqrtm, bs, localfactory);
+
+    if (nlindep.sum() == 0) {
+        // Symmetric orthogonalization, S^{-1/2} = U s^{-1/2} U^T.
+        auto sqrtm = std::make_shared<Matrix>("S^-1/2", SODIM, SODIM);
+        for (int h = 0; h < SODIM.n(); h++) {
+            for (int i = 0; i < SODIM[h]; i++) {
+                sqrtm->set(h, i, i, 1.0 / sqrt(evals->get(h, i)));
+            }
+        }
+        sqrtm->back_transform(evecs);
+        return OrbitalSpace(id, name, sqrtm, bs, localfactory);
+    }
+
+    // Symmetric orthogonalization cannot drop the dependent directions: zeroing their
+    // contribution leaves a rank-deficient square matrix, so the space keeps claiming
+    // the full basis dimension and passes null vectors on to whatever is built from it.
+    // Fall back to canonical orthogonalization, X = U_kept s_kept^{-1/2}, which really
+    // has the reduced column dimension.
+    Dimension NDIM = SODIM - nlindep;
+    auto X = std::make_shared<Matrix>("X (canonical orthogonalization)", SODIM, NDIM);
+    for (int h = 0; h < SODIM.n(); h++) {
+        int col = 0;
+        for (int i = 0; i < SODIM[h]; i++) {
+            double eval = evals->get(h, i);
+            if (std::fabs(eval) <= lindep_tol) continue;
+            double scale = 1.0 / sqrt(eval);
+            for (int j = 0; j < SODIM[h]; j++) {
+                X->set(h, j, col, evecs->get(h, j, i) * scale);
+            }
+            col++;
+        }
+    }
+
+    return OrbitalSpace(id, name, X, bs, localfactory);
 }
 
 OrbitalSpace orthogonal_complement(const OrbitalSpace &space1, const OrbitalSpace &space2, const std::string &id,
@@ -209,9 +235,14 @@ OrbitalSpace orthogonal_complement(const OrbitalSpace &space1, const OrbitalSpac
     // Overlap Matrix
     SharedMatrix O12 = OrbitalSpace::overlap(space1, space2);
 
-    // Half-transform to RIBS
+    // Transform the overlap into the orbital basis of both spaces, C1^T S12 C2.
+    // The row dimension of space1 is its number of orbitals, which is smaller than its
+    // number of basis functions whenever linear dependencies were removed from it.
+    auto O12C2 = std::make_shared<Matrix>("S12 C2", O12->rowspi(), space2.C()->colspi());
+    O12C2->gemm(false, false, 1.0, O12, space2.C(), 0.0);
+
     auto C12 = std::make_shared<Matrix>("C12", space1.C()->colspi(), space2.C()->colspi());
-    C12->gemm(false, false, 1.0, O12, space2.C(), 0.0);
+    C12->gemm(true, false, 1.0, space1.C(), O12C2, 0.0);
 
     // SVD of MO overlap matrix
     auto [U, S, Vt] = C12->svd_a_temps();
--- a/psi4/src/psi4/libmints/orbitalspace.cc
+++ b/psi4/src/psi4/libmints/orbitalspace.cc
@@ -203,11 +203,8 @@ OrbitalSpace orthogonal_complement(const OrbitalSpace &space1, const OrbitalSpac
     // Transform the overlap into the orbital basis of both spaces, C1^T S12 C2.
     // The row dimension of space1 is its number of orbitals, which is smaller than its
     // number of basis functions whenever linear dependencies were removed from it.
-    auto O12C2 = std::make_shared<Matrix>("S12 C2", O12->rowspi(), space2.C()->colspi());
-    O12C2->gemm(false, false, 1.0, O12, space2.C(), 0.0);
-
-    auto C12 = std::make_shared<Matrix>("C12", space1.C()->colspi(), space2.C()->colspi());
-    C12->gemm(true, false, 1.0, space1.C(), O12C2, 0.0);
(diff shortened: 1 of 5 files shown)


jturney:
I'd swap the hand-rolled orthogonalization for `BasisSetOrthogonalization::Automatic` (which also fixes the negative-eigenvalue NaN for free):

```cpp
BasisSetOrthogonalization orthog(BasisSetOrthogonalization::Automatic, overlap, lindep_tol,
                                 cholesky_tol, print);
auto X = orthog.basis_to_orthog_basis();
outfile->Printf("    %d linear dependencies will be 'removed'.\n", orthog.nlindep());
return OrbitalSpace(id, name, X, bs, std::make_shared<IntegralFactory>(bs));
```

JustinKyleKirkland:
Swapped it in, thanks. Automatic picks canonical on my test case (min eigenvalue 6.7e-11, rcond 4.4e-12, 3 of 324 dropped), energies unchanged to all printed digits, mp2f12-1/dfmp2f12-1 and quicktests still pass.

Two gotchas. `nlindep()` returns the number of functions kept, not removed, so that printf would say 1337 instead of 39 for the case in #3498. `dim()` and `orthog_dim()` don't help either: they return `Matrix::rowdim()`/`coldim()`, which take a defaulted irrep and return an int, and `Dimension(size_t)` isn't explicit, so you get a Dimension of N zeros. Both sum to 0. I counted off X instead:

```cpp
outfile->Printf("    %d linear dependencies will be 'removed'.\n",
                overlap->rowspi().sum() - X->colspi().sum());
```

Nothing calls those three today, hf.cc and dlpno.cc read their dimensions off `X_` directly, so I left the class alone. Say if you'd rather they were fixed here.

Second, there's no Cholesky tolerance in scope in orbitalspace.cc, so I added one to `build_ri_space` defaulting to 1e-8 and exposed it in the binding. Can pull `S_CHOLESKY_TOLERANCE` from the options instead if you prefer.

Print level is 1 so the method choice and eigenvalue diagnostics land in the output file. Easy to turn down.

jturney:
Oh good catch on the `nlindep`

psi4/src/psi4/libmints/orbitalspace.cc
```diff
@@ -209,9 +200,14 @@ OrbitalSpace orthogonal_complement(const OrbitalSpace &space1, const OrbitalSpac
...
-    // Half-transform to RIBS
+    // Transform the overlap into the orbital basis of both spaces, C1^T S12 C2.
+    // The row dimension of space1 is its number of orbitals, which is smaller than its
+    // number of basis functions whenever linear dependencies were removed from it.
+    auto O12C2 = std::make_shared<Matrix>("S12 C2", O12->rowspi(), space2.C()->colspi());
+    O12C2->gemm(false, false, 1.0, O12, space2.C(), 0.0);
+
     auto C12 = std::make_shared<Matrix>("C12", space1.C()->colspi(), space2.C()->colspi());
-    C12->gemm(false, false, 1.0, O12, space2.C(), 0.0);
+    C12->gemm(true, false, 1.0, space1.C(), O12C2, 0.0);
```
andyj10224: It may be better to do

```
auto O12C2 = linalg::triplet(space1.C(), O12, space2.C(), true, false, false);
```

Two advantages: the matrix multiplication can now be done in one line, and `triplet` gives you the ideal contraction order, allowing the multiplication to be done in the most efficient way :)

andyj10224 (approved):
LGTM, my `triplet` fix is a strong suggestion, do with it however you choose :)

psi4/src/psi4/libmints/orbitalspace.cc
JustinKyleKirkland: Nice, that's cleaner. Swapped in, and it drops the temporary along with both gemms. Energies identical, f12 ctests and the pytests still pass.

One aside while I was in there: the cost comparison in `triplet` accumulates into `int`, and this case is already at 1.4e9, so a big enough pair of spaces wraps it and the order gets picked at random. Nothing to do with this PR, but might be worth an issue.

psi4/src/psi4/libmints/orbitalspace.cc
```diff
@@ -165,34 +165,25 @@ void OrbitalSpace::print() const {
...
-            }
-        }
-    }
 
-    sqrtm->back_transform(evecs);
+    // Symmetric orthogonalization while the basis is well conditioned, canonical once
+    // vectors have to be dropped, partial Cholesky when the overlap is too ill conditioned
+    // for that. The dependent directions are really removed from X rather than zeroed, so
+    // the space that comes back has the reduced dimension instead of trailing null vectors.
+    BasisSetOrthogonalization orthog(BasisSetOrthogonalization::Automatic, overlap, lindep_tol, cholesky_tol, 1);
```
EricaCMitchell, line 177: > Print level is 1 so the method choice and eigenvalue diagnostics land in the output file. Easy to turn down.

This should follow the global option `Process::environment.options.get_int("PRINT")`

EricaCMitchell (changes requested):
Thanks for the fix! I had one comment on the printing.

For the other gotchas:
> nlindep() returns the number of functions kept, not removed, so that printf would say 1337 instead of 39

This should be fixed in a separate PR since it is [documented correctly](https://github.com/psi4/psi4/blob/f234e73af37c322a9c2af1a2e522daa1c5046516/psi4/src/psi4/libmints/orthog.h#L138-L141) but doesn't return what you'd think. I'm not sure if it will mess with anyone else's code and make this PR less focused on F12.

> there's no Cholesky tolerance in scope in orbitalspace.cc, so I added one to build_ri_space defaulting to 1e-8 and exposed it in the binding. Can pull S_CHOLESKY_TOLERANCE from the options instead if you prefer.

This is an SCF specific keyword so it shouldn't be used that keyword shouldn't be used here for F12.

EricaCMitchell:
Probably easiest to get this in before #3394 since it's smaller and would be a simple rebase

EricaCMitchell (approved):

loriab (approved):
Thanks for the lindep fix!

PR #3506, opened 2026-08-25
Higher Order DLPNO Methods (CCSDT, CCSDT(Q), and CCSDTQ)

andyj10224:
## Description

This draft PR extends Psi4’s in-tree domain-based local pair natural orbital (DLPNO) coupled-cluster hierarchy beyond DLPNO-CCSD(T), adding energy implementations for:

* DLPNO-CCSDT with iterative full triples
* DLPNO-CCSDT(Q0) with semicanonical perturbative quadruples
* DLPNO-CCSDT(Q) with iterative perturbative quadruples
* DLPNO-CCSDTQ with iterative full triples and quadruples

The implementation extends the existing localized occupied-orbital and pair-natural-orbital framework with triples natural orbital (TNO), quadruples natural orbital (QNO), and extended pair natural orbital (XPNO) spaces. It also provides screening thresholds, iterative solver controls, and disk-backed storage options needed for the higher-order amplitude equations and intermediates.

The scientific implementation was originally developed before DLPNO-CCSD(T) was incorporated into Psi4 (in version 1.11). This PR ports and reconciles that work with the current Psi4 DLPNO implementation, driver infrastructure, option system, documentation, and testing framework.

The implemented methods follow the formulations described in:

* [[DLPNO-CCSDT](https://doi.org/10.1021/acs.jctc.4c01716)](https://doi.org/10.1021/acs.jctc.4c01716)
* [[DLPNO-CCSDT(Q)](https://doi.org/10.1063/5.0257672)](https://doi.org/10.1063/5.0257672)
* [[DLPNO-CCSDTQ](https://doi.org/10.1021/acs.jctc.5c01910)](https://doi.org/10.1021/acs.jctc.5c01910)

The methods currently support closed-shell RHF reference wavefunctions and operate in (C_1) symmetry. They presently provide energies only.

* [x] Add `energy('dlpno-ccsdt')` for DLPNO-CCSDT energies with iterative full triples.
* [x] Add `energy('dlpno-ccsdt(q0)')` and `energy('dlpno-ccsdt(q)')` for semicanonical and iterative perturbative-quadruples corrections.
* [x] Add `energy('dlpno-ccsdtq')` for DLPNO-CCSDTQ energies with iterative full quadruples.
* [x] Add TNO, QNO, and XPNO truncation thresholds and expert controls for higher-order amplitude iterations, damping, DIIS, memory use, and disk storage.
* [x] Add the new methods to the energy driver and CBS method hierarchy.
* [x] Automatically request a (C_1) SCF reference when the reference wavefunction is constructed internally. A user-supplied reference wavefunction must already be in (C_1) symmetry.
* [x] Document that the new methods are currently restricted to closed-shell RHF energy computations.
* [x] Added chunking for DIIS when the number of T3 and T4 amplitudes that need to be extrapolated exceeds `INT_MAX` (2,147,483,647, or 2^31 - 1) doubles.

## Dev notes & details

* [x] Extend the DLPNO method enumeration, class hierarchy, wrapper dispatch, and energy-variable handling through CCSDTQ.
* [x] Implement iterative (T_3) residual equations and the associated TNO-domain transformations, screening, intermediates, and amplitude storage.
* [x] Implement semicanonical and iterative ((Q)) corrections in QNO spaces.
* [x] Implement iterative (T_4) residual equations and supporting QNO/XPNO-domain machinery for DLPNO-CCSDTQ.
* [x] Add driver procedures and method-table entries for `dlpno-ccsdt`, `dlpno-ccsdt(q0)`, `dlpno-ccsdt(q)`, and `dlpno-ccsdtq`.
* [x] Extend the CBS variable hierarchy through `CCSDT TOTAL ENERGY`, `CCSDT(Q) TOTAL ENERGY`, and `CCSDTQ TOTAL ENERGY`.
* [x] **C++ API:** Add the public PSIO file identifiers `PSIF_DLPNO_QIA_TNO` (314), `PSIF_DLPNO_QAB_TNO` (315), `PSIF_DLPNO_S_TNO` (316), `PSIF_DLPNO_QIA_QNO` (317), `PSIF_DLPNO_QAB_QNO` (318), and `PSIF_DLPNO_QUADS` (319) for disk-backed higher-order DLPNO integrals, overlaps, amplitudes, and intermediates.
* [x] Add the new DLPNO options to `read_options.cc`, including TNO, QNO, and XPNO thresholds and higher-order iterative-solver controls.
* [x] Link the DLPNO module against the Einsums tensor-algebra library used by the higher-order contractions.
* [x] Add an initial water/cc-pVDZ regression test covering the DLPNO-CCSDT and DLPNO-CCSDT(Q) energies.
* [x] Add initial Sphinx documentation and methodological bibliography entries.
* [x] Clean up option-type mismatches, variable names, comments, namespace qualification, and current-infrastructure compatibility without intentionally changing the underlying scientific algorithms.

## Questions

* [x] Should Einsums become a required/default-enabled dependency now that the higher-order DLPNO implementation uses it, or should these methods be conditionally compiled when `ENABLE_Einsums=ON` while preserving the existing lower-order DLPNO methods without Einsums?

## AI

* [x] AI-assisted source inspection, debugging, code generation, and software-maintenance work were used throughout this PR. Assistance included correcting option-type mismatches; consolidating DLPNO driver routing; diagnosing CMake/Einsums integration and namespace-lookup problems; refactoring DLPNO-CCSDT(Q) and DLPNO-CCSDTQ memory accounting and automatic disk-I/O selection; implementing disk-backed quadruples intermediates and native Einsums T4/R4 DIIS serialization; improving variable names, comments, headers, results reporting, scalar variables, and regression coverage; mapping implementation comments and terminology to the published DLPNO-CCSDT, DLPNO-CCSDT(Q), and DLPNO-CCSDTQ equations and supporting information; reviewing Sphinx documentation; performing code-quality and consistency audits; and drafting and updating this PR description. **The underlying DLPNO equations, scientific algorithms, and original implementation predate this assistance, and the AI-assisted refactoring was intended to preserve the mathematical contractions.** The author reviewed and tested the resulting changes and remains responsible for their scientific and software correctness.

## Checklist

* [x] Regression coverage added for DLPNO-CCSDT, DLPNO-CCSDT(Q), and DLPNO-CCSDTQ energies.
* [x] Initial narrative documentation, option documentation, and methodological references added.
* [x] Expand the narrative documentation with a dedicated DLPNO-CCSDTQ description and input example.
* [x] Resolve the intended default-build behavior when `ENABLE_Einsums=OFF`.
* Test running is well covered by CI. For running locally, see [[All or relevant fraction of full tests run](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests)](http://psicode.org/psi4manual/master/build_planning.html#how-to-run-a-subset-of-tests).

## Status

* [x] Ready for review
* [ ] Ready for merge


andyj10224:
Ready for a first round of reviews @loriab @jturney @TiborGY! In the meantime, I plan on running memory, runtime, and accuracy benchmarks on larger systems in comparison to previous results to validate the efficacy of the AI-assisted cleanups :)

andyj10224:
Can confirm! AI did not destroy my code, at least for DLPNO-CCSDT(Q)... will test Full Q more rigorously later. Compared results to semibullvalene/cc-pVDZ(p,p) they used in the LNO-CCSDT(Q)/Q benchmarks paper published recently: https://doi.org/10.1021/acs.jpca.6c00607

image

Input and output files with run of this branch are here... peak memory is not too terrible... 15-20 GB for 16 cores on an `AMD EPYC 9534`

[pr_benchmarks.zip](https://github.com/user-attachments/files/31432906/pr_benchmarks.zip)

susilehtola:
> image

why is the canonical ccsdt(q) value so bad? or is it just the canonical value i.e. not the error

psi4/src/psi4/dlpno/quads.cc
```diff
@@ -0,0 +1,5128 @@
...
+        for (int ij = 0; ij < n_lmo_pairs; ij++) {
+            int i, j;
+            std::tie(i, j) = ij_to_i_j_[ij];
+            if (i > j) continue;
+            for (int k : lmopair_to_lmos_[ij]) {
+                if (i > k || j > k) continue;
+                if (i == j && j == k) continue;
+                int ij_weak = i_j_to_ij_weak_[i][j], ik_weak = i_j_to_ij_weak_[i][k], kj_weak = i_j_to_ij_weak_[k][j];
+
+                int weak_pair_count = 0;
```
jturney: Quadruplet screening accumulates weak_pair_count across the l loop and silently drops valid quadruplets

quads.cc:315 declares the counter in the k scope; the l loop at quads.cc:322 adds to it at 329–331 and tests it at 333, and it is never reset per l.

```c++
int weak_pair_count = 0;                 // :315  — k scope
if (ij_weak != -1) weak_pair_count += 1;
if (ik_weak != -1) weak_pair_count += 1;
if (kj_weak != -1) weak_pair_count += 1;
if (weak_pair_count > max_weak_pairs) continue;

for (int l : lmopair_to_lmos_[ij]) {     // :322
    ...
    if (il_weak != -1) weak_pair_count += 1;   // :329  — accumulates across l
    if (jl_weak != -1) weak_pair_count += 1;
    if (kl_weak != -1) weak_pair_count += 1;
    if (weak_pair_count > max_weak_pairs) continue;   // :333
```

Concrete failure: `(i,j,k)` all-strong → count 0. First surviving `l₁` contributes 2 weak pairs → count 2, accepted. Second surviving l₂ also contributes 2 → count 4 > `QUADS_MAX_WEAK_PAIRS` (default 3) → rejected, although `(i,j,k,l₂)` itself has only 2 weak pairs. The counter never decreases, so every remaining l for that `(i,j,k)` is dropped too. The quadruplet list becomes dependent on iteration order and systematically truncated.

This is the construction path for the entire quadruplet list, so it affects CCSDT(Q), CCSDT(Q0), and CCSDTQ. It would not show up in a water/cc-pVDZ test because it has too few weak pairs to trip the threshold...which is consistent with the regression passing.

The existing triples analogue in `triples.cc:151-156` computes the count in the same scope it consumes it; the quads code copied that shape and added a nested loop without moving the accumulator. Fix is a per-l local seeded from the triple's count.

psi4/src/psi4/dlpno/dlpno.h
```diff
@@ -43,10 +43,25 @@
...
+#include "Einsums/Profile.hpp"
+#include "Einsums/TensorUtilities/RMSD.hpp"
+#endif
+
 namespace psi {
 namespace dlpno {
 
-enum class DLPNOMethod { MP2, CCSD, CCSD_T };
+#ifdef USING_Einsums
+using namespace einsums;
```
jturney: Bad practice to put `using namespace` statements inside a header file. I think it's tripped you up in places and caused you to scope `::psi` namespaces since `einsums::index` has a `psi` variable in it.

tests/dlpnocc-5/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+include(TestingMacros)
+
+add_regression_test(dlpnocc-5 "psi;dlpno;cc;quick")
```
jturney: `dlpnocc-5` is not guarded for `ENABLE_Einsums=OFF`

psi4/driver/procrouting/proc.py
```diff
@@ -4515,209 +4515,185 @@ def run_dfep2(name, **kwargs):
...
+        if "t0" in method:
+            core.set_local_option("DLPNO", "T0_APPROXIMATION", method["t0"])
+        if "q0" in method:
+            core.set_local_option("DLPNO", "Q0_APPROXIMATION", method["q0"])
+
+        if algorithm in {"CCSDT", "CCSDT(Q)", "CCSDTQ"}:
+            t_cut_tno = core.get_option("DLPNO", "T_CUT_TNO")
+            t_cut_tno_full = core.get_option("DLPNO", "T_CUT_TNO_FULL")
+            tno_scale = t_cut_tno_full / t_cut_tno
+            core.set_local_option("DLPNO", "T_CUT_TNO_STRONG_SCALE", tno_scale)
```
jturney: run_dlpno silently clobbers user-set TNO/QNO scale factors

```python
if algorithm in {"CCSDT", "CCSDT(Q)", "CCSDTQ"}:
    tno_scale = t_cut_tno_full / t_cut_tno
    core.set_local_option("DLPNO", "T_CUT_TNO_STRONG_SCALE", tno_scale)
    core.set_local_option("DLPNO", "T_CUT_TNO_WEAK_SCALE", tno_scale)
```

No `has_option_changed` guard, unlike the `T_CUT_XPNO` line a few statements later that does have one. A user who sets `T_CUT_TNO_STRONG_SCALE` or `T_CUT_QNO_STRONG_SCALE/_WEAK_SCALE` in their input has it overwritten with no warning. Same pattern in the CCSDTQ QNO block.

Two related notes on that arithmetic: `T_CUT_TNO * tno_scale` is identically `T_CUT_TNO_FULL`, so setting `T_CUT_TNO` for the CCSDT family is a no-op; worth stating in the docs, since it's a documented knob that quietly does nothing here. And `t_cut_tno` is unvalidated, so set `t_cut_tno 0` raises `ZeroDivisionError` out of the driver.

psi4/src/psi4/dlpno/dlpno.cc
```diff
@@ -184,9 +190,30 @@ void DLPNO::common_init() {
...
+        if (!T_CUT_ENERGY_MP2_changed) T_CUT_ENERGY_MP2_ = 0.999;
+        if (!T_CUT_DO_changed) T_CUT_DO_ = 5e-3;
+        if (!T_DIAG_SCALE_changed) T_CUT_PNO_DIAG_SCALE_ = 3e-2;
+        if (!T_CUT_MKN_changed) T_CUT_MKN_ = 1e-4;
+
+        // Post-CCSD(T) methods do not retain a separate weak-pair category. Honor an
+        // explicitly requested pair cutoff, but otherwise use the method default and
+        // collapse the MP2 screening boundary onto the same value.
+        if (!T_CUT_PAIRS_changed) T_CUT_PAIRS_ = 1.0e-8;
+        T_CUT_PAIRS_MP2_ = T_CUT_PAIRS_;
```
jturney, line 215: There isn't a guard on `T_CUT_PAIRS_MP2_` like the others have. Is that intended behavior?

psi4/src/psi4/dlpno/triples.cc
```diff
@@ -1506,5 +1508,2175 @@ void DLPNOCCSD_T::print_results() {
...
+        SharedMatrix psi_buffer = std::make_shared<Matrix>(n_tno_[ijk], n_tno_[ijk]);
+
+        double prefactor = 1.0;
+        if (i == j || j == k) {
+            prefactor = 0.5;
+        }
+
+        // Read integrals when disk-backed storage is enabled.
+        if (disk_ints_) {
+            q_ov_[ijk] = QIA_TNO(ijk);
```
jturney: `QIA_TNO/QAB_TNO/QIA_QNO/QAB_QNO` do the work twice

```c++
Tensor<double, 3> DLPNOCCSDT::QIA_TNO(const int ijk) {
    if (disk_ints_) {
        ...
        q_ov_[ijk] = Tensor<double, 3>(...);        // populates the member
        ::memcpy(q_ov_[ijk].data(), ...);
    }
    return q_ov_[ijk];                              // returns a full copy
}
```
called as
```c++
q_ov_[ijk] = QIA_TNO(ijk);        // triples.cc x2, quads.cc x1
q_vv_[ijk] = QAB_TNO(ijk);
```

The function already writes the member, then returns a deep copy of it, which the caller move-assigns back over the member. Net: two allocations and two full-size copies of the largest tensors in the algorithm per triplet per iteration, where one of each would do. Make them void (they already populate the member) or return const Tensor&. I checked the release side: the slots are correctly zeroed at the end of each loop body (`triples.cc` ~7349, ~8155; `quads.cc` ~4516, ~5798), so disk mode does bound its memory. It's cost, not a leak.

psi4/src/psi4/dlpno/quads.cc
```diff
@@ -0,0 +1,5128 @@
...
+        thread = omp_get_thread_num();
+#endif
+
+        std::unordered_set<int> computed_perms;
+
+        // Loop over all possible quadruplet permutations
+        einsums::for_sequence<24UL>([&](auto perm_idx) {
+            auto &[i_idx, j_idx, k_idx, l_idx] = quadruple_permutations_[perm_idx];
+            int i = ijkl_list[i_idx], j = ijkl_list[j_idx], k = ijkl_list[k_idx], l = ijkl_list[l_idx];
+            int ijkl_idx = i * std::pow(naocc, 3) + j * std::pow(naocc, 2) + k * naocc + l;
```
jturney: A number of places you do `std::pow(x, 3)` etc... floating-point `pow` in the innermost loops of the hottest contractions, narrowed back to `int`;  while other sites in the same files use plain `i * naocc * naocc + ...`.  Use the plain `i * noacc * naocc` version.

andyj10224:
> > image
> 
> why is the canonical ccsdt(q) value so bad? or is it just the canonical value i.e. not the error

It is the canonical value (not error)

psi4/src/psi4/dlpno/dlpno.cc
andyj10224: Yes! I have set the code up so that weak pairs should not exist with higher-order DLPNO codes (I should probably update that in the docs)... it does not make sense when MP2 amplitudes are factored into computations where CCSD is not even the rate-limiting step

andyj10224:
Thank you very much, @jturney, for taking the time to review this PR :)

I have worked through all of the review threads and made the following changes:

* Removed the quadruples weak-pair screening machinery, including `QUADS_MAX_WEAK_PAIRS`. For DLPNO-CCSDT and higher methods, `T_CUT_PAIRS_MP2` is intentionally collapsed onto `T_CUT_PAIRS`, so every retained pair is treated as strong. The documentation now records this behavior, and Psi4 warns if a user explicitly supplies a `T_CUT_PAIRS_MP2` value that will be overridden.
* Replaced the TNO/QNO scale-factor interface with absolute `T_CUT_TNO_STRONG`, `T_CUT_TNO_WEAK`, `T_CUT_QNO_STRONG`, and `T_CUT_QNO_WEAK` cutoffs. Higher-rank calculations explicitly use `T_CUT_TNO_FULL` or `T_CUT_QNO_FULL` for the iterative amplitudes preceding full triples or quadruples, with documentation, runtime warnings, and regression coverage for these overrides.
* Removed the `using namespace` directives from `dlpno.h`, fully qualified the Einsums tensor types in the header, and introduced only narrow declarations and namespace aliases in the implementation files. This also removes the need for the defensive `::psi` qualification caused by the `einsums::index::psi` collision.
* Guarded `dlpnocc-5` with `ENABLE_Einsums`.
* Converted the `QIA_TNO`, `QAB_TNO`, `QIA_QNO`, and `QAB_QNO` accessors into `void` disk loaders. Their callers now consume the populated member tensors directly or through const references, eliminating redundant allocations and deep copies of the largest integral tensors.
* Replaced floating-point `std::pow` calls used for flattened occupied-index arithmetic with explicit `triplet_key` and `quadruplet_key` helpers. The corresponding keys and lookup containers now use `size_t`, and the fourth-power XPNO tensor length is computed with size-safe integer multiplication.

psi4/src/psi4/dlpno/quads.cc
andyj10224: Fixed, added helper functions

psi4/src/psi4/dlpno/triples.cc
andyj10224: Fixed, changed to a const Tensor& return

psi4/driver/procrouting/proc.py
andyj10224: Changed the API to avoid this division by zero... a scaling is no longer applied, but absolute TNO/QNO cutoffs... see comment below

tests/dlpnocc-5/CMakeLists.txt
andyj10224: Fixed!

psi4/src/psi4/dlpno/dlpno.h
andyj10224: Fixed!

psi4/src/psi4/dlpno/quads.cc
andyj10224: Got rid of weak pairs for post-CCSD(T), see comment below

andyj10224:
In my latest series of pushes, I used AI to figure out how to reduce the number of evaluations in DLPNO-(Q) resulting from the P_{ijkl} operator, similar to some of the tricks I used in DLPNO-CCSDTQ to take advantage of the column symmetry of T2, T3, and T4 amplitudes to reduce the cost of certain amplitudes there! (I did not need to reinvent the wheel by manually redoing the same work I did for Full Q for (Q)). This will likely slightly reduce runtime, with the biggest impact on memory usage. Here is a technical report explaining these changes. Important note: The tensor contractions themselves are not fundamentally changed, just their organization.

[dlpno_q_permutation_symmetry_report.pdf](https://github.com/user-attachments/files/31473817/dlpno_q_permutation_symmetry_report.pdf)

andyj10224:
In this next round of changes, I've used AI to audit the tensor contraction order of all operations from DLPNO-CCSDT, DLPNO-CCSDT(Q), and DLPNO-CCSDTQ, as well as RAM, and I believe minor improvements were made runtime-wise, and major improvements made memory-wise. However, this process has also taught me that AI is not a magic bullet, as it initially messed up my `DLPNO-CCSDT` energy since the transpositions `abc -> bca` and `bca -> abc` do not represent the same mathematical operation (this is a known property for the "cyclical" permutations in P3). Nonetheless, after totally nuking my usage for two 5-hour periods in my $20/month ChatGPT subscription (and even needing to buy some extra credits), we finally figured it out! Since I completely decimated my weekly allocation now, this will likely be the last set of my major AI-based refactors before this code goes into 1.12 :)

I will now run some more benchmarks with my code, and present them as they finish to show to impact of these AI-refactors.

[dlpno_contraction_order_optimization_report.pdf](https://github.com/user-attachments/files/31493570/dlpno_contraction_order_optimization_report.pdf)

image

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -548,6 +552,12 @@ PSI Variables by Alpha
    for the requested full coupled-cluster (CCSD, CCSDT, up to CC\ *n*)
    level of theory.
 
+.. psivar:: CCSDTQ - CCSDT ENERGY
+   CCSDTQ - CCSDT(Q) ENERGY
+
+   The difference between the CCSDTQ total energy and the CCSDT or CCSDT(Q)
+   total energy, respectively [E_h].
```
TiborGY: So this psivar might contain either of two energies, what decides which one is it?

psi4/driver/procrouting/proc.py
```diff
@@ -4515,209 +4515,169 @@ def run_dfep2(name, **kwargs):
...
-
-    if name == 'scs-dlpno-mp2':
-        dlpnomp2_wfn.set_variable('CURRENT ENERGY', dlpnomp2_wfn.variable('SCS-MP2 TOTAL ENERGY'))
-        dlpnomp2_wfn.set_variable('CURRENT CORRELATION ENERGY', dlpnomp2_wfn.variable('SCS-MP2 CORRELATION ENERGY'))
-
-    elif name == 'dlpno-mp2':
-        dlpnomp2_wfn.set_variable('CURRENT ENERGY', dlpnomp2_wfn.variable('MP2 TOTAL ENERGY'))
-        dlpnomp2_wfn.set_variable('CURRENT CORRELATION ENERGY', dlpnomp2_wfn.variable('MP2 CORRELATION ENERGY'))
+def run_dlpno(name, **kwargs):
+    """Run any density-fitted local-PNO method from DLPNO-MP2 through DLPNO-CCSDTQ."""
```
TiborGY: IIRC with this PR, Psi4 will have two independent DLPNO-MP2 implementations and this function only handles the new one, correct? If yes that should be noted very briefly in this line, e.g, `the second implementation of DLPNO-MP2`

psi4/driver/procrouting/proc.py
```diff
@@ -4515,209 +4515,169 @@ def run_dfep2(name, **kwargs):
...
+                core.print_out(
+                    "\n  A requested method does not make use of molecular symmetry: "
+                    "further calculations in C1 point group.\n\n"
+                )
+                molecule.reset_point_group("c1")
+                molecule.update_geometry()
+            ref_wfn = scf_helper(method_name, use_c1=True, **kwargs)
+        elif ref_wfn.molecule().schoenflies_symbol() != "c1":
+            raise ValidationError(
+                f"{method['banner']} does not make use of molecular symmetry: "
```
TiborGY: Not a limitation of the method, just the implementation
```suggestion
                f"The {method['banner']} implementation does not make use of molecular symmetry: "
```

psi4/include/psi4/psifiles.h
```diff
@@ -284,6 +284,12 @@
 #define PSIF_DLPNO_QIA_PNO       311  /*- A PSIO file to store all (AUX | LMO PNO) integrals in a DLPNO computation -*/
 #define PSIF_DLPNO_QAB_PNO       312  /*- A PSIO file to store all (AUX | PNO PNO) integrals in a DLPNO computation -*/
 #define PSIF_DLPNO_TRIPLES       313  /*- A PSIO file to store all X_{ijk}^{abc} like terms in a DLPNO-CCSD(T) computation -*/
+#define PSIF_DLPNO_QIA_TNO       314  /*- A PSIO file to store all (AUX | LMO TNO) integrals in a DLPNO computation -*/
+#define PSIF_DLPNO_QAB_TNO       315  /*- A PSIO file to store all (AUX | TNO TNO) integrals in a DLPNO computation -*/
+#define PSIF_DLPNO_S_TNO         316  /*- A PSIO file to store all overlap integrals involving TNOs -*/
+#define PSIF_DLPNO_QIA_QNO       317  /*- A PSIO file to store all (AUX | LMO QNO) integrals in a DLPNO computation -*/
+#define PSIF_DLPNO_QAB_QNO       318  /*- A PSIO file to store all (AUX | QNO QNO) integrals in a DLPNO computation -*/
+#define PSIF_DLPNO_QUADS         319  /*- A PSIO file to store all X_{ijkl}^{abcd} like terms in DLPNO-(Q)/Full Q   -*/
```
TiborGY, line 292: This is probably a C++ API change, so therefore should be noted as such in the PR description.

psi4/src/psi4/dlpno/ccsd.cc
```diff
@@ -121,7 +121,7 @@ inline std::vector<SharedMatrix> DLPNOCCSD::QIA_PNO(const int ij) {
 
         auto q_ov = std::make_shared<Matrix>(toc_entry.str(), naux_ij, nlmo_ij * npno_ij);
 #pragma omp critical
-        q_ov->load(psio_, PSIF_DLPNO_QIA_PNO, psi::Matrix::SubBlocks);
+        q_ov->load(psio_, PSIF_DLPNO_QIA_PNO, ::psi::Matrix::SubBlocks);
```
TiborGY, line 124: Not opposed, just curious, why this change?

doc/sphinxman/source/glossary_psivariables.rst
andyj10224: These are two distinct psivars rather than one psivar whose meaning changes. A CCSDTQ calculation sets both: CCSDTQ - CCSDT ENERGY is $\(E_\mathrm{CCSDTQ}-E_\mathrm{CCSDT}\)$, while CCSDTQ - CCSDT(Q) ENERGY is $\(E_\mathrm{CCSDTQ}-E_\mathrm{CCSDT(Q)}\)$. The grouped glossary entry made that unclear, so I split it into two explicit definitions in 0e4648bd.

psi4/driver/procrouting/proc.py
andyj10224: There are indeed two MP2 code paths, but both predate this PR. energy("dlpno-mp2") constructs the independently callable `DLPNOMP2` implementation. DLPNO-CCSD and its subclasses instead perform a separate internal MP2 stage through `DLPNOCCSD::pno_lmp2_iterations()` for pair prescreening and PNO construction; that stage is not independently exposed through the driver. Since the consolidated `run_dlpno` function dispatches both the standalone DLPNO-MP2 method and the coupled-cluster hierarchy, describing it as handling only “the second implementation” would be inaccurate. I clarified the docstring accordingly.

psi4/driver/procrouting/proc.py
andyj10224: Agreed—this is a limitation of the current implementation rather than the DLPNO methods themselves. I updated the message accordingly in edf0047b.

psi4/include/psi4/psifiles.h
andyj10224: Agreed. These additions extend the public C++ API through psifiles.h. I added an explicit C++ API note to the PR description listing the six new PSIO identifiers and their purposes.

psi4/src/psi4/dlpno/ccsd.cc
andyj10224: This qualification was introduced because `einsums::index` defines an index object named `psi`. When that namespace was imported, it could shadow or collide with Psi4’s `psi` namespace, especially under unity builds. The leading `::` explicitly selects the global `psi` namespace. The broad Einsums import has since been removed, so it is now defensive rather than strictly necessary, but I retained it to keep the lookup unambiguous.

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -48,6 +48,10 @@ PSI Variables by Alpha
 
    The coupled-cluster bracket perturbative triples correction [E_h].
 
+.. psivar:: (Q) CORRECTION ENERGY
+
+   The coupled-cluster perturbative quadruples correction [E_h].
```
loriab:
```suggestion
   The coupled-cluster perturbative quadruples correction [E_h] as the
   difference between CCSDT(Q) and CCSDT.
```

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -548,6 +552,14 @@ PSI Variables by Alpha
    for the requested full coupled-cluster (CCSD, CCSDT, up to CC\ *n*)
    level of theory.
 
+.. psivar:: CCSDTQ - CCSDT ENERGY
```
loriab: `Q CORRECTION ENERGY` and `T CORRECTION ENERGY`, but optional necessary since they're not being computed on their own but by subtraction (iiuc).

doc/sphinxman/source/glossary_psivariables.rst
```diff
@@ -548,6 +552,14 @@ PSI Variables by Alpha
...
+
+   The difference between the CCSDTQ and CCSDT total energies [E_h].
+
+.. psivar:: CCSDTQ - CCSDT(Q) ENERGY
```
loriab: This is unusual to store.

psi4/driver/procrouting/proc.py
```diff
@@ -4515,209 +4515,169 @@ def run_dfep2(name, **kwargs):
...
-    dlpnomp2_wfn.compute_energy()
-
-    if name == 'scs-dlpno-mp2':
-        dlpnomp2_wfn.set_variable('CURRENT ENERGY', dlpnomp2_wfn.variable('SCS-MP2 TOTAL ENERGY'))
-        dlpnomp2_wfn.set_variable('CURRENT CORRELATION ENERGY', dlpnomp2_wfn.variable('SCS-MP2 CORRELATION ENERGY'))
-
-    elif name == 'dlpno-mp2':
-        dlpnomp2_wfn.set_variable('CURRENT ENERGY', dlpnomp2_wfn.variable('MP2 TOTAL ENERGY'))
-        dlpnomp2_wfn.set_variable('CURRENT CORRELATION ENERGY', dlpnomp2_wfn.variable('MP2 CORRELATION ENERGY'))
+def run_dlpno(name, **kwargs):
```
loriab, line 4508: Is this making the original dlpno-mp2 unreachable? Can do a select_dlpnomp2 function to make both accessible.

psi4/driver/procrouting/proc.py
```diff
@@ -4515,209 +4515,169 @@ def run_dfep2(name, **kwargs):
...
+            "algorithm": "CCSDT",
+            "banner": "DLPNO-CCSDT",
+            "aux_basis": "DF_BASIS_CC",
+            "energy_label": "CCSDT",
+        },
+        "dlpno-ccsdt(q0)": {
+            "algorithm": "CCSDT(Q)",
+            "banner": "DLPNO-CCSDT(Q0)",
+            "aux_basis": "DF_BASIS_CC",
+            "energy_label": "CCSDT(Q)",
```
loriab: It's customary to have the energy(mtd) match the CURRENT ENERGY = mtd TOTAL ENERGY. If t0/q0 are separate calls, need at least TOTAL and CORRELATION specially defined.

psi4/src/psi4/dlpno/CMakeLists.txt
```diff
@@ -6,9 +6,18 @@ list(APPEND sources
...
+endif()
+
 psi4_add_module(bin dlpno sources)
 if (psi4_ENABLE_PRECOMPILE_HEADERS)
     target_precompile_headers(dlpno REUSE_FROM psio)
 endif()
 
 target_link_libraries(dlpno PUBLIC diis)
+
+if(TARGET Einsums::Einsums)
```
loriab: this block shouldn't be needed -- psi4/src/psi4/CM does the linking.

psi4/driver/procrouting/proc.py
andyj10224: The original `dlpno-mp2` is reachable, I mean that the "DLPNO-MP2" called in the prescreening for DLPNO-CCSD+ cannot be called as a standalone algorithm

doc/sphinxman/source/glossary_psivariables.rst
andyj10224: Done!

doc/sphinxman/source/glossary_psivariables.rst
andyj10224: Made the changes!

doc/sphinxman/source/glossary_psivariables.rst
andyj10224: Agreed, got rid of this variable, but we are still letting it be print out in case it is needed for focal point.

psi4/driver/procrouting/proc.py
andyj10224: Total and Correlation are specially defined now, with `CCSD(T0) CORRELATION ENERGY` set when `dlpno-ccsd(t0)` is called and `CCSDT(Q0) CORRELATION ENERGY` called when `dlpno-ccsdt(q0)` is called. I hardly imagine anyone ever needing to call `dlpno-ccsd(t0)` or `dlpno-ccsdt(q0)`, but these options are available to keep continuity with ORCA.

psi4/src/psi4/dlpno/CMakeLists.txt
andyj10224: Fixed!

andyj10224:
@loriab @TiborGY @jturney Thank you all for your reviews! Once I finish my Full Q jobs on benzene dimer and confirm continuity with old results, as well as audit memory and runtime requirements, this code will be, in my opinion, ready for a second round of reviews and merging :)

andyj10224:
image
I verified benzene dimer computation from `doi.org/10.1063/1.3659142` (most recent computation) matches most recent results within reason (the scratch for the old computation has been deleted, so I'm not sure what exact cutoffs were used), but I am confident enough that the code is providing the right answer. What is even more encouraging is that CP and noCP results are generally in agreement post-CCSD(T). The zip files from the computation are also provided

[fullq_pr.zip](https://github.com/user-attachments/files/31702579/fullq_pr.zip)

andyj10224:
I changed `T_CUT_QNO_STRONG` back to `3.33e-6` since there was no reason to make it as tight at `1.0e-6` after doing some numerical tests

psi4/src/psi4/dlpno/dlpno.cc
```diff
@@ -231,6 +267,55 @@ void DLPNO::C_DGESV_wrapper(SharedMatrix A, SharedMatrix B) {
     }
 }
 
+void DLPNO::C_DGESV_shared_factorization(
```
TiborGY: I dont think I like this, will write more about it later

andyj10224:
WE DID IT!!! RUNNING DLPNO IN C1 AFTER RUNNING SCF IN SYMMETRY WAS AN ISSUE I COULD NEVER FIGURE OUT IN THE ORIGINAL DLPNO-CCSD(T) PR, BUT IT TURNS OUT THAT WITH SYMMETRIC ORBITALS, YOU CAN GET A DIVISION-BY-ZERO ERROR WHEN COMPUTING MULLIKEN POPULATIONS! 

TL;DR: We can now run DLPNO in c1 even if SCF is run with symmetry

psi4/src/psi4/dlpno/quads.cc
```diff
@@ -0,0 +1,6034 @@
...
+        auto A_solve = submatrix_rows_and_cols(*full_metric_, lmoquadruplet_to_ribfs_[ijkl], lmoquadruplet_to_ribfs_[ijkl]);
+        A_solve->power(0.5, 1.0e-14);
+
+        q_iv = linalg::doublet(q_iv, X_qno_[ijkl]);
+        q_jv = linalg::doublet(q_jv, X_qno_[ijkl]);
+        q_kv = linalg::doublet(q_kv, X_qno_[ijkl]);
+        q_lv = linalg::doublet(q_lv, X_qno_[ijkl]);
+
+        C_DGESV_shared_factorization(
+            A_solve->clone(), {q_io, q_jo, q_ko, q_lo, q_iv, q_jv, q_kv, q_lv, q_ov, q_vv});
```
TiborGY: What if you computed `A_solve->power(-0.5, 1.0e-14);` instead, and then applied A_solve with DGEMM or linalg doublet? That would avoid having to call  C_DGESV_shared_factorization.

psi4/src/psi4/dlpno/quads.cc
```diff
@@ -0,0 +1,6034 @@
...
+        q_jv = linalg::doublet(q_jv, X_qno_[ijkl]);
+        q_kv = linalg::doublet(q_kv, X_qno_[ijkl]);
+        q_lv = linalg::doublet(q_lv, X_qno_[ijkl]);
+        
+        // Multiply by (P|Q)^{-1/2}
+        auto A_solve = submatrix_rows_and_cols(*full_metric_, lmoquadruplet_to_ribfs_[ijkl], lmoquadruplet_to_ribfs_[ijkl]);
+        A_solve->power(0.5, 1.0e-14);
+
+        C_DGESV_shared_factorization(
+            A_solve->clone(), {q_io, q_jo, q_ko, q_lo, q_iv, q_jv, q_kv, q_lv, q_ov, q_vv});
```
TiborGY: Same here. The comment even says `(P|Q)^{-1/2}` but then `power(0.5,...` is used.

TiborGY (changes requested):

psi4/src/psi4/dlpno/quads.cc
andyj10224: Great idea! This has just been implemented for all `C_DGESV` calls, other than the ones where it is only called once

psi4/src/psi4/dlpno/quads.cc
andyj10224: Resolved :)

psi4/src/psi4/dlpno/wrapper.cc
```diff
@@ -44,6 +44,19 @@ SharedWavefunction dlpno(SharedWavefunction ref_wfn, Options& options) {
...
+        } else if (options.get_str("DLPNO_ALGORITHM") == "CCSDT(Q)") {
+            dlpno = std::make_shared<DLPNOCCSDT_Q>(ref_wfn, options);
+        } else if (options.get_str("DLPNO_ALGORITHM") == "CCSDTQ") {
+            dlpno = std::make_shared<DLPNOCCSDTQ>(ref_wfn, options);
+#else
+        } else if (options.get_str("DLPNO_ALGORITHM") == "CCSDT" ||
+                   options.get_str("DLPNO_ALGORITHM") == "CCSDT(Q)" ||
+                   options.get_str("DLPNO_ALGORITHM") == "CCSDTQ") {
+            throw PSIEXCEPTION("DLPNO-CCSDT, DLPNO-CCSDT(Q), and DLPNO-CCSDTQ require Psi4 built with Einsums support (ENABLE_Einsums=ON).");
+#endif
```
TiborGY: This is _fine_ but I feel that the modern solution would be to establish a `constexpr bool` set from the preprocessor symbol, maybe in pragma.h, and then use that in an if statement.

psi4/src/psi4/dlpno/dlpno.h
TiborGY: Why does the DLPNO class redeclare `std::shared_ptr<PSIO> psio_` when it is also inheriting the same symbol from its parent class, Wavefunction? There might be good reason for doing this that I am not seeing, but it could cause some surprising shadowing behavior. For example, since DLPNO does not define its own `psio()` getter, calling it on a DLPNO object returns the base class'  `std::shared_ptr<PSIO> psio_`, which might remain null even when the shadowing psio_ isn't.
I think the entire issue might be solvable by deleting this line. Then, DLPNO objects would only hold the psio_ they inherited, and `DLPNO::common_init()` would always init that.

psi4/src/psi4/dlpno/dlpno.cc
TiborGY, line 234: For example here, this is where DLPNO::psio_ gets set. This does not touch the shadowed psio_ from the parent class, which therefore might remain nullptr.

TiborGY (changes requested):
Some CI tests are failing. Might just be GH flakiness though.

psi4/src/psi4/dlpno/dlpno.h
andyj10224: Agreed! I did not realize `Wavefunction` had a `psio_` member variable as well! I just got rid of it

psi4/src/psi4/dlpno/dlpno.cc
andyj10224: Yeah, this is not good! I removed it

andyj10224:
DLPNO-CCSDTQ is too powerful 😂, it can run computations on molecules where the diis vector size exceeds `INT_MAX` doubles (over 2 billion amplitudes 😅). We implemented chunking for DIIS for T3 and T4 amplitudes! (Some DIIS changes are also made to the API)

TiborGY:
> DLPNO-CCSDTQ is too powerful 😂, it can run computations on molecules where the diis vector size exceeds `INT_MAX` doubles (over 2 billion amplitudes 😅). We implemented chunking for DIIS for T3 and T4 amplitudes! (Some DIIS changes are also made to the API)

We have had 64-bit CPUs since 2003 at the latest, so I think the proper fix would be to go through the entire call chain and replace whatever 32-bit int is overflowing with 64-bit ints. Might need build/conda adjustments to mandate ILP64 LAPACK/BLAS though, so that's out of scope for the PR given there is a workaround in place now.

psi4/src/psi4/dlpno/dlpno.h
TiborGY, line 111: The code does not seem to be actually using this, and instead it keeps calling options_.get_double("T_CUT_TNO") everywhere it needs this threshold. If that's intentional this member of the DLPNO class should be deleted.

TiborGY, line 174: Are these used anywhere in the codebase?
```suggestion
```

TiborGY, line 236: Are these two still in use?
```suggestion

```

TiborGY, line 163: These seem to be only used in `DLPNOCCSD` and derivatives, not in `DLPNOMP2`, so maybe they could be moved into `DLPNOCCSD`?

TiborGY, line 121: Could this be a private member of DLPNO?

TiborGY, line 204: lmopair_to_riatoms_ could probably go private

TiborGY, line 206: And also lmopair_to_paoatoms_

TiborGY (changes requested):

psi4/src/psi4/dlpno/dlpno.h
andyj10224: Agreed! Remove this variable (especially since different cutoffs are used in (T), and Full T now anyways)

andyj10224: Nem!

andyj10224: Nem!

andyj10224: Igen, Done!

andyj10224: Igen!

andyj10224: Igen!

andyj10224: Igen!

susilehtola:
@andyj10224 what's nem and igen?

andyj10224:
> @andyj10224 what's nem and igen?

Igen/Nem = "Yes/No" in Hungarian 😊 I can also say kyllä/ei next time 😉 (with added English subtitles)

psi4/src/psi4/dlpno/dlpno.h
andyj10224: These are not used, we got rid of them

andyj10224: Not used, same as above

andyj10224: Yes, we moved these to `DLPNOCCSD`

andyj10224: Yes, these can be private, we changed it

andyj10224: Yep, we made them private

andyj10224: Made private!

TiborGY, line 207: This only seems to be used once in `prep_sparsity`. Could this become a local in that function, instead of a class member?

TiborGY, line 207: This appears to be used only in `prep_sparsity` as a temporary, so perhaps it would be better to make it a local there instead of a member variable?

TiborGY, line 210: Same here

TiborGY, line 213: And here

TiborGY, line 187: like lmo_to_riatoms_ext_ this might be better to convert to a local

TiborGY, line 189: probably also local to prep_sparsity

psi4/src/psi4/dlpno/dlpno.h
```diff
@@ -205,15 +228,9 @@ class DLPNO : public Wavefunction {
     std::vector<std::vector<bool>> riatom_to_atoms2_dense_;
     std::vector<std::vector<int>> lmopair_to_lmos_dense_;
 
-    /// Useful for generating DF integrals (TODO: Replace this with "index_list" function)
-    std::vector<std::vector<std::vector<int>>> lmopair_lmo_to_riatom_lmo_;
-    std::vector<std::vector<std::vector<int>>> lmopair_pao_to_riatom_pao_;
     std::vector<std::vector<std::pair<int,int>>> riatom_to_pao_pairs_; ///< Which (u,v) pao pairs belong to an riatom
```
TiborGY, line 231: riatom_to_pao_pairs_ seems to be only used in compute_qab

psi4/src/psi4/dlpno/dlpno.h
```diff
@@ -132,9 +162,6 @@ class DLPNO : public Wavefunction {
     std::vector<SharedMatrix> X_pno_;   ///< global PAO -> canonical PNO transforms
     std::vector<SharedVector> e_pno_;   ///< PNO orbital energies
     std::vector<int> n_pno_;       ///< number of pnos
-    std::vector<double> occ_pno_;       ///< lowest PNO occupation number per PNO
-    std::vector<double> trace_pno_;     ///< total trace(Dij) recovered per PNO
-    std::vector<double> e_ratio_pno_;   ///< percentage of correlation energy recovered by PNOs
     std::vector<double> de_pno_;   ///< PNO truncation energy error
     std::vector<double> de_pno_os_;   ///< opposite-spin contributions to de_pno_
     std::vector<double> de_pno_ss_;   ///< same-spin contributions to de_pno_
```
TiborGY, line 167: these two could probably be locals in pno_transform

TiborGY (changes requested):

PR #3512, opened 2026-08-27
DF-MP2 memory checks

loriab:

- [x] dfocc stops and prints memory usage rather than proceeding when it knows it's going to require terabytes of memory for a df-mp2 energy or gradient

## Dev notes & details

  Two related pieces of memory work on the DF-MP2 paths, plus one driver fix.

  `dfocc`'s MP2 code is entirely in-core, but its only memory reporting ran *after*
  the integral transformation had already allocated. A calculation too large for the
  requested memory would print its costs and then die in `std::bad_alloc` — or, worse,
  take the machine down with it. This PR estimates each sequential stage's peak up
  front and stops with a diagnosable error before the first large allocation.

  `dfmp2` gets the integer-overflow and block-sizing bugs found while auditing the
  same class of code in `libdfmp2`.

  4 files, +207 / −151.

  ## dfocc: up-front memory estimate for DF-MP2

  `DFOCC::mp2_manager()` now estimates, before `df_corr()`, the peak of every stage it
  will run, and throws with the shortfall if the request can't cover the largest one.
  The stages are sequential and each frees before the next, so the requirement is the
  max over stages, not the sum:

  | stage | resident at peak |
  |---|---|
  | DF-CC AO build (`b_so`) | `Ap` + `Bp` + `J^-1/2` |
  | DF-CC transform, direct MP2 (`b_so`) | `B(Q\|mn)` + `J^-1/2` + `(Q\|mA)` + `(Q\|IA)` |
  | DF-CC transform, full (`b_oo`/`b_ov`/`b_vv`) | `B(Q\|mn)` + `(Q\|m x)` + `(Q\|x y)` |
  | DF-SCF AO build (`b_so_ref`) | `bQso` + `Ap` + `Bp` + `J^-1/2` |
  | DF-SCF transform (`trans_ref`) | as above, `nQ_ref` |
  | amplitudes | 3 restricted / 2 unrestricted energy / 4 density path |

  Output, printed before anything large is allocated:

  ```
        DFOCC MP2 in-core memory requirements:
        Available memory                     :        0.466 [GiB]
        B-CC (Q|mu nu)                       :        0.004 [GiB]
        DF-CC integral build/transform       :        0.007 [GiB]
        Amplitude/density (3 tensors)        :        0.001 [GiB]
        Estimated minimum peak (lower bound) :        0.007 [GiB]
  ```

  and on failure:

  ```
  The current restricted DFOCC MP2 energy algorithm is in-core and requires at
  least 4.2 GiB, but only 0.5 GiB is available. Use a memory-blocked MP2
  implementation or increase available memory.
  ```

  Details worth a reviewer's eye:

  - The AO build is 2 AO-sized blocks for DF-CC (`b_so` frees `Ap` before making
    `bQso`) but 3 for DF-SCF (`b_so_ref` allocates `bQso` up front). The estimate
    distinguishes them.
  - The full transform compares the `oo` and `vv` stages. `vv` alone under-sizes any
    case with more occupied than virtual orbitals. The `ov` stage needs no separate
    term: `nocc > nvir` makes both its factors smaller than `oo`'s, `nocc <= nvir`
    smaller than `vv`'s.
  - `trans_corr()` transforms all orbitals only on the density path and actives
    otherwise. Since `mp2_manager()` only runs with `orb_opt_ == FALSE`, one flag is
    both that branch and `omp2_tpdm()`'s guard, so it drives the dimension choice and
    the amplitude count together.
  - Amplitude counts are per-path rather than a single constant: restricted holds `T`,
    `K` and `U` at once; unrestricted energies work a spin block at a time and never
    hold more than two; the density path adds `omp2_tpdm`'s live `U`. Overestimating
    here refuses calculations that would have run.

  ## dfmp2: overflow and block-sizing fixes

  Genuine bugs:

  - **`apply_B_transpose` could hang.** `max_A = (rows <= 0 ? 1 : rows)` was
    immediately overwritten by `max_A = (rows > navir ? navir : rows)`, so `max_A`
    could reach 0 and the block-sizing loop `for (a = 0; a < navir; a += max_A)` would
    never terminate.
  - **`UV_helper` sized its blocks from the wrong variable** — the member `memory_`
    rather than the `memory` parameter it was passed.
  - **Unsigned underflow** in `form_Pab`, `form_Pij`, `form_L`, `fitting_helper` and
    `UV_helper`: subtracting fixed storage from `size_t` memory could wrap to an
    enormous value and sail past the function's own "Not enough memory" throw with a
    garbage block size.
  - **`int` overflow** in size computations that exceed 2^31 for large systems:
    `naux * naux` in the `Jm12` and `V` PSIO entries, `nrows * nso * nso`,
    `np * nso * nso` and `nrows * naocc * navir`, and `naocc * navir` for `nia` and
    the `Gia`/`Gmi`/`Gam` allocations. `apply_B_transpose`'s block loops moved to
    `size_t` throughout.
  - **`apply_G_transpose` under-costed its rows** at `naux` when it allocates two
    buffers per row; now `2 * naux`.

  #### driver: unmasked error message (this is trivial)

  `highest_analytic_properties_available()` passed `e.message` (a string) to
  `_alternative_methods_message()`, which indexes its argument as the stats dict —
  what every call site in `highest_analytic_derivative_available()` passes. A real
  `MissingMethodError` surfaced as `TypeError: string indices must be integers, not
  'str'`, hiding the cause.

  ```
  molecule { 0 2
  O
  H 1 0.97
  }
  set { basis cc-pvdz
        reference uhf
        mp2_type df }
  properties('mp2', properties=['dipole'])
  ```

  now reports that MP2 properties are unavailable for UHF/DF, with the conditions and
  the capabilities-table link.

  ## Two design decisions

  **1. Where to error, and where to keep going.** DFMP2's block-sizing clamps
  (`rows = (rows < maxP ? maxP : rows)`, `max_nia = (max_nia < 1 ? 1 : max_nia)`) look
  like clamps in the wrong direction — they let the code exceed the user's memory
  request. They're deliberate: proceed at the minimum useful block size rather than
  refuse to run. Per discussion in this PR, `mp2.cc` and `corr_grad.cc` keep that
  behavior everywhere, and `require_memory` throws only where the original threw, at
  the original's threshold. The underflow guards above are the exception, and only
  because the original *intended* to throw there and the wrap prevented it.

  **2. The dfocc estimate is a lower bound, deliberately.** It omits contributions it
  can't cheaply bound, and it compares against total requested memory rather than what
  is free — the SCF wavefunction is already resident. So it catches the clearly
  impossible and does not promise the calculation will fit. The label says so.

  ## Verification

  `dfmp2-grad1` through `-grad5`, `dfomp2-1/2`, `dfomp2-grad1/2`, `dfmp2-1/2` and
  `scf5` pass. Additionally checked by hand:

  - DF-MP2 gradient at 12 MB still runs (no regression from the clamp behavior).
  - Water/cc-pVQZ at 30 MB throws the new dfocc error before allocating.
  - Printed values hand-verified against the formulas for water/cc-pVTZ (frozen core)
    and benzene/STO-3G, the latter chosen for `nocc` (21) > `nvir` (15) so the `oo`
    stage is the one that binds.
  - Amplitude counts confirmed per path: RHF energy 3, UHF energy 2, UHF EKT-IP 4,
    UHF gradient 4.

## AI

- [x] Diagnosed by AI and rechecked by another AI

## Checklist
- [ ] Docstring or narrative docs/sphinxman/source updated if needed

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/dfmp2/mp2.cc
+++ b/psi4/src/psi4/dfmp2/mp2.cc
@@ -28,6 +28,9 @@
 
 #include "mp2.h"
 
+#include <algorithm>
+#include <sstream>
+
 #ifdef _OPENMP
 #include <omp.h>
 #endif
@@ -58,6 +61,26 @@
 
 namespace psi {
 namespace dfmp2 {
+
+namespace {
+
+void require_memory(size_t available, size_t required, const std::string& stage) {
+    if (available < required) {
+        constexpr double doubles_per_gib = 1024.0 * 1024.0 * 1024.0 / sizeof(double);
+        std::ostringstream message;
+        message << "DFMP2: " << stage << " requires at least " << required / doubles_per_gib
+                << " GiB, but only " << available / doubles_per_gib << " GiB is available.";
+        throw PSIEXCEPTION(message.str());
+    }
+}
+
+size_t auxiliary_block_rows(size_t available, size_t row_cost, size_t max_shell, size_t naux,
+                            const std::string& stage) {
+    require_memory(available, max_shell * row_cost, stage + " (one auxiliary shell)");
+    return std::min(naux, available / row_cost);
+}
+
+}  // namespace
 
 void DFMP2::compute_opdm_and_nos(const SharedMatrix Dnosym, SharedMatrix Dso, SharedMatrix Cno, SharedVector occ) {
     // The density matrix
@@ -400,7 +423,8 @@ SharedMatrix DFMP2::form_inverse_metric() {
         auto Jm12 = std::make_shared<Matrix>("SO Basis Fitting Inverse (Eig)", naux, naux);
         outfile->Printf("\t Will attempt to load fitting metric from file %d.\n\n", PSIF_DFSCF_BJ);
         psio_->open(PSIF_DFSCF_BJ, PSIO_OPEN_OLD);
-        psio_->read_entry(PSIF_DFSCF_BJ, "DFMP2 Jm12", (char*)Jm12->pointer()[0], sizeof(double) * naux * naux);
+        psio_->read_entry(PSIF_DFSCF_BJ, "DFMP2 Jm12", (char*)Jm12->pointer()[0],
+                          sizeof(double) * static_cast<size_t>(naux) * naux);
         psio_->close(PSIF_DFSCF_BJ, 1);
 
         timer_off("DFMP2 Metric");
@@ -417,7 +441,8 @@ SharedMatrix DFMP2::form_inverse_metric() {
         if (options_.get_str("DF_INTS_IO") == "SAVE") {
             outfile->Printf("\t Will save fitting metric to file %d.\n\n", PSIF_DFSCF_BJ);
             psio_->open(PSIF_DFSCF_BJ, PSIO_OPEN_OLD);
-            psio_->write_entry(PSIF_DFSCF_BJ, "DFMP2 Jm12", (char*)Jm12->pointer()[0], sizeof(double) * naux * naux);
+            psio_->write_entry(PSIF_DFSCF_BJ, "DFMP2 Jm12", (char*)Jm12->pointer()[0],
+                               sizeof(double) * static_cast<size_t>(naux) * naux);
             psio_->close(PSIF_DFSCF_BJ, 1);
         }
 
@@ -428,7 +453,7 @@ SharedMatrix DFMP2::form_inverse_metric() {
 }
 void DFMP2::apply_fitting(SharedMatrix Jm12, size_t file, size_t naux, size_t nia) {
     // Memory constraints
-    size_t Jmem = naux * naux;
+    size_t Jmem = naux * static_cast<size_t>(naux);
     size_t doubles = (size_t)(options_.get_double("DFMP2_MEM_FACTOR") * (memory_ / 8L));
     if (doubles < 2L * Jmem) {
         throw PSIEXCEPTION("DFMP2: More memory required for tractable disk transpose");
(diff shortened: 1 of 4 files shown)


psi4/src/psi4/dfmp2/mp2.cc
```diff
@@ -1531,10 +1558,7 @@ void RDFMP2::form_Amn_x_terms() {
     row_cost += nso * (size_t)nso;
     row_cost += nso * (size_t)naocc;
     row_cost += naocc * (size_t)navir;
-    size_t rows = memory / row_cost;
-    rows = (rows > naux ? naux : rows);
-    rows = (rows < maxP ? maxP : rows);
-    max_rows = (int)rows;
```
loriab, line 1537: So AI seizes upon `rows = (rows < maxP ? maxP : rows);` (for an array you want to block, if the number of rows that fit in memory is less than the max shell size, block over the max shell size) as a clamp-in-the-wrong-direction error that violates user-given-memory limitation. I agree it's violating the user-given-memory, but this occurs in half-a-dozen places, so I wonder if Rob was rather making a choice to keep the calculation going at minimum blocking size rather than believing that the user set memory too deliberately. 

Thoughts on whether to handle this as a bug?

jturney: Yeah, that was Rob's idea ... keep the calculation going with a minimum expected number of rows (maxP) even if that is larger than what the user requested. If you want to honor the user-given-memory then it should error out.

Looks like row_cost is roughly 4 * nso * nso or 2*naux in places. So I think the user would have to specify a really small amount of memory to trigger an error.

loriab: Ok, thanks. I think I agree with keep-it-going. Thanks for the clarification.

Issue #3514, opened 2026-08-28
Investigate further cleanups in libpsio


TiborGY:
After #3452 is in, these ideas could be investigated and be follow-up PRs:

- [ ] Remove multivolume vestiges (eg, merge the members of psio_vol into psio_ud )
- [ ] use std::string more instead of eg. char *name
- [ ] factor out duplicated filename generation code from open() and exists()

susilehtola:
Started on two of these (stacked on #3452):

- [x] **Item 3** — factor out duplicated filename generation from `open()`/`exists()`: #3515
- [x] **Item 1** — merge `psio_vol` members into `psio_ud`: #3516

Both are mechanical/no-behaviour-change and pass the round-trip test + a normal SCF/MP2 that round-trips PSIO to disk.

One note on **Item 2** (`std::string` for the paths): it has a prerequisite not on the list — `psio_unit` (and `psio_readlen`/`psio_writlen`) are `malloc`'d in `init.cc`, so no constructors run; a `std::string` member of `psio_ud`/`psio_vol` would be UB as-is. Item 2 should therefore be bundled with converting those arrays to a C++ container (`std::vector`/`new[]`) first, then the `char *path` → `std::string` conversion is safe. Flagging so whoever picks it up sequences it that way.

susilehtola:
- [x] **Item 2** — `std::string` for the paths / `char *name`: #3517

Done as flagged above — it bundles the allocation modernization (`psio_unit`/`psio_readlen`/`psio_writlen` `malloc` → `std::vector`) first, since that's the prerequisite for a `std::string` member, then converts `psio_ud::path` to `std::string` and `get_filename()` to return one. Net −69 lines; also fixes a latent path leak for units still open at teardown. Verified through SCF/DF-MP2/CCSD (heavy multi-unit PSIO).

That covers all three boxes (#3515, #3516, #3517), each stacked on #3452.

PR #3515, opened 2026-08-28
libpsio: factor out unit-filename generation (on #3452)

susilehtola:
## Description

Follow-up cleanup from #3514 (item 3), now that #3452 is merged.

`PSIO::open()` and `PSIO::exists()` each rebuilt the on-disk filename with the same `get_filename` + `PSIOManager::get_file_path` + `malloc`/`sprintf`/`strdup`/`free` sequence. This factors it into a single private helper `PSIO::get_unit_filename(unit)` that composes the path with `std::string` (`<path><name>.<unit>`), removing the duplication and the manual C-string juggling on the file-open path. Behaviour is unchanged.

- No user-facing change (internal refactor).

## Dev notes & details
- New private member `std::string PSIO::get_unit_filename(size_t unit)` composes `<path><name>.<unit>` via `std::string`.
- `PSIO::open()` and `PSIO::exists()` call it instead of each re-implementing the same path construction with `malloc`/`sprintf`/`strdup`/`free`.
- Pure refactor: no behavioural change on the file-open/exists paths.

## Questions
- None.

## AI
- Refactor drafted with Claude Code (identifying the duplicated path-construction, extracting the helper); reviewed and verified by me.

## Checklist
- [x] Tests added for any new features — n/a (no new feature; covered by existing `psio_roundtrip` test)
- [x] Docstring or narrative docs/sphinxman/source updated if needed — n/a

## Status
- [x] Ready for review

---

Verified: the `psio_roundtrip` standalone test passes (4/4), and a normal SCF (which round-trips PSIO to disk) gives the expected energy.

### DIFF
--- a/psi4/src/psi4/libpsio/open.cc
+++ b/psi4/src/psi4/libpsio/open.cc
@@ -41,8 +41,15 @@
 #include "psi4/psi4-dec.h"
 namespace psi {
 
-void PSIO::open(size_t unit, int status) {
+std::string PSIO::get_unit_filename(size_t unit) const {
     char *name;
+    get_filename(unit, &name);
+    std::string full_path = PSIOManager::shared_object()->get_file_path(unit) + name + "." + std::to_string(unit);
+    free(name);
+    return full_path;
+}
+
+void PSIO::open(size_t unit, int status) {
     psio_ud* this_unit;
 
     /* check for too large unit */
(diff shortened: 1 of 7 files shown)


loriab (approved):
Reviewed 2nd commit and lgtm

psi4/src/psi4/libpsio/psio.hpp
```diff
@@ -313,6 +313,9 @@ class PSI_API PSIO {
     /// grab the filename of unit and strdup into name.
     void get_filename(size_t unit, char **name, bool remove_namespace = false);
 
+    /// Compose the full on-disk path for a unit: <path><name>.<unit>.
+    std::string get_unit_filename(size_t unit);
+
```
TiborGY: Could this be a private member function instead of public?

TiborGY (changes requested):

TiborGY:
Please rebase and edit the PR description to match the template.

psi4/src/psi4/libpsio/psio.hpp
susilehtola: Done — moved `get_unit_filename` into the `private:` section (it is only called from `open()`/`exists()` within libpsio). Also rebased onto merged #3452 (the striping-removal commit drops out) and rewrote the PR description to match the template.

susilehtola:
Rebased onto master (now that #3452 is merged, its commit drops out) and rewrote the description to follow the PR template. Also addressed the inline comment by making `get_unit_filename` private.

psi4/src/psi4/libpsio/psio.hpp
susilehtola: Force-pushed a small follow-up on top of the private-move: `get_unit_filename` and its read-only call chain (`get_filename`, the `filecfg_kwd` getter) are now `const`, and I dropped an unused `<queue>` include from `psio.hpp`. Still a pure refactor.

psi4/src/psi4/libpsio/open.cc
```diff
@@ -41,8 +41,15 @@
 #include "psi4/psi4-dec.h"
 namespace psi {
 
-void PSIO::open(size_t unit, int status) {
+std::string PSIO::get_unit_filename(size_t unit) const {
     char *name;
+    get_filename(unit, &name);
```
TiborGY, line 46: "return result by char**, allocated by the callee" is such a grotesque interface, but let's not scope creep this PR

TiborGY (approved):

psi4/src/psi4/libpsio/open.cc
susilehtola: Agreed...

PR #3516, opened 2026-08-28
libpsio: merge psio_vol into psio_ud (on #3452, #3515)

susilehtola:
Follow-up cleanup from #3514 (item 1), stacked on #3452 and #3515 — **merge those first**; until then this PR's diff also shows their commits.

With multi-volume striping removed (#3452), a `psio_ud` holds a single `psio_vol`, so the separate struct is a vestige that only adds a `.vol.` indirection at every access. This:

- inlines `psio_vol`'s members (`char *path`, `int stream`) directly into `psio_ud` and drops the `psio_vol` struct;
- rewrites every `unit->vol.path` / `unit->vol.stream` to `unit->path` / `unit->stream`;
- changes `psio_volseek()` to take the stream `int` (all it ever used) instead of a `psio_vol*`.

Purely mechanical — no behaviour change. `psio_vol` is used nowhere outside libpsio. Verified: the `psio_roundtrip` test (4/4), an SCF, and a DF-MP2 (both round-trip PSIO to disk) all give the expected results.

Part of #3514.


TiborGY (changes requested):
this needs a rebase

susilehtola:
Superseded by #3540, which deletes the structure this PR reshapes.

#3540 reimplements libpsio's I/O core as a shim over [libspill](https://github.com/ReLibQC/libspill). Once libspill owns the file handle and the on-disk layout, `psio_vol` and `psio_ud` have nothing left to describe — the path, the descriptor and the table-of-contents head are all libspill's. Neither type is named anywhere outside `libpsio` (checked), and after the port the only code touching them was `PSIO`'s constructor, which `malloc`'d a `PSIO_MAXUNIT`-long array per instance, initialised it, and never read it again.

So they are removed outright in psi4/psi4@f0486cad2, along with `PSIO::psio_unit` and the dangling `psio_volseek` declaration. Deleting beats improving.

Nothing here is wrong — merging `psio_vol` into `psio_ud` is the right call for the code as it stands, and the reasoning in this PR is what made it obvious that the struct was vestigial after the port. It is just that the port removes the subject. Closing rather than merging so the two do not conflict over the same lines; happy to reopen if #3540 does not land.

TiborGY:
#3540 will probably not land anytime soon, Claude...
So I am reopening this since it is a good improvement in the interim

susilehtola:
> #3540 will probably not land anytime soon, Claude... So I am reopening this since it is a good improvement in the interim

why not? I am not sure it is smart to waste developer and reviewer resources on code that will likely be removed in the near future

TiborGY:
> > #3540 will probably not land anytime soon, Claude... So I am reopening this since it is a good improvement in the interim
> 
> why not? I am not sure it is smart to waste developer and reviewer resources on code that will likely be removed in the near future

I dont see it landing in 1.12, but fair enough, if it lands in 1.13 then this patch does indeed get superseded quickly.

PR #3517, opened 2026-08-28
libpsio: return the unit filename as a std::string

susilehtola:
## Description

`PSIO::get_filename()` returned its result the old way: a `char **` out-param, `strdup`'d by the callee, `free`'d by every caller. A reviewer on #3515 called the convention horrible, which it is — three of the four call sites existed only to allocate a buffer, copy into it and release it again. It now returns a `std::string`.

```c++
-    void get_filename(size_t unit, char **name, bool remove_namespace = false) const;
+    std::string get_filename(size_t unit, bool remove_namespace = false) const;
```

The three callers compose paths with `std::string` instead of `malloc`/`sprintf`/`free`, and `rename_file()` drops its own path assembly in favour of the `get_unit_filename()` helper #3515 factored out — the two were building the same string by different means.

**Net −70 lines**, no behaviour change: the composed paths are byte-identical. `get_filename` is used nowhere outside `libpsio`, so nothing else sees the signature change.

## Rebased and reduced

This PR previously carried the whole of #3514 item 2 and was stacked on #3452 → #3515 → #3516. #3452 and #3515 have merged, and the other two thirds are **superseded by #3540**, which reimplements libpsio's I/O core on [libspill](https://github.com/ReLibQC/libspill):

| previously in this PR | status |
| --- | --- |
| `psio_unit` / `psio_readlen` / `psio_writlen` → `std::vector` | dropped — #3540 deletes the array outright rather than converting it |
| `psio_ud::path` → `std::string` | dropped — `psio_vol` and `psio_ud` go with it |
| **`get_filename()` → `std::string`** | **kept — this PR** |

Once libspill owns the file handle and the on-disk layout, the per-unit descriptor has nothing left to describe, so modernizing it is effort spent on something about to be removed. `get_filename.cc` is a file that port *keeps*, so this half stands on its own and no longer depends on anything unmerged.

One difference from the earlier revision worth flagging: it had dropped `const` from `get_filename()` and `get_unit_filename()`, both of which are `const` on master. That is preserved here.

## Checklist
- [x] Tests added for any new features — n/a, no behaviour change; covered by the existing `psio_roundtrip` test
- [ ] Docstring or narrative docs updated if needed — n/a

## Status
- [x] Ready for review

### DIFF
--- a/psi4/src/psi4/libpsio/change_namespace.cc
+++ b/psi4/src/psi4/libpsio/change_namespace.cc
@@ -32,8 +32,7 @@
  */
 
 #include <cstdio>
-#include <cstdlib>
-#include <cstring>
+#include <string>
 #include "psi4/pragma.h"
 #include <memory>
 #include "psi4/libpsio/psio.h"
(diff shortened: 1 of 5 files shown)


susilehtola:
Two thirds of this is superseded by #3540; the last third is not, and is the part worth keeping.

#3540 reimplements libpsio's I/O core as a shim over [libspill](https://github.com/ReLibQC/libspill). Mapping this PR's three bullets onto that:

| this PR | after #3540 |
| --- | --- |
| `psio_unit` / `psio_readlen` / `psio_writlen` → `std::vector` | **gone** — the array is unreachable once libspill owns the handle, and psi4/psi4@f0486cad2 deletes it rather than converting it |
| `psio_ud::path` → `std::string` | **gone** — `psio_vol` and `psio_ud` go with it |
| **`get_filename()` returns `std::string`** instead of `strdup`ing into a `char **` out-param | **still wanted** — `get_filename.cc`, `rename_file.cc` and `change_namespace.cc` are files the port keeps, and they still have the `char **` convention |

The first two are careful modernizations of a structure that has no reason to exist afterwards. Worth saying explicitly that the reasoning here is what made that obvious: the note that `psio_unit` is `malloc`'d so a `std::string` member would be UB is exactly the property that shows the struct is a C-era artifact rather than something to improve.

The third bullet is the one a reviewer on #3515 actually asked for — "returning a result by `char **`, allocated by the callee, is horrible" — and #3540 does not touch it.

Suggestion: rather than close this, reduce it to that third bullet, on top of #3540. That drops it to `get_filename.cc` + the two composition sites and makes it independent of the port rather than stacked under it. I have the same change sitting on an unopened branch (`libpsio/get-filename-string`), so whichever version is convenient — no need for both.

Leaving this open rather than closing it, since unlike #3516 there is something here that survives.

susilehtola:
Rebased onto current master and reduced to the one part #3540 does not supersede.

- #3452 and #3515 have merged, so those two commits are gone from the diff.
- The `psio_unit` → `std::vector` and `psio_ud::path` → `std::string` work is dropped: #3540 deletes `psio_vol`, `psio_ud` and `PSIO::psio_unit` outright once libspill owns the handle and the layout, so converting them is effort spent on something about to be removed.
- What remains is `get_filename()` returning `std::string` — 5 files, +26/−96 — which stands on its own, since `get_filename.cc` is a file that port keeps.

No longer stacked on anything unmerged.

One thing found while reducing it, worth recording: the earlier revision had dropped `const` from both `get_filename()` and `get_unit_filename()`. Both are `const` on master (`get_unit_filename` since #3515), so that would have been a quiet regression. It is preserved in the rebased version.

loriab (approved):
lgtm. moreover, lgt windows and v2rdm.

PR #3530, opened 2026-09-03
Add fixed-reference variant to the DirectJK incremental Fock build

susilehtola:
Adds a fixed-reference variant to the `DirectJK` incremental Fock build, exposed through a new `INCFOCK_FIXED_REFERENCE` option.

## Background

The default incremental Fock build keeps the reference density updated every iteration: it builds J/K from `ΔD = D − D_prev` and accumulates the result in place. This tightens density-based screening as the SCF converges, but small errors accumulate over many iterations, so a full Fock matrix must be rebuilt every `INCFOCK_FULL_FOCK_EVERY` iterations, and the incremental build is switched off once `INCFOCK_CONVERGENCE` is reached.

## The fixed-reference variant

With `INCFOCK_FIXED_REFERENCE true`, the density and the J/K/wK matrices from the first full build are captured and held fixed. Each subsequent iteration:

1. builds J/K/wK from the difference against the **fixed** reference, `ΔD = D − D_ref`, and
2. adds the stored reference contributions back: `J = J[D − D_ref] + J[D_ref]`.

By linearity of the Fock build this is exact regardless of iteration count (up to screening), so it runs on **every** iteration and needs no periodic full rebuild — `INCFOCK_FULL_FOCK_EVERY` does not apply. The trade-off is that roughly the same number of integrals are computed at each incremental step, with no danger of accumulated numerical creep. See [Parrish:2016:131101] for inspiration.

## Implementation notes

- The default (updated-reference) path is **behaviorally unchanged** — the fixed-reference logic is a separate branch in `incfock_setup`, `incfock_postiter`, the per-iteration decision, and the `build_JK_matrices` zeroing (fixed reference always zeros, since it recomputes a fresh delta and adds the reference back).
- New members on `DirectJK`: `fixed_reference_` and `D_reference_ / J_reference_ / K_reference_ / wK_reference_`.
- `print_header` reports the fixed-reference setting when incfock is on.

## Testing

Adds a `density-screen-3` regression test that runs wB97X/cc-pVTZ three ways — no incfock, incremental with a fixed reference, and incremental with an updating reference — and checks all three energies agree to 1e-9. wB97X exercises J, K, and wK simultaneously, so all three reference matrices are covered.

## Notes for reviewers

- This is a clean reimplementation on current `master` of a fixed-reference variant I first prototyped ~2 years ago (branch `directjk_reference`); that branch had since been overtaken by the IncFock standardization (#2808), so it is reimplemented from scratch on top of the current incfock structure rather than rebased.
- I was not able to build/run locally in this environment, so **CI is the compile-and-test gate**. In particular the `density-screen-3` reference energy and the exact-agreement tolerance should be confirmed by CI; `output.ref` is intentionally not committed (the runner drives `input.dat`'s `compare_values` calls, not a byte comparison) and can be regenerated from a passing run.

### DIFF
--- a/psi4/src/psi4/libfock/jk.h
+++ b/psi4/src/psi4/libfock/jk.h
@@ -750,10 +750,10 @@ class PSI_API DirectJK : public JK {
     std::vector<SharedMatrix> D_prev_;
 
     /// Fixed reference density and the J, K, and wK matrices built from it (used by the fixed-reference incremental build)
-    std::vector<SharedMatrix> D_reference_;
-    std::vector<SharedMatrix> J_reference_;
-    std::vector<SharedMatrix> K_reference_;
-    std::vector<SharedMatrix> wK_reference_;
+    std::vector<SharedMatrix> D_fixed_ref_;
+    std::vector<SharedMatrix> J_fixed_ref_;
+    std::vector<SharedMatrix> K_fixed_ref_;
+    std::vector<SharedMatrix> wK_fixed_ref_;
 
     /// Pseudo-density matrix to be used this iteration
     std::vector<SharedMatrix> D_ref_;
--- a/psi4/src/psi4/libfock/jk.h
+++ b/psi4/src/psi4/libfock/jk.h
@@ -740,12 +740,20 @@ class PSI_API DirectJK : public JK {
     
     /// Perform Incremental Fock Build for J and K Matrices? (default false)
     bool incfock_;
+    /// Use a fixed reference for the incremental Fock build instead of updating it every iteration? (default false)
+    bool fixed_reference_;
     /// The number of times INCFOCK has been performed (includes resets)
     int incfock_count_;
     bool do_incfock_iter_;
 
-    /// Previous iteration pseudo-density matrix
+    /// Previous iteration pseudo-density matrix (used by the default, updated-reference incremental build)
     std::vector<SharedMatrix> D_prev_;
+
+    /// Fixed reference density and the J, K, and wK matrices built from it (used by the fixed-reference incremental build)
+    std::vector<SharedMatrix> D_reference_;
+    std::vector<SharedMatrix> J_reference_;
+    std::vector<SharedMatrix> K_reference_;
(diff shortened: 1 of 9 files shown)


psi4/src/psi4/libfock/jk.h
```diff
@@ -740,13 +740,21 @@ class PSI_API DirectJK : public JK {
...
+    bool fixed_reference_;
     /// The number of times INCFOCK has been performed (includes resets)
     int incfock_count_;
     bool do_incfock_iter_;
 
-    /// Previous iteration pseudo-density matrix
+    /// Previous iteration pseudo-density matrix (used by the default, updated-reference incremental build)
     std::vector<SharedMatrix> D_prev_;
 
+    /// Fixed reference density and the J, K, and wK matrices built from it (used by the fixed-reference incremental build)
```
andyj10224, line 752: Having `D_reference_` and `D_ref_` simultaneously as instance variables seem confusing. I think it would be better to name it something like `D_fixed_` or even `D_fixed_ref_`.

andyj10224 (changes requested):
Good work @susilehtola! This would be great for addressing some of the numerical instabilities associated with incremental Fock, and it would also be worth getting it working for `CompositeJK` in the future as well (of course after the CFMM PRs are merged :)) My only comment is on some of the naming conventions, but that should probably take less than 5 minutes to fix on Claude 😂

psi4/src/psi4/libfock/jk.h
susilehtola: Good call — renamed `D/J/K/wK_reference_` → `*_fixed_ref_` in 64c94ace6 so they no longer collide with the difference-density `D_ref_`. No functional change.

andyj10224 (approved):
LGTM! Thank you for this new feature!

susilehtola:
> LGTM! Thank you for this new feature!

Given that I did this originally in December 2023 (https://github.com/psi4/psi4/pull/3111), not sure if it counts as new ;)

